PR #542, opened 2024-08-16
[menu] Do not select an item when space is pressed during typeahead

michaldudak:
Changed the <kbd>Space</kbd> behavior to not trigger the menu item when pressed during text navigation.

### DIFF
--- a/packages/mui-base/src/Menu/Item/useMenuItem.ts
+++ b/packages/mui-base/src/Menu/Item/useMenuItem.ts
@@ -4,6 +4,7 @@ import { FloatingEvents } from '@floating-ui/react';
 import { useButton } from '../../useButton';
 import { mergeReactProps } from '../../utils/mergeReactProps';
 import { GenericHTMLProps } from '../../utils/types';
+import { MuiCancellableEvent } from '../../utils/MuiCancellableEvent';
 
 /**
  *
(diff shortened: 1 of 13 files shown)


packages/mui-base/src/Menu/Item/useMenuItem.ts
```diff
@@ -86,6 +93,10 @@ export namespace useMenuItem {
      * If `true`, the menu item will listen for mouseup events and treat them as clicks.
      */
     treatMouseupAsClick: boolean;
+    /**
+     * A ref that is set to `true` when the user is using the typeahead feature.
+     */
+    userIsTyping: React.RefObject<boolean>;
```
atomiks, line 99: Should add a `Ref` suffix, I named it `typingRef` instead

packages/mui-base/src/Menu/Item/useMenuItem.ts
```diff
@@ -36,6 +38,11 @@ export function useMenuItem(params: useMenuItem.Parameters): useMenuItem.ReturnV
           id,
           role: 'menuitem',
           tabIndex: highlighted ? 0 : -1,
+          onKeyUp: (event: React.KeyboardEvent & MuiCancellableEvent) => {
```
michaldudak, line 41: I'm going to refactor `useButton` to use the new method of preventing our logic in a separate PR

atomiks (approved):

PR #549, opened 2024-08-21
[MenuRoot] Remove wrong default value from docs

sai6855:
I was looking at source code of Menu and docs, [docs](https://deploy-preview-546--base-ui.netlify.app/base-ui/react-menu/components-api/#menu-root-props) says default value of `openOnHover` prop is `nested` but `nested` value is not defined in code and even if `openOnHover` prop is defined as `nested` following code (see line 27) would be illogical as `openOnHoverProp` will always be true. hence i removed default prop as `nested` from docs. 

https://github.com/mui/base-ui/blob/919d9ca513b3b2c52b175a33a858529408593a5f/packages/mui-base/src/Menu/Root/MenuRoot.tsx#L21-L30

### Before: 
image

### After
image

### Preview: 
https://deploy-preview-549--base-ui.netlify.app/base-ui/react-menu/components-api/#menu-root-props


packages/mui-base/src/Menu/Root/MenuRoot.tsx
```diff
@@ -139,8 +139,6 @@ namespace MenuRoot {
     delay?: number;
     /**
      * Whether the menu popup opens when the trigger is hovered after the provided `delay`.
-     *
-     * @default nested
```
michaldudak, line 143: I used to have such a prop in one of the previous implementation attempts and forgot to remove the `@default` annotation. We can add in plain text that by default `openOnHover` is set on nested menus.

sai6855: Updated https://deploy-preview-549--base-ui.netlify.app/base-ui/react-menu/components-api/#menu-root-props

packages/mui-base/src/Menu/Root/MenuRoot.tsx
```diff
@@ -139,6 +139,7 @@ namespace MenuRoot {
     delay?: number;
     /**
      * Whether the menu popup opens when the trigger is hovered after the provided `delay`.
+     * By default, openOnHover is set to `true` for nested menus.
```
michaldudak, line 142:
```suggestion
     * By default, `openOnHover` is set to `true` for nested menus.
```

michaldudak (approved):

michaldudak:
Thanks!

PR #583, opened 2024-09-05
[Slider] Modernize implementation

mj12albert:
### Summary

1. Replaced `useCompound` with `Composite`

Previously the `id` attribute of all the `input` elements were passed to `useCompound` as metadata to generate the `for` attribute of the `output` element.

However the refs provided by Composite can't be used during the render of `Output` for this, so the `id`s have to be held in a new state

2. Converted types to namespaces
3. Fix https://github.com/mui/base-ui/issues/635
4. Hooks/context are no longer exported for now - it's no longer possible to (easily) implement a Material Design slider as it would require custom subcomponents
5. Matches latest export convention in https://github.com/mui/base-ui/pull/700


docs/app/experiments/slider-marks.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 import * as React from 'react';
 import * as Slider from '@base_ui/react/Slider';
-import { useSliderContext } from '@base_ui/react/Slider';
+import { useSliderContext } from '../../../packages/mui-base/src/Slider/Root/SliderContext';
```
michaldudak, line 4: If there is a legitimate use case for the context, we should export it from the package.

docs/data/components/slider/UnstyledSliderIntroduction/css-modules/index.tsx
```diff
@@ -12,159 +13,36 @@ export default function UnstyledSliderIntroduction() {
...
-        <Slider.Output className="Slider-output" />
-        <Slider.Control className="Slider-control">
-          <Slider.Track className="Slider-track">
-            <Slider.Indicator className="Slider-indicator" />
-            <Slider.Thumb className="Slider-thumb" />
+        <Slider.Output className={classes.output} />
+        <Slider.Control className={classes.control}>
+          <Slider.Track className={classes.track}>
+            <Slider.Indicator className={classes.indicator} />
+            <Slider.Thumb className={classes.thumb} inputId=":slider-thumb-input:" />
```
michaldudak, line 31: The `inputId` prop is not documented

packages/mui-base/src/Slider/Control/SliderControl.tsx
```diff
@@ -71,6 +71,12 @@ const SliderControl = React.forwardRef(function SliderControl(
   return renderElement();
 });
 
+export namespace SliderControl {
+  export interface Props extends BaseUIComponentProps<'span', SliderRoot.OwnerState> {}
```
michaldudak, line 75:
```suggestion
  export interface OwnerState extends SliderRoot.OwnerState {}
  
  export interface Props extends BaseUIComponentProps<'span', OwnerState> {}
```

+ other cases where the OwnerState isn't defined explicitly

packages/mui-base/src/Slider/Indicator/useSliderIndicator.ts
```diff
@@ -21,9 +19,19 @@ const axisProps = {
   },
 };
 
-function useSliderIndicator(
-  parameters: UseSliderIndicatorParameters,
-): UseSliderIndicatorReturnValue {
+/**
```
michaldudak, line 22: Hooks are not documented, so this header is invalid. The current docs:api script doesn't create it anymore.

packages/mui-base/src/Slider/Root/SliderContext.tsx
```diff
@@ -0,0 +1,20 @@
+'use client';
+import * as React from 'react';
+import type { SliderRoot } from './SliderRoot';
+
+/**
+ * @ignore - internal component.
+ */
+export const SliderContext = React.createContext<SliderRoot.Context | undefined>(undefined);
```
michaldudak, line 8: Context type should be defined in this file (and be called `SliderContext`)

mj12albert: Updated ~ I've also renamed it to `SliderRootContext`/`useSliderRootContext` to match other components

docs/data/components/slider/UnstyledSliderIntroduction/css-modules/index.tsx
mj12albert: Updated!

docs/app/experiments/slider-marks.tsx
mj12albert: We discussed this and decided to revisit exporting Context(s) after the alpha phase

packages/mui-base/src/Slider/Root/useSliderRoot.ts
```diff
@@ -190,10 +188,31 @@ function useSliderRoot(parameters: UseSliderParameters): UseSliderReturnValue {
...
+      if (index > -1 && inputId !== undefined) {
+        setInputMap((prevMap) => new Map(prevMap).set(index, inputId));
+      }
+
+      return {
+        deregister: deregisterInputId,
+      };
+    },
+    [deregisterInputId],
+  );
```
atomiks, line 215: Is it possible to modify the `Composite` source to allow for metadata instead?

mj12albert: Should be! How about doing this when refactoring Tabs to use Composite? Tabs uses metadata more extensively, and I wanted to get this PR merged ASAP to unblock the ESM work @michaldudak is doing 🙏

atomiks: Seems ok to me as of now

packages/mui-base/src/Slider/Root/SliderContext.tsx
michaldudak: Yeah, that's actually what I meant 😅

michaldudak (approved):

PR #604, opened 2024-09-12
[useScrollLock] Avoid scrollbar layout shift issues

### LINKED ISSUE #603: [useScrollLock] Layout shift occurs with certain layouts
The `useScrollLock` hook currently has [issues](https://github.com/mui/base-ui/pull/541#issuecomment-2343556722):

- `position: fixed` elements need to have their `width` or `right` properties compensated by the removed scrollbar width, otherwise they shift over to the right once scroll locking is activated on the body. Regular content is unaffected. To complicate matters more, in some systems, the `<body>` scrollbar appears on the left of the screen; this means the `left` property, if specified, would need to be compensated under certain conditions.
- When `overflow-y: scroll` is applied to the `<body>`, the removed scrollbar `padding` compensation itself causes layout shift and needs to be ignored. There may be other layout settings where this is the case as well.

https://github.com/radix-ui/primitives/issues/1925

The `scrollbar-gutter: stable` CSS property, when applied to `<html>`:

- Is unsupported in Safari
- Fully solves this issue in Firefox
- Is bugged in Chrome, as initial shift still occurs, then works on subsequent locking until a resize occurs - making it unusable in practice. 

## Possible solutions

1. @vladmoroz has a demo as an alternative that could "just work" and require no manual intervention: https://codesandbox.io/p/sandbox/remove-scroll-poc-forked-pqfyhf
2. If the above doesn't work out - first, we need to remove the `padding` compensation under certain layouts. However, `position: fixed` will still need their layout adjusted for the removed scrollbar width. We need to add a CSS variable to the root such as `--base-ui-removed-scrollbar-width` that users can use to apply to their `position: fixed` elements: `width: calc(100% - var(--base-ui-removed-scrollbar-width, 0px))` or `right: var(--base-ui-removed-scrollbar-width, 0)`.

**Search keywords**: scroll-lock

atomiks:
Closes #603

---

- This applies the internal iOS technique to all browsers _aside_ from iOS, removing the need for scrollbar compensation.
- iOS implementation is pending

## Problem 1

This currently breaks the positioning of (anchored) floating elements. **Edit: works if using the `fixed` positioning strategy instead of `absolute` + fixed in latest Floating UI version.**

This raises some minor interoperability concerns with extensions and third party widgets that use outdated versions of Floating UI, Popper (assuming it's broken there) or custom positioning methods that don't handle this edge case (where `html` has an offset).

## Problem 2

The forced overflow of the `<html>` scrollbar appears on top of backdrops, which looks a bit unsightly if it's not a custom scrollbar

Screenshot 2024-09-12 at 4 29 06 PM

### DIFF
--- a/packages/mui-base/src/utils/useScrollLock.ts
+++ b/packages/mui-base/src/utils/useScrollLock.ts
@@ -24,8 +24,8 @@ function preventScrollStandard() {
   function lockScroll() {
     const offsetLeft = window.visualViewport?.offsetLeft || 0;
     const offsetTop = window.visualViewport?.offsetTop || 0;
-    const hasConstantOverflowY = rootStyle.overflowY === 'scroll';
-    const hasConstantOverflowX = rootStyle.overflowX === 'scroll';
+    const hasConstantOverflowY = getComputedStyle(html).overflowY === 'scroll';
+    const hasConstantOverflowX = getComputedStyle(html).overflowX === 'scroll';
 
     scrollX = rootStyle.left ? parseFloat(rootStyle.left) : window.scrollX;
     scrollY = rootStyle.top ? parseFloat(rootStyle.top) : window.scrollY;
--- a/packages/mui-base/src/utils/useScrollLock.ts
+++ b/packages/mui-base/src/utils/useScrollLock.ts
@@ -47,7 +47,7 @@ export function useScrollLock(enabled: boolean = true) {
           overflowY: '',
         });
 
-        if ('scrollTo' in window) {
+        if (window.scrollTo.toString().includes('[native code]')) {
           window.scrollTo(scrollX, scrollY);
         }
       }
--- a/packages/mui-base/src/utils/useScrollLock.ts
+++ b/packages/mui-base/src/utils/useScrollLock.ts
@@ -1,6 +1,5 @@
 import { useEnhancedEffect } from './useEnhancedEffect';
 import { useId } from './useId';
-import { isIOS } from './detectBrowser';
 
 const activeLocks = new Set<string>();
 
@@ -10,9 +9,8 @@ const activeLocks = new Set<string>();
  * @param enabled - Whether to enable the scroll lock.
  */
 export function useScrollLock(enabled: boolean = true) {
-  // Based on Floating UI's FloatingOverlay
-
   const lockId = useId();
+
   useEnhancedEffect(() => {
     if (!enabled) {
       return undefined;
@@ -20,55 +18,35 @@ export function useScrollLock(enabled: boolean = true) {
 
     activeLocks.add(lockId!);
 
-    const rootStyle = document.documentElement.style;
-    // RTL <body> scrollbar
-    const scrollbarX =
-      Math.round(document.documentElement.getBoundingClientRect().left) +
-      document.documentElement.scrollLeft;
-    const paddingProp = scrollbarX ? 'paddingLeft' : 'paddingRight';
-    const scrollbarWidth = window.innerWidth - document.documentElement.clientWidth;
+    const html = document.documentElement;
+    const rootStyle = html.style;
     const scrollX = rootStyle.left ? parseFloat(rootStyle.left) : window.scrollX;
     const scrollY = rootStyle.top ? parseFloat(rootStyle.top) : window.scrollY;
+    const offsetLeft = window.visualViewport?.offsetLeft || 0;
+    const offsetTop = window.visualViewport?.offsetTop || 0;
 
-    rootStyle.overflow = 'hidden';
-
-    if (scrollbarWidth) {
-      rootStyle[paddingProp] = `${scrollbarWidth}px`;
-    }
-
-    // Only iOS doesn't respect `overflow: hidden` on document.body, and this
-    // technique has fewer side effects.
-    if (isIOS()) {
-      // iOS 12 does not support `visualViewport`.
-      const offsetLeft = window.visualViewport?.offsetLeft || 0;
-      const offsetTop = window.visualViewport?.offsetTop || 0;
-
-      Object.assign(rootStyle, {
-        position: 'fixed',
-        top: `${-(scrollY - Math.floor(offsetTop))}px`,
-        left: `${-(scrollX - Math.floor(offsetLeft))}px`,
-        right: '0',
-      });
-    }
+    Object.assign(rootStyle, {
+      position: 'fixed',
+      top: `${-(scrollY - Math.floor(offsetTop))}px`,
+      left: `${-(scrollX - Math.floor(offsetLeft))}px`,
+      right: '0',
+      overflowY: html.scrollHeight > html.clientHeight ? 'scroll' : 'hidden',
+      overflowX: html.scrollWidth > html.clientWidth ? 'scroll' : 'hidden',
+    });
 
     return () => {
       activeLocks.delete(lockId!);
 
       if (activeLocks.size === 0) {
         Object.assign(rootStyle, {
-          overflow: '',
-          [paddingProp]: '',
+          position: '',
+          top: '',
+          left: '',
+          right: '',
+          overflowX: '',
+          overflowY: '',
         });
-
-        if (isIOS()) {
-          Object.assign(rootStyle, {
-            position: '',
-            top: '',
-            left: '',
-            right: '',
-          });
-          window.scrollTo(scrollX, scrollY);
-        }
+        window.scrollTo(scrollX, scrollY);
       }
     };
   }, [lockId, enabled]);
--- a/packages/mui-base/src/utils/useScrollLock.ts
+++ b/packages/mui-base/src/utils/useScrollLock.ts
@@ -1,6 +1,5 @@
 import { isIOS } from './detectBrowser';
 import { useEnhancedEffect } from './useEnhancedEffect';
-import { useId } from './useId';
 
 let originalStyles = {};
 
@@ -78,10 +77,8 @@ function preventScrollStandard() {
  * @param enabled - Whether to enable the scroll lock.
  */
 export function useScrollLock(enabled: boolean = true) {
-  const id = useId();
-
   useEnhancedEffect(() => {
-    if (!enabled || !id) {
+    if (!enabled) {
       return undefined;
     }
 
@@ -96,5 +93,5 @@ export function useScrollLock(enabled: boolean = true) {
         restore();
       }
     };
-  }, [enabled, id]);
+  }, [enabled]);
 }
--- a/packages/mui-base/src/utils/useScrollLock.ts
+++ b/packages/mui-base/src/utils/useScrollLock.ts
@@ -30,6 +30,7 @@ export function useScrollLock(enabled: boolean = true) {
       const scrollbarWidth = window.innerWidth - html.clientWidth;
       const offsetLeft = window.visualViewport?.offsetLeft || 0;
       const offsetTop = window.visualViewport?.offsetTop || 0;
+
       scrollX = rootStyle.left ? parseFloat(rootStyle.left) : window.scrollX;
       scrollY = rootStyle.top ? parseFloat(rootStyle.top) : window.scrollY;
 
@@ -57,10 +58,13 @@ export function useScrollLock(enabled: boolean = true) {
       bodyStyle.overflow = 'hidden';
 
       Object.assign(rootStyle, {
+        // Handle `scrollbar-gutter` in Chrome when there is no scrollable content.
         position: html.scrollHeight > html.clientHeight ? 'fixed' : '',
         top: `${-(scrollY - Math.floor(offsetTop))}px`,
         left: `${-(scrollX - Math.floor(offsetLeft))}px`,
         right: '0',
+        // On iOS, the html can't be scrollable at it allows "pull-to-refresh" to still work. This
+        // is only necessary when scrollbars are present.
         ...(scrollbarWidth && {
           overflowY: html.scrollHeight > html.clientHeight ? 'scroll' : 'hidden',
           overflowX: html.scrollWidth > html.clientWidth ? 'scroll' : 'hidden',
@@ -87,11 +91,11 @@ export function useScrollLock(enabled: boolean = true) {
       }
     }
 
-    const handleResize = () => {
+    function handleResize() {
       cleanup();
       cancelAnimationFrame(resizeRaf);
       resizeRaf = requestAnimationFrame(lockScroll);
-    };
+    }
 
     lockScroll();
     window.addEventListener('resize', handleResize);
--- a/packages/mui-base/src/utils/useScrollLock.ts
+++ b/packages/mui-base/src/utils/useScrollLock.ts
@@ -3,6 +3,7 @@ import { useId } from './useId';
 
 const activeLocks = new Set<string>();
 let originalStyles = {};
+let originalBodyOverflow = '';
 
 /**
  * Locks the scroll of the document when enabled.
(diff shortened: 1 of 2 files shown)


atomiks:
Problem 1: https://github.com/floating-ui/floating-ui/pull/3045

vladmoroz (commented):
> However, this raises the concern of interoperability with extensions and third party widgets that use outdated versions of Floating UI, Popper (assuming it's broken there) or custom positioning methods that don't handle this edge case (where html has an offset).

I think this is fine given that Base UI will take some time to get real traction, and a newer Floating UI version will be around during this time. Given that Floating UI is a dependency, I think we should recommend to use a matching Floating UI version in the docs either way.

> The forced overflow of the <html> scrollbar appears on top of backdrops, which looks a bit unsightly if it's not a custom scrollbar

I agree this is not ideal. I'd take that over any layout shift though.

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -10,63 +9,45 @@ const activeLocks = new Set<string>();
...
       if (activeLocks.size === 0) {
         Object.assign(rootStyle, {
-          overflow: '',
-          [paddingProp]: '',
+          position: '',
+          top: '',
+          left: '',
+          right: '',
+          overflowX: '',
+          overflowY: '',
```
vladmoroz, line 47: Do we need to restore any original styles?

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -10,63 +9,45 @@ const activeLocks = new Set<string>();
...
-        top: `${-(scrollY - Math.floor(offsetTop))}px`,
-        left: `${-(scrollX - Math.floor(offsetLeft))}px`,
-        right: '0',
-      });
-    }
+    const offsetLeft = window.visualViewport?.offsetLeft || 0;
+    const offsetTop = window.visualViewport?.offsetTop || 0;
+
+    Object.assign(rootStyle, {
+      position: 'fixed',
```
vladmoroz, line 29: Having `scrollbar-gutter: stable` on `html` and no scrollable content is bugged in Chrome with this approach because of Chrome's bug:
https://issues.chromium.org/issues/40835364

This seems to work regardless of `scrollbar-gutter: stable`:

```suggestion
    position: html.scrollHeight > html.clientHeight ? 'fixed' : '',
```

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -10,63 +9,45 @@ const activeLocks = new Set<string>();
...
+      top: `${-(scrollY - Math.floor(offsetTop))}px`,
+      left: `${-(scrollX - Math.floor(offsetLeft))}px`,
+      right: '0',
+      overflowY: html.scrollHeight > html.clientHeight ? 'scroll' : 'hidden',
+      overflowX: html.scrollWidth > html.clientWidth ? 'scroll' : 'hidden',
```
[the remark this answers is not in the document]
vladmoroz: It's doable, check this out:

https://github.com/user-attachments/assets/e956ff80-f4d4-4dd5-b745-55a648369997

On each resize, we can do `cleanup()` before making the new measurements and resetting the scroll lock.

I haven't tested thoroughly if performance in complex documents would be an issue, but I imagine that if that's the case, we can schedule the style change in a RAF and always cancel any previous RAFs that didn't run.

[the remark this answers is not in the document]
vladmoroz: Sure
https://github.com/atomiks/base-ui/compare/fix/lockScroll...vladmoroz:base-ui:lock-scroll

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -1,75 +1,104 @@
...
       Object.assign(rootStyle, {
-        position: 'fixed',
+        position: html.scrollHeight > html.clientHeight ? 'fixed' : '',
         top: `${-(scrollY - Math.floor(offsetTop))}px`,
         left: `${-(scrollX - Math.floor(offsetLeft))}px`,
         right: '0',
+        ...(scrollbarWidth && {
+          overflowY: html.scrollHeight > html.clientHeight ? 'scroll' : 'hidden',
+          overflowX: html.scrollWidth > html.clientWidth ? 'scroll' : 'hidden',
+        }),
```
vladmoroz, line 67: Any reason the styles are now contingent on `scrollbarWidth`?

The `scrollbarWidth` variable doesn't seem to recompute correctly when downsizing from viewport height that fits all contents to a viewport height that has a scrollbar

https://github.com/user-attachments/assets/3e168495-af56-4d96-b2e6-b9db18d83497

(Works fine the other way around)

atomiks: On iOS, it kind of scrolls the document html in the background in a strange way. It could be limited to use a platform check instead.

vladmoroz: Sounds good, should we leave a couple comments about these weird cases we are handling for the historical context? Like this one, and the one with conditional `position: fixed` that handles the scrollbar gutter bug in Chrome

atomiks: Added some comments and moved the variable inside the function so it should stay fresh

vladmoroz (commented):
Alright, desktop looks fantastic.

Checking iOS now and I can get it to do weird jerky right after zooming in or out... Is there any hope we can nail this?

https://github.com/user-attachments/assets/c19a39a6-0e75-4317-a461-610d5d4db94d

vladmoroz:
Doing another test:
- On an iOS device, I can't seem to scroll or zoom at all
- There's a shift when resizing the viewport in Chrome with `scrollbar-gutter: stable`. This used to works fine in 51fd6823679735a6df62d45109a55e3ed38a07ce

https://github.com/user-attachments/assets/cc1090a7-9c2e-4560-9d07-b93c962d95af

atomiks:
> On an iOS device, I can't seem to scroll or zoom at all

You mean once the scroll lock is activated and using pinch-zoom? I can also reproduce this. I suppose it's a limitation of react-aria's solution then, unfortunately. If you mean scrolling or zooming while it's not activated, I can't reproduce that.

> There's a shift when resizing the viewport in Chrome with scrollbar-gutter: stable. This used to works fine in https://github.com/mui/base-ui/commit/51fd6823679735a6df62d45109a55e3ed38a07ce

Will have to check that.

vladmoroz:
> You mean once the scroll lock is activated and using pinch-zoom? I can also reproduce this. I suppose it's a limitation of react-aria's solution then, unfortunately

Wow I'm shocked that that's what they have... if you open a dialog zoomed in, your only way to get out of it is basically to close the page.

atomiks:
I'll try to complete the `touch-action` prototype I had which still allowed pinch-zooming. The logic to find the scrollable container is possible with DOM traversal on a touch event

vladmoroz:
@atomiks let's catch up on Monday and mull it over as well. I checked a few other implementations (Ariakit working the best so far) and have a few new thoughts; not doing anything on iOS doesn't feel like a bad idea actually.

For Chrome's scroll gutter issue, I think it's caused by that same bug and not our code. We probably just want to check for the scrollbar-gutter value and if it's not "auto", set overflow to hidden because the browser will maintain a stable gutter itself.

oliviertassinari (commented):
Regarding #611, it doesn't work in this PR, there is a layout shift:

https://github.com/user-attachments/assets/27db52ba-092d-42d9-81b6-931426358e85

---

I wonder about forcing to show the scrollbar for Windows users:

SCR-20240914-oxvz

It's not the best feeling when it comes to UX. I guess the idea is that it's easier for developers not to have to add `.mui-fixed` like class names?
It would feel better like this though:

SCR-20240914-oyds

So this looks like a tradeoff between the developer experience, and the user experience. We either make it possible for developers to have peak UX, at the expense of making the worst case worse, or the other way around.

atomiks:
@oliviertassinari are you on the right deploy preview? The values you've listed don't have `position: fixed` which the [deploy link](https://deploy-preview-604--base-ui.netlify.app/components/react-dialog/) has. It preserves the value and has no layout shift when `overflow-y: scroll` is on `<html>`:

Screenshot 2024-09-15 at 1 46 42 AM

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -1,75 +1,384 @@
...
+      left: rootStyle.left,
+      right: rootStyle.right,
+      ...(scrollbarWidth && {
+        overflowX: rootStyle.overflowX,
+        overflowY: rootStyle.overflowY,
+      }),
+    };
+    originalBodyOverflow = bodyStyle.overflow;
+
+    bodyStyle.overflow = 'hidden';
```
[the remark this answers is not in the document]
atomiks: This should just be removed as it's already applied to `<html>`

oliviertassinari:
@atomiks To reproduce, you need to empty the page to not have enough content to create a scrollbar:

https://github.com/user-attachments/assets/0f07629c-a4e2-420e-9568-35de1099f742

https://deploy-preview-604--base-ui.netlify.app/components/react-dialog/

---

@joshwcomeau Cool hack in https://www.joshwcomeau.com/css/has/#global-detection-6 by the way:

SCR-20240915-cfmj

but we can't use this here 😄.

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -1,75 +1,110 @@
...
+function preventScrollIOS() {
+  // To implement
+  return () => {};
+}
+
+function preventScrollStandard() {
+  const html = document.documentElement;
+  const body = document.body;
+  const rootStyle = html.style;
+  const bodyStyle = body.style;
```
vladmoroz, line 18: We need to read the computed styles, otherwise the checks below work only when html/body styles are inline

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -1,75 +1,110 @@
...
+
+  let resizeRaf: number;
+  let scrollX: number;
+  let scrollY: number;
+
+  function lockScroll() {
+    const offsetLeft = window.visualViewport?.offsetLeft || 0;
+    const offsetTop = window.visualViewport?.offsetTop || 0;
+    const hasConstantOverflowY = rootStyle.overflowY === 'scroll';
+    const hasConstantOverflowX = rootStyle.overflowX === 'scroll';
```
vladmoroz, line 28: Let's also check whether body has `overflow: scroll`, that's common

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -1,75 +1,110 @@
...
+    originalBodyStyles = {
+      overflow: bodyStyle.overflow,
+    };
+
+    Object.assign(rootStyle, {
+      // Handle `scrollbar-gutter` in Chrome when there is no scrollable content.
+      position: html.scrollHeight > html.clientHeight ? 'fixed' : '',
+      top: `${-(scrollY - Math.floor(offsetTop))}px`,
+      left: `${-(scrollX - Math.floor(offsetLeft))}px`,
+      right: '0',
```
vladmoroz, line 50: - We don't need to set `bottom`?
- Nit: if we are setting position conditionally, should we also set top/left/right conditionally? If someone sets a `position` value on their `html` element, for whatever weird reason

atomiks: - `bottom` isn't necessary
- Changed

vladmoroz (approved):
Phew, looks great. Added a slight comment on the `scrollbar-gutter` edge case, otherwise good to go.

I'm super excited about this, reliable scroll locking without layout shifts was such a pain with Radix and other libs.

***

<details>
<summary>
For posterity, here's what I was testing across browsers
</summary>

`html` with `overflow: scroll`
- [x] Scrollable page
- [x] Non-scrollable page
- [x] Reducing viewport size to make the page scrollable
- [x] Expanding viewport size to make the page not scrollable

`html` with `scrollbar-gutter: stable`
- [x] Scrollable page
- [x] Non-scrollable page
- [x] Reducing viewport size to make the page scrollable
- [x] Expanding viewport size to make the page not scrollable

body with `overflow: scroll`
- [x] Scrollable page
- [x] Non-scrollable page
- [x] Reducing viewport size to make the page scrollable
- [x] Expanding viewport size to make the page not scrollable

Default browser styles on `body` and `html`
- [x] Scrollable page
- [x] Non-scrollable page
- [x] Reducing viewport size to make the page scrollable
- [x] Expanding viewport size to make the page not scrollable

</details>

***

> not cleanest feeling UX. I guess the idea is that it's easier for developers not to have to add .mui-fixed like class names? I would feel better like this though:

@oliviertassinari check this out:

https://github.com/user-attachments/assets/c40f7c58-7459-429e-89e7-ae5560ee5292

https://github.com/user-attachments/assets/64199710-606e-4dc8-afcc-a14af5f0a783

Developers consistently forget about hard scrollbars. Even in the teams of the highest calibre, these often end up being completely unhandled.

I agree that it would be ideal if the backdrop covered the scrollbar area. I am confident though that's not what the end user would be looking at or care about, since their attention is going to be on the part of the page that presents new information. Any risk of having a layout shift is not worth it in my opinion.

With other component libraries, we've been living with the common scrollbar compensation method that requires you to use a CSS variable to set a `right` value for `position: fixed` elements, and realistically people just forget to do this all the time.

As another example, here is a UI I had built:

https://github.com/user-attachments/assets/ed6e454d-d031-4f92-b2f7-9f57e68bc941

The header doesn't move because I did handle that, but the input got messed up over time. Maybe it was me, maybe another developer forgot that; it's natural because you don't anticipate any relationship between a random element on the page and an unrelated dialog.

My guiding rule here is that great DX and great UX go hand in hand. Compromises in DX tend to lead to end user issues.

oliviertassinari:
> Developers consistently forget about hard scrollbars. Even in the teams of the highest calibre, these often end up being completely unhandled.

@vladmoroz My guess it's because everyone on those teams works on macOS, so they don't see it. They forgot that 70% of the end-users are on Windows. And even for developers, it's still a majority https://www.notion.so/mui-org/Product-Analytics-f2e2576203564e6c8dc3429cf305a36f?pvs=4#8a38433d21814a15acfb99ce2eac7d6b.

> Any risk of having a layout shift is not worth it in my opinion.

Yeah, hard dilemma. Your rationale makes sense. Not peak of UX but we avoid the worst-case scenario, and since this is what people tend to remember, it's probably for the best: https://www.notion.so/mui-org/DevEx-DX-044f844fbda74826b2dea32a6e6d8a38?pvs=4#2a825570dd744a5a90b75e014100562e (point made by https://github.com/rsms).

I guess in this case, it's critical to have `color-scheme: dark` set for the top-level scrollbar to be dark in dark mode 😄.

---

@atomiks `overflow-y: scroll` on the html works now 👌

packages/mui-base/src/utils/useScrollLock.ts
```diff
@@ -1,75 +1,127 @@
...
+let restore: () => void = () => {};
+
+function preventScrollIOS() {
+  // To implement
+  return () => {};
+}
+
+function preventScrollStandard() {
+  const html = document.documentElement;
+  const body = document.body;
```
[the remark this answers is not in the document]
atomiks: That doesn't apply here since it's locking document level of the current script. There's no element to grab an owner of.

vladmoroz (approved):
iOS looks good too

I think we got to remove the `theme_color` from `manifest.json` because it's super obnoxious when testing iOS; kind of spoils the impression when it doesn't have to.

packages/mui-base/src/utils/useScrollLock.ts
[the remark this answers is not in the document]
oliviertassinari: > There's no element to grab an owner of.

This is the point I get to, this API can't work:

https://github.com/mui/base-ui/blob/1ab27b0510e0ba19157597f17642022a6208ae09/packages/mui-base/src/Dialog/Popup/useDialogPopup.tsx#L49

we should pass it context on which document it needs to scroll lock. Issue created #629.

PR #605, opened 2024-09-12
[core] React 19 compatibility

michaldudak:
Work in progress to ensure Base UI is compatible with React 19.

After this is approved, I'll revert all the changes to package.json files, and, if the checks are green, merge it into master.


packages/mui-base/src/utils/reactVersion.ts
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+
+let version = 17;
```
atomiks, line 3: React has a `version` export that's supported across all the majors

michaldudak: TIL, thanks!

packages/mui-base/src/utils/useComponentRenderer.ts
```diff
@@ -79,7 +79,7 @@ export function useComponentRenderer<
 
   let resolvedRenderProp:
     | ComponentRenderFn<React.HTMLAttributes<any>, OwnerState>
-    | React.ReactElement;
+    | React.ReactElement<any>;
```
oliviertassinari, line 82: ? https://github.com/mui/material-ui/pull/43402

```suggestion
    | React.ReactElement<unknown>;
```

packages/mui-base/src/utils/useRenderPropForkRef.ts
```diff
@@ -8,7 +10,13 @@ import { useForkRef } from './useForkRef';
...
   ...refs: Array<React.Ref<any>>
-) {
-  const childRef = typeof render !== 'function' ? render.ref : null;
+): React.RefCallback<any> | null {
+  let childRef;
+  if (isReactVersionAtLeast(19)) {
+    childRef = typeof render !== 'function' ? render.props.ref : null;
+  } else {
+    childRef = typeof render !== 'function' ? render.ref : null;
+  }
```
oliviertassinari, line 19: Better fork after the `typeof`. But also not sure this is wise: https://github.com/mui/material-ui/pull/42818#discussion_r1664248303. The explore we are exploring in https://github.com/mui/material-ui/pull/43022. To decide then use the same solution through the codebase.

packages/mui-base/src/utils/useAnchorPositioning.ts
```diff
@@ -275,3 +275,19 @@ export function useAnchorPositioning(
     ],
   );
 }
+
+function isRef(param: unknown): param is React.RefObject<any> {
+  return {}.hasOwnProperty.call(param, 'current');
```
oliviertassinari, line 280: Why not? It takes less space:
```suggestion
  return param.hasOwnProperty('current');
```

oliviertassinari (commented):
Connecting the dots with the other projects: Aaron went through a lot of those changes https://github.com/mui/material-ui/pull/42824. I think it would be great to have his review.

packages/mui-base/src/utils/useAnchorPositioning.ts
michaldudak: It's not a part of this PR anymore

packages/mui-base/src/utils/useRenderPropForkRef.ts
michaldudak: > Better fork after the typeof

You mean `const childRef = typeof render !== 'function' ? isReactVersionAtLeast(19) ? render.props.ref : render.ref : null;`? This makes it a nested ternary which I believe we disallow in our ESLint rules.

> I'm sure this is best: https://github.com/mui/material-ui/pull/42818#discussion_r1664248303.

How do you measure "best"? The version check is faster than property check (https://jsbench.me/mrm16egz7n/1) and the function call explains pretty well that we're branching because something changed in React version.

docs/src/components/GoogleAnalytics.tsx
```diff
@@ -19,15 +20,17 @@ export function GoogleAnalytics() {
...
-      <BaseGoogleAnalytics
-        productId="base-ui"
-        productCategoryId="core"
-        currentRoute={currentRoute}
-        codeLanguage={demoVariantSelectorContext.selectedLanguage}
-        codeStylingVariant={demoVariantSelectorContext.selectedVariant}
-        packageManager={packageManagerSnippetContext.packageManager}
-        userLanguage="en"
-      />
+      <NoSsr>
```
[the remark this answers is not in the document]
michaldudak: No, likely not. I was trying out different things (it's still a draft PR).

docs/src/design-system/ToggleButtonGroup.tsx
```diff
@@ -47,4 +47,4 @@ export const ToggleButtonGroup = React.forwardRef(function ToggleButtonGroup<
   );
 }) as <Option extends { value: string; label: string }>(
   props: ToggleButtonGroupProps<Option> & { ref?: React.Ref<HTMLDivElement> },
-) => JSX.Element;
+) => React.JSX.Element;
```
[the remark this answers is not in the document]
michaldudak: I intend to have everything (except for package.json changes) merged to master.

packages/mui-base/src/utils/useRenderPropForkRef.ts
oliviertassinari: I think that we should reverse  the condition, from

```js
  let childRef;
  if (isReactVersionAtLeast(19)) {
    childRef = typeof render !== 'function' ? render.props.ref : null;
  } else {
    childRef = typeof render !== 'function' ? render.ref : null;
  }
```

to 

```js
  let childRef;

  if (typeof render !== 'function') {
    childRef = isReactVersionAtLeast(19) ? render.props.ref : render.ref;
  } else {
    childRef = null;
  }
```

michaldudak: > I was proposing to go...

All right, makes sense :+1:

> The only ask is that we are moving to a place where Base UI is exporting one helper and Material UI, etc. using it. I don't think that this should be duplicated.

I see your point. I don't feel comfortable sharing Base UI utils just yet, though, even internally. When we're closer to the release and breaking changes are less likely, we can review the whole codebase and move utilities where they ultimately belong.

oliviertassinari: > I don't feel comfortable sharing Base UI utils just yet, though, even internally

We could duplicate the logic until this happen. The outcome I'm trying to go after is: learn once fix everywhere.

packages/mui-base/src/utils/useComponentRenderer.ts
michaldudak: We don't really care about the shape of the props here, especially that it's an internal hook and its type shouldn't be publicly visible. To constrain it a bit more, I changed it to `Record<string, unknown>`

docs/data/components/tooltip/UnstyledTooltipFollowCursor.js
```diff
@@ -7,7 +7,7 @@ import { styled } from '@mui/system';
 export default function UnstyledTooltipFollowCursor() {
   return (
     <div style={{ display: 'flex', gap: 12 }}>
-      <Tooltip.Root followCursorAxis="both">
+      <Tooltip.Root trackCursorAxis="both">
```
michaldudak, line 10: It's not clear to me why it started to fail now, but I guess it's a good thing it fails.

packages/mui-base/src/Field/Validity/FieldValidity.test.tsx
```diff
@@ -23,7 +23,7 @@ describe('<Field.Validity />', () => {
     fireEvent.change(input, { target: { value: 'test' } });
     fireEvent.blur(input);
 
-    const [data] = handleValidity.args[8];
+    const [data] = handleValidity.lastCall.args;
```
michaldudak, line 26: @atomiks, could you please verify that with these changes the intent of these checks is the same?

atomiks: Yes, that's fine

atomiks (approved):

michaldudak:
@aarongarciah, I'd appreciate it if you could take a look at it as well.

docs/app/experiments/dialog.tsx
```diff
@@ -179,6 +179,7 @@ function ReactSpringTransition(props: { open: boolean; children?: React.ReactEle
   }, [api, open, mounted, setMounted]);
 
   return mounted ? (
+    /* @ts-ignore */
```
aarongarciah, line 182: We could use `@ts-expect-error` and add a description on why is needed.

aarongarciah (commented):
Nice! I couldn't find any potential improvement related React 19 support.

docs/app/experiments/dialog.tsx
michaldudak: I added an explanation, but I can't use @ts-expect-error as there is no error in React 18 and this code must work in both versions.

PR #609, opened 2024-09-13
[menu] Fix custom anchor positioning

michaldudak:
Using the `anchor` prop works only in development mode, when React runs the effect twice. This is because we read the anchor ref in useLayoutEffect, when it hasn't been set yet. Additionally, passing the Element directly didn't work as well.

This PR adds an extra effect after the first render, that checks if the ref value changed after useLayoutEffect was called. If so, it calls `setPositionReference` again with the ref value.

### DIFF
--- a/packages/mui-base/src/utils/useAnchorPositioning.ts
+++ b/packages/mui-base/src/utils/useAnchorPositioning.ts
@@ -221,10 +221,13 @@ export function useAnchorPositioning(
   React.useEffect(() => {
     // Refs from parent components are set after useLayoutEffect runs and are available in useEffect.
     // Therefore, if the anchor is a ref, we need to update the position reference in useEffect.
-    const resolvedAnchor = typeof anchor === 'function' ? anchor() : anchor;
-    if (isRef(resolvedAnchor) && resolvedAnchor.current !== registeredPositionReference.current) {
-      refs.setPositionReference(resolvedAnchor.current);
-      registeredPositionReference.current = resolvedAnchor.current;
+    if (typeof anchor === 'function') {
+      return;
+    }
+
+    if (isRef(anchor) && anchor.current !== registeredPositionReference.current) {
+      refs.setPositionReference(anchor.current);
+      registeredPositionReference.current = anchor.current;
     }
   }, [refs, anchor, rerender]);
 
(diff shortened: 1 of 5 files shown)


packages/mui-base/src/utils/useAnchorPositioning.ts
```diff
@@ -205,20 +204,29 @@ export function useAnchorPositioning(
...
-    if (resolvedAnchor && !isElement(resolvedAnchor)) {
-      refs.setPositionReference(isRef(resolvedAnchor) ? resolvedAnchor.current : resolvedAnchor);
+  }, [refs, anchor]);
+
+  React.useEffect(() => {
+    // Refs from parent components are set after useLayoutEffect runs and are available in useEffect.
+    // Therefore, if the anchor is a ref, we need to update the position reference in useEffect.
+    const resolvedAnchor = typeof anchor === 'function' ? anchor() : anchor;
+    if (isRef(resolvedAnchor) && resolvedAnchor.current !== registeredPositionReference.current) {
+      refs.setPositionReference(resolvedAnchor.current);
```
atomiks, line 226: Is `flushSync` necessary now, if there will be a flicker?

michaldudak: I haven't noticed it flickering. With the menu initially open, even with 20x CPU slowdown, the popup appears only near the anchor.

packages/mui-base/src/utils/useAnchorPositioning.ts
```diff
@@ -205,20 +204,32 @@ export function useAnchorPositioning(
     nodeId,
   });
 
-  // We can assume that element anchors are stable across renders, and thus can be reactive.
-  const reactiveAnchorDep = anchor == null || isElement(anchor);
-  const anchorRef = useLatestRef(anchor);
+  const rerender = useForcedRerendering();
+
+  const registeredPositionReference = React.useRef<Element | VirtualElement | null>(null);
```
atomiks, line 209:
```suggestion
  const registeredPositionReferenceRef = React.useRef<Element | VirtualElement | null>(null);
```

packages/mui-base/src/utils/useAnchorPositioning.ts
```diff
@@ -205,20 +204,32 @@ export function useAnchorPositioning(
     nodeId,
   });
 
-  // We can assume that element anchors are stable across renders, and thus can be reactive.
-  const reactiveAnchorDep = anchor == null || isElement(anchor);
-  const anchorRef = useLatestRef(anchor);
+  const rerender = useForcedRerendering();
+
```
atomiks, line 208: This doesn't seem to be actually used? It's just in the dep array

michaldudak: Right, it's a leftover from a previous attempt.

atomiks (approved):

colmtuite (approved):

PR #663, opened 2024-09-30
[POC] render="asChild"

michaldudak:


packages/mui-base/src/utils/useComponentRenderer.ts
```diff
@@ -63,6 +65,8 @@ export function useComponentRenderer<
     customStyleHookMapping,
   } = settings;
 
+  const children = extraProps?.children;
```
michaldudak, line 68: Temporary. The proper solution will require an explicit `children` parameter (preferably required TS-wise when `render === 'asChild'`).

PR #700, opened 2024-10-02
[core] Improve DX of importing Base UI components

### LINKED ISSUE #650: Clean up exports
The DX of importing Base UI components should be cleaned up to avoid unnecessary noise:
- Each component should be exported once, and not its parts
- Types should be namespaces under component name
- Hooks should not be exported

For example, typing `Dialog` in VSC should _not_ suggest anything but the `Dialog` component.

Expected import styles:
```jsx
import { Dialog } from '@base_ui/react';
import * as Dialog from '@base_ui/react/Dialog';
```

Shouldn't be a thing:
```tsx
import { DialogRoot } from '@base_ui/react';
import type { DialogRootProps } from '@base_ui/react/Dialog/...'
```

michaldudak:
Changed how components are exported. Previously we had two index files: `index.ts` containing unprefixed component names (such as Root, Input, Indicator) and `index.barrel.ts` with full names (SwitchRoot, CheckboxIndicator). The former was supposed to be used as `import * as Foo from '@base_ui/react/Foo`, while the latter `import { FooRoot } from '@base_ui/react'`.

Now we still have two index files per component, but only index.ts is meant to be imported. `index.parts.ts` aggregates and exports all parts of a component, while `index.ts` reexports them and names the exported namespace.

This makes all component parts accessible only through the component namespace:
```ts
import { Checkbox } from '@base_ui/react/Checkbox'; // or '@base_ui/react'
const CheckboxRoot = Checkbox.Root;
```

Closes #650


atomiks:
This still works with tree-shaking if you test a minimal Vite repo and build for production?

michaldudak:
Yes, the resulting bundle does not contain the code of unused parts.

atomiks (approved):

PR #701, opened 2024-10-03
[core] Restore development playground

michaldudak:
This restores the development playground we used to have with the pages router. Unfortunately, the app router requires a bit more setup, but the experience should now be the same—every component with a default export placed in the /docs/app/playground directory will be exposed under the /playground/[filename] path.

Git ignores the contents of the directory.

Also fixed the Google Analytics script that was failing on the not-found page.


mj12albert (approved):
👍 just tried this and it works ~

Side note: maybe we could make GA fail silently if it doesn't load 😬
Screenshot 2024-10-14 at 3 14 21 PM

PR #722, opened 2024-10-08
[code-infra] Update renovate config

michaldudak:
Updated Renovate config:
- included (previously omitted by mistake) devDependencies of public packages
- grouped ESLint plugins together and enabled automerge to reduce the number of PRs
- enabled automerge for `@types/*` packages
- bundled MUI packages with Playwright, as they usually have to be updated together
- removed JSS group as it's unused in this repo

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -32,7 +32,7 @@
     },
     {
       "groupName": "eslint",
-      "matchPackagePatterns": ["eslint", "eslint-*"],
+      "matchPackagePatterns": ["eslint", "eslint-*", "@typescript-eslint/*"],
       "automerge": true
     },
     {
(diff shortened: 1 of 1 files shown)


LukasTy (approved):
LGTM. 👍 

> bundled MUI packages with Playwright, as they usually have to be updated together
Do they?
We don't have them bundled on `mui-x` and all is good. 👌 
In our case, we have a unique test relying on host machine print window working that keeps on flakily failing or behaving different after certain updates. 🙈

michaldudak:
Do you use @mui/internal-test-utils? Whenever it gets updated with its own Playwright dependency, we have a mismatch.

LukasTy:
> Do you use @mui/internal-test-utils? Whenever it gets updated with its own Playwright dependency, we have a mismatch.

Yes, we do.
We end up with two playwright installations, but in our case—that's basically a necessity. 🙈

PR #734, opened 2024-10-13
[Checkbox, Menu, Switch] Audit focus states on disabled components

onehanddev:
Changes - 
- Removed focusableWhenDisabled from Menu trigger
- The disabled prop has been passed down to the component renderer, allowing the Checkbox and Switch components to appropriately reflect their disabled state in the DOM elements.

[] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

### DIFF
--- a/packages/mui-base/src/Checkbox/Root/CheckboxRoot.tsx
+++ b/packages/mui-base/src/Checkbox/Root/CheckboxRoot.tsx
@@ -98,7 +98,6 @@ const CheckboxRoot = React.forwardRef(function CheckboxRoot(
     extraProps: {
       ...otherProps,
       ...otherGroupProps,
-      disabled
     },
   });
 
--- a/packages/mui-base/src/Checkbox/Root/CheckboxRoot.tsx
+++ b/packages/mui-base/src/Checkbox/Root/CheckboxRoot.tsx
@@ -98,7 +98,7 @@ const CheckboxRoot = React.forwardRef(function CheckboxRoot(
     extraProps: {
       ...otherProps,
       ...otherGroupProps,
-      disabled
+      disabled,
     },
   });
 
--- a/packages/mui-base/src/Checkbox/Root/CheckboxRoot.tsx
+++ b/packages/mui-base/src/Checkbox/Root/CheckboxRoot.tsx
@@ -97,8 +97,7 @@ const CheckboxRoot = React.forwardRef(function CheckboxRoot(
     customStyleHookMapping,
     extraProps: {
       ...otherProps,
-      ...otherGroupProps,
-      disabled,
+      ...otherGroupProps
     },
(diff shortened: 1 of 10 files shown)


packages/mui-base/src/Checkbox/Root/CheckboxRoot.tsx
```diff
@@ -98,6 +98,7 @@ const CheckboxRoot = React.forwardRef(function CheckboxRoot(
     extraProps: {
       ...otherProps,
       ...otherGroupProps,
+      disabled,
```
michaldudak, line 101: Instead of passing `disabled` here, it should be set by `useCheckboxRoot`'s `getButtonProps`. You can remove `aria-disabled` and have just `disabled` prop returned from there.
Same with the Switch.

michaldudak (commented):
Thanks for working on this! With a few tweaks this will be good to merge.

packages/mui-base/src/Checkbox/Root/CheckboxRoot.tsx
onehanddev: Thank you for pointing that out! 🙏 I’ve updated the code.
Please let me know if there are any further changes needed. 😊

onehanddev: It appears that removing aria-disabled is causing CircleCI’s browser tests to fail. I suspect this might be an accessibility (a11y) issue. Do you think we should reintroduce aria-disabled alongside the disabled prop?

onehanddev: Based on the various a11y resources, we are not required to use aria-disabled if we already are using native form elements like button with appropriate roles like switch or checkbox. what do you think @michaldudak ?

michaldudak: No, having just the `disabled` attribute is enough. Using both `disabled` and `aria-disabled` is redundant and may trigger warnings in some a11y validators.
IMO the way to go would be updating the tests to expect the `disabled` attribute.

onehanddev: Hi @michaldudak, I’m noticing a different assertion-related failure in CircleCI at the moment. Could you please check it and re-run the test for me, as I don’t have the access to do it myself?

onehanddev: @michaldudak I have fixed the tests, Please review it again. Thanks !

michaldudak (approved):
Looks good! Thanks for your contribution!

PR #742, opened 2024-10-17
Rename "docs-base" path alias to "docs"

vladmoroz:


michaldudak (dismissed):

docs/scripts/formattedTSDemos.js
```diff
@@ -88,7 +88,9 @@ async function transpileFile(tsxPath, project) {
     if (enableJSXPreview) {
       transformOptions.plugins = transformOptions.plugins.concat([
         [
-          require.resolve('docs/src/modules/utils/babel-plugin-jsx-preview'),
+          require.resolve(
+            'node_modules/@mui/monorepo/docs/src/modules/utils/babel-plugin-jsx-preview',
```
michaldudak, line 92: This should work as well:

```suggestion
            '@mui/monorepo/docs/src/modules/utils/babel-plugin-jsx-preview',
```

docs/scripts/formattedTSDemos.js
```diff
@@ -88,7 +88,9 @@ async function transpileFile(tsxPath, project) {
...
+          ),
```
vladmoroz, line 93: So this was the only place where the monorepo docs package was referenced
I am not even sure we want to display the JS demos, but for now I think this is fine

docs/scripts/formattedTSDemos.js
vladmoroz: Yep that works, thanks!

Issue #662, opened 2024-09-30
[meter] Implement Meter


colmtuite:
Meter pattern: https://www.w3.org/WAI/ARIA/apg/patterns/meter/

### TBD
- The [`optimum`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter#optimum) attribute and related feature is only relevant to sighted users [^1]. Is there a use-case for making a style hook out of it? e.g. set `[data-optimum]` when the value is in the "optimum" segment (high, medium, or low, as determined by `high`/`low`/`min`/`max` props)
- 

[^1]: https://www.htmhell.dev/adventcalendar/2022/5/

oliviertassinari:
Reminds me a bit of https://github.com/mui/material-ui/issues/22472 about the Stat component and which Sebastian asked about Meter. I guess success would be if it Base UI Meter component helps build a Stat component.

colmtuite:
A generic stat like this is prob not an appropriate use of `<meter>`. Meter is typically used to indicate how much of x is left. Like a battery indicator or disk space indicator. But in any case, we'll just implement it based on ARIA and other guidelines and people can build on it however they want.

PR #743, opened 2024-10-17
[meter] Create `Meter` component

mj12albert:
Closes: https://github.com/mui/base-ui/issues/662

Docs: https://deploy-preview-743--base-ui.netlify.app/components/react-meter/

Demo: https://deploy-preview-743--base-ui.netlify.app/experiments/meter

### Summary
- Does not use the `meter` element due to cross-browser inconsistencies, attributes like `optimum` actually make the VO experience worse [^1], styling is also annoying [^2]
- Uses the same anatomy as `Progress` since they are closely related
- Handles RTL per https://github.com/mui/base-ui/issues/831. Built-in styles on the `Indicator` use `inset-inline-start` instead of `left`/`right` that automatically account f or this

- ~Implements the "high/medium/low segmenting" feature from the HTML [spec](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-meter-element.html) - the `high`/`low` props can be used to set the boundaries between the 3 segments, and sets a `data-segment` attribute indicating which segment the current value falls in. Due to inconsistent and incorrect browser implementations of this feature, we assume the prop value is inclusive [^3]~
- ~The value of the `optimum` prop, in combination with segmenting, indicates which segment of the range is "preferred". If the current value is in the preferred segment, a `[data-optimum]` attribute is set~

[^1]: https://www.htmhell.dev/adventcalendar/2022/5/
[^2]: https://scottaohara.github.io/a11y_styled_form_controls/src/meter/
[^3]: https://www.ctrl.blog/entry/html-meter-segment-boundaries.html
[^4]: https://github.com/whatwg/html/issues/3520


mj12albert:
@colmtuite I've added a [demo](https://deploy-preview-743--base-ui.netlify.app/experiments/meter) here with some controls to tweak the `value`, `min`, `max`, `low`, `high` and `optimum` props that define the range represented by the component, and pretty much demos all the features.

The combination of `min`, `max`, `low`, `high` divides the `Meter`'s range into 3 segments, `low` (red), `medium` (yellow), `high` (green):
Screenshot 2024-10-28 at 5 23 12 PM

A `[data-segment='low|medium|high']` attribute is set depending on where the `value` falls in the range.

The `optimum` prop defines whether the low, medium, or high segment of the range is "preferable", e.g. for "battery health" higher is better, but for "CPU temperature" lower is better. A `[data-optimum]` attribute is set when the value is in the "preferable" segment.

NOTE: The corresponding `optimum` HTML attribute does nothing if the "optimum value" is in the medium segment, while the implementation here does allow the "medium" segment to be "preferable"

mj12albert:
Updated RTL handling per https://github.com/mui/base-ui/issues/831

Added a demo to: https://deploy-preview-743--base-ui.netlify.app/experiments/meter

Instead of using `left`/`right` to position the `Indicator` in the built-in styles, `inset-inline-start` is used which accounts for text direction and matches our browser support

packages/react/src/meter/root/useMeterRoot.ts
```diff
@@ -0,0 +1,179 @@
...
+     */
+    max?: number;
+    /**
+     * The minimum value
+     * @default 0
+     */
+    min?: number;
+    /**
+     * Indicates the optimal point in the numeric range represented by the component.
+     * If unspecified, it will fall back to the midpoint between `min` and `max`.
```
michaldudak, line 136: This API feels a bit weird to me, but I understand it reflects the native `<meter>` element.
I think I wouldn't set the default value for `optimum` if not provided and not set the `data-optimum` attribute in this case.

packages/react/src/meter/root/useMeterRoot.ts
```diff
@@ -0,0 +1,179 @@
...
+    min,
+    value,
+    percentageValue,
+    segment,
+    isOptimal,
+  };
+}
+
+namespace useMeterRoot {
+  export type Segment = 'low' | 'medium' | 'high';
```
michaldudak, line 85: Have we considered going beyond the native meter capabilities and allow custom ranges? I can think of a gauge that has a preferable range in the middle. A bit lower and a bit higher would be less preferable (yellow) and the extremes on both sides could be red. The built-in range doesn't support such a scenario (but on the other hand is simpler to use in more common cases)

packages/react/src/meter/root/useMeterRoot.ts
mj12albert: Sounds good, it does seem a bit arbitrary anyway

packages/react/src/meter/root/useMeterRoot.ts
michaldudak: Well, the whole concept of ranges is about styling, isn't it? I might not want a gradient, but have the whole indicator change color depending on which range the value is in, for example:

```css
.indicator {
  &[data-range-invalid] {
    color: red;
  }

  &[data-range-so-so] {
    color: yellow;
  }

 &[data-range-ok] {
   color: green
 }
```

where `invalid`, `so-so`, and `ok` are names defined by the developer.

mj12albert:
https://github.com/mui/base-ui/pull/1435

PR #744, opened 2024-10-17
Update folder structure

vladmoroz:
- Remove `/examples`
  - CRA/TS example moved into `.codesandbox/template` as it's referenced in the CodeSandbox CI config
- Remove unused static files in the docs
- Move `/docs/app` to `/docs/src/app`


docs/data/getting-started/support/support.mdx
```diff
@@ -36,25 +36,6 @@ You have a few possible options to provide it:
...
-<span className="only-dark-mode">
-  <img
-    src="/static/docs-infra/forking-an-example-dark.png"
-    alt="Forking an example"
-    loading="lazy"
-    width="1628"
-    style={{ display: 'block', maxWidth: '774px' }}
-  />
-</span>
-
```
vladmoroz, line 57: This was outdated and the doc will be reworked anyway

docs/app/playground/[slug]/page.tsx
Comment (michaldudak) [inline docs/app/playground/[slug]/page.tsx:1]: We use the playground to develop and test components. This shouldn't be removed.

.codesandbox/ci.json
```diff
@@ -6,6 +6,6 @@
   "publishDirectory": {
     "@base_ui/react": "packages/mui-base/build"
   },
-  "sandboxes": ["/examples/base-ui-cra-ts"],
+  "sandboxes": ["react-vite-ts"],
```
michaldudak, line 9: This doesn't really make sense, as `react-vite-ts` doesn't depend on `@base_ui/react`. I'd ask the team if they would find the sandboxes generated per PR useful, and if yes, I'd leave just this one directory in /examples. Otherwise, we can remove the `sandboxes`.

docs/public/static/readme/browserstack-darkmode.svg
michaldudak, line 1: These files should be referenced in the README. I just realized we link to assets under mui.com, but I think we should have our own copy of these files.

.codesandbox/ci.json
vladmoroz: I think they are going to be useful, yeah. I hadn't used CodeSandbox CI before, but from reading the description it sounds like something that would be good to set up once we go a bit more public.

Thoughts on putting the template it `/.codesandbox/template/`?
No big deal but `/examples/` sets up the expectation to see something that resembles a less or more complete app built with Base UI

colmtuite (approved):

.codesandbox/ci.json
oliviertassinari: I have never used this `"sandboxes"` feature of CodeSandbox CI 

SCR-20241021-cddi

https://ci.codesandbox.io/status/mui/base-ui/pr/744/builds/552351

I suspect that we can delete this. Starting from https://deploy-preview-744--base-ui.netlify.app/components/react-field/ feels much more convenient:

SCR-20241021-cdgh

PR #745, opened 2024-10-18
[core] Add exports field and build proper ES modules

michaldudak:
Builds "true" ES modules alongside CommonJS output.

## Summary of changes

- ESM build output is placed under `/build/esm`
- CommonJS build output is placed under `/build/cjs`
- Per-component package.json files were removed
- The `exports` field was added to the package
- The `typesVersion` field was added to package.json for compatibility with older tsconfig settings
- Type definitions are built per target - the CJS one with `module: "nodenext"`, the other one with `module: "esnext"`.
- Are The Types Wrong CLI is executed on CI to check if no issues with the package are introduced.

### Type definitions

To verify if different methods of module resolution in TypeScript can resolve the exported types, the [Are The Types Wrong (ATTW)](https://github.com/arethetypeswrong/arethetypeswrong.github.io/tree/main/packages/cli) tool was used. Its reports led to building the type definitions per build output (to avoid errors like [False CJS](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md) and [False ESM](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md).
Another reported issue was the inability to resolve types under the `node10` module resolution when the `exports` field is used. While we don't support Node 10, Typescript uses this resolution strategy by default (see https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/NoResolution.md#false-positive-unsupported-file-extension).
To work around this, I introduced the `typesVersions` field in package.json, as described in https://github.com/andrewbranch/example-subpath-exports-ts-compat/tree/main/examples/node_modules/types-versions-wildcards.

## Testing

I created a repo to test the package with different bundlers and plain Node.js: https://github.com/michaldudak/base-ui-bundler-tests. There is an issue with running a Node.js app in ESM mode due to Base UI dependency on @mui/utils. The utils package does not currently conform to the ESM spec and cannot be used without a bundler this way.
As we're planning to move away from it and define our utilities locally, this won't be a problem for long.

--- 

This is built on top of https://github.com/mui/base-ui/pull/812


michaldudak:
Superseded by #821

PR #746, opened 2024-10-18
Clean up Tailwind config and docs styles

vladmoroz:
Changes:
- Create a `/new` route that includes no styles from the current setup
  - Existing CSS styles will be removed altogether once the new design can replace the existing pages. This PR just prepares a better separation between the implementation of the existing and upcoming design.
- Include Tailwind preflight everywhere
  - We need it for the demos anyway
  - Although too aggressive for my taste, this reset is useful
- Set Tailwind `content` option to match all relevant files in `docs/src`
- Add Tailwind's Prettier plugin for automatically sorting the classes 
- Remove CSS utilities folder
- Rename `size-...` classes to `s-...` to avoid conflicts with Tailwind `size` utility

There is a few minor visual regressions in the content styling, but it's all going to be reworked anyway.


colmtuite (approved):

michaldudak (approved):
Tailwind demos seem to be rendered in dark mode at all times. Make sure to address this when redesigning the site.

PR #747, opened 2024-10-18
Clean up stray docs types

vladmoroz:


docs/src/components/ApiReference.tsx
```diff
@@ -1,9 +1,24 @@
...
+  defaultValue: string | null;
+  description: string;
+}
+
+export interface ComponentAPIReference {
+  name: string;
+  description: string | undefined;
+  props: ComponentPropReference[];
+}
+
```
vladmoroz, line 21: Not a fan of placing type files separately from the code that uses them (colocation of things that change together, etc.)

michaldudak (approved):

PR #748, opened 2024-10-18
[docs] Remove trailing slash from the URLs

vladmoroz:


docs/next.config.mjs
```diff
@@ -20,6 +20,7 @@ const rootPackage = loadPackageJson();
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  trailingSlash: false,
```
vladmoroz, line 23: Not sure when I last saw a website with trailing slashes URLs or if I ever saw anyone manually type out a trailing slash in a link. Seems a bit old-school

@oliviertassinari do you remember why trailing slash is used across the board? Is there a hard reason it's still needed? I dug up [this](https://github.com/mui/material-ui/pull/16705) as the PR that introduced it originally

oliviertassinari: > do you remember why trailing slash is used across the board?

This was Next.js < 9.0.0 behavior. Those are two different URLs so to retain SEO value, we never changed the URLs.

vladmoroz: Thanks, sounds like it's not a concern for the Base UI site then

oliviertassinari: Comment updated. 

> sounds like it's not a concern for the Base UI site then

Yes and no.

- Can Base UI remove trailing slash: yes, there is no existing SEO hit possible.
- Does Base UI can not throw when using trailing slash: no, Netlify and Next.js have silent redirections. We need to break so contributors don't add regressions without realizing.
- If Base UI hosts docs-infra, can it ignore trailing slashes support: no, we will likely only migrate Material UI, MUI X with a domain name change, e.g. mui.com/material-ui to material-ui.com or material-ui.newco.com. So need dual support.

vladmoroz: All right, thanks for the context. Now that I am seeing there is a broken link detection script in place, I updated it to account for no trailing slashes in these docs, so just Base UI itself should be good to go now.

Up next we can revisit trailing slashes when Material UI and X docs are moved to a new domain and/or when we do a proper docs-infra package.

Side note: I'd expect that docs-infra will be a standalone repo + npm package. It's not an intrinsic part of the Base UI product, it's just a company-wide concern across our projects.

flaviendelangle: How would we handle Base UI X
Since the same content would be used both for Base UI X and MUI X, it will be hard to have trailing slashes on MUI X but not on Base UI X.

colmtuite: None of our products should have trailing slashes, they make us look unprofessional.

oliviertassinari: > Now that I am seeing there is a broken link detection script in place, I updated it to account for no trailing slashes in these docs, so just Base UI itself should be good to go now.

@vladmoroz this is a nice improvement 👍. 

I think that we will also need https://github.com/mui/material-ui/blob/6679d01eed664b4949d495e2b0221b99a9fde966/packages/markdown/parseMarkdown.js#L82. This should be a much better DX than the script. As far as I remember, the script is a lifeline but it's hard to debug when making a mistake.

I have run a quick scan with https://www.screamingfrog.co.uk/seo-spider/. The only link issue he can find is with https://deploy-preview-748--base-ui.netlify.app/components/react-separator

SCR-20241021-bynq

those returns 404. It looks like a regression from this PR.

> I'd expect that docs-infra will be a standalone repo + npm package. It's not an intrinsic part of the Base UI product, it's just a company-wide concern across our projects.

There is https://www.notion.so/mui-org/docs-infra-Code-location-c292d34ed5b0487a9ade846ef11d4019 on this topic specifically. As I see this, we have to pick between one of those two repositories: mui-public (to be renamed to newco-public) or base-ui. I think it's whatever helps the docs-infra team to be the most effective. Historically, it's in material-ui and not in another repository so we can very quickly propagate the changes. Without this, we have to do two extra steps:

- have a solid test case in the other repository, something that reproduces most of the edge cases
- sync at least daily changes from that other repository to the others (forcing as few breaking changes as possible)

> How would we handle Base UI X Since the same content would be used both for Base UI X and MUI X, it will be hard to have trailing slashes on MUI X but not on Base UI X.

@flaviendelangle I don't think this will be a problem. https://base-ui.com/x/react-data-grid and https://mui.com/x/react-data-grid/ are powered by the same markdown file and demos files (or almost) should do it.

> None of our products should have trailing slashes, they make us look unprofessional.

@colmtuite Yeah, I think there is an agreement that this is the right direction.

flaviendelangle: >  @flaviendelangle I don't think this will be a problem. We could have https://base-ui.com/x/react-data-grid and https://mui.com/x/react-data-grid/ powered by the same markdown file and demos files.

And when writting internal links in the md, we would always use one that would be transformed at build time if needs be?

vladmoroz: > I have run a quick scan with https://www.screamingfrog.co.uk/seo-spider/. The only link issue he can find is with https://deploy-preview-748--base-ui.netlify.app/components/react-separator
> 
> SCR-20241021-bynq
>
> those returns 404. It looks like a regression from this PR.

These never worked btw, this demo renders `<a href=".">` which I'll fix when reworking the demos

colmtuite (approved):

docs/next.config.mjs
oliviertassinari: > And when writting internal links in the md, we would always use one that would be transformed at build time if needs be?

@flaviendelangle This makes sense. I think in that world, we would write links without trailing slashes, and docs-infra would add them where needed. I believe Next.js has logic around this. This way, we default toward being close to what we aim to reach.

PR #750, opened 2024-10-21
[docs] Use the new logo

vladmoroz:
- New logo and favicons
- Tweaked homepage


docs/src/styles.css
```diff
@@ -0,0 +1,47 @@
...
+    --color-content: #000;
+    --color-background: #000;
+    --color-foreground: oklch(90% 2% 264);
+    --color-selection: oklch(50% 0.2 264 / 80%);
+    --color-gray: oklch(90% 6% 264 / 70%);
+    --color-link: oklch(69% 50% 264);
+    --color-border: oklch(90% 12% 264 / 35%);
+    --color-gridline: oklch(24% 1% 264);
+  }
+}
```
vladmoroz, line 28: The functional colours are provisional, it's likely I'll end up doing a normal Radix-style palette once there is a bit more UI to work with

.eslintrc.js
```diff
@@ -38,6 +38,9 @@ module.exports = {
       },
     ],
     '@typescript-eslint/no-redeclare': 'off',
+
+    // We LOVE non-breaking spaces here
+    'no-irregular-whitespace': 'off',
```
michaldudak, line 43: Wouldn't setting `{ skipJSXText: true, skipStrings: true }` be enough?

docs/src/app/(new)/page.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const logo = await readFile('public/static/logo.svg', 'utf-8');
+  return (
+    <div className="relative flex grow flex-col justify-between gap-10 lg:justify-center [@media(min-height:700px)]:justify-center">
+      <div className="flex px-7 pt-7 md:px-16 md:pt-14 lg:items-center lg:pb-[min(10rem,max(3.5rem,10vh))]">
+        <div className="max-w-[26rem] 2xl:max-w-[40rem]">
+          <h1 className="mb-2 text-balance text-2xl font-medium 2xl:mb-3 2xl:text-3xl">
+            {description}
+          </h1>
+          <p className="text-color-gray mb-4 text-pretty 2xl:text-lg">
+            From the creators of Radix, Floating UI, and MUI.
```
michaldudak, line 19:
```suggestion
            From the creators of Radix, Floating UI, and Material UI.
```

docs/src/app/layout.tsx
```diff
@@ -1,23 +1,26 @@
-import * as React from 'react';
 import { Metadata } from 'next';
-import { GoogleAnalytics } from 'docs/src/components/GoogleAnalytics';
+import * as React from 'react';
```
michaldudak, line 2: We have a convention to import React first.

docs/src/components/Logo.css
```diff
@@ -0,0 +1,19 @@
+@layer components {
```
michaldudak, line 1: So what's wrong with CSS modules after all? Are we going to have plain CSS, CSS modules, and Tailwind at the same time, or do you want to get rid of modules completely?

michaldudak:
Side note: could you prefix your PR titles with the main tag (`[docs]` in this case)? It's a convention we use across our repos (and it makes it easier to write the changelog).

vladmoroz:
> Side note: could you prefix your PR titles with the main tag ([docs] in this case)? It's a convention we use across our repos (and it makes it easier to write the changelog).

Sure!

docs/src/app/layout.tsx
vladmoroz: Can I add an ES lint rule to do import order so that it's fixed automatically then? I agree with the convention, I did similar for Radix Themes with a couple other ordering rules in the mix, just don't think a human should waste time doing it manually

docs/src/components/Logo.css
vladmoroz: My personal preference would be just plain CSS + Tailwind, and get rid of the current CSS Modules usage as the docs are built out.

I would have went for CSS Modules in a larger team with developers of varying skill levels, but here I don't think there's any significant problem that the tech solves for us. (I always prefix classes with component name so clashes won't happen with basic hygiene).

Do you feel OK about this or would you strongly want to use CSS Modules anyway?

docs/src/app/(new)/page.tsx
colmtuite: @vladmoroz It will be MUI next year, but yeah it should be "Material UI" for now.

vladmoroz: > @vladmoroz It will be MUI next year, but yeah it should be "Material UI" for now.

Looks like I also don't understand the difference between MUI and Material UI 🫠
How come it will change next year? Material UI library will be renamed to MUI?

docs/src/components/Logo.css
michaldudak: I would use CSS modules and not go with Tailwind at all, but it's more of a preference than an objective advantage of one tech over the other. Since you'll mostly work on the docs, it's your call.

atomiks (commented):
New homepage looks super nice 👍

docs/src/app/(new)/page.tsx
michaldudak: Current state:
MUI - the company
Material UI - the library

Who knows what the future will bring ;)

docs/src/app/layout.tsx
michaldudak: sure!

PR #751, opened 2024-10-22
[tabs] Modernize implementation

mj12albert:
### Summary
- A custom metadata object can be passed into `Composite` components and hooks
- Replace `useCompound` with `Composite`
- ~Rename `direction` prop to `dir`~ not yet, need to discuss the overall approach further https://github.com/mui/base-ui/issues/831


mj12albert:
Depends on https://github.com/mui/base-ui/pull/763 for adding RTL support to `Composite`

packages/mui-base/src/Composite/Root/useCompositeRoot.ts
```diff
@@ -170,7 +172,9 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
           }
 
           if (nextIndex !== activeIndex && !isIndexOutOfBounds(elementsRef, nextIndex)) {
-            event.stopPropagation();
```
mj12albert, line 173: @atomiks Why is `stopPropagation()` needed here, and is it ever ok to force propagate it? This breaks a [test](https://github.com/mui/base-ui/pull/751)

Given this structure:
```jsx
<Tabs.Root onKeyDown={handleKeyDown}>
  <Tabs.List>
    <Tabs.Tab value={0} />
    <Tabs.Tab value={1} />
  </Tabs.List>
  {/* tab panels */}
</Tabs.Root>
```
`Tabs.List` is a `CompositeRoot`, so on arrow keys, the event doesn't propagate and `handleKeyDown` doesn't get called

atomiks: I'm not certain on that. It may not be necessary in most cases

packages/react/src/Tabs/Tab/useTab.ts
```diff
@@ -56,28 +86,52 @@ function useTab(parameters: useTab.Parameters): useTab.ReturnValue {
...
     index,
     rootRef: handleRef,
-    // the `selected` state isn't set on the server (it relies on effects to be calculated),
-    // so we fall back to checking the `value` prop with the selectedValue from the TabsContext
-    selected: selected || value === selectedValue,
-    totalTabsCount,
-    orientation,
+    selected,
+    // TODO: recalculate this using Composite stuff, but is it really needed?
+    totalTabsCount: -1,
```
mj12albert, line 126: @michaldudak Is it ok to drop this? I found it wasn't actually used in the Tab component

packages/react/src/Composite/Root/CompositeRoot.tsx
```diff
@@ -74,9 +76,14 @@ namespace CompositeRoot {
     itemSizes?: Dimensions[];
     dense?: boolean;
     enableHomeAndEndKeys?: boolean;
+    onMapChange?: (newMap: Map<Node, CompositeMetadata<Metadata> | null>) => void;
+    stopEventPropagation?: boolean;
+    rootRef?: React.RefObject<HTMLElement | null>;
```
mj12albert, line 81: @michaldudak in the end James suggested using a custom ref prop to keep the types simple as this is an internal component

packages/react/src/utils/useOnMount.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * @ignore - internal hook.
+ */
+export { default as useOnMount } from '@mui/utils/useOnMount';
```
michaldudak, line 4: We're getting rid of @mui/utils dependency (https://github.com/mui/base-ui/pull/827), so let's define our utilities locally.

packages/react/src/Composite/Root/CompositeRoot.tsx
michaldudak: That's pretty non-standard for our codebase. The types aren't that much complex, so my preference would be to use `forwardRef`, but I'll be able ti live with this approach as well. Your call.

michaldudak:
You can remove the `useList` and `useCompound` hooks, as they are no longer used anywhere.

packages/react/src/Composite/Item/CompositeItem.tsx
```diff
@@ -10,23 +10,20 @@ import type { BaseUIComponentProps } from '../../utils/types';
...
+  const { render, className, itemRef, metadata, ...otherProps } = props;
 
   const { activeIndex } = useCompositeRootContext();
-  const { getItemProps, ref, index } = useCompositeItem();
+  const { getItemProps, ref, index } = useCompositeItem({ metadata });
 
   const ownerState: CompositeItem.OwnerState = React.useMemo(
     () => ({
-      active: index === activeIndex,
+      highlighted: index === activeIndex,
```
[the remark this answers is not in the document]
mj12albert: This originally came up [here](https://github.com/mui/base-ui/pull/763#discussion_r1835964235) ~ I'll change the names to match `highlighted`

michaldudak (approved):

PR #761, opened 2024-10-24
[docs] Render demos on static MDX routes

vladmoroz:
In static app routes, the following MDX will now render demos located in the `demo-hero` folder relative to the current file:

```mdx
<Demo path="./demo-hero" />
```

See `/new/components/dialog` route for a functional example (no style tweaks yet, should look visually broken but technically functional)


.eslintrc.js
```diff
@@ -39,8 +39,10 @@ module.exports = {
     ],
     '@typescript-eslint/no-redeclare': 'off',
 
-    // We LOVE non-breaking spaces here
+    // We LOVE non-breaking spaces, and both straight and curly quotes here
     'no-irregular-whitespace': [1, { skipJSXText: true, skipStrings: true }],
+    'react/no-unescaped-entities': [1, { forbid: ['>', '}'] }],
+    'material-ui/straight-quotes': 'off',
```
vladmoroz, line 45: We need to be able to write text with proper typography in JSX

docs/src/app/new/(content)/components/dialog/NestedDialogs.tsx
vladmoroz, line 1: Copy-pasta of the current nested dialogs demo, for testing purposes

docs/src/app/new/(content)/components/dialog/demo-hero/css-modules/index.js
vladmoroz, line 1: Copy-pasta of the current hero dialog demo, for testing purposes

docs/src/app/new/(content)/components/dialog/demo-hero/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as CssModules } from './css-modules';
+export { default as Tailwind } from './tailwind';
```
vladmoroz, line 2: We'll need to do an `index.ts` file for each demo folder so that it can be automatically imported into the MDX file

docs/src/components/demo/Demo.tsx
```diff
@@ -19,14 +19,12 @@ import { StackBlitzLink } from './StackBlitzLink';
...
 
 export interface DemoProps {
-  componentName: string;
-  demoName: string;
   variants: BaseDemo.DemoVariant[];
   defaultCodeOpen?: boolean;
 }
 
 export function Demo(props: DemoProps) {
-  const { componentName, demoName, variants: demoVariants, defaultCodeOpen = true } = props;
```
vladmoroz, line 29: imo not worth it to require a component name prop just so we can display a different CodeSandbox title, seems like unnecessary legwork for us

docs/src/components/demo/NewDemoLoader.tsx
vladmoroz, line 1: This is essentially a copy of `DemoLoader` and `loadDemos` with the following differences:
- If a demo can't be rendered for whatever reason, it will break the build
- The demo loader expects an absolute `path` to the demo to load and `scope` with components to add to scope  (handled by a rehype plugin, which injects an import statement into the MDX file and rewrites the path to absolute)
- It doesn't rely on dynamic Webpack import that currently processes all tsx/jsx files anymore. Docs should work faster when we are developing locally and ship 100 kB less into the first load JS bundle

docs/src/components/demo/rehypeDemos.mjs
vladmoroz, line 1: This is JS because it's imported into the Next.js config

If we upgrade Next.js to v15, we can convert this file to TypeScript, since they've just added TS config support:
https://nextjs.org/blog/next-15#support-for-nextconfigts

docs/src/app/new/(content)/components/dialog/UnstyledDialogIntroduction/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as CssModules } from './css-modules';
+export { default as Tailwind } from './tailwind';
```
vladmoroz, line 2: A small price to pay for static imports, we'll need to do copy around an `index.ts` file for each demo folder so that it can be imported into the MDX file

vladmoroz: All demos will have the same set of CSS options so this should be a non-issue

michaldudak (commented):
I'll have to dig deeper into the implementation but it's great to see it working!

docs/src/components/demo/rehypeDemos.mjs
```diff
@@ -0,0 +1,102 @@
+import { visit } from 'unist-util-visit';
```
michaldudak, line 1: Add `@ts-check` so it's easy to port this file to TS when it's moved to a separate library.

.eslintrc.js
michaldudak: FYI, there's a discussion on Notion about this: https://www.notion.so/Writing-style-guide-2a957a4168a54d47b14bae026d06a24b?d=aff8a8dc579c48f8a2500de4253c6949&pvs=4#b17ab6d6eb6f4589ba084207caf33443

vladmoroz: 2022! 😀

docs/src/components/demo/rehypeDemos.mjs
vladmoroz: Added (no difference)

vladmoroz: Actually nah, the types check fail now because a bunch of stuff is untyped.

I am not sure this is a good time to add these because we need to hand-write the types and it seems  a pretty poor experience to author and read them with complex type annotations. Will be easy to add them when we make it into a separate package, do you mind if we keep it as JS right now?

.eslintrc.js
michaldudak: Does anything changed significantly in this matter since then?

docs/src/components/demo/rehypeDemos.mjs
michaldudak: All right, let's do it then

michaldudak (approved):

PR #762, opened 2024-10-25
[docs] Upgrade to Next.js 15

vladmoroz:
Upgrade to Next.js 15, ~mainly for the TypeScript config.~

~Since we are going to use built-in @next/mdx features that are configured through the Next.js config, we need this in order to use rehype/remark plugins written in TypeScript.~

Edit: forget this, `next.config.ts` module resolution is CJS only, and all rehype/remark ecosystem is ESM

*** 

I also briefly tried enabling Turbopack, but it doesn't work with the current demos which rely on Webpack to dynamically load the components. I'll migrate to Turbopack when we don't do the dynamic Webpack import anywhere.


docs/package.json
```diff
@@ -70,8 +70,8 @@
     "@types/hast": "^3.0.4",
     "@types/node": "^18.19.54",
     "@types/prop-types": "^15.7.13",
-    "@types/react": "^18.3.10",
-    "@types/react-dom": "^18.3.0",
+    "@types/react": "npm:types-react@19.0.0-rc.1",
+    "@types/react-dom": "npm:types-react-dom@19.0.0-rc.1",
```
vladmoroz, line 74: This comes from their codemod

docs/package.json
```diff
@@ -39,14 +39,14 @@
...
     "postcss": "^8.4.47",
     "postcss-import": "^16.1.0",
     "prop-types": "^15.8.1",
-    "react": "^18.3.1",
-    "react-dom": "^18.3.1",
+    "react": "19.0.0-rc-69d4b800-20241021",
+    "react-dom": "19.0.0-rc-69d4b800-20241021",
     "react-error-boundary": "^4.0.13",
-    "react-is": "^18.3.1",
+    "react-is": "19.0.0-rc-69d4b800-20241021",
```
vladmoroz, line 49: This comes from their codemod

docs/shiki.config.ts
```diff
@@ -0,0 +1,4 @@
+export const shikiTheme = {
+  light: 'github-light-default',
+  dark: 'material-theme-ocean',
+};
```
vladmoroz, line 4: Next.js was complaining about exporting a `config` from the project root, seems like this used to be some pattern that is now deprecated

For context, this file will go completely when I'm done with code blocks updates

michaldudak:
I have a feeling that making the regression tests pass will require upgrading React in all our packages, including `test` and Base UI.

vladmoroz:
> I have a feeling that making the regression tests pass will require upgrading React in all our packages, including test and Base UI.

@michaldudak upgraded to React 19 everywhere and the regression tests now pass

~Should I worry about this?~

Edit: never mind, it passes after the rebase

image

~Running `pnpm proptypes` locally doesn't produce any changes~

michaldudak:
We can disable the static route indicator (https://nextjs.org/docs/app/api-reference/next-config-js/devIndicators#appisrstatus-static-indicator). All our routes are prerendered at build time, so it doesn't add any value.

michaldudak:
I configured the CI to also run tests against React 18, as 19 won't be widely adopted yet

vladmoroz:
> We can disable the static route indicator (https://nextjs.org/docs/app/api-reference/next-config-js/devIndicators#appisrstatus-static-indicator). All our routes are prerendered at build time, so it doesn't add any value.

@michaldudak nice, updated. Good to merge?

michaldudak (approved):

michaldudak:
FYI, ESM support in next.config.ts is coming: https://github.com/vercel/next.js/pull/68365

PR #764, opened 2024-10-28
[docs] New code blocks

vladmoroz:
See https://deploy-preview-764--base-ui.netlify.app/new/components/dialog


docs/src/app/(content)/layout.tsx
```diff
@@ -3,7 +3,8 @@ import * as React from 'react';
 import { AppBar } from 'docs/src/components/AppBar';
 import { Navigation } from 'docs/src/components/Navigation';
 import routes from 'docs/data/pages';
-import '../../styles/style.css';
+import 'docs/src/styles.css';
+import 'docs/src/styles/style.css';
```
vladmoroz, line 7: This is temporary just to support the current setup in place, `'docs/src/styles/style.css'` will go away eventually

vladmoroz:
@michaldudak do you have a clue why `docs` in paths breaks deploys? https://github.com/mui/base-ui/pull/764/commits/5c2549dffcf8e178e8c70ff5fd52f8d01ac50c40

I was having minor problems with path resolution that were fixed by adding `docs` path to tsconfig, but Netlify builds break then

michaldudak:
tsconfig fields aren't merged with parent config settings but replaced, so by adding `compilerOptions.paths`, you overwrote the paths defined in `/tsconfig.base.json`. Try adding the path to the base config instead.

michaldudak:
The demos on old pages are getting messed up a bit. Would it be problematic to actually edit the styles on the existing pages instead of the `/new` ones?

docs/src/styles.css
```diff
@@ -1,46 +1,52 @@
...
-@import 'tailwindcss/base';
-@import 'tailwindcss/components';
-@import 'tailwindcss/utilities';
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
 
 :root {
-  --color-content: #fff;
+  --font-mono: 'Menlo', 'Consolas', monospace;
```
michaldudak, line 11: Is the default font OK for Linux users?

docs/src/components/demo/loadDemo.ts
```diff
@@ -127,9 +126,9 @@ async function loadSimpleDemo(path: string, variantName: string): Promise<DemoVa
   const jsFilePath = mainFilePath.replace(/\.tsx?$/, '.js');
   if (mainFileLanguage === 'ts' && existsSync(jsFilePath)) {
     const jsContent = await readFile(jsFilePath, 'utf-8');
-    const jsPrettyPromise = await codeToHtml(jsContent, {
+    const jsPrettyPromise = highlighter.codeToHtml(jsContent, {
```
michaldudak, line 129:
```suggestion
    const jsPrettyContent = highlighter.codeToHtml(jsContent, {
```

docs/src/styles.css
vladmoroz: Good point, I'll add a couple more fallbacks

vladmoroz:
> The demos on old pages are getting messed up a bit. Would it be problematic to actually edit the styles on the existing pages instead of the `/new` ones?

Oh man this annoys me the hell out of me, but I don't know if it's worth our time to refactor styles on the old pages. The root issue is that `.Content ...` selector that targets all the elements on the page indiscriminately and messes up styles of anything you put on the page. I think we'll have to live with it for a bit, because otherwise I just need to refactor a thing we won't use soon at all

Anything particular that's messed up and bothers you?

vladmoroz:
@michaldudak looks like unrelated tests are failing after rebasing, I assume I can ignore them?

michaldudak:
>  I think we'll have to live with it for a bit, because otherwise I just need to refactor a thing we won't use soon at all
> 
> Anything particular that's messed up and bothers you?

It's nothing blocking. The main issue is that demos render in dark mode (following my `prefers-color-scheme`), while the rest of the page is light, and they have some extra horizontal scrollbars.

---

Aside from this, I just realized it could be nice to detect cmd/ctrl + a when a demo source is focused and select just the demo source instead of the whole page (could be done in a separate PR, of course)

michaldudak:
As for the broken tests - this looks like an issue after migrating to vitest. I'll see what's going on right away.

EDIT: likely fixed in https://github.com/mui/base-ui/pull/765

vladmoroz:
> Aside from this, I just realized it could be nice to detect cmd/ctrl + a when a demo source is focused and select just the demo source instead of the whole page (could be done in a separate PR, of course)

Agree, added that

> As for the broken tests - this looks like an issue after migrating to vitest. I'll see what's going on right away.
>
> EDIT: likely fixed in https://github.com/mui/base-ui/pull/765

Thanks!

colmtuite (approved):

michaldudak:
One thing I'm missing is the Reset button. It used to come in handy to reload the demo to its initial state.

vladmoroz:
> One thing I'm missing is the Reset button. It used to come in handy to reload the demo to its initial state.

We decided to remove it, among other things, since there was too much going on in the toolbar. I agree it has its uses, but it's probably more so for us as we maintain and test the components in the docs way more often than the end user would.

PR #771, opened 2024-10-30
[docs] Upgrade to Tailwind v4-alpha

vladmoroz:
Upgrading to remove unnecessary CSS variables spam in dev tools.

No changes to the produced CSS.

More details on the version:
https://tailwindcss.com/blog/tailwindcss-v4-alpha


docs/src/app/experiments/dialog.module.css
```diff
@@ -188,22 +188,22 @@
   padding: 16px;
   font-family: IBM Plex Sans;
 
-  h1 {
+  & h1 {
```
vladmoroz, line 191: All of a sudden CSS Modules started acting out, saying the selector isn't pure? Not sure how these are connected. Anyway, I like explicit `&`

docs/tailwind.config.ts
vladmoroz, line 1: - Config is now in CSS
- There's no `content` configuration anymore, they are saying it knows what to look at automatically

colmtuite (approved):

PR #804, opened 2024-11-05
[docs] Quick nav and base content styles

vladmoroz:


docs/src/components/quick-nav/QuickNav.tsx
vladmoroz, line 1: All the trouble here so we don't need a dedicated scrollbar on the quick nav here and it feels more like an intrinsic part of the document contents

https://github.com/user-attachments/assets/cea7e6bd-bfab-48d8-a56a-0f021ff62fd2

colmtuite (approved):

docs/src/components/ToolbarButton.css
```diff
@@ -24,8 +24,9 @@
...
         &::before {
           background-color: oklch(20% 12% 264 / 5%);
-
-          @media (prefers-color-scheme: dark) {
+        }
+        @media (prefers-color-scheme: dark) {
+          &::before {
             background-color: oklch(90% 5% 264 / 8%);
           }
         }
```
vladmoroz, line 32: Safari doesn't pick up `@media` nested in `::before`, seems like a browser bug

vladmoroz:
@atomiks @michaldudak do you want me to wait for your reviews before I merge? There's some thick logic going on here and there. I feel confident about it, but wouldn't mind an extra pair of eyes if you plan to take a look.

docs/src/components/demo/Demo.tsx
```diff
@@ -24,7 +24,11 @@ export function Demo({ className, defaultOpen = false, title, ...props }: DemoPr
   return (
     <BaseDemo.Root className={clsx('DemoRoot', className)} {...props}>
       <ErrorBoundary FallbackComponent={DemoErrorFallback}>
-        <BaseDemo.Playground aria-label="Component demo" className="DemoPlayground" />
+        <BaseDemo.Playground
+          suppressHydrationWarning
```
michaldudak, line 28: What's going on during hydration?

docs/src/components/demo/Demo.tsx
```diff
@@ -66,14 +70,13 @@ export function Demo({ className, defaultOpen = false, title, ...props }: DemoPr
 
             <Collapsible.Panel
               hidden={false}
+              aria-hidden={!open}
```
michaldudak, line 73: Shouldn't this be handled by Collapsible itself?

docs/src/components/quick-nav/QuickNav.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  // When nav height is larger than viewport, we make document scroll push it around
+  // so it sticks to top or bottom depending on scroll direction as if it was a real
+  // physical object constrained by the viewport
+  React.useEffect(() => {
+    if (!ref.current) {
+      return undefined;
+    }
+
+    let top: number;
```
michaldudak, line 16: It seems that this whole code block can be defined outside of the component, so it's not unnecessarily evaluated on every render.

docs/src/components/quick-nav/rehype-slug.mjs
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Converts a string into a well-formatted URL material, taking into account common contractions
+ * `"1. Here’s a wicked example & more." => "1-here-is-a-wicked-example-and-more"`
+ */
+function stringToUrl(string) {
+  return string
+    .replace(/([a-z])('|’|‘)ll([^a-z])/gi, '$1-will$3') // "you'll, we'll" => "you-will, we-will"
+    .replace(/won('|’|‘)t([^a-z])/gi, 'will-not$2') // "won't" => "will-not"
+    .replace(/can('|’|‘)t([^a-z])/gi, 'cannot$2') // "can't" => "cannot"
+    .replace(/n('|’|‘)t([^a-z])/gi, '-not$2') // "don't, doesn't, wouldn't, etc" => "do not, does not, would not, etc"
```
michaldudak, line 57: nit
```suggestion
    .replace(/n('|’|‘)t([^a-z])/gi, '-not$2') // "don't, doesn't, wouldn't, etc" => "do-not, does-not, would-not, etc"
```

docs/src/components/quick-nav/QuickNav.tsx
vladmoroz: Hmm we are inside `useEffect` with empty deps, so it's not evaluated on every render anyway, right? I'd prefer to keep these close to the stateful stuff if it doesn't matter

docs/src/components/demo/Demo.tsx
vladmoroz: Oh weird, I used to have hydration warnings here related to some control ID attributes in the rendered demos, but not seeing this anymore. Had shut it off while I was working on something else and forgot to get back to this, but it's not giving me the hydration warnings now 🤔 

I'll keep an eye on it, for now removed `suppressHydrationWarning`

docs/src/components/demo/Demo.tsx
vladmoroz: Collapsible does the `hidden` attribute out of the box which hides the panel altogether and handles the accessibility.

In this case I needed it to be visibly rendered on the page, but hidden from the accessibility tree when it's not expanded, so I had to add `aria-hidden`

colmtuite:
Bug discovered, not sure how to recreate or what is causing it.

https://github.com/user-attachments/assets/ef9b3071-9477-4561-aeed-5bb6cbafb22b
https://github.com/user-attachments/assets/f2f9f258-5eb2-4c58-8ed7-ca629899f733

colmtuite:
Strange white flash bug in top right corner of the viewport when I scroll up quickly. Probably just a Chrome issue because it's a long document. UFO?

https://github.com/user-attachments/assets/21e7e39c-42ed-45eb-9b1d-6346489acc35

vladmoroz:
> Strange white flash bug in top right corner of the viewport when I scroll up quickly. Probably just a Chrome issue because it's a long document. UFO?

Yeah I'm seeing this too, it's not related to quick nav. Same for me when quick nav is not rendered at all:

https://github.com/user-attachments/assets/4f94d2e2-74ea-480f-95dc-77071c12cb9a

Nothing in the profiler either. I had spent some time trying to figure this one out yesterday but gave up. Anyone have ideas?

> Bug discovered, not sure how to recreate or what is causing it.
>
> https://github.com/user-attachments/assets/ef9b3071-9477-4561-aeed-5bb6cbafb22b
> https://github.com/user-attachments/assets/f2f9f258-5eb2-4c58-8ed7-ca629899f733

These are legitimate bugs, need help getting into this state though

atomiks:
Twitter's right sidebar uses `position: sticky` and pins to `top: -{value}px` when scrolling down or `bottom: -{value}px` when scrolling up, which should replicate this effect? 

I can't see the white flash on my end however

vladmoroz:
> Twitter's right sidebar uses `position: sticky` and pins to `top: -{value}px` when scrolling down or `bottom: -{value}px` when scrolling up, which should replicate this effect?
> 
> I can't see the white flash on my end however

Yep this is similar and the only similar experience I ever saw

We have different layout constraints here so the way it's done is somewhat different / more complete too

docs/src/components/quick-nav/QuickNav.tsx
michaldudak: It's not run on every render but it is evaluated (new functions are being allocated, etc.)

docs/src/components/demo/Demo.tsx
michaldudak: gotcha!

docs/src/components/quick-nav/QuickNav.tsx
vladmoroz: Is this right?
https://github.com/mui/base-ui/pull/804/commits/b087f7caa8e727a43a29ab095af61f484d8a9d4b?diff=split&w=1

vladmoroz: Fair, it's rendered once per page load so in my understanding it won't impact anything tangible

Issue #728, opened 2024-10-09
[collapsible] Support`keepMounted` prop


mj12albert:
Follow up to https://github.com/mui/base-ui/pull/577#pullrequestreview-2345745521

> Do we have an API to not mount the children of an accordion? Something similar to https://mui.com/material-ui/react-accordion/#performance. When rendering heavy components under an accordion, e.g. in a dashboard: charts, data grid, etc. It can be a lifesaver for the performance responsiveness on the page.
> 
> Or is the idea to solve this problem with a new component of this kind: https://ark-ui.com/react/docs/utilities/presence?

colmtuite:
This seems back-to-front to me. Wouldn't it be better—and more in-line with our other components—if the panel contents unmounted by default, and then you can use the `keepMounted` prop if you need it?

cc @michaldudak

michaldudak:
I'd name the prop `keepMounted` for sure. As for its default value, I don't have a strong opinion, but I lean towards making it consistent with other components (so `false`).

mj12albert:
@colmtuite I agree with naming this as `keepMounted` instead

I think it makes sense for the default for this to be `true`, because [generally](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden#the_hidden_state) browsers implement the behavior of the `hidden` attribute with CSS `display: none`, which doesn't remove the Panel (and contents) from the DOM

And for other components, `keepMounted` is needed for animations using Framer; while for `Collapsible`, it would be for performance, e.g. if you have a lot of `Collapsible`s with heavy components in the Panel as @oliviertassinari mentioned, what do you think?

PR #807, opened 2024-11-06
[collapsible][accordion] Add `keepMounted` prop

mj12albert:
Closes https://github.com/mui/base-ui/issues/728

We decided to default `keepMounted` to `false` for consistency with other components.

Since `hiddenUntilFound` requires `keepMounted={true}` to work, using it will override `keepMounted` and show a warning in dev mode if both `hiddenUntilFound` and `keepMounted={false}` are specified.


mj12albert:
@colmtuite @vladmoroz Do you think it's an issue that 2 props are required to use hidden-until-found now?
```jsx
<Collapsible.Panel hiddenUntilFound keepMounted>
  {/* content */}
</Collapsible.Panel>
```

vladmoroz:
@mj12albert I think we should always mount the panel when `hiddenUntilFound` is used, regardless of `keepMounted`

michaldudak:
If `keepMounted` is undefined, then yes, we can make `hiddenUntilFound` make the component always mounted. But when it's explicitly set to `false`, I'd show a warning (in dev mode only, though).

packages/mui-base/src/Collapsible/Panel/CollapsiblePanel.test.tsx
```diff
@@ -37,4 +49,86 @@ describe('<Collapsible.Panel />', () => {
...
+        panel?.dispatchEvent(event);
+      });
+
+      expect(handleOpenChange.callCount).to.equal(1);
+      expect(panel).to.have.attribute('data-open');
+    });
+
+    // toWarnDev doesn't work reliably with async rendering. To re-enable after it's fixed in the test-utils.
+    // eslint-disable-next-line mocha/no-skipped-tests
+    it.skip('warns when setting both `hiddenUntilFound` and `keepMounted={false}`', async function test() {
```
mj12albert, line 118: I tested this warning manually 😓

mj12albert:
Updated ~ would appreciate a review @vladmoroz @michaldudak 🙏

mj12albert:
Also includes a lot of rework to the panel internals as the Select PR changed the timing of the `useTransitionStatus` hook and set `data-entering` 1 animation frame sooner than before: https://github.com/atomiks/base-ui/commit/cfd185a36784341e7165d17c5fec927fc710e4a8#diff-ecb73c88ec0196a8b56a56bb2b147a4d57e6f461c854c51e7a3109408336fb6c

This broke CSS transition handling, as it was using that 1 frame between the open state changing and `data-entering` to measure the fully expanded size of the panel contents before applying the transition (which starts when `data-entering` is applied)

mj12albert:
Continued in https://github.com/mui/base-ui/pull/889

Requires https://github.com/mui/base-ui/pull/887

PR #810, opened 2024-11-07
[docs] Basic API reference tables

vladmoroz:
image


docs/src/components/ApiTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+            </Table.Cell>
+            <Table.Cell>
+              {prop.defaultValue ? (
+                <code className="text-blue">{prop.defaultValue}</code>
+              ) : (
+                <code className="text-pale">undefined</code>
+              )}
+            </Table.Cell>
+            <Table.Cell>
+              <Popover.Root>
```
atomiks, line 42: Add `openOnHover` here

docs/src/syntax-highlighting/rehype-inline-code.mjs
```diff
@@ -1,5 +1,6 @@
 import { visitParents } from 'unist-util-visit-parents';
 
+/** Adds basic syntax highlighting to inline `<code>` elements */
 export function rehypeInlineCode() {
```
atomiks, line 4: Unrelated, but I'm curious why you're doing this rather than using `rehype-pretty-code`'s inline highlighting feature? Rather than only a single color for the inline code, it accurately highlights it (punctuation differences, etc.) like it were a regular code block.

vladmoroz: Gave it another go, it's pretty good! Thanks for pointing in out

Initially I couldn't figure out why it wasn't working and didn't have the patience to debug; was just missing the default language option

docs/src/components/PropsTableTooltip.tsx
```diff
@@ -0,0 +1,98 @@
...
+import * as React from 'react';
+import { ToolbarButton } from './ToolbarButton';
+import { Popup } from './Popup';
+
+export function PropsTableTooltip({ children }: React.PropsWithChildren) {
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    // TODO: this was more effort than I expected, we need to chat about popovers that open on hover
+    <Tooltip.Root
```
atomiks, line 12: Curious what happened here? Is it just that `<Tooltip.Provider>` doesn't exist for popovers? This current pattern isn't accessible unlike the `<Popover.Root openOnHover>` one

vladmoroz: Yeah I have a point to chat through this on Monday, going to be easier to show.

For now I made it so that the interactions feel good and SR will read out the prop description when entering the cell as if it was regular text.

vladmoroz: Reverted to popovers that open on click while we figure it out

colmtuite (approved):

PR #815, opened 2024-11-12
[internal] Use `hidden` prop for `keepMounted` elements

atomiks:

---

Popup animations: https://deploy-preview-815--base-ui.netlify.app/experiments/tooltip

- Prevents the need to apply CSS to use `keepMounted` by default
- Styling animations is now the same as conditional rendering, so it's overall more consistent. They need to use `[data-entering]` and wait 1 frame as well, once `hidden` gets removed on open
- `inert` is no longer necessary for these components
- Fixes anchor positioning performance for `keepMounted` components, both on initial mount and for the lifetime of the mount


packages/mui-base/src/utils/useTransitionStatus.ts
```diff
@@ -19,6 +19,9 @@ export function useTransitionStatus(open: boolean, enabled = true) {
   if (enabled) {
     if (open && !mounted) {
       setMounted(true);
+      if (transitionStatus !== 'entering') {
+        setTransitionStatus('entering');
+      }
```
atomiks, line 24: This is necessary after the change, and also seems to fix the CSS Transition Group demo in the popups experiment. Likely should have been the case originally.

packages/mui-base/src/Popover/Positioner/usePopoverPositioner.tsx
```diff
@@ -9,13 +9,15 @@ import type {
 import { mergeReactProps } from '../../utils/mergeReactProps';
 import { useAnchorPositioning } from '../../utils/useAnchorPositioning';
 import type { GenericHTMLProps } from '../../utils/types';
-import { getInertValue } from '../../utils/getInertValue';
```
michaldudak, line 12: Is `getInertValue` still used anywhere? If not, we can remove it.

packages/mui-base/src/Popover/Positioner/usePopoverPositioner.tsx
```diff
@@ -26,7 +28,7 @@ export function usePopoverPositioner(
     renderedSide,
     renderedAlignment,
     positionerContext,
-  } = useAnchorPositioning(params);
+  } = useAnchorPositioning({ ...params, mounted });
```
michaldudak, line 31: Destructuring the object manually could be faster at runtime.

michaldudak (commented):
Looks good overall. I have just some minor comments.

packages/mui-base/src/Popover/Positioner/usePopoverPositioner.tsx
atomiks: Looks like it's used for `CheckboxRoot` input, but given it has `tabIndex: -1` and `aria-hidden: true`, I don't think that's necessary?

packages/mui-base/src/Popover/Positioner/usePopoverPositioner.tsx
atomiks: What do you mean destructuring? To avoid the spread here, it'd need to be part of `params` (which is doable but not really necessary)

atomiks: It looks like `mounted` already exists in the params for some other components, so changing to that for consistency

michaldudak: I meant avoiding the `...` operator and creating the object (as per https://www.notion.so/mui-org/Performance-guide-memory-allocations-134cbfe7b66080cba325db614278fc20?pvs=4#134cbfe7b66080cc998de00ce75070a7), so
```
useAnchorPositioning({
  anchor: params.anchor,
  side: params.side,
  // ...
  mounted
});
```

atomiks: I refactored it so that `mounted` now exists in params so it can avoid the spread operator (some other components already had it like that)

michaldudak:
Dialog and AlertDialog were not updated. The rest looks OK.

michaldudak (approved):

PR #819, opened 2024-11-15
[popover] Don't stick upon click with `openOnHover` prop

atomiks:

The upgrade fixes bugs and adds this prop to turn off.

### DIFF
--- a/packages/react/src/Popover/Trigger/PopoverTrigger.test.tsx
+++ b/packages/react/src/Popover/Trigger/PopoverTrigger.test.tsx
@@ -51,9 +51,29 @@ describe('<Popover.Trigger />', () => {
       expect(trigger).not.to.have.attribute('data-pressed');
     });
 
+    it('should not have the data-popup-open and data-pressed attributes when open by click when `openOnHover=true` and `delay=0`', async () => {
+      const { user } = await render(
+        <Popover.Root delay={0} openOnHover animated={false}>
+          <Popover.Trigger />
+        </Popover.Root>,
+      );
+
+      const trigger = screen.getByRole('button');
+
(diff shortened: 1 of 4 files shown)


michaldudak:
Seems to work fine.

> adds this prop to turn off.

There's no prop exposed to control this behavior?

atomiks:
> There's no prop exposed to control this behavior?

Not currently but there could be. @vladmoroz

vladmoroz:
I wouldn't want to add a prop unless necessary—is there a typical design where it makes sense to keep it open on click?

atomiks:
> I wouldn't want to add a prop unless necessary—is there a typical design where it makes sense to keep it open on click?

Back when I was developing tippy.js, the "stick" behavior was actually the requested default (a while ago, 2019), and it wasn't configurable. I didn't hear of complaints. It's also the default in Floating UI now (working better with the recent fixes).

packages/react/src/Popover/Trigger/PopoverTrigger.test.tsx
```diff
@@ -51,9 +51,29 @@ describe('<Popover.Trigger />', () => {
...
+      );
+
+      const trigger = screen.getByRole('button');
+
+      await user.hover(trigger);
+
+      // Closes instantly after opening due to `useClick` prop `stickIfOpen=false`
+      await act(async () => {
+        trigger.click();
+      });
```
atomiks, line 68: Perhaps this is why it's configured by default, but it depends on the `delay` you use. If you happen to click <50ms after it opens coincidentally, it's pretty unexpected

vladmoroz (approved):

PR #820, opened 2024-11-15
[core] Rename CSS vars to include component name

atomiks:

### DIFF
--- a/docs/src/components/Popup.css
+++ b/docs/src/components/Popup.css
@@ -1,7 +1,7 @@
 @layer components {
   .Popup {
-    max-width: var(--available-width);
-    max-height: var(--available-height);
+    max-width: var(--popover-available-width);
+    max-height: var(--popover-available-height);
     border-radius: 0.375rem;
     background-color: var(--color-popup);
     overflow: hidden;
@@ -16,7 +16,7 @@
       outline-offset: -1px;
     }
 
-    transform-origin: var(--transform-origin);
+    transform-origin: var(--popover-arrow-origin);
     transition-property: opacity, transform;
     transition-timing-function: cubic-bezier(0.3, 1.065, 0.01, 0.975);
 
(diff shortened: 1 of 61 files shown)


docs/src/components/Popup.css
```diff
@@ -1,7 +1,7 @@
 @layer components {
   .Popup {
-    max-width: var(--available-width);
-    max-height: var(--available-height);
+    max-width: var(--popover-available-width);
+    max-height: var(--popover-available-height);
```
atomiks, line 4: If we add component prefixes, this makes classes like this non-reusable across different popups... @vladmoroz. In fact, I'm not sure if this popup is used for popover specifically.

packages/mui-base/src/utils/useAnchorPositioning.ts
```diff
@@ -179,14 +181,14 @@ export function useAnchorPositioning(
...
-        const transformOrigin = {
+        const arrowOrigin = {
           top: `${transformX}px calc(100% + ${arrowHeight}px)`,
           bottom: `${transformX}px ${-arrowHeight}px`,
           left: `calc(100% + ${arrowHeight}px) ${transformY}px`,
           right: `${-arrowHeight}px ${transformY}px`,
         }[currentRenderedSide];
 
-        elements.floating.style.setProperty('--transform-origin', transformOrigin);
+        elements.floating.style.setProperty(`--${name}-arrow-origin`, arrowOrigin);
```
atomiks, line 192: The arrow origin and transform origin are a bit different. For edge-aligned popups, you want the `transform-origin` to be at the edge, not arrow's point/anchor center.

docs/src/components/Popup.css
vladmoroz: Hmm yeah good point, I was thinking that the Popup component would be used across different popups, such as Popovers, Preview Cards, Dialogs, etc. Let's discuss it on Monday in detail. We need to consider whether prefixing everything maps well to how components would be styled.

Issue #824, opened 2024-11-15
[Composite] Support navigating to the first/last items with the Home/End keys


mj12albert:
Used for
- Tabs: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction
- Toolbar: https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/#keyboardinteraction

**Search keywords**:

PR #825, opened 2024-11-15
[Composite] Support RTL navigation, Home/End keys

mj12albert:
Split from https://github.com/mui/base-ui/pull/763 to unblock https://github.com/mui/base-ui/pull/751
Closes https://github.com/mui/base-ui/issues/824

- Adds support for RTL navigation with arrow keys by detecting the CSS `direction` property that could be inherited from the HTML `dir` attribute
- Adds support for navigating to the first/last items with Home/End keys, enable-able with the `enableHomeAndEndKeys` prop

### DIFF
--- a/packages/mui-base/src/Composite/Root/useCompositeRoot.ts
+++ b/packages/mui-base/src/Composite/Root/useCompositeRoot.ts
@@ -172,29 +172,19 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
             ] as number; // navigated cell will never be nullish
           }
 
-          const toEndKeys = isRtl
-            ? {
-                horizontal: [ARROW_LEFT],
-                vertical: [ARROW_DOWN],
-                both: [ARROW_LEFT, ARROW_DOWN],
-              }[orientation]
-            : {
-                horizontal: [ARROW_RIGHT],
-                vertical: [ARROW_DOWN],
-                both: [ARROW_RIGHT, ARROW_DOWN],
-              }[orientation];
+          const horizontalEndKey = isRtl ? ARROW_LEFT : ARROW_RIGHT;
+          const toEndKeys = {
+            horizontal: [horizontalEndKey],
+            vertical: [ARROW_DOWN],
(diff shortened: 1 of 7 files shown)


packages/mui-base/src/Composite/Root/useCompositeRoot.ts
```diff
@@ -178,10 +229,7 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
 
             onActiveIndexChange(nextIndex);
 
-            // Wait for FocusManager `returnFocus` to execute.
-            queueMicrotask(() => {
-              elementsRef.current[nextIndex]?.focus();
-            });
```
mj12albert, line 184: @atomiks This broke a RadioGroup test... https://app.circleci.com/pipelines/github/mui/base-ui/4867/workflows/dbe5234b-0110-4b20-8b7e-d8dfd310df29/jobs/34622/parallel-runs/0/steps/0-106
Any idea why?

mj12albert: The test breaks but it still works in the browser, I've separated this for now

atomiks (approved):

packages/mui-base/src/Composite/Root/useCompositeRoot.ts
```diff
@@ -130,31 +166,54 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
...
+            ? {
+                horizontal: [ARROW_LEFT],
+                vertical: [ARROW_DOWN],
+                both: [ARROW_LEFT, ARROW_DOWN],
+              }[orientation]
+            : {
+                horizontal: [ARROW_RIGHT],
+                vertical: [ARROW_DOWN],
+                both: [ARROW_RIGHT, ARROW_DOWN],
+              }[orientation];
```
atomiks, line 185: This could be a bit simpler by assigning `const horizontalEndKey = isRtl ? ARROW_LEFT ? ARROW_RIGHT` and same for `start`

PR #827, opened 2024-11-18
[core] Remove @mui/utils and @mui/types dependencies

michaldudak:


packages/mui-base/src/utils/useId.test.tsx
```diff
@@ -0,0 +1,101 @@
+import * as React from 'react';
+import { expect } from 'chai';
+import { createRenderer, screen } from '@mui/internal-test-utils';
+import useId from '@mui/utils/useId';
```
oliviertassinari, line 4: Looks wrong ⚠️

packages/mui-base/src/utils/useControlled.test.tsx
```diff
@@ -0,0 +1,125 @@
+import * as React from 'react';
+import { expect } from 'chai';
+import { act, createRenderer } from '@mui/internal-test-utils';
+import { useControlled } from './useControlled';
```
oliviertassinari, line 4: What if we were to already expose them as a new npm package https://www.notion.so/mui-org/engineering-mui-utils-purpose-9a9fc9da3a004864b6c4e1f4d1f24f95?p=56e98feced4d4fe4a73df99e0b0361b4&pm=s while we are at it? so that Material UI, MUI X, can migrate to those and @mui/utils can deprecate those utils?

Or maybe it's too big of a jump at once?

packages/mui-base/src/utils/useControlled.ts
```diff
@@ -1,4 +1,77 @@
...
+  const { current: isControlled } = React.useRef(controlled !== undefined);
+  const [valueState, setValue] = React.useState(defaultProp);
+  const value = isControlled ? controlled : valueState;
+
+  if (process.env.NODE_ENV !== 'production') {
+    React.useEffect(() => {
+      if (isControlled !== (controlled !== undefined)) {
+        console.error(
+          [
+            `MUI: A component is changing the ${
```
oliviertassinari, line 41: Maybe?

```suggestion
            `Base UI: A component is changing the ${
```

 or we could remove this, and hope that https://github.com/facebook/react/issues/28992#issuecomment-2104395451 will get fixed? 

```suggestion
            `A component is changing the ${
```

michaldudak: It's an interesting problem. If this is imported in Material UI, users might get confused where does this "Base UI" come from. But it's equally true for all the other error messages (such as missing context providers). We might be better off removing the prefix in all cases. CC @atomiks

packages/mui-base/src/utils/useControlled.test.tsx
michaldudak: Let's do it incrementally. As I noted in https://www.notion.so/engineering-mui-utils-purpose-9a9fc9da3a004864b6c4e1f4d1f24f95?d=143cbfe7b66080e1ad16001c70691469&pvs=4#7cb73e22aaee40eb9c35cca853ffc73f, I'm not particularly keen on placing this kind of utilities in a separate package. I think they are useful to build styled libraries and should be exposed from @base-ui-components/react directly.

packages/mui-base/src/utils/useControlled.ts
atomiks: It could be possible to configure the prefix name at runtime before any messages are logged

```js
import {log} from '@base-ui-components/react';

log.prefix('Base UI: ');
```

Then we could use utilities to throw errors:

```js
log.error('...'); // throws, prefixed
log.warn('...'); // dedupes, prefixed
```

michaldudak: :+1: I like it. I wouldn't do this in this PR, though. I'll change the prefix to Base UI here, since for now only we use these utils, and create a follow-up one to configure it.

atomiks (approved):

PR #832, opened 2024-11-19
[core] Make callback parameters consistent

michaldudak:
- Renamed the first parameter to match the callback name (`value` in `onValueChange`, `checked` in `onCheckedChange`, etc.).
- Using native events in all public event handlers.
- Renamed `SliderRoot.onValueChange` second parameter to `activeThumbIndex` and moved it to the third position.

### DIFF
--- a/docs/data/api/slider-root.json
+++ b/docs/data/api/slider-root.json
@@ -15,8 +15,8 @@
     "onValueChange": {
       "type": { "name": "func" },
       "signature": {
-        "type": "function(value: number | Array<number>, activeThumb: number, event: Event) => void",
-        "describedArgs": ["value", "activeThumb", "event"]
+        "type": "function(value: number | Array<number>, event: Event, activeThumbIndex: number) => void",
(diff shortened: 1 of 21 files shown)


vladmoroz (approved):

vladmoroz (commented):
Not seeing `docs/reference` changes, could this branch be based off an older commit?

michaldudak:
No, it's from the lastest master. Is the `docs:api` script supposed to update these files?

vladmoroz:
>No, it's from the lastest master. Is the docs:api script supposed to update these files?

Duh I realised it's not updated because there is an override on Slider's `onValueChange` type

image

Check `docs/reference/overrides/slider-root.json`

michaldudak:
:/ that's an unfortunate consequence of having the definitions in several places. I hope we'll have time to do this properly soon.

vladmoroz:
Yeah totally agree @michaldudak 
Will start getting a lot easier though when we don't have two sets of API data for the docs

Issue #623, opened 2024-09-16
[dialog][popover][menu][select] Modality and focus changes


vladmoroz:
**Dialogs**
- Tab keypresses loop focus regardless of the `modal` prop value
  - `modal` will only mean "pointer" modality going forwards, which includes preventing outside clicks and scroll locking
  - Tabbing out should never close the popup
  - Focus is not trapped per se, it should still be able to escape the non-modal non-dismissible popups (for example, in Figma's non-dismissible popups keyboard focus is looped, but can be moved outside of the popup via pointer interactions).
- `Backdrop` part becomes optional and is only for presentation
  - Outside clicks should be prevented regardless of whether `Backdrop` is present
- Expose `reason` argument in `onOpenChange` to make it possible to prevent close on interactions like Esc keypress

**Popover**
- `modal` prop is added, false by default. Modality should work exactly the same as in dialogs.
  - The final decision re. whether default is true/false is contingent on whether [touch should be treated as always modal](https://github.com/mui/base-ui/issues/653)
  - Popovers that open on hover shouldn't steal current focus so that you can "unhover" the trigger; however, once focus is manually placed into such popover, keyboard focus is looped normally.
  - We need to explore/research how to make it work with touch screen readers if an explicit close button isn't rendered

**Menu, Select**
- `modal` prop is added, false by default
  - Same as popovers re. the final decision re. whether default is true/false
  - Menus should still close on the first Tab keypress
- Menus that open on hover also shouldn't steal focus until it's placed into the popup manually via hovering an item or arrow keypress

Context:
https://www.figma.com/design/Y6VArf17CAhMUc1DBWKHIi/Modality?node-id=0-1&t=kN6j5qsqKlWGNn1B-1
https://www.notion.so/mui-org/Vlad-James-modality-107cbfe7b66080f5a68cfd7450da9433?pvs=4

PR #841, opened 2024-11-20
[Dialog, Popover, Menu, Select, PreviewCard] Unify backdrop implementation

michaldudak:
Unified and simplified backdrop implementation:

- removed inline styles
- set `role: 'presentation'` in all backdrops
- removed use*Backdrop hooks
- configured FloatingFocusManager to add `inert` attributes to elements outside modal dialogs. This removes the need for an internal backdrop
- added the `container` prop to DialogBackdrop and AlertDialogBackdrop
- added the MenuBackdrop

part of https://github.com/mui/base-ui/issues/623


atomiks (commented):
`Menu` didn't have a Backdrop in the first place, but should?

michaldudak:
Added.

packages/react/src/Dialog/Popup/DialogPopup.tsx
```diff
@@ -114,6 +114,7 @@ const DialogPopup = React.forwardRef(function DialogPopup(
 
   return (
     <FloatingPortal root={container}>
+      {modal && mounted && <FloatingOverlay />}
```
atomiks, line 117: What about `z-index`?

atomiks (commented):
Is the bug that causes the `Backdrop` to be inserted into the DOM after the `Popup` fixed with this? Not sure why it would occur exactly but I'm assuming the way `FloatingPortal` creates a root container could be an issue. There is one minor part that has asynchronicity related to cleanup, but not sure if that would be the cause.

packages/react/src/Dialog/Popup/DialogPopup.tsx
michaldudak: Yeah, right, this is going to be tricky. We could figure out the z-index of the popup and set this one lower, but this will be unreliable in some cases. We might have to expose the `className` of this part, or the whole part itself, to be consistent with the rest of the API.

vladmoroz: Is this another problem solved by a separate Portal part? 😅

michaldudak:
I don't know TBH. I could reproduce it neither on master nor here.
@vladmoroz, do you have a reliable repro?

vladmoroz:
> I don't know TBH. I could reproduce it neither on master nor here.
> @vladmoroz, do you have a reliable repro?

If its about Backdrop overlaying the Popup, I think I had a transition on the Popup, and no transition on the Backdrop

packages/react/src/Dialog/Popup/DialogPopup.tsx
michaldudak: Not really, unless I'm missing something. Having an explicit Portal wouldn't help determining the right z-index of the internal backdrop.
Perhaps a better solution is not rendering the internal backdrop, but adding the `inert` attribute to all elements except the popup?

atomiks:
> If its about Backdrop overlaying the Popup, I think I had a transition on the Popup, and no transition on the Backdrop

I was able to reproduce it consistently (once every ~5-10 times) on master previously when clicking very quickly on the trigger and backdrop - the backdrop did have a normal transition along with the popup. But on latest master, and this PR on the new dialog page, I haven't been able to reproduce it at all. Might be fixed?

packages/react/src/Dialog/Popup/DialogPopup.tsx
atomiks: > Perhaps a better solution is not rendering the internal backdrop, but adding the inert attribute to all elements except the popup?

This could work. `guards={false}` on `FloatingFocusManager` does this, but it allows escaping to the browser address bar of course, which we didn't want to allow. Guess this is relevant: https://github.com/floating-ui/floating-ui/discussions/3098

The other option is to get the computed `z-index` style of the `Popup` and apply the `z-index` to be just one less on `FloatingOverlay`?

michaldudak:
I changed the implementation from the internal backdrop to making all the outside elements inert (with https://github.com/floating-ui/floating-ui/pull/3131).

atomiks (approved):

PR #844, opened 2024-11-20
[core] Rename the library directory

### LINKED ISSUE #482: [core] Confusing source folder name
### Steps to reproduce

The sources are located at https://github.com/mui/base-ui/tree/master/packages/mui-base/ but the npm package name is `@base_ui/react`. So I would expect the source to be located at https://github.com/mui/base-ui/tree/master/packages/base-ui-react/.

michaldudak:
Renamed the packages/mui-base to packages/react.
Also updated the docs domain name in API docs gen config (from base-ui.netlify-app to base-ui.com).

Closes #482


vladmoroz:
Totally a turbo nit, but while we are here, should the root package.json use something like "base-ui-components" or "@base-ui-components/monorepo" instead of "@base_ui/monorepo" for the name field?

atomiks (approved):

vladmoroz (approved):

PR #845, opened 2024-11-20
[docs] Place Quick Nav after the main demo in the DOM

vladmoroz:
Place Quick Nav after the main demo in the DOM so that keyboard focus follows the reading order

https://deploy-preview-845--base-ui.netlify.app/new/components/dialog

### DIFF
--- a/docs/src/components/demo/Demo.css
+++ b/docs/src/components/demo/Demo.css
@@ -3,6 +3,15 @@
     background-color: var(--color-content);
     border: 1px solid var(--color-gray-200);
     border-radius: var(--radius-md);
+
+    /* Coordinate the placement of Quick Nav that follows the main Demo */
+    &[data-before-quick-nav] {
+      float: left;
+      width: 100%;
+      & + * + * {
+        clear: left;
+      }
+    }
   }
 
   .DemoPlayground {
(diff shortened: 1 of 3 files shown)


docs/src/components/demo/index.css
```diff
@@ -3,6 +3,15 @@
...
     border: 1px solid var(--color-gray-200);
     border-radius: var(--radius-md);
+
+    /* Coordinate the placement of Quick Nav that follows the main Demo */
+    &[data-before-quick-nav] {
+      float: left;
+      width: 100%;
+      & + * + * {
+        clear: left;
+      }
```
vladmoroz, line 13: A thinker figure looking at the lobotomised owl

colmtuite:
@vladmoroz Imo, this won't be expected behaviour when headings become links.

vladmoroz:
> @vladmoroz Imo, this won't be expected behaviour when headings become links.

- Hmm what's the expected behaviour in that case?
- I wouldn't do heading links personally, I think quick nav is perfectly sufficient for grabbing a link to a section. I like plain, non-clickable headings. Would this be too much of a contentious choice?

colmtuite:
Discussed on a call:

Imo, headings should be links in some fashion. I appreciate that it's annoying when the actual heading is clickable, it's nice to be able to select or click headings for no reason at all while you're reading. But we could put a link icon beside each heading, or something similar.

Imo, when focus is in the main content pane, it should remain there. Sidebar will/should be considered a separate pane. So focus should either go to sidebar first or last.

vladmoroz:
Resolved differently by another PR

PR #843, opened 2024-11-20
[docs] Rework colors and refactor components

vladmoroz:
https://deploy-preview-843--base-ui.netlify.app/new/components/dialog

- Rework colors
  - Introduce a gray scale aligned with Tailwind grays in lightness.
    - This is so that the demos can be built with the same color system as the rest of the docs
    - Remove some of the unnecessary functional colors
  - Improve the appearance in dark mode
- Fix various design issues
  - Make sure table rows have consistent height
  - Use larger line height in tables
  - Fix unintended code letter spacing
  - Style the Select component for the code variant picker
  - Expand the demo collapsible when choosing a different code variant
  - Rework `ToolbarButton` into a `GhostButton`, remove unreliable styles
  - Make reference table popovers open on hover

PR #847, opened 2024-11-21
[select] Fix `min-height` fallback check

atomiks:

When the natural scroll height of the select popup is less than the default `min-height` for fallback (`100px`), it always falls back. This changes the check to only fallback if its natural height exceeds the `min-height` (which can also be customized already) and is less than it.

ref: https://github.com/mui/base-ui/pull/843#discussion_r1850378253


vladmoroz (approved):
Works great

PR #855, opened 2024-11-22
[popover] Disable focus management when opened via hover

atomiks:

### DIFF
--- a/packages/react/src/Popover/Root/usePopoverRoot.ts
+++ b/packages/react/src/Popover/Root/usePopoverRoot.ts
@@ -31,7 +31,6 @@ export function usePopoverRoot(params: usePopoverRoot.Parameters): usePopoverRoo
     open: externalOpen,
     onOpenChange: onOpenChangeProp = () => {},
     defaultOpen = false,
-    keepMounted = false,
     delay,
     closeDelay,
     openOnHover = false,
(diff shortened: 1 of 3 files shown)


packages/react/src/Popover/Root/usePopoverRoot.ts
```diff
@@ -69,22 +68,21 @@ export function usePopoverRoot(params: usePopoverRoot.Parameters): usePopoverRoo
     (nextOpen: boolean, event?: Event, reason?: OpenChangeReason) => {
       onOpenChange(nextOpen, event, reason);
       setOpenUnwrapped(nextOpen);
-      if (!keepMounted && !nextOpen) {
```
atomiks, line 72: Need to check how `setMounted(false)` with this `keepMounted` check operates now for the other components with the `hidden` prop

atomiks: Looks like `keepMounted` was never passed to the hooks, hence it worked like `keepMounted=false`. I've removed the prop for Popover but it should be done for all components.

michaldudak:
It feels wrong that the popover closes on mouseout after an input inside it has been focused. Clicking on anything inside the popover should make it sticky

atomiks:
@michaldudak that's a separate discussion as the previous focus management also worked like this

michaldudak (approved):

PR #861, opened 2024-11-25
[ScrollArea] Fix margin/padding scrollbar calculations

atomiks:

The user can specify padding or margin in their CSS to offset the scrollbar, but these weren't being handled correctly by the calculations; specifically:
- `padding` on the `Scrollbar` part
- `margin` on the `Thumb` part


vladmoroz:
Would `margin` work on the Scrollbar part?

atomiks:
> Would `margin` work on the Scrollbar part?

I tested it and it works as-is since it changes the `height`/`width` of the whole scrollbar part, so the calculation works correctly by default

vladmoroz (approved):

PR #865, opened 2024-11-25
[docs] Add a basic header

vladmoroz:
No logo yet, will add that when we have a few more pages (we'll be comparing a couple logo options / asking around)

https://deploy-preview-865--base-ui.netlify.app/new/components/dialog


colmtuite:
@vladmoroz 

imo links should just have default behaviour rather than opening in new tab. If you want to open in a new tab you can use `cmd` or right-click menu.

Bug: click the "Base UI" homepage link, press back, header is fucked.

vladmoroz:
> Bug: click the "Base UI" homepage link, press back, header is fucked.

This looks like a Next.js app router bug related to how CSS is loaded, not us. It will go away when we move things from `/new` to root, let's just ignore it for now.

> imo links should just have default behaviour rather than opening in new tab. If you want to open in a new tab you can use cmd or right-click menu.

OK, updated

colmtuite (approved):

PR #866, opened 2024-11-25
[docs] Mobile layout

vladmoroz:
https://deploy-preview-866--base-ui.netlify.app/new/components/dialog

### DIFF
--- a/docs/src/components/SideNav.tsx
+++ b/docs/src/components/SideNav.tsx
@@ -6,7 +6,7 @@ import { usePathname } from 'next/navigation';
 import { ScrollArea } from '@base-ui-components/react/ScrollArea';
 import scrollIntoView from 'scroll-into-view-if-needed';
 // eslint-disable-next-line no-restricted-imports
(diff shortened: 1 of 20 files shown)


colmtuite (approved):

docs/src/components/SideNav.tsx
```diff
@@ -6,7 +6,7 @@ import { usePathname } from 'next/navigation';
 import { ScrollArea } from '@base-ui-components/react/ScrollArea';
 import scrollIntoView from 'scroll-into-view-if-needed';
 // eslint-disable-next-line no-restricted-imports
-import { useEnhancedEffect } from '@base-ui-components/react/utils/useEnhancedEffect';
+import { SCROLL_TIMEOUT } from '@base-ui-components/react/ScrollArea/constants';
```
vladmoroz, line 9: @atomiks had a few questions about this

Why is the default `SCROLL_TIMEOUT` so generous? I couldn't figure out when `data-scrolling` gets removed until I looked into the source code. Wouldn't it be more intuitive to have the consumer do their own `transition-delay`?

PR #867, opened 2024-11-25
[core] Change the import paths to kebab case

mnajdova:
This PR adds support for kebab cases for import pathnames and updates all components. 

The infra changes are inside `scripts/buildApiDocs/config/getComponentImports.ts`. 

Note: This PR does not changes the file names, only directories, we can do this separately if we want to, as this will be an internal change.


michaldudak (approved):
Could you please rename useButton and useMediaQuery as well for consistency?
Feel free to merge afterwards.

michaldudak:
Just noticed that the docs refrence the old path. The InstallationInstructions component in .mdx files renders uppercased path.

mnajdova:
> Just noticed that the docs refrence the old path. The InstallationInstructions component in .mdx files renders uppercased path.

Nice catch, fixed, I just checked the JSON files.

PR #870, opened 2024-11-25
[core] Explicitly define exported subpaths

michaldudak:
- Explicitly defined identifiers exported from the package, as we don't intend to make every top-level package subdirectory public.
- Removed babel and webpack package aliases. There is now a single place that defines them: tsconfig
- Updated the build scripts to transform the package.json exports.


.eslintignore
```diff
@@ -2,6 +2,7 @@
 .nyc_output
 /coverage
 /docs/.next
+/docs/data/**/*.js
```
michaldudak, line 5: No need to lint autogenerated files

docs/src/app/experiments/slider-marks.tsx
```diff
@@ -1,263 +0,0 @@
-'use client';
```
michaldudak, line 1: Removed the two Slider experiments that use internal API

packages/react/package.json
```diff
@@ -26,27 +26,47 @@
...
+    "./select": "./src/select/index.ts",
+    "./separator": "./src/separator/index.ts",
+    "./slider": "./src/slider/index.ts",
+    "./switch": "./src/switch/index.ts",
+    "./tabs": "./src/tabs/index.ts",
+    "./text-input": "./src/text-input/index.ts",
+    "./tooltip": "./src/tooltip/index.ts",
+    "./unstable-no-ssr": "./src/unstable-no-ssr/index.ts",
+    "./unstable-use-media-query": "./src/unstable-use-media-query/index.ts",
+    "./utils": "./src/utils/index.ts"
```
michaldudak, line 56: Added the unstable- prefix to NoSsr and useMediaQuery as they are not documented yet (and their API might change).

vladmoroz (approved):

Issue #831, opened 2024-11-19
[core] Use a `DirectionProvider` to configure text direction (RTL/LTR)


mj12albert:
We have decided that RTL setup should be:
1. Use a `DirectionProvider` to configure RTL component behavior for an app (or part of an app), and
2. Set `dir="rtl"` accordingly in the html to affect styling, ideally in the `<html>` tag for a whole app

```jsx
<html dir="rtl">
  <body>
    <DirectionProvider direction="rtl">
        <Slider /> {/* automatically applies RTL keyboard behavior */}
    <DirectionProvider>
  </body>
</html>
```

We do not set the `dir` attribute on components by default, but users are free to do so if they wish.

For styling, the CSS [`:dir()` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:dir) could be used:
```css
.MySlider:dir(rtl) {
  /* RTL styles */
}
```

Looking at our [`.browserslistrc`](https://github.com/mui/base-ui/blob/master/.browserslistrc), only Chrome 109 and 119 don't support `:dir()` and would need an attribute selector instead:
```css
[dir='rtl'] .MySlider {
  /* RTL styles */
}
```

While auto-detecting the inherited `dir` worked for determining keyboard behavior, some components have lot of positioning styles involving CSS transforms, which do not have logical property counterparts yet (https://github.com/w3c/csswg-drafts/issues/7646) and require knowing the `dir` at render time

### Relevant components:
- Accordion
- Menu
- Progress
- RadioGroup
- Tabs
- Slider

oliviertassinari:
Does this work in the first place? There seem to be two cases:

1. Greenfield project. I set the direction on the html element, once at the root `<html dir="rtl">`. I don't want the underlying DOM node to have the dir attribute to avoid bloat in the DOM tree.
2. Brownfield project. I've added a few new components. I haven't made my whole app RTL-compliant yet. I want to add `dir="rtl"` a pages at a time.

Now, case 2. likely doesn't make a lot of sense. You wouldn't release RTL partially done.
But in case 1. doesn't this hold?

> I don't want the underlying DOM node to have the dir attribute to avoid bloat in the DOM tree.

If I set `<Slider dir="rtl">` I would expect this to be on the DOM node, but I would likely not want this attribute, only the RTL behavior. If it's not on the DOM node, how do I add it for case 2?

[...]

PR #871, opened 2024-11-26
[Composite] Detect text direction in event handler instead

mj12albert:
Part of https://github.com/mui/base-ui/issues/831

### DIFF
--- a/packages/react/src/Composite/Root/useCompositeRoot.ts
+++ b/packages/react/src/Composite/Root/useCompositeRoot.ts
@@ -85,16 +85,7 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
   const textDirectionRef = React.useRef<TextDirection | null>(null);
 
   const rootRef = React.useRef<HTMLElement | null>(null);
-  const handleRootRef = useEventCallback((element: HTMLElement) => {
-    if (!element) {
-      return;
-    }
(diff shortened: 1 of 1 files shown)


mj12albert:
@atomiks This is all of the lazy check we discussed yesterday right? (I'll do all the public components in a separate PR)

atomiks (approved):
Yeah this is the lazy check. There's no need to cache it as it isn't expensive enough to care when it's just 1 in an event handler, Floating UI has read from it multiple times on every update for years and does ok.

packages/react/src/Composite/Root/useCompositeRoot.ts
```diff
@@ -105,12 +96,19 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
...
 
-          const isRtl = textDirectionRef?.current === 'rtl';
+          const element = rootRef.current;
+          if (!element) {
+            return;
+          }
+
+          const isRtl =
+            textDirectionRef?.current == null
+              ? getTextDirection(element) === 'rtl'
```
oliviertassinari, line 110: This function triggers a reflow, proof: https://gist.github.com/paulirish/5d52fb081b3570c81e3a?permalink_comment_id=5299127#gistcomment-5299127

PR #874, opened 2024-11-26
[code-infra] Automatically get info for CSS variables

mnajdova:
This PR proposes a more structured way of defining CSS variables used in the components. The benefits are:
- avoid typos, we need to review only one place for checking if the variables are written correctly
- we can extract automatically the info for the CSS variables and use it in the API table (this was the main goal for proposing the change)

Depends on https://github.com/mui/material-ui/pull/44587

Preview: https://deploy-preview-874--base-ui.netlify.app/new/components/dialog#popup

I will update the rest of the CSS variables usages in a follow up PR.


packages/react/src/accordion/cssVars.ts
```diff
@@ -0,0 +1,6 @@
+export const accordionCssVars = {
+  accordionPanelHeight: '--accordion-panel-height',
+  accordionPanelWidth: '--accordion-panel-width',
```
mnajdova, line 3:
```suggestion
  /** Description for the CSS Varaible. */
  accordionPanelHeight: '--accordion-panel-height',
  /** Description for the CSS Varaible. */
  accordionPanelWidth: '--accordion-panel-width',
```

packages/react/src/accordion/cssVars.ts
```diff
@@ -0,0 +1,8 @@
+export const accordionCssVars = {
+  /** Description for the CSS Varaible. */
```
flaviendelangle, line 2:
```suggestion
  /** Description for the CSS Variable. */
```

packages/react/src/accordion/cssVars.ts
```diff
@@ -0,0 +1,8 @@
...
+  accordionPanelHeight: '--accordion-panel-height',
+  /** Description for the CSS Varaible. */
```
flaviendelangle, line 4:
```suggestion
  /** Description for the CSS Variable. */
```

flaviendelangle (commented):
:angel:

packages/react/src/dialog/popup/cssVars.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export type DialogPopupCssVars = {
+  /**
+   * Indicates how many dialogs are nested within.
+   * @type number
+   * */
+  nestedDialogs: string;
+};
+
+export const cssVars: DialogPopupCssVars = generateCssVariables(['nestedDialogs']);
```
mnajdova, line 11: I added this util similar to the utility for generating classes we have in other projects: https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Dialog/dialogClasses.ts#L43, to make sure we don't have typos while generating the CSS variables.

vladmoroz (approved):

packages/react/src/dialog/popup/dialogPopupCssVars.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export type DialogPopupCssVars = {
+  /**
+   * Indicates how many dialogs are nested within.
+   * @type number
+   * */
+  nestedDialogs: string;
+};
+
+export const dialogPopupCssVars: DialogPopupCssVars = generateCssVariables(['nestedDialogs']);
```
michaldudak, line 11: I'm thinking if there's a better way to define it and avoid specifying the same names in two places. A potentially slightly better version could be using enums:

```ts
enum PopupCssVars {
    maxWidth = '--max-width',
    foo = '--foo',
}
```

It still requires defining two things, but at least they are next to each other.
What do you think?

mnajdova: > I'm thinking if there's a better way to define it and avoid specifying the same names in two places. A potentially slightly better version could be using enums

Makes sense, I'll experiment to see how we can process enums and extract the info.

> It still requires defining two things, but at least they are next to each other. What do you think?

The downside is that there still may be typos (between the name and the value), but if we can extract the info from the enum directly, at least we won't have to "generate" the CSS variable value two times (once on the infra side and once in the generateCssVariables util), which is better. I will look into this tomorrow and report back.

mnajdova: I've updated to use enum, however I can't add the enum in the namespace, it throws an error:
```
Cannot redeclare block-scoped variable 'DialogPopup'.ts(2451)
```

packages/react/src/dialog/popup/dialogPopupCssVars.ts
```diff
@@ -0,0 +1,11 @@
+import { generateCssVariables } from '../../utils/generateCssVariables';
+
+export type DialogPopupCssVars = {
```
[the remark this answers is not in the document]
mnajdova: See https://github.com/mui/base-ui/pull/874#discussion_r1861916356.

packages/react/src/dialog/popup/dialogPopupCssVars.ts
michaldudak: Yeah, this is unfortunate. It's because our components are constants, not functions (thanks to `forwardRef`). It's fine to augment a function with additional values (using the `namespace` syntax, not so much a const/variable.

I found a workaround, but it requires some gymnastics. I'm not sure if it's worth it (but on the other hand, exporting cssVars differently to props feels wrong as well) 
https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQdA9I3LsgHZnwCuqWcMABb8AQhAAecdshD9hlOrgjtU8MeIDCJSOyzt4AXkxUYAOgLQA7sigATbAQA8ACQAqAWQAyAEWAA3AFEAGyxZfQAaOHVTAAViMFQAPgAKAm52fGBlOGSweNRIygIASjgAbzo4KrhKGG4odjhHW38arAIDMqKAX3LTfryIBN7GRLpu4vo6dS1wZT0zFtQwIOQATwA5GX4jMnVaBiUVeA1UVAA1G3QjCuq4eYAucgBaZ-mycMrqmEsIJ7JXj8IGRxgxmAJBMB0HxQugYBA4AAjfi6XBYM42NYCBEAaywWDAcFcAGU4IJkGAwGtFMpVFEJHAjAB5REAKyw+FMaFQwAA5uxkjNtPMIuU4KcLlc4BMptJZMs8KIGbdqlhxJBYHBgPosFACIq4HEhugVXcqktkIiQrYAPxPREQCAhDhfKrdBh3NUa+AwNZgfgSy5Qa4CP1YCAEcVnIOoejuuhe6DwMr0yTdKZMFi8ZC8rAPBhpDIwLKNACClOSpVNcFdcBC8D8Vye0UDVxocHBbEayLgvCwthQ6GQof9tfrcEbwcuQW4O1TplbwfbnY4SP4faHQ4nyBnWFrtdq9UayVrd0c6g7YzuxXGQA

Another solution would be not to export cssVars publicly (and only use them in our code to prevent typos and generate the docs).

cc @atomiks

mnajdova: > Another solution would be not to export cssVars publicly (and only use them in our code to prevent typos and generate the docs).

Not sure if this is possible as the API generation is looking for the type to exists in the TS project.

mnajdova: > I found a workaround, but it requires some gymnastics. I'm not sure if it's worth it (but on the other hand, exporting cssVars differently to props feels wrong as well)

Maybe we shouldn't have rushed with using namespaces for the types. We could hit similar issue in the future too even if we decide not to export the CSS vars type at this moment.

atomiks (approved):

PR #878, opened 2024-11-27
[popups] Look for animations when `open` is set to `false` externally

michaldudak:
When popups are closed programatically, the code that unmounts them after their exit animations finish wasn't run. 
In this PR I moved the exit animation detection to an effect, so it runs whenever the state changes. Added tests in all affected components (and fixed few other issues with the Dialog: `animated` prop not being propagated to `useDialogRoot` and `aria-hidden` being applied next to `inert`)

### DIFF
--- a/packages/react/src/dialog/root/DialogRoot.test.tsx
+++ b/packages/react/src/dialog/root/DialogRoot.test.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { expect } from 'chai';
 import { spy } from 'sinon';
-import { act, describeSkipIf, fireEvent } from '@mui/internal-test-utils';
+import { act, describeSkipIf, fireEvent, screen, waitFor } from '@mui/internal-test-utils';
 import { Dialog } from '@base-ui-components/react/dialog';
 import { createRenderer } from '#test-utils';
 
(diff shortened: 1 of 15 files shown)


packages/react/src/utils/useUnmountAfterCloseAnimation.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useEnhancedEffect } from './useEnhancedEffect';
+import { useLatestRef } from './useLatestRef';
+
+export function useUnmountAfterExitAnimation(parameters: useUnmountAfterExitAnimation.Parameters) {
+  const {
+    open,
+    animated,
+    animatedElementRef,
+    setMounted,
+    flushUpdatesSynchronously = false,
```
atomiks, line 12: This shouldn't actually be necessary. Think it was a leftover from the old select `alignOptionToTrigger` implementation.

packages/react/src/dialog/root/DialogRoot.test.tsx
```diff
@@ -44,43 +44,93 @@ describe('<Dialog.Root />', () => {
...
-          <Dialog.Root modal animated={false}>
-            <Dialog.Popup />
-          </Dialog.Root>,
-        ),
-      ).toWarnDev([
-        'Base UI: The Dialog is modal but no backdrop is present. Add the backdrop component to prevent interacting with the rest of the page.',
-        'Base UI: The Dialog is modal but no backdrop is present. Add the backdrop component to prevent interacting with the rest of the page.',
-      ]);
-    });
+    it('should remove the popup when animated=true and there is no exit animation defined', async function test(t = {}) {
```
michaldudak, line 48: We can create a reusable test suite for veryfing behavior of popup-like components in the future, so we don't need to copy the same tests several times.

atomiks (approved):

PR #529, opened 2024-08-04
[core] Make pnpm version permissive

atomiks:
- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

https://x.com/rickhanlonii/status/1790015586744795353

Avoids:

warning

PR #879, opened 2024-11-27
[core] Pin pnpm's version

mnajdova:
I noticed some changes when running pnpm i/pnpm dedupe whe using a newer version of pnpm. Pinning the pnpm's version in the engines definition will eliminate this problem.


michaldudak (approved):
This field's version was loosened in https://github.com/mui/base-ui/pull/529, but it appears that the lockfile can be different between minor versions, so we have to pin it.

atomiks (approved):
I also noticed this exact issue with dedupe, so pinning makes sense

PR #880, opened 2024-11-27
[Slider] Detect text direction instead of using `direction` prop

mj12albert:
Part of https://github.com/mui/base-ui/issues/831


packages/react/src/slider/thumb/useSliderThumb.ts
```diff
@@ -110,17 +109,16 @@ export function useSliderThumb(parameters: useSliderThumb.Parameters): useSlider
     return {
       position: 'absolute',
       [{
-        horizontal: 'left',
-        'horizontal-reverse': 'right',
+        horizontal: 'insetInlineStart',
```
mj12albert, line 112: The CSS logical property counterparts of positioning properties like `left`/`top` are supported by all our supported browsers
https://caniuse.com/mdn-css_properties_inset-inline-start

PR #882, opened 2024-11-27
[popups] Add data-closed style hook

atomiks:
Adds `hidden` to popup components' `State`, and when `true` (i.e. `hidden` is applied with `keepMounted`) it has a `data-closed` style hook.


packages/react/src/alert-dialog/popup/AlertDialogPopup.tsx
```diff
@@ -90,10 +90,11 @@ const AlertDialogPopup = React.forwardRef(function AlertDialogPopup(
   const state: AlertDialogPopup.State = React.useMemo(
     () => ({
       open,
+      mounted,
```
michaldudak, line 93: Not strictly related to these changes, but this now becomes public API - I feel like "mounted" doesn't describe the behavior very well, especially when `keepMounted = true`. "visible" could be a better option.

vladmoroz: Perhaps `hidden` to align with the DOM attribute that's being set?

vladmoroz: `visible` sounds good too though

atomiks: I think `hidden` makes the most sense to align with the attr. `visible` adds yet another term to deal with

packages/react/src/alert-dialog/popup/AlertDialogPopup.tsx
```diff
@@ -90,7 +90,7 @@ const AlertDialogPopup = React.forwardRef(function AlertDialogPopup(
   const state: AlertDialogPopup.State = React.useMemo(
     () => ({
       open,
-      mounted,
+      hidden: !mounted,
```
michaldudak, line 93: Can we also use the same term internally?

atomiks (commented):
The `hidden` value returned from `useAnchorPositioning()` refers to it being hidden _but still mounted_. This needs to use a `visibility` style, so it can still be measured. We don't expose a style hook for this state, but the functional style is inlined in the hook itself. I've refactored `pointerEvents` style (in the commit after this comment) to be inside the hook as well.

packages/react/src/utils/useAnchorPositioning.ts
```diff
@@ -272,14 +272,17 @@ export function useAnchorPositioning(
...
+  const anchorHidden = Boolean(hideWhenDetached && middlewareData.hide?.referenceHidden);
 
   const positionerStyles = React.useMemo(
     () => ({
       ...floatingStyles,
-      ...(hidden && { visibility: 'hidden' as const }),
+      ...(anchorHidden && {
+        visibility: 'hidden' as const,
+        pointerEvents: 'none' as const,
+      }),
```
atomiks, line 283: `pointerEvents` was being set inside each positioner, but since `visibility` was inlined as a functional style, it can be done here too. We don't expose a style hook for this state but could later, but I don't think it needs to be customized.

atomiks:
@vladmoroz thinking `data-hidden` (not transforming the hook name) makes more sense? Since `data-closed` seems like it should match `open: false` to match the other components' symmetric counterpart

vladmoroz:
> Since `data-closed` seems like it should match `open: false` to match the other components' symmetric counterpart

Wait I don't follow, isn't it symmetric to `open: false`? Maybe an example would help

`data-hidden` would have made sense if the other one was `data-visible`

atomiks:
> Wait I don't follow, isn't it symmetric to open: false? Maybe an example would help

This PR represents when it's fully unmounted. You were mentioning `[hidden]` isn't enough to target the relevant state since you can do `hidden={false}`. Maybe it should be `data-closed` (`open: false` state) AND `data-hidden` (`[hidden]` / `mounted: false` state)?

colmtuite (approved):

vladmoroz:
> > Wait I don't follow, isn't it symmetric to open: false? Maybe an example would help
> 
> This PR represents when it's fully unmounted. You were mentioning `[hidden]` isn't enough to target the relevant state since you can do `hidden={false}`. Maybe it should be `data-closed` (`open: false` state) AND `data-hidden` (`[hidden]` / `mounted: false` state)?

I'd expect that just `data-closed` that corresponds to `open: false`, and then if they are force mounting or adding a manual `hidden={false}`, they can add their own styling attributes as needed

atomiks:
Going to close this and open a new one that matches the `open: false` boolean.

PR #887, opened 2024-11-28
[core] Optionally delay `useTransitionStatus`'s `entering` status with a param

mj12albert:
Split from https://github.com/mui/base-ui/pull/807#issuecomment-2500516332 since `keepMounted={false}` needs some rework

Collapsible depends on having 1 frame between `open` and `entering` to measure the size of the panel contents

https://deploy-preview-887--base-ui.netlify.app/experiments/collapsible-transitions

### DIFF
--- a/packages/react/src/utils/useTransitionStatus.ts
+++ b/packages/react/src/utils/useTransitionStatus.ts
@@ -8,9 +8,12 @@ export type TransitionStatus = 'entering' | 'exiting' | undefined;
  * Provides a status string for CSS animations.
  * @param open - a boolean that determines if the element is open.
  * @param enabled - a boolean that determines if the logic is enabled.
+ * @param delayEnteringStatus - a boolean that set the `entering` status one
+ *     tick later. Example use-case: collapsible needs an extra frame in order
+ *     to measure the panel contents.
  * @ignore - internal hook.
  */
-export function useTransitionStatus(open: boolean, enabled = true) {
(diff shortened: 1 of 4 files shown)


packages/react/src/utils/useTransitionStatus.ts
```diff
@@ -8,9 +8,12 @@ export type TransitionStatus = 'entering' | 'exiting' | undefined;
  * Provides a status string for CSS animations.
  * @param open - a boolean that determines if the element is open.
  * @param enabled - a boolean that determines if the logic is enabled.
+ * @param delayStartingStatus - a boolean  that set the `entering` status one
+ *     tick later. For edge cases e.g. collapsible needs an extra frame in order
+ *     to measure the panel contents.
  * @ignore - internal hook.
  */
-export function useTransitionStatus(open: boolean, enabled = true) {
+export function useTransitionStatus(open: boolean, enabled = true, delayStartingStatus = false) {
```
mj12albert, line 16: @atomiks or `delayEnteringStatus`? If we're not renaming it internally https://github.com/mui/base-ui/pull/877

atomiks: Yeah I haven't renamed it internally, so we should match `entering`

mj12albert: Updated ~

atomiks (approved):

PR #888, opened 2024-11-28
[popover] Fix regression on disabling focus management on hover

michaldudak:
#878 brought back the bug causing page scrolling after a hover-triggered Popover is closed.


packages/react/src/popover/root/PopoverRoot.test.tsx
```diff
@@ -394,61 +394,94 @@ describe('<Popover.Root />', () => {
     });
   });
 
-  it('focuses the trigger after the popover is closed but not unmounted', async () => {
-    await render(
-      <div>
-        <input type="text" />
-        <Popover.Root animated={false}>
+  describe('focus manangment', () => {
```
atomiks, line 397:
```suggestion
  describe('focus manangment', () => {
```

atomiks (approved):

PR #897, opened 2024-11-29
[docs] Content 2: Collapsible, demo fixes

vladmoroz:
https://deploy-preview-897--base-ui.netlify.app/new/components/collapsible

- Add Collapsible demo (transitions are broken, supposedly fixed by https://github.com/mui/base-ui/pull/889)
- Fix dialog demo
- Make sure all demos use colors included in a default Tailwind config
- Unset our custom docs CSS variables in the demos
- Fix a Safari rendering issue with icons

Collapsible should look like this when the transitions are fixed:

https://github.com/user-attachments/assets/535920c7-bba7-4714-b67e-64154ee10bcf


docs/src/app/new/(content)/components/collapsible/demos/hero/tailwind/index.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { Collapsible } from '@base-ui-components/react/collapsible';
+
+export default function ExampleCollapsible() {
+  return (
+    <Collapsible.Root
+      className="flex min-h-36 w-56 flex-col justify-center"
+      render={<div />}
+    >
```
vladmoroz, line 9: @mj12albert Collapsible Root accepts DOM props but doesn't render anything by default. Feels like an oversight? I expected it'd render a div like Accordion

mj12albert: It was because the ARIA pattern is just a trigger and the panel, though I guess for practical purposes you'd usually want an element, and maybe for consistency too
We can have it render a div but also allow `render={null}` instead

colmtuite:
@vladmoroz No focus styles?

colmtuite (approved):

docs/src/app/new/(content)/components/collapsible/demos/hero/tailwind/index.tsx
vladmoroz: Let's just render a DOM element by default. It feels unexpected in comparison to other components that have Root and don't have portalled children

> allow render={null} instead

We don't allow this for other components, no? Don't see a need for it to be fair

PR #901, opened 2024-11-29
[docs] Logo B

vladmoroz:
https://deploy-preview-901--base-ui.netlify.app/new/components/accordion


oliviertassinari:
**Off-topic**
On the accordion demo, maybe it would make sense to position the demo to not be vertically centered but have a fixed top position, so when we expand the demo, it's more fluid? Not 100% sure

Before:

https://github.com/user-attachments/assets/e72953c5-c733-4a40-9123-c348eee21dfe

After:

https://github.com/user-attachments/assets/4d3ac11a-1f99-48b0-915d-b456643c8952

colmtuite (approved):

PR #903, opened 2024-11-29
[docs] Content 3: Text Input, Field, Fieldset, Form demos

vladmoroz:
https://deploy-preview-903--base-ui.netlify.app/new/components/field


docs/src/app/new/(content)/components/field/demos/hero/tailwind/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <Field.Root className="flex flex-col gap-2">
+      <div className="flex flex-col items-start gap-1">
+        <Field.Label className="text-sm font-medium text-gray-900">Name</Field.Label>
+        <Field.Control
+          required
+          // @ts-ignore
+          placeholder="Required"
+          className="w-64 rounded-md border border-gray-200 px-3.5 py-2 text-gray-900 focus:outline-2 focus:-outline-offset-1 focus:outline-blue-800"
+        />
```
vladmoroz, line 14: Something is off with `Field.Control` type definitions, it errors for `placeholder`

colmtuite:
@vladmoroz 
- The input is 42px tall but other buttons/controls are 40px. Heights should be consistent across components. I would use `height` rather than `padding` to achieve this.
- The input hint seems too far from the input. It also seems odd to me that the label and hint are not equidistant from the input.

docs/src/app/new/(content)/components/field/demos/hero/tailwind/index.tsx
atomiks: This is because it also tries to support types of `select` and `textarea` along with `input` as these can be rendered as a control as well. Maybe it should just use `input` types?

PR #928, opened 2024-12-02
[CheckboxGroup] Fix controlled usage

atomiks:


mnajdova (approved):

packages/react/src/checkbox/root/useCheckboxRoot.ts
```diff
@@ -26,9 +27,14 @@ export function useCheckboxRoot(params: UseCheckboxRoot.Parameters): UseCheckbox
...
+  const groupContext = useCheckboxGroupRootContext();
+  const groupValue = groupContext?.value;
+  const setGroupValue = groupContext?.setValue;
+  const defaultGroupValue = groupContext?.defaultValue;
+
   const [checked, setCheckedState] = useControlled({
-    controlled: externalChecked,
-    default: defaultChecked,
+    controlled: name && groupValue ? groupValue.includes(name) : externalChecked,
+    default: name && defaultGroupValue ? defaultGroupValue.includes(name) : defaultChecked,
```
mnajdova, line 37: On second thought, shouldn't the props added on the element itself have precedence?

atomiks: I'm not fully sure. In the codebase so far, the parent takes precedence, e.g. with `disabled` states. I wonder if it's confusing or potentially a footgun to allow individual overriding if you're controlling from a parent - I think that's why the parent overrides any children currently.

Issue #59, opened 2023-12-25
[Tabs] Missing reference for RTL keyboard navigation direction


ojbaharav:
### Search keywords

RTL keyboard navigation

### Related page

https://mui.com/base-ui/react-tabs/

### Kind of issue

Missing information

### Issue description

I found that in the Tabs API props section, there is a 'direction' prop but it says that it affects "the direction of the text" which is true but incomplete, cause it also affects keyboard navigation direction.

Also, it would be helpful if the 'direction' prop were mentioned in the main page's Accessibility > keyboard navigation section as a relevant prop to the subject like is currently doing with the 'selectionFollowsFocus' prop.

### Context 🔦

I was trying to figure out how to change the keyboard navigation direction on an RTL page so that I could use the right and left arrow keys according to the reading direction.

mj12albert:
> it would be helpful if the 'direction' prop were mentioned in the main page's Accessibility > keyboard navigation section

I agree it could be expanded in the docs more, but I don't think it's really an a11y topic, maybe `Customization > RTL`, what do you think? @samuelsycamore 

> I found that in the Tabs API props section, there is a 'direction' prop but it says that it affects "the direction of the text" which is true but incomplete

You are right, we should add that it also controls the navigation direction, would you be interested in creating a PR? @ojbaharav

mapache-salvaje:
@mj12albert agreed, it would be nice to have a section with a demo under Customization that shows how this prop works.

ojbaharav:
> > it would be helpful if the 'direction' prop were mentioned in the main page's Accessibility > keyboard navigation section
> 
> I agree it could be expanded in the docs more, but I don't think it's really an a11y topic, maybe `Customization > RTL`, what do you think? @samuelsycamore

It would concern only people who want to make it accessible for keyboard use, not any RTL customization in general. In order to make this component display as RTL, I only need to set the direction of the page or the Tabs container with CSS or dir=rtl. For pointer (mouse) users, you don't really need the 'direction' prop in RTL websites.

> > I found that in the Tabs API props section, there is a 'direction' prop but it says that it affects "the direction of the text" which is true but incomplete
> 
> You are right, we should add that it also controls the navigation direction, would you be interested in creating a PR? @ojbaharav

Never done that before, but I'll try :)

ojbaharav:
Well, it took me some time to get to it and now I want to modify the direction's description. Does the direction have any impact on the text or is it only useful for keyboard navigation purposes?

[...]

PR #931, opened 2024-12-02
[core] Add DirectionProvider

mj12albert:
Closes https://github.com/mui/base-ui/issues/831
Closes https://github.com/mui/base-ui/issues/59

- Docs: https://deploy-preview-931--base-ui.netlify.app/components/react-direction-provider
- Updates [Menu](https://deploy-preview-931--base-ui.netlify.app/components/react-menu#rtl), [RadioGroup](https://deploy-preview-931--base-ui.netlify.app/components/react-radio-group#rtl), [Slider](https://deploy-preview-931--base-ui.netlify.app/components/react-slider#rtl), [Tabs](https://deploy-preview-931--base-ui.netlify.app/components/react-tabs#rtl), [Accordion](https://deploy-preview-931--base-ui.netlify.app/components/react-accordion#rtl), [ToggleGroup](https://deploy-preview-931--base-ui.netlify.app/components/react-toggle-group#rtl)

### DIFF
--- a/packages/react/src/radio-group/root/RadioGroupRoot.tsx
+++ b/packages/react/src/radio-group/root/RadioGroupRoot.tsx
@@ -5,7 +5,7 @@ import type { BaseUIComponentProps } from '../../utils/types';
 import { CompositeRoot } from '../../composite/root/CompositeRoot';
 import { useComponentRenderer } from '../../utils/useComponentRenderer';
 import { useEventCallback } from '../../utils/useEventCallback';
-import { useDirectionContext } from '../../direction-provider/DirectionContext';
+import { useDirection } from '../../direction-provider/DirectionContext';
 import { useRadioGroupRoot } from './useRadioGroupRoot';
 import { RadioGroupRootContext } from './RadioGroupRootContext';
 import { useFieldRootContext } from '../../field/root/FieldRootContext';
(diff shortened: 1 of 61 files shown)


packages/react/src/radio-group/root/RadioGroupRoot.tsx
```diff
@@ -33,6 +34,9 @@ const RadioGroupRoot = React.forwardRef(function RadioGroupRoot(
     ...otherProps
   } = props;
 
+  const directionContext = useDirectionContext();
+  const direction = directionContext?.direction ?? 'ltr';
```
atomiks, line 38: Since it only has one value, we could simplify this by defaulting `ltr` inside `useDirectionContext()` and returning the value directly, `const direction = useDirectionContext()`

atomiks (commented):
`ScrollArea` also checks for RTL. Should we default to checking for computed styles if it's cheap, like inside an event, and falling back to the provider otherwise? 

Floating elements force read the style and would ignore the provider, unless we add it here: https://floating-ui.com/docs/platform#isrtl

mj12albert:
> Should we default to checking for computed styles if it's cheap, like inside an event, and falling back to the provider otherwise?

I thought about this, but we have to document DirectionProvider as a required step when changing the behavior is involved anyway to maintain that "there is one single way to set up RTL", even though technically it's not necessary (e.g. for ToggleGroup/RadioGroup)

And if a component doesn't actually use the provider internally it would obscure issues when someone used the provider but forgot to set `dir`

mj12albert:
> Floating elements force read the style and would ignore the provider,

@atomiks I think this should be fine for our components since only `Menu` is the only one that requires the provider (it has keyboard nav).

For the rest (popover, preview card etc), they should first use logical `side`s instead of `left`/`right`

I suppose since `align: 'start'|'end'` is already direction-aware without the provider, maybe `side` should be as well? https://github.com/mui/base-ui/pull/950/commits/0a51ad14684945498ff846e5370e832f8cedeb7f#diff-3edb30ade872773deda81b8799e55170e5a0a58fab8e2bee97a2210169e99066R95-R109

atomiks (approved):

PR #933, opened 2024-12-02
[CheckboxGroup] Remove `preserveChildStates` prop and default to `true`

atomiks:


packages/react/src/checkbox-group/root/CheckboxGroupRoot.test.tsx
```diff
@@ -13,122 +11,4 @@ describe('<CheckboxGroup.Root />', () => {
...
-
-      fireEvent.click(parent);
-
-      expect(red).to.have.attribute('aria-checked', 'true');
-      expect(green).to.have.attribute('aria-checked', 'true');
-      expect(blue).to.have.attribute('aria-checked', 'true');
-      expect(parent).to.have.attribute('aria-checked', 'true');
-    });
-
-    it('should preserve child states when true', () => {
```
mnajdova, line 76: Shouldn't we keep this test as a default behavior?

vladmoroz (approved):

packages/react/src/checkbox-group/root/CheckboxGroupRoot.test.tsx
atomiks: It's already tested in `useCheckboxGroupParent.test.tsx`

PR #934, opened 2024-12-02
[popups] Use `data-anchor-hidden` attribute

atomiks:

`hideWhenDetached` -> always-present `data-anchor-hidden` attribute

### DIFF
--- a/docs/data/api/menu-positioner.json
+++ b/docs/data/api/menu-positioner.json
@@ -31,7 +31,6 @@
       "default": "5"
     },
     "container": { "type": { "name": "union", "description": "HTML element<br>&#124;&nbsp;func" } },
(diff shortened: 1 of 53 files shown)


vladmoroz (approved):

vladmoroz:
Not a strong opinion, but should we consider `data-trigger-hidden`? Kind of how `data-popup-open` on triggers references the Popup part.

vladmoroz:
Although we do `--anchor-width`, `--anchor-height`. Maybe that's another term that can be replaced with "trigger", but we probably can discuss it later

atomiks:
> Although we do --anchor-width, --anchor-height. Maybe that's another term that can be replaced with "trigger", but we probably can discuss it later

The anchor can be a custom element not just the trigger (that's the default though)

Issue #896, opened 2024-11-29
[Checkbox] Unmount indicator


colmtuite:
Unmount by default. Add same api as popup for animation status.

**Search keywords**:

PR #936, opened 2024-12-02
[Checkbox] Default `Indicator` to `keepMounted=false`

atomiks:

Closes https://github.com/mui/base-ui/issues/896


packages/react/src/checkbox/indicator/CheckboxIndicator.tsx
```diff
@@ -22,19 +27,57 @@ const CheckboxIndicator = React.forwardRef(function CheckboxIndicator(
...
-    ref: forwardedRef,
+    ref: mergedRef,
     state,
     className,
     customStyleHookMapping,
-    extraProps: otherProps,
+    extraProps: {
+      hidden,
+      ...otherProps,
+    },
```
atomiks, line 80: One consequence of `hidden` here is that you can't rely on the outer `Root` checkbox animation, because it looks for animations on the indicator itself: `animatedElementRef: indicatorRef`. So you need to duplicate styles to keep it mounted; `hidden={false}` of course can be used

packages/react/src/utils/styleHookMapping.ts
```diff
@@ -0,0 +1,14 @@
+import type { TransitionStatus } from './useTransitionStatus';
+import type { CustomStyleHookMapping } from './getStyleHookProps';
+
+export const transitionStatusMapping = {
+  transitionStatus(value): Record<string, string> | null {
+    if (value === 'starting') {
+      return { 'data-starting-style': '' };
```
michaldudak, line 7: Nit: you could create the const object beforehand and reuse it in each call.

michaldudak (approved):

PR #937, opened 2024-12-02
[popups] Create `Portal` part

atomiks:


docs/data/api/accordion-panel.json
```diff
@@ -9,13 +9,6 @@
   "imports": [
     "import { Accordion } from '@base-ui-components/react/accordion';\nconst AccordionPanel = Accordion.Panel;"
   ],
-  "cssVariables": {
-    "--accordion-panel-height": {
-      "description": "The accordion panel's height.",
-      "type": "number"
-    },
-    "--accordion-panel-width": { "description": "The accordion panel's width.", "type": "number" }
-  },
```
atomiks, line 18: @mnajdova is `docs:api` supposed to create/preserve these?

michaldudak:
It seems to me we could gain quite a lot by inverting the dependencies. Instead of a separate `*Portal` component, we could have one (internal) `Portal` (re-exported as a part of each component) with its own `PortalContext`. Then, each Root part would set up this context.
The benefits I can think of are a smaller package, one place to introduce changes and fewer unnecessary rerenders (Portal would rerender only if its context changes, not the large *RootContext).
Backdrops could also benefit from this, and ideally Positioners, but this could be more tricky as they differ between components.

packages/react/src/alert-dialog/portal/AlertDialogPortal.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * API:
+ *
+ * - [AlertDialogPortal API](https://base-ui.com/components/react-alert-dialog/#api-reference-AlertDialogPortal)
+ */
+const AlertDialogPortal: AlertDialogPortal = function AlertDialogPortal(
+  props: AlertDialogPortal.Props,
+) {
+  const { children, container, keepMounted = false } = props;
+
+  const { mounted } = useAlertDialogRootContext();
```
michaldudak, line 23: We agreed to use the term `visible` instead of `mounted`.

atomiks: `mounted` is used everywhere currently and doesn't relate to this PR. We can change it but it's not necessary/not exposed to publicly right now

atomiks:
> It seems to me we could gain quite a lot by inverting the dependencies. Instead of a separate *Portal component, we could have one (internal) Portal (re-exported as a part of each component) with its own PortalContext. Then, each Root part would set up this context.

Not sure I fully understand - these separate Portal components are really small, I think the context would add more size if anything. Portal re-rendering is a non-issue since it's cheap

michaldudak:
They are small, but they are copied 7 times. Defining it just once (plus the context, which will be even smaller, I suppose) would make the package lighter.

I realize that rendering a Portal isn't a perf bottleneck, but if we can easily save a few CPU cycles, why not?

packages/react/src/dialog/index.parts.ts
```diff
@@ -2,6 +2,7 @@ export { DialogBackdrop as Backdrop } from './backdrop/DialogBackdrop';
 export { DialogClose as Close } from './close/DialogClose';
 export { DialogDescription as Description } from './description/DialogDescription';
 export { DialogPopup as Popup } from './popup/DialogPopup';
+export { Portal } from '../portal/Portal';
```
michaldudak, line 5: I'm not sure if the new docs handle it differently, but I think you might need to create a DialogPortal file that exports the Portal as DialogPortal to make the API gen recognize it. I'll rework it so it recognizes the actual parts, but in the meantime, this might be the necessary workaround.

packages/react/src/portal/Portal.tsx
```diff
@@ -0,0 +1,66 @@
...
+     * Whether to keep the portal mounted in the DOM when the popup is closed.
+     * @default false
+     */
+    keepMounted?: boolean;
+  }
+
+  export interface State {}
+}
+
+interface Portal {
```
michaldudak, line 41: Why is this needed?

atomiks: Prevent Root components from accepting invalid values

packages/react/src/dialog/index.parts.ts
atomiks: This doesn't seem to work with "No suitable component definition found." 

`MenuSeparator` also isn't exported separately

atomiks: Looks like we need to wrap, destructure etc. in order for the current gen to work so we don't gain much from the separate Portal file anyway as of right now.

michaldudak: It's still slightly better IMO now as you don't have to literally copy the component source 7 times, and will get progressively better as we fix the docgen.

packages/react/src/portal/Portal.tsx
```diff
@@ -0,0 +1,63 @@
...
+import PropTypes from 'prop-types';
+import { FloatingPortal } from '@floating-ui/react';
+import { usePortalContext } from './PortalContext';
+import { HTMLElementType, refType } from '../utils/proptypes';
+
+/**
+ * Renders a portal element into the DOM.
+ *
+ * Documentation: [Base UI Portal](https://base-ui.com/react/components/portal)
+ */
```
atomiks, line 12: This doesn't really exist (and probably shouldn't). Can't do `@ignore - internal component.` either

scripts/buildApiDocs/config/getBaseUiComponentInfo.ts
```diff
@@ -27,7 +27,12 @@ export function getBaseUiComponentInfo(filename: string): ComponentInfo {
   const componentUrl = overrides[name] ?? `/components/${parentComponentPath}`;
   const url = urlJoin(baseUrl, componentUrl);
 
-  const customAnnotation = `Documentation: [Base UI ${parentComponentName}](${url})`;
+  // Foreign components that aren't associated with any given component.
+  const EXCLUDED_NAMES = ['Portal'];
+
+  const customAnnotation = EXCLUDED_NAMES.includes(name)
+    ? `Documentation: https://base-ui.com`
+    : `Documentation: [Base UI ${parentComponentName}](${url})`;
```
atomiks, line 35: Exclude foreign components from having links. The alternative is to wrap each of them, but don't think a link is that important here 🤷

atomiks, line 35: Foreign standalone components (which aren't used standalone in terms of the public API and have no dedicated page) can't be associated to their component without wrapping. We can either always wrap (which _is_ already required for `Select`) or just link to the homepage?

packages/react/src/portal/PortalContext.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+
+export const PortalContext = React.createContext<boolean | undefined>(undefined);
```
michaldudak, line 3: Add the displayName

michaldudak (approved):

PR #946, opened 2024-12-03
[core] Remove animated prop

### LINKED ISSUE #655: Remove `animated` prop in favour of a provider component
`animated` prop is primarily needed to make tests run faster and easier to write

We need to remove it from the components and expose a provider to set the value globally

**Search keywords**:

atomiks:

Closes #655

Removes `animated` prop from all components. Exit animations are now ignored in JSDOM, and for browser tests, you can use the `globalThis.BASE_UI_ANIMATIONS_DISABLED = true` flag, which can be conditionally turned on/off when you want to test animations explicitly. This makes tests a bit simpler to write, though users still need to set `requestAnimationFrame` to be sync and wait for microtasks when rendering some elements.


test/setupVitest.ts
```diff
@@ -55,6 +55,11 @@ if (isVitestJsdom) {
...
-
-  globalThis.Element.prototype.getAnimations = () => [];
+} else if (typeof HTMLElement !== 'undefined') {
+  globalThis.originalGetAnimations = HTMLElement.prototype.getAnimations;
+  Object.defineProperty(HTMLElement.prototype, 'getAnimations', {
+    value: undefined,
+    configurable: true,
+    writable: true,
+  });
 }
```
atomiks, line 65: For in-browser tests, you'll need to explicitly set it to `undefined`

michaldudak: That's not great DX, especially if someone would want to control it per test.

atomiks: Yeah I had to 'undo' it but only for each of the animation tests you made. This is a lot simpler and a better trade-off than requiring a provider for every tree or a prop for every test

atomiks: That can work, as long as we don't set it ourselves in the package source (we could just read for its existence) it shouldn't cause any problems. A combo of checking for `getAnimations` existence (a good idea anyway) to handle the default jsdom scenario, and that variable for browser tests, seems ok.

packages/react/src/dialog/root/DialogRoot.test.tsx
```diff
@@ -175,4 +175,41 @@ describe('<Dialog.Root />', () => {
       });
     });
   });
+
+  it('when `true`, waits for the exit transition to finish before unmounting', async function test(t = {}) {
```
michaldudak, line 179: The "when `true`" part doesn't make much sense now.

michaldudak (approved):
Looks good now, assuming the test is renamed.

PR #949, opened 2024-12-04
[core] Fix proptypes script

atomiks:

After changing to kebab-case paths, the script could no longer find the component source files


packages/react/src/toggle/root/ToggleRoot.tsx
```diff
@@ -160,4 +154,9 @@ ToggleRoot.propTypes /* remove-proptypes */ = {
    * @ignore
    */
   type: PropTypes.oneOf(['button', 'reset', 'submit']),
+  /**
+   * A unique string that identifies the component when used
+   * inside a ToggleGroup
```
oliviertassinari, line 159:
```suggestion
   * inside a ToggleGroup.
```

PR #950, opened 2024-12-04
[Positioner] Support logical values for `side` prop

mj12albert:
Closes https://github.com/mui/base-ui/issues/951

**Demo:** https://deploy-preview-950--base-ui.netlify.app/experiments/rtl


packages/react/src/preview-card/positioner/usePreviewCardPositioner.ts
```diff
@@ -97,7 +102,7 @@ export namespace usePreviewCardPositioner {
      * The side of the anchor element that the preview card element should align to.
      * @default 'bottom'
      */
-    side?: Side;
+    side?: Side | InlineSide;
```
vladmoroz, line 105: Nit: can we use a single type instead of a union? Should be slightly easier to navigate the types for the developer

docs/src/app/experiments/rtl.tsx
```diff
@@ -0,0 +1,202 @@
...
+        <PreviewCard.Root>
+          <PreviewCard.Trigger href="#" className={c.trigger}>
+            PreviewCard.Trigger
+          </PreviewCard.Trigger>
+          <PreviewCard.Positioner
+            sideOffset={8}
+            side="inline-end"
+            align="center"
+            className={c.positioner}
+            dir={dir}
```
mj12albert, line 161: The only quirk is if you don't put `dir` on `html`, you need to put it on all the positioners manually because of portaling

packages/react/src/preview-card/positioner/usePreviewCardPositioner.ts
mj12albert: Updated!

docs/src/app/experiments/rtl.module.css
```diff
@@ -0,0 +1,239 @@
...
+    opacity: 0;
+
+    &[data-side='inline-end'] {
+      transform: scale(0.9) translateX(16px);
+    }
+
+    &[data-side='inline-start'] {
+      transform: scale(0.9) translateX(-16px);
+    }
+  }
```
mj12albert, line 97: @atomiks Having either physical only or logical sides in the rendered data attr works for these, but it does make more sense for it to be consistent with the prop that was provided 
CC @vladmoroz

vladmoroz: Yeah, it has to be consistent with the provided prop

mj12albert:
@vladmoroz @atomiks This is ready for review now ~

atomiks (approved):
My main concern with returning logical sides was animations, when you want the animation to play the opposite way based on the physical side, since transforms use physical values. But it looks like `:dir(rtl)` can handle that.

PR #953, opened 2024-12-04
[docs] Work around iOS Safari flicker

vladmoroz:
Fixes the flicker on some scroll containers in iOS Safari when the mobile nav is opened

Before

https://github.com/user-attachments/assets/1f252424-0f1f-4ae7-b65e-f45df0b36383

After

https://github.com/user-attachments/assets/17cbc01e-0a18-4c23-98d9-f3badbc506f6

### DIFF
--- a/docs/src/components/MobileNav.tsx
+++ b/docs/src/components/MobileNav.tsx
@@ -35,6 +35,7 @@ export function Popup({ children, className, ...props }: Dialog.Popup.Props) {
 
   return (
     <Dialog.Popup className={clsx('MobileNavPopup', className)} {...props}>
+      <MountEffect />
       <div className="MobileNavBottomOverscroll" />
       <div
         className="MobileNavViewport"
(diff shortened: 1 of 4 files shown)


docs/src/components/MobileNav.tsx
```diff
@@ -119,6 +119,28 @@ export function Popup({ children, className, ...props }: Dialog.Popup.Props) {
...
+    window.clearTimeout(timeout.current);
+    document.body.setAttribute('data-mobile-nav-open', '');
+
+    return () => {
+      window.clearTimeout(timeout.current);
+      timeout.current = window.setTimeout(() => {
+        document.body.removeAttribute('data-mobile-nav-open');
+        // Safari seems to need some arbitrary time to be done with whatever causes the flicker
+        // Using setTimeout 0, RAFs, or even double RAFs doesn't work reliably
+      }, 100);
```
vladmoroz, line 137: Setting an arbitrary timeout makes me incredibly angry, but I can't find anything deterministic that works instead.

The effect runs right before the Popup is unmounted, so the animations are done by that time.

Any ideas?

docs/src/styles.css
```diff
@@ -215,12 +215,6 @@
     -moz-osx-font-smoothing: grayscale;
   }
 
-  svg {
-    /* Safari messes up geometry of many basic SVGs, such as check icons in checkboxes or chevrons on menus */
-    /* Putting them on the GPU fixes that */
-    transform: translateZ(0);
-  }
-
```
vladmoroz, line 223: Ahem

Another Safari bug (macOS and iOS) that compounded here was that sometimes it renders the SVG geometry wrong (!). I had been debugging it on and off for a couple days before I settled on putting SVG's on the GPU, but now it's mostly gone?!

Anyway we can't have this either way, since the icons with this style also flicker when the nav is opened.

Love Safari 🫠

docs/src/components/MobileNav.tsx
mj12albert: Crazy suggestion, have you tried `setTimeout 0.5`?

vladmoroz: @mj12albert small delays like `0.5` work about half the time for me 🤷‍♂️

colmtuite (approved):

michaldudak (approved):

PR #964, opened 2024-12-05
[docs] Fix API info generation for Menu.* and Tabs.* components

mnajdova:
Preview: https://deploy-preview-964--base-ui.netlify.app/new/components/menu#submenu-trigger

Renamed the following components:
- SubmenuTrigger -> MenuSubmenuTrigger
- Tab -> TabsTab
- TabIndicator -> TabsTabIndicator
- TabPanel -> TabsTabPanel


michaldudak:
Do you know what is the root cause of the issue? Why did the docgen ignore the component with its previous name?

mnajdova:
The logic comes from here: https://github.com/mui/base-ui/blob/master/docs/src/components/reference/rehypeReference.mjs#L52 it expects that each part starts with the component name. I noticed that only this component didn't follow this pattern, so I renamed it.

michaldudak:
It's also the case with Tabs (we don't have TabsTab, just Tab)

mnajdova:
NIce catch, I guess Vlad didn't get to the Tabs doc yet :) I will fix it as part of this PR too.

vladmoroz (approved):

PR #972, opened 2024-12-06
[select] Retain highlight while transitioning out

### LINKED ISSUE #965: [Select] Retain the highlight state when closing
This is mostly a nit/polish item.
When closing the select, there's a frame of slight flicker/unnecessary motion:

https://github.com/user-attachments/assets/0465c913-1dca-4a5f-bde0-f7db8eefbfce

Frame by frame:

1. Right after clicking an option: there's a highlight on the clicked option (possibly retained from the hover). The previous value still has the indicator.

2. Next frame: the check icon is moved to its new place, but the highlight is lost

3. The popup is closed

Ideally:
- If animated, I think we need to make sure that from frame (2), the highlight is set and persisted until mouse movement and/or popup is fully closed. When animating the popup, it should be animating with the newly highlighted item visible.
- If not animated, the Select should close immediately without frame (2)

**Search keywords**:

atomiks:

Closes #965


mnajdova (approved):
Works as expected 👌

PR #973, opened 2024-12-06
[popover] Better handle impatient clicks

### LINKED ISSUE #955: [popups] When opening on hover, wait for animations before impatient clicks
For popups than open on hover, we've added a slight timeout that prevents impatient clicks from closing the popup immediately. It still feels like not enough if the popup is animating. Should the delay be contingent on the completion of animations/transitions? Possibly work like `Math.max(ourThreshold, animationDuration)`

**Search keywords**:

atomiks:

Closes #955

Currently if you click right as the popover opens it unexpectedly closes. The alternative, leaving it stay "stuck open" if it was already open for a while also feels off. This blends the two modes so that if they 'impatiently' clicked within 500ms after it opened on hover, it simply ignores the click entirely. This doesn't make it feel sticky, and it doesn't unexpectedly close, feeling the most natural. If it was open for >500ms, it does close though, making it feel responsive to clicks.

### DIFF
--- a/packages/react/src/popover/root/usePopoverRoot.ts
+++ b/packages/react/src/popover/root/usePopoverRoot.ts
@@ -13,7 +13,7 @@ import {
 import { useControlled } from '../../utils/useControlled';
 import { useEventCallback } from '../../utils/useEventCallback';
 import { useTransitionStatus } from '../../utils/useTransitionStatus';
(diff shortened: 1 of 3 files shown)


vladmoroz (approved):
Feels great
I'd maybe bump it to 500ms

PR #977, opened 2024-12-06
[dialog] Modality changes

michaldudak:
- added the `reason` parameter to Dialog's `onOpenChange`
- trapping keyboard focus regardless of the `modal` prop value

part of https://github.com/mui/base-ui/issues/623


packages/react/src/dialog/close/useDialogClose.ts
```diff
@@ -1,14 +1,18 @@
...
-  }, [open, onOpenChange]);
+  const { open, setOpen } = params;
+  const handleClick = React.useCallback(
+    (event: React.MouseEvent) => {
+      if (open) {
+        setOpen(false, event.nativeEvent, 'click');
+      }
+    },
+    [open, setOpen],
+  );
```
atomiks, line 15: `useEventCallback`

packages/react/src/dialog/popup/useDialogPopup.tsx
```diff
@@ -27,9 +36,16 @@ export function useDialogPopup(parameters: useDialogPopup.Parameters): useDialog
     titleElementId,
   } = parameters;
 
+  const handleFloatingUIOpenChange = React.useCallback(
+    (isOpen: boolean, event: Event | undefined, reason: FloatingUIOpenChangeReason | undefined) => {
+      setOpen(isOpen, event, translateOpenChangeReason(reason));
+    },
+    [setOpen],
+  );
```
atomiks, line 44: `useEventCallback`

michaldudak: This depends on `setOpen` only, which is stable.

packages/react/src/dialog/close/useDialogClose.ts
michaldudak: Same here - `setOpen` doesn't change between renders.

packages/react/src/dialog/popup/useDialogPopup.tsx
atomiks: Yeah, but you need to specify the deps needlessly, plus for consistency it's good to wrap event handlers in it since it's used elsewhere

packages/react/src/dialog/root/useDialogRoot.ts
```diff
@@ -46,21 +51,34 @@ export function useDialogRoot(parameters: useDialogRoot.Parameters): useDialogRo
...
+  const handleFloatingUIOpenChange = React.useCallback(
+    (
+      nextOpen: boolean,
+      event: Event | undefined,
+      reason: FloatingUIOpenChangeReason | undefined,
+    ) => {
+      setOpen(nextOpen, event, translateOpenChangeReason(reason));
+    },
+    [setOpen],
+  );
```
atomiks, line 76: Another here, though this one (and the one above) actually doesn't need to be wrapped, since it's done internally already

atomiks (approved):

PR #978, opened 2024-12-06
[docs] about page

colmtuite:
Sorry for all the messing folks


vladmoroz (approved):
Don't feel strongly about the restrictions on "our" / "we" usage, but I agree with the linter about the non-breaking spaces

oliviertassinari:
> Don't feel strongly about the restrictions on "our" / "we" usage

@vladmoroz Yeah, I agree, in the context of an about page, this rule is noise, we have to use "we".

I think it's why this rule is only a warning, it can't be enforced as an error.
The reason this rule exists though is because it's common to see engineer use "we" to describe how the code behaves, as "we" = "me and my code" 😆.

> I agree with the linter about the non-breaking spaces

Opened one https://github.com/mui/base-ui/pull/1228 to illustrate the value.

PR #988, opened 2024-12-09
[useAfterExitAnimation] `flushSync` the `onFinished` call

### LINKED ISSUE #979: [tooltip] Open tooltips without skipping a frame, if possible
Nitpick/polish item:

When moving focus or hovering another tooltip in a group, sometimes a frame is skipped

https://github.com/user-attachments/assets/bd15aeb5-134d-4c81-8a30-5d5e0bf11884

Frame by frame:

**Search keywords**:

atomiks:

Closes #979
Addresses the `alignOptionToTrigger` non-animated use case of #965

`runOnceAnimationsFinish` is wrapped in a `requestAnimationFrame` (to know the ending styles) and then a microtask. By flushing the `setMounted` state update synchronously, the browser won't paint, leading to no flicker.

### DIFF
--- a/packages/react/src/menu/root/MenuRoot.test.tsx
+++ b/packages/react/src/menu/root/MenuRoot.test.tsx
@@ -726,7 +726,7 @@ describe('<Menu.Root />', () => {
       });
     });
 
-    it('should remove the popup and the animation finishes', async function test(t = {}) {
+    it('should remove the popup when the animation finishes', async function test(t = {}) {
       if (/jsdom/.test(window.navigator.userAgent)) {
         // @ts-expect-error to support mocha and vitest
         // eslint-disable-next-line @typescript-eslint/no-unused-expressions
--- a/packages/react/src/menu/root/MenuRoot.test.tsx
+++ b/packages/react/src/menu/root/MenuRoot.test.tsx
@@ -694,7 +694,7 @@ describe('<Menu.Root />', () => {
   });
 
   describe('controlled mode', () => {
-    it('should remove the popup when animated=true and there is no exit animation defined', async function test(t = {}) {
+    it('should remove the popup when and there is no exit animation defined', async function test(t = {}) {
       if (/jsdom/.test(window.navigator.userAgent)) {
         // @ts-expect-error to support mocha and vitest
         // eslint-disable-next-line @typescript-eslint/no-unused-expressions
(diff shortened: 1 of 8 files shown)


packages/react/src/menu/root/MenuRoot.test.tsx
```diff
@@ -722,7 +726,7 @@ describe('<Menu.Root />', () => {
       });
     });
 
-    it('should remove the popup when animated=true and the animation finishes', async function test(t = {}) {
+    it('should remove the popup and the animation finishes', async function test(t = {}) {
```
michaldudak, line 729:
```suggestion
    it('should remove the popup when the animation finishes', async function test(t = {}) {
```

packages/react/src/popover/root/PopoverRoot.test.tsx
```diff
@@ -174,7 +178,7 @@ describe('<Popover.Root />', () => {
       expect(handleChange.firstCall.args[0]).to.equal(false);
     });
 
-    it('should remove the popup when animated=true and there is no exit animation defined', async function test(t = {}) {
+    it('should remove the popup and there is no exit animation defined', async function test(t = {}) {
```
michaldudak, line 181:
```suggestion
    it('should remove the popup when there is no exit animation defined', async function test(t = {}) {
```

packages/react/src/menu/root/MenuRoot.test.tsx
```diff
@@ -761,7 +765,7 @@ describe('<Menu.Root />', () => {
             <style dangerouslySetInnerHTML={{ __html: style }} />
             <button onClick={() => setOpen(false)}>Close</button>
             <Menu.Root open={open}>
-              <Menu.Positioner>
+              <Menu.Positioner keepMounted>
```
michaldudak, line 768: This now tests a different case. Does it not work without `keepMounted` anymore?

atomiks: Nope. It unmounts before it gets a chance to fire

atomiks: `queryByRole` tests for visibility though, which is why the test still passes. It sees `hidden` and thinks it's unmounted and will return `null`. So it tests almost the same thing still.

michaldudak (approved):

Issue #961, opened 2024-12-05
[meter][progress][slider] Unify formatting API


mj12albert:
NumberField uses [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat), we can follow the `format` prop and adapt it for these range widgets:

#### Meter/Progress:
- ✅ ~~Add `Value` subcomponent~~ https://github.com/mui/base-ui/pull/1355
- ✅ ~~Add `format` prop~~ https://github.com/mui/base-ui/pull/1355
- ✅ ~Add `locale` prop~
- 
#### Slider:
- ✅ ~~Add `format` prop~~ https://github.com/mui/base-ui/issues/459
- Add `locale` prop to match https://github.com/mui/base-ui/pull/1488

mj12albert:
@vladmoroz Wonder what you think about this quirk for formatting Slider? I'm not sure if feels similar enough to a conditional type to trigger you 😅

mj12albert:
@colmtuite What do you think about `Progress.Value`/`Meter.Value` ?

vladmoroz:
I agree with the suggestions

I'd also prefer if we converge on `Slider.Value` instead of `Slider.Output` too for consistency across the board.

mj12albert:
> Slider.Value instead of Slider.Output

The thing is Slider.Output is a `<output>` element which isn't suitable for the other components though

vladmoroz:
> The thing is Slider.Output is a <output> element which isn't suitable for the other components though

Maybe it's just my opinion, but it's an internal detail that’s opaque to the user. To me, consistency of the API within the library is more important because there's fewer part names to remember.

I feel the same way about `Select.Option` vs `Select.Item`. The`Option` name sticks out compared to everything else in Base UI.

colmtuite:
@vladmoroz @mj12albert Agree on both. `Select.Item` and `Slider.Value` ftw.

mj12albert:
I've found that the array form for Slider `<Slider format={[options1, options2} />` is not good, it will cause the `aria-valuetext` of each input element will be announced with a different format which sounds weird.

Instead the (display value) rendering should use `children`:
```jsx
<Slider.Value>{(formattedValues, values) => ''/* e.g. return a string */}</Slider.Value>
```

PR #990, opened 2024-12-09
[Slider] Change `Slider.Output` to `Slider.Value` and add `format` prop

mj12albert:
Part of https://github.com/mui/base-ui/issues/961

- `Slider.Value` matches the rendering API of [`Select.Value`](https://master--base-ui.netlify.app/react/components/select#value)
- `format` matches `NumberField` and does not take an array for range sliders, as it would make `aria-valuetext` for range sliders inconsistent ([comment](https://github.com/mui/base-ui/issues/961#issuecomment-2527118439))


packages/react/src/slider/value/SliderValue.tsx
```diff
@@ -6,34 +6,49 @@ import { useComponentRenderer } from '../../utils/useComponentRenderer';
...
 /**
- * Displays the current value of the slider as text.
- * Renders an `<output>` element.
+ * Demos:
+ *
+ * * [Slider](https://base-ui.com/components/react-slider/)
+ *
+ * API:
+ *
+ * * [SliderValue API](https://base-ui.com/components/react-slider/#api-reference-SliderValue)
```
mnajdova, line 17: This shouldn't be here, it's using the old format.

mnajdova (commented):
Can you please update the demo on the new docs page and ask Vlad for a review? The implementation itself looks good, but the comment I added above causes issues on the docs.

Screenshot 2024-12-11 at 19 30 16

docs/reference/overrides/slider-root.json
```diff
@@ -1,6 +1,9 @@
 {
```
mj12albert, line 1: @vladmoroz The `*.Value` components also support children as a function (with some props) for formatting: https://github.com/mui/base-ui/blob/master/docs/data/components/select/select.mdx?plain=1#L179-L183

Should we add them to `docs/reference/overrides` and displayed as well?

vladmoroz: > Should we add them to docs/reference/overrides to show them in the props table as well?

Yes please!

vladmoroz (approved):

PR #992, opened 2024-12-09
[Menu, Popover, PreviewCard, Select, Tooltip] Force `transition: none` while in `starting-style` status

atomiks:

This lets the correct transition play while waiting for the correct side to be calculated as the result of collision detection. Animations appear to be unaffected by this.


vladmoroz (approved):
@atomiks do we need to remove the animations too?

atomiks:
@vladmoroz I made a demo with animations and they appear to work normally already
https://deploy-preview-992--base-ui.netlify.app/experiments/anchor-side-animations

vladmoroz:
OK!

docs/src/app/(private)/experiments/tooltip.tsx
```diff
@@ -188,6 +188,12 @@ export default function TooltipTransitionExperiment() {
       </Tooltip.Root>
 
       <h3>CSS Transition with `@starting-style`</h3>
+      <p>
+        No longer supported due to{' '}
+        <a href="https://github.com/mui/base-ui/pull/992">
+          https://github.com/mui/base-ui/pull/992
+        </a>
+      </p>
```
atomiks, line 196: `@starting-style` no longer works with this change. Considering it doesn't handle this issue either, it shouldn't be encouraged. This is better for consistency if they always use `[data-starting-style]` and `[data-ending-style]` hooks anyway.

PR #1015, opened 2024-12-09
[docs, core] Update component annotations URLs and copy

vladmoroz:
~Pending https://github.com/mui/material-ui/pull/44707 first~

- Update component annotations
- Remove unused data


packages/react/src/direction-provider/DirectionProvider.tsx
```diff
@@ -6,13 +6,7 @@ import { DirectionContext, type TextDirection } from './DirectionContext';
...
  * A provider that configures RTL/LTR behavior for part of an app or a whole app.
  *
- * Demos:
- *
- * - [Direction Provider](https://base-ui.com/components/react-direction-provider/)
- *
- * API:
- *
- * - [DirectionProvider API](https://base-ui.com/components/react-direction-provider/#api-reference-DirectionProvider)
+ * Documentation: [Base UI Direction Provider](https://base-ui.com/react/components/direction-provider)
```
michaldudak, line 9: This should be under /react/utilites

vladmoroz: Nice catch, I managed to fumble `DirectionProvider` as `DirectoryProvider` somehow
https://github.com/mui/base-ui/pull/1015/commits/aad149816529a318271909182ffc02cf9e4e234d

scripts/buildApiDocs/config/getBaseUiComponentInfo.ts
```diff
@@ -1,74 +1,44 @@
...
     throw new Error(`Could not find the component name from: ${filename}`);
   }
 
-  const demos = getBaseUiDemos(name);
+  const parentComponentPath = filename.split(componentsDir)[1]!.split('/')[0];
+  const parentComponentName = startCase(parentComponentPath);
+  const componentUrl = overrides[name] ?? `/components/${parentComponentPath}`;
+  const url = urlJoin(baseUrl, componentUrl);
+
+  const customAnnotation = `Documentation: [Base UI ${parentComponentName}](${url})`;
```
vladmoroz, line 30: Just to be clear: the generated URL is based on the path in `/packages/react/src`

E.g. components created in `/packages/react/src/dialog/*` will be annotated with `https://base-ui.com/react/components/dialog`

scripts/buildApiDocs/config/getBaseUiComponentInfo.ts
```diff
@@ -1,74 +1,44 @@
...
 
   return {
     filename,
     name,
     muiName: name,
-    apiPathname: getApiPath(demos, name) ?? '',
-    apiPagesDirectory: path.join(process.cwd(), `docs/data/api`),
+    apiPathname: name, // Not used, but we pass `name` just for logging when the script runs
+    apiPagesDirectory: path.join(process.cwd(), 'docs/reference/temp/components'),
```
mnajdova, line 37: From here, later they are merged with overrieds and added in generated is that correct? Just making sure I understand the flow.

mnajdova (approved):

scripts/buildApiDocs/config/getBaseUiComponentInfo.ts
vladmoroz: Yes, `buildReference.ts` runs after, on top of the output of the monorepo infra

PR #1016, opened 2024-12-09
[select] Handle additional option -> item changes

atomiks:


michaldudak:
There are more occurrences of the word "option" in code, especially in JSDoc comments and docs.

atomiks:
@michaldudak yeah it's such a common word used everywhere it's hard to spot. The remaining ones seem to just be in the old docs now and don't need to be replaced

michaldudak (approved):
I found two more in useSelectItem (lines 208 and 212)

Issue #1012, opened 2024-12-09
[slider] API improvements


colmtuite:
- Render divs instead of spans, for easier styling.
- Set `position: relative` inline on Track.
- Remove `position: absolute` from Indicator.

**Search keywords**:

PR #1017, opened 2024-12-10
[Slider] API Tweaks

mj12albert:
- Remove `position: absolute` on indicator, add `position: relative` on track
- Use divs instead of spans

Closes https://github.com/mui/base-ui/issues/1012


packages/react/src/slider/thumb/useSliderThumb.ts
```diff
@@ -110,17 +109,16 @@ export function useSliderThumb(parameters: useSliderThumb.Parameters): useSlider
     return {
       position: 'absolute',
       [{
-        horizontal: 'left',
-        'horizontal-reverse': 'right',
+        horizontal: 'insetInlineStart',
```
mj12albert, line 112: @vladmoroz Since we decided that with DirectionProvider, and users are still responsible for setting HTML `dir` themselves, I wanted to use a logical property here instead of left/right, what do you think?
(The new DirectionProvider doc should mention this as well)

vladmoroz: This makes sense, but I'm still seeing `dir="rtl"` on the slider in the direction provider demo. Are you planning to remove it? (I don't mind it)

https://deploy-preview-1017--base-ui.netlify.app/react/utilities/direction-provider

image

mj12albert: Yup, I've updated the (new) demo as well: https://github.com/mui/base-ui/pull/1017/commits/569bdf0a065d07a4d2914bbc0cd04d8443d98edd
@vladmoroz If its ok I'll merge it later

vladmoroz: Sure. That'd be a separate PR?

mj12albert: Nope! updated here in this commit: https://github.com/mui/base-ui/commit/569bdf0a065d07a4d2914bbc0cd04d8443d98edd @vladmoroz

docs/src/app/(public)/(content)/react/utils/direction-provider/page.mdx
```diff
@@ -10,15 +10,19 @@
...
-  {/* Your app or a group of components */}
-</DirectionProvider>
+<html dir="rtl">
+  <body>
+    {/* prettier-ignore */}
+    <DirectionProvider>
+      {/* Your app or a group of components */}
+    </DirectionProvider>
+  </body>
+</html>;
```
vladmoroz, line 25: I'd keep the anatomy and the description as it was before since we don't typically show the API or the required props to make the components work in the Anatomy section. 

I think a code block also might suggest a few things that aren't really true:
- that we expect `dir="rtl"` on `html` specifically
- that Direction Provider depends on `dir`

Also, introducing `<html dir="rtl">` in the anatomy begs to add `direction="rtl"` in the `<DirectionProvider>` below, at which point the docs get kinda muddy 

imo they'll do native `dir` anyway so should be a non-issue

colmtuite (approved):

Issue #1004, opened 2024-12-09
[docs] Document data attributes


vladmoroz:

vladmoroz:
@mnajdova do you think we can set up something closer to the components source similarly to how you did it with the CSS variables?

PR #1022, opened 2024-12-10
[code-infra] Add support for automatic generation of data-attributes API

mnajdova:
Preview: https://deploy-preview-1022--base-ui.netlify.app/react/components/collapsible#trigger

Part of https://github.com/mui/base-ui/issues/1004


mnajdova:
@vladmoroz @michaldudak would be great if you can check it. If we are good I will proceed with getting a review & merging the core's PR first.

docs/reference/overrides/dialog-trigger.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "DialogTrigger",
-  "attributes": {
-    "[data-modal]": {
+  "dataAttributes": {
+    "data-modal": {
```
vladmoroz, line 4: I was kind of deliberate about naming it "attributes" since in the old docs, I saw certain components exposed some `aria-*` attributes in the documentation, so it's not just data-attributes. Should we keep it this way?

packages/react/src/collapsible/trigger/CollapsibleTriggerDataAttributes.ts
```diff
@@ -0,0 +1,6 @@
+export enum CollapsibleTriggerDataAttributes {
+  /**
+   * Indicates whether the collapsible panel is opened.
+   */
+  panelOpen = 'data-panel-open',
+}
```
vladmoroz, line 6: Should this be imported by the actual component code?

docs/reference/overrides/dialog-trigger.json
michaldudak: We agreed to only use custom data attributes for style hooks. I don't have a preference about `attributes` vs `dataAttributes` here, though.

docs/src/components/reference/AttributesTable.tsx
```diff
@@ -38,10 +38,12 @@ export async function AttributesTable({ data, ...props }: AttributesTableProps)
             useMDXComponents: () => inlineMdxComponents,
           });
 
+          const attributeName = name.startsWith('[') ? name : `[${name}]`;
```
michaldudak, line 41: Do we need to show brackets in the docs? Since they will be around every attribute, it's more noise IMO.

docs/reference/overrides/dialog-trigger.json
vladmoroz: Cool I am probably mixing it up with non-style attributes that we had documented on some components
image

Not sure how I feel about those, but that's separate

docs/src/components/reference/AttributesTable.tsx
vladmoroz: I think they mainly help you identify that you are looking at attributes for styling, check it in the context:

https://master--base-ui.netlify.app/react/components/dialog

packages/react/src/collapsible/trigger/CollapsibleTriggerDataAttributes.ts
mnajdova: We could replace the actual code instead of hard-coding this to use the enum. I haven't updated the code as part of the PR tough (btw, this is why I didn't include the brackets in the value.

docs/src/components/reference/AttributesTable.tsx
mnajdova: I was hesitant about this too, I had to remove it from the enum so we can use it in the code as an actual value. I would personally remove, but I didn't wanted to change previous behavior.

docs/reference/overrides/dialog-trigger.json
mnajdova: I don't think we document any aria attributes we add, and yep, the rest of them are just props that we accept. I would keep `dataAttributes` to be explicit, if in the future we want to document other things, we anymay may want to re-think the API section.

docs/src/components/reference/AttributesTable.tsx
vladmoroz: Cool agreed

packages/react/src/collapsible/trigger/CollapsibleTriggerDataAttributes.ts
vladmoroz: Makes sense

vladmoroz (approved):

mnajdova:
Is it us or Next.js doing some weird caching, I am seeing the attributes table with outdated info: https://deploy-preview-1022--base-ui.netlify.app/react/components/collapsible#trigger. The JSONs are updated as expected (looks like it remembers the first format it had). It happens locally too. I even tried removing the `dataAttributes` item from the JSON, but this info was still there.

----

Edit: Cleaning the cache worked.

PR #1027, opened 2024-12-10
[tabs] Do not display the indicator before layout is settled

### LINKED ISSUE #998: [tabs] Indicator animates on page load
Transitions and animations should be either prevented until after the CSS variables are set, or the indicator should not be rendered at all until the measurements are made

https://deploy-preview-996--base-ui.netlify.app/new/components/tabs

**Search keywords**:

michaldudak:
Hide the indicator if the active tab's width or height is 0 (meaning its layout is not settled just yet). Also fixed the prehydration script

Fixes #998


docs/src/app/(public)/(content)/react/components/tabs/demos/hero/css-modules/index.tsx
```diff
@@ -15,7 +15,7 @@ export default function ExampleScrollArea() {
         <Tabs.Tab className={styles.Tab} value="account">
           Account
         </Tabs.Tab>
-        <Tabs.Indicator className={styles.Indicator} />
+        <Tabs.Indicator className={styles.Indicator} renderBeforeHydration />
```
vladmoroz, line 18: Just double checking: `renderBeforeHydration` isn't required for tabs that don't animate on page load? If yes, this is probably unnecessary in the demos too?

atomiks: > Any reason to not set this problem to true by default? (And even remove the prop, potentially)

Agree here

michaldudak: It adds a script to every instance of the Indicator, so the application would grow in size. I can look into including just one instance of the script, though.

atomiks (approved):

Issue #1024, opened 2024-12-10
[Collapsible] Invalid `aria-controls` on the trigger when closed and `keepMounted={false}`


mj12albert:
![image](https://github.com/user-attachments/assets/8c11657b-f6ef-4450-8e38-42eb4051dbdf)

**Search keywords**:

mj12albert:
`aria-controls` can reference a hidden element though there are strong objections: https://github.com/w3c/aria/issues/531

PR #1028, opened 2024-12-10
[Collapsible] Fix trigger `aria-controls` referencing nonexistent `id`

mj12albert:
Closes https://github.com/mui/base-ui/issues/1024

### DIFF
--- a/packages/react/src/collapsible/panel/useCollapsiblePanel.ts
+++ b/packages/react/src/collapsible/panel/useCollapsiblePanel.ts
@@ -77,10 +77,14 @@ export function useCollapsiblePanel(
 
   useEnhancedEffect(() => {
     setPanelId(id);
+
+    if (!keepMounted && !open) {
+      setPanelId(undefined);
+    }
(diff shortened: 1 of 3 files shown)


packages/react/src/collapsible/panel/useCollapsiblePanel.ts
```diff
@@ -77,10 +77,14 @@ export function useCollapsiblePanel(
 
   useEnhancedEffect(() => {
     setPanelId(id);
+
+    if (!keepMounted && !open) {
+      setPanelId(undefined);
+    }
```
michaldudak, line 83: No need to set the state twice:

```suggestion
    if (!keepMounted && !open) {
      setPanelId(undefined);
    } else {
      setPanelId(id);
    }
```

mj12albert: @michaldudak Fixed ~

michaldudak (approved):

Issue #1003, opened 2024-12-09
[docs] Use new demos for the visual tests


vladmoroz:
@mnajdova @michaldudak is this an alpha thing?

For context, I have already seen regressions in the new demos due to recent changes, and plan to remove the `data` folder altogether, recreating as many advanced demos as possible

colmtuite:
I don't think demos and tests should be connected. Testing should be independent of presentation.

mj12albert:
@vladmoroz You are referring to the Argos check in our CI?

vladmoroz:
> @vladmoroz You are referring to the Argos check in our CI?

Yep

> I don't think demos and tests should be connected. Testing should be independent of presentation.

There are visual tests that already are though. Every old demo is visually diffed as part of the CI on PRs.

michaldudak:
It's expected that Argos will report issues often now that we're recreating the demos. Once the demos are stable, Argos helps us catch bugs that can result in small changes in presentation.

vladmoroz:
Right, @michaldudak, so what I'm saying is that I want Argos to check against the new demos since they aren't hooked up yet

mj12albert:
Happy to take this one 👋

PR #1030, opened 2024-12-10
[core] Run argos tests on new docs

mj12albert:
Updated the test setup to read the new directory structure and updated the Webpack config to load CSS modules and font files

Example of the updated screenshots from the [CI run](https://app.argos-ci.com/mui/base-ui/builds/5712):
image

Screenshot 2024-12-11 at 5 45 18 PM

Closes https://github.com/mui/base-ui/issues/1003


test/regressions/index.js
```diff
@@ -69,7 +69,12 @@ function excludeDemoFixture(suite, name, path) {
 }
 
 // Also use some of the demos to avoid code duplication.
-const importDemos = require.context('docs/data', true, /\.js$/, 'lazy');
+const importDemos = require.context(
+  'docs/src/app/(public)/(content)/react/components',
+  true,
+  /\.tsx$/,
```
mj12albert, line 75: Have we decided to have all the demos in the current dir structure going forward instead of `docs/data`?
And also no more `.jsx` version of the demos?
@vladmoroz @michaldudak

vladmoroz: Yeah, I don't think the JSX version is needed. It's virtually the same almost in all cases.

No plans for `docs/data` other than to migrate its contents to app directory and delete it 🙂

Also while we are here
- We might have demos in the `utils` folder too
- Also as a rule I've also been putting demos into the `demos` folder to make search easier in VS Code, maybe you can rely on this here. We can then safely add components beside individual pages if we ever need to.

mnajdova: > No plans for docs/data other than to migrate its contents to app directory and delete it 🙂

We can leave this for the end (maybe even after the release, as it would break some things in the infra, it may require some mork work).

mj12albert: I've set the path to `docs/src/app/(public)/(content)/react` to keep it simple, even in the old docs structure the contents were 100% demos anyway

test/regressions/index.js
```diff
@@ -161,12 +166,9 @@ function App(props) {
...
   React.useEffect(() => {
     webfontloader.load({
-      google: {
-        families: ['Roboto:300,400,500,700', 'Inter:300,400,500,600,700,800,900', 'Material+Icons'],
-      },
       custom: {
-        families: ['Font Awesome 5 Free:n9'],
-        urls: ['https://use.fontawesome.com/releases/v5.1.0/css/all.css'],
+        families: ['Unica 77'],
+        urls: ['../../docs/src/styles.css'],
```
mj12albert, line 171: The @font-face url is relative to this file, so directly using `fonts/index.css` doesn't work

test/regressions/index.js
```diff
@@ -4,8 +4,8 @@ import * as ReactDOMClient from 'react-dom/client';
 import { BrowserRouter as Router, Routes, Route, Link } from 'react-router-dom';
 import webfontloader from 'webfontloader';
 import TestViewer from './TestViewer';
+import 'docs/src/styles.css';
 import 'docs/src/demo-theme.css';
```
vladmoroz, line 8: We don't need this if we are loading `docs/src/styles.css`
`demo-theme.css` is just for presenting the theme.css tab in the demos

vladmoroz (approved):
Looks great, thanks for handling it @mj12albert

mnajdova (approved):

PR #1032, opened 2024-12-10
[test] Fix warnings from flushSync in tests

michaldudak:
After the recent addition of `flushSync` to Collapsible, there were many warnings in tests related to it. When animations are disabled, `runOnceAnimationsFinish` is called synchronously during rendering, which is what `flushSync` complains about.


mj12albert:
`flushSync` was always there, it was the default `keepMounted` that changed from true to false recently 🤔
(I also got this warning today running jsdom tests locally)

atomiks:
https://github.com/mui/base-ui/pull/988 already fixes this

Issue #993, opened 2024-12-09
[core] Render function props override `className`


vladmoroz:
This JSX:

```jsx
import * as React from 'react';
import { Toggle } from '@base-ui-components/react/toggle';
import styles from './index.module.css';

export default function ExampleToggle() {
  return (
    <div className={styles.Panel}>
      <Toggle
        aria-label="Favorite"
        render={(props, state) => {
          if (state.pressed) {
            return (
              <button type="button" className={styles.Button} {...props}>
                <HeartFilledIcon className={styles.Icon} />
              </button>
            );
          }

          return (
            <button type="button" className={styles.Button} {...props}>
              <HeartOutlineIcon className={styles.Icon} />
            </button>
          );
        }}
      />
    </div>
  );
}
```

produces the following DOM:

<img width="814" alt="image" src="https://github.com/user-attachments/assets/4565ed74-6d50-425c-9600-939850700efd">

CSS class is missing.

Haven't tested other components.

**Search keywords**:

mj12albert:
@vladmoroz Since you didn't put a class on `Toggle` (only inside the render fn), I think it should be a simple fix of not including `className: undefined` in the render props.

But if you were to put classes in both places, we wouldn't be able to merge the classes, do you think it matters 🤔 

```jsx
// i know this is kind of a contrived example
<Toggle
  className={styles.Toggle}
  render={(props, state) => {
    if (state.pressed) {
      return (
        <button type="button" className={styles.pressed} {...props}>
          heart filled
        </button>
      );
    }

    return (
      <button type="button" className={styles.unpressed} {...props}>
        heart outline
      </button>
    );
  }}
/>
```

vladmoroz:
> But if you were to put classes in both places, we wouldn't be able to merge the classes, do you think it matters 🤔

Why not? Radix merges classes when using `asChild` and even event handlers. I'd expect the same on all components

mj12albert:
The short form `render={<MyComponent />}` will merge everything! But having finer control with the long form also means you have ensure you don't overwrite the render props yourself.
(I suppose the render prop would be explained in more detail, probably in [composition](https://github.com/mui/base-ui/issues/1007)?)

vladmoroz:
Oh 🤦‍♂️
I'm just realising my mistake. Of course it makes sense, we don't merge things in the long form.

Dismiss the issue then

mj12albert:
No worries, still found a minor bug from this ~

PR #1041, opened 2024-12-11
[core] Exclude `undefined` className from render fn props

mj12albert:
Fixes https://github.com/mui/base-ui/issues/993

When `className` is unspecified on the component, className inside the render function should not be overwritten by the renderFn's `props.className`:

```jsx
<Toggle
  aria-label="Favorite"
  render={(props, state) => {
    if (state.pressed) {
      return (
        <button type="button" className={styles.Button} {...props}>
          <HeartFilledIcon />
        </button>
      );
    }

    return (
      <button type="button" className={styles.Button} {...props}>
        <HeartOutlineIcon />
      </button>
    );
  }}
/>
```


mj12albert:
> Test would be nice

Added ~

PR #1046, opened 2024-12-11
[docs] Remove transpiled TS → JS demo files

vladmoroz:
We don't plan to show JS versions in the docs since they are virtually the same as TS


oliviertassinari:
Analytics: today ~15% of the people change the default to have the JavaScript version: https://www.notion.so/mui-org/Product-Analytics-f2e2576203564e6c8dc3429cf305a36f?pvs=4#95d2b8b3873b404f951545a099d40cf4, trending downward from before.

vladmoroz:
Deleted that in the cleanup PR

PR #1052, opened 2024-12-12
[Checkbox, Select, Switch] Point `id` to control element

### LINKED ISSUE #1051: [Field] Base UI components link to hidden input `id`
In Checkbox, Switch, Select, the `id` points to the hidden `<input>` which causes a violation when clicking `Field.Label` and focus is not placed on the rendered control/button

**Search keywords**:

atomiks:
Closes #1051

### DIFF
--- a/packages/react/src/select/root/SelectRoot.tsx
+++ b/packages/react/src/select/root/SelectRoot.tsx
@@ -68,6 +68,10 @@ const SelectRoot: SelectRoot = function SelectRoot<Value>(
         </PortalContext.Provider>
         <input
           {...rootContext.fieldControlValidation.getInputValidationProps({
+            onFocus() {
+              // Move focus to the trigger element when the hidden input is focused.
+              rootContext.triggerElement?.focus();
+            },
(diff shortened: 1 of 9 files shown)


michaldudak (approved):

packages/react/src/select/root/SelectRoot.tsx
```diff
@@ -69,7 +69,7 @@ const SelectRoot: SelectRoot = function SelectRoot<Value>(
         <input
           {...rootContext.fieldControlValidation.getInputValidationProps({
             onFocus() {
-              // Move focus from the hidden <select> to the trigger element.
+              // Move focus to the trigger element when the hidden input is focused.
               rootContext.triggerElement?.focus();
```
atomiks, line 73: Small amendment: unlike Checkbox and Switch, with Select it's still pointing to the hidden `<input>`, but focused is moved straight away to the trigger. This prevents the warning and also prevents the select from opening when clicking on the label unexpectedly.

PR #1053, opened 2024-12-12
[field controls] Respect `validationMode`

### LINKED ISSUE #1050: [Field] Base UI components don't respect `validationMode` consistently
`validationMode="onChange"` is respected in `NumberField` and `Select`, but always also fires on blur. Some components like `Slider` also have some behavioral differences

**Search keywords**:

atomiks:
Closes #1050


packages/react/src/checkbox-group/CheckboxGroup.tsx
```diff
@@ -7,6 +7,7 @@ import { CheckboxGroupContext } from './CheckboxGroupContext';
 import type { FieldRoot } from '../field/root/FieldRoot';
```
mj12albert, line 7: Don't think it's directly related to this PR, but I couldn't make a required checkbox group because `required` is a prop on individual checkboxes but not on CheckboxGroup:
```jsx
// a combination of the CheckboxGroup and Field demos
export default function FieldCheckbox() {
  return (
    <Field.Root>
      <Field.Label>Pick one or more</Field.Label>

      <CheckboxGroup defaultValue={[]}>
        <label>
          <Checkbox.Root name="fuji-apple">
            <Checkbox.Indicator>
              <CheckIcon />
            </Checkbox.Indicator>
          </Checkbox.Root>
          Fuji
        </label>

        <label>
          <Checkbox.Root name="gala-apple">
            <Checkbox.Indicator>
              <CheckIcon />
            </Checkbox.Indicator>
          </Checkbox.Root>
          Gala
        </label>

        <label>
          <Checkbox.Root
            name="granny-smith-apple"
          >
            <Checkbox.Indicator>
              <CheckIcon />
            </Checkbox.Indicator>
          </Checkbox.Root>
          Granny Smith
        </label>
      </CheckboxGroup>

      <Field.Error match="valueMissing">
        You must pick at least one
      </Field.Error>
    </Field.Root>
  );
}
```

packages/react/src/radio-group/RadioGroup.tsx
```diff
@@ -9,6 +9,8 @@ import { useDirection } from '../direction-provider/DirectionContext';
 import { useRadioGroup } from './useRadioGroup';
```
mj12albert, line 9: https://codesandbox.io/p/sandbox/sweet-brattain-q9ptqr

When I make a required RadioGroup that has no initially checked radio and tab through it without checking any of the radios, it doesn't trigger an error state (I logged some stuff and it seems the validation functions are being called)

Also when the required RadioGroup is wrapped in a Form (with a submit button), if I click the submit button without touching the RadioGroup the error doesn't appear which is probably the same issue.

However when a radio is checked, the `console.log` in the Form's submit handler doesn't log the value 🤔

atomiks: Are these bugs (with CheckboxGroup as well) related to this PR or are they on master as well? If on master, we can create a new issue for them to separate the work

mj12albert: Sounds good, I think it's just something wrong with validating a required field with certain types of values (non-string?)

mj12albert (approved):
Looks good, I tested all of the controls here with the validate function (because `required`+`valueMissing` wasn't consistent, likely a different bug) and the validationMode behavior seems consistent now:
```jsx
<Field.Root
  className={styles.field.Field}
  validate={(value) => {
    if (value == null) {
      return 'You must pick something';
    }
    if (value === 'comic-sans') {
      return 'Comic Sans is not allowed';
    }
    return null;
  }}
>
  <Select.Root required>
    {/* rest of the select */}
  </Select.Root>
</Field.Root>
```

Side note: I think the using ValidityState terminology for the [match prop](https://master--base-ui.netlify.app/react/components/field#error) isn't ideal, some of the values seem vague/unintuitive, e.g. `match="badInput"`, `match="valid"`
Many users may end up just using a `validate` function on Field and `match="customError"` for everything just because of convenience CC @colmtuite

atomiks:
> Side note: I think the using ValidityState terminology for the [match prop](https://master--base-ui.netlify.app/react/components/field#error) isn't ideal, some of the values seem vague/unintuitive, e.g. match="badInput", match="valid"
Many users may end up just using a validate function on Field and match="customError" for everything just because of convenience (and because many form libraries are like this) CC @colmtuite

I see your point - though in the most basic usage, it allows them to leverage the native HTML validation while still customizing the error message. e.g. `<input pattern="..." />` without using JavaScript inside the `validate` function. Maybe it'd be more intuitive/obvious if each string was documented?

Issue #1031, opened 2024-12-10
[docs] Preserve part name case


vladmoroz:
Preserve part name case but make sure anchor URLs stay readable
https://mui-org.slack.com/archives/C02P87NQLJC/p1733837465743209

**Search keywords**:

PR #1056, opened 2024-12-12
[docs] Remove space from part names but preserve anchor urls

mj12albert:
Closes https://github.com/mui/base-ui/issues/1031

After the change: https://deploy-preview-1056--base-ui.netlify.app/react/components/menu#radio-item-indicator

### DIFF
--- a/docs/src/components/reference/rehypeReference.mjs
+++ b/docs/src/components/reference/rehypeReference.mjs
@@ -2,7 +2,6 @@
 import { readFileSync, existsSync } from 'node:fs';
 import { visitParents } from 'unist-util-visit-parents';
 import kebabCase from 'lodash/kebabCase.js';
-import startCase from 'lodash/startCase.js';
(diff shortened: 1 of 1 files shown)


docs/src/components/reference/rehypeReference.mjs
```diff
@@ -88,7 +88,10 @@ export function rehypeReference() {
             subtree.push(
               createMdxElement({
                 name: 'h3',
-                children: [{ type: 'text', value: name }],
+                children: [{ type: 'text', value: name.replace(/\s/g, '') }],
+                props: {
+                  id: name.toLowerCase().replace(/\s/g, '-'),
+                },
```
vladmoroz, line 94: Check `startCase()` just above where the original conversion from PascalCase happens. We can preserve the original name and just use `kebabCase()` to create a hyphenated anchor id

vladmoroz (approved):

PR #1060, opened 2024-12-12
[docs] Various fixes and cleanup

vladmoroz:
- Add focus outline to the switch demo
- Fix the demo code block accessibility violation
- Fix horizontal space on mobile
- Fix a screen reader issue with the docs select component
- Reduce docs min width to `320px`
- Bump up the timeout to force hide the scrollbar after the programmatic scroll in side nav

### DIFF
--- a/docs/src/components/demo/Demo.css
+++ b/docs/src/components/demo/Demo.css
@@ -152,9 +152,18 @@
       display: flex;
       cursor: text;
 
-      /* Closed state */
-      overflow: hidden;
-      max-height: calc(8.6lh + 0.5rem); /* Show 8.6 of code lines and account for top padding */
+      /* Scroll */
+      overflow: auto;
+      overscroll-behavior-x: contain;
+      scrollbar-width: thin;
+
+      /* Scroll containers may be focusable */
+      &:focus-visible {
+        position: relative;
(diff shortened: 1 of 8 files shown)


docs/src/components/demo/Demo.css
```diff
@@ -163,57 +172,35 @@
...
-      }
+    & code {
+      /* Different fonts may introduce vertical align issues */
+      display: block;
+      /* Make sure selection highlight spans full container width in Safari */
+      flex-grow: 1;
     }
 
-    &[data-open] {
+    &[data-closed] {
```
vladmoroz, line 182: We didn't have the closed state before iirc, much easier to style now. Not undoing any styles anymore

atomiks:
In Chrome responsive mode it stops resizing correctly below <360px still

Screenshot 2024-12-12 at 8 45 29 pm

colmtuite (approved):

vladmoroz:
Hmm @colmtuite it might be a bug of the responsive mode in Chrome, I don't have another explanation otherwise.

I also see it, but it doesn't happen in Safari's responsive mode and doesn't happen when resizing the Chrome's window with dev tools normally:

image

Issue #1061, opened 2024-12-12
[Select] The `onValueChange` and `onOpenChange` typing are not correct on the doc


flaviendelangle:
# Docs feedback

## How can we improve Base UI documentation?

Whenever possible, the props should have a detailed type in the API section of the doc, not `function`.
Looking at the code I can't tell why only the Select one does not work :shrug: 

The Select API doc:

![image](https://github.com/user-attachments/assets/e8fd2fb5-585d-4be4-b44c-686d79ba4bb6)

The Menu API doc:  (same for Dialog and the Popover)

![image](https://github.com/user-attachments/assets/384780e8-b2f7-4136-93ab-2fa30de49f01)

I did not check if they were other occurences of `function` in the doc 

**Search keywords**:

mj12albert:
So `function` is just generated by `pnpm proptypes`, but we now have this `overrides` directory with some json ([Menu for example](https://github.com/mui/base-ui/blob/master/docs/reference/overrides/menu-root.json)) that overrides the output of that for the props table in the docs site, so the one for Select was missing!

PR #1062, opened 2024-12-12
[docs-infra] Tweak `buildReference` script

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1061
Also fixes https://github.com/mui/base-ui/pull/990#discussion_r1881740914


scripts/buildApiDocs/buildReference.ts
```diff
@@ -70,7 +70,7 @@ export async function buildReference() {
     const descriptionData: PropsTranslations = JSON.parse(descriptionJsonContents);
 
     for (const prop in props) {
-      props[prop].description = descriptionData.propDescriptions[prop]?.description;
+      props[prop].description ??= descriptionData.propDescriptions[prop]?.description;
```
mj12albert, line 73: Fixes when a `description` is specified in `overrides/*.json` but not reflected in the generated one

scripts/buildApiDocs/buildReference.ts
```diff
@@ -93,7 +93,7 @@ export async function buildReference() {
     const json: ComponentDef = {
       name: componentData.name,
       description: descriptionData.componentDescription,
-      props,
+      props: Object.fromEntries(Object.entries(props).sort()),
```
mj12albert, line 96: If an "overrides" is added manually like `children` (and isn't actually overriding anything generated) the props would not be sorted alphabetically

scripts/buildApiDocs/buildReference.ts
mj12albert: (Also probably only happens to something that is not literally an "override" like `children`)

docs/reference/overrides/select-value.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "SelectValue",
+  "props": {
+    "children": {
+      "type": "React.ReactElement | (value) => React.ReactElement",
+      "description": "Allows you to customize the rendering of the text.\n\nAccepts a `ReactElement` or a function that returns the element to render."
```
mj12albert, line 6: @vladmoroz If you're ok with this description I'll use the same in https://github.com/mui/base-ui/pull/990

vladmoroz: I’d say:

```
Override children if you need to customize how the value is displayed.
By default, renders the current value as is.

Accepts a `ReactElement` or a function that returns the element to render.
```

***

Also, shouldn't the children render function allow `React.ReactNode` instead of `React.ReactElement`?

PR #1069, opened 2024-12-12
[docs] API reference fixes/improvements

vladmoroz:
Fixes unnecessary empty props tables:

image

Fixes occasional stray pixels when table cells overflow

image

Increases the width of the first column in API reference tables to fit the CSS variable names:

image


mnajdova:
Looks better on desktop. Maybe on mobile we can move the type info for the css vars and attributes to the more info popover?

IMG_8935

vladmoroz:
> Looks better on desktop. Maybe on mobile we can move the type info for the css vars and attributes to the more info popover?

Reworked that area, should be much better now

PR #1077, opened 2024-12-12
[visuallyHidden] Set top left position

vladmoroz:
Prevents scrollbars from appearing when the natural position of the visually hidden element exceeds viewport width (e.g. it is in a scroll container)

https://mui-org.slack.com/archives/C02P87NQLJC/p1734038721374039

Issue at hand

Uploading Screen Recording 2024-12-12 at 22.20.42.mov…

### DIFF
--- a/packages/react/src/utils/visuallyHidden.ts
+++ b/packages/react/src/utils/visuallyHidden.ts
@@ -8,6 +8,8 @@ export const visuallyHidden: React.CSSProperties = {
   overflow: 'hidden',
   padding: 0,
   position: 'absolute',
+  top: 0,
+  left: 0,
   whiteSpace: 'nowrap',
   width: '1px',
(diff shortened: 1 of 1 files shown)


packages/react/src/utils/visuallyHidden.ts
```diff
@@ -8,6 +8,8 @@ export const visuallyHidden: React.CSSProperties = {
   overflow: 'hidden',
   padding: 0,
   position: 'absolute',
```
atomiks, line 10:
```suggestion
  position: 'fixed',
```

vladmoroz: Updated

atomiks (approved):

colmtuite (approved):

PR #1083, opened 2024-12-13
[docs] Accessibility and Quick Start overview pages

### LINKED ISSUE #1008: [docs] Overview content
Create pages for Quick Start and a11y

colmtuite:
Closes #1008


docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
```diff
@@ -0,0 +1,47 @@
...
+        </Popover.Positioner>
+      </Popover.Portal>
+    </Popover.Root>
+  );
+}
+```
+
+## Next steps
+
+This walkthrough outlines the basics of putting together a Popover component with Base UI. Check out the Popover component documentation to learn how to configure it further. For example, you might want to add the `Popover.Arrow` part, or configure some props.
```
mnajdova, line 47: Maybe we can rephrase the Next steps to point people to browser the available components instead of the Popover page. What do you think?

mnajdova (commented):
Can we avoid using "our" and "we"? We have this recommendation in the writing  style guide: https://www.notion.so/mui-org/Writing-style-guide-2a957a4168a54d47b14bae026d06a24b

docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
vladmoroz: One other thing we can add as a semi-optional step is to add `isolation: isolate` to your root layout in order for portals to layer automatically (this is the new `position: relative; z-index: 0`)

colmtuite: > point people to browser the available components

I agree this would be better, but we don't have a page listing available components, and won't have time to create one.

> using a simple component that’s familiar to everyone
> Using a simple component on the Quick start makes sense
I stuck with Popover because I don't think using a simple component like a button is a good idea. When exploring a new headless UI lib, it's not helpful or exciting to learn how to style a button. I think it's more exciting and educational to jump into a popup component. It's also useful to learn some Base UI concepts like composition and some APIs. I chose Popover because I think it's the simplest component that's interesting.

> One other thing we can add as a semi-optional step is to add isolation: isolate

I added this, not sure if I demonstrated it in the most useful way?

atomiks: > One other thing we can add as a semi-optional step is to add isolation: isolate to your root layout in order for portals to layer automatically (this is the new position: relative; z-index: 0)

Curious what this achieves? Also in the PR it was made to `<html>`, did you mean the root app div where the portals go outside?

docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
```diff
@@ -0,0 +1,65 @@
...
+        </Popover.Positioner>
+      </Popover.Portal>
+    </Popover.Root>
+  );
+}
+```
+
+## Next steps
+
+This walkthrough outlines the basics of putting together a Popover component with Base UI. Check out the [Popover component documentation](/react/components/popover) to learn how to configure it further.
```
mnajdova, line 65:
```suggestion
This walkthrough outlines the basics of putting together a Popover component with Base UI. Check out the [Popover component documentation](/react/components/popover) to learn how to configure it further, or browse the rest of the components to check similar demos of the other components available in the library.
```

I still think we should slightly extend this section so that the call to action is towards the other components as well. Otherwise, it looks like the page is all about the popover.

mnajdova (approved):
Left one final comment for consideration, the rest looks good. 

We should also update the landing page's link to point to the quick start page, now that we have it.

colmtuite:
> we should slightly extend this section so that the call to action is towards the other components
> We should also update the landing page's link to point to the quick start page

Done ✅

vladmoroz:
- We need titles on the code blocks. They look too bare without them
- We need a CSS modules version on the code block. You can't copy paste this code into your app if you don't use Tailwind (I'd argue that CSS modules should be the default too)

colmtuite:
> We need titles on the code blocks. They look too bare without them

Added.

> We need a CSS modules version on the code block.

I added the Popover hero demo instead.

docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+```jsx title="Terminal"
+npm i @base-ui-components/react
+```
+
+> All components are included in a single package. Base UI is tree-shakeable, so your app bundle will contain only the components that you actually use.
+
+## Set up portals
+
+Base UI uses portals for components that render popups, such as Dialog and Popover. To make portalled components always appear on top of the entire page, add the following style to your application layout root:
```
vladmoroz, line 18: Should we link to Dialog and Popover pages or do you think that's unnecessary?

docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+```tsx title="layout.tsx"
+<body>
+  <div className="Root">{children}</div>
+</body>
```
vladmoroz, line 23: I don't like how prettier puts children inline

We can override it ("prettier-ignore" won't show up in the docs)

```suggestion
<body>
  <div className="Root">
    {/* prettier-ignore */}
    {children}
  </div>
</body>
```

docs/src/app/(public)/page.tsx
```diff
@@ -18,7 +18,7 @@ export default function Homepage() {
         </p>
         <Link
           className="-m-1 inline-flex items-center gap-1 p-1"
-          href="/react/components/accordion"
+          href="/react/overview/quick-start"
```
vladmoroz, line 21: Could you update `not-found.tsx` too?

vladmoroz (approved):

docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
colmtuite: Popover is linked below. Distracting/unnecessary to have many links in a walkthrough.

PR #1086, opened 2024-12-13
[docs] Expose data attrs descriptions and improve table code formatter

vladmoroz:
- Expose descriptions for data attributes and CSS variables
- Tweak reference tables column widths
- Improve table code formatter
  - Handle union types when returned by a solo function:
  - Sturdier handling of nested union types

https://deploy-preview-1086--base-ui.netlify.app/

### DIFF
--- a/docs/src/components/TableCode.tsx
+++ b/docs/src/components/TableCode.tsx
@@ -22,19 +22,21 @@ export function TableCode({ children, printWidth = 40, ...props }: TableCodeProp
         return;
       }
 
-      if (getChildrenText(child).trim() === '|' && depth < 1) {
+      const str = getChildrenText(child);
+
+      if (str.trim() === '|' && depth < 1) {
         groupIndex += 1;
         unionGroups.push([]);
         return;
       }
 
-      if (getChildrenText(child).trim() === '(') {
+      str.split('(').forEach(() => {
         depth += 1;
-      }
+      });
 
-      if (getChildrenText(child).trim() === ')') {
+      str.split(')').forEach(() => {
         depth -= 1;
-      }
+      });
 
       unionGroups[groupIndex].push(child);
     });
--- a/docs/src/components/TableCode.tsx
+++ b/docs/src/components/TableCode.tsx
@@ -18,15 +18,20 @@ export function TableCode({ children, printWidth = 40, ...props }: TableCodeProp
     let groupIndex = 0;
     parts.forEach((child, index) => {
       if (index === 0) {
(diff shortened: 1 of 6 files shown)


docs/src/components/TableCode.tsx
```diff
@@ -18,23 +18,30 @@ export function TableCode({ children, printWidth = 40, ...props }: TableCodeProp
...
 
-      if (getChildrenText(child).trim() === '|' && depth < 1) {
-        groupIndex += 1;
+      const str = getChildrenText(child);
+
+      // Solo function return values shouldn't be broken up, e.g. in:
+      // `(value) => string | string[] | null | Promise`
+      if (str.includes('=>') && depth < 1) {
+        depth += 1;
+      }
```
vladmoroz, line 30: before (incorrect, shows `validate` prop as a union type that includes a function):

image


after (correct, `validate` prop is a function that returns a union type):

image

docs/src/components/TableCode.tsx
```diff
@@ -18,23 +18,30 @@ export function TableCode({ children, printWidth = 40, ...props }: TableCodeProp
...
+      str.split('(').forEach(() => {
         depth += 1;
-      }
+      });
 
-      if (getChildrenText(child).trim() === ')') {
+      str.split(')').forEach(() => {
         depth -= 1;
-      }
+      });
```
vladmoroz, line 44: before (incorrect, `anchor` function return value members are on the same level as the function itself)

image


after (correct, `anchor` is a union type, and one of its members is a function that returns a union type):

image

colmtuite (approved):

mnajdova:
@vladmoroz CSS variables look great, however data attributes (at least some of them) have types. Should we add this info available in a tooltip?

vladmoroz:
> @vladmoroz CSS variables look great, however data attributes (at least some of them) have types

Have tried that, honestly feels too weird when it's just a type in the data attributes as almost all of them are empty strings (and when not, they are really simple when you look at them in the DOM and/or match the corresponding props).

I think I'd just go through the notable data attributes that are not empty and mention whether values match a certain prop.

Let me know if this sounds good to you

mnajdova:
> I think I'd just go through the notable data attributes that are not empty and mention whether values match a certain prop.
> 
> Let me know if this sounds good to you

I am ok with including this info in the description, or maybe explaining them a bit better 👌. I think we have either boolean or union, and we can likely easily add in the desiption something like: These are the available values: ... or something.

PR #1090, opened 2024-12-13
[docs] Custom props order

vladmoroz:
After having worked in the docs for the past couple weeks, I realised that a custom semi-arbitrary order will feel much less arbitrary than alphabetic.

I'm ordering most related props together, and placing the essential props at the top (like `defaultValue, value, onValueChange`)


docs/reference/order.json
vladmoroz, line 1: This file defines the order of things

colmtuite (approved):

PR #1093, opened 2024-12-15
[Checkbox] Handle the `value` prop correctly

michaldudak:
The button rendered by the checkbox used to have `value=off` regardless of its state. This PR removes the value from the button and propagates the `value` prop from `Checkbox.Root` to the underlying input.


packages/react/src/checkbox/root/useCheckboxRoot.ts
```diff
@@ -80,7 +81,6 @@ export function useCheckboxRoot(params: UseCheckboxRoot.Parameters): UseCheckbox
     (externalProps = {}) =>
       mergeReactProps<'button'>(getValidationProps(externalProps), {
         ref: buttonRef,
-        value: 'off',
```
atomiks, line 83: I thought Switch ahd this at some point as well, but can't check the history easily

michaldudak: It could make some sense if the value on the `button` was placed conditionally (only when unchecked) and also the `name` was set to the same value as the input. This way we could avoid rendering the hidden input conditionally, I suppose.

docs/reference/generated/checkbox-root.json
```diff
@@ -26,6 +26,10 @@
       "default": "false",
       "description": "Whether the checkbox is in a mixed state: neither ticked, nor unticked."
     },
+    "value": {
+      "type": "number | string",
+      "description": "The value of the selected checkbox."
+    },
```
atomiks, line 32: `CheckboxGroup` uses `name` to link, but maybe it should now be `value` (which is like `RadioGroup`)

michaldudak: From my experience it's not very common to set values on checkboxes (at least when posting to a server) - it's the name what's important and the value is just the default `on`/`off`. With radio buttons, there is no other way than using values.
I'd personally leave it as is.

One thing that we can consider, though, is the customization of the unchecked value. Currently, it's hardcoded to `off`. I never needed to customize it myself, but maybe there is a use case. We can wait for community feedback on it.

atomiks (approved):

PR #1094, opened 2024-12-15
[Menu, Select] Do not ignore the Trigger's `label` prop

### LINKED ISSUE #1072: [Menu, Select] Trigger `label` prop appears to be unused
Not sure what the original purpose of the prop was, but it's not used in any way. Should we remove it?

https://github.com/mui/base-ui/blob/master/packages/react/src/select/trigger/SelectTrigger.tsx
https://github.com/mui/base-ui/blob/master/packages/react/src/menu/trigger/MenuTrigger.tsx

**Search keywords**:

michaldudak:
Fixes #1072


atomiks (approved):
`title`?

michaldudak:
`label`, of course.

PR #1097, opened 2024-12-15
[docs] Account for the scroll lock in quick nav

vladmoroz:
Fixes this issue when the quick nav positioning math wasn't aware of scroll locking

https://github.com/user-attachments/assets/e10c779f-8f08-41dc-8ca4-5f845badb2b0

### DIFF
--- a/docs/src/components/quick-nav/QuickNav.tsx
+++ b/docs/src/components/quick-nav/QuickNav.tsx
@@ -26,7 +26,7 @@ function onMounted(ref: React.RefObject<HTMLDivElement | null>) {
 
   let top: number;
   let bottom: number;
-  let prevScrollY = window.scrollY;
+  let prevScrollY = getScrollY();
   let resizeObserver: ResizeObserver | undefined;
   let state: 'Scrollable' | 'StickyTop' | 'StickyBottom' = 'StickyTop';
   let raf = 0;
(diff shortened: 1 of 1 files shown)


docs/src/components/quick-nav/QuickNav.tsx
```diff
@@ -250,6 +250,11 @@ function onMounted(ref: React.RefObject<HTMLDivElement | null>) {
   };
 }
 
+/** Get window.scrollY accounting for a potential scroll lock offset top style */
+function getScrollY() {
+  return window.scrollY - parseFloat(document.documentElement.style.top || '0');
+}
```
atomiks, line 256: This scroll locking technique is looking to cause a bunch of interop issues with positioning math, such as in custom extensions or code on users' websites. I guess we'll need to see how large the impact is, but it may not be the bulletproof solution we're looking for compared to a simple `overflow: hidden` (at least on non-iOS browsers...)

atomiks (approved):

docs/src/components/quick-nav/QuickNav.tsx
vladmoroz: Let's discuss it today with https://github.com/mui/base-ui/pull/1098

PR #1098, opened 2024-12-15
[docs] Mobile nav fixes

vladmoroz:
- Wait for the URL to change before closing the mobile nav dialog. Otherwise, there might be a slight flash of content during the close animation as the navigation happens. This guarantees that closing the nav dialog smoothly “reveals” the new page.
- If navigating to the same page, close the nav and scroll up smoothly.
- Fix the flick down gesture animation after https://github.com/mui/base-ui/pull/1109 (we relied on the scroll lock to be on during the flick down)

### DIFF
--- a/docs/src/components/CodeBlock.tsx
+++ b/docs/src/components/CodeBlock.tsx
@@ -79,6 +79,7 @@ export function Pre({ className, ...props }: React.ComponentProps<'pre'>) {
       }}
     >
       <ScrollArea.Viewport
+        style={{ overflow: undefined }}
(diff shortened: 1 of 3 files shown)


michaldudak:
Fixed in #1109

michaldudak:
> Wait for the URL to change before closing the mobile nav dialog.

On a slow network, this doesn't give any feedback that the link was clicked and something is actually loading

vladmoroz:
> On a slow network, this doesn't give any feedback that the link was clicked and something is actually loading

I can add a spinner, but mind if I do it during the week after the release? Want to prioritise the demos

michaldudak (approved):

PR #1102, opened 2024-12-16
[menu] Remove the `keepMounted` prop from the Positioner

michaldudak:
As the Portal part has the `keepMounted` prop, it's no longer necessary to have it in Positioner.


atomiks (commented):
This should apply to all the Positioners, but considering you don't need to use `Portal` necessarily, I decided to keep the prop since it still has a purpose

PR #1112, opened 2024-12-16
 [docs] Custom code block scrollbars and misc fixes

vladmoroz:
- Custom code block scrollbars
- Add aria label to the reference table popover triggers
- Fix accordion icon demo shrink


docs/src/blocks/Demo/DemoSourceBrowser.tsx
vladmoroz, line 1: cc @michaldudak
was super annoying we didn't have direct access to the `<pre>` element like in the code block

colmtuite (approved):

PR #1113, opened 2024-12-16
[core] Use stable React packages

michaldudak:
Bumped React version to stable and Next.js to the latest version.


mnajdova (commented):
We should also update the peer dependencies, we can remove ^19.0.0-rc, no?

mnajdova (approved):

PR #1115, opened 2024-12-16
[docs] Automatic nice typography

vladmoroz:
Automatically use curly quotes, em dashes, etc.

### DIFF
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -7,7 +7,7 @@ import withDocsInfra from '@mui/monorepo/docs/nextConfigDocsInfra.js';
 import nextMdx from '@next/mdx';
 import rehypeExtractToc from '@stefanprobst/rehype-extract-toc';
 import remarkGfm from 'remark-gfm';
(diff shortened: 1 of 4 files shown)


mnajdova (commented):
What problem does this package solve? Can you share some examples of things that changed? I would avoid adding the dependency, mostly because it's a v0 and it has around 130 weekly downloads.

vladmoroz:
> What problem does this package solve? Can you share some examples of things that changed? I would avoid adding the dependency, mostly because it's a v0 and it has around 130 weekly downloads.

It converts straight quotes and hyphens into their proper typographic equivalent.

Here's an example - "this is a quote"
→
Here’s an example—“this is a quote”

We don't need to update/maintain it, it works well from what I've tested. If we ever need to remove it, nothing depends on it either.

This is the source code
https://github.com/codsen/codsen/blob/main/packages/remark-typography/src/main.ts

It's from codsen.com which maintain quite a few other popular packages

mnajdova (approved):
Ok, let's try it.

PR #1118, opened 2024-12-16
[docs] Improve API reference copy

samuelsycamore:
"place its parts the following way" struck me as awkward phrasing—how about "assemble ~it as shown below~ its parts"?

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/accordion/page.mdx
+++ b/docs/src/app/(public)/(content)/react/components/accordion/page.mdx
@@ -10,7 +10,7 @@
 
 ## API reference
 
-Import the component and place its parts the following way:
+Import the component and assemble it as shown below:
 
 ```jsx title="Anatomy"
 import { Accordion } from '@base-ui-components/react/accordion';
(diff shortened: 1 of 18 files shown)


vladmoroz:
Actually crossed my mind the other day that this is more specific than it needs to be. Can we say something like "Import the component and put it together"? What I'm getting at is that in some cases, the order/composition matters a lot, and in some there is flexibility to it, so I don't want this to sound too prescriptive.

mapache-salvaje:
Ah that's interesting @vladmoroz. Maybe something like "After importing the component, you can render the following DOM nodes (or subcomponents, or elements, have you settled on a common term for these pieces?):"? This would convey the idea that you have options rather than a strict prescription of what must be included.

vladmoroz:
Hmm I think we need to keep it a bit ambiguous so it works regardless whether there is some or absolutely no flexibility to the anatomy. "You can render the following DOM nodes..." sounds like there always are options.

How about `Import the component and assemble its parts`? (To your question, those pieces are called "parts")

mapache-salvaje:
Makes sense, I'll run with that 🫡

vladmoroz (approved):

PR #1120, opened 2024-12-16
[docs] Add menu demos

vladmoroz:
- Add code block line highlights
- Add menu checkbox items demo
  - Note: check icon isn't unmounted yet, I'll create an issue
- Add menu open on hover demo
- Add menu with submenu demo

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/menu/demos/checkbox-items/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/menu/demos/checkbox-items/css-modules/index.module.css
@@ -1,0 +1,166 @@
+.Button {
+  box-sizing: border-box;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 0.375rem;
+  height: 2.5rem;
+  padding: 0 0.875rem;
+  margin: 0;
+  outline: 0;
+  border: 1px solid var(--color-gray-200);
+  border-radius: 0.375rem;
+  background-color: var(--color-gray-50);
+  font-family: inherit;
+  font-size: 1rem;
+  font-weight: 500;
+  line-height: 1.5rem;
+  color: var(--color-gray-900);
+  user-select: none;
+
+  @media (hover: hover) {
+    &:hover {
+      background-color: var(--color-gray-100);
+    }
+  }
+
+  &:active {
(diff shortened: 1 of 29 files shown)


colmtuite:
@vladmoroz Checkbox items demo is not working. Why is the nested menu demo split into two separate code blocks?

vladmoroz:
> Checkbox items demo is not working

https://github.com/mui/base-ui/issues/1121

vladmoroz:
> Why is the nested menu demo split into two separate code blocks?

I think it's important to show a high-level overview of the submenu anatomy first so it's easy to understand how it's composed before looking at the detailed demo

Also have just updated the example demos to use a more compact view—how do you feel about this altogether?
https://deploy-preview-1120--base-ui.netlify.app/react/components/menu

mnajdova (changes requested):
When choosing a menu item from the hover example, it scrolls to the top of the page.
Checkbox item demo is still not working (probably we just need to rebase with master, as https://github.com/mui/base-ui/issues/1121 is fixed)

michaldudak:
The checkboxitems demo should have the positioner/portal always mounted so the checked state is remembered when the menu is closed.

vladmoroz:
> The checkboxitems demo should have the positioner/portal always mounted so the checked state is remembered when the menu is closed.

Hmm probably a more realistic example would be to show controlled demos instead of keeping stuff mounted

atomiks:
The submenu demo right at the bottom feels awkward because it's near the bottom of the screen, feels like there should be more breathing room below it

vladmoroz:
> The submenu demo right at the bottom feels awkward because it's near the bottom of the screen, feels like there should be more breathing room below it

Agree it's awkward

I think it just needs a logical way to the next page, like next/prev component links.

Adding more space makes it more awkward too.

vladmoroz:
Updates:
- Added radio items
- Made radio and checkbox items controlled
- Added a menu item + link example
- Added a `closeOnClick` example
- Added an example with group labels

@mnajdova the bugs are gone after merging with master

mnajdova (approved):
Demos look good. I noticed an issue with the table of content (it jumps when you open a menu). We can fixed it separately


https://github.com/user-attachments/assets/c0e48065-7e30-498d-bb5b-88a7ea03c018

PR #1126, opened 2024-12-17
[select] Fix instant close with no set value

### LINKED ISSUE #1125: [Select] Closes instantly when no `value`/`defaultValue` is set

atomiks:
Closes #1125

### DIFF
--- a/packages/react/src/select/popup/SelectPopup.tsx
+++ b/packages/react/src/select/popup/SelectPopup.tsx
@@ -14,7 +14,6 @@ import type { TransitionStatus } from '../../utils/useTransitionStatus';
 import { useSelectPositionerContext } from '../positioner/SelectPositionerContext';
 import { mergeReactProps } from '../../utils/mergeReactProps';
 import { transitionStatusMapping } from '../../utils/styleHookMapping';
-import { useSelectIndexContext } from '../root/SelectIndexContext';
 
 const customStyleHookMapping: CustomStyleHookMapping<SelectPopup.State> = {
   ...popupStateMapping,
@@ -35,7 +34,6 @@ const SelectPopup = React.forwardRef(function SelectPopup(
 
   const { id, open, popupRef, transitionStatus, alignItemToTrigger, mounted, modal } =
     useSelectRootContext();
-  const { selectedIndex } = useSelectIndexContext();
 
   const positioner = useSelectPositionerContext();
 
@@ -90,8 +88,6 @@ const SelectPopup = React.forwardRef(function SelectPopup(
         modal={false}
         disabled={!mounted}
         visuallyHiddenDismiss={modal ? 'Dismiss popup' : undefined}
-        initialFocus={selectedIndex === null ? popupRef : undefined}
-        order={['content', 'reference']}
       >
         {renderElement()}
       </FloatingFocusManager>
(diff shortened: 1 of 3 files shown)


packages/react/src/select/popup/SelectPopup.tsx
```diff
@@ -88,6 +90,7 @@ const SelectPopup = React.forwardRef(function SelectPopup(
         modal={false}
         disabled={!mounted}
         visuallyHiddenDismiss={modal ? 'Dismiss popup' : undefined}
+        initialFocus={selectedIndex === null ? popupRef : undefined}
         order={['content', 'reference']}
       >
```
atomiks, line 95: Removing `order` also fixes it, which was added with modality changes, but not sure if it's truly necessary here @michaldudak

packages/react/src/select/root/SelectRoot.test.tsx
```diff
@@ -315,10 +315,6 @@ describe('<Select.Root />', () => {
       await user.click(trigger);
       expect(handleOpenChange.callCount).to.equal(1);
       expect(handleOpenChange.args[0][0]).to.equal(true);
-
-      await user.click(trigger);
-      expect(handleOpenChange.callCount).to.equal(2);
-      expect(handleOpenChange.args[1][0]).to.equal(false);
```
atomiks, line 321: I don't think this test is valid, because the trigger doesn't need to be the click target to dismiss it, any outside area works fine

mnajdova:
What exactlly is the issue. I try reproducing it here but I can't: https://codesandbox.io/p/sandbox/laughing-shamir-k8hmn8?file=%2Fsrc%2FApp.tsx%3A11%2C32 (it's version from another random PR package).

packages/react/src/select/root/SelectRoot.test.tsx
atomiks: For touch it does require `outsidePressEvent: 'mousedown'`, which matches `Menu`

atomiks:
@mnajdova the trigger needs to be moved down so `alignItemToTrigger` is activated, with the popup on the top of the trigger

`position: relative, top: 100` inline styles on the trigger reveal it

packages/react/src/select/root/SelectRoot.test.tsx
mnajdova: > I don't think this test is valid, because the trigger doesn't need to be the click target to dismiss it, any outside area works fine

It's not that is not valid, but both clicking the trigger or any outside area should close it. Maybe we can actually test both of these two scenarios.

packages/react/src/select/popup/SelectPopup.tsx
```diff
@@ -88,7 +88,6 @@ const SelectPopup = React.forwardRef(function SelectPopup(
         modal={false}
         disabled={!mounted}
         visuallyHiddenDismiss={modal ? 'Dismiss popup' : undefined}
-        order={['content', 'reference']}
```
mnajdova, line 91: 👍

mnajdova (approved):
Looks good. I would add test cases for both scenarios I mentioned in https://github.com/mui/base-ui/pull/1126/files#r1888154490, just to be safe.

packages/react/src/select/root/SelectRoot.test.tsx
atomiks: This particular test fails due to `pointer-events: none;` style, so you can't click the trigger, only the area above it (which is `inert`, similar to a backdrop being present). The end result is the same though

PR #1131, opened 2024-12-17
[docs] Fix CSS vars generation

mj12albert:
Fixed: https://deploy-preview-1131--base-ui.netlify.app/react/components/accordion#panel

### DIFF
--- a/scripts/buildApiDocs/buildReference.ts
+++ b/scripts/buildApiDocs/buildReference.ts
@@ -125,8 +125,16 @@ export async function buildReference(grep: RegExp | null = null) {
   rmSync(TEMP_DIR, { recursive: true });
 }
 
+function isObjectEmpty(object: Record<any, any>) {
+  // eslint-disable-next-line
+  for (const _ in object) {
+    return false;
+  }
+  return true;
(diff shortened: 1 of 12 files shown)


vladmoroz (approved):

scripts/buildApiDocs/buildReference.ts
```diff
@@ -125,7 +125,19 @@ export async function buildReference(grep: RegExp | null = null) {
...
+function isObjectEmpty(object: Record<any, any>) {
+  // eslint-disable-next-line
+  for (const _ in object) {
+    return false;
+  }
+  return true;
+}
+
 function sortObjectByKeys<T>(obj: Record<string, T>, order: string[]): Record<string, T> {
+  if (isObjectEmpty(obj) || order.length === 0) {
```
mnajdova, line 137: My fix was to update the order.json to:
```
  "cssVariables": [
    "__EVERYTHING_ELSE__"
  ],
```
with the thought that we will update this array later. This works too, but we probably wouldn't even use this function if the order array is empty (it means this data does not need to be sorted).

PR #1133, opened 2024-12-17
[useScrollLock] Handle Safari pinch-zooming

### LINKED ISSUE #1130: [useScrollLock] Glitch in Safari when opening while pinch-zoomed
Started happening after 83b738d26f9f909def971e5542cb145fc25a6a22

https://github.com/user-attachments/assets/06b5f745-2143-4b1f-b59b-8106a68e3d6b

atomiks:
Partially addresses #1130. There are still some details to work out regarding small jumps

### DIFF
--- a/packages/react/src/utils/useScrollLock.ts
+++ b/packages/react/src/utils/useScrollLock.ts
@@ -7,6 +7,15 @@ let originalBodyStyles = {};
 let preventScrollCount = 0;
 let restore: () => void = () => {};
 
+function getVisualOffsets(doc: Document) {
+  const win = ownerWindow(doc);
+  const vV = win.visualViewport;
+  return {
+    offsetLeft: vV?.offsetLeft || 0,
+    offsetTop: vV?.offsetTop || 0,
+  };
+}
+
 function preventScrollIOS(referenceElement?: Element | null) {
   const doc = ownerDocument(referenceElement);
   const html = doc.documentElement;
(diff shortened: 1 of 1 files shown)


michaldudak:
It's much better now, but there's still a slight movement ~~to the bottom-right~~ (EDIT: actually it's quite random) when the menu popup opens while zoomed in. IMO we can live with it, but let's see if it's easily solvable.

atomiks:
@vladmoroz Safari doesn't like the side nav when pinch zoomed

https://github.com/user-attachments/assets/cbcb6980-0c5d-4e0d-a1ca-f856ea55d2a7

atomiks:
@michaldudak yeah it's probably floating point errors... I think `scrollY` returning an integer is a problem, but not sure if that's solvable

packages/react/src/utils/useScrollLock.ts
```diff
@@ -94,11 +94,17 @@ function preventScrollStandard(referenceElement?: Element | null) {
     // Handle `scrollbar-gutter` in Chrome when there is no scrollable content.
     const hasScrollbarGutterStable = htmlComputedStyles.scrollbarGutter?.includes('stable');
 
+    // Safari needs visual viewport offsets added to account for pinch-zoom
+    const webkit = isWebKit();
+    const vV = doc.defaultView?.visualViewport;
```
michaldudak, line 99: I'm curious: is `defaultView` different to `window`?

atomiks: It's the owning document's window

michaldudak: :+1:
We have a utility to return this - `ownerWindow` in `utils/owner`

michaldudak (approved):

atomiks:
There's still some slight jumps, particularly when changing the zoom after opening. But it's much better than before and not entirely broken/inaccessible, I'll mark it as partially addressed.

PR #1143, opened 2024-12-17
[useScrollLock] Patch firefox

atomiks:
@vladmoroz ⚠️  `<html>` must have `overflow-y: visible`, we can do it here or in the docs


mnajdova (commented):
~I think we can just add this in the docs to avoid last minute library changes. @vladmoroz what do you think?~ Actually, we may as well just merge this, it only affects Firefox, we can fix it later if we find some more issues.

vladmoroz (approved):

PR #1144, opened 2024-12-17
[Dialog, AlertDialog] Add the `data-has-nested-dialogs` hook

michaldudak:
Adds the style hook when the Dialog/AlertDialog has open subdialogs (of the same type)


mnajdova (commented):
We should update the DataAttributes files and run docs:api

vladmoroz (approved):

PR #1146, opened 2024-12-17
[docs] Nested dialog demo

vladmoroz:
- Nested dialog demo
- Remove quick nav pointer events auto

https://deploy-preview-1146--base-ui.netlify.app/react/components/dialog


colmtuite (approved):

michaldudak:
IMO nested menus would look better if they were slightly more apart. The double border between them doesn't look pretty.

image

vladmoroz:
Good spot @michaldudak 

Probably will do this though

image

michaldudak:
Looks great as well!

michaldudak:
Something is broken on iOS - when the nested trigger is clicked the whole menu closes immediately.

vladmoroz:
Yeah I saw it @michaldudak—don't think it's at the example level

PR #1155, opened 2024-12-17
[menu] Do not close the menu when touching the subtrigger on iOS

michaldudak:
This is a quick workaround. We should look for the root cause.


atomiks (approved):

atomiks:
I can't reproduce this on master with a real iOS device or Simulator (iOS 18), tapping the submenu trigger opens the submenu as expected

colmtuite:
I can't reproduce this either. Tapping the submenu trigger opens the submenu.

michaldudak:
Yeah, I don't see it now either. Perhaps it was fixed otherwise in the meantime.

PR #1159, opened 2024-12-18
[useScrollLock] New implementation

### LINKED ISSUE #1156: [useScrollLock] Address Firefox and iOS issues
### Firefox

`position: sticky` elements disappear due to a layout bug in Firefox when `position: fixed` ancestors are offset (which is used to maintain the scroll position). We patched it with a quick alternative solution in #1143, but this isn't as robust as the one used for Chrome/Safari due to `position: fixed` elements being allowed to shift. Vlad's early demo doesn't have some of these issues, let's try to work off it: https://codesandbox.io/p/sandbox/remove-scroll-poc-forked-pqfyhf?file=%2Fsrc%2FApp.tsx%3A61%2C2

### iOS

iOS' current implementation randomly causes scroll containers to flicker when the lock is disabled. Look into an alternative solution for iOS so this doesn't happen

atomiks:
Closes #1156

A new custom implementation is used across non-iOS browsers, handling the previous edge cases on Firefox as well, removing the patch we gave it.

### iOS

As far as I can tell, scroll containers are optimized in iOS similar to the `will-change` property (adding it causes the flicker everywhere), possibly due to some compositing step in the browser when animations are occurring when scrolled down(?). The new implementation on iOS also flickered, but even worse and had other issues.

For iOS, there seems to be 3 options:

1. Accept that iOS has this flicker issue and has workarounds (like we did on the docs for the main nav). We could make this easier, maybe?
2. Use a scroll lock dependency just for iOS, or browsers without inset scrollbars.
3. Accept that iOS can't lock scroll at all without causing problems

I personally feel like number 2 is the best option, so I've done this in the PR. `@react-aria/overlays`' `usePreventScroll` hook seems like the best one and handles an enormous number of edge cases, and is kept up to date. It also seems to address the macOS "scrollbar popping" issue in most cases

@vladmoroz the side nav seems to need updating with the new non-iOS implementation?

### DIFF
--- a/packages/react/src/utils/useScrollLock.ts
+++ b/packages/react/src/utils/useScrollLock.ts
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { usePreventScroll } from '@react-aria/overlays';
-import { isIOS } from './detectBrowser';
+import { isFirefox, isIOS } from './detectBrowser';
 import { ownerDocument, ownerWindow } from './owner';
 import { useEnhancedEffect } from './useEnhancedEffect';
 
@@ -15,6 +15,15 @@ function supportsDvh() {
     typeof CSS.supports === 'function' &&
     CSS.supports('height', '1dvh')
   );
+}
+
+function hasInsetScrollbars(referenceElement?: Element | null) {
+  if (typeof document === 'undefined') {
+    return false;
+  }
+  const doc = ownerDocument(referenceElement);
+  const win = ownerWindow(doc);
+  return win.innerWidth - doc.documentElement.clientWidth > 0;
 }
 
 function preventScrollStandard(referenceElement?: Element | null) {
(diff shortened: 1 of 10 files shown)


packages/react/src/utils/useScrollLock.ts
```diff
@@ -63,117 +63,70 @@ function preventScrollIOS(referenceElement?: Element | null) {
...
-      Object.assign(bodyStyle, {
-        overflow: '',
-        [paddingProp]: '',
-      });
-      Object.assign(htmlStyle, {
-        overflow: '',
-      });
-      return;
-    }
+    const heightProperty = CSS.supports('height', '1dvh') ? 'dvh' : 'vh';
```
atomiks, line 114: Old Android browsers may be an issue here since `dvh` is pretty new. We could probably just get away with using `overflow: hidden` since inset scrollbars presumably aren't an issue there?

packages/react/src/utils/useScrollLock.ts
```diff
@@ -8,172 +9,87 @@ let originalBodyStyles = {};
...
+    scrollTop = html.scrollTop;
 
     originalHtmlStyles = {
-      position: htmlStyle.position,
-      top: htmlStyle.top,
-      left: htmlStyle.left,
-      right: htmlStyle.right,
-      overflowX: htmlStyle.overflowX,
-      overflowY: htmlStyle.overflowY,
+      overflowY: html.style.overflowY,
```
vladmoroz, line 37: We aren't cleaning up overflowX

packages/react/src/utils/useScrollLock.ts
```diff
@@ -8,172 +9,80 @@ let originalBodyStyles = {};
...
-        [paddingProp]: '',
-      });
-      Object.assign(htmlStyle, {
-        overflow: '',
-      });
-      return;
-    }
+    Object.assign(body.style, {
+      position: 'relative',
+      height: `calc(100dvh - ${bodyVerticalMargins}px)`,
```
vladmoroz, line 72: A total turbo nit, but since in the vast majority of websites the margins will be reset to 0, can we check for that and if yes, set height to `100dvh`? Just so it's not `calc(100dvh - 0px)` when inspecting

packages/react/src/utils/useScrollLock.ts
```diff
@@ -8,172 +9,104 @@ let originalBodyStyles = {};
...
+  let scrollTop = 0;
+  let scrollLeft = 0;
+  let resizeRaf = -1;
 
-      return;
-    }
+  // Pinch-zoom in Safari causes a shift. Just don't lock scroll if there's any pinch-zoom.
+  if (isWebKit() && (win.visualViewport?.scale ?? 1) !== 1) {
+    return () => {};
+  }
```
vladmoroz, line 42: Should we add a resize handler on the visual viewport to catch when they unzoom?

atomiks: They can still scroll the whole page while pinch-zoomed, so it doesn't really fix the root issue

PR #1161, opened 2024-12-18
[Menu, Select, Dialog] Use internal backdrop for pointer modality

### LINKED ISSUE #1107: [popups] Clicking outside of nested popups closes all of them
Whenever there are nested popups open, such as a Menu or a Select in a Dialog, clicking outside of them closes all of them. The correct behavior would be to close just the topmost one (as nested Dialogs currently behave).

### LINKED ISSUE #1119: [menu] Highlight is slow
I can feel a delay when moving through the items with the mouse:

https://github.com/user-attachments/assets/67707aa4-9d4f-408f-95f6-9360e0f14f8d

Radix:

https://github.com/user-attachments/assets/5d14aeaf-cd23-4f18-8352-0d763b3dcfb1

atomiks:
Closes #1119
Also looks to close #1107 [demo](https://deploy-preview-1161--base-ui.netlify.app/experiments/popups-in-popups)

1. `pointer-events: none` on "outside" elements causes perf issues in Firefox and Safari, degrading increasingly worse as the page size increases.

2. `pointer-events: none` with an `auto` "hole" on the popup isn't feasible due to interop issues with other popups (extensions, other libraries or custom elements) as they will be unusable since they don't know to specify `auto`.

The main issue with an internal backdrop is `z-index` layering. With [Set up portals](https://base-ui.com/react/overview/quick-start) in our docs, this issue is essentially gone as the portaled backdrop will always cover the isolated root content.


mnajdova (commented):
The direction makes sense to me. Just to make sure, will we still have the *Dialog exported components so people can style their backdrops?

atomiks:
> will we still have the *Dialog exported components so people can style their backdrops?

The `Backdrop` parts are meant to be visual-only for the purposes of dimming the background and such, so `Dialog.Backdrop` etc still exist

mnajdova:
> are meant to be visual-only for the purposes of dimming the background and such, so Dialog.Backdrop etc still exist

Makes total sense, this is what I wanted to confirm 👌

packages/react/src/dialog/root/DialogRoot.test.tsx
```diff
@@ -388,4 +390,186 @@ describe('<Dialog.Root />', () => {
 
     expect(notifyTransitionEnd.callCount).to.equal(1);
   });
+
+  describe('prop: modal', () => {
+    it('should render an internal backdrop when `true`', async () => {
```
mnajdova, line 395: These two tests are identical.

atomiks: Thanks for spotting, you're right. Seems like the Menu and Select ones are different but this one wasn't changed.

mnajdova (approved):

PR #1166, opened 2024-12-18
[dialog] Fix nesting different kinds of dialogs

### LINKED ISSUE #1165: [dialog] Nested alert dialogs aren't recognised as nested dialogs and vice versa
`[data-has-nested-dialogs]` attribute isn't applied when a dialog composes an alert dialog:
https://codesandbox.io/p/sandbox/7yw3n3

When fixed, please update the confirmation dialog demo (https://github.com/mui/base-ui/pull/1164) to use an alert dialog within.

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1165

### DIFF
--- a/packages/react/src/alert-dialog/root/AlertDialogRoot.tsx
+++ b/packages/react/src/alert-dialog/root/AlertDialogRoot.tsx
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import type { DialogRoot } from '../../dialog/root/DialogRoot';
+import { DialogRootContext } from '../../dialog/root/DialogRootContext';
 import { AlertDialogRootContext } from './AlertDialogRootContext';
 import { useDialogRoot } from '../../dialog/root/useDialogRoot';
 import { PortalContext } from '../../portal/PortalContext';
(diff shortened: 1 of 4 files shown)


packages/react/src/alert-dialog/root/AlertDialogRoot.tsx
```diff
@@ -15,19 +16,22 @@ import { PortalContext } from '../../portal/PortalContext';
...
 
   const dialogRoot = useDialogRoot({
     open,
     defaultOpen,
     onOpenChange,
     modal: true,
     dismissible: false,
-    onNestedDialogClose: parentDialogRootContext?.onNestedDialogClose,
-    onNestedDialogOpen: parentDialogRootContext?.onNestedDialogOpen,
+    onNestedDialogClose: parentContext?.onNestedDialogClose,
```
flaviendelangle, line 30: Wouldn't it be cleaner to have a shared context for the element that are shared across the various dialog types?
Or are we confident that we won't have another type of dialog (not sure if a drawer falls in the category) that we will have to remember to add here?

mnajdova: Yes, I started working on this coincidently, let me open a PR. @mj12albert please assign an issue to you if you plan to work on it to avoid duplicated work.

mnajdova: Alternative PR: https://github.com/mui/base-ui/pull/1167

PR #1167, opened 2024-12-18
[Dialog, AlertDialog] Fix the nesting of different dialogs

mnajdova:
Fixes https://github.com/mui/base-ui/issues/1165

The idea is to use common context between all dialog components. I extracted only the common things inside this context.

### DIFF
--- a/packages/react/src/alert-dialog/root/AlertDialogRootContext.ts
+++ b/packages/react/src/alert-dialog/root/AlertDialogRootContext.ts
@@ -1,29 +1,19 @@
 'use client';
 import * as React from 'react';
-import { type useDialogRoot } from '../../dialog/root/useDialogRoot';
+import { DialogContext } from '../../dialog/utils/DialogContext';
 
-export interface AlertDialogRootContext extends useDialogRoot.ReturnValue {
-  /**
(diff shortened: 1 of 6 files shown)


packages/react/src/alert-dialog/root/AlertDialogRootContext.ts
```diff
@@ -1,27 +1,14 @@
...
-}
+export { DialogContext as AlertDialogRootContext };
 
 export function useAlertDialogRootContext() {
-  const context = React.useContext(AlertDialogRootContext);
+  const context = React.useContext(DialogContext);
   if (context === undefined) {
     throw new Error(
-      'Base UI: AlertDialogRootContext is missing. AlertDialog parts must be placed within <AlertDialog.Root>.',
+      'Base UI: AlertDialogRootContext is missing. Alert dialogs parts must be placed within <AlertDialog.Root>.',
```
mnajdova, line 11:
```suggestion
      'Base UI: AlertDialogRootContext is missing. AlertDialog parts must be placed within <AlertDialog.Root>.',
```

atomiks (approved):

PR #1170, opened 2024-12-18
[docs] Skip unnecessary position changes when quick nav height is similar to viewport

vladmoroz:
Prevent quick nav from adjusting the position if its height is similar to viewport height and the scroll difference would be unimportant

Avoids unnecessary movement

https://github.com/user-attachments/assets/ef9a5107-82db-4b9b-958d-c9102db06970


colmtuite (approved):

colmtuite:
👌 👨‍🍳

PR #1171, opened 2024-12-18
[docs] Quick nav actually first

vladmoroz:
Put quick nav before the h1 so that clicking anything in the main page contents sets initial focus after the quick nav

(MDN does a similar thing)

### DIFF
--- a/docs/src/components/QuickNav/QuickNav.tsx
+++ b/docs/src/components/QuickNav/QuickNav.tsx
@@ -61,7 +61,7 @@ function onMounted(ref: React.RefObject<HTMLDivElement | null>) {
 
     ref.current.style.top = '0px';
     ref.current.style.bottom = '';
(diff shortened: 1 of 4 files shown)


colmtuite (approved):

docs/src/components/QuickNav/QuickNav.tsx
```diff
@@ -252,8 +252,8 @@ function onMounted(ref: React.RefObject<HTMLDivElement | null>) {
   };
 }
 
-export function Title({ className, ...props }: React.ComponentProps<'h2'>) {
-  return <div aria-hidden className={clsx('QuickNavTitle', className)} {...props} />;
```
vladmoroz, line 256: Not aria-hidden anymore since this element is announced as almost the first thing when entering the page contents  (it matches the page title; MDN uses an h2 there but I disagree with it)

Issue #1106, opened 2024-12-16
[useButton] Remove `event.defaultMuiPrevented` and `MuiCancellableEvent`


vladmoroz:
There is a couple references to `event.defaultMuiPrevented` in the codebase. Should these be using `event.defaultPrevented` like others?

mj12albert:
This was the old way to prevent our (internal) handlers from running without affecting custom event handlers, the new way is with a `event.preventBaseUIHandler` method: https://github.com/mui/base-ui/blob/master/packages/react/src/utils/mergeReactProps.ts#L6-L7

I see the old references are all in useButton, some of that code may be unused now and could be cleaned up a bit

mj12albert:
Turns out `MenuItem` was the last thing depending on this: https://github.com/mui/base-ui/pull/542/files#r1721346832

PR #1177, opened 2024-12-19
[useButton] Remove `defaultMuiPrevented` and `MuiCancellableEvent`

mj12albert:
Closes https://github.com/mui/base-ui/issues/1106


packages/react/src/menu/item/useMenuItem.ts
```diff
@@ -34,9 +33,9 @@ export function useMenuItem(params: useMenuItem.Parameters): useMenuItem.ReturnV
           id,
           role: 'menuitem',
           tabIndex: highlighted ? 0 : -1,
-          onKeyUp: (event: React.KeyboardEvent & MuiCancellableEvent) => {
+          onKeyUp: (event: BaseUIEvent<React.KeyboardEvent>) => {
             if (event.key === ' ' && typingRef.current) {
-              event.defaultMuiPrevented = true;
+              event.preventBaseUIHandler();
```
vladmoroz, line 38: cc @colmtuite you might want to mull over what the public API is for cancelling Base UI event handlers.

It's an advanced use case, but an unavoidable need if you are building a complex, highly interactive app

As an example from Radix, there you can do things like:
```jsx
onEscKeydown={event => event.preventDefault()}
```

It overloads the native `event.preventDefault()` to mean "prevent browser default AND the Radix handler", which may have been a bit of an overkill.

We relied on these escape hatches quite a lot back at Modulz when we needed more control over interactions in the product

packages/react/src/use-button/useButton.ts
```diff
@@ -1,145 +1,120 @@
...
+          // calling preventDefault in keyUp on a <button> will not dispatch a click event if Space is pressed
+          // https://codesandbox.io/p/sandbox/button-keyup-preventdefault-dn7f0
+          // Keyboard accessibility for non interactive elements
+          if (
+            event.target === event.currentTarget &&
+            !isNativeButton() &&
+            !disabled &&
+            event.key === ' '
+          ) {
+            externalProps.onClick?.(event);
```
mj12albert, line 91: I'd prefer not making onClick a param just for this case

atomiks (approved):

PR #1182, opened 2024-12-19
[docs] Sturdier quick nav math

vladmoroz:
Avoid back and forth quick nav positioning updates when:
- Rounding errors occur due to zoom / pinch-zoom
- Current position can be either sticky or non-sticky without any difference
- Bouncing back from overscroll


vladmoroz:
@atomiks you've mentioned some issues in Safari when pinch zoomed before, should be better now

colmtuite (approved):

Issue #1186, opened 2024-12-19
[docs] Code snippets are not accessible


Studio384:
# Docs feedback

## How can we improve Base UI documentation?

There are various issues with the code examples.

- The scrollbar doesn't appear on mouse hover, instead you must scroll before the scrollbar becomes visible. Scrolling is only possible by selecting text and dragging the cursor left and right, middle clicking the scroll wheel or - if the mouse supports it - pushing the scroll wheel left and right. There is seemingly no keyboard navigation.
- When visible, the scroll bar is imo too easily dismissed by accidentally clicking just slightly above the bar, but being able to interact with the bar requires to be relatively fast so the scroll bar doesn't hide again.
- On large code examples that require scrolling, this experience is just incredibly bad. Mayhaps the maximum height of the code examples should be limited.

This scrollbar seems to be based on the ScrollArea component, whose example does show the scroll bar on hovering over the area. It would be great if that behavior is used in the code blocks as well. Or better yet; just use the native scroll bars.

- Selecting code but leaving the code block with the cursor will result in the selection being dismissed. This only happens in the collapsible code blocks.

https://github.com/user-attachments/assets/66125534-8084-4aa1-bb7a-6cd27ea2e770

- Selecting code and going with the cursor outside the codeblock results in unexpected behavior where the selection seems to flip. This makes that some extra precision is required, because moving back into the codeblock won't restore the previous selection. This also only happens with the collapsible code blocks. This bug does not appear on Firefox.

https://github.com/user-attachments/assets/97ce97b7-58ad-48c4-b61f-c2b24a626236

Furthermore, moving even further out away from the block removes the selection entirely from the code and instead only selects the contents above it instead.

Issue #982, opened 2024-12-06
[ScrollArea] Make it possible to recreate the macOS overlay scrollbar behaviour


vladmoroz:
I've tried to recreate how the macOS overlay scrollbar works, but couldn't.

Mainly, it's shown when scrolling, and once shown, it persists if the cursor is placed over the scrollbar itself in time before it hides—we don't have a style hook for this, nor it seems to be possible with CSS

**Search keywords**:

atomiks:
> we don't have a style hook for this, nor it seems to be possible with CSS

The original demo on `components/react-scroll-area` does this though? It uses `:hover` on the `Scrollbar` part itself to persist it.

```css
  &[data-hovering],
  &[data-scrolling],
  &:hover {
    visibility: visible;
    opacity: 1;
  }

  &[data-scrolling]:not(:hover) {
    transition: none;
  }
```

vladmoroz:
@atomiks I still couldn't figure out how to be able to re-hover a scrollbar after hovering it and then unhovering it. macOS provides a grace period for that

One more detail is that the native overlay scrollbars show only for the direction that is being scrolled—I think this one is more important

atomiks:
@vladmoroz I notice in the linked PR there's a grace period, can `transition-delay` be used?

> One more detail is that the native overlay scrollbars show only for the direction that is being scrolled—I think this one is more important

This makes sense

PR #1187, opened 2024-12-19
[docs] Code block scroll improvements

vladmoroz:
- Make sure scrollbar appears on hover
- Make sure the scrollarea is focusable
- Fix a bug when scrollarea renders would cause contents to lose text selection (not as much a bug as an unfortunate coincidence of rendering the viewport and dangerously setting inner html on it)
- ~Add max height to the code block scrollarea~
  - Update: we'll want to live with these code blocks for a bit before making changes

Closes https://github.com/mui/base-ui/issues/1186


vladmoroz:
@atomiks I think it's clear we need to be able to show only the scrollbars that match the direction you are scrolling in

https://github.com/user-attachments/assets/8c0faec6-ee95-4f32-8b4e-2f0e2c87c9cc

Mentioned/tracking in https://github.com/mui/base-ui/issues/982

atomiks (approved):

colmtuite:
Feels like a pretty bad experience getting get trapped in the code block when trying to scroll up/down the page. Of course, the alternative is very long code blocks. Currently, I prefer no max-height rather than being trapped in scroll areas. Not sure yet if there's a different way to solve it. I want to live with the docs for 1–3 months before tackling problems like this.

vladmoroz:
OK, removed max height for now. Let's live with it, I personally find getting trapped in code blocks more annoying than long continuous scroll.

PR #1188, opened 2024-12-20
[ScrollArea] Differentiate `x`/`y` orientation `data-scrolling`

### LINKED ISSUE #982: [ScrollArea] Make it possible to recreate the macOS overlay scrollbar behaviour

atomiks:
Closes #982

Previously both scrollbars received `data-scrolling` even if the orientation of the scroll was for the other scrollbar. This splits the states so they time out independently.


vladmoroz (approved):
very nice

PR #1191, opened 2024-12-20
[menu] Fix `openOnHover` issues

### LINKED ISSUE #1180: [Menu] `openOnHover` issues
For [Menu `openOnHover` demo](https://base-ui.com/react/components/menu)

### 1. First item incorrectly highlighted when mixing pointer/keyboard use

- Focus trigger using keyboard
- Press <kbd>esc</kbd>
- Hover the trigger with pointer
- First item highlighted incorrectly (nothing should be highlighted when using pointer)

### 2. Returns focus and scrolls to element out of view

- Focus 'Group Labels' demo trigger (View) at the bottom with keyboard
- Scroll up and hover the `openOnHover` demo
- Mouseout of the trigger
- Browser focuses and scrolls down to the 'Group Labels' trigger (focus management should be disabled like Popover when using `openOnHover`)

### 3. Stick behavior

- Hover `openOnHover` demo Menu
- Click
- Mouseout of trigger
- Stays stuck but should not (it also shouldn't stick if you moused into the popup, but didn't click anything)

Side note: I think Popover's implementation can also be improved in that clicking within the `patient threshold` makes it stick, but after 500ms (the current value) it closes, but they should at least match each other.

atomiks:
Closes #1180

Includes `data-instant` style hook for `Menu`

### DIFF
--- a/docs/reference/generated/menu-popup.json
+++ b/docs/reference/generated/menu-popup.json
@@ -19,7 +19,7 @@
       "description": "Present when the menu is closed."
     },
     "data-instant": {
-      "description": "Indicates the instant type of the popover popup.",
+      "description": "Indicates the instant type of the menu popup.",
       "type": "'click' | 'dismiss'"
     },
     "data-side": {
--- a/docs/reference/generated/menu-popup.json
+++ b/docs/reference/generated/menu-popup.json
@@ -19,7 +19,7 @@
       "description": "Present when the menu is closed."
     },
     "data-instant": {
-      "description": "Indicates the instant type of the menu popup.",
+      "description": "Present if animations should be instant.",
       "type": "'click' | 'dismiss'"
     },
     "data-side": {
--- a/docs/reference/generated/menu-popup.json
+++ b/docs/reference/generated/menu-popup.json
@@ -18,6 +18,10 @@
     "data-closed": {
       "description": "Present when the menu is closed."
     },
+    "data-instant": {
+      "description": "Indicates the instant type of the popover popup.",
+      "type": "'click' | 'dismiss'"
+    },
     "data-side": {
       "description": "Indicates which side the menu is positioned relative to the trigger.",
       "type": "'top' | 'bottom' | 'left' | 'right' | 'inline-end' | 'inline-start'"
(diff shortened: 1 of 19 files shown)


packages/react/src/popover/root/usePopoverRoot.ts
```diff
@@ -103,11 +104,11 @@ export function usePopoverRoot(params: usePopoverRoot.Parameters): usePopoverRoo
...
-        // Prevent impatient clicks from unexpectedly closing the popover.
-        setClickEnabled(false);
-        clearTimeout(clickEnabledTimeoutRef.current);
-        clickEnabledTimeoutRef.current = window.setTimeout(() => {
-          setClickEnabled(true);
+        // Only allow "patient" clicks to close the popover if it's open.
+        // If they clicked within 500ms of the popover opening, keep it open.
+        clearTimeout(stickIfOpenTimeoutRef.current);
+        stickIfOpenTimeoutRef.current = window.setTimeout(() => {
+          setStickIfOpen(false);
```
atomiks, line 111: While this new implementation makes more sense logically, it makes infotips like the ones on the docs work less well. Ignoring clicks is favorable there because it prevents multiple ones from being open simultaneously. The only way to get that to happen with the master version is to click it after closing/multiple times. cc: @vladmoroz

atomiks: Using `Popover.Backdrop` (or adding modality in general, at least for clicks) would actually solve this, since it would block pointer-events of everything underneath until they intentionally dismiss with outside press

colmtuite (approved):

docs/reference/generated/menu-popup.json
```diff
@@ -18,6 +18,10 @@
     "data-closed": {
       "description": "Present when the menu is closed."
     },
+    "data-instant": {
+      "description": "Indicates the instant type of the menu popup.",
```
michaldudak, line 22: TBH this description doesn't help much to understand what this attribute is for.

atomiks: Yep, agreed. The one for Popover is the same but I can update both of them.

packages/react/src/popover/root/usePopoverRoot.ts
benface: > Being able to hover open a menu/popover when there is another menu/popover open feels terrible

I totally agree @vladmoroz, and unfortunately it seems that even after this PR this is still possible (by hovering an item, waiting for the popover to open, clicking the item to close the popover, clicking it again to open it and make it sticky, and moving the cursor over a different item that also has a popover on hover). Is that intentional @atomiks?

benface: Makes sense, I ended up keeping track (in a context provider) of the last popover whose `onOpenChange` was called with `true`, and only passing `true` to `open` if it matches. Thanks @atomiks!

Issue #1189, opened 2024-12-20
[ScrollArea] Does not read `DirectionProvider` for `rtl`


atomiks:
`ScrollArea` uses the previous auto `getComputedStyle` check on mount

PR #1194, opened 2024-12-20
[ScrollArea] Read `DirectionProvider` and use logical positioning CSS props

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1189

RTL horizontal scroll area demo: https://codesandbox.io/p/sandbox/eloquent-nightingale-s74566


atomiks (approved):

packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
```diff
@@ -35,10 +38,10 @@ const ScrollAreaRoot = React.forwardRef(function ScrollAreaRoot(
 
   const contextValue = React.useMemo(
     () => ({
-      dir,
+      direction,
```
mnajdova, line 41: I am curious, what's the benefit of adding the direction in a context, when each subcomponent that need it can read the DirectionProvider's context?

mj12albert: > what's the benefit of adding the direction in a context

It doesn't need to be in the root context anymore, I've updated this ~

PR #1212, opened 2024-12-22
[docs] Remove `rehypeEmptyLines`

### LINKED ISSUE #1211: [docs] Missing line on Composition page in "Composing multiple components"
# Docs feedback

In the "Composing multiple components" section on the Composition page, there is a missing line that is not showing in the code example. The line is `<Menu.Item render={<a href="base-ui.com" />}>` (source [here](https://github.com/mui/base-ui/blob/master/docs/src/app/(public)/(content)/react/handbook/composition/page.mdx?plain=1#L69)). Perhaps the `prettier-ignore` comment on the line above is causing it to be skipped somehow?

atomiks:
Closes #1211. Not sure what this is for but it's breaking the code blocks.

### DIFF
--- a/docs/src/syntax-highlighting/index.mjs
+++ b/docs/src/syntax-highlighting/index.mjs
@@ -1,6 +1,5 @@
 import { createHighlighter } from 'shiki';
 import rehypePrettyCode from 'rehype-pretty-code';
-import { rehypeEmptyLines } from './rehypeEmptyLines.mjs';
 import { rehypeInlineCode } from './rehypeInlineCode.mjs';
 import { rehypePrettierIgnore } from './rehypePrettierIgnore.mjs';
 import { rehypeJsxExpressions } from './rehypeJsxExpressions.mjs';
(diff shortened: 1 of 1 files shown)


aarongarciah (approved):

docs/src/syntax-highlighting/index.mjs
```diff
@@ -1,6 +1,5 @@
 import { createHighlighter } from 'shiki';
 import rehypePrettyCode from 'rehype-pretty-code';
-import { rehypeEmptyLines } from './rehypeEmptyLines.mjs';
```
oliviertassinari, line 3: So, we should delete this file too? We `rehypeEmptyLines.mjs` is orphan now?

atomiks: It has some purpose, but I'm not sure what - Vlad knows. I kept it in case it'll be re-added soon with a fixed implementation. Since git history keeps it here, it's not crucial to be kept around though.

PR #1209, opened 2024-12-20
[Popover, Tooltip] Fix PopoverTrigger and TooltipTrigger prop types

okmr-d:
Fix wrong prop types.
スクリーンショット 2024-12-21 5 09 47

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #1213, opened 2024-12-23
[core] Fix render prop types

atomiks:

Continues #1209

Switches to `GenericHTMLProps` for the `render` props type. This includes the `ref` which was previously missing. Using `'div'` for Tooltip uses the `GenericHTMLProps` regardless.

### DIFF
--- a/packages/react/src/tooltip/trigger/TooltipTrigger.tsx
+++ b/packages/react/src/tooltip/trigger/TooltipTrigger.tsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import { useTooltipRootContext } from '../root/TooltipRootContext';
 import { useComponentRenderer } from '../../utils/useComponentRenderer';
 import { useForkRef } from '../../utils/useForkRef';
-import type { BaseUIComponentProps } from '../../utils/types';
+import type { BaseUIComponentProps, GenericHTMLProps } from '../../utils/types';
 import { triggerOpenStateMapping } from '../../utils/popupStateMapping';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/react/src/tooltip/trigger/TooltipTrigger.tsx
```diff
@@ -46,14 +46,18 @@ namespace TooltipTrigger {
     open: boolean;
   }
 
-  export interface Props extends BaseUIComponentProps<any, State> {}
+  export interface Props extends BaseUIComponentProps<'div', State> {}
```
mnajdova, line 49: Do we expect here people to always need to cast if they use different element?

mnajdova (commented):
Can we add types test please? We could add `TooltipTrigger.spec.ts` file with different use-cases.

packages/react/src/tooltip/trigger/TooltipTrigger.tsx
atomiks: The `render` prop's `props` parameter is actually loosely typed in order to allow you to compose different tags. 

We need to ensure we document when a tag can't be replaced if we aren't polyfilling native behavior (e.g. if it relies in a specific tag's native props in order to work, like `<button>` being replaced with a `<span>`, but internally is not using `useButton()` to keep it accessible).

We could actually remove the first `tag` type since it serves no purpose in this case

mnajdova: Let's see if this will be enough, if we see more issues around this we can revisit it.

mnajdova (approved):

PR #1223, opened 2024-12-24
[useAnchorPositioning] Add `OffsetFunction` for `sideOffset` and `alignOffset`

### LINKED ISSUE #1195: [Popover] How to align the start of the popover with the center of the trigger?
First of all, amazing work on this library. The alpha version looks super solid already.

I've come across a potential limitation when using `Popover`. My design requires me to align the left / start of the popover with the center of the trigger (minus an offset, because actually the goal is to align the center of the first column of the popover, which has a fixed width... but that's irrelevant). With the `align` prop, it only seems possible to align the `start` of the popover with the start of the trigger, the `center` of the popover with the center of the trigger, or the `end` of the popover with the end of the trigger. Is there a solution I'm not seeing, given that the trigger has an unknown/dynamic width? If not, I have 2 potential (but very rough) ideas of how it could be handled (assuming you agree that it's a worthy use case!)

1. Another prop named `triggerAlign` or something, which would default to the same value as `align`, but in my case I would set it to `center`.
2. I have no idea if this is technically feasible, but maybe the `alignOffset` prop could accept a function that would receive the dimensions of the trigger and of the popup, and could do some math based on those to return an offset value?

Thanks so much again, I'm excited to see how this library evolves.

atomiks:
Closes #1195


packages/react/src/utils/useAnchorPositioning.ts
```diff
@@ -118,12 +122,29 @@ export function useAnchorPositioning(
   // presence.
   const arrowRef = React.useRef<Element | null>(null);
 
+  // Keep these reactive if they're not functions
+  const sideOffsetDep = typeof sideOffset === 'number' ? side : 0;
+  const alignOffsetDep = typeof alignOffset === 'number' ? align : 0;
```
atomiks, line 127: Positioning reactivity is lost when they specify a function; we can't tell if they're going to `useCallback` or not, usually the function gets inlined like:

```jsx
<Popover.Positioner sideOffset={sizes => sizes.anchor.height / 2}>
```

packages/react/src/utils/useAnchorPositioning.ts
```diff
@@ -118,12 +122,29 @@ export function useAnchorPositioning(
   // presence.
   const arrowRef = React.useRef<Element | null>(null);
 
+  // Keep these reactive if they're not functions
+  const sideOffsetDep = typeof sideOffset !== 'function' ? sideOffset : 0;
+  const alignOffsetDep = typeof alignOffset !== 'function' ? alignOffset : 0;
+
```
atomiks, line 128: Positioning reactivity is lost when they specify a function. We can't tell if they're going to `React.useCallback` or not - usually the function gets inlined like:

```jsx
<Popover.Positioner sideOffset={(sizes) => sizes.anchor.height / 2}>
```

So it can't be specified as a dependency as it's always changing. This means if they update the math inside the function, the position won't update live on fast refresh (but it will whenever the anchoring updates at some point)

docs/reference/generated/tooltip-positioner.json
```diff
@@ -18,9 +18,9 @@
       "description": "Which side of the anchor element to align the popup against.\nMay automatically change to avoid collisions."
     },
     "sideOffset": {
-      "type": "number",
+      "type": "number | (data) => number",
       "default": "0",
-      "description": "Distance between the anchor and the popup."
+      "description": "Distance between the anchor and the popup in pixels.\nAlso accepts a function that returns a number to read the dimensions of the anchor and popup,\nalong with its side and alignment.\n\n- `data.anchor`: the dimensions of the anchor element with properties `width` and `height`.\n- `data.popup`: the dimensions of the popup element with properties `width` and `height`.\n- `data.side`: which side of the anchor element the popup is aligned against.\n- `data.align`: how the popup is aligned relative to the specified side."
```
atomiks, line 23: Screenshot 2024-12-24 at 7 24 35 pm

packages/react/src/utils/useAnchorPositioning.ts
```diff
@@ -118,12 +138,31 @@ export function useAnchorPositioning(
...
-      crossAxis: alignOffset,
-      alignmentAxis: alignOffset,
-    }),
+    offset(
+      ({ rects, placement: currentPlacement }) => {
+        const data = {
+          side: getLogicalSide(sideParam, getSide(currentPlacement), isRtl),
+          align: getAlignment(currentPlacement) || 'center',
+          anchor: { width: rects.reference.width, height: rects.reference.height },
+          popup: { width: rects.floating.width, height: rects.floating.height },
```
atomiks, line 152: Technically this is the `Positioner` element, not `Popup`, but in practically every case they should be the same since it's a presentational wrapper

packages/react/src/utils/useAnchorPositioning.ts
michaldudak: If this ever becomes a problem, the `sideOffset`'s argument can include an `update` callback function that will force updating the layout.

michaldudak:
Is it possible to place a popup outside of the trigger's bounds?
I'm trying to align the start of the popup with the right-hand side of the trigger with some margin, but it seems that the offset can only be as high as the anchor width (the `+ 5` is ignored):

```tsx
<Popover.Positioner
  sideOffset={5}
  alignOffset={(sizes) => sizes.anchor.width + 5}
  align="start"
>
```

atomiks:
@michaldudak the `limitShift` logic forcefully prevents that from happening so it doesn't "detach". I'm not sure if it's important to support that in practice

michaldudak:
OK, it's fine. We could write a few words about it in the docs, so it's not surprising. I miss having a dedicated section per prop/group of props

atomiks:
Renamed `popup` to `positioner` so it explicitly matches the relevant element's dimensions, even though popup could be simpler and should usually match the size - it may not always

michaldudak (approved):
Could you add the description of the `data` structure to the `alignOffset` prop as well?

Besides that, it looks good to me.

atomiks:
> Could you add the description of the data structure to the alignOffset prop as well?

It already is from what I see?

michaldudak:
OK, weird, it doesn't show up on the docs: https://deploy-preview-1223--base-ui.netlify.app/react/components/menu#positioner

atomiks:
@michaldudak yeah I've noticed the docs preview sometimes doesn't update due to some cache issue. Locally, it's there and is after merging

PR #1231, opened 2024-12-26
[POC] Anchored modality changes

atomiks:
This changes the modality of anchored popups that primarily open on click, and adds the `modal` prop to `Popover`.

- Scroll locking is only enabled if the `Backdrop` part is rendered (where obscuring content justifies the lock)
- The internal backdrop becomes inert (no longer blocks `pointer-events`) if the anchor is hidden from view (possible alternative: if the popup is hidden, but Floating UI doesn't provide a hook for this out of the box). This allows the user to instantly click any item beneath once they've scrolled away from the popover and no longer 'care' about it being open. Outside clicks are presumed to be careful and intentional at this point


colmtuite:
I've lived with this the past few days (checking it intermittently). I'm not seeing much value/sense in this yet.

When a Menu is open, it's usually important to me that clicks outside it are blocked. When clicks outside are blocked, it's important that scroll is locked. These things have nothing to do with a backdrop being rendered.

I can potentially see an argument for closing the Menu when you scroll it out of view, or when you scroll x pixels away from where you were when it opened, or something like that. But just changing the behaviour of the Menu when its state has not changed, seems bad/odd.

atomiks:
@colmtuite to make https://github.com/mui/base-ui/pull/1234 work well for infotips, Popovers also need to add a `modal` prop which is `true` by default. When this is added, it will also by default lock the scroll for Popover. This doesn't match Radix (which doesn't block outside clicks) and feels overly restrictive.

> When clicks outside are blocked, it's important that scroll is locked. These things have nothing to do with a backdrop being rendered.

The main reason to block outside pointer events is to prevent sloppy outside presses from causing any interaction. The only reason to lock the scroll in conjunction then is because when scrolling away from the popup, you may be confused why you can't instantly click anything beneath it on first click, since the popup is no longer visible/relevant. This PR makes it so the overlay becomes inert when it isn't visible, so this concern is mostly gone (depending on the anchor's visibility).

The only justified reason to lock outside scroll is because you've obscured the content beneath the popup and can't see it properly. It makes sense to prevent scrolling in that case since you'll become disoriented after dismissing the popup. But if the backdrop is fully transparent, there's no justification to lock scroll when you can see everything beneath the popup - the fact that outside presses are prevented on first click isn't related to the ability to scroll around

atomiks:
Supeseded by https://github.com/mui/base-ui/pull/1459

PR #1234, opened 2024-12-26
[popover] New stick behavior

atomiks:
- If the user interacts with anything inside the popover after it opens on hover (detected with `pointerdown`), it sticks like it was clicked. This ensures sloppy mouseleaves don't unexpectedly close the popover, including when attempting to select text inside the popover, focusing an input and needing to navigate to a native browser popup, etc.
- Copies #1191 `openOnHover` stick behavior, where pressing within the patient threshold doesn't entirely ignore the click. This relies on #1231 (`modal` prop and related changes) to work well in terms of user experience.


atomiks:
Superseded by https://github.com/mui/base-ui/pull/1459

PR #1236, opened 2024-12-27
[popups] `actionsRef` prop

### LINKED ISSUE #1160: JavaScript exit animations aren't properly supported
With `keepMounted` + `AnimatePresence` in (Framer) Motion, `node.getAnimations()` detects `opacity` transitions, but not `transform`:

This won't work, with the element unmounting instantly:
```js
<motion.div exit={{ x: '-100%' }} />
```

This does however:

```js
<motion.div exit={{ x: '-100%', opacity: 0.9999 }} />
```

### Possible solution

A new value for `keepMounted`,  such as `keepMounted="controlled"`, indicating the element's mounting is being controlled by another component like `AnimatePresence`. This will prevent `hidden` from being added, and remove the `disabled` prop from `FocusManager`.

atomiks:
Closes #1160

[JavaScript animations page](https://deploy-preview-1236--base-ui.netlify.app/react/handbook/animation#javascript-animations). I focused on `Motion` fairly exclusively. The `actionsRef` prop holds imperative methods, currently just `unmount`. `action.current.unmount()` is for when animating `opacity` doesn't work in other libraries for the most part, as `opacity: 0.9999` is really the simplest solution.

[Motion experiments](https://deploy-preview-1236--base-ui.netlify.app/experiments/motion)


packages/react/src/utils/useAfterExitAnimation.tsx
```diff
@@ -8,13 +8,17 @@ import { useLatestRef } from './useLatestRef';
...
 export function useAfterExitAnimation(parameters: useAfterExitAnimation.Parameters) {
-  const { open, animatedElementRef, onFinished: onFinishedParam } = parameters;
+  const { enabled = true, open, animatedElementRef, onFinished: onFinishedParam } = parameters;
 
   const onFinished = useEventCallback(onFinishedParam);
   const runOnceAnimationsFinish = useAnimationsFinished(animatedElementRef);
   const openRef = useLatestRef(open);
 
-  useEnhancedEffect(() => {
+  React.useEffect(() => {
```
atomiks, line 17: Change to regular `useEffect` as the timing lets Motion know the exit animation styles have been applied inside a single `requestAnimationFrame` when checking for `element.getAnimations()`

packages/react/src/popover/root/usePopoverRoot.ts
```diff
@@ -75,15 +75,20 @@ export function usePopoverRoot(params: usePopoverRoot.Parameters): usePopoverRoo
...
     open,
     animatedElementRef: popupRef,
-    onFinished: () => {
-      setMounted(false);
-      setOpenReason(null);
-    },
+    onFinished: handleUnmount,
   });
 
+  React.useImperativeHandle(params.unmountRef, () => ({ unmount: handleUnmount }), [handleUnmount]);
```
atomiks, line 90: An API with a bunch of imperative methods could be useful in general, not just for this, so it could be named something more generic to future-proof it. This concept may be similar to using hooks, or component stores as in Ariakit. One difference is you can't "read" values in render, it's just for accessing internal methods/state inside event handlers & effects (since it's a ref).

michaldudak: :+1:
A couple of Material UI components use this pattern and call the prop `action` (like https://github.com/mui/material-ui/blob/master/packages/mui-material/src/ButtonBase/ButtonBase.d.ts#L13)

packages/react/src/dialog/root/useDialogRoot.ts
```diff
@@ -226,11 +233,18 @@ export interface SharedParameters {
    * @default true
    */
   dismissible?: boolean;
+  /**
+   * A ref to imperative actions.
+   */
+  action?: React.RefObject<{ unmount: () => void }>;
```
michaldudak, line 239: The type of the ref should be exported so it's easy for consumers to pass a correct type to `React.useRef`

packages/react/src/menu/root/MenuRoot.test.tsx
```diff
@@ -877,6 +877,46 @@ describe('<Menu.Root />', () => {
...
+        expect(screen.queryByRole('menu')).not.to.equal(null);
+      });
+
+      await user.click(trigger);
+
+      await waitFor(() => {
+        expect(screen.queryByRole('menu')).not.to.equal(null);
+      });
+
+      act(() => actionRef.current.unmount());
```
michaldudak, line 912: In the future version of React Testing Library `act` will require awaiting. 

```suggestion
      await act(async () => actionRef.current.unmount());
```

packages/react/src/dialog/root/useDialogRoot.ts
atomiks: Hmm, how should it be exported? Like `Dialog.ActionsRef`?

michaldudak: Yeah, should be fine like that.

packages/react/src/menu/root/MenuRoot.tsx
```diff
@@ -13,7 +13,7 @@ import type { OpenChangeReason } from '../../utils/translateOpenChangeReason';
  *
  * Documentation: [Base UI Menu](https://base-ui.com/react/components/menu)
  */
-const MenuRoot: React.FC<MenuRoot.Props> = function MenuRoot(props) {
```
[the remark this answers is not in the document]
michaldudak: Yeah, `DialogRoot` has the same issue - it has `propTypes` in its built type definitions:
```
declare const DialogRoot: {
    (props: DialogRoot.Props): React.JSX.Element;
    propTypes: any;
};
```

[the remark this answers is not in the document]
atomiks: Fixed Roots that were missing it here

michaldudak (approved):

mj12albert:
🤔 I wonder if this is a viable method for cleanly closing menus in this situation: https://github.com/mui/base-ui/pull/1349#issuecomment-2642086279

Issue #1230, opened 2024-12-25
[Slider] `value` type should narrow `onValueChange` value type


aarongarciah:
# Feature request

## Summary

The `Slider` accepts values of type `Array<number> | number` but the `onValueChange` and `onValueCommitted` callbacks don't narrow the type of the `value` parameter to `number` or `Array<number>` and users need to type cast it manually.

Example of code throwing a TS error:

```jsx
//  ❌ Argument of type 'number | number[]' is not assignable to parameter of type 'SetStateAction<number>'.
//    Type 'number[]' is not assignable to type 'SetStateAction<number>'.

const [value, setValue] = React.useState(25);
<Slider.Root value={value} onValueChange={(value) => setValue(value}>
```

Example of type casting needed to overcome the TS error:

```jsx
const [value, setValue] = React.useState(25);
<Slider.Root value={value} onValueChange={(value) => setValue(value as number}>
```

## Motivation

Avoid users needing to do manual type casting.

good-jinu:
I'm planning to resolve this issue using a similar approach as the previous [metiral-ui PR](https://github.com/mui/material-ui/pull/44777).

Here's the plan:

- Update Type Definitions
- Update Demos
- Update Documentation

PR #1241, opened 2024-12-28
[Slider] Narrow the type of `value` in callbacks

seloner:
Trying to fix
image

Closes https://github.com/mui/base-ui/issues/1230


packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -10,15 +10,16 @@ import { sliderStyleHookMapping } from './styleHooks';
...
 
 /**
  * Groups all parts of the slider.
  * Renders a `<div>` element.
  *
  * Documentation: [Base UI Slider](https://base-ui.com/react/components/slider)
  */
-const SliderRoot = React.forwardRef(function SliderRoot(
-  props: SliderRoot.Props,
+const SliderRoot = fixedForwardRef(function SliderRoot<TValue extends number | number[]>(
```
seloner, line 21: ReactForwardRef does not work well with generics.
So we need to create a wrapper function to apply the generic.

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -157,28 +158,25 @@ export namespace SliderRoot {
     values: ReadonlyArray<number>;
   }
 
-  export interface Props
+  export interface Props<TValue extends number | number[]>
```
seloner, line 161: My first approach here was to try to carry this generic all the way down in the slider components but it resulted to just pollute  the code with generics.
Instead I simple just try to modify the top level component that the consumer will use.

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -192,12 +190,34 @@ export namespace SliderRoot {
...
+    /**
+     * The component orientation.
+     * @default 'horizontal'
+     */
   }
 }
 
 export { SliderRoot };
 
+// @ts-expect-error
```
seloner, line 220: I don't know how to handle this I am getting ts error for proptypes.
Tried the `proptypes` script but no luck.

good-jinu:
Nice work! @seloner

This issue originated from https://github.com/mui/material-ui/pull/44777. Therefore, referring to https://github.com/mui/material-ui/pull/44777 might be helpful.

IMHO, it would be better to run `pnpm proptypes` and fix the ts errors since the CI: test_static check has failed.

image
*It has failed CI check*

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -192,12 +190,34 @@ export namespace SliderRoot {
...
+     *
+     * @param {number | number[]} value The new value.
+     * @param {Event} event The corresponding event that initiated the change.
+     * **Warning**: This is a generic event not a change event.
+     */
+    onValueCommitted?: (value: TValue, event: Event) => void;
+    /**
+     * The component orientation.
+     * @default 'horizontal'
+     */
```
good-jinu, line 214: @seloner
It seems like this should be removed.

packages/react/src/slider/root/SliderRoot.tsx
good-jinu: @seloner 
I believe we can define Props without using generics. By utilizing intersection types, we can potentially make the necessary changes internally without affecting the external API. This approach could simplify the type definitions and improve readability.

Here’s an example of how we can implement this:

```diff
@@ -158,7 +158,10 @@ export namespace SliderRoot {
     values: ReadonlyArray<number>;
   }
 
-  export interface Props<TValue extends number | number[]>
+  export type Props = PropsTemplate<number> &
+    PropsTemplate<number[]> &
+    PropsTemplate<number | number[]>;
+  export interface PropsTemplate<TValue extends number | number[]>
     extends Pick<
         useSliderRoot.Parameters,
         | 'disabled'
```

Please refer to https://github.com/mui/material-ui/pull/44777#issuecomment-2564268790

packages/react/src/slider/root/SliderRoot.test.tsx
```diff
@@ -49,7 +49,7 @@ function TestSlider(props: SliderRoot.Props) {
   );
 }
 
-function TestRangeSlider(props: SliderRoot.Props) {
+function TestRangeSlider(props: SliderRoot.Props<number[]>) {
```
good-jinu, line 52: @seloner I think we can define Props without using generics.

Check https://github.com/mui/base-ui/pull/1241/files#r1899037802, please

packages/react/src/slider/root/SliderRoot.tsx
good-jinu: @seloner I think we can remove generic SliderRoot component by removing generic in Props interfacce.

Check https://github.com/mui/base-ui/pull/1241/files#r1899037802, please

packages/react/src/slider/root/SliderRoot.tsx
good-jinu: if you remove fixedForwardRef and replace it with `React.forwardRef`, then `// @ts-expect-error` can be removed.

```diff
@@ -18,8 +18,8 @@ import { fixedForwardRef } from '../utils';
  *
  * Documentation: [Base UI Slider](https://base-ui.com/react/components/slider)
  */
-const SliderRoot = fixedForwardRef(function SliderRoot<TValue extends number | number[]>(
-  props: SliderRoot.Props<TValue>,
+const SliderRoot = React.forwardRef(function SliderRoot(
+  props: SliderRoot.Props,
   forwardedRef: React.ForwardedRef<HTMLDivElement>,
 ) {
   const {
```

good-jinu (changes requested):
I’d like to suggest the following tasks to address this issue:

1. **Convert `SliderRoot.Props` to an Intersection Type**: Change the Props definition to use an intersection type, eliminating the need for generics.
```diff
@@ -158,7 +158,10 @@ export namespace SliderRoot {
     values: ReadonlyArray<number>;
   }
 
-  export interface Props<TValue extends number | number[]>
+  export type Props = PropsTemplate<number> &
+    PropsTemplate<number[]> &
+    PropsTemplate<number | number[]>;
+  export interface PropsTemplate<TValue extends number | number[]>
     extends Pick<
         useSliderRoot.Parameters,
         | 'disabled'
```
2. **Remove Generic Type from SliderRoot**: Simplify SliderRoot by removing the generic type.
3. **Replace `fixedForwardRef` with `React.forwardRef`**: Update the code to use `React.forwardRef` directly, which will allow us to remove the `// @ts-expect-error`.
```diff
@@ -18,8 +18,8 @@ import { fixedForwardRef } from '../utils';
  *
  * Documentation: [Base UI Slider](https://base-ui.com/react/components/slider)
  */
-const SliderRoot = fixedForwardRef(function SliderRoot<TValue extends number | number[]>(
-  props: SliderRoot.Props<TValue>,
+const SliderRoot = React.forwardRef(function SliderRoot(
+  props: SliderRoot.Props,
   forwardedRef: React.ForwardedRef<HTMLDivElement>,
 ) {
   const {
```
4. **Remove `// @ts-expect-error`**: With the changes above, the TypeScript error should be resolved, making the `// @ts-expect-error` unnecessary.

These changes should help resolve the type-related issues. Let me know your thoughts!

seloner:
@good-jinu 
I have applied your suggestions but I can't make this work 🤔 
SlideRoot.test.tsx broke

image

good-jinu:
@seloner My bad

> An intersection of function types is treated as overloaded signatures in TypeScript.

So It seems like just a field like value doesn't work with it.
How about union type?

For example: 

```typescript
export type Props =  PropsTemplate<number> |
  PropsTemplate<number[]> |
  PropsTemplate<number | number[]>;
```

seloner:
> ```ts
>  PropsTemplate<number> |
>   PropsTemplate<number[]> |
>   PropsTemplate<number | number[]>;
> ```

It does not work
I think the generic is needed.
Another approach is to export two SliderRoots one SliderRoot and one RangeSliderRoot to support both types without generic?

mj12albert:
> Another approach is to export two SliderRoots one SliderRoot and one RangeSliderRoot to support both types without generic?

@seloner Thanks for working on this, I'm afraid this isn't an option though

> So It seems like just a field like value doesn't work with it.

@good-jinu Not sure if I understand you correctly, by "field like value" did you mean a single non-array value?

seloner:
> > Another approach is to export two SliderRoots one SliderRoot and one RangeSliderRoot to support both types without generic?
> 
> 
> 
> @seloner Thanks for working on this, I'm afraid this isn't an option though
> 
> 
> 
> > So It seems like just a field like value doesn't work with it.
> 
> 
> 
> @good-jinu Not sure if I understand you correctly, by "field like value" did you mean a single non-array value?
> 
> 

We can still do the solution with the generics

mj12albert:
> We can still do the solution with the generics

I think this is ok as long as it doesn't affect the public API, just wonder why @good-jinu is against generics (I prefer to avoid them too, but I'm just curious here)

A while back we did decide that for components with values that could sometimes be an array (accordion, slider, toggle group), we'd make the value in callbacks always be an array so it's predictable, also curious to see what you all think about this!

seloner:
> > We can still do the solution with the generics
> 
> I think this is ok as long as it doesn't affect the public API, just wonder why @good-jinu is against generics (I prefer to avoid them too, but I'm just curious here)
> 
> A while back we did decide that for components with values that could sometimes be an array (accordion, slider, toggle group), we'd make the value in callbacks always be an array so it's predictable, also curious to see what you all think about this!

We could use a default value so it does no really affect the public API compared to now .
```tsx
type Props<T extends number | number[] = number | number[]> = {
  value: T;
  onValueChange: (value: T) => void;
};

const Component: <T extends number | number[]>(props: Props<T>) => {};

```

I like your suggestion to remove the number case and always go with an array.
It will work for this specific case.
The only issue I see that it tries to bypass the problem and not solve it.
In this case it will be fine but in general for a more complex case it will not work.
For example if there is  a case with possible inputs string/number.

I am new to this project so I can't really answer though what is the best solution here.

good-jinu:
> I think this is ok as long as it doesn't affect the public API, just wonder why @good-jinu is against generics (I prefer to avoid them too, but I'm just curious here)



I actually agree with using generics; however, I wanted to avoid changing the public API. My main intention was to eliminate the `@ts-expect-error` from previous commits due to the TypeScript error.

I also think it's fine as long as it doesn't change the public API.

seloner:
@good-jinu  @mj12albert 

Updated.
Used the defaults for generics.
The only issue that I can't solve is proptypes.

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -157,16 +158,14 @@ export namespace SliderRoot {
     values: ReadonlyArray<number>;
   }
 
-  export interface Props
+  export interface Props<TValue extends number[] | number = number | number[]>
```
mj12albert, line 161:
```suggestion
  export interface Props<Value extends readonly number[] | number = number | readonly number[]>
```
`Value` is fine ~ we don't use the `T*` convention, also the array form of value/defaultValue should be readonly arrays

packages/react/src/slider/root/SliderRoot.tsx
mj12albert: Instead of making `fixedForwardRef`, we could cast the result of `React.forwardRef`, e.g. something like this:

```jsx
type SliderRootType = {
  <Value>(props: SliderRoot.Props<Value>): React.JSX.Element;
  propTypes?: any;
}

const SliderRoot = React.forwardRef(function SliderRoot<Value>(
  props: SliderRoot.Props<Value>,
  forwardedRef: React.ForwardedRef<HTMLDivElement>
) {
  //
}) as SliderRootType;
```

We did something similar in the legacy package: https://github.com/mui/material-ui/blob/master/packages/mui-base/src/Select/Select.tsx#L235

mj12albert (dismissed):

packages/react/src/slider/root/SliderRoot.tsx
mj12albert: https://github.com/mui/base-ui/pull/1241#discussion_r1906510445

mj12albert:
> I like your suggestion to remove the number case and always go with an array.

@seloner Since this would be a breaking change, let's keep going with the solution using generics!

I've left some comments, you should also merge/rebase master first btw ~

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -12,16 +12,24 @@ import { useSliderRoot } from './useSliderRoot';
 import { SliderRootContext } from './SliderRootContext';
 import { useFieldRootContext } from '../../field/root/FieldRootContext';
 
+type SliderRootType = {
+  <Value extends number | readonly number[]>(
+    props: SliderRoot.Props<Value> & {
+      ref?: React.RefObject<HTMLDivElement>;
```
seloner, line 18: This was needed to include the ref into types @mj12albert

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -197,12 +204,27 @@ export namespace SliderRoot {
...
+     *
+     * @param {number | number[]} value The new value.
+     * @param {Event} event The corresponding event that initiated the change.
+     * **Warning**: This is a generic event not a change event.
+     */
+    onValueCommitted?: (value: Value, event: Event) => void;
   }
 }
 
-export { SliderRoot };
```
seloner, line 204: @mj12albert 
This was needed because ts error 
`Parameter props of call signature from exported interface has or is using private name`

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -288,7 +310,7 @@ SliderRoot.propTypes /* remove-proptypes */ = {
   /**
    * Callback function that is fired when the slider's value changed.
    *
-   * @param {number | readonly number[]} value The new value.
+   * @param {number | number[]} value The new value.
```
seloner, line 313: This was auto generated.

seloner:
> > I like your suggestion to remove the number case and always go with an array.
> 
> @seloner Since this would be a breaking change, let's keep going with the solution using generics!
> 
> I've left some comments, you should also merge/rebase master first btw ~

Updated

packages/react/src/slider/root/SliderRoot.tsx
seloner: I don't think the types work correctly with your recommendation 🤔 
image

packages/react/src/slider/root/SliderRoot.tsx
seloner: I managed to bypass this with 
https://github.com/mui/base-ui/pull/1241/commits/1b5a14dc3d29a5e8c404d71fc913fc5878788508

good-jinu (approved):
Great work on the improvements! Since it has no more ts error, I guess it is ready.

seloner:
image
@good-jinu 
I don't have access here. What is wrong ? 🤔 
Can you help me?

good-jinu:
> image @good-jinu I don't have access here. What is wrong ? 🤔 Can you help me?

It seems like the error `There was an error fetching the storage URL during POST: 503` suggests network connectivity issues between CircleCI and the Codecov server. I think we need help fixing this CI.

@mj12albert  can you help us fixing this?

packages/react/src/slider/root/SliderRoot.tsx
mj12albert: 👍 `readonly` is only TS and not in `prop-types`

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -59,8 +59,9 @@ const SliderRoot = React.forwardRef(function SliderRoot(
     min,
     minStepsBetweenValues,
     name: nameProp ?? '',
-    onValueChange: onValueChangeProp ?? NOOP,
-    onValueCommitted: onValueCommittedProp ?? NOOP,
+    onValueChange: (onValueChangeProp as useSliderRoot.Parameters['onValueChange']) ?? NOOP,
+    onValueCommitted:
+      (onValueCommittedProp as useSliderRoot.Parameters['onValueCommitted']) ?? NOOP,
```
mj12albert, line 64: @seloner It should be possible without casting these two props (the legacy `useSelect` could also be used as a reference)

1. Add a type parameter to `useSliderRoot` as well, e.g. https://github.com/mui/material-ui/blob/master/packages/mui-base/src/useSelect/useSelect.types.ts#L22
2. Export a type (from `useSliderRoot`) to use in `SliderRoot` as the generic, e.g. https://github.com/mui/material-ui/blob/master/packages/mui-base/src/useSelect/useSelect.types.ts#L14

Let me know if you need any help ~

mj12albert: @seloner See if this helps: https://github.com/mj12albert/base-ui/commits/slider-value-type/

seloner: I tried yesterday and I was not able to make it work.
Typescript was not correctly infering  the type.
I will check again today with your suggestion and update 👍

seloner: @mj12albert 
Are you sure your approach is working?
I created a minimal sandbox with the recommended approach but types are not working correctly 🤔 
https://stackblitz.com/edit/vitejs-vite-k75jr9vi?file=src%2FApp.tsx
Am I missing something? 
I have tried your fork also but it does not work 🤔

mj12albert: @seloner You are right ~ it "works" as in there are no errors but (similar to the now deprecated useSelect) we still need to cast some stuff in the end (https://github.com/mui/base-ui/pull/1333/commits/16ff230e0bea0cdd170dd8d9627b5b4d0af6b8a8), and _still_ the DX isn't good 😢 

With this approach you'd still need to manually specify this: https://github.com/mui/material-ui/blob/master/docs/data/base/components/select/UnstyledSelectMultiple.tsx#L25

RA is similar: https://react-spectrum.adobe.com/react-aria/Slider.html#controlled-value

Can't believe I didn't notice before but Radix's slider always takes arrays: https://github.com/radix-ui/website/blob/main/components/demos/Slider/css-modules/index.jsx#L7

So anyway, I'm ok with casting the callbacks like that to make it work, what do you think @michaldudak ?

(I'm pretty sure @vladmoroz would absolutely hate anything that requires users to touch generics)

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -59,8 +59,9 @@ const SliderRoot = React.forwardRef(function SliderRoot(
     min,
     minStepsBetweenValues,
     name: nameProp ?? '',
-    onValueChange: onValueChangeProp ?? NOOP,
-    onValueCommitted: onValueCommittedProp ?? NOOP,
+    onValueChange: (onValueChangeProp as useSliderRoot.Parameters['onValueChange']) ?? NOOP,
```
mj12albert, line 62: Testing in vscode locally, I noticed that when using `defaultValue` and `onValueChange` the type isn't `number`:
Screenshot 2025-01-23 at 3 51 04 PM

Using controlled mode (`value`) it's correct:
Screenshot 2025-01-23 at 3 55 12 PM

Uncontrolled range slider also seems correct:
Screenshot 2025-01-23 at 3 49 58 PM

michaldudak:
I added a couple of type tests using the infrastructure we use in Material UI and fixed the issue outlined in https://github.com/mui/base-ui/pull/1241#discussion_r1926521076

mj12albert (approved):
Thanks again everybody who helped with this 🙏

Issue #1246, opened 2024-12-30
Bundle size improvements


atomiks:
Inspecting the production bundle with `create-vite-app` using `@base-ui-components/react` reveals some areas of improvement:

- **Extra `prop-types` logic is included in the final bundle**: Even though `.propTypes` assignments are removed in prod, there's some extra code related to it when inspecting the bundle. We should remove `prop-types` entirely if possible in the final build, and only use it for documentation
- **Minified errors**: Context error messages aren’t minified for production, we can use a simple short/reusable message for production instead
- **Duplicated logic between Floating UI and Base UI**: Another version of `mergeReactProps` is used in `useInteractions`, so we should re-create `useInteractions` to use our own version instead of importing it. There is also duplicated logic between our internal `Composite` component and `useListNavigation`. There's also a duplicated version of `useId`
- **Needless destructuring and re-passing of variables**: we can pass the whole object in most cases, reducing unnecessary variable assignments. This is done for some components but not others
- **Avoid unnecessary dependencies**: importing only `Dialog` needlessly includes all anchor positioning logic since `useFloating()` brings it in. If `FloatingFocusManager` can accept `floatingRootContext` object, then that hook won’t need to be used. Importing only `Tooltip` needlessly includes the `tabbable` dependency and related focus guard logic since `FloatingPortal` brings it in. This matters less if you're using multiple components, however, since this logic will be brought in at some point anyway

oliviertassinari:
> 1. Even though .propTypes assignments are removed in prod, there's some extra code related to it when inspecting the bundle.

@atomiks What logic is left in production? 

> We should remove prop-types entirely if possible in the final build, and only use it for documentation

Did you mean `process.env.NODE_ENV`? Recently, I found two cases using our APIs that frustrated be because they didn't correctly use the prop-types:

- https://github.com/mui/material-ui/issues/43437#issuecomment-2504280775
- https://github.com/mui/mui-x/pull/16017

What I anticipate to be great DX is for us to have our own custom prop-types package: https://github.com/mui/material-ui/issues/43138.

> 2. Minified errors:

@Janpot recently iterated on this, helping a bit bring this in Base UI: https://github.com/mui/material-ui/pull/43904.

> 3. Duplicated logic between Floating UI and Base UI

👍  related to #454

> 4. Needless destructuring and re-passing of variables

@romgrk has been doing a lot of changes around this in the data grid. It's a performance killer too. 

---

If we want to use this issue as an umbrella, more ideas:

[...]

PR #1247, opened 2024-12-30
[core] Use local `useInteractions`

atomiks:
Part of #1246. This prevents another `mergeProps` utility from being included in the bundle as we reuse the one locally.


atomiks:
This hook has special properties inside Floating UI and can't be replicated locally

PR #1248, opened 2024-12-30
[POC] Minified context error message in prod

atomiks:
Part of #1246

### DIFF
--- a/packages/react/src/utils/productionContextError.ts
+++ b/packages/react/src/utils/productionContextError.ts
@@ -1,0 +1,3 @@
+export function productionContextError(): never {
+  throw new Error('Base UI: Invalid component tree');
+}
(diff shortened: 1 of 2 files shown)


packages/react/src/utils/productionContextError.ts
```diff
@@ -0,0 +1,3 @@
+export function productionContextError(): never {
+  throw new Error('Base UI: Invalid component tree');
+}
```
atomiks, line 3: React's minified errors link to a webpage that details the error, but this seems unnecessary for invalid component trees since it's certain to be encountered while developing beforehand

michaldudak:
This will surely shave off a few bytes but looks quite bad in source code. Have you considered using a similar solution to what's in Material UI? We have a Babel macro that minifies these messages.

atomiks:
> Have you considered using a similar solution to what's in Material UI? We have a Babel macro that minifies these messages.

Do you have a link/example of the source?

mj12albert:
> link/example of the source

Check out this PR https://github.com/mui/material-ui/pull/43904 😬

atomiks:
Replaced by #1463

PR #1249, opened 2024-12-30
[core] Refactor data passing

atomiks:
Part of #1246. Focuses just on anchored popups in this PR


packages/react/src/tooltip/root/TooltipRoot.tsx
```diff
@@ -13,77 +13,43 @@ import { PortalContext } from '../../portal/PortalContext';
...
+    trackCursorAxis = 'none',
+  } = props;
 
   const delayWithDefault = delay ?? OPEN_DELAY;
   const closeDelayWithDefault = closeDelay ?? 0;
 
-  const {
+  const tooltipRoot = useTooltipRoot({
+    ...props,
+    defaultOpen,
```
michaldudak, line 31: What's the point of passing `onOpenChange`, `open`, and other props explicitly if they are a part of `...props`?

michaldudak:
Some cases could be simplified even more (less destructuring) if we passed all the props to hooks as they are, without providing default values on a component level. I'm also thinking about reconsidering the context access pattern we use and having the hooks instead of components read context values. This way, we might not need to create new objects just to pass data to hooks as a single parameter.

packages/react/src/tooltip/root/TooltipRoot.tsx
atomiks: Yeah, you're right. Forgot that the linting allows unused variables (as they need to be destructured to be documented)

atomiks:
> Some cases could be simplified even more (less destructuring) if we passed all the props to hooks as they are, without providing default values on a component level.

Are you referring to `useAnchorPositioning` or other ones too?  A lot of the defaults are the same for that one, although a couple are different for each component (like `side`). Also, doesn't the documentation require the destructuring?

michaldudak:
> Are you referring to useAnchorPositioning or other ones too?

I mean all our hooks in general (especially the simpler ones). In cases like https://github.com/mui/base-ui/blob/master/packages/react/src/popover/title/PopoverTitle.tsx#L25, we might simplify passing parameters by passing the whole props object to `usePopoverTitle` and moving `usePopoverRootContext` to the hook. I see that the Select hooks already read context. The perf gains won't be huge, so it's likely not worth going through the whole codebase, but it could be something to consider in future components.

> Also, doesn't the documentation require the destructuring

Yup, we'd have to change the docgen if we wanted to do this.

packages/react/src/tooltip/root/TooltipRoot.tsx
atomiks: Well not anymore with our latest linting config:

> /tmp/base-ui/packages/react/src/popover/root/PopoverRoot.tsx
  16:11  error  'defaultOpen' is assigned a value but never used   @typescript-eslint/no-unused-vars
  16:24  error  'open' is assigned a value but never used          @typescript-eslint/no-unused-vars
  16:30  error  'onOpenChange' is assigned a value but never used  @typescript-eslint/no-unused-vars
...
✖ 12 problems (12 errors, 0 warnings)

michaldudak (approved):

PR #1250, opened 2024-12-31
[Select, Menu] Handle pseudo-element bounds in mouseup detection

### LINKED ISSUE #1232: [Menu, Select] Closes on `mouseup` while cursor is still on trigger if clickable bounds are larger than trigger rect
Note: `alignItemToTrigger` must be `false` for Select

On the docs the `Select` uses `::before` and `::after` pseudo-elements to make the clickable target larger than the actual highlighted trigger itself. This area is clickable, but the `mouseup` check can't "measure" this part since `getBoundingClientRect` measures the inner trigger part, and so it closes on mouseup unexpectedly if you click inside the extended area.

https://github.com/user-attachments/assets/d9b61779-e4b1-4021-8d62-32201fdaf5fd

https://github.com/mui/base-ui/blob/1d4206e80fc00eadb8bd9bd3f80ab34c4bbdf6f8/packages/react/src/select/trigger/useSelectTrigger.ts#L110

onehanddev:
Below changes were made for Select and Menu components for their use[Menu/Select]trigger hook -
- Add getPseudoElementPadding utility to measure ::before and ::after dimensions
- Update mouseup detection logic to include pseudo-element padding in clickable area
- Fix unexpected Select closing when clicking within extended pseudo-element bounds

Fixes #1232


onehanddev:
@atomiks I believe these changes might work better if we move the entire handleMouseUp and getPseudoElementPadding logic to a common place, as the behavior seems consistent. What do you think?

onehanddev:
CircleCI is failing JSDOM test, I guess rerunning it should the fix the issue.

packages/react/src/select/trigger/useSelectTrigger.ts
```diff
@@ -100,25 +100,48 @@ export function useSelectTrigger(
...
+                    parseInt(beforeStyles.width || '0', 10),
+                    parseInt(afterStyles.width || '0', 10),
+                    parseInt(beforeStyles.height || '0', 10),
+                    parseInt(afterStyles.height || '0', 10),
+                  ) / 2
+                : 0;
+
+              return padding;
+            }
+
```
atomiks, line 121: Thanks for your contribution!

This could be refactored into a utility function (and technically, parts of the rest of the function too) to reuse between Menu and Select.

- Given the padding can be asymmetric, this wouldn't actually measure it correctly I think, e.g. if `left` has `0px` while `right` has `20px`. Maybe instead of considering it as padding, you just check the max between the trigger rect width/height along with the pseudo elements' width/height
- `parseFloat(beforeStyles.width) || 0` is a simpler and more accurate function to use than `parseInt`

onehanddev: Thanks for the detailed review @atomiks 
I have made these changes except I was a bit skeptical about moving the final checks to a utility function (see: _technically, parts of the rest of the function too_) as it was abstracting the code a bit much, Please take a look at the updated code and let me know if you think it still needs to be more reusable.

atomiks: That's ok, and I see that determining asymmetry on a given axis might not be feasible, so it's ok to assume it's symmetric. 

However I still see the bug with the CSS modules selector on the docs, if I click on the bottom-left corner or right at the left edge, it still closes. Maybe floating point errors or something else?

onehanddev: Yes its indeed a minor floating point error, I assume it should be fixed with Math.round, let me try that and push the fix.

onehanddev: Can you please check the latest commit @atomiks Thanks !

atomiks: Seems to work correctly now

onehanddev: Ah, thats correct and I think the latest commit should fix this, please let me know if this still has some issues.

onehanddev: @atomiks Please do checkout the fix, and let me know if that works.

onehanddev: After several tries, I still don't find the issue which was there before the last commit, If thats not too much to ask for, can you please share a screen recording where you are encountering the same issue, That would be helpful. Thanks !
https://github.com/user-attachments/assets/92f308dc-a20f-43af-a211-d08d27d5884b

atomiks: For what it's worth, it doesn't happen with a single 'tap' on the trackpad, only a force click. I tried each one in this recording.

https://github.com/user-attachments/assets/2535e9e7-1898-400f-81fa-0942158a4f28

Also, I'll help debug this soon locally myself

atomiks: You might be able to get away with just adding `+1` and `-1` to the bounds manually, instead of using floor/ceil. I did this in an earlier PR to fix floating point errors.

onehanddev: Yeah, I am only using force click, but its interesting that I am still not able to see the issue anymore even though we both are on the same preview URL.
I am not sure if this would help -- but can you clear caches / try on incognito once or check if my latest changes are visible in sources ?

atomiks: Yeah I cleared cache, and incognito is the same

onehanddev: Just saw your earlier comment, Let me just add `+1 -1` to see if the issue fixes for you.

onehanddev: Just made above change. Feel free to recheck the preview URL.

atomiks: Debugged locally - we'll need `2` pixels, not `1`. In fact I remember needing to use `2px` for other things related to mouse events, and it works in all cases.

Screenshot 2025-01-06 at 10 56 49 pm

onehanddev: Okay, Done 👍

atomiks (approved):

PR #1285, opened 2025-01-05
[menu] Fixed Menu item not toggling highlighted prop when submenu popup is opened using keyboard

### LINKED ISSUE #1197: [menu] Submenu triggers retain `data-highlighted` when focus has moved to submenu popup using keyboard
Pointer usage is unaffected since the root popup takes the focus when moving the pointer off, but this doesn't happen with keyboard usage. 

Screenshot 2024-12-20 at 8 39 26 pm

A workaround appears to be adding `:focus` to `[data-highlighted]`, as real focus is used:

```css
&[data-highlighted]:focus {
  /* ... */
}
```

onehanddev:
Improve Menu Navigation and Fix Highlighting Issue

- Fixed bug where parent expandable menu item is getting highlighted along with the its sub menu with keyboard navigation.
- Improved navigation by resetting parent menu's highlight state when entering a submenu.

Fixes #1197


packages/react/src/menu/submenu-trigger/useMenuSubmenuTrigger.ts
```diff
@@ -38,11 +44,20 @@ export function useMenuSubmenuTrigger(
...
           ...externalProps,
+          onKeyDown: (event: MenuKeyboardEvent) => {
+            if (event.key === 'ArrowRight' && highlighted) {
+              // Clear parent menu's highlight state when entering submenu
+              // This prevents multiple highlighted items across menu levels
+              setActiveIndex(null);
+            }
+            onKeyDown?.(event);
+            externalProps?.onKeyDown?.(event);
+          },
```
atomiks, line 55: Using the `mergeReactProps` util will work better here

packages/react/src/menu/submenu-trigger/useMenuSubmenuTrigger.ts
```diff
@@ -38,11 +44,20 @@ export function useMenuSubmenuTrigger(
         ...getMenuItemProps({
           'aria-haspopup': 'menu' as const,
           ...externalProps,
+          onKeyDown: (event: MenuKeyboardEvent) => {
+            if (event.key === 'ArrowRight' && highlighted) {
```
atomiks, line 48: There's also regular `onClick` to check for

onehanddev: Added, Please check.

packages/react/src/menu/submenu-trigger/useMenuSubmenuTrigger.ts
onehanddev: Done, Please check.

packages/react/src/menu/submenu-trigger/useMenuSubmenuTrigger.ts
```diff
@@ -82,6 +103,21 @@ export namespace useSubmenuTrigger {
...
+    /**
+     * Callback function that is triggered on key down events.
+     * @param event - The keyboard event that triggered the callback.
+     */
+    onKeyDown?: (event: MenuKeyboardEvent) => void;
+    /**
+     * Callback function that is triggered on click events.
+     * @param event - The click event that triggered the callback.
+     */
+    onClick?: (event: React.MouseEvent) => void;
```
atomiks, line 115: These aren't necessary with `mergeReactProps`, it will automatically call them from the external props

onehanddev: Missed it, Its being removed in latest commit.

packages/react/src/menu/submenu-trigger/useMenuSubmenuTrigger.ts
```diff
@@ -34,15 +40,22 @@ export function useMenuSubmenuTrigger(
 
   const getRootProps = React.useCallback(
     (externalProps?: GenericHTMLProps) => {
-      return {
+      return mergeReactProps(externalProps, {
         ...getMenuItemProps({
           'aria-haspopup': 'menu' as const,
-          ...externalProps,
+          onKeyDown: (event: MenuKeyboardEvent) => {
+            if (event.key === 'ArrowRight' && highlighted) {
```
atomiks, line 47: This works well overall, but in `rtl` mode, it's `ArrowLeft` instead of `ArrowRight`

I wonder if it's more robust to check for focus/blur, or the submenu opening? Explicitly checking for opening actions like onClick/ArrowLeft/Right can work ok, but there's edge cases to consider since it's not as general

onehanddev: I see, Is it fine to just check the direction and use `ArrowLeft | ArrowRight` ? What are the other edge cases which could potentially not respect these events ?

onehanddev: Yes, I agree. I debugged further and found that we get the latest open state for the submenu in the useMenuRoot hook. Even when we use it in useMenuSubmenuTrigger, it only works for keyboard events, not hover ones.

onehanddev: I implemented the direction check using useDirection and toggled the keyboard event key accordingly.

Apologies for the commit mess in this PR—it happened while pulling the latest changes and resolving conflicts. To keep things clean, I created a fresh PR here: [#1310](https://github.com/mui/base-ui/pull/1310).

Please have a look at it when you get a chance, @atomiks. Thanks!

PR #1288, opened 2025-01-06
[tabs] Expose width/height state in tabs indicator

### LINKED ISSUE #1286: [Tabs] Missing width/height state in `Tabs.Indicator` render callback
# Feature request

## Summary

The `Tabs.Indicator` element exposes the following CSS vars in order to let users style/animate the element:

```
--active-tab-top
--active-tab-right
--active-tab-bottom
--active-tab-left

--active-tab-width
--active-tab-height
```

The `state` argument in the `render` callback of the `Tabs.Indicator` component includes a `selectedTabPosition` object which contains `{top, right, bottom , left}`. These coordinates correspond to the first four CSS vars, but `--active-tab-width` and `--active-tab-height` have no corresponding state keys.

## Motivation

Some users might need to access the width/height of the selected tab in JS. For example, it's useful if someone is relying on [Motion](https://motion.dev/) to animate the tabs indicator. See this example: https://codesandbox.io/p/sandbox/black-platform-4zphr8?workspaceId=ws_HE9JRzFW36Y2kpBoSwdvZ5. The coordinates of the tabs indicator can be animated using Motion but not the width because width/height is not exposed in the `render` callback state.

Further context: https://discord.com/channels/1287292451308048406/1287292451308048409/1325583630956888135

aarongarciah:
Closes https://github.com/mui/base-ui/issues/1286

### DIFF
--- a/packages/react/src/tabs/indicator/TabsIndicator.test.tsx
+++ b/packages/react/src/tabs/indicator/TabsIndicator.test.tsx
@@ -57,8 +57,8 @@ describe('<Tabs.Indicator />', () => {
         right: tabRight,
         top: tabTop,
         bottom: tabBottom,
-        width,
-        height,
+        width: tabWidth,
+        height: tabHeight,
       } = activeTab.getBoundingClientRect();
 
       const relativeLeft = tabLeft - listLeft;
@@ -78,8 +78,8 @@ describe('<Tabs.Indicator />', () => {
       assertSize(actualRight, relativeRight);
       assertSize(actualTop, relativeTop);
       assertSize(actualBottom, relativeBottom);
-      assertSize(actualWidth, width);
-      assertSize(actualHeight, height);
+      assertSize(actualWidth, tabWidth);
+      assertSize(actualHeight, tabHeight);
     }
 
     it('should set CSS variables corresponding to the active tab', async () => {
--- a/packages/react/src/tabs/indicator/TabsIndicator.test.tsx
+++ b/packages/react/src/tabs/indicator/TabsIndicator.test.tsx
@@ -57,6 +57,8 @@ describe('<Tabs.Indicator />', () => {
         right: tabRight,
         top: tabTop,
         bottom: tabBottom,
+        width,
+        height,
       } = activeTab.getBoundingClientRect();
 
       const relativeLeft = tabLeft - listLeft;
@@ -69,11 +71,15 @@ describe('<Tabs.Indicator />', () => {
       const actualRight = bubbleComputedStyle.getPropertyValue('--active-tab-right');
       const actualTop = bubbleComputedStyle.getPropertyValue('--active-tab-top');
       const actualBottom = bubbleComputedStyle.getPropertyValue('--active-tab-bottom');
+      const actualWidth = bubbleComputedStyle.getPropertyValue('--active-tab-width');
+      const actualHeight = bubbleComputedStyle.getPropertyValue('--active-tab-height');
 
       assertSize(actualLeft, relativeLeft);
       assertSize(actualRight, relativeRight);
       assertSize(actualTop, relativeTop);
       assertSize(actualBottom, relativeBottom);
+      assertSize(actualWidth, width);
+      assertSize(actualHeight, height);
     }
 
     it('should set CSS variables corresponding to the active tab', async () => {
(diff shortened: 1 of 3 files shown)


packages/react/src/tabs/indicator/TabsIndicator.test.tsx
aarongarciah, line 1: Should a test be added to check the returned state in the `render` callback?

packages/react/src/tabs/indicator/TabsIndicator.test.tsx
```diff
@@ -69,11 +71,15 @@ describe('<Tabs.Indicator />', () => {
       const actualRight = bubbleComputedStyle.getPropertyValue('--active-tab-right');
       const actualTop = bubbleComputedStyle.getPropertyValue('--active-tab-top');
       const actualBottom = bubbleComputedStyle.getPropertyValue('--active-tab-bottom');
+      const actualWidth = bubbleComputedStyle.getPropertyValue('--active-tab-width');
+      const actualHeight = bubbleComputedStyle.getPropertyValue('--active-tab-height');
```
aarongarciah, line 75: These were missing.

packages/react/src/tabs/indicator/TabsIndicator.test.tsx
michaldudak: It won't hurt, for sure. Feel free to add it if you have time. Otherwise, we'll do it.

packages/react/src/tabs/indicator/TabsIndicator.tsx
```diff
@@ -44,9 +48,10 @@ const TabsIndicator = React.forwardRef<HTMLSpanElement, TabsIndicator.Props>(
       () => ({
         orientation,
         selectedTabPosition,
+        selectedTabSize,
```
michaldudak, line 51: All fields from `state` are automatically translated into data attributes. This is not really needed (and incorrect in its current form, as it's serialized to `[object Object]`), so it must be explicitly excluded from style hooks (via `customStyleHookMapping` in line 67, similarly to `selectedTabPosition`).

packages/react/src/tabs/indicator/useTabsIndicator.ts
```diff
@@ -124,6 +136,11 @@ export interface ActiveTabPosition {
   bottom: number;
 }
 
+export interface ActiveTabSize {
```
aarongarciah, line 139: I was in doubt if "dimensions" was a better name than "size". If you have a preferred termn, let me know.

packages/react/src/tabs/indicator/TabsIndicator.tsx
aarongarciah: Fixed https://github.com/mui/base-ui/pull/1288/commits/8736ded65cf3b33c83be3b95ac9933887253106b

packages/react/src/tabs/indicator/TabsIndicator.test.tsx
aarongarciah: I was about to add a couple of tests like the ones below but I think this should ideally be automated at the conformance test level. I'll leave it out of this PR if you are ok with that.

```tsx
it('should pass state to render callback', async () => {
  const renderCb = spy((props, _state) => <span {...props} />);
  const { getByTestId } = await render(
    <Tabs.Root value={2}>
      <Tabs.List>
        <Tabs.Tab value={1}>One</Tabs.Tab>
        <Tabs.Tab value={2}>Two</Tabs.Tab>
        <Tabs.Tab value={3}>Three</Tabs.Tab>
        <Tabs.Indicator data-testid="bubble" render={renderCb} />
      </Tabs.List>
    </Tabs.Root>,
  );

  const bubble = getByTestId('bubble');

  assertState(bubble, renderCb.lastCall.args[1]);
});

it('should pass state to className callback', async () => {
  const classNameCb = spy();
  const { getByTestId } = await render(
    <Tabs.Root value={2}>
      <Tabs.List>
        <Tabs.Tab value={1}>One</Tabs.Tab>
        <Tabs.Tab value={2}>Two</Tabs.Tab>
        <Tabs.Tab value={3}>Three</Tabs.Tab>
        <Tabs.Indicator data-testid="bubble" className={classNameCb} />
      </Tabs.List>
    </Tabs.Root>,
  );

  const bubble = getByTestId('bubble');

  assertState(bubble, classNameCb.lastCall.args[0]);
});
```

packages/react/src/tabs/indicator/useTabsIndicator.ts
michaldudak: Size is ok (it's shorter and equally descriptive imo).

michaldudak (approved):

Issue #1181, opened 2024-12-19
[docs] Headings should be linkable


atomiks:
# Docs feedback

It's not possible to click headings to link to subsections of a page. Headings should have `<a>` tags around them with id hashes.

colmtuite:
You can click the quick nav link

atomiks:
> You can click the quick nav link

I forgot that was possible, doesn't seem easily discoverable + doesn't work on narrow viewports/mobile?

imMatheus:
Still feel like the headers should be linkable IMO, also, would be fairly easy to add no? I could make a quick PR for it

mnajdova:
I noticed this and forgot to mention it. The most common use-case for me is to click on the header and share a link with someone. I am already looking at the title, finding it again in the table of content seems like an overkill.

mj12albert:
> The most common use-case for me is to click on the header and share a link with someone

This is also a big use-case for me, e.g. adding links to GH comments while doing support on-the-go

I think API docs generally have this, examples of some docs I looked at recently:

[Radix docs](https://www.radix-ui.com/primitives/docs/components/select#root): 

https://github.com/user-attachments/assets/45611ec1-8910-4218-8ae2-1fc7e6d84de4

[Ariakit docs](https://ariakit.org/reference/select): 

The subcomponents in the anatomy are links so that counts? I like that individual props are linkable too

https://github.com/user-attachments/assets/10ba19db-e943-4c0d-ac95-1c0b614444c3

[RA docs](https://react-spectrum.adobe.com/react-aria/Select.html#anatomy):

The links exist but they are invisible on mobile...

https://github.com/user-attachments/assets/a693b71a-dab0-44e6-91f5-6fce3d80a08b

[Fastify](https://fastify.dev/docs/latest/Reference/Hooks/#onregister):

https://github.com/user-attachments/assets/755cade7-9f7c-4292-a576-15ba971a8201

PR #1302, opened 2025-01-08
[docs] Link h2 and h3 in docs

mj12albert:
Closes https://github.com/mui/base-ui/issues/1181


mj12albert:
Dunno if it's a bit much, but I made the icon always show for touch devices, none of [the examples](https://github.com/mui/base-ui/issues/1181#issuecomment-2577000730) I looked at looked obviously like links to me

https://github.com/user-attachments/assets/ed11e09c-f477-4834-86eb-c3701a9ee4d3

colmtuite:
Closing this for now because we'll need some time to consider whether to add this, and if so, how to handle it.

It's possible that only the icon should be inside `<a>`. It will also need `text-underline-offset`, different colors, and a larger icon etc. It will be better to batch docs improvements.

Issue #1208, opened 2024-12-20
[Dialog, Popover] Provide a callback for open/close completion


vladmoroz:
A prop like `onOpenChangeComplete` is needed for callbacks that need to run when a component gets fully hidden with all animations done, e.g. to reset its internal state. Doing that in `onOpenChange` would yield a flash of contents.

atomiks:
I think `onHidden` makes sense as a prop. There isn't value in knowing when `open` has finished animating in, only animating out?

The `State` of these elements can contain `hidden`, which is `!mounted` internally, which represents when it's unmounted from the DOM or has the `hidden` prop applied with `keepMounted`.

PR #1305, opened 2025-01-08
[popups] `onOpenChangeComplete` prop

atomiks:
Closes https://github.com/mui/base-ui/issues/1208

New alternative to https://github.com/mui/base-ui/pull/1235 to support both symmetric open/close cases, which has a more intuitive API.

- Reworks `useAfterExitAnimation` to support both open/close animations
- `onOpenChangeComplete` is called with `open: boolean`


packages/react/src/utils/useOpenChangeComplete.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useAnimationsFinished } from './useAnimationsFinished';
+import { useEnhancedEffect } from './useEnhancedEffect';
+import { useEventCallback } from './useEventCallback';
+import { useLatestRef } from './useLatestRef';
+
+/**
+ * Calls the provided function when the CSS open animation or transition completes.
+ */
+export function useOpenChangeComplete(parameters: useOpenChangeComplete.Parameters) {
+  const { open, change = 'close', ref, onComplete: onCompleteParam } = parameters;
```
mj12albert, line 11: It may be a bit more readable if this `change` param didn't have a default, but it probably doesn't matter outside of looking at this code in Github

anyway I tested `onOpenChangeComplete` in a bunch of our hero demos and it works well (they all use CSS transitions) 👍 

CC @michaldudak to review the implementation as well 🙏

mj12albert (approved):

packages/react/src/utils/useOpenChangeComplete.tsx
atomiks: Was a small optimization since it's called in many of the root components

packages/react/src/utils/useOpenChangeComplete.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from 'react';
+import { useAnimationsFinished } from './useAnimationsFinished';
+import { useEnhancedEffect } from './useEnhancedEffect';
+import { useEventCallback } from './useEventCallback';
+import { useLatestRef } from './useLatestRef';
+
+/**
+ * Calls the provided function when the CSS open animation or transition completes.
```
michaldudak, line 8:
```suggestion
 * Calls the provided function when the CSS open/close animation or transition completes.
```

packages/react/src/tooltip/popup/TooltipPopup.tsx
```diff
@@ -30,10 +31,19 @@ const TooltipPopup = React.forwardRef(function TooltipPopup(
 ) {
   const { className, render, ...otherProps } = props;
 
-  const { open, instantType, transitionStatus, getRootPopupProps, popupRef } =
+  const { open, instantType, transitionStatus, getRootPopupProps, popupRef, onOpenChangeComplete } =
     useTooltipRootContext();
   const { side, align } = useTooltipPositionerContext();
 
+  useOpenChangeComplete({
```
michaldudak, line 38: Why this can't be handled in useTooltipRoot, as the close animation?

packages/react/src/utils/useOpenChangeComplete.tsx
michaldudak: I agree with Albert, not having this parameter present when calling the hook makes it more confusing: we're calling `useOpenChangeComplete`, but its `onComplete` handles just the closing part.
TBH the `change` prop itself isn't super clear and I had to spend some time to understand what it's for. Would it be possible to remove it and make the hook handle both cases at the same time? The second parameter to `runOnceAnimationsFinish` could perhaps be derived from `open`?

packages/react/src/tooltip/popup/TooltipPopup.tsx
atomiks: The timing of the ref assignment means it needs to be called in the popup component to be accessible

packages/react/src/utils/useOpenChangeComplete.tsx
atomiks: Changed it so the caller with `onComplete` specified needs to check the change direction

packages/react/src/alert-dialog/popup/AlertDialogPopup.tsx
```diff
@@ -63,9 +63,10 @@ const AlertDialogPopup = React.forwardRef(function AlertDialogPopup(
   useOpenChangeComplete({
     open,
     ref: popupRef,
-    change: 'open',
     onComplete() {
```
michaldudak, line 66: The most straightforward and idiomatic way to use this would be to have a parameter on the `onComplete` callback that determines if the popup was opened or closed.

atomiks: I thought of that, but it's not actually necessary because you can just use the outer variable that's passed as `open`. The `useEventCallback` ensures it's not stale.

Issue #201, opened 2024-03-17
[core] Add `test_bundle_size_monitor` back


oliviertassinari:
Continuous bundle size monitoring, same as https://github.com/mui/mui-x/issues/5550, it's important.

mj12albert:
Here's a more recent example:
- https://github.com/mui/material-ui/pull/45398#issuecomment-2679520233
- https://mui-dashboard.netlify.app/size-comparison?circleCIBuildNumber=799816&baseRef=v6.x&baseCommit=3bfa70e29b3521f746bbe75d1ee36a31e6a840d4&prNumber=45398

I wonder if there any challenges in measuring it like this, e.g. mentioned in this [comment](https://github.com/mui/base-ui/issues/602), what do you think? @michaldudak @atomiks

michaldudak:
Implemented in #1791

PR #1307, opened 2025-01-08
[core] Remove dangerfile

michaldudak:
Replaced custom danger-based site deployment notifications with the one coming from Netlify for better DX.

Before:
image

After:
image


mnajdova (approved):

oliviertassinari:
> Replaced custom danger-based site deployment notifications with the one coming from Netlify for better DX.

I'm not following, what's the advantage? This looks like the DX regressed here. The problems with this approach:

- Now we can't implement https://github.com/mui/mui-public/issues/260.
- Now it takes more vertical space to scroll.
- Now contributors will have to silence two bot spammy notifications (Netlify + Danger). We need the danger notifications for https://github.com/mui/base-ui/issues/201. And hopefully, we can implement performance regression tests too.

I have preventively removed the Netlify comment, assuming the case is convincing and we will take the step backward. But maybe I'm missing something, and the points I listed aren't less of a pain than the reason behind this change.

michaldudak:
The main advantage for me is having the deployment URL available as a QR code. I'm sure it would be possible to implement it using danger, but this was the simplest way. We currently don't use danger for anything else. Could you please re-enable Netlify comments? When we implement any new features with danger, we can come back to it.

oliviertassinari:
> available as a QR code

Ok, restored, and issue created for where we are heading: https://github.com/mui/mui-public/issues/261.

oliviertassinari:
If we move to Vercel to host the docs, this issue would make sense too, e.g. https://github.com/chakra-ui/chakra-ui/pull/9470#issuecomment-2590409609 the QR code is not there.

PR #1310, opened 2025-01-08
[menu] Fixed Menu item not toggling highlighted prop when submenu popup is opened using keyboard

### LINKED ISSUE #1197: [menu] Submenu triggers retain `data-highlighted` when focus has moved to submenu popup using keyboard

onehanddev:
Improve Menu Navigation and Fix Highlighting Issue

- Fixed bug where parent expandable menu item is getting highlighted along with the its sub menu with keyboard navigation.
- Improved navigation by resetting parent menu's highlight state when entering a submenu.

Fixes #1197


onehanddev:
@atomiks Please disregard the [older PR](https://github.com/mui/base-ui/pull/1285) and review this one instead.

atomiks (dismissed):
A test would be helpful to prevent future regressions

onehanddev:
> A test would be helpful to prevent future regressions

Unrelated to this PR -- but when I try to run the tests using `pnpm t [file-name]` I am getting below error -
image

Tried several other commands but it doesn't seem to work, @atomiks It would be great If you can suggest how to correctly run the tests. Thanks !

onehanddev:
> A test would be helpful to prevent future regressions

@atomiks Just did, Please check, Thanks !

packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+      const submenuItems = await screen.findAllByRole('menuitem');
+      const submenuItem1 = submenuItems.find((item) => item.textContent === '2.1');
+
+      await waitFor(() => {
+        expect(submenuItem1).toHaveFocus();
+      });
+
+      submenuItems.forEach((item) => {
+        if (item === submenuItem1) {
+          expect(item).to.have.attribute('tabindex', '0');
```
atomiks, line 62: Small nit: `data-highlighted` is the attribute used for styling so it'd be better to test that, even if they're functionally the same

onehanddev: @atomiks Done 👍

atomiks:
It looks like setting the active index to `null`  has a problem with tabbing. When opening the submenu, tabbing backwards should tab to the submenu trigger first, and then the root trigger. Tabbing forward from the root trigger should focus the submenu trigger and then the last known highlighted item in the submenu popup.  (This is what happens currently)

But this no longer happens since `tabindex=0` is removed with this change when setting the `activeIndex` to `null`

onehanddev:
> It looks like setting the active index to `null` has a problem with tabbing. When opening the submenu, tabbing backwards should tab to the submenu trigger first, and then the root trigger. Tabbing forward from the root trigger should focus the submenu trigger and then the last known highlighted item in the submenu popup. (This is what happens currently)
> 
> But this no longer happens since `tabindex=0` is removed with this change when setting the `activeIndex` to `null`

@atomiks That was a good find, I have made the change to update the tabindex while nulling the activeIndex, Please checkout my latest commit, Thanks !

packages/react/src/menu/submenu-trigger/useMenuSubmenuTrigger.ts
```diff
@@ -32,17 +41,37 @@ export function useMenuSubmenuTrigger(
...
     (externalProps?: GenericHTMLProps) => {
-      return {
+      const openKey = direction === 'rtl' ? 'ArrowLeft' : 'ArrowRight';
+      const handleOpenSubmenu = () => {
+        if (highlighted) {
+          setActiveIndex(null);
+          setIsSubmenuOpen(true);
+        }
+      };
+      return mergeReactProps(externalProps, {
```
atomiks, line 57: `mergeReactProps<'div'>(...)` will be able to type the `onKeyDown` event automatically

packages/react/src/menu/submenu-trigger/useMenuSubmenuTrigger.ts
```diff
@@ -32,17 +41,37 @@ export function useMenuSubmenuTrigger(
 
   const menuTriggerRef = useForkRef(menuItemRef, setTriggerElement);
 
+  const direction = useDirection();
+
+  const [isSubmenuOpen, setIsSubmenuOpen] = useState(false);
```
atomiks, line 46: How does this state get reset back to `false` when `keepMounted=true` is used?

onehanddev: Ah, at the moment, it seems we need to access the ultimate Menu.root open property or state. Based on this, we can reset the isSubmenuOpen to false. However, this could be challenging given that the menu could be nested any number of times. Do you have any approach or solution in mind for handling this?

atomiks: Setting state in render checking either `!open` (or `!mounted`, if it works better with unmounting animations) should work:
```js
if (!open && isSubmenuOpen) {
  setIsSubmenuOpen(false);
}
```

atomiks:
@onehanddev I understand you may not have had time to finish this up, so I opened a new PR to finish it

Issue #1312, opened 2025-01-09
[useButton] When `focusableWhenDisabled` the button can still be clicked or pressed


mj12albert:
Regression from https://github.com/mui/base-ui/pull/1177

Before we were manually calling external handlers, but using `mergeReactProps` external handlers are called before our handlers

Affected: collapsible trigger, menu items (regular, checkbox, radio), select items, tabs (`TabsTab`)

PR #1313, opened 2025-01-09
[useButton] Fix `focusableWhenDisabled` components

mj12albert:
Using `mergeReactProps`, our event handlers can't (easily) prevent external handlers since external ones are called first, it seems best to just bypass mergeReactProps here and manually call the external `onClick`.

Additionally found these issues along the way and fixed them:
- `<Menu.RadioGroup disabled />` didn't work
- `<Select.Trigger disabled />` didn't work

Closes https://github.com/mui/base-ui/issues/1312

### DIFF
--- a/packages/react/src/menu/radio-group/MenuRadioGroup.tsx
+++ b/packages/react/src/menu/radio-group/MenuRadioGroup.tsx
@@ -1,13 +1,11 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import { MenuRadioGroupContext } from './MenuRadioGroupContext';
+import { NOOP } from '../../utils/noop';
 import { BaseUIComponentProps } from '../../utils/types';
 import { useComponentRenderer } from '../../utils/useComponentRenderer';
 import { useControlled } from '../../utils/useControlled';
 import { useEventCallback } from '../../utils/useEventCallback';
-
-const EMPTY_OBJECT = {};
-const NOOP = () => {};
 
 /**
  * Groups related radio items.
@@ -22,7 +20,8 @@ const MenuRadioGroup = React.forwardRef(function MenuRadioGroup(
     className,
     value: valueProp,
     defaultValue,
-    onValueChange: onValueChangeProp = NOOP,
+    onValueChange: onValueChangeProp,
+    disabled = false,
     ...other
   } = props;
 
@@ -32,7 +31,7 @@ const MenuRadioGroup = React.forwardRef(function MenuRadioGroup(
     name: 'MenuRadioGroup',
   });
 
-  const onValueChange = useEventCallback(onValueChangeProp);
+  const onValueChange = useEventCallback(onValueChangeProp ?? NOOP);
 
   const setValue = React.useCallback(
     (newValue: any, event: Event) => {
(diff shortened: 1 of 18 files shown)


mj12albert:
`<Select.Trigger disabled />` doesn't seem to work, though it's not related to focusableWhenDisabled

packages/react/src/select/item/SelectItem.test.tsx
```diff
@@ -117,6 +117,29 @@ describe('<Select.Item />', () => {
...
+            </Select.Popup>
+          </Select.Positioner>
+        </Select.Portal>
+      </Select.Root>,
+    );
+
+    const item = screen.getByText('two');
+    act(() => item.focus());
+    expect(item).toHaveFocus();
+  });
```
atomiks, line 141: `Select` should now follow `Menu` with `useListNavigation`'s `disabledIndices`  prop

packages/react/src/use-button/useButton.test.tsx
```diff
@@ -3,12 +3,59 @@ import { expect } from 'chai';
...
+      const button = getByRole('button');
+
+      act(() => button.focus());
+      expect(button).toHaveFocus();
+
+      fireEvent.keyDown(button, { key: 'Enter' });
+      expect(handleKeyDown.callCount).to.equal(1);
+      expect(handleClick.callCount).to.equal(0);
+
+      fireEvent.keyDown(button, { key: 'Space' });
```
michaldudak, line 48: Space normally activates the button on keyup

atomiks:
> <Select.Trigger disabled /> doesn't seem to work, though it's not related to focusableWhenDisabled

I think this could also be fixed in this PR. The interaction hooks like `useClick` only check for `readOnly` for the `enabled` state, but not `disabled` like Menu does

packages/react/src/menu/radio-group/MenuRadioGroup.tsx
```diff
@@ -92,9 +95,17 @@ namespace MenuRadioGroup {
      * @default () => {}
      */
     onValueChange?: (value: any, event: Event) => void;
+    /**
+     * Whether the component should ignore user interaction.
+     *
+     * @default false
+     */
+    disabled?: boolean;
```
mj12albert, line 103: Added back `<Menu.RadioGroup disabled />` for consistency with regular `RadioGroup`

mj12albert:
> The interaction hooks like `useClick` only check for `readOnly` for the `enabled` state, but not `disabled` like Menu does

Fixed, it seems ok to have `listNavigation` and `typeahead` ignore a disabled trigger (not disabled Select.Root)

packages/react/src/select/item/SelectItem.test.tsx
```diff
@@ -117,6 +117,29 @@ describe('<Select.Item />', () => {
     });
   });
 
+  it('should focus disabled items', async () => {
```
michaldudak, line 120: It seems that hovering over disabled options doesn't focus them (unlike menu items)

mj12albert: I found that SelectItem inlines `pointer-events: 'none'` when disabled which causes this, is it safe to handle it in the handlers instead like this @atomiks ? https://github.com/mui/base-ui/pull/1313/commits/52838ad058107cf8bd80bfa94b55c42c520bed43

atomiks: @mj12albert yes, should be

atomiks (approved):

PR #1314, opened 2025-01-09
[test] Use vitest's `userEvent` implementation in browser tests

michaldudak:
Vitest docs don't recommend using React Testing Library's `userEvent` in browser tests (see the warning box in https://vitest.dev/guide/browser/#examples).

This PR changes the imported library depending on the environment in which the tests run.


packages/react/src/accordion/root/AccordionRoot.test.tsx
```diff
@@ -65,15 +65,15 @@ describe('<Accordion.Root />', () => {
       expect(trigger).to.have.attribute('aria-expanded', 'false');
       expect(queryByText(PANEL_CONTENT_1)).to.equal(null);
 
-      await user.pointer({ keys: '[MouseLeft]', target: trigger });
+      await user.click(trigger);
```
michaldudak, line 68: Vitest doesn't have the `pointer` method but clicks should work the same way in these cases

packages/react/src/dialog/popup/DialogPopup.test.tsx
```diff
@@ -282,27 +282,25 @@ describe('<Dialog.Popup />', () => {
...
+          <Dialog.Root modal={false}>
+            <Dialog.Trigger>Trigger 0</Dialog.Trigger>
+            <Dialog.Portal>
+              <Dialog.Popup data-testid="popup0">
+                {showNested && (
+                  <Dialog.Root modal={false}>
+                    <Dialog.Trigger>Trigger 1</Dialog.Trigger>
+                    <Dialog.Portal>
+                      <Dialog.Popup data-testid="popup1">
+                        <button onClick={() => setShowNested(!showNested)}>toggle</button>
```
michaldudak, line 294: Vitest functions are more strict about visibility and accessibility of the element to interact with.

packages/react/test/createRenderer.ts
```diff
@@ -23,13 +30,41 @@ export function createRenderer(globalOptions?: CreateRendererOptions): BaseUITes
...
-    act(async () => {
+  const render = async (element: React.ReactElement<DataAttributes>, options?: RenderOptions) => {
+    const renderResult = await act(async () => {
       const result = await originalRender(element, options);
       await flushMicrotasks();
       return result;
     });
 
+    if (!isJSDOM) {
+      const { userEvent: vitestUserEvent } = await import('@vitest/browser/context');
```
michaldudak, line 41: This must be imported dynamically as the static import fails when executed outside of the browser environment.

packages/react/test/createRenderer.ts
```diff
@@ -23,13 +30,41 @@ export function createRenderer(globalOptions?: CreateRendererOptions): BaseUITes
...
+      const originalUser = vitestUserEvent.setup();
+
+      const proxyHandler: ProxyHandler<any> = {
+        get: (target, prop) => {
+          if (typeof target[prop] !== 'function') {
+            return target[prop];
+          }
+
+          return async (...args: unknown[]) => {
+            await act(async () => target[prop](...args));
```
michaldudak, line 52: This isn't ideal. Perhaps we could change the implementation of `act` to no-op when running in the browser and get rid of this.

PR #1318, opened 2025-01-10
[tabs] Fix tabs activating incorrectly on non-primary button clicks

### LINKED ISSUE #1317: [Tabs] `Tab` activates when clicking non-primary mouse button
# Bug report

## Current behavior

Tabs switch on any mousedown / pointerdown on the Tabs.Tab component.

## Expected behavior

Presumably this should only happen on mousedown / pointerdown being the main button not also right or middle mouse click and such.

## Reproducible example

https://master--base-ui.netlify.app/react/components/tabs

Use right or middle mouse button on the Tabs.Tab component

## Base UI version

[1.0.0-alpha.4](https://www.npmjs.com/package/@base-ui-components/react)

Brave

## Which OS are you using?

Windows

None

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1317

### DIFF
--- a/packages/react/src/tabs/root/TabsRoot.test.tsx
+++ b/packages/react/src/tabs/root/TabsRoot.test.tsx
@@ -168,6 +168,23 @@ describe('<Tabs.Root />', () => {
   });
 
   describe('prop: onValueChange', () => {
+    it('should call onValueChange on pointerdown', async () => {
+      const handleChange = spy();
+      const handlePointerDown = spy();
+      const { getAllByRole, user } = await render(
+        <Tabs.Root value={0} onValueChange={handleChange}>
+          <Tabs.List>
+            <Tabs.Tab value={0} />
+            <Tabs.Tab value={1} onPointerDown={handlePointerDown} />
+          </Tabs.List>
+        </Tabs.Root>,
+      );
+
+      await user.pointer({ keys: '[MouseLeft>]', target: getAllByRole('tab')[1] });
+      expect(handleChange.callCount).to.equal(1);
+      expect(handlePointerDown.callCount).to.equal(1);
+    });
+
     it('should call onValueChange when clicking', async () => {
       const handleChange = spy();
       const { getAllByRole } = await render(
@@ -184,7 +201,7 @@ describe('<Tabs.Root />', () => {
       expect(handleChange.firstCall.args[0]).to.equal(1);
     });
 
-    it('should not call onValueChange on non-primary button clicks', async () => {
+    it('should not call onValueChange on non-main button clicks', async () => {
       const handleChange = spy();
       const { getAllByRole } = await render(
         <Tabs.Root value={0} onValueChange={handleChange}>
--- a/packages/react/src/tabs/root/TabsRoot.test.tsx
+++ b/packages/react/src/tabs/root/TabsRoot.test.tsx
@@ -184,6 +184,21 @@ describe('<Tabs.Root />', () => {
       expect(handleChange.firstCall.args[0]).to.equal(1);
     });
 
+    it('should not call onValueChange on non-primary button clicks', async () => {
+      const handleChange = spy();
+      const { getAllByRole } = await render(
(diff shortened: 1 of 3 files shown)


packages/react/src/tabs/tab/useTabsTab.ts
```diff
@@ -75,32 +76,56 @@ function useTabsTab(parameters: useTabsTab.Parameters): useTabsTab.ReturnValue {
...
+            }
+
+            isPressingRef.current = true;
+
+            function handlePointerUp() {
+              isPressingRef.current = false;
+              isPrimaryButtonRef.current = false;
+            }
+
+            if (!event.button || event.button === 0) {
```
oliviertassinari, line 120: We use

```suggestion
            if (event.button === 0) {
```

in all Material UI codebase I have never seen an issue with `event.button` being missing. Can't we reduce bundle size here?

packages/react/src/tabs/tab/useTabsTab.ts
```diff
@@ -75,32 +76,56 @@ function useTabsTab(parameters: useTabsTab.Parameters): useTabsTab.ReturnValue {
...
+          },
+          onFocus(event) {
+            if (!activateOnFocus || selected) {
+              return;
+            }
+
+            if (!isPressingRef.current || (isPressingRef.current && isPrimaryButtonRef.current)) {
               onTabActivation(tabValue, event.nativeEvent);
-            },
-            onFocus(event) {
```
oliviertassinari, line 92: Isn't that broken in the first place?

Why does this activate on mouse down? Isn't this missing the point of why onClick exists in the first place: cancellability?

elieven: John Carmack had a good argument for acting on press wherever possible in that the UI feels insanely faster and more responsive. I've switched most things to act on press (mouse/key down) and it is 10000% true. Some tabbed form component for example had some quite slow tabs since it was quite complex and this hid its slowness.

mj12albert: Briefly discussed [here](https://github.com/mui/base-ui/issues/1323) ~ we'll keep using pointerdown

packages/react/src/tabs/tab/useTabsTab.ts
mj12albert: Tried removing this check in NumberField [here](https://github.com/mui/base-ui/pull/1318/commits/6e63c2bf352e0cb382a1f6ac7b89a35534e1b961) too but without it a test fails: https://app.circleci.com/pipelines/github/mui/base-ui/7042/workflows/cc6f84c4-5cc2-488c-b741-ebb30c275d4a/jobs/61996/parallel-runs/0/steps/0-106

Not sure if it's just a test issue but I'd prefer to keep the main button check the same for now CC @atomiks

atomiks (approved):

packages/react/src/tabs/root/TabsRoot.test.tsx
```diff
@@ -168,6 +168,23 @@ describe('<Tabs.Root />', () => {
...
+      const { getAllByRole, user } = await render(
+        <Tabs.Root value={0} onValueChange={handleChange}>
+          <Tabs.List>
+            <Tabs.Tab value={0} />
+            <Tabs.Tab value={1} onPointerDown={handlePointerDown} />
+          </Tabs.List>
+        </Tabs.Root>,
+      );
+
+      await user.pointer({ keys: '[MouseLeft>]', target: getAllByRole('tab')[1] });
```
atomiks, line 183: Interesting syntax?

colmtuite (approved):

packages/react/src/tabs/root/TabsRoot.test.tsx
mj12albert: https://testing-library.com/docs/user-event/pointer/#pressing-a-button-or-touching-the-screen
pressing and releasing, also works for keyboard 😬 `user.keyboard('{a>}') // press a without releasing it`

PR #1319, opened 2025-01-10
[wip][experiments] Fluent UI submenus

mnajdova:
wip

### DIFF
--- a/docs/src/app/(private)/experiments/menu-fluent-ui.tsx
+++ b/docs/src/app/(private)/experiments/menu-fluent-ui.tsx
@@ -1,0 +1,130 @@
+import * as React from 'react';
+import { Menu } from '@base-ui-components/react/menu';
+import styles from './menu-fluent-ui.module.css';
+
+export default function ExampleMenu() {
+  return (
+    <div>
+      <div style={{ marginBottom: '10px' }}>
+        Reference:{' '}
+        <a href="https://developer.microsoft.com/en-us/fluentui#/controls/web/contextualmenu">
+          ContextualMenu with subemnus
+        </a>{' '}
+      </div>
+      <Menu.Root>
+        <Menu.Trigger className={styles.Button}>
+          Contextual menu <ChevronDownIcon className={styles.ButtonIcon} />
+        </Menu.Trigger>
+        <Menu.Portal>
+          <Menu.Positioner align="start" className={styles.Positioner}>
+            <Menu.Popup className={styles.Popup}>
+              {/* Submenu */}
+              <Menu.Root>
+                <Menu.SubmenuTrigger className={styles.ItemWithSubmenu}>
+                  New{' '}
+                  <span className={styles.SubmenuTrigger}>
+                    <ChevronRightIcon />
+                  </span>
+                </Menu.SubmenuTrigger>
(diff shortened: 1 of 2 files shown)


colmtuite:
> On the composite grids that have cols, we may need a different loop-behavior, looping trough rows/cols, vs looping trough all elements.

When you're focused on "5" and press left arrow, I expected to move to "4". Is that an example of what you're talking about here?

Other notes:

- When using the pointer to hover menuitems in the grid, the focus doesn't move as you hover each menuitem.
- I can't move focus in the grid using the arrow keys while the pointer is hovering over the submenu.
- When you close the submenu using the left arrow key, number "4" becomes highlighted as the submenu is closing.

mnajdova:
> When you're focused on "5" and press left arrow, I expected to move to "4". Is that an example of what you're talking about here?

Yep exactly, we could build it as a new feature in the Composite component.

mnajdova:
> Other notes

I will look into what changes we would need to do to fix these things without doing any library changes

mnajdova:
I was wrong initially on using Composition as we use `useListNavigation` from Floating UI in the menu. I tried extending a bit our API to add support for it, but there is more work involved in making the keyboard navigation work as expected. The mouse interactions work tough. If we want to support this, I would recommend adding an issue for a new feature and add this natively supported as building it on top of the existing primitives is not possible.

It's going to require some changes in Floating UI as well, as seems like this is not a supported feature, see https://codesandbox.io/p/sandbox/morning-haze-ztpr88?file=%2Fsrc%2FDropdownMenu.tsx&workspaceId=ws_LCQafiBEJ7ynffHYRMAZL

mnajdova:
I investigated a bit the Floating UI repo and found the issue. When deciding on opening submenus/stepping inside etc, Floating UI uses the orientation of the submenu to decide whether it should open/step into instead of the orientation of the parent. This is why when those two are different it's behaving unexpectedly. I will talk with James to see if it makes sense to change this behavior and open a PR there.

mnajdova:
I've created PR on the Floating UI side: https://github.com/floating-ui/floating-ui/pull/3199, I will test the changes here and see if everything work as expected, but this should fix all issues. @colmtuite you can check the behavior in the linked PR.

docs/src/app/(private)/experiments/menu-fluent-ui.tsx
```diff
@@ -0,0 +1,159 @@
...
+import { Menu } from '@base-ui-components/react/menu';
+import styles from './menu-fluent-ui.module.css';
+
+export default function ExampleMenu() {
+  return (
+    <div style={{ display: 'flex', gap: '16px', flexDirection: 'column' }}>
+      <div>
+        Reference:{' '}
+        <a href="https://developer.microsoft.com/en-us/fluentui#/controls/web/contextualmenu">
+          ContextualMenu with subemnus
```
michaldudak, line 12:
```suggestion
          ContextualMenu with submenus
```

michaldudak (commented):
These menus suffer from the same problem as described here: https://github.com/mui/base-ui/pull/1330#issuecomment-2588679397

Once #1330 is merged, it should work OK.

PR #1321, opened 2025-01-10
[code-infra] CircleCI maintenance

michaldudak:
- Removed the React 17 jobs (as in the Core repo). The shared test utils don't work with this version anymore.
- Corrected how TS is installed in the typescript@next pipeline
- Used more human-friendly names for all jobs for consistency
- Commented out the bundle-size-monitor job as it wasn't doing anything


.circleci/config.yml
```diff
@@ -350,71 +333,58 @@ jobs:
...
     jobs:
       - checkout:
           <<: *default-context
+          name: 'Checkout'
       - test_unit:
           <<: *default-context
           name: 'JSDOM tests'
           requires:
-            - checkout
+            - Checkout
```
LukasTy, line 347: These should probably remain `checkout` as it relies on `key`, not the `name`. 🤔 
The jobs seem stuck now. :/
```suggestion
            - checkout
```

.circleci/config.yml
```diff
@@ -245,18 +233,13 @@ jobs:
       - run:
           name: Resolve typescript version
           command: |
-            pnpm add typescript@next -d -w
+            pnpm install typescript@next -d -w
```
Janpot, line 236: I think the best command I know for this is

```
pnpm update -r typescript@next
```

it should update every workspace in the repo where typescript is installed, including the root.

.circleci/config.yml
oliviertassinari: +1 to keep snake case to match with the rest of the codebase and with https://circleci.com/docs/introduction-to-yaml-configurations/

michaldudak: It does rely on a name. The job seemed stuck as Github checks expect a job with the old name. I will update the GH configuration once the PR is merged so that other PRs aren't stuck.

michaldudak: @oliviertassinari I haven't found anything in the CircleCI that forces or recommends any particular format. In fact, they use different conventions in different parts of the docs (https://circleci.com/docs/configuration-reference/#requires has kebab-case)

I prioritized the readability of GH status checks (especially for external contributors).
Compare
image
with
image

michaldudak:
@mui/code-infra can I please have this re-reviewed? I'd like to get rid of red builds on master every morning after the broken pipeline runs.

.circleci/config.yml
LukasTy: I'm all for more readability if it's just a GH config. 👍

LukasTy (approved):
LGTM. 👍

PR #1338, opened 2025-01-15
[menu] Support submenus with `openOnHover` prop

### LINKED ISSUE #1332: [Menu] `openOnHover` doesn't support submenus
https://deploy-preview-1330--base-ui.netlify.app/experiments/menu-fully-featured

- Open Settings (on right)
- Toggle `Open on hover`
- Hover trigger without clicking
- Try to hover submenu
- Root menu closes and the submenu stays stuck open with no anchor (to the top-left)

Previously it worked since it stayed stuck open when hovering, but changed due to #1191. If a submenu was opened it should 'stick' like before.

atomiks:
Closes #1332

### DIFF
--- a/packages/react/src/menu/positioner/useMenuPositioner.ts
+++ b/packages/react/src/menu/positioner/useMenuPositioner.ts
@@ -56,7 +56,7 @@ export function useMenuPositioner(
     return () => {
       menuEvents.off('openchange', onMenuOpenChange);
     };
-  }, [menuEvents, nodeId, parentNodeId, setOpen]);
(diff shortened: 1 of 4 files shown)


mnajdova:
Behavior looks good: https://codesandbox.io/p/sandbox/gsy7q8. Can you please add a test for this. Maybe an e2e test would be best suited, but if we can cover it with unit test, even better 👌

atomiks:
@mnajdova had an issue writing a test for this before, but looks like I solved it 👍

michaldudak (approved):

PR #1339, opened 2025-01-15
[docs][Slider] Prevent long presses from triggering context menu or text selection

### LINKED ISSUE #1295: [Slider] Holding the thumb after `pointerdown` on touch devices opens context menu
Devtools responsive mode:

https://github.com/user-attachments/assets/29067d82-8162-4dcc-986e-57c8db735ed5

iOS:

https://github.com/user-attachments/assets/10db8708-eadf-4e33-8f63-e93fdcdd92fd

onehanddev:
Fixes #1295 

Preview: https://deploy-preview-1339--base-ui.netlify.app/react/components/slider


mj12albert:
@onehanddev Would you mind trying to move the styles to the `Track`? It's still possible to text-select the track 😅 though much harder than before, at 15s here:

https://github.com/user-attachments/assets/7b523679-64d0-40d1-ba77-526cfd80801c

onehanddev:
> @onehanddev Would you mind trying to move the styles to the `Track`? It's still possible to text-select the track 😅 though much harder than before, at 15s here:
> 
>  slider-user-select.mp4

@mj12albert I’ve applied the styles to all slider elements, as moving them to just the Track didn’t resolve the issue. This is because touch events are registered separately for each element. The issue should now be fully resolved. Let me know if you notice anything else! 😊

docs/src/app/(public)/(content)/react/components/slider/demos/hero/css-modules/index.module.css
```diff
@@ -4,6 +4,8 @@
   align-items: center;
   width: 14rem;
   padding-block: 0.75rem;
+  touch-action: none;
```
mj12albert, line 7: 👍 makes total sense to put this on `Control` actually, though it isn't needed on any descendents:

> This means that in practice, touch-action is typically applied only to top-level elements which have some custom behavior, without needing to specify touch-action explicitly on any of that element's descendants

https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action

(`user-select` is though, it doesn't inherit)

mj12albert (approved):
LGTM ~ just need to remove the redundant `touch-action`s @onehanddev 

Though on iOS it seems impossible to fully prevent unwanted text selection in CSS alone:

https://github.com/user-attachments/assets/122fc554-c3b8-409c-9a19-16b1d3efafb1

As described in the RA codebase [here](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/src/textSelection.ts#L15-L19):
> Safari on iOS starts selecting text on long press. The only way to avoid this, it seems, is to add user-select: none to the entire page. Adding it to the pressable element prevents that element from being selected, but nearby elements may still receive selection. We add user-select: none on touch start, and remove it again on touch end to prevent this. This must be implemented using global state to avoid race conditions between multiple elements.

docs/src/app/(public)/(content)/react/components/slider/demos/hero/css-modules/index.module.css
onehanddev: Good point @mj12albert, I have removed the touch-action property from all of the slider controls's children elements in the latest commit, please do have a look.

PR #1343, opened 2025-01-17
[useScrollLock] Fix jump with `scroll-behavior` style

### LINKED ISSUE #1340: [Dialog] Page is scrolled to top when opening the dialog if `html { scroll-behavior: smooth }`
# Bug report

## Current behavior

Super weird one, but like the title says! I can reproduce in Chrome and Safari; it works fine in Firefox.

## Expected behavior

The page should _not_ be scrolled to the top when the dialog opens.

## Reproducible example

https://codesandbox.io/p/sandbox/competent-chandrasekhar-8wcv88

## Base UI version

1.0.0-alpha.5

Chrome 132.0.6834.84, Safari 18.2

## Which OS are you using?

macOS

None

## Additional context

Thank you 🙏

atomiks:
Fixes #1340


packages/react/src/utils/useScrollLock.ts
```diff
@@ -52,6 +53,7 @@ function preventScrollStandard(referenceElement?: Element | null) {
       overflowY: html.style.overflowY,
       overflowX: html.style.overflowX,
     };
+    originalHtmlScrollBehavior = html.style.scrollBehavior;
```
michaldudak, line 56: Why isn't it a part of `originalHtmlStyles`?

atomiks: It has to be assigned after the `.scrollTop/Left` assignments

michaldudak (approved):

PR #1349, opened 2025-01-21
feat/toolbar

mj12albert:


docs/src/app/(private)/experiments/toolbar/text-editor.tsx
```diff
@@ -0,0 +1,324 @@
...
+  icon: (props: React.ComponentProps<'svg'>) => React.JSX.Element;
+  disabled?: boolean;
+}) {
+  const { key, label, icon: Icon, disabled = false } = args;
+  return (
+    <Tooltip.Root key={key}>
+      <Tooltip.Trigger
+        render={
+          <Toolbar.Button
+            render={<Toggle />}
```
mj12albert, line 90:
```jsx
<Toolbar.Button
  render={
    <Toggle
      pressed={dialogOpen}
      onPressedChange={(nextPressed) => {
        setDialogOpen(nextPressed);
      }}
    >
      <Icon />
    </Toggle>
  }
/>
```
If I tried to make this a controlled toggle, this part of `mergeReactProps` throws when it tries to merge our custom handlers, as the event arg is not the first: https://github.com/mui/base-ui/blob/master/packages/react/src/utils/mergeReactProps.ts#L44

mj12albert: https://github.com/mui/base-ui/issues/1430

aarongarciah:
I found a strange behavior: the menu won't close when a submenu is open and you move the roving focus to the next toolbar element by pressing ➡️. Only the submenu is closed. See attached video.

https://github.com/user-attachments/assets/ac5f6a87-1066-498e-8787-9ce037722433

mj12albert:
> Only the submenu is closed

Hmm this is tricky, it probably has to be handled internally in Menu somehow. I think there could be a big overlap with `Menubar` functionality here

packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -90,7 +93,11 @@ const ToggleGroup = React.forwardRef(function ToggleGroup(
 
   return (
     <ToggleGroupContext.Provider value={contextValue}>
-      <CompositeRoot direction={direction} loop={loop} render={renderElement()} />
+      {toolbarContext ? (
+        renderElement()
+      ) : (
+        <CompositeRoot direction={direction} loop={loop} render={renderElement()} />
+      )}
```
mj12albert, line 100: Is it a bit sketchy to need a component to have to check another somewhat-arbitrary component's context?

docs/src/app/(private)/experiments/toolbar/text-editor.tsx
```diff
@@ -0,0 +1,413 @@
...
+                      )}
+                    >
+                      <NumberField.Decrement
+                        className={styles.numField.Decrement}
+                        disabled={settings.toolbarDisabled}
+                      >
+                        <MinusIcon />
+                      </NumberField.Decrement>
+
+                      <Toolbar.Input
```
mj12albert, line 234: Possibly check for NumberFieldContext inside Toolbar.Input so this configuration be disabled like:
```jsx
<NumberField.Root disabled>
  <Toolbar.Input render={<NumberField.Input />} />
</NumberField.Root>
```

docs/src/app/(private)/experiments/toolbar/text-editor.tsx
mj12albert: Fixed in https://github.com/mui/base-ui/pull/1440

mj12albert:
> * https://deploy-preview-1349--base-ui.netlify.app/experiments/toolbar/basic
> * https://deploy-preview-1349--base-ui.netlify.app/experiments/toolbar/text-editor
> * https://deploy-preview-1349--base-ui.netlify.app/experiments/toolbar/triggers

@colmtuite What do you think about adapting one of these to be the hero demo or do you have something else in mind?

mj12albert:
> I found a strange behavior: the menu won't close when a submenu is open and you move the roving focus to the next toolbar element by pressing ➡️. Only the submenu is closed. See attached video.

I think it's better to solve this in [Menubar](https://github.com/mui/base-ui/issues/1407), perhaps something like a shared `CompositeMenuContext` to handle multiple menus grouped into a composite component

CC @michaldudak

mj12albert:
There are still a handful tests to add but otherwise this is ready for review @colmtuite @atomiks @michaldudak ~

colmtuite:
I tested Toolbar in Chrome desktop a bit:

- When an input receives focus, shouldn't its whole value become focused?
- When focus is on a menuitem, and you press right/left, it doesn't move focus to the next/previous menu trigger. Is there a specific reason for this? Same goes for Select.
- In the Triggers experiment, it seems like disabling controls doesn't work properly? When I disable the switch individually, that works. But when I disable the Dialog individually, that doesn't work. I saw this was called out before, at least in some capacity.
- What do we think about treating numberfield differently to other inputs? Handling its behaviour differently, given that its value is usually just 1–2 chars long? Is there value in it, or better to leave it the same?
- Why is there a "Menubar" experiment? Is this trying to be a menubar? Is it supposed to avoid the need for a Menubar component?

cc @Albert Yu

mj12albert:
> * When an input receives focus, shouldn't its whole value become focused?

https://github.com/user-attachments/assets/1ad46073-efed-471f-afbc-91cb7c170263

Following the reference Michal shared, we decided that the focus should move directly to a text insertion cursor placed at the start/end of the textbox value depending on the direction you're arrowing in ~

> When focus is on a menuitem, and you press right/left, it doesn't move focus to the next/previous menu trigger. Is there a specific reason for this?

This was a side effect of something else. We decided that this should be different from Menubar, for toolbar you should be able to arrow left/right from (possibly nested) menu items to an adjacent toolbar item. If that item is a trigger (menu or otherwise) it should not activate the trigger.

> Why is there a "Menubar" experiment?

It's just an experiment to put 3 menus in toolbar which is mostly a generic roving focus group to see what gaps there are between this and a proper menubar, what can be shared etc

mj12albert:
> Following the reference Michal shared, we decided that the focus should move directly to a text insertion cursor placed at the start/end of the textbox value depending on the direction you're arrowing in

Hey @michaldudak just wondering if text navigation behavior feels correct to you now (in the [input demo](https://deploy-preview-1349--base-ui.netlify.app/experiments/toolbar/basic) and  [number field demo](https://deploy-preview-1349--base-ui.netlify.app/experiments/toolbar/text-editor))

> When focus is on a menuitem, and you press right/left, it doesn't move focus to the next/previous menu trigger. Is there a specific reason for this? Same goes for Select

Also wondering what the behavior is like for this in your MS Office example, what happens when the select/combobox popup is open, one of the items is focused and then ArrowLeft/Right is pressed? 

(Ariakit does not let you arrow sideways out of a select in a toolbar: https://ariakit.org/examples/toolbar-select)

mj12albert:
> * In the Triggers experiment, it seems like disabling controls doesn't work properly? When I disable the switch individually, that works. But when I disable the Dialog individually, that doesn't work. I saw this was called out before, at least in some capacity

OK it turns out various triggers have issues with the disabled state when they are not native `button` elements disabled with the `disabled` attribute (i.e. a custom element disabled using aria-disabled/data-disabled)

The fix depends on https://github.com/mui/base-ui/pull/1445

Before some of these experiments seemed to work occasionally because I was passing the disabled state 2-3 times, to the trigger, the root, and the toolbar button at the same time, but it should only be passed once to the toolbar button

michaldudak:
I find restoring focus to the previously active item quite weird. I know ARIA guidelines suggest this ("If the toolbar has previously contained focus, focus is **optionally** set on the control that last had focus. Otherwise, it is set on the first control that is not disabled."), but the applications I checked don't work like that. cc @colmtuite

packages/react/src/toolbar/link/ToolbarLink.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { CompositeItem } from '../../composite/item/CompositeItem';
+import type { ToolbarItemMetadata } from '../root/ToolbarRoot';
+import { useToolbarRootContext } from '../root/ToolbarRootContext';
+
+const TOOLBAR_LINK_METADATA = {
+  focusableWhenDisabled: true,
+};
+/**
+ * A link component.
+ * Renders a `<a>` element.
```
michaldudak, line 16:
```suggestion
 * Renders an `<a>` element.
```

packages/react/src/toggle/useToggle.ts
```diff
@@ -38,18 +38,16 @@ export function useToggle(parameters: useToggle.Parameters): useToggle.ReturnVal
 
   const getRootProps = React.useCallback(
     (externalProps?: GenericHTMLProps): GenericHTMLProps => {
-      return mergeReactProps(
-        externalProps,
-        {
+      return getButtonProps(
```
michaldudak, line 41: #1445 is now merged, so you can write `mergeReactProps(getButtonProps, externalProps, {...})`

packages/react/src/toolbar/button/ToolbarButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const state: ToolbarButton.State = React.useMemo(
+    () => ({
+      disabled,
+      orientation,
+      focusable: focusableWhenDisabled,
+    }),
+    [disabled, focusableWhenDisabled, orientation],
+  );
+
+  const extraProps = React.useMemo(
```
michaldudak, line 54: This doesn't have to be memoized as `useComponentRenderer` destructures it anyway and doesn't care about object reference.

packages/react/src/use-button/useButton.ts
```diff
@@ -22,6 +23,8 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
     rootElementName: elementNameProp,
   });
 
+  const isCompositeItem = useCompositeRootContext(true) !== undefined;
```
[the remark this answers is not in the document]
michaldudak: I wonder if this could cause problems with buttons nested inside other components within a toolbar (for example, a button inside a dialog opened from a toolbar would have this set to `true`, even though it's not a composite item).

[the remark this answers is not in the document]
mj12albert: > for example, a button inside a dialog opened from a toolbar

Only if that button happened to use `useButton` as well, so in this case, a disabled `Dialog.Close` isn't affected, but a MenuTrigger in that dialog would be

What this affects is that when `disabled`, the HTML `disabled` attribute is never applied. Generally I think our components that use `useButton` already don't rely on the `disabled` attr to disable functionality anyway in order to support component/element replacement so it should be OK in most cases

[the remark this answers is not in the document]
mj12albert: I also tried making it a `isCompositeItem` param instead of a check but it wasn't enough to handle this more common case: https://github.com/mui/base-ui/pull/1349/commits/575dc96ef3d9ec058a8592df8858a2398f9d1834

packages/react/src/use-button/useButton.ts
```diff
@@ -42,27 +46,50 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
...
         additionalProps['aria-disabled'] = disabled;
-      } else {
+      } else if (!isCompositeItem) {
         additionalProps.disabled = disabled;
       }
     } else if (elementName !== '') {
-      additionalProps.role = 'button';
-      additionalProps.tabIndex = tabIndex ?? 0;
+      if (elementName !== 'A') {
+        additionalProps.role = 'button';
```
mj12albert, line 61: This overwrites `Select.Trigger`'s built-in `role` 😩

mj12albert:
> * But when I disable the Dialog individually, that doesn't work. I saw this was called out before, at least in some capacity.

I've opened a PR to fix this issue for Dialog/AlertDialog/Popover on their end: https://github.com/mui/base-ui/pull/1473

Update: it's merged

packages/react/vitest.config.mts
```diff
@@ -1,10 +1,16 @@
+import * as path from 'path';
 import { mergeConfig, defineProject } from 'vitest/config';
 // eslint-disable-next-line import/no-relative-packages
 import sharedConfig from '../../vitest.shared.mts';
 
 export default mergeConfig(
   sharedConfig,
   defineProject({
+    resolve: {
```
michaldudak, line 9: Why is this needed now?

packages/react/src/use-button/useButton.ts
michaldudak: I don't think we need `useButton` to set a role at all. Our components using `useButton` are not regular buttons and should override it anyway.

packages/react/vitest.config.mts
mj12albert: Oh it's actually not needed, the actual issue was I forgot to add toolbar to the package.json exports field

packages/react/src/use-button/useButton.ts
mj12albert: Without changing the order of Floating UI `getReferenceProps` and our `getButtonProps`, this seems like the easiest solution: https://github.com/mui/base-ui/pull/1349/commits/5103ac6ff6c56b02f90d7e078c75d9f3e9e6ecfa

mj12albert:
https://github.com/user-attachments/assets/9964a8c4-3f12-4bc1-8aac-e27b3b6b7f5b

☝️ I've updated the toolbar input keyboard interaction behavior, see what you think @colmtuite 

Also: what do you think about Michal's [comment](https://github.com/mui/base-ui/pull/1349#issuecomment-2673872992) about restoring focus to the last focused item:

> I find restoring focus to the previously active item quite weird. I know ARIA guidelines suggest this ("If the toolbar has previously contained focus, focus is optionally set on the control that last had focus. Otherwise, it is set on the first control that is not disabled."), but the applications I checked don't work like that.

michaldudak:
> I've updated the toolbar input keyboard interaction behavior

It feels OK to me. It's also consistent with the usual behavior of text fields when they receive focus, so IMO we can go with it.

mj12albert:
Added a [simple demo](https://deploy-preview-1349--base-ui.netlify.app/react/components/toolbar) for now:

image

mj12albert:
image

I've made the [demo](https://deploy-preview-1349--base-ui.netlify.app/react/components/toolbar) smaller and removed the icons/input @colmtuite

colmtuite:
@mj12albert 

**Content**
- Change "Auto" and "Text" toggles to "Align Left" and "Align Right".
- Remove the "Using with Toggles" example, that's covered in the hero example.
- Change "Using with popups" to "Using with Menu", and remove the Select example.
- Add a Select menu into the hero demo. Just 2 options: "Arial" and "Helvetica".

**Styling**
- Computed font-size on all text in the hero example should be 14px.
- Add `font-family: inherit` to controls/buttons, so it inherits the font stack.
- Add `font-weight: 500` to controls/buttons.
- Change parent from `inline-flex` to `flex`. Add `align-items: center` to it.
- Remove `align-self: stretch` from separators. Give separators `height: 16px`. Set their `margin` to `0 .25rem`.
- Remove `cursor: pointer` from controls/buttons.

mj12albert:
image

@colmtuite Updated the demo and docs: https://deploy-preview-1349--base-ui.netlify.app/react/components/toolbar

If this is ok I'll add the tailwind version as well

Issue #559, opened 2024-08-30
[tooltip] Bring mobile support back?


oliviertassinari:
### Steps to reproduce

Link to live example: (required)

Steps:
1. https://master--base-ui.netlify.app/base-ui/react-tooltip/

### Expected behavior

Worked like in https://mui.com/material-ui/react-tooltip/

Also missing those tests: https://github.com/mui/material-ui/blob/eea5079a00bedb68e27f22b7f94d9a71cc97ee1e/packages/mui-material/src/Tooltip/Tooltip.test.js#L404

### Context

- #558
- https://github.com/adobe/react-spectrum/issues/1136

### Your environment

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

**Search keywords**: -

atomiks:
Tooltips intentionally don't work on touch devices in the new Base UI. [Hoverable popovers (for "infotips"/contextual help)](https://base-ui.com/react/components/popover#PopoverTrigger-openOnHover) are a valid pattern though, and should be used instead when the trigger's only purpose is to show the "tooltip" (infotip for more correct terminology). If the trigger performs an action, then it's impossible for a tooltip to be attached in a valid way for touch input. They're inherently secondary in nature.

oliviertassinari:
> Tooltips intentionally don't work on touch devices in the new Base UI. 

Ok, turning this into a waiting for upvotes issue then.

oliviertassinari:
I faced a pain today that seems to be a good motive for the need to bring back tooltip support for touch devices.

I was on my phone, and as usual, I have the desktop mode enabled (nowadays, I almost never use the mobile web version of websites on my phone, most of the time it has: /2 features, a different UI organization making me so lost on them, has everything too big).

- On Ashby, I wanted to understand what this icon was for: https://app.ashbyhq.com/admin/custom-tracked-links/6efb2ea4-7a66-4046-9e4f-656a82a23053, to fix the application source association
  <img width="202" alt="Image" src="https://github.com/user-attachments/assets/e78c3fed-844b-45af-a8db-7d671e44552f" />
  It's a Radix Primitive Tooltip

  <img width="1166" alt="Image" src="https://github.com/user-attachments/assets/48ee3feb-f2e3-4c05-a1e1-bc6a9cc1debc" />

[...]

PR #1356, opened 2025-01-27
[docs] Tooltip guidelines

atomiks:
Before the API reference, it's a good idea to list guidelines to ensure users are using the right component. This is currently a plain section/paragraph, but we can add callouts or styling/colors.

Part of https://github.com/mui/base-ui/issues/559

[Preview](https://deploy-preview-1356--base-ui.netlify.app/react/components/tooltip#guidelines)


colmtuite:
@atomiks Perhaps "Accessibility" or "Accessibility Guidelines" is a better heading? Also, this should come after the API Reference. The first thing you see for any component should be the demo and the anatomy.

atomiks:
Hmm yeah, "Accessibility guidelines" is likely better. Other guidelines like "Design guidelines" can be separate from accessibility guidelines, if we decide to add them at some point.

> Also, this should come after the API Reference.

I somewhat disagree here (even though I think it looks cleaner). The user may not reach the end of the API reference before deciding to start using the component since they didn't see the guidelines - it's important that they understand they're using the right component up front/as soon as possible before diving in.

colmtuite (approved):

PR #1358, opened 2025-01-27
[test] Extract common popup tests

michaldudak:
To help ensure all our popup-based components work in a similar way, I added a popup test suite that's shared across all these components. I have included just the basic tests so far. It can be extended to test modality going forward.

Additional changes:
Created async versions of `rerender` and `setProps` functions returned from `render` in tests. They now behave similarly to `render` as they flush the microtask queue before returning.


packages/react/src/preview-card/positioner/usePreviewCardPositioner.ts
```diff
@@ -21,7 +21,6 @@ export function usePreviewCardPositioner(
         }
 
         return mergeReactProps<'div'>(externalProps, {
-          role: 'presentation',
           hidden: !mounted,
```
atomiks, line 24: Why was this moved?

michaldudak: So it's consistent with other roles - they are placed on popups, not positioners. Is it an issue?

atomiks: For some screen readers, the positioner wrapper needs to have `role="presentation"` to not interfere with the child popup `role` from what I've gathered. It makes it 'inert'/ignored by them. The popup is what should have the ARIA role, but never role=presentation

atomiks (approved):

PR #1359, opened 2025-01-27
[tabs] Fix being able to activate a disabled tab

michaldudak:
Placing a `disabled` prop on a `Tabs.Tab` only prevented it from being activated via a keyboard, but not with a pointer.

Before: https://codesandbox.io/p/sandbox/gallant-sunset-rq8gt4?file=%2Fsrc%2Findex.module.css
After: https://codesandbox.io/p/sandbox/kind-newton-xmpj8v?workspaceId=ws_QAeLcZEcDnfGaFA47etEUH


mj12albert (approved):
Looks good 👍

PR #1364, opened 2025-01-29
[Form] Fix focusing of invalid field controls on errors prop change

### LINKED ISSUE #1361: [Form] Input focus shift for onClearErrors with multiple Fields
# Bug report

## Current behavior

[Live example](https://codesandbox.io/p/sandbox/busy-sound-9mspmd)

This is a copy of the [Base UI form documentation example](https://base-ui.com/react/components/form), but then with 2 form fields.

Steps:
1. Create a Form with multiple Fields.
2. In the onSubmit of the form, set the errors object with an error for both Fields. 
3. Both Fields show the error correctly, the first Field is focussed.
4. When changing the first input, the error for this field is immediatly cleared (which makes sense, the user is working on fixing the error), except where it goes wrong is that the moment a single change is made, the focus immediatly switches to the second Field that has an error, instead of letter the using fix the input content.

## Expected behavior

In step 4, I would expect the focus to remain in the first field. I cannot change more than 1 character before my focus automatically goes to the next field that has an error.

## Reproducible example

Codesandbox example: https://codesandbox.io/p/sandbox/busy-sound-9mspmd

## Base UI version

For example v1.0.2

Chrome

## Which OS are you using?

Mac OS/Windows

atomiks:
Fixes #1361

This uses a ref to check when `onSubmit` is called and assigns it to `true`. The effect to react to `errors` changing only runs once after the event fires. Also removed the manual `onSubmit` call, since `mergeReactProps` already does this now

### DIFF
--- a/packages/react/src/form/Form.tsx
+++ b/packages/react/src/form/Form.tsx
@@ -17,20 +17,13 @@ const Form = React.forwardRef(function Form(
   props: Form.Props,
   forwardedRef: React.ForwardedRef<HTMLFormElement>,
 ) {
-  const {
-    render,
-    className,
-    errors,
-    onClearErrors: onClearErrorsProp,
-    onSubmit: onSubmitProp,
-    ...otherProps
-  } = props;
+  const { render, className, errors, onClearErrors: onClearErrorsProp, ...otherProps } = props;
 
   const formRef = React.useRef<FormContext['formRef']['current']>({
     fields: new Map(),
   });
+  const submittedRef = React.useRef(false);
 
-  const onSubmit = useEventCallback(onSubmitProp || (() => {}));
   const onClearErrors = useEventCallback(onClearErrorsProp || (() => {}));
 
   const getFormProps = React.useCallback(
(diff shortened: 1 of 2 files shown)


michaldudak:
It is now possible to submit a form with errors - such as with value not matching the required pattern.

atomiks:
> It is now possible to submit a form with errors - such as with value not matching the required pattern.

Yeah the order matters here with `onSubmit`, it's called always with `mergeReactProps`.
Added a regression test

packages/react/src/form/Form.tsx
```diff
@@ -17,13 +17,21 @@ const Form = React.forwardRef(function Form(
...
+    onSubmit: onSubmitProp,
+    ...otherProps
+  } = props;
 
   const formRef = React.useRef<FormContext['formRef']['current']>({
     fields: new Map(),
   });
   const submittedRef = React.useRef(false);
 
+  const onSubmit = useEventCallback(onSubmitProp || (() => {}));
```
michaldudak, line 34: nit: we could define a NOOP constant and reuse it (also in onClearErrors below) instead of defining a new function on each render.

michaldudak (approved):

packages/react/src/form/Form.tsx
atomiks: I'd prefer if `useEventCallback` handled nil values actually

michaldudak: Yeah, that's a good approach.

PR #1371, opened 2025-01-30
[test] Fix flaky browser tests

atomiks:
Skips flaky Menu (likely due to how `safe-polygon` works) and Avatar tests in browser, adds `waitFor` for other Avatar flakes that occurred.


mj12albert (approved):
👍 nice ~ I started to notice the avatar one yesterday too

Issue #1406, opened 2025-02-03
[Slider] Thumb position doesn't update on external value change when controlled


jaggaer-mplanic:
# Bug report

## Current behavior

Slider component, when used as controlled component does not update its value on state change outside of its self.

Here is example where state is changed on button click but slider still shows initial value even it is passed as **value** prop

<img width="1209" alt="Image" src="https://github.com/user-attachments/assets/719df000-592f-434c-b271-c043d6943942" />

## Expected behavior

Slider should be updated, thumb should be moved to the new position according to new value of the state. 

## Reproducible example

[Link to a CodeSandbox Example](https://codesandbox.io/p/sandbox/sad-surf-zhg7ts)

## Base UI version

1.0.0-alpha.5

## Which browser are you using?

Chrome, Firefox

## Which OS are you using?

Mac OS

mj12albert:
Thanks for reporting this @jaggaer-mplanic, it's a bug (related to https://github.com/mui/base-ui/issues/1360)

mj12albert:
If anybody would like to test the fix locally in the mean time:
```sh
npm install https://pkg.csb.dev/mui/base-ui/commit/6599a23d/@base-ui-components/react
```

Issue #1360, opened 2025-01-27
[Slider] Ends are unreachable with mouse drags


SimeonRolev:
# Bug report

## Current behavior

Drag a `Slider` to one of its ends. Value is properly set to 0 or 100, but the `Slider.Thumb` doesn't get positioned at the very start/end of the `Slider.Track` (its `inset-inline-start` CSS property is set to 0.4242% or 99.4242%, for example). A proper positioning is only achievable with a quick drag. It's more obvious with wider sliders or multiple sliders one under the other.

https://github.com/user-attachments/assets/064dc37f-145c-43f1-9ca3-d5bc77d396aa

## Expected behavior

The positioning of the thumb should reflect the value passed to the `onValueChange(value, event)` - a value of `0 / 100` positions the `Slider.Thumb` at `inset-inline-start: 0% / 100%;` without the need of a quick mouse drag.

## Reproducible example

https://base-ui.com/react/components/slider 

## Base UI version

^1.0.0-alpha.5

## Which browser are you using?

Chrome

mj12albert:
Thanks for reporting this @SimeonRolev 

This would be caused by https://github.com/mui/base-ui/pull/1219, I think we need to round the position values when the pointer is released

SimeonRolev:
@mj12albert That's what I initially thought of as a workaround and I gave it a shot yesterday, utilizing the `onValueCommitted` callback. As the `min/max` `value`s of the `Slider` might not be 0 to 100, I parsed the value of the `thumbNode.style.insetInlineStart` css property instead to determine whether we are close to an edge and re-adjust the `Slider.Thumb` only in that case to avoid re-adjusting on every mouse up. This approximation can be improved for example by using the `min`, `max` and `step` values properly, instead of just comparing <1 || >99, but you get the point.

Unfortunately, I don't think it's optimal. A user sets a slider's value during a mouse drag and seeing the slider re-adjust on its own upon mouse release is awkward - even if it happens only at the edges. Makes you not trust the component - did it only glitch visually or did it slightly tweak my value, too?

```jsx
const MySlider = (props) => {
    const ref = useRef(null);

    const onValueCommitted = (value, event) => {
        handleEdges();
        return props.onValueCommitted?.(value, event);
    };

    const handleEdges = () => {
        if (props.orientation === 'vertical') return;

        const thumbNode = ref.current.querySelector('.thumb');
        if (!thumbNode) return;

        const cssValue = thumbNode.style.insetInlineStart;
[...]

PR #1411, opened 2025-02-04
[Slider] Fix thumb positioning

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1360
Fixes https://github.com/mui/base-ui/issues/1406

**Demo:** https://stackblitz.com/edit/vitejs-vite-zsatgbgk?file=src%2FApp.tsx

The demo shows both fixes:
- when setting the value with the NumberField, the corresponding slider should reposition correctly
- when dragging a thumb to match the same value set by a NumberField (of an adjacent row) the thumbs should line up exactly and have the same CSS `width` value

### DIFF
--- a/packages/react/src/slider/root/useSliderRoot.ts
+++ b/packages/react/src/slider/root/useSliderRoot.ts
@@ -363,6 +363,24 @@ export function useSliderRoot(parameters: useSliderRoot.Parameters): useSliderRo
   );
 
   useEnhancedEffect(() => {
+    if (valueProp === undefined || dragging) {
+      return;
+    }
+
+    if (typeof valueUnwrapped === 'number') {
+      const newPercentageValue = valueToPercent(valueUnwrapped, min, max);
+      if (newPercentageValue !== percentageValues[0]) {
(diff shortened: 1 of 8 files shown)


packages/react/src/slider/root/useSliderRoot.ts
```diff
@@ -535,34 +574,27 @@ export namespace useSliderRoot {
...
      */
     active: number;
     'aria-labelledby'?: string;
-    handleInputChange: (
-      valueInput: number,
-      index: number,
-      event: React.KeyboardEvent | React.ChangeEvent,
-    ) => void;
+    /**
+     * Callback fired when drag ends and invokes onValueCommitted.
```
michaldudak, line 583: nit: it's really up to the consumer to decide when this function is really called (as useSliderRoot doesn't call it itself as far as I can see). IMO a better description would be "Function to be called when drag ends. Invokes `onValueCommitted`"

michaldudak (approved):
Nice, bugs seem to be fixed. I have a minor remark, but it's good to go anyway.

PR #1418, opened 2025-02-05
[useRenderer] Add public hook

mnajdova:
Added a public `useRenderer` hook as an simpler adapter that uses the `useComponentRenderer` hook. Let me know if this is what we want, and if we agree I will add a documentation for it.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -25,3 +25,4 @@ export * from './tabs';
 export * from './toggle';
 export * from './toggle-group';
 export * from './tooltip';
+export * from './use-renderer';
--- a/packages/react/src/use-renderer/useRenderer.ts
+++ b/packages/react/src/use-renderer/useRenderer.ts
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import type { ComponentRenderFn } from '../utils/types';
+import { useComponentRenderer } from '../utils/useComponentRenderer';
+import { defaultRenderFunctions } from '../utils/defaultRenderFunctions';
+
+/**
+ * Returns a function that renders a Base UI component.
+ */
+export function useRenderer<
+  State extends Record<string, any>,
+  RenderedElementType extends Element,
+>(settings: useRenderer.Settings<State, RenderedElementType>) {
+  return useComponentRenderer(settings);
+}
+
+namespace useRenderer {
+  export interface Settings<State, RenderedElementType extends Element> {
+    /**
+     * The class name to apply to the rendered element.
+     * Can be a string or a function that accepts the state and returns a string.
+     */
+    className?: string | ((state: State) => string);
+    /**
+     * The render prop or React element to override the default element.
+     */
+    render:
+      | ComponentRenderFn<React.HTMLAttributes<any>, State>
+      | React.ReactElement<Record<string, unknown>>
+      | keyof typeof defaultRenderFunctions;
+    /**
+     * The state of the component. It will be used as a parameter for the render and className callbacks.
+     */
+    state: State;
+    /**
+     * The ref to apply to the rendered element.
+     */
+    ref?: React.Ref<RenderedElementType>;
+    /**
+     * A function that returns props for the rendered element.
+     * It should accept and merge additional props.
+     */
+    propGetter?: (
+      externalProps: Record<string, any>,
+    ) => React.HTMLAttributes<any> & React.RefAttributes<RenderedElementType>;
+    /**
+     * Additional props to be spread on the rendered element.
+     */
+    extraProps?: Record<string, any>;
+  }
+}
--- a/packages/react/src/use-renderer/index.ts
+++ b/packages/react/src/use-renderer/index.ts
@@ -1,0 +1,1 @@
+export { useRenderer } from './useRenderer';


mj12albert (commented):
Nice ~ I imagine the use-case is something like this: https://codesandbox.io/p/sandbox/flamboyant-fog-9x266p?file=%2Fsrc%2FApp.tsx%3A29%2C63
(for some reason importing the hook in CSB doesn't work, this works fine locally in a playground though)

This feels more intuitive to me (biased though because I never used asChild extensively) to use than a Slot component

mnajdova:
> for some reason importing the hook in CSB doesn't work, this works fine locally in a playground though

I haven't updated the exports field, will fix it. Here is the updated sandbox: https://codesandbox.io/p/sandbox/exciting-paper-9x266p-use-renderer-forked-ry269y?file=%2Fpackage.json

packages/react/src/use-renderer/useRenderer.ts
```diff
@@ -0,0 +1,49 @@
...
+    state: State;
+    /**
+     * The ref to apply to the rendered element.
+     */
+    ref?: React.Ref<RenderedElementType>;
+    /**
+     * A function that returns props for the rendered element.
+     * It should accept and merge additional props.
+     */
+    propGetter?: (
```
michaldudak, line 41: IMO `propGetter` is pretty specific to our setup, where we have hooks with logic and components using them.

mj12albert:
> I removed only the customStyleHookMapping setting from the original hook

It may be useful to provide a (alternative) way to opt-out individual props placed in state from generating a corresponding data attribute

packages/react/src/use-renderer/useRenderer.ts
```diff
@@ -0,0 +1,49 @@
...
+      externalProps: Record<string, any>,
+    ) => React.HTMLAttributes<any> & React.RefAttributes<RenderedElementType>;
+    /**
+     * Additional props to be spread on the rendered element.
+     */
+    extraProps?: Record<string, any>;
```
atomiks, line 47: This could possibly just be the top-level object props (along with removing `propGetter`)

mnajdova:
Thanks for the initial review, I updated the API to include:
- className
- render
- state
- ref
- props (passed as extraProps)
- excludedStyleHookStates [string[]] - list of state keys that should not be generated as a styled hooks (data-attributes) - based on these I am creating a customStyleHookMapping that is passed in the internal hook. I've also added usage example under: https://deploy-preview-1418--base-ui.netlify.app/experiments/custom-components

packages/react/src/use-renderer/useRenderer.ts
mnajdova: Makes sense, done.

michaldudak:
I'd leave `customStyleHookMapping` as in `useComponentRenderer`. It should be quite common to customize the data attributes (we do this often).

mnajdova:
> I'd leave customStyleHookMapping as in useComponentRenderer. It should be quite common to customize the data attributes (we do this often).

Yeah, fair enough, ok I was thinking initially to make the API simpler. I've changed it back to use the `customStyleHookMapping` and added a test for it.

packages/react/src/use-renderer/useRenderer.ts
```diff
@@ -0,0 +1,59 @@
...
+     */
+    ref?: React.Ref<RenderedElementType>;
+    /**
+     * Props to be spread on the rendered element.
+     */
+    props?: Record<string, any>;
+    /**
+     * A mapping of state to style hooks.
+     */
+    customStyleHookMapping?: CustomStyleHookMapping<State>;
```
michaldudak, line 55: This could be called just `styleHookMapping` to be simpler.

packages/react/src/use-renderer/useRenderer.ts
```diff
@@ -0,0 +1,59 @@
...
+     */
+    state: State;
+    /**
+     * The ref to apply to the rendered element.
+     */
+    ref?: React.Ref<RenderedElementType>;
+    /**
+     * Props to be spread on the rendered element.
+     */
+    props?: Record<string, any>;
```
michaldudak, line 51: Would `Record<string, unknown>` work here as well?

docs/src/app/(private)/experiments/custom-components.tsx
```diff
@@ -0,0 +1,78 @@
+'use client';
+import * as React from 'react';
+import { useRenderer } from '@base-ui-components/react/use-renderer';
+import styles from './custom-components.module.css';
+
+type TextProps = {
+  weight?: 'light' | 'regular' | 'bold';
+  className: any;
+  render?: any;
```
aarongarciah, line 9: How would users type this properly? Like this?

```suggestion
  render?: useRenderer.Settings<any, Element>['render'];
```

packages/react/src/use-renderer/useRenderer.ts
mj12albert: 👍 I was thinking the same thing, we could shorten it in the internal hook as well ~

docs/src/app/(private)/experiments/custom-components.tsx
mj12albert: We'd export a specific type for render as Michal commented: https://github.com/mui/base-ui/pull/1418#discussion_r1944835602

packages/react/src/use-renderer/useRenderer.ts
mnajdova: Renamed

packages/react/src/use-renderer/useRenderer.ts
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react';
+import type { ComponentRenderFn } from '../utils/types';
```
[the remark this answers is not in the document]
mnajdova: I've reexported them from this file. I also renamed CustomStyleHookMapping to StyleHookMapping, to match the name of the input param.

docs/src/app/(private)/experiments/custom-components.tsx
mnajdova: I've exported a RenderProp<State> type, and I've updated the example to use all the right types.

mnajdova:
I've resolved all feedback and added a docs page: https://deploy-preview-1418--base-ui.netlify.app/react/utils/use-renderer

docs/src/app/(public)/(content)/react/utils/use-renderer/page.mdx
```diff
@@ -0,0 +1,60 @@
...
+## API reference
+
+### Input parameters
+
+<PropsReferenceTable
+  data={{
+    className: {
+      type: 'string | ((state: State) => string)',
+      description:
+        'The class name to apply to the rendered element. Can be a string or a function that accepts the state and returns a string.',
```
aarongarciah, line 26: Should we use the same description used in component prop tables?

```suggestion
        'CSS class applied to the element, or a function that returns a class based on the component’s state.',
```

docs/src/app/(public)/(content)/react/utils/use-renderer/page.mdx
```diff
@@ -0,0 +1,60 @@
...
+    },
+    render: {
+      type: 'RenderProp<State>',
+      description: 'The render prop or React element to override the default element.',
```
aarongarciah, line 30: In components' prop tables we say:

```suggestion
      description: 'Allows you to replace the component’s HTML element\nwith a different tag, or compose it with another component.\n\nAccepts a `ReactElement` or a function that returns the element to render.',
```

docs/src/app/(public)/(content)/react/utils/use-renderer/page.mdx
```diff
@@ -0,0 +1,60 @@
...
+  content="Utility for adding Base UI like features in a custom build components."
+/>
+
+Base UI provides few common features across all components:
+
+- support for a [callback](/react/handbook/styling#css-classes) on the `className` prop where developers can generate classes based on the state of the components
+- [render](/react/handbook/composition) prop that allows developers to override the default rendered element of the component
+- adds data-attributes for the component's state that can be used as style hooks
+
+The `useRenderer` hook allows you to add these feature on your custom built components, so that you can provide consistenc experience across your library.
```
aarongarciah, line 15:
```suggestion
The `useRenderer` hook allows you to support the same features Base UI provides across all components so that you can also have a consistent experience in your custom components:

- A [render](/react/handbook/composition) prop to override the default rendered element.
- A [callback](/react/handbook/styling#css-classes) on the `className` prop that enables passing dynamically generated CSS classes based on the component's state.
- Adds [data-attributes](https://base-ui.com/react/handbook/styling#data-attributes) that map to the component's state that can be used as style hooks.
```

docs/src/app/(private)/experiments/custom-components.tsx
```diff
@@ -0,0 +1,87 @@
...
+      state,
+      className,
+      ref: forwardedRef,
+      props: {
+        ...otherProps,
+        style: {
+          ...style,
+          fontWeight,
+        },
+      },
```
atomiks, line 60: `ref` is considered a "regular" prop now. We do still need to support React 17-18, so not sure if sticking in the same object causes issues. Also, for `extraProps`, I was referring to spreading them inside the same top-level object passed to `useRenderer`

docs/src/app/(private)/experiments/custom-components.tsx
```diff
@@ -0,0 +1,87 @@
...
+      styleHookMapping: {
+        excludedProp() {
+          return null;
+        },
+      },
```
atomiks, line 65: I'm not the biggest fan of this prop name (`styleHookMapping`) - not sure if "style hook" is suitable for a public API, and `Map` instead of `Mapping` might be better

docs/src/app/(private)/experiments/custom-components.tsx
```diff
@@ -0,0 +1,87 @@
...
+      bold: 700,
+    }[weight];
+
+    const state = React.useMemo(
+      () => ({ weight, size, excludedProp }),
+      [weight, size, excludedProp],
+    );
+
+    const { renderElement } = useRenderer({
+      render: render ?? 'p',
```
atomiks, line 50: I've been wondering if we should get rid of the string fallbacks. We have to maintain a map in `defaultRenderFunctions` which increases bundle size and is not necessary given `render` only supports `HTMLAttributes<any>` now. i.e., they should just do `render ?? <p />`

docs/src/app/(private)/experiments/custom-components.tsx
mnajdova: I liked the separation, just so it's clear what's top level API vs other props spread, e.g. state is not passed on the element, but everything else is, it may be confusing.

mnajdova: For ref, maybe yes, we can include it in the props, I don't see how it could cause issues even for React 17-18, we are still receiving it in the hook, it doesn't matter where it came from, no? @michaldudak can you think of any issue with this?

docs/src/app/(private)/experiments/custom-components.tsx
mnajdova: The most clear name would be: `stateToDataAttributesMap`, to be very explicit, but it's a strange name 😅 @aarongarciah or @colmtuite do you have some suggestions?

docs/src/app/(private)/experiments/custom-components.tsx
mnajdova: This is a great idea 💯 I also wondered why we have that defaultRenferFunction mapping. I think we can update the demo and just not support it for the external use-cases. Is there any benefit on having this as a string?

michaldudak: `"p"` is slightly better than `<p />` as it's mapped to `(props) => <p {...props} />`. Since we use the longhand versions internally, it's better to have these functions defined in one place to avoid recreating them in each component. So a publicly exported string => render fn map doesn't really add to bundle size.

aarongarciah: It probably appeals to users that are used to `<Box as="p">`

atomiks: The problem is the map is non-exhaustive and doesn't cover every tag. I'm finding I also have to constantly add new ones. By default, `<tag />` does and works out of the box. The string doesn't seem worth it

docs/src/app/(private)/experiments/custom-components.tsx
michaldudak: Yeah, "hook" has a strong different meaning in React world, so we can think of something else. `stateAttributes`?

docs/src/app/(private)/experiments/custom-components.tsx
michaldudak: > I liked the separation, just so it's clear what's top level API vs other props spread, e.g. state is not passed on the element, but everything else is, it may be confusing.

I like it as well. Plus, it removes the need to spread and extract props from the hook's parameters, so it should save a few CPU cycles.

As for the ref, I can't think of a scenario where it won't work when placed among other props.

docs/src/app/(private)/experiments/custom-components.tsx
michaldudak: OK, it's a good point for external users as we won't map all the tags. I'd leave them for `useComponentRenderer` (or replace them with a direct reference to `defaultRenderFunctions`)

docs/src/app/(private)/experiments/custom-components.tsx
michaldudak: Technically, you could create any attribute (including `class`), so "dataAttributes" is not entirely correct. We just happen to use only data attributes.

docs/src/app/(private)/experiments/custom-components.tsx
mnajdova: Move the ref to the props object.

docs/src/app/(private)/experiments/custom-components.tsx
aarongarciah: I see! `stateAttributes` makes sense then.

mnajdova: Renamed to `stateDataAttributes`, I am still missing the mapping in it, but people can figure out how to define them by using the type. Maybe it's even worth adding demo about it on the page. What do you think?

docs/src/app/(public)/(content)/react/utils/use-renderer/demos/usage/css-modules/index.tsx
```diff
@@ -0,0 +1,73 @@
+'use client';
+import * as React from 'react';
+import { useRenderer, RenderProp } from '@base-ui-components/react/use-renderer';
+import styles from './index.module.css';
+
+type Weight = 'light' | 'regular' | 'bold';
+type Size = 'small' | 'medium' | 'large';
+
+type TextState = {
+  weight: Weight;
```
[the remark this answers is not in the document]
mnajdova: I added color inside this too, which is a state that changes when clicking the text. I am out of ideas what else to include, maybe we can have a demo with a different component, more complex to showcase these features. I felt character counter, although could be state, it's not really valuable while styling.

colmtuite:
Fine to me pending changing the name to useRender as per discussion. 🚀

colmtuite (approved):

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.module.css
```diff
@@ -0,0 +1,5 @@
+.Text {
+  font-size: 0.875rem;
+  line-height: 1rem;
+  color: var(--color-gray-900);
+}
```
aarongarciah, line 5: `strong` has `font-weight: 700` by default, but it looks ugly. I'd use `500` or `600` instead.


```suggestion
  strong& {
    font-weight: 500;
  }
}
```

|before|after|
|---|---|
|deploy-preview-1418--base-ui netlify app_react_utils_use-render (3)|localhost_3005_react_utils_use-render|

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+The hook returns a function that when called returns the element that should be rendered.
+
+## Usage
+
+### Supporting the `render` prop
+
+This is an example of a Text component that provides the support for the render prop.
+
+<Demo path="./demos/render" />
```
aarongarciah, line 54: I think this example makes it look more complex than it really is, due to imports, types, styles, etc. I wonder if it'd be better to just have a code snippet and not a live example for this. Something like:

````suggestion
This is an example on how to create your own `render` prop:

```jsx
import { useRender } from '@base-ui-components/react/use-render';

function Text({ render, ...otherProps }) {
  const { render, ...otherProps } = props;

  const { renderElement } = useRender({
    render: render ?? <p />,
    props: otherProps,
  });

  return renderElement();
}
```

It'd be used as follows:

```jsx
<Text>I'm a paragraph</Text>
<Text render={<span />}>I'm a span</Text>
```
````


The downside is not being able to open it on CodeSandbox, but not sure if it's that important.

File: docs/src/app/(public)/(content)/react/utils/use-render/page.mdx [hunk 16]
```diff
@@ -0,0 +1,89 @@
...
+
+### Generating data attributes
+
+In the following demo, the Text component provides more Base UI features, like adding data attributes and `className` callback where developers can have access to the internal state of the component.
```
aarongarciah, line 58:
```suggestion
`useRender` can also be used to automatically add data attributes based on the component's internal state and support a callback in the `className` prop where developers can access the component's internal state.
```

It's weird that the paragraph talks about the className callback but the it's not present in the demo.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,89 @@
...
+  state,
+  stateAttributesMap: {
+    size: (value) => ({ [`data-size-${value}`]: '' }),
+    color: (value) => ({ [`data-color-${value}`]: '' }),
+  },
+  props,
+});
+```
+
+Then, developers can target the data attributes like this:
```
aarongarciah, line 76: Seeing `data-size-small` might confuse users into thinking the generated data attributes have this shape by default (instead of `data-size='small'`). I'm not sure if this paragraph and the code block below are necessary.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+## Usage
+
+### Supporting the `render` prop
+
+This is an example of a Text component that provides the support for the render prop.
+
+<Demo path="./demos/render" />
+
+### Generating data attributes
```
aarongarciah, line 56: I think we could come up with a better demo. Clicking a text is not very common and it's hard to discover. We could also simplify it and only leave one "state" (color or size, but not both).

We could also use this demo only for data attributes and leave the className callback as paragraph + code snippet in a new section. This could also improve the discoverability of the className callback docs.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  return renderElement();
+}
+
+export default function ExampleText() {
+  return (
+    <div>
+      <Text className={styles.Text}>Text component rendered as a paragraph tag</Text>
+      <Text className={styles.Text} render={<strong />}>
```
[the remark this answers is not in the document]
mnajdova: For sure 👍

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
michaldudak: It could be an (another) argument for having the ability to select code to display in a demo preview. This way we could have the best of both - a concise code snippet that's runnable and can be opened in CSB. I could take a look at this next if you agree this could be handy.

aarongarciah: > It could be an (another) argument for having the ability to select code to display in a demo preview

This would be perfect.

> code snippets are not verified in any way (not executed, linted, typechecked, etc.), so they could be invalid (or inadvertently become invalid in the future after changes to the API)

Yeah, it's a trade-off. Sometimes live demos are an overkill and makes things harder to convey. I'll leave the live demo for now.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
mnajdova: Yeah, the thing I was trying to add here is that people can use internal state as well, I couldn't come up with an idea for internal state for Text component :) I am ok with simplifying it.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
mnajdova: Maybe we need a better description of what it is about.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
mnajdova: Maybe we can extend the demo on the usage to show className callback where a class is added fort he large variant. What do you think?

docs/src/app/(public)/(content)/react/utils/use-render/demos/data-attributes/css-modules/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+type Size = 'small' | 'medium' | 'large';
+type Color = 'default' | 'active';
+
+type TextState = {
+  size: Size;
+  color: Color;
+};
+
+type TextProps = {
+  className: string | ((state: TextState) => string);
```
aarongarciah, line 15: Can `useRender` export a type for `className`?

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
aarongarciah: I changed the demo to be a counter. Also added a separate demo for the className callback.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
aarongarciah: I updated this section and removed the CSS code block.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
aarongarciah: The className callback is now a separate demo.

docs/src/app/(public)/(content)/react/utils/use-render/demos/data-attributes/css-modules/index.tsx
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable react/button-has-type, jsx-a11y/control-has-associated-label */
```
aarongarciah, line 1: `render ?? <button />` requires to disable these rules, or add the required props to `<button />` but it's also weird.

packages/react/src/index.ts
```diff
@@ -26,4 +26,5 @@ export * from './tabs';
 export * from './toggle';
 export * from './toggle-group';
 export * from './tooltip';
+export * from './use-render';
```
aarongarciah, line 29: Why `useRender` is not inside `/utils`?

aarongarciah:
Docs updated. Highlights:
- The basic `render` prop demo has been simplified.
- The data-attrs and className demos are now two separate demos, for better discoverability. The end result in both demos is the same, but achieved differently.
- The data-attrs and className use a Counter button instead of a clickable Text component. Not my favorite example but I think it's better than the previous one.
- New "Migrating from Radix" section added.
- Other copy updates e.g. replacing the "Usage" headline with "Examples", which we use in all components.

---

I'm still not happy with the subtitle ("Utility for adding Base UI-like features to custom components") and the intro bullet points below. I think 99% os users will use `useRender` only to implement their own `render` prop, and the state-based data-attrs and className callback are only nice to have features for a minority of users i.e. only useful for users building unstyled primitives, which are not very numerous? The subtitle and bullet points probably makes `useRender` look more complex that it should because of this.

One of the reasons why it's so hard to come up with simple examples for the data-attrs and className callback demos is because it's mostly useful for unstyled primitives, and unstyled primitives already exist on Base UI, and tend to be complex (or at least require a lot of boilerplate to be fully accessible).

Note: state-based data attributes can be useful for styled components that want to have self-contained styles based on data attributes (`.Text[data-size="large"]`), but IMO that doesn't justify such a feature in `useRender`.

After having though about `useRender` for a while, shipping `useRender` without data-attrs and className callback could be an option. I'm curious to know your thoughts cc @mnajdova @michaldudak @colmtuite

docs/src/app/(public)/(content)/react/utils/use-render/demos/data-attributes/css-modules/index.tsx
mnajdova: Why not moving the type and aria-label from the props here?

mnajdova: Maybe then list the file it the eslint ignore config? I don't want people to have to see it in the demo.

docs/src/app/(public)/(content)/react/utils/use-render/demos/data-attributes/css-modules/index.module.css
```diff
@@ -0,0 +1,38 @@
+.Button {
```
mnajdova, line 1: Can we make the counter text width fixed, e.g. limit it to support up to 3 digits, so that it doesn't jump when we click the button.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
```diff
@@ -0,0 +1,32 @@
+'use client';
+import * as React from 'react';
+import { useRender } from '@base-ui-components/react/use-render';
+import styles from './index.module.css';
+
+type TextProps = {
+  className?: string;
+  render?: useRender.RenderProp<Record<string, any>>;
```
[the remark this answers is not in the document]
aarongarciah: Updated to use `unknown`. What do you mean by state? Using a separate state type as in the other demos? This demo has no state. In scenarios where state is not expected like this Text component, I think users should be able to use the `useRender.RenderProp` type without having to pass a state type.

docs/src/app/(public)/(content)/react/utils/use-render/demos/data-attributes/css-modules/index.tsx
aarongarciah: Done in https://github.com/mui/base-ui/pull/1418/commits/ab396658da035c6564374797c397480976051ff3. Not sure if the preference would be to ignore individual demos instead of the whole /react folder for the `jsx-a11y/control-has-associated-label` rule.

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * The render prop or React element to override the default element.
+     */
+    render: RenderProp<State>;
+    /**
+     * The state of the component. It will be used as a parameter for the render and className callbacks.
+     */
+    state?: State;
+    /**
+     * Props to be spread on the rendered element.
```
michaldudak, line 48:
```suggestion
     * Props to be spread on the rendered element.
     * They are merged with the internal props of the component, so that event handlers
     * are merged, class names and styles are joined, and other external props overwrite the
     * internal ones.
```

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,58 @@
...
+    | keyof typeof defaultRenderFunctions;
+
+  export interface Settings<State, RenderedElementType extends Element> {
+    /**
+     * The class name to apply to the rendered element.
+     * Can be a string or a function that accepts the state and returns a string.
+     */
+    className?: string | ((state: State) => string);
+    /**
+     * The render prop or React element to override the default element.
```
michaldudak, line 40:
```suggestion
     * The React element or a function that returns one to override the default element.
```

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
+import type { ComponentRenderFn } from '../utils/types';
+import { useComponentRenderer } from '../utils/useComponentRenderer';
+import { defaultRenderFunctions } from '../utils/defaultRenderFunctions';
+import { CustomStyleHookMapping as StateDataAttributes } from '../utils/getStyleHookProps';
```
michaldudak, line 5:
```suggestion
import { CustomStyleHookMapping as StateAttributes } from '../utils/getStyleHookProps';
```

mj12albert:
> I'm still not happy with the subtitle ("Utility for adding Base UI-like features to custom components")

How about "Utility for creating custom components that support a render prop"?

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * The state of the component. It will be used as a parameter for the render and className callbacks.
+     */
+    state?: State;
+    /**
+     * Props to be spread on the rendered element.
+     */
+    props?: Record<string, unknown> & { ref?: React.Ref<RenderedElementType> };
+    /**
+     * A mapping of state to data attributes.
```
michaldudak, line 52:
```suggestion
     * An object that maps the state (passed with the `state` parameter) to attributes placed on the rendered component.
```

michaldudak:
> After having though about useRender for a while, shipping useRender without data-attrs and className callback could be an option.

It's not a bad idea, especially considering that state attributes and class names implementation is pretty simple and should be easily done in userland if needed (or we could provide separate utilities for them if devs want them).

packages/react/src/use-render/useRender.ts
aarongarciah: Done

packages/react/src/use-render/useRender.ts
aarongarciah: Done

packages/react/src/use-render/useRender.ts
aarongarciah: Done. Also updated in the props table: https://github.com/mui/base-ui/pull/1418/files#diff-a4b963fd83e20c9225af018585e14e171301cda8819bf4a23096b15a66ecdcb6R38

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,124 @@
...
+      description:
+        "CSS class applied to the element, or a function that returns a class based on the component's state.",
+    },
+    state: {
+      type: 'State',
+      description:
+        'The state of the component. It will be used as a parameter for the render and className callbacks.',
+    },
+    stateAttributesMap: {
+      type: 'StateAttributes<State>',
```
aarongarciah, line 38: AFAIK this type is not available for users so not sure if having `StateAttributes` here is helpful.

packages/react/src/use-render/useRender.ts
mnajdova: Yes, this is not auto-generated. Once we have Michal's work on api generation for hooks & functions, we can automatically extract it.

mnajdova:
> After having though about useRender for a while, shipping useRender without data-attrs and className callback could be an option.
> > It's not a bad idea, especially considering that state attributes and class names implementation is pretty simple and should be easily done in userland if needed (or we could provide separate utilities for them if devs want them).

Yeah, we can do this in the first iteration. @aarongarciah I will do the change on the hook & docs, if we decide to extend it in the future, we can use your examples.

mnajdova:
I've simplified the hook & the docs page to only cover the render prop, we can add more utils in the future if people ask for them.

.eslintrc.js
```diff
@@ -95,6 +95,12 @@ module.exports = {
         'import/no-relative-packages': 'off',
       },
     },
+    {
+      files: ['docs/src/app/(public)/(content)/react/utils/use-render/demos/**/*{.tsx,.js}'],
+      rules: {
+        'jsx-a11y/control-has-associated-label': 'off',
+      },
+    },
```
aarongarciah, line 103: We can undo this change.

```suggestion
```

packages/react/src/index.ts
mnajdova: @michaldudak @atomiks what are your thoughts? I though it as a "first class export" similar to the components, but having it in the utils makes sense as well, especially if we want to publicly make available more of the hooks.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,68 @@
...
+## API reference
+
+### Input parameters
+
+<PropsReferenceTable
+  data={{
+    render: {
+      type: 'RenderProp<State>',
+      description:
+        "Allows you to replace the component's HTML element\nwith a different tag, or compose it with another component.\n\nAccepts a \`ReactElement\` or a function that returns the element to render.",
```
aarongarciah, line 17:
```suggestion
        "The React element or a function that returns one to override the default element.",
```

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+import type { ComponentRenderFn } from '../utils/types';
+import { useComponentRenderer } from '../utils/useComponentRenderer';
+import { defaultRenderFunctions } from '../utils/defaultRenderFunctions';
+import { CustomStyleHookMapping as StateAttributes } from '../utils/getStyleHookProps';
```
aarongarciah, line 5: This import is unused.

```suggestion
```

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+function useRender<RenderedElementType extends Element>(
+  settings: useRender.Settings<RenderedElementType>,
+) {
+  const { render, props } = settings;
+  const { ref, ...extraProps } = props ?? {};
+
+  return useComponentRenderer({
+    render,
+    state: {} as Record<string, any>,
```
aarongarciah, line 18: off-topic: I wonder if `state` should be optional in `useComponentRenderer`

mnajdova: We probably didn't have use-case for it so far, as all components have some state, but we could consider it if we hit this need again.

packages/react/src/index.ts
mnajdova: We decided that it's best to export from a top level directory.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+import * as React from 'react';
+import { useRender } from '@base-ui-components/react/use-render';
+import styles from './index.module.css';
+
+type TextProps = {
+  render?: useRender.RenderProp<Record<string, unknown>>;
```
michaldudak, line 7: `Record<string, unknown>` is the default. Omitting it makes the code cleaner.

```suggestion
  render?: useRender.RenderProp;
```

michaldudak (commented):
It looks good in general. What do you think about adding another demo to show how to use the callback form with state?

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children: React.ReactNode;
+};
+
+function Text(props: TextProps) {
+  const { render, ...otherProps } = props;
```
atomiks, line 12: Default assignment `render = <p />` seems like a cleaner example to use throughout the docs? (Same for `button` ones)

```jsx
function Text(props: TextProps) {
  const { render = <p />, ...otherProps } = props;

  const { renderElement } = useRender({
    render,
    props: { ...otherProps, className: styles.Text },
  });
  // ...
}
```

packages/react/src/use-render/useRender.ts
atomiks: A lot of them actually define an empty object outside the component (in fact, this needs to be done here to prevent a new reference every render). It would be good to make it optional since many components have empty state objects

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
aarongarciah: The default is `Record<string, any>`, should it be updated to `Record<string, unknown>`?

https://github.com/mui/base-ui/pull/1418/files#diff-1ed79dc7239102d3f0a279612e8c0e98471ef4821c8427d59d47c46d16b90b13R24-R33

aarongarciah:
> What do you think about adding another demo to show how to use the callback form with state?

@michaldudak AFAIK `useRender` doesn't accept a `state` setting anymore so not sure if it makes sense. I guess this is weird since the `render` callback still has the state argument? 😅

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
mnajdova: Done.

mnajdova:
> @michaldudak AFAIK useRender doesn't accept a state setting anymore so not sure if it makes sense. I guess this is weird since the render callback shape still has the state argument? 😅

Yeah, maybe we should indeed keep the state attribute, as it is used not only for data attributes generation & className callback, but for the render prop too.

michaldudak:
> Yeah, maybe we should indeed keep the state attribute, as it is used not only for data attributes generation & className callback, but for the render prop too.

👍 yup, let's make the render prop fully featured as we won't be able to add this in another util later.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
mnajdova: Yep, changed.

mnajdova:
I added back the state and added demo with callback. I brought back the counter demo and added emoji based on the state, yeah I am not creative 😅 @aarongarciah let me know if you have a better idea.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,78 @@
...
+      type: 'RenderProp<State>',
+      description:
+        'The React element or a function that returns one to override the default element.',
+    },
+    props: {
+      type: 'Record<string, unknown>',
+      description:
+        'Props to be spread on the rendered element.\n\nThey are merged with the internal props\n of the component, so that event handlers are merged,\nclass names and styles are joined, and other external props overwrite the internal ones.',
+    },
+    state: {
```
aarongarciah, line 24: Should we mention that this parameter is optional?

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,53 @@
...
+  return useComponentRenderer({
+    render,
+    state: state as State,
+    ref: ref as React.Ref<RenderedElementType>,
+    extraProps,
+    customStyleHookMapping: Object.keys(state).reduce((acc, key) => {
+      acc[key as keyof State] = () => null;
+      return acc;
+    }, {} as CustomStyleHookMapping<State>),
+  });
```
atomiks, line 25: This calculation and `state` should be memo'd

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+function Counter(props: CounterProps) {
+  const { render, className, ...otherProps } = props;
+  const [count, setCount] = React.useState(0);
+  const odd = count % 2 === 1;
+  const state = React.useMemo(() => ({ odd }), [odd]);
+
+  const { renderElement } = useRender({
+    render: render ?? <button type="button" />,
```
aarongarciah, line 22: We should `<button type="button">` as a default assignment as we do in the other demo.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+export default function ExampleCounter() {
+  return (
+    <Counter
+      render={(props, state) => (
+        <button
+          {...props}
+          className={`${state.odd ? styles.odd : styles.even} ${props.className}`}
+        >
+          {props.children}
+          <span className={styles.suffix}>{state.odd ? '👎' : '👍'}</span>
```
aarongarciah, line 50: I'd choose between the class name and the emoji. One of them is enough. We can leave the emoji and remove the class name to simplify the demo.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [count, setCount] = React.useState(0);
+  const odd = count % 2 === 1;
+  const state = React.useMemo(() => ({ odd }), [odd]);
+
+  const { renderElement } = useRender({
+    render: render ?? <button type="button" />,
+    state,
+    props: {
+      ...otherProps,
+      className: `${className} ${styles.Button}`,
```
aarongarciah, line 26:
```suggestion
      className: `${styles.Button} ${className}`,
```

PS: some people will realize about `undefined` ending up in the DOM if we didn't pass one `className` in the demo 😅

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+The hook returns a function that when called returns the element that should be rendered.
+
+## Examples
+
+This is an example of a Text component supporting the `render` prop.
+
+<Demo path="./demos/render" />
+
+Developers can also use the callback version of the `render` prop.
```
aarongarciah, line 42: Should we mention `state`?

```suggestion
Developers can also use the callback version of the `render` prop to access the internal `state` of a component.
```

atomiks:
I think we need to ship `mergeReactProps` with this - the `Counter` demo's props spreading highlights why it'd be good. Maybe even built into the hook itself somehow instead?

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
mnajdova: What do you mean?

aarongarciah: Move it to `const { render = <button type="button" />, className, ...otherProps } = props;`

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
mnajdova: Fixed.

packages/react/src/use-render/useRender.ts
mnajdova: I've added the calculation in a useMemo, I will add a note for the state to be memoized in the docs, is that what you had in mind?

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+import * as React from 'react';
+import { useRender } from '@base-ui-components/react/use-render';
+import styles from './index.module.css';
+
+type TextProps = {
+  render?: useRender.RenderProp;
```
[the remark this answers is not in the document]
mnajdova: Fixed.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const [count, setCount] = React.useState(0);
+  const odd = count % 2 === 1;
+  const state = React.useMemo(() => ({ odd }), [odd]);
+
+  const { renderElement } = useRender({
+    render: render,
+    state,
+    props: {
+      ...otherProps,
+      className: `${styles.Button} ${className ?? ''}`,
```
aarongarciah, line 26: We can remove the `className`

```suggestion
  render?: useRender.RenderProp<CounterState>;
};

function Counter(props: CounterProps) {
  const { render = <button type="button" />, ...otherProps } = props;
  const [count, setCount] = React.useState(0);
  const odd = count % 2 === 1;
  const state = React.useMemo(() => ({ odd }), [odd]);

  const { renderElement } = useRender({
    render,
    state,
    props: {
      ...otherProps,
      className: styles.Button,
```

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const odd = count % 2 === 1;
+  const state = React.useMemo(() => ({ odd }), [odd]);
+
+  const { renderElement } = useRender({
+    render,
+    state,
+    props: {
+      ...otherProps,
+      className: `${styles.Button} ${className ?? ''}`,
+      type: 'button',
```
aarongarciah, line 27: We're setting `type: 'button'` here but also in the props destructuring above. Can we leave just one? Ideally this one.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
mnajdova: But this way we are overriding the user's class name no? How we can omit it?

mnajdova:
> I think we need to ship mergeReactProps with this - the Counter demo's props spreading highlights why it'd be good. Maybe even built into the hook itself somehow instead?

I think I would skip this for now, until we see people asking for it. We can add it separately, it doesn't change how the hook works.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
aarongarciah: The example doesn't pass a `className` to the Counter component anymore, so we can remove it.

mnajdova: But regardless, if we are showing a usable component, people will need to do this, no?

aarongarciah: > people will need to do this, no?

You would also need to handle the role, keyboard interactions, etc. if you change the rendered element. But I think that goes way beyond the purpose of the demo. I consider not managing class name in the same category as this stuff. We don't manage class names in the first demo (the Text one) to keep it focused on what's important.

It also depends on wether you want to showcase the Counter component as an unstyled primitive (`styles.Button` should be passed to the component) or as a styled component (as it is currently, class name is defined inside the component). Also, not all components are meant to be overridden, it depends on the use case.

I think the `className` adds more noise to the demo (which in complete honesty, is kinda silly but I couldn't come up with something better that doesn't exist in Base UI today e.g. collapsible, toggle, etc).

atomiks:
> I think I would skip this for now, until we see people asking for it. We can add it separately, it doesn't change how the hook works.

We can do it in a new PR, but I don't think we should skip it as it's key to making the hook useful for building primitive components. 

The problem is evident with the `onClick` handler here, which is actually promoting a bad practice for "primitive" components (those that accept any DOM prop):

```js
  const { renderElement } = useRender({
    render,
    state,
    props: {
      ...otherProps,
      // ...
      onClick: () => setCount((prev) => prev + 1),
    },
  });
```

```jsx
// Doesn't work:
<Counter onClick={console.log} />
```

A good way to solve this is to accept an array for `props`, which calls `mergeReactProps` internally, if we don't want to expose it. But if they use the function `render` form, they'll probably need a merging util anyway.

```js
useRender({
  props: [externalProps, internalProps] 
})
```

---

For types, perhaps instead of just `RenderProp`, they should have access to the `BaseUIComponentProps` type? Like:

```js
function Button(props: Button.Props) {
}

namespace Button {
  export interface Props extends useRender.ElementProps<'button'> {
    // ...
  }
} 
```

mnajdova:
> We can do it in a new PR, but I don't think we should skip it as it's key to making the hook useful for building primitive components.

Yeah fair enough, I will do it in a follow up PR, for now, I fixed the demo.

> For types, perhaps instead of just RenderProp, they should have access to the BaseUIComponentProps type? Like:

Makes sense, exported an interface that can be extended.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
mnajdova: On the contrary we discussed just now that the demo should also show how people should merge props. For now I manually merged all props, but we are discussing adding this as a first class support in the hook, see https://github.com/mui/base-ui/pull/1418#issuecomment-2692010558. This way we would have feature parity with the Slot component from Radix UI.

mnajdova:
> But if they use the function render form, they'll probably need a merging util anyway.

This for me was the most important bit. Even if we do handle the merging internally, it will anyway be needed for the end-user.

michaldudak:
I'm also leaning towards making `mergeReactProps` public (it could be called `mergeProps` BTW). It'll emphasize we support the function form of render props the same way we do the element one.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
```diff
@@ -14,37 +15,28 @@ interface CounterProps extends useRender.ElementProps<CounterState> {
...
+        children: (
+          <React.Fragment>
+            Counter: <span>{count}</span>
+          </React.Fragment>
+        ),
+        onClick: () => setCount((prev) => prev + 1),
+        'aria-label': `Count is ${count}, click to increase.`,
+      },
+      otherProps,
+    ),
```
atomiks, line 39: `otherProps` goes first

atomiks: It's neater with Prettier to place the variable first

mnajdova: > Devs nowadays may be more familiar with object spread pattern that works from left to right than Object.assign that mergeProps currently mimics.

This was my initial thought as well, if I spread this would be the order.

mnajdova: Also for props that should replace (like aria-label in the demo), you want other props to override, so it needs to comes second, no?

atomiks: In the `mergeReactProps` implementation, the ones on the left take precedence. So `aria-label` from the user will override our `aria-label` when it's specified in the object later

Re-thinking it, I guess it does make sense for ones specified later in the arguments to override earlier ones instead. Prettier formatting is only a bit worse

mnajdova: I feel strongly that we should change this then, as Michal mentioned, object spread, Object.assign etc all work the other way around. Are we ok with changing this? If yes, I can apply the change as part of this PR.

mnajdova: I opened a new PR for this, there were too many changes: https://github.com/mui/base-ui/pull/1533

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+Developers can also use the callback version of the `render` prop to access the internal `state` of a component.
+
+<Demo path="./demos/render-callback" />
+
+### Merging the props
+
+You can use the `mergeProp` utility to safely merge the props.
+It takes care of concatenating the class names, as well as merging the style prop and the event handlers.
+You can use it both in the `useRender` hook and the callback version of the render prop, e.g.
```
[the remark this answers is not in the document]
mnajdova: Makes sense 👌

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
mnajdova: I don't think we show optional/required parameter anywhere on the docs yet. We could think of how we want to do this separately. I think it would be better if we just mention the required params/props on the docs instead of the other way around. The type should be enough to guide people.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,99 @@
...
+  return <Comp {...props} />;
+}
+
+// Usage
+<Button asChild>
+  <a href="/contact">Contact</a>
+</Button>;
+```
+
+In Base UI, you use `useRender` to implement your own `render` prop. The example below is the equivalent implementation to the Radix example above.
```
aarongarciah, line 87: re "The example below is the equivalent implementation to the Radix example above", do you think this example should use the `mergeProps` util to be fully equivalent to the Radix example?

mnajdova: Hmm, the example is to simple to show the usage of mergeProps as we are not defining any props in the component itself. Maybe the other examples should go in more details and we can keep this one simple, what do you think?

aarongarciah: 👍

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,99 @@
+# useRender
+
```
atomiks, line 2: My take on the docs:

````mdx
# useRender

<Subtitle>Hook for enabling a `render` prop in custom components.</Subtitle>

<Meta name="description" content="Utility for supporting a render prop in custom components." />

The `useRender` hook enables custom components to provide a `render` prop to override the default rendered element.

## API reference

### Input parameters

<PropsReferenceTable
data={{
    render: {
      type: 'RenderProp<State>',
      description:
        'The React element or a function that returns one to override the default element.',
    },
    props: {
      type: 'Record<string, unknown>',
      description:
        'Props to be spread on the rendered element.\n\nThey are merged with the internal props\n of the component, so that event handlers are merged,\nclass names and styles are joined, and other external props overwrite the internal ones.',
    },
    state: {
      type: 'State',
      description:
        'The state of the component. It will be used as a parameter for the render callback.',
    },
  }}
/>

### Return value

The hook returns a `renderElement` function that returns the element to be rendered when called.

## Examples

A `render` prop for a custom Text component lets consumers use the `render` prop to replace the default rendered `p` element with a different element tag.

<Demo path="./demos/render" />

The callback version of the `render` prop also enables accessing the `state` of a component.

<Demo path="./demos/render-callback" />

### Merging props

The `mergeProps` function merges two or more sets of React props together. It safely merges three types of props:

1. Event handlers, so that all are invoked
2. `className` strings
3. `style` properties

The `ref` prop and other mergeable properties like `aria-describedby` are not handled.

It merges objects from left to right, so that subsequent objects' propertes in the arguments overwrite previous ones' properties.

Merging props is useful in custom components, as well as the callback version of the `render` prop for any Base UI component.

```jsx {7-7} title="Using mergeProps in the render callback"
import { mergeProps } from "@base-ui-components/react/merge-props";
import styles from "./index.module.css";

function Button() {
  return (
    <Component
      render={(props, state) => (
        <button {...mergeProps(props, { className: styles.Button })} />
      )}
    />
  );
}
```

### Migrating from Radix

Radix uses the `asChild` prop, while Base UI uses the `render` prop. Learn more about how composition works in Base UI in the [composition guide](/react/handbook/composition).

In Radix, the `Slot` component lets you implement an `asChild` prop.

```jsx title="Custom asChild prop with Radix"
import { Slot } from "radix-ui";

function Button({ asChild, ...props }) {
  const Comp = asChild ? Slot.Root : "button";
  return <Comp {...props} />;
}

// Usage
<Button asChild>
  <a href="/contact">Contact</a>
</Button>;
```

In Base UI `useRender` lets you implement a `render` prop. The example below is the equivalent implementation to the Radix example above.

```jsx title="Custom render prop with Base UI"
import { useRender } from "@base-ui-components/react/use-render";

function Button({ render = <button />, ...props }) {
  const { renderElement } = useRender({ render, props });
  return renderElement();
}

// Usage
<Button render={<a href="/contact">Contact</a>} />;
```
````

File: docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx [hunk 49]
```diff
@@ -0,0 +1,29 @@
...
+function Text(props: TextProps) {
+  const { render = <p />, ...otherProps } = props;
+
+  const { renderElement } = useRender({
+    render,
+    props: mergeProps({ className: styles.Text }, otherProps),
+  });
+
+  return renderElement();
+}
```
atomiks, line 20: I think we need to mention `React.forwardRef` is required in React 17 and 18.

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,61 @@
...
+  const customStyleHookMapping = React.useMemo(() => {
+    return Object.keys(state ?? {}).reduce((acc, key) => {
+      acc[key as keyof State] = () => null;
+      return acc;
+    }, {} as CustomStyleHookMapping<State>);
+  }, [state]);
+
+  return useComponentRenderer({
+    render,
+    state: (state ?? {}) as State,
```
atomiks, line 25: The fallback object `?? {}` should be defined outside of the component

atomiks (commented):
I forgot about the fact that we frequently need to use the `useForkRef` utility internally when building our components. `mergeProps` doesn't handle merging refs as mentioned, and doesn't memo it. Could we handle merging multiple refs in `useRender`?

It actually _should_ handle this by default because it internally uses `useForkRef` in `useRenderPropForkRef`, but it doesn't allow an array of refs. So we could easily let it accept an array and make this happen automatically

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  ['aria-label']?: string;
+}
+
+function Counter(props: CounterProps) {
+  const { render = <button />, ...otherProps } = props;
+  const [count, setCount] = React.useState(0);
+  const odd = count % 2 === 1;
+  const state = React.useMemo(() => ({ odd }), [odd]);
+
+  const defaultProps: mergeProps.Props<'button'> = {
```
atomiks, line 23: I suppose it would be good to explain the types of `useRender.ElementProps` and `mergeProps.Props` and how they're different

mnajdova:
> It actually should handle this by default because it internally uses useForkRef in useRenderPropForkRef, but it doesn't allow an array of refs. So we could easily let it accept an array and make this happen automatically

@michaldudak @atomiks do you think we should add additional param refs that allows an array, and if people don't need it they can just add the ref as part of the props.

I implemented ☝️, the useComponentRender, which is internal accepts both one and array of refs in the 'ref' param.

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,71 @@
...
+     * Props to be spread on the rendered element.
+     * They are merged with the internal props of the component, so that event handlers
+     * are merged, class names and styles are joined, and other external props overwrite the
+     * internal ones.
+     */
+    props?: Record<string, unknown> & { ref?: React.Ref<RenderedElementType> };
+  }
+}
+
+export { useRender };
```
atomiks, line 71: We use `export namespace` to still use `export function ...` which exports both

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,71 @@
...
+     * The state of the component. It will be used as a parameter for the render callback.
+     */
+    state?: State;
+    /**
+     * Props to be spread on the rendered element.
+     * They are merged with the internal props of the component, so that event handlers
+     * are merged, class names and styles are joined, and other external props overwrite the
+     * internal ones.
+     */
+    props?: Record<string, unknown> & { ref?: React.Ref<RenderedElementType> };
```
atomiks, line 67: Maybe our internal `GenericHTMLProps` type is better than `Record<string, unknown>` here (I think they should support any tag's props without a TS error 🤔)

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,71 @@
...
+) {
+  const { render, props, state, refs } = settings;
+  const { ref, ...extraProps } = props ?? {};
+
+  const customStyleHookMapping = React.useMemo(() => {
+    return Object.keys(state ?? {}).reduce((acc, key) => {
+      acc[key as keyof State] = () => null;
+      return acc;
+    }, {} as CustomStyleHookMapping<State>);
+  }, [state]);
```
atomiks, line 23: Wonder if `useComponentRenderer` should have an option to turn off applying `data-*` attributes to avoid runtime overhead here

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,110 @@
...
+      type: 'RenderProp<State>',
+      description:
+        'The React element or a function that returns one to override the default element.',
+    },
+    props: {
+      type: 'Record<string, unknown>',
+      description:
+        'Props to be spread on the rendered element.\n\nThey are merged with the internal props\n of the component, so that event handlers are merged,\nclass names and styles are joined, and other external props overwrite the internal ones.',
+    },
+    state: {
```
atomiks, line 25: Should be `refs`*? Wonder if it's confusing that `params.props.ref` and params.refs` are both a thing?

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+1. Event handlers, so that all are invoked
+2. `className` strings
+3. `style` properties
+
+The `ref` prop is not handled by `mergeProps`, but refs are handled when the ref is contained within the `props` object passed to `useRender`.
+
+`mergeProps` merges objects from left to right, so that subsequent objects' propertes in the arguments overwrite previous ones' properties.
+
+Merging props is useful in custom components, as well as the callback version of the `render` prop for any Base UI component.
```
atomiks, line 63: This should be updated based on the new `refs` option param. I also made a typo in `subsequent objects' properties`

atomiks: I think somehow mentioning that React 19 doesn't require `forwardRef` anymore, and that the default forwarded ref simply goes inside `props` could be useful?

````suggestion
`mergeProps` merges objects from left to right, so that subsequent objects' properties in the arguments overwrite previous ones. Merging props is useful when creating custom components, as well as inside the callback version of the `render` prop for any Base UI component. 

The `ref` prop is merged separately using the `refs` option in `useRender`, which holds an array of refs.

```js
// React 19
function Text({ render = <p />, ...props }: TextProps) {
  const internalRef = React.useRef<HTMLElement | null>(null);

  const { renderElement } = useRender({
    render,
    props,
    refs: [internalRef],
  });
  
  return renderElement();
}

// React 17 and 18
const Text = React.forwardRef(function Text(
  { render = <p />, ...props }: TextProps, 
  forwardedRef: React.ForwardedRef<HTMLElement>
) {
  const internalRef = React.useRef<HTMLElement | null>(null);

  const { renderElement } = useRender({
    render,
    props,
    refs: [forwardedRef, internalRef],
  });
  
  return renderElement();
});
```
````

**Comment by user-0a** [comment]
+1, Looking forward to this. Our use case is we have a custom Button component built w/ our styling and we occasionally want to render it as a Link.

File: packages/react/src/use-render/useRender.ts (hunk 54 above)
Comment (mnajdova) [reply packages/react/src/use-render/useRender.ts]: Makes sense, added optional boolean setting for this.

File: docs/src/app/(public)/(content)/react/utils/use-render/page.mdx (hunk 55 above)
Comment (mnajdova) [reply docs/src/app/(public)/(content)/react/utils/use-render/page.mdx:25]: I wanted to make it simple for both scenarios, e.g. ref is already in the props and people don't have additional regs, they don't need to specify anything, if they have additional refs, they can use the refs array.

File: packages/react/src/use-render/useRender.ts (hunk 52 above)
Comment (mnajdova) [reply packages/react/src/use-render/useRender.ts]: Ah nice, I added it, I think I saw this in some menu components.

File: docs/src/app/(public)/(content)/react/utils/use-render/page.mdx (hunk 55 above)
Comment (atomiks) [reply docs/src/app/(public)/(content)/react/utils/use-render/page.mdx:25]:
```suggestion
    refs: {
```
(`state` is duplicated below)

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
```diff
@@ -0,0 +1,36 @@
+'use client';
+import * as React from 'react';
+import { useRender } from '@base-ui-components/react/use-render';
+import { mergeProps } from '@base-ui-components/react/merge-props';
+import styles from './index.module.css';
+
+// ElementProps contains the 'render' prop type
+interface TextProps extends useRender.ElementProps {
+  children: React.ReactNode;
+}
```
atomiks, line 10: `ElementProps` is only an object with a `render` prop, but I was thinking it should be similar to `BaseUIComponentProps` which accepts the HTML/DOM props. Otherwise, they can just add `render: useRender.RenderProp` themselves in the interface

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const Text = React.forwardRef(function Text(
```
atomiks, line 12: Not sure how we can default to React 19 and show that React 17/18 need to be different? (Of course, this works for all the versions, but is using a soon-deprecated API)

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -0,0 +1,139 @@
...
+
+  const { renderElement } = useRender({
+    render,
+    props,
+    refs: [forwardedRef, internalRef],
+  });
+
+  return renderElement();
+});
+```
```
atomiks, line 92:
````suggestion
#### Merging refs

When building custom components, you often need to control a ref internally while still letting external consumers pass their own ref. Ref merging lets both parties have access to the underlying DOM element—the `refs` option in `useRender` enables this, which holds an array of refs to be merged together.

In the React 19, `React.forwardRef()` is not needed when building primitive components, as the external ref prop is already contained inside `props`. Your internal ref can be passed to `refs` to be merged with `props.ref`:

```tsx title="React 19"
function Text({ render = <p />, ...props }: TextProps) {
  const internalRef = React.useRef<HTMLElement | null>(null);

  const { renderElement } = useRender({
    render,
    props,
    refs: [internalRef],
  });

  return renderElement();
}
```

In older versions of React, you need to use `React.forwardRef()` and add the forwarded ref to the `refs` array along with your own internal ref:

```tsx title="React 18 and 17"
const Text = React.forwardRef(function Text(
  { render = <p />, ...props }: TextProps,
  forwardedRef: React.ForwardedRef<HTMLElement>,
) {
  const internalRef = React.useRef<HTMLElement | null>(null);

  const { renderElement } = useRender({
    render,
    props,
    refs: [forwardedRef, internalRef],
  });

  return renderElement();
});
```
````

File: packages/react/src/utils/useComponentRenderer.ts [hunk 60]
```diff
@@ -61,12 +65,16 @@ export function useComponentRenderer<
     propGetter = (props) => props,
     extraProps,
     customStyleHookMapping,
+    skipGeneratingStyleHooks = false,
   } = settings;
 
   const className = resolveClassName(classNameProp, state);
   const styleHooks = React.useMemo(() => {
+    if (skipGeneratingStyleHooks) {
+      return {};
```
michaldudak, line 74: `{}` can be defined outsite of the hook so it's not recreated every time it runs.

packages/react/src/merge-props/mergeProps.ts
```diff
@@ -193,3 +193,5 @@ function mergeClassNames(ourClassName: string | undefined, theirClassName: strin
 function isSyntheticEvent(event: unknown): event is React.SyntheticEvent {
   return event != null && typeof event === 'object' && 'nativeEvent' in event;
 }
+
+export { mergeProps };
```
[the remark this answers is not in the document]
michaldudak: Could you show an example where this is a problem? I was going to standardize how we export component and hooks and this can change the approach.

packages/react/src/use-render/useRender.ts
```diff
@@ -0,0 +1,62 @@
...
+  export type RenderProp<State = Record<string, unknown>> =
+    | ComponentRenderFn<React.HTMLAttributes<any>, State>
+    | React.ReactElement<Record<string, unknown>>
+    | keyof typeof defaultRenderFunctions;
+
+  export interface ElementProps<State = Record<string, unknown>> {
+    render?: RenderProp<State>;
+  }
+
+  export interface Settings<State, RenderedElementType extends Element> {
```
michaldudak, line 41: We usually call it `Parameters`

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
mnajdova: Maybe I can revert back this change, now that we have the isntructions in the mdx.

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
mnajdova: I defined it as this: https://github.com/mui/base-ui/pull/1418/files#diff-1ed79dc7239102d3f0a279612e8c0e98471ef4821c8427d59d47c46d16b90b13R38-R52. I used `React.ComponentPropsWithRef` instead of `ComponentPropsWithoutRef` which is used in the `BaseUIComponentProps`, do you know why we didn't added the ref in that type?

atomiks: If the props object can contain a ref, I think it should be the case for `BaseUIComponentProps` too—unless there were typing issues when rendering as a different tag and spreading the props?

atomiks (approved):
This looks good! I have a few various nits, mainly with the docs, but I'm going to create a new PR with the changes, as this one has become really long

docs/src/app/(public)/(content)/react/utils/use-render/demos/render/css-modules/index.tsx
mnajdova: Yeah, I though the same. @michaldudak are you aware of any issues around this?

mnajdova:
I am merging, let's create a follow up PR on https://github.com/mui/base-ui/pull/1418#discussion_r1988713894 if there was a reason for not defining the ref as part of the props.

PR #1422, opened 2025-02-05
[core] Fix stylelint violations

michaldudak:
A follow-up to #1184. I disabled a couple of rules (per https://github.com/mui/base-ui/pull/1184#issuecomment-2569022722) and fixed the others. We'll see if this ruleset feels OK to work with and update the Core stylelint config.


atomiks (approved):
Wasn't sure if modern notation has acceptable enough browser support (`color: rgb(0 0 0 / 0.87)`) to use for docs, but I think it's a few years old now 🤔

PR #1428, opened 2025-02-06
[Avatar] Add missing Avatar export

Gomah:
Thank you for the `alpha.6 ` release, just noticed that `Avatar` was missing from the main export when testing it


mj12albert (approved):
@Gomah Thanks for this ~

Luckily `import { Avatar } from '@base-ui-components/react/avatar'` still works

PR #1431, opened 2025-02-07
[test] Verify root exports

michaldudak:
Added a test that verifies if everything with a named export is also exported from the root to prevent problems like #1428.


packages/react/src/index.test.ts
```diff
@@ -0,0 +1,44 @@
...
+              undefined,
+              `${exportKey} (from ${importSpecifier}) was not found in root exports`,
+            );
+          });
+        }),
+    );
+
+    // Ensure the utils are also exported
+    expect(BaseUI.utils).not.to.equal(undefined);
+  });
```
atomiks, line 43: In our releases/npm package I see all the utils are published. Should any utils even be exported to the consumer? I thought they were intended for private/internal usage

michaldudak: We need a subset of utils published as they are required to build applications: `useForkRef`, `useId`, maybe `useControlled`.

mnajdova: Maybe we can have a list of utils we expect to be exported here and fail if there are more or less exported, that way when we add a new util, we will need to make decision of whether it should be public as the test would automatically fail.

michaldudak: It might be easier to manage if we had internal utils in a separate directory. This way we'll be able to treat all contents of ./utils as public (which will make us more careful about introducing breaking changes).

michaldudak: Anyway, this is outside the scope of this PR. We can discuss it separately.

packages/react/src/index.ts
```diff
@@ -26,3 +26,4 @@ export * from './tabs';
 export * from './toggle';
 export * from './toggle-group';
 export * from './tooltip';
+export * as utils from './utils';
```
mnajdova, line 29: Why did we export all utils? Was it intentional?

michaldudak: We don't export all. index.ts in `./utils` has just a few of them.

atomiks (approved):

Issue #661, opened 2024-09-30
[toolbar] Implement Toolbar


colmtuite:
Toolbar pattern: https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/

### API
```jsx
<Toolbar.Root>
  <Toolbar.Group>
    <Toolbar.Button />
    {/* Focusability of disabled controls is handled via a focusableWhenDisabled prop (default true) */}
    <Toolbar.Button focusableWhenDisabled={false} />
  </Toolbar.Group>
  <Toolbar.Separator />
  <Toolbar.Link />
  <Toolbar.Separator />
  <ToggleGroup>
    <Toolbar.Button render={<Toggle />} />
    <Toolbar.Button render={<Toggle />} />
  </ToggleGroup>
  <Toolbar.Separator />
  <Menu.Root>
    <Toolbar.Button render={<Menu.Trigger />} />
    <Menu.Portal>
      <Menu.Positioner>
        <Menu.Popup>…</Menu.Popup>
      </Menu.Positioner>
    </Menu.Portal>
  </Menu.Root>
  <Toolbar.Separator />
  <NumberField.Root>
    <NumberField.Group>
      <NumberField.Decrement />
      <Toolbar.Input render={<NumberField.Input />} />
      <NumberField.Increment />
    </NumberField.Group>
  </NumberField.Root>
  <Toolbar.Separator /> 
  <Select.Root>
    <Toolbar.Button render={<Select.Trigger />} />
    <Select.Portal>
      <Select.Positioner>
        <Select.Popup>…</Select.Popup>
      </Select.Positioner>
    </Select.Portal>
  </Select.Root>
  <Toolbar.Separator />
  <Dialog.Root>
    <Toolbar.Button render={<Dialog.Trigger />} />
    <Dialog.Portal>
      <Dialog.Positioner>
        <Dialog.Popup>…</Dialog.Popup>
      </Dialog.Positioner>
    </Dialog.Portal>
  </Dialog.Root>
  <Toolbar.Separator />
  <Toolbar.Button render={<Switch.Root/>}/>
    <Switch.Thumb />
  </Toolbar.Button>
</Toolbar.Root>
```

**Dependencies**
- https://github.com/mui/base-ui/issues/11
- https://github.com/mui/base-ui/issues/31
- https://github.com/mui/base-ui/issues/1303
- https://github.com/mui/material-ui/issues/17203

PR #1437, opened 2025-02-11
[Toolbar] POC with context checking API

mj12albert:
**Demo:** https://deploy-preview-1437--base-ui.netlify.app/experiments/toolbar/text-editor

```jsx
<Toolbar.Root>
  <Select.Root>
    <Select.Trigger />
    <Select.Portal>
     {/* the rest of the select */}
    </Select.Portal>
  </Select.Root>
  <Toolbar.Separator />
  <ToggleGroup>
    <Toggle value="bold" />
    <Toggle value="italics" />
    <Toggle value="underline" />
  </ToggleGroup>
  <Menu.Root>
    <Menu.Trigger />
    <Menu.Portal>
     {/* the rest of the menu */}
    </Menu.Portal>
  </Menu.Root>
</Toolbar.Root>
```
Full source: https://github.com/mui/base-ui/blob/2eae6834348d70804ca94c93f97c04b2f4738adf/docs/src/app/(private)/experiments/toolbar/text-editor.tsx

Forked from https://github.com/mui/base-ui/pull/1349


mj12albert:
Issues so far:
1. Seems impossible for TS to enforce the `focusableWhenDisabled` prop on various triggers depending on whether it's placed in a Toolbar or not
    - `Input` and `NumberField` can be toolbar items, `focusableWhenDisabled` will be exposed on these too?
2. Manually merging the `disabled` states is tricky, `mergeReactProps` could do it automatically but this API intentionally wants to bypass it (also requires extra code). In the [demo](https://deploy-preview-1437--base-ui.netlify.app/experiments/toolbar/text-editor), if the MenuTrigger is disabled (use the settings panel) aria-disabled/data-disabled are not applied correctly right now even though the trigger is functionally disabled
3. Menu/Select trigger [intentionally sets tabIndex](https://github.com/mui/base-ui/blob/master/packages/react/src/menu/trigger/useMenuTrigger.ts#L46), if they are a CompositeItem, that tabIndex needs to override the one on the trigger, resulting in the same issue as 2

2/3 could possibly be worked around by not using `<CompositeItem />` and using `useCompositeItem` directly, though it will result in more duplication

Not sure if anything can be done about 1, it doesn't seem ideal to expose `focuseableWhenDisabled` on regular triggers

packages/react/src/toolbar/root/useCompositeButton.ts
```diff
@@ -0,0 +1,48 @@
+'use client';
+import * as React from 'react';
+// import { GenericHTMLProps } from '../../utils/types';
+// import { mergeReactProps } from '../utils/mergeReactProps';
+import { useToolbarRootContext, type ToolbarRootContext } from './ToolbarRootContext';
+import type { ToolbarItemMetadata } from './ToolbarRoot';
+
+function useCompositeButton(parameters: useCompositeButton.Parameters) {
```
mj12albert, line 8: (this is super rough right now)

mj12albert:
Going with the original API in https://github.com/mui/base-ui/issues/661

PR #1440, opened 2025-02-12
[core] Make `mergeReactProps` work with non-native event handlers

### LINKED ISSUE #1430: [core] `mergeReactProps` cannot merge non-native event handlers
Sandbox: https://codesandbox.io/p/sandbox/fervent-chaplygin-779dcw

Click the toggle and it will throw: `Uncaught TypeError: Cannot create property 'preventBaseUIHandler' on boolean 'true'`

```jsx
<Tooltip.Trigger
  render={
    <Toggle
      pressed={isPressed}
      onPressedChange={(nextPressed) => {
        setPressed(nextPressed);
      }}
    >
      <Icon />
    </Toggle>
  }
/>
```
The uncontrolled counterpart of this works: https://codesandbox.io/p/sandbox/epic-hofstadter-62pnf3

The issue is that `mergeReactProps` expects the event as the first argument like native handlers, but our custom handlers usually pass event as the second: https://github.com/mui/base-ui/blob/master/packages/react/src/utils/mergeReactProps.ts#L44

michaldudak:
Fixed the `mergeReactProps` utility to handle event handlers that don't have the event as the first parameter. It now checks if the parameter is an object and contains the `nativeEvent` field before adding the `preventBaseUIHandler` method.

Additionally, I refactored the implementation to be more readable.

Fixes #1430

### DIFF
--- a/packages/react/src/utils/mergeReactProps.ts
+++ b/packages/react/src/utils/mergeReactProps.ts
@@ -32,21 +32,21 @@ function merge<T extends React.ElementType>(
   }
 
   return Object.entries(externalProps).reduce(
-    (mergedProps, [externalPropName, externalPropValue]) => {
-      if (isEventHandler(externalPropName, externalPropValue)) {
-        mergedProps[externalPropName] = mergeEventHandlers(
-          externalPropValue,
-          internalProps[externalPropName],
-        );
-      } else if (externalPropName === 'style') {
-        mergedProps[externalPropName] = mergeStyles(
+    (mergedProps, [propName, externalPropValue]) => {
+      if (isEventHandler(propName, externalPropValue)) {
+        mergedProps[propName] = mergeEventHandlers(externalPropValue, internalProps[propName]);
+      } else if (propName === 'style') {
+        mergedProps[propName] = mergeStyles(
           externalPropValue as React.CSSProperties,
           internalProps.style,
         );
-      } else if (externalPropName === 'className') {
-        mergeClassNames(externalPropValue as string, internalProps.className);
+      } else if (propName === 'className') {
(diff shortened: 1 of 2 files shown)


packages/react/src/utils/mergeReactProps.ts
```diff
@@ -1,87 +1,140 @@
...
+    key[0] === 'o' &&
+    key[1] === 'n' &&
+    thirdCharCode >= 65 /* A */ &&
+    thirdCharCode <= 90 /* Z */ &&
+    typeof value === 'function'
   );
 }
+
+function mergeEventHandlers(theirHandler: Function, ourHandler: Function) {
+  return (event: unknown) => {
```
mj12albert, line 90: @michaldudak I was wondering if this was a valid use-case: if 
you use useRender (or Slot) to create a custom component 
with event handlers that match ours like this, and want to conditionally cancel "our" handler:
```jsx
<MySlider
  onValueChange={(_value, event) => {
    if (/* something */) {
      event.preventBaseUIDefault()
    }
  }}
  render={
    <Slider.Root />
  }
/>
```

But I realized (generally) `<NumberField.Root onValueChange={(_value, event) => { event.preventBaseUIDefault() }} />` isn't possible either

michaldudak: This API was designed to handle native events only. It's meant to prevent our logic from running on DOM elements (which only accept native events). If there's a real-world scenario that requires skipping logic in out custom handlers, we may revisit it. Most (if not all) of our events are called in reaction to something happening, and there's nothing to prevent. Considering the example you posted: `onValueChange` is called as a last step of `setValue`, so it can't prevent any code from being run.

PR #1445, opened 2025-02-14
[menu] Fix prop merging issues

### LINKED ISSUE #1441: [Menu] Trigger's default behavior cannot be prevented
Calling `preventBaseUIHandler` in MenuTrigger's `onMouseDown` doesn't prevent the menu from opening.

Repro: https://codesandbox.io/p/sandbox/winter-cloud-7kyh2j?file=%2Fsrc%2FApp.tsx%3A7%2C16

This happens because events coming from Floating UI are not merged with `mergeReactProps`.

michaldudak:
1. Made it possible to prevent default Base UI logic by calling `preventBaseUIHandler` in event handlers (fixes #1441) ([before](https://codesandbox.io/p/devbox/shy-fire-rpfyxk?file=%2Fsrc%2FApp.tsx%3A13%2C22&workspaceId=ws_QAeLcZEcDnfGaFA47etEUH) / [after](https://codesandbox.io/p/sandbox/friendly-rgb-28p6wf?file=%2Fsrc%2FApp.tsx%3A10%2C62)).
2. Fixed the behavior of disabled non-interactive triggers (i.e., `<Menu.Trigger disabled render={<span />} />`). Clicking on such a trigger used to open the menu anyway. ([before](https://codesandbox.io/p/devbox/gallant-agnesi-f98dt3?file=%2Fsrc%2FApp.tsx%3A7%2C16&workspaceId=ws_QAeLcZEcDnfGaFA47etEUH) / [after](https://codesandbox.io/p/devbox/hardcore-framework-rfxjvv?file=%2Fsrc%2FApp.tsx%3A10%2C22&workspaceId=ws_QAeLcZEcDnfGaFA47etEUH))
3. Removed `mergeReactProps` calls where they were not necessary.
4. Removed the `disabled` attribute from `SubmenuTrigger`. The submenus should be disabled by setting `disabled` on the Root part.
5. Updated the [fully featured menu experiment](https://deploy-preview-1445--base-ui.netlify.app/experiments/menu/menu-fully-featured) to cover more cases.

The main problem in 1. was that event handlers returned from Floating UI's `useInteractions` didn't respect preventBaseUIHandler as they have their own prop merging logic.
Instead of using these prop getters directly, I now call them without arguments and pass the result to our `mergeReactProps` function.

Another issue I found was that we had to nest `mergeReactProps` calls and props getters, such as `getButtonProps(mergeReactProps(external, internal))`. In some cases this was confusing and error-prone (point 2.), so I updated `mergeReactProps` to accept either props directly or a props getter function.

When a function is provided to `mergeReactProps`, it will be called with all the props defined to the right of it, so `mergeReactProps(obj1, obj2, fn, obj3, obj4)`, will be equivalent to `mergeReactProps(obj, obj2, fn(mergeReactProps(obj3, obj4)))`.
I assume that in such cases it is the props getter functionality to merge incoming and internal props and prevent default logic if needed.


mj12albert:
Hmm I think this `<Field.Root /> > prop: validationMode > onBlur > validates Select on blur` test seems to be flaking https://app.circleci.com/pipelines/github/mui/base-ui/7639/workflows/ecfe76f3-3b49-428a-b074-34a9ceed8e65/jobs/69093/parallel-runs/0/steps/0-107

(same in my PR)

mj12albert:
> When a function is provided to `mergeReactProps`, it will be called with all the props defined to the right of it

If I understand correctly, this: `getButtonProps(mergeReactProps(external, internal))`
now becomes this? `mergeReactProps(getButtonProps, externalProps, internalProps)`

michaldudak:
I noticed it on `master` as well. Fixed (or worked around it) in https://github.com/mui/base-ui/pull/1446

michaldudak:
> If I understand correctly, this: getButtonProps(mergeReactProps(external, internal))
now becomes this? mergeReactProps(getButtonProps, externalProps, internalProps)

Yup, that's right.

packages/react/src/use-button/useButton.ts
```diff
@@ -67,22 +66,40 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
...
       return mergeReactProps(otherExternalProps, buttonProps, {
         type,
         onClick(event: React.MouseEvent) {
           if (!disabled) {
-            onClickProp(event);
+            externalOnClick?.(event);
           }
         },
-        onKeyDown(event: React.KeyboardEvent) {
+        onMouseDown(event: React.MouseEvent) {
```
mj12albert, line 84: > Fixed the behavior of disabled non-interactive triggers

I also encountered this when these are still `<button>`s but "disabled" using aria-disabled+data-disabled and not the disabled attr to keep them focusable-when-disabled: https://github.com/mui/base-ui/blob/6a00daf335a771fd5a1119318cf59de35d1bd300/packages/react/src/use-button/useButton.ts#L92-L129

This fix looks more robust 😄 
In mine, when the disabled menu trigger is focused, Up/Down arrows are still opening the menu (Enter/Space doesn't)

packages/react/src/utils/mergeReactProps.ts
```diff
@@ -1,36 +1,74 @@
...
+ * the conflicting ones from others. This doesn't apply to event handlers, `className` and `style` props.
+ * Event handlers are merged such that they are called in sequence (the leftmost one being called first),
+ * and allows the user to prevent the subsequent event handlers from being
  * executed by attaching a `preventBaseUIHandler` method.
  * It also merges the `className` and `style` props, whereby the classes are concatenated
  * and the leftmost styles overwrite the subsequent ones.
+ *
+ * Props can either be provided as objects or as functions that take the previous props as an argument.
+ * The function will receive the merged props up to that point (going from right to left):
+ * so it the case of `(obj1, fn, obj2, obj3)`, `fn` will receive the merged props of `obj2` and `obj3`.
```
atomiks, line 22:
```suggestion
 * so in the case of `(obj1, fn, obj2, obj3)`, `fn` will receive the merged props of `obj2` and `obj3`.
```

packages/react/src/utils/mergeReactProps.ts
```diff
@@ -1,36 +1,74 @@
...
+ * The function is responsible for chaining event handlers if needed (i.e. we don't run the merge logic).
+ *
+ * Event handlers returned by the functions are not automatically prevented when `preventBaseUIHandler` is called.
+ * They must check `event.baseUIHandlerPrevented` themselves and bail out if it's true.
```
atomiks, line 26: Can you elaborate more on this? Will all event handlers specified in our internal prop getters for components work by default without needing a check?

michaldudak: Whenever a function is passed to `mergeReactProps`, it will be called with the merged props and the `preventBaseUIHandler` is not taken into consideration. Most of our prop getter functions actually call `mergeReactProps` internally, with one notable exception: `useButton`. In its case, the logic for prop merging is more complex as it depends on the button's disabled state.

In the case of functions returned from Floating UI's `useInteractions`, they don't know anything about `preventBaseUIHandler`, so I'm calling them without parameters and provide the result to `mergeReactProps` so our merging logic is run on them. This alone should be sufficient to fix 1., but I refactored `mergeReactProps` to avoid having to nest prop getters and help prevent bugs like 2.

atomiks (commented):
The ordering logic makes sense to me

mj12albert:
> 2\. Fixed the behavior of disabled non-interactive triggers

Was just testing this and found that dialog also has this issue:
 https://codesandbox.io/p/sandbox/clever-alex-6472d7

I think triggers that currently don't use useButton (Dialog, AlertDialog, Popover) need to start using it with the fixes in this PR

michaldudak:
Yes, there's possibly more components with these problems. I started with the Menu to keep the PR size managable.

atomiks (approved):

PR #1446, opened 2025-02-14
[test] Skip flaky FieldRoot tests in real browsers

michaldudak:
The FieldRoot tests verifying Select are flaky when run on a fast machine using browser mode. Adding an artificial delay after a trigger is pressed does also seem to help, but it further increases the duration of the test. Since they work fine in JSDOM, I opted for skipping them in browsers.


mj12albert (approved):

atomiks:
There's also https://github.com/mui/base-ui/pull/1371 but it's rarer it seems. Not sure if that actually helps since it doesn't occur too often 🤔

Issue #1425, opened 2025-02-06
[Field] Does not invalidate certain required controls


mj12albert:
# Bug report

## Reproducible example

RadioGroup: https://codesandbox.io/p/sandbox/sweet-brattain-q9ptqr
Select: https://codesandbox.io/p/sandbox/eloquent-pond-cgnlg5

## Current behavior

```jsx
<Field.Root validationMode="onBlur">
  <Select.Root required>
    {/* the rest of the select */}
  </Select.Root>
  <Field.Error match="valueMissing">This field is required</Field.Error>
</Field.Root>
```

When there is no initial value (e.g. no `defaultValue`/`value` or `defaultValue=""`) and the validation mode is `"onBlur"`, Tabbing through the controls without making any selection doesn't not trigger the error

## Expected behavior

Validate different types of empty values correctly

## Base UI version

master & `1.0.0-alpha.5`

## Additional context

Only tried Input, RadioGroup and Select so far

mj12albert:
A workaround is to use a validation function and `match="customError"`:

```jsx
<Field.Root
  validationMode="onBlur"
  validate={(value) => {
    if (value == null) {
      return "This field is required";
    }
    return null;
  }}
>
  <Select.Root required>
    {/* the rest of the select */}
  </Select.Root>
  <Field.Error match="customError" />
</Field.Root>
```

atomiks:
For basic inputs at least, this is intentional. The input needs to be marked dirty first (changed then emptied) to trigger validation on blur, to reduce error noise while tabbing. Maybe there needs to be a setting to configure this?

mj12albert:
> The input needs to be marked dirty first (changed then emptied) to trigger validation on blur, to reduce error noise while tabbing

This seems like a sensible default behavior, maybe it's time to expand the Field docs a bit like https://github.com/mui/base-ui/pull/1356 😬 

> Maybe there needs to be a setting to configure this?

To keep the API surface lean, we could provide an extra `fieldState` argument to the `validate` function to have the noisier validation-on-blur instead of an extra prop, what do you think?
```
<Field.Root validate={(value, { dirty, touched, filled }) => {/* validation logic */}}>
```

PR #1448, opened 2025-02-17
[RadioGroup] Fix `Form`/`Field` validation integration

atomiks:
1. The `name` prop wasn't being forwarded from `Field` to `RadioGroup` `<input>`
2. The `required` prop wasn't being forwarded from `RadioGroup` to `<input>`, and native HTML validation requires it be visually hidden, not true `type=hidden`

Part of #1425, mentioned in https://github.com/mui/base-ui/pull/1053/#discussion_r1942478141


mj12albert (approved):
👍 validation on submit works now: https://codesandbox.io/p/devbox/gallant-brahmagupta-q9ptqr-forked-j7dhrx

michaldudak (approved):

PR #1449, opened 2025-02-17
[tabs] Fix keyboard navigation involving disabled Tabs

### LINKED ISSUE #1442: [Tabs] Arrow keys cannot navigate to disabled Tabs
Repro: https://codesandbox.io/p/sandbox/tender-bassi-w74kv9

`Tabs.Tab` are intentionally focusable when disabled. It's possible to initially Tab into the first disabled Tab, but subsequent arrow left/right keys will not move focus to other disabled Tabs (or back to the first one)

`disabledIndices` needs to be used to properly handle focusable-when-disabled items that are also composite (roving focus) items: https://github.com/mui/base-ui/blob/master/packages/react/src/composite/root/useCompositeRoot.ts#L57-L58

This is used in https://github.com/mui/base-ui/pull/1349

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1442

Before: https://codesandbox.io/p/sandbox/tender-bassi-w74kv9

After: https://codesandbox.io/p/devbox/restless-framework-g3j43t?file=%2Fsrc%2FApp.tsx

Tabs are intentionally focusable when disabled, so it should be possible to move focus to disabled Tabs using the arrow keys. The disabled state should only disable tab activation.


mj12albert:
`activateOnFocus` complicates this, might need to have separate states for "highlighted index" and "selected index" since they could be different

packages/react/src/tabs/list/TabsList.tsx
```diff
@@ -94,6 +94,7 @@ const TabsList = React.forwardRef(function TabsList(
         onHighlightedIndexChange={setHighlightedTabIndex}
         onMapChange={setTabMap}
         render={renderElement()}
+        disabledIndices={[]}
```
michaldudak, line 97: Do not recreate the array on each render.

packages/react/src/tabs/tab/useTabsTab.ts
```diff
@@ -58,11 +58,15 @@ function useTabsTab(parameters: useTabsTab.Parameters): useTabsTab.ReturnValue {
     return valueParam === selectedTabValue;
   }, [index, selectedTabValue, valueParam]);
 
-  // when activateOnFocus is `true`, ensure the active item in Composite's roving
-  // focus group matches the selected Tab
+  const selectedValueSyncedWithHighlightedTabIndexRef = React.useRef(false);
```
michaldudak, line 61: It's a mouthful :)
But too long is much better than too short, so let's leave it.

michaldudak (commented):
:+1: I haven't found any issues in behavior.

packages/react/src/tabs/list/TabsList.tsx
mj12albert: Fixed!

packages/react/src/tabs/tab/useTabsTab.ts
mj12albert: this is all within a few lines so I shortened it by 12 chars 😅
`isSelectionSyncedWithHighlightRef`

michaldudak (approved):

PR #705, opened 2024-10-04
[core] Add explicit type annotations to components

michaldudak:
Components defined as functions (i.e. without `forwardRef`) with added propTypes ends up having an unexpected type in resulting .d.ts (as described in #651). This also causes #622.
Adding an explicit type annotation solves both of these problems.

Closes #622
Closes #651

PR #1456, opened 2025-02-17
[internal] Add explicit type annotations to components

michaldudak:
A follow-up to https://github.com/mui/base-ui/pull/705.
New components had the same issue. I added a test to prevent it from happening again.


atomiks (approved):

michaldudak:
This is no longer necessary after we removed prop-types.

PR #1459, opened 2025-02-18
[popover] `modal` prop

### LINKED ISSUE #623: [dialog][popover][menu][select] Modality and focus changes

atomiks:
- Adds the `modal` prop to `Popover`, like `Menu`
- Makes `Popover` match `Menu`'s "stick" behavior. Prevents multiple popovers from overlapping each other when one is clicked and hovering over if they have `openOnHover`

Scroll locking and internal backdrop aren't active if `openReason === 'hover'` for either one, even if `modal=true`.

Supersedes https://github.com/mui/base-ui/pull/1234 and https://github.com/mui/base-ui/pull/1231

Closes #623


michaldudak:
The new prop is missing from the docs

atomiks:
> The new prop is missing from the docs

Fixed – caused by the issue of the root not being identified

michaldudak (approved):

vladmoroz:
Do you guys still plan to make popovers trap keyboard focus regardless of the `modal` prop? (Don't think I'm seeing that behavior on the preview link).

Asking because we are using non-modal Popovers in Paper, but keyboard focus closing the popover feels unexpected

https://github.com/user-attachments/assets/408ec0a3-3042-4897-96d1-ce4b73fef9c4

atomiks:
@vladmoroz we need some options to control the four "modality" side-effects:

1. Focus trapping
2. Screen reader block with `aria-hidden` (you mentioned a non-modal popover, so this would be disabled)
3. Pointer interaction block
4. **Document** scroll lock (pointer interaction block prevents scroll containers from being scrolled already)

A separate prop like `modalOptions` doesn't really make sense if you set `modal={false}`. I think the `modal` prop may need to be named something different entirely to incorporate these config options

vladmoroz:
I wouldn't need the extra granularity beyond pointer + scroll modality on/off, and personally think that `modal` should just control 2-4 altogether and keyboard focus is always expected to be locked. If you are still set on providing the options, making it `modal: boolean | ModalityOptions` sounds reasonable to me though, as well as just individual props for each component of the modality

atomiks:
I guess this seems nice actually?

```tsx
<Popover.Popup 
  modal={{
    focus: boolean,
    scroll: boolean, // or documentScroll?
    pointer: boolean,
  }}
>
```

The concept of "modal" seems to span three distinct user input modalities that you might want to turn on/off:

- `focus`: "modal" here means means tabbing is trapped within the element. We have not applied this to  `Menu`, `Select`, and `Popover` when `modal=true`, but you want this one. One part is enforcing that either the `Close` component was rendered, or a visually-hidden dismiss button is available (for touch screen readers)
- `scroll`: "modal" here means you can't scroll
- `pointer`: "modal" here means you can't click anything outside the element with your pointer

(Removing the fourth I mentioned, as I think if tabbing is trapped, the screen reader virtual cursor should be too. It's unexpected that you'd escape with tabbing or moving the virtual cursor in both scenarios since they're a similar action)

I suppose `true` would be some combination of options we think makes sense, not making all of them `true` (how `Menu` acts now, while `false` turns them all off, always)

mnajdova:
I like having an options, not sure if they should be part of a modal prop though, but I can live with it. Iw ould propose the following names: 
- `trapFocus`
- `lockScroll`
- `trapPointer`

atomiks:
@mnajdova I think those names would make more sense if they were separated, and we removed the `modal` prop entirely. With the `modal` prop accepting options, we need to name it based on the input that's being made modal:



```js
<Popover.Popup modal={{ focus: true, scroll: true, pointer: true }}>
// vs
<Popover.Popup trapFocus lockScroll trapPointer>
```

The main benefit of the `modal` prop is it can accept `true`/`false` in addition to the options, which is a bit simpler than specifying props separately in simple cases

atomiks:
With the word "modal" being confusing and overloaded, I actually like `trap` instead, with `modal` being removed.

```jsx
<Popover.Popup trap={{ focus: true, scroll: true, pointer: true }}>
<Popover.Popup trap>
<Dialog.Popup trap={false}>
```

`trap=true` would set all to `true`, while `false` the opposite. Each component would specify its own configuration for the options as a default internally.

I think `trap` works for `scroll` too even though "lock" is more common, since the popover's scroll (if it has any) isn't "locked".

michaldudak:
I like the "trap" idea, but the downside may be that it's not what users would search for so it'll be hard to discover.

atomiks:
> I like the "trap" idea, but the downside may be that it's not what users would search for so it'll be hard to discover.

Good point, though we could make a section on modality for the components to improve discoverability

vladmoroz:
"trap" sounds weird tbh, only "trap focus" makes sense

Also I don't see how scroll and pointer modality can be disconnected from one another? That'd be an incoherent user experience

atomiks:
I also thought it sounded a bit weird initially, but it actually makes perfect sense and is the same as trapping focus:

- Trapping their focus means they can't tab to anything except what's in the popup
- Trapping their pointer means they can't click anything except what's in the popup
- Trapping their scroll means they can't scroll anything except what's in the popup

> Also I don't see how scroll and pointer modality can be disconnected from one another? That'd be an incoherent user experience

Some of GitHub's popups trap the pointer, but not scroll. It's what I was trying to achieve with https://github.com/mui/base-ui/pull/1231, with the benefit of removing the pointer trap when the popup has been scrolled out of view (since locking scroll for anchored popups that don't have a dimming backdrop can feel overly restrictive & annoying for certain apps/sites).

Issue #1443, opened 2025-02-13
[number field] Increment and Decrement can be clicked when disabled


mj12albert:
Repro: https://codesandbox.io/p/sandbox/cocky-kepler-y5j9tz

This sets the `disabled` attribute in the DOM but they are still clickable:
```jsx
<NumberField.Decrement disabled>

<NumberField.Increment disabled>
```

GalindoSVQ:
If you want to disable, you need to disable the root element instead of the decrement/increment buttons

mj12albert:
Maybe this is not so likely in userland, but in my case (https://github.com/mui/base-ui/pull/1349) the disabled state can't be passed directly to the Root: https://github.com/mui/base-ui/pull/1349/files#diff-72aed59a3c9228e3356be5fc00fd2ee07c218049d48127e5e75e2dca9252c5bfR207-R248

mj12albert:
BTW disabling the buttons seems to work as expected in controlled mode: https://codesandbox.io/p/devbox/blissful-tess-86zkjw?file=%2Fsrc%2FApp.tsx%3A38%2C55&workspaceId=ws_WwF21DhBng6HdaYyXQWC1w

PR #1462, opened 2025-02-18
[NumberField] Fix disabled state on increment/decrement buttons

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1443

Before: https://codesandbox.io/p/sandbox/cocky-kepler-y5j9tz

After: https://codesandbox.io/p/devbox/billowing-cookies-gnsfn5

The issue was that pointer event handlers didn't consider the disabled prop on the buttons, only other disabled states (e.g. root)

Also extracted hooks/prop getters from the root hook to individual hooks.


packages/react/src/number-field/decrement/NumberFieldDecrement.test.tsx
```diff
@@ -315,4 +316,22 @@ describe('<NumberField.Decrement />', () => {
 
     expect(input).to.have.value('-2');
   });
+
+  it('prop: disabled', async () => {
```
michaldudak, line 320: nit: the `it` function should describe the correct behavior. We usually place "prop: X" on the `describe` blocks.

packages/react/src/number-field/root/useNumberFieldRoot.ts
```diff
@@ -452,7 +452,8 @@ export function useNumberFieldRoot(
         },
         onPointerDown(event) {
           const isMainButton = !event.button || event.button === 0;
-          const isDisabled = disabled || (isIncrement ? isMax : isMin);
+          const isDisabled =
+            (externalProps.disabled ?? false) || disabled || (isIncrement ? isMax : isMin);
```
michaldudak, line 456: I'm not entirely comfortable with this pattern. Not because it produces wrong results, but because we generally never rely on anything from `externalProps`. Is there a good reason for keeping all the buttons' handlers in the root hook instead of moving them closer to components?

mj12albert: > Is there a good reason for keeping all the buttons' handlers in the root hook instead of moving them closer to components?

Yeah I agree, I thought reaching into externalProps was starting to get weird 😅
Probably this was the earliest rewritten component and later we moved to have them closer to components, do you think its ok to move this to its own hook now @atomiks ?

atomiks: Yeah this was written before we moved hooks/logic into their owning components - it's okay to change it to follow the latest patterns

mj12albert:
I've extracted the hooks/prop getters from the root hook except for the scrub related ones to not conflict with https://github.com/mui/base-ui/pull/1485

packages/react/src/number-field/input/NumberFieldInput.tsx
```diff
@@ -20,9 +21,57 @@ const NumberFieldInput = React.forwardRef(function NumberFieldInput(
...
+    min,
+    name,
+    readOnly,
+    required,
+    setValue,
+    valueRef,
+    setInputValue,
+    locale,
+    inputRef: forwardedRef,
+  });
```
atomiks, line 72: I think we can just keep the logic inside the components, as this inflates the bundle size by passing all the params (same with the other components)

atomiks: Makes sense about the buttons as they previously were shared in the root - I think it's ok then

packages/react/src/number-field/scrub-area-cursor/NumberFieldScrubAreaCursor.test.tsx
```diff
@@ -5,20 +5,66 @@ import sinon from 'sinon';
...
+import { NOOP } from '../../utils/noop';
 import { NumberFieldRootContext } from '../root/NumberFieldRootContext';
 
-const defaultTestContext: NumberFieldRootContext = {
-  getScrubAreaCursorProps: (externalProps) => externalProps,
+const defaultTestContext = {
+  allowInputSyncRef: { current: false },
+  autoFocus: false,
+  disabled: false,
+  formatOptionsRef: { current: undefined },
```
atomiks, line 15: Let's change this to use the real component pattern with `NumberField.Root` instead of a mock context

mj12albert: @atomiks `NumberFieldScrubAreaCursor` needs `isScrubbing: true` in the fake context to render something for the conformance test to work, what do you think about using a global flag similar to `BASE_UI_ANIMATIONS_DISABLED` to force render it in tests?

atomiks: I think in cases like that, the fake context is fine - just when it's duplicated across each test when we can use the new root pattern it's better to do so

atomiks (approved):

Issue #1470, opened 2025-02-22
[AlertDialog][Dialog][Popover] Handle custom buttons


mj12albert:
Example: https://codesandbox.io/p/sandbox/clever-alex-6472d7

Dialog, AlertDialog, Popover have triggers that don't use useButton internally, causing their disabled state to not work properly when rendering a non-interactive element, or used in a composite component (https://github.com/mui/base-ui/pull/1349#issuecomment-2672070302)

_Originally posted by @mj12albert in https://github.com/mui/base-ui/pull/1445#issuecomment-2672052422_

PR #1473, opened 2025-02-24
[AlertDialog][Dialog][Popover] Fix non-interactive button disabled state

mj12albert:
Closes https://github.com/mui/base-ui/issues/1470

Demo: https://codesandbox.io/p/devbox/suspicious-bessie-khdjkj?file=%2Fsrc%2FApp.tsx%3A8%2C1

Use useButton internally in (Alert)Dialog.Trigger, (Alert)Dialog.Close, Popover.Trigger to fix their disabled states:
- always sets `data-disabled` when disabled for consistency, even when it's a native `button` element
- when a non-interactive element (e.g. a span) is rendered, functionality is disabled accordingly and `aria-disabled` is used instead of the HTML `disabled` attribute


atomiks (approved):
Looks great

Issue #1471, opened 2025-02-22
[Slider] Maximum depth exceeded bug when `min` and `max` is 1


jcheese1:
# Bug report

## Current behavior

https://codesandbox.io/p/sandbox/slider-bug-min-and-max-the-same-yh935r

maximum depth error when `min` and `max` is 1.

Provide a clear and concise description of the current behavior you're experiencing. If applicable, provide screenshots and/or videos.

## Expected behavior

Does not error out

Provide a clear and concise description of the expected behavior.

## Reproducible example

Link to a CodeSandbox, StackBlitz, or other IDE.

https://codesandbox.io/p/sandbox/slider-bug-min-and-max-the-same-yh935r

## Base UI version

For example v1.0.2

## Which browser are you using?

1.0.0 alpha 6

Chrome/Safari/Firefox/Safari iOS etc.

chrome

## Which OS are you using?

Mac OS/Windows/Other etc.

mac

## Which assistive tech are you using (if applicable)?

Voiceover/JAWS etc.

## Additional context

Provide any additional context that might help us identify the problem and find a solution.

mj12albert:
Thanks for reporting this, it should obviously not crash 😅 

Just to be sure, you don't have a real use-case around this do you? @jcheese1

mj12albert:
https://codesandbox.io/p/sandbox/affectionate-mestorf-2ldw9y?file=%2Fsrc%2FDemo.tsx

Material UI just silently fails when the range is invalid (min = max, min > max)

PR #1475, opened 2025-02-24
[Slider] Warn when `min` is not less than `max`

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1471

Log a console warning when `min` is not less than `max` and don't update the state. When min/max are the same something divides by zero internally and causes a loop.

### DIFF
--- a/packages/react/src/slider/root/useSliderRoot.ts
+++ b/packages/react/src/slider/root/useSliderRoot.ts
@@ -10,6 +10,7 @@ import { useControlled } from '../../utils/useControlled';
 import { useEnhancedEffect } from '../../utils/useEnhancedEffect';
 import { useForkRef } from '../../utils/useForkRef';
 import { valueToPercent } from '../../utils/valueToPercent';
+import { warn } from '../../utils/warn';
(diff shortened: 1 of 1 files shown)


mj12albert:
https://codesandbox.io/p/sandbox/slider-bug-min-and-max-the-same-forked-kz8xgy

It doesn't crash and logs a warning, I think this should be enough

michaldudak:
I can imagine a case where the min/max values are set based on some state set in other parts of the application (i.e., not hardcoded), and with certain conditions, the Slider could allow just a single value. IMO we should support such a scenario.

mj12albert:
> with certain conditions, the Slider could allow just a single value

All we could do differently though is not warn if min/max/value are the same? If min/max/value are all the same, it's still not possible to determine a position on the track @michaldudak

michaldudak:
Yeah, ignore what I wrote. For some reason I thought min/max restricts the slider position within the track (that the track renders the full range, but you're not allowed to slide the thumb across the whole track).

michaldudak (approved):

PR #1482, opened 2025-02-26
[code-infra] Fix build:types not copying on some setups

Janpot:
* Fix `copyDeclarations` as per https://github.com/mui/material-ui/pull/45422
* Also noticed a set of promises that weren't at all awaited for


mj12albert (approved):
Tested this and I can output both:
```
/Users/mj12albert/dev/base-ui.git/packages/react/build/cjs/index.d.ts 

/Users/mj12albert/dev/base-ui.git/packages/react/build/esm/index.d.ts 
```

michaldudak (approved):

PR #1488, opened 2025-02-28
[NumberField] Add `locale` prop

### LINKED ISSUE #1474: [number field] Allow configuring locale with a prop
# Get help

## Ask a question

I can not seem to figure out, how to use the format options in NumberInput, especially changing the locale of the number formatting.

atomiks:
Closes #1474

### DIFF
--- a/packages/react/src/number-field/utils/parse.ts
+++ b/packages/react/src/number-field/utils/parse.ts
@@ -37,10 +37,12 @@ export function parseNumber(
   options?: Intl.NumberFormatOptions,
 ) {
   let computedLocale = locale;
-  if (ARABIC_RE.test(formattedNumber)) {
-    computedLocale = 'ar';
-  } else if (HAN_RE.test(formattedNumber)) {
-    computedLocale = 'zh';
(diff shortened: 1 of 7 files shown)


packages/react/src/number-field/utils/parse.ts
```diff
@@ -31,15 +31,19 @@ export function getNumberLocaleDetails(
...
-  let locale: string | undefined;
+export function parseNumber(
+  formattedNumber: string,
+  locale?: Intl.LocalesArgument,
+  options?: Intl.NumberFormatOptions,
+) {
+  let computedLocale = locale;
   if (ARABIC_RE.test(formattedNumber)) {
-    locale = 'ar';
+    computedLocale = 'ar';
```
mnajdova, line 41: Do we really want to override this now that it is specified by the user?

atomiks: Yeah you're right, it should check for `undefined`

mnajdova (approved):

PR #1491, opened 2025-02-28
[NumberField] Prevent virtual cursor overlapping native one

atomiks:
The virtual cursor currently can overlap  the native one since the state can be set before the pointer lock promise (previously had wrong types in TS itself - it's now correct) finishes. Removed the flicker timeout as a result


mnajdova:
Can you share a screen-recording of this? I can't reproduce it, it was make it clear what's the PR about.

atomiks:
@mnajdova it's hard to see in the screen recording (I tried) because the flicker of the two overlapping cursors is so quick. Essentially, you briefly saw two simultaneous cursors (native and our fake HTML one - can repro old on live `base-ui.com`), now you should not.

As long as you see it looks good now, that's fine

mj12albert (approved):
The code looks fine, but I tested this vs master and couldn't see any difference either

PR #1529, opened 2025-03-04
[tabs][RadioGroup] Handle modifier keys

### LINKED ISSUE #1465: [tabs] Arrow keys should not move Tab focus if any modifier keys are pressed
Discovered via https://github.com/mui/material-ui/issues/45308

It's reproducible in the hero demo: https://base-ui.com/react/components/tabs

We should check for Shift, Alt, Ctrl, and meta keys

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1465

All composite components will now prevent arrow key navigation if any modifier key (Shift, Ctrl, Alt, Meta) is pressed by default unless it's allowed with the `modifierKeys` prop.

The exception is RadioGroup since HTML radio groups work normally when `Shift` is held ([example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#try_it))

Demo:
- Tabs will no longer navigate if any modifier key is pressed: https://deploy-preview-1529--base-ui.netlify.app/react/components/tabs
- RadioGroup works normally if `Shift` is pressed: https://deploy-preview-1529--base-ui.netlify.app/react/components/radio

### DIFF
--- a/packages/react/src/composite/root/useCompositeRoot.ts
+++ b/packages/react/src/composite/root/useCompositeRoot.ts
@@ -25,9 +25,11 @@ import {
   HORIZONTAL_KEYS_WITH_EXTRA_KEYS,
   isDisabled,
   isIndexOutOfBounds,
+  MODIFIER_KEYS,
   VERTICAL_KEYS,
   VERTICAL_KEYS_WITH_EXTRA_KEYS,
   type Dimensions,
+  type ModifierKey,
 } from '../composite';
 
 export interface UseCompositeRootParameters {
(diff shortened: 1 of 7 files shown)


packages/react/src/composite/root/useCompositeRoot.ts
```diff
@@ -76,6 +94,7 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
     enableHomeAndEndKeys = false,
     stopEventPropagation = false,
     disabledIndices,
+    modifierKeys = [],
```
atomiks, line 97: Should this default be specified outside so it's a stable reference?

mj12albert: 👍 ~ fixed

atomiks (approved):

PR #1533, opened 2025-03-05
[utils] Change order of args in `mergeReactProps`

mnajdova:
This PR changes the order in which args are processed in mergeReactProps, so we can prepare it as a public API.

Review with hidden whitespaces: https://github.com/mui/base-ui/pull/1533/files?diff=split&w=1


atomiks:
The ordering makes more sense here, but formatting is more of an issue than I thought

It's hard to notice `externalProps` the very end, and there's a lot of syntax and nesting which makes it tricky to understand.

Here's one example where it seems like it's just "dangling" randomly:
https://github.com/mui/base-ui/blob/a187ec52e918deacfc515dbc8949f38a4ea7ee71/packages/react/src/accordion/root/useAccordionRoot.ts#L158-L165

mnajdova:
> The ordering makes more sense here, but formatting is more of an issue than I thought
> 
> It's hard to notice externalProps the very end, and there's a lot of syntax and nesting which makes it tricky to understand.

This is more of a code style, we can extract the first arg as `defaultProps` in a constant and it would be much more readable. People are used to props spreading, having a reverse order is just unintuitive. I haven't seen any merge util in any library that has the reverse order. I feel strongly about this for the external app, if you, @michaldudak and @mj12albert really prefer the previous version, we can change the order only for the public API, but I would be honestly against this.

atomiks:
> This is more of a code style, we can extract the first arg as defaultProps in a constant and it would be much more readable

If so, I think we should encourage doing that instead throughout the codebase for readability (assigning a variable with `React.ComponentPropsWithoutRef<tag>` beforehand)

atomiks (approved):

PR #1537, opened 2025-03-05
[docs-infra] New API reference generator

michaldudak:


docs/reference/generated/accordion-panel.json
```diff
@@ -41,12 +40,10 @@
   },
   "cssVariables": {
     "--accordion-panel-height": {
-      "description": "The accordion panel's height.",
-      "type": "number"
+      "description": "The accordion panel's height."
```
mnajdova, line 43: Can we keep the previous logic for the @type info?

docs/reference/generated/accordion-root.json
```diff
@@ -20,27 +21,31 @@
...
+      "type": "boolean | undefined",
       "default": "true",
+      "required": true,
       "description": "Whether multiple items can be open at the same time."
     },
     "disabled": {
-      "type": "boolean",
+      "type": "boolean | undefined",
       "default": "false",
+      "required": true,
```
mnajdova, line 32: This is a bit strange. It says it's required, but we allow undefined as value. Is there a way to change the props that have default values not to be flagged as required?

docs/reference/generated/alert-dialog-description.json
```diff
@@ -3,11 +3,11 @@
   "description": "A paragraph with additional information about the alert dialog.\nRenders a `<p>` element.",
   "props": {
     "className": {
-      "type": "string | (state) => string",
+      "type": "string | ((state: unknown) => string)",
       "description": "CSS class applied to the element, or a function that\nreturns a class based on the component’s state."
     },
     "render": {
-      "type": "React.ReactElement | (props, state) => React.ReactElement",
+      "type": "React.ReactElement | ((props: unknown, state: unknown) => React.ReactElement)",
```
mnajdova, line 10: We probably need to specify the state & props for each component.

mnajdova (commented):
It would be nice to set up a test case for hook types (maybe after useRender is merged). Also, could be try running this in other repos?

Would it make sense to make the extractor part this GitHub org, to make it easier to make changes for other people?

docs/reference/generated/accordion-panel.json
michaldudak: Yes, the extractor doesn't handle the @type tag yet, but will soon.

docs/reference/generated/alert-dialog-description.json
michaldudak: Yeah, this doesn't look right. These aren't really the props of a component but attributes to be spread on the rendered HTML element. I'll try to display in a better way

docs/reference/generated/composite-list.json
michaldudak, line 1: I'm going to exclude internal components.

docs/reference/generated/accordion-header.json
```diff
@@ -3,11 +3,11 @@
   "description": "A heading that labels the corresponding panel.\nRenders an `<h3>` element.",
   "props": {
     "className": {
-      "type": "string | (state) => string",
+      "type": "string | ((state: State) => string)",
       "description": "CSS class applied to the element, or a function that\nreturns a class based on the component’s state."
     },
     "render": {
-      "type": "React.ReactElement | (props, state) => React.ReactElement",
+      "type": "ReactElement | ((props: GenericHTMLProps, state: State) => ReactElement)",
```
mnajdova, line 10: Was it intentional that 'React.' is removed as a prefix?

docs/reference/generated/accordion-item.json
```diff
@@ -2,8 +2,11 @@
   "name": "AccordionItem",
   "description": "Groups an accordion header with the corresponding panel.\nRenders a `<div>` element.",
   "props": {
+    "value": {
```
mnajdova, line 5: Nice, this was missed before.

docs/reference/generated/accordion-root.json
```diff
@@ -35,12 +35,12 @@
       "description": "Whether to loop keyboard focus back to the first item\nwhen the end of the list is reached while using the arrow keys."
     },
     "orientation": {
-      "type": "'horizontal' | 'vertical'",
+      "type": "\"horizontal\" | \"vertical\"",
```
mnajdova, line 38: Nitpick, could we use consistently single or double quotes for the type and the default value? The default value uses single quite, while the type is now double quote, it looks a bit off in the docs: https://deploy-preview-1537--base-ui.netlify.app/react/components/accordion#root

docs/reference/generated/alert-dialog-description.json
```diff
@@ -3,11 +3,11 @@
   "description": "A paragraph with additional information about the alert dialog.\nRenders a `<p>` element.",
   "props": {
     "className": {
-      "type": "string | (state) => string",
+      "type": "string | ((state: {}) => string)",
```
mnajdova, line 6:
```suggestion
      "type": "string | ((state: State) => string)",
```
It's strange that it is not consistent with the rest.

docs/reference/generated/alert-dialog-root.json
```diff
@@ -12,16 +12,19 @@
       "description": "Whether the dialog is currently open."
     },
     "onOpenChange": {
-      "type": "(open, event, reason) => void",
+      "type": "((open: boolean, event: Event | undefined, reason: \"click\" | \"hover\" | \"focus\" | \"focus-out\" | \"escape-key\" | \"outside-press\" | \"trigger-press\" | undefined) => void)",
```
mnajdova, line 15: We'll need to update the UI on the docs table, this is cut off, you need to scroll to read it,  I don't think it's the best experience (can be done in a follow up)

Screenshot 2025-03-17 at 10 36 52

docs/reference/generated/alert-dialog-root.json
```diff
@@ -12,16 +12,19 @@
...
       "description": "Event handler called when the dialog is opened or closed."
     },
     "actionsRef": {
-      "type": "{ current: { unmount: func } }",
+      "type": "RefObject",
```
mnajdova, line 19: Was this overriden before? Maybe we can improve the type here?

docs/reference/generated/field-validity.json
```diff
@@ -3,7 +3,7 @@
   "description": "Used to display a custom message based on the field’s validity.\nRequires `children` to be a function that accepts field validity state as an argument.",
   "props": {
     "children": {
-      "type": "(validity) => React.ReactNode",
+      "type": "((state: FieldValidityState) => ReactNode)",
```
mnajdova, line 6: In a follow up we may want to improve and show the values in this enums, I saw similar one around the `ImageLoadingStatus`

docs/reference/generated/form.json
```diff
@@ -3,19 +3,19 @@
   "description": "A native form element with consolidated error handling.\nRenders a `<form>` element.",
   "props": {
     "errors": {
-      "type": "Record<string, string | string[]>",
+      "type": "{}",
```
mnajdova, line 6: Where does this type difference come from?

docs/reference/generated/menu-radio-group.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "MenuRadioGroup",
-  "description": "Groups related radio items.\nRenders a `<div>` element.",
```
mnajdova, line 3: The description is missing here.

mnajdova: I don't know if it is related, but the JSdocs on this component misses the "Documentation" information.

docs/reference/generated/number-field-root.json
```diff
@@ -11,15 +11,15 @@
       "description": "The uncontrolled value of the field when it’s initially rendered.\n\nTo render a controlled number field, use the `value` prop instead."
     },
     "value": {
-      "type": "number",
+      "type": "null | number",
```
mnajdova, line 14:
```suggestion
      "type": "number | null",
```

It's strate null to be the first option. Do you know why it was generated this way? The prop is defined as `value?: number | null;`

docs/reference/generated/slider-root.json
```diff
@@ -2,30 +2,29 @@
   "name": "SliderRoot",
   "description": "Groups all parts of the slider.\nRenders a `<div>` element.",
   "props": {
-    "aria-labelledby": {
```
mnajdova, line 5: This prop is missing.

docs/reference/generated/slider-thumb.json
```diff
@@ -6,29 +6,106 @@
...
-      "type": "function(index: number) => string",
+      "type": "null | ((index: number) => string)",
       "description": "Accepts a function which returns a string value that provides a user-friendly name for the input associated with the thumb"
     },
     "getAriaValueText": {
-      "type": "function(formattedValue: string, value: number, index: number) => string",
+      "type": "null | ((formattedValue: string, value: number, index: number) => string)",
       "description": "Accepts a function which returns a string value that provides a user-friendly name for the current value of the slider.\nThis is important for screen reader users."
     },
+    "handleInputChange": {
```
mnajdova, line 31: Hmm, why were these missing before?

mnajdova (commented):
Did you check that we already fix all needed overrides JSON files?

docs/reference/generated/accordion-header.json
michaldudak: Yes, I decided to display just the type name, as the fully qualified name sometimes looks incorrect - it contains the path to the file it's declared in. I played it safe by using just the type name.

docs/reference/generated/alert-dialog-description.json
michaldudak: In this case `State` is an empty interface, that's why it's processed differently. I'll make it consistent, though.

docs/reference/generated/alert-dialog-root.json
michaldudak: Yeah, I'm not a fan of the current layout.

docs/reference/generated/alert-dialog-root.json
michaldudak: Refs are pretty hard to handle because of a certain TS weirdness. I'll try to make it more readable somehow, though.

docs/reference/generated/menu-radio-group.json
michaldudak: MenuRadioGroup is not what is directly exported - it's `const MemoizedMenuRadioGroup = React.memo(MenuRadioGroup);` which doesn't have any description.

docs/reference/generated/slider-root.json
michaldudak: The API extractor documents only properties that are defined in the project it analyses (otherwise we'd have a ton of HTML attributes in all components).

docs/reference/generated/slider-thumb.json
michaldudak: Good question. Could be a bug in the previous implementation?

michaldudak:
This seems to be good enough for review now.

To do in follow-up PRs:
- update the layout of the docs to display longer type descriptions
- expand types such as named unions, aliased functions and objects so their full definitions appear (in a popup, perhaps?)
- remove proptypes
- update the useRender docs to use the generated API reference

docs/reference/generated/form.json
michaldudak: It was a bug in the API extractor

docs/reference/generated/alert-dialog-root.json
flaviendelangle: :laughing: Welcome to the never ending discussion about API page layout when the prop's types are getting more complex

flaviendelangle: I wonder if the type are actually that valuable on the default UI.
Most of the widely used parameters can have their types deduced very easily from their name.

One could say that it's harder to read with the type because we have more characters.
Maybe we could try to stick with a minimalist typing on the default layout, and have a more detailed UI (tooltip ? accordion ? something else ?) that displays the full version.

Just thinking out loud :thinking:

colmtuite: Is it possible to drop onto a new line at the `|`, the same way it's done in the `render` prop type on the Trigger component?

michaldudak: There is a new line placed before each union member:

image

Doing this in function parameters would make it less readable IMO.

packages/react/src/use-render/useRender.ts
```diff
@@ -73,4 +75,8 @@ export namespace useRender {
      */
     props?: Record<string, unknown> & { ref?: React.Ref<RenderedElementType> };
   }
+
+  export interface ReturnValue {
```
mnajdova, line 79: 👌

docs/reference/generated/use-render.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "useRender",
```
mnajdova, line 2: Let's add a TODO to update the documentation page to use this.

packages/react/src/use-render/useRender.ts
```diff
@@ -7,11 +7,13 @@ const emptyObject = {};
 
 /**
  * Returns an object with a `renderElement` function that renders a Base UI element.
+ *
+ * @public
```
mnajdova, line 11: I am a bit confused, why do we need both @public and @private tags on hooks. Was the @internal on hooks a left over? For example `packages/react/src/utils/useAnimationsFinished.ts` has the @internal tag.

michaldudak: Yeah, this was unnecessary. Hooks are considered internal by default. I removed the redundant tags.

PR #1548, opened 2025-03-10
[popover][Menu] Ensure `stickIfOpen` is reset to `true` correctly

atomiks:
When impatiently clicking the trigger of Menu or Popover, it would unexpectedly close the popup if either had already been opened and the `PATIENT_CLICK_THRESHOLD` time had elapsed the first time before deciding to click when either is opened on hover.

Steps:

1. Open on hover
2. Wait >500ms
3. Dismiss
4. Re-open on hover
5. Click within 500ms
6. Now sticks (previously would close)


mj12albert (approved):
Tested popover and it stays open correctly now: https://codesandbox.io/p/sandbox/frosty-einstein-y5dhtw

But following the steps found a separate (currently reproducible on master) issue with Menu , it sticks open indefinitely and can't be dismissed with clicks:

https://github.com/user-attachments/assets/43473d31-5fa0-4fdf-88ac-e80a2655bd56

mj12albert:
@atomiks using `data-floating-ui-inert` fixed the issue with menu 👌 

https://github.com/user-attachments/assets/180f4f78-8c75-4507-b31a-dedfe08ed2c2

atomiks:
There's a small bug remaining for Popover (Menu doesn't have it), but I don't think it is critical as it's very rare to see. I can open a separate issue to address that. It's related to how the mouse move events get blocked in Floating UI so it could also be fixed as a transitive release by updating it as I'll see if I can debug it in there.

Patient click w/ `openOnHover`:
- Hover trigger and click after threshold
- Click again to open
- Click again to dismiss
- Move mouse slowly
- Opens but shouldn't

Impatient click w/ `openOnHover`:
- Hover trigger and click before threshold
- Click to dismiss
- Click again to open
- Click again to dismiss
- Move mouse slowly
- Opens but shouldn't

Issue #1458, opened 2025-02-17
[accordion][collapsible] With CSS transitions the panel does not resize when content size changes


M1dnight-ofcl:
# Bug report

## Current behavior
I have a resizable sidebar and a group of collapsible. The problem is, that whenever I resize the collapsible or delete an element from the collapsible, it doesn't resize, and it leads to annoying bugs
Idle:
![Image](https://github.com/user-attachments/assets/db340f55-a300-4ed8-929b-be905d624b50)
Deleted Element:
![Image](https://github.com/user-attachments/assets/3c7447a8-8765-44a6-8698-8d92f1eeb33f)
Fully Expanded Sidebar:
![Image](https://github.com/user-attachments/assets/7dfba589-f7b3-4b36-9f62-df099888f227)

Now when I close the collapsible and reopen it, it does account for the new size, but I wish that it would observe the size change (maybe using [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver)) and account for it

## Expected behavior
I wanted the collapsible to resize according to the new change

## Reproducible example
Just create a collapsible, and try using states to remove one of the elements (like using a map) or shrink it

## Base UI version
v1.0.0-alpha.6

## Which browser are you using?

Microsoft Webview2 (on wails)

## Which OS are you using?

Windows 11

## Which assistive tech are you using (if applicable)?

Voiceover/JAWS etc.

## Additional context

Provide any additional context that might help us identify the problem and find a solution.

mj12albert:
@M1dnight-ofcl Seems to be an issue with using CSS transitions since it requires explicitly setting height in CSS, that's what you're using right?

Using CSS keyframe animations avoids this: https://codesandbox.io/p/sandbox/bold-sun-7h6xm6?file=%2Fsrc%2Findex.module.css%3A20%2C1

mj12albert:
The reworked CSS transition handling in https://github.com/mui/base-ui/pull/1549 is quite promising so far and could possibly solve this by setting `height: 'auto'` inline at the end of opening transitions

dartess:
Hey guys

I've encountered the same problem with `beta.0`. I initially discovered it when I was using nested `Collapsible`: https://codesandbox.io/p/sandbox/hj432g

[...]

Issue #1334, opened 2025-01-14
[collapsible] Opacity starting style not working


benface:
# Bug report

## Current behavior

The following Tailwind classes in `Collapsible.Panel` lead to a fade out transition when the panel closes, but no fade in transition when it opens:

`transition data-[starting-style]:opacity-0 data-[ending-style]:opacity-0`

## Expected behavior

There should be a fade in transition when the panel opens.

## Reproducible example

https://codesandbox.io/p/sandbox/winter-cdn-vxmxzn

## Base UI version

1.0.0-alpha.5

## Which browser are you using?

Reproduced in Chrome, Firefox, and Safari

## Which OS are you using?

macOS

## Which assistive tech are you using (if applicable)?

None

## Additional context

Thank you! 🙏

mj12albert:
Thanks for reporting @benface it is indeed a bug 😪 , working on it ~

PR #1549, opened 2025-03-10
[Collapsible] Rework CSS transitions

mj12albert:
Demo: https://deploy-preview-1549--base-ui.netlify.app/experiments/collapsible/transition

Exploring a solution for both https://github.com/mui/base-ui/issues/1334 and https://github.com/mui/base-ui/issues/1458

Findings:

- Tailwind v4 defaults started applying [`display: none !important` on `[hidden]`](https://github.com/tailwindlabs/tailwindcss/blame/main/packages/tailwindcss/preflight.css#L382) which we must override in order to measure the node
- ~~New gotcha: when transitioning multiple properties e.g. `opacity` and `height` together, `transition-property` must list individual properties (`transition-property: height, opacity`) instead of `all`~~ solved and not a limitation anymore ✌️ 
- Used an `AbortController` to cancel pending unmounts if the open state changes back to open before the pending unmount actually occurs (which is after the duration of the exit transition). I wonder if CSS animations can be made interruptible like transitions with AbortController.
- Using `element.style.removeProperty('height')` leads to huge improvements in the timing/consistency vs `element.style.height = ''` (or even `'0px'`)
- There is a weird edge case when `keepMounted={false}` and the panel is initially closed, when it opens for the first time and only the first time `data-starting-style` is missing or off by a frame and it needs to be manually set: `element.setAttribute('data-starting-style', '')`


docs/src/app/(private)/experiments/collapsible/plain.tsx
```diff
@@ -0,0 +1,220 @@
...
+    panel.style.setProperty('display', 'block', 'important');
+
+    if (nextOpen) {
+      if (abortControllerRef.current != null) {
+        abortControllerRef.current.abort();
+        abortControllerRef.current = null;
+      }
+
+      /* opening */
+      panel.style.opacity = '0';
```
mj12albert, line 89: Need to read/parse all transition-able properties somehow and detect the "closed" value from computedStyles ?
This seems only possible if when `keepMounted={true}` and when it's initially closed

docs/src/styles.css
```diff
@@ -222,6 +222,10 @@
   [data-demo]:not([data-demo='tailwind']) * {
     all: revert-layer;
   }
+
```
mj12albert, line 225: Tailwind v4 preflight contains:
```css
[hidden]:where(:not([hidden="until-found"])) {
  display: none !important;
}
```
which interferes when we need to set `display: block` while the panel is still `hidden` to measure it's rendered size before the browser actually renders the change

Setting `!important` in our inline styles should be enough as long as they are removed accordingly

mj12albert:
Continues in https://github.com/mui/base-ui/pull/1601

PR #1551, opened 2025-03-12
[useRender] Refine docs and APIs

atomiks:
This is a continuation of #1418.

Docs: https://deploy-preview-1551--base-ui.netlify.app/react/utils/use-render

APIs refined:

1. `useRender.ElementProps` becomes `useRender.ComponentProps`. Since it's used for _components_ and has a special `render` prop, it doesn't make sense to use the word `Element`
2. `mergeProps.Props` -> `useRender.ElementProps`. These are raw element props, and it's confusing to have a namespace on `mergeProps` and two separate things to remember. We might as well reuse `useRender` for this purpose.
3. ~`mergeProps` defaults to `'div'` tag to get type inference out of the box for the most common case (I'm not sold on this being a good default, but it's easy to forget adding a tag and losing inference and types)~
4. Small tweaks to utility naming
5. Docs improvements

### DIFF
--- a/docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
@@ -4,11 +4,11 @@ import { useRender } from '@base-ui-components/react/use-render';
 import { mergeProps } from '@base-ui-components/react/merge-props';
 import styles from './index.module.css';
 
-type CounterState = {
+interface CounterState {
   odd: boolean;
-};
+}
 
-interface CounterProps extends useRender.ElementProps<'button', CounterState> {
+interface CounterProps extends useRender.ComponentProps<'button', CounterState> {
   className?: string;
   onClick?: (event: React.MouseEvent) => void;
   ['aria-label']?: string;
@@ -16,11 +16,11 @@ interface CounterProps extends useRender.ElementProps<'button', CounterState> {
 
 function Counter(props: CounterProps) {
   const { render = <button />, ...otherProps } = props;
+
   const [count, setCount] = React.useState(0);
   const odd = count % 2 === 1;
   const state = React.useMemo(() => ({ odd }), [odd]);
 
-  // The Props type is an alias of React.ComponentPropsWithoutRef
   const defaultProps: mergeProps.Props<'button'> = {
     className: styles.Button,
     type: 'button',
@@ -29,7 +29,9 @@ function Counter(props: CounterProps) {
         Counter: <span>{count}</span>
       </React.Fragment>
     ),
-    onClick: () => setCount((prev) => prev + 1),
+    onClick() {
+      setCount((prev) => prev + 1);
+    },
     'aria-label': `Count is ${count}, click to increase.`,
   };
 
--- a/docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
@@ -21,7 +21,7 @@ function Counter(props: CounterProps) {
   const odd = count % 2 === 1;
   const state = React.useMemo(() => ({ odd }), [odd]);
 
-  const defaultProps: mergeProps.Props<'button'> = {
+  const defaultProps: useRender.ElementProps<'button'> = {
     className: styles.Button,
     type: 'button',
     children: (
@@ -38,7 +38,7 @@ function Counter(props: CounterProps) {
   const { renderElement } = useRender({
     render,
     state,
-    props: mergeProps(defaultProps, otherProps),
+    props: mergeProps<'button'>(defaultProps, otherProps),
   });
 
   return renderElement();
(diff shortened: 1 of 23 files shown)


packages/react/src/dialog/popup/useDialogPopup.tsx
```diff
@@ -51,13 +46,6 @@ export function useDialogPopup(parameters: useDialogPopup.Parameters): useDialog
     return initialFocus;
   }, [defaultInitialFocus, initialFocus, openMethod]);
 
-  useEnhancedEffect(() => {
-    setPopupElementId(id);
-    return () => {
-      setPopupElementId(undefined);
-    };
-  }, [id, setPopupElementId]);
-
```
atomiks, line 60: Reverting this to match other popups, as I added `useRole`.

Issue to make them all handle `id` correctly: #1552

docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
```diff
@@ -4,39 +4,37 @@ import { useRender } from '@base-ui-components/react/use-render';
...
 
-type CounterState = {
+interface CounterState {
   odd: boolean;
-};
-
-interface CounterProps extends useRender.ElementProps<'button', CounterState> {
-  className?: string;
-  onClick?: (event: React.MouseEvent) => void;
-  ['aria-label']?: string;
```
mnajdova, line 14: Nice catch, I forgot to remove these once they were not handled any more 👍

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -58,13 +74,38 @@ The `mergeProps` function merges two or more sets of React props together. It sa
...
+```jsx title="Using mergeProps in the render callback"
+import { mergeProps } from '@base-ui-components/react/merge-props';
+import styles from './index.module.css';
+
+function Button() {
+  return (
+    <Component
+      render={(props, state) => (
+        <button
+          {...mergeProps(props, {
```
mnajdova, line 86:
```suggestion
          {...mergeProps<'button'>(props, {
```

Or change the tag to be `<div />`

packages/react/src/merge-props/mergeProps.ts
```diff
@@ -29,11 +29,11 @@ type MergableProps<T extends React.ElementType> =
  * @param props props to merge.
  * @returns the merged props.
  */
-function mergeProps<T extends React.ElementType>(
-  ...props: MergableProps<T>[]
-): WithBaseUIEvent<React.ComponentPropsWithRef<T>> {
+export function mergeProps<Tag extends React.ElementType = 'div'>(
```
mnajdova, line 32: I would make the generic required without default, it will force people to type it correctly.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -78,9 +119,11 @@ function Text({ render = <p />, ...props }: TextProps) {
 }
 ```
 
-In older versions of React, you need to use `React.forwardRef()` and add the forwarded ref to the `refs` array along with your own internal ref:
+In older versions of React, you need to use `React.forwardRef()` and add the forwarded ref to the `refs` array along with your own internal ref.
 
-```tsx title="React 18 and 17"
+The [Examples](#examples) above assume React 19, and should be modified to use `React.forwardRef()` to support React 18 and 17.
```
mnajdova, line 124:
```suggestion
The [example](#examples) above assume React 19, and should be modified to use `React.forwardRef()` to support React 18 and 17.
```

This reads a bit strange.

mnajdova (commented):
Changes look good in general, the only thing I am not sure about is the generic default value to be 'div'

atomiks:
> Changes look good in general, the only thing I am not sure about is the generic default value to be 'div'

Yeah, going to just remove this for now as it can be problematic. We can re-evaluate it later if necessary

mnajdova (approved):
I am not aware of why the popup changes are part of this PR, but the rest looks fine. If this PR does not break those changes, I would add them in a dedicated PR.

atomiks:
> I am not aware of why the popup changes are part of this PR, but the rest looks fine. If this PR does not break those changes, I would add them in a dedicated PR.

With the `'div'` default for `mergeProps` and  required override, I found the bug. It enforces consistency across them and I opened the issue related to it - might as well retain the changes here as they shouldn't cause actual runtime changes

packages/react/src/use-render/useRender.ts
```diff
@@ -25,17 +24,19 @@ export function useRender<
...
+    styleHooks: false,
   });
 }
 
 export namespace useRender {
   export type RenderProp<State = Record<string, unknown>> =
     | ComponentRenderFn<React.HTMLAttributes<any>, State>
-    | React.ReactElement<Record<string, unknown>>
-    | keyof typeof defaultRenderFunctions;
+    | React.ReactElement<Record<string, unknown>>;
```
atomiks, line 34: Removed `keyof typeof defaultRenderFunctions` since we shouldn't allow strings

mnajdova: Good call!

PR #1555, opened 2025-03-13
[POC] New renderer usage

atomiks:
Simplifies internal `useComponentRenderer` usage to more similarly match the public `useRender` hook

- Manual usage of `useForkRef` is abandoned where possible, the renderer `refs` option is instead used
- No more prop getters, use `mergeProps` at the call site with `extraProps` becoming `props`
- No more `defaultRenderFunctions` strings, instead just use the JSX element like `<p />` except for moderately complex ones like `button` which has `type` attached causing eslint to warn


packages/react/src/tooltip/root/useTooltipRoot.ts
```diff
@@ -158,7 +158,7 @@ export function useTooltipRoot(params: useTooltipRoot.Parameters): useTooltipRoo
       positionerElement,
       setPositionerElement,
       popupRef,
-      getRootTriggerProps,
+      triggerProps: getRootTriggerProps(),
```
michaldudak, line 161: This might not work as intended in all cases. Components that use `useButton` need to have `externalProps` passed into the prop getter, as merging props isn't trivial there (event handlers are not chained, but `onClick` can be called from other handlers).

michaldudak (commented):
Passing an array of refs instead of calling useForkRef looks nice.
What's the rationale for dropping defaultRenderFunctions? Using React elements instead of functions will require using `cloneElement` unnecessarily.

packages/react/src/tooltip/root/useTooltipRoot.ts
atomiks: Just saw that `mergeProps` can also accept a function. So we should do `mergeProps(getRootTriggerProps, other, getButtonProps)`?

atomiks:
> What's the rationale for dropping defaultRenderFunctions? Using React elements instead of functions will require using cloneElement unnecessarily.

A few reasons:

1. It's not tree-shakeable. Passing `button` as a separate module ensures the other render functions aren't included when they don't have to be
2. `render = button` or `render = <p />` in the default props is much cleaner than `render: render ?? 'button'`, especially at the `useComponentRenderer` callsite where we can use the shortform object property, and matches what we recommend with our `useRender` hook
3. Is there a reason not to use `cloneElement`? I think I remember you mentioning perf, but I don't see why it would be slower necessarily (do you have a measurement of the difference?). With our `useRender` hook, we suggest to use a JSX element which defaults to that since it's the simplest case

michaldudak:
> Is there a reason not to use cloneElement? I think I remember you mentioning perf, but I don't see why it would be slower necessarily (do you have a measurement of the difference?)

`cloneElement`-based implementation is about 15% slower in my tests (https://gist.github.com/michaldudak/0850391c0a29f2ea23d39fafaf6a8168, run on Node 22.14 on Windows)

atomiks:
> cloneElement-based implementation is about 15% slower in my tests

I guess this could be case-by-case, like in list components where hundreds/thousands can be rendered, we must use a function, but it's likely we can get away with it for ones that are less frequently rendered?

Creating separate module functions is better than baking it into `useComponentRender` anyway if we want to always use them: `render = div`, where `div` is a function imported similar to `button` in this PR already, etc for other tags

Perhaps something like this would be better (we don't need to care about the tag-specific prop types):

```js
export function createRenderElement(Tag: string) {
  return (props: GenericHTMLProps) => <Tag {...props} />
}
```

atomiks:
New usage:

- `useComponentRenderer` -> `useRenderElement`, and it returns the function directly. The object is only useful if we need to add more properties later, which is more beneficial for a public API but not an internal one. We haven't had a use for this.
- The component's element (tag or function) is passed as a first argument (acting as the required fallback), while the `render` and `className` component props object is passed as a second argument. Parameters (previously settings) are last. The main reason to use a tag as a first argument is to bypass the `@default` jsdoc issue with the `render` prop while keeping the usage clean—since it's a required option, it makes more sense to put it as a first arg.
- `state` is optional
- A string creates a performant render function using the `tag` generic util, while `button` and `img` are separate module functions (since they have props on them)
- `props` accepts an array of native/intrinsic DOM props (not component props) to be merged with `mergeProps`—separate the naming from the upper component level `props` by using `...intrinsicProps`, not `...otherProps` which is ambiguous as to what it contains
- Prop getters are only from `getReferenceProps` and `getFloatingProps`, just use a plain object otherwise

This seems to be the cleanest for creating components internally and best for bundle size

```tsx
const renderElement = useRenderElement('div', props, {
  state,
  ref: [ref, arrowRef],
  props: [{ style: arrowStyles, 'aria-hidden': true }, intrinsicProps],
});
```

michaldudak:
I like it!

> Prop getters are only from getReferenceProps and getFloatingProps, just use a plain object otherwise

Can `getReferenceProps` and `getFloatingProps` be called without arguments and merged alongside other props? I don't think I saw any non-trivial logic of merging props in them (passing them to `mergeProps` should give the same result).

atomiks:
> Can getReferenceProps and getFloatingProps be called without arguments and merged alongside other props? I don't think I saw any non-trivial logic of merging props in them (passing them to mergeProps should give the same result).

Most likely - though the prop merging logic we have is a bit confusing with how `useButton` operates

I'm going to open a new PR and do components separately since this touches a lot of code and is prone to conflicts and causing a regression somewhere.

PR #1560, opened 2025-03-14
[NumberField] `snapOnStep` prop

### LINKED ISSUE #1010: [number field] The `largeStep` prop should not round up/down
Instead, it should just add/subtract the largeStep value to/from the current value. We could have a prop perhaps, to change this behaviour to rounding.

**Search keywords**:

### LINKED ISSUE #1542: [number field] Prop to disable rounding of value when setting step increments
# Feature request

## Summary

Could we add a prop to to enable/disable rounding of the increment steps to the current value.

## Motivation / Use Case

When a value is passed to the component (in my case often a decimal value, from an API), I would like to be able to increment on a decimal base, but without rounding the other non-incremental decimals. 

As you can see in the example, when the `step` is undefined, and defaults to 1+, then the decimals aren't rounded. Would be amazing if you could make that behavior optionally achievable.

https://github.com/user-attachments/assets/fdc981fa-85c6-4b74-a9bd-7cac756cc751

atomiks:
Closes #1542
Closes #1010

The new `snapOnStep` prop is a boolean:

- `true`: the value is snapped to the nearest step when stepping (incrementing/decrementing). So `1.07` with a step of `1` becomes `2` when incrementing
- `false` (new default): the value isn't snapped at all. So `1.07` with a step of `1` becomes `2.07`.

Also new behavior for `true`:

- If you type a decimal then increment/decrement, it snaps to the nearest regular `step`, just like `largeStep` already did.
- It will snap the nearest step rather than round and unexpectedly go "more" than wanted.


michaldudak:
Using small or large step with the custom function doesn't work. Decrement doesn't work at all, while increment adds a double value. See https://codesandbox.io/p/devbox/silly-moser-t85lpm?file=%2Fsrc%2FApp.tsx%3A12%2C22&workspaceId=ws_QAeLcZEcDnfGaFA47etEUH

atomiks:
@michaldudak sandbox is private

michaldudak:
Try now

michaldudak:
It now works fine with `stepBehavior="snap"` for small and large steps, but the normal step doesn't snap (10.1 + step = 11.1)

atomiks:
@michaldudak yeah - hadn't yet changed the implementation after our chat. Just changed the math in a new commit and updated PR description with additional details

PR #1561, opened 2025-03-14
[select] Ensure switching controlled value to `null` updates `Select.Value` label

### LINKED ISSUE #1556: [select] Controlled select does not update when the external value changes
# Bug report

## Current behavior

When using a controlled select (`value` and `onValueChange` props are provided to `Select.Root`), the select updates for non-null changes to `value`, but when it is set to `null`, the previous state persists.

https://github.com/user-attachments/assets/03eb1ec4-f84e-4934-ab4d-fe5443ae9894

You can see that the placeholder is only visible upon initial load, before the state has been modified, even if it changes back to its original value.

## Expected behavior

Updating the value to `null` should cause the previously selected item to no longer appear selected, and the placeholder should be shown.

## Reproducible example

[CodeSandbox](https://codesandbox.io/p/devbox/cool-gagarin-tf8ngr?workspaceId=ws_KUu2LH4pxNiE4obMakXjek)

## Base UI version

1.0.0-alpha.6 (`@base-ui-components/react`)

Firefox 136

## Which OS are you using?

MacOS 15.3.1

N/A

## Additional context

N/A

atomiks:
Fixes #1556

Ensures the label is set back to empty string if value changes to `null` and there's no `null` `Select.Item` available in the options list


michaldudak:
It seems that the bug still appears with the updated version: https://codesandbox.io/p/devbox/purple-violet-nhqpmq?file=%2Fpackage.json

atomiks:
@michaldudak the referenced commit is wrong in the devbox

The user in the discord says the PR fixed their issue and the test also should confirm that

atomiks:
@michaldudak hmm are you sure the devbox is fresh/doesn't have some caching issue/referencing a wrong commit? Copy and pasting the example locally works fine. The test also wouldn't pass otherwise

michaldudak (approved):
OK, right, when I ran the sandbox locally it works fine.

PR #1564, opened 2025-03-16
[select] Improve `ScrollArrow` behavior

### LINKED ISSUE #1568: [Select] When `alignItemsToTrigger` is `false`, the popup overflows and scroll arrows won't show
# Bug report

## Current behavior

https://codesandbox.io/p/sandbox/competent-morning-mfkyry?file=%2Fsrc%2FApp.tsx%3A8%2C56-8%2C61

When `alignItemsToTrigger` is `false`, the popup overflows and scroll arrows won't show. Changing this prop's value to `true` and the issue goes away.

https://github.com/user-attachments/assets/b36b3c7b-96b6-4b22-a4be-69dba7e87b8e

## Expected behavior

Overflow restriction and scrolling for the Popup should work regardless of `alignItemsToTrigger` value.

## Reproducible example

https://codesandbox.io/p/sandbox/competent-morning-mfkyry?file=%2Fsrc%2FApp.tsx%3A8%2C56-8%2C61

## Base UI version

1.0.0-alpha.6

Chrome

## Which OS are you using?

Mac OS

atomiks:
From reviewing https://github.com/mui/mui-next/pull/67

- Mousing over scroll arrows removes stuck highlight
- Scrolling with the arrows now moves by full items every 40ms, matching native macOS behavior (like in a terminal where it scrolls full lines without landing on partial lines) which otherwise obscures items. Natural wheeling doesn't use this behavior still though
- Updates Select hero docs to 1) add the scroll arrows by default, and 2) add `overflow-y: auto` & `max-height: var(--available-height)` to the `Popup`

Preview: https://deploy-preview-1564--base-ui.netlify.app/experiments/long-select

Closes #1568


michaldudak:
I found one more issue: when the first visible item is partially obscured, focusing it should scroll to make it fully visible, but currently it doesn't:

https://github.com/user-attachments/assets/8cf80823-412c-48e7-980c-5542db29fc6c

atomiks:
Yeah I've seen that—the `scroll-margin` CSS only detects the edges of the scroll container without considering overlaid elements to create extra padding. I think we may need some kind of CSS variable to `calc` the `scroll-margin` to consider the height of the scroll arrow elements 🤔 

Edit: never mind, I remember trying to work on a solution to this earlier. The problem is the `.scrollIntoView()` function doesn't actually scroll the container to get `scroll-margin-block` to activate its offset, which seems to be an unfortunate shortcoming of the API (it should take into account `scroll-margin` when detecting whether to scroll). Can file a separate issue for that if necessary as I'm not sure how easy it is to fix (it might require changing how tall the popup is in order to preserve "full lines", but is quite a bit more invovled than should be handled in this PR)

michaldudak:
Sure, makes sense to do it in a separate PR.

packages/react/src/select/scroll-arrow/SelectScrollArrow.tsx
```diff
@@ -67,67 +70,121 @@ const SelectScrollArrow = React.forwardRef(function SelectScrollArrow(
           style: {
             position: 'absolute',
           },
-          onMouseEnter() {
-            if (!alignItemToTrigger) {
+          onMouseMove(event) {
+            if (
+              (event.movementX === 0 && event.movementY === 0) ||
```
michaldudak, line 75: I'm curious in what cases `onMouseMove` fires with `event.movementX === 0 && event.movementY === 0`?

atomiks: Screenshot 2025-03-19 at 7 00 47 pm

When moving slowly it can happen frequently

michaldudak (approved):

PR #1565, opened 2025-03-16
[Field] Fix `FieldControl` [data-filled] not reacting to external value changes

atomiks:
Previously, `FieldControl` only checked if the value exists on mount and `onChange`, but not when the `value` prop is changed externally

### DIFF
--- a/packages/react/src/field/control/useFieldControl.ts
+++ b/packages/react/src/field/control/useFieldControl.ts
@@ -39,10 +39,13 @@ export function useFieldControl(params: useFieldControl.Parameters) {
   }, [id, setControlId]);
 
   useEnhancedEffect(() => {
-    if (inputRef.current?.value) {
+    const hasExternalValue = valueProp != null;
+    if (inputRef.current?.value || (hasExternalValue && valueProp !== '')) {
       setFilled(true);
+    } else if (hasExternalValue && valueProp === '') {
+      setFilled(false);
     }
-  }, [inputRef, setFilled]);
+  }, [inputRef, setFilled, valueProp]);
 
   const [value, setValueUnwrapped] = useControlled({
     controlled: valueProp,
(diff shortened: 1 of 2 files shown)


packages/react/src/field/control/useFieldControl.ts
```diff
@@ -39,10 +39,13 @@ export function useFieldControl(params: useFieldControl.Parameters) {
   }, [id, setControlId]);
 
   useEnhancedEffect(() => {
-    if (inputRef.current?.value) {
+    const hasExternalValue = valueProp != null;
```
flaviendelangle, line 42: How should this behave when your field receives `value={null}`?
On the picker, `null` is still a controlled value which indicates that the component is empty, unlike `undefined` that indicates that the component is not controlled.
This is the behavior of the `useControlled` hook of `@mui/utils` as well.

But maybe it's irrelevant here.

atomiks: For `FieldControl` (like `<input>`), `null` isn't a valid type, only `undefined`  - the `!=` checks for null/undefined simultaneously in case of lack of TS

flaviendelangle: OK
And when we build more advanced fields like `<NumberField />` or a `<DateField />` we can support `null`?
Or is it an intrisic limitation right now?

atomiks: `NumberField` supports `null`, but `Field.Control` is just a wrapper over `input`. Rendering other Base UI controls as a `Field.Control` isn't recommended as they instead should just be placed inside `Field.Root` with regular nesting,  as they use context to read the field state

flaviendelangle: OK thanks for the clarification :ok_hand:

mj12albert (approved):
Looks good!

PR #1570, opened 2025-03-17
[select] Fast hover highlight perf

atomiks:
Preview: https://deploy-preview-1570--base-ui.netlify.app/experiments/select-perf

- Debounces `setActiveIndex(...)` until only necessary (to synchronize the hover and focus states)
- Mutates the item ref directly to set the `[data-highlighted]` attribute without needing to wait for React to re-render—main downside is unable to change `data-highlighted` in your component API as this is forcefully set
- Important: highlights should not be able to get stuck or be duplicated in any scenario

The highlights should be as fast as CSS `:hover` (@michaldudak, to check on your PC)

As a result of this PR:

- The `highlighted` state is now removed. You can't customize `[data-highlighted]` and can't rely on it as a state since we can't wait for React to re-render
- A bug with highlights remaining stuck if the mouseleave fired on the popup rather than the item is fixed
- A bug with highlight not being removed when expanding the popup with `alignItemToTrigger=true` is fixed
- A bug with the highlight not appearing when the initial value is `null` and there is no `null` item in the list with `alignItemToTrigger=true` is fixed
- `Menu` highlighted state has been removed, to sync with Select. However its implementation is still the same (we can apply it to that, if necessary, but it has a lot fewer items than a Select typically)


packages/react/src/toolbar/button/ToolbarButton.test.tsx
```diff
@@ -380,7 +380,7 @@ describe('<Toolbar.Button />', () => {
           expect(trigger).toHaveFocus();
         });
 
-        expect(handleValueChange.callCount).to.equal(1);
+        expect(handleValueChange.callCount).to.equal(2);
```
atomiks, line 383: Should be fixed after https://github.com/mui/base-ui/pull/1597, changed it here for now

michaldudak (approved):

PR #1571, opened 2025-03-18
[dialog][Popover] `modal="trap-focus"` prop value

### LINKED ISSUE #1554: [popups] Rework `modal` prop
# Feature request

## Summary

Discussion starting at https://github.com/mui/base-ui/pull/1459#issuecomment-2714057692 (@vladmoroz)

The word `modal` is quite overloaded and confusing, since there are actually three different user inputs that can be "modal", which can have varying combinations of being on or off:

- modal `focus` —  you can't focus anything behind (i.e. except what's in) the popup while it's open
- modal `scroll` — you can't scroll anything behind (i.e. except what's in) the popup while it's open
- modal `pointer` — you can't click anything behind  (i.e. except what's in) the popup while it's open

 Typically when people say something is "modal", they mean all three are modal simultaneously, such as in the case with most dialogs that are positioned in the center of the screen, have a dimming backdrop, and aren't anchored to another element.

But it can be desirable to have a popover be "non-modal" (with respect to pointers) but still trap focus, i.e. be modal to keyboard focus. 

Instead of the `modal` prop, I proposed the less-overloaded word `trap` instead, typed as `boolean | TrapOptions`:

```jsx
// fully modal popover
<Popover.Popup trap />

// "non-modal" popover with trapped (modal) focus
<Popover.Popup trap={{ focus: true, scroll: false, pointer: false }} />
```

Although "lock" is typically used for scroll modality, and "pointer" doesn't really have a word describing its modality, I think `trap` correctly describes what's happening to all three user inputs:

- Trapping their focus means their focus is _trapped_ inside the popup — they can't tab to anything except what's in the popup
- Trapping their pointer means their pointer is _trapped_ inside the popup — they can't click anything except what's in the popup
- Trapping their scroll means their scroll is _trapped_ inside the popup — they can't scroll anything except what's in the popup

### Combinations

One of the problems with this config object is that some combinations lead to an incoherent user experience.

| #  | `trap` value                                       | Description                                                                                   |
|----|----------------------------------------------------|-----------------------------------------------------------------------------------------------|
| 1  | `focus: true` \| `scroll: true` \| `pointer: true`    | ✅ All three are trapped — no focus, scroll, or pointer interactions allowed behind.            |
| 2  | `focus: true` \| `scroll: true` \| `pointer: false`   | ❌ Focus and scroll are trapped, but pointer interactions are allowed behind.                  |
| 3  | `focus: true` \| `scroll: false` \| `pointer: true`   | ⚠️ Focus and pointer are trapped, but scrolling is allowed behind. (Seen on GitHub's popups, less restrictive and just avoids sloppy pointer outside clicks. However, if the pointer is trapped, inner HTML scroll containers can't be scrolled, only document scroll is locked, so this can be confusing)                            |
| 4  | `focus: true` \| `scroll: false` \| `pointer: false`  | ✅ Only focus is trapped; scrolling and pointer interactions are allowed behind. (Vlad's popover)               |
| 5  | `focus: false` \| `scroll: true` \| `pointer: true`   | ✅ Scroll and pointer are trapped, but focus is allowed behind.  (used in Menu & Select)                                 |
| 6  | `focus: false` \| `scroll: true` \| `pointer: false`  | ❌ Only scroll is trapped; focus and pointer interactions are allowed behind.                  |
| 7  | `focus: false` \| `scroll: false` \| `pointer: true`  | ⚠️ Only pointer is trapped; focus and scrolling are allowed behind.  (Seen on GitHub's popups, variant of the other one)                            |
| 8  | `focus: false` \| `scroll: false` \| `pointer: false` | ✅ No trapping — focus, scrolling, and pointer interactions are all allowed behind.             |

atomiks:
Closes #1554 @vladmoroz 

This lets the `modal` prop accept a `'trap-focus'` value. New descriptions were added for clarity.

- **Dialog**: `modal={false}` now disables focus trapping, whereas previously it was always enabled. `modal="trap-focus"` only traps focus to add that behavior back (ignoring scroll & pointer)
- **Popover**: `modal="trap-focus"` now only enables focus trapping


atomiks:
@michaldudak thanks for spotting all those, this PR changed APIs several times and got messed up by the end

michaldudak (approved):
Works well now!

dartess:
Based on the description, modal=true shouldn't have changed. But in my popover with modal=true, the focus trap doesn't work anymore. That is, the background is blocked, the scroll is blocked, but Tab easily closes the popup when losing focus. 

Is this the intended behavior? How can I now both enable the focus trap AND block interaction with the rest of the page? I don't want to let the user leave the component until they explicitly select an action or explicitly close the popup.

atomiks:
@dartess Popover with `modal=true` didn't trap focus to begin with. You can see in the diff that `modal={false}` on `FloatingFocusManager` which means it didn't trap focus

The main reason for this is that (like Menu and Select), a Popover without an explicit close button can't be closed on touch devices using a screen reader. An alternative is a visually-hidden dismiss button by default, but it's not internationalized, or checking if `Popover.Close` was rendered, but this doesn't support custom close buttons that avoid using that part

dartess:
@atomiks I probably missed something. Thanks for the resonse.

PR #1573, opened 2025-03-18
[popups] Fix selection on outside press on Firefox with modal prop

atomiks:
This was reported on Discord and is worse on Firefox. When simply pressing outside (with a hard click on a trackpad) it selects the document's text unexpectedly. 

The internal backdrop `pointer-events: none` style seems to cause the bug while closing on FF & Safari; `modal={false}` is the current workaround since it avoids rendering the backdrop

In order to maintain the fix in https://github.com/mui/base-ui/pull/1221, I changed the style hook for `[data-popup-open]` to match the `mounted` state to prevent the flicker instead. It doesn't map 1:1 with the `open` state, but I'm not sure if that matters here since in practice you want it kept when mounted, not open

### DIFF
--- a/packages/react/src/alert-dialog/popup/AlertDialogPopup.tsx
+++ b/packages/react/src/alert-dialog/popup/AlertDialogPopup.tsx
@@ -112,7 +112,7 @@ const AlertDialogPopup = React.forwardRef(function AlertDialogPopup(
 
   return (
     <React.Fragment>
-      {mounted && modal && <InternalBackdrop ref={internalBackdropRef} inert={!open} />}
+      {mounted && modal && <InternalBackdrop ref={internalBackdropRef} />}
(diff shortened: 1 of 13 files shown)


mj12albert (approved):
I tested on a few versions of Firefox and the issue is fixed, though  I couldn't reproduce this in Safari

atomiks:
@mj12albert on Safari you kind of have to drag the cursor a bit when clicking outside, hence the "sloppy" in that issue, but it's harder to repro  indeed

Edit: I tested a bit more in Safari and it actually seems like it's the scroll lock that causes it there, as this issue only changes the selection behavior a bit in Safari but it's still there. Or maybe it's fixed in a new version since you can't reproduce it 🤔  I will unmark the issue being closed for now

Issue #1576, opened 2025-03-19
NumberInput resets to default value when input exceeds five digits


KasiaSmolarska:
# Bug report

## Current behavior

When entering a five-digit value in the `NumberInput` component and clicking the **Increase Value** or **Decrease Value** button, the input value resets to the default value. This happens regardless of the `max` value set for the component.

You can reproduce this issue using the following example:  
[Base UI NumberField Example](https://base-ui.com/react/components/number-field).

## Expected behavior

When a five-digit value is entered, clicking **Increase Value** or **Decrease Value** should correctly update the value based on the step logic, without resetting it to the default.

## Reproducible example

[Base UI NumberField Example](https://base-ui.com/react/components/number-field).

## Base UI version

1.0.0-alpha.6

## Which browser are you using?

Chrome

## Which OS are you using?

Windows

## Additional context

This issue occurs regardless of the `max` value specified. Even if `max` is set higher than five digits, the input still resets unexpectedly.

mj12albert:
https://codesandbox.io/p/sandbox/elated-hill-rf4q4z?file=%2Fsrc%2FApp.tsx

@KasiaSmolarska I couldn't reproduce this, it works fine for me using these props:
```jsx
<NumberField.Root
  defaultValue={9999}
  min={1}
  max={999999}
/>
```

Can you share a sandbox with a repro?

atomiks:
@mj12albert perhaps related to the runtime locale since that changes the formatting. What's your lang setting @KasiaSmolarska?

KasiaSmolarska:
My browser interface is in Polish, but navigator.language is set to en-US. Could this mismatch cause issues with number formatting or how values are being processed? 

After changing the browser interface language to English, it started working correctly.

KasiaSmolarska:
Hi @mj12albert,

Thanks for the suggestion! I noticed that when navigator.language is set to pl (or pl-PL), it may cause issues with number formatting due to a mismatch in locale settings.

After changing the browser interface language to en-US, the issue disappeared, and everything started working correctly. It seems that the library may not handle number formatting properly for pl-PL.

Is there a way to enforce a specific number format or better handle different locales, regardless of the system language settings?

Thanks!

atomiks:
In our upcoming release, `NumberField` will have a `locale` prop, but this overrides the user's runtime one. I'll need to investigate this to see why the mismatch causes a problem

michaldudak:
[...]

PR #1577, opened 2025-03-19
[NumberField] Fix parse of numbers with spaces as thousands separators

michaldudak:
Certain locales define space as a thousands separator in numbers. The NumberField incorrectly parsed numbers from these locales. If a thousands separator is a space-like character (including non-breaking  and different width spaces), we treat every space-like character as a possible separator.

Fixes https://github.com/mui/base-ui/issues/1576

### DIFF
--- a/packages/react/src/number-field/utils/parse.ts
+++ b/packages/react/src/number-field/utils/parse.ts
@@ -12,7 +12,7 @@ export function getNumberLocaleDetails(
   locale?: Intl.LocalesArgument,
   options?: Intl.NumberFormatOptions,
 ) {
-  const parts = getFormatter(locale, options).formatToParts(1111.1);
+  const parts = getFormatter(locale, options).formatToParts(11111.1);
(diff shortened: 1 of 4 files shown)


packages/react/src/number-field/utils/parse.ts
```diff
@@ -46,9 +46,15 @@ export function parseNumber(
   }
 
   const { group, decimal, currency, unit } = getNumberLocaleDetails(computedLocale, options);
+  let groupRegex: RegExp | null = null;
+  if (group) {
+    // Check if the group separator is a space-like character.
+    // If so, we'll replace all such characters with an empty string.
+    groupRegex = /\p{Zs}/u.test(group ?? '') ? /\p{Zs}/gu : new RegExp(`\\${group}`, 'g');
+  }
```
atomiks, line 54: `if (group)` and `group ?? ''` seems redundant?

michaldudak: Copy-pasta...

atomiks (approved):

PR #1578, opened 2025-03-19
[NumberField] Fix `large/smallStep` getting stuck

### LINKED ISSUE #1064: [number field] Shift modifier key is weirdly sticky
# Bug report

## Current behavior

- Open https://master--base-ui.netlify.app/react/components/number-field
- Use the built-in macOS hotkeys to either:
  - Start a screen recording
  - Make a screenshot of the page via cmd shift 4 → spacebar → mouse click on the browser window
- Click increment/decrement
- Increments in steps of 10

https://github.com/user-attachments/assets/ba3188c9-4ad3-465c-b914-371b312eb474

## Expected behavior

Increments in steps of 1

**Search keywords**:

atomiks:
Replaces the global listeners by checking for `event.shfitKey/altKey` on the event if available. 

The only times it's not available on the event is when pressing and holding the increment button and then pressing shift midway through, but I don't consider that to be very important/relevant

Fixes #1064

### DIFF
--- a/packages/react/src/number-field/input/NumberFieldInput.tsx
+++ b/packages/react/src/number-field/input/NumberFieldInput.tsx
@@ -226,7 +226,7 @@ const NumberFieldInput = React.forwardRef(function NumberFieldInput(
             // We need to commit the number at this point if the input hasn't been blurred.
             const parsedValue = parseNumber(inputValue, locale, formatOptionsRef.current);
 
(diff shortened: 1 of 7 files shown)


michaldudak (approved):
> The only times it's not available on the event is when pressing and holding the increment button and then pressing shift midway through, but I don't consider that to be very important/relevant

Agree. I'm pretty sure everyone will be able to live with it.

PR #1583, opened 2025-03-20
[menu] Ensure submenu triggers respond to clicks when `openOnHover=false`

### LINKED ISSUE #1582: [Menu] can't open submenu with click if openOnHover is false
# Get help

## Ask a question

Seems to me that if we set the `openOnHover` props to false in the Root component of the SubMenu we can't open it with click anymore, only keyboard shortcuts work. Is this intended ?

I don't like the hover behavior for submenu and I'd like to be able to just click it to open it with mouse or use the keyboard shortcuts. 

https://codesandbox.io/p/sandbox/small-river-7w7hf9

atomiks:
Closes #1582

### DIFF
--- a/packages/react/src/menu/root/MenuRoot.test.tsx
+++ b/packages/react/src/menu/root/MenuRoot.test.tsx
@@ -486,6 +486,45 @@ describe('<Menu.Root />', () => {
         expect(submenuTrigger).toHaveFocus();
       });
     });
+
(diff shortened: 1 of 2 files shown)


mj12albert (approved):
It works correctly now: https://codesandbox.io/p/sandbox/cocky-tess-q4rpr3

PR #1584, opened 2025-03-20
[NumberField] Fix ScrubArea on Safari

atomiks:
There should be no WebKit check for pointerup release. After the promise changes, Safari no longer cancels scrubbing.


mj12albert (approved):
👍 Tested this and it no longer gets stuck in the scrubbing state after releasing the pointer

PR #1586, opened 2025-03-20
[core] Clean up Babel and next.js configuration

michaldudak:
- removed usused Babel config options (as we now use Babel only to build the Base UI package)
- removed unused plugin packages
- configured the build for modern browsers
- removed unused settings from Next.js configuration


babel.config.js
```diff
@@ -35,15 +37,7 @@ module.exports = function getBabelConfig(api) {
...
-    'babel-plugin-optimize-clsx',
-    [
-      '@babel/plugin-transform-runtime',
-      {
-        useESModules,
-        // any package needs to declare 7.4.4 as a runtime dependency. default is ^7.0.0
-        version: '^7.4.4',
-      },
-    ],
+    ['@babel/plugin-transform-runtime', { regenerator: false, version: '^7.26.10' }],
```
Janpot, line 40: in core/X we read the runtime version from the package.json

Janpot:
This is great. Is babel still used for anything other than production builds? Eventually I'd love to eliminate these babel configs altogether and add them into the build script.

babel.config.js
michaldudak: :+1: updated

michaldudak:
> Is babel still used for anything other than production builds? Eventually I'd love to eliminate these babel configs altogether and add them into the build script.

No, we use it only to build the library (unless some scripts from the Core repo use it). Getting rid of the config could be a good next step.

babel.config.js
```diff
@@ -11,7 +12,7 @@ module.exports = function getBabelConfig(api) {
       '@babel/preset-env',
       {
         bugfixes: true,
-        browserslistEnv: process.env.BABEL_ENV || process.env.NODE_ENV,
+        browserslistEnv: process.env.NODE_ENV,
```
Janpot, line 15: Can this be `api.env()`? Two different flags (NODE_ENV and BABEL_ENV) made this file hard to reason about in core.

Janpot (approved):
Looks great. If possible, I'd remove the NODE_ENV from babel config altogether, it'll be easier to reason about.

babel.config.js
michaldudak: Sure thing!

PR #1592, opened 2025-03-21
[test] Setup tests on all browsers

michaldudak:
Prepare npm scripts for running the test in each supported browser (Chromium, Firefox and WebKit) separately or in all at once.

I initially intended to run all browsers on CI, but they are quite unstable.


michaldudak:
Firefox connection timeouts look like https://github.com/vitest-dev/vitest/issues/7377

package.json
```diff
@@ -43,11 +43,16 @@
     "test:regressions:dev": "vite --config test/regressions/vite.config.mjs",
     "test:regressions:run": "cross-env VITEST_ENV=chromium vitest --project regressions",
     "test:regressions:server": "serve test/regressions -p 5173",
-    "test:jsdom": "cross-env NODE_ENV=test VITEST_ENV=jsdom vitest --project @base-ui-components/react --project docs",
+    "test:_unit": "cross-env NODE_ENV=test vitest --project @base-ui-components/react --project docs",
```
mj12albert, line 46: Is the underscore in `test:_unit` intentional?

michaldudak: yes, to indicate it's a subtask, not meant to be used directly

mj12albert (approved):

PR #1596, opened 2025-03-23
[select] Reset `selectedIndex` when set to `null`

### LINKED ISSUE #1556: [select] Controlled select does not update when the external value changes

atomiks:
Closes #1556

### DIFF
--- a/packages/react/src/select/root/SelectRoot.test.tsx
+++ b/packages/react/src/select/root/SelectRoot.test.tsx
@@ -496,7 +496,7 @@ describe('<Select.Root />', () => {
               opacity: 0;
             }
           }
-  
+
           .animation-test-indicator[data-starting-style] {
             animation: test-anim 1ms;
           }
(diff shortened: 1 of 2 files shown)


atomiks:
@shayypy can you confirm this fixes 'all' cases? Hopefully not missing anything else. It should probably work like the `key={}` workaround

shayypy:
@atomiks hi, yes! Cloned locally to test and this allows a reset of both the value and the `ItemIndicator` as expected. Thank you!

mj12albert (approved):
👍 
Tested this too and it works: https://codesandbox.io/p/devbox/angry-sun-jmsglw

PR #1597, opened 2025-03-23
[select] Avoid double commit on value change

### LINKED ISSUE #1589: Select onValueChange handler fires twice
# Bug report

## Current behavior
The `Select` component's `onValueChange` handler fires twice

## Expected behavior
It should only fire once. Disabling React's strict mode does not solve the problem.

## Reproducible example
https://codesandbox.io/p/sandbox/bold-fire-sjwsw4?file=%2Fsrc%2FApp.tsx%3A11%2C11

## Base UI version
1.0.0-alpha.7

Chrome.

## Which OS are you using?
Mac OS

atomiks:
Closes #1589

Solution here is to only run `onClick` if there wasn't a `pointerdown` on the item when the click occurred (since `onMouseUp` also runs). `onClick` doesn't run if a pointerdown didn't occur first.

Also ensures the `event` is passed for `onValueChange`

### DIFF
--- a/packages/react/src/field/root/FieldRoot.test.tsx
+++ b/packages/react/src/field/root/FieldRoot.test.tsx
@@ -839,6 +839,14 @@ describe('<Field.Root />', () => {
   });
 
   describe('style hooks', () => {
+    const { render: renderFakeTimers, clock } = createRenderer({
+      clockOptions: {
+        shouldAdvanceTime: true,
+      },
+    });
+
+    clock.withFakeTimers();
+
     describe('touched', () => {
       it('should apply [data-touched] style hook to all components when touched', async () => {
         await render(
@@ -1141,7 +1149,7 @@ describe('<Field.Root />', () => {
       });
 
       it('supports Select', async () => {
-        const { user } = await render(
+        const { user } = await renderFakeTimers(
           <Field.Root>
             <Select.Root>
               <Select.Trigger data-testid="trigger" />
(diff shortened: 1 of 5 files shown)


packages/react/src/field/root/FieldRoot.test.tsx
```diff
@@ -1162,16 +1170,14 @@ describe('<Field.Root />', () => {
         expect(trigger).not.to.have.attribute('data-dirty');
 
         await userEvent.click(trigger);
-
         await flushMicrotasks();
+        clock.tick(200);
```
atomiks, line 1174: This is unexpected when trying to test item selections since the logic blocks it from selecting within 200ms for UX reasons — not sure how we should make this easier in testing envs

packages/react/src/select/item/useSelectItem.ts
```diff
@@ -34,14 +34,14 @@ export function useSelectItem(params: useSelectItem.Parameters): useSelectItem.R
...
-    handleSelect();
+  const commitSelection = useEventCallback((event: MouseEvent) => {
+    handleSelect(event);
     setOpen(false, event);
   });
 
   const lastKeyRef = React.useRef<string | null>(null);
   const pointerTypeRef = React.useRef<'mouse' | 'touch' | 'pen'>('mouse');
-
+  const didMouseDownRef = React.useRef(false);
```
michaldudak, line 44: To be more precise

```suggestion
  const didPointerDownRef = React.useRef(false);
```

michaldudak (approved):

PR #1598, opened 2025-03-24
[ScrollArea] Handle visibility change and nesting

atomiks:
When using ScrollArea inside Select, all the wrappers need `role="presentation"` to not interfere with the `listbox` role. 

Further, the thumb was not computed when the ScrollArea became visible, and incorrectly reset the scrollbar offset if it had no width/height (such as on hide/unmount).

https://deploy-preview-1598--base-ui.netlify.app/experiments/scroll-area/inside-select


michaldudak:
There's an issue when the viewport is resized while the Select is open. Once the viewport edge reaches the thumb, the native scrollbar appears.

https://github.com/user-attachments/assets/dfaed2df-5ece-44b5-b791-f7f190303a71

atomiks:
@michaldudak updated - looks like that's just because `overflow-y: auto` is on `.Popup` in the experiment, but should be removed when using ScrollArea inside of it

michaldudak:
Now the scrollbar appears on the viewport:

https://github.com/user-attachments/assets/0aef3123-550e-4e67-bfca-00bf3f817dcb

(https://codesandbox.io/p/devbox/determined-cache-rm6n4x?file=%2Fsrc%2FApp.tsx&workspaceId=ws_QAeLcZEcDnfGaFA47etEUH)

Instead of recalculating the thumb position:

https://github.com/user-attachments/assets/5dc9a6d1-4c69-43ba-a763-091fd384c4fa

(https://codesandbox.io/p/sandbox/hopeful-babycat-nvh9pq?file=%2Fsrc%2FApp.tsx)

atomiks:
@michaldudak I can't repro the first one on macOS with the scrollbars always visible setting - is it also visible on the current v1.0.0-alpha.7? 

Second one, I can

michaldudak:
The second video shows alpha.7

atomiks:
@michaldudak Oh right, I thought you meant it was a second issue not recalculating the position. But it seems like they're related: because it doesn't recalculate the position it overflows the bottom and causes the main scrollbar to appear. Seems like that's due to the `Content` part being removed by default, and the ResizeObserver ends up bailing out of both the viewport and content nodes

michaldudak:
Sorry I wasn't clear enough! It seems to work well now. I'll do a final pass of the implementation now.

michaldudak (approved):

PR #1601, opened 2025-03-24
[accordion][Collapsible] Rework animations and transitions

### LINKED ISSUE #740: [collapsible] Layout shift when using CSS transitions and the `Panel` is initially open
only the one using CSS transitions is consistently flagged for CLS here

https://deploy-preview-577--base-ui.netlify.app/experiments/collapsible-cls

Screenshot 2024-10-25 at 9 30 35 PM

### LINKED ISSUE #1334: [collapsible] Opacity starting style not working

mj12albert:
Closes https://github.com/mui/base-ui/issues/1334
Closes https://github.com/mui/base-ui/issues/740

### Some highlights:
- All the accordion/collapsible experiments in `/experiments` are updated, for example [CSS transitions](https://deploy-preview-1601--base-ui.netlify.app/experiments/collapsible/transitions), check out the sidebar
- Tailwind v4 defaults started applying [`display: none !important` on `[hidden]`](https://github.com/tailwindlabs/tailwindcss/blame/main/packages/tailwindcss/preflight.css#L382) which we must override in order to measure the node. This also complicates integration with motion.
- Used an `AbortController` (as a new optional param to `useAnimationsFinished`) to cancel pending unmounts if the open state changes back to open before the pending unmount actually occurs (which is after the duration of the exit transition). It's still not possible to reverse CSS animations smoothly using with this.
- There is a weird edge case when `keepMounted={false}` and the panel is initially closed, when it opens for the first time and only the first time `data-starting-style` is missing or off by a frame and it needs to be manually set: `element.setAttribute('data-starting-style', '')`
- When using CSS transitions, the side that opens (height or width) needs to be set to `0px` for a frame, but not both together or it will cause unwanted stretching. If height/width are both transitioned (may not even be a real use case), `transition-property` must be `height, width` instead of `all`.


atomiks:
`Accordion` -> `Transitions` all examples seem perfect

`Accordion` -> `Horizontal` (RTL) has some glitches

https://github.com/user-attachments/assets/cca27f6f-55ff-4978-a0da-3d2b0cb60c61

`Accordion` -> `Animations` `keepMounted: true` seems to replay the animation on mount after it finishes

https://github.com/user-attachments/assets/27a4ad52-4ed5-43a3-a6aa-74382c3c3689

Not sure if some of these are to do with the experiments' CSS implementation or the source code though

packages/react/src/collapsible/root/CollapsibleRoot.tsx
```diff
@@ -29,10 +29,12 @@ const CollapsibleRoot = React.forwardRef(function CollapsibleRoot(
     ...otherProps
   } = props;
 
+  const onOpenChange = useEventCallback(onOpenChangeProp ?? NOOP);
```
atomiks, line 32: `useEventCallback` can accept undefined values now

packages/react/src/accordion/root/AccordionRoot.test.tsx
```diff
@@ -129,7 +129,10 @@ describe('<Accordion.Root />', () => {
   });
 
   describe('controlled', () => {
-    it('open state', async () => {
+    it('open state', async ({ skip }) => {
+      if (isJSDOM) {
+        skip();
+      }
```
atomiks, line 135: `it.skipIf` should be possible now

mj12albert:
> Accordion -> Animations keepMounted: true seems to replay the animation on mount after it finishes

Fixed in https://github.com/mui/base-ui/pull/1601/commits/0b659f30e00a93736f5df14e4becc54d5435c18c, an `abort()` was missing

> `Accordion` -> `Horizontal` (RTL) has some glitches

I think it's just the demo CSS, I cleaned it up a little though I think even with vertical scripts, a horizontal opening accordion with variable width doesn't seem like a real use-case

However I did find a another bug when `openMultiple={false}` and `keepMounted={true}` and then switching from one open panel to another, the exit transition of the closing panel is skipped while the enter transition of the opening runs fine 🤔

mj12albert:
> when `openMultiple={false}` and `keepMounted={true}` and then switching from one open panel to another, the exit transition of the closing panel is skipped while the enter transition of the opening runs

Fixed in ~~https://github.com/mui/base-ui/pull/1601/commits/3c5b1c21a5d41de316c9d9f52601de01d15b82d2~~ https://github.com/mui/base-ui/pull/1601/commits/7aa9bc53f1e3b213d57351a945f878b42f75a478 and changed the `hidden` attribute to match popups (set after the exit animation runs when closing)

atomiks:
Collapsible native CSS ones look good, but Motion examples seem broken. Clicking "Animate presence" example trigger quickly shows the bug.

mj12albert:
> Motion examples seem broken

Tailwind v4 still gets in the way here, motion sets styles with `element.style[key] = '…'` which doesn't support `!important`: https://github.com/motiondivision/motion/issues/1285#issuecomment-934332108

On the other hand we can't (internally) detect when motion's exit animation finishes to defer setting `hidden`, so probably the best workaround is to put [`hidden={undefined}` on the panel](https://github.com/mui/base-ui/blob/7aa9bc53f1e3b213d57351a945f878b42f75a478/docs/src/app/(private)/experiments/collapsible/animate-presence.tsx#L28), as we're unmounting after exit anyway

With Tailwind v3 (or just CSS) this is enough: https://codesandbox.io/p/sandbox/gracious-tdd-c5ysdd
```jsx
<Collapsible.Panel
  className="flex …"
  keepMounted
  key="MyPanel"
  render={
    <motion.div
      initial={{ height: 0 }}
      animate={{
        height: "auto",
      }}
      exit={{
        height: 0,
        // `display: block` here is critical to ensure the panel
        // is visible during the exit animation
        display: "block",
      }}
    />
  }
/>
```

docs/src/app/(private)/experiments/collapsible/transitions.tsx
```diff
@@ -0,0 +1,91 @@
+'use client';
+import * as React from 'react';
+import { Collapsible } from '@base-ui-components/react/collapsible';
+import { ChevronIcon } from './_icons';
+import styles from './transitions.module.css';
```
atomiks, line 5: I sometimes noticed a flash on this example when refreshing

With 20x CPU slowdown it actually plays the transition on page load/mount

mj12albert: > With 20x CPU slowdown it actually plays the transition on page load/mount

I see the panel just flicker into existence, and the height is not actually transitioning at all, is that what you see? @atomiks 

I thought the issue was related to suppressing the enter transition in an initial open state, but it seems like it's because the styles from `data-starting-style` are applied for a frame

`useTransitionStatus` always starts undefined so when initially open panels mount, the status always goes from undefined -> starting -> idle which causes that 

I think the hook needs to be able to start at idle for these cases, even though for popups its quite rare for them to start fully open and transitioned

mj12albert: Updated the hook in https://github.com/mui/base-ui/pull/1601/commits/1d1c0c711f8d4922095402847a7e7bb3b845d9f9

Quickly tested the docs/some experiments, doesn't seem to affect popups 🤞

benface:
@mj12albert, you're doing God's work here. Thank you so much for sweating the small stuff and making sure every edge case is covered.

atomiks (approved):
All the examples work great now

I'm still a bit confused why animating collapsible/accordion components ends up being more complex than popups, especially since both measure element size with JS and support `keepMounted`/`hidden` already using the old methods. 

Maybe there's a simpler approach here, but either way this is a big improvement from before

PR #1606, opened 2025-03-26
[ScrollArea] Correct thumb sizing with scrollbar margins

atomiks:
When there's slight overflow and the scrollbar itself has margins, the sizing was not correct. It needs to take into account the margin of the scrollbar for the new thumb size calculation - noticed this on the documentation


michaldudak:
Could you point me to where the issue appears currently?

atomiks:
@michaldudak I noticed it on one of the toast codeblocks: https://deploy-preview-1467--base-ui.netlify.app/react/components/toast (text rendering may cause the overflow to be different):

https://github.com/user-attachments/assets/41819af3-b1b4-47bb-b018-07e799b7e89b

Also added an experiment with this PR: https://deploy-preview-1606--base-ui.netlify.app/experiments/scroll-area/scroll-area-slight

michaldudak (approved):

Issue #1602, opened 2025-03-24
[Slider] Range slider thumbs can be dragged past the other thumb


mj12albert:
# Bug report

## Current behavior

From https://github.com/mui/base-ui/pull/1600#issuecomment-2747568092

https://github.com/user-attachments/assets/0ab23197-7984-43ef-bb78-bca9ba7ccbf8

## Expected behavior

Do not go past the another thumb

## Reproducible example

https://deploy-preview-1600--base-ui.netlify.app/react/components/slider#range-slider

## Base UI version

alpha.7/latest

mj12albert:
The issue is that since the value and positioning are separate, when dragging `validateMinimumDistance` needs to compare the % value here as well: https://github.com/mui/base-ui/blob/master/packages/react/src/slider/control/useSliderControl.ts#L105

PR #1612, opened 2025-03-28
[Slider] Prevent range slider thumbs from being dragged past each other

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1602

Demo: https://codesandbox.io/p/sandbox/great-wu-p6trjf

The issue was that `percentageValues` which determines the positioning also need to be clamped with adjacent values in range sliders like the value itself

### DIFF
--- a/packages/react/src/slider/root/useSliderRoot.ts
+++ b/packages/react/src/slider/root/useSliderRoot.ts
@@ -350,18 +350,20 @@ export function useSliderRoot(parameters: useSliderRoot.Parameters): useSliderRo
       }
 
       const closestThumbIndex = closestThumbIndexRef.current ?? 0;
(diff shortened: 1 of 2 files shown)


packages/react/src/slider/root/useSliderRoot.ts
```diff
@@ -358,13 +358,19 @@ export function useSliderRoot(parameters: useSliderRoot.Parameters): useSliderRo
         values[closestThumbIndex + 1] - minStepsBetweenValues || Infinity,
       );
 
+      const newPercentageValue = clamp(
+        valueRescaled * 100,
+        percentageValues[closestThumbIndex - 1] ?? -Infinity,
```
mj12albert, line 363: need to account for `minSteps...` here

atomiks (approved):

Issue #1494, opened 2025-03-01
How does it work with Motion?


subproject22:
So I tried to use it similar way as Radix, but it's doesn't work

```
"use client"

import * as React from "react"
import { useState } from "react"

import { Tabs } from "@base-ui-components/react/tabs"
import { AnimatePresence, motion } from "motion/react"

export function NewTabsDemo() {
  const [active, setActive] = useState("one")
  return (
    <Tabs.Root value={active} onValueChange={setActive} defaultValue="one">
      <Tabs.List>
        <Tabs.Tab value="one">One</Tabs.Tab>
        <Tabs.Tab value="two">Two</Tabs.Tab>
        <Tabs.Tab value="three">Three</Tabs.Tab>
      </Tabs.List>

      <AnimatePresence mode="popLayout">
        {active === "one" && (
          <Tabs.Panel
            className="bg-red-500 p-12"
            value="one"
            keepMounted
            render={
              <motion.div
                initial={{
                  opacity: 0,
                  y: -100,
                }}
                animate={{
                  opacity: 1,
                  y: 0,
                }}
                exit={{
                  opacity: 0,
                  y: 100,
                }}
              />
            }
          >
            one
          </Tabs.Panel>
        )}

        {active === "two" && (
          <Tabs.Panel
            keepMounted
            className="bg-green-500 p-12"
            value="two"
            render={
              <motion.div
                initial={{
                  opacity: 0,
                  y: -100,
                }}
                animate={{
                  opacity: 1,
                  y: 0,
                }}
                exit={{
                  opacity: 0,
                  y: 100,
                }}
              />
            }
          >
            two
          </Tabs.Panel>
        )}

        {active === "three" && (
          <Tabs.Panel
            keepMounted
            className="bg-yellow-500 p-12"
            value="three"
            render={
              <motion.div
                initial={{
                  opacity: 0,
                  y: -100,
                }}
                animate={{
                  opacity: 1,
                  y: 0,
                }}
                exit={{
                  opacity: 0,
                  y: 100,
                }}
              />
            }
          >
            three
          </Tabs.Panel>
        )}
      </AnimatePresence>
    </Tabs.Root>
  )
}
```

I haven't found a single examples of usage with Motion.

atomiks:
Hi @subproject22, in our next release the docs will be updated with some Motion examples but this mainly pertains to popup-related components, not ones like Tabs.

[...]

Issue #1486, opened 2025-02-28
[Tabs] Add `Tabs.Panels` component with `value` state in `render` function?


benface:
# Feature request

## Summary

I would like to have more flexibility in how tab panels are rendered, and one way to achieve that would be a `Tabs.Panels` component that, like other components, would accept a `render` function that would get passed some `state`... however, unlike `Tabs.Root` and `Tabs.List`, that state would include the current `value` so I can use something like [Motion's `AnimatePresence` component to transition from the previous tab panel to the current one](https://motion.dev/docs/react-animate-presence#changing-key).

## Examples in other libraries

Headless UI has [a `TabPanels` component](https://headlessui.com/react/tabs#tab-panels) that has a `selectedIndex` render prop.

## Motivation

Kind of answered in "Summary". Thank you!

mj12albert:
> however, unlike Tabs.Root and Tabs.List, that state would include the current `value`

@benface Would it help your use-case and avoid adding a new subcomponent if `value` was added to the state (the state arg of the render prop) of the existing Tabs subcomponents?

benface:
Thank you for the quick response @mj12albert!

I'm not sure it would help, unless it was possible to add it to `Tabs.Root` itself, but even then, I'm not 100% sure. I'll try to better explain my problem.

I just made [this CodeSandbox](https://codesandbox.io/p/sandbox/4n48xl) as a simple reproduction / test case. Notice that when you select a tab, there is an animation, but both the exiting and the entering contents are the same. I would expect the exiting contents to be the previously selected tab panel, and the entering one to be the newly selected one. It seems that by the time `App` re-renders with a new `selectedTab`, the `Tabs.Panel` components have already re-rendered, so Motion's `AnimatePresence` transitions from and to the same contents.

Now, I'm guessing that if you add `value` to `Tabs.Root`'s `render`'s state, it would execute that render function _before_ the `Tabs.Panel` re-render and so it would animate properly _in this case_, but I am working on a styled TabSet component that exposes its own `Tabs`, `Tab`, `Panels`, and `Panel` parts. The `Panels` component renders the `AnimatePresence` and `motion.div` wrapping the `children` passed to it, so if `value` was simply passed to `Tabs.Root`'s render function, I would need to pass it down to `Panels` via a context provider, and then I'm not sure it would re-render fast enough?

mj12albert:
https://deploy-preview-1613--base-ui.netlify.app/experiments/tabs/motion

[...]

PR #1613, opened 2025-03-28
[docs] Tabs examples with motion

mj12albert:
https://github.com/mui/base-ui/issues/1494
https://github.com/mui/base-ui/issues/1486


docs/src/app/(private)/experiments/tabs/motion.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const { children, value, ...otherProps } = props;
+  return (
+    <Tabs.Panel
+      ref={forwardedRef}
+      key={value}
+      value={value}
+      keepMounted
+      hidden={undefined}
+      {...otherProps}
+      render={(renderProps, state) => {
```
mj12albert, line 23: It would be handy for the `state` here to provide all the `--active-tab-*` values as well to plug into motion

mj12albert: Actually it doesn't matter, it accepts CSS variables anyway

https://motion.dev/docs/react-animation#css-variables

mj12albert: Couldn't get motion to animate based on these value, raw values are actually already in `state.selectedTabSize/Position.*`

Maybe motion values have to be used: https://motion.dev/docs/react-motion-value

mj12albert:
Closing this since motion has examples now

PR #1617, opened 2025-03-31
[core] Remove clashing props

### LINKED ISSUE #984: [core] Consider removing the built-in HTML attributes that clash with the component API
There's a few incredibly annoying HTML attributes that are present on every HTML element and can lead to confusion about what is the actual API of our components:

- `defaultValue`
- `defaultChecked`
- `color`

It gets confusing because autocomplete is eager to suggest them on components that are supposed to be used with different props, or when they are presented on parts that don't support either:

image

image

There might be others, specific to certain tags:
- The `value` attribute is inherited from `<button>` on `Checkbox.Root`. This clashes with how `CheckboxGroup` expects the `name` of the checkbox and not value
- Number Field input has `value`, but the value should be set on Root. Is this potentially confusing?
- etc

***

As a reference, in Radix Themes I had done this:
https://github.com/search?q=repo%3Aradix-ui%2Fthemes%20removedprops&type=code
https://github.com/radix-ui/themes/blob/main/packages/radix-ui-themes/src/components/checkbox.tsx#L22

**Search keywords**:

atomiks:
Closes #984

- `NumberField` incorrectly had an `invalid` prop that should belong to `Field`
- `Positioner` components incorrectly had `open` as a prop available
- Removed `color`, `defaultValue`, and `defaultChecked` from every component type (`BaseUIComponentProps`). Those that actually do have them for their API specify it manually to begin with
- Removed documented native props (`aria-label`, `aria-labelledby`, `autoFocus`, ...) that are already implicitly part of the native HTML element types. For `autoFocus`, we discourage it for focus management reasons inside popups (`initialFocus` is preferred), but also they can just place it on the `Input` subcomponent which is recommended
- Fixes empty string attrs on `Slider`


docs/reference/generated/slider-thumb.json
```diff
@@ -2,14 +2,6 @@
...
   "description": "The draggable part of the the slider at the tip of the indicator.\nRenders a `<div>` element.",
   "props": {
-    "aria-label": {
-      "type": "string",
-      "description": "The label for the input element."
-    },
-    "aria-valuetext": {
-      "type": "string",
-      "description": "A string value that provides a user-friendly name for the current value of the slider."
-    },
```
atomiks, line 12: Not 100% on removing the ARIA attributes even though they're already implicitly part of the native HTML types @mj12albert - is it just for documentation purposes to let them know it can be changed for i18n reasons?

atomiks: From what I can tell they all get forwarded to the owning element, with the exception of `aria-labelledby` on `Slider.Root` which can also get forwarded to a subcomponent but not sure if that's worth keeping documented as an explicit prop.

mj12albert: I think `aria-label`/`aria-valuetext` were explicit params because there are `getAriaLabel` and `getAriaValueText` props: https://github.com/mui/base-ui/blob/master/packages/react/src/slider/thumb/useSliderThumb.ts#L273

Basically the same as just forwarding them so I think it's ok to not document it

packages/react/src/utils/types.ts
```diff
@@ -41,7 +41,10 @@ export type BaseUIComponentProps<
   ElementType extends React.ElementType,
   State,
   RenderFunctionProps = GenericHTMLProps,
-> = Omit<WithBaseUIEvent<React.ComponentPropsWithoutRef<ElementType>>, 'className'> & {
+> = Omit<
+  WithBaseUIEvent<React.ComponentPropsWithoutRef<ElementType>>,
+  'className' | 'color' | 'defaultValue' | 'defaultChecked'
```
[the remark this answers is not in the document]
atomiks: In Vlad's issue, he also removed `color` for Radix Themes. I think it's because the attribute makes it seem like it's part of a styling API from Base UI, therefore just shouldn't exist / is never used anyway.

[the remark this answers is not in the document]
michaldudak: OK, let's move the discussion [to the issue](https://github.com/mui/base-ui/issues/984#issuecomment-2800813426).

michaldudak (approved):

PR #1618, opened 2025-03-31
[popover][Menu] Fix `inert` prop compatibility in React <19

sebinsua:
This is a follow-up to https://github.com/mui/base-ui/pull/1548.

[Older versions of React will give warnings](https://github.com/facebook/react/issues/17157#issuecomment-2003750544) like:

```
Received `false` for a non-boolean attribute `inert`.

If you want to write it to the DOM, pass a string instead: inert="false" or inert={value.toString()}.

If you used to conditionally omit it with inert={condition && value}, pass inert={condition ? value : undefined} instead.
```

I found [a fix within `react-spectrum`](https://github.com/adobe/react-spectrum/pull/7519) which I've ported into this repository.

### DIFF
--- a/packages/react/src/utils/inertValue.ts
+++ b/packages/react/src/utils/inertValue.ts
@@ -1,9 +1,7 @@
-import * as React from 'react';
+import { isReactVersionAtLeast } from './reactVersion';
 
 export function inertValue(value?: boolean): string | boolean | undefined {
-  const pieces = React.version.split('.');
-  const major = parseInt(pieces[0], 10);
-  if (major >= 19) {
+  if (isReactVersionAtLeast(19)) {
     return value;
   }
   // compatibility with React < 19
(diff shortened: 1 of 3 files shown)


packages/react/src/utils/inertValue.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function inertValue(value?: boolean): string | boolean | undefined {
+  const pieces = React.version.split('.');
+  const major = parseInt(pieces[0], 10);
+  if (major >= 19) {
+    return value;
+  }
+  // compatibility with React < 19
+  return value ? 'true' : undefined;
+}
```
atomiks, line 11: There's already a `reactVersion.ts` util (`isReactVersionAtLeast`)

sebinsua: OK, let me go and update this!

atomiks: There's a type error curently, you can just cast the fallback to boolean to silence it

sebinsua: Sorry, I made assumptions that what I was inlining from `react-spectrum` didn't have bad types...!

atomiks (approved):

PR #1662, opened 2025-04-01
[core] Re-use latest code infra scripts from core repo

brijeshb42:
Removed all the local code wherever possible and used scripts from the core repo.


scripts/testBuiltTypes.mjs
```diff
@@ -1,38 +1 @@
...
-      }
-    }),
-  );
-}
-
-main().catch((error) => {
-  console.error(error);
-  process.exit(1);
-});
+import '@mui/monorepo/scripts/testBuiltTypes.mjs';
```
Janpot, line 1: Can keep it for now, but looking at the script, I assume this is unnecessary if your declarations are built from typescript files.

scripts/useReactVersion.mjs
```diff
@@ -37,7 +37,9 @@ async function main(version) {
     return;
   }
 
-  const packageJsonPath = path.resolve(getWorkspaceRoot(), 'package.json');
+  const currentDirectory = url.fileURLToPath(new URL('.', import.meta.url));
+  const workspaceRoot = path.resolve(currentDirectory, '..');
```
Janpot, line 41: Can't this script just default to `process.cwd()` and add the script to root package.jsons scripts. Then it will always run with cwd of the root, regardless of whether it's executed from a subfolder.

scripts/validateTypescriptDeclarations.mts
```diff
@@ -1,28 +1 @@
...
- * Validates if there are no missing exports from TS files that would
- * result in an import from a local file.
- */
-function validateFiles() {
-  const declarationFiles = globbySync(['packages/*/build/**/*.d.ts'], {
-    followSymbolicLinks: false,
-  });
-  const invalidFiles = declarationFiles.filter((file) => {
-    const content = fs.readFileSync(file, 'utf8');
-    const regex = /import\(["']packages\//gm;
```
Janpot, line 15: 🤔 I feel like this is trying to do the exact same thing as testBuiltTypes. Should probably do a bit of an audit on all these scripts.

Janpot (commented):
Personally, I would remove the `getWorkspaceRoot` logic in favor for `process.cwd()`. Then never run the script directly, but always through a package.json script. That makes this script more portable within and across projects.

Looks great otherwise, I have a feeling some of those type validation scripts are a bit unnecessary.

scripts/validateTypescriptDeclarations.mts
michaldudak: This script was created to prevent issues such as https://github.com/mui/material-ui/issues/33321 (caused by referencing a type that's not exported).

scripts/useReactVersion.mjs
michaldudak: My personal preference is not to include everything in package.json scripts to avoid cluttering them, especially if the script is only used by the CI.

package.json
```diff
@@ -10,10 +10,10 @@
     "build:codesandbox": "NODE_OPTIONS=\"--max_old_space_size=4096\" lerna run --concurrency 8 --no-private build",
     "release:version": "lerna version --no-changelog --no-push --no-git-tag-version --no-private",
     "release:build": "lerna run --concurrency 8 --no-private build --skip-nx-cache",
-    "release:changelog": "node scripts/releaseChangelog.mjs",
+    "release:changelog": "node scripts/releaseChangelog.mjs --repo base-ui",
```
michaldudak, line 13: We plan to change the changelog script significantly to generate the changelog according to our format. Let's leave the local version so it's easier to change it.

packages/react/package.json
```diff
@@ -69,9 +69,9 @@
   "scripts": {
     "prebuild": "rimraf --glob build build-tests \"*.tsbuildinfo\"",
     "build": "pnpm build:node && pnpm build:stable && pnpm build:types && pnpm build:copy-files && pnpm build:manifest",
-    "build:node": "node ../../scripts/build.mjs node",
-    "build:stable": "node ../../scripts/build.mjs stable",
-    "build:types": "tsx ../../scripts/buildTypes.mts --project tsconfig.build.json --copy build/cjs",
+    "build:node": "node ../../scripts/build.mjs node --cjsDir ./cjs",
+    "build:stable": "node ../../scripts/build.mjs stable --cjsDir ./cjs",
+    "build:types": "tsx ../../scripts/buildTypes.mts --copy build/cjs",
```
michaldudak, line 74: Does it pick up the right tsconfig automatically?

brijeshb42: Yes. You can also pass explicitly if it not the default which is always `tsconfig.build.json`.

scripts/testBuiltTypes.mjs
brijeshb42: > Actually, it's looking for mui-* segments in the paths
True. I wonder if it was even working as intended.

brijeshb42:
> That makes this script more portable within and across projects.

By portable, you mean copy pastable to new repos in any directory, right ?

scripts/validateTypescriptDeclarations.mts
Janpot: 🤔 I'm suspecting the root cause of that issue is the use of [`baseUrl`](https://github.com/mui/base-ui/blob/45c8bd1102c66dff234feebc06bb253790a6db5d/tsconfig.base.json#L11). We should try if we can still reproduce this bug with that option removed.

Janpot:
> By portable, you mean copy pastable to new repos in any directory, right ?

No, I mean the location in the file system of the script makes no difference on whether it behaves correctly. So we can simply move it to a published package and make it available in the path with package.json bin and it'll still work. I don't think we should encourage copy/pasting these tools.

brijeshb42:
The way this particular script works right now, it'll need to be copy pasted unless we change that.
Rest can be used from an installable package.

Janpot (commented):
Great job. I can't reproduce any problem that would break the types validation functions. @michaldudak for a final approval.

scripts/useReactVersion.mjs
```diff
@@ -37,7 +36,8 @@ async function main(version) {
     return;
   }
 
-  const packageJsonPath = path.resolve(getWorkspaceRoot(), 'package.json');
+  const workspaceRoot = process.cwd();
```
michaldudak, line 39: Now we're in a state where we must run the script from the root directory and don't have it added to npm scripts to enforce the correct working directory. My preference would be to leave the previous version, but if you and @Janpot like reading `process.cwd` and calling the script from npm, let's have it your way.

Janpot: > but if you and @Janpot like reading process.cwd and calling the script from npm, let's have it your way.

FYI: I don't really have some bizarre fetish to see `process.cwd()` 😄. My preference would be that we centralize this script instead of copying it over multiple repositories. In that scenario, the `getWorkspaceRoot()` as implemented in this repo would never work as it would exist in a different location relative to the workspace root (such as in the node_modules folder). Any CLI tool we use operates on the current working dir, or its parents. That's why I advocate for not writing our scripts in a way that its location in the file system matters. That way we make it easy for them to be distributed across the company.

Personally, for this script, I think it's a bit overkill to look for a root, but as always, code infra can only work when the project teams buy in to the idea. So if there is a strong desire for the functionality we can certainly add it back, but I believe the correct implementation should be to start from `process.cwd()` and progressively check parent folders for a `pnpm-workspace.yaml` file.

package.json
```diff
@@ -64,11 +65,12 @@
     "@babel/preset-react": "^7.26.3",
     "@babel/preset-typescript": "^7.27.0",
     "@babel/register": "^7.25.9",
+    "@mui/internal-babel-plugin-resolve-imports": "2.0.0",
     "@mui/internal-docs-utils": "^2.0.1",
     "@mui/internal-markdown": "^2.0.4",
     "@mui/internal-scripts": "^2.0.7",
     "@mui/internal-test-utils": "^2.0.7",
-    "@mui/monorepo": "github:mui/material-ui#v7.0.2",
+    "@mui/monorepo": "github:mui/material-ui#259b71c8847a5a0a08976fd501a7207b7d5e8fb1",
```
michaldudak, line 73: Haven't these scripts been released yet?

brijeshb42: Not yet. Might be released this week.

brijeshb42: It's been released. Reverted.

michaldudak (approved):
I couldn't find any issues with this. Thanks for your contribution!

PR #1665, opened 2025-04-02
[ContextMenu] Create new `ContextMenu` component

### LINKED ISSUE #51: [menu] Create Context Menu component
It should be possible to create a context menu using the existing Dropdown and Menu components. Instead of a MenuButton to toggle the menu's visibility, another trigger should be created.

atomiks:
Closes #51


michaldudak:
We haven't discussed the API of it. I'm wondering if an ordinary Menu with a different Trigger wouldn't be enough:
```tsx
<Menu.Root>
  <Menu.ContextTrigger />
  ...
```

atomiks:
@michaldudak discoverability-wise, it might be better as a component for Radix parity? The Root needs to know if it should act as a context menu, where a component provider works nicely in the API, since otherwise it would need a prop in conjunction with using `Menu.ContextTrigger` (unless we set the state in an effect to know it's been rendered, which is a bit annoying).

~It's also better if it's part of `ContextMenu` for tree-shaking reasons~. (Edit: if `ContextTrigger` isn't used, where most of the logic is contained, then tree-shaking still works under your suggestion) - I think many extraneous features, like Tooltip's `trackCursorAxis` may be better as a component/provider than a prop if it has non-negligible logic.

mj12albert:
Testing on an iPhone SE (iOS 17.7.2), a long press on the trigger area when the context menu is already open will make a large text selection:

https://github.com/user-attachments/assets/d2ad1be0-4ea7-44d3-b9c2-4b9c3b8dfff4

I screen-capped Safari only but it repros on Firefox and Chrome as well

According to react-aria it requires quite a heavy handed approach to handle this for all browsers: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/src/textSelection.ts

atomiks:
@mj12albert looks like `WebkitUserSelect: none` on the internal backdrop is enough to fix that

mj12albert:
> looks like `WebkitUserSelect: none` on the internal backdrop is enough to fix that

Nice, glad we can avoid that extra Safari handling ~

michaldudak:
We could add an attribute indicating how the menu was open (touch or right-click). It seems that on mobile menus are often placed above the trigger (so they are not covered by user's hand). A callback variant of `side`/`placement` props could be useful here.

packages/react/src/context-menu/trigger/ContextMenuTrigger.tsx
```diff
@@ -0,0 +1,181 @@
...
+          style: {
+            WebkitTouchCallout: 'none',
+          },
+        },
+        externalProps,
+      ),
+    [handleContextMenu, handleTouchStart, handleTouchMove, handleTouchEnd],
+  );
+
+  const { renderElement } = useComponentRenderer({
```
michaldudak, line 140: Can be updated to use `useRenderElement`

packages/react/src/menu/popup/MenuPopup.tsx
```diff
@@ -49,6 +50,8 @@ const MenuPopup = React.forwardRef(function MenuPopup(
     onOpenChangeComplete,
   } = useMenuRootContext();
   const { side, align, floatingContext } = useMenuPositionerContext();
+  const contextMenuContext = useContextMenuRootContext();
+  const hasContextMenuParent = Boolean(contextMenuContext) && !nested;
```
michaldudak, line 54: In the Menubar PR I implemented a more universal solution to determine the menu's parent and its capabilities: https://github.com/mui/base-ui/pull/1684/files#diff-6da7b84ff007a487d62a9d48ba9fb32ad4b286ef666fd3d0419b14e9ebbe472a.

atomiks:
>  A callback variant of side/placement props could be useful here.

This could also be used instead of `useMediaQuery` to determine what side to use in narrower viewports. Would be good if we found a robust solution to https://github.com/mui/base-ui/pull/817 though.

packages/react/src/menu/popup/MenuPopup.tsx
atomiks: Context menus don't have any parents though, so I think this should suffice?

michaldudak: I'd like to avoid having all kinds of boolean values like `nested`, `isInContextMenu`, `isInMenubar`, (possibly) `isInNavigationMenu`. The approach I took in the Menubar PR lets a menu part know what kind of menu it's in in a standard way and provides a type-safe way to access the parent context (without having to check for its existence)

atomiks: Alright 👍 I'll wait for Menubar to be merged and then use its logic for `Menu`

oliviertassinari:
From https://x.com/rsms/status/1522583516265029633, the click-drag-release flow doesn't work properly:

https://github.com/user-attachments/assets/2668a1af-fb22-428f-a4b4-9c99062d3d48

https://www.radix-ui.com/primitives/docs/components/context-menu, https://ariakit.org/examples/menu-context-menu don't support this correctly either. So maybe OK to ignore fore a v0

---

Great to see this, we will need this for 👍 https://github.com/mui/mui-x/issues/12021

ImSingee:
Is it possible to stack two context menus? For example, can one context menu appear when clicking on a div and another when clicking outside of it?

I try to do it directly, but sadly it doesn't work:

https://codesandbox.io/p/sandbox/sweet-wave-h4lygt

image

I have two boxes here. Right-clicking the outer one displays the context menu as expected, but right-clicking the inner one does not show its context menu.

docs/reference/generated/context-menu-root.json
```diff
@@ -0,0 +1,67 @@
...
+      "description": "A ref to imperative actions.\n- `unmount`: When specified, the menu will not be unmounted when closed.\nInstead, the `unmount` function must be called to unmount the menu manually.\nUseful when the menu's animation is controlled by an external library."
+    },
+    "closeParentOnEsc": {
+      "type": "boolean",
+      "default": "true",
+      "description": "When in a submenu, determines whether pressing the Escape key\ncloses the entire menu, or only the current child menu."
+    },
+    "modal": {
+      "type": "boolean",
+      "default": "true",
```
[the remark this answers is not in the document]
atomiks: i think a merge went wrong somewhere. It shouldn't even have a `modal` prop (and always be `true`)

michaldudak:
The menu should return focus to the previously focused element when an item is clicked or <kbd>Esc</kbd> is pressed. This is pretty annoying when keyboard is used to open the menu.

packages/react/src/context-menu/root/ContextMenuRoot.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Menu } from '../../menu';
+
+/**
+ * A component that creates a context menu activated by right clicking or long pressing.
+ * Doesn’t render its own HTML element.
+ *
+ * Documentation: [Base UI Context Menu](https://base-ui.com/react/components/context-menu)
+ */
+export function ContextMenuRoot(props: ContextMenuRoot.Props) {
+  const { children, ...otherProps } = props;
```
romgrk, line 13: Doesn't need to be destructured, it's just being passed as-is below:
```
<Menu.Root {...otherProps}>{children}</Menu.Root>
```

colmtuite:
Testing in Chrome on iOS, when I open it and scroll before I release my longpress, it gets all shaky.

colmtuite (approved):

michaldudak:
The menu is not modal on mobile devices (tested on iPhone and iPad). Is this expected?

atomiks:
@michaldudak if you mean it doesn't lock scroll in all situations, that's from #1890

atomiks:
@ImSingee sorry for the super late reply but I see why this doesn't work (but should). Working to fix

michaldudak:
No, I mean you are able to click on elements outside of it. As if the internal backdrop doesn't exist.

atomiks:
@michaldudak ah right. I switched outside press dismissal from `mousedown` to `pointerdown` since there was a bug with the menu closing after long press if you held for 500ms but less than some X time, say 800ms, it would close immediately

atomiks:
Changed to a 500ms grace period to block outside press after long press, so we can still use `mousedown`
 
Fixes inner/outer nesting: https://codesandbox.io/p/sandbox/base-ui-context-menu-stack-forked-z2hclk

michaldudak:
~~The click handler now fires twice **sometimes**. It's pretty hard to reproduce. In this example I was click-and-dragging, but I also noticed it a couple of times when just clicking~~


https://github.com/user-attachments/assets/1720d921-9957-4bb7-9b71-77286b6da254


EDIT: Never mind, it's an issue with the CSB built-in console. The browser console doesn't show duplicates.

michaldudak (approved):

atomiks:
There's a bug with return focus causing a scroll to the previously focused element if it was out of the viewport, but will be updated in Floating UI before release

PR #1671, opened 2025-04-04
[core] `useRenderElement`

atomiks:
`useRenderElement` is a new and more DX friendly hook for internal usage:

```jsx
const renderElement = useRenderElement(
  tagName,
  { className, render }, 
  { state, props, ref }
);
```

Usage:

```jsx
const renderElement = useRenderElement('button', componentProps, {
  state, // now optional
  ref: forwardedRef, // or [forwardedRef, customRef]
  props: intrinsicProps, // or [defaultProps, intrinsicProps]
});
```

- `tagName`: All components require a fallback element to render, so a string as a first argument makes this cleaner. 
- `{ className, render }`: These are common to all components, and can passed in as the main `componentProps` object (props belonging to a component that renders a HTML element)
- `{ state, props, ref }`: These are specific to each component. `state` is now optional, `props` can be an object or an array which internally calls `mergeProps` for us, while `ref` performs ref merging by also accepting an array, preventing the need to use `useForkRef` internally 

`useComponentRenderer` still exists but as a middle migration function, and is now deprecated. This lets us migrate over several PRs to reduce conflicts and accidental changes


packages/react/src/use-render/useRender.ts
```diff
@@ -13,19 +11,21 @@ export function useRender<
   RenderedElementType extends Element,
 >(params: useRender.Parameters<State, RenderedElementType>) {
   const { render, props, state, refs } = params;
-  const { ref, ...extraProps } = props ?? {};
+  const { ref: intrinsicRefProp, ...intrinsicProps } = props || {};
 
   const refsArray = React.useMemo(() => {
-    return [...(refs ?? []), ref].filter(Boolean);
-  }, [refs, ref]) as React.Ref<RenderedElementType>[];
+    return [...(refs || []), intrinsicProps].filter(Boolean);
```
michaldudak, line 17: Shouldn't this array have `intrinsicRefProp` instead of `intrinsicProps`?

michaldudak (approved):

PR #1672, opened 2025-04-04
[docs] Remove duplicate outline class from tailwind demos

brijeshb42:
Something I found when going through the codebase. I am sure there are more such instances.


michaldudak (changes requested):
This is required for Tailwind 3. `outline` sets the style to `solid` and `outline-2` sets its width. See https://play.tailwindcss.com/APw65C5uTH

We want our demos to work well with both TW 3 and 4 to be easily copy-pastable into different projects.

brijeshb42:
Got it. I was wondering if it was deliberate.

PR #1676, opened 2025-04-07
[NumberField] Correct percentage parse handling

### LINKED ISSUE #1675: [NumberField] Bug with increment/decrement buttons for percent unit
**Description**: I am using NumberField from version 1.0.0-alpha.7 and encountered an issue with the increment/decrement buttons and input value behavior when the format is set to { style: "percent" } or { style: "unit", unit: "percent" }.

**Current Behavior:**

When I set format={{ style: "percent" }}, the increment and decrement buttons work correctly if I set step={0.01}. However, when typing a value directly into the input field, it is automatically multiplied by 100. For example, typing 50 results in 5000%, but I expect it to be 50% without multiplication.

**When I set format={{ style: "unit", unit: "percent" }}, the input field works fine (typing 50 results in 50% as expected). However, the increment and decrement buttons do not work properly. They do not respond as expected when trying to increment or decrement the value by 1.**

For other unit types like format={{ style: "unit", unit: "liter" }}, both the buttons and the input field behave as expected. The increment/decrement buttons work correctly, and typing values works without any unwanted multiplication.

**Expected Behavior:**

When style: "unit", unit: "percent" is used, the increment and decrement buttons should work correctly, allowing changes by increments of 1.

**Steps to Reproduce:**

Set the NumberField component with format={{ style: "percent" }}.

Try using the increment/decrement buttons (they work correctly).

Manually type a value like 50 into the input field and observe that it gets multiplied by 100, resulting in 5000%.

Set the NumberField component with format={{ style: "unit", unit: "percent" }}.

Observe that the input works correctly (typing 50 results in 50%), but the increment and decrement buttons do not work as expected.

Set the NumberField component with other unit types, like format={{ style: "unit", unit: "liter" }}, and observe that everything works as expected.

Link to Sandbox Example: [Example on CodeSandbox](https://codesandbox.io/p/sandbox/goofy-elion-zkf866?file=%2Fsrc%2FApp.tsx%3A12%2C7-12%2C73)

Browser Information:

Browser: Chrome

Operating System:

OS: Windows

Please let me know if you need any further details!

atomiks:
Fixes #1675

When `style: "percent"`, it's expected that the number is divided by 100 (0.1 = 10%, etc.)
When `style: "unit"`, then the percent is not divided, so 40 = 40%.

https://codesandbox.io/p/sandbox/base-ui-cra-ts-forked-7xzgdc?file=%2Fsrc%2FApp.tsx

### DIFF
--- a/packages/react/src/number-field/utils/parse.test.ts
+++ b/packages/react/src/number-field/utils/parse.test.ts
@@ -44,5 +44,13 @@ describe('NumberField parse', () => {
     it('returns null for an invalid number', () => {
       expect(parseNumber('invalid')).to.equal(null);
     });
+
+    it('handles percentages with style: "percent"', () => {
+      expect(parseNumber('12%', 'en-US', { style: 'percent' })).to.equal(0.12);
+    });
+
+    it('handles percentages with style: "unit" and unit: "percent"', () => {
+      expect(parseNumber('12%', 'en-US', { style: 'unit', unit: 'percent' })).to.equal(12);
+    });
   });
 });
--- a/packages/react/src/number-field/utils/parse.ts
+++ b/packages/react/src/number-field/utils/parse.ts
@@ -71,7 +71,11 @@ export function parseNumber(
 
   let num = parseFloat(unformattedNumber);
 
-  if (PERCENT_RE.test(formattedNumber)) {
+  const style = options?.style;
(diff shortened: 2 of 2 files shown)


mj12albert (approved):
Looks good ~ it works as expected now

PR #1682, opened 2025-04-08
[Tooltip] Add `disabled` prop

### LINKED ISSUE #1680: [tooltip] Support Conditional Rendering/Disabling of Tooltip Without Breaking Context
# Feature request
Hello 👋

I'm trying to conditionally show or hide tooltips in my app depending on screen size and sidebar state. Specifically, I want tooltips to appear only when my sidebar is collapsed on desktop

Here's what I tried initially:

```tsx
<Tooltip isVisible={isCollapsed && isDesktop}>
  <TooltipTrigger render={<NavItem ... />} />
  <TooltipContent>Label</TooltipContent>
</Tooltip>
```

To implement this, I added an `isVisible` prop to my wrapper around `TooltipPrimitive.Root`, and when `isVisible` is false, I skip rendering the `Tooltip` entirely. However, this causes the following error:

`Error: Base UI: TooltipRootContext is missing. Tooltip parts must be placed within <Tooltip.Root>`

Which makes sense but can Base UI add support for conditionally rendering tooltips (e.g. via visible, enabled, or shouldRender props) without breaking context, similar to how other primitive toolkits handle this?

This is how it is currently done in [React Aria Components](https://react-spectrum.adobe.com/react-spectrum/Tooltip.html#options):

```tsx
<TooltipTrigger isDisabled>
  <ActionButton
    aria-label="Danger Will Robinson"
    onPress={() => alert('pressed trigger')}
  >
    <Delete />
  </ActionButton>
  <Tooltip variant="negative" showIcon>Dangerous action.</Tooltip>
</TooltipTrigger>
```

atomiks:
Closes #1680


mj12albert (approved):
Looks good ~

PR #1683, opened 2025-04-08
[form controls] Add `inputRef` props

### LINKED ISSUE #1619: Supply an `inputProps` prop to access hidden `<input>` elements
`Checkbox` and `Switch` have an `inputRef` prop, but other form controls don't. Instead of just a ref, it's even better if any prop can be passed e.g. `onChange`

atomiks:
Closes #1619

- `RadioGroup` and `Radio`
- `Slider`
- `Select`


packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -41,6 +41,7 @@ const SliderRoot = React.forwardRef(function SliderRoot<Value extends number | r
     step = 1,
     tabIndex: externalTabIndex,
     value,
+    inputRef,
```
mj12albert, line 44: For slider, I think we need to put this on the Thumb to support range sliders

atomiks: It gets forwarded to the `<input>` in Thumb via context

mj12albert: I meant in a range slider each thumb has it's own input element, so they can't all take the same ref

atomiks: Actually had it on `Slider.Thumb` already, just removed it from the Root

mj12albert (approved):

Issue #1466, opened 2025-02-20
[Dialog] Focus gets lost under certain conditions with nested dialogs


atomiks:
# Bug report

## Current behavior

1. [Nested dialogs](https://base-ui.com/react/components/dialog#nested-dialogs): open nested dialog and press outside with a pointer. Focus is ignored and not returned to its trigger, so subsequent tabs aren't trapped unless pressing inside the dialog with the pointer
2. [Close confirmation](https://base-ui.com/react/components/dialog#close-confirmation): type a char in the textarea and press Cancel. Press the <kbd>esc</kbd> key 3 times quickly. Focus gets returned to the root trigger, not the Cancel trigger

## Expected behavior

1. Focus should return to the trigger with `preventScroll: true` if it's available for outside presses instead of being ignored
2. Focus shouldn't return to the root and should stay trapped

## Reproducible example

Link to a CodeSandbox, StackBlitz, or other IDE.

## Base UI version

1.0.0-alpha.6

## Which browser are you using?

Chrome/Safari/Firefox/Safari iOS etc.

## Which OS are you using?

Mac OS/Windows/Other etc.

## Which assistive tech are you using (if applicable)?

Voiceover/JAWS etc.

## Additional context

Provide any additional context that might help us identify the problem and find a solution.

atomiks:
Fixed in `@floating-ui/react@0.27.5`

- https://github.com/floating-ui/floating-ui/pull/3235
- https://github.com/floating-ui/floating-ui/pull/3237

PR #1684, opened 2025-04-08
[menubar] Create the Menubar component

### LINKED ISSUE #1407: [menu] Implement Menubar

michaldudak:
Preview: https://deploy-preview-1684--base-ui.netlify.app/react/components/menubar

Experiment: https://deploy-preview-1684--base-ui.netlify.app/experiments/menubar

Closes #1407

### To do
- [x] modality
- [x] tests
- [x] docs and demos
- [x] vertical orientation
- [x] implementation using `CompositeRoot`
- [x] `data-instant`
- [x] Use the released version of Floating UI


packages/react/src/menubar/root/useMenubarRoot.ts
```diff
@@ -0,0 +1,107 @@
...
+    nested: false,
+    loop,
+    orientation,
+    rtl: direction === 'rtl',
+    onNavigate: (index) => {
+      if (index !== null) {
+        setActiveIndex(index);
+      }
+    },
+  });
```
atomiks, line 44: I notice this has a lot of bugs, wouldn't it be better just to use `Composite`? This is missing the hover functionality if a menu is already open, but it avoids many issues by default

https://github.com/floating-ui/floating-ui/blob/master/packages/react/test/visual/components/Menubar.tsx

michaldudak: I'll see if it's any better, thanks.

packages/react/src/utils/InternalBackdrop.tsx
```diff
@@ -1,29 +1,59 @@
...
+      100% 100%,
+      0% 100%,
+      0% 0%,
+      ${rect.left}px ${rect.top}px,
+      ${rect.left}px ${rect.bottom}px,
+      ${rect.right}px ${rect.bottom}px,
+      ${rect.right}px ${rect.top}px,
+      ${rect.left}px ${rect.top}px
+    )`;
+  }
```
atomiks, line 28: This is cool! I remember we discussed modality using backdrops while excluding certain elements, but can this support multiple "holes"?

michaldudak: This implementation assumes just one "hole" but it can easily be expanded to support more. It may be more tricky if elements overlap (due to clip-path fill rules), but doable with some maths.

colmtuite:
A few visual things:

1. Total container height should be 38px computed. Check ToggleGroup and Toolbar for reference. Best accomplished by a fixed height on the triggers.
2. Horizontal trigger padding should be 12px computed.
3. Drop `cursor: pointer` from the triggers.
4. Drop the `:hover` styles from the triggers.
5. We can probably drop all of the animations and transitions. It feels much snappier without them, and menubars typically don't have animations.

michaldudak:
@colmtuite I ended up leaving transitions when the first menu appears (and the last one disappears), just like in MacOS menubars.

atomiks:
1. There should be an offset between the menus and the menubar, right? In macOS, there's a gap:

Screenshot 2025-04-16 at 11 45 51 pm

2. Submenus don't react to their parents closing/don't fade out with them

https://github.com/user-attachments/assets/71756ff6-8072-48fb-80d1-cddfad0351d5

docs/src/nav.ts
```diff
@@ -88,6 +88,10 @@ export const nav = [
         label: 'Menu',
         href: '/react/components/menu',
       },
+      {
+        label: 'Menubar',
```
flaviendelangle, line 92: Nitpick: What is that rational for `Menubar` vs `Menu Bar` here?
To me the component name is similar to `Preview Card` more than `Fieldset` (which is the only component with 2 words that we don't split in the navbar AFAIK).

packages/react/src/composite/root/useCompositeRoot.ts
```diff
@@ -152,7 +157,10 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
...
             }
 
-            if ((event.target as HTMLElement).closest('[data-floating-ui-portal]') != null) {
+            if (
+              ignoreNavigationFromPopups &&
+              (event.target as HTMLElement).closest('[data-floating-ui-portal]') != null
+            ) {
               // don't navigate if the event came from a popup
               return;
             }
```
mj12albert, line 166: Discussed with @atomiks about this and decided we could drop it completely from composite (it was kind of a hack anyway), and instead use `event.stopPropagation` in other components onKeyDown instead (e.g. SliderThumb)

I tested this a bit more and currently it doesn't work for some setups like Toolbar > Popover > RadioGroup

michaldudak:
> There should be an offset between the menus and the menubar, right? In macOS, there's a gap

Only on the screen with a notch. On an external monitor the menu opens just below the trigger:
Screenshot 2025-04-16 at 20 18 01

But anyway, it shouldn't overlap the trigger as it does now.

> Submenus don't react to their parents closing/don't fade out with them

That's likely a bug with data-instant. I'll figure it out.

> Nitpick: What is that rational for Menubar vs Menu Bar here?

I was wondering the same at the beginning but chose to follow what's in https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menubar_role (although they are not 100% consistent). Also, we have a Toolbar, not ToolBar.

docs/src/nav.ts
colmtuite: Mostly just adhering to ARIA. They do "Menubar", "Toolbar", "Alert Dialog", "Fieldset".

I think the rationale is that "Bar" is not a thing on its own. Whereas "Dialog" is a thing.

Preview Card is not an ARIA pattern, we just made it up. But "Card" would be a thing on its own, so it follows the same logic.

colmtuite:
@michaldudak It seems like there's some sort of issue with arrow key behaviour. In this video, I'm clicking a trigger each time, then pressing the right arrow key. Sometimes I leave the pointer over the trigger, and sometimes I move the pointer off the trigger before pressing the right arrow key.

Sometimes, pressing the right arrow actually opens the menu to the left of the trigger I opened originally. When I move the pointer off the trigger before pressing the right arrow key, the next menu doesn't open, the trigger of the currently open menu just becomes focused.

Separately—and not shown in the video—when you click to open a menu, then leave the pointer where it is and press right to open the next menu, then move the pointer slightly, it correctly opens the previous menu without playing any animations. However, if you follow the same steps but instead click (rather than moving the pointer slightly), the enter animation plays.

https://github.com/user-attachments/assets/3b127804-0f28-4a61-b422-6c5024cb8df4

michaldudak:
Good catch. I can reproduce it locally, so I should be able to fix it.

docs/src/nav.ts
flaviendelangle: OK, the naming does not feel super natural to me, but adhering to ARIA is a good reason to stick with it

packages/react/src/menu/root/MenuRoot.tsx
michaldudak, line 1: I moved all the MenuRoot logic to its hook so it's more centralized.

packages/react/src/menubar/Menubar.tsx
```diff
@@ -63,7 +63,7 @@ const Menubar = React.forwardRef(function Menubar(
 
   const renderElement = useRenderElement('div', props, {
     state,
-    props: [otherProps, { onFocus: handleFocus, onBlur: handleBlur }],
+    props: [otherProps, { onFocus: handleFocus, onBlur: handleBlur, role: 'menubar' }],
```
atomiks, line 66: This order is incorrect (with the latest `mergeProps`), the internal props should come first

packages/react/src/menu/popup/MenuPopup.tsx
```diff
@@ -98,7 +98,8 @@ const MenuPopup = React.forwardRef(function MenuPopup(
       context={floatingContext}
       modal={false}
       disabled={!mounted}
-      initialFocus={nested ? -1 : 0}
+      returnFocus={parent.type === undefined}
+      initialFocus={parent.type === 'menu' ? -1 : 0}
```
atomiks, line 102: Playing with this, it doesn't seem to make sense not to return focus in a web context. Radix and Ariakit return focus to the trigger. In the OS, menubar has different accessibility semantics most likely regarding where focus is placed, but inside a browser the focus should be restored, otherwise it just restarts at the beginning of the document and loses the context

atomiks:
Bug: 

1. Tab into menubar
2. Hover over any item
3. Should not open

michaldudak:
> Bug:

> Tab into menubar
> Hover over any item
> Should not open

This is by design to follow the pattern from MacOS menubar.

atomiks:
Hmm, it feels off to me. It's another thing Radix & Ariakit do differently to this, possibly making it different in a web context

cc: @colmtuite

atomiks:
The Tailwind CSS demo has no transitions. I think it'd be best if the transitions matched macOS: no fade in transition, but a fade out transition (no scaling, though)

atomiks:
On macOS, the open menu hides itself if you're still within the menubar bounds and not hovering over any trigger, but not if you go below the menubar in the "popups region" 🤔

michaldudak:
> The Tailwind CSS demo has no transitions. I think it'd be best if the transitions matched macOS: no fade in transition, but a fade out transition (no scaling, though)

:+1: I updated the demos.

> On macOS, the open menu hides itself if you're still within the menubar bounds and not hovering over any trigger, but not if you go below the menubar in the "popups region" 🤔

Right. I don't think it's strictly necessary to replicate this, though.
Let's decide what to do with returning focus and hover behavior first, then we can consider what if this is worth implementing (@colmtuite, I'd love to hear your feedback).

flaviendelangle:
https://github.com/user-attachments/assets/9f9e4b25-e648-45c5-a4bb-5cb395bf197f

On Webstorm (Ubuntu), when I'm focusing an element in the menu and press "Escape", the focus moves back to the menu trigger.

With the Base UI component, is it something we want to have?
I saw that Arrow Left / Arrow Right allowed to move to the next / previous menu, but as a user my first reflex was to press "Escape" first which was a bad idea :laughing:

atomiks:
## Bug with `data-instant`

1. Tab to first menu trigger
2. <kbd>ArrowDown</kbd> (doesn't happen with <kbd>Enter</kbd>)
3. <kbd>ArrowRight</kbd>
4. The first menu fades out but shouldn't transition

## Bug with `data-highlighted`

1. Tab to first menu trigger
2. <kbd>Enter</kbd> (doesn't happen with <kbd>ArrowDown</kbd>)
3. <kbd>ArrowRight</kbd>
4. <kbd>ArrowLeft</kbd>
5. `data-highlighted` is on the first menu item but shouldn't be

atomiks:
I think the `:focus` style should be `:focus-visible`. I wouldn't expect it to be highlighted after pressing outside to dismiss:

https://github.com/user-attachments/assets/e711cde2-d316-4dd9-b451-aaba3fc25d34

michaldudak:
> I think the :focus style should be :focus-visible

`focus-visible` has the same effect (as the focus is placed programmatically). It's consistent with other modal components such as Menu and Select, though. We might need to update this behavior across the board.

atomiks:
> focus-visible has the same effect (as the focus is placed programmatically).

Programmatic focus depends on what the user modality was and respects `:focus-visible` and it appears less frequently (though #1726 is still needed to fully fix it). 

At the very least, it shouldn't flicker when closing with an exit animation (the style gets removed and then added back after it unmounts - this might relate to #1761 where we need a `data-popup-mounted` hook

Another couple bugs:

1.  It's possible for two highlights to be present:

https://github.com/user-attachments/assets/2328452e-8f7c-418f-84d7-3ccb8634a712

2. Sometimes the exit animation doesn't play at all when clicking outside

michaldudak:
What do you think about not restoring focus to the trigger at all on outside press? Radix Menubar works like that.

atomiks:
> What do you think about not restoring focus to the trigger at all on outside press? Radix Menubar works like that.

This seems okay, the main issue is modal dialogs must return focus on outside press to allow tabbing to work in a nested context: https://github.com/mui/base-ui/issues/1466

michaldudak:
Yeah, this would apply just to the Menubar.

packages/react/src/menubar/Menubar.tsx
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Whether to loop keyboard focus back to the first item
+     * when the end of the list is reached while using the arrow keys.
+     * @default true
+     */
+    loop?: boolean;
+  }
+}
+
+export { Menubar };
```
atomiks, line 156: Should use new export pattern

packages/react/src/menubar/Menubar.tsx
```diff
@@ -0,0 +1,156 @@
...
+    </MenubarContext.Provider>
+  );
+});
+
+function MenubarContent(props: React.PropsWithChildren<{}>) {
+  const nodeId = useFloatingNodeId();
+  const { events: menuEvents } = useFloatingTree()!;
+  const openSubmenus = React.useRef(new Set<string>());
+  const rootContext = useMenubarContext();
+  React.useEffect(() => {
```
atomiks, line 93:
```suggestion
  const nodeId = useFloatingNodeId();
  const { events: menuEvents } = useFloatingTree()!;
  const openSubmenusRef = React.useRef(new Set<string>());
  const rootContext = useMenubarContext();

  React.useEffect(() => {
```

atomiks: Add `Ref` suffix

packages/react/src/menubar/Menubar.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const contentRef = React.useRef<HTMLDivElement>(null);
+
+  const renderElement = useRenderElement('div', props, {
+    state,
+    props: [{ role: 'menubar' }, otherProps],
+    ref: [forwardedRef, setContentElement, contentRef],
+  });
+
+  // TODO: use hasSubmenuOpen instead
+  const shouldOpenOnHover = hasSubmenuOpen;
```
atomiks, line 57: Is this still relevant?

atomiks:
Sometimes it's possible to hover a new menu item where the click-to-dimsiss behavior doesn't work on the first click

https://github.com/user-attachments/assets/5bf4bee5-f302-49fd-8c2c-41ccc7d9e423

atomiks:
> Sometimes it's possible to hover a new menu item where the click-to-dimsiss behavior doesn't work on the first click

This now works (clicking menu trigger, hovering to a new one, then clicking the new trigger to dismiss), but the exit animation doesn't play, though it does if you press outside

atomiks:
In the demo, the submenu `alignOffset` should subtract the padding so they line up when highlighted:

Screenshot 2025-05-05 at 9 42 00 pm

---

I also don't think it should have `:active` styles since they don't work on open until we polyfill it.

---

Sometimes it's possible for `[data-instant]` not to be applied when it should be when mixing keyboard+pointer modalities

https://github.com/user-attachments/assets/8a838ab0-cffa-48cb-8fdc-40892c9737b4

atomiks:
If you mousedown, move and mouseup (or off) in a different menu, it doesn't close or select the item

https://github.com/user-attachments/assets/4c80af92-9b1f-4d1a-9299-59cdc43423a1

packages/react/src/menu/root/useMenuRoot.ts
```diff
@@ -134,22 +159,47 @@ export function useMenuRoot(parameters: useMenuRoot.Parameters): useMenuRoot.Ret
...
       function changeState() {
-        setOpen(openValue, eventValue, translateOpenChangeReason(reasonValue));
+        onOpenChange?.(nextOpen, event, reason);
+        setOpenUnwrapped(nextOpen);
+
+        setLastOpenChangeReason(reason ?? null);
       }
 
-      if (isHover) {
+      if (reason === 'hover') {
```
atomiks, line 202: Hover (leave) events can also generate `'safe-polygon'`

michaldudak: We're translating Floating UI's `safe-polygon` to `hover`, so this should be covered.

docs/src/app/(public)/(content)/react/components/menubar/demos/hero/tailwind/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+        </Menu.Trigger>
+      </Menu.Root>
+    </Menubar>
+  );
+}
+
+function handleClick(event: React.MouseEvent) {
+  // eslint-disable-next-line no-console
+  console.log(`${(event.currentTarget as HTMLElement).innerText} clicked`);
+}
```
atomiks, line 182: Type the `MouseEvent` with `<HTMLElement> and use `textContent` (`innerText` is not the standard property)

atomiks:
Doesn't happen on iPhone, but Chrome DevTools has that default "300ms delay" (when the meta tag is usually not set) when tapping and when that occurs on touch it breaks opening the next menu when one is open.

https://github.com/user-attachments/assets/c15be237-b58a-4a9a-8078-7a4b5e0aeb1b

michaldudak:
I wonder why it happens - we have the meta tag on the docs. Anyway, this should work now.

packages/react/src/menu/root/useMenuRoot.ts
```diff
@@ -174,6 +175,25 @@ export function useMenuRoot(parameters: useMenuRoot.Parameters): useMenuRoot.Ret
...
+
+      // Prevent the menu from closing on mobile devices that have a delayed click event.
+      // In some cases the menu, when tapped, will fire the focus event first and then the click event.
+      // Without this guard, the menu will close immediately after opening.
+      if (nextOpen && reason === 'focus') {
+        allowTouchToCloseRef.current = false;
+        setTimeout(() => {
+          allowTouchToCloseRef.current = true;
+        }, 300);
+      }
```
atomiks, line 195: I think this timeout should have cleanup, at least right before it's called again

atomiks (approved):
Found one minor issue with the Tailwind CSS demo (icon isn't white/visible when highlighted), but the implementation looks good and I wasn't able to spot any other bugs. I think you can revert `@floating-ui/react` to the npm version since a release will be done to fix the issue before Base UI releases. 

Screenshot 2025-05-08 at 9 51 14 pm

colmtuite (approved):

packages/react/src/utils/floating-ui/useClick.ts
```diff
@@ -112,7 +112,8 @@ export function useClick(context: FloatingRootContext, props: UseClickProps = {}
           open &&
           toggle &&
           (dataRef.current.openEvent && stickIfOpen
-            ? dataRef.current.openEvent.type === 'click'
+            ? dataRef.current.openEvent.type === 'click' ||
```
michaldudak, line 115: @atomiks, I just added this to fix an issue with tests (https://app.circleci.com/pipelines/github/mui/base-ui/9809/workflows/65a23a64-619c-4531-897d-d40332476f6b/jobs/94635).

atomiks:
Probably some merge issue: clicking now causes the menu to immediately close again

packages/react/src/utils/floating-ui/useClick.ts
michaldudak: No, it's unrelated. I'm figuring it out.

packages/react/src/menu/positioner/MenuPositioner.test.tsx
```diff
@@ -315,29 +322,31 @@ describe('<Menu.Positioner />', () => {
...
+      await user.click(trigger, { delay: 20 });
       await waitFor(() => {
-        expect(queryByRole('menu', { hidden: false })).not.to.equal(null);
+        expect(screen.queryByRole('menu', { hidden: false })).not.to.equal(null);
       });
-      expect(queryByRole('menu', { hidden: false })).not.toBeInaccessible();
+      expect(screen.queryByRole('menu', { hidden: false })).not.toBeInaccessible();
 
-      await user.click(trigger);
+      await user.click(trigger, { delay: 20 });
```
michaldudak, line 336: This is an interesting case. I spend quite some time debugging why everything works perfectly fine and yet tests fail when clicking on the trigger twice. It turns out that test tools (both RTL and Playwright) generate mouse events so quickly one after the other (mousedown + click), that our code doesn't have a chance to run in between. In real-life usage, the handler for `mousedown` would always fire before the `click` event, setting the `ignoreClickRef` value (and preventing the menu from closing right after it was opened). In tests, `click` was fired before the ref was set and the menu closed immediately.

PR #1701, opened 2025-04-11
[Tooltip][Popover] Refactor to `useRenderElement`

atomiks:


packages/react/src/tooltip/arrow/TooltipArrow.tsx
```diff
@@ -16,27 +15,14 @@ import { mergeProps } from '../../merge-props';
  * Documentation: [Base UI Tooltip](https://base-ui.com/react/components/tooltip)
  */
 const TooltipArrow = React.forwardRef(function TooltipArrow(
-  props: TooltipArrow.Props,
+  componentProps: TooltipArrow.Props,
   forwardedRef: React.ForwardedRef<HTMLDivElement>,
 ) {
-  const { className, render, ...otherProps } = props;
+  const { className, render, ...intrinsicProps } = componentProps;
```
atomiks, line 21: `elementProps` may be better than `intrinsicProps` here, which matches `useRender.ComponentProps` vs `useRender.ElementProps`. React itself treats native tags as `ComponentProps` but it's a bit ambiguous/misleading in this scenario

mj12albert: The reasoning is sound but I will miss the simplicity of `props` and `otherProps` 😅

mj12albert (approved):

PR #1702, opened 2025-04-11
[Backdrops] Fix text selection & right-clicks

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1705

- Adding `WebkitUserSelect: 'none'` and `userSelect: 'none'` to the inlined styles prevents text selection on touch
- Only main-button clicks on dialog backdrop should close it


mj12albert:
@atomiks I've only updated the dismiss behavior for Dialog so it doesn't close on right click (matches Radix)

Will leave the rest alone in this PR given the larger outside press issue

atomiks (commented):
Do you need regular `userSelect` for Android?

Also wonder if `pointerEvents: undefined` (instead of conditional rendering for `openReason === 'hover'` could be a problem

mj12albert:
> Also wonder if `pointerEvents: undefined` (instead of conditional rendering for `openReason === 'hover'` could be a problem

I tested Popover with a mouse and a phone and it seems ok, but `openReason === 'hover' ? 'none' : 'auto'` could work too, what do you think? @atomiks

atomiks:
`undefined` should be ok

PR #1704, opened 2025-04-11
[Collapsible][Accordion] Resize panel on window resize

### LINKED ISSUE #1458: [accordion][collapsible] With CSS transitions the panel does not resize when content size changes

atomiks:
Closes #1458

### DIFF
--- a/packages/react/src/accordion/panel/AccordionPanel.tsx
+++ b/packages/react/src/accordion/panel/AccordionPanel.tsx
@@ -13,6 +13,7 @@ import type { AccordionItem } from '../item/AccordionItem';
 import { useAccordionItemContext } from '../item/AccordionItemContext';
 import { accordionStyleHookMapping } from '../item/styleHooks';
 import { AccordionPanelCssVars } from './AccordionPanelCssVars';
+import { ownerWindow } from '../../utils/owner';
 
 /**
  * A collapsible panel with the accordion item contents.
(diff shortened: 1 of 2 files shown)


packages/react/src/utils/usePanelResize.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    // Ensures the panel will expand to the correct height when the window is resized.
+    // This prevents content from being cut off or the panel not fitting to the content.
+    function handleWindowResize() {
+      if (panel) {
+        const originalHeight = panel.style.height;
+        const originalWidth = panel.style.width;
+        panel.style.height = 'auto';
+        panel.style.width = 'auto';
+        setDimensions({ height: panel.scrollHeight, width: panel.scrollWidth });
```
atomiks, line 24: The idea of removing the CSS variable when the transition completes (so it uses implicit `auto`) may be a better idea, depending if we want the CSS variable present with the values all the time or not

packages/react/src/utils/usePanelResize.ts
```diff
@@ -0,0 +1,70 @@
...
+      // Avoid size transitions when the window is resized.
+      if (panel) {
+        cancelAnimationFrame(frame);
+        const cleanup = setAutoSize(panel);
+        frame = requestAnimationFrame(cleanup);
+      }
+    }
+
+    const win = ownerWindow(panel);
+    win.addEventListener('resize', handleWindowResize);
```
mj12albert, line 62: Testing this a bit and it works well, just wondering is it not enough to only use ResizeObserver?

atomiks: In this case the ResizeObserver fires when transitioning the height, which is why I added `if (panel.getAnimations().length > 0) {` to prevent its logic from running, but when resizing the window, I don't want any animations to occur so the resizing is 1:1 with layout.

For this purpose we can't reuse the ResizeObserver since we want regular transitions to work when opening/closing

mj12albert (approved):

PR #1713, opened 2025-04-14
[select] Move item anchoring prop to `Positioner`

### LINKED ISSUE #1710: [docs][Select] Update `Positioner` and `anchor` prop descriptions
# Bug report

`Select` inherits the type description that's common to all `Positioner`s, but it acts special by default due to `alignItemToTrigger=true`.

> Positioner
Positions the select menu popup against the trigger. Renders a `<div>` element.

Image

### LINKED ISSUE #1711: [Select] Move `alignItemToTrigger` from `Root` to `Positioner`
# Bug report

The previous implementation required the prop to be on the `Root`, but this is no longer the case with the latest one, as it can exist on `Positioner` instead which makes more sense in the API.

@DiegoAndai @aarongarciah you mentioned possibly changing the name, which we could also do simultaneously here.

An actual proper description of the behavior is more like:

`<Select.Positioner alignSoThatSelectedItemTextIsOverTriggerValueText>`

Which is obviously unrealistically verbose, but is more accurate as the prop aligns the `Positioner` element, not the `Item` element, so the selected item text inside of it _visually appears_ directly aligned with the `Trigger` `Value` text.  

It also disables the other anchoring-related props and enters a "special positioning mode"

```jsx
<>
  <Select.Positioner align="selected" />
  <Select.Positioner overlayTriggerValue />
  <Select.Positioner alignToTrigger />
  <Select.Positioner anchorMode="selected-item|trigger" />
  <Select.Positioner positioningMode="selected-item|trigger" />
</>
```

atomiks:
1. Removes `alignItemToTrigger` prop from `Root`, to be configured on `Positioner` instead
2.  Changes the boolean prop name to `alignItemWithTrigger` (`To` -> `With`)

Closes #1710
Closes #1711


packages/react/src/select/item/useSelectItem.ts
```diff
@@ -152,7 +152,7 @@ export function useSelectItem(params: useSelectItem.Parameters): useSelectItem.R
                 cursorMovementTimerRef.current = -1;
               }
 
-              // With `alignItemToTrigger`, avoid re-rendering the root due to `onMouseLeave`
+              // With `anchor="item"`, avoid re-rendering the root due to `onMouseLeave`
```
michaldudak, line 155:
```suggestion
              // With `anchorMode="overlap-trigger"`, avoid re-rendering the root due to `onMouseLeave`
```

packages/react/src/select/root/useSelectRoot.ts
```diff
@@ -243,7 +225,7 @@ export function useSelectRoot<T>(params: useSelectRoot.Parameters<T>): useSelect
       setActiveIndex(nextActiveIndex);
     },
     // Implement our own listeners since `onPointerLeave` on each option fires while scrolling with
-    // the `alignItemToTrigger` prop enabled, causing a performance issue on Chrome.
+    // the `anchor="item"`, causing a performance issue on Chrome.
```
michaldudak, line 228:
```suggestion
    // the `anchorMode="overlap-trigger"`, causing a performance issue on Chrome.
```

michaldudak (approved):

PR #1715, opened 2025-04-14
[select] fix: update select value children type

### LINKED ISSUE #1558: [Select] Value component children documented incorrectly
On the Select docs, the [Value component's children](https://base-ui.com/react/components/select#value) is documented as such:

Image

But that doesn't match the type:

https://github.com/mui/base-ui/blob/376df1d27a9fdff150bad474ca5627bd8dd8f426/packages/react/src/select/value/SelectValue.tsx#L53

I think the type is incorrect.

joaom00:

Closes #1558


docs/reference/overrides/select-value.json
```diff
@@ -2,7 +2,7 @@
   "name": "SelectValue",
   "props": {
     "children": {
-      "type": "React.ReactNode | (value) => React.ReactNode",
+      "type": "null | (value) => React.ReactNode",
       "description": "Override children if you need to customize how the value is displayed.\nBy default, renders the current value as is.\n\nAccepts a `ReactNode` or a function that returns the node to render."
```
atomiks, line 6: This override actually needed to be updated in several ways:

- `label` was always first, not `value`
- In new Base UI versions, `value`  is available now but comes second

This is the current source of truth:

```jsx
children?: null | ((label: string, value: any) => React.ReactNode);
```

joaom00: What would be a label in this case? the placeholder?

atomiks: `label` is the raw `textContent` of `Select.Item` as presented to the user (or `label` prop on `Select.Item`), while `value` is the unique identifier

joaom00: makes sense!

atomiks:
Looks like https://github.com/mui/base-ui/pull/1537 already ended up fixing this, this branch is now sync'd with with main already. Going to close, but thank you for the contribution!

Issue #1731, opened 2025-04-16
[Slider] Thumbs that overlap at `max` cannot be dragged


mj12albert:
From https://github.com/mui/base-ui/pull/1600#issuecomment-2807946605

https://github.com/user-attachments/assets/f4f49a88-2f9c-4e7c-b36b-c75b94c35f6b

PR #1732, opened 2025-04-16
[Slider] Fix overlapping slider thumbs stuck at min or max

mj12albert:
Fixes https://github.com/mui/base-ui/issues/1731

Demo: https://codesandbox.io/p/sandbox/fast-glitter-7tck3z

When 2 or more thumbs overlap in the same position they can be dragged apart now

### DIFF
--- a/packages/react/src/slider/root/useSliderRoot.ts
+++ b/packages/react/src/slider/root/useSliderRoot.ts
@@ -37,15 +37,17 @@ function areValuesEqual(
   return false;
 }
 
-function getClosestThumbIndex(values: readonly number[], currentValue: number, min: number) {
+function getClosestThumbIndex(values: readonly number[], currentValue: number, max: number) {
   let closestIndex;
   let minDistance;
   for (let i = 0; i < values.length; i += 1) {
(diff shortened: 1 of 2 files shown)


atomiks:
I think whichever one was used last should be the one that's on top? e.g. I'd expect max to be on top here but it flips to the min

https://github.com/user-attachments/assets/27ee0ca8-54da-4cc0-9ab6-6981d22e6a65

mj12albert:
> I think whichever one was used last should be the one that's on top?

Actually assuming all the thumbs have equal z-index, it seems better to consider the right-most thumb (closest to `max`) to be "closest" based on the natural stacking order. The exception is when they overlap at `max`, drags need to move the left-most thumb first, or it will block other thumbs from moving, what do you think @atomiks ?

https://github.com/user-attachments/assets/3ce14ee9-6eb8-4984-9893-1bba7c5003bb

atomiks (approved):
This should also work fine with RTL right?

mj12albert:
> This should also work fine with RTL right?

Yep ~ here it's only comparing values vs each other and min/max so direction and orientation don't matter

PR #1735, opened 2025-04-17
[dialog] Allow document to slide input into view on iOS when keyboard opens

### LINKED ISSUE #1455: [dialog] Dialogs should adjust position when mobile keyboard is open
# Feature request

## Summary

Dialogs on mobile could be more responsive, particularly in handling input fields. This is not limited to dialogs but applies broadly to modals. The document should adjust its position dynamically to keep the input field visible when the keyboard is open.

Additionally, there's a noticeable jumping effect when the keyboard appears, which affects the user experience. This behavior is demonstrated in the attached video.

I’ve included three examples from Base UI, 9ui (my project built on top of Base UI), and shadcn/ui to illustrate how other libraries handle this.

https://github.com/user-attachments/assets/95f7dbfd-f401-468f-8d05-17a7c448213e (Base UI)
https://github.com/user-attachments/assets/68efb3b7-431e-4758-82de-2b85c6cc236b (9ui)
https://github.com/user-attachments/assets/c761bbb1-22c0-43df-ac2b-c58c1c538958 (shadcn/ui)

## Motivation

Enhancing the user experience on mobile devices. This feature would primarily benefit mobile users and developers building mobile-first applications, ensuring smoother interactions when dealing with modals and input fields.

atomiks:
Fixes #1455

The way inputs slide inside Popover/anchored elements is a lot stranger so this is just restricted to Dialog, for now

Test on iOS: https://deploy-preview-1735--base-ui.netlify.app/experiments/mobile-scroll-lock

### DIFF
--- a/packages/react/src/popover/root/usePopoverRoot.ts
+++ b/packages/react/src/popover/root/usePopoverRoot.ts
@@ -26,7 +26,6 @@ import {
 import { useOpenChangeComplete } from '../../utils/useOpenChangeComplete';
 import { PATIENT_CLICK_THRESHOLD } from '../../utils/constants';
 import { useScrollLock } from '../../utils/useScrollLock';
-import { useIOSKeyboardSlideFix } from '../../utils/useIOSKeyboardSlideFix';
 
 export function usePopoverRoot(params: usePopoverRoot.Parameters): usePopoverRoot.ReturnValue {
   const {
@@ -50,7 +49,6 @@ export function usePopoverRoot(params: usePopoverRoot.Parameters): usePopoverRoo
   const [positionerElement, setPositionerElement] = React.useState<HTMLElement | null>(null);
   const [openReason, setOpenReason] = React.useState<OpenChangeReason | null>(null);
   const [stickIfOpen, setStickIfOpen] = React.useState(true);
-  const [allowIOSLock, setAllowIOSLock] = React.useState(true);
 
   const popupRef = React.useRef<HTMLElement>(null);
   const stickIfOpenTimeoutRef = React.useRef(-1);
@@ -170,28 +168,11 @@ export function usePopoverRoot(params: usePopoverRoot.Parameters): usePopoverRoo
   const dismiss = useDismiss(context);
   const role = useRole(context);
 
-  const enableScrollLock = open && modal === true;
-  const enableScrollLockIOS = enableScrollLock && allowIOSLock;
-
-  const iOSDocumentSlide = useIOSKeyboardSlideFix({
-    enabled: enableScrollLock,
-    setLock: setAllowIOSLock,
-    popupRef,
-  });
-
-  useScrollLock({
-    enabled: enableScrollLockIOS,
-    mounted,
-    open,
-    referenceElement: positionerElement,
-  });
-
   const { getReferenceProps, getFloatingProps: getPopupProps } = useInteractions([
     hover,
     click,
     dismiss,
     role,
-    iOSDocumentSlide,
   ]);
 
   const getTriggerProps = React.useCallback(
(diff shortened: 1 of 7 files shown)


packages/react/src/popover/root/usePopoverRoot.ts
```diff
@@ -164,20 +166,32 @@ export function usePopoverRoot(params: usePopoverRoot.Parameters): usePopoverRoo
...
+  const role = useRole(context);
 
-  const click = useClick(context, {
-    stickIfOpen,
-  });
+  const enableScrollLock = open && modal === true;
+  const enableScrollLockIOS = enableScrollLock && allowIOSLock;
 
-  const dismiss = useDismiss(context);
+  const iOSDocumentSlide = useIOSKeyboardSlideFix({
```
mj12albert, line 176: Isn't this only for Dialog for now?

atomiks: Yeah, committed the work I was trying mistakenly - removed

michaldudak:
It only works on the first interaction. After I dismiss the keyboard and focus the textarea again, the dialog doesn't shift.

atomiks:
@michaldudak it works for me every time

michaldudak:
Here's how it looks on Edge on iPhone 16 Pro (iOS 18.3.2):

https://github.com/user-attachments/assets/b1784448-4c16-4522-829e-221bafb84b52

It's the same on Chrome, but it works well on Safari.

mj12albert:
> It's the same on Chrome, but it works well on Safari.

I was about to record the same thing on Chrome + iOS, after the textarea is blurred and the keyboard goes away, "View notifications" is no longer visible just underneath the popup

atomiks:
> I was about to record the same thing on Chrome + iOS, after the textarea is blurred and the keyboard goes away, "View notifications" is no longer visible just underneath the popup

That's this issue

>  Scroll needs to be manually restored since the sliding causes the page to scroll in the background.

I think the tradeoff is acceptable in this case; the scroll position is restored manually so it's in the same position as when it was opened

packages/react/src/utils/useIOSKeyboardSlideFix.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  const floating: ElementProps['floating'] = React.useMemo(
+    () => ({
+      onFocus: handlePress,
+      onClick: handlePress,
+    }),
+    [handlePress],
+  );
+
+  return React.useMemo(() => ({ floating }), [floating]);
```
michaldudak, line 77: Let's merge these two `useMemo`s

atomiks: Each one needs to be memo'd separately

atomiks:
> I was about to record the same thing on Chrome + iOS, after the textarea is blurred and the keyboard goes away, "View notifications" is no longer visible just underneath the popup

This restoration issue actually happens with non-modal dialogs already since it slides the background a bit each time the textarea gets focused. 

In Chrome iOS it doesn't restore it properly in this PR, but Safari/Edge work correctly. We can either match non-modal dialogs (accept that the background slides each time) or restore it using the existing logic which is not foolproof

packages/react/src/utils/useIOSKeyboardSlideFix.ts
michaldudak: Why can't we make it
```
return React.useMemo(
  () => ({
    floating: {
      onFocus: handlePress,
      onClick: handlePress,
    },
  }),
  [handlePress],
);
```

atomiks: `useInteractions` uses `.floating` property as a dependency in `useCallback`

michaldudak: Does it matter? The value returned from this hook is the same regardless of how many memos are used, no? It's just a matter of optimization.
The memoized value won't ever change anyway as `handlePress` is stable.

atomiks: Yeah you're right - the nested object should also be stable in this case since the top level object itself is stable between renders

michaldudak (approved):

PR #1738, opened 2025-04-17
[docs-infra] Create llms.txt (WIP)

Janpot:
WIP


colmtuite:
@Janpot Is it better to have just a single llms.txt file in the sidebar that documents the entire project, rather than individual files for each page like this?

Janpot:
> @Janpot Is it better to have just a single llms.txt file in the sidebar that documents the entire project, rather than individual files for each page like this?

I regularly find myself pasting a link to a documentation page in a chat window to ask questions or to make it not hallucinate when generating code. The idea in this PR is the simplest form of AI [chat integration](https://www.notion.so/mui-org/docs-infra-AI-integration-1e8cbfe7b660805f833efc5c01048146?source=copy_link#231cbfe7b6608098952ade30b0c0c8ec) such as "copy markdown", "link to markdown", "ask questions about this page in chatgpt/claude",... shortcuts to bring the page content into a LLM context.

I tried to keep it minimal but still discoverable. Happy to do something else, from removing it altogether to a full fledged menu as in those benchmarks, to adding a llms.txt link somewhere.

colmtuite:
@Janpot Yes but do you have a good sense of the utility of copying markdown/url for individual pages, versus a single url for the entire docs? If a single llms.tsx for the entire docs worked well, that would be better obviously.

Ime, after being fed an entire docs llms.txt, they will often still mess up, or seemingly forget documentation, and still need to be reminded of specific docs sections. But also, ime, I've felt like simply sending regular urls can work well from that point.

Also, I wonder about the utility of linking to the markdown url versus just having a button to copy the url? I'm not sure I can think of a use case for visiting the url?

I feel like the best approach might be to have an "llms.txt" link in the left sidebar, under "Handbook", and then a "Copy for llms" button on each page. But I'm not sure, curious what you think.

Janpot:
> Yes but do you have a good sense of the utility of copying markdown/url for individual pages, versus a single url for the entire docs? If a single llms.tsx for the entire docs worked well, that would be better obviously.

Depends but quite often, on documentation websites, I want to show the llm a specific page. e.g. "read this specific page so you can stop hallucinating random details". I do that because I'm sure for many use-cases, the llms.txt doesn't contain enough information for it to find that page on its own. I see llms.txt more as something we can feed to an MCP, but I don't have a lot of experience with that.

> But also, ime, I've felt like simply sending regular urls can work well from that point.

Yes, but the point of the markdown is to reduce noise and amount of input tokens. I wouldn't be surprised if some users want to copy+paste a specific section from the markdown (to preserve some basic formatting).

> Also, I wonder about the utility of linking to the markdown url versus just having a button to copy the url? I'm not sure I can think of a use case for visiting the url?

I'd 90% of the time use it as "right-click > copy link address". I'm sure some people would rather copy the markdown itself and paste it in their chat. So for them opening the markdown page would be beneficial. I'm not really the biggest fan of copy buttons, I prefer it if websites let me manage my clipboard myself. But I have no strong opinion, I'll do whatever the base team wants.

> But I'm not sure, curious what you think.

Personally I'd stick to a more neutral "copy markdown" or "copy markdown link", but all is fine for me, I don't think it matters too much UX-wise whether it's a link, or a copy button and what the text is. I think we should do whatever you feel like is most "base ui docs like"

colmtuite:
Right yes ok. Copying sections of markdown is a good use case. Ok well I think a library-wide llms.txt link will be useful too, but can do it in a separate PR if you wish.

For this PR, the link should be a regular light gray link (same as other light gray links on the site). It should be located underneath the component description on mobile, and baseline-aligned with the component description on larger viewports.

Screenshot 2025-07-16 at 12 22 05

colmtuite:
@Janpot looks like the npm link in the header is now missing the version number. Also the route for the llms.txt link in the sidebar is missing `.txt`

Janpot:
😄  well,... linking to /llms.txt works in Next.js dev mode only apparently. Not a surprise I guess, the next.js router doesn't know about these static files and does weird stuff anyway with routes that contain file extensions.
npm version regression was introduced in https://github.com/mui/base-ui/pull/2167 (cc @michaldudak in case there was a specific reason to remove the version from root package.json?)

dav-is (approved):

docs/src/nav.ts
```diff
@@ -35,6 +47,11 @@ export const nav = [
         label: 'Composition',
         href: '/react/handbook/composition',
       },
+      {
+        label: 'llms.txt',
```
brijeshb42, line 51: I feel we should highlight this somewhere in the content as well. Like in the quick start guide, we can add an `llm` section.

docs/scripts/generateLlmTxt/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+        let content = [
+          '---',
+          `title: ${title || 'Untitled'}`,
+          subtitle ? `subtitle: ${subtitle}` : '',
+          description ? `description: ${description}` : '',
+          '---',
+          '',
+          markdown,
+        ]
+          .filter(Boolean)
```
brijeshb42, line 149: Nit: This filter will remove the empty string on line 68 as well.

docs/scripts/generateLlmTxt/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+    const llmsFilePath = path.join(OUTPUT_BASE_DIR, 'llms.txt');
+    const prettierOptions = await prettier.resolveConfig(llmsFilePath);
+
+    llmsTxtContent = await prettier.format(llmsTxtContent, {
+      ...prettierOptions,
+      filepath: llmsFilePath,
+      parser: 'markdown',
+    });
+
+    await fs.writeFile(path.join(OUTPUT_BASE_DIR, 'llms.txt'), llmsTxtContent, 'utf-8');
```
brijeshb42:
```suggestion
    await fs.writeFile(llmsFilePath, llmsTxtContent, 'utf-8');
```

docs/scripts/generateLlmTxt/demoProcessor.mjs
```diff
@@ -0,0 +1,121 @@
...
+  const demoPath = path.resolve(mdxDir, pathAttr);
+
+  // Check if the demo folder exists
+  if (!fs.existsSync(demoPath)) {
+    throw new Error(`Demo folder not found at "${demoPath}"`);
+  }
+
+  // Define implementation types and their configurations
+  const implementationTypes = [
+    {
```
brijeshb42, line 66: I don't know if this will affect the llm in anyway since it'll consume the whole mdx (and not partially), but do you think tailwind should be showcased before css modules ?

dav-is: Why Tailwind first when CSS Modules is the default?

brijeshb42: Since llms consume tailwind better. Atleast this is what I have been led to believe in all the llm discourse.

Janpot: I don't think it'll matter but I'm happy to move it around

brijeshb42 (approved):
👏🏼

colmtuite:
@Janpot 

For the sidebar llms.text link, I was thinking of a link to the entire markdown for the whole docs website. Here's an example from [Zero docs](https://zero.rocicorp.dev/llms.txt). Screenshot attached.

Screenshot 2025-07-22 at 10 04 00

Apparently this is commonly called llms-full.txt? idk, I'm not experienced with it, just figured I'd mention it. If you know that linking to the navigation links list works well, then let's go with that.

docs/src/nav.ts
colmtuite: Good idea. Even something super simple at the bottom of the Quick Start page might help:

**Working with LLMs**
For those of you working with LLMs, each docs page has a "View as Markdown" link at the top, which can share share with AI chat assistants to help them understand Base UI concepts and component APIs.

Additionally, there is an "llms.txt" link in the "Handbook" section of the navigation sidebar, which you can feed to AI chat assistants to help them navigate the docs.

Janpot:
> If you know that linking to the navigation links list works well, then let's go with that.

If I'm honest, nothing about llms.txt is properly standardized, and as far as I know, no major vendor is supporting any of this. I'm happy to create an additional llms-full.txt, it's not a lot of work. I think we should just do what the industry converges on (I have no idea what that is exactly though).

The problem I see with llms-full.txt is that it creates a massive amount of input tokens.

docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
```diff
@@ -47,3 +47,9 @@ There are examples for both Tailwind and CSS Modules below, but since Base UI i
 
 This walkthrough outlines the basics of putting together a Base UI component.
 Browse the rest of the documentation to see what components are available in the library and how to use them.
+
+## Working with LLMs
+
+For those of you working with LLMs, each docs page has a "View as Markdown" link at the top, which can share share with AI chat assistants to help them understand Base UI concepts and component APIs.
```
brijeshb42:
```suggestion
For those of you working with LLMs, each docs page has a "View as Markdown" link at the top, which can be shared with AI chat assistants to help them understand Base UI concepts and component APIs.
```

colmtuite:
@Janpot right ok, happy to go with whatever you think is best. ✅

atomiks (commented):
It looks like the links must be absolute, not relative.

- https://vercel.com/docs/llms.txt
- https://circleci.com/llms.txt

Supporting both `llms.txt` and `llms-full.txt` seems fine, as they have different use cases with different AIs based on their strength or weaknesses

docs/scripts/generateLlmTxt/demoProcessor.mjs
colmtuite: yes tailwind should be the default whether it helps ai or not.

Janpot:
🤷  There's no proper spec so who really knows what kind of links are allowed? imo, since it's a document published to the web, it must have an origin. Resolving urls in that context should be very well defined, given how well this is specced, I bet even an LLM would do a great job at that task (though it should probably be just done by a tool). Anyway, the vercel markdown version of the docs are full of relative urls, and the circle ci version points to html files. None of them is doing the same thing.

But ok. Let's try to find what's the most common sensible thing. Here's what I propose we do:

* inside .md files, all internal urls resolved to their .md variant, but we keep them root relative. This keeps them nice and portable and creates an interconnected set of markdown files. And none of the example seems to suggest it's important to have absolute urls there. (so `./popover` becomes `/react/components/popover.md`)
* inside llms.txt all urls resolved to their absolute url. (so `/react/components/popover` becomes `https://base-ui.com/react/components/popover.md`)
* inside llms-full.txt 
  * all urls that would resolve to another markdown file, use a hash url to their title (so `/react/components/popover` becomes `#popover`, technically there could be duplicates between titles/subtitles I guess). This can signal an llm that the info the link points to is already in the document, and nothing needs to be fetched,
  * all other internal urls, resolve to their absolute url. (so `/some/other/url` becomes `https://base-ui.com/some/other/url` but we currently have none of these)

wdyt?

> Supporting both llms.txt and llms-full.txt seems fine, as they have different use cases with different AIs based on their strength or weaknesses

Which one would you link to from the sidebar?

atomiks:
> Which one would you link to from the sidebar?

`llms.txt` is likely best there

I think this looks good now aside from that link ✅

docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
```diff
@@ -47,3 +47,9 @@ There are examples for both Tailwind and CSS Modules below, but since Base UI i
 
 This walkthrough outlines the basics of putting together a Base UI component.
 Browse the rest of the documentation to see what components are available in the library and how to use them.
+
+## Working with LLMs
+
+For those of you working with LLMs, each docs page has a "View as Markdown" link at the top, which can be shared with AI chat assistants to help them understand Base UI concepts and component APIs.
+
+Additionally, there is an "llms-full.txt" link in the "Handbook" section of the navigation sidebar, which you can feed to AI chat assistants to help them navigate the docs.
```
atomiks:
```suggestion
Additionally, there is an "llms.txt" link in the "Handbook" section of the navigation sidebar, which you can feed to AI chat assistants to help them navigate the docs.
```

Janpot: Good catch, I'll also turn that into a link.

michaldudak (approved):

Issue #1214, opened 2024-12-23
[Popover, Menu, Select] Improve outside press using touch modality


atomiks:
Upstream issue: https://github.com/floating-ui/floating-ui/issues/3138

On iOS for popups like Select and Menu, the outside press behavior is `touchend` + `touchmove` with an ~8px movement threshold. If you pressed and held the screen for >1s, then `touchend` also gets ignored.

This:

1. Makes it easier to dismiss, requiring a less "intentional" press, allowing you to scroll away to dismiss the popup much more easily and feels far less restrictive and annoying
2. Makes it a bit harder to unintentionally dismiss accidentally (the `1s` threshold)

Modal dialogs and drawers on iOS match the native `mousedown` event, which requires an intentional/non-sloppy outside press however.

- `Popover` currently uses `pointerdown` which doesn't match either behavior. It's less restrictive than `mousedown`, but also too permissive/loose compared to iOS' native behavior.
- `Menu` and `Select` currently use `mousedown`, which feels too restrictive — you can't easily scroll away when it's open.
- `Dialog`/`AlertDialog` should continue using `mousedown` as-is, as they take up more of the screen with an obstructive backdrop and are modal-like.

benface:
> `Dialog`/`AlertDialog` should continue using `mousedown`

@atomiks – I feel like it's more friendly to close dialogs on _release_ (mouseup) rather than on press (mousedown), as long as the press was also on the backdrop. Would you be open to make that customizable? 🙏

atomiks:
@benface it needs a bit of thought into the API as to how it works on touch vs. mouse.

Internally it's configurable via Floating UI in three ways but the upstream issue details why those aren't sufficient:
- `click` means it closes on "release" on both mouse and touch, but if you scrolled a bit (pointermove on touch) it cancels it from having an effect. This is part of the original issue^ - this should only be the behavior for dialogs with dimming backdrops, not transparent ones. Not used by any of the components yet, but you want this one.
- `mousedown` means it closes on "release" on touch, but on "press" on mouse (`Select`, `Menu`, `Dialog`)
- `pointerdown` means it closes on "press" on both touch and mouse (`Popover`). Not suitable for dialogs with backdrops since touch events fall through and aren't blocked with sloppy outside touches

Most likely it needs some configuration option to separately handle touch and mouse.

```js
outsidePressBehavior={{
  mouse: 'down', // or start/end?
  touch: 'up', // or start/end?
}}
```

[...]

PR #1741, opened 2025-04-18
[NavigationMenu] New `NavigationMenu` component

atomiks:


mj12albert:
https://github.com/user-attachments/assets/33f49757-3ed1-47db-8a81-5ad6a6362393

I noticed the scrollbar on the 2nd item flashed when hovering from trigger 1 > trigger 2 (at ~4s)

atomiks:
@mj12albert  I see this issue happens with `Always` scrollbars. I will try to see if I can do something else to fix this problem, but using our `ScrollArea` component can avoid it since it uses overlay scrollbars that can be hidden while transitioning the scrollable content

atomiks:
I decided to remove the scrollability of the popups from the demos by default and instead add a section of how to handle it.

michaldudak:
I'd intuitively expect the <kbd>DownArrow</kbd> to open the menu, as in Menubar. Also, when the menu is open, the focus should be placed within the popup.

atomiks:
@michaldudak I think the `<nav>` role has a few different focus semantics with composite keys + tabbing, in this case I did what Radix does, but there's definitely some debate about how this should be handled. 

cc: @colmtuite 

- https://www.makethingsaccessible.com/guides/site-navigation-is-not-an-aria-menu/
- https://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html

michaldudak:
Radix works slightly better, as when the popup is open, it allows focusing its contents with the <kbd>DownArrow</kbd> while our solution focuses the next nav menu item.

I like interactions in Ariakit's [Navigation Menubar](https://ariakit.org/examples/menubar-navigation) a bit more, although it's not perfect either - I wouldn't open the menu on focus and left/right arrow keys should work differently when menu contents are laid out in a grid.

atomiks:
On https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/

> The menubar pattern requires complex functionality that is unnecessary for typical site navigation that is styled to look like a menubar with expandable sections or fly outs.

If you don't use `role="menubar"`, those interactions are not strictly necessary

> Radix works slightly better, as when the popup is open, it allows focusing its contents with the DownArrow while our solution focuses the next nav menu item.

This seems correct, will fix

michaldudak:
There are a couple of things that can be improved regarding keyboard navigation:
- I'd expect the <kbd>DownArrow</kbd> to open the menu, not to move between triggers (especially if I see down arrows/chevrons). Or, if we decide it shouldn't work like this, it should at least prevent default when the last element is focused (same with <kbd>UpArrow</kbd> and the first item
- <kbd>Space</kbd> should focus the content of the opened menu
- nice to have: we could showcase grid navigation in the Overview menu in the demo (have <kbd>UpArrow</kbd> / <kbd>DownArrow</kbd> navigate between rows)

---

What's the use of the `open` prop on the Root? Wouldn't `value` be enough?

atomiks:
> What's the use of the open prop on the Root? Wouldn't value be enough?

Hmm interesting. A non-nullish value could indeed signal that it's open.

> nice to have: we could showcase grid navigation in the Overview menu in the demo (have UpArrow / DownArrow navigate between rows)

It's a bit hard in the demo only using public APIs due to the fact that it turns into a column on small screens

michaldudak:
Focus restoration bug:
1. tab into the first trigger
2. click away to close the menu
3. hover over items
4. when you hover over the Link, the previous Trigger draws the outline


https://github.com/user-attachments/assets/bfaffc19-25cd-491f-86cb-3770c84ee64f

packages/react/src/navigation-menu/root/NavigationMenuRoot.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+  return (
+    <NavigationMenuTreeContext.Provider value={nodeId}>
+      {element}
+    </NavigationMenuTreeContext.Provider>
+  );
+}
+
+export namespace NavigationMenuRoot {
+  export interface State {
```
michaldudak, line 234: `nested` should also be a part of State so they can render different components based on this (as we do by default).

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
```diff
@@ -0,0 +1,401 @@
...
+    delay,
+    closeDelay,
+    orientation,
+    listRef,
+  } = useNavigationMenuRootContext();
+  const itemValue = useNavigationMenuItemContext();
+  const nodeId = useNavigationMenuTreeContext();
+  const tree = useFloatingTree();
+
+  const timeout = useTimeout();
```
michaldudak, line 82: This seems unused

packages/react/src/navigation-menu/positioner/useNavigationMenuPositioner.ts
michaldudak, line 1: Is the hook reused somewhere? If not, its contents can be moved to the component.

packages/react/src/navigation-menu/content/NavigationMenuContent.tsx
```diff
@@ -0,0 +1,150 @@
...
+      setFocusInside(true);
+    },
+    onBlur(event) {
+      if (!contains(event.currentTarget, event.relatedTarget)) {
+        setFocusInside(false);
+      }
+    },
+  };
+
+  const renderElement = useRenderElementLazy('div', componentProps, {
```
michaldudak, line 105: Use the `enabled` option of `useRenderElement` instead.

mj12albert:
Testing on my (small) iPhone, the popup closes immediately (pointerdown it seems?) when I try to scroll:

https://github.com/user-attachments/assets/b3196278-e23f-4591-ba70-b5c1c403911e

atomiks:
@mj12albert I'm assuming you mean scroll outside? This is due to the `pointerdown` event, which Popover also uses by default. There's no modality with this component

In terms of improving outside press behavior on touch input, there's already an issue #1214

Edit: nvm, I see it's reproducible on iPhone by scrolling the content also

mj12albert:
> I see it's reproducible on iPhone by scrolling the content also

Yeah I was trying to touch inside to scroll!

mj12albert:
> I see it's reproducible on iPhone by scrolling the content also

Yeah I was trying to touch inside to scroll!

atomiks:
@mj12albert should be fixed - it was caused by the window `resize` causing it to close, I don't think it's strictly necessary except in extreme edge cases

michaldudak (approved):
Looks good to me!

PR #1753, opened 2025-04-21
[Radio] Fix composite focus of initially selected radio item

### LINKED ISSUE #1740: [Radio] First `Radio`, rather than checked one, is focused when navigating to `RadioGroup` without previously interacting with it
# Bug report

## Current behavior

If a `RadioGroup` has a `value` or a `defaultValue` that matches a `Radio` that is not the first one in the group, navigating to the `RadioGroup` with the keyboard (no matter the direction: from above or from below) moves the focus to the _first_ `Radio` rather than the checked one. However, after interacting with any of the radios, that is, clicking on any of them or using the arrow keys when one of them is focused, it works as expected the following time that the group is navigated to.

## Expected behavior

I expect the checked `Radio`, if any, to get the focus when the `RadioGroup` is navigated to using the keyboard. Only if there is no checked `Radio` do I expect the first one to get the focus.

## Reproducible example

https://codesandbox.io/p/sandbox/infallible-sun-wtln3v

## Base UI version

1.0.0-alpha.8

Chrome, Safari, and Firefox

## Which OS are you using?

macOS

None

atomiks:
Fixes #1740

- `queueMicrotask` lets the effect wait for the composite to have set the indices of the refs from their DOM position
- `data-composite-active` attr adentifier on a composite item can inform it to set a default index different from `0`, which can work on arbitrary composite items

### DIFF
--- a/packages/react/src/composite/root/useCompositeRoot.ts
+++ b/packages/react/src/composite/root/useCompositeRoot.ts
@@ -33,6 +33,8 @@ import {
   type Dimensions,
   type ModifierKey,
 } from '../composite';
+import { ACTIVE_COMPOSITE_ITEM } from '../constants';
+import { useEnhancedEffect } from '../../utils';
 
 export interface UseCompositeRootParameters {
   orientation?: 'horizontal' | 'vertical' | 'both';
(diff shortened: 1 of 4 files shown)


packages/react/src/composite/root/useCompositeRoot.ts
```diff
@@ -113,6 +115,22 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
...
+  const hasSetDefaultIndexRef = React.useRef(false);
+
+  const onMapChange = useEventCallback((map: Map<Element, CompositeMetadata<any>>) => {
+    if (map.size === 0 || hasSetDefaultIndexRef.current) {
+      return;
+    }
+    hasSetDefaultIndexRef.current = true;
+    const sortedElements = Array.from(map.keys());
+    // Set the default highlighted index of an arbitrary composite item.
+    const activeIndex = sortedElements.findIndex((compositeElement) =>
```
mj12albert, line 127: Nice ~ in Tabs I used a controlled composite and an effect to do this: https://github.com/mui/base-ui/blob/master/packages/react/src/tabs/tab/useTabsTab.ts#L63-L71

mj12albert (approved):

PR #1755, opened 2025-04-21
[form][field] Fix integration of Base UI components

### LINKED ISSUE #1745: [form][select] `data-invalid` not cleared on Select and RadioGroup fields.
# Bug report

Errors set to radioGroups and Select elements from within a Form component do not seem to clear. The fields keep `data-invalid`, even though the element itself has been changed. This means the error set from the Form component does not clear.

## Expected behavior

After setting errors on form submit, when the field is being interacted with (e.g. an option is selected in the select field), I expected the `data-invalid` tag to disappear similar to how the Field component works.

## Reproducible example

**Example 1**
Example setting form errors with Zod: https://codesandbox.io/p/sandbox/confident-benz-prlpnt

First, fill in all fields and click submit, the form data is logged correctly in the console.

Refresh and try again, now click submit when not filling in any field. All fields are set to `data-invalid` and a validation message is shown below each field. When editing the regular text inpput Fields, the `data-invalid` is removed from the field the moment the content of the input changes. The error is also cleared this moment.

When the select is changed, the `data-invalid` remains on the select field. The error is not cleared.

When a radio option is selected, the same happens. 

The Form won't submit, likely because there are invalid fields in the Form.

**Example 2**
Example without Zod, using setErrors in the onSubmit of a Form element to manually set errors for each field the first time the form is submitted: https://codesandbox.io/p/devbox/peaceful-meitner-5hpx4j

Interacting with the regular text Fields clears the error and removes the `data-invalid`, interacting with the select does not clear the `data-invalid` or remove the error message.

## Base UI version

Latest version

### LINKED ISSUE #1758: [select] Required option does not prevent form submit
# Bug report

## Current behavior

Currently setting required on the `Select.Root` inside a `Form`, does not prevent the form from submitting when the select is empty and no option has been chosen. The validity does not return a valueMissing error.

The workaround shown here: https://github.com/mui/base-ui/issues/1425 does also not seem to work to check for a missing value on form submit.

## Expected behavior

I would expect an empty select field to prevent a form submit and show a missing value error

## Reproducible example

Codesandbox example: https://codesandbox.io/p/devbox/boring-bush-mrf8r5

Entering value in the radio and text field without interacting with the select field, then clicking the submit button, triggers the onsubmit (form data is logged in the console)

## Base UI version

Latest version

atomiks:
All form parts: https://deploy-preview-1755--base-ui.netlify.app/experiments/form

- Fixes #1745 — `onClearErrors` was not integrated with Base UI components
- Fixes #1758 — `Select` was missing `useField` integration
- Now with native validation, it will commit any revalidation on change, matching the revalidation of the controlled `errors` prop

Moved `Field` tests to each individual component suite, much easier to see what is being tested for each Base UI component


packages/react/src/form/Form.tsx
```diff
@@ -92,9 +92,22 @@ const Form = React.forwardRef(function Form(
...
+    }
+  });
+
   const contextValue: FormContext = React.useMemo(
-    () => ({ formRef, errors: errors ?? {}, onClearErrors }),
-    [formRef, errors, onClearErrors],
+    () => ({
+      formRef,
+      errors: errors ?? {},
+      onClearErrors,
```
flaviendelangle, line 107: Do we still need `onClearErrors` on the context?
I can't see any other usage, removing it would avoid some confusion between two similarly named methods.

atomiks: Yeah good point! Will remove it.

mj12albert:
I think it would feel more responsive if the errors were cleared on change (or for the validation to re-run on change) even when `validationMode="onBlur"` - https://www.smashingmagazine.com/2022/09/inline-validation-web-forms-ux/#4-reward-early-punish-late

https://github.com/user-attachments/assets/88867abe-5be7-4d30-bffe-f0baca5824ce

(RHF lets you separately configure re-validation mode: https://react-hook-form.com/docs/useform#reValidateMode)

atomiks:
@mj12albert yeah this is a good point and something I noticed but wasn't sure how to handle with `validationMode="onBlur"`, but a new prop like `revalidationMode` would help in this case. Since the `errors` controlled prop clears on change already, I will match that for revalidation, but we can make it configurable later.

packages/react/src/radio-group/useRadioGroup.ts
```diff
@@ -53,6 +55,26 @@ export function useRadioGroup(params: useRadioGroup.Parameters) {
...
+    clearErrors(name);
+
+    if (validationMode === 'onChange') {
+      fieldControlValidation.commitValidation(checkedValue);
+    } else {
+      fieldControlValidation.commitValidation(checkedValue, true);
+    }
+  }, [name, clearErrors, validationMode, checkedValue, fieldControlValidation]);
+
+  useModernLayoutEffect(() => {
```
mj12albert, line 74: Perhaps worth making a `usePreviousValueRef` util?

mj12albert (approved):
It works pretty good now, though I agree revalidation needs more thought (providing configurability vs being opinionated about the UX, maybe we could be opinionated because we don't have to be as flexible as a general form lib)

One thing (that I think can be done in a separate PR) is maybe only revalidate `required` on change, so the user isn't scolded by another error before finishing typing (but it's much better than feeling unresponsive)

https://github.com/user-attachments/assets/bb01abe2-9c6d-4163-924d-96f326c34758

atomiks:
@mj12albert good point, will do in a separate PR 👍

packages/react/src/radio-group/useRadioGroup.ts
atomiks: It's a bit hard to make one of those because you can't read refs in render and so you need to make a callback hook, but then `NumberField` requires _two_ different values to be watched for changes, making the abstraction more complex. I think it's okay to leave it like this for now

PR #1756, opened 2025-04-21
[docs] Mark required props in API docs

### LINKED ISSUE #1694: [docs] Radio docs are incorrect/lacking
# Docs feedback

The docs for the [Radio.Root](https://base-ui.com/react/components/radio#root) component mention:

> Represents the radio button itself. Renders a <button> element and a hidden <input> beside.

But that's incorrect; the hidden `<input>` seems to be rendered as a sibling of the `<button>` element rather than inside it.

Which begs the question... why can't `<Radio.Root>` actually render a `<label>` that contains both the `<button>` and the `<input>`? Seems strange to have to wrap it in a `<label>` ourselves, as per the example... it's not really documented so maybe that's optional, but if we don't wrap it in a `<label>`, how is the radio going to be labelled?

ZeeshanTamboli:

Closes #1694 (See https://github.com/mui/base-ui/issues/1694#issuecomment-2790649653)

Preview:
- `Radio.Root` -  https://deploy-preview-1756--base-ui.netlify.app/react/components/radio#root
- `Menu.RadioItem` - https://deploy-preview-1756--base-ui.netlify.app/react/components/menu#radio-item


colmtuite:
Why does this need to be documented? It becomes clear pretty quickly in the IDE, no? I'm not a fan of adding superfluous content to the actual prop name in the docs, it could be misleading.

ZeeshanTamboli:
> Why does this need to be documented? It becomes clear pretty quickly in the IDE, no? I'm not a fan of adding superfluous content to the actual prop name in the docs, it could be misleading.

Then why document any props at all if they're visible in the IDE? Also, did you see this comment? https://github.com/mui/base-ui/issues/1694#issuecomment-2790649653

colmtuite:
Well because not all props are equal in this regard. I'm not saying nothing should change here, I'm just saying that making the prop names inaccurate in the table is not the right solution.

ZeeshanTamboli:
> I'm just saying that making the prop names inaccurate in the table is not the right solution.

I assume you're referring to the asterisk after the prop names. From what I've seen, it's a common convention — for example:

- Material UI: [`children` prop](https://mui.com/material-ui/api/tooltip/) is marked as required.
- Mantine UI: [Props table](https://mantine.dev/core/tooltip/?t=props) uses a similar approach.

Do you have a different suggestion in mind?

atomiks:
I think the asterisk is ok but it needs to be in a `<sup>`, maybe red too (Mantine does this at least). This matches the common convention of required fields in forms all over the web. The `*` on the same level as the prop name looks a bit odd

ZeeshanTamboli:
> I think the asterisk is ok but it needs to be in a `<sup>`, maybe red too (Mantine does this at least). This matches the common convention of required fields in forms all over the web. The `*` on the same level as the prop name looks a bit odd

@atomiks Made the changes.

atomiks (approved):
This change looks fine to me 👍 

cc: @colmtuite  @mnajdova @michaldudak

Perhaps `title="required"` or something else is required for screen readers.

michaldudak:
I'd place the asterisk slightly lower (`top: -.3em;` looks better), but let @colmtuite decide.

colmtuite:
Yeah `top: -.3em;` is more natural. With that tweak we're good 👍

PR #1768, opened 2025-04-23
[internal] Use `useRenderElement()` in Composite Item and Composite Root

flaviendelangle:
I noticed it while updating my Calendar PoC.


packages/react/src/composite/item/CompositeItem.tsx
```diff
@@ -1,16 +1,15 @@
...
 import { useCompositeItem } from './useCompositeItem';
 import type { BaseUIComponentProps } from '../../utils/types';
 
 /**
  * @internal
  */
-function CompositeItem<Metadata>(props: CompositeItem.Props<Metadata>) {
-  const { render, className, itemRef, metadata, ...otherProps } = props;
+function CompositeItem<Metadata>(componentProps: CompositeItem.Props<Metadata>) {
+  const { render, className, itemRef = null, metadata, ...elementProps } = componentProps;
```
flaviendelangle, line 12: `useRenderElement` accepts `React.Ref<RenderedElementType>` which does not include `undefined`.

I can update the type or pass `null` as you prefer

packages/react/src/composite/item/CompositeItem.tsx
```diff
@@ -22,15 +21,10 @@ function CompositeItem<Metadata>(props: CompositeItem.Props<Metadata>) {
...
-  const { renderElement } = useComponentRenderer({
-    propGetter: getItemProps,
-    ref: mergedRef,
-    render: render ?? 'div',
+  const renderElement = useRenderElement('div', componentProps, {
     state,
-    className,
-    extraProps: otherProps,
+    ref: [itemRef, ref],
+    props: [getItemProps, elementProps],
```
mj12albert, line 27: `getItemProps` could be replaced by returning a `props` object from `useCompositeItem` (and same for getRootProps/useCompositeRoot) e.g. https://github.com/mui/base-ui/blob/master/packages/react/src/popover/popup/usePopoverPopup.ts#L11-L17

I think it should be a bit faster as mergeProps is only done once

packages/react/src/composite/root/CompositeRoot.tsx
```diff
@@ -49,12 +49,9 @@ function CompositeRoot<Metadata extends {}>(props: CompositeRoot.Props<Metadata>
       modifierKeys,
     });
 
-  const { renderElement } = useComponentRenderer({
-    propGetter: getRootProps,
-    render: render ?? 'div',
+  const renderElement = useRenderElement('div', componentProps, {
     state: {},
```
mj12albert, line 53: Unrelated to the PR but noticed that this empty object should be outside the component

packages/react/src/composite/item/CompositeItem.tsx
mj12albert: I think `null` is ok, as I remember, we didn't bother to forwardRef here since it's an internal component and the generic makes it a bit annoying

packages/react/src/composite/item/CompositeItem.tsx
flaviendelangle: Same logic applied to `useCompositeRoot`

mj12albert (approved):
Looks good ~

PR #1770, opened 2025-04-23
[Collapsible] POC using `grid-template-*` properties as open/close animation target

mj12albert:
Demo: https://deploy-preview-1770--base-ui.netlify.app/experiments/collapsible2/transitions

### API

The anatomy remains the same, except `Collapsible.Panel` must render an extra `div` (internally) around its children

CSS transitions:
```css
.Panel {
  display: grid;
  grid-template-rows: 1fr;

  transition: all var(--duration);

  &[data-starting-style],
  &[data-ending-style] {
    opacity: 0;
    grid-template-rows: 0fr;
  }
}
```

CSS animations:
```css
@keyframes slide-down {
  from {
    grid-template-rows: 0fr;
  }

  to {
    grid-template-rows: 1fr;
  }
}

@keyframes slide-up {
  from {
    grid-template-rows: 1fr;
  }

  to {
    grid-template-rows: 0fr;
  }
}

.Panel {
  display: grid;

  &[data-open] {
    animation: slide-down var(--duration) ease-out;
  }

  &[data-closed] {
    animation: slide-up var(--duration) ease-in;
  }
}
```

### 👍 
- Don't need to manage resizing when using CSS transitions

### 👎 
- When transitioning, the size of the panel isn't 1:1 with the actual layout. Notice the size of the gap between the two collapsibles as it transitions:

https://github.com/user-attachments/assets/c11935be-04fb-458d-b656-d867b30314a1

When transitioning `height` it looks like this:

https://github.com/user-attachments/assets/31605424-ce69-4784-a10a-b932cf5298cf

Ariakit has the same issue: https://codesandbox.io/p/sandbox/hungry-goldwasser-mktxk5

- It's even more difficult to suppress the initial open animation when using CSS animations

### DIFF
--- a/docs/src/app/(private)/experiments/collapsible2/transitions.module.css
+++ b/docs/src/app/(private)/experiments/collapsible2/transitions.module.css
@@ -1,0 +1,76 @@
+.Root {
+  --width: 320px;
+  --duration: 800ms;
+
+  width: var(--width);
+
+  & + .Root {
+    margin-top: 2rem;
+  }
+}
+
+.Panel {
+  display: grid;
+  box-sizing: border-box;
+  width: 100%;
+  grid-template-rows: 1fr;
+
+  transition: all var(--duration);
+
+  &[data-starting-style],
+  &[data-ending-style] {
+    opacity: 0;
+    grid-template-rows: 0fr;
+  }
+}
+
+.Content {
+  display: flex;
+  flex-direction: column;
+  gap: 0.5rem;
+  margin-top: 0.25rem;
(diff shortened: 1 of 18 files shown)


mj12albert:
@colmtuite I think these are all the major downsides of this approach, though the upside of not having to manage resizing is very attractive, it doesn't seem good enough

atomiks:
This is a really nice breakdown of the pros/cons!

The fact that it doesn't resize with layout 1:1 might be a dealbreaker in my opinion, although maybe it's not that bad in practice

PR #1772, opened 2025-04-24
[Slider] Use `useRenderElement`

mj12albert:
Also fixed some minor issues:
- `event.stopPropagation` on the thumb's keydown handler when it is a composite key
- the internal active thumb index wasn't set when interacting with a keyboard
- onValueCommitted was duplicated in the root context


packages/react/src/utils/useRenderElement.tsx
```diff
@@ -73,7 +74,11 @@ export function useRenderElement<
 }
 
 export namespace useRenderElement {
-  export interface Parameters<State, RenderedElementType extends Element> {
+  export interface Parameters<
+    State,
+    RenderedElementType extends Element,
+    RenderFunctionProps extends GenericHTMLProps,
```
mj12albert, line 80: `GenericHTMLProps` didn't include `htmlFor` for the `output` element, for Slider.Value `React.OutputHTMLAttributes` is needed

atomiks: Is it possible to infer it from the tag passed as the first argument? Also encountered this with `Form`

mj12albert: Yes ~ got it working in https://github.com/mui/base-ui/pull/1772/commits/187276b190a2b07c5caefec018db10f8f35cf9de

mj12albert:
Also added a couple of demos in https://deploy-preview-1772--base-ui.netlify.app/experiments/slider/slider

PR #1776, opened 2025-04-24
[test] Remove deprecated playwright package

michaldudak:
Removed the `playwright` package imports (as mentioned in https://github.com/mui/material-ui/pull/45967#discussion_r2057786156)


renovate.json
```diff
@@ -55,12 +55,7 @@
     },
     {
       "groupName": "MUI",
-      "matchPackagePatterns": [
-        "@mui/*",
-        "playwright",
-        "@playwright/test",
-        "mcr.microsoft.com/playwright"
-      ],
+      "matchPackagePatterns": ["@mui/*", "@playwright/test", "mcr.microsoft.com/playwright"],
```
Janpot, line 58: Not hugely important, but this caught my eye. What a strange group? If it's intentional, would suggest to add a [`description`](https://docs.renovatebot.com/configuration-options/#description) field for clarification.

Janpot (approved):

renovate.json
JCQuintas: I believe the issue is that `@mui/internal-test-utils` has a lot of test dependencies, one of them being `playwright`. If we have multiple versions installed, playwright doesn't work.

IMO we should remove a lot of the packages there from deps to peerDependencies to prevent issues.

There is another issue regarding chai and vitest that happens due to the deps there.

Janpot: Ah yes, because `@mui/internal-test-utils` is not a real dependency and therefore can't declare a peer dependency on playwright...

michaldudak: Actually, it is a real npm dependency. I don't remember why playwright wasn't added as a peer dependency there. Perhaps it was an omission.

michaldudak: Anyway, the docker image version (mcr.microsoft.com/playwright) must be the same as @playwright/test and I don't think there's another way to enforce this.

PR #1779, opened 2025-04-25
[Field] Fix validation inconsistency

### LINKED ISSUE #1778: [Field] the initial value is not validated and the validation is inconsistent
# Bug report

## Current behavior

`Field` does not validate the initial value and the validation is inconsistent and incorrect.

With this code:

```tsx
<Field.Root>
  <Field.Label>Email</Field.Label>
  <Field.Control type="email" defaultValue="not_an_email@" />
</Field.Root>

<Field.Root>
  <Field.Label>Email</Field.Label>
  <Field.Control type="email" value="not_an_email@" />
</Field.Root>
```

The root `div` initially has only `data-filled`. When I focus then blur without typing anything, it gets `data-valid`. But if I type something (to make it dirty) and revert back to old value, it gets `data-invalid`. The `input` always has `:invalid` pseudo-class in these scenarios.

## Expected behavior

- `Field` should validate the initial value (I'm not sure if this behavior is intended or not).
- The validation should be consistent between dirty value and non-dirty value.

## Base UI version

v1.0.0-alpha.8

Chrome

## Which OS are you using?

Mac OS

atomiks:
Fixes #1778 

- For `valueMissing`, require that the input have been dirtied at some point to produce an error on blur. This reduces errors for empty fields to begin with
- For any other error, it always produces an error on blur even if not dirtied yet

### DIFF
--- a/packages/react/src/field/control/useFieldControlValidation.ts
+++ b/packages/react/src/field/control/useFieldControlValidation.ts
@@ -42,18 +42,34 @@ export function useFieldControlValidation() {
     }
 
     function getState(el: HTMLInputElement) {
-      return validityKeys.reduce(
+      const computedState = validityKeys.reduce(
         (acc, key) => {
           acc[key] = el.validity[key];
-
-          if (!el.validity.customError && !markedDirtyRef.current) {
-            acc[key] = key === 'valid';
-          }
-
           return acc;
         },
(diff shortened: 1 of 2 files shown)


mj12albert (approved):
Looks good, it works well with Form too: https://codesandbox.io/p/devbox/sleepy-nash-tcfgpl

PR #1782, opened 2025-04-25
[popups] Refine `OpenChangeReason`

### LINKED ISSUE #1773: Refine `OpenChangeReason` values
We currently translate `OpenChangeReason` from Floating UI, but:

1. We don't consider Base UI specific reasons, like menu item click.
2. `trigger-press` vs. `click` is confusing in the Base UI API as one refers to `useDismiss` hook vs. the `useClick` hook internally, but `open: false` is possible with both reasons and it's not clear why they're differentiated.

We currently either expose one of the following strings, or just skip it with `undefined`, but should instead provide more useful reasons:
```jsx
export type OpenChangeReason =
  | 'click'
  | 'hover'
  | 'focus'
  | 'focus-out'
  | 'escape-key'
  | 'outside-press'
  | 'trigger-press';
```

atomiks:
Closes #1773

- `click` -> `trigger-press`
- `hover` -> `trigger-hover`
- `focus` -> `trigger-focus`

```ts
// Preview Card
OpenChangeReason

// Tooltip
OpenChangeReason | 'disabled'

// Popover
// Dialog
// Alert Dialog
OpenChangeReason | 'close-press'

// Select
OpenChangeReason | 'window-resize';

// Menu
OpenChangeReason | 'sibling-open';
```

- `close-press` occurs when clicking the the `Close` button part
- `item-press` occurs when clicking `Menu.Item` or `Select.Item`
- `cancel-open` occurs when they `mouseup` after `mousedown` outside of the menu or select


docs/reference/generated/alert-dialog-root.json
```diff
@@ -12,7 +12,7 @@
       "description": "Whether the dialog is currently open."
     },
     "onOpenChange": {
-      "type": "((open: boolean, event: Event | undefined, reason: OpenChangeReason | undefined) => void)",
+      "type": "((open: boolean, event: Event | undefined, reason: {} | undefined) => void)",
```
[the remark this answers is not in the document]
michaldudak: Yeah, there was a bug in how nested unions were handled. Could you update the extractor to https://pkg.pr.new/michaldudak/react-api-extractor@a21d7b8 and see if it helps?

packages/react/src/menu/root/useMenuRoot.ts
```diff
@@ -22,11 +22,12 @@ import { useOpenChangeComplete } from '../../utils/useOpenChangeComplete';
 import type { TextDirection } from '../../direction-provider/DirectionContext';
 import { useScrollLock } from '../../utils/useScrollLock';
 import {
-  type OpenChangeReason,
+  type BaseOpenChangeReason,
   translateOpenChangeReason,
 } from '../../utils/translateOpenChangeReason';
 
 const EMPTY_ARRAY: never[] = [];
+export type MenuOpenChangeReason = BaseOpenChangeReason | 'item-select' | 'cancel-open';
```
michaldudak, line 30: These types should be exported from the package so they can be used in userland functions

packages/react/src/dialog/root/useDialogRoot.ts
```diff
@@ -248,7 +250,7 @@ export namespace useDialogRoot {
     onOpenChange?: (
       open: boolean,
       event: Event | undefined,
-      reason: OpenChangeReason | undefined,
+      reason: DialogOpenChangeReason | undefined,
```
michaldudak, line 253: We should use the publicly exported type here (`AlertDialog.Root.OpenChangeReason`). If `DialogOpenChangeReason` is used, devs won't know where to import it from.

atomiks: I suppose type only imports don't lead to circular dependency issues?

atomiks: The generator can't handle the public type of `Select.Root.OpenChangeReason`. To be fair, the `Actions` type is also not clear where it exists from at the moment, so I could just do `OpenChangeReason`

michaldudak: Once I update the API extractor to handle types in namespaces, we shouldn't need to use `@type` annotations manually. Still, an ability to override could be useful sometimes.

atomiks: It does seem like they're still needed here

michaldudak: This doesn't look right. I'm fixing the API extractor to handle type aliases better.

docs/reference/generated/alert-dialog-root.json
```diff
@@ -12,7 +12,7 @@
       "description": "Whether the dialog is currently open."
     },
     "onOpenChange": {
-      "type": "((open: boolean, event: Event | undefined, reason: OpenChangeReason | undefined) => void)",
+      "type": "(open: boolean, event?: Event, reason?: AlertDialog.Root.OpenChangeReason) => void",
```
michaldudak, line 15: This isn't 100% accurate and goes against https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters-in-callbacks but since it's just for the docs, we can leave it.

packages/react/src/menu/root/useMenuRoot.ts
```diff
@@ -26,11 +26,13 @@ import { useOpenChangeComplete } from '../../utils/useOpenChangeComplete';
 import { useDirection } from '../../direction-provider/DirectionContext';
 import { useScrollLock } from '../../utils/useScrollLock';
 import {
-  type OpenChangeReason,
+  type BaseOpenChangeReason,
   translateOpenChangeReason,
 } from '../../utils/translateOpenChangeReason';
 import { ownerDocument } from '../../utils/owner';
 
+export type MenuOpenChangeReason = BaseOpenChangeReason | 'item-press' | 'cancel-open';
```
michaldudak, line 34: `item-press` and `cancel-open` are already defined in BaseOpenChangeReason. `sibling-open` can be moved here instead.

docs/reference/generated/alert-dialog-root.json
atomiks: Yeah, it looks shorter for the docs

michaldudak (approved):

PR #1785, opened 2025-04-26
[Toast] Fix swipe jump on iOS

atomiks:
On iOS, the `pointermove` doesn't start firing until swiping ~10px after `pointerdown`. This led to a glitchy jump after swiping past that threshold, which was mentioned by @colmtuite. But the solution to fix this is quite straightforward (as commented), the swipe movement is adjusted from when it starts moving, not where the finger is after swiping


michaldudak (approved):

colmtuite:
legend @atomiks

PR #1787, opened 2025-04-27
[useAnchorPositioning] Fix function dependency handling

### LINKED ISSUE #1353: [menu] Resetting the anchor doesn't reattach the menu to its trigger
When the Menu.Positioner's anchor prop is set to a custom element and then reset to undefined, the popup appears in the corner of the viewport instead of being attached to the trigger.

Reproduction steps:
1. Open https://deploy-preview-1330--base-ui.netlify.app/experiments/menu-fully-featured
2. Open settings and check "Custom anchor"
3. Open and close the menu
4. Uncheck "Custom anchor"
5. Open the menu - it will not appear next to the trigger, as it should 

https://github.com/user-attachments/assets/b90dab29-bfe3-4fe2-9659-253eacea5c31

atomiks:
- Fixes #1353
- Fixes passing a non-memo'd function as the `anchor` causing a loop


michaldudak (approved):
Nice!

PR #1791, opened 2025-04-28
[code-infra] Add bundle size monitor

Janpot:
Porting the bundle size monitor from core. This establishes a baseline. This bundles each entrypoint with its dependencies. Bundles are stored in S3 for comparison. Peer dependencies are ignored. To inspect the created bundles run

```
pnpm size:why
```

examples:

Screenshot 2025-05-09 at 10 06 31

Screenshot 2025-05-15 at 15 01 28


packages/react/scripts/createPackageManifest.mts
```diff
@@ -40,7 +40,10 @@ export async function createPackageManifest() {
         '*': ['./cjs/*/index.d.ts'],
       },
     },
-    exports: retargetExports(exports),
+    exports: {
+      './package.json': './package.json',
```
Janpot, line 44: commonly used export to advertise the package root to tooling

michaldudak:
These changes look good. However, I noticed that the icons in the bundle size report look unreadable in light mode:

image

bundle-size/bundle-size-checker.config.mjs
michaldudak, line 1: Can this be placed under `scripts/bundle-size`?

dangerFileContent.ts
michaldudak, line 1: I wouldn't mind hiding this under `scripts` as well to reduce the number of files in the repo root.

Janpot:
> I noticed that the icons in the bundle size report look unreadable in light mode:

Will change it to something that works better for both themes:

Screenshot 2025-05-19 at 16 20 48

Screenshot 2025-05-19 at 16 21 27

oliviertassinari:
For #201

bundle-size/bundle-size-checker.config.mjs
Janpot: you mean the whole `bundle-size` workspace? The workspace and the config file belong together because the bundles need to be generated from a folder where it can resolve to the built base ui node module, linked from its workspace in the node_modules folder of this workspace.

Janpot:
Oh, and mind you that I updated it to allow for arbitrary code if you want to check something custom:

```tsx
// bundle-size/bundle-size-checker.config.mjs

      {
        id: 'custom-entry',
        code: `
          import { Field } from '@base-ui-components/react/field';
          console.log(Field.Root, Field.Label);
        `,
        externals,
      },
```
(Doesn't accept jsx yet I just noticed, but will [look](https://github.com/mui/mui-public/pull/335) into that)

bundle-size/bundle-size-checker.config.mjs
```diff
@@ -0,0 +1,41 @@
...
+  const entrypoints = Object.keys(exports).map((exportKey) => {
+    // Convert from "./accordion" to "@base-ui-components/react/accordion"
+    const entrypoint =
+      exportKey === '.'
+        ? '@base-ui-components/react'
+        : `@base-ui-components/react${exportKey.slice(1)}`;
+    return {
+      id: entrypoint,
+      import: entrypoint,
+      externals,
```
Janpot, line 33: This is unnecessary if `@base-ui-components/react` adds an export condition for `./package.json`. The you can just return `entrypoint` from this function and the bundle size monitor will infer the externals from the peer dependencies. Right now it has no real way to find the package.json location reliably for a module. (besides using a custom resolver)

bundle-size/bundle-size-checker.config.mjs
Janpot: Sure, also happy to put it under test/bundle-size, then we won't even have to declare a new workspace in pnpm-workspace.yaml

bundle-size/bundle-size-checker.config.mjs
michaldudak: I don't mind exporting package.json. Can you do it in this PR or prefer me to open a new one?

Janpot: had it before in this PR, will bring it back.

michaldudak:
I checked https://bundlephobia.com/package/@base-ui-components/react@1.0.0-alpha.8 and it shows drastically different values. Do you have any idea where this difference might come from?
One reason could be measuring a different version (alpha.8 vs. the latest master), but it's hard to believe we almost doubled the library size since the latest release.

Janpot:
bundlejs also [measures](https://bundlejs.com/?q=%40base-ui-components%2Freact&treeshake=%5B*%5D)  around 100KB gzipped

```
Bundle size is 316 kB -> 100 kB (gzip)
```

In general we should be wary of bundlephobia stats. They for instance also [measure](https://bundlephobia.com/package/@mui/x-data-grid@8.4.0) the data grid at 47B.

michaldudak (approved):
All right, let's see it in action.

Janpot:
Also working on an [experimental vite version](https://github.com/mui/mui-public/pull/337) of the bundle size monitor, it shows similar results to the webpack one within a tolerance of ~1KB.

bundlephobia is built with [`package-build-stats`](https://www.npmjs.com/package/package-build-stats). Looks like they also do a webpack build and analyze the stats.json, with externals from peer dependencies. Haven't looked any deeper yet how they calculate their sizes.

PR #1795, opened 2025-04-29
[docs] Link `h2` and `h3` in docs

mj12albert:
Example: https://deploy-preview-1795--base-ui.netlify.app/react/components/checkbox#api-reference

Closes https://github.com/mui/base-ui/issues/1181


atomiks:
Think the icon should be revealed on focus as well

colmtuite:
@mj12albert 
- remove the icon
- Do `text-decoration-color: currentColor`
- Do `text-underline-offset: 4px;`

mj12albert:
> * remove the icon
> * Do `text-decoration-color: currentColor`
> * Do `text-underline-offset: 4px;`

Updated! @colmtuite

colmtuite:
@mj12albert We need responsive scroll-margin-top now, because the mobile header is 48px, but scroll-margin is just 24px. So gotta set it for mobile-first, then tweak it back to 24px for larger screens.

oliviertassinari (commented):
In the execution, could we also bring this back from Material UI https://github.com/mui/material-ui/pull/41994? I frequently select the header text and using a `<a>` break this flow

colmtuite:
@oliviertassinari Wouldn't that break native <a> functionality of dragging an `<a>` into the browser search bar? What's the use case for selecting a range inside a heading?

oliviertassinari:
> Wouldn't that break native functionality of dragging an <a> into the browser search bar?

@colmtuite  It does (I never use this). If we can only have the ability to double-click to select the whole text, it would help, but I have a workaround (selecting text, starting outside of the element, so it's kind of ok).

> What's the use case for selecting a range inside a heading?

I think the use case is maintainer-related, not about end-users. I usually use this to leave comments on the heading structure, propose changes.

mj12albert:
Updated scroll-margin-top: https://deploy-preview-1795--base-ui.netlify.app/react/components/checkbox#api-reference @colmtuite

colmtuite:
@mj12albert On mobile now when I click a heading, it's touching the header/appbar.

mj12albert:
> @mj12albert On mobile now when I click a heading, it's touching the header/appbar.

@colmtuite Fixed: https://deploy-preview-1795--base-ui.netlify.app/react/components/checkbox#api-reference

image

colmtuite (approved):

Issue #1800, opened 2025-04-29
[docs] "New" badge for new components


colmtuite:
In the sidebar, give new components a "New" badge.

Probably best achieved by adding `display: flex; align-items: center; gap: 8px;` to `.SideNavLink`. Then adding a `<span class="Badge">` with `color: var(--color-red); line-height: 1; user-select: none;`.

These px value suggestions are for computed values.

This doesn't have to be magic, like on an "x amount of days" timer, or automatically removed with the next release or anything. We can probably just handle adding/removing manually.

<img width="717" alt="Image" src="https://github.com/user-attachments/assets/2364123a-e0fe-4390-9a1a-7c5816951d98" />

atomiks:
This design seems a bit odd to me, I think the text should be smaller, wrapped in a rounded border or background, and maybe be all uppercase?

colmtuite:
Prob too fancy, design style should be minimal af. Will tweak it in the PR though

PR #1811, opened 2025-04-30
[docs] Add `SideNav.Badge` to SideNav

mj12albert:
Preview: https://deploy-preview-1811--base-ui.netlify.app/react/components/meter

Closes https://github.com/mui/base-ui/issues/1800


docs/src/components/SideNav.tsx
```diff
@@ -132,7 +141,7 @@ export function Item({ children, className, href, ...props }: ItemProps) {
           }
         }}
       >
-        <span className="SideNavLinkInner">{children}</span>
```
mj12albert, line 135: Renamed `SideNavLinkInner` to `SideNavLabel` since it only holds the text label

atomiks:
@colmtuite something like this feels better to me

Screenshot 2025-04-30 at 4 19 13 pm

mj12albert:
Or maybe this:
Screenshot 2025-04-30 at 2 26 19 PM

michaldudak:
James' suggestion looks the best to me. The current version is very weird as it's shifted 1px above the baseline:
image

docs/src/components/SideNav.tsx
colmtuite: Should change it back. `Inner` is better since it's abstract and can support future content.

colmtuite:
Any chance you can do the improve typography issue before this one? It will impact the vibe of this one a lot.

mj12albert:
new-badges

I rebased with the typography fixes, and added all the suggestions to the [preview](https://deploy-preview-1811--base-ui.netlify.app/react/components/input) @colmtuite @atomiks @michaldudak 

With the "corner" alignment, I think that when the item is active, the badge seems too close to the top border, and there's not enough room to both space it out more and offset it's original position (to be in the corner) at the same time

Also I find that on a crappier screen this font doesn't look as good in all caps, even with the maximum available font weight (600) it looks too close to the normal casing

BTW `line-height: inherit` instead of `1` seems to fix the baseline misalignment

atomiks:
The NEW on `Input` but with `0.7rem` (instead of `0.75rem)` + `500` weight looks best to me. I don't think we should optimize for low PPI displays though

colmtuite:
Ok, happy to go with the consensus. @mj12albert can you make the following updates please:

- Decrease badge font size to 11px computed.
- Change badge font-weight to 500.
- Increase badge letter-spacing to `.035em`, or something similar (whatever we have available that is close).
- Decrease the link's `gap` to 4px.
- Add 2px horizontal padding to the badge.
- Remove the horizontal translate from the badge, and decrease the vertical translate to `-2px`.

mj12albert:
Updated:

IMG_0225
IMG_0226

docs/src/components/SideNav.tsx
```diff
@@ -68,9 +68,18 @@ export function List({ className, ...props }: React.ComponentProps<'ul'>) {
...
 }
 
+export function Label({ className, ...props }: React.ComponentProps<'span'>) {
+  return <span className={clsx('SideNavLabel', className)} {...props} />;
+}
+
+export function Badge({ className, ...props }: React.ComponentProps<'span'>) {
+  return <span className={clsx('SideNavBadge', className)} {...props} />;
+}
+
```
romgrk, line 78: In line with the style/perf changes I've been doing, I'd update the style for these to this more performant and terser version:

```suggestion
export function Label(props: React.ComponentProps<'span'>) {
  return <span {...props} className={clsx('SideNavLabel', props.className)} />;
}

export function Badge(props: React.ComponentProps<'span'>) {
  return <span {...props} className={clsx('SideNavBadge', props.className)} />;
}

```

atomiks (approved):

docs/src/components/SideNav.tsx
mj12albert: Thanks updated ~ I suppose we should do this for all the docs components then

PR #1812, opened 2025-04-30
[useAnchorPositioning] `side`/`align` callbacks

### LINKED ISSUE #1769: [Positioners] Callback values for `side`/`align`
# Feature request

Mentioned in https://github.com/mui/base-ui/pull/1665#issuecomment-2824141973

## Summary

You sometimes want to change the `side` or `align` based on the viewport size (e.g. `right`/`left` sides never really work on mobile screens - you need to use a `useMediaQuery` hook to change it to `top`/`bottom`) or interaction type (e.g. context menus should be positioned on top, not bottom, to avoid the fingers obscuring it).

The `fallbackAxisSideDirection` option for `flip()` in use Floating UI is supposed to solve this by default for the viewport size scenario, but it has some issues with [undesirable repositioning while scrolling](https://github.com/mui/base-ui/pull/817), although this setting might still suffice for nested menus (positioned on the `right` for `Menu` and `NavigationMenu`) on narrow viewports

`sideOffset` and `alignOffset` take callbacks with the side/align/dimensions, so `side` and `align` taking a callback with the interaction type and viewport size would be consistent and useful.

atomiks:
Closes #1769

### DIFF
--- a/packages/react/src/utils/useAnchorPositioning.ts
+++ b/packages/react/src/utils/useAnchorPositioning.ts
@@ -23,9 +23,32 @@ import { useModernLayoutEffect } from './useModernLayoutEffect';
 import { useDirection } from '../direction-provider/DirectionContext';
 import { useLatestRef } from './useLatestRef';
 import { useEventCallback } from './useEventCallback';
+import { InteractionType } from './useEnhancedClickHandler';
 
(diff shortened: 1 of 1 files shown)


atomiks:
Replaced by #1849. For `interactionType` I'll work on a new PR separately

PR #1840, opened 2025-05-01
[Field] Revalidate only `required` on change

atomiks:
Addresses https://github.com/mui/base-ui/pull/1755#pullrequestreview-2807036034


mj12albert:
When valueMissing is the first error triggered, it doesn't get cleared until all other rules become valid:

https://github.com/user-attachments/assets/c751d88d-2f0b-4f16-9d41-a1a2f7c99a49

atomiks:
@mj12albert this case feels a little ambiguous when there are two+ errors simultaneously. 

1. If `required` is cleared and no errors appear while typing, then it implies that there are no other errors (they only see `typeMismatch` on blur)
2. If we validate the `required` on change, it "scolds" them with `typeMismatch` only once, then doesn't again, which is strange(?)
3. Currently, it goes away when both errors are gone even though it doesn't live-update the first error (linked to point no. 2), as you point out

I suppose an alternative might be: clear on change and claim it's valid even though it's not (due to `typeMismatch`), then on blur, it will show `typeMismatch`. But this is difficult to get working - the logic is a lot simpler to handle with how it's done here and it's not clear if that is actually better

atomiks:
@mj12albert I implemented the alternative suggestion I wrote, the behavior seems more expected

mj12albert:
> @mj12albert I implemented the alternative suggestion I wrote, the behavior seems more expected

I tested this again (and prior to this change as well) and it works great now ~ also tried to collect some good examples in my actual web use but most aren't good for one reason or another...

And in a realistic use-case for inputs with complicated validation rules is that the requirements should primarily be communicated via `Field.Description` and not via error messages, so in practice being scolded too early isn't _that_ bad. When the valueMissing error is still showing even though there's some input its very noticeable and feels more wrong 😅

mj12albert:
@atomiks When using NumberField, the issue with the validate function receiving an outdated value returned 🤔 

https://codesandbox.io/p/sandbox/wonderful-sunset-jd598f?file=%2Fsrc%2FApp.tsx

https://github.com/user-attachments/assets/a2d787b6-20b6-4160-b383-2186db70d3b4

atomiks:
@mj12albert updated

Can't add a test for this due to how `event.isTrusted` is used to set the input value, but blocking revalidation after blur looks to work as that's when the `value` changes and revalidation shouldn't occur in the effect (only for `inputValue`)

mj12albert (approved):
Works very nicely now 🙌

Issue #1726, opened 2025-04-16
Polyfill CSS `:focus-visible` and `:active` states manually


atomiks:
# Bug report

### `:focus-visible`

The way browsers handle `:focus-visible` is annoying because it requires the pointer to be used on an element that doesn't `preventDefault` the event. As a result, the focus-visible outline can appear on elements when using the pointer if nothing else was interacted with to begin with.

https://github.com/user-attachments/assets/44842eba-94c3-4ad1-83be-486c3a0600c8

### `:active`

The CSS `:active` state doesn't work when popups are triggered by `mousedown`. This should be polyfillable so it actually does still work, though not sure if that is fully desirable https://github.com/mui/base-ui/issues/1720

oliviertassinari:
So the focus-visible polyfill https://github.com/mui/material-ui/pull/42467 was removed prematurely? Related issues:

- https://github.com/w3c/csswg-drafts/issues/5885#issuecomment-2227138661
- https://github.com/radix-ui/primitives/issues/2953#issuecomment-2497463101

OliverJAsh:
👋 We're also running into this problem at Unsplash. Is there any workaround?

https://github.com/user-attachments/assets/e53435e8-02cb-4a23-a763-05554f28ed11

atomiks:
@OliverJAsh opened https://github.com/mui/base-ui/pull/1846

atomiks:
#1846 should solve most cases of the native `:focus-visible` CSS state appearing undesirably.

There still remains these ones:
-  `:active` not working when the popup opens on `mousedown`
- `:hover` appearing on hybrid devices (which `@media (hover: hover)` doesn't address)
- A [generic `pressed` state](https://github.com/mui/base-ui/issues/1747) that works across mouse, keyboard, touch input modalities

atomiks:
- `:active` -> `[data-popup-open]` flickers on iOS: https://ff6hgj.csb.app/
- `[data-highlighted]` should likely not appear when using touch input

oliviertassinari:
> `@media (hover: hover)`

We did a good dive on this problem in https://github.com/mui/mui-x/issues/10039 in 2023. This media selector was not implemented correctly in many platforms.

PR #1846, opened 2025-05-01
[Select][Menu] Avoid `:focus-visible` style appearing

atomiks:
This one avoids calling `event.preventDefault()` in `onMouseDown`, which causes the `:focus-visible` bug in #1726. Since the focus manager is async it doesn't cause focus to be stolen. Also, we can remove a bunch of unused logic that `useButton` already handles.

I added a new folder `utils/floating-ui/` so we can start writing custom hooks that better supports what we need until Floating UI's API is refined in v1. I exported a lot of utils in 0.27.8 specifically for this purpose.

Preview: https://deploy-preview-1846--base-ui.netlify.app/react/components/menu
Clicking either the CSS Modules Select or the Menu demo with your pointer won't cause `:focus-visible` style upon closing either one.


packages/react/src/utils/floating-ui/useClick.ts
```diff
@@ -0,0 +1,122 @@
...
+          toggle &&
+          (dataRef.current.openEvent && stickIfOpen
+            ? dataRef.current.openEvent.type === 'mousedown'
+            : true)
+        );
+        // Wait until focus is set on the element. This is an alternative to
+        // `event.preventDefault()` to avoid :focus-visible from appearing when using a pointer.
+        frameRef.current = requestAnimationFrame(() => {
+          onOpenChange(nextOpen, event.nativeEvent, 'click');
+        });
```
atomiks, line 94: I don't like adding async behavior as it makes testing harder, but this fixes the issue

OliverJAsh:
Will this also fix issues with focus rings appearing around menus?

https://github.com/user-attachments/assets/7a2b8e05-2d2a-4faf-ae82-00ccef2edb87

atomiks:
@OliverJAsh yes it should since the browser checks the first pointer down to determine the `:focus-visible` state. Though the docs use `outline: 0` on the menu popup since an item is always focused with keyboard navigation

packages/react/src/utils/floating-ui/useClick.ts
```diff
@@ -0,0 +1,126 @@
...
+        onOpenChange(nextOpen, event.nativeEvent, 'click');
+      },
+      onKeyDown() {
+        pointerTypeRef.current = undefined;
+      },
+    }),
+    [dataRef, eventOption, ignoreMouse, onOpenChange, open, stickIfOpen, toggle],
+  );
+
+  return React.useMemo(() => (enabled ? { reference } : {}), [enabled, reference]);
```
michaldudak, line 125: Return a stable (const) object instead of `{}` to avoid recreating the return value when `enabled=false` and `reference` changing.

michaldudak (commented):
Nice, it works much better!

One issue I can still see is if you focus the trigger with keyboard (so the outline appears), then click on it, then select an item, the outline is reapplied, even though mouse was used to interact.

atomiks:
> One issue I can still see is if you focus the trigger with keyboard (so the outline appears), then click on it, then select an item, the outline is reapplied, even though mouse was used to interact.

This might be the browser's behavior - but I think it actually is fine/makes sense given it had focus-visible initially and has focus returned to it afterwards. 

If you have focus on another element, click with pointer to open, then click with pointer to close, there's no outline on the trigger.

michaldudak:
> This might be the browser's behavior - but I think it actually is fine/makes sense given it had focus-visible initially and has focus returned to it afterwards.

I don't think there's a definitive answer to this. I'd say that since the last interaction was done using a mouse, the outline shouldn't be applied, but it doesn't bother me much if it is.

michaldudak (approved):

OliverJAsh:
Would it be possible to cut a release for this? We really need it 🤞 

I tried using the Code Sandbox package created by this PR, but for some reason the bundle size increased a lot. Assuming that won't happen for a proper release.

atomiks:
We set up pkg.pr.new canary releases: https://pkg.pr.new/view/mui/base-ui. These should be the same as the final release

OliverJAsh:
All of the canary releases are ~18 KB (gzip) bigger than 1.0.0-alpha.8. Any ideas why that is?

atomiks:
18 KB bigger how? The whole package or just when using specific component(s)? The library is tree-shakeable and testing the latest master commit package in a Vite project shows e.g. `Menu` treeshakes the same as it did in `alpha.8` and is actually smaller.

```js
"use client";
import { Menu } from "@base-ui-components/react/menu";

  return (
    <Menu.Root>
      <Menu.Trigger>Open</Menu.Trigger>
      <Menu.Portal>
        <Menu.Positioner>
          <Menu.Popup>
            <Menu.Item />
          </Menu.Popup>
        </Menu.Positioner>
      </Menu.Portal>
    </Menu.Root>
  );
}
```

`https://pkg.pr.new/@base-ui-components/react@04c08fb`

> dist/assets/index-DI-xZtX5.js   243.89 kB │ gzip: 81.16 kB

`1.0.0-alpha.8`

> dist/assets/index-CLkWVFjf.js   247.07 kB │ gzip: 82.14 kB

OliverJAsh:
Sorry, please disregard. It looks like my bundler just decided to shuffle things into different chunks by coincidence.

OliverJAsh:
This might be an edge case but I'm noticing that `:focus-visible` styles still appear when pressing escape to close a menu and then _using the mouse_ to re-open the menu. Is that expected?


https://github.com/user-attachments/assets/2ac01b28-d7b2-4ae2-adac-3d0904020742

atomiks:
Maybe a limitation of the browser's implementation of `:focus-visible`? Most basic vanilla example: https://codesandbox.io/p/sandbox/kg6vdg

OliverJAsh:
Yeah looks like it! Sorry to bother you but I think I may have found another instance of this. On iOS, using [the official popover example only modified to use the latest canary release](https://codesandbox.io/p/sandbox/elated-microservice-d3qvyj?file=%2Fpackage.json%3A3%2C89):


https://github.com/user-attachments/assets/ffc5131f-334d-4b19-907f-0dd0c9695a64


Furthermore, when using tooltips, the tooltip is also shown when the popover is closed:

https://codesandbox.io/p/devbox/silly-mountain-qg4lqv?file=%2Fsrc%2FApp.tsx&workspaceId=ws_4qKVDKGuRrrvd8oLfoRWSz


https://github.com/user-attachments/assets/6f9af6e0-f77f-4cf1-9e47-96914e7fbe4a

atomiks:
Does it happen if you add `tabIndex={0}` to the `<button>` (`Trigger` part)? I think the button needs to receive focus on click in Safari which only happens with that attribute (in the latest versions?) to prevent the `:focus-visible` style.

This doesn't seem to repro for me on iOS 18.4 unlike your CodeSandbox
https://687j7c.csb.app/

We add `tabIndex={0}` to some buttons already for this purpose, so likely we'll need to do it everywhere for Safari

OliverJAsh:
> This doesn't seem to repro for me on iOS 18.4 unlike your CodeSandbox
> https://687j7c.csb.app/

Yeah, that seems to fix it! That also somehow fixes the issue with the tooltip appearing.

> We add `tabIndex={0}` to some buttons already for this purpose, so likely we'll need to do it everywhere for Safari

That sounds reasonable to me.

OliverJAsh:
I've also just noticed that—whilst this PR fixes the original issue for menus ([reduced test case](https://codesandbox.io/p/sandbox/vigorous-dust-ffvvsg?file=%2Fpackage.json))—popovers still show `:focus-visible` styles when they're opened on iOS ([reduced test case](https://codesandbox.io/p/sandbox/elated-microservice-d3qvyj?file=%2Fpackage.json)):

https://github.com/user-attachments/assets/0a0e1cd8-8faf-4603-a14d-edc77525df7d

atomiks:
Isn't that just the `tabIndex={0}` fix?

OliverJAsh:
Ah yes, seems so! Just to check, you plan to add that to the library? Or is it a workaround to end users need to adopt themselves?

PR #1847, opened 2025-05-02
[menu][Select] Ensure `null` items are removed from composite lists

### LINKED ISSUE #1842: [menu] Changing the rendered items after the menu is open breaks keyboard navigation
# Bug report

## Current behavior

In my app, I have a menu with a search input that allows filtering the items inside it. I managed to hack the focus management, but I'm running into an issue where navigating through the filtered items with the keyboard seems to consider all the items that were present when the menu was first opened. In other words, it seems to allow me to highlight items that are no longer rendered; maybe it should "refresh" the items every time the up or down arrow key is pressed? As a workaround, I re-mount the menu every time the search value changes (with the `key` prop), which I'm sure takes a considerable hit on performance.

## Expected behavior

Only the menu items that are currently rendered should be focusable/highlightable with the keyboard.

## Reproducible example

https://codesandbox.io/p/sandbox/broken-hill-4cmpp6

## Base UI version

1.0.0-alpha.8

Chrome, Safari, and Firefox

## Which OS are you using?

macOS

None

atomiks:
Closes #1842

### DIFF
--- a/packages/react/src/menu/root/useMenuRoot.ts
+++ b/packages/react/src/menu/root/useMenuRoot.ts
@@ -208,7 +208,7 @@ export function useMenuRoot(parameters: useMenuRoot.Parameters): useMenuRoot.Ret
     loop,
     orientation,
     rtl: direction === 'rtl',
-    disabledIndices: (index) => (itemDomElements.current[index] == null ? [index] : EMPTY_ARRAY),
+    disabledIndices: EMPTY_ARRAY,
     onNavigate: setActiveIndex,
   });
 
--- a/packages/react/src/menu/root/useMenuRoot.ts
+++ b/packages/react/src/menu/root/useMenuRoot.ts
@@ -208,7 +208,7 @@ export function useMenuRoot(parameters: useMenuRoot.Parameters): useMenuRoot.Ret
     loop,
     orientation,
     rtl: direction === 'rtl',
-    disabledIndices: EMPTY_ARRAY,
+    disabledIndices: (index) => (itemDomElements.current[index] == null ? [index] : EMPTY_ARRAY),
     onNavigate: setActiveIndex,
   });
 
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -83,7 +83,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.27.0",
-    "@floating-ui/react": "^0.27.8",
+    "@floating-ui/react": "https://pkg.pr.new/@floating-ui/react@3311",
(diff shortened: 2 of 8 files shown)


packages/react/src/menu/root/useMenuRoot.ts
```diff
@@ -208,7 +208,7 @@ export function useMenuRoot(parameters: useMenuRoot.Parameters): useMenuRoot.Ret
     loop,
     orientation,
     rtl: direction === 'rtl',
-    disabledIndices: EMPTY_ARRAY,
+    disabledIndices: (index) => (itemDomElements.current[index] == null ? [index] : EMPTY_ARRAY),
```
atomiks, line 211: I think it's better to leave the `null` items in the array like

```jsx
[element, element, element, null, null, null]
```

Since trying to slice the array to remove them is a big perf hit when an item is removed (similar to the `key={}` workaround mentioned in the issue), but memory-wise, this shouldn't be a problem

atomiks: The other alternative is to guard the ref callback not to add a `null` node

```jsx
  const ref = React.useCallback(
    (node: HTMLElement | null) => {
      componentRef.current = node;

      if (index !== null && node !== null) { // <----
        elementsRef.current[index] = node;
        if (labelsRef) {
          const isLabelDefined = label !== undefined;
          labelsRef.current[index] = isLabelDefined ? label : (node?.textContent ?? null);
        }
      }
    },
    [index, elementsRef, labelsRef, label],
  );
```

Then when the composite map gets sorted:

```jsx
useModernLayoutEffect(() => {
  elementsRef.current.length = sortedMap.size; // <----
  labelsRef.current.length = sortedMap.size; // <----
  onMapChange?.(sortedMap);
}, [sortedMap, onMapChange, elementsRef]);
```

michaldudak:
When a Select is open and the currently highlighted item is removed, the keyboard navigation doesn't work anymore until another item is hovered.

Try https://codesandbox.io/p/sandbox/angry-thunder-qq76c7?file=%2Fsrc%2FApp.tsx. The list changes every 5s. If you open the Select and highlight "two", on the next change keyboard navigation will be broken

atomiks:
@michaldudak A lot of other stuff also breaks when it's done while open, like the positioning and scroll arrows. I guess that can be tackled in a separate PR (assuming it's also broken on main)

michaldudak:
Sure, I'm OK with that.

michaldudak (approved):

PR #1849, opened 2025-05-02
[useAnchorPositioning] Add `collisionAvoidance` prop

### LINKED ISSUE #1546: [Popover / Positioner] Add prop to customize whether to flip or to shift when there's no more space
# Feature request

## Summary

Again, this request is for `Popover.Positioner`, but it probably applies to the other `Positioner` components as well.

I noticed that when the popover reaches `collisionPadding` distance or less from the nearest `collisionBoundary`, it flips if there's enough space on the other side. That's nice, but it's not always the desired behavior. In my case, I would prefer if the popover was shifted instead, as I would rather keep the trigger as close as possible to the left side of the popover:

https://github.com/user-attachments/assets/1d5db707-4807-4eba-bf5c-151a4cbda522

Would it be possible to add a new prop to customize that behavior? Maybe something like `collisionAvoidanceStrategy?: 'flip' | 'shift'`?

I believe `@floating-ui` itself has a few different strategies for avoiding collisions.

## Motivation

Sorry, I don't know what to add that I haven't said above. :D Thank you very much!

atomiks:
Closes #1546

New options in the demo: https://deploy-preview-1849--base-ui.netlify.app/experiments/anchor-positioning

The `collisionAvoidance` prop allows users to control the collision _preference_. The default setting is:

```js
{
  side: 'flip',
  align: 'flip',
  fallbackAxisSide: 'end',
}
```

This means: 

- For `side` collision avoidance, prefer to flip e.g. from `bottom` to `top`.
- For `align` collision avoidance, prefer to flip e.g. from `end` to `start`.
- If the chosen side axis doesn't fit (e.g. `left` / `right` on narrow viewports), choose the logical `end` side. So if you specify `side="left"`, there's no need to worry about it overflowing on narrow viewports since it will flip to `bottom` (`end`) by default. `@floating-ui/core@1.7.0` solves the viewport size scenario in #1769 when you specify `crossAxis: 'alignment'`, which works better if `shift()` is being used.

If you specify: 

```js
{
  align: 'shift',
}
```

That solves #1546, since it will shift on the align axis instead.

If you want `Select` in the `alignItemWithTrigger=false` mode to work like Material UI's, then you can specify:

```js
{
  side: 'shift',
}
```

So it won't flip, but rather overlap the trigger


atomiks:
This config object has one impossible state:

```js
{
  side: 'shift',
  align: 'flip',
}
```

`align` can't `flip` when `side` prefers `shift`, though the reverse is possible (`align=shift`, `side=flip`). I don't think that setting is useful, but the API shouldn't allow such states

---

A discriminated union can handle this by preventing some strings from being set in combination with others:

```tsx
interface SideFlipMode {
  side?: 'flip' | 'none';
  align?: 'flip' | 'shift' | 'none';
  fallbackAxisSide?: 'start' | 'end' | 'none';
}

interface SideShiftMode {
  side?: 'shift' | 'none';
  align?: 'shift' | 'none';
  fallbackAxisSide?: 'start' | 'end' | 'none';
}

export type CollisionAvoidance = SideFlipMode | SideShiftMode;
```

docs/reference/generated/menu-positioner.json
```diff
@@ -2,6 +2,10 @@
   "name": "MenuPositioner",
   "description": "Positions the menu popup against the trigger.\nRenders a `<div>` element.",
   "props": {
+    "collisionAvoidance": {
+      "type": "CollisionAvoidance",
+      "description": "Determines how to handle collisions when positioning the popup."
+    },
```
atomiks, line 8: The type is too big to display in our current table design, but will work with the new Accordion-style one

benface:
@atomiks – I finally got to test this and it's more than I could ask for, thank you so much! 🫶 Only issue is the docs don't specify the default:

CleanShot 2025-05-29 at 21 22 32@2x

atomiks:
@benface we're overhauling the props tables to be way more user friendly soon, so this will be fixed (it's not possible to display long defaults/types/examples etc in the current setup.)

PR #1851, opened 2025-05-02
[menu] Add missing `'use client'` to `RadioGroup` part

### LINKED ISSUE #1841: `Menu` and probably other components can't be used in server components without marking as `use client` (NextJS)
# Bug report

## Current behavior

When I use Menu components in a server component, my page crashes because the Menu tries to use `useContext`. I have to mark my whole parent component as a client component which is annoying if I am using server-component only behaviour like async fetches. (As a workaround I created a separate file marked 'use client' and re-exported all of the Menu components, which fixed it).

I am a bit confused why this is happening as in the source code I do see `use client`. I am not sure if this is a bundling issue or a NextJS issue, but thought I'd start here first in case I missed something.

```
 ⨯ TypeError: createContext only works in Client Components. Add the "use client" directive at the top of the file to use it. Read more: https://nextjs.org/docs/messages/context-in-server-component
    at __webpack_require__ (.next/server/webpack-runtime.js:33:42)
    at __webpack_require__ (.next/server/webpack-runtime.js:33:42)
    at __webpack_require__ (.next/server/webpack-runtime.js:33:42)
    at __webpack_require__ (.next/server/webpack-runtime.js:33:42)
    at eval (webpack-internal:///(rsc)/./app/page.tsx:9:88)
    at <unknown> (rsc)/./app/page.tsx (/project/workspace/.next/server/app/page.js:33:1)
    at Function.__webpack_require__ (.next/server/webpack-runtime.js:33:42) {
  page: '/'
}
```

## Expected behavior

`Menu.xxxx` components can be imported because they themselves are client components.

## Reproducible example

https://codesandbox.io/p/devbox/3kdwz2

## Base UI version

`1.0.0-alpha.8`

Chrome

## Which OS are you using?

Mac

Voiceover/JAWS etc.

## Additional context

atomiks:
I confirmed that if even one re-exported component is missing `'use client'`, the entire export breaks, even if you don't use that part.

Fixes #1841. We could consider marking the index files with `'use client'` again instead, unless granularly choosing which components to mark is better still.

### DIFF
--- a/packages/react/src/menu/radio-group/MenuRadioGroup.tsx
+++ b/packages/react/src/menu/radio-group/MenuRadioGroup.tsx
@@ -1,3 +1,4 @@
+'use client';
 import * as React from 'react';
 import { MenuRadioGroupContext } from './MenuRadioGroupContext';
 import { BaseUIComponentProps } from '../../utils/types';


atomiks:
I went and tested every component we had and found a few others:

- `Toolbar`
- `Toggle`
- `Toast`

Will update these as well

atomiks:
I went and tested every component we have and only `Menu.RadioGroup` is missing it. This also breaks the barrel file imports for any other component.

michaldudak (approved):

PR #1858, opened 2025-05-04
[Form] Select inputs on focus

### LINKED ISSUE #1068: [number field][form] Select input value when returning focus
Number Field focuses the input when increment/decrement is clicked
Form (?) focuses the invalid input when it fails validation, maintaining the previous text cursor/selection position—which is odd/random

The UX would be better if in both cases, the inputs were selected instead of just being focused

https://base-ui.com/react/components/form
https://base-ui.com/react/components/number-field

**Search keywords**:

atomiks:
Closes #1068

This doesn't apply to `NumberField` since the caret should be at the end of the input when it's incrementing/decrementing, as selecting the entire field is obscuring. It also doesn't select `<textarea>` as preserving caret location may be important here

Also ensures that `errors` is defined for the Form `clear` function

### DIFF
--- a/packages/react/src/form/Form.tsx
+++ b/packages/react/src/form/Form.tsx
@@ -33,6 +33,13 @@ export const Form = React.forwardRef(function Form(
   const onSubmit = useEventCallback(onSubmitProp);
   const onClearErrors = useEventCallback(onClearErrorsProp);
 
+  const focusControl = useEventCallback((control: HTMLElement) => {
+    control.focus();
+    if (control.tagName === 'INPUT') {
+      (control as HTMLInputElement).select();
+    }
+  });
+
   React.useEffect(() => {
     if (!submittedRef.current) {
       return;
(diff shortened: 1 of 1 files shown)


mj12albert (approved):
Works well for textboxes, I think other controls like switch etc could also support focus-if-first-error when possible (not sure if it will play well with radio group and activateOnFocus?)

atomiks:
All controls already receive focus if they're first, this just selects the text inside them if a textbox input

mj12albert:
> All controls already receive focus if they're first, this just selects the text inside them if a textbox input

Oh you are right, I was testing in https://master--base-ui.netlify.app/experiments/form and didn't see because the styles use :focus-visible

https://github.com/user-attachments/assets/0d792e0e-c91d-4c4b-8667-81aa412e06b1

mj12albert:
> All controls already receive focus if they're first, this just selects the text inside them if a textbox input

Oh you are right, I was testing in https://master--base-ui.netlify.app/experiments/form and didn't see because the styles use :focus-visible

https://github.com/user-attachments/assets/0d792e0e-c91d-4c4b-8667-81aa412e06b1

PR #1861, opened 2025-05-05
[select][menu] Better handle dynamic and non-string items

### LINKED ISSUE #1331: [select] Controlled components should not change internal values when change handler is missing
# Bug report

## Current behavior

When a state variable is passed to the `value` prop of the Base UI `Select` component, the `Select` menu can still be opened, and a different option can be selected without using the `onValueChange` function. This behavior is unexpected for a controlled component since changes to its value should only occur through the `onValueChange` handler.

## Expected behavior

The `Select` component should function as a controlled component. Changes to the selected value should only be possible when handled via the `onValueChange` function. The `value` prop should strictly determine the displayed selected value.

## Reproducible example

[Link to a CodeSandbox](https://codesandbox.io/p/sandbox/festive-sid-h8syd4?workspaceId=ws_Eork3eV5WBjoZPqsrajEAW)

Steps to reproduce:
1. Pass a state variable to the `value` prop of the `Select` component. (Already done)
2. Do not handle the `onValueChange` function. (Already commented out)
3. Open the `Select` menu and try selecting a different option.
4. Notice that the selected value changes without using `onValueChange`.

## Base UI version

v1.0.0-alpha.5

Chrome, Safari

## Which OS are you using?

Mac OS

## Additional context

This behavior contradicts the expected behavior of controlled components. It might cause inconsistencies in state management when using `Select` in forms or other UI workflows.

### LINKED ISSUE #1853: [Select]: "The value `` is not present in the select items." when using numbers as item values.
# Bug report

## Current behavior

When using number as an Item value, there is a warning "The value `` is not present in the select items.", when selected from the popup.

## Expected behavior

Select should work with numbers.

## Base UI version

1.0.0-alpha.8

Chrome

## Which OS are you using?

Windows

None

### LINKED ISSUE #1922: Select with dynamic items fails to show popup upon update
# Bug report

## Current behavior

I have a setup where the items in one select depend on the value of another. After changing the first select's value, the dependent select updates, but its popup won't show again. In Chrome DevTools, I can see that the popup is rendered in the DOM, but it's not visible in the viewport. It seems like the styles aren't being recalculated after the change.

## Expected behavior

After changing the value of the first select, the dependent select should update its items and the popup should appear correctly.

## Reproducible example

[CodeSandbox](https://codesandbox.io/p/devbox/determined-snowflake-tcg4qw?file=%2Fsrc%2FApp.tsx)

## Base UI version

v1.0.0-alpha.8

Chrome

## Which OS are you using?

Mac OS

atomiks:
- Select: Fixes #1853 to ensure non-string values don't warn, clears styles and scroll arrows state when the item list size changes
- Select: Fixes #1922 to ensure the `selectedItemTextRef.current` is set when the previous text element gets disconnected, fixing the positioning
- Select + Menu: Fixes focus restoration (add `restoreFocus` prop) to ensure keyboard navigation works when item is removed while menu is open
- Select: Fixes error due to `scrollArrow` visibility state changing state of parent component in render by moving it locally
- Select: Fixes #1331 (based on https://github.com/mui/base-ui/issues/1331#issuecomment-2898238781)


packages/react/src/select/positioner/SelectPositioner.tsx
```diff
@@ -114,12 +120,58 @@ export const SelectPositioner = React.forwardRef(function SelectPositioner(
...
   );
 
+  const prevMapSize = React.useRef(0);
+
   return (
-    <CompositeList elementsRef={listRef} labelsRef={labelsRef}>
+    <CompositeList
+      elementsRef={listRef}
+      labelsRef={labelsRef}
+      onMapChange={(map) => {
```
michaldudak, line 143: Use a stable function as `onMapChange` is on `useModernLayoutEffect`'s dependency list in CompositeList.

mj12albert (approved):
Looks good, I tested all except for the scroll arrow issue but the code looks ok

PR #1865, opened 2025-05-05
[code-infra] Set up canary releases

michaldudak:
Publish the canary package on pkg.pr.new

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,4 +37,4 @@ jobs:
       - run: pnpm validate-declarations
       - name: Publish to pkg-pr-new
         if: matrix.os == 'ubuntu-latest'
-        run: pnpm dlx pkg-pr-new publish './packages/react' './packages/dom' './packages/react' --compact --pnpm
+        run: pnpm dlx pkg-pr-new publish './packages/react' --compact --pnpm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,3 +35,5 @@ jobs:
       - run: pnpm install
       - run: pnpm release:build
       - run: pnpm validate-declarations
+      - name: Publish to pkg-pr-new
+        run: pnpm dlx pkg-pr-new publish './packages/react' './packages/dom' './packages/react' --compact --pnpm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,4 +36,5 @@ jobs:
       - run: pnpm release:build
       - run: pnpm validate-declarations
       - name: Publish to pkg-pr-new
+        if: matrix.os == 'ubuntu-latest'
         run: pnpm dlx pkg-pr-new publish './packages/react' './packages/dom' './packages/react' --compact --pnpm


.github/workflows/ci.yml
```diff
@@ -35,3 +35,6 @@ jobs:
       - run: pnpm install
       - run: pnpm release:build
       - run: pnpm validate-declarations
+      - name: Publish to pkg-pr-new
+        if: matrix.os == 'ubuntu-latest'
+        run: pnpm dlx pkg-pr-new publish './packages/react' './packages/dom' './packages/react' --compact --pnpm
```
atomiks, line 40:
```suggestion
        run: pnpm dlx pkg-pr-new publish './packages/react' --compact --pnpm
```

atomiks (approved):

.github/workflows/ci.yml
```diff
@@ -35,3 +35,6 @@ jobs:
       - run: pnpm install
       - run: pnpm release:build
       - run: pnpm validate-declarations
+      - name: Publish to pkg-pr-new
+        if: matrix.os == 'ubuntu-latest'
+        run: pnpm dlx pkg-pr-new publish './packages/react' --compact --pnpm
```
oliviertassinari, line 40: This should match with https://github.com/mui/material-ui/blob/394fa56eb0085a4fd7502447513caa836de8f1e8/package.json#L15, no?

1. We didn't add nocomments, so we get noise: https://github.com/mui/base-ui/pull/1865#issuecomment-2851031174
2. Using - -peerDeps seems to be only win-win: https://github.com/stackblitz-labs/pkg.pr.new/pull/156
3. For the use of compact, https://github.com/stackblitz-labs/pkg.pr.new?tab=readme-ov-file no preference, I assume the other should copy it (seems better?)
4.

Janpot: > 4. The others should use --pnpm as it's more accurate?

Not sure why yet, but `--pnpm` doesn't work correctly in core/X repo,

michaldudak: > We didn't add nocomments

Not sure what others think, but I actually prefer having a comment with the published package URL. It appears shortly after the PR is created, so I don't consider it noise.  But if it's annoying, we can remove it (cc @atomiks).

> Using - -peerDeps seems to be only win-win

:+1: We don't have any workspace peer deps, but I can add this parameter for consistency.

atomiks: I think the comments are useful - one of the reasons we set it up is to allow users to try things out before they're released if necessary, and the comments make it obvious how to do so

oliviertassinari: > one of the reasons we set it up is to allow users to try things out before they're released if necessary, and the comments make it obvious how to do so

I'm missing something. How is https://github.com/mui/material-ui/blob/master/CONTRIBUTING.md#how-can-i-use-a-change-that-hasnt-been-released-yet not better? Meaning,

1. If you want to test something, you open the Netlify preview, open the relevant demo in CodeSandbox, have the pkg used. 
2. If you really need to test something, you go to the GitHub status check, and find the URL to use.

The comment in the PR seems too much, most of the time we don't need it, instead, we want to be focused on the PR comment.

atomiks: It's about visibility and accessibility - those options require many more steps and more knowledge of how continuous releases work in order to use them, particularly for new/unfamiliar users of the library

The `pkr.pr.new` comment is very straightforward and obvious how to install the package containing the PR's changes (in fact, we could change the comment to make this better cc: @michaldudak)

PR #1868, opened 2025-05-05
[Toast] Refactor to `useRenderElement`

atomiks:


packages/react/src/toast/root/useToastRoot.ts
```diff
@@ -453,76 +453,72 @@ export function useToastRoot(props: useToastRoot.Parameters): useToastRoot.Retur
...
-          return {
-            [ToastRootCssVars.swipeMovementX]: '0px',
-            [ToastRootCssVars.swipeMovementY]: '0px',
-          };
-        }
-
-        const deltaX = dragOffset.x - initialTransform.x;
-        const deltaY = dragOffset.y - initialTransform.y;
-
+  const props: GenericHTMLProps = React.useMemo(() => {
```
mj12albert, line 456: Do you keep these (likely non-reusable) hooks because it returns more than just the props?

packages/react/src/toast/provider/ToastProvider.tsx
```diff
@@ -1,7 +1,22 @@
...
+import { generateId } from '../../utils/generateId';
+import { resolvePromiseOptions } from '../utils/resolvePromiseOptions';
+
+interface TimerInfo {
+  timeoutId?: ReturnType<typeof setTimeout>;
+  start: number;
+  delay: number;
+  remaining: number;
+  callback: () => void;
+}
```
romgrk, line 19: Should match the existing `interface` here and use the new `Timeout` class with the guard: https://github.com/mui/base-ui/blob/master/packages/react/src/toast/provider/useToastProvider.ts#L15-L21

romgrk: Not sure what's the relation between this file and that hook file I linked, but there is a lot of repeated code and afaict that hook file is left untouched by this PR.

atomiks: Yes you're right, the changes to timeouts made in the hook weren't applied back into the main component. The hook should also be deleted.

PR #1871, opened 2025-05-06
[composites] Stop event propagation

### LINKED ISSUE #1759: [toolbar] Prevent navigation when arrow keys are pressed on an item in a popup in a toolbar
Currently `Composite` ignores any keypresses from inside popups to prevent toolbar navigation, which doesn't work for Menubar and Navigation Menu, or other composite components inside the popup: https://github.com/mj12albert/base-ui/blob/7ce819e138b5e06ef3657ed500abab2d0efa2bd0/packages/react/src/composite/root/useCompositeRoot.ts#L158-L161

Components that use the same keys (arrow keys, Home+End) should `event.stopPropgation()` in their keydown handlers instead:
- Toggle/Toggle group
- ~Slider~ done
- Radio group
- Number field
- Button/trigger components (useButton?)

atomiks:
Fixes #1759

Triggers preview: https://deploy-preview-1871--base-ui.netlify.app/experiments/toolbar/triggers

- Inside `Dialog`/`AlertDialog`, it prevents composite behavior from escaping itself. `Popover`/anything that doesn't trap focus or close when focus escapes it allows it
- Tabs and any other composite works inside a Dialog/Popover correctly

### DIFF
--- a/packages/react/src/number-field/input/NumberFieldInput.tsx
+++ b/packages/react/src/number-field/input/NumberFieldInput.tsx
@@ -241,6 +241,7 @@ export const NumberFieldInput = React.forwardRef(function NumberFieldInput(
               'ArrowRight',
               'Tab',
               'Enter',
+              'Escape',
             ].includes(event.key);
 
             if (
--- a/packages/react/src/number-field/input/NumberFieldInput.tsx
+++ b/packages/react/src/number-field/input/NumberFieldInput.tsx
@@ -1,6 +1,6 @@
 'use client';
 import * as React from 'react';
-import { useModernLayoutEffect } from '@floating-ui/react/utils';
+import { useModernLayoutEffect, stopEvent } from '@floating-ui/react/utils';
 import { useNumberFieldRootContext } from '../root/NumberFieldRootContext';
 import { useComponentRenderer } from '../../utils/useComponentRenderer';
 import { useForkRef } from '../../utils/useForkRef';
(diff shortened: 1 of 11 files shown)


packages/react/src/number-field/input/NumberFieldInput.tsx
```diff
@@ -1,6 +1,6 @@
 'use client';
 import * as React from 'react';
-import { useModernLayoutEffect } from '@floating-ui/react/utils';
+import { useModernLayoutEffect, stopEvent } from '@floating-ui/react/utils';
```
mj12albert, line 3: Shouldn't this import from the local utils?

packages/react/src/tabs/root/TabsRoot.test.tsx
```diff
@@ -1062,4 +1064,84 @@ describe('<Tabs.Root />', () => {
       expect(root).to.have.attribute('data-activation-direction', 'up');
     });
   });
+
+  describe('popups', () => {
```
mj12albert, line 1068: Maybe we could consider grouping these tests with setups involving other components in `/packages/react/test/integrationTests/` or something like that? CC @michaldudak

packages/react/src/number-field/input/NumberFieldInput.tsx
atomiks: Not strictly necessary, for some of them I did for backwards-compat throughout the codebase

mj12albert (approved):
Looks good 👍

PR #1873, opened 2025-05-06
[Tabs] fix keepMounted for panel

### LINKED ISSUE #1872: Tabs.Panel: keepMounted doesn't work in
# Bug report

## Current behavior

The whole dom for non-active tab panel shouldn't exist when keepMounted is false (or is omitted).

But the dom still exists for non-active tab.

Image

## Expected behavior

The last two dom nodes from the previous screenshot shouldn't exist.

## Reproducible example

Your doc site https://base-ui.com/react/components/tabs

Or its CodeSandbox https://codesandbox.io/s/gzrqk5?file=src/App.tsx

## Base UI version

1.0.0-alpha.8

Chrome 135.0.3179.98 (Official build) (arm64) 

## Which OS are you using?

macOS

N/A

## Additional context

The related component: https://base-ui.com/react/components/tabs#panel

ImSingee:

Fix #1872

### DIFF
--- a/packages/react/src/tabs/panel/TabsPanel.tsx
+++ b/packages/react/src/tabs/panel/TabsPanel.tsx
@@ -47,9 +47,13 @@ export const TabsPanel = React.forwardRef(function TabPanel(
     render: render ?? 'div',
     className,
     state,
-    extraProps: { ...other, children: hidden && !keepMounted ? undefined : children },
+    extraProps: { ...other, children },
     customStyleHookMapping: tabsStyleHookMapping,
   });
+
+  if (hidden && !keepMounted) {
+    return null;
+  }
 
   return renderElement();
 });


packages/react/src/tabs/panel/TabsPanel.tsx
```diff
@@ -47,10 +47,14 @@ export const TabsPanel = React.forwardRef(function TabPanel(
     render: render ?? 'div',
     className,
     state,
-    extraProps: { ...other, children: hidden && !keepMounted ? undefined : children },
+    extraProps: { ...other, children },
```
mj12albert, line 50:
```suggestion
    extraProps: other,
```
This should be enough, no need to destructure children anymore ~

mj12albert:
Was checking out the failing tests and found that this change would break the functionality when `value` isn't explicitly specified on both the `Tab` and the `Panel` ([example](https://github.com/mui/base-ui/blob/master/packages/react/src/tabs/root/TabsRoot.test.tsx#L307-L316)). It's probably affecting the internal Composite in some way and I guess this could be why it was only unmounting the children

michaldudak:
Implemented in #3372

PR #1609, opened 2025-03-27
[docs] Add inset slider experiments

mj12albert:
Relates to https://github.com/mui/base-ui/issues/1530

Demo: https://deploy-preview-1609--base-ui.netlify.app/experiments/slider/inset

Update: after https://github.com/mui/base-ui/pull/1661 pseudo-elements are no longer necessary for functionality, one is needed to visually extend the indicator a bit

~~It's possible with the current API though a bit tricky:~~
- ~~Put `Control` inside `Track`, instead of track inside control as documented; this doesn't affect any functionality since track is just a plain div~~
- ~~Add semi-circular pseudo-elements to `Control` to extend the hitbox~~
- ~~Add semi-circular pseudo-elements to the `Track` so borders can wrap around them~~
- ~~To put a border around the whole thing, use box-shadow to apply top/bottom borders to the track, and borders around 3 sides of the `Track`s pseudo-elements~~

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #1874, opened 2025-05-06
[Slider] Set data-dragging on touchstart and pointerdown

mj12albert:
A small fix from https://github.com/mui/base-ui/pull/1609

It matches the native range input more when `data-dragging` is applied before any movement: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range#try_it

https://github.com/user-attachments/assets/8185d1c1-2fa0-46de-b634-a776c293ba53


atomiks (approved):
I wonder if this acts like/should be the unified `[data-pressed]` style hook that we're looking to add

atomiks:
Present also in alpha.8: If you move the cursor outside the iframe and let go, it doesn't remove the `[data-dragging]` style

mj12albert:
> Present also in alpha.8: If you move the cursor outside the iframe and let go, it doesn't remove the `[data-dragging]` style

I made a separate issue, `onValueCommitted` also gets stuck 🤔

PR #1876, opened 2025-05-06
[core] perf: avoid calling clearTimeout & cancelAnimationFrame

### LINKED ISSUE #1539: [tooltip] Lazy floating UI/don't execute any floating UI logic until mouse over on uncontrolled variant.
# Feature request

## Summary

Our website uses a ton of Tooltips, and using the profiler, seems that a decent amount of time is spent hitting floating-ui code even if the user hasn't interacted with the site or hovered anything. I'm wondering if it makes sense for the tooltip to have some prop that defers any floating UI logic until interaction? Of course, I could do this manually with a Controlled Tooltip but wondering if this would be beneficial as part of the library. Thanks!

### LINKED ISSUE #1875: `useAnimationsFinished` degrades performance
## Current behavior

When navigating to a page, most of the CPU time is taken by `useAnimationsFinished`. The 3 blobs of yellow bars represent `useAnimationsFinished`'s CPU time.

Image

## Expected behavior

Not using most of the CPU time.

## Reproducible example

Build & serve the mui-next repo docs, then navigate to `/select`.

## Base UI version

1.0.0-alpha.8

Brave (chromium)

## Which OS are you using?

Linux (gnome/archlinux)

romgrk:
Closes #1539
Closes #1875

Calling native functions unnecessarily degrades performance.

Reproduction: build & serve the mui-next repo docs, then navigate to the `/select` page.

| Before | After |
| --- | --- |
| Image | image |

Note: the 3 big stacks in the before picture are caused by `setTimeout` or `cancelAnimationFrame` calls. I haven't looked at why so many calls are made, but that might be another cause of perf degradation (not as big as actually making those native calls though).


packages/react/src/utils/useAnimationsFinished.ts
```diff
@@ -12,12 +12,18 @@ export function useAnimationsFinished(
...
-    cancelAnimationFrame(frameRef.current);
-    clearTimeout(timeoutRef.current);
+    if (frameRef.current) {
+      cancelAnimationFrame(frameRef.current);
+      frameRef.current = 0;
+    }
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+      timeoutRef.current = 0;
+    }
```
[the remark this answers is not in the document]
romgrk: For the animation frame class, I'd probably implement it as a JS-side scheduler to avoid even calling `requestAnimationFrame` too much (at most once per frame), I'm pretty sure it will also improve performance.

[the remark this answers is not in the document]
romgrk: To highlight how the encapsulating class can help, I also went through the other pages of the docs for mui-next and `clearTimeout` takes around 20% of CPU time for mounting each of those pages, e.g. here is `/popover`:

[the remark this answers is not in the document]
romgrk: The transformation was easy enough to apply that I went ahead with it and updated the final profiling benchmark in the top comment, it saves roughly 60% of the CPU time for mounting `/select` compared to the baseline.

atomiks:
This was mentioned in #1539 - so this will also close that.

In Floating UI it just used a simple `clearTimeoutIfSet` util to guard `-1`, but there wasn't any scheduling done

packages/react/src/utils/useAnimationFrame.ts
```diff
@@ -0,0 +1,103 @@
...
+const EMPTY = 0 as unknown as AnimationFrameId;
+
+class Scheduler {
+  /* This implementation uses an array as a backing data-structure for frame callbacks.
+   * It allows `O(1)` callback cancelling by inserting a `null` in the array, though it
+   * never calls the native `cancelAnimationFrame` if there are no frames left. This can
+   * be much more efficient if there is a call pattern that alterns as
+   * "request-cancel-request-cancel-…".
+   * But in the case of "request-request-…-cancel-cancel-…", it leaves the final animation
+   * frame to run anyway. We turn that frame into a `O(1)` no-op via `callbacksCount`. */
```
romgrk, line 16: I only implemented the scheduler for animation frames, but if we have a situation where we call `setTimeout(*, N)` a lot where `N` is constant across calls during one tick of the event loop, we could also optimize that use-case to limit the number of `setTimeout` calls.

romgrk:
> In Floating UI it just used a simple clearTimeoutIfSet util to guard -1, but there wasn't any scheduling done

That would also work, but I like to encapsulate it in a hook because it looks cleaner when reading the code, and it guarantees that the cleanup effect will be called (whereas it would need to be manually setup with a `clearTimeout`-like wrapper). One disadvantage of the hook however is that it forces allocating an object in memory for the `Timeout` instance, so in some cases using simple wrappers over the native functions could be better. That's what I did for a few `(request|cancel)AnimationFrame` cases in this PR, where I call `AnimationFrame.scheduler.(request|cancel)` directly rather than use an `AnimationFrame` instance.

packages/react/src/utils/useTimeout.ts
```diff
@@ -0,0 +1,49 @@
+'use client';
+import { useLazyRef } from './useLazyRef';
+import { useOnMount } from './useOnMount';
+
+type TimeoutId = ReturnType<typeof setTimeout>;
```
michaldudak, line 5: A similar code [caused issues in MUI X](https://github.com/mui/mui-x/pull/17533#discussion_r2058930177). Using `number` should be safer.

packages/react/src/collapsible/panel/useCollapsiblePanel.ts
```diff
@@ -132,9 +133,9 @@ export function useCollapsiblePanel(
     let frame = -1;
     let nextFrame = -1;
 
-    frame = requestAnimationFrame(() => {
+    frame = AnimationFrame.scheduler.request(() => {
```
michaldudak, line 136: It might be more convenient to exporse the scheduler methods directly on AnimationFrame (so we can call `AnimationFrame.schedule()`). WDYT?

packages/react/src/utils/useAnimationFrame.ts
```diff
@@ -0,0 +1,108 @@
...
+      this.currentId = EMPTY;
+    }
+  };
+
+  disposeEffect = () => {
+    return this.cancel;
+  };
+}
+
+export function useAnimationFrame() {
```
michaldudak, line 102: Some JSDocs could be useful, especially for external contributors.

packages/react/src/collapsible/panel/useCollapsiblePanel.ts
romgrk: I did think about it, but I went with this because it keeps the implementation for `AnimationFrame` and `Scheduler` in two separate classes, which makes it easier to read & understand the boundary. Both classes also have `request` and `cancel` methods, so I didn't like the name collision only differentiated with `static`. But I could be convinced to mash it all in one class.

I wouldn't call it `AnimationFrame.schedule()` though, I like keeping the API names as they are.

romgrk: I could also just add static methods on `AnimationFrame` that carry the call to scheduler, that would keep the classes separate.

romgrk: I've added `static` methods on `AnimationFrame` that forward the calls to `Scheduler`, looks like a good option for readability internally & externally.

packages/react/src/utils/useAnimationFrame.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.callbacks.push(fn);
+    this.callbacksCount += 1;
+
+    /* In a test environment with fake timers, a fake `requestAnimationFrame` can be called
+     * but there's no guarantee that the animation frame will actually run before the fake
+     * timers are teared, which leaves `isScheduled` set, but won't run our `tick()`. */
+    const didRAFChange =
+      process.env.NODE_ENV === 'test' &&
+      LAST_RAF !== requestAnimationFrame &&
+      ((LAST_RAF = requestAnimationFrame), true);
```
romgrk, line 64: This took a while to figure out :/

packages/react/src/tooltip/root/TooltipRoot.test.tsx
```diff
@@ -421,90 +421,6 @@ describe('<Tooltip.Root />', () => {
     });
   });
 
-  describe.skipIf(isJSDOM)('prop: onOpenChangeComplete', () => {
```
romgrk, line 424: I ran into this as I was debugging the scheduler, these tests seem to be a duplicate (with one block of code less inside one of the `it()`, and a few less `it()` blocks) of the tests just below. I've kept the longer version of these below.

romgrk:
This PR is ready for review.

michaldudak:
These changes look sensible. I haven't noticed anything breaking so far. Thanks for your contribution!

> I haven't looked at why so many calls are made, but that might be another cause of perf degradation

These are React's effect calls. We rely on effects quite a lot across all components, so they are likely unavoidable (I haven't checked them in more depth, though).

michaldudak (approved):

atomiks (approved):
This seems good! The auto-cleanup on unmount is much better DX

packages/react/src/select/trigger/useSelectTrigger.ts
```diff
@@ -44,20 +45,24 @@ export function useSelectTrigger(
...
-      const timeoutId2 = window.setTimeout(() => {
+      timeout1.start(200, () => {
         selectionRef.current.allowUnselectedMouseUp = true;
-      }, 200);
+
+        // mousedown -> mouseup on selected item should not select within 400ms.
+        timeout2.start(200, () => {
+          selectionRef.current.allowSelectedMouseUp = true;
+        });
+      });
```
atomiks, line 61: I don't think these should be nested as before?

romgrk: I'm pretty sure it's equivalent to the unnested version. It felt a bit more efficient perf-wise to nest it, but reading the context again (select open event) I think we can ignore the perf aspect for this case. I've reverted this change.

PR #1877, opened 2025-05-06
[number field] Refactor to `useRenderElement`

atomiks:


packages/react/src/number-field/root/NumberFieldRoot.tsx
```diff
@@ -3,10 +3,10 @@ import * as React from 'react';
 import { NumberFieldRootContext } from './NumberFieldRootContext';
 import { useNumberFieldRoot } from './useNumberFieldRoot';
```
mj12albert, line 4: `useNumberFieldRoot` could be dropped as well?

mj12albert (approved):

PR #1883, opened 2025-05-07
[code-infra] Do not wait for `dedupe` check before running other CI jobs

michaldudak:
Porting https://github.com/mui/mui-x/pull/17729 (including https://github.com/mui/mui-x/pull/17729#issuecomment-2857546512).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,9 +157,6 @@ jobs:
       - install_js:
           react-version: << parameters.react-version >>
       - run:
-          name: Should not have any unstaged files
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -180,6 +156,19 @@ jobs:
       - checkout
       - install_js:
           react-version: << parameters.react-version >>
+      - run:
```
Janpot, line 159: find this step rather questionable so I'm not against removing it as well

Janpot (approved):

PR #1885, opened 2025-05-07
[examples] Create the example to be used on StackBlitz previews

michaldudak:
Added the Vite + CSS modules example project.
This can serve as a starting point for exploring Base UI, and, more importantly, will be used to create a StackBlitz environment for every PR, so that PRs can be verified more conveniently.
See the [next comment](https://github.com/mui/base-ui/pull/1885#issuecomment-2858536639) for the link to the created environment.


atomiks:
I'm curious how this is better than navigating to a demo and using the CodeSandbox link (or at least linking to that automatically without needing to browse)? For instance, it doesn't set up the relevant component as a starting point. It also takes a while to install the deps and boot it up 🤔

michaldudak:
A couple of reasons:
- The link in the pkg.pr.new comment currently points to a project that's essentially unusable - it has Base UI installed but no React.
- StackBlitz, while being slower to boot, sets up a more robust environment (as it's basically Node running in the browser). Dependency resolution works better (it uses actual npm, not a custom resolver like CodeSandbox) and TypeScript support in the editor seems to be more reliable. Because of this, we'd like to move away from CodeSandbox at all (across all projects)
- We can set up multiple examples, with Next.js, Vite, Remix, etc. They can serve as both starting points for setting up projects and be be run on StackBlitz.

The startup time is indeed much worse, but for now we're not removing sandboxes from demos.
We might consider including all components in an example so it's easier to edit them.

atomiks:
> We might consider including all components in an example so it's easier to edit them.

This seems good

michaldudak:
I'll work on it in a follow-up PR.

PR #1890, opened 2025-05-08
[core] Use basic scroll lock on iOS

### LINKED ISSUE #1888: [Dialog][iOS Safari] Background scrolls to top of page when focusing a textarea within a dialog
# Bug report

## Current behavior

If I go to the `Close confirmation` example at https://base-ui.com/react/components/dialog#close-confirmation on iOS Safari, open the dialog, and tap within the text area, the background behind the dialog scrolls to the top of the page. The scroll position is then restored when the dialog is closed.

https://github.com/user-attachments/assets/13ea4bd8-7975-4861-b4ac-437187f9124b

## Expected behavior

The background should not move when focusing an element within the dialog.

## Reproducible example

https://base-ui.com/react/components/dialog#close-confirmation

## Base UI version

1.0.0-alpha.8

iOS Safari

## Which OS are you using?

iOS 18.4

None.

## Additional context

None.

atomiks:
Until we find a solution that works **without** any (dealbreaker) side effects, it's best if this is simply disabled on iOS.

Closes #1888


packages/react/src/utils/useScrollLock.ts
```diff
@@ -149,11 +149,10 @@ export function useScrollLock(params: {
 }) {
   const { enabled = true, mounted, open, referenceElement = null } = params;
 
-  const isReactAriaHook = React.useMemo(
+  const isBasicLock = React.useMemo(
```
michaldudak, line 153: A comment with a few words of explanation about basic and standard locks would be nice.

michaldudak (approved):

atomiks:
`overflow: hidden` on iOS does work if the navbar is expanded at least

PR #1896, opened 2025-05-08
[core] Rename `GenericHTMLProps` to `HTMLProps`

atomiks:
Works better on the docs and less confusing given it's `React.HTMLAttributes` plus `ref`


mj12albert (approved):
Much better for docs 👍

Issue #1886, opened 2025-05-07
[docs] Add a "Skip to Content" link


colmtuite:
Add a "Skip to Content" link.

![Image](https://github.com/user-attachments/assets/d89c1117-4753-4799-86b5-53fd5bfb61c9)

PR #1897, opened 2025-05-08
[docs] Add skip to main content link

mj12albert:
Closes https://github.com/mui/base-ui/issues/1886

https://github.com/user-attachments/assets/f889839a-1883-4397-b21d-7b93ef9e030e

Preview: https://deploy-preview-1897--base-ui.netlify.app/.


colmtuite:
Just change "contents" to "content" (singular) and we're good to go

mj12albert:
> Just change "contents" to "content" (singular)

Updated in https://github.com/mui/base-ui/pull/1897/commits/1862e89cf8a6436d6dfa28cebc4411c35182f6d9

Though I actually made it plural on purpose because Voiceover (with US English) on both MacOS and iPad keeps pronouncing "content" as content the adjective (e.g. "I feel content")

Could you give it a try? @colmtuite 

https://deploy-preview-1897--base-ui.netlify.app/react/components/avatar

colmtuite:
@mj12albert Interesting, great catch. If that's the case, then I guess "contents" is best.

docs/src/app/(public)/(content)/layout.tsx
```diff
@@ -25,7 +25,7 @@ export default function Layout({ children }: React.PropsWithChildren) {
         ))}
       </SideNav.Root>
 
-      <main className="ContentLayoutMain">
+      <main className="ContentLayoutMain" id="main-content">
```
oliviertassinari, line 28: Nice, to see it uses the same id name as in Material UI. A different name would be a pain to manage in the Algolia config. More context: https://github.com/algolia/docsearch/issues/1745.

docs/src/components/Header.tsx
```diff
@@ -13,6 +14,7 @@ export function Header() {
   return (
     <div className="Header">
       <div className="HeaderInner">
+        <SkipNav href="#main-content">Skip to content</SkipNav>
```
oliviertassinari, line 17: How about we move the ahrefs inside the SkipNav component so we enforce that it can only be single value?

oliviertassinari (commented):
We need this on the homepage too, no?

docs/src/components/Header.tsx
mj12albert: Makes sense, will export it as a constant as well then

mj12albert:
> We need this on the homepage too, no?

We don't need it right now, but it depends on the actual design and what the main content will be. As long as the design doesn't start with a 20+ tab stops menu like the docs it may not be as important ~

mj12albert:
> If that's the case, then I guess "contents" is best.

👌 Reverted back to "contents" @colmtuite

docs/src/components/SkipNav.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+import clsx from 'clsx';
+
+export function SkipNav({ className, ...props }: React.ComponentProps<'a'>) {
+  return <a className={clsx('SkipNav', className)} href={MAIN_CONTENT_ID} {...props} />;
+}
+
+export const MAIN_CONTENT_ID = '#main-content';
```
oliviertassinari, line 8:
```suggestion
export const MAIN_CONTENT_ID = 'main-content';
```

atomiks:
I think the focus ring shouldn't be cut off at the top and also match the color/style of the rest of the docs?

docs/src/components/SkipNav.tsx
mj12albert: Oops fixed

mj12albert:
> I think the focus ring shouldn't be cut off at the top and also match the color/style of the rest of the docs?

Oops thought it was Safari again but I forgot to copy the focus ring styles

PR #1899, opened 2025-05-09
[code-infra] Replace babel-plugin-add-import-extension

Janpot:
Noticed in https://github.com/mui/joy-ui/pull/116 that this plugin can't handle `.css` imports, it turns them into `.css.js`. Our plugin does resolve it correctly. base ui doesn't need to import css, but the plugin is quite outdated, we can align babel configs.


package.json
```diff
@@ -20,7 +20,7 @@
     "docs:size-why": "cross-env DOCS_STATS_ENABLED=true pnpm docs:build",
     "docs:start": "pnpm --filter docs serve",
     "docs:link-check": "pnpm --filter docs link-check",
-    "extract-error-codes": "cross-env MUI_EXTRACT_ERROR_CODES=true lerna run --concurrency 8 build:modern",
+    "extract-error-codes": "cross-env MUI_EXTRACT_ERROR_CODES=true lerna run --concurrency 1 build:stable",
```
Janpot, line 23: Not that it's important for this PR, but there is no modern build, and the command can't run concurrently

PR #1900, opened 2025-05-09
[test] Shorten Testing Library error messages

michaldudak:
React Testing Library often includes the HTML of the whole document when a test fails. This is unnecessary and makes the test output long and hard to read.


michaldudak:
This messes up the stack trace. Needs more work.

PR #1901, opened 2025-05-09
[core] Refactor: remove raw requestAnimationFrame

romgrk:
This was merged in parallel to #1876 so I didn't have a chance to change it.


packages/react/src/utils/floating-ui/useClick.ts
```diff
@@ -53,16 +54,7 @@ export function useClick(context: FloatingRootContext, props: UseClickProps = {}
   } = props;
 
   const pointerTypeRef = React.useRef<'mouse' | 'pen' | 'touch'>(undefined);
-  const frameRef = React.useRef(-1);
```
romgrk, line 56: Just fyi, `-1` is not recommended as an empty value for rAF: https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame#return_value

atomiks (approved):

PR #1902, opened 2025-05-10
[core] Performance: remove memory allocations

romgrk:
Refactor a few low-level utilities to avoid hitting the Garbage Collector so much.

Methodology to find optimization targets: check which functions are allocating large amount of objects via the memory profiler:

## Test case

Run the mui-next docs, start the profiler, and navigate to the `/select` page. Code is warmed up before the profile is ran by mounting/unmounting the page a few times. No CPU slowdown is set as it adds noise from one run to another.

## Results

> [!WARNING]
> Benchmarking the GC is imprecise and non-deterministic.

The various individual profiles that I ran result in a 15-35% decrease in total render time.

| Before | After |
| --- | --- |
| Screenshot From 2025-05-12 19-23-12 | Screenshot From 2025-05-12 19-20-52 |


packages/react/src/utils/useEventCallback.ts
```diff
@@ -1,4 +1,44 @@
 'use client';
-import { useEffectEvent } from '@floating-ui/react/utils';
```
romgrk, line 2: For reference:

[Source link](https://github.com/floating-ui/floating-ui/blob/ae1329d3173f8fd57c3a8808c6d1ae6a5e6d7825/packages/react/src/utils/hooks.ts#L17-L38)

```typescript
const useInsertionEffect = SafeReact.useInsertionEffect as
  | AnyFunction
  | undefined;

const useSafeInsertionEffect = useInsertionEffect || ((fn) => fn());

export function useEffectEvent<T extends AnyFunction>(callback?: T) {
  const ref = React.useRef<AnyFunction | undefined>(() => {
    if (__DEV__) {
      throw new Error('Cannot call an event handler while rendering.');
    }
  });

  useSafeInsertionEffect(() => {
    ref.current = callback;
  });

  return React.useCallback<AnyFunction>(
    (...args) => ref.current?.(...args),
    [],
  ) as T;
}
```

packages/react/src/utils/useLatestRef.ts
```diff
@@ -1,2 +1,27 @@
 'use client';
-export { useLatestRef } from '@floating-ui/react/utils';
```
romgrk, line 2: For reference:

[Source link](https://github.com/floating-ui/floating-ui/blob/ae1329d3173f8fd57c3a8808c6d1ae6a5e6d7825/packages/react/src/utils/hooks.ts#L2-L13)

```typescript
import useModernLayoutEffect from 'use-isomorphic-layout-effect';
import {SafeReact} from './safeReact';

export {useModernLayoutEffect};

export function useLatestRef<T>(value: T) {
  const ref = React.useRef<T>(value);
  useModernLayoutEffect(() => {
    ref.current = value;
  });
  return ref;
}
```

packages/react/src/utils/useEventCallback.ts
romgrk, line 1: Went from 3x allocations per render cycle down to a single allocation at initialization. The floating-ui version adds a devmode assertion during the first render phase. This new version extends that assertion to any render phase, not just the first one. That allows to skip the insertion effect.

packages/react/src/utils/useForkRef.ts
romgrk, line 1: Made `useForkRef` accept between 2-4 ref arguments. Use `useForkRefN` for the variadic version, which is only used once in the codebase. Went from 4x allocations per render cycle down to a single allocation at initalization time (plus one when refs change).

packages/react/src/utils/useLatestRef.ts
romgrk, line 1: Functionally equivalent. From 1 allocation per render cycle down to a single allocation at initalization.

packages/react/src/select/item-text/SelectItemText.tsx
```diff
@@ -59,11 +53,10 @@ export const SelectItemText = React.forwardRef(function SelectItemText(
 ) {
   const { selected, indexRef } = useSelectItemContext();
   const { selectedItemTextRef } = useSelectRootContext();
-  const mergedRef = useForkRef<HTMLElement>(forwardedRef);
 
   return (
     <InnerSelectItemText
-      ref={mergedRef}
+      ref={forwardedRef}
```
atomiks, line 59: The forwarded ref may not be memoized

packages/react/src/utils/useEventCallback.ts
```diff
@@ -1,4 +1,44 @@
...
+  } else {
+    // In development, we enable strict assertion that the event callback is
+    // never called during the render phase.
+    stable.callback = ASSERT_NOT_CALLED as unknown as T;
+
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useInsertionEffect(() => {
+      stable.callback = callback;
+    });
+  }
```
atomiks, line 33: This seems somewhat questionable, it's safer if there aren't runtime differences between production and development regarding the phase in which it's set (writing in render doesn't abide by the Rules of React).

atomiks (commented):
It seems better if these changes were made in `floating-ui` so we can both reuse the utils to reduce bundle size, and also because `floating-ui` still uses the deoptimized versions which we still get penalized for

packages/react/src/select/item-text/SelectItemText.tsx
romgrk: Then `useForkRef` would still just re-create a new `mergedRef` because `forwardedRef` would change on every render. So it's just pure overhead to use it here, if I'm not mistaken.

romgrk:
> It seems better if these changes were made in floating-ui

Maybe, but I first need to prove that the changes provide measurable gains, it's easier to make the changes here and test with our internal repositories (joy-ui mainly).

I also don't love having the code live in an external repository. It might be nice to create instead a small hooks package in the mui org, and then send a PR to floating-ui to use it? It's not the first time we mention a hooks package, and it could also benefit material-ui. The versions implemented here are faster than some of the hooks I wrote/tweaked in material-ui.

flaviendelangle:
Floating UI is maintained mostly by @atomiks

packages/react/src/merge-props/mergeProps.test.ts
```diff
@@ -199,7 +199,12 @@ describe('mergeProps', () => {
 
   describe('props getters', () => {
     it('calls the props getter with the props defined after it', () => {
-      const propsGetter = spy((props) => props);
+      let observedProps;
+      const propsGetter = spy((props) => {
+        observedProps = { ...props };
+        return props;
+      });
+
```
romgrk, line 207: `mergeProps` now uses a single accumulator props object and progressively mutates it in-place as it merges a new bundle of props. Which is fine, except that prop-getters see that object while they run, so if they keep a reference to that props object and later use it like these tests do (in other words, if prop getters are impure), then they'll see the final merged props instead of what they saw when the function was called.

`mergeProps` could fall back to copying props before passing them to a prop-getter, but that would make it slower. What do we think? Do we want to support prop-getters that keep a ref to props they see, or are we ok with this optimization?

packages/react/src/merge-props/mergeProps.ts
romgrk, line 1: Both `mergeProps` and `useForkRef` have become non-variadic. The only place in the codebase where the variadic versions are used is `useRenderElement`.

packages/react/src/utils/useRenderElement.tsx
romgrk, line 1: This hook has been split into lazy & non-lazy variants. The only place in the codebase where the lazy variant is used is inside `useRender()`.

packages/react/src/utils/useEventCallback.ts
```diff
@@ -1,4 +1,41 @@
 'use client';
-import { useEffectEvent } from '@floating-ui/react/utils';
+import * as React from 'react';
+import { useLazyRef } from './useLazyRef';
 
-export const useEventCallback = useEffectEvent;
+// https://github.com/mui/material-ui/issues/41190#issuecomment-2040873379
+const useInsertionEffect =
+  (React as any)[`useInsertionEffect${Math.random()}`.slice(0, 5)] || ((fn: any) => fn());
```
atomiks, line 7: The slice needs to match the length of the string, but I think we should just go with `{...React}` instead? It's a bit more costly but close to negligible (at least this is used in Floating UI).

romgrk: I have updated to slice away the number with a negative index. I have seen to many `...` in the codebase, I'm trying to stay away from them.

romgrk:
This PR is ready for review.

packages/react/src/utils/useLatestRef.ts
```diff
@@ -1,2 +1,27 @@
...
+import * as React from 'react';
+import { useLazyRef } from './useLazyRef';
+
+const useEffect = typeof document !== 'undefined' ? React.useLayoutEffect : React.useEffect;
```
atomiks, line 5: `useModernLayoutEffect`

packages/react/src/merge-props/mergeProps.ts
```diff
@@ -29,110 +32,142 @@ type MergableProps<T extends React.ElementType> =
...
+  d: InputProps<T>,
+): PropsOf<T>;
+export function mergeProps<T extends ElementType>(
+  a: InputProps<T>,
+  b: InputProps<T>,
+  c: InputProps<T>,
+  d: InputProps<T>,
+  e: InputProps<T>,
+): PropsOf<T>;
+export function mergeProps(a: any, b: any, c?: any, d?: any, e?: any) {
```
atomiks, line 56: Although 6+ sets of props is unlikely, this technically allowed an infinite number of objects to merge before, so it'd be a breaking change?

atomiks (commented):
It would be great if the perf improvements here were ported to Floating UI as well, even if we don't share some of the utils yet. Especially the `useInteractions`/merging hook which is used a lot implicitly by us, and can't be shared since the logic is a bit different. That should result in another substantial gain in performance.

packages/react/src/merge-props/mergeProps.ts
romgrk: Yes, but isn't base-ui still in alpha? If it really is a problem, we could export `mergePropsN` as `mergeProps`.

romgrk:
Sure, but I'd rather do that in a subsequent step. Although floating-ui's runtime is surprisingly high, I'm going through the points in [the notion docs](https://www.notion.so/mui-org/Select-profiling-1eecbfe7b660803ebfc0c466a3e48e99) in order of importance.

packages/react/src/merge-props/mergeProps.ts
atomiks: Yeah, we just need to mark them as `breaking change`, which is okay still. However I think such a utility should probably be variadic for the public export, so `mergePropsN` should be good.

romgrk: I think we shouldn't expose the slower variadic version by default. There's not a single case inside the codebase where we need to merge 6+ set of props, I don't think it's more likely to happen outside the codebase. If it does, it's nice that users have a moment to stop and think about what they're doing.

romgrk:
Just added a comment in the discussion above about that.

atomiks:
Fair enough, 6 plus is pretty excessive indeed.

PR #1905, opened 2025-05-12
[number field] Ensure `onValueChange` is called with already-formatted parsed value

### LINKED ISSUE #1817: [number field] refocusing and blurring input can cause value to change
# Bug report

## Current behavior

If you enter a value, say `"1.23456"`, which would be truncated by `formatNumber`, `onValueChange` gets called with `1.23456`, and the input text is set to e.g., "1.234".

The next time you focus and blur, `onValueChange` is called with `1.234`, despite the user not having input.

## Expected behavior

Up for discussion, but IMO focusing and blurring an input should not change the underlying value.

## Reproducible example

https://codesandbox.io/p/sandbox/base-ui-number-input-blurring-xj5xsd

## Base UI version

1.0.0-alpha.8

Edge

## Which OS are you using?

Windows

## Additional context

In my personal number input component, to avoid this, I'm just roundtripping the value before calling `onValueChange`, so you do `user input -> value -> formatted text -> value`. Currently it's doing `user input -> value` and `value -> formatted text` separately.

atomiks:
Fixes #1817

- Uses `maximumFractionDigits: 20` when checking to ensure full precision
- Values coming externally have their precision respected as priority _until_ changed via the event handlers, at which point the formatting is respected

Preview: 

- https://codesandbox.io/p/sandbox/base-ui-number-input-blurring-2-forked-crsnw2?file=%2Fsrc%2FApp.tsx
- https://codesandbox.io/p/sandbox/base-ui-number-input-blurring-forked-zsk4gx?file=%2Fsrc%2FApp.tsx


packages/react/src/number-field/root/NumberFieldRoot.test.tsx
```diff
@@ -135,6 +135,39 @@ describe('<NumberField />', () => {
...
+        );
+      }
+
+      await render(<App />);
+      const input = screen.getByRole('textbox');
+
+      const formatted = new Intl.NumberFormat().format(1.23456);
+      expect(input).to.have.value(formatted);
+
+      // First focus → blur: should round-trip and fire once
```
mj12albert, line 160: @atomiks Tried to test this in the [sandbox](https://codesandbox.io/p/devbox/base-ui-number-input-blurring-2-forked-67dq8p?workspaceId=ws_WwF21DhBng6HdaYyXQWC1w) with react-query, after clicking `externally set to 1.23456` the console.log in `onValueChange` never runs (after focusing and blurring)

(The fix works in the [original sandbox](https://codesandbox.io/p/sandbox/base-ui-number-input-blurring-forked-n2nppx) though)

mj12albert: OK I see ~

PR #1906, opened 2025-05-13
[select] Defer mounting until typeahead is needed

atomiks:
This defers mounting the `<Select.Portal>` component until it's needed for closed typeahead (when the trigger is focused). After it opens for the first time, it stays mounted, which helps performance when opening/closing large lists.

## `Select.Value` API change
The `placeholder` prop has been changed to `children` as a plain string. This must always be specified because the items are no longer rendered upon mount, which was previously used to grab the selected item value's `.textContent` to render the value in the trigger while it was closed. However, they always needed to use `placeholder` for SSR too, and it was confusing how that API worked. The docs were using:

```jsx
<Select.Value>
  {(label) => label || ssrFallback}
</Select.Value>
```

So we could make `children: React.ReactNode` the "default/initial" label present during SSR, and now with this PR, before the select menu is opened for the first time. It will get overwritten by the selected label when it needs to after being interacted with:

```jsx
<Select.Value>
  {initialValue}
</Select.Value>
```

## Performance 

For 1,000 items: since the items don't need to mount on page load, scripting time is 35% of what it was before:

This PR: https://deploy-preview-1906--base-ui.netlify.app/experiments/select-perf
Screenshot 2025-05-14 at 4 34 40 pm

Master: https://master--base-ui.netlify.app/experiments/select-perf
Screenshot 2025-05-14 at 4 34 20 pm

However, there's a tradeoff here. When opening the Select for the first time, it's noticeably slow compared to before, since the work was already done on load when the user is less likely to notice (depending on the page/flow):

With 20x slowdown:
Master: ~1,168 ms to open first time (~40ms without slowdown)
This PR: ~2,544 ms to open first time (~136ms without slowdown)


atomiks:
@romgrk the main issue here is that `<Select.Value>` uses the label of the rendered item (by default its `textContent`) after hydration. It swaps from `placeholder` to the rendered value, but this can't be deferred - so it seems that this would need to be either opt-in, _or_ require the user specify `placeholder` (or some better-named prop) for the first render

docs/reference/generated/select-value.json
romgrk, line 1: > So we could make children: string the "default/initial" label present during SSR, and now with this PR, before the select menu is opened for the first time.

Any chance we can change that from `children` to some prop with a more semantic name? I would expect `children` to always be rendered if present, whereas a prop like `initial` would convey more clearly the expectation.

docs/src/app/(private)/experiments/select-perf.tsx
romgrk, line 1: > However, there's a tradeoff here. When opening the Select for the first time, it's noticeably slow compared to before, since the work was already done on load when the user is less likely to notice (depending on the page/flow)

That can be improved in a different PR, there's a few things that can be optimized inside select items.

> since the work was already done on load when the user is less likely to notice

Browsing the joy-ui docs for the first time, I could noticeably feel that the select docs page was slow to render compared to the other pages, and it barely has 18 selects. LCP, TTI & all other related web metrics are part of lighthouse a11y scores, we can't just throw work at load-time. It might not be noticeable enough for you on a fast laptop, but on a slow phone those delays are definitely consciously perceived by users.

docs/reference/generated/select-value.json
atomiks: I felt like it was more confusing to have `initial` + `children` callback as possible options. If you specify a children string you can't also specify a callback. This one feels quite natural once you know how it works (there's no need to have an "always-present" children value, because then you don't need to use `Select.Value` at all)

docs/src/app/(private)/experiments/select-perf.tsx
romgrk: > If you have something like the Twitter feed which is virtualized, it's likely better to do the work while scrolling

For having re-done the datagrid virtualization engine not too long ago, I can guarantee you that with virtualization, you don't want any kind of work done by the virtualized items during scrolling. The fundamental problem with virtualization in JS is that the spec behavior for the `scroll` event is for JS to run _after_ the user has already scrolled the viewport, so you're already late for rendering when JS work starts. So you want any CPU time to be spent on rendering actually visible content, not invisible items nor their JSX. You can do "clever" stuff like predicting/guessing by how much the user will scroll to do something viable with the `scroll` event, but virtualization is an unsatisfying problem to solve in JS. Anyway I digress.

> If you're coming to the page for the first time (not from a client-side link), it seems doing the work on page load is better because it's not really noticeable

I don't understand how you're coming to this conclusion.

> But in the case of the Joy UI docs, it feels better to defer the work until opening when navigating with the client-side routed links on the navbar.

I don't understand how that use-case is different from the one above. In both cases, the user is trying to navigate to page X (be it via an initial page load or a client-side link) and we're blocking them by doing upfront work.

atomiks: Hmm, then maybe you want to try to mount the items when the user is idle or in such a way that the work won't cause any adverse effects, but _still_ before they manage to open it. I thought of trigger hover, but that doesn't solve touch input and it might cause hover transitions on the trigger to be laggy (haven't tested)

> I don't understand how you're coming to this conclusion. 
> I don't understand how that use-case is different from the one above.

With first page load, the user is more tolerant of blocking JS work since the screen has still painted and are still adjusting to the actual website layout, so it's less noticeable in two ways. In the Joy UI docs, the page remains the same (so no disorientation tolerance effect), but most importantly it doesn't navigate/paint after clicking until the JS work is complete which feels much worse than seeing the prerendered page that has yet to complete hydration

romgrk: > With first page load, the user is more tolerant of blocking JS work since the screen already painted (with SSR pre-rendering at least), and they're still adjusting to the actual website layout

That's an assumption that SSR is enabled. Besides, with large select lists it still causes render-blocking time: once in the server which needs to render all those select items (even if they're not in the output HTML) and once after hydration. The wasted server time is render blocking.

And the idle time after hydration isn't that long anyway. For frequently used interfaces where the user already has an action planned, I'd estimate the idle time can be as low as 500ms after first paint (at least for me on say, gmail, that's the case). That 500ms could easily be overspent in hydrating select items for large lists (in addition to whatever other work also needs to happen).

michaldudak:
> However, there's a tradeoff here. When opening the Select for the first time, it's noticeably slow compared to before, since the work was already done on load when the user is less likely to notice (depending on the page/flow)

One way to solve it would be to use an intersection observer to mount only visible selects. This should help, especially on mobile devices, as their viewport is much smaller. We could also make this behavior controllable with a prop (`mountItems='eager|lazy|when-visible'`)

romgrk:
> One way to solve it would be to use an intersection observer to mount only visible selects

All it takes is a couple of select with lots of items at the top of the page and the end-user will feel the delays. I don't think there's a viable solution that allows us to decide for users that we're going to dedicate CPU runtime for non-visible content.

> it's noticeably slow compared to before

I've been holding off on optimizing select items while this PR is open, but I think them being split into outer/inner components contributes to their high initial mount time. If we spend time optimizing the item component, I think we can reduce the INP for select open to acceptable levels. If it's really necessary, we could also look into virtualizing large item lists.

atomiks:
I'm ok with the tradeoff of larger lists taking longer to open on first mount for now. It's better than the current default.

michaldudak:
Can we then see how much the current implementation can be sped up and then decide on deferring mounting strategy? We also considered using an idle callback (or timeout) to make allow the page to load and only then mount the items.

atomiks:
Thinking over `Select.Value` API more: `children` as a string represents the initially selected item label during SSR and pre-mount. There should be no need for an extra `placeholder` prop?

https://codesandbox.io/p/sandbox/cranky-dream-7xc64p

Case 1: There is no way to unselect anything/no empty value

```jsx
<Select.Root defaultValue="red">
  <Select.Value>
    Red {/* SSR/pre-mount label */}
  </Select.Value>
  <Select.Portal>
    <Select.Positioner>
      <Select.Popup>
        <Select.Item value="red">
          Red
        </Select.Item>
      </Select.Popup>
    </Select.Positioner>
  </Select.Portal>
</Select.Root>
```

Case 2: There is an empty value but only initially

```jsx
<Select.Root>
  <Select.Value>
    Select value {/* SSR/pre-mount label, can't revert back */}
  </Select.Value>
  <Select.Portal>
    <Select.Positioner>
      <Select.Popup>
        <Select.Item value="red">
          Red
        </Select.Item>
      </Select.Popup>
    </Select.Positioner>
  </Select.Portal>
</Select.Root>
```

Case 3: There is an empty value, including as an item

```jsx
<Select.Root>
  <Select.Value>
    Select value {/* SSR/pre-mount label */}
  </Select.Value>
  <Select.Portal>
    <Select.Positioner>
      <Select.Popup>
        <Select.Item> {/* null value by default */}
          Select value {/* Acts as the placeholder, which is the same */}
        </Select.Item>
        <Select.Item value="red">
          Red
        </Select.Item>
      </Select.Popup>
    </Select.Positioner>
  </Select.Portal>
</Select.Root>
```

atomiks:
> Can we then see how much the current implementation can be sped up and then decide on deferring mounting strategy?

I think this is a better default so we can still go with this and decide on a mounting strategy in a separate PR

packages/react/src/select/value/SelectValue.test.tsx
```diff
@@ -15,11 +15,11 @@ describe('<Select.Value />', () => {
     },
   }));
 
-  describe('prop: placeholder', () => {
+  describe('prop: initial', () => {
```
michaldudak, line 18:
```suggestion
  describe('children', () => {
```

michaldudak (approved):

PR #1908, opened 2025-05-13
[scroll area] Ignore `data-scrolling` during programmatic scroll

### LINKED ISSUE #983: [ScrollArea] Ignore state changes during programmatic scroll
Can Scroll Area ignore state changes during programmatic scroll? I don't think it's useful to display `data-scrolling` styles when the viewport was scrolled programmatically

In our docs side nav, on page load we are scrolling the active link into view. I attempted to create a custom workaround to hide the scrollbars while that's happening, but it's ugly and fragile, and actually stopped working at some point (not even sure when!):

https://github.com/mui/base-ui/blob/b352ac3f102e55dd0af77316e5f6112d0cf1ea6a/docs/src/components/SideNav.tsx

atomiks:
Closes #983

### DIFF
--- a/docs/src/components/SideNav.tsx
+++ b/docs/src/components/SideNav.tsx
@@ -3,57 +3,23 @@ import * as React from 'react';
 import clsx from 'clsx';
 import NextLink from 'next/link';
 import { usePathname } from 'next/navigation';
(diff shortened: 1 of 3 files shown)


atomiks:
Yeah the code in SideNav.tsx (removed in this PR) handles that but it sometimes doesn't work, quite random though

PR #1909, opened 2025-05-13
[useButton] `nativeButton` prop

### LINKED ISSUE #1843: Disabled button/trigger is missing disabled attribute during SSR
# Bug report

## Current behavior

```tsx
import { renderToString } from 'react-dom/server';

console.log(
  renderToString(
    <Menu.Root>
      <Menu.Trigger disabled={true}>Menu</Menu.Trigger>
      <Menu.Portal>
        <Menu.Positioner>
          <Menu.Popup>
            <Menu.Item>Add to Library</Menu.Item>
          </Menu.Popup>
        </Menu.Positioner>
      </Menu.Portal>
    </Menu.Root>
  )
);

```

Result: no `disabled` attribute, only `data-disabled`.

```html
<button aria-haspopup="menu" tabindex="0" aria-expanded="false" id="«R0H1»" data-disabled="">Menu</button>
```

## Expected behavior

Result: `disabled` attribute _as well as_ `data-disabled`.

```html
<button disabled aria-haspopup="menu" tabindex="0" aria-expanded="false" id="«R0H1»" data-disabled="">Menu</button>
```

## Base UI version

1.0.0-alpha.8

Chrome

## Which OS are you using?

Mac OS

Voiceover/JAWS etc.

## Additional context

atomiks:
Fixes #1843

Adds `nativeButton: boolean` prop to components that use `useButton`.

When using the `render` prop, you must specify `nativeButton={boolean}` if the tag that's rendered differs from the default one (these are documented already as `Renders a <button> element` or `Renders a <div> element`).


packages/react/src/utils/useRootElementName.ts
```diff
@@ -1,7 +1,5 @@
 'use client';
```
mj12albert, line 1: I think this hook was for in case we ever wanted to expose the element/tag name as a prop, since we're not doing that it can just be dropped and merged into useButton

atomiks: The event handler still uses it inside the events for e.g. `event.preventDefault()`

michaldudak:
We don't really need to know the actual rendered tag name, but just the fact that they want to render something different than the native `<button />`. So a boolean prop would suffice.

atomiks:
@michaldudak it also checks for `a` when computing the rendered props. It will avoid `role="button"` in that case

https://github.com/mui/base-ui/blob/1634afed3bce1d2920c1909b8cd2a56302c07df2/packages/react/src/use-button/useButton.ts#L59-L60

michaldudak:
I don't think we should treat links the same way as buttons. I'd rather have separate `useLink` hook. This code was ported over from Material UI, where it may make more sense to have a more generic notion of a button, but we are low-level enough to separate buttons and links.

atomiks:
Seems fair enough. Our components usually have a `role=""` specified (like `menuitem` or `option`) to begin with so it might not be an issue. I'll see what I can change.

atomiks:
@michaldudak updated PR description

Adds `nativeButton: boolean` to any component that uses `useButton`. We could check the simple case (a plain element like `<div />`) or always require it to be specified if it differs from the default rendered tag (probably better as it's more explicit/less magical/more consistent).

packages/react/src/use-button/useButton.ts
```diff
@@ -103,9 +85,16 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
         ...otherExternalProps
       } = externalProps;
 
+      let type: 'button' | 'text' | undefined;
+      if (native === 'input') {
+        type = 'text';
```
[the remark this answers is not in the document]
michaldudak: Why are we handling Toolbar.Input with `useButton`, though?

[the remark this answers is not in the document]
mj12albert: It's for the `focusableWhenDisabled` related features, maybe it can be extracted out of useButton though nothing else uses it standalone I think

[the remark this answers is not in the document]
michaldudak: Yeah, I don't like it being handled by this hook only because it happens to require similar functionality. Let's create another hook that can be called by `useButton` and Toolbar.Input.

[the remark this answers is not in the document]
mj12albert: Yeah I think it would be a good idea to split off `useLink` and `useFocusableWhenDisabled` from this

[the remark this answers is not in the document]
atomiks: To reduce the scope of this PR, it's likely a good idea to do that in a new one

packages/react/src/toolbar/link/ToolbarLink.tsx
```diff
@@ -26,7 +27,8 @@ export const ToolbarLink = React.forwardRef(function ToolbarLink(
 
   const { getButtonProps } = useButton({
     buttonRef: forwardedRef,
-    elementName: 'a',
+    native: 'a',
+    focusableWhenDisabled: true,
```
mj12albert: Links cannot be disabled, thats why `focusableWhenDisabled` wasn't passed here; I realize the `TOOLBAR_LINK_METADATA` above is super misleading, but the `true` is only used for deriving `disabledIndices`

I added a comment about this [here](https://github.com/mui/base-ui/pull/1999/commits/bfb22041f3d174e9136e8225ad6b4dba7734956c), sorry for conflicting this PR again btw 🙏

michaldudak: I'd also remove support for links in useButton (could be in a separate PR) and create another hook if necessary.

docs/reference/generated/accordion-trigger.json
```diff
@@ -2,6 +2,11 @@
   "name": "AccordionTrigger",
   "description": "A button that opens and closes the corresponding panel.\nRenders a `<button>` element.",
   "props": {
+    "nativeButton": {
+      "type": "boolean",
+      "default": "true",
+      "description": "Determines whether the component is being rendered as a native button."
```
michaldudak: IMO we need a more detailed description what it's for. How about adding "Set it to `false` if the `render` prop renders a non-interactive element, like `div` or `span`."?

packages/react/src/toolbar/link/ToolbarLink.tsx
mj12albert: I agree to do it separately; at a glance there's actually nothing to put in a useLink hook and we can just remove the link logic

`ToolbarLink` is the only link component using this; the only other link component we have is `NavigationMenu.Link` and that doesn't use `useButton`

docs/reference/generated/accordion-trigger.json
atomiks: It can technically be replaced with an interactive element, so I think any different tag should be the doc

packages/react/src/use-button/useButton.ts
```diff
@@ -4,45 +4,51 @@ import { useForkRef } from '../utils/useForkRef';
...
-      elementName === 'BUTTON' ||
-      (elementName === 'INPUT' && BUTTON_TYPES.has((element as HTMLInputElement)?.type))
-    );
-  });
+    if (hasError) {
+      throw new Error(
+        `Set the \`nativeButton\` prop to \`${isButtonButFalse}\` when rendering a <${tagNameLower}> element.`,
+      );
+    }
+  }, [native]);
```
atomiks: Since this prop is quite obscure and not shown in the types when using `render` by default, a runtime error is likely required

atomiks: Nevermind, doesn't play nicely with the conformance tests

docs/src/app/(public)/(content)/react/components/popover/demos/hero/css-modules/index.tsx
```diff
@@ -5,7 +5,7 @@ import styles from './index.module.css';
 export default function ExamplePopover() {
   return (
     <Popover.Root>
-      <Popover.Trigger className={styles.IconButton}>
+      <Popover.Trigger className={styles.IconButton} render={<div />}>
```
michaldudak: Is this a leftover from testing?

atomiks: Yeah woops

michaldudak (approved):

Issue #1912, opened 2025-05-13
[input] defaultValue missing from Input component props


Alvi24:
i want to give the input a defaultValue (like the native web <input/>) but the defaultValue is missing in Base UI Input component

mj12albert:
@Alvi24 defaultValue should work: https://codesandbox.io/p/sandbox/jovial-visvesvaraya-7cmkv4

We removed it from the types in https://github.com/mui/base-ui/pull/1617

Since the component is just an alias for `Field.Control` we could probably `extend Field.Control.Props` for `Input.Props` as well

PR #1915, opened 2025-05-13
[input] Fix `Input.Props` type

mj12albert:
Closes https://github.com/mui/base-ui/issues/1912

### DIFF
--- a/packages/react/src/input/Input.tsx
+++ b/packages/react/src/input/Input.tsx
@@ -18,7 +18,13 @@ export const Input = React.forwardRef(function Input(
 });
 
 export namespace Input {
-  export interface Props extends BaseUIComponentProps<'input', State> {}
+  export interface Props extends BaseUIComponentProps<'input', State> {
+    /**
+     * Callback fired when the `value` changes. Use when controlled.
+     */
+    onValueChange?: Field.Control.Props['onValueChange'];
+    defaultValue?: Field.Control.Props['defaultValue'];
+  }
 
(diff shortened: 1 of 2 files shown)


packages/react/src/input/Input.tsx
```diff
@@ -18,7 +18,13 @@ export const Input = React.forwardRef(function Input(
 });
 
 export namespace Input {
-  export interface Props extends BaseUIComponentProps<'input', State> {}
+  export interface Props extends BaseUIComponentProps<'input', State> {
+    /**
+     * Callback fired when the `value` changes. Use when controlled.
+     */
+    onValueChange?: Field.Control.Props['onValueChange'];
+    defaultValue?: Field.Control.Props['defaultValue'];
```
atomiks, line 26: This shouldn't need to be specified, `defaultValue` on `Field.Control` already is allowed:

```js
<Field.Control defaultValue="" />
```

Maybe just extend `Field.Control.Props`?

mj12albert: > Maybe just extend `Field.Control.Props`?

That's what I tried first, but it causes the generated input.json to show `Field.Control.State` (e.g. in className, render) instead of `Input.State` @atomiks

atomiks: @mj12albert Alright, makes sense, `Field.Control` also specifies it since it's omitted as a default for everything.

atomiks (approved):

PR #1919, opened 2025-05-14
[field] Consolidate `Field.Error` `forceShow` into `match` prop

### LINKED ISSUE #1200: [field] Is `forceShow` prop necessary?
iirc we've discussed that having the `forceShow` prop can be avoided altogether by just rendering the field error yourself:
```jsx
{myCondition && <Field.Error>My special error</Field.Error>}
```

atomiks:
Closes #1200, based on https://github.com/mui/base-ui/issues/1200#issuecomment-2636790851

`match` now has `boolean` type, consolidating the always-show behavior into it rather than the separate `forceShow` prop


packages/react/src/field/error/FieldError.tsx
```diff
@@ -87,11 +87,8 @@ export namespace FieldError {
     /**
      * Determines whether to show the error message according to the field’s
      * [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState).
+     * Specifying `true` will always show the error message.
```
mj12albert, line 90: Works well, maybe worth mentioning this is useful for letting external libraries control it?

mj12albert (approved):

PR #1921, opened 2025-05-14
[perf] Fix `useLockScroll` forced reflows

romgrk:
The current implementation is causing lots of reflows.

Test case: https://www.notion.so/mui-org/Menu-profiling-1eecbfe7b66080c9a810c82f2308f04d

The "menu click" portion highlighted below is the render-blocking JS that needs to run before the browser repaints. After these changes, floating-ui becomes an interesting optimization target as it's the next big render-blocking task.

### Before

image

### After

image

### DIFF
--- a/packages/react/src/utils/useScrollLock.ts
+++ b/packages/react/src/utils/useScrollLock.ts
@@ -162,7 +162,7 @@ class ScrollLocker {
       this.restore?.();
       this.restore = null;
     }
-  }
+  };
 
   private lock(referenceElement: Element | null) {
     const isOverflowHiddenLock = isIOS() || (isFirefox() && !hasInsetScrollbars(referenceElement));
--- a/packages/react/src/utils/useScrollLock.ts
+++ b/packages/react/src/utils/useScrollLock.ts
@@ -157,7 +157,7 @@ class ScrollLocker {
     this.unlockFrame.request(this.unlock);
   };
 
-  private unlock() {
+  private unlock = () => {
     if (this.lockCount === 0 && this.restore) {
       this.restore?.();
       this.restore = null;
--- a/packages/react/src/utils/useScrollLock.ts
+++ b/packages/react/src/utils/useScrollLock.ts
@@ -1,4 +1,3 @@
-import * as React from 'react';
 import { isFirefox, isIOS, isWebKit } from './detectBrowser';
 import { ownerDocument, ownerWindow } from './owner';
 import { useModernLayoutEffect } from './useModernLayoutEffect';
@@ -7,12 +6,8 @@ import { AnimationFrame } from './useAnimationFrame';
 let originalHtmlStyles: Partial<CSSStyleDeclaration> = {};
 let originalBodyStyles: Partial<CSSStyleDeclaration> = {};
 let originalHtmlScrollBehavior = '';
-let preventScrollCount = 0;
-let restore: () => void = () => {};
 
-export function getPreventScrollCount() {
-  return preventScrollCount;
-}
+const NOOP = () => {};
 
 function hasInsetScrollbars(referenceElement: Element | null) {
   if (typeof document === 'undefined') {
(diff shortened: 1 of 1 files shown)


packages/react/src/utils/useScrollLock.ts
```diff
@@ -49,6 +49,8 @@ function preventScrollStandard(referenceElement: Element | null) {
   }
 
   function lockScroll() {
+    /* DOM reads: */
+
```
romgrk, line 53: Grouped reads & writes to avoid this:
Screenshot From 2025-05-14 03-32-31

packages/react/src/utils/useScrollLock.ts
```diff
@@ -137,66 +137,89 @@ function preventScrollStandard(referenceElement: Element | null) {
...
+    this.lockCount -= 1;
+    this.unlockFrame.request(this.unlock);
+  };
+
+  private unlock = () => {
+    if (this.lockCount === 0 && this.restore) {
+      this.restore?.();
+      this.restore = null;
+    }
+  }
```
romgrk, line 165: Avoids unlock-lock sequences that create a lot of work. This is however a bit brittle, for example if React splits its work in multiple frames, the the unlock-lock sequence could happen and it would trigger reflows. We could instead use a timeout with a small delay to make it a bit more resilient. Wdyt?

Screenshot From 2025-05-14 04-16-43

packages/react/src/utils/useScrollLock.ts
romgrk: This is a good start, but floating-ui is also read-then-writing the DOM, so we still have forced reflows, but it will required fixing in or sync'ing with floating-ui. Maybe something like [fastdom](https://github.com/wilsonpage/fastdom).

packages/react/src/utils/useScrollLock.ts
```diff
@@ -137,66 +138,94 @@ function preventScrollStandard(referenceElement: Element | null) {
...
+    this.lockCount -= 1;
+    this.timeoutUnlock.start(0, this.unlock);
+  };
+
+  private unlock = () => {
+    if (this.lockCount === 0 && this.restore) {
+      this.restore?.();
+      this.restore = null;
+    }
+  };
```
romgrk, line 167: The unlock timeout avoids unlock-lock sequences that create a lot of work.

Screenshot From 2025-05-14 04-16-43

packages/react/src/utils/useScrollLock.ts
```diff
@@ -137,66 +138,94 @@ function preventScrollStandard(referenceElement: Element | null) {
...
+  timeoutLock = Timeout.create();
+  timeoutUnlock = Timeout.create();
+
+  acquire(referenceElement: Element | null) {
+    this.lockCount += 1;
+    if (this.lockCount === 1 && this.restore === null) {
+      this.timeoutLock.start(0, () => this.lock(referenceElement));
+    }
+    return this.release;
+  }
```
romgrk, line 155: The lock timeout moves the lock logic from the middle of the render phase to after react is done rendering. This has the dual effect of letting rendering have the priority, and making the read/writes much cheaper (because they happen after react has updated the DOM).

### Before

image

### After

Note: the lock scroll runtime is down to 10ms.

image

packages/react/src/utils/useScrollLock.ts
```diff
@@ -85,18 +83,23 @@ function preventScrollStandard(referenceElement: Element | null) {
     const scrollbarWidth = Math.max(0, win.innerWidth - html.clientWidth);
     const scrollbarHeight = Math.max(0, win.innerHeight - html.clientHeight);
 
+    // Avoid shift due to the default <body> margin. This does cause elements to be clipped
+    // with whitespace. Warn if <body> has margins?
+    const marginY = parseFloat(bodyStyles.marginTop) + parseFloat(bodyStyles.marginBottom);
+    const marginX = parseFloat(bodyStyles.marginLeft) + parseFloat(bodyStyles.marginRight);
```
romgrk, line 89: `bodyStyles` is a computed styles object, which is [live](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle#return_value).

packages/react/src/utils/useScrollLock.ts
```diff
@@ -137,66 +140,92 @@ function preventScrollStandard(referenceElement: Element | null) {
...
+        doc.body.style.userSelect = 'none';
+        doc.body.style.webkitUserSelect = 'none';
+        return () => {
+          doc.body.style.userSelect = originalUserSelect;
+          doc.body.style.webkitUserSelect = originalWebkitUserSelect;
+        };
+      }
+      return undefined;
+    }, [mounted, open, referenceElement]);
+  }
```
atomiks, line 223: Not sure if this is worth it? We should try to be React Compiler-compliant as much as possible and avoid disabling lint rules. It also seems a bit risky based on how `isWebKit` is calculated and could lead to a crash in some scenarios (I thought server pre-rendering -> client hydration in Safari might be an issue, but can't reproduce)

romgrk: > It also seems a bit risky based on how isWebKit is calculated

Making those functions into constants is on my todo list, but I didn't want to busy this PR with those changes.

> We should try to be React Compiler-compliant as much as possible and avoid disabling lint rules

I don't insist much on keeping this optimization, but out of curiosity are you aware of any upcoming issues with that?

atomiks: Last time I tested it (awhile ago, last year) it bailed out of optimizing a hook or component if a lint rule was disabled

packages/react/src/utils/useScrollLock.ts
```diff
@@ -137,66 +140,92 @@ function preventScrollStandard(referenceElement: Element | null) {
   };
 }
 
+class ScrollLocker {
+  lockCount = 0;
+  restore = null as (() => void) | null;
```
flaviendelangle, line 145:
```suggestion
  restore: (() => void) | null = null;
```

romgrk: I find the first style more readable. I think the `$identifier = $value [as $optional_type]` is more parsable to the eye because it has the optional part at the end, whereas `$identifier[: $optional_type] = $value` has the optional part in the middle, and sometimes types are hard to read.

romgrk: Wdym? Afaiu, it should be equivalent due to type inference.

romgrk: >  "sufficiently" is quite blurry

Not that blurry, for a cast the `Y` in `X as Y` needs to be an element of the `X` set. Types are like sets that contain a number of elements, which are other types - `never` is the empty set, `any` is the set containing all types (and so is `unkown` for practical reasons, but that's a gory detail of TS).

In your example, `a` is infered as `number` in the first line. `2` is an element of the `number` type, therefore the cast is valid. `a` is not typed as `1` because inference doesn't narrow types lower than JS primitives, unless specified with `as const`.

`Map` is not an element of the `number` set, thefore the second cast is invalid.

> In your code, I'm not sure sure you have an actual scenario where it could be problematic.

You make an interesting point, but I'd rather see an actual case where it's problematic before starting to write less readable code.

atomiks (approved):
This looks good, but I think the `isWebkit()` conditional hook check should be removed. I don't think the micro-optimization is worth potential issues, and we should try to keep lint rules enabled as much as possible.

Issue #1904, opened 2025-05-12
[field] Option or way to run the validation function after checkValidity


benjavicente:
# Feature request

## Summary

In a Field, I want to run a validation only after the native input validation has run, such that a `customError`is only shown if there aren't any non-custom errors. 

Here is a codepen: https://codesandbox.io/p/sandbox/clever-chihiro-ssw6fd?file=%2Fsrc%2FApp.tsx

See that `Field.Error` with `customError` is rendered when the pattern is also invalid.

On those cases, the preferred way would be to use a validation library like zod, or there is an intended way to run a validation function **after** the validity of the input was checked?

A way to hack around this limitation would be to get a ref of the input element and run `checkValidity` inside the validation function. Uncomment the first line of the validation function to see that.

## Examples in other libraries

Couldn't find a similar example.

## Motivation

My use case is to run a [chek digit](https://en.wikipedia.org/wiki/Check_digit) function **after** the field is completed (passing required validation) and if it matches a regular expression (patter validation). I want to avoid repeating code, ej validating the pattern on the validation function and on the element itself.

mj12albert:
Thanks for the sandbox @benjavicente, I've been testing a bunch of similar field validation setups 

It may be possible to use [`Field.Validity`](https://base-ui.com/react/components/field#validity) instead of `Field.Error` to achieve this, it accepts a function child with the validity state as an arg, though I haven't tested this myself

For Field.Error if you also consider how the errors are cleared/revalidated, I've found that the UX feels better when using only the native validation for required, and doing everything else in the validate function, that way you have more control over the order and timing of all the checks

mj12albert:
https://codesandbox.io/p/sandbox/busy-shockley-yrnzyn?file=%2Fsrc%2FApp.tsx%3A1%2C1-2%2C1

Here's a sandbox using a build from https://github.com/mui/base-ui/pull/1840

To get the behavior you describe it seems easiest to do everything with the `validate` function. `<Field.Control required />` is only for setting the HTML required attribute, the function will do the actual check as well.

PR #1926, opened 2025-05-15
[field] Run validate function after native validations

mj12albert:
Demo: https://codesandbox.io/p/devbox/serene-brahmagupta-xgncdd

Currently a custom validation function will run before any native validations (e.g. required) 

It makes more sense for the basic native checks to run before custom (presumably more complicated) checks, otherwise you'd have to recreate the native checks in the custom function

Closes https://github.com/mui/base-ui/issues/1904


atomiks (approved):
Agree this makes more sense - would be good to add a test case for this

mj12albert:
> Agree this makes more sense - would be good to add a test case for this

Added, originally tried to use `minLength` in the test like in the sandbox but it wouldn't trigger using fireEvent or userEvent.keyboard (waitFor, flushMicrotasks etc) 🤔

Issue #1864, opened 2025-05-05
[Field] Support array values


mj12albert:
Currently it's not possible to integrate Field with a range slider (2+ thumbs, the value is an array), there are distinct `<input>` elements for each thumb and they can't be given different `name`s (the one received via Field can't be overwritten by setting a different name with a ref or something)

This will only include the value of the last thumb in the form values:
```jsx
<Form
  onSubmit={async (event) => {
    event.preventDefault();
    const formData = new FormData(event.currentTarget);
    const formValue = Object.fromEntries(formData as any);
    console.log('submit', formValue);
  }}
>
  <Field.Root name="priceRange">
    <Slider.Root defaultValue={[33, 50]} format={FORMAT}>
      <Slider.Value />
      <Slider.Control>
        <Slider.Track>
          <Slider.Indicator />
          <Slider.Thumb/>
          <Slider.Thumb />
        </Slider.Track>
      </Slider.Control>
    </Slider.Root>
  </Field.Root>
  <button type="submit">Submit</button>
</Form>
```

Array values would also be needed for components like multi-select or combobox

mj12albert:
Currently for CheckboxGroup, passing `name` to `Field` causes a bug where all the checkboxes will have the same value (and become checked/unchecked together inseparably): https://github.com/mui/base-ui/blob/master/packages/react/src/checkbox/root/useCheckboxRoot.ts#L70-L72

To use it in a form, it only works like this:

```jsx
<Form>
  <Field.Root>
    <CheckboxGroup defaultValue={[]}>   
      <Checkbox.Root name="fuji-apple" />
      <Checkbox.Root name="gala-apple" />
      <Checkbox.Root name="granny-smith-apple" />
    </CheckboxGroup>      
  </Field.Root>
</Form>
```

Resulting in this FormData:
```js
{
    "fuji-apple": "off",
    "gala-apple": "on",
    "granny-smith-apple": "on"
}
```

Given FormData's limitations, I think the best would be:
```js
<Field.Root name="apple />

FormData.get('apple')

// return `'["gala-apple","granny-smith-apple"]'`
```

Though because of some React 19 features we can't completely dismiss FormData: https://react.dev/reference/react-dom/components/form#handle-form-submission-with-a-server-function

mj12albert:
This was completed in https://github.com/mui/base-ui/pull/1948

We decided that array values should work well with `formData.getAll()`

PR #1929, opened 2025-05-15
[slider][form][field] Integrate range sliders with Form and Field

mj12albert:
Part of https://github.com/mui/base-ui/issues/1864

Preview: https://deploy-preview-1929--base-ui.netlify.app/experiments/slider/form

Use ARIA-hidden/visually-hidden `<input>`s in the root that is wired to Field and Form instead of the thumb's `<input type="range">`.

Range sliders render one hidden input per value in the value array.

There are some limitations:

1) When the slider's numerical value goes through `FormData` as part of form submission, the values will become strings.
```js
formData.get('volume') // => '50'

// range slider
formData.getAll('priceRange') // => ['25','50']
```

2) `Object.fromEntries(formData)` no longer works:
```js
Object.fromEntries(formData) // => {priceRange: '50'}
```
`for...of` is needed to iterate the form data correctly:
```js
for (let [key, value] of data) {
  // `key` is the `name` and isn't unique
}
```


packages/react/src/slider/thumb/SliderThumb.tsx
```diff
@@ -139,15 +137,9 @@ export const SliderThumb = React.forwardRef(function SliderThumb(
...
-  const {
-    getInputValidationProps,
-    inputRef: inputValidationRef,
-    commitValidation,
-  } = useFieldControlValidation();
+  const { commitValidation, getValidationProps } = useFieldControlValidation();
 
   const thumbRef = React.useRef<HTMLElement>(null);
-  const inputRef = React.useRef<HTMLInputElement>(null);
-  const mergedInputRef = useForkRef(inputRef, inputRefProp, inputValidationRef);
```
mj12albert, line 150: @atomiks `inputValidationRef` was being set twice, here and `inputValidationRef.current = element.querySelector<HTMLInputElement>('input[type="range"]')` in the root

Now that it's only set once with `useForkRef` in the root, the ref isn't registered when `commitValidation()` runs 🤔

mj12albert: Fixed by calling `useFieldControlValidation` once in the root and forwarding via context as needed

atomiks:
> One limitation is that if the range slider's value goes through FormData as part of form submission, it'll become a string ('[2,4]') and will need to be manually JSON.parse-ed

I added multiple hidden inputs for the multi-select feature

```jsx
{multiple &&
  Array.isArray(value) &&
    value.map((v) => <input key={v} type="hidden" name={name} value={v} />)}
```

```jsx
const data = new FormData(event.currentTarget);
const submitted = data.getAll(name); // normal array
```

mj12albert:
> I added multiple hidden inputs for the multi-select feature

For slider the values are always numbers but the formData methods will always return a string so this would be more inconvenient to convert back to number though: `priceRange: ['10', '20']` vs `priceRange: '[10,20]'`. It would be the same for multiselect with non-string values I guess? @atomiks 

Though I think we should have `onFormSubmit` so you don't have to manually process a FormData if you don't want to

mj12albert:
Changed it to use multiple hidden inputs (one per value in the value array)

https://react.dev/reference/react-dom/components/form#handle-form-submission-with-a-server-function

Since these features are stable in React 19 and use FormData it's probably better to stick closely to it

packages/react/src/form/Form.tsx
```diff
@@ -55,10 +57,8 @@ export const Form = React.forwardRef(function Form(
     }
   }, [errors, focusControl]);
 
-  const state = React.useMemo<Form.State>(() => ({}), []);
-
   const element = useRenderElement('form', componentProps, {
-    state,
+    state: EMPTY,
```
atomiks, line 61: `state` is optional now

atomiks:
> Object.fromEntries(formData) no longer works:
Object.fromEntries(formData) // => {priceRange: '50'}

So `.getAll()` is recommended instead for form inputs that have multiple values like multiselect and range sliders?

mj12albert:
> So `.getAll()` is recommended

Not a lot of specs or documentation about this, but all signs suggest that the intended use is for array-like values, for example native `<select multiple>`: https://codesandbox.io/p/sandbox/4kmkjl

```html
<form>
  <select multiple name="toppings">
    <option>Pineapple</option>
    <option>Prosciutto</option>
  </select>
</form>
```
```js
formData.getAll("toppings") // => ['Pineapple', 'Prosciutto']
```
Same works for `<input type="file" multiple>`, so that gives me some confidence that this is the right way ~

packages/react/src/slider/root/SliderRoot.tsx
```diff
@@ -336,6 +332,38 @@ export const SliderRoot = React.forwardRef(function SliderRoot<
...
+                key={`${name}-input-${index}`}
+                {...fieldControlValidation.getInputValidationProps({
+                  type: 'hidden',
+                  disabled,
+                  name,
+                  ref: inputRef,
+                  value,
+                  style: visuallyHidden,
+                  tabIndex: -1,
+                  'aria-hidden': true,
```
atomiks, line 348: Extraneous props

atomiks (approved):

PR #1934, opened 2025-05-16
[core] Performance/refactor: `useRender`

romgrk:
This is both perf/refactor. Make `useRender` match the `useRenderElement` API.

### DIFF
--- a/docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
+++ b/docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
@@ -128,7 +128,7 @@ const Text = React.forwardRef(function Text(
     render,
   });
 
-  return element();
+  return element;
 });
 ```
 
--- a/docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
+++ b/docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
@@ -47,11 +47,9 @@ The `useRender` hook lets you build custom components that provide a `render` pr
 />
 
 ```tsx title="renderElement function" "renderElement"
-const { renderElement } = useRender({
+const element = useRender({
   // Input parameters
 });
-
-const element = renderElement();
 ```
 
 ## Examples
@@ -103,13 +101,13 @@ In React 19, `React.forwardRef()` is not needed when building primitive componen
 function Text({ render = <p />, ...props }: TextProps) {
   const internalRef = React.useRef<HTMLElement | null>(null);
 
-  const { renderElement } = useRender({
-    render,
+  const element = useRender({
+    ref: internalRef,
     props,
-    refs: [internalRef],
+    render,
   });
 
-  return renderElement();
+  return element;
 }
 ```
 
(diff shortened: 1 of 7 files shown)


packages/react/src/use-render/useRender.ts
```diff
@@ -12,25 +12,14 @@ export function useRender<
...
-      ),
-      disableStyleHooks: true,
-    },
-  );
+  const element = useRenderElement(undefined, renderParams, renderParams);
 
-  return {
-    renderElement,
-  };
+  return element;
```
romgrk, line 22: I don't insist on the mutable `.disableStyleHooks` line, I just wrote it this way to highlight what this refactor made evident: `useRender` is really just `useRenderElement` but with style hooks turned off.

The prop/ref destructuring & handling logic was all unnecessary: it's duplicated with the logic in `useRenderElement` whic already does all that.

packages/react/src/use-render/useRender.ts
```diff
@@ -63,9 +52,9 @@ export namespace useRender {
      */
     render: RenderProp<State>;
     /**
-     * Refs to be merged together to access the rendered DOM element.
+     * The ref to apply to the rendered element.
      */
-    refs?: React.Ref<RenderedElementType>[];
+    ref?: React.Ref<RenderedElementType> | React.Ref<RenderedElementType>[];
```
romgrk, line 57: I standardized this with `useRenderElement`, both use `ref` for the propname, while still allowing an array.

docs/src/app/(public)/(content)/react/utils/use-render/page.mdx
```diff
@@ -124,13 +122,13 @@ const Text = React.forwardRef(function Text(
...
-    render,
+  const element = useRender({
+    ref: [forwardedRef, internalRef],
     props,
-    refs: [forwardedRef, internalRef],
+    render,
   });
 
-  return renderElement();
+  return element();
```
atomiks, line 131: Extra `()` here

atomiks (commented):
What about the conditional rendering optimization with an early `null` return, or is what not much of a concern in your perf tests?

romgrk:
> What about the conditional rendering optimization with an early null return, or was that not much of a concern in your perf tests?

I have not seen a case where the lazyness was actually needed, we seem to call `renderElement()` right away every time. If we do need it, we can introduce a matching wrapper for `useRenderElementLazy`.

atomiks:
Searching `if (!shouldRender)` reveals most of the cases where an early return is used.

romgrk:
Ah right, I didn't look at the deprecated `useComponentRender`. I've added the lazy variant here and updated the docs.

There could be more lazyness though. Laziness is set just at the `evaluateRenderProp()` boundary, but the whole props merging logic before that could be avoided as well. I'll keep this PR focused on the current change though, I'll do that change separately.

michaldudak (approved):

romgrk:
About the lazy `useRender`, I just realized that `useRender` is exclusively for the public API. The internal `useRenderElement` already has a lazy variant. To avoid increasing the public API and because the docs generator script was failing on the new function and I don't want to spend time debugging it now, I've reverted that change. If we really need to publicly expose the lazy variant, we can add it anytime.

PR #1937, opened 2025-05-17
[core] Refactor Popover root

romgrk:
I was reading the code here and I noticed there is a lot of logic repetition between the hook and the component. It might be beneficial to have a convention for who handles setting up the context value: the root component or the root hook. This will be much easier for maintainability than having duplicated logic, and it also avoids a couple of destructuring & `useMemo` calls.

I think a similar refactor can be applied elsewhere (the Dialog component seems to suffer from similar problems, there might be more).


atomiks:
We're removing hooks (that aren't reusable) entirely now. That also reduces repetition

romgrk:
Gotcha, should I update this PR to move all the logic inside `PopoverRoot`?

atomiks:
Yeah — this would ideally include all other components as well, but that's a slightly bigger refactor

atomiks (approved):

PR #1938, opened 2025-05-19
[popover] Fix nested `openOnHover`

### LINKED ISSUE #1932: [popover] Nested popover when both have `openOnHover` closes when hovering over it
# Bug report

When both the root popover and nested popover have `openOnHover`, the nested one closes when trying to enter it.

## Reproducible example

https://codesandbox.io/p/sandbox/musing-sun-4wg359?file=%2Fsrc%2FApp.tsx

## Base UI version

1.0.0-alpha.8

## Workaround

Portal the child into the parent `Popup` with a ref:
https://codesandbox.io/p/sandbox/musing-sun-4wg359

Of course, this has some limitations with clipping/default styles.

## Potential fix

A `FloatingTree` setup would solve this. It would be great though if Floating UI used React props instead of native event listeners; this way we might be able to check if the event originated inside the React tree even despite differing DOM contexts. This is already done for click outside behavior. https://github.com/floating-ui/floating-ui/issues/3060#issuecomment-2885609644

atomiks:
Fixes #1932

Possible questions:

- Should the nested popover disable the `useHover` of the parent hovered popover like `Menu` does? This means it requires an outside press to dismiss rather than a hover out.
- `FloatingTree` can't generalize between different floating components (e.g. Popover `openOnHover` -> Menu `openOnHover` will close the Popover unexpectedly). It is possible that dot point 1 would solve this if we detected it landed in a `data-floating-ui-portal` (but maybe not all cases where we land in a new popup should prevent closing the parent).

https://codesandbox.io/p/devbox/happy-wiles-2ymwmv


michaldudak:
> FloatingTree can't generalize between different floating components (e.g. Popover openOnHover -> Menu openOnHover will close the Popover unexpectedly). It is possible that dot point 1 would solve this if we detected it landed in a data-floating-ui-portal (but maybe not all cases where we land in a new popup should prevent closing the parent).

We can introduce a "container" context across the library, similarly to how menu `parent` context works. Components that can have other components nested (like dialogs, toolbars, popovers, etc.) could set up a context the will make children aware if they are placed inside. Such context could expose methods such as `close` so that children can control parent's behavior to a degree.

PR #1940, opened 2025-05-19
[core] Refactor: "style hooks" to "state attributes"

romgrk:
Get rid of the "style hooks" terminology which is unclear.

Now that I read the PR, I feel like it could be simply `stateMapping` instead of `stateAttributesMapping`. The "attributes" token would be missing, but it would be less verbose and "state mapping" is clear enough as a concept, imo. Opinions welcome, I could go with either.


michaldudak:
@colmtuite, are you OK with changing this term?

colmtuite:
@michaldudak @romgrk Sure

michaldudak:
@romgrk are you going to update this PR? I can take it over if you'd like.

romgrk:
I'm working on the new virtualization engine for the time being, so you can take over this PR if you want it merged.

PR #1941, opened 2025-05-19
[field] Enable custom validation based on other form values

mj12albert:
Demo: https://codesandbox.io/p/sandbox/eloquent-wright-sff4fm

Adds a `formValues` object as the 2nd argument to `validate`, which enables validating a field based on the value of another field, e.g. `confirmPassword` and `password`


packages/react/src/radio-group/useRadioGroup.ts
```diff
@@ -0,0 +1,188 @@
+'use client';
```
atomiks: Stale file

atomiks (approved):

PR #1943, opened 2025-05-20
[select] Use `Select.ItemText` ref to grab default text content

### LINKED ISSUE #1942: Select: `Select.Value` shows SVG description
# Bug report

## Current behavior

When:

- Composing `Select.Item` with React Router's `Link`
- Using `desc` for SVG within `Select.ItemIndicator`

`Select.Value` shows the SVG `desc`:

https://github.com/user-attachments/assets/f5340693-a1a3-47c7-9f8c-ca8437664749

## Expected behavior

`Select.Value` does not show the SVG `desc`.

## Reproducible example

https://codesandbox.io/p/sandbox/wonderful-banzai-83wrmg

## Base UI version

1.0.0-alpha.8

Chrome
## Which OS are you using?

Mac OS

N/A

## Additional context

For context on why we are trying to use `Link` with `Select`, I am trying to implement this:

https://github.com/user-attachments/assets/54c75c76-6c46-43a7-88fb-da46067978ee

atomiks:
Fixes #1942

### DIFF
--- a/packages/react/src/composite/list/useCompositeListItem.ts
+++ b/packages/react/src/composite/list/useCompositeListItem.ts
@@ -6,6 +6,7 @@ import { useCompositeListContext } from './CompositeListContext';
 export interface UseCompositeListItemParameters<Metadata> {
   label?: string | null;
   metadata?: Metadata;
+  getItemText?: () => string | null;
 }
 
 interface UseCompositeListItemReturnValue {
@@ -19,7 +20,7 @@ interface UseCompositeListItemReturnValue {
 export function useCompositeListItem<Metadata>(
   params: UseCompositeListItemParameters<Metadata> = {},
 ): UseCompositeListItemReturnValue {
-  const { label, metadata } = params;
+  const { label, metadata, getItemText } = params;
 
   const { register, unregister, map, elementsRef, labelsRef } = useCompositeListContext();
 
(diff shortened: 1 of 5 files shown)


packages/react/src/composite/list/useCompositeListItem.ts
```diff
@@ -33,13 +34,23 @@ export function useCompositeListItem<Metadata>(
 
       if (index !== null && node !== null) {
         elementsRef.current[index] = node;
+
         if (labelsRef) {
           const isLabelDefined = label !== undefined;
-          labelsRef.current[index] = isLabelDefined ? label : (node?.textContent ?? null);
+          let textContent = node.textContent;
```
michaldudak, line 40: Don't call `.textContent` if the value will be replaced by `getItemText` return value.

packages/react/src/composite/list/useCompositeListItem.ts
romgrk, line 1: Just a general comment, but building a `Select` that has absolutely no knowledge or access to the data model seems to be causing a lot of hacks. Having to mount items to retrieve their `.textContent` is one of them. It's also going to make some optimizations impossible, for example we can't use virtualization if we don't know anything about the options length. I wonder if this API will be workable in the long-term.

atomiks: It's a tricky one, for sure. But it's considered the cleanest API that most closely matches the native HTML select and is used by most headless libraries

packages/react/src/composite/list/useCompositeListItem.ts
atomiks: It might not necessarily be replaced if it's nullish

packages/react/src/composite/list/useCompositeListItem.ts
romgrk: Sure but the browser is much more low-level, so it has more tricks and it can afford spending more operations on that sort of processing thanks to using a native language.

A clean API is nice, but some folks (me included) would favor optimizing INP instead of prioritizing a clean API. I wouldn't be surprised to have requests for a more performant API once/if bigger players start using base-ui for large traffic websites.

packages/react/src/select/item/SelectItem.tsx
```diff
@@ -125,7 +128,13 @@ export const SelectItem = React.forwardRef(function SelectItem(
...
   const { value: valueProp = null, label, ...otherProps } = props;
 
-  const listItem = useCompositeListItem({ label });
+  const itemTextRef = React.useRef<HTMLElement | null>(null);
+
+  const getItemText = React.useCallback(() => {
+    return itemTextRef.current?.textContent ?? null;
+  }, []);
+
+  const listItem = useCompositeListItem({ label, getItemText });
```
romgrk, line 137: Isn't this logic duplicated with the `.textContent` read in `useCompositeListItem()`?

Issue #1067, opened 2024-12-12
[checkbox] Parent checkbox API suggestion: CheckboxGroup


vladmoroz:
I've found that the relationship between `<Checkbox parent>` and `<CheckboxGroup allValues={...}>` isn't obvious/easily memorable. You need to maintain in your head that two props on separate components are needed to make it work.

My proposal, instead of:
```jsx
<CheckboxGroup allValues={['a', 'b', 'c']} value={['a']} onValueChange={...}>
  <Checkbox.Root parent>
    <Checkbox.Indicator />
  </Checkbox.Root>

  <Checkbox.Root name="a">
    <Checkbox.Indicator />
  </Checkbox.Root>

  <Checkbox.Root name="b">
    <Checkbox.Indicator />
  </Checkbox.Root>

  <Checkbox.Root name="c">
    <Checkbox.Indicator />
  </Checkbox.Root>
</CheckboxGroup>
```

Set child names on the parent checkbox directly via a `group` prop:

```jsx
<CheckboxGroup value={["a"]} onValueChange={...}>
  <Checkbox.Root group={['a', 'b', 'c']} >
    <Checkbox.Indicator />
  </Checkbox.Root>

  <Checkbox.Root name="a">
    <Checkbox.Indicator />
  </Checkbox.Root>

  <Checkbox.Root name="b">
    <Checkbox.Indicator />
  </Checkbox.Root>

  <Checkbox.Root name="c">
    <Checkbox.Indicator />
  </Checkbox.Root>
</CheckboxGroup>
```

`CheckboxGroup` would pass down the current group value to the parent checkbox which then would be responsible for rendering itself as ticked/unticked/indeterminate, and the `group` prop is memorable because of its relationship to `CheckboxGroup`. Also, one fewer prop to have in the API.

I thought about `groupValues` too, but decided that I'm not a fan of how the `values` part misleads you in thinking that you might need to set a `value` on the child checkboxes instead of the `name` prop; also, `group` is shorter and feels somewhat more declarative

**Search keywords**:

mj12albert:
IIRC the reason for this is that putting `allValues={['a', 'b', 'c']}` (by any name) one level higher than `<Checkbox.Root name="a">` is necessary to SSR the checked state correctly

(I think if it weren't for this, the group value could even just be derived from the `name`s of the individual checkboxes instead of being a prop)

vladmoroz:
[...]

PR #1948, opened 2025-05-20
[form][checkbox group] Enable submitting checkbox group value as one field

mj12albert:
Part of https://github.com/mui/base-ui/issues/1864#issuecomment-2892623576

Before: https://codesandbox.io/p/sandbox/naughty-cookies-9yvq86

```jsx
<Form>
  <Field.Root name="apple>
    <CheckboxGroup defaultValue={[]}>
      <Checkbox.Root
        // `value` doesn't work either
        name="fuji-apple"
      >
    </CheckboxGroup>
  </Field.Root>
</Form>
```

(Native forms don't really have the concept of checkbox groups at all)


mj12albert:
This adds a new option to pass `name` to `Field.Root` (in addition to `name` on the `Checkbox`) to result in single Form Data entry with a string value:
```js
formData.get('apple') // => 'gala-apple,fuji-apple'

Object.fromEntries(formData) // => {apple: 'gala-apple,fuji-apple'}
```
Or maybe `name` on `Field.Root` + `value` on `Checkbox.Root`?

Otherwise it could just be the same as before, though maybe it's odd that `name` goes on the individual CheckboxRoots and not the Field, what do you think? @atomiks

packages/react/src/checkbox/root/useCheckboxRoot.ts
```diff
@@ -85,14 +86,15 @@ export function useCheckboxRoot(params: useCheckboxRoot.Parameters): useCheckbox
   }, [id, setControlId]);
 
   useField({
+    enabled: !groupContext,
```
mj12albert, line 89: When there's a checkbox group, the group will be registered as a single field instead of registering each checkbox individually

Otherwise, a custom `validate` function will be called incorrectly with individual checkbox values upon form submission instead of the group value

docs/src/app/(public)/(content)/react/components/checkbox-group/page.mdx
```diff
@@ -40,3 +40,21 @@ A checkbox that controls other checkboxes within a `CheckboxGroup` can be create
 ### Nested parent checkbox
 
 <Demo compact path="./demos/nested" />
+
+### Form integration
+
+To use checkbox group in a form, pass the checkbox group's `name` to a `Field`, then use the `render` prop to render the field as a `Fieldset` in order to label the group.
+
+```tsx title="Using checkbox group in a form"
```
atomiks, line 48: "`CheckboxGroup`" instead of "checkbox group" reads better here

packages/react/src/checkbox-group/CheckboxGroup.test.tsx
```diff
@@ -254,12 +255,13 @@ describe('<CheckboxGroup />', () => {
...
-        'fuji-apple',
-        'granny-smith-apple',
-      ]);
+
+      // expect(validateSpy.callCount).to.equal(4);
+      // expect(validateSpy.args[3][0]).to.deep.equal([
+      //   'gala-apple',
+      //   'fuji-apple',
+      //   'granny-smith-apple',
+      // ]);
```
atomiks, line 264: Should be removed?

packages/react/src/checkbox-group/CheckboxGroup.tsx
```diff
@@ -25,20 +30,41 @@ export const CheckboxGroup = React.forwardRef(function CheckboxGroup(
...
   });
 
+  const id = useBaseUiId(idProp);
+
+  const controlRef = React.useRef<HTMLButtonElement>(null);
+  const handleRef = useEventCallback((element: HTMLElement | null) => {
+    if (element) {
+      controlRef.current = element.querySelector<HTMLButtonElement>('button[role="checkbox"]');
+    }
+  });
```
atomiks, line 57: Is this just linking the first checkbox? Would be good to avoid querySelector, if possible.

docs/src/app/(public)/(content)/react/components/checkbox-group/demos/form/css-modules/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+          className={styles.CheckboxGroup}
+        >
+          <label className={styles.Item}>
+            <Checkbox.Root value="anchovies" className={styles.Checkbox}>
+              <Checkbox.Indicator className={styles.Indicator}>
+                <CheckIcon className={styles.Icon} />
+              </Checkbox.Indicator>
+            </Checkbox.Root>
+            Anchovies
+          </label>
```
atomiks, line 40: Can each checkbox have its own `Field`? That means they can use `Field.Label`, which prevents selection on double click

packages/react/src/checkbox-group/CheckboxGroup.tsx
mj12albert: Updated in https://github.com/mui/base-ui/pull/1948/commits/97088b9440fe83b773a2635fd7f6858ae750e039

atomiks:
I'm assuming `name` is still fine for standalone checkboxes outside of a group?

How about parent checkboxes which still use `name` instead of `value`? https://deploy-preview-1948--base-ui.netlify.app/react/components/checkbox-group#parent-checkbox

mj12albert:
> I'm assuming `name` is still fine for standalone checkboxes outside of a group?

Yup ~ there were already tests for single checkboxes integrating with field/form already

> How about parent checkboxes

@atomiks Updated in https://github.com/mui/base-ui/pull/1948/commits/b3edef5c3571a68657cd3ba3f5dff3b2bbc87939, child checkboxes should also use `value`, and parent checkbox is excluded from form submission: https://codesandbox.io/p/sandbox/sleepy-smoke-s8kdnn

atomiks (approved):
Much nicer as the `allValues` prop now makes more sense for parent checkboxes. We could close this issue then, based on my comment the current API is needed for SSR: https://github.com/mui/base-ui/issues/1067

PR #1952, opened 2025-05-20
[core] Performance: inline style tags

romgrk:
Avoid HTML parsing/reflow/restyle costs. The blue stacks here are the inline `<style>` tags' content being re-parsed everytime.

Screenshot From 2025-05-20 18-07-54

### DIFF
--- a/packages/react/src/utils/styles.tsx
+++ b/packages/react/src/utils/styles.tsx
@@ -1,0 +1,8 @@
+export const STYLE_DISABLE_SCROLLBAR = {
+  className: 'base-ui-disable-scrollbar',
+  element: (
+    <style href="/base-ui-disable-scrollbar">
+      {`.base-ui-disable-scrollbar{scrollbar-width:none}.base-ui-disable-scrollbar::-webkit-scrollbar{display:none}`}
+    </style>
+  ),
+};
--- a/packages/react/src/utils/styles.tsx
+++ b/packages/react/src/utils/styles.tsx
@@ -1,3 +1,5 @@
+import * as React from 'react'
+
 export const STYLE_DISABLE_SCROLLBAR = {
   className: 'base-ui-disable-scrollbar',
   element: (
--- a/packages/react/src/merge-props/mergeProps.ts
+++ b/packages/react/src/merge-props/mergeProps.ts
@@ -200,7 +200,10 @@ export function makeEventPreventable<T extends React.SyntheticEvent>(event: Base
   return event;
 }
 
-function mergeClassNames(ourClassName: string | undefined, theirClassName: string | undefined) {
+export function mergeClassNames(
+  ourClassName: string | undefined,
+  theirClassName: string | undefined,
+) {
   if (theirClassName) {
     if (ourClassName) {
       // eslint-disable-next-line prefer-template
--- a/packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
+++ b/packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
@@ -8,6 +8,7 @@ import { useEventCallback } from '../../utils/useEventCallback';
 import { SCROLL_TIMEOUT } from '../constants';
 import { getOffset } from '../utils/getOffset';
 import { ScrollAreaScrollbarDataAttributes } from '../scrollbar/ScrollAreaScrollbarDataAttributes';
+import { STYLE_DISABLE_SCROLLBAR } from '../../utils/styles';
 import { useBaseUiId } from '../../utils/useBaseUiId';
 import { useTimeout } from '../../utils/useTimeout';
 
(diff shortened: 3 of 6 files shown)


packages/react/src/utils/styles.tsx
romgrk, line 1: If you're adding inline style tags for functional styles:
 - Avoid dynamic styles, prefer static style with classnames.
 - Avoid `dangerouslySetInnerHtml`
 - Avoid conditionally adding the style to the DOM (e.g. the select style was added back & forth when the popup was opened/closed) unless it's a stable condition.
 - Always add an `href` value, so that at least on R19 it gets deduplicated.

packages/react/src/utils/styles.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+
+export const STYLE_DISABLE_SCROLLBAR = {
+  className: 'base-ui-disable-scrollbar',
+  element: (
+    <style href="/base-ui-disable-scrollbar">
```
Janpot, line 6: Just a question in the margin, but has there been any discussion about CSP yet? Should base ui be built with strict csp in mind?

michaldudak: We touched the topic briefly some time ago but we were using style attributes at that time. We can revisit it now that the approach changed.

packages/react/src/utils/styles.tsx
michaldudak: We used `dangerouslySetInnerHtml` to prevent hydration errors in React <19. During SSR some characters (notably ">" or "'") are HTML-encoded and don't match the non-encoded version on the client, causing an error. Using `dangerouslySetInnerHtml` prevents this as the content is not encoded.
This seems to be fixed in React 19.

It doesn't apply to this particular script as it doesn't contain any of the characters that are subject to encoding, but I'd not make this a general rule. Is `dangerouslySetInnerHtml` much worse in terms of perf?

packages/react/src/utils/styles.tsx
Janpot: This method won't be practical nor scalable as you'd have to provide a hash for every possible tag and you're going to have to ask users to update it on every version update, even patches, unless you consider changes to those tags a breaking change.

romgrk: Last time I mentionned that issue I also suggested using CSS stylesheet imports, it could be a nice option. It solves pretty much any perf concern, and static bundled stylesheets work well with strict CSP.

brijeshb42, line 6: It should also have the precedence attribute for React 19 to hoist it to `head`.

brijeshb42: What's the starting `/` for in `href` ?

romgrk: Fixed, thanks

michaldudak: The main issue here is that we need dynamic styles in some places (like popup positioners). We currently use style attributes to handle it, but I assume an inline style tag could work as well.

packages/react/src/utils/styles.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+
+export const STYLE_DISABLE_SCROLLBAR = {
```
michaldudak, line 3: Let's not overuse all caps. It doesn't matter much for consumers that it's a constant.

```suggestion
export const disableScrollbarStyle = {
```

packages/react/src/utils/styles.tsx
brijeshb42: Inline `<style>` tag will also allow us to mitigate any `nonce` related issues given that we provide a way for users to pass nonce to the components somehow.

Janpot: > given that we provide a way for users to pass nonce to the components somehow.

Really annoying there isn't something like React Context that works on the server. And `React.cache` doesn't seem to be usable for it. I mean, I don't need it to pass the client boundary, just a server version of React Context would be fine for me. I feel like I keep jumping through hoops to get around this limitation.

Janpot: > Maybe an assumption of existing key on globalThis like globalThis.nonce will work

It needs to be tied to the request context in RSC, to avoid multiple concurrent requests poisoning each other. A single global `nonce` would not work for that. This is the part that `React.cache` handles. But it doesn't seem to guarantee being called in strict order in certain server setups.

romgrk: > This is the part that React.cache handles. But it doesn't seem to guarantee being called in strict order in certain server setups.

Do you mind expanding on that? I don't understand the issue.

If we were to disregard warnings, I think it would be possible to access what's needed to hook up a working read-only dual context implementation with [react internals](https://github.com/facebook/react/blob/c0464aedb16b1c970d717651bba8d1c66c578729/packages/react-server/src/flight/ReactFlightAsyncDispatcher.js#L16-L19), something like this:

image

We wouldn't be the first ones to access the internals, [microsoft does it here](https://github.com/microsoft/use-disposable/blob/da61bd108a1451233616cdfdf16e4f304cc6d6ef/src/useIsStrictMode.ts), and I think there's a popular package that also does it but I don't remember the name anymore.

michaldudak (approved):

packages/react/src/utils/styles.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+
+export const styleDisableScrollbar = {
+  className: 'base-ui-disable-scrollbar',
+  element: (
+    <style href="base-ui-disable-scrollbar" precedence="low">
```
brijeshb42: Just an FYI, the actual value of `precedence` (`low`/`high`/`medium` etc) doesn't actually mean anything. The value is just used to determine where the other styles with the same precedence values should end up in the `head`. They can either be clubbed in the same `<style>` tag with the same precendence or go below another `<style>` tag with the same value.

So I'd propose to use namespaced values like `precedence="base-ui:low"` or just simply `precedence="base-ui"`. This way, we can ask users to add an empty `<style precedence="base-ui" />` at the correct location in the `<head>` based on their framework so that base-ui styles have a lower/higher priority than user styles as required.

Janpot: yep, probably want to treat precedence like css layers.

PR #1958, opened 2025-05-21
[core] perf: state classes [draft]

romgrk:


mnajdova:
If we see this affecting and creating perf issues on the styled layer, would it make sense to add this as some kind of config in a centralized place for the library? If the config is enabled, Base UI uses classes as style hooks instead of data attributes. We could try it internally first. Unless of course, adding this overhead is more than using data attributes directly.

PR #1961, opened 2025-05-22
[select] Performance improvements

romgrk:


packages/react/src/composite/list/useCompositeListItem.ts
```diff
@@ -7,32 +7,54 @@ export interface UseCompositeListItemParameters<Metadata> {
   label?: string | null;
   metadata?: Metadata;
   textRef?: React.RefObject<HTMLElement | null>;
+  /** Enables guessing the indexes. This avoids a re-render after mount, which is useful for
+   * large lists. This should be used for lists that are likely flat and vertical, other cases
+   * might trigger a re-render anyway. */
+  indexGuessBehavior?: IndexGuessBehavior;
```
romgrk, line 13: In addition to this, I would like to introduce an `index` prop on item navigation components for users to tell us what the index is, that would avoid having to guess. I plan to do that in a separate PR.

packages/react/src/composite/list/CompositeList.tsx
romgrk, line 1: This now uses reactivity logic similar to `Store`. I didn't re-use store/selectors because this PR is already big enough, but we should aim to migrate our reactivity needs towards that solution eventually.

packages/react/src/utils/store/createSelector.ts
romgrk, line 1: Lots of gymnastics to avoid `...args` here, which is what the whole `a1, a2, a3` logic is about. We accept at most 3 arguments on selectors. We can add more later if needed.

packages/react/test/popupConformanceTests.tsx
```diff
@@ -149,6 +149,9 @@ export function popupConformanceTests(config: PopupTestConfig) {
       });
 
       it('removes the popup when the animation finishes', async ({ skip }) => {
+        // XXX: revisit after feedback from the team
+        skip();
+
```
romgrk, line 154: This is about the popup conformance test using `[data-ending-style]` instead of `[data-closed]` for CSS animation: https://mui-org.slack.com/archives/C02P87NQLJC/p1748454606756699

Afaiu, the test is flawed and needs to be reworked.

packages/react/src/select/positioner/SelectPositioner.tsx
```diff
@@ -44,43 +47,57 @@ export const SelectPositioner = React.forwardRef(function SelectPositioner(
...
+  const modal = useSelector(store, selectors.modal);
+  const value = useSelector(store, selectors.value);
+  const touchModality = useSelector(store, selectors.touchModality);
+  const positionerElement = useSelector(store, selectors.positionerElement);
+  const controlledItemAnchor = useSelector(store, selectors.controlledItemAnchor);
+  const alignItemWithTriggerActive = useSelector(store, selectors.alignItemWithTriggerActive);
 
-  const alignItemWithTriggerActive = mounted && controlledItemAnchor && !touchModality;
+  useFirstRender(() => {
+    // XXX: Should those options be applied on the root component?
```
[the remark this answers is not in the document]
atomiks: It _was_ previously on the Root in `alpha.8` but it made less sense than being on the Positioner  😅

packages/react/src/select/trigger/useSelectTrigger.ts
```diff
@@ -85,12 +93,20 @@ export function useSelectTrigger(
...
         }
+
+        // Saves a re-render on initial click: `typeaheadReady === true` mounts
+        // the items before `open === true`. We could sync those cycles better
+        // without a timeout, but this is enough for now.
+        //
+        // XXX: might be causing `act()` warnings.
+        timeoutFocus.start(0, () => {
+          store.set('typeaheadReady', true);
+        });
```
atomiks, line 109: With a large select, this is much better as focus immediately gets set on the element. Previously it was blocked synchronously

docs/src/components/NoSSR.tsx
```diff
@@ -0,0 +1,18 @@
+'use client';
+import * as React from 'react';
+
+export function NoSSR(props: { defer?: boolean; children: any }): React.JSX.Element {
```
Janpot, line 4: In the past I've sometimes used something like

```tsx

const subscribe = () => () => {}
const getSnapshot = () => false
const getServerSnapshot = () => true

export function NoSsr({ children }) {
  // true during SSR and hydration
  const isSsr = React.useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot);
  return isSsr ? null : children;
}
```

This avoids the extra renders caused by `useLayoutEffect`. Haven't benchmarked it though.

romgrk: Don't care too much about perf, this is just to avoid SSR so I can benchmark the component's rendering client-side.

packages/react/src/select/positioner/SelectPositioner.tsx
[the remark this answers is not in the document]
romgrk: Any objection to moving it back? Might not make as much sense but it's not out of place either, and it lets us avoid some hooks & hacks.

docs/src/components/NoSSR.tsx
Janpot: Sure, but it's a new component introduced in the codebase. 6 months from now, another maintainer may want to use this not realizing this is only to be used under the `DISABLE_DEMO_SSR` feature flag.

packages/react/src/select/store.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  scrollUpArrowVisible: boolean;
+  scrollDownArrowVisible: boolean;
+  controlledItemAnchor: boolean;
+  alignItemWithTriggerActive: boolean;
+};
+
+export type SelectStore = Store<State>;
+
+export const selectors = {
```
flaviendelangle, line 34: This object has a very nice DX.
I only have one concern regarding tree shaking if we have lot of selectors that are only used in parts that most people won't use.
I guess that, with that pattern, they will be bundled anyway.

Maybe the solution would then be to add a 2nd object specific to this feature to avoid always bundling them.

atomiks: Exporting it as module object instead should help

packages/react/src/select/positioner/SelectPositioner.tsx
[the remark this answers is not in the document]
atomiks: It was confusing to users so I'd prioritize them over internal DX - I was using a ref to pass the state to the trigger from the positioner since it occurs inside an event and doesn't need state. That prevented any weird hacks.

docs/src/components/NoSSR.tsx
michaldudak: There is an `unstable_NoSSR` component in Base UI already.

packages/react/src/composite/list/useCompositeListItem.ts
michaldudak: We are considering passing the items array directly to the root for some components (like Select). If this is accepted, this code shouldn't be necessary.

packages/react/src/utils/useFirstRender.ts
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export function useFirstRender(fn: Function) {
```
michaldudak: For consistency with `useOnMount`
```suggestion
export function useOnFirstRender(fn: Function) {
```

packages/react/src/select/root/SelectRootContext.ts
```diff
@@ -67,3 +50,11 @@ export function useSelectRootContext() {
   }
   return context;
 }
+
+export function useSelectFloatingContext() {
+  const context = React.useContext(SelectFloatingContext);
+  if (context === null) {
+    throw new Error('useSelectFloatingContext must be used within a SelectRoot');
```
michaldudak: Error messages should be written according to the pattern: `Base UI: [Context] is missing. [Component] parts must be placed within <[Context provider component]>.`

packages/react/src/select/store.ts
romgrk: Module objects can create live bindings in some transpilers/bundlers (to support `export let mutableVariable = x`), which are implemented with `get` property descriptors and are bad for perf. I'd leave as it is for now, and we can keep thinking how to optimize the bundle size if those really affect it.

<details><summary>Live-binding transpilation example</summary>

image

</detail>

romgrk: <details><summary>Live-binding transpilation example</summary>

image

</detail>

packages/react/src/select/positioner/SelectPositioner.tsx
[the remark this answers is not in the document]
romgrk: Positioner is a child of root so it doesn't render before root has render, therefore if a prop is needed to initialize the store it won't be available at the first render, which will create a re-render. I don't see a way to prevent it. This works for us because we test with the default select values (`alignItemWithTrigger=true`), but if users set different ones from the defaults, that will affect their performance.

> It was confusing to users 

How confusing? What kind of feedback do we have on that?

packages/react/src/composite/list/useCompositeListItem.ts
romgrk: Ok, we can remove that logic then.

packages/react/src/select/root/SelectRootContext.ts
romgrk: That doesn't seem to be consistent, this one was copy-pasted from above. I've added the `Base UI:` prefix for now, and we can make all the messages consistent with one another in a different PR.

image

atomiks (approved):

PR #1963, opened 2025-05-23
[tooltip] Optional `TrackCursorAxis` component

### LINKED ISSUE #1792: Optional features tree-shaking
# Feature request

Currently, there are optional features that are part of components that add a bunch of bundle size even if not used. `trackCursorAxis` on `Tooltip` is one such example.

If we want to add more features to Tooltip like truncation logic, `inline()` logic, etc, this adds further cost but most users don't need that by default and shouldn't need to incur.

One possible solution is a `features` prop where you pass in 'features' using imports so bundlers can remove them if they aren't called:

```jsx
<Tooltip.Root 
  features={[
    Tooltip.trackCursorAxis('both'),
    Tooltip.checkTruncation(),
  ]}
>
```

Another is component providers:

```jsx
<Tooltip.TrackCursorAxis value="both">
  <Tooltip.CheckTruncation>
    <Tooltip.Root />
  </Tooltip.CheckTruncation>
</Tooltip.TrackCursorAxis>
```

This would only apply to non-essential features, as there will still be a baseline of expected behavior and props

atomiks:
Closes #1792

### DIFF
--- a/packages/react/src/tooltip/root/useTooltipRoot.ts
+++ b/packages/react/src/tooltip/root/useTooltipRoot.ts
@@ -2,7 +2,6 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import {
   safePolygon,
-  useClientPoint,
   useNextDelayGroup,
   useDismiss,
   useFloatingRootContext,
@@ -32,7 +31,6 @@ export function useTooltipRoot(params: useTooltipRoot.Parameters): useTooltipRoo
     onOpenChange: onOpenChangeProp,
     defaultOpen = false,
     hoverable = true,
-    trackCursorAxis = 'none',
     delay,
     closeDelay,
     onOpenChangeComplete,
@@ -128,7 +126,7 @@ export function useTooltipRoot(params: useTooltipRoot.Parameters): useTooltipRoo
     enabled: !disabled,
     mouseOnly: true,
     move: false,
-    handleClose: hoverable && trackCursorAxis !== 'both' ? safePolygon() : null,
+    handleClose: hoverable ? safePolygon() : null,
     restMs() {
       const providerDelay = providerContext?.delay;
       const groupOpenValue =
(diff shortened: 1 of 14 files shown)


packages/react/src/tooltip/trigger/TooltipTrigger.tsx
```diff
@@ -18,13 +19,14 @@ export const TooltipTrigger = React.forwardRef(function TooltipTrigger(
...
   const { open, setTriggerElement, triggerProps } = useTooltipRootContext();
+  const trackCursor = useTooltipTrackCursorAxisContext();
 
   const state: TooltipTrigger.State = React.useMemo(() => ({ open }), [open]);
 
   const element = useRenderElement('button', componentProps, {
     state,
     ref: [forwardedRef, setTriggerElement],
-    props: [triggerProps, elementProps],
+    props: [triggerProps, trackCursor.elementProps?.reference, elementProps],
```
atomiks, line 29: We can bypass the cost (runtime + bundle) of Floating UI's `useInteractions` hook if we pass in the element props manually. This will require 5+ sets of props to be merged though @romgrk

packages/react/src/tooltip/root/useTooltipRoot.ts
```diff
@@ -128,7 +126,7 @@ export function useTooltipRoot(params: useTooltipRoot.Parameters): useTooltipRoo
     enabled: !disabled,
     mouseOnly: true,
     move: false,
-    handleClose: hoverable && trackCursorAxis !== 'both' ? safePolygon() : null,
+    handleClose: hoverable ? safePolygon() : null,
```
atomiks, line 129: Currently unsolved: the root hook needs to know the value of the context provider but it's a child of the Root component

atomiks: Two options:

1. Use a ref to synchronize the `trackCursorAxis` provider value. Requires forking `useHover` to make `handleClose` a callable function so it can read a ref value inside the event handler (added as a new commit)
2. Use state + a re-render to synchronize the `trackCursorAxis` provider value

packages/react/src/tooltip/trigger/TooltipTrigger.tsx
romgrk: uRE already uses the variadic `mergeProps`. It's actually the only place in the codebase that does.

PR #1966, opened 2025-05-23
[menu] Refactor to `useRenderElement`

michaldudak:
Use `useRenderElement` instead of `useComponentRenderer`


packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx
```diff
@@ -61,40 +60,33 @@ export const MenuSubmenuTrigger = React.forwardRef(function SubmenuTriggerCompon
...
         onBlur() {
           if (highlighted) {
             setActiveIndex(null);
           }
         },
       };
     },
-    [getItemProps, highlighted, menuTriggerRef, open, setActiveIndex],
+    [getItemProps, highlighted, open, setActiveIndex],
   );
```
atomiks, line 77: Should inline all prop getters (if possible) now

atomiks (approved):

PR #1967, opened 2025-05-23
[performance] Skip most of useRenderElement logic when unnecessary

michaldudak:
Instead of using `useRenderElementLazy` which skips just a small part of processing when there's nothing to render, I introduced an `enabled` parameter to `useRenderElement`. It skips most of the logic (including prop merging).

There were some TS acrobatics necessary to ensure `useRenderElement` actually returns a `React.ReactElement` when enabled and `null` when disabled.

### DIFF
--- a/packages/react/src/number-field/scrub-area-cursor/NumberFieldScrubAreaCursor.tsx
+++ b/packages/react/src/number-field/scrub-area-cursor/NumberFieldScrubAreaCursor.tsx
@@ -52,7 +52,7 @@ export const NumberFieldScrubAreaCursor = React.forwardRef(function NumberFieldS
     customStyleHookMapping: styleHookMapping,
   });
 
-  return ReactDOM.createPortal(renderedElement, ownerDocument(element).body);
+  return renderedElement && ReactDOM.createPortal(renderedElement, ownerDocument(element).body);
 });
 
 export namespace NumberFieldScrubAreaCursor {
--- a/packages/react/src/number-field/scrub-area-cursor/NumberFieldScrubAreaCursor.tsx
+++ b/packages/react/src/number-field/scrub-area-cursor/NumberFieldScrubAreaCursor.tsx
@@ -29,13 +29,13 @@ export const NumberFieldScrubAreaCursor = React.forwardRef(function NumberFieldS
   const { isScrubbing, isTouchInput, isPointerLockDenied, scrubAreaCursorRef } =
     useNumberFieldScrubAreaContext();
 
-  const [element, setElement] = React.useState<Element | null>(null);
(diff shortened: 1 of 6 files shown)


packages/react/src/number-field/scrub-area-cursor/NumberFieldScrubAreaCursor.tsx
```diff
@@ -31,7 +31,10 @@ export const NumberFieldScrubAreaCursor = React.forwardRef(function NumberFieldS
 
   const [element, setElement] = React.useState<Element | null>(null);
 
-  const renderElement = useRenderElementLazy('span', componentProps, {
+  const shouldRender = isScrubbing && !isWebKit && !isTouchInput && !isPointerLockDenied;
+
+  const renderedElement = useRenderElement('span', componentProps, {
```
romgrk, line 36: The rest of the codebase uses `element` consistently as the name for the variable, as far as I remember.

romgrk (approved):
Neat work :)

packages/react/src/use-render/useRender.ts
```diff
@@ -64,7 +64,13 @@ export namespace useRender {
      * internal ones.
      */
     props?: Record<string, unknown>;
+    /**
+     * If `false`, the hook will skip most of its internal logic and return `null`.
+     * This is useful for rendering a component conditionally.
+     * @default true
+     */
+    enabled?: boolean;
```
romgrk, line 72: I'm not sure if `shouldRender` wouldn't be better? More clear & semantic, and matches the variable that we're going to assign to this prop most of the time.

michaldudak: I went with `enabled` as it's commonly used in libraries (Floating UI, TanStack, etc.) for this purpose.

flaviendelangle:
Am I the only one for which `useRenderElement` return value is now `never` whenever `enabled` property is not defined?

If you don't pass `enabled`, the generic is set to `never` which causes the entire hook to return `never`.
Since the component is enabled when the property is not passed, I think the default value should be `true` not `never`.

image

michaldudak:
The return value of the hook is defined as `Enabled extends false ? null : React.ReactElement<Record<string, unknown>>`, so it should never be `never`. I'll see what went wrong.

flaviendelangle:
image

A simplified reproduction case

flaviendelangle:
image

Never is a weird beast :laughing: 
So I guess your `boolean | undefined | never` type doesn't work as intended

michaldudak:
https://github.com/mui/base-ui/pull/1968 should fix it.

PR #1974, opened 2025-05-27
[dialog] Refactor implementation

michaldudak:
Replaced `useComponentRenderer` with `useRenderElement`.

### DIFF
--- a/packages/react/src/alert-dialog/close/AlertDialogClose.tsx
+++ b/packages/react/src/alert-dialog/close/AlertDialogClose.tsx
@@ -23,8 +23,7 @@ export const AlertDialogClose = React.forwardRef(function AlertDialogClose(
 
   return useRenderElement('button', componentProps, {
     state,
-    propGetter: getRootProps,
-    props: elementProps,
+    props: [elementProps, getRootProps],
   });
 });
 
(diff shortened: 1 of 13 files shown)


packages/react/src/alert-dialog/trigger/AlertDialogTrigger.tsx
```diff
@@ -29,24 +28,17 @@ export const AlertDialogTrigger = React.forwardRef(function AlertDialogTrigger(
...
-    render: render ?? 'button',
-    className,
+  return useRenderElement('button', componentProps, {
     state,
-    propGetter: (externalProps) => getButtonProps(getTriggerProps(externalProps)),
-    extraProps: other,
+    ref: [forwardedRef, buttonRef, setTriggerElement],
     customStyleHookMapping: triggerOpenStateMapping,
-    ref: mergedRef,
+    props: [elementProps, getTriggerProps, getButtonProps],
```
atomiks, line 40: The `getTriggerProps`/`getFloatingProps` getters should be already-called in the array, before `elementProps`

packages/react/src/alert-dialog/close/AlertDialogClose.tsx
```diff
@@ -12,24 +12,20 @@ import type { BaseUIComponentProps } from '../../utils/types';
...
 
-  const { renderElement } = useComponentRenderer({
-    render: render ?? 'button',
-    className,
+  return useRenderElement('button', componentProps, {
     state,
-    propGetter: getRootProps,
-    extraProps: other,
+    ref: [forwardedRef, ref],
+    props: [elementProps, getRootProps],
```
atomiks, line 27: Remove prop getters

michaldudak: It relies on `useButton` that must return a prop getter.

atomiks: Ah, I see

PR #1981, opened 2025-05-28
[menu] Close submenu on mouseout when parent opened on hover

### LINKED ISSUE #1946: [menu] Close nested menus on mouseout when the root was open on hover

michaldudak:
When a menu is opened on hover, hovering out of the submenu now closes it.
It does not, however, close the whole tree, for two reasons:

- It might be annoying to open a whole tree of menus and inadvertently close the whole thing by accidentally moving your mouse outside
- Menus and submenus can be separated by a gap. When user wants to hover over the parent menu and moves the mouse over the gap, the whole menu would close. This could be avoided similarly to how "safe-polygon" is implemented for opening menus, but it would require quite complex custom logic.

Closes #1946

### DIFF
--- a/packages/react/src/menu/root/MenuRoot.tsx
+++ b/packages/react/src/menu/root/MenuRoot.tsx
@@ -100,6 +100,7 @@ export const MenuRoot: React.FC<MenuRoot.Props> = function MenuRoot(props) {
     } else {
       parent = {
         type: undefined,
+        context: undefined,
       };
     }
   }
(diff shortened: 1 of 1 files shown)


atomiks:
> It does not, however, close the whole tree, for two reasons:
> It might be annoying to open a whole tree of menus and inadvertently close the whole thing by accidentally moving your mouse outside

This makes sense but also feels a bit confusing 🤔  I wonder if the current behavior is actually better

michaldudak:
@colmtuite, as you proposed this change - what do you think? IMO the existing behavior (don't close on hover out) is slightly better, but I could live with either.

colmtuite:
where can I see an example

michaldudak:
https://deploy-preview-1981--base-ui.netlify.app/experiments/menu/menu-fully-featured

Set **Open on hover** in the settings panel

michaldudak:
We decided to leave the old behavior

Issue #1976, opened 2025-05-27
[menu] What is disabling `Menu.Root` supposed to do?


benface:
# Bug report

## Current behavior

When setting a `Menu.Root` to `disabled`, the `Menu.Trigger` becomes disabled, but it can already be `disabled` on its own. If the menu happens to be open while disabled (for instance if forced with the `open` prop), it seems like the intention is to also disable all the `Menu.Item`s it contains because they don't get `data-highlighted` when hovered, but accessibility-wise they are still enabled (no `aria-disabled="true"`, and they can be clicked).

## Expected behavior

Not sure:

- Most likely, the `Menu.Item`s are meant to be disabled when the `Menu.Root` is disabled.
- ...or `Menu.Root`'s `disabled` prop is only meant to prevent the menu from opening (but if that's the case, it seems redundant with `Menu.Trigger`'s `disabled` prop)

## Reproducible example

https://codesandbox.io/p/sandbox/cool-worker-58fxvw

## Base UI version

v1.0.0-alpha.8

## Which browser are you using?

Chrome, Safari, and Firefox

## Which OS are you using?

macOS

## Which assistive tech are you using (if applicable)?

None

benface:
I just noticed that `Menu.Root`'s `disabled` prop was the only way to disable an item that opens a submenu. I feel like it would make sense for `Menu.SubmenuTrigger` to accept a `disabled` prop, but it doesn't.

mj12albert:
I expected `<Menu.Root disabled>` to prevent the menu from being opened externally but it doesn't 🤔 
```jsx
<button onClick={() => setOpen(!open)}>ext trigger</button>
// still opens
<Menu.Root open={open} onOpenChange={setOpen} disabled />
```

aarongarciah:
Potential related bug: it appears that `Menu.Trigger` only receives the `data-disabled` attribute when the `disabled` prop is applied to `Menu.Trigger`, but not when it is applied to `Menu.Root`.

michaldudak:
We should remove the `disabled` props from Trigger and SubmenuTrigger.

> I expected <Menu.Root disabled> to prevent the menu from being opened externally but it doesn't 

Since you're controlling it, we assume you know better. We can't say no to you setting the state explicitly.

benface:
> We should remove the `disabled` props from `Trigger` and `SubmenuTrigger`.

I was hoping you would _add_ it to `SubmenuTrigger`, haha.

[...]

PR #1984, opened 2025-05-28
[menu] Remove `disabled` prop from triggers

michaldudak:
Experimenting with a solution to #1976. Do not merge.


michaldudak:
I'm going to continue supporting the prop on both the Trigger and the Root, as it makes more sense when there are multiple triggers present (#3031). Setting `disabled` on a trigger will disable just this particular trigger, while setting it on the root disables all the triggers.

PR #1985, opened 2025-05-29
[select] Revert `placeholder` API

atomiks:
Continuation of https://github.com/mui/base-ui/pull/1906#issuecomment-2897274740

### DIFF
--- a/docs/reference/generated/select-value.json
+++ b/docs/reference/generated/select-value.json
@@ -2,14 +2,17 @@
   "name": "SelectValue",
   "description": "A text label of the currently selected item.\nRenders a `<span>` element.",
   "props": {
-    "initialLabel": {
+    "initialSelectedLabel": {
       "type": "ReactNode",
-      "required": true,
-      "description": "Specifies the initial value label before hydration and before the popup is opened for the first time.\nIf no item is selected, it acts as a placeholder label. If an item is pre-selected, it should match\nthe selected item’s label.\n\n```tsx\n<Select.Value initialLabel=\"Select an item\" />\n```"
+      "description": "When an item is pre-selected, this specifies the initial value label before hydration and\nbefore the popup is opened for the first time. It should be identical to the selected item’s label.\n\n```tsx\n<Select.Value initialSelectedLabel=\"Red\" />\n```"
+    },
+    "placeholder": {
+      "type": "ReactNode",
+      "description": "Specifies the placeholder label when no item is selected.\n\n```tsx\n<Select.Value placeholder=\"Select an item\" />\n```"
     },
     "children": {
       "type": "ReactNode | ((label: ReactNode, value: any) => ReactNode)",
-      "description": "Specifies a controlled label or a callback to customize the value label when uncontrolled.\n\n```tsx\n<Select.Value initialLabel=\"Select an item\">\n  {(label, value) => value !== null ? `${label} (${value})` : label}\n</Select.Value>\n```"
+      "description": "Specifies a controlled label or a callback to customize the value label when uncontrolled.\n\n```tsx\n<Select.Value initialSelectedLabel=\"Select an item\">\n  {(label, value) => value !== null ? `${label} (${value})` : label}\n</Select.Value>\n```"
     },
     "className": {
       "type": "string | ((state: Select.Value.State) => string)",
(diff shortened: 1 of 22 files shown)


packages/react/src/select/value/SelectValue.tsx
```diff
@@ -14,18 +14,23 @@ export const SelectValue = React.forwardRef(function SelectValue(
...
+    initialLabel,
+    ...elementProps
+  } = componentProps;
 
-  const { value, label, valueRef } = useSelectRootContext();
+  const { value, label: contextLabel, valueRef } = useSelectRootContext();
+
+  const label = contextLabel || initialLabel;
+  const children =
+    typeof childrenProp === 'function' ? childrenProp(label, value) : childrenProp || label;
```
atomiks, line 29: `children` can be a controlled label here:

```jsx
<Select.Value>
  {labelState}
</Select.Value>
```

docs/reference/generated/select-value.json
```diff
@@ -2,10 +2,14 @@
   "name": "SelectValue",
   "description": "A text label of the currently selected item.\nRenders a `<span>` element.",
   "props": {
-    "children": {
-      "type": "ReactNode | ((label: string, value: any) => ReactNode)",
+    "initialLabel": {
+      "type": "ReactNode",
       "required": true,
-      "description": "Specifies the initial value label before choosing an item. A callback can be used to\ncustomize the value label.\n\n```tsx\n<Select.Value>\n  {(label, value) => label ? `${label} (${value})` : 'Select an item'}\n</Select.Value>\n```"
+      "description": "Specifies the initial value label.\nRequired to display the label before mounting the popup and during pre-rendering.\n\n```tsx\n<Select.Value initialLabel=\"Select an item\" />\n```"
```
mj12albert, line 8: It may be hard to understand `initialLabel` fully just by this description without reading this PR description (and all the related discussions). 

Maybe add that `initialLabel` isn't 100% equivalent to "placeholder" but can be used to make a placeholder? (Or even a snippet in ### Examples)

romgrk: Users are missing a lot of context indeed.

> Maybe add that initialLabel isn't 100% equivalent to "placeholder" but can be used to make a placeholder?

I'm not sure that's enough. I think they need to understand the SSR issue for the whole thing to make sense.

michaldudak:
As the prop is called "initial", I would expect it to only cover the SSR/pre-mount cases. Null values can be handled by the children callback. This will cover the case where a value is already selected on the server (so `initialValue` will display the actually selected label), while children called with null would cover the no-value-selected case.

atomiks:
@michaldudak "initial" also sounds like it covers placeholder styles as well. With your change I'd expect it to be called `initialSelectedLabel` maybe?

So, that would be (case 2/3):
```jsx
<Select.Value>
  {(label) => label || 'Placeholder'}
</Select.Value>
```
Or case 1:
```jsx
<Select.Value initialSelectedLabel="Red" />
```

Not sure about forcing the callback to be used for placeholders as it's a little bit less ergonomic. So, maybe we can actually have two props:

Case 1:
```jsx
<Select.Value initialSelectedLabel="Red" />
```

Case 2/3:
```jsx
<Select.Value placeholder="Select item" />
```

Case 2/3 with callback means there's no need for `||` as `label` is always non-empty:
```jsx
<Select.Value placeholder="Select item">
  {(label) => label.toUpperCase()}
</Select.Value>
```

michaldudak:
Then perhaps we shouldn't get rid of the `placeholder` prop. It would apply anytime (server/client) when there's no value. The `initialValue` would then work only on the server/pre-mounting and override the `placeholder` if defined.

packages/react/src/select/value/SelectValue.tsx
```diff
@@ -33,17 +36,11 @@ export const SelectValue = React.forwardRef(function SelectValue(
...
-     *
-     * ```tsx
-     * <Select.Value>
-     *   {(label, value) => label ? `${label} (${value})` : 'Select an item'}
-     * </Select.Value>
-     * ```
+     * A placeholder to display before the popup is opened for the first time.
      */
-    children: React.ReactNode | ((label: string, value: any) => React.ReactNode);
+    placeholder: React.ReactNode;
```
atomiks, line 43: `placeholder` is now required compared to before

atomiks:
Deploy has the cache bug, here's the latest docs:

Screenshot 2025-05-29 at 8 00 46 pm

michaldudak (approved):

PR #1988, opened 2025-05-29
[docs] Highlight new components

michaldudak:
Mark Menubar, Context Menu and Navigation Menu as new.

PR #1989, opened 2025-05-29
[docs] Fix issues in useRender docs

michaldudak:
Fixes https://github.com/mui/base-ui/pull/1988#issuecomment-2918589913


atomiks:
Found a couple more issues: 

```
When building custom components, you often need to control a ref internally while still letting external consumers pass their own—merging refs lets both parties have access to the underlying DOM element. The refs option in useRender enables this, which holds an array of refs to be merged together.
```

> The refs option

And the type displayed in the table shows only an array

michaldudak:
I pressed merge too soon. Sorry, I opened yet another PR: https://github.com/mui/base-ui/pull/1991.

PR #2002, opened 2025-05-30
[core] Import `@floating-ui/react`

romgrk:
Changes:
 - Import `@floating-ui/react` codebase at `./src/floating-ui-react`
 - Add its dependencies here
 - Renamed `./src/utils/floating-ui` to `./src/utils/interactions` (plan to move interaction hooks here in a later PR)
 - Merge duplicate hooks (`useClick`, `useEventCallback`, `useId`, etc)
 - Merge duplicate components (`FocusGuard`)
 - Small perf touchup on `mergeProps` (I really could not resist)


packages/react/src/utils/interactions/useClick.ts
```diff
@@ -1,7 +1,7 @@
 'use client';
 import * as React from 'react';
-import type { ElementProps, FloatingRootContext } from '@floating-ui/react';
-import { isMouseLikePointerType } from '@floating-ui/react/utils';
+import type { ElementProps, FloatingRootContext } from '../../floating-ui-react';
+import { isMouseLikePointerType } from '../../floating-ui-react/utils';
 import { useAnimationFrame } from '../useAnimationFrame';
 
 const EMPTY_OBJECT = {};
```
atomiks, line 7: These `utils/interactions` hooks can go in the floating-ui folder?

romgrk: I plan to move all of those to `utils/interactions` in a later PR.

packages/react/src/floating-ui-react/components/FloatingDelayGroup.tsx
romgrk, line 1: This is `NextFloatingDelayGroup`.

packages/react/package.json
```diff
@@ -86,8 +86,10 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.27.0",
-    "@floating-ui/react": "^0.27.9",
+    "@floating-ui/react-dom": "^2.1.2",
     "@floating-ui/utils": "^0.2.9",
+    "tabbable": "^6.2.0",
+    "use-isomorphic-layout-effect": "^1.2.1",
```
atomiks, line 92: We might as well just inline this as `typeof document !== 'undefined' ? React.useLayoutEffect : noop`

packages/react/src/floating-ui-react/components/FloatingDelayGroup.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+interface UseNextDelayGroupOptions {
+  /**
+   * Whether delay grouping should be enabled.
+   * @default true
+   */
+  enabled?: boolean;
+}
+
+interface UseNextDelayGroupReturn {
```
atomiks, line 92: Type names should be updated

atomiks:
These should be imported  now as well

- [fix(useDismiss): bail out of blur to mark inside React tree if floating tree exists](https://github.com/floating-ui/floating-ui/commit/df494e763383d80917b29d1da8bab4731651b91a)
- https://github.com/floating-ui/floating-ui/pull/3334

atomiks:
I don't think this affects us right now, but it may in the future:

[fix(markOthers): use independent WeakMaps per control attribute (](https://github.com/floating-ui/floating-ui/commit/72f0c781613456ff8b5d12fb247603ff4972a22e)

michaldudak:
There seems to be an issue with Next - building the docs fails with `Error: EMFILE: too many open files`

romgrk:
I have no idea why NextJS suddenly times out while building the docs with this PR. NextJS doesn't output enough logs and doesn't have a `--verbose` option, so it's very hard to understand why it fails when it does. This seem to only happen for "experiments" pages, I could try to remove those. Or I could try to debug NextJS again, though by experience that's a losing battle - NextJS ships all its code minified so it's impossible to step through it. Looking at `htop`, a `jest-worker` process seems to be running non-stop during the `Generating static pages` step of `next build`.

romgrk:
If I start a debugger on the running process (`kill -s SIGUSR1 [jest-worker-pid]` then `chrome://inspect`), it seems to be stopping into a WASM version of oniguruma (the C regex library) every time, so I'm guessing they have a very inefficient regex-based code parser.

image

It seems to be trying to parse the code from the new hooks from floating-ui to output HTML:
image

atomiks:
Oniguruma powers Shiki, the code highlighter used on our docs. Have you tried updating `rehype-pretty-code` and `shiki` to the latest? That could fix it.

romgrk:
To start with, it shouldn't be trying to output those specific files to HTML. I'll put this on hold for now, but that issue might be slowing down the build on master already.

atomiks:
@romgrk I find myself needing to edit the internals of `@floating-ui/react` for Combobox, and would be great if this was merged

I can fork what I need in the meantime but of course this would still be better

romgrk:
I was awaiting the result of your discussions to resume this work. I'm currently working on the new virtualization engine and don't have much bandwidth for the next week(s), but feel free to take over this PR.

atomiks:
We were fine with forking `/react` but not the `/dom` or `/core` libs

I'll see what I can do here

atomiks:
`DemoLoader` was causing an explosion of runaway recursion after the `floating-ui` deps were referenced locally, now it builds

packages/react/src/floating-ui-react/test-components/Button.tsx
michaldudak, line 1: These files should be moved to the `test` directory or excluded from the bundle by other means.

PR #2008, opened 2025-05-30
[code-infra] Update bundle monitor report

Janpot:
* remove redundant checkbox snippet
* Only show tracked bundles in the report
* use vite instead of webpack


LukasTy (approved):
LGTM. 👌

Issue #1982, opened 2025-05-28
[field][checkbox group][radio group] Support multiple `Field.Label`s in a single field


mj12albert:
From https://github.com/mui/base-ui/pull/1948#discussion_r2111203510

We currently use a plain `<label>` in the demos, so it would be good to reduce the need for that when in a `Field`:

```jsx
<Field.Root>
  <CheckboxGroup>
    <Field.Label>
      <Checkbox.Root value="fuji" />
      Fuji
    </Field.Label>
    <Field.Label>
      <Checkbox.Root value="gala" />
      Gala
    </Field.Label>
  </CheckboxGroup>
</Field.Root>
```

Issue #1308, opened 2025-01-08
[docs][checkbox] Update the Hero demo to use Field


mnajdova:
I know we aim at having a simple hero demo, but leaving out the usage of the field for the checkbox may be a bit too much, if we want to teach people that the components should be used together in real-life apps. We got this question as one of the first on Discord (if the checkbox should be used together with the field component).

onehanddev:
Hi @mnajdova ! I’d like to pick this up.
Let me know if there’s any specific direction you’d like me to follow. 😊

mj12albert:
I think in most cases `Field` should be used with `Form`, and we could have a (single) comprehensive demo under the `Form` component's page showing integration with all form controls, then link to there from each component page

mj12albert:
I've added an example in https://github.com/mui/base-ui/pull/2036

Preview: https://deploy-preview-2036--base-ui.netlify.app/react/components/checkbox#form-integration

Issue #2005, opened 2025-05-30
[select] ID prop is not forwarded to hidden input


tomshane:
# Bug report

## Current behavior

When setting ID for CheckBox.Root, Switch.Root and Radio.Root, it is not forwarded to inner hidden input.

## Expected behavior

Select.Root ID should be applied and should override ID of the inner input element.

## Base UI version

1.0.0-beta.0

## Which browser are you using?

Chrome

## Which OS are you using?

Windows

## Additional context

I need to set ID for the component's inner input to be used together with my custom label (attribute hmltFor).

mj12albert:
> I need to set ID for the component's inner input to be used together with my custom label (attribute hmltFor).

@tomshane Could you share a bit more about your use-case? I don't think a label should point to hidden inputs 🤔

tomshane:
Every `<label>` (if it doesn't have nested input) has to have `htmlFor` attribute pointing to `ID` of an associated `<input>`, it doesn't matter if it is hidden or not. Important is it holds a value.

Also, forwarding the ID would be in line with the latest changes in Select API (https://github.com/mui/base-ui/pull/1862).

atomiks:
In the case of the Select, it moves focus to the trigger when the hidden input is focused. (In that case, it needs to be visually hidden, not `type="hidden"`).

mj12albert:
> Every `<label>` (if it doesn't have nested input)

For checkbox, radio, and switch, the label (`<label>` or `Field.Label`) is expected to enclose the whole component which includes the labelable element (the 'button' in most cases) so the id association should be implicit ~

The [HTML spec does explicitly state](https://html.spec.whatwg.org/multipage/forms.html#categories) that hidden inputs are not labelable:

> Some elements, not all of them [form-associated](https://html.spec.whatwg.org/multipage/forms.html#form-associated-element), are categorized as labelable elements. These are elements that can be associated with a [label](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) element.
> [button](https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element), [input](https://html.spec.whatwg.org/multipage/input.html#the-input-element) (if the [type](https://html.spec.whatwg.org/multipage/input.html#attr-input-type) attribute is not in the [Hidden](https://html.spec.whatwg.org/multipage/input.html#hidden-state-(type=hidden)) state)

atomiks:
[...]

PR #2036, opened 2025-06-02
[field][switch][checkbox][radio] Support implicit `Field.Label`

mj12albert:
**💥 Breaking:** If `Field.Label` encloses `Switch`/`Checkbox`/`Radio`, the `htmlFor`/`id` attributes are no longer explicitly set to associate them

Demo: https://deploy-preview-2036--base-ui.netlify.app/experiments/form

This enables multiple `Field.Label`s in a single field for radio/checkbox groups. Implicit association of the label and label-able element is preferred by wrapping `Field.Label` around the component:
```jsx
<Field.Label>
  <Switch.Root />
</Field.Label>

<RadioGroup>
  <Field.Label>
    <Radio.Root />
  </Field.Label>
  <Field.Label>
    <Radio.Root />
  </Field.Label>
</RadioGroup>
```

For radio/checkbox group the implicit association works even when `nativeButton={false}`, so the `for`/`aria-labelledby` attributes don't need to be added back unless even the `label` is replaced with a `span`, in which case it should be on the user to ensure the labelling is correct.

Closes https://github.com/mui/base-ui/issues/1982
Closes https://github.com/mui/base-ui/issues/1308


mj12albert:
https://github.com/mui/base-ui/issues/2005 made me think could all these inputs be a hidden input since its not the labelable element for these components?

If the `button` element is replaced (determined by the `nativeButton` prop in https://github.com/mui/base-ui/pull/1909), we probably need to set the `id`s, `for` and `aria-labelledby` like before though

atomiks:
> 💥 Breaking: htmlFor/id attributes are no longer set to (explicitly) associate Field.Label and Switch/Checkbox/Radio

Is it not possible to still support this?

mj12albert:
Hmm one way to avoid the breaking change is if `Field.Label` also had it's own context, do you think it's worth it? @atomiks

atomiks:
imo it's a bit unexpected that it doesn't work consistently across every part, so perhaps

mj12albert:
Makes sense, let me try it out; it's not a big deal if the attributes are always set (even when supposedly implicit) if it's able to work both ways

mj12albert:
Actually it isn't technically breaking because multiple labels doesn't work to begin with 😅 

But if the label is sibling to the control, it would need an index prop or something like that to pair them together:
```jsx
<Field.Root>
  <CheckboxGroup>
    <Checkbox.Root index={0} />
    <Field.Label index={0} />

    <Checkbox.Root index={1} />
    <Field.Label index={1} />

    <Checkbox.Root index={2} />
    <Field.Label index={2} />
  </CheckboxGroup>
</Field.Root>
```
It's a bit better than having to come up with `id`s, otherwise it's not really possibly to pair them together automagically

atomiks:
I think it should be possible to have 1 `Field.Label` as a sibling of a single `Checkbox`, but otherwise, requiring them to wrap the control inside a Group seems fine (passing `htmlFor` and `id` manually should still work, though)

mj12albert:
OK, so it should support these configs:
```jsx
<Field.Root>
  <CheckboxGroup>
    <Field.Label>
      <Checkbox.Root />
    </Field.Label>

    <Field.Label>
      <Checkbox.Root />
    </Field.Label>
  </CheckboxGroup>
</Field.Root>

<Field.Root>
  <Field.Label>
    <Checkbox.Root />
  </Field.Label>
</Field.Root>

<Field.Root>
  <Checkbox.Root />
  <Field.Label />
</Field.Root>
```

packages/react/src/switch/root/SwitchRoot.tsx
```diff
@@ -69,19 +70,29 @@ export const SwitchRoot = React.forwardRef(function SwitchRoot(
...
   useModernLayoutEffect(() => {
-    setControlId(id);
+    const element = switchRef?.current;
+    if (!element) {
+      return undefined;
+    }
+
+    if (element.closest('label') != null) {
+      setControlId(idProp ?? null);
+    } else {
```
mj12albert, line 88: Used `element.closest('label')` instead of making a Label context to detect enclosing labels

packages/react/src/switch/root/SwitchRoot.tsx
```diff
@@ -69,19 +70,29 @@ export const SwitchRoot = React.forwardRef(function SwitchRoot(
...
+  const inputRef = React.useRef<HTMLInputElement>(null);
+  const handleInputRef = useForkRef(inputRef, externalInputRef, inputValidationRef);
+
+  const switchRef = React.useRef<HTMLButtonElement | null>(null);
+
   const id = useBaseUiId(idProp);
 
   useModernLayoutEffect(() => {
-    setControlId(id);
+    const element = switchRef?.current;
```
atomiks: `?` check shouldn't be necessary here

packages/react/src/switch/root/SwitchRoot.tsx
atomiks: I suppose we can assume they can't render a non-`<label>` element

packages/react/src/radio/root/RadioRoot.tsx
```diff
@@ -104,12 +106,17 @@ export const RadioRoot = React.forwardRef(function RadioRoot(
...
   });
 
+  const id = useBaseUiId();
+
   const inputProps: React.ComponentPropsWithRef<'input'> = React.useMemo(
     () => ({
       type: 'radio',
       ref,
+      // Set `id` to stop Chrome warning about an unassociated input
+      id,
```
atomiks, line 120: How does focus work here? In Select, it has to move focus to the trigger (visible control)

packages/react/src/switch/root/SwitchRoot.tsx
mj12albert: You can, but it can't be implicitly associated and always needs `id` etc, I updated some tests for this

packages/react/src/radio/root/RadioRoot.tsx
mj12albert: Fixed in https://github.com/mui/base-ui/pull/2036/commits/848ff69ae1a7e292dac7bb631f86741e8f942fd3

It matches checkbox and uses the first Radio (`<button>`) as the `controlRef`

atomiks (approved):

atomiks:
Found a regression (from 1.0.0-beta.0) but it's also on master already: clicking the select label doesn't seem to focus the control anymore - doesn't seem like there's any association

Screenshot 2025-06-12 at 9 34 10 pm

PR #2037, opened 2025-06-02
[dialog][alert dialog] Add `forceRender` prop to `Backdrop` part

### LINKED ISSUE #1477: [Dialog] Provide a scape hatch for displaying backdrops on nested dialogs
# Feature request

## Summary

Currently nested dialogs hide their backdrop, which is a GREAT default. However, in certain situations we need to display the backdrop on the nested dialog, e.g. a dialog that is being fired from a Sidebar drawer. 

Something like `forceNestedBackdrop` should be declarative enough to get this working.

## Motivation

I need to fire a nested dialog from within a sidebar drawer (currently built with the Dialog primitive) - the drawer doesn't have its own backdrop, and the nested dialog needs it to be presented correctly. Both dialogs have different shapes so sharing backdrop settings from the parent is not possible.

(https://share.cleanshot.com/5H8GlXdn)

Current Backdrop design forcefully hides the overlay when nested and there is no way to easily opt out of this behavior, so there is no easy way to position an overlay without working around nesting and breaking composition.

### LINKED ISSUE #2279: [dialog] Allow rendering of nested backdrop in Dialog
# Feature request

## Summary

For our use-case, we actually want to render backdrops in nested Dialogs. Currently it's hard-set to be disabled via `enabled: !nested`

We preferably want to toggle if the nested backdrops should be rendered or noet.

atomiks:
Closes #1477
Closes #2279

Adds a new `forceRender` prop to `Backdrop` parts that provides the ability to escape preventing rendering nested backdrops


packages/react/src/alert-dialog/backdrop/AlertDialogBackdrop.tsx
```diff
@@ -67,7 +73,17 @@ export const AlertDialogBackdrop = React.forwardRef(function AlertDialogBackdrop
...
+    /**
+     * How to render the backdrop when dialogs are nested in the React tree.
+     * @default 'root'
+     * @description
+     * - `root`: Render only one backdrop at the root dialog.
+     * - `leaf`: Render only one backdrop at the leaf dialog.
+     * - `always`: Always render the backdrop regardless of nesting.
+     */
+    renderMode?: 'root' | 'leaf' | 'always';
+  }
```
atomiks: Explanation

flaviendelangle:
Not a huge fan of the prop name.
I saw quite a lot of `renderXXX` props that were render props (`renderItem={item => <Item item={item} />}`).
So I wouldn't use the same type of name for a prop like the one created in this PR.

atomiks:
@flaviendelangle yeah I don't like it either but it describes whether it doesn't render or not, so what would it be called?

flaviendelangle:
I don't have a very good proposal :/
`nestedDialogRenderMode` is very verbose

atomiks:
`nestedRenderMode` seems ok though 🤔 @colmtuite

colmtuite:
Why would you want `leaf` instead of `root` or `always`?

atomiks:
You want `leaf` when you don't want stacking effects and always want the topmost dialog to dim everything behind it - the linked issue shows a screenshot: 

#1477



The stacked dialogs in the docs do dim a parent nested dialog already, but use an `::after` overlay instead:

Screenshot 2025-06-04 at 3 55 49 pm

---

It turns out that due to re-parenting, `leaf` doesn't play nicely with transitions. Even if it animated here, the cross-fade may not be perfect with the regular `ease` timing function? So maybe the `::after` pseudo-element is actually the best technique to dim parent dialogs already 🤔 


https://github.com/user-attachments/assets/7b6f04b1-2f3c-4d38-b2b1-efcbb7c6643b

marcojacobsNL:
As you mentioned in my feature request; I agree that a boolean should suffice for the use cases.

Would be nice if this gets merged :)

marcojacobsNL:
This is currently preventing us from using the library for our UI. When do you think it could get a review? @colmtuite @michaldudak

atomiks:
@marcojacobsNL you can use the canary pre-releases before it's released https://github.com/mui/base-ui/pull/2037#issuecomment-2929112103

atomiks:
@michaldudak practically I don't think that works: https://jakearchibald.com/2021/dom-cross-fade/. I also think it's complex enough that it won't be used, as the `::after` technique already used on the docs is significantly simpler to handle

michaldudak (approved):
Yeah, ok, it doesn't look perfect indeed.

PR #2042, opened 2025-06-02
[menu][context menu] Add `SubmenuRoot` part

### LINKED ISSUE #2038: [context menu] `Sub` root part
Currently, it's not possible to disambiguate a submenu and a new root menu that's nested in the React tree while reusing the `ContextMenu` namespace. 

This is more obvious/clear with a `ContextMenu` than a regular `Menu`, because they can easily be nested.

As a result of a missing `.Sub` or `.Subroot` part, the current API requires you use `Menu` to create a submenu, rather than additional parts directly on `ContextMenu`:

```js
<ContextMenu.Root>
  <ContextMenu.Trigger>
    <ContextMenu.Root> {/* this is a new root, not a submenu *}
      ...
    </ContextMenu.Root>
  </ContextMenu.Trigger>
</ContextMenu.Root>
```

While it may only apply to `ContextMenu` in most cases, for consistency, it may be better for all `Menu` components to work the same

atomiks:
Closes #2038


docs/reference/generated/menu-submenu-root.json
```diff
@@ -0,0 +1,72 @@
...
+    "actionsRef": {
+      "type": "RefObject<Actions>",
+      "description": "A ref to imperative actions.\n- `unmount`: When specified, the menu will not be unmounted when closed.\nInstead, the `unmount` function must be called to unmount the menu manually.\nUseful when the menu's animation is controlled by an external library."
+    },
+    "closeParentOnEsc": {
+      "type": "boolean",
+      "default": "true",
+      "description": "When in a submenu, determines whether pressing the Escape key\ncloses the entire menu, or only the current child menu."
+    },
+    "modal": {
```
michaldudak: Modal prop hasn't been supported on submenus anyway, so we can remove it.

docs/reference/generated/menu-submenu-root.json
```diff
@@ -0,0 +1,72 @@
...
+      "description": "Event handler called after any animations complete when the menu is closed."
+    },
+    "disabled": {
+      "type": "boolean",
+      "default": "false",
+      "description": "Whether the component should ignore user interaction."
+    },
+    "openOnHover": {
+      "type": "boolean",
+      "description": "Whether the menu should also open when the trigger is hovered.\n\nDefaults to `true` for nested menus."
```
michaldudak: This can be annotated with `@default true`

michaldudak:
The Anatomy and Nested menu sections in the docs need an update.

PR #2043, opened 2025-06-02
[accordion][collapsible] Allow content to resize naturally

### LINKED ISSUE #1458: [accordion][collapsible] With CSS transitions the panel does not resize when content size changes

atomiks:
Better version of #1704, which automatically handles the panel's dimensions changing from content being inserted/removed. 

Closes #1458 

Collapsible transitions: https://deploy-preview-2043--base-ui.netlify.app/experiments/collapsible/transitions
Accordion transitions: https://deploy-preview-2043--base-ui.netlify.app/experiments/accordion/transitions

- Unsets dimensions once open transition finishes
- Recalculates fixed dimension upon close, which requires deferring `[data-ending-style]` until after the fixed size was set (controlled mode complicates this)
- Moves all dimensions handling to the effect, which simultaneously handles controlled + uncontrolled modes instead of mixing them, so there's a single place to update


mj12albert:
Works great ~ the only issue I found is that when stuff is inserted into the panel (what MutationObserver should handle) the transition may not run: https://codesandbox.io/p/sandbox/late-wind-37gwkv

Though I think we could ignore it for now, the nested one works good and that's a much more realistic use-case

PR #2044, opened 2025-06-02
[menu] Remove disabled prop from Root

michaldudak:
Experimenting with a solution to https://github.com/mui/base-ui/issues/1976. Do not merge.

### DIFF
--- a/docs/src/app/(private)/experiments/menu/menu-fully-featured.tsx
+++ b/docs/src/app/(private)/experiments/menu/menu-fully-featured.tsx
@@ -38,12 +38,12 @@ export default function MenuFullyFeatured() {
   return (
     <div>
       <h1>Fully featured menu</h1>
-      <Menu.Root
-        openOnHover={settings.openOnHover}
-        modal={settings.modal}
-        disabled={settings.disabled}
-      >
-        <Menu.Trigger className={classes.Button} render={triggerRender}>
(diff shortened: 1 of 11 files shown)


atomiks:
Other components also have a similar issue, so it seems we'd need to change them all together. 

But I'm also not convinced yet that it's not worthwhile to still have a `disabled` prop on the root to allow granular disabled states without wholesale disabling the entire trigger. 

It also seems that just because the trigger is disabled, doesn't mean the the root should also be disabled. Tooltips on disabled triggers can be an acceptable use case: https://www.reddit.com/r/UXDesign/comments/1cifb9k/using_tooltips_on_disabled_surfaces_what_is_the/

Though, Tooltip for instance handles the linked issue like this:
https://github.com/mui/base-ui/blob/c54a7829999a175ff4c6b6bb9273c648686f6209/packages/react/src/tooltip/root/useTooltipRoot.ts#L87-L89 

I think this needs a bit more thought

michaldudak:
All right, I'm putting this on hold for now. With https://github.com/mui/base-ui/pull/2336 adapted to Menu, we might need `disabled` on both triggers and root parts.

michaldudak:
I'm going to continue supporting the prop on both the Trigger and the Root, as it makes more sense when there are multiple triggers present (#3031). Setting `disabled` on a trigger will disable just this particular trigger, while setting it on the root disables all the triggers.

PR #2047, opened 2025-06-03
[menu] Unset `role` from Trigger

### LINKED ISSUE #2039: [menu] trigger inside popovers have `role="menuitem"`
# Bug report

## Current behavior

Menu trigger inside popovers have `role="menuitem"`. This appears to be new behaviour after upgrading from the alpha to the beta.

## Expected behavior

Menu trigger inside popovers do not have `role="menuitem"` because the trigger is not a menu item.

(I could be totally wrong!)

## Reproducible example

https://codesandbox.io/p/devbox/exciting-voice-l9nz94?file=%2Fsrc%2FApp.tsx%3A10%2C22&workspaceId=ws_4qKVDKGuRrrvd8oLfoRWSz

## Base UI version

1.0.0-beta.0

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

N/A

atomiks:
Fixes #2039

The problem is the `useRole` hook checks `FloatingTree` by default, not the `MenuRootContext`, and since the Popover now has one, it adds the nested role incorrectly

### DIFF
--- a/packages/react/src/menu/trigger/MenuTrigger.test.tsx
+++ b/packages/react/src/menu/trigger/MenuTrigger.test.tsx
@@ -3,6 +3,7 @@ import { expect } from 'chai';
 import userEvent from '@testing-library/user-event';
 import { act, fireEvent, flushMicrotasks, screen } from '@mui/internal-test-utils';
 import { Menu } from '@base-ui-components/react/menu';
+import { Popover } from '@base-ui-components/react/popover';
 import { describeConformance, createRenderer } from '#test-utils';
 import { PATIENT_CLICK_THRESHOLD } from '../../utils/constants';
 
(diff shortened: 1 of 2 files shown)


packages/react/src/menu/trigger/MenuTrigger.test.tsx
```diff
@@ -328,4 +329,24 @@ describe('<Menu.Trigger />', () => {
       expect(queryByRole('menu', { hidden: false })).to.equal(null);
     });
   });
+
+  it('does not have role prop inside a Popover', async () => {
```
michaldudak, line 333: Could you also test if `<Menu.Trigger render={<span />} />` still has its role set correctly?

michaldudak (approved):

PR #2054, opened 2025-06-03
[code-infra] Migrate to eslint v9 and flat config

brijeshb42:

### DIFF
--- a/docs/scripts/reportBrokenLinks.mts
+++ b/docs/scripts/reportBrokenLinks.mts
@@ -7,11 +7,9 @@ import * as jsxRuntime from 'react/jsx-runtime';
 import { evaluate } from '@mdx-js/mdx';
 import rehypeExtractToc, { type Toc, type TocEntry } from '@stefanprobst/rehype-extract-toc';
 import rehypeExportToc from '@stefanprobst/rehype-extract-toc/mdx';
-/* eslint-disable import/extensions */
 import { rehypeSlug } from 'docs/src/components/QuickNav/rehypeSlug.mjs';
 import { rehypeQuickNav } from 'docs/src/components/QuickNav/rehypeQuickNav.mjs';
 import { rehypeReference } from 'docs/src/components/ReferenceTable/rehypeReference.mjs';
-/* eslint-enable import/extensions */
 import remarkGfm from 'remark-gfm';
 import { rehypeExtractLinkUrls } from './rehypeExtractLinkUrls.mts';
 
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -110,6 +110,16 @@ export default defineConfig(
     extends: createDocsConfig(),
     rules: {
       '@typescript-eslint/no-use-before-define': 'off',
+      'import/extensions': [
+        'error',
+        // Ignores extensions in package imports as well as local ts/tsx imports but .mjs is always required
+        'ignorePackages',
+        {
+          ts: 'never',
+          tsx: 'never',
+          mjs: 'always',
+        },
+      ],
       'no-restricted-imports': [
         'error',
         {
--- a/packages/react/src/floating-ui-react/utils/nodes.test.ts
+++ b/packages/react/src/floating-ui-react/utils/nodes.test.ts
@@ -4,100 +4,98 @@ import { getNodeAncestors, getNodeChildren } from './nodes';
 const contextOpen = { open: true } as FloatingContext;
 const contextClosed = { open: false } as FloatingContext;
 
-describe('floating-ui-react/utils/nodes', () => {
-  test('getNodeChildren returns an array of children, ignoring closed ones when onlyOpenChildren=true', () => {
-    expect(
-      getNodeChildren(
-        [
-          { id: '0', parentId: null, context: contextOpen },
-          { id: '1', parentId: '0', context: contextOpen },
-          { id: '2', parentId: '1', context: contextOpen },
-          { id: '3', parentId: '1', context: contextOpen },
-          { id: '4', parentId: '1', context: contextClosed },
-        ],
-        '0',
-        true,
-      ),
-    ).toEqual([
-      { id: '1', parentId: '0', context: contextOpen },
-      { id: '2', parentId: '1', context: contextOpen },
-      { id: '3', parentId: '1', context: contextOpen },
-    ]);
-  });
+test('getNodeChildren returns an array of children, ignoring closed ones when onlyOpenChildren=true', () => {
+  expect(
+    getNodeChildren(
+      [
+        { id: '0', parentId: null, context: contextOpen },
+        { id: '1', parentId: '0', context: contextOpen },
+        { id: '2', parentId: '1', context: contextOpen },
+        { id: '3', parentId: '1', context: contextOpen },
+        { id: '4', parentId: '1', context: contextClosed },
+      ],
+      '0',
+      true,
+    ),
+  ).toEqual([
+    { id: '1', parentId: '0', context: contextOpen },
+    { id: '2', parentId: '1', context: contextOpen },
+    { id: '3', parentId: '1', context: contextOpen },
+  ]);
+});
 
-  test('getNodeChildren returns an array of children, including closed ones when onlyOpenChildren=false', () => {
-    expect(
-      getNodeChildren(
-        [
-          { id: '0', parentId: null, context: contextOpen },
-          { id: '1', parentId: '0', context: contextOpen },
-          { id: '2', parentId: '1', context: contextOpen },
-          { id: '3', parentId: '1', context: contextOpen },
-          { id: '4', parentId: '1', context: contextClosed },
-        ],
-        '0',
-        false,
-      ),
-    ).toEqual([
-      { id: '1', parentId: '0', context: contextOpen },
-      { id: '2', parentId: '1', context: contextOpen },
-      { id: '3', parentId: '1', context: contextOpen },
-      { id: '4', parentId: '1', context: contextClosed },
-    ]);
-  });
+test('getNodeChildren returns an array of children, including closed ones when onlyOpenChildren=false', () => {
+  expect(
+    getNodeChildren(
+      [
+        { id: '0', parentId: null, context: contextOpen },
+        { id: '1', parentId: '0', context: contextOpen },
+        { id: '2', parentId: '1', context: contextOpen },
+        { id: '3', parentId: '1', context: contextOpen },
+        { id: '4', parentId: '1', context: contextClosed },
+      ],
+      '0',
+      false,
+    ),
+  ).toEqual([
+    { id: '1', parentId: '0', context: contextOpen },
+    { id: '2', parentId: '1', context: contextOpen },
+    { id: '3', parentId: '1', context: contextOpen },
+    { id: '4', parentId: '1', context: contextClosed },
+  ]);
+});
 
-  test('getNodeChildren handles deep parent structures correctly (onlyOpenChildren=true)', () => {
-    const nodes = [
-      { id: '0', parentId: null, context: contextOpen },
-      { id: '1', parentId: '0', context: contextOpen },
-      { id: '2', parentId: '1', context: contextClosed },
-      { id: '3', parentId: '2', context: contextOpen },
-      { id: '4', parentId: '2', context: contextClosed },
-      { id: '5', parentId: '0', context: contextOpen },
-      { id: '6', parentId: '5', context: contextOpen },
-    ];
+test('getNodeChildren handles deep parent structures correctly (onlyOpenChildren=true)', () => {
+  const nodes = [
+    { id: '0', parentId: null, context: contextOpen },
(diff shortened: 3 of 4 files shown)


eslint.config.mjs
```diff
@@ -0,0 +1,186 @@
...
+      'no-restricted-imports': [
+        'error',
+        {
+          paths: NO_RESTRICTED_IMPORTS_PATHS_TOP_LEVEL_PACKAGES,
+          patterns: NO_RESTRICTED_IMPORTS_PATTERNS_DEEPLY_NESTED,
+        },
+      ],
+    },
+  },
+  {
```
Janpot, line 132: One way we could clean this up is to put local config closer to where it should apply.

```tsx
// eslint.config.js
import experiments from 'docs/src/app/(private)/experiments/eslint.config.js';

export default [
  ...experiments
]
```

```tsx
// docs/src/app/(private)/experiments/eslint.config.js
import { getWorkspaceRelative } from '@mui/internal-code-infra';
const ROOT = getWorkspaceRelative(import.meta.dirname);
export default [
  {
    files: [`${ROOT}/**/*{.tsx,.js}`],
    rules: {
      '@typescript-eslint/no-use-before-define': 'off',
      'no-alert': 'off',
      'no-console': 'off',
      'import/no-relative-packages': 'off',
    },
  }
]
```

Perhaps for a follow-up if there is enough interest in this pattern?

eslint.config.mjs
```diff
@@ -0,0 +1,186 @@
...
+    name: 'MUI ESLint config for docs',
+    files: ['docs/**/*'],
+    extends: createDocsConfig(),
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          paths: NO_RESTRICTED_IMPORTS_PATHS_TOP_LEVEL_PACKAGES,
```
Janpot, line 126: @brijeshb42 Not to be part of this PR, but this may be something we could write our own lint rule for. We've been hijacking `no-restricted-imports` for this for a while and it's just such a pain to configure and easy to accidentally overwrite. And we use it in all of our repos. wdyt?

eslint.config.mjs
```diff
@@ -0,0 +1,186 @@
...
+    },
+  },
+  {
+    files: ['test/**/*{.ts,.tsx}'],
+    rules: {
+      'guard-for-in': 'off',
+    },
+  },
+  {
+    files: ['test/**/*{.ts,.tsx,.js}'],
```
Janpot: should we group this with the previous rule?

eslint.config.mjs
```diff
@@ -0,0 +1,186 @@
...
+          patterns: NO_RESTRICTED_IMPORTS_PATTERNS_DEEPLY_NESTED,
+        },
+      ],
+      '@typescript-eslint/no-use-before-define': 'off',
+    },
+  },
+  {
+    name: 'Disable image rule for demos',
+    files: [
+      'docs/src/app/(public)/(content)/**/demos/**/*{.tsx,.js}',
```
Janpot: @brijeshb42 I was thinking, perhaps we should standardize on an extension pattern? I know in core we started using `?(c|m)[jt]s?(x)`, which works quite well. How do you feel about exporting a few presets from the code infra package? e.g.

```tsx
export const EXTENSION_JS = '?(c|m)js?(x)';
export const EXTENSION_TS = '?(c|m)[jt]s?(x)';
export const EXTENSION_TS_ONLY = '?(c|m)ts?(x)';
export const EXTENSION_DTS = `.d.${EXTENSION_TS_ONLY}`;
export const EXTENSION_TEST_FILE = `.test.${EXTENSION_TS}`;
// ...
```

wdyt?

eslint.config.mjs
brijeshb42: I agree. This has been a recurring theme during migration across the repos to the point that its almost code duplication.
I'll create an issue to track this.

eslint.config.mjs
brijeshb42: Added.

eslint.config.mjs
brijeshb42: Makes sense for a followup. I'll also wait for it to be actually required in a repo by a team before doing this.

eslint.config.mjs
michaldudak: Instead of looking for a specific number of nesting levels, it would be more correct to see if a requested identifier is present in the package.json's exports field.

packages/react/src/floating-ui-react/utils/nodes.test.ts
```diff
@@ -4,98 +4,100 @@ import { getNodeAncestors, getNodeChildren } from './nodes';
...
-      '0',
-      true,
-    ),
-  ).toEqual([
-    { id: '1', parentId: '0', context: contextOpen },
-    { id: '2', parentId: '1', context: contextOpen },
-    { id: '3', parentId: '1', context: contextOpen },
-  ]);
-});
+describe('floating-ui-react/utils/nodes', () => {
```
brijeshb42: I am not sure if this code should be added or the rule should be ignored here.

eslint.config.mjs
```diff
@@ -0,0 +1,183 @@
...
+          project: ['tsconfig.json'],
+        },
+      },
+    },
+    /**
+     * Sorted alphanumerically within each group. built-in and each plugin form
+     * their own groups.
+     */
+    rules: {
+      'import/export': 'off', // Mostly handled by Typescript itself. ESLint produces false positives with declaration merging.
```
Janpot: base ui authors in typescript exclusively I believe? but when I try to remove I get an exception

```suggestion
      'import/export': 'off', // FIXME: Maximum call stack exceeded
```

eslint.config.mjs
```diff
@@ -0,0 +1,183 @@
...
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: NO_RESTRICTED_IMPORTS_PATTERNS_DEEPLY_NESTED,
+        },
+      ],
+      '@typescript-eslint/no-redeclare': 'off',
+      // We LOVE non-breaking spaces, and both straight and curly quotes here
+      'no-irregular-whitespace': [1, { skipJSXText: true, skipStrings: true }],
```
Janpot: Let's avoid magic numbers

```suggestion
      'no-irregular-whitespace': ['warn', { skipJSXText: true, skipStrings: true }],
```

There's another one of those, right below

eslint.config.mjs
```diff
@@ -0,0 +1,183 @@
...
+    rules: {
+      'import/export': 'off', // Mostly handled by Typescript itself. ESLint produces false positives with declaration merging.
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: NO_RESTRICTED_IMPORTS_PATTERNS_DEEPLY_NESTED,
+        },
+      ],
+      '@typescript-eslint/no-redeclare': 'off',
+      // We LOVE non-breaking spaces, and both straight and curly quotes here
```
Janpot, line 58: Why not relax this on the company level?

brijeshb42: Noted for next PR in `mui-public`.

eslint.config.mjs
```diff
@@ -0,0 +1,183 @@
...
+      'material-ui/straight-quotes': 'off',
+      // This prevents us from creating components like `<h1 {...props} />`
+      'jsx-a11y/heading-has-content': 'off',
+      'jsx-a11y/anchor-has-content': 'off',
+
+      // This rule doesn't recognise <label> wrapped around custom controls
+      'jsx-a11y/label-has-associated-control': 'off',
+
+      // An overzealous rule that shouts at <a href="#"> in demos.
+      'jsx-a11y/anchor-is-valid': 'off',
```
Janpot: Perhaps we can bring this to the docs preset?

eslint.config.mjs
```diff
@@ -0,0 +1,183 @@
...
+      'react/no-unescaped-entities': [1, { forbid: ['>', '}'] }],
+      'react/prop-types': 'off',
+      'react-hooks/exhaustive-deps': [
+        'error',
+        {
+          additionalHooks: 'useModernLayoutEffect',
+        },
+      ],
+      'react/no-unused-prop-types': 'off',
+      'material-ui/straight-quotes': 'off',
```
Janpot: Same: Why not relax this on the company level?

eslint.config.mjs
```diff
@@ -0,0 +1,183 @@
...
+    },
+  },
+  {
+    files: [
+      // matching the pattern of the test runner
+      `**/*${EXTENSION_TEST_FILE}`,
+    ],
+    extends: createTestConfig(),
+    rules: {
+      'mocha/no-top-level-hooks': 'off', // not needed
```
Janpot: We could parametrize `createTestConfig`, or instead create a `createVitestConfig` and `createMochaConfig `?

eslint.config.mjs
Janpot: I believe this is caused by overly aliasing in tsconfig. It basically aliases the whole package structure regardless of exports, which makes the import rules happily going along. I've been contemplating that we could write a tool that generates the exact aliases in `tsconfig.json` for each package workspace dependencies.

eslint.config.mjs
```diff
@@ -0,0 +1,183 @@
...
+      'react/button-has-type': 'off',
+    },
+  },
+  {
+    files: ['docs/**/*{.ts,.tsx,.js}'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: NO_RESTRICTED_IMPORTS_PATTERNS_DEEPLY_NESTED,
```
Janpot: Isn't this already covered by an earlier rule?

eslint.config.mjs
```diff
@@ -0,0 +1,183 @@
...
+    files: [`packages/**/*.test.${EXTENSION_TS}`],
+    rules: {
+      'testing-library/prefer-screen-queries': 'off', // TODO: enable and fix
+      'testing-library/no-container': 'off', // TODO: enable and fix
+      'testing-library/render-result-naming-convention': 'off', // False positives
+    },
+  },
+  {
+    name: 'MUI ESLint config for docs',
+    files: ['docs/**/*'],
```
Janpot: Shouldn't there be an extension?

Janpot (commented):
Added a bunch of small comments, but nothing blocking. Should we switch to using the file extension constants?

Final review for @mui/base-ui

eslint.config.mjs
brijeshb42: The rationale is that both core and x have this as error. Only base has it off.

eslint.config.mjs
brijeshb42: Adding to both docs and test presets in code-infra.

eslint.config.mjs
brijeshb42: Let's wait for more cases (besides base-ui). I think its fine for now.

eslint.config.mjs
brijeshb42: Merged.

eslint.config.mjs
Janpot: Ok, but there's nothing fundamentally different between these projects that requires us to fork this specific config. If for base ui this config is not necessary, why would it be necessary for the others? We can just relax the rule.

brijeshb42:
@michaldudak Can I get a final approval ?

docs/scripts/reportBrokenLinks.mts
```diff
@@ -7,9 +7,11 @@ import * as jsxRuntime from 'react/jsx-runtime';
 import { evaluate } from '@mdx-js/mdx';
 import rehypeExtractToc, { type Toc, type TocEntry } from '@stefanprobst/rehype-extract-toc';
 import rehypeExportToc from '@stefanprobst/rehype-extract-toc/mdx';
+/* eslint-disable import/extensions */
```
michaldudak: We should be able to explicitly import `.mjs` everywhere. IIRC Typescript doesn't resolve these automatically if they are extensionless.

package.json
```diff
@@ -22,14 +22,14 @@
     "docs:link-check": "pnpm --filter docs link-check",
     "extract-error-codes": "cross-env MUI_EXTRACT_ERROR_CODES=true lerna run --concurrency 1 build:stable",
     "install:codesandbox": "pnpm install --no-frozen-lockfile",
-    "jsonlint": "node ./scripts/jsonlint.mjs",
+    "jsonlint": "code-infra jsonlint",
```
michaldudak, line 25: ❤️

packages/react/src/floating-ui-react/utils/nodes.test.ts
michaldudak: This whole directory (packages/react/src/floating-ui-react) was copied from the Floating UI repo as we decided to maintain a local fork. It may not conform to our internal code standards fully, so we might create a separate config

docs/scripts/reportBrokenLinks.mts
brijeshb42: Addressed

PR #2055, opened 2025-06-03
[core] Remove useComponentRenderer

atomiks:


packages/react/src/alert-dialog/description/AlertDialogDescription.tsx
```diff
@@ -31,9 +29,8 @@ export const AlertDialogDescription = React.forwardRef(function AlertDialogDescr
   }, [id, setDescriptionElementId]);
 
   return useRenderElement('p', componentProps, {
-    state,
     ref: forwardedRef,
-    props: [elementProps, { id }],
+    props: [{ id }, elementProps],
```
michaldudak, line 33: This actually doesn't matter as `elementProps` can't contain `id`.

michaldudak (approved):

packages/react/src/alert-dialog/description/AlertDialogDescription.tsx
romgrk: Sidenote on perf, I've been looking into having uRE automatically handle merging `componentProps` minus `render` & `className` into props, it would avoid destructuring into `elementProps` for some components (those that don't extract more props out of `componentProps`). The only thing stopping me right now is that some components don't have `elementProps` as the final element in the `props` array.

atomiks: @michaldudak yeah, but each component was doing something different - just changed to be the most consistent

PR #2057, opened 2025-06-03
[all components] Do not pass refs to useButton

michaldudak:
Removed the deprecated `buttonRef` parameter of `useButton`. Refs are now merged in `useRenderElement`.


packages/react/src/use-button/useButton.ts
```diff
@@ -12,7 +12,6 @@ const BUTTON_TYPES = new Set(['button', 'submit', 'reset']);
 
 export function useButton(parameters: useButton.Parameters = {}): useButton.ReturnValue {
```
mj12albert, line 9: I think it would be helpful to update the return value buttonRef JSDoc to mention that it's mandatory to pass it to useRenderElement

atomiks (approved):

PR #2058, opened 2025-06-03
[buttons] Extract `useFocusableWhenDisabled` from `useButton`

mj12albert:
Follow up https://github.com/mui/base-ui/pull/1909#discussion_r2122495104

- Removed `'input'` tag support from `useButton`, `ToolbarInput` uses `useFocusableWhenDisabled` standalone; we could generally consider buttons made with input tags e.g. `input type="submit"` obsolete.
- `ToolbarLink` doesn't need to use `useButton`


packages/react/src/utils/useFocusableWhenDisabled.ts
```diff
@@ -0,0 +1,80 @@
+'use client';
+import * as React from 'react';
+
+export interface UseFocusableWhenDisabledParameters {
+  focusableWhenDisabled: boolean | undefined;
+  disabled: boolean;
```
mj12albert: `focusableWhenDisabled` is effectively the `enabled` param for this hook but it would be weird to have both `enabled` and `disabled` params...

packages/react/src/use-button/useButton.ts
```diff
@@ -200,7 +171,7 @@ export namespace useButton {
      * Whether the component is being rendered as a native button or specific native tag.
      * @default true
      */
-    native?: boolean | 'input' | 'a';
+    native?: boolean | 'a';
```
michaldudak: Is `'a'` still used anywhere?

packages/react/src/utils/useFocusableWhenDisabled.ts
michaldudak: Could you document this in this parameter's JSDoc?

packages/react/src/use-button/useButton.ts
mj12albert: Nope ~ I removed the `'a'` param

But found that MenuItem could still indirectly be rendering links through useButton: https://github.com/mui/base-ui/issues/2126#issuecomment-2981473327

packages/react/src/utils/useFocusableWhenDisabled.ts
mj12albert: Updated ~

packages/react/src/utils/useFocusableWhenDisabled.ts
michaldudak, line 1: Let's use the namespace pattern we have in other hooks and components.

mj12albert: Updated!

Issue #1970, opened 2025-05-27
[Slider] capture pointer while dragging


ziloen:
# Feature request

## Summary

In some cases, the slider is only visible when its container is hovered over.

## Examples in other libraries

[Radix UI Slider](https://www.radix-ui.com/themes/docs/components/slider)

mj12albert:
> In some cases, the slider is only visible when its container is hovered over

@ziloen I don't really understand the issue you're describing 😅 is it a bug or a feature request? 

Would help if you could provide a sandbox or a screencap ~

ziloen:
Sure, here's an [example](https://stackblitz.com/edit/vitejs-vite-s9lprtgx?file=src%2FApp.tsx), when dragging the slider outside the container, the `group-hover` state disappears without pointer capture.

mj12albert:
@ziloen You can style the `[data-dragging]` state in addition to the hover state:
```jsx
<Slider.Root
  defaultValue={25}
  className="group-hover:block data-[dragging]:block hidden"
>
```
Demo: https://stackblitz.com/edit/vitejs-vite-47n36f1a?file=src%2FApp.tsx

ziloen:
Thank you, this solves my problem, but I still believe that capturing the pointer would be more intuitive and better align with the behavior of the HTML `<input type="range">`.

mj12albert:
Pointer capture could possibly help with https://github.com/mui/base-ui/issues/1910

Issue #1910, opened 2025-05-13
[slider] When drag crosses an iframe and ends, `[data-dragging]` and `onValueCommitted` get stuck


mj12albert:
Present also in alpha.8: If you move the cursor outside the iframe and let go, it doesn't remove the `[data-dragging]` style

_Originally posted by @atomiks in https://github.com/mui/base-ui/issues/1874#issuecomment-2870403169_

PR #2059, opened 2025-06-03
[slider] Use pointer capture when dragging

mj12albert:
Demo: https://codesandbox.io/p/devbox/condescending-lamarr-8trt73

Fixes `[data-dragging]` and `onValueCommitted` from being stuck when the drag crosses an iframe

Closes https://github.com/mui/base-ui/issues/1970
Closes https://github.com/mui/base-ui/issues/1910

### DIFF
--- a/packages/react/src/slider/control/SliderControl.tsx
+++ b/packages/react/src/slider/control/SliderControl.tsx
@@ -284,7 +284,7 @@ export const SliderControl = React.forwardRef(function SliderControl(
     onValueCommitted(lastChangedValueRef.current ?? finger.value, nativeEvent);
 
     if (
-      nativeEvent instanceof PointerEvent &&
+      'pointerType' in nativeEvent &&
       controlRef.current?.hasPointerCapture(nativeEvent.pointerId)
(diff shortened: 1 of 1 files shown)


mj12albert:
Tests failed but it fixes the issues: https://codesandbox.io/p/devbox/condescending-lamarr-8trt73

packages/react/src/slider/control/SliderControl.tsx
```diff
@@ -283,6 +283,13 @@ export const SliderControl = React.forwardRef(function SliderControl(
     fieldControlValidation.commitValidation(lastChangedValueRef.current ?? finger.value);
     onValueCommitted(lastChangedValueRef.current ?? finger.value, nativeEvent);
 
+    if (
+      nativeEvent instanceof PointerEvent &&
```
atomiks: This possibly has issues in different window contexts. Checking if the type starts with `pointer` could work?

mj12albert: 👍 ~ updated the check

PR #2060, opened 2025-06-03
[NavigationMenuTrigger] Fix positioner height when opening the list using the keyboard

juliomerisio:
**Before**

https://github.com/user-attachments/assets/93f82d76-c980-43c9-9c3b-b15307b615d8

**After**

https://github.com/user-attachments/assets/2abc0f3c-cfd2-4db2-9002-059b0fc0d3d0


atomiks (approved):
The fix works correctly 👍 

I'll let @michaldudak decide on the infra addition - we usually write our tests with Vitest browser mode with RTL if we need to calculate real sizes for in the browser at the moment (using `skipIf(isJSDOM)`), which is why the e2e is largely empty

test/e2e/index.test.ts
```diff
@@ -129,4 +132,48 @@ describe('e2e', () => {
       await expect(screen.getByTestId('three')).toHaveFocus();
     });
   });
+
+  describe('<NavigationMenu />', () => {
```
michaldudak: This test can be added to our unit test suite. We run it in the browser using Vitest's browser mode (using Playwright under the hood).

test/e2e/fixtures/NavigationMenu.tsx
```diff
@@ -0,0 +1,185 @@
...
+      />
+      <path
+        d="M10.3333 3.34539L5.47654 7.71648C4.55842 8.54279 3.36693 9 2.13172 9H0V8H2.13172C3.11989 8 4.07308 7.63423 4.80758 6.97318L9.66437 2.60207C10.0447 2.25979 10.622 2.2598 11.0023 2.60207L15.8591 6.97318C16.5936 7.63423 17.5468 8 18.5349 8H20V9H18.5349C17.2998 9 16.1083 8.54278 15.1901 7.71648L10.3333 3.34539Z"
+        className="dark:fill-gray-300"
+      />
+    </svg>
+  );
+}
+
+const triggerClassName =
```
michaldudak: We don't use Tailwind CSS in our tests. Also, we don't need all the bells and whistles in test code, only the minimum to verify the functionality.

juliomerisio: Good to know! code removed.

test/e2e/index.test.ts
juliomerisio: Nice! I just replaced the Playwright test with the unit test.

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.test.tsx
```diff
@@ -17,4 +19,75 @@ describe('<NavigationMenu.Trigger />', () => {
...
   }));
+
+  it('handles focus and positioner height', async () => {
+    await render(
+      <NavigationMenu.Root>
+        <NavigationMenu.List>
+          <NavigationMenu.Item>
+            <NavigationMenu.Trigger>Overview</NavigationMenu.Trigger>
+            <NavigationMenu.Content>
+              <NavigationMenu.Link href="#">Quick Start Install and</NavigationMenu.Link>
```
michaldudak:
```suggestion
              <NavigationMenu.Link href="#">Quick Start</NavigationMenu.Link>
```

PR #2063, opened 2025-06-03
[core] Refactor: EMPTY_OBJ => EMPTY_OBJECT

romgrk:
We usually avoid abbreviations like "obj" in all the MUI codebases.


mj12albert (approved):
There's a bunch that's just called `EMPTY` 😅 

https://github.com/search?q=repo%3Amui%2Fbase-ui+%22const+EMPTY+%22&type=code

romgrk:
Those are local values so the reader can understand it refers to the local semantic/logical context. It would be fine to use `const EMPTY = EMPTY_OBJECT`. This one requires `_OBJECT` because we'd likely want to also have `EMPTY_ARRAY`.

PR #2067, opened 2025-06-04
[menu] Emit `close` event on `cancel-open`

### LINKED ISSUE #1947: [menu] Menu doesn't close properly on mouseout
Repro:
In Menu, mousedown and enter a submenu without mouseing up, then move the pointer quickly off the right side of the submenu popup, and release the mouse quickly. The Menu should explode.

atomiks:

Fixes #1947 

https://deploy-preview-2067--base-ui.netlify.app/experiments/menu/menu-fully-featured

### DIFF
--- a/packages/react/src/menu/item/useMenuItem.ts
+++ b/packages/react/src/menu/item/useMenuItem.ts
@@ -17,6 +17,7 @@ export function useMenuItem(params: useMenuItem.Parameters): useMenuItem.ReturnV
     allowMouseUpTriggerRef,
     typingRef,
     nativeButton,
+    trigger,
   } = params;
 
   const itemRef = React.useRef<HTMLElement | null>(null);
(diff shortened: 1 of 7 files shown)


atomiks:
An alternative that's possibly more robust is to notify all children that their parent is closed and so they must also close - but it's a little trickier

michaldudak:
There's still an issue when you mouseup on a submenu trigger. Sometimes the menu it opens stays after the root closes:


https://github.com/user-attachments/assets/45f3aecf-c99a-4d14-9724-62694804e212

packages/react/src/menu/item/useMenuItem.ts
```diff
@@ -115,6 +127,10 @@ export namespace useMenuItem {
      * @default false
      */
     nativeButton: boolean;
+    /**
+     * Whether the item is a submenu trigger.
+     */
+    trigger: boolean;
```
michaldudak: `trigger` alone doesn't say much here. How about `isSubmenuTrigger` or `triggersSubmenu`?

michaldudak (approved):
It works fine. I just have a naming suggestion.

PR #2070, opened 2025-06-04
[navigation menu] Handle layout resize while open

### LINKED ISSUE #2035: [navigation menu] Handle layout resize while open
# Bug report

## Current behavior

If the navigation menu is open and the layout of the popup changes due to window resize, the content remains clipped and doesn't resize, as the CSS variables don't recalculate. This is quite rare to see (unlike e.g. Collapsible), but is possible.

Didn't get time to resolve this edge case before release. 

## Expected behavior

The fixed width CSS variables should correctly adapt to resizing.

## Base UI version

1.0.0-beta.0

atomiks:
Fixes #2035

https://deploy-preview-2070--base-ui.netlify.app/experiments/navigation-menu


atomiks:
One of the benefits of `MutationObserver` over `auto` is that it should in theory be able to handle content being inserted or removed mid-transition and also allow the container to transition its height rather than instantly snapping to the new size. However it doesn't handle window resizing by itself

This also applies to Collapsible/Accordion for https://github.com/mui/base-ui/pull/2043

ChrisCates:
Was having issues with resizing.
Then I realized changes were not in the latest release.

Need to update `package.json` to point to:

```json
...
   "@base-ui-components/react": "https://pkg.pr.new/@base-ui-components/react@2070",
```

Just keeping this as a reference in the future for people who might have the same issue.

PR #2071, opened 2025-06-05
[popups] Fix close toggle when rendering non-native button

### LINKED ISSUE #2065: [popover] Does not toggle closed when rendering non-native `<button>` trigger
# Bug report

This likely applies to other components as well.

## Current behavior

`Popover.Trigger` is rendered as a `<div>`:

- Open with <kbd>Enter</kbd>
- <kbd>Shift+Tab</kbd> to tab back to the button
- <kbd>Enter</kbd>
- Doesn't toggle closed

## Expected behavior

Has the same behavior as when rendering a `<button>`.

## Reproducible example

https://codesandbox.io/p/sandbox/6l9c5j

## Base UI version

1.0.0-beta.0

atomiks:
Closes #2065

### DIFF
--- a/packages/react/src/utils/floating-ui/useClick.ts
+++ b/packages/react/src/utils/floating-ui/useClick.ts
@@ -100,13 +100,16 @@ export function useClick(context: FloatingRootContext, props: UseClickProps = {}
           return;
         }
 
-        const openEvent = dataRef.current.openEvent;
+        const openEventType = dataRef.current.openEvent?.type;
 
         const nextOpen = !(
           open &&
           toggle &&
-          (openEvent && stickIfOpen
-            ? ['click', 'mousedown', 'keydown', 'keyup'].includes(openEvent.type)
+          (openEventType && stickIfOpen
+            ? openEventType === 'click' ||
+              openEventType === 'mousedown' ||
+              openEventType === 'keydown' ||
+              openEventType === 'keyup'
             : true)
         );
         onOpenChange(nextOpen, event.nativeEvent, 'click');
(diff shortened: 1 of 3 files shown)


packages/react/src/utils/floating-ui/useClick.ts
```diff
@@ -100,12 +100,13 @@ export function useClick(context: FloatingRootContext, props: UseClickProps = {}
...
+        const openEvent = dataRef.current.openEvent;
+
         const nextOpen = !(
           open &&
           toggle &&
-          (dataRef.current.openEvent && stickIfOpen
-            ? dataRef.current.openEvent.type === 'click' ||
-              dataRef.current.openEvent.type === 'mousedown'
+          (openEvent && stickIfOpen
+            ? ['click', 'mousedown', 'keydown', 'keyup'].includes(openEvent.type)
```
flaviendelangle: I think Romain would prefer to avoid arrays here :laughing: 
To keep the lise of `||` as before, or at least to have a `Set` outside of the function to avoid `O(n)`

atomiks: Yeah I knew 😄  - I just figured since it's a discrete click event that fires infrequently, the perf concerns are not an issue

michaldudak: I don't think there will be any significant difference for just 4 elements. One thing I'd suggest, though is to have a constant array, not define it every time the handler runs.

michaldudak (approved):

PR #2074, opened 2025-06-06
[preview card] Integrate inline positioning

### LINKED ISSUE #2064: [preview card][tooltip] Support inline positioning
# Feature request

## Summary

When the anchor element is an inline element that spans over two lines, the positioner doesn't point to one of the `ClientRect`s.

https://floating-ui.com/docs/inline

## Motivation

Wikipedia's preview cards correctly choose the client rect even when the anchor link spans over two lines.

atomiks:

Closes #2064

Previously it would choose the whole bounding client rect rather than choosing an individual client rect. `align="start"` sort of alleviates this problem but you can't use center positioning or other sides.

https://deploy-preview-2074--base-ui.netlify.app/experiments/inline-positioning

https://github.com/user-attachments/assets/c73bdb71-7625-4c1c-9a9c-f54ccf7e3b53

### DIFF
--- a/docs/src/app/(private)/experiments/inline-positioning.module.css
+++ b/docs/src/app/(private)/experiments/inline-positioning.module.css
@@ -1,0 +1,174 @@
+.Experiment {
+  display: flex;
+  flex-direction: column;
+  gap: 2.5rem;
+  padding: clamp(3rem, 8vw, 6rem) 1.5rem 5rem;
+}
+
+.Section {
+  display: flex;
+  flex-direction: column;
+  gap: 0.75rem;
+  max-width: 32rem;
+}
+
+.SectionLabel {
+  margin: 0;
+  font-size: 0.75rem;
+  font-weight: 700;
+  letter-spacing: 0.14em;
+  text-transform: uppercase;
+  color: var(--color-gray-500);
+}
+
+.Paragraph {
+  margin: 0;
+  max-width: 30rem;
+  font-size: 1rem;
+  line-height: 1.7;
+  color: var(--color-gray-900);
+  text-wrap: pretty;
+}
+
+.Link {
+  outline: 0;
+  color: var(--color-blue);
+  text-decoration-line: underline;
+  text-decoration-thickness: 1px;
+  text-decoration-color: color-mix(in oklab, var(--color-blue), transparent 40%);
+  text-underline-offset: 2px;
+
+  @media (hover: hover) {
+    &:hover {
+      text-decoration-color: var(--color-blue);
+    }
+  }
+
+  &[data-popup-open] {
+    text-decoration-color: var(--color-blue);
+  }
+
+  &:focus-visible {
+    border-radius: 0.125rem;
+    outline: 2px solid var(--color-blue);
+    text-decoration-line: none;
+  }
+}
+
+.PreviewPopup {
+  box-sizing: border-box;
+  border-radius: 0.5rem;
+  background-color: canvas;
+  transform-origin: var(--transform-origin);
+  transition:
+    transform 150ms,
+    opacity 150ms;
+
+  &[data-starting-style],
+  &[data-ending-style] {
+    opacity: 0;
+    transform: scale(0.9);
+  }
+
+  @media (prefers-color-scheme: light) {
+    outline: 1px solid var(--color-gray-200);
+    box-shadow:
+      0 10px 15px -3px var(--color-gray-200),
+      0 4px 6px -4px var(--color-gray-200);
+  }
+
+  @media (prefers-color-scheme: dark) {
+    outline: 1px solid var(--color-gray-300);
+    outline-offset: -1px;
+  }
+}
+
+.PreviewPopup {
+  width: 256px;
+  display: flex;
+  flex-direction: column;
+  gap: 0.75rem;
+  padding: 0.65rem;
+}
+
+.Arrow {
+  display: flex;
+
+  &[data-side='top'] {
+    bottom: -8px;
+    rotate: 180deg;
+  }
+
+  &[data-side='bottom'] {
+    top: -8px;
+    rotate: 0deg;
+  }
+
+  &[data-side='left'] {
+    right: -13px;
+    rotate: 90deg;
+  }
+
+  &[data-side='right'] {
+    left: -13px;
+    rotate: -90deg;
+  }
+}
+
+.ArrowFill {
+  fill: canvas;
+}
+
+.ArrowOuterStroke {
+  @media (prefers-color-scheme: light) {
+    fill: var(--color-gray-200);
+  }
+}
+
+.ArrowInnerStroke {
+  @media (prefers-color-scheme: dark) {
+    fill: var(--color-gray-300);
+  }
+}
+
+.Artwork {
+  min-height: 126px;
+  border-radius: 0.375rem;
+  display: flex;
+  align-items: flex-end;
+  padding: 0.75rem;
+}
+
+.ArtworkLabel {
+  display: inline-flex;
+  padding: 0.2rem 0.45rem;
+  border-radius: 999px;
+  background-color: color-mix(in oklab, white, transparent 25%);
+  font-size: 0.7rem;
+  font-weight: 700;
+  letter-spacing: 0.08em;
+  text-transform: uppercase;
+  color: rgb(24 24 27 / 0.88);
+}
+
+.CardBody {
+  display: flex;
+  flex-direction: column;
+  gap: 0.35rem;
+}
+
+.CardTitle {
+  margin: 0;
+  font-size: 0.95rem;
+  line-height: 1.2;
+  font-weight: 600;
+  color: var(--color-gray-900);
+}
+
+.Summary {
+  margin: 0;
+  font-size: 0.875rem;
+  line-height: 1.35;
+  color: var(--color-gray-700);
+  text-wrap: pretty;
+}
--- a/docs/src/app/(private)/experiments/inline-positioning.module.css
+++ b/docs/src/app/(private)/experiments/inline-positioning.module.css
@@ -56,7 +56,12 @@
 }
 
 .PreviewPopup {
+  width: 256px;
   box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  gap: 0.75rem;
+  padding: 0.65rem;
   border-radius: 0.5rem;
   background-color: canvas;
   transform-origin: var(--transform-origin);
@@ -81,14 +86,6 @@
     outline: 1px solid var(--color-gray-300);
     outline-offset: -1px;
   }
-}
-
-.PreviewPopup {
-  width: 256px;
-  display: flex;
-  flex-direction: column;
-  gap: 0.75rem;
-  padding: 0.65rem;
 }
 
 .Arrow {
--- a/docs/src/app/(private)/experiments/inline-positioning.tsx
+++ b/docs/src/app/(private)/experiments/inline-positioning.tsx
@@ -1,0 +1,169 @@
+'use client';
+import * as React from 'react';
+import { PreviewCard } from '@base-ui/react/preview-card';
+import { type SettingsMetadata, useExperimentSettings } from './_components/SettingsPanel';
+import styles from './inline-positioning.module.css';
+
+interface Settings {
+  use300msDelay: boolean;
+}
+
+const DELAY_MS = 300;
+
+function getDelay(settings: Settings) {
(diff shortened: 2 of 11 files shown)


oleksandr-danylchenko:
Hello 👋🏻 
Do I understand correctly that the `Popover` and the `Tooltip` components will need to go through a similar treatment for positioning with [`inline`](https://floating-ui.com/docs/inline) FloatingUI middleware?

Currently, we are experiencing a similar issue with the `Popover` over multiline text:
a popover is rendered in the middle above multi-line textual content

oleksandr-danylchenko (commented):
Thanks! The extracted `getInlineRectHoverCoords` and `createInlineMiddleware` look pretty good! I like how nicely they fit into an existing infrastructure. I believe that the next step will be adding the `inline` condition prop for the `*Positioner` components. Ideally, the consumers may be able to control which `clientRect` will be picked. Either only the first one, or the hovered one.

atomiks:
@oleksandr-danylchenko what's a reason to not choose the hovered one?

oleksandr-danylchenko:
> @oleksandr-danylchenko what's a reason to not choose the hovered one?

I'd say "design preference". Consumers might wanna show the popover in a "static" place, over the 1st client rectangle. That would correspond to the [`inline` middleware](https://floating-ui.com/docs/inline) behavior in Floating UI. It might make the migration from Floating easier for the existing components.

packages/react/src/tooltip/positioner/TooltipPositioner.tsx
```diff
@@ -67,6 +70,7 @@ export const TooltipPositioner = React.forwardRef(function TooltipPositioner(
     keepMounted,
     collisionAvoidance,
     adaptiveOrigin,
+    inline: createInlineMiddleware(inlineRectCoordsRef),
```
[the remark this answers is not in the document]
atomiks: This is fine

atomiks:
## Codex Review

### Overview
This patch adds line-aware positioning for wrapped inline Tooltip and Preview Card triggers so the popup can follow the relevant text fragment instead of the whole text block. The latest revision also guards stored hover coords by their source element, which closes the stale controlled trigger-switch case.

### Findings
#### 1. ℹ️ [Informational] Runtime cost remains small after the stale-coord fix

**Impact**: The custom inline middleware still adds measurable work to each positioning update, but the post-fix hot path remains in the low-microsecond range and is still cheaper than Floating UI's generic `inline()` on wrapped inline triggers.

**Evidence**: Full branch diff review plus targeted regression runs with `pnpm test:jsdom inlineRect --no-watch`, `pnpm test:chromium TooltipPositioner --no-watch`, `pnpm test:chromium PreviewCardPositioner --no-watch`, and `pnpm typescript`. I also reran a lightweight `@floating-ui/core` micro-benchmark against latest `master` semantics (offset-only baseline) and the current branch head, with 1,200 measured iterations, 200 warmup iterations, and 9 samples per scenario.

| Scenario | Ref | Mean ms/update | p50 | p95 | Samples |
| --- | --- | ---: | ---: | ---: | ---: |
| Single-line trigger | latest `master` baseline | 0.00081 | 0.00069 | 0.00134 | 9 |
| Single-line trigger | PR head custom inline | 0.00088 | 0.00081 | 0.00119 | 9 |
| Wrapped inline trigger | latest `master` baseline | 0.00067 | 0.00070 | 0.00077 | 9 |
| Wrapped inline trigger | PR head custom inline | 0.00125 | 0.00108 | 0.00154 | 9 |
| Wrapped inline trigger | Floating UI `inline()` | 0.00147 | 0.00126 | 0.00189 | 9 |

The deltas were about `+0.00007 ms/update` on single-line triggers and `+0.00059 ms/update` on wrapped inline triggers.

**Recommendation**: No runtime change is necessary based on these numbers.

### Confidence: 4/5
High confidence based on a full branch diff review, targeted JSDOM and Chromium regression coverage for the new controlled-switch path, and a rerun of the middleware micro-benchmark on the current branch tip. Residual risk is mostly around unusual inline fragmentation shapes that still are not exhaustively covered.

### Notes
- Latest changes since the previous review: stored inline hover coords are now tied to their source element, stale coords are ignored when the reference trigger changes, and new controlled-switch regressions were added for both Tooltip and Preview Card.
- The inline-positioning experiment now also exposes a sidebar toggle that switches both `delay` and `closeDelay` between `0` and `300ms`.

packages/react/src/popover/positioner/PopoverPositioner.tsx
oleksandr-danylchenko, line 1: Why was the Popover support removed? 🤔

atomiks: The link-as-a-trigger pattern isn't really accessible in the example for Popover. Because it's not a link, it's a button. 

Technically you can make buttons inline (text spanning over multiple lines) but I think it should be an "infotip" there, instead of styled looking like a link (separate info icon circle/button) because it doesn't go anywhere.

And FWIW, you can always use the `anchor` prop to position it yourself over one of the client rects

oleksandr-danylchenko: Thanks for the detailed breakdown. The reasoning is sensible. Missing navigation after clicking the link might confuse the SR users with an unexpected focus shift. Using the "infotip" accompanied by an explicit `aria-label` on the link might be clearer. 
Therefore, we'd wait for this PR to be released before making the centralized transition.

oleksandr-danylchenko:
Hey, @atomiks 👋🏻 
Was the `Tooltip` inline rendering stripped out, too? Are there still plans to implement it as a part of a separate PR?

atomiks:
I'm just trying to keep it scoped to preview card for now, as it's the most commonly affected component where this inline positioning is needed as the cost of this is not free

michaldudak:
## Review summary

Aggregated findings from a multi-agent review (code, tests, error handling, comments) plus one external finding. Recurring concerns cluster around how `lineIndex` / coords are captured and reused.

### Critical

- **Stale hovered-line coords leak across close/reopen of the same trigger.** `inlineRectCoordsRef` is not cleared when the card closes. Hover line 2 → close → reopen the same trigger via controlled `open` / `triggerId` (no new hover or focus event) → the trigger-identity check at `inlineRect.ts:240-247` accepts the old coords and the popup anchors to the previously hovered line instead of the side-aligned fallback expected for non-hover opens. The existing tests cover stale coords from a *different* element, not stale coords from the *same* element across lifecycles. **Fix:** invalidate `inlineRectCoordsRef.current` on close (carefully, to avoid exit-position jumps) and add a regression: hover wrapped trigger line 2 → close → controlled-open the same `triggerId` → assert side-aligned rect.

- **Stale `lineIndex` is trusted across re-layouts within a single open lifecycle** — `inlineRect.ts:143-145`. The middleware indexes `lines[coords.lineIndex]` without verifying the captured `(x, y)` still lies inside that line. If the trigger reflows after capture (font load, container resize, content insertion above), the popup silently anchors to the wrong line. Re-validate the captured coord against the candidate rect before reusing `lineIndex`, falling back to the x/y re-find.

  These two are related (both stem from coords persisting too long) but neither fix subsumes the other: clear-on-close handles the cross-lifecycle case, re-validate-on-use handles the mid-session re-layout case. Both are needed.

- **Untested middleware safeguards** — three branches that protect against runaway behavior have no direct test:
  - The early-return when computed rects equal `state.rects.reference` (`inlineRect.ts:264-270`) — guards against Floating UI infinite reset loops.
  - The fast-path when `reference.getClientRects` is missing (`inlineRect.ts:233-237`) — needed for custom virtual anchors.
  - The single-line short-circuit (`inlineRect.ts:135-137`) — every non-wrapping trigger goes through this.

### Important

- **Asymmetric coord refresh while open** — `inlineRect.ts:208-222`. `onMouseMove` early-returns when `isOpen`, but `onMouseEnter` always overwrites coords. With `safePolygon`, users routinely re-enter the trigger from the popup, which would reposition it mid-open. Either gate `onMouseEnter` on `!isOpen` for symmetry, or add a comment + test asserting the intended re-anchor.

- **Inverted-comparison branch needs explanation or removal** — `inlineRect.ts:155-157`. The condition `lines[0].left > lines[1].right` reads like a likely typo (after a top-sort, this is "first line starts right of where second line ends" — unusual geometry). It silently returns the union bounding box, masking stale coords. Add a comment naming the geometry it detects, add a test (multi-column / disjoint two-rect fallback), or delete it.

- **Combined error condition hides integration bugs** — `inlineRect.ts:249-251`. `if (!rect || typeof state.platform.getElementRects !== 'function')` collapses an expected fall-through (`!rect`) with a configuration bug (missing platform method). Split them and dev-warn on the second.

- **Synthetic `(0, 0)` mouse events overwrite valid coords** — `inlineRect.ts:200-223`. Programmatic `MouseEvent`s (tests, accessibility tools) often have `clientX/Y === 0`. `getInlineRectCoords` treats these as real coordinates and writes `lineIndex: -1` into the ref, replacing earlier valid data. Reject when the coord falls outside the trigger's bounding rect.

- **Left-side placement untested** — symmetric to the covered right-side branch (`inlineRect.ts:167-189`). A copy-paste error (swapped `<`/`>` or `±Infinity`) wouldn't be caught. Mirror the existing right-placement test.

- **Focus-clears-coords integration test doesn't exercise the regression** — `PreviewCardPositioner.test.tsx:680`. The test tabs in from a clean state, so `coordsRef` is already `undefined` whether or not `onFocus` clears it. To assert the actual behavior, hover line 2 first, then blur, then keyboard-open.

### Suggestions

- **File header comment overclaims** — `inlineRect.ts:5-7`. "Improved left/right edge grouping" isn't substantiated. Either point at the concrete scenario or drop the claim. The actually non-obvious WHY (why a local copy exists rather than wrapping `@floating-ui/react-dom`'s `inline()`) is missing.
- **Magic constants lack WHY comments** — `getLineRects` half-height threshold (`inlineRect.ts:76`) and the `±2px` slop in `findLineIndex` (`inlineRect.ts:99-102`) are exactly the heuristics future maintainers will second-guess.
- **Floating-point equality in edge grouping** — `inlineRect.ts:184`. `nextEdge === edge` will miss subpixel-different rects on justified text. Matches Floating UI's own behavior, but `Math.abs(nextEdge - edge) < 0.5` would be more robust.
- **`findLineIndex` returns the first match** — `inlineRect.ts:96-104`. With merged lines containing mixed font sizes (e.g. an `<a>` containing a `<sup>`), two lines can both contain the point. Prefer the line whose center is closest to `y`.
- **Redundant comments in test file** — `PreviewCardPositioner.test.tsx:120-129`. Three comments restate what the assertions already say. The single intent comment at line 102 is enough.
- **Loose X-tolerance windows** — integration assertions use ~20px-wide acceptance windows for X-coords; combined with small popup widths, off-by-one-rect regressions could slip through.
- **Trigger-switch test name is misleading** — `PreviewCardPositioner.test.tsx:467`. The test actually asserts "stale coords for the previous trigger don't bleed over." Rename or comment so a future contributor doesn't read it as a no-op.

### Strengths

- Solid coverage of the happy-path scenarios: hovered-line positioning, side-aligned fallback, virtual-reference `contextElement`, scroll handling, and the `keepMounted` + custom anchor + collision-boundary regression net.
- Trigger-identity check correctly handles the multi-trigger Preview Card model (within a single lifecycle).
- Project-conventions clean: no `setTimeout`/`rAF`/`document`/`window` global access, owner-doc-aware where it matters, browser-only tests gated with `describe.skipIf(isJSDOM)`, no Chai/Sinon leakage, no `flushMicrotasks` after `await render`.
- Public-API discipline: `inline` is internal-only on `useAnchorPositioning`, types exported via the popups barrel, store integration matches the `closeDelayRef` pattern.
- Bundle-cost commit shows deliberate care about size impact when reimplementing Floating UI's middleware locally.
- `InlineRectCoords` JSDoc is genuinely useful — `lineIndex` documentation explains a non-obvious capture detail.

### Recommended action

1. Address both stale-coords paths: clear `inlineRectCoordsRef` on close (cross-lifecycle) **and** re-validate `lineIndex` against captured `(x, y)` at middleware time (mid-session re-layout). Add a regression test for each.
2. Add tests for the three untested middleware safeguards.
3. Resolve the `onMouseEnter` / `onMouseMove` asymmetry one way or the other and clarify or delete the inverted-comparison branch.
4. Split the combined `!rect || !getElementRects` early-out and reject `(0, 0)` synthetic coords.
5. Add the missing left-side placement test and rework the focus-coords-clearing integration test so it actually proves what it claims.
6. Comment / suggestion cleanups can be folded in at the same time.

michaldudak (approved):

PR #2079, opened 2025-06-07
[select][menu] Add `highlighted` to item `State`

### LINKED ISSUE #2076: [menu] add `highlighted` into `MenuItem` state
# Feature request

## Summary

I want to know if the current `MenuItem` is highlighted in its `render` prop.

## Motivation

I need to display some explanatory information why the element is disabled. When using a mouse I can handle the hover, but I would also like to show this information when navigating with a keyboard. So, it would be nice to have a `highlighted` boolean next to `disabled` in second argument of `render` prop.

atomiks:
This seems a lot slower than before, assuming it's implemented correctly. 

Instead of a state like `highlighted: boolean`, it could be an event handler like `<Select.Item onHighlightChange>`

Closes #2076

---

20x slowdown, hover items quickly

Before: https://master--base-ui.netlify.app/experiments/select-perf
This PR: https://deploy-preview-2079--base-ui.netlify.app/experiments/select-perf


packages/react/src/select/item/SelectItem.tsx
romgrk, line 1: Screenshot From 2025-06-07 00-42-33

Screenshot From 2025-06-07 00-42-44

---

`SelectItem` is chunky, re-rendering it will be expensive there's no getting around that.

You're interleaving DOM reads & writes, it's destroying performance. DOM reflows are a common issue in base-ui, you need to be more careful about those.

An example of both of those issues, in `onMouseMove` you're reading the popup's `.offsetHeight` on every event. Is that DOM read really necessary? Note that with this PR the read/write operations order has been reversed (master is "set data-highlighted, then read offsetHeigh", this PR is "read offsetHeight, then set data-highlighted"), that could explain some issues.  
You're also storing that value in one `useRef` per item (each item has its own `prevPopupHeightRef`). A `useRef()` call by itself adds 2-4ms on slow devices, if you have one ref per item (for the same value), you're adding overhead that could be avoided. React hooks are not free, I've annotated below a zone where react hooks add overhead.

Other than that, I see that floating-ui is doing some work, not sure what.

Last issue, there's a `store.update()` on `mouseout` and one on `mousemove`, that's two in quick succesion. Store updates are also not free with large lists, that's the one [downside of selector-based reactivity](https://www.notion.so/mui-org/Fine-grained-reactivity-1fdcbfe7b6608034b2d3da1281c637e7?source=copy_link#1fdcbfe7b66080678321d8724214e4b0). You should try to combine them in one. Avoid the store update if the `mouseout` event has a [`relatedTarget`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget) that's another list item (which will trigger a `mousemove`).

image

romgrk: I can take a look at optimizing this component, I could commit on this PR if you don't mind.

atomiks: Sure, thanks for the info here! Very helpful

atomiks:
One thing to look out for is the highlights getting stuck. On the latest commit, this can happen:

https://github.com/user-attachments/assets/20bfbbf2-ca9f-4573-8305-b405fb6b19b1

packages/react/src/select/item/SelectItem.tsx
romgrk: I've added a commit, feel free to discard or rework it as you like. The `hover` latency is down to 100ms, compared to master's 65ms and this PR's previous 255ms. Still a downgrade, but more acceptable.

Here is a graph of the event (with proper debug symbols). It's down to a single store update.

The largest part of the remaining overhead is caused by floating-ui:
 - The `Select.Root` needs to re-render because FUI requires the `activeIndex` in the `useListNavigation` and `useTypeahead` hooks, nothing to do there until we can refactor FUI more agressively. Ideally FUI would use a reactivity model that doesn't rely on re-renders.
 - FUI's focus handling is causing re-renders and delays, does it really need to be async? Anyway, not much to do there either, can't refactor that easily. Ideally we'd manage focus with layout effects so that the `.focus()` call happens right after React updates the DOM, we want grouped DOM operations.

Screenshot From 2025-06-07 02-00-00

atomiks: The asynchronicity should only be on the first render (waiting for the positioning to be ready), which itself would be solved if the positioning ran sync for the DOM platform in the next major version 😅 

What part of Floating UI is running for 27ms? The select itself tries to bypass Floating UI's `mousemove` item handlers (only Menu doesn't)

romgrk: It starts with `enqueueFocus()`, it's the async focus handling code.

atomiks: Right yeah `forceSyncFocusRef` internally in `useListNavigation` doesn't get assigned to `true` since the item handlers aren't used. Moving the assignment lower after the early returns seems to fix it.

atomiks:
@romgrk  [fix(Select.Item): different strategy to avoid double update](https://github.com/mui/base-ui/pull/2079/commits/9e41599d83be2b2cc097334a448cde8f53de2428)

Pretty hard  to reproduce but it's possible for the highlight to get stuck when moving quickly when it's clipped by the popup.

Edit: can't reproduce this w latest commit on dev mode

https://github.com/user-attachments/assets/0a3c6510-7396-4a5a-8ee3-9531643ded88

PR #2082, opened 2025-06-09
[select] POC: items array API

michaldudak:
https://deploy-preview-2082--base-ui.netlify.app/experiments/select/select-array-api


docs/src/app/(private)/experiments/select/select-array-api.tsx
```diff
@@ -0,0 +1,88 @@
...
+import styles from './select-perf.module.css';
+
+const items = Array.from({ length: 100 }, (_, i) => ({
+  label: `Item ${i + 1}`,
+  value: `item-${i + 1}`,
+}));
+
+export default function Experiment() {
+  return (
+    <Select.Root defaultValue="item-42" items={items}>
```
atomiks, line 25: `item-42` isn't being selected initially (when opening the popup there's no selection, then when closing, the value becomes blank)

atomiks:
How do you create groups?

michaldudak:
> How do you create groups?

You can't yet, but I'm planning to make the Select.Root accept objects in shape of:
```tsx
{
  group: string;
  disabled?: boolean;
  items: SelectOption[];
}
```

michaldudak:
Added support for groups. Also moved the templates under Popup, closer to where they're actually rendered.

PR #2084, opened 2025-06-09
[select] Fix programmatic value changes and autofill handling

atomiks:
From #2083. Since the items aren't rendered on mount anymore, they need to be force mounted when either:

1. `onChange` on the hidden input (for autofill handling) is invoked
2. The value changes externally (controlled mode) before the popup is opened to register the values/labels

### DIFF
--- a/packages/react/src/select/portal/SelectPortal.tsx
+++ b/packages/react/src/select/portal/SelectPortal.tsx
@@ -17,9 +17,9 @@ export function SelectPortal(props: SelectPortal.Props) {
 
   const { store } = useSelectRootContext();
   const mounted = useSelector(store, selectors.mounted);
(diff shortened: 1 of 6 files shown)


michaldudak:
Will force-mounting be necessary after we implement https://github.com/mui/base-ui/pull/2087?

atomiks:
Yes as the prop is optional

michaldudak:
Still, isn't it a similar case when there is `defaultValue` and user is responsible for figuring out the right label and controlled `value` that changes before the Select is open for the first time? In both cases they'll have to either pass in `items` or use the callback in Select.Value.

atomiks:
This bug is related to the actual values being registered internally (the selected state doesn't work for these cases at the moment). It also only applies to the controlled `value` prop not `defaultValue` since it has to change externally by the user

PR #2086, opened 2025-06-09
[internal] Restore Preact support

### LINKED ISSUE #1995: Does it work with Preact?
Preact is gaining popularity with the recent announcements from Remix, Fresh and apparently multiple other frameworks, I haven't found anything in the docs regarding compatibility with Preact, it might be worth mentioning if these two both work together well

atomiks:
Closes #1995

### DIFF
--- a/packages/react/src/utils/useEventCallback.ts
+++ b/packages/react/src/utils/useEventCallback.ts
@@ -6,11 +6,13 @@ import { useLazyRef } from './useLazyRef';
 const useInsertionEffect = (React as any)[
   `useInsertionEffect${Math.random().toFixed(1)}`.slice(0, -3)
 ];
-// Support Preact, which replaces useInsertionEffect with useLayoutEffect and fires too late.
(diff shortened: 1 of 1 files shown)


packages/react/src/utils/useEventCallback.ts
```diff
@@ -38,5 +41,7 @@ function createStableCallback() {
 }
 
 function assertNotCalled() {
-  throw new Error('Base UI: Cannot call an event handler while rendering.');
+  if (process.env.NODE_ENV !== 'production') {
+    throw new Error('Base UI: Cannot call an event handler while rendering.');
+  }
```
atomiks, line 50: There's no need for the throw to occur in prod as it will be caught by tests & when developing

atomiks: Also, the fact that this throws if assigned in `useLayoutEffect` makes it incorrect 😅 (but it's good that it throws when the phase is wrong, anyway)

Issue #2011, opened 2025-05-31
[select] Refine `Select.Value` `placeholder` API


atomiks:
# Feature request

## Summary

Currently `<Select.Value placeholder="" />` represents the "initial" label before an item has been chosen, even if one was initially chosen on mount. It  therefore conflates text shown when `value == null`, and the value during SSR pre-rendering (and now, in `beta.0`, before the popup is opened to grab the labels based on the default value).

## Motivation

`placeholder` should only represent `value == null`, while the initial label when `defaultValue` or `value` is pre-selected (during SSR/client pre-mount) must be distinguished to prevent confusion. 

In `alpha.8`, it only affected SSR pre-rendering (it was blank before hydration), but now with the perf improvements to Select, it also affects client rendering too (so it's always blank). As a result we made `placeholder` required instead of optional. Either way it must be specified manually.

atomiks:
The `placeholder` prop conflates an "initial" label (due to SSR and pre-mount limitations) with text shown when `value == null`

## Initial label (default value)

### Case 1: Default value, never a placeholder

```jsx
<Select.Root defaultValue="red">
  <Select.Value placeholder="Red" /> {/* SSR/pre-mount label */}
  <Select.Portal>
    <Select.Positioner>
      <Select.Popup>
        <Select.Item value="red">
          Red
        </Select.Item>
      </Select.Popup>
    </Select.Positioner>
  </Select.Portal>
</Select.Root>
```

### Case 2: Default value, can change to an empty value

```jsx
<Select.Root defaultValue="red">
  <Select.Value placeholder="Red" /> {/* SSR/pre-mount label */}
  <Select.Portal>
    <Select.Positioner>
      <Select.Popup>
        <Select.Item>
          Select value {/* Acts as the placeholder */}
        </Select.Item>
        <Select.Item value="red">
          Red
        </Select.Item>
      </Select.Popup>
    </Select.Positioner>
  </Select.Portal>
</Select.Root>
```

## Placeholder label (empty value)

### Case 3: No default value, cannot revert back after selection

```jsx
<Select.Root>
  <Select.Value placeholder="Select value" /> {/* SSR/pre-mount label, can't revert back */}
  <Select.Portal>
    <Select.Positioner>
      <Select.Popup>
        <Select.Item value="red">
          Red
        </Select.Item>
      </Select.Popup>
    </Select.Positioner>
  </Select.Portal>
</Select.Root>
```

### Case 4: No default value, but can revert back after selection

[...]

PR #2087, opened 2025-06-09
[select] Print raw value in `Select.Value`

atomiks:
Closes https://github.com/mui/base-ui/issues/2011


docs/src/app/(public)/(content)/react/components/select/demos/label/css-modules/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  mono: <span style={{ fontFamily: 'monospace' }}>Monospace</span>,
+  cursive: <span style={{ fontFamily: 'cursive' }}>Cursive</span>,
+};
+
+function renderValue(value: string | null) {
+  if (value == null) {
+    return 'Select a font';
+  }
+  return valueToLabel[value];
+}
```
mj12albert: I think it would be a bit easier to read if this was all inlined in the component:
```jsx
<Select.Value>
  {(value: string | null) => {
    // ...
  }}
</Select.Value>
```
otherwise to me it reads upside down and stands out more than it needs to 😅 

Maybe we could even consider showing this in the hero demo instead of having a separate demo, I feel like in most real use-cases the raw value will be something that's not meant for display at all like a UUID

mj12albert: On the other hand, for many use-cases the children fn is usually even simpler than this example, like a country select:
```jsx
<Select.Value>
  {(value: string | null) => {
    if (value == null) {
      return 'Select a country';
    }
    const country = countries.find((c) => c.iso2 === value);
    return country?.name;
  }}
</Select.Value>
```

atomiks: The `.find()` case is likely not good to encourage by default due to perf concerns, but yeah it can be relatively simple

michaldudak:
Haven't we decided to drop `placeholder` altogether after all?

I think the current solution is a pretty good compromise. It's worth having a section in the docs about it, though, as it's not immediately clear how it works.

atomiks:
> Haven't we decided to drop placeholder altogether after all?
> 
> I think the current solution is a pretty good compromise. It's worth having a section in the docs about it, though, as it's not immediately clear how it works.

I've removed it; I think the only time it's useful is when you default to `null` but aren't also rendering a `null` item in the list (so there's no label to grab). It's easy enough to workaround and avoids the confusion when using both `placeholder` but also using `children` prop

michaldudak (approved):

PR #2089, opened 2025-06-10
[select] Fix re-render warning in `Positioner` on fallback

atomiks:
The set-state-in-render pattern only works if the state is local to the component, otherwise it needs to use an effect now that it's in the store.

To reproduce the warning before, just open the select close to the edge of the viewport so it enters fallback mode


mj12albert:
By fallback mode you mean item-not-aligned-to-trigger like this right? I couldn't reproduce a warning 🤔 @atomiks 

Screenshot 2025-06-12 at 9 37 52 AM

atomiks:
@mj12albert yeah that's right, but I think it only happens in dev mode

romgrk (approved):

Issue #2085, opened 2025-06-09
[tabs] Support scrolling TabsList


ciampo:
# Bug report

## Current behavior

When the `Tabs.List` component is set to `overflow: scroll` in order to show all children `Tabs.Tab`, the `Tabs` component stops functioning as expected:

- if the selected tab is hidden, the user won't have a visual clue that it is selected
- the `--active-tab-*` CSS variables used to position the `Tabs.Indicator` have incorrect values that cause the indicator to go out of sync

https://github.com/user-attachments/assets/e7579717-dadc-4795-b96c-4e10a99e3a3c

## Expected behavior

- Whenever a new tab is selected, the tablist should scroll it into view
- The `--active-tab-*` CSS variables should be correct even when the tablist is scrollable

For reference, this is a screen recording of the current implementation that we have in the `@wordpress/components` package:

https://github.com/user-attachments/assets/ac3904c0-343b-429f-b243-78a8a6f8bb4a

## Reproducible example

https://codesandbox.io/p/sandbox/xenodochial-cray-68ycy4

## Base UI version

`1.0.0-beta.0`

## Which browser are you using?

Chrome

## Which OS are you using?

Mac OS

## Which assistive tech are you using (if applicable)?

N/A

## Additional context

Currently, in Gutenberg, our `Tabs` component is build on top of `Ariakit`.

While `Ariakit` doesn't provide support for this feature out of the box (not even the CSS variables), its APIs allow us to grab a reference to the selected tab's DOM element, which we feed to some custom code enable support for scrolling tablist:

- Code: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/tabs/tablist.tsx
- Example: https://wordpress.github.io/gutenberg/?path=/story/components-tabs--size-and-overflow-playground

While we could probably manage to implement something similar with `BaseUI`, the absence of APIs to easily get `Tabs` internal state and the selected Tab's DOM element makes it more convolute.

Also, since Base UI exposes the `--active-tab-*`  CSS variables, it feels like this scenario should be also handled internally by Base UI.

mj12albert:
```diff
diff --git a/packages/react/src/tabs/indicator/TabsIndicator.tsx b/packages/react/src/tabs/indicator/TabsIndicator.tsx
index 457e120d18..df49265829 100644
--- a/packages/react/src/tabs/indicator/TabsIndicator.tsx
+++ b/packages/react/src/tabs/indicator/TabsIndicator.tsx
[...]

PR #2093, opened 2025-06-10
[tabs] Fix indicator positioning when TabsList overflows

mj12albert:
Demo: https://codesandbox.io/p/sandbox/dreamy-leftpad-hxv4w8

- `CompositeRoot`s will scroll the active item into view if it has overflowed. It needs to know the text direction to do so, so now it always call `useDirection` internally and the `direction` prop is removed
- `Tab.Indicator`'s position vars needed to be calculated differently to account for potential scrollbars

Fixes https://github.com/mui/base-ui/issues/2085

### DIFF
--- a/docs/reference/generated/field-control.json
+++ b/docs/reference/generated/field-control.json
@@ -6,7 +6,7 @@
       "type": "string | number | string[]"
     },
     "onValueChange": {
-      "type": "((value: string | number | string[] | undefined, event: Event) => void)",
+      "type": "((value: string, event: Event) => void)",
       "description": "Callback fired when the `value` changes. Use when controlled."
     },
     "className": {
--- a/docs/reference/generated/input.json
+++ b/docs/reference/generated/input.json
@@ -6,7 +6,7 @@
       "type": "string | number | string[]"
     },
     "onValueChange": {
-      "type": "((value: string | number | string[] | undefined, event: Event) => void)",
+      "type": "((value: string, event: Event) => void)",
       "description": "Callback fired when the `value` changes. Use when controlled."
     },
     "className": {
--- a/docs/reference/generated/radio-root.json
+++ b/docs/reference/generated/radio-root.json
@@ -7,6 +7,11 @@
       "required": true,
       "description": "The unique identifying value of the radio in a group."
     },
+    "nativeButton": {
+      "type": "boolean",
+      "default": "true",
+      "description": "Whether the component renders a native `<button>` element when replacing it\nvia the `render` prop.\nSet to `false` if the rendered element is not a button (e.g. `<div>`)."
+    },
     "disabled": {
       "type": "boolean",
       "default": "false",
--- a/docs/reference/generated/switch-root.json
+++ b/docs/reference/generated/switch-root.json
@@ -19,6 +19,11 @@
       "type": "((checked: boolean, event: Event) => void)",
       "description": "Event handler called when the switch is activated or deactivated."
     },
+    "nativeButton": {
+      "type": "boolean",
+      "default": "true",
+      "description": "Whether the component renders a native `<button>` element when replacing it\nvia the `render` prop.\nSet to `false` if the rendered element is not a button (e.g. `<div>`)."
+    },
     "disabled": {
       "type": "boolean",
       "default": "false",
--- a/docs/src/app/(private)/experiments/form.module.css
+++ b/docs/src/app/(private)/experiments/form.module.css
@@ -1,9 +1,9 @@
 .Form {
   display: flex;
   flex-direction: column;
-  gap: 1rem;
+  gap: 24px;
   width: 100%;
-  max-width: 16rem;
+  max-width: 20rem;
 }
 
 .Field {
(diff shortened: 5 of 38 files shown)


docs/src/app/(private)/experiments/tabs.module.css
```diff
@@ -69,7 +71,10 @@
     &::before {
       content: '';
       position: absolute;
-      inset: 0.25rem 0;
+      top: 0;
+      left: 0;
+      right: 0;
+      bottom: 0;
```
mj12albert: This fixes a misalignment in vertical orientation:

Screenshot 2025-06-10 at 2 25 50 PM

docs/src/app/(public)/(content)/react/components/tabs/demos/hero/css-modules/index.module.css
```diff
@@ -27,6 +27,8 @@
   line-height: 1.25rem;
   font-weight: 500;
   user-select: none;
+  white-space: nowrap;
+  word-break: keep-all;
```
mj12albert, line 31: I think it's better to have these in the styles so when people copy or fork the demo and change the text to include a space, the layout won't immediately break

mj12albert:
`scrollIntoView()` will scroll all scrollable ancestors: https://github.com/w3c/csswg-drafts/issues/9452

In the sandbox, the window scrolls vertically downward when the lower TabsList/Tab calls scrollIntoView

There's a proposed `container` option but nothing that can be used today, so I think we need to do it manually 😪

packages/react/src/tabs/list/TabsList.tsx
```diff
@@ -52,6 +52,26 @@ export const TabsList = React.forwardRef(function TabsList(
...
+      (tabsList.clientWidth < tabsList.scrollWidth &&
+        ((direction === 'ltr' &&
+          activeTab.offsetLeft + activeTab.offsetWidth > tabsList.clientWidth) ||
+          (direction === 'rtl' && activeTab.offsetLeft < tabsList.scrollLeft))) ||
+      (tabsList.clientHeight < tabsList.scrollHeight &&
+        activeTab.offsetTop + activeTab.offsetHeight > tabsList.clientHeight)
+    ) {
+      activeTab.scrollIntoView({ behavior: 'auto', block: 'nearest', inline: 'nearest' });
+    }
+  }, [activeTab, direction]);
```
atomiks: Let's put this in Composite instead, as it should handle everything else (Toolbar, etc.) And use a little polyfill for `.scrollIntoView()` to handle the ancestor issue

docs/src/app/(private)/experiments/tabs.module.css
atomiks: `inset: 0` should be the same as this style

packages/react/src/tabs/list/TabsList.tsx
ciampo: We also do it manually https://github.com/Automattic/wp-calypso/blob/9bea85d93dc040ac8683aa30456964ffb3bbd140/packages/components/src/tabs/tablist.tsx#L16-L56

ciampo:
@mj12albert Thank you for working on this!

I'd like to suggest another couple of related improvements:

- update indicator positioning when the selected tab or the tablist changes size, position (or any other forced update)
  - This is how we're doing it in `@wordpress/components`: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/tabs/tablist.tsx#L90-L94
- add a way for the consumers to know whether the tablist has got any "scrollable" space to either sides — this may be useful for consumers of the component to display visual hints to the user, enable/disable scrolling buttons, etc etc
  - This is how we're doing it in `@wordpress/components`: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/tabs/tablist.tsx#L96-L100

Also happy to open separate issues for these list items.

mj12albert:
> @mj12albert Thank you for working on this!
> 
> I'd like to suggest another couple of related improvements:
> 

Hey thanks ~ please go ahead and open other issues and we'll check it out ~ @ciampo

atomiks:
Using the latest CodeSandbox CI commit here with the demo:

https://github.com/user-attachments/assets/a13c38b3-6b29-4d0f-8b09-8520073925a3

I'd expect it to scroll into view if the item overflows a bit, like Menu/Select do with the native `.scrollIntoView()`.

atomiks:
- `prehydrationScript.template.js` should likely be updated
- The logic is still not right on the latest commit; the scroll "sticks" to the wrong side (e.g. when the last item is highlighted and pressing arrow left on the container, the highlighted item is always against the right edge, but this doesn't match `inline/block: 'nearest'` which only sticks in the direction of the arrow key)

mj12albert:
> * match `inline/block: 'nearest'` which only sticks in the direction of the arrow key

Should be correct now, basically it should find the minimum distance to scroll right?

https://github.com/user-attachments/assets/c6ee24c2-8936-423a-a4a8-fcca3e75fcfd

atomiks:
@mj12albert yeah it looks right now - although I'm assuming the native `scroll-margin` (on the items) / `scroll-padding` (on the container) don't work and would also need to be checked for. 

Also would be good if it didn't have the slight jumps while navigating, maybe due to subpixel/fractional size calculation issues (`getBoundingClientRect` would handle those in theory but I saw it was removed)

mj12albert:
> `getBoundingClientRect` would handle those in theory but I saw it was removed

Yeah because distances from the absolute edge of the container are needed, but the Rect only measures from the visible edge and mixing it with `offset*` in the calculations causes issues

I updated the indicator positioning again and the prehydration script to not use a Rect as well, which also avoids having to check the border width

packages/react/src/tabs/tab/TabsTab.tsx
```diff
@@ -166,6 +167,7 @@ export const TabsTab = React.forwardRef(function Tab(
         onClick,
         onFocus,
         onPointerDown,
+        [ACTIVE_COMPOSITE_ITEM as string]: selected ? '' : undefined,
```
atomiks, line 170: This should fix https://github.com/mui/base-ui/issues/2099 as well

PR #2104, opened 2025-06-11
[code-infra] Update typescript-api-extractor

### LINKED ISSUE #2048: [popover] Type definition of `onOpenChange` appears to be inaccurate
# Docs feedback
For components that have an `onOpenChange()` callback prop, that prop's types are inaccurate. The documentation site says that `onOpenChange` has this signature:

```ts
onOpenChange?: (
    open: boolean,
    event?: Event | undefined,
    reason?: PopoverOpenChangeReason | undefined,
) => void;
```

... but the exported type has this signature:

```ts
onOpenChange?: (
    open: boolean,
    event: Event | undefined,
    reason: PopoverOpenChangeReason | undefined,
) => void;
```

In the exported type, `event` and `reason` params are required; on the docs page, they are optional.

At the same time, a ~~legacy~~ typedef in the code looks like this:

```
 /**
   * Event handler called when the popover is opened or closed.
   * @type (open: boolean, event?: Event, reason?: Popover.Root.OpenChangeReason) => void
   */
```

Ensuring that `onOpenChange`'s exported types match the documentation site would be a huge help! Based on my understanding, the documentation site is accurate, but the exported types are not.

I would be happy to implement a fix for this if maintainers are interested!

michaldudak:
Updated the Typescript API Extractor to 1.0.0-alpha.5 and solved a few issues with API docs generation. Custom `@type` annotation are no longer needed (except for data attributes and CSS variables)

Closes #2048


docs/reference/generated/slider-root.json
```diff
@@ -15,11 +15,11 @@
       "description": "The value of the slider.\nFor ranged sliders, provide an array with two values."
     },
     "onValueChange": {
-      "type": "((value: (number | number[]), event: Event, activeThumbIndex: number) => void)",
+      "type": "((value: number | number | readonly number[], event: Event, activeThumbIndex: number) => void)",
```
mj12albert: `number` is repeated here and in `onValueCommitted` below

docs/reference/generated/toast-root.json
```diff
@@ -8,7 +8,7 @@
       "description": "Direction(s) in which the toast can be swiped to dismiss."
     },
     "toast": {
-      "type": "Toast.Root.ToastObject",
+      "type": "Toast.Root.ToastObject<any>",
```
mj12albert, line 11: If `Data` [here](https://github.com/mui/base-ui/blob/master/packages/react/src/toast/root/ToastRoot.tsx#L605) was moved to the namespace could this be shown as `<Data>`?

michaldudak: No, this won't work. Handling type aliases is tricky and having them as type only adds complexity, so this isn't supported now. We could use the `@type` annotation if you'd like `<Data>` to appear here.

docs/reference/generated/alert-dialog-root.json
```diff
@@ -12,11 +12,11 @@
       "description": "Whether the dialog is currently open."
     },
     "onOpenChange": {
-      "type": "(open: boolean, event?: Event, reason?: AlertDialog.Root.OpenChangeReason) => void",
+      "type": "((open: boolean, event: Event | undefined, reason: AlertDialog.Root.OpenChangeReason | undefined) => void)",
       "description": "Event handler called when the dialog is opened or closed."
     },
     "actionsRef": {
-      "type": "RefObject<{ unmount: () => void; }>",
+      "type": "RefObject<{ unmount: (() => void) }>",
```
atomiks: Not sure if related but `actionsRef` types could be made consistent (`RefObject<AlertDialog.Root.Actions>`)

Like in Menu:

```
      "type": "RefObject<Menu.Root.Actions>",
```

michaldudak: :+1: Fixed

atomiks (approved):

Issue #602, opened 2024-09-11
[docs-infra] Document bundle size of each component


oliviertassinari:
### Search keywords

-

### Latest version

- [X] I have tested the latest version

### Summary

Bundle size matters. I think it's important we document it in the docs. Documenting:

- Forces us to measure it
- Means developers know we know what it is
- Means developers easily know if it's good enough for their use cases

For example, see https://github.com/propeldata/ui-kit/issues/157 who are looking for a select component, there is a reputation for bloat to debunk (Base UI) and solve (Material UI).

<img src="https://github.com/mui/material-ui/assets/3165635/ec2b46c6-bb07-4de8-abcd-cfa611dc9034" width="300">

### Examples

I tried two iterations:

- https://github.com/mui/material-ui/pull/40592
- https://github.com/mui/material-ui/pull/40549

but never had time to work on this.

### Motivation

It's related to #201.

atomiks:
Displaying bundle size accurately is a difficult problem. If you only use one single component from the library, then it’s fully accurate. However once you start using multiple components, the numbers become inaccurate as the sizes don’t add up simply/cumulatively as displayed due to shared dependencies/logic/compression, etc.

I propose we take a non-naive approach to this. https://bundlejs.com allows you to export the specific components you're actually using, treeshaking enabled, to see a true picture. We could have a UI for this on the website where they select all the components they're using/will decide to use to get a true picture of the bundle size.

oliviertassinari:
Oh, I forgot I had opened https://github.com/mui/material-ui/issues/40591. The best I could figure out in the past was https://github.com/pastelsky/bundlephobia/issues/389.

Link2Twenty:
Something Radix did well was allowing you to install just the parts you needed, if you want the select component you run `npm install @radix-ui/react-select` and that's all you get (with it's dependencies of course). This also means I can look at the [npm page](https://www.npmjs.com/package/@radix-ui/react-select) and see that the component is going to cost 328 kB. I will say they could do some work to make each npm page feel like it's real when it's too bare it feels like somewhere you've ended up by mistake.

By comparison if I want to use the base-ui select (`@base-ui-components/react/select`) I have to get the whole of `@base-ui-components/react` which I know rationally will treeshake on build but at a glance looking at the [npm page](https://www.npmjs.com/package/@base-ui-components/react) I see 3.48 MB which feels like a 10× increase.

[...]

PR #2105, opened 2025-06-12
[combobox] New `Combobox` and `Autocomplete` components

### LINKED ISSUE #222: [Combobox] Implement Combobox

atomiks:
Closes #222

## Combobox

Preview: https://deploy-preview-2105--base-ui.netlify.app/react/components/combobox

Perf: https://deploy-preview-2105--base-ui.netlify.app/experiments/combobox-perf

## Autocomplete

Preview: https://deploy-preview-2105--base-ui.netlify.app/react/components/autocomplete

## FilterableMenu

~Preview: https://deploy-preview-2105--base-ui.netlify.app/react/components/filterable-menu~ (will ship later)

--- 

To investigate in the future, possibly as separate components or new parts: 

- `Combobox` + `Tabs` - different semantics to the regular `Tabs` with virtual focus


docs/src/app/(public)/(content)/react/components/combobox/demos/hero/css-modules/index.tsx
```diff
@@ -0,0 +1,289 @@
...
+      country.name.toLowerCase().includes(inputValue.toLowerCase()),
+    );
+  }, [inputValue]);
+
+  return (
+    <Combobox.Root
+      value={value}
+      onValueChange={(nextValue) => {
+        setValue(nextValue);
+        React.startTransition(() => setInputValue(nextValue));
```
atomiks: No idea why but I noticed `React.startTransition` (or `React.useDeferredValue`) can break, but only with Fast Refresh(?) after editing the files. It ends up not updating the value.

Not sure if it's a React, Next.js, MDX, the demo, or source code bug causing it though

michael-land:
it would be really nice this component will also support autoHighlight, similar to MUI Autocomplete.

atomiks:
@michael-land in this current implementation, the `selectable` prop enables the selected item to be remembered as the `value` state, which then shows as the highlight upon opening (which can be styled with `data-selected` and/or `data-highlighted`)

This should enable creating the same effect as in the Material UI autocomplete `autoHighlight` prop:

Screenshot 2025-06-13 at 9 15 28 am

packages/react/src/menu/root/MenuRoot.tsx
```diff
@@ -385,6 +382,8 @@ export const MenuRoot: React.FC<MenuRoot.Props> = function MenuRoot(props) {
     disabledIndices: EMPTY_ARRAY,
     onNavigate: setActiveIndex,
     openOnArrowKeyDown: parent.type !== 'context-menu',
+    virtual: true,
+    virtualItemRef,
```
atomiks: When you use a `Combobox` inside a `Menu`, the menu's navigation must become virtual since real focus needs to remain on the `Combobox.Input`, which gets stolen by the menu items otherwise

msholmesdev:
Good morning! I am not sure where to best ask this so I apologize if this is the wrong location. I see this is actively being worked on. Do yall have a rough timeline you plan on it being merged in? Thank you!

colmtuite:
@msholmesdev We don't operate on fixed timelines like that. But you should expect it in a couple months. Our Discord is the best place for questions like this.

packages/react/src/combobox/root/ComboboxRootContext.tsx
```diff
@@ -0,0 +1,79 @@
...
+    value: any,
+    event: Event | undefined,
+    reason: ValueChangeReason | undefined,
+  ) => void;
+  open: boolean;
+  setOpen: (
+    open: boolean,
+    event: Event | undefined,
+    reason: BaseOpenChangeReason | undefined,
+  ) => void;
```
romgrk: You should aim to have a single point of truth for state. `open` and `selectedValue` should come from the store, those are unstable values that change the context, so they force a re-render for all its consumers. I also see more state values below that could be moved to the store.

The context should ideally contain only stable values like refs or setters, otherwise you'll end up with re-renders and the whole `Store` logic is wasted.

michaldudak:
A couple of issues I found:

- Add option demo: when you type something followed by a space, the suggestion list doesn't show the Add message. And when you type in an existing item followed by a space (for example "Angular "), the suggestion box is empty
- Grid navigation works inconsistently. Up/down arrows wrap around and remain in the same column while left/right go to the previous/next row.
- I don't find the basic example useful. I know it's similar to the ARIA one, but I don't think I've seen such combobox in the wild. IMO input autofill and select should be at the top, as they might be the most common
- I don't know how I did it, but I got the multiple selection example close whenever I pressed Enter. Can't reproduce it now, though.
- Input autofill: I intuitively expected the popup to close when I enter a text that isn't present on the list and I press enter.

atomiks:
> Grid navigation works inconsistently. Up/down arrows wrap around and remain in the same column while left/right go to the previous/next row.

This is expected no? (Grids allow wrapping with ArrowRight/Left - it's how Calendar works for instance)

> I don't find the basic example useful. I know it's similar to the ARIA one, but I don't think I've seen such combobox in the wild. IMO input autofill and select should be at the top, as they might be the most common

Yeah, fair enough

> Input autofill: I intuitively expected the popup to close when I enter a text that isn't present on the list and I press enter.

The popup doesn't show when it's empty in this example? Unless you mean in general for the other demos

michaldudak:
> This is expected no? (Grids allow wrapping with ArrowRight/Left - it's how Calendar works for instance)

I don't think there's a gold standard for this. macOS emoji picker doesn't wrap vertically. Google calendar doesn't wrap at all and Slack's emoji picker works like your implementation. I likely wouldn't notice if I wasn't testing it specifically so maybe it's ok.

> The popup doesn't show when it's empty in this example? Unless you mean in general for the other demos

I mean this:

https://github.com/user-attachments/assets/43d8e28b-b3ec-48f6-8f94-33b2f064416a

When no option is selected, <kbd>Enter</kbd> doesn't do anything. I'm not sure if there's any spec for this, but intuitively I expected the popup to close.

mnajdova:
I just notice that the selected tags spacing is not the consistent

Screenshot 2025-07-14 at 13 16 26

atomiks:
@mnajdova I'm guessing this is because the style uses `gap: 0.1rem` but should instead be an evenly divisible number (by 16 - `0.125rem`), otherwise rounding artifacts can occur; will update

docs/src/app/(public)/(content)/react/components/combobox/page.mdx
```diff
@@ -0,0 +1,143 @@
...
+<Reference
+  component="Combobox"
+  parts="Root, Input, Clear, Chips, Chip, ChipRemove, SelectedValue, List, Trigger, Portal, Positioner, Popup, Status, Empty, Collection, Row, Item, ItemIndicator, Group, GroupLabel, Separator"
+/>
+
+## Examples
+
+### Select
+
+The combobox can require the input value to match one of its predefined options. Trigger and clear buttons are also added to the input container.
```
mj12albert:
```suggestion
The combobox can require the input value to match one of its predefined options. `Combobox.Trigger` and `Combobox.Clear` buttons are also added to the input container.
```
I think explicitly naming the part names involved helps to learn the anatomy, maybe the relevant lines in the demo could be highlighted as well since the demos are all pretty long

docs/src/app/(public)/(content)/react/components/combobox/page.mdx
```diff
@@ -0,0 +1,143 @@
...
+
+### Inside `Popover`
+
+The combobox can be rendered inline inside a parent popover to create a searchable select menu.
+
+<Demo path="./demos/inside-popover" compact />
+
+### Grid layout
+
+The combobox items can be navigated in a grid layout by wrapping each row in `Combobox.Row` adding the `cols` prop to `Combobox.Root`.
```
mj12albert: I think it would help if the description of each demo explicitly named/highlighted the key subcomponents or props involved (and highlight in the code as well) like this one whenever possible

That way it will be easier to get a summary of how each demo/behavior differs from another without having to manually look for differences between them ~

mj12albert:
```suggestion
The combobox items can be navigated in a grid layout by wrapping each row in `Combobox.Row` and adding the `cols` prop to `Combobox.Root`.
```

docs/src/app/(public)/(content)/react/components/combobox/demos/async/css-modules/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+async function searchMovies(query: string): Promise<MovieItem[]> {
+  // Simulate network delay
+  await new Promise((resolve) => {
+    setTimeout(resolve, Math.random() * 500 + 100);
+  });
+
+  // Simulate occasional network errors (1% chance)
+  if (Math.random() < 0.01) {
```
Janpot: Wanted to try out error handling, but there's no way to simulate. Perhaps instead you could make it error on a specific `query`?

docs/src/app/(public)/(content)/react/components/combobox/page.mdx
atomiks: > maybe the relevant lines in the demo could be highlighted as well since the demos are all pretty long

Yeah I've wanted this, but the docs don't support it currently. Only regular MDX code blocks support it, not the demo code blocks

docs/src/app/(public)/(content)/react/components/combobox/demos/async/css-modules/index.tsx
atomiks: Added `will_error` special string

Dakkers:
@atomiks Is it possible to add the "clear button" to the "inside popover" example? I ask because in the Select example, the interactive element is an `input role="combobox"` with buttons inside, while the Inside Popover example uses a button as the interactive element, and putting a button inside a button is a no-no from my understanding.

atomiks:
@Dakkers usually selects have an option specifically for deselection (with a `null` value), such as in the regular Select hero demo (`Select font`)

Dakkers:
@atomiks yeah, I understand that – I was hoping it'd be possible to have an easy way for a user to clear the Combobox's selected value(s) when the popover is inside the menu. _edit_: this is supported for the other Combobox style (where the input is visible without clicking into the menu) and it's a good UX, especially for when the Combobox has multiple values

packages/react/src/combobox/root/ComboboxRootContext.tsx
```diff
@@ -0,0 +1,99 @@
...
+  handleEnterSelection: (event: Event) => void;
+  name: string | undefined;
+  disabled: boolean;
+  readOnly: boolean;
+  required: boolean;
+  fieldControlValidation: ReturnType<typeof useFieldControlValidation>;
+  cols: number;
+  isGrouped: boolean;
+  virtualized: boolean;
```
romgrk: When writing new components, if there is a store, all its mutable state should be in the store. `Select` was only partially refactored because I wanted to focus that PR on the perf aspect, but new components should have only one state management model rather than multiple ones.

packages/react/src/combobox/root/utils/useFilter.ts
```diff
@@ -0,0 +1,61 @@
...
+  cache.set(optionsString, collator);
+
+  return collator;
+}
+
+/**
+ * Matches items against a query using `Intl.Collator` for robust string matching.
+ */
+export function useFilter(options: Intl.CollatorOptions = {}) {
+  const collator = React.useMemo(() => getCollator(options), [options]);
```
romgrk: Why `useMemo`? `getCollator` already memoizes, that's 2 levels of memoization for the same value.

Also, all the uses I've seen of `Combobox.useFilter` specify `options` inline (as an unstable object across renders), so the `useMemo` is just pure overhead. If you really want to have a `useMemo`, you should make use of it in the examples.

romgrk: Furthermore, you could also cache the full filter object `({ contains, startsWith, endsWith })` rather than just the `collator` instance, you could avoid using any `useMemo` at all in `useFilter`.

atomiks:
Small update for this `Combobox` PR, after some internal discussion (I'll be doing some refactoring here)

Names and actual behaviors are still in flux. 

- `Combobox` will change to only have filterable select behavior (in this PR, currently `selectionMode="single"|"multiple"`). The input is restricted to a predefined set of values.
- `Autocomplete`, a new component, will be the current Combobox implementation with `selectionMode="none"`. Used when the input can have any value, not restricted to a predefined set. For example, Google Search.
- New `Menu` filtering behavior (possibly `FilterableMenu`), where the items fire off comands when clicked, unrelated to the input, and there's no selection remembered. Things like `cmd+k`, Notion's menu widget, an emoji picker, etc. Similar to `selectionMode="none"` in this PR, but the input doesn't reflect the value chosen as it's just used for filtering menu items.
- Textarea + Combobox (e.g. slash command / mentions widget) is currently in an unknown state. It would be great if it was supported externally for the time being using one of the given components, though.

This is because Combobox in its current form is taking on too many roles to the point of becoming overly monolithic (despite `role="combobox"` being flexible). It is hard to pin down exactly what a "Combobox" should be doing - we believe the select behavior is the most intuitive and common for this name

packages/react/src/combobox/root/utils/useFilter.ts
```diff
@@ -0,0 +1,77 @@
+interface Filter {
+  contains: (item: string, query: string) => boolean;
+  startsWith: (item: string, query: string) => boolean;
+  endsWith: (item: string, query: string) => boolean;
+}
+
+const collatorCache = new Map<string, Intl.Collator>();
```
romgrk: You don't need a collator cache, this logic can be removed. Just keep the filter cache.

packages/react/src/combobox/root/ComboboxRootContext.tsx
atomiks: Are you referring to derived/prop values though? All these ones that were highlighted get passed through context because they aren't managed as state

packages/react/src/combobox/group-label/ComboboxGroupLabel.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const { setLabelId } = useComboboxGroupContext();
+
+  const id = useBaseUiId(idProp);
+
+  useIsoLayoutEffect(() => {
+    setLabelId(id);
+    return () => {
+      setLabelId(undefined);
+    };
+  }, [id, setLabelId]);
```
romgrk, line 28: Is the cleanup function really useful to have with this pattern?

packages/react/src/combobox/group-label/ComboboxGroupLabel.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const element = useRenderElement('div', componentProps, {
+    ref: forwardedRef,
+    props: [{ id }, elementProps],
```
romgrk, line 32: Feels like we could instead easily just set `elementProps.id = ...` instead of creating a wrapper object & array.

packages/react/src/combobox/item/ComboboxItem.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+    useIsoLayoutEffect(() => {
+      if (hasRegistered && value === rootSelectedValue) {
+        registerItemIndex(index);
+
+        if (selectable && allowActiveIndexSyncRef.current) {
+          frame.request(() => {
+            itemRef.current?.scrollIntoView?.({ inline: 'nearest', block: 'nearest' });
+            store.set('activeIndex', index);
+            store.set('selectedIndex', index);
```
romgrk: The store is currently synchronous by design, so state updates are not batched. It's more efficient to update the state in one call with `store.apply()` (that I'll refactor to `store.update()` in the near future). Eventually we might have opt-in explicit batching (e.g. `store.batch(() => { ... })`), but for now we don't because it makes state updates predictable (as soon as you call it, you know all the listeners have been called).

packages/react/src/combobox/item/ComboboxItem.tsx
```diff
@@ -0,0 +1,271 @@
...
+            onItemHighlighted(value, { type: 'none', index });
+            allowActiveIndexSyncRef.current = false;
+          });
+          return () => {
+            frame.cancel();
+          };
```
romgrk:
```suggestion
          return frame.cancel;
```

The low-level utils I wrote all have cleanup functions that can be detached from their instance, this avoids the need to allocate a closure.

packages/react/src/combobox/list/ComboboxList.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }, [hasPositionerContext, store, popupRef]);
+
+  // Support "closed template" API: if children is a function, implicitly wrap it
+  // with a Combobox.Collection that reads items from context/root.
+  const resolvedChildren = React.useMemo(() => {
+    if (typeof children === 'function') {
+      return <ComboboxCollection>{children}</ComboboxCollection>;
+    }
+    return children;
+  }, [children]);
```
romgrk: In theory I like the idea of [memoizing the children](https://attardi.org/why-we-memo-all-the-things/), in practice I wonder if it really ever makes a difference for a case like this. Most devs never think about memoizing the children, and it looks like none of the docs example in this PR make use of that feature either, and there's a few props down below that aren't memoized, so it feels like the `useMemo` is just overhead.

packages/react/src/combobox/positioner/ComboboxPositioner.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  const resolvedAnchor = React.useMemo(() => {
+    if (anchor != null) {
+      return anchor;
+    }
+    if (defaultAnchor === 'trigger') {
+      return triggerElement;
+    }
+    return inputElement;
+  }, [anchor, defaultAnchor, triggerElement, inputElement]);
```
romgrk: The computation that is wrapped in `useMemo` is more expensive than the overhead of creating the allocations (closure, deps array) and checking the dependencies. I think you're overusing `useMemo`, it should be reserved for either non-primitive values that need to be stable, or computationally expensive operations. Calling `useMemo` is not a free operation - on a slow phone, it's 2-4ms a pop.

packages/react/src/combobox/positioner/ComboboxPositioner.tsx
```diff
@@ -0,0 +1,195 @@
...
+      hiddenStyles.pointerEvents = 'none';
+    }
+
+    return {
+      role: 'presentation',
+      hidden: !mounted,
+      style: {
+        ...positioning.positionerStyles,
+        ...hiddenStyles,
+      },
```
romgrk: No need to create an object just to destructure it right away into another. Just use one object.

```suggestion
    const style: React.CSSProperties = {
      ...positioning.positionerStyles,
    };
    if (!open) {
      style.pointerEvents = 'none';
    }

    return {
      role: 'presentation',
      hidden: !mounted,
      style,
```

packages/react/src/combobox/positioner/ComboboxPositioner.tsx
```diff
@@ -0,0 +1,195 @@
...
+    }),
+    [
+      positioning.side,
+      positioning.align,
+      positioning.arrowRef,
+      positioning.arrowUncentered,
+      positioning.arrowStyles,
+      positioning.anchorHidden,
+    ],
+  );
```
romgrk: 1. Could we structure the return value of `useAnchorPositioning` in a way that avoids having to recreate a separate object for the context?
2. I see a similar but just a bit different pattern in [other uses of `useAnchorPositioning`](https://github.com/mui/base-ui/blob/d24991095efcbaa9587b1f610b7638d3ca48256c/packages/react/src/preview-card/positioner/PreviewCardPositioner.tsx#L88-L103). Would it make sense to standardize the "positioner context" values into one format?

packages/react/src/combobox/root/ComboboxPublicRoot.tsx
romgrk, line 1: Why not keep the filenames matching their top-level component? (`ComboboxRoot.tsx` and `ComboboxRootInternal.tsx`)

And why is the component split that way?

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -0,0 +1,1135 @@
...
+  const multiple = selectionMode === 'multiple';
+
+  const frame = useAnimationFrame();
+
+  useIsoLayoutEffect(() => {
+    setControlId(id);
+    return () => {
+      setControlId(undefined);
+    };
+  }, [id, setControlId]);
```
romgrk: This pattern is repeated often, would it make sense to add it to `useBaseUiId(idProp, setControlId)`?

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -0,0 +1,1135 @@
...
+  const [selectedValue, setSelectedValueUnwrapped] = useControlled<any>({
+    controlled: selectedValueProp,
+    default: multiple ? (defaultSelectedValue ?? EMPTY_ARRAY) : defaultSelectedValue,
+    name: 'Combobox',
+    state: 'selectedValue',
+  });
+
+  const [queryChangedAfterOpen, setQueryChangedAfterOpen] = React.useState(false);
+
+  const collatorFilter = useFilter({ sensitivity: 'base' });
```
romgrk: Static object can be extracted.

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -0,0 +1,1135 @@
...
+  });
+
+  const query = React.useMemo(() => {
+    if (inputValue === '') {
+      return '';
+    }
+    return String(inputValue).trim().toLocaleLowerCase();
+  }, [inputValue]);
+
+  const isGrouped = React.useMemo(() => isGroupedItems(items), [items]);
```
romgrk: In both of these cases, `useMemo` is more expensive than the operation it wraps.

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -0,0 +1,1135 @@
...
+  } = useOpenInteractionType(open);
+
+  const initialList = React.useMemo(() => {
+    if (virtualized) {
+      return Array.from({ length: flatFilteredItems.length }, () => null);
+    }
+    return [];
+  }, [virtualized, flatFilteredItems]);
+
+  const listRef = React.useRef<Array<HTMLElement | null>>(initialList);
```
romgrk: I find this logic a bit hard to understand, it's not clear to me how it's used or for what. Also, I see that sometimes `initialList === listRef.current`, but sometimes not (when `listRef.current = initialList.slice()`), is that intentional?

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -0,0 +1,1135 @@
...
+    ) {
+      return;
+    }
+
+    setQueryChangedAfterOpen(true);
+  }, [open, query, defaultInputValue]);
+
+  useIsoLayoutEffect(() => {
+    prevQueryRef.current = query;
+  }, [query]);
```
romgrk: Any way to simplify/merge these effects? Just like `useMemo`, calling `useLayoutEffect` is not a free operation.

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -0,0 +1,1135 @@
...
+      : selectedValue !== null && selectedValue !== undefined && selectedValue !== '';
+    setFilled(hasValue);
+    if (prevValueRef.current !== selectedValue) {
+      updateValue(selectedValue);
+    }
+  }, [setFilled, updateValue, selectedValue, multiple]);
+
+  useIsoLayoutEffect(() => {
+    prevValueRef.current = selectedValue;
+  }, [selectedValue]);
```
romgrk: For both `prevQueryRef` and `prevValueRef`, it would be preferable to encapsulate this behavior in a hook analogous to [useLatestRef](https://github.com/mui/base-ui/blob/master/packages/utils/src/useLatestRef.ts). It will be both more performant (no allocations), and have a better DX, its interface could look something like this:

```ts
const previousQuery = usePreviousValue(query)
```

edit: something like [this (untested) implementation](https://gist.github.com/romgrk/3e4026ce4acda65211cd14b5a6ed4bff) - feel free to rename it, maybe just `usePrevious` could be fine as well

packages/react/src/combobox/root/utils/useFilter.ts
```diff
@@ -0,0 +1,62 @@
...
+  filterCache.set(optionsString, filter);
+  return filter;
+}
+
+/**
+ * Matches items against a query using `Intl.Collator` for robust string matching.
+ */
+export function useFilter(options: Intl.CollatorOptions = {}) {
+  return getFilter(options);
+}
```
romgrk: So technically, `const useFilter = getFilter` would be correct. This highlights that `useFilter` is actually not a hook, even though it has "use" in its name. I think the function should be kept as just `getFilter` and used as-is. The "use" token has a specific semantic and constraints attached to it in the react world, it would be better to maintain that convention. If it's just a plain function, it should look like a plain function.

packages/utils/src/useControlled.ts
```diff
@@ -3,6 +3,20 @@
...
+ */
+function safeStringifyForDeps(value: unknown): string | unknown {
+  try {
+    return JSON.stringify(value);
+  } catch {
+    // If JSON.stringify fails (e.g., circular structure),
+    // return the original value for reference comparison
+    return value;
+  }
+}
```
romgrk: If comparing by reference is correct, then why are we stringifying in the first place? Stringify has a possibly huge overhead compared to just a comparison by reference.

If an actually correct stringify function is needed, we have [this one in X](https://github.com/mui/mui-x/blob/9eee74b01531b95bdecf056ace37f4421771b8fc/packages/x-internals/src/hash/stringify.ts).

docs/src/app/(public)/(content)/react/components/autocomplete/demos/async/css-modules/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  const shouldRenderPopup = searchValue !== '';
+
+  return (
+    <Autocomplete.Root
+      items={searchResults}
+      inputValue={searchValue}
+      onInputValueChange={setSearchValue}
+      itemToString={itemToString}
+      filter={filterByTitleOrYear}
```
JeromeDeLeon: Sorry for jumping in. Is this prop needed here? I thought the point of async search is to search via API (be it through TanStack Query or `useEffect`), and then the result will be provided via the `items` prop.

packages/react/src/combobox/positioner/ComboboxPositioner.tsx
atomiks: I was tending to use it as an IIFE for the `const` sometimes, rather than calculating it externally, but since it's not free I'll do the calculations in the component body itself

packages/react/src/combobox/group-label/ComboboxGroupLabel.tsx
atomiks: If the label is removed dynamically, the id needs to be unset

packages/react/src/combobox/root/ComboboxPublicRoot.tsx
atomiks: Combobox is an internal 'god' component that `Autocomplete` and `FilterableMenu` are using. The public `Combobox` name is a filterable select but internally it supports more use cases.

packages/react/src/combobox/root/ComboboxRoot.tsx
atomiks: Interesting. I thought `trim()`/`toLocaleLowerCase()` would be more expensive?

packages/react/src/combobox/root/ComboboxRoot.tsx
atomiks: I struggled creating something that didn't mutate the ref in render. It seems like yours is doing that?

packages/react/src/combobox/root/utils/useFilter.ts
atomiks: I tend to agree, but in Floating UI I had named a hook with `use` despite not using React hooks, which later was refactored to use them, and I didn't need to cause a breaking change 🙈 I'm not sure if that would happen here, but still

docs/src/app/(public)/(content)/react/components/autocomplete/demos/async/css-modules/index.tsx
atomiks: Are you referring to `filter`? The items are in object format and the internal filter compares strings or `object.value`

Edit: ah nvm, I see what you're asking. The demos were refactored quite a lot and I need to re-review them, so I'll have to see why `filter` is specified.

packages/utils/src/useControlled.ts
atomiks: This is an error - removed

packages/react/src/combobox/root/ComboboxRoot.tsx
romgrk: Not sure what you mean by that. If you want to have the initial value be `undefined` rather than `value`, that's doable.

packages/react/src/combobox/root/utils/useFilter.ts
romgrk: Is this function public?

packages/react/src/combobox/root/ComboboxPublicRoot.tsx
romgrk: Makes sense for the split. Why not keep the filenames consistent with their components though?

packages/react/src/combobox/root/utils/useFilter.ts
atomiks: It was the `useInteractions` hook. Originally it had no memoization at all, but then I added it. It hasn't changed since then - and it may not apply here unless we want to start using `React.useMemo()` instead of the global cache/custom memo we have

`NumberField` also does a similar thing

packages/react/src/combobox/root/ComboboxRoot.tsx
atomiks: The value is being written to the ref in render:
```js
instance.next = value;
instance.dependencies[0] = value;
```

I don't know how this might affect concurrent features

packages/react/src/combobox/root/utils/useFilter.ts
romgrk: I meant "is `useFilter` public?", and it seems that it's not, so we shouldn't have to worry about breaking changes.

packages/react/src/combobox/root/ComboboxRoot.tsx
romgrk: > The value is being written to the ref in render:

That's incorrect, the "ref" is the outer wrapper, `instance` is not the ref. Besides, the observed value outside of the hook doesn't change following those statements, it changes after the layout effect has run. Note that the return value of that hook should read `return instance.current` rather than `return instance`, I copy/pasted `useLatestRef` as a starting point, but we don't need the consumers of the hook to be aware of implementation details.

> I don't know how this might affect concurrent features

Could you state a specific case/example you're worried about?

packages/react/src/combobox/root/ComboboxRoot.tsx
romgrk: Definitely not unless you have a huge string, which the query string will not be. Those operations aren't even likely to modify the query (users are unlikely to prepend/append spaces, or to type in uppercase letters).

packages/react/src/combobox/root/utils/useFilter.ts
atomiks: Ah, yeah it is public - there are various demos where you want to use the filter externally (controlled)

mnajdova:
The docs pages look awesome! I found some things that I am not sure about based on the demos:

I was a bit surprised that clicking on x on one of the selected elements in https://deploy-preview-2105--base-ui.netlify.app/react/components/combobox#multiple-select closes the list, shouldn't it be kept open while interacting with the elements inside the input? Same while navigating between the selected items with the left/right arrow.

This jumps too much when searching: https://deploy-preview-2105--base-ui.netlify.app/react/components/filterable-menu#command-palette, can we make sure the input is always in the same location on the screen and only change the height of the dialog?

I will start reviewing the code next.

atomiks:
@mnajdova we can skip review of FilterableMenu demos/logic for now since it's not in the upcoming release. I will just hide it from the docs and package and work on it in a new PR.

mnajdova:
> @mnajdova we can skip review of FilterableMenu demos/logic for now since it's not in the upcoming release. I will just hide it from the docs and package and work on it in a new PR.

Yeah fair enough, do you agree for the behavior for the multiple combobox tough?

atomiks:
@mnajdova yeah I do agree

I think https://mui.com/material-ui/react-autocomplete/#multiple-values is a good reference for this behavior and it doesn't close when clicking the chips. It does however close in a `multiple` select when selecting a single item - I suppose this should be configurable

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -0,0 +1,1131 @@
...
+  props: ComboboxRootConditionalProps<Item, Mode>,
+): React.JSX.Element {
+  const {
+    id: idProp,
+    onOpenChangeComplete: onOpenChangeCompleteProp,
+    defaultSelectedValue = null,
+    selectedValue: selectedValueProp,
+    onSelectedValueChange,
+    defaultInputValue = '',
+    inputValue: inputValueProp,
```
mnajdova:
```suggestion
    inputValue: inputValueProp,
    setInputValueChange,
```

I'm curious, why is it not destructed? I was wondering why we are missing this prop, and I saw it used later down in the code.

mnajdova:
> It does however close in a multiple select when selecting a single item - I suppose this should be configurable

It's up for discussion, I would keep it open in that case too, as otherwise it is inconsistent with the keyboard selection, if I press enter on an item it doesn't close the list, why would click close the list, it's just weird experience.

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -0,0 +1,1131 @@
...
+        />
+      );
+    });
+  }, [multiple, selectedValue, name, itemToValue]);
+
+  const children = (
+    <React.Fragment>
+      {props.children}
+      <input
+        {...fieldControlValidation.getInputValidationProps({
```
mnajdova: Can we move the creation of this props object above? It would be more readable then having it inlined in the JSX.

packages/react/src/combobox/root/ComboboxPublicRoot.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * Doesn't render its own HTML element.
+ */
+export function ComboboxRoot<Item = any, Multiple extends boolean | undefined = false>(
+  props: ComboboxRoot.Props<Item, Multiple>,
+): React.JSX.Element {
+  const { multiple = false as Multiple, ...rest } = props;
+
+  type Mode = ModeFromMultiple<Multiple>;
+  const mode = multiple ? 'multiple' : 'single';
+  return <ComboboxRootInternal<Item, Mode> {...(rest as any)} selectionMode={mode} />;
```
mnajdova: API consideration, considering we have an adapter for the ComboBox (filterable select), can we rename the `items` prop to `options`? It's really weird to see items as a name in a context of select/combobox. It could make sense for the FilterableMenu, even for the autocomplete I would use options honestly

docs/src/app/(public)/(content)/react/components/combobox/demos/input-inside-popup/css-modules/index.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+import { Combobox } from '@base-ui-components/react/combobox';
+import styles from './index.module.css';
+import { countries, type Country } from './data';
+
+export default function ExamplePopoverCombobox() {
```
mnajdova, line 5: This demo looks a bit weird in CodeSandbox, I haven't investigated what may be the issue, we can look into it in a follow up:

Screenshot 2025-08-19 at 18 56 11

packages/react/src/combobox/chip/ComboboxChip.tsx
```diff
@@ -0,0 +1,151 @@
...
+    state,
+    props: [
+      {
+        tabIndex: -1,
+        'aria-disabled': disabled || undefined,
+        'aria-readonly': readOnly || undefined,
+        onKeyDown(event) {
+          const nextIndex = handleKeyDown(event);
+
+          ReactDOM.flushSync(() => {
```
mnajdova, line 105: Why do we need `flushSync` here?

packages/react/src/combobox/clear/ComboboxClear.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const element = useRenderElement('button', componentProps, {
+    state,
+    enabled: shouldRender,
+    ref: [forwardedRef, buttonRef],
+    props: [
+      {
+        disabled,
+        'aria-readonly': readOnly || undefined,
+        onMouseDown(event) {
```
mnajdova, line 95: I am curious why do we have the onMouseDown event too?

packages/react/src/combobox/selected-value/ComboboxSelectedValue.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { useStore } from '@base-ui-components/utils/store';
+import { useComboboxRootContext } from '../root/ComboboxRootContext';
+import { stringifyItem } from '../root/utils';
+import { selectors } from '../store';
+
+/**
+ * The currently selected value.
+ * Doesn't render its own HTML element.
+ */
+export function ComboboxSelectedValue(props: ComboboxSelectedValue.Props) {
```
mnajdova: Should we have different components for single vs multiple selection. Having this component when multiple is set reads really weirdly in my opinion, e.g.:

```
<Combobox.SelectedValue>
  {(value: Lang[]) => (
    <React.Fragment>
      {value.map((language) => (
        <Combobox.Chip
          key={language.id}
          className={styles.Chip}
          aria-label={language.value}
        >
          {language.value}
          <Combobox.ChipRemove className={styles.ChipRemove} aria-label="Remove">
            <XIcon />
          </Combobox.ChipRemove>
        </Combobox.Chip>
      ))}
      <Combobox.Input
        id={id}
        placeholder={value.length > 0 ? '' : 'e.g. TypeScript'}
        className={styles.Input}
      />
    </React.Fragment>
  )}
</Combobox.SelectedValue>
```

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -665,7 +667,11 @@ export function ComboboxRoot<Item = any, Mode extends SelectionMode = 'none'>(
       anchorElement !== inputElement ? { mouse: 'intentional', touch: 'sloppy' } : undefined,
     outsidePress(event) {
       const target = getTarget(event) as Element | null;
-      return !contains(triggerElement, target);
+      return (
+        !contains(triggerElement, target) &&
+        !contains(clearRef.current, target) &&
+        !contains(chipsContainerRef.current, target)
```
mnajdova: Do we expect the ComboBox.Input to always be inside the chips container?

docs/src/app/(public)/(content)/react/components/combobox/demos/input-inside-popup/css-modules/index.tsx
atomiks: Ah, I always have this issue because the docs specify a CSS reset by default but CodeSandbox doesn't, and they need to have things like `box-sizing: border-box` added manually.

imo the demos container should `unset` all the default styles back to their defaults so we can catch this while developing

packages/react/src/combobox/root/ComboboxRoot.tsx
atomiks: Possibly not, but I wonder if clicking into the container / onto the chips should never close the popup regardless

packages/react/src/combobox/selected-value/ComboboxSelectedValue.tsx
atomiks: Yeah though it does give control over the rendering. What did you have in mind?

packages/react/src/combobox/clear/ComboboxClear.tsx
atomiks: Prevents focus from being stolen from the input. `event.preventDefault()` or relocating the focus could both work here - relocating focus isn't as nice with touch though

packages/react/src/combobox/chip/ComboboxChip.tsx
atomiks: The highlighted chip index is set at the same time as the `.focus()` call, synchronously

packages/react/src/combobox/root/ComboboxPublicRoot.tsx
atomiks: Our `Select` API uses `items` prop on the root, and we use `Select.Item`

packages/react/src/combobox/root/ComboboxRoot.tsx
atomiks: > There is a bug in the circular keyboard navigation, there is one more step (I think it's the input) between the last and first option, e.g. it doesn't go from the first to the last option and vice versa immediately.

Whether referring to the items or chips: this isn't a bug. It intentionally includes the input in the focus loop so it can be reset; useful for screen readers to return to the input after reaching the end of the list, or in autofill demos to reset the input, etc. Google Search for instance does this. I believe focusing the first item immediately also fails to announce the item in VoiceOver.

michaldudak:
First of all: great work! Most of the functionality is flawless.

Random things I noticed:

- In the Creatable demo, pressing <kbd>Enter</kbd> should work the same as choosing to add a new item. Otherwise, you can end up having both chips and free text, which doesn't look right
  image
- Similarly, the multiple example allows free text to be present. I'd expect invalid values to be cleared on blur (or is this somehow configurable?)
- The Input inside popup example could use typeahead on a closed combobox (similarly to how Select works), so for example, focusing the combobox and pressing "D" would select "Denmark" without opening the popup. That's how I usually operate with selects when I know what options are there.
- Why is the controlled value prop called `selectedValue`, and not `value` as in Select?
- I haven't found in the docs what the Row part is for and why we may need it around Item
- In Autocomplete, can we limit the number of displayed suggestions?

packages/react/src/combobox/list/ComboboxList.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const setPositionerElement = useEventCallback((element) => {
+    store.set('positionerElement', element);
+  });
+
+  const setListElement = useEventCallback((element) => {
+    store.set('listElement', element);
+  });
+
+  const handleKeyDown = useEventCallback((event: React.KeyboardEvent<HTMLDivElement>) => {
```
romgrk: Why does this use `useEventCallback`? Other event handlers in the props are specified as inline closures, so this seems to be wasted.

packages/react/src/combobox/popup/ComboboxPopup.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    return initialFocus;
+  }, [initialFocus, openMethod, isAnchorInput, inputRef, popupRef]);
+
+  const resolvedFinalFocus = React.useMemo(() => {
+    if (finalFocus != null) {
+      return finalFocus;
+    }
+    return isAnchorInput ? false : triggerRef;
+  }, [finalFocus, isAnchorInput, triggerRef]);
```
romgrk: Idem about `useMemo` overhead.

packages/react/src/combobox/item/ComboboxItem.tsx
```diff
@@ -0,0 +1,264 @@
...
+      readOnly,
+      virtualized,
+      listRef,
+    } = useComboboxRootContext();
+
+    const { flatFilteredItems } = useComboboxDerivedItemsContext();
+
+    const selectable = selectionMode !== 'none';
+    const multiple = selectionMode === 'multiple';
+    const index = virtualized ? flatFilteredItems.indexOf(value) : listItem.index;
```
romgrk: Looks quite unefficient both ways. The index is a known property for a list item, it should be passed as a prop to the item.

atomiks: We can make `index` an optional property; to support an open API use case it needs to use the DOM order index still

atomiks:
> Why is the controlled value prop called selectedValue, and not value as in Select?

I originally had something like this but realized it was super confusing what `value` referred to. I saw different implementations used `value` for the selected value and others used it for the input value. So I decided that the slightly more verbose name was better to remove the ambiguity. 

However, that may have been more of a concern when Combobox was monolithic. Now that they're split, `value` for selected value and `inputValue` for Combobox could work; while for Autocomplete, `value` could refer to the `inputValue` instead?

michaldudak:
Yes, this is what I'd intuitively expect. It'll also make transitioning from Select to Combobox easier.

packages/react/src/combobox/item/ComboboxItem.tsx
romgrk: > to support an open API use case it needs to use the DOM order index

Or a required `index` property. What's the rationale for making `index` required on `Slider.Thumb` but not here? It solves the same underlying issue.

packages/react/src/combobox/root/utils/useFilter.ts
romgrk: Ok. Might still be nice to `const useFilter = getFilter`, less bytes, less runtime.

On a sidenote, would the naming make more sense as `useCollatorFilter`? The params make it a very `Intl.Collator`-specific function, and it would leave space for other (non-`Intl.Collator`) filters if there ever are more.

packages/react/src/floating-ui-react/hooks/useListNavigation.ts
```diff
@@ -508,32 +508,32 @@ export function useListNavigation(
   const hasActiveIndex = activeIndex != null;
 
   const item = React.useMemo(() => {
-    function syncCurrentTarget(currentTarget: HTMLElement | null) {
+    function syncCurrentTarget(currentTarget: HTMLElement | null, event: React.SyntheticEvent) {
```
romgrk: Just keep `event` as a parameter, no need for both.

oliviertassinari:
Did we look into the bundle size? Does it really need to be one of the heaviest? It's tiny in Material UI:

- Material UI unstyled autocomplete: **3.5kB** gzipped https://bundlejs.com/?q=%40mui%2Fmaterial%2FuseAutocomplete&treeshake=%5B*%5D&text=%22export+%7B+useAutocomplete+%7D%3B%22. https://mui.com/material-ui/react-autocomplete/#customized-hook
- Downshift: **14.7kB** gzipped: https://bundlejs.com/?q=downshift&treeshake=%5B*%5D&text=%22export+%7B+useCombobox%C2%A0%7D%22 https://www.downshift-js.com/use-combobox
- React Select: **34.6kB** gzipped: https://bundlejs.com/?q=react-select&treeshake=%5B*%5D&text=%22export+%7BSelect%7D%22 https://react-select.com/home
- Headless UI ComboBox: **42.5kB** gzipped: https://bundlejs.com/?q=%40headlessui%2Freact&treeshake=%5B*%5D&share=KYDwDg9gTgLgBAbzgYQgWwEYSyANC9LHAIQFcYYIA7fVTbCEASSrHNsIZAHkwYBLah3o5eA6gGc4AXyA https://headlessui.com/react/combobox
- Base UI ComboBox: **43.4kB** gzipped https://frontend-public.mui.com/size-comparison/mui/base-ui/diff?prNumber=2105&baseRef=master&baseCommit=0f2b7a1e4220a61f496257c7497e926e2b69a03a&headCommit=ae29f2720c1c4b6670e2e0e68f3f751b448b997b&circleCIBuildNumber=135868
- React Aria ComboBox: **52.9kB** gzipped https://bundlejs.com/?q=react-aria-components&treeshake=%5B*%5D&text=%22export+%7B%C2%A0ComboBox%7D%22 https://react-spectrum.adobe.com/react-aria/ComboBox.html

atomiks:
@oliviertassinari the upfront cost is substantial but once you're using it with e.g. Select, then the cost is only +5-10 kb roughly. If you use Dialog and add Combobox after, it's close to around half the cost. I mentioned the shared dependencies here a bit: https://github.com/mui/base-ui/issues/602#issuecomment-2345217478

And some things ours includes that are lacking in the smaller ones:

- Proper anchor positioning
- `Field`/`Form` integration
- Animation API integration
- Grid list navigation (grid/gridcell/rows)
- The component API (not just a hook with some spreading), which has a bunch of overhead to make the DX nice
- Correctness and bug handling that is lacking in many of those. e.g. pressing Tab in Material UI loses focus to the body and doesn't support iOS VoiceOver, which ours does (adds nearly 5 kB to the total)
- Supports an open API not only closed
- The size presented is if you're using **every** component - if you only use some, treeshaking eliminates a bunch of the size

oliviertassinari:
> then the cost is only +5-10 kb roughly

@atomiks The size of a component once used with another component is something that feels like we should not look at much (but still looked at a bit to make sure we don't duplicate absurd stuff).
1. As a user, what I care is the standalone size, I will weight the pros and cons with using another library. I think react-select is a proof that people are willing to do the effort to cherry pick. People's codebase is a mix of so many different libraries, where evergreen projet, starting from starch are the minority.
2. The same argument of size with more components applies to Headless UI and Material UI. To be fair, the example of Material UI I used is the most barebone version, I think it has 80% of what people need in real projets, but still, it's missing the anchor position to work within dialogs. And maybe to fix https://github.com/mui/material-ui/issues/21661.

So I think success is about being clearly smaller than Headless UI, and I think smaller than React Select since they also bundle emotion. So it feels like 25-30KB should be the maximum JS size budget we allow ourselves, because a. competition won't be able to go a lot lower, and b. it shouldn't need more.

> The size presented is if you're using every component - if you only use some, treeshaking eliminates a bunch of the size

Ok, so this size number needs to be updated to reflect a comparable canoncial use case as the other libraries.

> Supports an open API not only closed

This makes me connect the dot that the demos in the docs feels wrong: that it shouldn't be about picking one value among 7 options, but 300 to reflect how people will evaluate their options.

michaldudak:
1. I found it unexpected to have the clear and trigger buttons focusable in the hero demo. Usually, when I use the keyboard, I select the value and expect the <kbd>Tab</kbd> key to move to the next field immediately. Focusing the trigger seems unnecessary, especially since there is already an alternative way to open the list (using the down arrow).
As for the clear button, the most similar example I could find is the clear button in native search fields. It also doesn't create a tab stop, and its hotkey is <kbd>Esc</kbd>.

2. In the creatable demo, there's a difference in behavior depending on whether you press <kbd>Enter</kbd> or select the "Create..." command - only the latter opens the Dialog. Since the item can theoretically be an object with multiple fields, the Dialog should be displayed in both cases.

3. I'm building a tag selector (quite similar to the Creatable demo, so multi-select with the ability to specify new options, but without a Dialog). One feature I'd find useful would be custom confirmation keys. In this specific scenario, users might type "red, blue" and have the comma act like Enter (either creating a new tag or selecting one from existing).

4. This is more general feedback about our demos - IMO some users might struggle to create reusable DS components based on the demos (especially complex components like the combobox). The main reason is that it's not immediately clear what parts of code are inherent to the reusable component and what are demo-specific (for example, I don't know without deeper analysis if state variables in the creatable demo support this specific scenario or can be used to create a generic component).

docs/reference/generated/autocomplete-root.json
```diff
@@ -0,0 +1,141 @@
...
+      "description": "Filter function used to match items vs input query.\nThe `itemToLabel` function is provided to help convert items to strings for comparison.",
+      "detailedType": "| ((\n    value: Value,\n    query: string,\n    itemToLabel: ((value: Value) => string) | undefined,\n  ) => boolean)\n| null\n| undefined"
+    },
+    "itemToValue": {
+      "type": "((value: Value) => string)",
+      "description": "Converts an item's value to its string representation for display in the input.",
+      "detailedType": "((value: Value) => string) | undefined"
+    },
+    "items": {
+      "type": "Group<Value>[] | Value[]",
```
michaldudak, line 69: This shows as `any[] | Group[] | undefined` in the docs. Is this the stale cache issue again?

atomiks:
@mnajdova it's intentional - without `autoHighlight`, the highlight doesn't remain "trapped" in the listbox and can be removed (so virtual focus returns to the input). This also applies to opening the listbox for the first time (the Google Search combobox for example works like this). If there is a selection made in Combobox, then the highlight is always on the selected item though when opening

mnajdova:
I feel this is wrong (at least for a default behavior). Based on [ARIA](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/):
> Down Arrow: If the popup is available, moves focus into the popup:
> - If the autocomplete behavior automatically selected a suggestion before Down Arrow was pressed, focus is placed on the suggestion following the automatically selected suggestion.
> - Otherwise, places focus on the first focusable element in the popup.

React Aria's combobox also has that as a default behavior: https://react-spectrum.adobe.com/react-aria/ComboBox.html

atomiks:
When the input isn't part of the focus loop, that is the correct behavior. Otherwise, it opens without moving focus initially, as [Ariakit](https://ariakit.org/components/combobox) or Google Search as mentioned. Maybe this is more applicable to Autocomplete not Combobox?

mnajdova:
> Maybe this is more applicable to Autocomplete not Combobox?

Yes, this makes sense 👌

michaldudak:
I'm getting linter warnings ([@typescript-eslint/no-unsafe-member-access](https://typescript-eslint.io/rules/no-unsafe-member-access)) whenever I try to access the event details parameter (on both Autocomplete and Combobox) when the package is installed via npm:

image

atomiks:
@michaldudak maybe https://stackoverflow.com/a/79386184?

michaldudak:
Never mind, it was caused by an invalid setup in my tsconfig.

docs/src/app/(public)/(content)/react/components/autocomplete/demos/async/css-modules/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+  );
+}
+
+interface Movie {
+  id: string;
+  title: string;
+  year: number;
+}
+
+const top100Movies: Movie[] = [
```
LukasTy, line 137: General nitpick: Is there a way to share this data between CSS Modules and Tailwind examples? 🤔

mnajdova:
Home & End keyboard actions are not implemented - they should highlight the first/last option respectively, see https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

atomiks:
@mnajdova for "editable" comboboxes (i.e. an input rather than a trigger button), Home/End don't move focus to an item, but instead the beginning/end of the input

> Home (Optional): Either moves focus to and selects the first option or, **if the combobox is editable, returns focus to the combobox and places the cursor on the first character.** [...]

mnajdova:
Fair enough, but the end button for me always navigates to the end of the page (at least on the hero demo).

mnajdova:
Here is a recording: https://github.com/user-attachments/assets/5af57546-ff71-4c0a-aebc-ccbc67c12234

packages/react/src/combobox/empty/ComboboxEmpty.tsx
```diff
@@ -0,0 +1,39 @@
+'use client';
+import * as React from 'react';
+import { BaseUIComponentProps } from '../../utils/types';
+import { useRenderElement } from '../../utils/useRenderElement';
+import { useComboboxDerivedItemsContext } from '../root/ComboboxRootContext';
+
+/**
+ * Renders its children only when the list is empty with
+ * screen reader support.
+ */
```
michaldudak: How about "renders its children with screen reader support only when the list is empty"?
The current version sounds like the list (is empty with screen reader support)

atomiks: Updated both Empty + Status to be more clear

docs/src/app/(public)/(content)/react/components/autocomplete/page.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+<Subtitle>An input that suggests options as you type.</Subtitle>
+<Meta
+  name="description"
+  content="A high-quality, unstyled React autocomplete component that renders an input with a list of filtered options."
+/>
+
+<Demo path="./demos/hero" />
+
+## Accessibility guidelines
```
michaldudak: "Usage guidelines"?

packages/react/src/combobox/status/ComboboxStatus.tsx
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Displays a status message whose content changes are announced politely to screen readers.
+ * Renders a `<div>` element.
+ */
+export const ComboboxStatus = React.forwardRef(function ComboboxStatus(
+  componentProps: ComboboxStatus.Props,
+  forwardedRef: React.ForwardedRef<HTMLDivElement>,
+) {
+  const { render, className, ...elementProps } = componentProps;
+
```
michaldudak, line 16: This needs a bit more explanation - what kind of status messages? Why is it different to `Empty`?

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
```diff
@@ -0,0 +1,200 @@
...
+    disabled: fieldDisabled,
+    labelId,
+    setTouched,
+    setFocused,
+    validationMode,
+  } = useFieldRootContext();
+  const {
+    store,
+    setOpen,
+    disabled: comboboxDisabled,
```
michaldudak: Why `disabled`, `readOnly`, etc. aren't in the store?

packages/react/src/filterable-menu/root/FilterableMenuRoot.tsx
michaldudak, line 1: Can we remove these files since they are out of scope of this PR?

examples/vite-css/.eslintcache
michaldudak, line 1: This looks like it shouldn't be committed.

packages/react/src/autocomplete/root/AutocompleteRoot.tsx
```diff
@@ -0,0 +1,202 @@
...
+    ...rest
+  } = props;
+
+  const enableInline = mode === 'inline' || mode === 'both';
+  const staticItems = mode === 'inline' || mode === 'none';
+
+  // Mirror the typed value for uncontrolled usage so we can compose the inline overlay.
+  const isControlled = value !== undefined;
+  const [internalValue, setInternalValue] = React.useState(defaultValue ?? '');
+  const [inlineOverlay, setInlineOverlay] = React.useState('');
```
michaldudak, line 41: What's this overlay?

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
atomiks: Since these values are props, not state, they should be simple derived values placed in the context, since otherwise they need to be synchronized with an effect to the store, which feels off. (Select also currently does this)

packages/react/src/autocomplete/root/AutocompleteRoot.tsx
atomiks: It's a temporary 'inline' overlay to be displayed with the input value while navigating with keyboard, so it autofills (`aria-autocomplete="both|inline"`)

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
michaldudak: They are props of the root component, but for everything else, it's just an external value. Having them directly in the context loses update granularity (although they shouldn't change often, so maybe it's not a big deal). Anyway, I'd be in favor of keeping the whole state in the store for consistency.

atomiks: Yeah, I figured they had low update frequency and usually the list is not mounted when they change, meaning the perf is not a concern specifically.

I agree consistency is better, though sync'ing props to state always feels wrong to me 😅 as it's considered an anti-pattern. But for the store, maybe it's fine.

packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
```diff
@@ -0,0 +1,1003 @@
...
+                  )}
+                </Autocomplete.List>
+              </Autocomplete.Popup>
+            </Autocomplete.Positioner>
+          </Autocomplete.Portal>
+        </Autocomplete.Root>,
+      );
+
+      const input = screen.getByRole<HTMLInputElement>('combobox');
+      fireEvent.click(input);
```
LukasTy, line 100: Nitpick
```suggestion
      await user.click(input);
```

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
michaldudak: Yes, syncing local props to **local** state is pointless and an antipattern. But here, by placing it in a context, we expose it anyway (plus, context really acts as state as it causes rerenders). I'd rather have one structure (store) to keep the whole state together.

packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
```diff
@@ -0,0 +1,1003 @@
...
+                    </Autocomplete.Item>
+                  )}
+                </Autocomplete.List>
+              </Autocomplete.Popup>
+            </Autocomplete.Positioner>
+          </Autocomplete.Portal>
+        </Autocomplete.Root>,
+      );
+
+      const input = screen.getByTestId('input');
```
LukasTy, line 580: WDYT about sticking with the `getByRole` instead?
Usage of `data-testid` should be reserved for cases where an accessible way to query the element doesn't work.
```suggestion
      const input = screen.getByRole('textbox');
```
P.S. This applies to the whole test file and would allow dropping the `data-testid="input"` prop.

packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
```diff
@@ -0,0 +1,1003 @@
...
+            </Autocomplete.Portal>
+          </Autocomplete.Root>
+        </Field.Root>,
+      );
+
+      const input = screen.getByTestId('input');
+
+      expect(input).not.to.have.attribute('data-focused');
+
+      fireEvent.focus(input);
```
LukasTy, line 802: Nitpick: `user.keyboard('{Tab}')` would probably work as well. 🤔

packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
```diff
@@ -0,0 +1,1003 @@
...
+      );
+
+      const input = screen.getByTestId('input');
+
+      expect(input).not.to.have.attribute('data-valid');
+      expect(input).not.to.have.attribute('data-invalid');
+
+      // Type a non-empty value and blur to trigger validation
+      fireEvent.focus(input);
+      fireEvent.change(input, { target: { value: 'ok' } });
```
LukasTy: Looks cleaner with user events, since you don't need explicit focus management. 🤔 
```suggestion
      await user.type(input, 'ok')
```

packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
```diff
@@ -0,0 +1,1003 @@
...
+
+      const input = screen.getByTestId('input');
+
+      expect(input).not.to.have.attribute('aria-invalid');
+
+      await user.type(input, 'invalid');
+
+      fireEvent.blur(input);
+
+      await flushMicrotasks();
```
LukasTy: Nitpick: Is this necessary if the assertion below is within a `waitFor`? 🤔 
I would assume that one or the other would be needed...

packages/react/src/combobox/index.parts.ts
```diff
@@ -0,0 +1,27 @@
...
+export { ComboboxItemIndicator as ItemIndicator } from './item-indicator/ComboboxItemIndicator';
+export { ComboboxChips as Chips } from './chips/ComboboxChips';
+export { ComboboxChip as Chip } from './chip/ComboboxChip';
+export { ComboboxChipRemove as ChipRemove } from './chip-remove/ComboboxChipRemove';
+export { ComboboxRow as Row } from './row/ComboboxRow';
+export { ComboboxCollection as Collection } from './collection/ComboboxCollection';
+export { ComboboxEmpty as Empty } from './empty/ComboboxEmpty';
+export { ComboboxClear as Clear } from './clear/ComboboxClear';
+
+export { Separator as ComboboxSeparator } from '../separator';
```
LukasTy: Shouldn't this be just `Separator` as with `Autocomplete`? 🤔 
```suggestion
export { Separator } from '../separator';
```

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
atomiks: Updated to use the store for all values - DX seems solid aside from the sync'ing effect

packages/react/src/combobox/chip-remove/ComboboxChipRemove.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+  const element = useRenderElement('button', componentProps, {
+    ref: [forwardedRef, buttonRef],
+    state,
+    props: [
+      {
+        tabIndex: -1,
+        disabled,
+        'aria-readonly': readOnly || undefined,
+        onClick(event) {
```
LukasTy, line 52: Nitpick: The contents of this function is very similar to the `onKeyDown` handler.
Some things could probably be extracted. 🤔

packages/react/src/combobox/chips/ComboboxChips.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const chipsContainerRef = useStore(store, selectors.chipsContainerRef);
+  const open = useStore(store, selectors.open);
+
+  const [highlightedChipIndex, setHighlightedChipIndex] = React.useState<number | undefined>(
+    undefined,
+  );
+
+  if (open && highlightedChipIndex !== undefined) {
+    setHighlightedChipIndex(undefined);
```
LukasTy, line 31: Could be worth adding a comment, because initially reading the code it seems like this might be a mistake. 🙈

packages/react/src/combobox/clear/ComboboxClear.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+                <Combobox.Item value="a">a</Combobox.Item>
+              </Combobox.List>
+            </Combobox.Popup>
+          </Combobox.Positioner>
+        </Combobox.Portal>
+      </Combobox.Root>,
+    );
+
+    const input = screen.getByTestId('input') as HTMLInputElement;
+    fireEvent.click(screen.getByTestId('clear'));
```
LukasTy:
```suggestion
    const input = screen.getByRole('textbox');
    await user.click(screen.getByTestId('clear'));
```

packages/react/src/combobox/clear/ComboboxClear.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    await flushMicrotasks();
```
LukasTy: Nitpick: is it necessary if the previous action is async? 🤔

packages/react/src/combobox/clear/ComboboxClear.tsx
LukasTy: WDYT about adding a comment to signal the intent of this event listener? 🤔

packages/react/src/combobox/clear/ComboboxClear.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  export interface Props extends BaseUIComponentProps<'button', State> {
+    /**
+     * Whether the component renders a native `<button>` element when replacing it
+     * via the `render` prop.
+     * Set to `false` if the rendered element is not a button (e.g. `<div>`).
+     * @default true
+     */
+    nativeButton?: boolean;
```
LukasTy: extends `NativeButtonProps`? 🤔

packages/react/src/combobox/empty/ComboboxEmpty.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+          </Combobox.Positioner>
+        </Combobox.Portal>
+      </Combobox.Root>,
+    );
+
+    expect(screen.getByTestId('empty')).to.have.text('No results');
+    expect(screen.getByTestId('empty')).to.have.attribute('role', 'status');
+  });
+
+  it('does not render when there are items', async () => {
```
LukasTy, line 53: WDYT, would it make sense to add another test checking that empty status is rendered when no items match the search query? 🤔

packages/react/src/combobox/item/ComboboxItem.tsx
```diff
@@ -0,0 +1,254 @@
...
+     * @default false
+     */
+    disabled?: boolean;
+    /**
+     * Whether the component renders a native `<button>` element when replacing it
+     * via the `render` prop.
+     * Set to `false` if the rendered element is not a button (e.g. `<div>`).
+     * @default false
+     */
+    nativeButton?: boolean;
```
LukasTy: extends `NonNativeButtonProps`

packages/react/src/combobox/positioner/ComboboxPositioner.tsx
```diff
@@ -0,0 +1,174 @@
...
+      )}
+      {element}
+    </ComboboxPositionerContext.Provider>
+  );
+});
+
+export namespace ComboboxPositioner {
+  export interface State {
+    /**
+     * Whether the preview card is currently open.
```
LukasTy:
```suggestion
     * Whether the combobox is currently open.
```

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  export interface Props extends BaseUIComponentProps<'button', State> {
+    /**
+     * Whether the component renders a native `<button>` element when replacing it
+     * via the `render` prop.
+     * Set to `false` if the rendered element is not a button (e.g. `<div>`).
+     * @default true
+     */
+    nativeButton?: boolean;
```
LukasTy: extends `NativeButtonProps`

packages/react/src/combobox/root/ComboboxRootInternal.tsx
```diff
@@ -0,0 +1,1267 @@
...
+
+  const [closeQuery, setCloseQuery] = React.useState<string | null>(null);
+
+  const query =
+    closeQuery ?? (inputValue === '' ? '' : String(inputValue).trim().toLocaleLowerCase());
+  const isGrouped = isGroupedItems(items);
+
+  const flatItems: Value[] = React.useMemo(() => {
+    if (!items) {
+      return [];
```
LukasTy, line 179: Wouldn't `EMPTY_ARRAY` work here?
```suggestion
      return EMPTY_ARRAY;
```

packages/react/src/combobox/root/ComboboxRootInternal.tsx
```diff
@@ -0,0 +1,1267 @@
...
+    if (isGrouped) {
+      return items.flatMap((group) => group.items);
+    }
+
+    return items;
+  }, [items, isGrouped]);
+
+  const filteredItems: Value[] | Group<Value>[] = React.useMemo(() => {
+    if (!items) {
+      return [];
```
LukasTy, line 191:
```suggestion
      return EMPTY_ARRAY;
```

packages/react/src/combobox/root/ComboboxRootInternal.tsx
atomiks: I did this before - the array can be mutated and break the entire const for every module 😬 I've refactored it a lot since, so maybe it works differently now, but still

LukasTy: Gotcha. 👌

packages/react/src/combobox/chips/ComboboxChips.tsx
atomiks: This should just be in the store now! Will refactor

packages/react/src/combobox/clear/ComboboxClear.tsx
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  export interface Props extends NativeButtonProps, BaseUIComponentProps<'button', State> {
+    /**
+     * Whether the component renders a native `<button>` element when replacing it
+     * via the `render` prop.
+     * Set to `false` if the rendered element is not a button (e.g. `<div>`).
+     * @default true
+     */
+    nativeButton?: boolean;
```
LukasTy:
```suggestion
```

packages/react/src/combobox/item/ComboboxItem.tsx
```diff
@@ -0,0 +1,256 @@
...
```
LukasTy:
```suggestion
```

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  export interface Props extends NativeButtonProps, BaseUIComponentProps<'button', State> {
+    /**
+     * Whether the component renders a native `<button>` element when replacing it
+     * via the `render` prop.
+     * Set to `false` if the rendered element is not a button (e.g. `<div>`).
+     * @default true
+     */
+    nativeButton?: boolean;
```
LukasTy:
```suggestion
```

PR #2107, opened 2025-06-12
[tabs] Fix focus going out of sync when selected value is changed externally

### LINKED ISSUE #2100: [tabs] Active item in the tablist composite widget can go out of sync when a new tab is selected with a controlled update
# Bug report

## Current behavior

When a new tab is selected while the tablist doesn't have focus (eg via a controlled update), the tablist doesn't update its internal "active" composite item to match the selected item, and instead continues to keep as the "active" item the last tab focused by the user (or the first tab).

This is particularly problematic when the `activateOnFocus` prop is `true`, since a user using the keyboard may inadvertently select a different tab just by tabbing into the tablist.

https://github.com/user-attachments/assets/a04f89c0-796a-43cb-b631-509f602d7a4e

## Expected behavior

When a new tab is selected while the tablist doesn't have focus (eg via a controlled update), the tablist should update its internal "active" composite item to match the selected item.

This will allow the user to tab back into the tablist and have the keyboard focus on the selected tab.

This behaviour should be implemented regardless of the `activateOnFocus` prop.

## Reproducible example

https://codesandbox.io/p/sandbox/gallant-grass-h5fvk6:
- Select a tab other than "account"
- Click on the "Select "Account"" button
- Using the keyboard, tab into the tablist
- Notice how the tab receiving focus is not the selected tab (account), but the tab that was last focused
- This is particularly problematic when `activateOnFocus` is `true`, since the tab receiving focus will become selected

## Base UI version

`1.0.0-beta.0`

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

N/A

### LINKED ISSUE #2101: [tabs] Tablist active composite item can go out of sync with focused tab if a new tab is selected
# Bug report

## Current behavior

When:
- keyboard focus is on a tab in the tablist
- the selected tab changes via a controlled update

The tablist composite widget "active" element is seemingly updating to the newly selected tab, and therefore goes out of sync with the keyboard focus

This causes a buggy experience for the user when pressing the arrow keys after the controlled update.

https://github.com/user-attachments/assets/b85c4ce0-e448-4745-b656-a4eb47923f06

## Expected behavior

When such a controlled update happens, the internal "active" composite item should be updated to be the same as the focused tab

## Reproducible example

https://codesandbox.io/p/sandbox/dry-firefly-fdc529

## Base UI version

`1.0.0-beta.0`

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

We currently handle this edge case in `@wordpress/components` [this way](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/tabs/index.tsx#L90-L110)

atomiks:
Fixes #2100
Fixes #2101


packages/react/src/tabs/tab/TabsTab.tsx
```diff
@@ -166,6 +167,9 @@ export const TabsTab = React.forwardRef(function Tab(
         onClick,
         onFocus,
         onPointerDown,
+        onKeyDownCapture() {
+          isNavigatingRef.current = true;
+        },
```
atomiks, line 174: This is needed when moving from a disabled tab to a non-disabled one with the new effect logic, since it stays stuck on the disabled item otherwise (`selected` remains on the non-disabled tab, but the highlighted index doesn't)

packages/react/src/tabs/tab/TabsTab.tsx
```diff
@@ -108,7 +109,7 @@ export const TabsTab = React.forwardRef(function Tab(
       return;
     }
 
-    if (index > 1 && index !== highlightedTabIndex) {
+    if (index > -1) {
```
atomiks, line 113: This seems like a bug but it didn't affect the tests

Issue #2106, opened 2025-06-12
[tabs] Inconsistency between "selected", "highlighted" and "active" tab states


ciampo:
# Bug report

## Current behavior

I find that some of the variable names used in the `Tabs` component are confusing, especially about the difference between:
- the selected tab
- the tab that is the active item in the composite tablist widget

In particular:
- when it's selected, a tab receives the `aria-selected="true"` and `data-selected` attributes
- when it's the "active" composite item, a tab receives the `data-highlighted` attribute
  - the `data-highlighted` attribute is also added when keyboard focus is not on the tablist
- consumers can use `--active-tab-*` CSS variables to style the indicator to follow the size and dimensions of the selected tab

The result is that different terms are used for overlapping concepts: `selected`, `highlighted`, `active`

## Expected behavior

Establish a clear nomenclature for composite widgets and apply consistently — for example, `selected` and `active`. The result would be the following:
- `aria-selected="true"` and `data-selected` attributes for selected tabs
- `data-active` attribute for tabs when they are the active composite item
- `--selected-tab-*` CSS variables to style the indicator

Two extra thoughts:
- I guess the `data-active` attribute (currently `data-highlighted`) should not be added when the keyboard focus is not on the tablist?
- Are `data-*` attributes needed if an equivalent `aria-*` attribute is already specified?

## Reproducible example

N/A

## Base UI version

`1.0.0-beta.0`

## Which browser are you using?

Chrome

## Which OS are you using?

Mac OS

## Which assistive tech are you using (if applicable)?

N/A

## Additional context

I assume that any changes should be applied consistently across all composite-based components.

atomiks:
I think `[data-highlighted]` should only exist when hover is synchronized with focus (as in Select/Menu). Otherwise, since Composite components don't synchronize hover and focus, and since they use roving focus, the `:focus` pseudo-selector should be enough? Meaning `Composite` components shouldn't have an attribute to indicate the current "active item" to begin with

> Are data-* attributes needed if an equivalent aria-* attribute is already specified?

I think it's better to use data attributes since aria attributes are more of an implementation detail and not a style target

ciampo:
The `data-highlighted` attribute could still be helpful to abstract over instances that use roving tabinxed vs virtual focus (ie. focus stays on the parent composite item, and the `aria-activedescendant⁠` attribute is used instead).

[...]

PR #2109, opened 2025-06-13
[composite] Remove highlighted state

atomiks:
Part of #2106

### DIFF
--- a/packages/react/src/composite/item/CompositeItem.tsx
+++ b/packages/react/src/composite/item/CompositeItem.tsx
@@ -1,7 +1,6 @@
 'use client';
 import * as React from 'react';
 import { useRenderElement } from '../../utils/useRenderElement';
-import { useCompositeRootContext } from '../root/CompositeRootContext';
 import { useCompositeItem } from './useCompositeItem';
 import type { BaseUIComponentProps } from '../../utils/types';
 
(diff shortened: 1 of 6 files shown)


mj12albert (approved):

ciampo:
Cross-posting [this comment](https://github.com/mui/base-ui/issues/2106#issuecomment-2972755306) for visibility:

> The `data-highlighted` attribute could still be helpful to abstract over instances that use roving tabinxed vs virtual focus (ie. focus stays on the parent composite item, and the `aria-activedescendant⁠` attribute is used instead).

atomiks:
@ciampo agreed but since this component currently doesn't support virtual focus, it can be removed now. We can settle properly on how the attribute works once/if we add virtual focus

ciampo:
> since this component currently doesn't support virtual focus, it can be removed now. We can settle properly on how the attribute works once/if we add virtual focus

Gotcha — one last piece of feedback that I also shared in #2106 is that **`active` could be a better name instead of `highlighted`**:

- it more closely aligns with the WAI-ARIA composite widget nomenclature (especially with the `aria-activedescendant` prop)
- since BaseUI doesn't "highlight" the element on screen (consumers are expected to pass styles), `highlighted` could be slightly confusing

Thank you for being so responsive on the feedback I shared!

PR #2110, opened 2025-06-13
[select] Fix position flicker when no default value is set

atomiks:
After the store introduction, the `alignItemWithTriggerActive` values can't be in the store since it updates too late (and isn't necessary since the items don't need it). It causes the position to flicker into place as it's not aligned for the first/initial frame. This also returns the select-state-in-render-pattern in `Select.Positioner`

Remove the `defaultValue` from https://master--base-ui.netlify.app/experiments/long-select to see the bug before and after


atomiks:
@romgrk yes it's still performant - in fact it seems slightly faster in this PR

michaldudak (approved):

PR #2111, opened 2025-06-13
[scroll area] Handle touch momentum scrolling for programmatic scroll check

atomiks:
Since touch devices lack `onWheel` the scrolling state is unset while in momentum scroll phase: https://github.com/mui/base-ui/pull/1908

### DIFF
--- a/packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
+++ b/packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
@@ -11,6 +11,7 @@ import { MIN_THUMB_SIZE } from '../constants';
 import { clamp } from '../../utils/clamp';
 import { useModernLayoutEffect } from '../../utils/useModernLayoutEffect';
 import { onVisible } from '../utils/onVisible';
+import { useTimeout } from '../../utils/useTimeout';
 
 /**
  * The actual scrollable container of the scroll area.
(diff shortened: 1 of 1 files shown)


packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
```diff
@@ -225,7 +227,16 @@ export const ScrollAreaViewport = React.forwardRef(function ScrollAreaViewport(
...
+      // Debounce the restoration of the programmatic flag so that it only
+      // flips back to `true` once scrolling has come to a rest. This ensures
+      // that momentum scrolling (where no further user-interaction events fire)
+      // is still treated as user-driven.
+      scrollEndTimeout.clear();
+      // 100 ms without scroll events ≈ scroll end
+      // https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event
+      scrollEndTimeout.start(100, () => {
+        programmaticScrollRef.current = true;
+      });
```
romgrk: `.clear()` is redundant. A `Timeout` instance can only contain one timeout at once, it clears the old one if a new one is started.

michaldudak:
What problem does it solve? Momentum scrolling can last for much more than 100ms - is it enough to treat everything after this duration as programmatic scroll?

atomiks:
@michaldudak `onScroll` fires during momentum scrolling in quick succession so it works correctly until the end

michaldudak (approved):

Issue #1752, opened 2025-04-20
[docs] Fix API table design


oliviertassinari:
## Current behavior

The new design of the Base UI API reference:

<img width="794" alt="Image" src="https://github.com/user-attachments/assets/e089b404-96d6-471c-8841-7dccc2001281" />

_https://base-ui.com/react/components/toast#provider_

seems to deliver low usability in practice. I landed here trying to use a bit the product.

As far as I'm aware, here are the possible user journeys to use the API reference, sorted by importance/frequency:

1. "I don't know the prop to use, but I want to see if an option to do what I want already exists."
  For example:
  - Let's say I'm using Notistack and I want to migrate to the new Toast component. If I'm using Material Design, I can't use stacking: https://m3.material.io/components/snackbar/guidelines.
  <img width="380" alt="Image" src="https://github.com/user-attachments/assets/05d9b3ad-0978-4bac-8e87-802ff89f6440" />

  So I searched for "maximum" to find something similar to https://notistack.com/features/basic#maximum-snackbars. No results. 

  <img width="800" alt="Image" src="https://github.com/user-attachments/assets/7c248a21-b26c-4a19-a639-65413dad35b2" />

  It's not available in Google nor in Algolia search. Still, the content exists: 

  <img width="390" alt="Image" src="https://github.com/user-attachments/assets/9ebb257a-5868-4d3c-8880-daf55500e6d7" /> 

  - Or let's say I want to implement an "auto close" behavior, same pain.

  So I don't see how this solves the design constraint that the table should solve.

2. "I see a component that use a prop, I don't know what the prop does, I follow the link from the IDE".
   For example, I'm in:

<img width="570" alt="Image" src="https://github.com/user-attachments/assets/c09dc109-b40d-4987-b3cd-cd7e8f6bcb38" />

(_I think it's missing a link to the API reference like MUI Base has, but that's a detail_). I'm opening the docs page, now I need to move my cursor all the way to the info tooltip to see what's important.

  It feels much slower compared to before.

3. "I don't know the props available, I'm curious to learn what's possible to do"

  I have to hover over each info tooltip. I don't see how this is an optimal design solution. It feels much slower compared to before.

4. ? I'm not aware of other use cases.

## Expected behavior

Right now, it feels like we have something that is 1. looks great 2. usable. I would expect the opposite priority. So: 1. it has to be usable, 2. it has to look great.

[...]

PR #2116, opened 2025-06-14
[docs] Expandable props `<table>`

mj12albert:
Preview: https://deploy-preview-2116--base-ui.netlify.app/react/components/toggle

Testing out a new design for https://github.com/mui/base-ui/issues/1752

### DIFF
--- a/docs/src/components/CollapsibleReferenceTable/CollapsibleReferenceTable.tsx
+++ b/docs/src/components/CollapsibleReferenceTable/CollapsibleReferenceTable.tsx
@@ -1,0 +1,155 @@
+import * as React from 'react';
+import clsx from 'clsx';
+import { Collapsible } from '@base-ui-components/react/collapsible';
+import { createMdxComponent } from 'docs/src/mdx/createMdxComponent';
+import { inlineMdxComponents } from 'docs/src/mdx-components';
+import { rehypeSyntaxHighlighting } from 'docs/src/syntax-highlighting';
+import * as Table from '../Table';
+import { TableCode } from '../TableCode';
+import { GhostButton } from '../GhostButton';
+
+const DATA: Record<
+  string,
+  { type: string; description?: string; default?: boolean; required?: boolean }
+> = {
+  className: {
+    type: 'string | ((state: Fieldset.Root.State) => string)',
+    description:
+      'CSS class applied to the element, or a function that\n' +
+      'returns a class based on the component’s state.',
+  },
+  render: {
+    type: 'ReactElement | ((props: HTMLProps, state: Fieldset.Root.State) => ReactElement)',
+    description:
+      'Allows you to replace the component’s HTML element\n' +
+      'with a different tag, or compose it with another component.\n' +
+      '\n' +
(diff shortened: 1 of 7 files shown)


docs/src/components/CollapsibleReferenceTable/CollapsibleReferenceTable.tsx
romgrk, line 1: I played a bit with the UX, these changes would be nice:
 - Make the whole row a click target, the small toggle button at the end is tedious to click (particularly with a touchpad)
 - Merge the `type` and `default` columns. For large type descriptions, that leaves more space for the type
 - Don't show the default if there is none (when `prop.default === undefined`)

https://github.com/user-attachments/assets/ccf36b35-7307-48da-9e90-47e91e883dc1

atomiks:
The entire row should be clickable imo, as the button is quite obscure and small

Also would be good to in this PR:

1. Amend a component's prop JSDoc with an `@example` code block to see how examples render for each row
2. Add a manual `@type` or such to show expanded types (until the extractor can handle it). So for `collisionAvoidance: CollisionAvoidance` it'd be `collisionAvoidance: { side: 'flip' | 'shift', align: 'flip' | 'shift', fallbackAxisSide: 'none' | 'start' | 'end' }` in the expanded panel

It's nice to keep the JSDocs in sync with the documentation as much as possible (and keep it as a source of truth)

mj12albert:
> The entire row should be clickable imo, as the button is quite obscure and small

Currently I started with one of the patterns from https://adrianroselli.com/2019/09/table-with-expando-rows.html

From this example it seems possible to have clickable rows as well: https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html#Update02

Just need to test if combining both maintains a good SR experience; in the [“Disclosure Control as Entire Row” example](https://codepen.io/aardrian/pen/VoQbLm?editors=1100), the trigger row is special and doesn't have "content cells" matching the headings

atomiks:
We really need a footer 😅 

https://github.com/user-attachments/assets/82f346a3-2277-4d44-be4b-8ca85e7d5f55

mj12albert:
A huge issue I found (also present in the [reference example](https://codepen.io/aardrian/pen/wvKWVBW?editors=1100)) is that when VO moves to an expanded details row, the row number of the new row is always announced incorrectly, also it always announces the row header contents of a different row

Seems like these are bugs that Adrian already filed in the webkit tracker:
- [Tables with show/hide rows report wrong counts and block access to rows in VoiceOver](https://bugs.webkit.org/show_bug.cgi?id=275366) - supposedly fixed in MacOS 15.1, which I don't have
- [Incorrect row header announcement in VoiceOver macOS](https://bugs.webkit.org/show_bug.cgi?id=235740) - opened in 2022, no response/activity

mj12albert:
Screenshot 2025-06-19 at 6 09 38 PM

Using `rowspan` fixes 90% of the a11y issue but makes the design impossible, since overriding the `rowspan`-based height requires changing `display: table-cell` which breaks the whole table's semantics

Closing this as the accordion based implementations are better

PR #2119, opened 2025-06-16
[dialog] Fix popup prop merging

atomiks:
With #2042, merging with main I noticed two issues:

- `nativeButton={false}` must be specified on `Dialog.Trigger` since the `Menu.Item` is rendering it, which is not a native button (the keyboard handlers didn't work, only the pointer).  I wonder if this can be made more obvious/clear.
- The prop merging was incorrect in `useDialogPopup`, causing the `onKeyDown` handler to overwrite and break the handlers that handled the `Escape` event key bubbling

### DIFF
--- a/docs/src/app/(private)/experiments/menu/complex-nesting.tsx
+++ b/docs/src/app/(private)/experiments/menu/complex-nesting.tsx
@@ -48,9 +48,10 @@ export default function MenuComplexNestingExperiment() {
 
                 <Dialog.Root>
                   <Menu.Item
-                    render={<Dialog.Trigger nativeButton={false} />}
+                    render={<Dialog.Trigger />}
                     className={styles.Item}
                     closeOnClick={false}
+                    nativeButton
                   >
                     Open Settings
                   </Menu.Item>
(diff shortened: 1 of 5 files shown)


docs/src/app/(private)/experiments/menu/complex-nesting.tsx
```diff
@@ -48,7 +48,7 @@ export default function MenuComplexNestingExperiment() {
 
                 <Dialog.Root>
                   <Menu.Item
-                    render={<Dialog.Trigger />}
+                    render={<Dialog.Trigger nativeButton={false} />}
```
michaldudak: This doesn't seem correct. Dialog.Trigger renders a native button. If there was `<Dialog.Trigger render={<Menu.Item />}>` we'd need `nativeButton={false}`.

michaldudak: Actually, it should be `<Menu.Item render={<Dialog.Trigger />} nativeButton />`, no?

atomiks: Yes, right - imo this area needs work in the API as it's pretty confusing and an easy footgun

michaldudak: It could be alleviated by us using native buttons everywhere, including menu items. Then, only if devs want to use noninteractive elements, they'll have to explicitly add the prop.

PR #2123, opened 2025-06-17
[useScrollLock] Fix height calculation on Android

### LINKED ISSUE #2122: [dialog] Incorrect scroll-lock style calculation
# Bug report

Currently when opening a modal in Chrome on Android the bottom part of the screen is being cut-off due to an incorrect calculation of the `height` property. This is only happening if the top navigation / URL bar of the browser is hidden because the user scrolled down.

## Current behavior

https://github.com/user-attachments/assets/0d6506d0-23ee-4b83-94df-76d51fd38ef5

In the video notice the bottom part of the page, its being cut-off when a dialog is open. The reason for that is the `height: calc(-57px + 100dvh)` style on the `body element.`

## Expected behavior

The height and width of the body element should remain visually unchanged.

## Reproducible example
[
calc(-57px + 100dvh)](https://base-ui.com/react/components/dialog)

## Base UI version

1.0.0-beta.0

Chrome on Android

## Which OS are you using?

Android

## Additional context

Taking a look at the code this bug seems to be related to this line: https://github.com/mui/base-ui/blob/d07961b40bd474f26db632421c5e346a29eabaef/packages/react/src/utils/useScrollLock.ts#L87 where the scrollbarHeight is being calculated as:

```ts
const scrollbarHeight = Math.max(0, win.innerHeight - html.clientHeight);
```

atomiks:
Fixes #2122


packages/react/src/utils/useScrollLock.ts
```diff
@@ -83,8 +83,8 @@ function preventScrollStandard(referenceElement: Element | null) {
       htmlStyles.overflowX === 'scroll' || bodyStyles.overflowX === 'scroll';
 
     // Values can be negative in Firefox
-    const scrollbarWidth = Math.max(0, win.innerWidth - html.clientWidth);
-    const scrollbarHeight = Math.max(0, win.innerHeight - html.clientHeight);
+    const scrollbarWidth = isScrollableY ? Math.max(0, win.innerWidth - html.clientWidth) : 0;
+    const scrollbarHeight = isScrollableX ? Math.max(0, win.innerHeight - html.clientHeight) : 0;
```
atomiks: Caveat: this check doesn't work if the content is horizontally scrollable (`html.scrollWidth > html.clientWidth`) but there's still no vertical scrollbar size

packages/react/src/utils/useScrollLock.ts
```diff
@@ -191,7 +191,7 @@ class ScrollLocker {
       return;
     }
 
-    const isOverflowHiddenLock = isIOS || (isFirefox && !hasInsetScrollbars(referenceElement));
+    const isOverflowHiddenLock = isIOS || !hasInsetScrollbars(referenceElement);
```
atomiks, line 194: This applies the `hasInsetScrollbars` check (previously Firefox-only) to all browsers, which checks `win.innerWidth - doc.documentElement.clientWidth` to determine if it can use the standard scroll lock (where no layout shift occurs due to disappearing scrollbars), which is unaffected by mobile browser navbars

mnajdova: We should update the comment below, as it is not related to Firefox only now.

mnajdova (approved):
Please check the comment above before merging, the fix works 👍

PR #2124, opened 2025-06-17
[accordion][tabs] Make `value` a required prop

### LINKED ISSUE #1588: [Accordion] Opening animation isn't cancelled when a panel is initially open
https://codesandbox.io/p/sandbox/eloquent-gwen-dpp375

Tested with transitions but likely to happen with animations as well, the underlying collapsible's open state can't be determined before the first render when all the panels are indexed and the accordion's value prop `['value']` can be resolved to the panel's initial states

### LINKED ISSUE #1872: Tabs.Panel: keepMounted doesn't work in

### LINKED ISSUE #1880: [Accordion][Tabs] Consider making `value` a required prop
Currently it's possible to omit the `value` prop in accordion and tabs entirely, using the index (by position) implicitly as the value instead.

This causes a few small issues:

- Unable to "truly" unmount a TabPanel when `keepMounted={false}` https://github.com/mui/base-ui/issues/1872
- Mis-timed accordion animations https://github.com/mui/base-ui/issues/1588
- Unable to SSR initially open accordion or initially active tab

### LINKED ISSUE #2096: [tabs] `aria-controls` not applied to the first tab if no `value` is set
# Bug report

## Current behavior

When the first tab (and respective tabpanel) are not given ane explicit `value`, the first tab doesn't not get the `aria-controls` attribute — even though clicking on the tab correctly shows the associated tabpanel.

## Expected behavior

Each tab should have the `aria-controls` attribute, referencing the ID of the associated tabpanel.

## Reproducible example

[Link to a CodeSandbox, StackBlitz, or other IDE.](https://codesandbox.io/p/sandbox/suspicious-noether-czy2zs)

## Base UI version

`1.0.0-beta.0`

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

N/A

michaldudak:
Made `value` in Accordion.Item, Tabs.Tab, and Tabs.Panel a mandatory prop. This helps to solve several issues with these components:
- closes #1880
- closes #1872
- closes #1588
- closes #2096

### DIFF
--- a/packages/react/src/tabs/tab/TabsTab.tsx
+++ b/packages/react/src/tabs/tab/TabsTab.tsx
@@ -27,27 +27,20 @@ export const TabsTab = React.forwardRef(function Tab(
     className,
     disabled = false,
     render,
-    value: valueProp,
+    value,
     id: idProp,
     nativeButton = true,
     ...elementProps
   } = componentProps;
 
-  const {
-    value: selectedTabValue,
-    getTabPanelIdByTabValueOrIndex,
-    orientation,
-  } = useTabsRootContext();
+  const { value: selectedTabValue, getTabPanelIdByTabValue, orientation } = useTabsRootContext();
 
   const { activateOnFocus, highlightedTabIndex, onTabActivation, setHighlightedTabIndex } =
     useTabsListContext();
 
   const id = useBaseUiId(idProp);
 
-  const tabMetadata = React.useMemo(
-    () => ({ disabled, id, value: valueProp }),
-    [disabled, id, valueProp],
-  );
+  const tabMetadata = React.useMemo(() => ({ disabled, id, value }), [disabled, id, value]);
 
   const {
     props: compositeItemProps,
@@ -57,17 +50,11 @@ export const TabsTab = React.forwardRef(function Tab(
     // because the index is needed for Tab internals
   } = useCompositeItem<TabsTab.Metadata>({ metadata: tabMetadata });
 
-  const tabValue = valueProp ?? index;
-
   // the `selected` state isn't set on the server (it relies on effects to be calculated),
   // so we fall back to checking the `value` param with the selectedTabValue from the TabsContext
   const selected = React.useMemo(() => {
-    if (valueProp === undefined) {
-      return index < 0 ? false : index === selectedTabValue;
-    }
-
-    return valueProp === selectedTabValue;
-  }, [index, selectedTabValue, valueProp]);
+    return value === selectedTabValue;
+  }, [selectedTabValue, value]);
 
   const isSelectionSyncedWithHighlightRef = React.useRef(false);
 
@@ -87,9 +74,7 @@ export const TabsTab = React.forwardRef(function Tab(
     focusableWhenDisabled: true,
   });
 
-  // const handleRef = useForkRef(compositeItemRef, buttonRef, externalRef);
-
-  const tabPanelId = index > -1 ? getTabPanelIdByTabValueOrIndex(valueProp, index) : undefined;
+  const tabPanelId = index > -1 ? getTabPanelIdByTabValue(value) : undefined;
 
   const isPressingRef = React.useRef(false);
   const isMainButtonRef = React.useRef(false);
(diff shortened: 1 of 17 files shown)


atomiks:
Would be good to add tests for each of those cases

michaldudak:
Updated.

packages/react/src/tabs/tab/TabsTab.tsx
```diff
@@ -57,17 +50,11 @@ export const TabsTab = React.forwardRef(function Tab(
...
   // so we fall back to checking the `value` param with the selectedTabValue from the TabsContext
   const selected = React.useMemo(() => {
-    if (valueProp === undefined) {
-      return index < 0 ? false : index === selectedTabValue;
-    }
-
-    return valueProp === selectedTabValue;
-  }, [index, selectedTabValue, valueProp]);
+    return value === selectedTabValue;
+  }, [selectedTabValue, value]);
```
atomiks: Not sure, but I wonder if this makes the `isNavigatingRef.current` redundant? It fixes an issue with disabled tabs specifically due to how this value got calculated

michaldudak: Yup, it seems to be working without this ref.

atomiks: This doesn't need the `useMemo` wrapper anymore

atomiks:
`NavigationMenu` uses `useId()` inside of its `Item` component as a fallback when `value` is missing. I wonder if something similar could be used without requiring `value`

https://github.com/mui/base-ui/blob/8ba89eadb183833239310328ccf98af2953e17d5/packages/react/src/navigation-menu/item/NavigationMenuItem.tsx#L20-L21

michaldudak:
`useBaseUiId` can be undefined before effects run (in React 17) and we need to have this value defined during SSR to get the tab panel content in the prerendered response.
NavigationMenu doesn't need to be rendered on the server, so `useBaseUiId` is sufficient.

atomiks:
> useBaseUiId can be undefined before effects run (in React 17) and we need to have this value defined during SSR to get the tab panel content in the prerendered response.

So it's only a problem on React 17? I don't like that we need to change to a worse API to support a version of React that's fast becoming irrelevant; it's not future-looking. I'd rather change the type to be optional and just throw a runtime warning for React 17.

mj12albert:
With useId is it possible to "fully" unmount the TabPanel when keepMounted={false}? https://github.com/mui/base-ui/issues/1872

michaldudak:
It's not just that. The value must be possible to be figured out by developer so it can be used in `defaultValue` or `value` props. There's no way to do it with `useId` - the only sensible options are explicit `value` or index.

This approach is possible in NavigationMenu as `defaultValue` will often be null. In Tabs it's not common to see no tabs selected.
Even in NavigationMenu you should provide explicit values if you want to control its state from the outside, right?

atomiks:
>  It's not just that. The value must be possible to be figured out by developer so it can be used in defaultValue or value props. There's no way to do it with useId - the only sensible options are explicit value or index.

Yeah but imo it seems intuitive/expected that the `value` prop becomes required when you need to control it or select an initial tab. For Accordion this would also be less common than Tabs I imagine.

> This approach is possible in NavigationMenu as defaultValue will often be null. In Tabs it's not common to see no tabs selected. Even in NavigationMenu you should provide explicit values if you want to control its state from the outside, right?

Yes, though it seems like with Tabs defaulting to the first tab is ok in many cases and you don't need to control it?

ciampo:
I think that defaulting to the first (non disabled) tab being selected feels reasonable, if no `value` (or `defaultValue`) are passed.

michaldudak:
Closed in #3373

PR #2132, opened 2025-06-18
[utils] Make Romain's utils public

flaviendelangle:
I added some of the utils that we are using on the Scheduler to the public utils list.

For `useOnMount`, `useLazyRef` and `useTimeout`, I don't see a big reason not to make them public since you already expose stuff like `useForkRef` or `useModernLayoutEffect`.

For the store, it's clearly a bigger addition to the public API.
I'm strongly in favor of having a single implementation of this logic in the company in the future, but maybe we don't want it to be exported like this from the `@base-ui-components/react` package.

Sidenote, we are also using `useRenderElement` and `useButton` but I'm keeping those for later since they are more tight to the Base UI logic and less generic.


packages/react/src/utils/index.ts
romgrk, line 1: > but maybe we don't want it to be exported like this from the @base-ui-components/react package.

We don't, that makes those hooks subject to semver. If we need to refactor them, we can't until BUI releases another major. I want to publish a hooks/utils package separately from BUI, but I haven't had time to do it.

atomiks:
We shouldn't have _any_ public utils (it should be removed from the exports map.) 

I think some were made public just for the docs, but we should have a `utils` package instead.

If we make a util public, it's documented for public use (`useRender` and `mergeProps`). So `useControlled`, `useForkRef` etc. should all be hidden.

michaldudak:
I'm going to create a separate internal utils package to use across MUI products

michaldudak:
I opened a PR that moves the utils to a new library (https://github.com/mui/base-ui/pull/2167), so we can close this one.

PR #2134, opened 2025-06-18
[useRender] Make useRender RSC-friendly

michaldudak:
Conditionally calls `useForkRef` depending on whether the code runs on a server.
This makes the `useRender` compatible with React Server Components.
While it theoretically breaks the rules of hooks, the condition is constant per environment, so the it should work fine.


packages/react/src/utils/useRenderElement.tsx
```diff
@@ -82,13 +82,17 @@ function useRenderElementProps<
...
+  // This also skips the `useForkRef` call on the server, which is fine because
+  // refs are not used on the server side.
   /* eslint-disable react-hooks/rules-of-hooks */
-  if (!enabled) {
-    useForkRef(null, null);
-  } else if (Array.isArray(ref)) {
-    outProps.ref = useForkRefN(outProps.ref, getChildRef(renderProp), ...ref);
-  } else {
-    outProps.ref = useForkRef(outProps.ref, getChildRef(renderProp), ref);
+  if (typeof window !== 'undefined') {
```
atomiks: `document` is used by the modern layout effect util and is more robust I believe

atomiks (approved):

codeplott:
Is there a plan for the next release yet? 🤞
Really great work! As a longtime Radix UI user, I'm already loving this project 💪 
Keep it up!

atomiks:
@sawden we set up `pkg.pr.new` canary releases for when you need a release before it's official: 

https://github.com/mui/base-ui/pull/2134#issuecomment-2983965389

You can also use the latest master commit hash instead of the PR number (which don't have all the latest changes synchronized)

---

Official release will be either this week or early next week

codeplott:
@atomiks Thanks a lot! That helps a ton!

PR #2120, opened 2025-06-17
[docs] Accordion props reference

mj12albert:
Preview: https://deploy-preview-2120--base-ui.netlify.app/react/components/toggle

Testing an Accordion implementation for https://github.com/mui/base-ui/issues/1752

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #2135, opened 2025-06-18
[docs] New props table

mj12albert:
Preview: https://deploy-preview-2135--base-ui.netlify.app/react/components/toggle

Uses native `<details>`+`<summary>` for the new props table design; also applies to CSS vars and data attrs tables where the popover is replaced with a `<details>`.

Notable differences from [Accordion](https://github.com/mui/base-ui/pull/2120):
- VO reads the trigger as "_disclosure triangle, group_" in Chrome, "_summary, group_" in Firefox and Safari; overall it's a bit shorter because it doesn't have a heading element adding to the announcement
Screenshot 2025-06-18 at 11 25 40 PM
- Can't arrow between triggers like Accordion but is lightly reimplemented here
- It's `hidden="until-found"` by default when supported, no need to fight React 👍 

Closes https://github.com/mui/base-ui/issues/1752


docs/src/components/Accordion.tsx
```diff
@@ -0,0 +1,28 @@
+'use client';
+import * as React from 'react';
+import { Accordion } from '@base-ui-components/react/accordion';
+import clsx from 'clsx';
+
+export function Root({ className, ...props }: Accordion.Root.Props) {
+  return <Accordion.Root className={clsx('AccordionRoot', className)} {...props} />;
+}
```
romgrk: It's the docs so it matters less, but you should take the habit of avoiding destructuring if it's not necessary. You don't need to destructure the props parameter, you can use it as-is.

docs/src/components/ReferenceAccordion/PropsReferenceAccordion.tsx
```diff
@@ -0,0 +1,250 @@
...
+    return type;
+  }
+
+  if (['children'].includes(name)) {
+    return type;
+  }
+
+  if (['boolean', 'string', 'number'].includes(type)) {
+    return type;
+  }
```
romgrk: Same as above. The `[...].includes(...)` syntax is bad for performance

docs/src/components/Accordion.tsx
mj12albert: Copy pasta 🙈 fixed in the other PR, this Accordion is actually not needed in this one

docs/src/components/ReferenceAccordion/PropsReferenceAccordion.tsx
mj12albert: For now this is just a throwaway ~
But should we generate the short prop type here or via the api extractor? @michaldudak

michaldudak: If API extractor returns an IntrinsicNode, we should use it as is.

If the returned type has both the name an either types (like a union or intersection) or properties (object type), use the name as a short type and drill into the details for the expanded view.

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
romgrk, line 1: I wonder if we should avoid repeating the `default` value in the expanded details. It adds noise if we repeat information, and noise is not great for UX.

And I wonder if we shouldn't mark the `Type` in the expanded details as something like `Type (detailed)`, or inversely mark the column header as `Type (simplified)`, to convey that there's possibly more information in the expanded details.

image

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
```diff
@@ -0,0 +1,188 @@
...
+          <Accordion.Item>
+            <Accordion.Trigger
+              index={index}
+              aria-label={`prop: ${name}, type: ${shortPropTypeName} ${prop.default !== undefined ? `(default: ${prop.default})` : ''}`}
+              className="max-xs:gap-4 md:gap-4"
+            >
+              <TableCode className="text-navy">{name}</TableCode>
+              {prop.type && (
+                <span className="flex items-baseline gap-4 text-sm">
+                  <ShortPropType />
```
romgrk: I'm on the fence about hiding the detailed type in the details in some cases. For `render` & `className` it makes sense to always hide it, but for other cases (like below) I think it might degrade the UX a bit to have to click to see the full type. I wonder if we could try to hide it based on its length.

image

docs/src/components/DescriptionList.css
romgrk, line 1: I played a bit with the styling, sharing in case it inspires you to change things a bit. Imho the current box-shadow is too intense.

| Before | After |
| --- | --- |
| Screenshot From 2025-06-24 17-06-05 | Screenshot From 2025-06-24 17-06-52 |

mj12albert:
Needs https://github.com/mui/base-ui/pull/2104 to generate the json in the new format

docs/src/components/DescriptionList.css
mj12albert: CC @colmtuite , this continues from https://github.com/mui/base-ui/pull/2120#issuecomment-2978916211

I toned down the shadow, but there's still the issue of the open panel background being very similar to the hover state of the trigger

atomiks:
Seems like it shouldn't open while trying to select text. (The Accordion example made this impossible, but that's not accessible)

https://github.com/user-attachments/assets/a4660d49-d49f-45ad-a06f-79231a4bf907

mj12albert:
> Seems like it shouldn't open while trying to select text

When clicking the text selection, should it cancel the text selection only or toggle the element as well? 

https://github.com/user-attachments/assets/b0f5dadc-c22a-49eb-9206-6fbaa538b40c

Or should we use `user-select:none` for consistency, in the docs we seem to do this for all interactive elements

atomiks:
The main issue with disabling text selection is that you can't copy the name of the prop when trying to select it. Double clicking also allows the entire text to be selected. Maybe trying to click on text shouldn't open it, but everywhere else is fine

romgrk:
> Maybe trying to click on text shouldn't open it, but everywhere else is fine

That doesn't sound like a good UX. The text is right in the middle of the element, it makes for a weird click target if the text doesn't also open the element.

colmtuite:
@mj12albert Some suggestions:

1. Remove the gray box styling from the prop name `<code>`, and don't prevent the click. Revert it to how it was before.
2. Add a "Name" `dt` to the panel.
3. Remove the `:active` style from the `<summary>`.
4. Make the `<dt>` 400 weight.
5. Bump up the lightness of the `<summary>` `:hover` in dark mode. Light mode is good.
6. `<ul>` inside the panel should not have bottom margin.
7. Reduce the `gap` on the (default) thing to `gap-2`.
8. Rows are much larger now than on prod. We need to get them back to 37.5px tall (I assume this is the new `<code>` styling causing this)
9. Afaik, we have no control over the obnoxious orange highlight when using hidden until found?

michaldudak:
> Needs https://github.com/mui/base-ui/pull/2104 to generate the json in the new format

@mj12albert FYI, this is now merged.

mj12albert:
> 6\. `<ul>` inside the panel should not have bottom margin.

There's no margins on the list element, did you mean this padding?
Screenshot 2025-07-08 at 6 45 50 PM

> Afaik, we have no control over the obnoxious orange highlight when using hidden until found?

Not at the moment, the closest is a `::search-text` pseudo-element that's experimental in Chrome: https://github.com/Igalia/explainers/blob/main/css/find-in-page/README.md

Fixed your other comments as well @colmtuite

mj12albert:
> > Needs #2104 to generate the json in the new format
> 
> @mj12albert FYI, this is now merged.

Thanks I think we should do the new format and changes to the formatter in a separate PR actually ~

atomiks (approved):
Looks/feels really nice to me. Implementation also looks good

colmtuite:
@mj12albert the padding yes

michaldudak:
WDYT about displaying expanded type information in a tooltip on hover (so when I hover over `function` I get the full signature in a tooltip)? This would be an additional improvement for mouse users and other will still be able to expand the details.

michaldudak:
It seems the opening and closing parentheses are displayed in a different font and color
image

colmtuite:
> WDYT about displaying expanded type information in a tooltip

Could work yep. Set `hoverable` to `false`, so it doesn't block clicks on triggers above.

Do you think there's value in adding a "Copy" icon button beside some of the info inside the expanded panel?

@michaldudak @mj12albert

mj12albert:
> Do you think there's value in adding a "Copy" icon button beside some of the info inside the expanded panel?

I think if the panel can accommodate one more clickable/interactive thing, the most useful thing could be a link to the individual prop, like the Material UI docs have: https://mui.com/material-ui/api/checkbox/#checkbox-prop-defaultChecked

For users, I don't think it's that useful to be able to copy entire blocks of docs content (like description) that isn't code

colmtuite:
@mj12albert not the description no, but the prop name and type definition? Would it be useful to be able to copy those? Linking to the prop name could be useful yep. But ok let's ignore these ideas for now until we fix up the other parts.

**Additional feedback:**
- The `dt` and `dd` have different top padding; 6px and 8px respectively. Currently, their contents are misaligned.
- The tooltip arrow appears later than the tooltip popup.
- The tooltip needs `hoverable` set to `false`, so it doesn't block clicks on triggers above.
- We need consistent spacing between text in the triggers and in the panel (as discussed elsewhere)

atomiks:
> The tooltip arrow appears later than the tooltip popup.

This is due to `overflow: auto` on the `Popup`, and the transform style temporarily creates a containing block, which causes it to be clipped off. The infotips had this style but the tooltip shouldn't need it. I also think the delay should be more than 100ms (maybe 300) as it seems to pop up a bit too easily.

colmtuite:
@mj12albert Much better, thanks. A couple things left in this round:

1. The props still jump a tiny bit when the panel expands. `.AccordionContent` needs 1px padding on both top and bottom.
2. Code blocks inside panels are misaligned. Those margin props need to be changed to `mt-1` and `mb-1`.
3. If it's easy to tweak the tooltip without affecting other tooltips/infotips on the docs, it would be nice to reduce the padding inside the tooltip popup a tiny amount, just down to the next step in the scale.

I'll send more feedback soon on visual styling (shadows etc.) and also the mobile styles, they need some tweaking.

michaldudak:
Literal unions such as AccordionRoot's `orientation` should be expandable like function definitions

atomiks:
It's probably a good idea to make the tooltip have a wider max-width to accomodate most function definition lengths (without them spilling onto a new line and becoming less readable)

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
colmtuite: if reducing noise is the goal, then I would remove the default from the summary and only have it in the panel.

In the screenshot below, the visual noise is coming from the repetition of the word "default".

Screenshot 2025-07-22 at 07 09 50

So I see 3 options:

1. Remove the default from the summary, and just have it in the expanded panel. Of course, the trade-off is that you can't see the default at a glance.
2. Add a new column for default.
3. Just leave it as-is.

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
colmtuite: The full type is available in a tooltip

Screenshot 2025-07-22 at 07 19 28

colmtuite:
@mj12albert I mentioned setting `scroll-margin-top` to `0` on the meeting, but didn't understand the reasoning or outcome. Isn't this better? 

https://github.com/user-attachments/assets/edc8cc8b-e48b-453f-b394-edaee54ce21e

mj12albert:
> @mj12albert I mentioned setting `scroll-margin-top` to `0` on the meeting, but didn't understand the reasoning or outcome. Isn't this better?

I updated it to 0, what about in the mobile layout do you want the scroll margin to make the trigger flush against the header? @colmtuite 

Screenshot 2025-07-22 at 10 42 01 PM

colmtuite:
@mj12albert yea, looks nice 👌

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
romgrk: 1. Sounds like a worst UX, there's less information available without clicking. When I refer to a props table, it's nice to be able to keep it open as a reference while I write code and see everything at a glance.
2. That's the way it was before, and it left less place for the type description, and in some cases it was nearly all empty. Tbh I suggested merging type & default in a single column because I was hoping to have more space to show the type, but now that long function types are always simplified to `function` and hidden, it's less relevant, but imho hidding the function types isn't necessary.
3. Doesn't feel too bad. We could also find a terser notation, e.g. name the column `Type [Default]` and use values like `boolean [false]`. Or integrate the default to the name column, e.g. `defaultChecked = false`, but we have a lot of unused space in the type column.
  edit: Or `boolean [=false]`, somewhat similar to jsdoc's `[name=default]` notation.

Sidenote but there's still `default: undefined` entries, we should not display the default if there is none or if it's `undefined`.

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
romgrk: Nice for mouse users, but keyboard-centric users (which are probably higher in proportion among devs) don't have that, and requiring a hover is not that far from requiring a click. I still think we should use the space if we have the space.

mj12albert:
For aligning the mobile layout of the various tables, the triggers all use `display: flex` by default (the base `Accordion` styles) unless there's 3 or more "cells" then it'll change to `grid`

Also added the overscroll graphic to the trigger as it could overflow in some cases:

https://github.com/user-attachments/assets/0870dd99-f812-4807-b9f9-ef4b40bdc77c

Any further feedback or is it good to go? @colmtuite

PR #2140, opened 2025-06-20
[docs] Tweak experiments settings panel

mj12albert:
Screenshot 2025-06-20 at 5 22 34 PM

- Usually switches are on the right of the label (unlike checkbox/radio which is to the left of the label) e.g.
Screenshot 2025-06-20 at 2 33 59 PM

- Fixed the `Select.Value` not displaying the initial value


docs/src/components/Experiments/Select.tsx
```diff
@@ -9,7 +9,9 @@ export function Select(props: Select.Props) {
   return (
     <BaseSelect.Root value={value} onValueChange={onChange}>
       <BaseSelect.Trigger {...otherProps} className={clsx(classes.Select, otherProps.className)}>
-        <BaseSelect.Value placeholder="Select an option" />
+        <BaseSelect.Value placeholder="Select an option">
```
michaldudak, line 12: We can drop the placeholder as per our discussion about their sensibility.

mj12albert: @michaldudak It's still required in master 😅

michaldudak: ahh, right 😅

michaldudak (approved):

PR #2141, opened 2025-06-20
[select, menu, popover] Cut out internal backdrop to allow interacting with triggers

### LINKED ISSUE #1965: [menu] Trigger `onClick` is not called
# Bug report

## Current behavior

```tsx
      <Menu.Trigger
        className={styles.Button}
        onClick={() => {
          console.log("clicked trigger");
        }}
      >
        Song <ChevronDownIcon className={styles.ButtonIcon} />
      </Menu.Trigger>
```

`onClick` is never called.

## Expected behavior

`onClick` is called.

## Reproducible example

https://codesandbox.io/p/sandbox/vigorous-tree-nl9y5q?file=%2Fsrc%2FApp.tsx%3A17%2C46

## Base UI version

1.0.0-alpha.8

Chrome

## Which OS are you using?

Mac OS

Voiceover/JAWS etc.

## Additional context

michaldudak:
This brings the backdrop cutout functionality introduced for the Menubar to Menu, Select and Popover.
It cuts a hole in the internal backdrop (present when the popup is modal) so that interacting with the trigger is possible. This enables hover states on the trigger and makes `onClick` handlers work properly.

Fixes #1965


atomiks:
Click-to-toggle closed behavior is somewhat broken here: it should close on mousedown not mouseup still. Sometimes when tapping with the trackpad (no force click) it also fails to toggle closed.

michaldudak:
Both issues should be fixed now.

packages/react/src/popover/root/PopoverRoot.tsx
```diff
@@ -153,19 +161,23 @@ function PopoverRootComponent({ props }: { props: PopoverRoot.Props }) {
       close: closeDelay,
     },
   });
-  const click = useClick(floatingContext, { stickIfOpen });
+  const click = useClick(floatingContext, {
+    event: open ? 'mousedown' : 'click',
+    stickIfOpen,
+  });
```
atomiks: Popover closed on click when pressing the trigger but not outside before

michaldudak: Fixed. It'll always close on trigger click now, which is a change from the current behavior for modal popovers (since they make triggers inert), but it makes the experience consistent for both modal and nonmodal popovers.

atomiks (approved):

PR #2145, opened 2025-06-22
[select] Fix test errors

atomiks:

### DIFF
--- a/packages/react/src/utils/useTransitionStatus.ts
+++ b/packages/react/src/utils/useTransitionStatus.ts
@@ -1,5 +1,6 @@
 'use client';
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 import { useModernLayoutEffect } from './useModernLayoutEffect';
 import { AnimationFrame } from './useAnimationFrame';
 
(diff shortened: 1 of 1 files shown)


atomiks:
It looks like `process.env.NODE_ENV === 'test'` is already being used in the codebase, I assume this check is fine

packages/react/src/utils/useTransitionStatus.ts
```diff
@@ -48,6 +48,11 @@ export function useTransitionStatus(
   }, [open, mounted, transitionStatus, deferEndingState]);
 
   useModernLayoutEffect(() => {
+    // Avoid `act` warnings due to `store` apply updates
+    if (process.env.NODE_ENV === 'test') {
```
michaldudak: I'm not keen on introducing changes just for tests. Can you convert the failing tests to use Vitest browser mode utils (as in https://github.com/mui/base-ui/blob/c9e5c885916055c35139f2afd11f6be9213730bf/packages/react/src/menubar/Menubar.test.tsx#L97-L113)? It doesn't use `act`, so it shouldn't be a problem.

atomiks: Consumers will test it in the same environment but they'll also get these warnings, that's the main reason for this. 

This particular `transitionStatus` test doesn't seem necessary in tests in most cases unless you're specifically testing transitions in a real browser context 🤔

michaldudak: In some cases we are testing transitions, like in https://github.com/mui/base-ui/blob/93bb63f9bf80ca6608f3129c1686f00a90191579/packages/react/src/dialog/root/DialogRoot.test.tsx#L291-L328 or recently in https://github.com/mui/base-ui/pull/2124/files#diff-69be453ac22a3e59e15071ab0d2b90ecd3d255e7a87e2fda3c8b63d822559d3d

atomiks: Yeah these pass since this only affects the `data-starting-style` one (it fails to get removed in testing environments with this change)

atomiks:
@michaldudak it looks like `flushSync` also avoids it

PR #2150, opened 2025-06-23
[docs] Fix incorrect links in TOC on releases page

mj12albert:


docs/src/components/QuickNav/rehypeChangelog.mjs
```diff
@@ -0,0 +1,38 @@
...
+    visit(tree, 'element', (node, _, parent) => {
+      if (headingRank(node) === 1 && toString(node) !== 'Releases') {
+        return EXIT;
+      }
+
+      if (headingRank(node) === 3) {
+        let index = parent.children.indexOf(node);
+
+        // eslint-disable-next-line no-plusplus
+        while (index--) {
```
mj12albert, line 25: This is how the official util does it: https://github.com/syntax-tree/unist-util-find-before
Though I didn't think we need to add it as a dependency just for this

docs/src/components/QuickNav/rehypeSlug.mjs
```diff
@@ -67,3 +71,7 @@ function stringToUrl(string) {
     .replace(/^-|-$/gi, '')
     .toLowerCase();
 }
+
+// https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+const SEMVER_PATTERN =
+  /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/gm;
```
michaldudak: Let's define it just once.

mj12albert: oops, deleted it as it was moved to the other file

PR #2153, opened 2025-06-24
[select] Pass `value` as state

### LINKED ISSUE #2010: [Select] add "value" and other state to SelectValue.State
# Feature request

## Summary

In the `Select.Value` component, I would like a way to change my `className` or `render` values based on the current `value` of the component. Now, the only way to do that would be to make it a controlled component because `SelectValue.State` has nothing in it.

The concept of passing state to class/render callbacks is unique to Base UI as far as I know. `SliderRoot.State` and `NumberFieldRoot.State` are examples of components that have `value` in the state that's provided to callbacks. 

## Motivation

We would like to have the text of our `Value` component be a "dim" color if it is displaying placeholder text, but a normal color when a value is selected.

I also expect there will be a need for the `Value` to display more than just the text content of the selected `Item` (for example, showing an icon)

atomiks:
Closes #2010

### DIFF
--- a/packages/react/src/select/value/SelectValue.tsx
+++ b/packages/react/src/select/value/SelectValue.tsx
@@ -5,6 +5,11 @@ import type { BaseUIComponentProps } from '../../utils/types';
 import { useRenderElement } from '../../utils/useRenderElement';
 import { useSelectRootContext } from '../root/SelectRootContext';
 import { selectors } from '../store';
+import { CustomStyleHookMapping } from '../../utils/getStyleHookProps';
+
+const customStyleHookMapping: CustomStyleHookMapping<SelectValue.State> = {
+  value: () => null,
+};
 
 /**
  * A text label of the currently selected item.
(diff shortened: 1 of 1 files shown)


packages/react/src/select/value/SelectValue.tsx
romgrk, line 1: Reading the linked issue, the last comment is:

> I'll reopen this because I think it would still be helpful to change the class based on state

I wonder what's the use-case we're intending to solve with this feature? The original one (dimmed color for placeholder value) is already solved, I don't understand why we really need this.

I would advocate against adding features unless we can state a use-case for it, in particular if it adds overhead (perf-wise and maintainability-wise).

atomiks: `Select.Value` renders an element, so it's useful if you want to re-use the same element. It renders one because a ref gets attached for `alignItemWithTrigger` to work correctly.

PR #2155, opened 2025-06-24
[checkbox][CheckboxGroup] Fix `onCheckedChange` not running when parent checkbox is present

mj12albert:
Repro: https://codesandbox.io/p/sandbox/runtime-moon-m4snqz

In this case although checking the parent also changes the checked state of the child, the child checkbox's change handler does not run
```jsx
<CheckboxGroup>
  <Checkbox parent onCheckedChange={handleParentChange} />
  <Checkbox onCheckedChange={handleChildChange} />
</CheckboxGroup>
```

### DIFF
--- a/packages/react/src/select/value/SelectValue.tsx
+++ b/packages/react/src/select/value/SelectValue.tsx
@@ -5,6 +5,11 @@ import type { BaseUIComponentProps } from '../../utils/types';
 import { useRenderElement } from '../../utils/useRenderElement';
 import { useSelectRootContext } from '../root/SelectRootContext';
 import { selectors } from '../store';
+import { CustomStyleHookMapping } from '../../utils/getStyleHookProps';
+
+const customStyleHookMapping: CustomStyleHookMapping<SelectValue.State> = {
+  value: () => null,
+};
 
 /**
  * A text label of the currently selected item.
@@ -32,14 +37,23 @@ export const SelectValue = React.forwardRef(function SelectValue(
     return null;
   }, [items, value]);
 
+  const state: SelectValue.State = React.useMemo(
+    () => ({
+      value,
+    }),
+    [value],
+  );
+
   const children =
     typeof childrenProp === 'function'
       ? childrenProp(value)
       : (childrenProp ?? labelFromItems ?? value);
 
   const element = useRenderElement('span', componentProps, {
+    state,
     ref: [forwardedRef, valueRef],
     props: [{ children }, elementProps],
+    customStyleHookMapping,
   });
 
   return element;
@@ -59,5 +73,10 @@ export namespace SelectValue {
     children?: React.ReactNode | ((value: any) => React.ReactNode);
   }
 
-  export interface State {}
+  export interface State {
+    /**
+     * The value of the currently selected item.
+     */
+    value: any;
+  }
 }
--- a/packages/react/src/tabs/tab/TabsTab.tsx
+++ b/packages/react/src/tabs/tab/TabsTab.tsx
@@ -95,8 +95,6 @@ export const TabsTab = React.forwardRef(function Tab(
   const isPressingRef = React.useRef(false);
   const isMainButtonRef = React.useRef(false);
 
-  const highlighted = index > -1 && index === highlightedTabIndex;
-
   const onClick = useEventCallback((event: React.MouseEvent<HTMLButtonElement>) => {
     if (selected || disabled) {
       return;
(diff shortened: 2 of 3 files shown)


mj12albert:
Currently `onCheckedChange` doesn't run when the controlled value changes externally, and `Checkbox.Root` itself can't differentiate between a external change caused by a parent or this:
```jsx
<button onClick={() => setChecked(!checked)}>external</button>

<Checkbox.Root
  checked={checked}
  onCheckedChange={(nextChecked) => {
    console.log('checked change');
    setChecked(nextChecked);
  }}
/>
```

Another thing is if a child checkbox's `onCheckedChange` is caused by a parent checkbox change, there's no relevant event to pass to the 2nd arg of `onCheckedChange`. It would need to be `undefined` for this scenario, or provide the event from the parent checkbox which doesn't seem right

**UPDATE:** For consistency with other components controlled behavior, checking the parent should not fire the child's onCheckedChange

packages/react/src/checkbox-group/useCheckboxGroupParent.test.tsx
```diff
@@ -3,19 +3,22 @@ import { createRenderer, fireEvent, screen } from '@mui/internal-test-utils';
...
     function App() {
       const [value, setValue] = React.useState<string[]>([]);
       return (
         <CheckboxGroup value={value} onValueChange={setValue} allValues={allValues}>
-          <Checkbox.Root parent data-testid="parent" />
+          <Checkbox.Root parent data-testid="parent" onCheckedChange={parentCheckedChange} />
           <Checkbox.Root name="a" />
-          <Checkbox.Root name="b" />
+          <Checkbox.Root name="b" onCheckedChange={childCheckedChange} />
           <Checkbox.Root name="c" />
```
atomiks: Shouldn't these tests use `value` now?

mj12albert: Updated, the ~parent hook was still incorrectly checking `name` instaed of `value` as well~ child checkbox was incorrectly registering `name` instead of `value` to the parent

packages/react/src/checkbox-group/useCheckboxGroupParent.test.tsx
```diff
@@ -39,23 +42,30 @@ describe('useCheckboxGroupParent', () => {
...
+
     fireEvent.click(parent);
     expect(parent).to.have.attribute('aria-checked', 'false');
 
     checkboxes.forEach((checkbox) => {
       expect(checkbox).to.have.attribute('aria-checked', 'false');
     });
+
+    expect(parentCheckedChange.callCount).to.equal(2);
+    expect(childCheckedChange.callCount).to.equal(0);
```
atomiks, line 56: Why does/should the child one not get called here?

mj12albert: To the child checkbox, a state change caused by toggling the parent is the same as the controlled checked state changing externally: https://github.com/mui/base-ui/pull/2155#issuecomment-3000090633

I'm 50-50 on whether this should be an exception, my first thought on seeing Aaron's repro is it should run @atomiks

atomiks: Right, yeah that makes sense. For consistency, since the parent is acting as controlled state, this seems "correct"

atomiks (approved):

packages/react/src/checkbox/root/CheckboxRoot.tsx
```diff
@@ -249,10 +250,10 @@ export const CheckboxRoot = React.forwardRef(function CheckboxRoot(
   const computedIndeterminate = isGrouped ? groupIndeterminate || indeterminate : indeterminate;
 
   React.useEffect(() => {
-    if (parentContext && name) {
-      parentContext.disabledStatesRef.current.set(name, disabled);
+    if (parentContext && value) {
+      parentContext.disabledStatesRef.current.set(value, disabled);
     }
-  }, [parentContext, disabled, name]);
+  }, [parentContext, disabled, value]);
```
atomiks, line 256: I'm assuming the disabledStatesRef logic was already updated to use value?

mj12albert: Yep ~ in `useCheckboxGroupParent` it only needs `allValues` and was doing that already (I mis-worded my earlier [comment](https://github.com/mui/base-ui/pull/2155#discussion_r2168003284))

PR #2156, opened 2025-06-24
[menu] Fix (click, drag outside, release) not closing the menu

michaldudak:
#2094 broke cancelling opening of the menu by dragging outside.

Here I'm setting a root id on every menu popup and when a mouse pointer is released over an element I check if this element belongs to the root.


packages/react/src/menu/popup/MenuPopup.tsx
```diff
@@ -92,6 +93,7 @@ export const MenuPopup = React.forwardRef(function MenuPopup(
       popupProps,
       transitionStatus === 'starting' ? DISABLED_TRANSITIONS_STYLE : EMPTY_OBJECT,
       elementProps,
+      { 'data-rootownerid': rootId } as Record<string, string>, // This should be replaced with the actual root owner ID if available
```
atomiks: I'm struggling to understand  this comment

Also we've been using `{ ['data-rootownerid' as string]: rootId }`

michaldudak: I forgot to remove the comment.

As for `{ ['data-rootownerid' as string]: rootId }` - this makes the property name computed, which is unnecessary. `{ 'data-rootownerid': rootId } as Record<string, string>` does not affect runtime code at all.

atomiks:
This seems to break the exit animation when you mousedown -> mousemove into a submenu -> mouseup

michaldudak:
I can't reproduce it. Does it still happen to you after the last merge commit?


https://github.com/user-attachments/assets/e6d91f58-a390-45a3-854d-060a2a79a7c6

atomiks:
It happens with Menubar. It also happens when moving to a different menu, not just a submenu

michaldudak:
Alright, fixed.

packages/react/src/menu/trigger/MenuTrigger.tsx
```diff
@@ -188,3 +189,15 @@ export namespace MenuTrigger {
...
+  if (element.hasAttribute('data-rootownerid')) {
+    return element.getAttribute('data-rootownerid') ?? undefined;
+  }
+
+  if (!element.parentElement || element.parentElement === document.body) {
+    return undefined;
+  }
+
+  return findRootOwnerId(element.parentElement);
+}
```
atomiks: The `isLastTraversableNode` util from `floating-ui-react` should work better here

michaldudak: `isLastTraversableNode` does a whole lot more we don't really need here (checking the type, accessing node name, etc.). In this case it should even be enough to check for `!element.parentElement`.

atomiks: It's more robust by checking `html` / `#document` which can sometimes appear (especially cross-browser). It's used in Floating UI so it's performant enough to be used. Otherwise it's possible for an infinite loop to occur

michaldudak: `document` shouldn't ever appear here as it's not an `Element`. I'm curious in what cases this will lead to an infinite loop?

atomiks: It's edge-casey, but possible
For instance, `document.body` in this case isn't using the same owning window body as the element. It's simple enough to use that it's worth it

michaldudak: Even if `element.parentElement === document.body` would let some cases through, `!element.parentElement` will be always reliable, no? HTML elements can't form cycles.

atomiks: `.parentElement` may work differently to `.parentNode` here - I just remember this being something that needed to be used in Popper.js for some reason, possibly to patch out cross-browser issues years ago.

So even if it ends up falsy while traversing, the `body` check won't work in the iframe case, so why have it? And since it's used in Floating UI, this doesn't add any bundle size (Menu is already using it when it opens), nor is it slow, so I don't see a reason not to use it?

packages/react/src/menu/trigger/MenuTrigger.tsx
```diff
@@ -188,3 +189,15 @@ export namespace MenuTrigger {
...
+  if (element.hasAttribute('data-rootownerid')) {
+    return element.getAttribute('data-rootownerid') ?? undefined;
+  }
+
+  if (element.parentElement == null) {
+    return undefined;
+  }
+
+  return findRootOwnerId(element.parentElement);
+}
```
atomiks: I realized this also won't work in shadow DOM, which isn't common with React but is still a possible/valid use case

atomiks: How `getContainingBlock` does it:

```jsx
export function getContainingBlock(element: Element): HTMLElement | null {
  let currentNode: Node | null = getParentNode(element);

  while (isHTMLElement(currentNode) && !isLastTraversableNode(currentNode)) {
    if (isContainingBlock(currentNode)) {
      return currentNode;
    } else if (isTopLayer(currentNode)) {
      return null;
    }

    currentNode = getParentNode(currentNode);
  }

  return null;
}
```

These utils are available to use in `@floating-ui/utils`

michaldudak: Yeah, OK, this is a good point. Updated.

atomiks:
I was able to get into this state where dragging to a new menu trigger doesn't open it:

https://github.com/user-attachments/assets/2603f577-775c-4c21-8e9c-28ba4c0fa859

atomiks:
This also exists in master, looks like a regression.
It also happens if you try to `mouseenter` in general to the next trigger, regardless of the drag

atomiks (approved):

PR #2163, opened 2025-06-25
[menu] Fix active index sync on hover

atomiks:

1. Open a menu (with submenu triggers) with the keyboard while resting the cursor over where the submenu will appear
2. The menu won't open unexpectedly until `mousemove`
3. Moving the cursor from a submenu trigger to another submenu adjacent to it while stopping over it abruptly (to prevent triggering subsequent `mousemove`s) now synchronizes the highlight correctly on `mouseenter` while opening the submenu as before

https://deploy-preview-2163--base-ui.netlify.app/experiments/menu/menu-fully-featured


packages/react/src/menu/item/useMenuItem.ts
```diff
@@ -43,6 +43,17 @@ export function useMenuItem(params: useMenuItem.Parameters): useMenuItem.ReturnV
...
+          onMouseEnter() {
+            if (
+              !submenuTrigger.allowMouseEnterEnabled ||
+              !submenuTrigger.setActiveIndex ||
+              submenuTrigger.index == null
+            ) {
+              return;
+            }
+
+            submenuTrigger.setActiveIndex(submenuTrigger.index);
```
michaldudak: Nit: It looks a bit weird to expose both `setActiveIndex` and `index` and only call `setActiveIndex` with the associated `index`. A parameterless `setActive()` would look better.

Issue #2121, opened 2025-06-17
[popover] Popover becomes unanchored when trigger is invisible


ryanleichty:
# Bug report

## Current behavior

When a popover trigger becomes invisible (via `display: none` or anything else that would return false via `triggerElement.checkVisibility()`), the popover becomes unanchored.

![Image](https://github.com/user-attachments/assets/016761b8-1998-4c4c-a88b-8d7cd2b7d598)

## Expected behavior

I would expect the popover to close.

## Reproducible example

https://base-ui.com/react/components/popover

Open the popover, set the button display property to `none` (in the browser)

## Base UI version

Tested in the browser in the docs

## Which browser are you using?

Chrome

## Which OS are you using?

Mac OS

## Which assistive tech are you using (if applicable)?

None

## Additional context

I'm actually implementing a popover in a non-React context and was trying to follow a similar pattern as Base UI (also using Floating UI under the hood) and ran into the same issue where I was hiding the popover trigger on desktop, causing the popover to become unanchored. I recently stumbled upon [`checkVisibility()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility) and used it to solve the problem by running it in the `autoUpdate` function from Floating UI.

I was mainly curious to see how you might've handled this scenario.

mj12albert:
> I recently stumbled upon [`checkVisibility()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility)

Ohh I did not know about this 👀 maybe it could be used for https://github.com/mui/base-ui/pull/1907 as well @atomiks

atomiks:
Currently, `data-anchor-hidden` only checks clipping but this check (`checkVisibility`) could be added for that attribute for the `display: none` case. That would allow you to hide the popup when the anchor is hidden in this scenario

ryanleichty:
[Support isn't the greatest in Safari](https://caniuse.com/mdn-api_element_checkvisibility) since it's fairly new (Safari 17.4) so I created a basic polyfill to be on the safe side. Otherwise, it's fairly helpful

piyushzala158:
Raised a PR for above issue: https://github.com/mui/base-ui/pull/2166

OliverJAsh:
In case it helps, here's a real world example of where this happens:

https://github.com/user-attachments/assets/040b8bef-7969-43c1-8b46-3277996c38f1

PR #2166, opened 2025-06-25
fix[popover]:Popover becomes unanchored when trigger is invisible

piyushzala158:

Issue : https://github.com/mui/base-ui/issues/2121

Screenshot 2025-06-25 at 8 24 03 PM


michaldudak:
This won't work in cases where an ancestor of the trigger becomes hidden. Making this work reliably in all cases is quite tricky and may have a negative impact on performance (especially when observing all ancestors). I'd recommend using the same approach as in #1907 - ignoring it. We can update the docs to say that the popup should be hidden manually whenever the trigger gets hidden.
cc @atomiks

atomiks:
This may be more of a feature or extension of the [`hide()` middleware](https://github.com/floating-ui/floating-ui/blob/master/packages/core/src/middleware/hide.ts) - if the `autoUpdate` function runs to recalculate the position when the trigger hides, then `[data-anchor-hidden]` would be applied to the positioner to hide it with CSS (`visibility: hidden`)

michaldudak:
Fixed in #3267

PR #2167, opened 2025-06-25
[code-infra] Extract utilities to a separate package

michaldudak:
Extracted utilities that can be used in other MUI projects to a separate library - [@base-ui-components/utils](https://www.npmjs.com/package/@base-ui-components/utils).

Before we start using it across the company, I want to update JSDocs and possibly rename and reorganize few utilities.


Janpot:
> can be used in other MUI projects to a separate library

That means they will be semver stable? Should we document them?

cc @romgrk to get your attention on this PR, I believe you brought up the topic of shared libraries recently.

packages/react-utils/README.md
```diff
@@ -0,0 +1,3 @@
+# @base-ui-components/react-utils
```
romgrk: I think I'd go with just `@base-ui-components/utils` for the name. The package contains more than just react-specific utils, and the name is already long due to the prefix.

packages/react-utils/package.json
romgrk, line 1: > That means they will be semver stable? Should we document them?

Yes it should follow semver, and no I don't think we should document them beyond jsdoc.

packages/react-utils/README.md
michaldudak: Yeah, I'm fine with it too. I'm wondering if it shouldn't be `internal-utils`, though, to match the convention we use in other such libraries (@Janpot, any opinions?).

Janpot: According to the current [conventions](https://www.notion.so/mui-org/engineering-mui-utils-purpose-9a9fc9da3a004864b6c4e1f4d1f24f95?source=copy_link#d1699bb48f0f4ef58ca831d1ee9772ec), that would make it semver unstable and then it can only be depended in this repo, and only if the dependency is defined with exact version.
I think `@base-ui-components/utils` makes sense, just don't make breaking changes in it on a minor.

michaldudak:
Tested the release process and uploaded https://www.npmjs.com/package/@base-ui-components/utils

packages/react/src/alert-dialog/popup/AlertDialogPopup.tsx
```diff
@@ -1,5 +1,6 @@
 'use client';
 import * as React from 'react';
+import { useForkRef, InteractionType, inertValue } from '@base-ui-components/utils';
```
romgrk: We've been moving towards non-barrel imports in others parts of the codebase, I think for DX reasons. Might not matter as much for a small package like the utils, but it would be nice to agree on a direction.

michaldudak: I went this way to make the imports shorter, as there's usually more of them imported in a single file. I don't think this will affect dev-time performance as utils are small functions and might even reduce the bundle size a little bit.

Janpot: > but it would be nice to agree on a direction.

Barrel files has had too large impact on dev mode in vite and typescript compile times for our users in the past. Therefore I don't think the direction can ever be "we start using barrel files from now on for everything". At least in our public demos and our largest packages we should keep the current method recommendation of avoiding barrel files. Question remains whether for the rest of the imports we leave it up to the intuition of the programmer. Personally, I would recommend just doing it everywhere. I only can imagine it having positive impacts and it frees us from having to think and decide about every import whether ot should be barrel or not. It also ensures consistency across different authors, which improves maintainability.

> utils are small functions and might even reduce the bundle size a little bit.

🤔 I would expect either no impact or the opposite. Where do you base this info on?

michaldudak: I mean importing them can lead to shorter code:
```tsx
import { useLazyRef } from '@base-ui-components/utils/useLazyRef';
import { useModernLayoutEffect } from '@base-ui-components/utils/useModernLayoutEffect';
import { ownerDocument } from '@base-ui-components/utils/owner';
```
is longer than
```tsx
import { useLazyRef, useModernLayoutEffect, ownerDocument } from '@base-ui-components/utils';
```

(220 vs 93 characters in this particular example).

Considering that utils are used heavily in almost all components, this can lead to noticeable savings.

brijeshb42: But during bundling (on the end-users side), the paths won't be part of the final output. It only saves code in the source code and is immaterial to the built artifacts.

Janpot: But wouldn't those imports also exist in the barrel file and therefore pulled into the bundle anyway? I'm not sure whether they fully resolve imports and compile away the barrel file or include a chunk for the barrel file, but in the latter case you're definitely getting larger bundles, in the former it'd be the same as if you wrote it like your first snippet. (Some bundlers may just hoist everything and get rid of imports altogether). As far as I'm aware, bundlers will also rewrite those specifiers so the length of them doesn't influence the final bundle size.

michaldudak: Yeah, Brijesh is right that this will be bundled with the end application, so doesn't really matter.

I'm fine with individual imports. I just think barrel files offer slightly better DX when importing (at the possible expense of dev-time performance, but this is likely negligible).

docs/src/app/(private)/experiments/dialog.tsx
```diff
@@ -2,12 +2,9 @@
 import * as React from 'react';
 import clsx from 'clsx';
 import { Dialog } from '@base-ui-components/react/dialog';
-import { useTransitionStatus } from '@base-ui-components/react/utils';
-import {
-  animated as springAnimated,
-  useSpring,
-  useSpringRef,
-} from '@react-spring/web';
+// eslint-disable-next-line no-restricted-imports
```
LukasTy, line 5: If this package does not have problems importing files from more than 2 levels deep, then maybe the ESLint rules should be adjusted? 🤔

docs/src/blocks/GoogleAnalytics.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 import * as React from 'react';
 import { useMediaQuery } from '@base-ui-components/react/unstable-use-media-query';
-import { useModernLayoutEffect } from '@base-ui-components/react/utils';
+import { useModernLayoutEffect } from '@base-ui-components/utils';
```
LukasTy: Nitpick: At least on X we settled on using deeper imports on demos even on `@mui/utils` package.

packages/react/src/floating-ui-react/components/FloatingPortal.test.tsx
```diff
@@ -1,8 +1,8 @@
 import { fireEvent, flushMicrotasks, render, screen } from '@mui/internal-test-utils';
 import * as React from 'react';
 
+import { isJSDOM } from '@base-ui-components/utils';
```
LukasTy: This feels slightly out of place as a `util`. 🤔 
Feel more like a `internal-util` or `internal-test-utils`. 🙈 
Or maybe I'm biased by the usage of `@mui/internal-test-utils`. 😆

packages/utils/README.md
```diff
@@ -0,0 +1,3 @@
+# @base-ui-components/utils
+
+A collection of React utility functions for Base UI.
```
LukasTy:
```suggestion
A collection of React utility functions for Base UI.
```

docs/src/app/(private)/experiments/dialog.tsx
```diff
@@ -2,12 +2,9 @@
...
+import { useTransitionStatus } from '@base-ui-components/react/utils/useTransitionStatus';
```
brijeshb42, line 6: Shouldn't this be -

```suggestion
import { useTransitionStatus } from '@base-ui-components/utils/useTransitionStatus';
```

docs/src/components/MobileNav.tsx
```diff
@@ -4,7 +4,8 @@ import clsx from 'clsx';
 import NextLink from 'next/link';
 import { Dialog } from '@base-ui-components/react/dialog';
 import * as ReactDOM from 'react-dom';
-import { useScrollLock } from '@base-ui-components/react/utils';
+// eslint-disable-next-line no-restricted-imports
+import { useScrollLock } from '@base-ui-components/react/utils/useScrollLock';
```
brijeshb42, line 8: Same

docs/src/app/(private)/experiments/dialog.tsx
michaldudak: No, `useTransitionStatus` is Base UI-specific and lives in the main package.

LukasTy (commented):
Great work!
Thank you for exporting these. 🙏 
Leaving some minor nitpicks.

I've tested it on https://github.com/mui/mui-x/pull/18734 and have a question.
Have you considered exporting `useButton` hook? 🤔

packages/react/src/floating-ui-react/components/FloatingPortal.test.tsx
romgrk: Once we have the utils package, I plan to group all platform-detection logic in a single `platform` export, e.g. `platform.isJSDOM`. Note that `isJSDOM` is required in utils, jsdom is part of the platforms that we support and we need that value in non-test code sometimes (particularly in the grid).

In `x-internals` I've done that switch [in this PR](https://github.com/mui/mui-x/pull/18275/files#diff-95514cf15464b808890b1e2fcbfc0d18678f5be4301ac3fac3788abc10d5ad0f).

LukasTy: Thanks for expanding on this, Rom. 👍 
I completely forgot that some packages do feature detection based on this flag in production code. 🙈

docs/src/app/(private)/experiments/dialog.tsx
michaldudak: It doesn't really import these from the **package**. Docs use Base UI sources directly, so they can access private exports.

michaldudak:
> Have you considered exporting useButton hook? 🤔

We might actually have a Button component at some point, but it's not fully decided yet.
If you need to add button functions to a non-interactive component, I suppose `useButton` could be useful. However, before we export it, I'd like to clean it up properly and remove handling of HTML anchors (we could have `useLink` if there's a need, but I don't see it now).

docs/src/app/(private)/experiments/dialog.tsx
Janpot: > Docs use Base UI sources directly

Is this avoidable? this tends to mess up caching algorithms that rely only on npm dependencies. and all kinds of other subtle problems, e.g. typescript filles outside of the root folder and stuff like that

romgrk: It does create subtle problems, but it also makes the devmode DX much nicer. HMR works seamlessly if you use source files rather than built files, and you can import anything you want. The setup is harder to maintain, but the tradeoff is worth it, imho.

romgrk (approved):

docs/src/app/(private)/experiments/dialog.tsx
dav-is: > HMR works seamlessly if you use source files rather than built files

I think for this case, we should be using a watch script for typescript builds of the lib. HMR should work fine on the built files after they are updated by the typescript watcher.

I think another good purpose of the docs is testing that our packaging is working properly. I could see an issue if there were subtle differences in the typescript config between the docs and the lib.

Also, why do the docs need to access private exports?

Janpot: if necessary we can alias to sources for dev mode, I don't mind so much, as long as the aliasing can be removed and no code changes are necessary for regular pnpm workspaces resolution to work correctly.

each package building in watch mode has my preference, but we need to make sure it can scale to the size of core and X monorepos. We'll likely start work on a vite configuration for building our libraries soon.

romgrk: > I could see an issue if there were subtle differences in the typescript config between the docs and the lib.

I have witnessed such issues in the past, but I argued unsuccessfully for running our test suites (rather than the docs) in prod mode to deal with that.

> Also, why do the docs need to access private exports?

Private experiments that we don't want to publish, such as this case and anything else under `.../(private)/experiments/`.

> I think for this case, we should be using a watch script for typescript builds of the lib. HMR should work fine on the built files after they are updated by the typescript watcher.

If it can be shown to not affect the DX, I'm fine with it.

michaldudak: > Is this avoidable? this tends to mess up caching algorithms that rely only on npm dependencies. and all kinds of other subtle problems, e.g. typescript filles outside of the root folder and stuff like that

We can discuss it on the infra meeting. It's not in scope of this PR anyway.

PR #2168, opened 2025-06-25
feat[popover]:added trigger ref to support external trigger outside root

piyushzala158:

### DIFF
--- a/packages/react/src/popover/root/PopoverRoot.tsx
+++ b/packages/react/src/popover/root/PopoverRoot.tsx
@@ -43,7 +43,8 @@ function PopoverRootComponent({ props }: { props: PopoverRoot.Props }) {
   const [instantType, setInstantType] = React.useState<'dismiss' | 'click'>();
   const [titleId, setTitleId] = React.useState<string>();
   const [descriptionId, setDescriptionId] = React.useState<string>();
-  const [triggerElement, setTriggerElement] = React.useState<Element | null>(null);
+  const [internalTriggerElement, setInternalTriggerElement] = React.useState<Element | null>(null);
+  const triggerElement = props.triggerRef?.current ?? internalTriggerElement;
   const [positionerElement, setPositionerElement] = React.useState<HTMLElement | null>(null);
   const [openReason, setOpenReason] = React.useState<PopoverOpenChangeReason | null>(null);
   const [stickIfOpen, setStickIfOpen] = React.useState(true);
@@ -170,7 +171,7 @@ function PopoverRootComponent({ props }: { props: PopoverRoot.Props }) {
       setMounted,
       transitionStatus,
       triggerElement,
-      setTriggerElement,
+      setTriggerElement: props.triggerRef ? () => {} : setInternalTriggerElement,
       positionerElement,
       setPositionerElement,
       popupRef,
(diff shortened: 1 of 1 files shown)


oleksandr-danylchenko:
Thank you, @piyushzala158! 
Do I understand correctly that the `reference` passed down to the `useFloatingRootContext` will get its aria attributes and event listeners properly populated?

piyushzala158:
> Thank you, @piyushzala158! Do I understand correctly that the `reference` passed down to the `useFloatingRootContext` will get its aria attributes and event listeners properly populated?

Yes

packages/react/src/popover/root/PopoverRoot.tsx
```diff
@@ -170,7 +171,7 @@ function PopoverRootComponent({ props }: { props: PopoverRoot.Props }) {
       setMounted,
       transitionStatus,
       triggerElement,
-      setTriggerElement,
+      setTriggerElement: props.triggerRef ? () => {} : setInternalTriggerElement,
```
oleksandr-danylchenko: Also, there might be a conflict if you try both passing the `triggerRef` and rendering the `Popover.Trigger` within. I understand that it's an edge case and not the intended way to use the component. But maybe there's a way to be more defensive against that 🤔.

I believe that instead of using a noop for the `setTriggerElement`, some form of a warning message can be provided:
```suggestion
      setTriggerElement: props.triggerRef 
        ? () => { console.warn("WHAT THE HELL ARE YOU DOING? THERE'S A REF ALREADY!") } 
        : setInternalTriggerElement,
```
I don't think it should be a hard exception, though. As it might yield quite a few "uncaught exception" logs for the consuming apps.

piyushzala158: sure that will good 👍

atomiks:
Thanks for the contribution @piyushzala158

Right now, though, this isn't going to work 😅: the trigger's props (`getReferenceProps`) are generated inside `Popover.Root`, and they can't be passed to the external trigger without bubbling them using an effect.

https://github.com/mui/base-ui/blob/bb26cbe3af94c6c5c6400773c291c3675f13ef7d/packages/react/src/popover/root/PopoverRoot.tsx#L164

We discussed this somewhat briefly yesterday and thought something like [Toast's API](https://base-ui.com/react/components/toast) could possibly work for things like Dialogs and Popovers. 

This means you render your Popover(s) high up in the tree and use a hook to "create" them by passing structured data that the renderer uses as a template. 

That said, I don't know if this solves @oleksandr-danylchenko's use case with the `[data-page-family-id]` needing to be query selected?

oleksandr-danylchenko:
> We discussed this somewhat briefly yesterday and thought something like [Toast's API](https://base-ui.com/react/components/toast) could possibly work for things like Dialogs and Popovers.
> 
> This means you render your Popover(s) high up in the tree and use a hook to "create" them by passing structured data that the renderer uses as a template.

I like this idea 🙌🏻 
That provider can expose a context with methods like `popover(anchorEl, {})`, or `popoverManager.add(anchorEl, {})`. That way, we'll also be able to handle the presence of just a single popover at a time.

michaldudak:
Thanks for your work! We've been working on a more comprehensive solution in #2336, so I'm closing your PR in favor of that one.

PR #2169, opened 2025-06-26
[accordion] Fix transition status mapping

atomiks:
`Accordion.Trigger` had `data-transitionstatus="..."`


atomiks:
`State` should likely not contain `transitionStatus` as the `Trigger` shouldn't know about it (this is also how popups work)

PR #2170, opened 2025-06-26
[accordion][collapsible] Fix `aria-controls` reference

atomiks:


packages/react/src/collapsible/root/CollapsibleRoot.test.tsx
```diff
@@ -68,7 +68,6 @@ describe('<Collapsible.Root />', () => {
       const externalTrigger = getByRole('button', { name: 'toggle' });
       const trigger = getByRole('button', { name: 'trigger' });
 
-      expect(trigger).to.not.have.attribute('aria-controls');
```
atomiks: Not sure why these have `not` assertions @mj12albert?

mj12albert: `aria-controls` should be gone whenever the panel is closed/`hidden`, that's still the case after this fix

atomiks: I realized it was causing some test issue that is now fixed - I misread it as being `not.to.have.attribute()` for the `aria-expanded=true` lines, but it's just missing for those ones to begin with

Issue #1956, opened 2025-05-21
[select] Support multiple selections


OliverJAsh:
# Feature request

## Summary

I have the following design which appears to represent a `Select` with multiple selections. Base UI's `Select` doesn't appear to support multiple selections right now.

Perhaps this should be implemented as a different component, in which case I will close this issue.

<img width="553" alt="Image" src="https://github.com/user-attachments/assets/39ab1e1b-6af4-4556-a18f-c91ce3cb3a0d" />

## Examples in other libraries

- [`<select>`'s `multiple` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/multiple).
- [Open issue for the same feature in Radix](https://github.com/radix-ui/primitives/issues/1270), with lots of upvotes.
- [React Aria's `Select` has `selectionMode="multiple"`](https://react-spectrum.adobe.com/react-aria/Select.html#listbox).

## Motivation

Currently search filters on Unsplash only support one selection at a time. We would like to allow multiple selections.

Example: https://unsplash.com/s/photos/dog

<img width="591" alt="Image" src="https://github.com/user-attachments/assets/7f724078-e181-4c47-ac32-ec66e92e27d7" />

atomiks:
Open PR #1927 

I think in your example since there's no label in/near the trigger, you could perhaps use `Menu` + `Menu.CheckboxItem` 🤔  (there isn't any `Field`/`Form` integration for Menu, though)

OliverJAsh:
Ah, nice!

> I think in your example since there's no label in/near the trigger

There is when there's a selection e.g.:

<img width="566" alt="Image" src="https://github.com/user-attachments/assets/a19e9d2d-ec21-45fe-8e87-4b52f57efa02" />

colmtuite:
Let's say it was a multi-select, what would you put in the trigger when multiple options are selected?

OliverJAsh:
To answer that question, here's what our design team have in Figma. It's very possible this is a bad design and we need to go back to the drawing board. I would personally expect something like "Orientation (2)".

<img width="841" alt="Image" src="https://github.com/user-attachments/assets/f999348d-7ef3-4810-b446-aa60b07dc777" />

colmtuite:
> I would personally expect something like "Orientation (2)"

[...]

PR #2173, opened 2025-06-26
[select] Add `multiple` prop

atomiks:
Closes https://github.com/mui/base-ui/issues/1956

https://deploy-preview-2173--base-ui.netlify.app/react/components/select#multiple-selection


packages/react/src/select/value/SelectValue.tsx
```diff
@@ -23,19 +23,44 @@ export const SelectValue = React.forwardRef(function SelectValue(
...
+      const map = new Map<any, React.ReactNode>();
+      for (const { value: itemValue, label } of items) {
+        map.set(itemValue, label);
       }
-      return items[value];
+      return map;
+    }
+
+    return new Map<any, React.ReactNode>(Object.entries(items));
+  }, [items]);
```
romgrk: I don't insist on it, but this is a case where the store and `reselect` can help. You can put derived values in memoized selectors. Those are more efficient than React's `useMemo`.

```typescript
const itemsSelector = state => state.items
const valueSelector = state => state.value
const multipleSelector = state => state.multiple

const valueLabelMapSelector = createSelectorMemoized(
  itemsSelector,
  items => { /* ... */ }
)

const labelFromItems = createSelectorMemoized(
  valueSelector,
  multipleSelector,
  valueLabelMapSelector,
  (value, multiple, valueLabelMap) => { /* ... */ }
)
```

---

That being said, also note that I'm not sure this code is actually beneficial as a performance optimization. The main issue is that we're adding the cost of building a `Map` (an `O(n)` cost) for each `Select` on mount. If you have a page with a large number of those components, you're actually increasing the page-load time.

docs/reference/generated/select-root.json
```diff
@@ -45,6 +45,11 @@
       "default": "true",
       "description": "Determines if the select enters a modal state when open.\n- `true`: user interaction is limited to the select: document page scroll is locked and and pointer interactions on outside elements are disabled.\n- `false`: user interaction with the rest of the document is allowed."
     },
+    "multiple": {
+      "type": "false",
```
michaldudak, line 49: That's not quite right. I'll see what's going on in the API extractor

packages/react/src/select/value/SelectValue.tsx
michaldudak: Building a map could potentially be beneficial (or at least not harmful) when `items` is an array. Without the map, we'd need to find the label to display in the array, what also is O(n). However, when `items` is an object, it seems pointless.

atomiks: I changed it so it only creates a map under this condition:

`Array.isArray(items) && Array.isArray(value)`

As this is the one that creates a loop within a loop to find the label for each value. It should also not be so common to have many of these on one page. I _also_ think most will be using a custom renderer (`children` function) instead of using the default comma-separated behavior, which sidesteps this

ej-shafran:
Hey y'all - is there any update on this? What needs to be done to get this merged?

If needed, I can pick this up and help fix the relevant issues... Would really love to have this feature as it's very relevant for basically every project where I use a `Select`...

docs/reference/generated/select-root.json
michaldudak: https://github.com/mui/base-ui/pull/2104 should fix it.

michaldudak (approved):
Aside from https://github.com/mui/base-ui/pull/2173#discussion_r2180229584, this looks OK. If you merge this before #2104, I'll fix the docs in that PR.

packages/react/src/select/value/SelectValue.tsx
romgrk: > Building a map could potentially be beneficial (or at least not harmful)

That's false. Building a map upfront means you're forcing an `O(n)` cost to happen at mount time. This only saves time assuming that:
 - The user changes the value of the select multiple times (which I could argue is a rarer use-case than only setting its value once)
 - Every select on a page gets its value changed (otherwise, you're paying an `O(n)` cost for no reason), which is not necessarily true

>  It should also not be so common to have many of these on one page.

I feel like you're consistently thinking only of simple cases. If you want to build a general-purpose framework that can be used for a wide variety of cases, you need to consider complex/edge cases as well. For example, if a user wants `Select` in datagrid cells, I can guarantee you I'm going to be unhappy about the mount-time performance cost of creating those `Map`:s while the user is scrolling the grid - it directly competes with the virtualization logic for the available CPU time. 

An unopened `Select` is basically a `button` component. It should not have an `O(n)` cost associated with it at mount-time.

romgrk: And the reason why I said store selectors can be more efficient is that they would allow you to make that `O(n)` computation lazy, delaying it to until the map is needed for the first time. `React.useMemo()` forces the `O(n)` immediately. You would get all the benefits and no downsides if you moved that computation to a store selector (provided you don't `useSelector` it but call the selector directly when you need it).

atomiks: The latest commit removes the map entirely as they always need to use a custom renderer for multiple selection. They'll use their own lookup map to retrieve each label for the values

atomiks:
Note on hidden `<select>` vs. hidden `<input>`s:

From what I can tell the hidden `<select>` requires _all_ possible `<option>`s to be available upfront (the browser autofill test fails in particular if not), which would make the `items` prop required and also introduce some cost of creating all the `<option>` elements. The hidden inputs on the other hand don't have this issue

PR #2174, opened 2025-06-26
[select][number field] Fix revalidation on change

atomiks:

Select was a regression from `beta.0`, likely due to the Store refactoring - the `store` value is updated later than the internal `value`, so the hidden `<input>` was getting a slightly stale value - the internal hook to commit the validation therefore was executing sooner than React had updated the hidden input `value` in the DOM for native validation
NumberField issue existed in `beta.0`

https://deploy-preview-2174--base-ui.netlify.app/experiments/form

1. Press submit without filling anything in
2. NumberField: Press increment -> validates immediately (required blur before)
3. Select: choose a value -> validates immediately (required blur before)

### DIFF
--- a/packages/react/src/number-field/input/NumberFieldInput.tsx
+++ b/packages/react/src/number-field/input/NumberFieldInput.tsx
@@ -112,12 +112,7 @@ export const NumberFieldInput = React.forwardRef(function NumberFieldInput(
       blockRevalidationRef.current = false;
       return;
     }
-
-    // Defer the validation until after React has committed the DOM updates
-    // to ensure the hidden input element reflects the latest value.
-    queueMicrotask(() => {
-      commitValidation(value, true);
-    });
+    commitValidation(value, true);
   }, [commitValidation, validationMode, value]);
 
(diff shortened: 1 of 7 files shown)


packages/react/src/number-field/input/NumberFieldInput.tsx
```diff
@@ -112,7 +112,12 @@ export const NumberFieldInput = React.forwardRef(function NumberFieldInput(
       blockRevalidationRef.current = false;
       return;
     }
-    commitValidation(value, true);
+
+    // Defer the validation until after React has committed the DOM updates
+    // to ensure the hidden input element reflects the latest value.
+    queueMicrotask(() => {
+      commitValidation(value, true);
+    });
```
atomiks: Reading the validity state of the input ref `element.validity.valueMissing` means it needs to be deferred to wait for the parent components to finish rendering. This is because (for NumberField) the `value` on the input is the `inputValue` which is set in the `Root` from `syncFormattedInputValueOnValueChange`

mj12albert (approved):

Issue #2175, opened 2025-06-26
Toolbar.Button loses disabled attribute


dmcweeney:
# Bug report

## Current behavior

Setting disabled=true on Toolbar.Button should result in the disabled flag being written to the html but it doesnt.

## Expected behavior

disabled attribute should be passed down to the html so button is disabled.
When focusableWhenDisabled=false the and disabled===true the tabindex should be set to -1.

## Reproducible example

The following code results in:

```
<Toolbar.Button
    aria-label="Delete"
    className="flex select-none items-center justify-center"
    tabIndex={0}
    onClick={() => {
         logger.debug('Delete button clicked, selectedCount = ', selectedCount);
         deleteSelectedTask();
    }}
    disabled={selectedCount === 0}
    focusableWhenDisabled={false}
>
Delete
</Toolbar.Button>
```

Generated html for disabled button:

`<button data-orientation="horizontal" aria-label="Delete" tabindex="0" class="flex select-none items-center justify-center" type="button" aria-disabled="true" data-disabled="">Delete</button>
`

Generated html for enabled button

`<button data-orientation="horizontal" aria-label="Delete" tabindex="0" class="flex select-none items-center justify-center" type="button" aria-disabled="false">Delete</button>
`

as you can see the disabled attribute is missing from the disabled button but the aria-disabled attribute is correct. 
Also tabINdex=0 making it a tab stop.
onclick is disabled though!.

## Base UI version

1.0.0-beta.0

## Which browser are you using?

Edge/Chrome

## Which OS are you using?

Windows

## Which assistive tech are you using (if applicable)?

Voiceover/JAWS etc.

## Additional context

Provide any additional context that might help us identify the problem and find a solution.

mj12albert:
Need to triple check if this is a bug or expected behavior because there was a test for this: https://github.com/mui/base-ui/blob/master/packages/react/src/toolbar/root/ToolbarRoot.test.tsx#L176-L179

dmcweeney:
@mj12albert thank you... impressive turnaround!

PR #2176, opened 2025-06-26
[toolbar] Set `disabled` attr on toolbar button when `focusableWhenDisabled={false}`

mj12albert:
Preview: https://codesandbox.io/p/sandbox/priceless-einstein-w2qtyj

Fixes https://github.com/mui/base-ui/issues/2175

### DIFF
--- a/packages/react/src/utils/useFocusableWhenDisabled.ts
+++ b/packages/react/src/utils/useFocusableWhenDisabled.ts
@@ -12,6 +12,9 @@ export function useFocusableWhenDisabled(
     isNativeButton,
   } = parameters;
 
+  const isFocusableComposite = composite && focusableWhenDisabled !== false;
+  const isNonFocusableComposite = composite && focusableWhenDisabled === false;
+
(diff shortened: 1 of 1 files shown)


packages/react/src/utils/useFocusableWhenDisabled.ts
```diff
@@ -40,7 +40,11 @@ export function useFocusableWhenDisabled(
       additionalProps['aria-disabled'] = disabled;
     }
 
-    if (isNativeButton && !composite && !focusableWhenDisabled) {
+    if (
+      isNativeButton &&
+      ((!composite && focusableWhenDisabled !== true) ||
+        (composite && focusableWhenDisabled === false))
+    ) {
```
atomiks: I think it's worth extracting these to variables - the `boolean | undefined` value can be a bit confusing

> const focusableWhenDisabled: boolean | undefined
Whether the component should be focusable when disabled. When undefined, composite items are focusable when disabled by default.

PR #2193, opened 2025-06-30
[popover][tooltip] Inline hooks

atomiks:

Refactors away unneeded hook pattern


packages/react/src/popover/positioner/PopoverPositionerContext.ts
```diff
@@ -1,10 +1,21 @@
 'use client';
 import * as React from 'react';
-import { usePopoverPositioner } from './usePopoverPositioner';
+// REMOVE: import { usePopoverPositioner } from './usePopoverPositioner';
```
michaldudak:
```suggestion
```

PR #2221, opened 2025-07-01
[select] Fix scroll-item-into-view in fallback mode after keyboard interaction

atomiks:
(Post-`beta.0` regression fix, so doesn't need to be mentioned in release)

Repro:

1. Open select with long scrollable list such that it enters fallback mode with `alignItemWithTrigger` still enabled
2. Scroll to last item, interact using keyboard
3. Close, re-open select, did not scroll the item into view

This _only_ happens with `alignItemWithTrigger` being in fallback mode, and only after interacting with the keyboard

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useListNavigation.ts
+++ b/packages/react/src/floating-ui-react/hooks/useListNavigation.ts
@@ -412,6 +412,8 @@ export function useListNavigation(
     }
 
     if (activeIndex == null) {
+      forceSyncFocusRef.current = false;
+
       if (selectedIndexRef.current != null) {
         return;
       }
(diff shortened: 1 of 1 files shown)


packages/react/src/floating-ui-react/hooks/useListNavigation.ts
```diff
@@ -451,8 +453,6 @@ export function useListNavigation(
           }
         };
 
-        forceSyncFocusRef.current = false;
```
atomiks, line 454: This assignment was moved down to improve highlight hover perf, but it causes the mentioned bug. So instead, we can make the item mouse listeners be responsible for making the focus sync, which retains the perf improvement while addressing this issue. They still guard `focusItemOnHover`, and only let the refs be assigned — the `forceSyncFocusRef` flag needs to be `false` on first open, making `scrollIntoView` asynchronous with a `requestAnimationFrame`

Issue #2186, opened 2025-06-28
[dialog] portal stays "mounted" when popup is removed


KingSora:
# Bug report

Its a bit hard to explain, but I've created an example here: https://codesandbox.io/p/devbox/pedantic-grass-22zl8v?workspaceId=ws_9E7g8meGhxx9KyReSrgWiy

Steps to reproduce:
- Create a dialog where the `Dialog.Popup` is dynamically rendered depending on the `open`  state
- When the dialog is dismissed and the `Dialog.Popup` is switched out with anything else (in the examle its `null`)
- After the `Dialog.Popup` disappears the `Dialog.Portal` stays around indefinitely

## Current behavior

The `Dialog.Portal` stays around even when the `Dialog.Popup` is gone and thus no transition / animation can occur to unmount the `Dialog.Popup`.

## Expected behavior

The `Dialog.Portal` should not wait for any animations to complete when there is no `Dialog.Popup` to perform any animations.

## Reproducible example

https://codesandbox.io/p/devbox/pedantic-grass-22zl8v?workspaceId=ws_9E7g8meGhxx9KyReSrgWiy

## Base UI version

1.0.0-beta.0

## Which browser are you using?

Firefox 139.0.4 (64-Bit)

## Which OS are you using?

Windows 11

## Additional context

Provide any additional context that might help us identify the problem and find a solution.

atomiks:
Why not conditionally render the whole portal element tree instead?

Also: the devbox is private

KingSora:
@atomiks sorry, the devbox should be available now

> Why not conditionally render the whole portal element tree instead?

To give a bit more context, a dialog in the codebase I'm working in is implemented like this:

```ts
<Dialog content={<DialogPopupImplementation />} />
```

The `DialogPopupImplementation` component internally uses the `Dialog.Popup` element and the `Dialog` component provides the needed `Dialog.Portal` element for it. With this structure the `Dialog` component provides "the needed infrastructure" and its `content` prop is the visual implementation. Some more complex dialogs can change their `content` prop depending on some state (also the `close` state), and thus content can also be just `null` since the `content` prop can be any `ReactNode`.

I recognize that this is an edge case, but I would also expect that a `dialog` is not stuck in this state where I can't interact with the rest of the page (as you can see in the devbox which is now available)

KingSora:
[...]

PR #2233, opened 2025-07-02
Fix: useAnimationsFinished when ref is not present

KingSora:
Fixes: https://github.com/mui/base-ui/issues/2186

Previously the function passed into the `useAnimationsFinished` hook would only run if the passed `ref` value is an element. This PR adapts the hook to also run the function in case the refs value is `falsy`.


atomiks:
Thanks for contributing! The changes seem ok to me, can you add a test?

KingSora:
@atomiks I've added a test which would have failed before.. But its hard to add correct / better units tests because JSOM doesn't support the Web animation api.

atomiks:
@KingSora I think it would be better to test a real component e.g. Dialog instead, and you can also use `.skipIf(isJSDOM)` (imported) to run it in the browser only instead, which is a lot easier

KingSora:
@atomiks I've added tests for the `dialog` component but this issue spans accross more components.. so those need to be adapted as well (I'll take a look later as I have to shift my focus now)

I've also noticed that in the dialogs `'is called on open when there is no enter animation defined'` test the assertion said:
```ts
await waitFor(() => {
  expect(screen.queryByTestId('popup')).not.to.equal(null);
});

expect(onOpenChangeComplete.callCount).to.equal(2);
expect(onOpenChangeComplete.firstCall.args[0]).to.equal(true);
```

Where the call count for the `onOpenChangeComplete` is asserting to `2`, which should be just `1` correct?

atomiks:
> Where the call count for the onOpenChangeComplete is asserting to 2, which should be just 1 correct?

I believe it's related to `globalThis.BASE_UI_ANIMATIONS_DISABLED`, which causes it to be `2` but not `1` in the test env unless it's set to `false` in the test explicitly (it's `1` in a real browser environment). I don't think the call count was correctly updated after this property was added

KingSora:
Alright, updated all the components so far and the tests should run through.. Its a bit hard for me to keep track of "real" failing tests because all the format related tests in `NumberInput` and other components are failing because the default language is not english on my system.

packages/react/src/alert-dialog/popup/AlertDialogPopup.tsx
```diff
@@ -53,22 +52,11 @@ export const AlertDialogPopup = React.forwardRef(function AlertDialogPopup(
...
 
-  useOpenChangeComplete({
-    open,
-    ref: popupRef,
-    onComplete() {
-      if (open) {
-        onOpenChangeComplete?.(true);
-      }
-    },
-  });
```
atomiks, line 70: This can't be moved from this component. The root is unable to handle the `open: true` case (it just ends up firing it instantly instead of waiting for the animation to finish)

atomiks:
Lint errors: 

```
/tmp/base-ui/packages/react/src/select/root/SelectRoot.test.tsx
  699:7  error  Unused eslint-disable directive (no problems were reported from 'testing-library/no-node-access')
  707:7  error  Unused eslint-disable directive (no problems were reported from 'testing-library/no-node-access')

/tmp/base-ui/packages/react/src/tooltip/root/TooltipRoot.test.tsx
  652:7  error  Unused eslint-disable directive (no problems were reported from 'testing-library/no-node-access')
  660:7  error  Unused eslint-disable directive (no problems were reported from 'testing-library/no-node-access')
```

KingSora:
@atomiks after having a look again I must admit that this issue is a bit more involved than I initially thought. I'll close the PR for now since the fix I thought should work isn't working properly.

A few findings / thoughts I would like to share:

1. All tests where the `onOpenComplete` is tested with an open animation instead of an open transition are potentially flawed / inconsistent since animations are canceled when the selector changes so an animation like this:
```css
@keyframes test-anim {
  from {
    opacity: 0;
  }
}

.animation-test-indicator[data-starting-style] {
  animation: test-anim 1s;
}
```
Would only run for a few frames since the `data-starting-style` is removed pretty quickly (here is the [sandbox](https://codesandbox.io/p/devbox/blazing-pond-qxy624?workspaceId=ws_9E7g8meGhxx9KyReSrgWiy)) 
The same does **not** apply for the `data-ending-style` since its sticking around until the element is removed from the DOM.

Since the `useAnimationsFinished` hook is pretty frame dependent I would refactor those tests to use transitions instead.

This is not a bug in general since animations should not be used with the `data-xxxx-style` selector as stated in the docs.

2. If `useOpenChangeComplete` is used inside the `Dialog.Popup` the popup is required to be rendered, otherwise `onOpenComplete` will never be called with `true`. So either this has to be handled somewhere else (if the behavior should be supported) or the `Dialog.Popup` is required to always be rendered.

PR #2237, opened 2025-07-02
[select] Add `value` and `readOnly` to `Select.Trigger` state

### LINKED ISSUE #2196: [select] Select.Trigger.State should have value
# Feature request

## Summary

`Select.Trigger.State` should have value in its state like `Select.Value.State`.

```ts
<Select.Trigger className={({ disabled, valid, value }) => ''} /> 
```

N/A

## Motivation

I wanted to style `Select.Trigger` text in 2 mutually exclusive situation, `disabled` and `value === null`. Can't apply the style on `Select.Value` because it would not have disabled + valid states like `Select.Trigger` does. I do believe, though, that the set of states exposed by both Trigger and Value should be the same and consistent (similar to `Field.Label`, `Field.Description`, `Field.Error` having the same state as `Field.Root`)

FYI, I am using a later version of Base UI which has `Select.Value` API revamped (https://pkg.pr.new/@base-ui-components/react@2192)

atomiks:
Closes #2196

### DIFF
--- a/packages/react/src/select/trigger/SelectTrigger.tsx
+++ b/packages/react/src/select/trigger/SelectTrigger.tsx
@@ -24,6 +24,7 @@ const BOUNDARY_OFFSET = 2;
 const customStyleHookMapping: CustomStyleHookMapping<SelectTrigger.State> = {
   ...pressableTriggerOpenStateMapping,
   ...fieldValidityMapping,
+  value: () => null,
 };
 
 /**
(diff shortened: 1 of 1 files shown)


atomiks:
It seems the `readOnly` state should likely be part of `Field` like `disabled` is

mj12albert:
> It seems the `readOnly` state should likely be part of `Field` like `disabled` is

You think so? Right now `readOnly` isn't explicitly documented for all the components that could support it

mj12albert (approved):

atomiks:
> You think so? Right now readOnly isn't explicitly documented for all the components that could support it

I think it depends if every control can can accept `readonly` like they do `disabled`

mj12albert:
> I think it depends if every control can can accept `readonly` like they do `disabled`

It doesn't work for slider, possibly checkbox and radio https://codepen.io/aardrian/pen/poMXEGd

Tried to add `readOnly` to Slider, but like the native `<input type="range">`, VO can still change the value despite setting `readonly`/`aria-readonly` @atomiks

PR #2241, opened 2025-07-03
[select] Fix `alignItemWithTrigger` fallback scroll jump

atomiks:
Caused by https://github.com/mui/base-ui/pull/2110

Under these conditions:

- Select is scrollable & taller than viewport
- Using mouse (keyboard seems ok)
- Ensure fallback to `alignItemWithTrigger=false` can occur when the trigger is at the very bottom of the viewport
- Click trigger → scroll jumps to top

Changing the state to be local unfortunately changed the timing of how the styles are cleared

Experiment to test: https://deploy-preview-2241--base-ui.netlify.app/experiments/long-select


packages/react/src/select/popup/SelectPopup.tsx
```diff
@@ -317,7 +318,7 @@ export const SelectPopup = React.forwardRef(function SelectPopup(
         return;
       }
 
-      if (reachedMaxHeightRef.current || !alignItemWithTriggerActive) {
+      if (reachedMaxHeightRef.current) {
```
atomiks, line 323: Not relevant to the bug, but there's already an `!alignItemWithTriggerActive` check above so this is needless.

michaldudak (approved):
This seems to work well.

I've noticed a likely unrelated bug - using <kbd>Space</kbd> to open a Select or select an option causes the page to scroll.

PR #2246, opened 2025-07-03
[toast] Make `Toast.Viewport` an announce container

atomiks:

Currently the announcer div lives inside the `Toast.Root` as a visually hidden node that has delayed content (since screen readers don't announce content that's inserted with the status node simultaneously, and macOS Safari needs a 'substantial' delay - 50ms was used previously)

This makes the `Viewport` act as the announcer and has some possible benefits:

- Removes timeout/visually hidden hacks.
- Remove `Dialog` `aria-modal` attribute, and instead rely on the `aria-hidden` technique on "outside" elements, which ignores `aria-live` elements (the status announcer).
- For low priority toasts, this change also does announce other internal content, which might not be relevant however
- For high priority toasts, they're rendered in a visually hidden container next to the viewport announce container, as they can receive `role="alert"` when dynamically rendered and avoid being announced multiple times in the status container


packages/react/src/dialog/popup/useDialogPopup.tsx
```diff
@@ -48,7 +48,6 @@ export function useDialogPopup(parameters: useDialogPopup.Parameters): useDialog
   const popupProps: HTMLProps = {
     'aria-labelledby': titleElementId ?? undefined,
     'aria-describedby': descriptionElementId ?? undefined,
-    'aria-modal': mounted && modal === true ? true : undefined,
     role: 'dialog',
```
atomiks, line 50: https://github.com/floating-ui/floating-ui/issues/3359

`aria-modal` shouldn't be used because it renders outside content inaccessible when they're created after the dialog opens.  `aria-hidden` on "outside" elements ensures content injected after the dialog renders is accessible to screen readers

michaldudak:
The Promise demo announces "close" even twice: "Loading data. Close. Close.". I don't know what screen reader users prefer, but to me announcing buttons in the toast sounds unnecessary.

atomiks:
@michaldudak added `aria-hidden` on `Toast.Close` when the viewport isn't focused
It seems potentially useful for `Toast.Action`, though, but I don't think it's critical.

michaldudak:
Safari on macOS now says "Notifications (F6)" twice when the demo toasts are opened.

atomiks:
> Safari on macOS now says "Notifications (F6)" twice when the demo toasts are opened.

Seems like Chrome doesn't but Firefox and Safari both do. Changed it so it's a static label that doesn't get announced

michaldudak:
After these changes, VoiceOver announces "Notifications (F6)" after pressing F6, which seems pointless to me - since I know how to focus notifications, I don't need a hint.

atomiks:
@michaldudak I don't see that as a problem. Other libraries like Sonner use aria-label on the container. It labels the region that's currently focused which seems fine.

michaldudak:
It's not a problem per se, but it doesn't make a lot of sense to say "F6" right after you pressed it. "Notifications" alone should suffice.

michaldudak (approved):

PR #2248, opened 2025-07-03
[code-infra] Use shared push to argos cli

Janpot:
Align with the other projects


Janpot:
Aside, creating https://github.com/mui/mui-public/issues/392 to deal with the unstable bundle size comparison

Screenshot 2025-07-03 at 12 58 29

brijeshb42 (approved):

PR #2250, opened 2025-07-03
[tooltip] Memoize leftover object in tooltip

sai6855:
This PR adds memoization for an object that was missed in PR #1677.


atomiks:
This value doesn't need memoization as it's made into a ref before being passed to the context value

sai6855:
>  ref before being passed to the context value

Yes, `delay` object is converted to ref inside `FloatingDelayGroup` but the `delay` object here `<FloatingDelayGroup delay={{ open: delay, close: closeDelay }} />` is recreated on every render, which can trigger unnecessary re-renders.

Before memoizing the delay object, we can see in the flamegraph (screenshot 1) that it contributes to the rendering. After memoizing it, the delay object no longer appears in the flamegraph (screenshot 2), indicating reduced render cost.


### Before
image

### After

image

atomiks (approved):

PR #2256, opened 2025-07-04
[composites] Avoid double `useRenderElement` passes

### LINKED ISSUE #2255: [toggle][togglegroup] render function receives only part of the necessary props
# Bug report

## Current behavior

The keyboard navigation for toggle group is broken if I use only those props that render function provides.

## Expected behavior

Props from render function should be enough for the work of keyboard navigation.

## Reproducible example

https://codesandbox.io/p/devbox/damp-wind-j72vfx?workspaceId=ws_8iqbh1gg1GcLonjY38LEti
(just focus on any item and try navigate through keyboard arrows)

## Base UI version

1.0.0-beta.1

Chrome

## Which OS are you using?

Mac OS

Voiceover/JAWS etc.

## Additional context

I spent about two hours to figure out where the problem is. The element that is returned from the function is additionally added with props, such as another ref, some on-handlers and a tabindex. But I can't make a universal, independent and normally typed component because of this.

atomiks:

Fixes #2255 by avoiding two `useRenderElement` passes

### DIFF
--- a/packages/react/src/composite/item/CompositeItem.tsx
+++ b/packages/react/src/composite/item/CompositeItem.tsx
@@ -3,27 +3,45 @@ import * as React from 'react';
 import { useRenderElement } from '../../utils/useRenderElement';
 import { useCompositeItem } from './useCompositeItem';
 import type { BaseUIComponentProps } from '../../utils/types';
+import { EMPTY_OBJECT, EMPTY_ARRAY } from '../../utils/constants';
+import { CustomStyleHookMapping } from '../../utils/getStyleHookProps';
 
 /**
  * @internal
  */
-export function CompositeItem<Metadata>(componentProps: CompositeItem.Props<Metadata>) {
-  const { render, className, itemRef = null, metadata, ...elementProps } = componentProps;
+export function CompositeItem<Metadata, State extends Record<string, any>>(
+  componentProps: CompositeItem.Props<Metadata, State>,
+) {
+  const {
+    render,
+    className,
+    state = EMPTY_OBJECT as State,
+    props = EMPTY_ARRAY,
+    refs = EMPTY_ARRAY,
+    metadata,
+    customStyleHookMapping,
+    tag = 'div',
+    ...elementProps
+  } = componentProps;
 
   const { compositeProps, compositeRef } = useCompositeItem({ metadata });
 
-  return useRenderElement('div', componentProps, {
-    ref: [itemRef, compositeRef],
-    props: [compositeProps, elementProps],
+  return useRenderElement(tag, componentProps, {
+    state,
+    ref: [...refs, compositeRef],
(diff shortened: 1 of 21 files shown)


mj12albert:
Using the `CompositeRoot` component would also have the same double render issue? e.g. https://github.com/mui/base-ui/blob/master/packages/react/src/radio-group/RadioGroup.tsx#L224-L229

atomiks:
@mj12albert great point. I think maybe we need to "defer" the rendering to the `Composite{Root,Item}` components by forwarding the props to it instead of using `useRenderElement` before rendering it. (`CompositeRoot` less easy to inline in each component that renders it compared to the `Item` one)

packages/react/src/composite/root/CompositeRoot.tsx
```diff
@@ -8,16 +8,22 @@ import { useRenderElement } from '../../utils/useRenderElement';
...
  */
-export function CompositeRoot<Metadata extends {}>(componentProps: CompositeRoot.Props<Metadata>) {
+export function CompositeRoot<Metadata extends {}, State extends Record<string, any>>(
+  componentProps: CompositeRoot.Props<Metadata, State>,
+) {
   const {
     render,
     className,
+    refs = EMPTY_ARRAY,
+    props = EMPTY_ARRAY,
```
romgrk, line 24: `null` or `undefined` would be more efficient as default values, array `ref` makes `useRenderElement` use `useForkRefN`.

packages/react/src/composite/item/CompositeItem.tsx
```diff
@@ -3,27 +3,46 @@ import * as React from 'react';
...
-  const { render, className, itemRef = null, metadata, ...elementProps } = componentProps;
+export function CompositeItem<Metadata, State extends Record<string, any>>(
+  componentProps: CompositeItem.Props<Metadata, State>,
+) {
+  const {
+    render,
+    className,
+    state = EMPTY_OBJECT as State,
+    props = EMPTY_ARRAY,
+    refs = EMPTY_ARRAY,
```
romgrk, line 20: It would be nice to standardize the API of the `refs` prop with `useRender` and `useRenderElement`: call it `ref`, and accept an array optionally.

atomiks: Since it's a component, it doesn't work properly (especially in React <19)

mj12albert (approved):
Works 👍 
Tested with toggle and toggle group using the original repro: https://codesandbox.io/p/sandbox/pedantic-silence-6qgh9t

packages/react/src/composite/root/CompositeRoot.tsx
atomiks: They're spread inside the array, so it will always use `useForkRefN` regardless

PR #2258, opened 2025-07-04
[code-infra] Use CI action from mui-public

brijeshb42:


.github/workflows/ci.yml
```diff
@@ -12,6 +12,12 @@ on:
 permissions: {}
 
 jobs:
+  continuous-releases:
```
michaldudak: This actually builds the project for the second time. The original solution reused the build done previously. Also, from what I can see, the shared action has the `--comment=off` flag. We believe these comments are useful and would like to keep them (the deployment info can be included in Danger's or other comment, but it should be somewhere).

Janpot: Agreed, perhaps it would work better as a [composite action](https://docs.github.com/en/actions/tutorials/creating-a-composite-action) with just the publish command and a `comment` input? I'm also fine keeping it as is if the composite action would rely on `code-infra` implicitly installed.

brijeshb42: I had initially added arguments for `pre-build` and `post-build` commands to run in which case for this repo, we would provide `post-build: "pnpm validatedeclarations"` but then went against it to keep things simple.
Currently, both `material-ui` and `mui-x` function similarly to what is in this PR but maybe both can be revisited to reuse same logic.

Janpot (approved):

brijeshb42:
@michaldudak review please.

michaldudak (approved):
LGTM now!

PR #2260, opened 2025-07-04
[useCompositeRoot] Remove unnecessary type assertion in isModifierKey function

sai6855:


packages/react/src/composite/root/useCompositeRoot.ts
```diff
@@ -368,10 +368,10 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
 
 function isModifierKeySet(event: React.KeyboardEvent<any>, ignoredModifierKeys: ModifierKey[]) {
```
michaldudak: It should work without `any` here as well.

```suggestion
function isModifierKeySet(event: React.KeyboardEvent, ignoredModifierKeys: ModifierKey[]) {
```

michaldudak (approved):
:+1:

Issue #2146, opened 2025-06-22
[docs] Clarify difference between dialog and alert dialog


marcospgp:
so far I've only noticed that dialog will let user click outside the dialog to close it while alert dialog won't

would be great if docs were explicit about the difference

michaldudak:
The main reason for separating them is accessibility: see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alertdialog_role.

You're right that we should make it more clear in the docs, though.

ChrisCates:
@michaldudak, is it possible that Dialog and Alert Dialog are unified into the same component, with a prop being passed that can make it an `alert`?

The structure of both is very similar and you can meet ARIA compliance with a prop to simplify the flow.

mj12albert:
Duplicate of https://github.com/mui/base-ui/issues/1687

mj12albert:
> is it possible that Dialog and Alert Dialog are unified into the same component

@ChrisCates We are indeed considering this, though it's not a priority for now, and won't be something that we'll get to for a few months at least 🙏

PR #2265, opened 2025-07-06
[dialog] Unified Dialog <> Alert Dialog Component

ChrisCates:
# Unified Dialog, Alert Dialog Component

As per [this comment](https://github.com/mui/base-ui/issues/2146#issuecomment-3019448033). I decided to work on a PR for a unified `Dialog` <> `AlertDialog` component. Effectively deprecating the current `AlertDialog` component.

## What Changed

1. Added `alert` prop to DialogRootContext
2. `DialogPopup` uses `alert` from dialog context to set proper roles

## Behaviour Changes

1. Setting `alert` forces modal behavior for alert dialogs even if set to `false`
2. Setting `alert` prevents dismissing alert dialogs even if set to `true`
3. `AlertDialog` by default sets `alert` to `true` (reuses Dialog Components)

## Requirements for merge

1. What should happen to the docs? Deprecation notice?
2. Testing? What other tests should I create to ensure behaviour is consistent
4. Anything else I need to update (to meet the standards of the repository)


mj12albert:
@ChrisCates Thanks for the PR, we are still discussing this issue internally but it's not a priority for now so it may be a while before we come back to this 🙏

michaldudak:
We decided not to proceed with this for now. We may revisit it in the next major, but for now, I'm closing the PR.

Issue #2263, opened 2025-07-05
[meter] `aria-valuenow` should not be converted to percentage


brianle1301:
# Bug report

## Current behavior

`Meter` converts user provided `value` to a percentage within `max` and `min` for `aria-valuenow`, as can be seen on the documentation site:

<img width="618" height="106" alt="Image" src="https://github.com/user-attachments/assets/a342d9d0-c5ce-46d1-a6da-b1b813487373" />

This produces a value that is incorrect within the user provided range.

## Expected behavior

The user provided value should be used as is for `aria-valuenow`

## Reproducible example

https://base-ui.com/react/components/meter

## Base UI version

1.0.0-beta.1

## Which browser are you using?

N/A

## Which OS are you using?

N/A

## Which assistive tech are you using (if applicable)?

N/A

## Additional context

This is not a problem on the `Progress` component.

PR #2266, opened 2025-07-07
[meter] Fix aria-valuenow should not be converted to percentage

sai6855:
closes https://github.com/mui/base-ui/issues/2263

### DIFF
--- a/packages/react/src/meter/root/MeterRoot.tsx
+++ b/packages/react/src/meter/root/MeterRoot.tsx
@@ -59,7 +59,7 @@ export const MeterRoot = React.forwardRef(function MeterRoot(
     'aria-labelledby': labelId,
     'aria-valuemax': max,
     'aria-valuemin': min,
-    'aria-valuenow': value,
+    'aria-valuenow': value ?? undefined,
     'aria-valuetext': ariaValuetext,
     role: 'meter',
   };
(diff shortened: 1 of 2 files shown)


packages/react/src/meter/root/MeterRoot.tsx
```diff
@@ -59,7 +59,7 @@ export const MeterRoot = React.forwardRef(function MeterRoot(
     'aria-labelledby': labelId,
     'aria-valuemax': max,
     'aria-valuemin': min,
-    'aria-valuenow': percentageValue / 100,
+    'aria-valuenow': value ?? undefined,
```
sai6855, line 62: updated `aria-valuenow` to similar to `Progress`

https://github.com/mui/base-ui/blob/b5ed091e5ab52b9730d6cfe87646077a9e4099e3/packages/react/src/progress/root/ProgressRoot.tsx#L94

mj12albert:
Thanks @sai6855 but I did more cleanup to meter/progress in https://github.com/mui/base-ui/pull/2267 🙏

packages/react/src/meter/root/MeterRoot.tsx
mj12albert: Meter's `value` is never undefined, since it doesn't have a indeterminate state like Progress

Issue #2261, opened 2025-07-05
[number field] `NumberField.Decrement` and `NumberField.Increment` disabled state is false when the button itself is disabled but `NumberField.Root` is not


brianle1301:
# Bug report

## Current behavior

The disabled state inside the className prop for `NumberField.Decrement` or `NumberField.Increment` only reports the disabled state of `NumberField.Root` as a whole, not when the button itself is disabled (due to min/max constraints)

## Expected behavior

The disabled state for `NumberField.Decrement` or `NumberFied.Increment` should be true either when the field as a whole is disabled or when they are disabled individually due to min/max constraints.

## Reproducible example

```
<NumberField.Root min={0}>
  <NumberField.Group>
    <NumberField.Decrement className={({ disabled /* <- Should be true */ }) => ''} />
    <NumberField.Input />
    <NumberField.Increment />
  </NumberField.Group> 
</NumberField.Root>
```

## Base UI version

1.0.0-beta.1

## Which browser are you using?

N/A

## Which OS are you using?

N/A

## Which assistive tech are you using (if applicable)?

N/A

## Additional context

https://github.com/mui/base-ui/blob/master/packages/react/src/number-field/decrement/NumberFieldDecrement.tsx#L79 should be 

```
state: {
  ...state,
  disabled
}
```

Issue #2245, opened 2025-07-03
[docs] Display `aria-label` for components that should be internationalized


atomiks:
# Docs feedback

There are components like `Toast.Viewport` and `NumberField.Increment` that have default English `aria-label`s internally (in source), but the docs don't show that these should be internationalized. 

It's also possible we should remove the default English labels and defer it to the docs which makes it a lot more clear when looking at the examples that they need to be handled by the dev.

Details on auto ARIA translations: https://adrianroselli.com/2019/11/aria-label-does-not-translate.html#Update06

mj12albert:
Makes sense ~ there could be more strings like these, e.g. `aria-roledescription` in number field

PR #2268, opened 2025-07-07
[numberfield] Fix button disabled state only including root disabled state

mj12albert:
Fixes https://github.com/mui/base-ui/issues/2261


docs/reference/generated/number-field-input.json
```diff
@@ -2,6 +2,12 @@
   "name": "NumberFieldInput",
   "description": "The native input control in the number field.\nRenders an `<input>` element.",
   "props": {
+    "aria-roledescription": {
```
mj12albert, line 5: To document i18n-able strings https://github.com/mui/base-ui/issues/2245#issuecomment-3043606441

michaldudak:
The fix looks OK, but could you add a test to verify it?

docs/reference/generated/number-field-input.json
atomiks: Not sure if they should be documented explicitly, but rather just used in the demos. e.g. some demos do `aria-label="..."`, so we could require all our demos do this for other translatable ARIA attrs

mj12albert:
> The fix looks OK, but could you add a test to verify it?

Added! @michaldudak

michaldudak (approved):

PR #2275, opened 2025-07-08
[popover][dialog] Use `click` event for outside press dismissal

### LINKED ISSUE #2272: [popover] Prevent dismissal when a user moves the cursor back over the floating element
## Relations
- This issue is based on the https://github.com/mui/base-ui/issues/1214, where significant improvements for the touch modality were suggested. 
- Upstream `@floating-ui` discussion - https://github.com/floating-ui/floating-ui/discussions/3361

# Feature request
1. **_Allow dismissal prevention when a user moves the cursor back over the floating element._**
2. Allow configuring the [`outsidePressEvent`](https://floating-ui.com/docs/useDismiss#outsidepressevent) for the underlying `@floating-ui` library. Currently, the `Popover` components is getting immediately dismissed once a user _puts their pointer down_.

## Summary
The popover should close on `mouseUp`, and if a user drags back over the popover's content before releasing the mouse button, that should cancel the action.

## Motivation
In order to meet [Success Criterion 2.5.2 Pointer Cancellation](https://www.w3.org/TR/WCAG21/#pointer-cancellation), we need to adjust the way we handle mouse events for the following elements, because they close on `mouseDown` when users click outside of them. The popover closes on `mouseDown` instead of `mouseUp`. Instead, it should close on full tap or click or on mouse up, and _the action should not happen if they mouse down, but then move the cursor back over the popup._

Mouse users with mobility impairments (or anyone who taps their screen in the wrong place) have no way to cancel the action.

atomiks:

Closes #2272

### DIFF
--- a/packages/react/src/dialog/root/useDialogRoot.ts
+++ b/packages/react/src/dialog/root/useDialogRoot.ts
@@ -101,7 +101,7 @@ export function useDialogRoot(params: useDialogRoot.Parameters): useDialogRoot.R
   const role = useRole(context);
   const click = useClick(context);
   const dismiss = useDismiss(context, {
-    outsidePressEvent: 'mousedown',
+    outsidePressEvent: 'click',
     outsidePress(event) {
       if (event.button !== 0) {
         return false;
--- a/packages/react/src/popover/root/PopoverRoot.tsx
+++ b/packages/react/src/popover/root/PopoverRoot.tsx
@@ -157,7 +157,9 @@ function PopoverRootComponent({ props }: { props: PopoverRoot.Props }) {
   const click = useClick(floatingContext, {
     stickIfOpen,
   });
-  const dismiss = useDismiss(floatingContext);
+  const dismiss = useDismiss(floatingContext, {
+    outsidePressEvent: 'click',
+  });
   const role = useRole(floatingContext);
 
   const { getReferenceProps, getFloatingProps } = useInteractions([hover, click, dismiss, role]);


packages/react/src/utils/constants.ts
```diff
@@ -3,6 +3,7 @@ export const PATIENT_CLICK_THRESHOLD = 500;
 export const DISABLED_TRANSITIONS_STYLE = { style: { transition: 'none' } };
 export const EMPTY_OBJECT = {};
 export const EMPTY_ARRAY = [];
+export const CLICK_TRIGGER_IDENTIFIER = 'data-click-trigger';
```
atomiks: Not sure if we should prefix here (`data-base-ui-*`) - `NavigationMenu.Trigger` also has a similar attribute

packages/react/src/dialog/root/DialogRoot.test.tsx
```diff
@@ -479,19 +479,19 @@ describe('<Dialog.Root />', () => {
     await user.click(nestedButton2);
 
     const backdrops = Array.from(document.querySelectorAll('[role="presentation"]'));
-    await user.click(backdrops[backdrops.length - 1]);
+    fireEvent.click(backdrops[backdrops.length - 1]);
```
michaldudak: Why is this change needed?

packages/react/src/utils/constants.ts
michaldudak: We don't prefix other data attributes. Why would it be necessary here?

packages/react/src/dialog/root/DialogRoot.test.tsx
atomiks: Test fails otherwise

packages/react/src/utils/constants.ts
atomiks: It's sort of an "internal" (non-style hook/state attribute) attr. With the Floating UI fork, we have things like `data-base-ui-focusable` for such attributes.

michaldudak: OK, it makes sense to prefix it, then.

michaldudak (approved):

PR #2278, opened 2025-07-09
[navigation menu] Add `openOnHover` and `closeOnClick` props

### LINKED ISSUE #2254: [Navigation Menu] Support click-only or mixed trigger modes for item activation
# Feature request

## Summary

Add support for activating `NavigationMenu.Item` via **click only**, not just hover or hover+click. Ideally, allow configuring individual triggers to respond to either hover, click, or both. This would provide more control and flexibility in building complex navigation patterns.

For example, some menu items might require explicit user intent (click-only), while others benefit from faster access via hover or combined interaction.

Radix UI's related issue: [radix-ui/primitives#1630](https://github.com/radix-ui/primitives/issues/1630)

## Motivation

I'm building a menu with mixed interaction models — some items should open on **click only**, while others can open on **hover or click**. Currently, `NavigationMenu` doesn't offer fine-grained control over interaction type per trigger.

Supporting this would enable more accessible, predictable navigation behaviors for complex UIs, such as:

- Large dropdowns that shouldn't open on accidental hover
- Menus with nested navigation requiring deliberate activation

atomiks:

Adds two props:
- `openOnHover` to `NavigationMenu.Item` allows disabling the hover-to-open behavior for individual items/triggers
- `closeOnClick` to `NavigationMenu.Link` closes the popup by default when clicking a link, useful when it's in a shared layout across the links, defaulting to uncontrolled behavior

Closes #2254


michaldudak:
When an item has `openOnHover = false`, it can still be opened on hover after another item is opened. This seems OK, but the popup doesn't close on mouseout (it behaves as it was open by clicking).

atomiks:
> This seems OK

This seems unexpected to me; I've changed it so it closes instead when mousing out of the `openOnHover` trigger

colmtuite:
@atomiks Hold off on this please. Not sure I like it.

PR #2280, opened 2025-07-09
[accordion] Destructure `render` prop

atomiks:


romgrk (approved):
It would be neat if we could set up typings on `useRenderElement` to guard against this type of issue.

michaldudak (approved):

Issue #2157, opened 2025-06-24
[`POPOVER`] Is there a way to render the `Popover` oven an existing element in the DOM?


oleksandr-danylchenko:
# Get help

### Context:
- On the page, I have the special element with the `id="special-trigger"`.
- I have an existing `SpecialPopover` component that contains all the styles and content formatting logic.

I want to render the `SpecialPopover` component over the `div#special-trigger`, queried from the DOM using the `querySelector()`.

### Issue:

Unfortunately, the `Popover` doesn't allow me to use anything beyond its own React tree as a trigger. Also, I can't pass the `HTMLElement` as a trigger, too. _As it'll create a duplicate in the DOM._

### Example

Codesandbox - https://codesandbox.io/p/sandbox/external-popover-trigger-forked-g463pp

### Questions
Is there a recommended way to deal with rendering the `Popover` component over an arbitrary element in the DOM, using it as the `Trigger`? That includes forwarding the `Popover.Trigger` props onto that element, including the event listeners.

![Image](https://github.com/user-attachments/assets/400175d8-b1d7-43d4-a299-089d556ab5f0)

oleksandr-danylchenko:
I also tried using a combo with passing the `anchor` to the `Popover.Positioner` and manually handling the events/attributes/state over a queried element. However, it quickly turns into a clumsy pile of code that basically tries to replicate the `Popover.Trigger`. It's especially painful if you have a single styled `Popover` component that tries to handle both `trigger` and `trigger`-less variants.

atomiks:
`tippy` attaches native event listeners while this library primarily uses React props. This means in React, the data (props in this case) flow from top to bottom, unless you synchronize the props with an effect higher in the tree (pretty messy and not performant).

So the `Popover.Root` needs to be higher in the component tree than where the trigger is; this way `Popover.Trigger` can be used in the JSX without an error since it has the data available to it, and Popover.Portal etc can be used elsewhere in the child tree

What's the use case for an external trigger?

oleksandr-danylchenko:
> `tippy` attaches native event listeners while this library primarily uses React props. This means in React, the data (props in this case) flow from top to bottom, unless you synchronize the props with an effect higher in the tree (pretty messy and not performant).

That's right. I tried to implement the mentioned synchronization layer between the native element and the `Popover` rendered via the `React.createPortal`. But you're correct that it turns messy quite quickly. Also, there's no convenient way to forward the `Trigger` props up besides using a bunch of callbacks.

[...]

PR #2283, opened 2025-07-10
[POC] Detached triggers for Dialog

michaldudak:
Proof of concept for Dialog's detached triggers.

Ordinary triggers: https://deploy-preview-2283--base-ui.netlify.app/experiments/dialog/perf-ordinary-triggers
Detached triggers: https://deploy-preview-2283--base-ui.netlify.app/experiments/dialog/perf-detached-triggers

Usage: see docs/src/app/(private)/experiments/dialog/perf-detached-triggers.tsx

The idea of a "detached" trigger is to place it outside of Dialog.Root. Such a trigger references a target dialog by its id (we might allow refs as well). This helps to avoid rendering many Dialog.Root components in scenarios like data lists with actions (as shown in the demos). Many triggers can reuse a single Dialog instance providing different data to it.
This requires an extra Dialog.Provider component that sets up a global context so relation between dialogs and triggers can be established.

This implementation assumes the usage of DetachedTrigger is optional. Dialogs still work the same way using the existing Dialog.Trigger API (without requiring Dialog.Provider).

Detached triggers show significant perf improvements.

The charts display less scripting work needed for detached triggers. Results vary from run to run, but the trend is clear - for the example with 200 rows, scripting time after the `load` event takes about 2x more time with ordinary triggers (~300ms vs ~600ms on 6x CPU slowdown). Also memory usage is significantly lower (peaking at ~15MB for detached and ~24MB for ordinary triggers). This is mostly due to reducing the number of rendered React nodes and contexts.

cc @romgrk 

Screenshot 2025-07-10 113610

Screenshot 2025-07-10 113629

### DIFF
--- a/docs/reference/generated/alert-dialog-root.json
+++ b/docs/reference/generated/alert-dialog-root.json
@@ -19,12 +19,15 @@
       "type": "RefObject<AlertDialog.Root.Actions>",
       "description": "A ref to imperative actions.\n- `unmount`: When specified, the dialog will not be unmounted when closed.\nInstead, the `unmount` function must be called to unmount the dialog manually.\nUseful when the dialog's animation is controlled by an external library."
     },
+    "dialog": {
+      "type": "TypedDialogHandle<any>"
+    },
     "onOpenChangeComplete": {
       "type": "((open: boolean) => void)",
       "description": "Event handler called after any animations complete when the dialog is opened or closed."
     },
     "children": {
-      "type": "ReactNode"
+      "type": "ReactNode | Dialog.Root.ChildRenderFunction<any>"
     }
   },
   "dataAttributes": {},
--- a/docs/reference/generated/dialog-provider.json
+++ b/docs/reference/generated/dialog-provider.json
@@ -1,0 +1,6 @@
+{
+  "name": "DialogProvider",
+  "props": {},
+  "dataAttributes": {},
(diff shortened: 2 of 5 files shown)


atomiks:
Does this API also solve https://github.com/mui/base-ui/issues/2157?

KingSora:
Just as an thought, the keyword `detached` could be used to identify this API / Pattern for all of the sub-components related:

e.g. `Dialog.DetachedRoot`, `Dialog.DetachedTrigger`, `Dialog.createDetachedDialog()` instead of inventing new names like `Dialog.TypedRoot`.

This would only apply if `Dialog.TypedRoot` for example can't be used for any other pattern / API as well.

michaldudak:
The latest version doesn't use any new parts (except for Dialog.Provider) - only Dialog.Root and Dialog.Trigger are used. If a Dialog.Trigger has a `dialog` (naming TBD) prop, it acts as a detached trigger and doesn't have to be placed within the Root.

beeirl:
interesting poc! are you planning to make a detached trigger reactive to the open state?

michaldudak:
Yes, I'm going to. I'm also going to change the implementation significantly based on my learnings from #2336.

michaldudak:
Implemented in https://github.com/mui/base-ui/pull/2974

PR #2303, opened 2025-07-15
refactor: useModernLayoutEffect => useIsoLayoutEffect

romgrk:
Follow-up of #1931, with the consensus on slack regarding the naming for that hook.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,4 +38,4 @@ jobs:
       - run: pnpm validate-declarations
       - name: Publish to pkg-pr-new
         if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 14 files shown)


docs/src/blocks/GoogleAnalytics.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 import * as React from 'react';
 import { useMediaQuery } from '@base-ui-components/react/unstable-use-media-query';
-import { useModernLayoutEffect } from '@base-ui-components/utils/useModernLayoutEffect';
+import { useIsoLayoutEffect } from '@base-ui-components/utils/useIsoLayoutEffect';
```
romgrk, line 4: This is going to require a new release of the utils package, which should be `0.1.0`.

PR #2316, opened 2025-07-17
[safePolygon] Check for `performance` existence on server

atomiks:

Mentioned on Discord, some exotic runtimes (not Node.js/CF Workers) don't support `performance` and need a polyfill. It's not necessary to use here and can be replaced with `Date.now()`.

### DIFF
--- a/packages/react/src/floating-ui-react/safePolygon.ts
+++ b/packages/react/src/floating-ui-react/safePolygon.ts
@@ -54,10 +54,10 @@ export function safePolygon(options: SafePolygonOptions = {}) {
   let hasLanded = false;
   let lastX: number | null = null;
   let lastY: number | null = null;
-  let lastCursorTime = Date.now();
+  let lastCursorTime = typeof performance !== 'undefined' ? performance.now() : 0;
 
   function getCursorSpeed(x: number, y: number): number | null {
-    const currentTime = Date.now();
+    const currentTime = performance.now();
     const elapsedTime = currentTime - lastCursorTime;
 
     if (lastX === null || lastY === null || elapsedTime === 0) {
--- a/packages/react/src/floating-ui-react/safePolygon.ts
+++ b/packages/react/src/floating-ui-react/safePolygon.ts
@@ -54,10 +54,10 @@ export function safePolygon(options: SafePolygonOptions = {}) {
   let hasLanded = false;
   let lastX: number | null = null;
   let lastY: number | null = null;
-  let lastCursorTime = performance.now();
+  let lastCursorTime = Date.now();
 
   function getCursorSpeed(x: number, y: number): number | null {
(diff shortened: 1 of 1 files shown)


packages/react/src/floating-ui-react/safePolygon.ts
```diff
@@ -54,10 +54,10 @@ export function safePolygon(options: SafePolygonOptions = {}) {
   let hasLanded = false;
   let lastX: number | null = null;
   let lastY: number | null = null;
-  let lastCursorTime = performance.now();
+  let lastCursorTime = Date.now();
```
romgrk: `Date.now()` is not ideal. It's limited to 1ms precision, so there's less accuracy to calculate the cursor speed. It's also subject to time adjustments, so e.g. if a user is moving their cursor during Daylight Saving Time adjustment, time goes back/forward one hour, so you'd end up with a negative/positive one hour `elapsedTime`. Similar concern for negative leap seconds.

The environments you mention don't seem to be real users, so you could use a fallback like this:

```tsx
const now = typeof performance !== 'undefined' ? performance.now : () => 0;
```

atomiks: > if a user is moving their cursor during Daylight Saving Time adjustment, time goes back/forward one hour

Wild :) and probably extremely rare, but sure. I think in 2017 the precision was intentionally lost due to some kind of exploit that I remember reading way back when. It's now 1-2ms since then.

> The environments you mention don't seem to be real users, so you could use a fallback like this:

Yeah it's client-only as it runs in an event handler, so a basic fallback will work

romgrk: > I think in 2017 the precision was intentionally lost due to some kind of exploit 

Yes, the Spectre bug for Intel CPUs and the equivalent Meltdown for AMD. But the precision has been changed up/down at various points since then, for various reasons (e.g. Firefox anti-fingerprinting measures). The precision could come back for CPUs that are designed without that flaw, and tbh 1ms is overkill, even 100us snapping is enough to defeat the exploit.

But either way, `performance.now()` is designed to be a monotonical clock (only increases, never decreases), whereas `Date.now()` simply follows system time so it can decrease whenever the system time changes, manually or automatically.

romgrk: > even 100us snapping is enough to defeat the exploit

MDN says it's currently set to 5-100us: https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#security_requirements

romgrk: If you want to know too much about time, clocks & programming, I recommend https://zachholman.com/talk/utc-is-enough-for-everyone-right

romgrk (approved):

PR #2318, opened 2025-07-17
[docs] Improve focus outline for header links

Janpot:
Noticed this when working on https://github.com/mui/base-ui/pull/1738

Links have this nice padding:

Screenshot 2025-07-17 at 15 28 00
Screenshot 2025-07-17 at 15 28 13

but the header links look bad
Screenshot 2025-07-17 at 15 28 21

after this PR:
Screenshot 2025-07-17 at 15 29 17

### DIFF
--- a/docs/src/components/HeadingLink.css
+++ b/docs/src/components/HeadingLink.css
@@ -1,4 +1,7 @@
 .HeadingLink {
+  padding: 0.5em;
+  margin-inline: -0.5em;
+
   &:focus-visible {
     outline: 2px solid var(--color-blue);
     outline-offset: -2px;


docs/src/components/HeadingLink.css
```diff
@@ -1,4 +1,7 @@
 .HeadingLink {
+  padding: 0.25em;
```
mj12albert:
```suggestion
  padding: 0.5rem 0.75rem;
```
I think the padding shouldn't be equal on all sides, like the other focus rings you screencapped 😬

mj12albert:
> It's completely unclear to me how I should write styling in this project

Vlad actually wrote a readme here: https://github.com/mui/base-ui/blob/master/docs/src/styles.css#L1-L15

Maybe we should put this in the actual `docs/README.md`

docs/src/components/HeadingLink.css
Janpot: Would it make sense to create some variable for it? e.g. `--focus-outline-padding`, `--focus-outline-color`, `--focus-outline-radius`.

mj12albert (approved):

colmtuite:
The issue with `padding` is that it increases the click area such that you can be hovering empty whitespace, yet click a link. I'm aware this is how other links already work, which I'm not a fan of, but can live with in the sidebar and app header. I'd rather not see it spread to the main content.

One potential solution is to use a pseudo element.

PR #2319, opened 2025-07-17
[useButton] Error when `ref` and `nativeButton` prop don't match

atomiks:

Logs an error in the console if there's a mismatch between the rendered element and `nativeButton` prop. Makes it much  more clear that something is wrong instead of silently failing.

```js
// Errors
<Popover.Trigger render={<div />} />
```

```js
// Errors
<Select.Trigger render={<button />} />
```


packages/utils/src/error.ts
```diff
@@ -0,0 +1,14 @@
+let set: Set<string>;
+if (process.env.NODE_ENV !== 'production') {
+  set = new Set<string>();
+}
+
+export function error(...messages: string[]) {
```
romgrk, line 6: Is the variadic parameter really necessary? Looks like the use-cases could be just string templates.

packages/utils/src/error.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (process.env.NODE_ENV !== 'production') {
+    const messageKey = messages.join(' ');
+    if (!set.has(messageKey)) {
+      set.add(messageKey);
+      console.error(`Base UI: ${messageKey}`);
```
romgrk, line 11: This package is meant to be reused across all the repos, the `Base UI` prefix might need adjusting.

packages/utils/src/error.ts
romgrk, line 1: I wonder if we could find a better solution for this, IIUC, the error messages would stay in the bundle size even though they're never shown.

packages/utils/src/error.ts
atomiks: This is the same as the existing `warn` util, but just calls `error` not `warn`

packages/utils/src/error.ts
atomiks: It matches the `console` APIs

packages/react/src/use-button/useButton.ts
```diff
@@ -2,17 +2,19 @@
...
 import { useFocusableWhenDisabled } from '../utils/useFocusableWhenDisabled';
 
-export function useButton(parameters: useButton.Parameters = {}): useButton.ReturnValue {
+export function useButton(parameters: useButton.Parameters): useButton.ReturnValue {
   const {
     disabled = false,
     focusableWhenDisabled,
     tabIndex = 0,
     native: isNativeButton = true,
+    name,
```
romgrk: If this is really an error like the `error()` name suggests, then I would advise to simply throw an `Error`. It will let the user know this needs to be fixed, and React will tell them what's the component stack. It could throw in devmode and do nothing in prod mode, that would void the need for the `name` parameter. Strictly enforcing requirements is a good thing, we can always become more lax in the future if this causes a problem.

Otherwise, if this is a warning, knowing that this is for devmode only, have you tried fetching the component name from `new Error().stack` instead? It would avoid polluting all the components calling `useButton` with this concern, and the runtime cost of making those button param objects bigger.

atomiks: I'm not sure if it should throw an error - it certainly prevents things from working correctly but it doesn't break their app and can be fixed gradually. FWIW, the `useControlled` hook also has a `name` param and uses `console.error` as well when the controlled mode switches.

romgrk: > it certainly prevents things from working correctly

That sounds like an error. We shouldn't aim to tolerate user errors if there's no way to provide a working behavior. If it's user mistake, let them fix it now rather than let those errors accumulate. Anyway, if they're running a sensible test suite, console errors will be a failure. We might as well use the existing mechanism (throwing), because that mechanism is well integrated with the rest of the stack. For example, in dev mode React will print a useful component stack trace (which we can't print), and in prod mode they can catch errors with Error boundaries. Console messages are a subpar interface. There is value in being strict, both for us and our users, and we'll avoid the overhead (verbosity, runtime, bundle-size) of adding debugging symbols throughout our hooks.

atomiks: But things like `useControlled`, React `key` in a list, `Suspense` errors, calling `flushSync` inside `useLayoutEffect` etc. use `console.error`, which seem similar to this.

I took throwing errors as more like runtime errors will occur if you don't (e.g. consuming contexts/placing component parts in the wrong place), and there's a need to "catch" them (such as in promises). Whereas this is more of a passive error, and can be fixed gradually. Otherwise, breaking the whole app feels a bit overly aggressive. `console.error` component stack works well for the other errors already in React.

Just to be clear: it prevents the button from working with keyboard interaction (like `<div onClick>`), but works for pointer usage. It seems less like a throw due to this.

romgrk: > and can be fixed gradually

There is no need to fix gradually if the error is thrown. Errors won't accumulate in the first place, they're fixed on the spot when new code is written. Having to create follow-up PRs to fix accumulated errors is a worst DX for our users.

> Otherwise, breaking the whole app feels a bit overly aggressive.

We're not breaking the whole app, we're throwing an error. If the error is recoverable, Error boundaries exist for a reason. If it's not recoverable, then a console message would let the interface be silently broken. Imagine an end-user fills a whole form, just to find out that the submit button silently does nothing. That's a big UX/a11y issue. This is an error and I'm convinced that we need to fail early in this case.

atomiks: > There is no need to fix gradually if the error is thrown. 

When I say "gradually" I mean their entire app tree doesn't blow up as the result of this which is very disorienting, they'll fix it quickly, but while still viewing the app. Throwing makes sense for runtime/context errors but not this, imo. (fwiw both Claude and o3 agree that it should be `console.error`.)

I don't see devs tolerating console.errors and shipping an app with them present (Next.js shows a red banner, and as you mention, tests usually fail).

@michaldudak curious what you think here

---

The component stacks seem to show the same information - you need to expand the chevron in both cases in the demo code (which has an `ErrorBoundary`). Adding the name up front looks clearer at first glance without expanding.

`console.error()`
Screenshot 2025-07-23 at 8 17 13 am

`throw new Error()`
Screenshot 2025-07-23 at 8 17 38 am

romgrk: > When I say "gradually" I mean their entire app tree doesn't blow up as the result of this

Alright, DX seems better that way then. It still leaves the production case unsolved. We prevent users from using Error boundaries, and a clearly broken app is better than a silently broken one, as in the use-case of a form with a failing submit button.

> Adding the name up front looks clearer at first glance without expanding.

Yes but we're adding verbosity & bundle-size throughout the codebase just for that debugging symbol. I think we should at least get rid of that. As I mentioned before, if we really want to show it (in dev mode) we should try to fetch it from either `new Error().stack` or from react internals, that way we avoid shipping unused bytes.

atomiks: `new Error().stack` is a very long string but shows `useButton.useEffect` first, not the component as the first part, so it doesn't seem easily usable?

> "Error
    at useButton.useEffect (webpack-internal:///(app-pages-browser)/../packages/react/src/use-button/useButton.ts:49:46) ...

romgrk: It's a newline separated stack trace, I would expect the correct name to be after the `useButton` line, given that we're trying to get the symbol for the caller of `useButton`.

image

edit: and via react19 internals:
image

Anyway, it seems to me that there are less expensive ways to get the currently rendering component name than adding that param throughout our codebase.

michaldudak: I dislike the `name` parameter as well (both here and in `useControlled`). If there's other reliable way of getting the component name, IMO we should go for it.

packages/utils/src/error.ts
michaldudak: Aren't bundlers removing the whole `if (process.env.NODE_ENV !== 'production') {` blocks when building for production?

romgrk: That block is inside the `error()` function. The outside calls (`error('a long message here...')`) are still in the bundle, they call an empty `error()` function but I don't know that bundlers remove empty functions.

packages/react/src/use-button/useButton.ts
romgrk: I just realized that [with the implementation of `error()`](https://github.com/mui/base-ui/pull/2319/files#diff-29245811f5b029c9d81b5c1dd83267f3f577b97cc579ae20da2a4810458ab521) (it's a no-op in prod), in production mode there is absolutely no error message (and yet all the error-message bytes are in the bundle-size!), so the interface would be silently failing with no way for users to collect that failure other than by end-users reporting the issue. This approach needs to be reworked.

atomiks: It's probably not necessary to use `process.env` check _inside_ `error` as you need to guard it at the callsite anyway (which is being done.)

There's no useful info in `new Error().stack` when I check. Reaching into React internals, even in dev-only mode, doesn't feel great to me and I would rather avoid that (and imo it seems like it could potentially reference the incorrect name if they're wrapping the Base UI components with their own). The `name` param is predictable and guaranteed to be correct without any hacks. Maybe we could guard adding the `name` property to the object options with `process.env` but I feel like the strings are tiny enough that this isn't going to make a notable difference at all. I think https://github.com/mui/base-ui/pull/1463 didn't even have much impact and that minifies much larger strings that are more common.

romgrk: > There's no useful info in new Error().stack when I check.

Right, presumably because you're checking inside `useEffect`'s callback. You need to figure out the name when the call stack is set up, to use the example in `useButton`:

```diff
  if (process.env.NODE_ENV !== 'production') {
+   const name = figureOutNameFromErrorStackOrReactInternals();
    // eslint-disable-next-line react-hooks/rules-of-hooks
    React.useEffect(() => {
```

> and imo it seems like it could potentially reference the incorrect name if they're wrapping the Base UI components with their own

I cannot imagine a case where that affects any of the solutions above. The JS call stack works predictably, and the currently rendering component in react internals always refers to what's actively being rendered, regardless if there is a wrapping component around it. If you think about it, a "wrapping component'" is nothing more than a parent component, and components always have a parent (except the root one). If you have doubts about this, please describe a scenario that would cause an issue.

> but I feel like the strings are tiny enough that this isn't going to make a notable difference at all

My main concern is that we're polluting around 50 files with this concern in this PR, and presumably this means we'd give this approach the green light so we'd be inserting similar debugging code throughout our codebase. Ideally, debugging code does not affect production code. It's not just a bundle-size concern, it's a maintainability concern. And this is for information that is already available by just expanding the error details.

atomiks: In my latest commit, I found stack traces aren't all the same between browsers. Safari/Firefox are different to Chrome. Even if parsing `at` vs. `@`, the name is at index 1 in Chrome but index 2 in Safari/Firefox with the default `\n` split.

There's also the concern that browsers change the formatting of their stack traces between versions, meaning the error message can break at random. `name` on the other hand always works.

Lastly, the parsing is potentially expensive inside buttons like `Select.Item` and could slow down large selects in dev mode (even if it doesn't affect end users, devs could complain)

```js
    const stackLines = new Error().stack?.split('\n');

    let stackComponentName = '';
    if (stackLines) {
      const stackLine = stackLines[1];
      if (stackLine) {
        // Chrome
        if (stackLine.includes('at')) {
          const match = stackLine.match(/at (\w+)/);
          if (match && match[1]) {
            stackComponentName = match[1];
          }
        } else {
          // Safari/Firefox
          stackComponentName = stackLine.split('@')[0];
        }
      }
    }
    const componentName = stackComponentName ?? '';
```

romgrk: Why are we introducing that much complexity to save maybe one click on a DX use-case that's overall somewhat rare? I really don't see how it's worth it. The data is already there in the error details. Just make the message generic and avoid all of this.

atomiks: Alright, sure, let's just remove it. Since the stack trace reveals the name, it's not a big deal.

packages/react/src/number-field/decrement/NumberFieldDecrement.tsx
```diff
@@ -87,5 +94,13 @@ export const NumberFieldDecrement = React.forwardRef(function NumberFieldDecreme
...
-  export interface Props extends BaseUIComponentProps<'button', State> {}
+  export interface Props extends BaseUIComponentProps<'button', State> {
+    /**
+     * Whether the component renders a native `<button>` element when replacing it
+     * via the `render` prop.
+     * Set to `false` if the rendered element is not a button (e.g. `<div>`).
+     * @default true
+     */
+    nativeButton?: boolean;
+  }
```
romgrk: Placing the declaration in a shared interface would avoid a lot of repetition and manual work if there are adjustments to make to the typing or jsdoc.

atomiks: Yeah, good idea - will update

PR #2324, opened 2025-07-18
[code-infra] Demonstrate type portability problem with namespaces

Janpot:
Demo for https://github.com/mui/joy-ui/pull/175#discussion_r2214072945

The type namespace seems to cause issues in creating declaration files for joy. (And will ofcourse for anyone else trying to build a library on top of base)

relevant typescript config:

```
"skipLibCheck": false,
"declaration": true,
```

The `Checkbox.Root` namespace prevents typescript from associating the `Props` type with an actual declaration file. Causing the issues described by @DiegoAndai in joy-ui. This PR demonstrates the exact problem:

1. Run `pnpm release:build && pnpm -F ./test/public-types release:test`
2. Check that the build fails with:

    ```
    index.tsx:4:14 - error TS2742: The inferred type of 'MyCheckbox' cannot be named without a reference to './node_modules/@base-ui-components/react/cjs/checkbox/root/CheckboxRoot'. This is likely not portable. A type annotation is necessary.
    
    4 export const MyCheckbox = React.forwardRef<HTMLButtonElement, CheckboxFieldProps>(() => {
                   ~~~~~~~~~~
    ```
4. Uncomment the export in packages/react/src/checkbox/index.ts
5. Run again `pnpm release:build && pnpm -F ./test/public-types release:test`
6. Check the build succeeds

The other solution is to make library authors explicitly type their exports. But that can become very cumbersome. We should offer a better DX than that.

By exporting a type from a file instead of using a namespace we give typescript the ability to associate it with a declaration file.

Proposal: Let's change how types are exported. You can optionally keep the namespace if you want. The test setup in this repo can also serve as the basis for integration testing the behavior for the future.


test/public-types/checkbox-field.ts
```diff
@@ -0,0 +1,6 @@
+// ./checkbox-field
+import { Checkbox } from '@base-ui-components/react/checkbox';
```
flaviendelangle, line 2: I guess we have the same problem if we do:


```suggestion
import { type Checkbox } from '@base-ui-components/react/checkbox';
```

Janpot: Yes, I don't think that really influences how typescript generates declaration files.

flaviendelangle:
:cry: I really like how clean the exports are with the namespace approach
It would be a shame if we had to drop it

Janpot:
> It would be a shame if we had to drop it

Well,... you don't have to drop it if you don't want to. This PR doesn't drop it. You just need to also export those types from a public module instead of a namespace.

packages/react/src/checkbox/index.ts
```diff
@@ -1 +1,3 @@
 export * as Checkbox from './index.parts';
+
+// export { CheckboxRootProps } from './root/CheckboxRoot';
```
atomiks, line 3: This seems to match this fix:

https://github.com/typescript-eslint/typescript-eslint/pull/7690/files

Consumers can still use `Checkbox.Root.Props` on their wrapped components without issues, right?

Janpot: Yes, that's essentially what this PR demonstrates:

* define the interfaces top-level
* re-export them from the namespace
* re-export them from a public module (and namespace them in the export name to create base-wide unique name)

consumers can choose to either use the namespaced one or the top-level one, it would be the same.

It does seem to have a slight effect on the API docs though looking at CI

Janpot: btw, I tried the other way around

```tsx
// packages/react/src/checkbox/index.ts
import * as Checkbox from './index.parts';

export { Checkbox };
export type CheckboxRootProps = Checkbox.Root.Props;
```

but that doesn't seem to work

michaldudak:
Closing the PR as the issue has been solved in #2912

PR #2327, opened 2025-07-19
[useAnimationsFinished] Use second rAF instead of timeout

atomiks:

Fixes `open: true` transitions not being detected (`data-starting-style` hasn't yet been removed) in Safari and sometimes in Firefox.

### DIFF
--- a/packages/react/src/utils/useAnimationsFinished.ts
+++ b/packages/react/src/utils/useAnimationsFinished.ts
@@ -60,7 +60,7 @@ export function useAnimationsFinished(
 
           // `open: true` animations need to wait for the next tick to be detected
           if (waitForNextTick) {
-            timeout.start(0, exec);
+            frame.request(exec);
           } else {
             exec();
           }


packages/react/src/utils/useAnimationsFinished.ts
```diff
@@ -60,7 +60,7 @@ export function useAnimationsFinished(
 
           // `open: true` animations need to wait for the next tick to be detected
           if (waitForNextTick) {
-            timeout.start(0, exec);
```
romgrk, line 63: Should the `timeout` instance be removed then?

atomiks:
On second look, this needs a more robust method to detect `open: true`. Safari works fine but it seems like it rarely still fires instantly in FF.

It should detect when `transitionStatus: 'starting'` gets removed before `element.getAnimations()` gets checked - so there'll need to be some orchestration between `useAnimationsFinished`/`onOpenChangeComplete` and `useTransitionStatus`

PR #991, opened 2024-12-09
[Menu, Select] Add the `modal` prop

michaldudak:
- Added the `modal` prop to Menu and Select
- Refactored the Menu to be more in line with Select regarding click-drag-release support

Closes #1078

PR #2331, opened 2025-07-20
[floating-ui] Enhance `useDismiss` to check for node ancestors before dismissing

### LINKED ISSUE #2330: [popover] Does not close on outside click when nested
# Bug report

When a Popover is nested in another one, it can't be closed on outside click

## Reproducible example

https://codesandbox.io/p/sandbox/dklnw3

https://github.com/user-attachments/assets/16471b34-0e56-44fc-9a6c-0728a920f624

## Base UI version

v1.0.0-beta.1

Firefox

## Which OS are you using?

Mac OS

baptisteArno:
Fixes #2330

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useDismiss.ts
+++ b/packages/react/src/floating-ui-react/hooks/useDismiss.ts
@@ -25,6 +25,7 @@ import {
 import { useFloatingTree } from '../components/FloatingTree';
 import type { ElementProps, FloatingRootContext } from '../types';
 import { createAttribute } from '../utils/createAttribute';
+import { getNodeAncestors } from '../utils/nodes';
 
 const bubbleHandlerKeys = {
   pointerdown: 'onPointerDown',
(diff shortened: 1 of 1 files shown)


atomiks:
Thanks for the contribution - while this fix can work, it seems this line below (which was added last December) is what prevents the parent in the tree from being marked with `data-base-ui-inert`. This makes it think it's an element injected after the child rendered, and so it doesn't close when clicking the parent incorrectly.

https://github.com/mui/base-ui/blob/a2e10fa110733fad39d2003a7b9ba18421bd9317/packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx#L507-L508

@michaldudak was this line necessary due to the old implementation of `modal` with the `pointer-events` style, or something else? I wonder if we can remove this line entirely which also fixes this issue.  It seems marking the parent with `data-base-ui-inert` should be happening.

ref: https://github.com/mui/base-ui/pull/991#issuecomment-2541412049

michaldudak:
We used to have a bug that allowed two submenus to be opened at once using a mouse. I don't see it happening now with this PR, so we must have fixed it another way as well.

PR #2333, opened 2025-07-21
[select] Fix multiple ARIA behavior on touch

atomiks:
Testing `multiple` prop on iOS:

- `onMouseLeave` fires after selecting an item which moves the virtual cursor to the start of the list incorrectly, causing disorientation
- `selected` fails to get announced. `active && selected` was used previously with roving focus to fix VoiceOver bugs but this appears no longer necessary to have `active` as a guard (all three main browsers work fine on Mac with VoiceOver) 

Closes https://github.com/mui/base-ui/issues/2366


LukasTy (approved):
LGTM, great work. 👍 
Both changes look like very needed fixes for correctness. 👌

PR #2334, opened 2025-07-21
[popups] Improve outside press behavior with touch input

### LINKED ISSUE #1214: [Popover, Menu, Select] Improve outside press using touch modality

atomiks:
Closes #1214 

Significantly improves the UX of dismissing popups with touch input. Scroll is not locked on touch for components that use `sloppy`, so you can scroll away to dismiss it like iOS' native select popup. It no longer feels overly restrictive.

Extension to https://github.com/mui/base-ui/pull/2275. `click` is now `intentional` in this PR, which `Popover` + `Dialog` use (except Popover uses `sloppy` on touch). Other popup components use `sloppy`.

### DIFF
--- a/packages/react/src/dialog/root/useDialogRoot.ts
+++ b/packages/react/src/dialog/root/useDialogRoot.ts
@@ -107,7 +107,7 @@ export function useDialogRoot(params: useDialogRoot.Parameters): useDialogRoot.R
   const role = useRole(context);
   const click = useClick(context);
   const dismiss = useDismiss(context, {
-    outsidePressEvent: 'up',
+    outsidePressEvent: 'intentional',
     outsidePress(event) {
       if (event.button !== 0) {
         return false;
--- a/packages/react/src/floating-ui-react/hooks/useDismiss.test.tsx
+++ b/packages/react/src/floating-ui-react/hooks/useDismiss.test.tsx
@@ -898,9 +898,9 @@ describe.skipIf(!isJSDOM)('useDismiss', () => {
     });
   });
 
-  describe('outsidePressEvent up', () => {
+  describe('outsidePressEvent: intentional', () => {
     test('dragging outside the floating element does not close', () => {
-      render(<App outsidePressEvent="up" />);
+      render(<App outsidePressEvent="intentional" />);
       const floatingEl = screen.getByRole('tooltip');
       fireEvent.mouseDown(floatingEl);
       fireEvent.mouseUp(document.body);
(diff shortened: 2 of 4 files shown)


michaldudak:
Scrolling to dismiss a popover in the hero demo doesn't work on iOS. 

> Other popup components use down.

Menu and Select also seem to close on pointerup, not pointerdown

atomiks:
> Scrolling to dismiss a popover in the hero demo doesn't work on iOS.

Yeah it uses `up` (due to #2272) but maybe on touch it should still use `down`?

> Menu and Select also seem to close on pointerup, not pointerdown

That's intentional for touch, but maybe the `up`/`down` terms are confusing. (There is a JSDoc description, though)

michaldudak:
> Yeah it uses up (due to https://github.com/mui/base-ui/issues/2272) but maybe on touch it should still use down?

Native iOS popovers seem to close on `up` or scroll.

IMO it's worth explaining what "special iOS-like touch handling" means.

atomiks:
I think `down`/`up` are just the wrong terminology to encompass touch + mouse. 

What this really is is anchored vs non-anchored (with a dimming backdrop) outside press behavior. Dialogs and drawers should require an _intentional_ outside press, because you can't scroll away from them on touch input — they're truly modal-like and their backdrop obscures content behind them. While anchored popups like Popover, Menu, and Select don't require an _intentional_ outside press, allowing you to scroll away  (with the special iOS like handling).

I don't think the names matter much because it's not a public API but `intentional`/`sloppy` might be ok.

`mouse`:
- `intentional`: it dismisses on `mouseup`, allowing you to cancel the dismissal if you accidentally pressed outside or changed your mind
- `sloppy`: it dismisses on `mousedown`, hiding immediately when pressing down

`touch`:
- `intentional`: it dismisses on `touchend` but only if a certain number of touchmoves weren't fired between start and end
- `sloppy`: it fires on `touchend`, but only if you held for under 1 second when tapping. It also dismisses when scrolling away before a `touchend`.

`Popover` complicates this because it uses `intentional` on `mouse` but `sloppy` on `touch` (the `>1`s guard allows sloppy to be used, I believe)

michaldudak (approved):

PR #2336, opened 2025-07-21
[POC] Detached triggers for Popovers

michaldudak:


packages/react/src/popover/root/PopoverRoot.tsx
```diff
@@ -26,27 +24,52 @@ import {
...
   } = props;
 
-  const [instantType, setInstantType] = React.useState<'dismiss' | 'click'>();
-  const [titleId, setTitleId] = React.useState<string>();
-  const [descriptionId, setDescriptionId] = React.useState<string>();
-  const [triggerElement, setTriggerElement] = React.useState<Element | null>(null);
-  const [positionerElement, setPositionerElement] = React.useState<HTMLElement | null>(null);
-  const [openReason, setOpenReason] = React.useState<PopoverOpenChangeReason | null>(null);
-  const [stickIfOpen, setStickIfOpen] = React.useState(true);
+  const store = useLazyRef((localHandle) => {
```
romgrk: Tbh the first param of `useLazyRef` is there to avoid inline closures (and the resulting memory allocation on each render), e.g. `useLazyRef(initialize, param)`. If you're using an inline closure, you might as well pass the variable into the closure directly.

atomiks:
Nice work coming up with an API for this.

Checking experiments and noticing the animations need some work:

1. `Popover.Arrow` is clipped while transitioning (due to temporary `transform` style)
2. Content can collapse onto multiple lines
3. The inner content element should have the padding, not the popup, so there isn't clipping whitespace
4. The arrow doesn't transition
5. The Positioner can "jump" instead of transitioning, most noticeable when going from 3 -> 1 at the end

https://github.com/user-attachments/assets/4ae3bda5-f3e1-468b-94a2-2d2ce6cd1371

It was quite tricky getting NavigationMenu to be smooth here

packages/react/src/popover/store.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    this.set('triggers', new Map(triggers));
+  }
+
+  unregisterTrigger(triggerElement: HTMLElement) {
+    const triggers = this.state.triggers;
+    triggers.delete(triggerElement);
+
+    this.set('triggers', new Map(triggers));
+  }
```
romgrk: To avoid weird bugs, it would be preferable to create the new map before modifying it, rather than modifying the old one.
```suggestion
  registerTrigger(triggerElement: HTMLElement, getPayload?: () => Payload) {
    const triggers = new Map(this.state.triggers);
    triggers.set(triggerElement, getPayload);
    this.set('triggers', triggers);
  }

  unregisterTrigger(triggerElement: HTMLElement) {
    const triggers = new Map(this.state.triggers);
    triggers.delete(triggerElement);
    this.set('triggers', triggers);
  }
```

packages/utils/src/usePrevious.ts
romgrk, line 1: It could be implemented similarly to [this one](https://github.com/mui/base-ui/pull/2105#discussion_r2280642694).

The naming would be a concern, those two hooks have similar names but different behaviors. Maybe `usePreviousRender` vs `usePreviousValue` could be clear enough.

packages/react/src/popover/store.ts
michaldudak: Good point, thanks. Updated.

michaldudak:
> The Positioner can "jump" instead of transitioning, most noticeable when going from 3 -> 1 at the end

@atomiks FYI, we had a bug in `useAnimationsFinished` that caused this. When the target value of the animated property is changed in the meantime (in this case `left` is updated as `width` changes), the current animation is aborted and a new one is immediately set up. However, our code didn't see these new animations and ultimately set `instant = true`, that canceled the new animation. If you had to do any hacks to work around this in Navigation Menu, you should be able to remove them after this lands on master.

See https://github.com/mui/base-ui/pull/2336/files#diff-391ac91919dc0519485d440c415899b7c7c95296b05a5161487094a82151a60e for a fix.

packages/react/src/popover/viewport/PopoverViewport.tsx
```diff
@@ -0,0 +1,219 @@
...
+    height: number;
+  } | null>(null);
+
+  // Capture children HTML in a ref when not transitioning.
+  // We can't simply store previous `children` React node, as it might be stateful and doing so would lose this state.
+  useIsoLayoutEffect(() => {
+    if (currentContentRef.current && !previousChildrenHtml) {
+      capturedHtmlRef.current = currentContentRef.current.innerHTML;
+    }
+  });
```
atomiks, line 83: `innerHTML` can be dangerous 😬  is there a way to avoid this?

packages/react/src/popover/trigger/PopoverTrigger.tsx
```diff
@@ -26,17 +33,99 @@ export const PopoverTrigger = React.forwardRef(function PopoverTrigger(
...
+      store.registerTrigger(element, getPayload);
+      setTriggerElement(element);
+
+      return () => {
+        store.unregisterTrigger(element);
+        setTriggerElement(null);
+      };
+    },
+    [getPayload, store],
+  );
```
atomiks, line 147: Cleanup refs don't work in React <=18

packages/react/src/popover/viewport/PopoverViewport.tsx
michaldudak: I replaced this code with capturing the actual DOM node and cloning it into the `previous` container. Should also be faster than serializing and deserializing to HTML.

packages/react/src/popover/trigger/PopoverTrigger.tsx
michaldudak: 🤦

packages/utils/src/usePrevious.ts
michaldudak: I called it `usePreviousRenderValue` (https://github.com/mui/base-ui/pull/2336/commits/246dd66247a7adbf5d90c23ef3dc04f37eb7c233)

packages/utils/src/usePreviousRenderValue.ts
```diff
@@ -0,0 +1,30 @@
...
+export function usePreviousRenderValue<T>(value: T): T | null {
+  const previous = useRefWithInit(createPreviousRef, value).current;
+
+  previous.next = value;
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useIsoLayoutEffect(previous.effect);
+
+  return previous.current;
+}
```
atomiks: Reading refs in render is a problem

With combobox I took a different approach, though I'm not sure if it can apply to your case:

```js
export function useValueChanged<T>(
  valueRef: React.RefObject<T>,
  value: T,
  onChangeParam: (previousValue: T) => void,
) {
  const onChange = useEventCallback(onChangeParam);

  useIsoLayoutEffect(() => {
    if (valueRef.current === value) {
      return;
    }

    onChange(valueRef.current);
  }, [valueRef, value, onChange]);

  useIsoLayoutEffect(() => {
    valueRef.current = value;
  }, [valueRef, value]);
}
```

```js
useValueChanged(queryRef, query, (previous) => {
  // ...
});
```

docs/src/app/(public)/(content)/react/components/popover/page.mdx
```diff
@@ -35,5 +37,103 @@ import { Popover } from '@base-ui-components/react/popover';
...
+To animate the position, apply CSS transitions to the `left`, `right`, `top`, and `bottom` properties of the **Positioner**.
+Target the **Popup's** `width` and `height` properties to animate dimensions.
+
+The popover also supports transitions between content.
+Wrap the content you wish to transition in the `Viewport` part and apply the transition styles to it.
+The `Viewport` part is only required if you want to animate the content.
+It renders an element (`div` by default) with a `data-activation-direction` attribute set to `left` | `right` | `down` | `up` and a children divs marked with `data-current` (when no transition is active) or `data-next` and `data-previous`.
+The `data-activation-direction` attribute specifies the position of the new trigger relative to the previous one, which can be used to apply different styles based on the direction of the transition.
+
+<Demo path="./demos/detached-triggers-full" />
```
atomiks, line 161: This demo's transition styles (timing function, duration, etc.) should ideally closely match Navigation Menu - imo the spring effect looks a bit odd in certain cases.

docs/src/app/(public)/(content)/react/components/popover/page.mdx
```diff
@@ -35,5 +37,103 @@ import { Popover } from '@base-ui-components/react/popover';
...
+
+Since the popup part is reused between triggers, its position and dimensions can be animated.
+To animate the position, apply CSS transitions to the `left`, `right`, `top`, and `bottom` properties of the **Positioner**.
+Target the **Popup's** `width` and `height` properties to animate dimensions.
+
+The popover also supports transitions between content.
+Wrap the content you wish to transition in the `Viewport` part and apply the transition styles to it.
+The `Viewport` part is only required if you want to animate the content.
+It renders an element (`div` by default) with a `data-activation-direction` attribute set to `left` | `right` | `down` | `up` and a children divs marked with `data-current` (when no transition is active) or `data-next` and `data-previous`.
+The `data-activation-direction` attribute specifies the position of the new trigger relative to the previous one, which can be used to apply different styles based on the direction of the transition.
```
atomiks: The docs on detached triggers (and the section above) seems very verbose

docs/src/app/(public)/(content)/react/components/popover/page.mdx
atomiks: Also, the button styles should match the trigger button in the hero and also have a small gap between each.

packages/utils/src/usePreviousRenderValue.ts
michaldudak: Actually, now that I think of it, this value should be kept in state. It's used to affect rendering, so it can't be a ref.

atomiks:
Found a regression for the activation direction calculation under a certain condition. First is master (correct), while second is this PR.

https://github.com/user-attachments/assets/f9c93094-1715-49c6-af2e-f5852027a4d1

I also suppose the animation shouldn't restart when toggling before the animation finishes?

https://github.com/user-attachments/assets/1f32e9fc-f9a5-4e6c-a7fb-da57b8a1670f

michaldudak:
I updated the PR with fixes for the issues you found.

atomiks:
In the (first) detached trigger example, when the popover opens on top due to a collision, it glitches possibly due to a lack of fixed size specified (I had this bug with Navigation Menu at some points, which you mentioned at some point actually)

It doesn't happen for the animation example at the bottom though.

docs/reference/generated/popover-positioner.json
```diff
@@ -122,6 +122,12 @@
       "description": "The available width between the trigger and the edge of the viewport.",
       "type": "number"
     },
+    "--popover-height": {
+      "description": "The height of the popover.\nIt is important to set `height` to this value when using CSS to animate size changes."
+    },
+    "--popover-width": {
+      "description": "The width of the popover.\nIt is important to set `width` to this value when using CSS to animate size changes."
+    },
```
atomiks: `NavigationMenu` calls this `positioner-width/height`

docs/reference/generated/popover-root.json
```diff
@@ -34,27 +38,15 @@
...
-    },
-    "closeDelay": {
-      "type": "number",
-      "default": "0",
-      "description": "How long to wait before closing the popover that was opened on hover.\nSpecified in milliseconds.\n\nRequires the `openOnHover` prop.",
-      "detailedType": "number | undefined"
+    "triggerId": {
+      "type": "string | null",
+      "description": "Id of the trigger that the popover is associated with.\nThis is useful in conjuntion with the `open` prop to create a controlled popover.\nThere's no need to specify this prop when the popover is uncontrolled (i.e. when the `open` prop is not set).",
+      "detailedType": "string | null | undefined"
```
atomiks: `ID`

docs/src/app/(private)/experiments/motion.tsx
```diff
@@ -59,10 +59,17 @@ function AlwaysMounted() {
...
 
   return (
-    <Popover.Root open={open} onOpenChange={setOpen} actionsRef={actionsRef}>
+    <Popover.Root
+      open={open}
+      onOpenChange={(nextOpen, eventDetails) => {
+        setOpen(nextOpen);
+        eventDetails.preventUnmountOnClose();
+      }}
+      actionsRef={actionsRef}
```
atomiks, line 71: Can you explain this special method a bit?

packages/react/src/floating-ui-react/hooks/useHover.ts
```diff
@@ -481,37 +521,49 @@ export function useHover(context: FloatingRootContext, props: UseHoverProps = {}
       onPointerEnter: setPointerRef,
       onMouseMove(event) {
         const { nativeEvent } = event;
+        const trigger = event.currentTarget as HTMLElement;
+
+        // `true` when there are multiple triggers per floating element and user hovers over the one that
+        // wasn't used to open the floating element.
+        const isOverInactiveTrigger =
+          elements.domReference && !elements.domReference.contains(event.target as Element);
```
atomiks: Use the `contains` util wherever necessary

packages/react/src/popover/close/PopoverClose.tsx
```diff
@@ -31,7 +31,11 @@ export const PopoverClose = React.forwardRef(function PopoverClose(
     props: [
       {
         onClick(event) {
-          setOpen(false, createBaseUIEventDetails('close-press', event.nativeEvent));
+          setOpen(
+            false,
+            createBaseUIEventDetails('close-press', event.nativeEvent),
+            event.currentTarget,
```
atomiks: I'm curious about this signature. I think only one other component uses it, but it would probably be better if `eventDetails` could be extended with custom properties.

packages/react/src/popover/positioner/PopoverPositioner.tsx
```diff
@@ -98,14 +109,50 @@ export const PopoverPositioner = React.forwardRef(function PopoverPositioner(
...
-    [open, positioner.side, positioner.align, positioner.anchorHidden],
+    [open, positioner.side, positioner.align, positioner.anchorHidden, instant],
+  );
+
+  const setPositionerElement = React.useCallback(
+    (element: HTMLElement | null) => {
+      store.set('positionerElement', element);
+    },
+    [store],
   );
```
atomiks, line 154: Other setters in e.g. Combobox `useEventCallback` where possible if it's not called during render.

packages/react/src/popover/trigger/PopoverTrigger.test.tsx
```diff
@@ -181,8 +181,8 @@ describe('<Popover.Trigger />', () => {
 
     it('closes the popover if the user clicks patiently', async () => {
       await renderFakeTimers(
-        <Popover.Root delay={0} openOnHover>
-          <Popover.Trigger />
+        <Popover.Root>
+          <Popover.Trigger delay={0} openOnHover />
```
atomiks, line 185: How are we going to deal with `Dialog.Trigger`, `Tooltip.Trigger` having a different API now? Can this be backwards compatible, at least until all have the API simultaneously?

packages/react/src/popover/viewport/PopoverViewport.tsx
```diff
@@ -0,0 +1,302 @@
...
+    // When a trigger changes, set the captured children HTML to state,
+    // so we can render both new and old content.
+    if (
+      open &&
+      activeTrigger &&
+      previousActiveTrigger &&
+      activeTrigger !== previousActiveTrigger &&
+      lastHandledTriggerRef.current !== activeTrigger &&
+      capturedNodeRef.current &&
+      previousContentNode?.innerHTML !== capturedNodeRef.current.innerHTML
```
atomiks: Even just reading `innerHTML` can prevent browser extensions from accepting code with it  iirc

packages/utils/src/useControlled.ts
```diff
@@ -74,3 +75,66 @@ export function useControlled<T = unknown>({
...
+  name: keyof State;
+}
+
+export function useStoreControlled<Value, State>({
+  controlled,
+  default: defaultProp,
+  store,
+  selector,
+  name,
+}: UseStoreControlledParameters<Value, State>): [Value, (newValue: Value) => void] {
```
atomiks: Does this somehow bypass this issue: https://github.com/mui/base-ui/pull/2707#issuecomment-3272069030 

Or does it just avoid needing to use effects to synchronize the values to the store for the controlled props?

atomiks:
> I updated the PR with fixes for the issues you found.

`NavigationMenu` still has the issue I recorded

There's also this minor issue where the layout jumps if clicking mid-transition:

https://github.com/user-attachments/assets/a9b8dd29-c773-4671-b408-a4ab136625db

packages/react/src/popover/trigger/PopoverTrigger.test.tsx
michaldudak: We're going to release Dialog and Tooltip together with this (I'm already working on the Dialog), so there shouldn't be inconsistencies.

packages/utils/src/useControlled.ts
michaldudak: Actually this is unused. I've been experimenting with making it eaiser to handle controlled values with the Store but I'm going to open a separate PR for it.

packages/react/src/popover/positioner/PopoverPositioner.tsx
michaldudak: I found `useCallback` to be faster in cases where dependencies don't change, both for initial render and re-renders. In our case, `store` is constant, so there's no need to call `useEventCallback`.

docs/src/app/(private)/experiments/motion.tsx
michaldudak: Previously unmounting was prevented whenever `actionsRef` was present. As I added a new imperative action, this logic didn't make sense anymore, so I had to make preventing unmounting explicit.

packages/react/src/popover/close/PopoverClose.tsx
michaldudak: Yeah, OK. I was wondering whether to add a new parameter or a field to eventDetails. I don't really have a preference, so it might as well be a field.

michaldudak: I added the `trigger` field to BaseUIEventDetails and made Floating UI hooks aware of it as it'll be used more widely when we roll out detached triggers for other components.

atomiks:
Focus management is broken:

https://github.com/user-attachments/assets/99e154a5-c420-42fa-94a9-7a4bda16b7e9

michaldudak:
It's also broken this way on master (although it's not common to set up components in a way that it's visible): https://codesandbox.io/p/sandbox/keen-star-rm3lqw

If you tab out of the popover, the "work?" button will be focused.

atomiks:
I see, that makes sense...

It's assumed that the portal is placed directly after the trigger for 'normal' usage

I had to use custom focus guards instead of `FloatingFocusManager` for `NavigationMenu` to make it work correctly for the singular popup.

michaldudak:
I fixed it by allowing to override the next/previous focusable elements in FFM: https://github.com/mui/base-ui/pull/2336/commits/b3172e931d28e667a9e6224c6b97cec6a1b66b33#diff-b6932b4b98aafe6140f0445934c0858c737c3017131f90c6288e856b65dfbe7c

Let me know if it makes sense to you.

atomiks:
Nice idea! I guess NavigationMenu could potentially use that as well, although it had some extra special logic with nested inline menus where it might not work.

When pressing <kbd>Shift+Tab</kbd> inside the popover, it should return focus to the popover trigger, not the previous trigger/element.

It also seems to be incorrectly returning focus immediately on <kbd>Esc</kbd> now, without waiting for the transition to finish (unmount)

The content translation transitions work in the experiment, but not in the docs page (or if they're active, they look off/wrong when transitioning from the first to second trigger, or vice-versa. Transitioning from second to third looks fine). The Tailwind version is also broken (layout issue).

atomiks:
Also, I'm curious if this could work with the `modal` prop by cutting out multiple holes, or if that doesn't work with `clip-path`..

packages/react/src/popover/close/PopoverClose.tsx
atomiks: We'll have to align it with this PR afterwards: https://github.com/mui/base-ui/pull/2796

michaldudak:
> The content translation transitions work in the experiment, but not in the docs page (or if they're active, they look off/wrong when transitioning from the first to second trigger, or vice-versa. Transitioning from second to third looks fine). 

I think this is caused by the popups having different widths. Transitions to width, position and content combined makes the result look a bit off. I made the second popover slightly wider and it looks better.

> Nice idea! I guess NavigationMenu could potentially use that as well, although it had some extra special logic with nested inline menus where it might not work.

I'd prefer to refactor navigation menu separately, as I really would like to merge this ASAP. Resolving conflict has been a pain.

> When pressing Shift+Tab inside the popover, it should return focus to the popover trigger, not the previous trigger/element.

> It also seems to be incorrectly returning focus immediately on Esc now, without waiting for the transition to finish (unmount)

Both of these are now fixed, but a few tests broke in the meantime, so I still have to resolve this.

> Also, I'm curious if this could work with the modal prop by cutting out multiple holes, or if that doesn't work with clip-path.

It should work, in principle, but I don't think we should do this. Inactive triggers shouldn't be clickable when a modal popover is open.

atomiks:
> I think this is caused by the popups having different widths. Transitions to width, position and content combined makes the result look a bit off. I made the second popover slightly wider and it looks better.

Yeah I kinda noticed that each popup (content) needs to have translation values manually tweaked for it to look good (in Navigation Menu)

Maybe we need to provide variables for some kind of automatic translation offset so it always looks good, but that can be done separately.

> I'd prefer to refactor navigation menu separately, as I really would like to merge this ASAP. Resolving conflict has been a pain.

Sure

> It should work, in principle, but I don't think we should do this. Inactive triggers shouldn't be clickable when a modal popover is open.

Yeah, fair enough.

atomiks:
1. Pressing <kbd>Tab</kbd> incorrectly returns focus to the trigger rather than the next tabbable element now
2. Focusing the second popover trigger in the animations demo and pressing <kbd>Shift+Tab</kbd> twice no longer closes it when focus goes outside its trigger (to the first trigger in the row).

atomiks:
When focus is on the popover, pressing <kbd>Shift+Tab</kbd> (to go back to the trigger) then <kbd>Tab</kbd> (to go back to the popover) unexpectedly closes it and focuses the next trigger unlike before.

It also feels a bit weird to show animations while using the keyboard. It works ok in Navigation Menu because focusing a different trigger doesn't cause the popup to close, but given that it does with Popover, you can press <kbd>Enter</kbd> when on a new trigger while the popup is transitioning out and it will animate the size even if it was almost unmounted and it looks odd

packages/react/src/utils/useAnimationsFinished.ts
```diff
@@ -54,14 +58,36 @@ export function useAnimationsFinished(
...
+                  }
+
+                  ReactDOM.flushSync(fnToExecute);
+                } else if (
+                  element.getAnimations().length > 0 &&
+                  element
+
+                    .getAnimations()
+
+                    .some((anim) => anim.pending || anim.playState !== 'finished')
```
atomiks, line 74: Formatting issue

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-full/css-modules/index.module.css
```diff
@@ -0,0 +1,338 @@
...
+  50% {
+    opacity: 0;
+  }
+
+  100% {
+    transform: translateX(-50%);
+    opacity: 0;
+  }
+}
+
```
atomiks: Why are keyframes/CSS animations used instead of CSS transitions?

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-full/css-modules/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+                  <ArrowSvg />
+                </Popover.Arrow>
+
+                {USE_VIEWPORT ? (
+                  <Popover.Viewport className={styles.Viewport}>
+                    {Payload !== undefined && <Payload />}
+                  </Popover.Viewport>
+                ) : (
+                  <div className={styles.Viewport}>{Payload !== undefined && <Payload />}</div>
+                )}
```
atomiks: Is there a reason this const is used in the demo?

packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
```diff
@@ -892,3 +918,13 @@ export function FloatingFocusManager(props: FloatingFocusManagerProps): React.JS
...
+
+function resolveRef<T extends HTMLElement>(
+  ref: T | React.RefObject<T | null> | null | undefined,
+): T | null | undefined {
+  if (ref == null) {
+    return ref;
+  }
+
+  return 'current' in ref ? ref.current : ref;
+}
```
atomiks: This fn is used elsewhere in the codebase, it should be reused

atomiks:
The local state inputs don't seem to work anymore, if they're supposed to

michaldudak:
> The local state inputs don't seem to work anymore, if they're supposed to

I added them to test if the local state isn't reset when transitioning out and it seems to work. I don't intend to keep the state when a popover is closed.

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-full/css-modules/index.tsx
michaldudak: No, it was just for debugging

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-full/css-modules/index.module.css
michaldudak: No particular reason. I started with them and they stuck.

atomiks:
On iOS, the layout of the "Notifications" popup collapses (onto multiple lines) when tapping between the 1st and 3rd trigger

docs/src/app/(public)/(content)/react/components/popover/page.mdx
```diff
@@ -35,5 +37,114 @@ import { Popover } from '@base-ui-components/react/popover';
...
+
+## Examples
+
+### Detached triggers
+
+A popover can be controlled by a trigger located either inside or outside the `Popover.Root` component. For simple, one-off interactions, place the `Popover.Trigger` inside `Popover.Root`, as shown in the example at the top of this page.
+
+However, if defining the popover's content next to its trigger is not practical, you can use a detached trigger. This involves placing the `Popover.Trigger` outside of `Popover.Root` and linking them with a `handle` created by the `Popover.createHandle()` function.
+
+```jsx title="detached triggers"
```
atomiks: Capitalize titles

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-full/css-modules/index.module.css
atomiks: It works with transitions though right?

packages/react/src/floating-ui-react/utils/tabbable.ts
```diff
@@ -40,6 +40,36 @@ export function getPreviousTabbable(referenceElement: Element | null): Focusable
   );
 }
 
+function getTabableNearElement(referenceElement: Element | null, dir: 1 | -1) {
```
atomiks:
```suggestion
function getTabbableNearElement(referenceElement: Element | null, dir: 1 | -1) {
```

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-full/css-modules/index.module.css
michaldudak: I updated the demo to use them (and also made the TW classes slightly more readable by breaking them in multiple lines - I think we could adopt this pattern elsewhere).

michaldudak:
I updated the docs with a section about controlled mode.

As for the release, I won't merge this yet and prepare the Dialog PR ASAP. We'll see how much time it takes (it shouldn't be nearly as much as the Popover as I'll reuse most of the code).

Issue #2329, opened 2025-07-20
ScrollArea


sadetskayfu:
If i selected text in scrollArea and fast click on scrollBar, onPointerUp dont called and i have all time dragging = true. I dont know, maybe is normal

mj12albert:
> If i selected text in scrollArea and fast click on scrollBar, onPointerUp dont called and i have all time dragging = true

@sadetskayfu I couldn't reproduce this, are you able to share a recording or something like that?

sadetskayfu:
I use obs, but .bkw not supported in github, i posted in youtube https://www.youtube.com/watch?v=qWsGDvGBtXI 
The same behavior occurs if you grab the scrollbar at the very edge https://www.youtube.com/watch?v=9jZSGdJbRhQ

atomiks:
What browser/OS is that? I can't reproduce on macOS in any of the main browsers. It could be a bug with how that client fires `pointerup` or `releasePointerCapture`, as I don't see why it wouldn't get called

sadetskayfu:
I check scroll area in radix ui, It's not happening there. I use window 10 with google and edge

sadetskayfu:
I have pointerup and pointerCapture set up, and everything works fine. However, if I have selected text and do click on the scrollbar, my dragging remains true even after the mouse button is released. This only happens during text selection. 
https://www.youtube.com/watch?v=Snt3D8145Vk

atomiks:
@sadetskayfu thanks for the info, I can't reproduce it on macOS but maybe it's an issue on Windows. 

Radix sets `webkitUserSelect` to `none` (which works on Chromium), which perhaps prevents this issue from occurring, somewhat unrelatedly to why it was added.

>  // pointer capture doesn't prevent text selection in Safari
> // so we remove text selection manually when scrolling

sadetskayfu:
Yes thanks, i add for my scrollbar, and all work	
-webkit-user-select: none; /* Chrome/Safari */
-moz-user-select: none; /* Firefox */
-ms-user-select: none; /* IE10+ */

atomiks:
We can either update the docs demo here to include this style in the CSS, or we should make it a functional style like Radix does

PR #2338, opened 2025-07-22
[scroll area] Disable `user-select` on scrollbar and non-main button interactions

atomiks:

Closes https://github.com/mui/base-ui/issues/2329

Doesn't seem to repro on macOS or Windows, so likely caused by some specific esoteric setup

### DIFF
--- a/packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
+++ b/packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
@@ -84,6 +84,10 @@ export const ScrollAreaRoot = React.forwardRef(function ScrollAreaRoot(
   });
 
   const handlePointerDown = useEventCallback((event: React.PointerEvent) => {
+    if (event.button !== 0) {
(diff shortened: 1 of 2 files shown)


mj12albert (approved):

packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
```diff
@@ -84,6 +84,10 @@ export const ScrollAreaRoot = React.forwardRef(function ScrollAreaRoot(
   });
 
   const handlePointerDown = useEventCallback((event: React.PointerEvent) => {
+    if (event.button !== 0) {
```
mnajdova, line 87: How is this related to the issue?

atomiks: While testing the scrollbar with text selection/user-select styling I noticed this bug, too.  Updated title to better reflect both fixes

Issue #2225, opened 2025-07-01
Button component


vitalijalbu:
# Feature request
Would be great having button component as well, now you can tell ok u can do it by yourself, but buttons have a tons of aria ttributes and it's very useful everywhere.. 😁💪

## Summary

[]

https://intentui.com/docs/components/buttons/button

## Examples in other libraries

[]

https://intentui.com/docs/components/buttons/button

## Motivation

most useful component, used everywhere

colmtuite:
> buttons have a tons of aria attributes

@vitalijalbu what do you mean? which aria attributes?

mj12albert:
See https://github.com/mui/base-ui/issues/2138 for why we don't have this and what you could do instead in your project @vitalijalbu

mj12albert:
Duplicate of https://github.com/mui/base-ui/issues/2138

vitalijalbu:
> See [#2138](https://github.com/mui/base-ui/issues/2138) for why we don't have this and what you could do instead in your project [@vitalijalbu](https://github.com/vitalijalbu)

yeah but guys, it's not only a simple button, you have many variants of that:

- icon start
- icon end
- loading state
- disabled state

and many others props...

All ui librararies does it... I don't understand why you don't want to add it...

atomiks:
Base UI is a headless library, so icon stuff is out of scope (you already control the styles)

For loading/disabled states, isn't that straightforward to add yourself?

If the library would add a Button component, it would be to implement `focusableWhenDisabled` + `Enter`/`Space` functionality when rendering a non-native-`<button>` tag - but aside from that, there isn't a reason for an unstyled library to have a button component

aarongarciah:
> For loading/disabled states, isn't that straightforward to add yourself?

Implementing a loading/pending state is [not so straightforward](https://react-spectrum.adobe.com/react-aria/Button.html#pending), so offering a Button component could be interesting in that regard. I see it as a similar case as Avatar: it's not "needed" but it's nice to have.

atomiks:
> Implementing a loading/pending state is [not so straightforward](https://react-spectrum.adobe.com/react-aria/Button.html#pending)

As far as I can tell, this is just setting `aria-disabled="true"` on the `<button>` (i.e. the `focusableWhenDisabled` behavior); it doesn't do anything special necessarily

The `usePress` logic would affect the components that use `useButton` internally if we were to normalize some of these behaviors: https://github.com/mui/base-ui/issues/1726

[...]

PR #2363, opened 2025-07-23
[button] New `Button` component

atomiks:

Essentially this component allows two things:

- Render the button as another tag, like `div`, while remaining accessible
- Add a `pending` prop which replaces the `disabled` prop for buttons that are meant to be disabled while focus was on it while clicking, so focus isn't lost

It's a general-purpose button component, but is _not_ meant to solve "press events" (https://github.com/mui/base-ui/issues/1726), which is a separate issue


packages/react/src/button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+     * Set to `false` if the rendered element is not a button (e.g. `<div>`).
+     * @default true
+     */
+    nativeButton?: boolean;
+    /**
+     * Whether the button is currently pending, indicating an ongoing action.
+     * When `true`, the button will have a `data-disabled` attribute.
+     * @default false
+     */
+    pending?: boolean;
```
mnajdova: Have we consider loading as a name for the prop. I remember some prior discussions on Material UI, e.g. https://github.com/mui/material-ui/pull/21389#discussion_r438994929 and https://github.com/mui/material-ui/issues/21593#issuecomment-651180893 (the loading prop had x5 more votes). 

Both could work in my opinion, but from what I remember when introducing this in Material UI, the loading prop was much more used.

packages/react/src/button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Whether the component renders a native `<button>` element when replacing it
+     * via the `render` prop.
+     * Set to `false` if the rendered element is not a button (e.g. `<div>`).
+     * @default true
+     */
+    nativeButton?: boolean;
+    /**
+     * Whether the button is currently pending, indicating an ongoing action.
+     * When `true`, the button will have a `data-disabled` attribute.
```
mnajdova: Can we leave a note that it shouldn't be used in combination with disabled? Or explain what happen if both are used.

packages/react/src/button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+ */
+export const Button = React.forwardRef(function Button(
+  componentProps: Button.Props,
+  forwardedRef: React.ForwardedRef<HTMLButtonElement>,
+) {
+  const {
+    render,
+    className,
+    disabled: disabledProp = false,
+    nativeButton = true,
```
mnajdova, line 22: I thought `native` would be a better prop name as it's already applied on the `Button` component, but as we already have the prop in other components, it's probably better to use the same name.

packages/react/src/button/Button.test.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { Button } from '@base-ui-components/react/button';
+import { describeConformance, createRenderer } from '#test-utils';
+
+describe('<Button />', () => {
+  const { render } = createRenderer();
+
+  describeConformance(<Button />, () => ({
+    render: (node) => render(node),
+    refInstanceof: window.HTMLButtonElement,
+  }));
```
mnajdova, line 14: Let's add tests for the pending prop.

packages/react/src/button/Button.tsx
mnajdova: I mean, if a button that can show pending state is disabled because of some reason, it shouldn't be focusable, right?

packages/react/src/button/Button.tsx
atomiks: Yeah the `NativeButtonProps` type is reused everywhere, including here now.

packages/react/src/button/Button.tsx
atomiks: Yes, I think `disabled` + `loading` combined should prevent it from being focusable. Added tests for this.

packages/react/src/button/Button.tsx
atomiks: Changed to `loading`.

atomiks:
@mnajdova updated this PR - should be in a proper state now

packages/react/src/button/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+    loading = false,
```
michaldudak: Have you considered other names for this prop? 
This state will rarely be used when something is actually "loading" - more often it'll be "saving". WDYT about a more generic term such as "pending" or "busy"?

atomiks: I had `pending` before but `loading` seems more popular: https://github.com/mui/base-ui/pull/2363#discussion_r2230731816

hungdoansy:
Hey guys. Is this the PR for the Button of base-ui? Am I right or not when saying before this PR is merged, there's no Button in the system?

packages/react/src/button/Button.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+interface ButtonNonNativeProps
+  extends NonNativeButtonProps,
+    ButtonCommonProps,
+    Omit<BaseUIComponentProps<'button', ButtonState>, NonNativeAttributeKeys> {
+  nativeButton: false;
+}
+
+export type ButtonProps = ButtonNativeProps | ButtonNonNativeProps;
```
atomiks, line 92: Union here breaks the API extractor @michaldudak

packages/react/src/button/Button.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+interface ButtonCommonProps {
+  /**
+   * Whether the button should ignore user interaction.
+   *
+   * `'focusable'` disables the button but keeps it focusable.
+   * Use for loading states to prevent the button from losing focus.
+   */
+  disabled?: boolean | 'focusable';
```
battaglr: It seems like there's already a pattern for handling `focusable` state in the [`ToolbarButton`](https://github.com/mui/base-ui/blob/b70db666e2a35d9a16f29e03baca926b2a55ae28/packages/react/src/toolbar/button/ToolbarButtonDataAttributes.ts#L11-L14) component.

packages/react/src/button/Button.tsx
```diff
@@ -0,0 +1,94 @@
...
+    render,
+    className,
+    disabled: disabledProp = false,
+    nativeButton = true,
+    ...elementProps
+  } = componentProps;
+
+  const disabled = Boolean(disabledProp);
+
+  const { getButtonProps, buttonRef } = useButton({
```
battaglr, line 28: I see that `a` elements that have an `href` attribute also get `role="button"`. Is that correct? This is probably an issue within `useButton` itself.

packages/react/src/button/Button.tsx
```diff
@@ -0,0 +1,94 @@
...
+});
+
+export interface ButtonState {
+  /**
+   * Whether the button should ignore user interaction.
+   */
+  disabled: boolean;
+}
+
+interface ButtonCommonProps {
```
battaglr, line 55: Based on https://github.com/mui/base-ui/issues/2225#issuecomment-3195871750, I thought the idea was to add a `loading?: boolean` prop that renders an indeterminate `role="progressbar"` element (and also disables the button). If that's the case, maybe we could use [`Progress`](https://base-ui.com/react/components/progress) for that?

packages/react/src/button/Button.tsx
mj12albert: I suppose we should change Toolbar to match this since it's much nicer than [`focusableWhenDisabled: boolean`](https://base-ui.com/react/components/toolbar#ToolbarButton-focusableWhenDisabled)

packages/react/src/button/Button.tsx
atomiks: That's more so a feature for a wrapping design system button component, not a primitive headless button component - but yeah, you can add Progress inside to create a loading spinner

The problem with a `loading` prop here is it creates 8 different states with three boolean props (`disabled`, `focusableWhenDisabled`, `loading`) and some of them overlap or are impossible. So having just one prop with an extra `loading` string simplifies the mental model of what's actually happening (disabling the button but keeping it focusable).

packages/react/src/button/Button.tsx
battaglr: What about simply `focusable`? Although it could be a bit confusing or add complexity to the library, since it would need to handle cases like `aria-hidden="true"` when `focusable="true"`.

Considering all that, I think `focusableWhenDisabled`, although ugly as hell, might be the right prop for this.

I’m not a fan of extending the `disabled` type, since it’s a standard, well-known prop.

packages/react/src/button/Button.tsx
battaglr: Yeah, you are right. Then, if there's no specific component to show progress of a task, adding the `loading` prop makes absolutely no sense.

docs/reference/generated/button.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Button",
+  "description": "A button component that can be used to trigger actions.\nRenders a `<button>` element.",
+  "props": {},
+  "dataAttributes": {},
```
michaldudak: `data-disabled` is missing

michaldudak (approved):
Aside from the missing data attribute description, this looks good to me.

packages/react/src/button/Button.tsx
battaglr: @atomiks Happy to open a PR to fix this if you think it could help.

docs/src/app/(public)/(content)/react/components/button/page.mdx
```diff
@@ -0,0 +1,48 @@
...
+<Subtitle>
+  A button component that can be rendered as another tag or disabled while still being focusable.
+</Subtitle>
+<Meta
+  name="description"
+  content="A high-quality, unstyled React button component that can be rendered as another tag or disabled while still being focusable."
+/>
+
+<Demo path="./demos/hero" />
+
```
mj12albert, line 12: I think it would help to add some brief Usage guidelines here, e.g. why (when) you'd want a focusable disabled button, as this may not be obvious to everyone

atomiks:
> Aside from the missing data attribute description, this looks good to me.

This isn't really mergeable without the API extractor fixes - all the documented props are broken due to the union type bug I mentioned above

---

Also, I'm rethinking the `focusable` string value for `disabled`. It seems ok that there's no concept of `loading`, and only the concept of `focusableWhenDisabled`. 

For buttons that can be in a loading state at any point, you just add that prop unconditionally.

```js
<Button disabled={loading} focusableWhenDisabled />
```

Form submit buttons should likely always be focusable at any point, as disabling them when they aren't loading seems invalid (as validation should be used when clicked). It's also awkward to use the prop `<Button disabled={loading ? 'focusable' : false}>`

Being able to focus a disabled button is not detrimental to AT given they announce disabled buttons as `dimmed` so having them in tab order seems fine. For design system components that have a `loading` prop, they can also use `focusableWhenDisabled={loading}` if needed

battaglr:
@atomiks 

> I'm rethinking the `focusable` string value for `disabled`. It seems ok that there's no concept of loading, and only the concept of `focusableWhenDisabled`.

Agree 100% with keeping `disabled` with the standard value types and add `focusableWhenDisabled` instead of the `focusable` string.

michaldudak:
I'd see if there are other opinions about this in the team, but overall, I can live with either option.

junyper:
Would it be worth adding logic to prevent setting the `disabled` attribute if the button is currently focused? I think it could help prevent some common accessibility issues where focus is lost (e.g. in pagination and form submission). Here is a simplified example of what I mean: https://stackblitz.com/edit/vitejs-vite-j922lynm?file=src%2FButton.tsx

atomiks:
@junyper interesting idea. The browser (at least Chrome) blocks `aria-hidden` if focus lands on an element with it, and emits a warning, which is somewhat similar to that

Screenshot 2025-10-27 at 5 04 26 pm

junyper:
> @junyper interesting idea. The browser (at least Chrome) blocks `aria-hidden` if focus lands on an element with it, and emits a warning, which is somewhat similar to that
@atomiks I think a warning like that could help, but I was actually thinking about adding logic to use `aria-disabled` in that case instead so that focus is not lost. Something like this:

```
export function Button({
  disabled: disabledProp,
  'aria-disabled': ariaDisabledProp,
  onClick: onClickProp,
  ...props
}: ComponentProps<'button'>) {
  const buttonRef = useRef<HTMLButtonElement>(null);

  let ariaDisabled: typeof ariaDisabledProp;
  let disabled: typeof disabledProp;

  if (Boolean(disabledProp) && document.activeElement === buttonRef.current) {
    // disabled = undefined
    ariaDisabled = true;
  } else {
    disabled = disabledProp;
    ariaDisabled = ariaDisabledProp;
  }

  return (
    <button
      {...props}
      ref={buttonRef}
      onClick={(e: React.MouseEvent<HTMLButtonElement>) => {
        if (ariaDisabled) {
          e.preventDefault();
        } else if (onClickProp) {
          onClickProp(e);
        }
      }}
      aria-disabled={ariaDisabled}
      disabled={disabled && !ariaDisabledProp}
    />
  );
}
```

atomiks:
@junyper reading a ref and document in render like this is quite problematic for several reasons (async React rules, SSR), but it also seems like using an effect or event handler both fire too late. We can consider a different solution if this is requested later on.

junyper:
> @junyper reading a ref and document in render like this is quite problematic for several reasons (async React rules, SSR), but it also seems like using an effect or event handler both fire too late. We can consider a different solution if this is requested later on.

Thanks @atomiks. If you can point me in the direction you're thinking I can put up a separate PR to add this. Are you thinking something more like the following? Although that won't work for SSR because of the state... 

```
export function Button({
  disabled: disabledProp,
  'aria-disabled': ariaDisabledProp,
  onClick: onClickProp,
  ...props
}: ComponentProps<'button'>) {
  const buttonRef = useRef<HTMLButtonElement>(null);

  const [disabledProps, setDisabledProps] = useState<DisabledProps>({
    disabled: disabledProp,
    'aria-disabled': ariaDisabledProp,
  });

  useEffect(() => {
    if (Boolean(disabledProp) && document.activeElement === buttonRef.current) {
      setDisabledProps({
        disabled: undefined,
        'aria-disabled': true,
      });
    } else {
      setDisabledProps({
        disabled: disabledProp,
        'aria-disabled': ariaDisabledProp,
      });
    }
  }, [disabledProp, ariaDisabledProp]);

  return (
    <button
      {...props}
      ref={buttonRef}
      onClick={(e: React.MouseEvent<HTMLButtonElement>) => {
        if (disabledProps['aria-disabled']) {
          e.preventDefault();
        } else if (onClickProp) {
          onClickProp(e);
        }
      }}
      {...disabledProps}
    />
  );
}
```

PR #2369, opened 2025-07-25
[select] Support conditional `multiple` prop in types

### LINKED ISSUE #2368: [select] Canary Release Mis-Types the Multiple Option for Select.Root
# Bug report

Running on the canary release (as encouraged in this request for multi-select dropdowns: https://github.com/mui/base-ui/issues/1956), the multiple prop for `Select.Root` has the type of `true` instead of `boolean`, breaking conditional functionality. 

## Current behavior

SelectRoot.d.ts exports the following incorrect typing for MultipleProps: 
```
interface MultipleProps<Value> extends Omit<Props<Value>, 'multiple' | 'value' | 'defaultValue' | 'onValueChange'> {
    /**
     * Whether multiple items can be selected.
     * @default false
     */
    multiple: true; 
    value?: Value[] | null;
    defaultValue?: Value[] | null;
    onValueChange?: (value: Value[], event?: Event) => void;
  }
```

## Expected behavior

I would expect the multiple prop to be of type boolean, allowing true/false arguments to be passed. When manually changing this type, the typescript engine accepts expressions in the prop.
```
interface MultipleProps<Value> extends Omit<Props<Value>, 'multiple' | 'value' | 'defaultValue' | 'onValueChange'> {
    /**
     * Whether multiple items can be selected.
     * @default false
     */
    multiple: boolean;
    value?: Value[] | null;
    defaultValue?: Value[] | null;
    onValueChange?: (value: Value[], event?: Event) => void;
  }
```

## Base UI version

Canary Release (react@9c5f6d4)

atomiks:
Uses conditional types instead of overloads for `multiple` so as to support `multiple={condition}`

Closes #2368


packages/react/src/select/root/useSelectRoot.ts
```diff
@@ -28,11 +28,12 @@ import { translateOpenChangeReason } from '../../utils/translateOpenChangeReason
...
 import { useFormContext } from '../../form/FormContext';
 import { useField } from '../../field/useField';
-import { type SelectRoot } from './SelectRoot';
+import type { SelectRootConditionalProps, SelectRoot } from './SelectRoot';
+import { EMPTY_ARRAY } from '../../utils/constants';
 
-const EMPTY_ARRAY: never[] = [];
-
-export function useSelectRoot<T>(params: useSelectRoot.Parameters<T>): useSelectRoot.ReturnValue {
+export function useSelectRoot<T, Multiple extends boolean | undefined = false>(
```
michaldudak: nit: there's no need to define a default value to Multiple as it's always provided explicitly.

PR #2372, opened 2025-07-26
[select] Fix `onValueChange` type inference

atomiks:
It's possible to deselect an item (setting it back to `null`) but `onValueChange` didn't reflect the generic type correctly.

Now, `newValue` has `string | null`, matching the state generic:

```jsx
const [value, setValue] = React.useState<string | null>('');
return (
  <Select.Root
    value={value}
    onValueChange={(newValue) => {
                    ^string | null
    })
  >
```

```jsx
const [value, setValue] = React.useState('');
return (
  <Select.Root
    value={value}
    onValueChange={(newValue) => {
                    ^string
    })
  >
```


docs/reference/generated/select-root.json
```diff
@@ -16,7 +16,7 @@
       "description": "The value of the select."
     },
     "onValueChange": {
-      "type": "((value: Value[] | Value, event: Event | undefined) => void)",
+      "type": "((value: Value[] | Value | any | null, event: Event | undefined) => void)",
       "description": "Callback fired when the value of the select changes. Use when controlled."
```
atomiks: Bug in the API extractor I assume

PR #2382, opened 2025-07-29
[all components] Base UI event details

atomiks:
This changes how custom event callbacks pass events.

Previously, the native event and reason would be passed like so:
```js
onOpenChange={(open, event, reason) => {}}
```

They would potentially be undefined.

Now, a `BaseUIEventDetails` object is passed as the second param, which contains `eventDetails.reason`, `eventDetails.event`, `eventDetails.cancel()` and `eventDetails.allowPropagation()`.

```js
onOpenChange={(open, details) => {}}
```

The `details` is always defined; the `event` falls back to `new Event('base-ui')` if it's triggered inside an effect or during render (rather than an event handler).

`.cancel()` cancels the event from having an effect while leaving the component uncontrolled (previously it needed to be controlled). You can also only cancel `.stopPropagation()` from being called: `.allowPropagation()`

(We can add `.allowDefault()` in the future if it becomes needed)

## Unified change reasons

Every component that has a `change` handler (e.g. `onOpenChange`, `onValueChange`) has an exported `ChangeEventReason` type and `ChangeEventData` (the second parameter)

For `onValueChange` callbacks and others which never passed a reason before, I've initialized the reason to `'none'`. We could supply reasons for these if they become necessary in the future

## Event narrowing

`data.event` is now narrowed to the correct type based on the `reason` string without needing to cast (e.g. `MouseEvent`) to access detailed properties

## Experiment

1. The 'B' tooltip doesn't hide when pressing the trigger (once it's open).
2. The 'I' tooltip doesn't stop propagation of `escape-key` 
https://deploy-preview-2382--base-ui.netlify.app/experiments/tooltip-prevent-open

### DIFF
--- a/packages/react/src/input/Input.tsx
+++ b/packages/react/src/input/Input.tsx
@@ -26,4 +26,7 @@ export namespace Input {
   }
 
   export interface State extends Field.Control.State {}
+
+  export type ChangeEventReason = Field.Control.ChangeEventReason;
+  export type ChangeEventDetails = Field.Control.ChangeEventDetails;
 }


packages/react/src/utils/createBaseUIEvent.ts
```diff
@@ -0,0 +1,43 @@
...
+import { makeEventPreventable } from '../merge-props/mergeProps';
+import { NOOP } from './noop';
+
+/**
+ * Creates a SyntheticEvent that satisfies the `BaseUIEvent` contract.
+ */
+export function createBaseUIEvent<E extends Event = Event>(
+  eventParam?: React.SyntheticEvent<Element, E> | E,
+): BaseUIEvent<React.SyntheticEvent<Element, E>> {
+  if (eventParam && typeof eventParam === 'object' && 'persist' in eventParam) {
```
romgrk: In general, runtime type detection is a code smell. There seems to be two independent cases, which could be split into something like:
 - `BaseUIEvent.fromNative(event)`
 - `BaseUIEvent.fromSynthetic(event)`

atomiks: Adjusted

vladmoroz:
Will `event.preventBaseUIHandler()` remove Base UI side effects like this Esc propagation stopped?
https://github.com/mui/base-ui/blob/589511b4bf0836e01fd7706973469a12039abc51/packages/react/src/floating-ui-react/hooks/useDismiss.ts#L192

My problem / use case:

I've got a non-modal dialog which I'd like it to close on Esc only if an element in the dialog is focused, otherwise keep the dialog open and use Esc in my own handler.

It's easy to keep the dialog open after Esc keypress, but I can't process Esc on my side because dialogs swallow Esc and it doesn't reach my event listener

(I have to say in most cases it's very appreciated that dialogs stop Esc propagation when I do want them to close the way it works by default, but when I want to prevent the default behavior, I want complete control over everything)

atomiks:
@vladmoroz we're discussing this now! We might allow passing options like `.preventBaseUIHandler({ stopPropagation: boolean })`

vladmoroz:
Great to hear that. So in this case would passing `true` mean that Base UI still stops propagation, or that I'm preventing Base UI from stopping propagation?
```
event.preventBaseUIHandler({ stopPropagation: true })
```

fwiw if I am down this route, I think I can stop propagation myself, and I guess it doesn't apply to all events (I assume only some stop propagation)

atomiks:
@vladmoroz it probably makes more sense to call it `preventStopPropagation`. Because you can always stop it yourself in `onOpenChange`

vladmoroz:
and then in that case, does it mean that other default behaviors are kept as is? I'm confused a bit about the granularity of the options – are they extra on top of cancelling the default behavior, or exclusive with the default behavior once an option is provided?

atomiks:
Yeah good question. I imagine it should work like this:
```js
// stops the entire handler from having an effect
// event.stopPropagation() isn't called anyway
event.preventBaseUIHandler();

// *only* prevents event.stopPropagation() from being called
// but the logic still executes
event.preventBaseUIHandler({
  preventStopPropagation: true,
});
```

vladmoroz:
Are there any other configuration things to prevent?

Something like this came to my mind
```js
event.configure({
  dismissDialog: true,
  preventDefault: false,
  stopPropagation: false,
});
```

just spitballing though

atomiks:
`preventPreventDefault` is awkward yeah 😬 

Maybe `allowPreventDefault` / `allowStopPropagation`

Since this is meant to be generic for all events/components, we shouldn't make anything too specific. The `reason` strings per component should be enough to have control aside from this?

vladmoroz:
Yeah I like `allowPreventDefault` and `allowStopPropagation`, sounds spot on

atomiks:
@vladmoroz would you prefer

`onOpenChange={(open, data) => {})`

Where `data.event` is the native event (and now `data.reason`)

or 

`onOpenChange={(open, event, data) => {})`

I've changed approach since cloning events is messy, we should avoid mutating native DOM events with custom properties.

The handler is now attached to data either way: `data.preventBaseUIHandler()`

We could attach the React event as `data.reactEvent` and keep `data.event` as native (not all events are from React props)

vladmoroz:
> I've changed approach since cloning events is messy, we should avoid mutating native DOM events with custom properties.
> The handler is now attached to data either way: data.preventBaseUIHandler()

Yeah I much prefer this, the contract is much clearer when the methods like this are on a separate object instead of sneaking it into the event. If you are sticking to `allowPreventDefault, allowStopPropagation`, would a boolean option for preventing the Base UI handler also work?

I don't care about React events vs native events too much, having access to the native event is enough for me personally.

atomiks:
> would a boolean option for preventing the Base UI handler also work?

If you prevent the entire handler, then you shouldn't need to specify any options right? Options are basically if you want to prevent the handler from also doing something to the event while letting it execute

docs/reference/generated/accordion-root.json
```diff
@@ -13,9 +13,9 @@
       "detailedType": "any[] | undefined"
     },
     "onValueChange": {
-      "type": "((value: any[]) => void)",
+      "type": "((value: any[], event: Accordion.Root.ChangeEventData) => void)",
```
michaldudak: The parameter should be called `data`.
However, looking at the whole PR, I'm not sold on the `data` name. In particular, code like `data.cancel` looks ugly. Am I canceling data?
How about `eventData` or `eventDetails` instead?

packages/react/src/utils/createBaseUIEventData.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Discriminated union keyed by the `reason` string literal.
+ * Narrows `event` type based on `reason`.
+ */
+export type BaseUIEventData<Reason extends string = PopupChangeReason> = {
+  [K in Reason]: {
+    reason: K;
+    event: ReasonToEvent<K>;
+    cancel: () => void;
+    cancelStopPropagation: () => void;
```
michaldudak: `allowPropagation`?

packages/react/src/utils/createBaseUIEventData.ts
```diff
@@ -0,0 +1,63 @@
...
+            : Reason extends 'escape-key' | 'list-navigation'
+              ? KeyboardEvent
+              : Event;
+
+/**
+ * Discriminated union keyed by the `reason` string literal.
+ * Narrows `event` type based on `reason`.
+ */
+export type BaseUIEventData<Reason extends string = PopupChangeReason> = {
+  [K in Reason]: {
```
michaldudak: Add JSDocs to properties. Also, this type should be exported from the package.

docs/reference/generated/accordion-root.json
atomiks: > However, looking at the whole PR, I'm not sold on the data name. In particular, code like data.cancel looks ugly. Am I canceling data?

Yeah, agreed; I preferred `event.cancel()` but we shouldn't mutate event objects.  `eventDetails.cancel()` seems to work.

packages/react/src/utils/createBaseUIEventData.ts
atomiks: Each `Root` component has the event data attached. Where are we exporting generic/top-level types atm?

michaldudak: We're not AFAIK. Let's create a top-level `types.ts` file and import the type from there. Then add a /types export path in package.json.

michaldudak (approved):

PR #2393, opened 2025-07-30
Remove redundant Tailwind styles for ScrollArea Tailwind Example

JohnPhamous:
CleanShot 2025-07-30 at 14 12 49@2x

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/scroll-area/demos/hero/tailwind/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/scroll-area/demos/hero/tailwind/index.tsx
@@ -4,7 +4,7 @@ import { ScrollArea } from '@base-ui-components/react/scroll-area';
 export default function ExampleScrollArea() {
   return (
     <ScrollArea.Root className="h-[8.5rem] w-96 max-w-[calc(100vw-8rem)]">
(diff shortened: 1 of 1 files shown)


atomiks:
I think this is done for Tailwind v3 compatibility: https://github.com/mui/base-ui/pull/1185

vladmoroz:
@atomiks @colmtuite do you guys still need to support Tailwind v3? The syntax mix is indeed a little odd/confusing when pasting into your codebase.

There's a v4 for the browser runtime now (there was no v4 for the browser which meant CodeSandboxes wouldn't work which was the original reason for writing v3/v4 compatible classes)

colmtuite:
@Janpot can you think of a way to automate this? so we could upgrade to v4 and write all component styles in v4, but then have a tool automatically add a v3 example? and we could offer v3 and v4 options in the select menu?

Janpot:
If it's just about replacing patterns in classes as per https://github.com/mui/base-ui/pull/1185, then it shouldn't be too hard to automate. We could treat it the same way we do the TS->JS transformation. Isolating v3 from v4 in the demo runtime may be more challenging though. @dav-is thoughts?

I'm adding https://github.com/mui/mui-public/issues/495

PR #2396, opened 2025-07-31
[docs] Expanded types and examples

atomiks:
Todo:

- [x] Fix newline pipe `|` logic

I think in terms of examples, only the more complex/non-obvious props and types (e.g. `collisionAvoidance`) should have them. Controlled vs. uncontrolled mode is also a good use case for examples, though examples can be done in a new PR.

https://deploy-preview-2396--base-ui.netlify.app/react/components/tooltip#TooltipPositioner-collisionAvoidance


docs/reference/generated/tooltip-positioner.json
```diff
@@ -4,68 +4,85 @@
...
       "description": "Minimum distance to maintain between the arrow and the edges of the popup.\n\nUse it to prevent the arrow element from hanging out of the rounded corners of a popup."
     },
     "anchor": {
       "type": "Element | RefObject<Element | null> | VirtualElement | (() => Element | VirtualElement | null) | null",
+      "detailedType": "| Element\n| React.RefObject<Element | null>\n| VirtualElement\n| (() => Element | VirtualElement | null)\n| null\n| undefined",
       "description": "An element to position the popup against.\nBy default, the popup will be positioned against the trigger."
     },
     "collisionBoundary": {
       "type": "Boundary",
+      "detailedType": "| Element\n| any\n| Element[]\n| 'clipping-ancestors'\n| undefined",
```
atomiks: `any` seems to be a bug with the API extractor

docs/reference/generated/tooltip-positioner.json
```diff
@@ -4,68 +4,85 @@
   "props": {
     "collisionAvoidance": {
       "type": "CollisionAvoidance",
-      "description": "Determines how to handle collisions when positioning the popup."
+      "detailedType": "| {\n    side: 'none' | 'flip'\n    align: 'shift' | 'none' | 'flip'\n    fallbackAxisSide: 'none' | 'start' | 'end'\n  }\n| {\n    side: 'shift' | 'none'\n    align: 'shift' | 'none'\n    fallbackAxisSide: 'none' | 'start' | 'end'\n  }\n| undefined",
+      "description": "Determines how to handle collisions when positioning the popup.",
+      "example": "```jsx\n<Positioner\n  collisionAvoidance={{\n    side: 'shift',\n    align: 'shift',\n    fallbackAxisSide: 'none',\n  }}\n/>\n```"
```
atomiks: Each of these properties should have `?:` - possible bug with the extractor

docs/reference/generated/accordion-header.json
```diff
@@ -4,10 +4,12 @@
   "props": {
     "className": {
       "type": "string | ((state: Accordion.Item.State) => string)",
+      "detailedType": "string | ((state: Accordion.Item.State) => string)",
```
flaviendelangle: I'm wondering how many hundered of line we could save by just storing `detailedType` only when if's not exactly the same string as `type` 😆

docs/reference/generated/tooltip-positioner.json
michaldudak: The extractor handles these correctly. It was the type formatter not including the "?" in optional properties. I fixed it.

docs/reference/generated/tooltip-positioner.json
michaldudak: Fixed in 1.0.0-alpha.8

michaldudak:
Tooltips with expanded type definitions are missing (for example, compare Tooltip.Positioner's `anchor` to `collisionBoundary` types)

atomiks:
@michaldudak updated

It shows `| undefined` for simple types but I feel like that's ok (it's nice that tooltips will consistently show for all collapsed types)

michaldudak (approved):
LGTM

PR #2397, opened 2025-07-31
[select] Fix stale items prop

### LINKED ISSUE #2389: [select] after updating `items`, `value` is shown when new item selected
# Bug report

## Current behavior

When updating the items of the select component via React state, any newly added items that are selected display their value instead of the label in the select trigger.

## Expected behavior

The new items should have their label shown, once selected, in the trigger

## Reproducible example

[Link to a CodeSandbox, StackBlitz, or other IDE.
](https://codesandbox.io/p/sandbox/busy-engelbart-vrq9ys)

## Base UI version

1.0.0-beta.1

Chrome/Firefox

## Which OS are you using?

Windows

Using the children function `((value: any) => ReactNode)` fixes the issue but it feels like that shouldn't be needed

atomiks:
Fixes #2389


flaviendelangle (commented):
I don't know why the store is not updated. 

For me the solution would just be to add a `useIsoLayoutEffect` that updates the items in the store whenever the items prop changes.

That way only the parts that listen to this store key will re-render, compared to a full re-render if you store it directly in the context.

Unless I'm missing something.

flaviendelangle (commented):
You can improve slightly the memoization issue by computing `labelFromItems` inside the selector.
You still pay the computation cost of looping through all the items on every render of the root, but at least `<Select.Value />` don't have to re-render (if it has other costly operations)

atomiks:
> For me the solution would just be to add a useIsoLayoutEffect that updates the items in the store whenever the items prop changes.

This seems problematic if the `items` prop isn't referentially stable, as the effect to update `store.apply(...)` will fire continuously

> But in either case, I do agree we need memoization. Just like we need it for the rows and columns on the grid or for the items on the tree view.

Yeah when it's used during rendering, there's no way to avoid needing external memoization, unlike event callbacks/values turned into refs to be used inside an effect or event handler.

flaviendelangle:
But what would be the downside compared to putting it directly in the context?

atomiks:
I haven't reproduced it, but there was concern of a loop being induced vs. just unnecessary re-renders with the context method. Curious what @romgrk thinks

flaviendelangle:
Another approach that works, as long as you only use this prop for the label, you can move the `labelFromItems` from `Select.Value` to `Select.Root` and store this in the state.
AFAIK, you only get benefits compared to adding the items to the context. Nothing re-renders outside of the root unless the items really change (not just memoization issue). Of course you still pay the price of running the `React.useMemo` to compute `labelFromItems` often if they don't memoize the items but I don't see any good solution where you don't (unless you compare the items value, but it would probably be more costly than just re-running everything).

Regarding the loop, as long as you have your dependecy array set up correctly, you shouldn't have any loop.

romgrk:
> You can improve slightly the memoization issue by computing `labelFromItems` inside a memoized selector.

Indeed, which I also pointed out [back here](https://github.com/mui/base-ui/pull/2173#discussion_r2175158724). I wonder if it would be worth explaining better how memoized selectors integrate with the store. They should be the go-to solution for derived state. I would advise against storing derived state in the store directly.

> I haven't reproduced it, but there was concern of a loop being induced vs. just unnecessary re-renders with the context method.

What's the concern exactly? Where's this coming from? From what I can see, we need to add `items` to the store's update effect.

> vs. just unnecessary re-renders

If you're going to stop caring about re-renders in this component, you might as well just remove the whole `Store` logic. The whole point of it is to avoid re-renders to get proper performance. If it's not used, it's just unnecessary complexity & overhead.

atomiks:
> What's the concern exactly? Where's this coming from? From what I can see, we need to add items to the store's update effect.

Mainly just experience with effects having referentially unstable dependencies that also set state, but if the store implementation won't suffer from this, then let's go with it

> If you're going to stop caring about re-renders in this component

It's more so about the fact that they pretty much always _will_ pass a referentially stable `items` value (especially if they care about the perf), the only concern was an infinite loop being generated in cases where it isn't

packages/react/src/select/root/useSelectRoot.ts
```diff
@@ -441,6 +441,7 @@ export function useSelectRoot<Value, Multiple extends boolean | undefined>(
       transitionStatus,
       popupProps: getFloatingProps(),
       triggerProps: getReferenceProps(),
+      items,
```
atomiks, line 444: Possibly this effect should also be `useIsoLayoutEffect` not a plain one

romgrk (approved):

romgrk:
> > If you're going to stop caring about re-renders in this component
> 
> It's more so about the fact that they pretty much always _will_ pass a referentially stable `items` value (especially if they care about the perf), the only concern was an infinite loop being generated in cases where it isn't

Right, but you fixed it by adding the prop to the context. If the context re-renders, all item components re-render. Adding unstable values to the context makes the Store logic useless, so keeping that logic would be pointless.

flaviendelangle:
> Mainly just experience with effects having referentially unstable dependencies that also set state, but if the store implementation won't suffer from this, then let's go with it

To give some additional information here.
To get a loop, you need your effect to do something that triggers a re-render of the component that contains the effect.
Which is not the case here.

```tsx
// Will create a loop of renders
function App() {
  const [state, setState] = React.useState({ name: 'fake date', id: 1 });

  React.useEffect(() => {
    setState({ ...state, name: 'new-fake-name' })
  }, [state])
}
```

```tsx
// Won't create a loop of renders because SubComponent re-renders but not App
// On the 2nd render, `App` did not re-render so `props.item` is still the same object in `SubComponent`, so the effect does not re-trigger
// (the code is super ugly of course)
function App() {
  return <SubComponent item={{ name: 'fake date', id: 1 }}  />
}

function SubComponent({ item }) {
  const [state, setState] = React.useState(item);

  React.useEffect(() => {
    setState({ ...state, name: 'new-fake-name' })
  }, [item])
}
```

In our scenario, the component that contains the update `Select.Root` doesn't even re-render because it's not listening to `state.items`. But even if it were in the future, it wouldn't cause a loop.

atomiks:
@flaviendelangle though in theory it could be possible for the owning component to re-render at the same time depending on how callbacks are being passed down to the sub component - I don't think this is a possibility here for Select, it's just that I usually try to avoid synchronizing state (the store in this case) with an external prop, and instead prefer to always derive it (which my context change was). But of course, external memoization complicated that

flaviendelangle:
I agree that if in your `useEffect` you call a callback passed by the parent, then you expose yourself to potential loops depending on what the callback does.

Issue #1420, opened 2025-02-05
[tooltip] Trigger does not set aria-describedby


mathias-falkenberg:
# Feature request

## Summary

BaseUI’s `Tooltip` component currently lacks a built-in way to properly associate tooltips with their trigger elements using `aria-describedby`. This feature would improve accessibility by ensuring that assistive technologies, such as screen readers, can properly announce tooltips as descriptions for their related elements.

## Examples in other libraries

[MUI Tooltip](https://mui.com/material-ui/react-tooltip/#accessibility)
[React Aria Tooltip](https://react-spectrum.adobe.com/react-aria/Tooltip.html#example)

## Motivation

1. **Improved Accessibility** – Associating tooltips with their trigger elements using `aria-describedby` ensures that users relying on screen readers receive the tooltip information when focusing on the element.  
2. **Better Compliance with ARIA Practices** – The [W3C ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/) suggests linking tooltips to their trigger elements to provide clear context.  
3. **Consistency Across UI Libraries** – Other popular component libraries, like MUI and React Aria, follow this practice, making it a valuable enhancement for BaseUI to maintain parity.  
4. **Improved Developer Experience** – Developers would have an out-of-the-box solution to make tooltips more accessible, reducing the need for manual workarounds.

mj12albert:
@atomiks We're not intentionally leaving out aria-describedby right?
I remember we heavily discussed the verbosity of announcing these as it relates to the length of the text

atomiks:
We intentionally made Tooltip visual-only, as in most cases it's used as a visual label for the trigger (often an icon button). An accessible name (e.g. via `aria-label`) is preferred to communicate the same information for screen readers as the tooltip does for sighted users — `aria-describedby` repeats the accessible name twice when announced in this use case

There are cases where the tooltip offers additional information different from the accessible name though, where it could be useful. GitHub has a few examples:

<img width="419" alt="Image" src="https://github.com/user-attachments/assets/7aac02a8-79b8-4034-ae7e-8cba204998bd" />

The issue is differentiating when it's a label vs. when it's really a description to use `aria-describedby`

mathias-falkenberg:
It seems to be a very common pattern that most other libraries support.
It's even stated in the [WAI ARIA Tooltip pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/)

[...]

PR #2432, opened 2025-08-02
[tooltip] WAI-ARIA attributes 

seongminn:

In accordance with the tooltip pattern defined in the [W3C documentation](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/#wai-ariaroles,states,andproperties), the following attributes have been added to the Tooltip component.

- Added `role="tooltip"` to `<Tooltip.Popup />`.
- Added `aria-describedby` to `<Tooltip.Trigger />`.

Additionally, a test for the `aria-describedby` attribute has been added to the `TooltipPopup.test.tsx` file.

### DIFF
--- a/packages/react/src/tooltip/popup/TooltipPopup.tsx
+++ b/packages/react/src/tooltip/popup/TooltipPopup.tsx
@@ -58,6 +58,9 @@ export const TooltipPopup = React.forwardRef(function TooltipPopup(
     state,
     ref: [forwardedRef, popupRef],
     props: [
+      {
+        role: 'tooltip',
+      },
       popupProps,
       transitionStatus === 'starting' ? DISABLED_TRANSITIONS_STYLE : EMPTY_OBJECT,
       elementProps,
--- a/packages/react/src/tooltip/popup/TooltipPopup.tsx
+++ b/packages/react/src/tooltip/popup/TooltipPopup.tsx
@@ -1,5 +1,6 @@
 'use client';
 import * as React from 'react';
+import { useIsoLayoutEffect } from '@base-ui-components/utils/useIsoLayoutEffect';
 import { useTooltipRootContext } from '../root/TooltipRootContext';
 import { useTooltipPositionerContext } from '../positioner/TooltipPositionerContext';
 import type { BaseUIComponentProps } from '../../utils/types';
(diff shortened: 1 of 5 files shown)


packages/react/src/tooltip/popup/TooltipPopup.tsx
```diff
@@ -27,12 +29,28 @@ export const TooltipPopup = React.forwardRef(function TooltipPopup(
...
+    setPopupId,
+  } = useTooltipRootContext();
   const { side, align } = useTooltipPositionerContext();
 
+  useIsoLayoutEffect(() => {
+    setPopupId(id);
+    return () => {
+      setPopupId(undefined);
+    };
+  }, [setPopupId, id]);
```
romgrk, line 52: Perf note: this pattern forces a re-render right after the popup mounts/renders for the first time.

atomiks:
Hi @seongminn,

The tooltip component is intentionally visual-only and has no presence to screen readers.

Per the [accessibility guidelines](https://base-ui.com/react/components/tooltip#accessibility-guidelines), you should instead rely on `aria-label` on the trigger to label the element for screen readers, not rely on the tooltip (via `aria-describedby`/`aria-labelledby`). 

1. If your tooltip is acting purely as a label (which is usually the case, e.g. an icon button tooltip), `aria-label` can directly match the tooltip content
2. If your tooltip is acting as 'extended' information on top of the label, this is usually a design mistake. This is because touch users cannot see tooltips and have no way to access the information. Per the guideline, `<Popover.Root openOnHover>` is recommended for essential information to be accessible to all input modalities. If the information is non-essential, then there's no need for `aria-describedby` because omitting it doesn't matter as you accept that not everyone can see it and that this isn't an issue. This is questionable though and suggests an alternative design should be used.

https://github.com/mui/base-ui/issues/1420#issuecomment-2642620055

seongminn:
@atomiks 

I see there's a possibility that including the aria-describedby attribute in the Tooltip could result in the same content being announced twice (once for the label and once for the description).

As mentioned in the example from [this issue](https://github.com/mui/base-ui/issues/1420#issuecomment-2638911361), there can certainly be cases where a Tooltip provides additional, non-essential information about an element. In those situations, aria-describedby would be appropriate. However, I agree with your point that it should not be a default requirement.

I will close this PR. Thank you for the valuable insight and your detailed explanation!

PR #2441, opened 2025-08-04
[navigation menu] Unshare AbortController instance

tomokinat:

`AbortController` construction outside of a request context is not supported in Cloudflare Workers/workerd, which makes base-ui unusable in that environment: https://github.com/cloudflare/workerd/issues/3657.

I also noticed a likely unintentional bug (correct me if it's intended) where multiple `NavigationMenuTrigger` instances were sharing the same `AbortController`. I've simply inlined the initial value to fix both problems.


packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
```diff
@@ -99,7 +98,7 @@ export const NavigationMenuTrigger = React.forwardRef(function NavigationMenuTri
 
   const allowFocusRef = React.useRef(false);
   const prevSizeRef = React.useRef(DEFAULT_SIZE);
-  const animationAbortControllerRef = React.useRef(DEFAULT_ABORT_CONTROLLER);
+  const animationAbortControllerRef = React.useRef(new AbortController());
```
romgrk: We want to avoid useless memory allocations. The alternative would be to leave the ref empty by default.

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
```diff
@@ -108,6 +107,9 @@ export const NavigationMenuTrigger = React.forwardRef(function NavigationMenuTri
   const runOnceAnimationsFinish = useAnimationsFinished(popupElement);
 
   React.useEffect(() => {
+    if (animationAbortControllerRef.current === null) {
+      animationAbortControllerRef.current = new AbortController();
+    }
     animationAbortControllerRef.current.abort();
```
romgrk:
```suggestion
    animationAbortControllerRef.current?.abort();
```

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
```diff
@@ -194,6 +196,9 @@ export const NavigationMenuTrigger = React.forwardRef(function NavigationMenuTri
     }
 
     const mutationObserver = new MutationObserver(() => {
+      if (animationAbortControllerRef.current === null) {
+        animationAbortControllerRef.current = new AbortController();
+      }
       animationAbortControllerRef.current.abort();
```
romgrk:
```suggestion
      animationAbortControllerRef.current?.abort();
```

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
tomokinat: Oops, I didn't realize the default controller is just a placeholder value that's used nowhere. Done.

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
tomokinat: Done, as in the comments below. Thanks!

PR #2310, opened 2025-07-16
[docs] Enable highlighed lines in demos

mj12albert:
- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #2443, opened 2025-08-05
[docs-infra] Migrate to @mui/internal-docs-infra Demos

dav-is:
Migrates this repo to use `@mui/internal-docs-infra` Code Highlighter: https://github.com/mui/mui-public/pull/379


docs/next.config.mjs
```diff
@@ -50,6 +48,37 @@ const rootPackage = loadPackageJson();
 const nextConfig = {
```
brijeshb42, line 48: It'd be ideal if we could follow the accepted way of adding items to the default config like -

```js
export default withDocsInfraPipeline(nextConfig);
```

dav-is: Good idea, should we replace `withDocsInfra()`?

brijeshb42: I think the current one will still be there. We could have a new one, like `withDocsInfraDemo()` or similar since it mainly affects the demos.

dav-is: What would you think about `withDocsInfra(nextConfig, { demos: true })`

brijeshb42: Yeah. Works, assuming we rename the existing one to `withMonorepoDocsInfra`.

In that case, we could just use `withDocsInfra()` without the 2nd argument.

mj12albert:
@dav-is Just wondering what's the syntax for highlighting lines? (I couldn't find an example here or in the linked PR 😓 )

I was trying out the native Shiki features for this in https://github.com/mui/base-ui/pull/2310 but there are issues that needed to be manually worked around in userland (on our side)

michaldudak:
I'm not convinced that having the last selected file as a URL fragment by default is a good idea. What's the use case? How about a "Copy permalink" button instead?

colmtuite:
@michaldudak What's your beef with this?

michaldudak:
I find it odd that the URL (global state) reflects the state of only the last code panel a user has interacted with.

dav-is:
> @mj12albert Just wondering what's the syntax for highlighting lines?

This feature will come in a follow up PR. The proposed syntax will be similar to:

Single line:
```jsx
export default function Component() {
  return (
    <div>
      <h1>Heading 1</h1> {/* @demo see here */}
    </div>
  )
}
```

Multiline:
```jsx
export default function Component() {
  return (
    // @demo see below
    <div>
      <h1>Heading 1</h1>
    </div>
    // @demo see above
  )
}
```

Multiline with comment "we add a heading with an h1":
```jsx
export default function Component() {
  return (
    // @demo see below where we add a heading with an h1
    <div>
      <h1>Heading 1</h1>
    </div>
    // @demo see above
  )
}
```

Some pros to this approach is that it's easy to understand in plain markdown and is enhanced in the web version. It also avoids the shortcomings of manual line selections (`highlight="1,3-5,7...10"`) drifting from the code itself.

> @michaldudak I'm not convinced that having the last selected file as a URL fragment by default is a good idea. What's the use case?

To add some context here, Michał and I had a discussion about this and decided the behavior in this PR is ok. The main reason for using the URL hash is for rendering the correct file tab before hydration and easily reconciling tab changes that occurred before hydration. This hasn't been implemented yet, but is planned for the future. Additionally, poor UX is avoided because the back navigation stack is not updated with each file: pressing back will go to the previous page, not the previous file selected.

atomiks:
Given the state is now in the URL, is it possible to avoid a hydration flicker?

https://github.com/user-attachments/assets/51eb9e7e-baca-4815-938d-643ad91238d2

dav-is:
> @atomiks Given the state is now in the URL, is it possible to avoid a hydration flicker?

Yes, because it is stored in the URL hash, it will open up the possibility of CSS based tab selection using `:target` in a future iteration. The existing demos already have a flash of the wrong file when restoring the variant preference from localstorage.

docs/src/app/(public)/(content)/react/components/progress/demos/hero/css-modules/index.tsx
```diff
@@ -1,3 +1,4 @@
+'use client';
 import * as React from 'react';
 import { Progress } from '@base-ui-components/react/progress';
 import styles from './index.module.css';
```
atomiks, line 4: Curious why this `'use client'` addition necessary? Some demos don't use state or other client-only features, so they're directly pasteable into server components

Edit: never mind, I see this demo is actually using the client features

mnajdova:
I noticed that the colors used in the CSS vars in the previous demos were different, e.g. https://deploy-preview-2443--base-ui.netlify.app/react/components/accordion#hero:theme.css uses different color than https://base-ui.com/react/components/accordion. More precisely the new demos correctly used the values defined in theme.css, but we should double check the demos to ensure that all demos look good with this update. It's a fix in a way, but we need to validate if the CSS vars gives good output in all demos.

dav-is:
> @mnajdova I noticed that the colors used in the CSS vars in the previous demos were different

Good catch. I updated the demo's colors to match the rest of the page, reflecting today's behavior.

[`b6ff6e7` (#2443)](https://github.com/mui/base-ui/pull/2443/commits/b6ff6e7e2f4bd77633096597975b0281769f146e#diff-110006516b54267b00da7f709474542e15fdbe93d96186a3b72bda2f46bc7847)

docs/src/app/(public)/(content)/react/components/progress/demos/hero/css-modules/index.tsx
dav-is: > @atomiks Edit: never mind, I see this demo is actually using the client features

When I added `'use client';` to these files, it was because they were erroring at build time, so there are still some demos that intentionally do not have it.

brijeshb42:
@dav-is I found a bug where the demo (in url hash) is pre-expanded everytime the page is opened. Is this expected ?

dav-is:
@brijeshb42 If there is a file selected in the URL, it should automatically expand. Is it expanding with no hash in the URL? Or is it other demos on the page are also expanding?

One of the ideas behind the hash is if you refresh the page, it can restore to the view you had open before.

mnajdova:
Yeah, I would expact the code to be expanded if the URL has a file 👌

docs/src/app/(public)/(content)/react/components/accordion/page.mdx
```diff
@@ -6,7 +6,9 @@
   content="A high-quality, unstyled React accordion component that displays a set of collapsible panels with headings."
 />
 
-<Demo path="./demos/hero" />
+import { DemoAccordionHero } from './demos/hero';
```
mj12albert, line 9: Is this manual import unavoidable? It's not as convenient as `<Demo path="./demos/hero" />`

dav-is: This convention is what enables Ctrl/Cmd+Click on the import to open the demo and makes the behavior more transparent. It also allows the prop types for the demo component to be exposed when Ctrl/Cmd+Space, hover, or typing.

I think this might be slower to write, but faster to read / find the file you're looking for. Considering you only add the demo once, but update the demo contents many times, I think this is a fair tradeoff.

Make sure you have the VSCode plugin: https://marketplace.visualstudio.com/items?itemName=unifiedjs.vscode-mdx

mnajdova:
@dav-is can you please replace CodeSandbox with Stackblitz as part of this PR too?

Janpot:
I noticed a less than ideal overflow of the tabs in https://deploy-preview-2989--base-ui.netlify.app/react/handbook/forms#tanstack-form

https://github.com/user-attachments/assets/2d3b6b67-cf6d-43e2-bf3e-47b3442de134

mj12albert:
> I noticed a less than ideal overflow of the tabs in https://deploy-preview-2989--base-ui.netlify.app/react/handbook/forms#tanstack-form

Yeah one of the reasons for adding another sandbox button in the upper right corner

For the overflow, there is a overflow gradient from the props table that could probably used here

colmtuite:
> I'm not convinced that having the last selected file as a URL fragment by default is a good idea. What's the use case?

@michaldudak @dav-is The use case is sharing the docs to someone so they can open the docs using the correct styling lib (tailwind or css modules), and also share with the correct tab open (in case you're sharing something specific about the css for example)

I'm also not convinced it's a good idea. Often, it's worth missing features for overall simplicity.

> How about a "Copy permalink" button instead?

This reminds me of Youtube's "Start at xx:yy" UI, which is buried, and for good reason.

I quite dislike the url. It makes me feel like I'm in a complex app, when it's just a docs site. If I go to copy the URL, it makes me feel like I can't, because I'm confronted with complexity that I then need to think through before sharing.

Another issue for me is that I don't know that I'd ever want to be linked to the hero demo, but slightly scrolled down the page. I'd prefer to just arrive at scrollTop: 0. I find it disorienting. 

@atomiks Do you have any thoughts from a product/UX perspective?

colmtuite:
@dav-is Connor came up with an idea where we rely on native <a> functionality, so there are no url changes unless you right-click a tab and press "copy link address", then share that url.

I think this could be a nice solution?

dav-is:
> @colmtuite Connor came up with an idea where we rely on native functionality, so there are no url changes unless you right-click a tab and press "copy link address", then share that url.
>
> I think this could be a nice solution?

Ok I updated it to avoid adding the hash into the address bar: https://deploy-preview-2443--base-ui.netlify.app/react/components/autocomplete#grouped:index.module.css

See the a files that you can right-click to get the link, or Cmd/Ctrl+Click to open a new tab or drag it into a new tab

mnajdova:
Code wise it looks good to me. I was wondering if `compact` shouldn't be the default way the demo render, and maybe have an extended prop, but it's a nit, probably not that big of a deal.

Janpot:
> Ok I updated it to avoid adding the hash into the address bar

Is it supposed to immediately update the url after opening the page? I fully agree that it shouldn't follow the tabs changes. But my expectations when opening a url is that I can subsequently refresh my page without it behaving differently. Would it make sense to just maintain the whole hash from the url until another anchor (non-file) is clicked?

**edit**

It seems the page url changes to https://deploy-preview-2443--base-ui.netlify.app/react/components/autocomplete#inline, but that's not pointing to an existing id.

dav-is:
> Is it supposed to immediately update the url after opening the page?

Yes, if the url is maintained, then changing tabs would update the url. Colm and I were discussing that it would be inconsistent. Removing the file and variant from the url immediately also decreases the chances of accidentally sharing the verbose URL.

> It seems the page url changes to https://deploy-preview-2443--base-ui.netlify.app/react/components/autocomplete#inline, but that's not pointing to an existing id.

It looks like this example has a mismatch between the title (being `Autocomplete Inline` and the demo being stored at `demos/inline/`

This could be solved with linting or following authoring recommendations: https://github.com/mui/mui-public/blob/davis/add-docs-infra-docs/docs/app/docs-infra/components/code-highlighter/page.mdx#demo-titles

Janpot:
> Yes, if the url is maintained, then changing tabs would update the url.

In my expectation it's like scrolling away from a section header past another one. I don't like websites that automatically update the hash. If I didn't click a link that is meant to update the url, the url shouldn't change.

> This could be solved with linting or following authoring recommendations

This feels like quite the restriction on content authoring. i.e. in core there are [sections](https://mui.com/material-ui/react-autocomplete/#creatable) with multiple demos. Can't the demo just have its own id, maybe namespaced with `demo-` or something?

dav-is:
> This feels like quite the restriction on content authoring. i.e. in core there are sections with multiple demos. Can't the demo just have its own id, maybe namespaced with demo- or something?

The hashes are meant to extend one another, e.g. `#advanced` has `#advanced:index.module.css`

I see 3 options we could do to fix this:

1. Use `<DemoAlertDialogInline.Title />` pattern (title: 'Inline Autocomplete', path: `/components/autocomplete#inline`) - allows a title to have a different slug
2. Change `# Inline Autocomplete` to `# Inline` (path: `/components/autocomplete#inline`)
3. Change `./demos/inline` to `./demos/inline-autocomplete` (path: `/components/autocomplete#inline-autocomplete`)

As for the case Jan highlights where they are multiple demos under a single heading, we could add a prop `hasNoHeading` which would add the demo slug as an `id` to the demo's `div`

colmtuite:
> my expectations when opening a url is that I can subsequently refresh my page without it behaving differently.
> I don't like websites that automatically update the hash. If I didn't click a link that is meant to update the url, the url shouldn't change.

Yes seeing it in action, I agree. Just seeing the URL flash bothers me. It makes me think "oh god they're doing magic. I don't like magic. Is there magic in the components too? oh no!"

> Would it make sense to just maintain the whole hash from the url until another anchor (non-file) is clicked?

I don't think this would work because when you land on `/accordion#hero:index.module.css`, and then click the theme.css tab, the URL would still be `/accordion#hero:index.module.css` and would be inaccurate.

I like using `<a>` for progressive enhancement. I think that's more important than any other problem mentioned.

Janpot:
> I don't think this would work because when you land on /accordion#hero:index.module.css, and then click the theme.css tab, the URL would still be /accordion#hero:index.module.css and would be inaccurate.

Or you could clear the hash when clicking another tab?

dav-is:
> Or you could clear the hash when clicking another tab?

I think we can make that work 👍🏼

dav-is:
> Or you could clear the hash when clicking another tab?

Ok the hash is now removed when clicking another tab: https://deploy-preview-2443--base-ui.netlify.app/react/components/autocomplete#grouped:index.module.css

colmtuite:
@dav-is I like it ✅ Please wait for a few upvotes, I wanna make sure people are happy with this change.

Is it possible to set the scroll to the same location as if you clicked the heading? It's disorienting when you click the link and land in an area where you can't see the example's heading and description text.

Janpot:
Something seems to go wrong with this:

Steps:
1. go to https://deploy-preview-2443--base-ui.netlify.app/react/components/autocomplete
2. clear localstorage
3. pick tailwind
4. close the tab
5. go to https://deploy-preview-2443--base-ui.netlify.app/react/components/autocomplete#grouped:index.module.css
6. click `theme.css`
7. It opens the tailwind variant


https://github.com/user-attachments/assets/5ba0e31c-6604-4ffd-838c-8699d856fb87

For some reason that I can't immediately explain, clicking the links seems significant. Pasting in the omnibox does something else.

Also, when opening https://deploy-preview-2443--base-ui.netlify.app/react/components/autocomplete#grouped:index.module.css, only this demo is in css modules, the other remain in tailwind. This feels very surprising and inconsistent to me. My expectation would be:

1. css modules demo and tailwind demo have different hash target. e.g. `#tw-grouped:...`, and `#cm-grouped:...`
2. when the hash is in the url, all demos are in the variant of the hash. when the hash is not in the url, all demos are in the variant of local storage
3. when another hash link pointing to the same page is clicked, 2 things happen
    1. localstorage is updated with the variant that was in the hash. (If you don't want this, then only remove the file part from the hash and leave the demo id)
    2. the hash disappears if it's a demo link, or just updated if it's any other link

dav-is:
> Is it possible to set the scroll to the same location as if you clicked the heading?

We're using native functionality to scroll by adding `<span id="outline:index.module.css" />` to the beginning of the demo div. The demo doesn't have access to the heading and description elements, so we can't add any elements there.

dav-is:
> @Janpot My expectation would be:
> 1. css modules demo and tailwind demo have different hash target. e.g. #tw-grouped:..., and #cm-grouped:...
> 2. when the hash is in the url, all demos are in the variant of the hash. when the hash is not in the url, all demos are in the variant of local storage
> 3. when another hash link pointing to the same page is clicked, 2 things happen
>   i. localstorage is updated with the variant that was in the hash. (If you don't want this, then only remove the file part from the hash and leave the demo id)
>   ii. the hash disappears if it's a demo link, or just updated if it's any other link

Ok updated, should match your expecations: https://deploy-preview-2443--base-ui.netlify.app/react/components/autocomplete#grouped:css-modules:index.module.css 

https://github.com/mui/mui-public/pull/879

Old URLs wont work because we are always adding the variant name in the hash now

mnajdova:
> We're using native functionality to scroll by adding <span id="outline:index.module.css" /> to the beginning of the demo div. The demo doesn't have access to the heading and description elements, so we can't add any elements there.

Can the demo component/factory accept alternative element as a scroll target? We can do it in a follow up, no need to be handled in this PR.

Janpot (approved):
Looks like it works 👍 

<hr>

Some notes for follow-up:

It still doesn't feel 100% right to me that opening a hash link and switching a tab changes the setting for your whole browser from then on. I slept a bit on it and maybe the right approach is a persistence layer that first reads from sessionstorage first and cascades down into localstorage if nothing is in session storage. only the variant switcher can update the localstorage key, but opening a page through a hash link sets the session storage key. That way, after opening a link, your tab stays on that initial variant indefinitely, but never updates the global setting without explicitly doing so.

If this makes sense, we should try it in follow-up. This PR has been blocked on this bikeshed color for way too long. For future PRs, our reflex should be to scope down if we encounter a non-essential feature that requires more extensive discussion.

> Can the demo component/factory accept alternative element as a scroll target? We can do it in a follow up, no need to be handled in this PR.

🤔 Something analog to `aria-labelledby` could be nice, but it would require us to fully handle hash links in javascript. I guess we'd need a synchronous script as well for this, to avoid seeing the page jump when react effects kick in.

docs/src/app/(public)/(content)/react/components/alert-dialog/demos/detached-triggers-controlled/css-modules/index.tsx
```diff
@@ -1,6 +1,7 @@
+'use client';
 import * as React from 'react';
 import { AlertDialog } from '@base-ui-components/react/alert-dialog';
-import styles from '../../index.module.css';
+import styles from '../../_index.module.css';
```
LukasTy, line 4: Could you clarify if these reused CSS files were renamed because of a changed processing, which would treat anything starting with `index` differently?

docs/src/app/(public)/(content)/react/components/checkbox-group/page.mdx
```diff
@@ -35,11 +37,15 @@ A checkbox that controls other checkboxes within a `<CheckboxGroup>` can be crea
...
+import { DemoCheckboxGroupParent } from './demos/parent';
+
+<DemoCheckboxGroupParent />
 
 ### Nested parent checkbox
 
-<Demo compact path="./demos/nested" />
+import { DemoCheckboxGroupNested } from './demos/nested';
+
+<DemoCheckboxGroupNested />
```
LukasTy: Missing `compact` or conscious decision? 🤔 
```suggestion
<DemoCheckboxGroupNested compact />
```

docs/src/app/(public)/(content)/react/components/checkbox-group/page.mdx
```diff
@@ -71,4 +77,6 @@ To use checkbox group in a form:
 </Form>
 ```
 
-<Demo compact path="./demos/form" />
+import { DemoCheckboxGroupForm } from './demos/form';
+
+<DemoCheckboxGroupForm />
```
LukasTy: Same question as above.

```suggestion
<DemoCheckboxGroupForm compact />
```

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-controlled/tailwind/index.tsx
```diff
@@ -96,3 +96,30 @@ export default function PopoverDetachedTriggersSimpleDemo() {
     </React.Fragment>
   );
 }
+
+export function ArrowSvg(props: React.ComponentProps<'svg'>) {
```
LukasTy: Is common component splitting/sharing no longer possible? 🙈

dav-is:
> Can the demo component/factory accept alternative element as a scroll target?

Another idea might be making the title position sticky with a collapsed description. Could be done alongside making the demo content and collapse button position sticky when the code is expanded

docs/src/app/(public)/(content)/react/components/alert-dialog/demos/detached-triggers-controlled/css-modules/index.tsx
dav-is: The filenames are normalized when displayed, so `button/index.tsx` would become `button.tsx`. If you want `index.tsx` to be shown, it should be escaped in the filesystem with `_index`

docs/src/app/(public)/(content)/react/components/checkbox-group/page.mdx
dav-is: Looks like I missed it because it had `compact` before `path`

docs/src/app/(public)/(content)/react/components/alert-dialog/demos/detached-triggers-controlled/css-modules/index.tsx
LukasTy: Gotcha. Thanks for confirming. 👍 
Just checking: this applies to `CSS` files as well?

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-controlled/tailwind/index.tsx
dav-is: Common component sharing is still supported, but not with tree shaking and bundling into a single file

Edit: maybe this has changed since when I last looked at it?

docs/src/app/(public)/(content)/react/components/alert-dialog/demos/detached-triggers-controlled/css-modules/index.tsx
dav-is: Yes, it would apply to any files that are imported into a demo

docs/src/components/Demo/Demo.tsx
```diff
@@ -1,72 +1,119 @@
...
+                  className="contents"
+                  onVariantChange={demo.expand}
+                  variants={demo.variants}
+                  selectedVariant={demo.selectedVariant}
+                  selectVariant={demo.selectVariant}
+                  availableTransforms={demo.availableTransforms}
+                  selectedTransform={demo.selectedTransform}
+                  selectTransform={demo.selectTransform}
+                />
+                {fallbackToCodeSandbox ? (
```
LukasTy, line 79: Is there a problem with StackBlitz on Edge or Safari? 🤔 
Could we document the problem?

docs/src/app/(public)/(content)/react/components/checkbox-group/page.mdx
LukasTy: Oh yeah, those were tricky to spot when glancing over the diff. 🙈

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-controlled/tailwind/index.tsx
dav-is: It looks like it works fine, so I restored it as it was

dav-is:
> Is the change in certain CSS token highlighting colors expected?

Yes, we are using a new lib for highlighting and it seems to detect `oklch` differently. The changes between the two libraries is minimal

docs/src/components/Demo/Demo.tsx
dav-is: This is something @michaldudak and I were looking at today. It seems StackBlitz + Safari is in beta still

image

image

LukasTy: Ok, gotcha. 👍 
Sad to see it. 🙈

docs/src/app/(public)/(content)/react/components/popover/demos/detached-triggers-controlled/tailwind/index.tsx
LukasTy: Nice. 👍

docs/src/utils/demoExportOptions.ts
```diff
@@ -0,0 +1,232 @@
...
+    variant: transformedVariant,
+    globals: transformedGlobalsForCRA,
+  };
+};
+
+const versions = {
+  '@types/react': '^19',
+  '@types/react-dom': '^19',
+  react: '^19',
+  'react-dom': '^19',
```
LukasTy, line 174: Nitpick: Possible check for future improvement - could this be read from `package.json` instead of hardcoding?

docs/src/utils/demoExportOptions.ts
```diff
@@ -0,0 +1,232 @@
...
+      return { [packageName]: version };
+    }
+
+    case '@mui/material':
+    case '@mui/system':
+      return {
+        [packageName]: 'latest',
+        '@emotion/react': 'latest',
+        '@emotion/styled': 'latest',
+      };
```
LukasTy: Technically, this should not be needed.
We no longer have any demo using `@emotion` or `@mui/material`, or `@mui/system`. 🤔 
Naturally, not a blocker, just an idea for a cleanup.

LukasTy (approved):
Great work, looks and works great! 💯 
Thank you for the effort! 🙏

docs/package.json
```diff
@@ -21,6 +21,7 @@
     "@emotion/styled": "^11.14.1",
     "@mdx-js/loader": "^3.1.1",
     "@mdx-js/react": "^3.1.1",
+    "@mui/internal-docs-infra": "https://pkg.pr.new/mui/mui-public/@mui/internal-docs-infra@cdde8ab",
```
LukasTy: Nitpick: do you plan to update this to a released package version before merging?

dav-is: Yes, after https://github.com/mui/mui-public/pull/879 is merged

docs/src/utils/demoExportOptions.ts
dav-is: Ok, I removed these

PR #1534, opened 2025-03-05
Revert tailwind 4.0.9 to 4.0.3

mj12albert:
It breaks the styles: https://github.com/mui/base-ui/pull/1518#pullrequestreview-2656855691

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #2471, opened 2025-08-08
[internal] Automerge dev dependency renovate PRs

LukasTy:
This is to avoid the manual labour on non-risky renovate PRs, like: https://github.com/mui/base-ui/pull/2463.

It's already successfully used in `mui-x`: https://github.com/mui/mui-x/blob/49d5efda4018f7e50c4a2cb93ecc7ca4bad4a5f1/renovate.json#L9-L13

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -27,7 +27,7 @@
       "matchPackageNames": ["@mui/*", "!@mui/internal-*", "!@mui/docs"],
       "allowedVersions": "!/-dev/"
     },
-       {
+    {
       "matchDepTypes": ["devDependencies"],
       "matchUpdateTypes": ["minor", "patch", "pin", "digest"],
       "matchPackageNames": ["@mui/*"],


renovate.json
```diff
@@ -6,6 +6,11 @@
     "schedule": "* 0-5 3 * *"
   },
   "packageRules": [
+    {
```
mj12albert, line 9: > The order of rules matters, because later rules may override configuration options from earlier ones, if they both specify the same option

Just wondering does the tailwind group below need `"automerge": false` as it would match both groups? (first time I'm looking at the renovate config and docs btw)

Asking because this happened before 😅 https://github.com/mui/base-ui/pull/1534

LukasTy: Good point. 👍 
Hmm, that's a good question.
I'm not 100% sure if the `automerge` flag will affect any other custom `packageRules`.
Maybe @mui/infra knows if:
1. Will this rule only apply to PRs, which do not fall into any other group?
2. Will this rule apply to any dev dependency bump PR?

Janpot:
Wouldn't auto-merge make us more susceptible for supply-chain attacks?

LukasTy:
> Wouldn't auto-merge make us more susceptible for supply-chain attacks?

Possibly slightly - yes, but on the other hand - how would manual merging be different?
It's not that we manually check every bump in vulnerability database before merging. 🤔 
I would assume that it's dependebot, which is responsible for resolving any security issues.

Janpot:
> how would manual merging be different?

There would be more time between renovate PR opening and merging, to allow for npm to flag and pull the harmful update from the registry. Automerge essentially makes this time zero.

LukasTy:
> There would be more time between renovate PR opening and merging, to allow for npm to flag and pull the harmful update from the registry. Automerge essentially makes this time zero.

@Janpot, I take it you are proposing to close this PR and remove this rule from the `mui-x` repo to avoid any potential mishaps? 🤔

Janpot:
Personally, I [wouldn't](https://github.com/mui/mui-x/pull/18683#discussion_r2205471437) auto-merge any dependency we don't directly control. But I have to admit, I'm also less subjected to the pain of dealing with these PRs. To deal with it though I'd be more in favour for grouping more dependencies together and reducing the frequency, rather than to auto merge. 

> I take it you are proposing to close this PR and remove this rule from the `mui-x` repo

Not something I'm going to strongly advocate for.

LukasTy:
Thank you for your input, Jan.
I'll wait for input from other Base team members and then decide the best course of action in this regard. 👌

Janpot:
@LukasTy We discussed during code-infra chapter. We think it could be acceptable if we auto-merge with delay (`minimumReleaseAge`)

LukasTy:
@Janpot What strategy do you have in mind?
Set [`minimumReleaseAge`](https://docs.renovatebot.com/configuration-options/#minimumreleaseage) to let's say `3 days`, which would delay the creation of bump PRs, but override the `internalChecksFilter` on packages that we control? 🤔

Janpot:
The schema doesn't seem to allow it, but I see some [code samples](https://github.com/renovatebot/renovate/blob/7cf85c2d2490238f5d5a940b7972c84014325f88/docs/development/minimal-reproductions.md?plain=1#L71) in their repo that suggest we can use `minimumReleaseAge` in `packageRules` to set this value on a per-group basis. Would it be possible to set it to 3 days by default and reset to 0 for our own packages or security updates?

michaldudak:
Does automerge work with our setup? We used to have rules with it enabled, but since all our PRs require approval, a manual step was needed anyway.

LukasTy:
> Does automerge work with our setup? We used to have rules with it enabled, but since all our PRs require approval, a manual step was needed anyway.

It does work on `mui-x`, even though approvals are usually required.
Maybe this needs some special config, to allow the `renovatebot` a force-merge, bypassing the checks. 🤷‍♂️

LukasTy:
> The schema doesn't seem to allow it, but I see some [code samples](https://github.com/renovatebot/renovate/blob/7cf85c2d2490238f5d5a940b7972c84014325f88/docs/development/minimal-reproductions.md?plain=1#L71) in their repo that suggest we can use `minimumReleaseAge` in `packageRules` to set this value on a per-group basis. Would it be possible to set it to 3 days by default and reset to 0 for our own packages or security updates?

@Janpot Is this (https://github.com/mui/base-ui/pull/2471/commits/cbe929cfd787481fd26a32c7f797906b0f823380) the idea you had in mind?
I'm not sure if it works as expected, but probably the only way to test is to merge and see.
P.S. It passes renovate linting.

oliviertassinari:
> > Does automerge work with our setup? We used to have rules with it enabled, but since all our PRs require approval, a manual step was needed anyway.
>
> It does work on mui-x, even though approvals are usually required.
Maybe this needs some special config, to allow the renovatebot a force-merge, bypassing the checks. 🤷‍♂️

Ah, now I get it, it's why renovate was listed as someone who could bypass the approval request ruleset in mui-x. I have added this back for all repositories. There is one set of conditions, if all present that seems to make automerge a clear win:
- strong test coverage for the dependency
- patch or minor update
- PR opened with enough days of lag unless security fix or internally controlled
- dev dependency

Issue #2447, opened 2025-08-05
[radio] Does not match form data of native radio when submitted with no checked items


mirka:
# Bug report

## Current behavior

When a `Radio` is integrated into a form and is submitted without any items checked, the resulting form data for that `name` key is an empty string. This differs from a native radio group, where the form data would be `null`.

## Expected behavior

The form data for that key would be `null`, matching the native behavior.

(cf. [`Checkbox`](https://base-ui.com/react/components/checkbox#form-integration), which does return `null` for the key when submitted unchecked, matching the native checkbox behavior.)

## Reproducible example

[CodeSandbox](https://codesandbox.io/p/sandbox/gracious-dust-sjgrxs?file=%2Fsrc%2FApp.tsx)

## Base UI version

1.0.0-beta.2

## Which browser are you using?

Chrome

## Which OS are you using?

Mac OS

mj12albert:
@mirka Makes sense for us to use `null` for this as well ~ are you interested in opening a PR?

ingokpp:
I took a look at the code to understand this issue better as I'm interested in making my first contribution to the project. 👋🏻

After taking a look at the code, I think this issue could be tricky to fix. Achieving native-like behavior (returning null instead of "") could be done by conditionally rendering the input component only when there is a value. However, this would break other parts since we always need at least a hidden element for validation.

  Another solution could be to conditionally render the name attribute (when selected) rather than the entire input, but this would mean that we need to update all other tests which expect that the input always has a name attribute. The existing tests like expect(stringifiedFormData).to.equal('group=') are testing the bug, not the
  correct behavior. They were written to match the "incorrect" implementation rather than native HTML standards.

Since I'm not that deep into the project yet, I'm wondering: would conditionally rendering the name attribute be an acceptable change for the tests, or are there other considerations I should be aware of when it comes to how the hidden input's name property is used throughout the codebase?

PR #2473, opened 2025-08-08
[radio group] Return null in form data when no option selected

ingokpp:
Hey, this is my first contribution.

This PR is fixing the RadioGroup form submission to match native HTML behavior. When no radio button is selected, FormData.get() now returns null instead of an empty string. 

Changes

  - Conditionally set name attribute on hidden input only when a value is selected or when used within Field.Root
  - Updated tests to expect native behavior (null instead of "")
  - Added test specifically validating null return value

Closes https://github.com/mui/base-ui/issues/2447


packages/react/src/radio-group/RadioGroup.tsx
```diff
@@ -143,10 +143,10 @@ export const RadioGroup = React.forwardRef(function RadioGroup(
 
   const inputProps = mergeProps<'input'>(
     {
-      value: serializedCheckedValue,
+      value: serializedCheckedValue ?? '',
       ref: mergedInputRef,
       id,
-      name,
+      name: fieldName || serializedCheckedValue !== null ? name : undefined,
```
atomiks: Thanks for the contribution!

Bit confused by this part. Shouldn't it just be the following (plus, without the `null` change above in `serializedCheckedValue`):
```suggestion
      name: serializedCheckedValue ? name : undefined,
```

We can assume a falsy value should be undefined. And regardless of it being in a Field or not, it should have no `name` if there's no selected value, right?

packages/react/src/radio-group/RadioGroup.test.tsx
```diff
@@ -154,26 +154,67 @@ describe('<RadioGroup />', () => {
...
     expect(group.nextElementSibling).to.have.attribute('name', 'radio-group');
   });
 
+  it('should return null when no radio is selected (matching native behavior)', async ({
+    skip,
+  }) => {
+    if (isJSDOM) {
+      // FormData is not available in JSDOM
+      skip();
+    }
```
atomiks: `it.skipIf(isJSDOM)` is now a possible method
Not all tests were updated to this format yet, but we've been using this going forward

ingokpp: Ah, didn't know that. I updated the the test file to use the new way

packages/react/src/radio-group/RadioGroup.tsx
ingokpp: You are right! Removed the overcomplexity

ingokpp:
circleci failed due to api rate limit

packages/react/src/radio-group/RadioGroup.test.tsx
```diff
@@ -190,15 +222,15 @@ describe('<RadioGroup />', () => {
 
     submitButton.click();
 
-    expect(stringifiedFormData).to.equal('group=');
+    expect(formData!.get('group')).to.equal(null);
```
mj12albert:
```suggestion
    expect(formData.get('group')).to.equal(null);
```
I think the `!` shouldn't be necessary

packages/react/src/radio-group/RadioGroup.test.tsx
```diff
@@ -190,15 +222,15 @@ describe('<RadioGroup />', () => {
...
 
     await act(async () => {
       radioA.click();
     });
 
     submitButton.click();
 
-    expect(stringifiedFormData).to.equal('group=a');
+    expect(formData!.get('group')).to.equal('a');
```
mj12albert:
```suggestion
    expect(formData.get('group')).to.equal('a');
```

packages/react/src/radio-group/RadioGroup.test.tsx
ingokpp: Sadly no, becuase Formdata is declared like this `let formData: FormData;` so no initialization is happening which will cause TypeScript into thinking this might still be unassigned. 

Bildschirmfoto 2025-08-10 um 10 09 25

Even though we know that onSubmit will run before calling formData.get(...), the compiler can’t guarantee that.

mj12albert: Ah, in similar tests I actually put the `expect` inside the submit handler to get around this, e.g. https://github.com/mui/base-ui/blob/master/packages/react/src/slider/root/SliderRoot.test.tsx#L1851-L1857

(because we don't really use the non-null `!`)

packages/react/src/radio-group/RadioGroup.test.tsx
ingokpp: Sorry, I was on vacation and didn’t have my MacBook with me. I’ve now moved the except into the onSubmit handler to get rid of the ! usage. I also split the "should include the radio value in the form submission" test into two seperate tests, since otherwise we’d need to track the click count, which feels hacky to me.

PR #2476, opened 2025-08-09
[select][docs] Fix`scroll-margin/padding` styles

atomiks:

When a scroll arrow was visible in `alignItemWithTrigger` mode, the margin would not correctly take the absolutely-positioned scroll arrow element into account

1. Tailwind demos didn't have proper styles
3. CSS Modules was partially correct before, but this improves it with extra padding to ensure it shifts when it needs to scroll for the first time depending on the location of the item


atomiks:
Fixed in https://github.com/mui/base-ui/pull/2523

PR #2477, opened 2025-08-09
[docs] Fix iOS right swipe Tailwind styles

### LINKED ISSUE #2365: [toast][docs] Tailwind example animates down when swiping right on iOS
# Docs feedback

On the Tailwind CSS example, swiping right causes it to animate down instead of to the right on iOS. Doesn't happen on Mac Safari, Chrome, etc.

atomiks:
When swiping right on iOS, it should translate to the right, not down

Fixes #2365

---

Another thing I wanted to improve here is the `data-ending-style` translation. I noticed `calc(100% + 4rem)` is much better than `150%` because it works better for any toast height (small or large). Currently tall ones exit too fast, while small ones exit too sluggishly. 

However, it breaks the animation on iOS Safari when quickly creating toasts before the previous one's animation finishes (or rather makes it worse than it is currently). 

Either WebKit borks on the calc or something else is possibly wrong

```diff
&[data-ending-style] {
-  transform: translateY(150%);
+  transform: translateY(calc(100% + 5rem));
}
```

`translateY(100%) translateY(5rem)` also doesn't work

Curious if you had any thoughts/solutions @colmtuite @DiegoAndai @aarongarciah


atomiks:
> Another thing I wanted to improve here is the data-ending-style translation.

I'll try to address this in a separate PR

PR #2478, opened 2025-08-09
[scroll area] Overflow presence state attributes

### LINKED ISSUE #1100: [scroll area] Need a styling attribute to check for overflow presence
# Feature request

I attempted styling ScrollArea with a hard scrollbar and ended up with a selector that looked like this:
```css
  .ScrollAreaViewport {
    height: 100%;

    &:has(+ .ScrollAreaScrollbar[data-orientation='horizontal']) {
      padding-bottom: 1.5rem;
    }
}
```

I think we need data attributes both on Root and Viewport to target this comfortably.

Naming ideas:
- `data-has-overflow-x` / `data-has-overflow-y`
- `data-has-vertical-scrollbar` / `data-has-horizontal-scrollbar`

atomiks:
Closes #1100

Applies to `Root`, `Viewport`, `Scrollbar`
```jsx
/** Whether horizontal overflow is present. */
hasOverflowX: boolean;
/** Whether vertical overflow is present. */
hasOverflowY: boolean;
/** Whether there is overflow on the inline start side for the horizontal axis. */
overflowXStart: boolean;
/** Whether there is overflow on the inline end side for the horizontal axis. */
overflowXEnd: boolean;
/** Whether there is overflow on the block start (top) side. */
overflowYStart: boolean;
/** Whether there is overflow on the block end (bottom) side. */
overflowYEnd: boolean;
```


docs/reference/generated/scroll-area-root.json
```diff
@@ -11,7 +11,26 @@
...
     }
   },
-  "dataAttributes": {},
+  "dataAttributes": {
+    "data-has-overflow-x": {
+      "description": "Present when the scroll area has horizontal overflow."
+    },
+    "data-has-overflow-y": {
+      "description": "Present when the scroll area has vertical overflow."
+    },
```
atomiks: Present when `scrollWidth/Height` > `clientWidth/Height`

docs/reference/generated/scroll-area-root.json
```diff
@@ -11,7 +11,26 @@
...
+    "data-overflow-x-start": {
+      "description": "Present when there is overflow on the inline start side for the horizontal axis."
+    },
+    "data-overflow-y-end": {
+      "description": "Present when there is overflow on the block end (bottom) side."
+    },
+    "data-overflow-y-start": {
+      "description": "Present when there is overflow on the block start (top) side."
+    }
+  },
```
atomiks: Present when `.scrollTop/Left > 0` for `start`

I'm curious if there should also be CSS variables for the scroll amount if you want to transition the strength of e.g. an overflow gradient by the amount. Maybe CSS scroll-driven animations already solve this, though

docs/reference/generated/scroll-area-root.json
ciampo: "Present when the Scroll Area's content is taller than the viewport" ?

docs/reference/generated/scroll-area-root.json
ciampo: > I'm curious if there should also be CSS variables for the scroll amount if you want to transition the strength of e.g. an overflow gradient by the amount. Maybe CSS scroll-driven animations already solve this, though

I haven't needed such variable for my use case so far, but I agree that it would be interesting.

Here we could also decide to expose it only if/when needed?

ciampo:
@atomiks 

> Does the presence of data-overflow-{x,y}-start/data-overflow-{x,y}-end need to have a threshold? Right now if it's scrolled by 1px from the start, it gets added, and same if you're only 1px away from the end.

In my experience / for my use case, a threshold wouldn't be needed. I think we can be "lazy" and expose the option only if/when needed.

docs/reference/generated/scroll-area-root.json
```diff
@@ -12,7 +18,26 @@
...
   },
-  "dataAttributes": {},
+  "dataAttributes": {
+    "data-has-overflow-x": {
+      "description": "Present when the scroll area content is wider than the viewport."
+    },
+    "data-has-overflow-y": {
+      "description": "Present when the scroll area content is taller than the viewport."
+    },
+    "data-overflow-x-end": {
```
michaldudak, line 28: What's the advantage of having separate attributes for -start and -end?
We could use `data-has-overflow-x="start end"`.

atomiks: Edge scroll gradients work better with boolean attrs.

Compare:

```css
.fade-start[data-overflow-y-start],
.fade-end[data-overflow-y-end] {
  opacity: 1;
}
```

This breaks unless you use `~=` token matchers which isn't intuitive:

```css
.fade-start[data-has-overflow-y="start"],
.fade-end[data-has-overflow-y="end"] {
  opacity: 1;
}
```

michaldudak (approved):

PR #2479, opened 2025-08-09
[dialog][popover] Restore focus to popup when focused element is removed

### LINKED ISSUE #2475: Focus issues and inconsistencies between Popover and Dialog
# Bug report

Issues:
- `"trap-focus"` popovers close on focus out, but dialogs don't. Both close on focus out with `modal={false}`. Expected consistent behavior (to close popups).
- Removing elements from the popover/dialog DOM places focus on body in Chrome (I expected it to go to the popup)
- Removing an element from the popover DOM on pointerdown breaks close on focus out behavior for popovers (same expectation, move focus to the popup element)

## Reproducible example

[Link to CodeSandbox](https://codesandbox.io/p/sandbox/nervous-chandrasekhar-yhmpzg?file=%2Fsrc%2FApp.tsx%3A11%2C24)

## Base UI version

Latest (1.0.0-beta.2)

atomiks:
The `restoreFocus` prop uses the next tabbable element (in Menu/Select) but restoring to the `dialog` element root makes more sense for Popover/Dialog

For outside presses, it only uses `click` (intentional behavior) when a backdrop is present

Fixes #2475

https://codesandbox.io/p/sandbox/magical-mirzakhani-mkvfp2


mj12albert:
> pressing outside on an element that opens on mousedown (e.g. onto the CSS Modules select) causes both popups to be open simultaneously in the timeframe between mousedown and mouseup, which doesn't happen when focus isn't trapped

This case mentioned [in the issue](https://github.com/mui/base-ui/issues/2475#issuecomment-3170565202) is a Select inside `<Popover.Root modal="trap-focus" />` @atomiks ?

atomiks:
@mj12albert I was referring to the Popover hero demo on the docs and clicking outside onto the style selector

PR #2488, opened 2025-08-12
[tabs] Fix tab size rounding

### LINKED ISSUE #2235: [tabs] Rounding `--active-tab-width` and `--active-tab-height` can cause unwanted overflow
# Bug report

## Current behavior

When:
- styling the tab indicator with `--active-tab-width` or `--active-tab-height`
- the active tab has non-integer width/height that, if rounded, would be rounded up

The active tab indicator can cause the tablist to overflow, since the `--active-tab-width` or `--active-tab-height` can assume values that are larger than the underlying tab

## Expected behavior

No overflow is caused

## Reproducible example

N/A

## Base UI version

`1.0.0.beta.1`

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

The simplest solution is to `Math.floor` the active tab dimentions — ie. preferring a tab indicator up to `0.9px` smaller, but that never causes overflow.

atomiks:

Fixes #2235 @ciampo @cgatian


cgatian:
@atomiks rendering correctly now, and the logic is correct. Thank you!

PR #2508, opened 2025-08-15
[popover] Fix outside click after right clicking in popup

### LINKED ISSUE #2507: [popover] Right click in popover popup needs 2 outside clicks to close it
# Bug report

## Reproducible example

https://codesandbox.io/p/sandbox/dklnw3

1. Open the popover
2. Right click in the popup
3. First outside click: dismisses the system context menu, ✅
4. Second outside click does not close the popover popup
5. Needs another outside click

## Base UI version

beta-2

Firefox

## Which OS are you using?

Mac OS

Voiceover/JAWS etc.

## Additional context

baptisteArno:

Fixes #2507

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useDismiss.ts
+++ b/packages/react/src/floating-ui-react/hooks/useDismiss.ts
@@ -626,7 +626,8 @@ export function useDismiss(
 
   const handlePressedInside = useEventCallback((event: React.MouseEvent) => {
     const target = getTarget(event.nativeEvent) as Element | null;
-    if (!contains(elements.floating, target)) {
+    const isRightClick = event.button === 2;
+    if (!contains(elements.floating, target) || isRightClick) {
       return;
     }
     endedOrStartedInsideRef.current = true;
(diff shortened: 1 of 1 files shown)


packages/react/src/floating-ui-react/hooks/useDismiss.ts
```diff
@@ -626,7 +626,8 @@ export function useDismiss(
 
   const handlePressedInside = useEventCallback((event: React.MouseEvent) => {
     const target = getTarget(event.nativeEvent) as Element | null;
-    if (!contains(elements.floating, target)) {
+    const isRightClick = event.button === 2;
+    if (!contains(elements.floating, target) || isRightClick) {
```
atomiks: I guess `event.button !== 0` could also work. Do you have context on why this is needed?

michaldudak: `event.button !== 0` would actually be more correct, as it handles middle-clicks as well.

atomiks (approved):

atomiks:
Looks like Dialog still has the issue

baptisteArno:
Ok let me take a look

baptisteArno:
My bad, wrong condition operator 🙏

oliviertassinari:
I wonder if this doesn't deserve a regression test case.

baptisteArno:
> I wonder if this doesn't deserve a regression test case.

You're right, will add this :)

PR #2513, opened 2025-08-15
[code-infra] Replace codesandboxci with pkg.pr.new

Janpot:
Keep using codesandbox in the docs, but use pkg.pr.new preview published packages instead of csbci.

Remove codesandboxci

See https://github.com/mui/base-ui/pull/2503#issuecomment-3191484307.
One of https://github.com/mui/mui-public/issues/509.

### DIFF
--- a/docs/src/components/Demo/CodeSandboxLink.tsx
+++ b/docs/src/components/Demo/CodeSandboxLink.tsx
@@ -121,7 +121,7 @@ export function resolveDependencies(packageName: string): Record<string, string>
     case '@base-ui-components/utils': {
       const version =
         COMMIT_REF === undefined || SOURCE_CODE_REPO !== 'https://github.com/mui/base-ui'
-          ? 'latest'
+          ? 'latest' // #npm-tag-reference
           : `https://pkg.pr.new/mui/base-ui/${packageName}@${COMMIT_REF}`;
       return { [packageName]: version };
     }


docs/src/components/Demo/CodeSandboxLink.tsx
```diff
@@ -117,17 +117,13 @@ export function CodeSandboxLink({ title, description, ...props }: CodeSandboxLin
...
-      }
-      const shortSha = COMMIT_REF.slice(0, 8);
-      return {
-        '@base-ui-components/react': `https://pkg.csb.dev/mui/base-ui/commit/${shortSha}/@base-ui-components/react`,
-      };
+    case '@base-ui-components/react':
+    case '@base-ui-components/utils': {
+      const version =
+        COMMIT_REF === undefined || SOURCE_CODE_REPO !== 'https://github.com/mui/base-ui'
+          ? 'latest'
```
oliviertassinari:
```suggestion
          ? 'latest' // #npm-tag-reference
```

We should have this here, no? https://www.notion.so/mui-org/React-Major-release-process-0178410ef26941e6a4f333f80ded127a?source=copy_link#1dfcbfe7b660807580e2fab8f64096f4.

michaldudak:
Did Codesandbox add support for non-npm packages or are we now using VM-based devboxes now?

Janpot:
> Did Codesandbox add support for non-npm packages or are we now using VM-based devboxes now?

AFAIK, they always supported url dependencies. It was Stackblitz who doesn't support them in their "engineblock" runtime.

PR #2523, opened 2025-08-17
[select] Fix scroll arrows not propagating scroll fully to start/end of list

### LINKED ISSUE #2046: [select] Allow `ScrollArrow` parts to be rendered when `alignItemWithTrigger={false}`
# Feature request

## Summary

Usually you don't want these rendered in this mode, but in fallback mode to maintain a single pointer cycle, it can be desirable (macOS also does display them in fallback.)

In Radix, they are still rendered

## Motivation

Retain a single pointer cycle and could be less confusing as to why they aren't being rendered despite being specified in the JSX.

### LINKED ISSUE #2516: Select scroll arrows not visible when alignItemWithTrigger={false}
# Bug report

The `Select` component’s `ScrollUpArrow` and `ScrollDownArrow` are not visible when `alignItemWithTrigger={false}` is applied to the `Select.Positioner` component.

## Expected behavior

The scroll arrows should always be displayed whenever the Select popup content is scrollable, regardless of the `alignItemWithTrigger` setting.

## Base UI version

1.0.0-beta.2

Chrome

## Which OS are you using?

macOS

### LINKED ISSUE #2518: [select] Select.ScrollArrowUp fails to scroll to top if Select.GroupLabel is at the start
# Bug report

## Current behavior

`Select.ScrollUpArrow` doesn’t scroll all the way to the top when a `Select.GroupLabel` is present at the beginning of the `Select.Popup`.

## Expected behavior

When hovering over `Select.ScrollUpArrow`, the scroll position should continue moving until it reaches the very top of the `Select.Popup`, ensuring that any `Select.GroupLabel` at the start is fully visible.

## Reproducible example

[CodeSandbox](https://codesandbox.io/p/sandbox/bold-meadow-s2w3zr)

## Base UI version

1.0.0-beta.2

Chrome

## Which OS are you using?

Mac OS

atomiks:
Fixes #2518, scroll not propagating to the start/end of the list when there's content/labels at the start/end

https://deploy-preview-2523--base-ui.netlify.app/experiments/long-select

Also puts in initial work for #2516, #2046, but we need a `Select.List` component to properly solve it, see below:

---

A `Select.List` component, which will have the `overflow` style (this also applies to `Menu`), is needed in order to allow parts to escape the `Popup` element boundaries. `Select.Arrow` gets cut off while the `transform` is transitioning, and if you're rendering `Select.ScrollArrow` inside `Select.Popup` instead of `Select.Positioner`, you cannot create a "hitbox" (`::before` pseudo-element) to extend to the top / bottom of the viewport. A `Select.List` component will allow the scroll arrows to be placed inside `Select.Popup` while using `position: absolute` instead of `position: sticky` (which has some side effects with .scrollIntoView, since it affects the list layout).


packages/react/src/select/popup/SelectPopup.tsx
```diff
@@ -90,20 +101,41 @@ export const SelectPopup = React.forwardRef(function SelectPopup(
...
+    const list = listRef.current;
+    const firstItem = list.find((item) => item != null);
+    let lastItem: HTMLElement | undefined;
+    for (let i = list.length - 1; i >= 0; i -= 1) {
+      const item = list[i];
+      if (item) {
+        lastItem = item;
+        break;
+      }
+    }
```
LukasTy: Could [findLast](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast) be used?
Or don't we polyfill code for older browser support? 🤔

packages/react/src/select/scroll-arrow/SelectScrollArrow.tsx
```diff
@@ -84,9 +78,41 @@ export const SelectScrollArrow = React.forwardRef(function SelectScrollArrow(
...
+
+          const shouldShowUp = viewportTop > firstItemTop + 1;
+          const shouldShowDown = viewportBottom < lastItemBottom - 1;
+
+          if (direction === 'up') {
+            store.set('scrollUpArrowVisible', shouldShowUp);
+          } else {
+            store.set('scrollDownArrowVisible', shouldShowDown);
+          }
+        } else if (direction === 'up') {
```
LukasTy: This looks like a duplicate of code in `SelectPopup`.
Could it be extracted into a single place? 🤔

docs/src/app/(private)/experiments/scroll-area/inside-select.module.css
```diff
@@ -166,6 +166,7 @@
 }
 
 .ScrollArrow {
+  position: absolute;
```
LukasTy: The scroll arrow does not seem to work in this demo. 🤔 
Should it?
If not, maybe we could hide/remove it?

docs/src/app/(private)/experiments/menu/menu-fully-featured.tsx
```diff
@@ -55,7 +55,13 @@ export default function MenuFullyFeatured() {
             side={settings.side}
             align={settings.align}
           >
-            <Menu.Popup className={classes.Popup}>
+            <Menu.Popup
+              className={classes.Popup}
+              style={{
+                overflowY: 'auto',
+                maxHeight: 'var(--available-height)',
```
LukasTy: Nitpick: Would it make sense to provide these styles by default to avoid the need for everyone to specify them?
Otherwise, the menu doesn't work if it's taller than the viewport. 🙈 

P.S. The nested submenu also needs this style...

P.P.S. Such a fix could probably be extracted outside of this PR to avoid coupling context.

docs/src/app/(public)/(content)/react/components/select/demos/multiple/css-modules/index.module.css
```diff
@@ -144,6 +144,7 @@
 }
 
 .ScrollArrow {
+  position: absolute;
```
LukasTy: The number of breaking changes (things to adjust) for this to work is slightly concerning. 🙈 

Side question: Can users achieve a layout where scroll arrows do not overlap the scroll bars? 🤔 

By the way, I've checked the demo examples, WDYT about the following improvements?

### Before
Screenshot 2025-08-28 at 14 54 13

### After
Screenshot 2025-08-28 at 14 54 27

```diff
diff --git a/docs/src/app/(public)/(content)/react/components/select/demos/multiple/css-modules/index.module.css b/docs/src/app/(public)/(content)/react/components/select/demos/multiple/css-modules/index.module.css
index fb2a08cc..d63fca74 100644
--- a/docs/src/app/(public)/(content)/react/components/select/demos/multiple/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/select/demos/multiple/css-modules/index.module.css
@@ -145,17 +145,17 @@
 
 .ScrollArrow {
   position: absolute;
-  width: 100%;
+  width: calc(100% - 2px);
   background: canvas;
   z-index: 1;
   text-align: center;
   cursor: default;
-  border-radius: 0.375rem;
   height: 1rem;
   font-size: 0.75rem;
   display: flex;
   align-items: center;
   justify-content: center;
+  justify-self: center;
 
   &::before {
     content: '';
@@ -167,6 +167,8 @@
 
   &[data-direction='up'] {
     top: 0;
+    border-top: 1px solid var(--color-gray-200);
+    border-radius: 0.375rem 0 0;
 
     &::before {
       top: -100%;
@@ -175,6 +177,8 @@
 
   &[data-direction='down'] {
     bottom: 0;
+    border-bottom: 1px solid var(--color-gray-200);
+    border-radius: 0 0 0 0.375rem;
 
     &::before {
       bottom: -100%;
```

LukasTy (commented):
Nice work. 👍 
Leaving some questions and suggestions.

packages/react/src/select/scroll-arrow/SelectScrollArrow.tsx
atomiks: Yeah good spot, will see.

packages/react/src/select/popup/SelectPopup.tsx
atomiks: Hmm not sure the minimum of browsers we intend to support 🤔

docs/src/app/(private)/experiments/menu/menu-fully-featured.tsx
atomiks: This shouldn't be changed, removed. 

Some context: the overflow style causes `Menu.Arrow` to be clipped off unless we add a `Menu.List` component (as I mentioned in the original PR comment.) They need to be specified manually due to a tradeoff here for now.

docs/src/app/(public)/(content)/react/components/select/demos/multiple/css-modules/index.module.css
atomiks: The `multiple` demo shouldn't have scroll arrows since we recommend not using scroll arrows with `alignItemWithTrigger={false}`; I've removed them here

docs/src/app/(private)/experiments/scroll-area/inside-select.module.css
atomiks: Yeah, they shouldn't be specified in the JSX in this demo - removed

docs/src/app/(public)/(content)/react/components/select/demos/multiple/css-modules/index.module.css
atomiks: Also: I've exported `Select.DISABLE_SCROLLBAR_CLASS_NAME` because they need to disable the scrollbar when the scroll arrows are showing; they shouldn't both show simultaneously

docs/src/app/(private)/experiments/menu/menu-fully-featured.tsx
LukasTy: Gotcha. 👌

packages/react/src/select/popup/SelectPopup.tsx
LukasTy: Comparing https://github.com/mui/mui-x/blob/master/.browserslistrc with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast#browser_compatibility it looks like we should be good. 🤔 
And if not, Babel should polyfill those when building production code.

docs/src/app/(public)/(content)/react/components/select/demos/multiple/css-modules/index.module.css
LukasTy: Gotcha. 👌 
Well, but a similar comment regarding borders still applies to the main select demo as well as the long select experiment.
The arrow's background overlaps/cuts the border. 🙈

atomiks: Oh okay - they don't for me but I guess it can on other screens

Screenshot 2025-08-28 at 11 12 32 pm

LukasTy: Good spot. 👍 
It looks like this is exclusively a dark theme issue. 🤔

atomiks: Oh, yeah I see there. It uses a different border technique in dark theme - will update

docs/src/app/(private)/experiments/forms/form.module.css
```diff
@@ -296,6 +296,7 @@
 }
 
 .ScrollArrow {
+  position: absolute;
```
LukasTy: Complete nitpick, but scroll arrows can probably be removed from this module/page as well. 🤔

packages/react/src/utils/styles.tsx
```diff
@@ -1,10 +1,12 @@
 import * as React from 'react';
 
+export const DISABLE_SCROLLBAR_CLASS_NAME = 'base-ui-disable-scrollbar';
```
romgrk: This is already available as `styleDisableScrollbar.className`, why expose it in a different way? Also are we sure we want to expose this publicly now? IIUC, this is just for the experiment page, we should avoid adding public exports for that, they become subject to semver if we do.

atomiks: It's specifically for always disabling the scrollbar when you want the scroll arrows to show when `alignItemWithTrigger={false}`, as our docs demos and default is to allow them to show in that mode. The alternative is some kind of prop about scroll arrows or scrollbars on `Select.Positioner`/`Select.Popup`, perhaps `forceDisableScrollbar`

PR #2524, opened 2025-08-17
[useRender] Add support for data-* attributes

### LINKED ISSUE #2370: TypeScript Error with data-* Attributes in useRender.ElementProps
# Feature request

## Summary

I'm trying to handle the polymorphic behavior of a component using Base UI's useRender hook. I've structured my code based on the useRender documentation as follows:

```tsx
interface ButtonProps extends useRender.ComponentProps<'button'> {}
function SButton({ render = <button />, ...props }: ButtonProps) {
    const defaultProps: useRender.ElementProps<'button'> = {
        type: 'button',
        children: 'Click me',
        'data-disabled': false, // <---- TS Error
    };

    const element = useRender({
        render,
        props: mergeProps<'button'>(defaultProps, props),
    });
    return element;
}
```

However, I'm encountering a TypeScript error with the 'data-disabled' attribute. It appears that the useRender.ElementProps<'button'> type does not currently support dataset attributes.

I was wondering if it would be possible to add support for dataset properties to the useRender arguments, perhaps like this:

```tsx
interface ButtonProps extends useRender.ComponentProps<'button'> {}
function SButton({ render = <button />, ...props }: ButtonProps) {
    const defaultProps: useRender.ElementProps<'button'> = {
        type: 'button',
        children: 'Click me',

    };

    const dataset: useRender.DatasetProps = {
        'data-disabled': false,

        // or
        // disabled: false,
    }

    const element = useRender({
        render,
        dataset,
        props: mergeProps<'button'>(defaultProps, props),
    });
    return element;
}
```

I'm curious to hear your thoughts on this.

If there's a better way to add dataset attributes than what I've described, I'd be grateful if you could let me know.

Thank you!

Raghuboi:

Fixes #2370

Exposes the `stateAttributesMapping` parameter in `useRender` to enable converting component state into HTML data-* attributes.

## Changes
- Exposed `stateAttributesMapping` parameter in `useRender` hook
- Added documentation for state to data-* attribute conversion
- Added comprehensive test coverage

## Usage
```typescript
const element = useRender({
  render: <button>Submit</button>,
  state: {
    active: true,
    count: 42,
  },
});

const element = useRender({
  render: <button>Submit</button>,
  state: {
    isActive: true,
  },
  stateAttributesMapping: {
    isActive: (value) => (value ? { 'data-is-active': '' } : null),
  },
});
```


atomiks:
@Raghuboi `useRender` has a [`state`](https://base-ui.com/react/utils/use-render#undefined-state) property that can be mapped to data attributes. To do so, `useRender` should allow style hooks (`disableStyleHooks: false`)

The only issue there is `camelCase` state variables need to be converted to `kebab-case` using the `customStyleHookMapping` property, but I believe we wanted to call this `stateAttributesMapping` going forward (@romgrk). So we could instead use this name for the public hook only for the time being and pass it to `useRenderElement` like so: `customStyleHookMapping: stateAttributesMapping`

Though this doesn't solve things like `data-testid: '...'` which shouldn't really be part of `state`. Maybe you usually pass such attributes at the JSX callsite instead of in `useRender`, though.

atomiks (approved):
Thanks for the contribution! 

I may do some tweaks for the `CustomStyleHookMapping` type in a new PR

PR #2526, opened 2025-08-17
[navigation menu] Semantic element structure and `active` page prop

### LINKED ISSUE #2525: [navigation menu] enhance accessibility
# Feature request

### aria-current="page"

When screen reader users are navigating, it is necessary to use the aria-current="page" attribute to inform them of their current location. This helps prevent them from clicking the link to the page they are already on. Currently, the aria-current attribute is not being added, so I believe this area needs enhancement.

Using a method like that of [Radix UI](https://www.google.com/search?q=https://www.radix-ui.com/primitives/docs/components/navigation-menu%23link), which manages this via props, seems like a good approach.
Alternatively, the value prop currently only manages the open state of the Popup. If a Link is a child of NavigationMenu.Item, we could consider using the value prop to also manage whether the link is for the current page.

### Semantic HTML

A navigation element is a hierarchical structure containing a list of links. For that reason, I think it's better to use `<ul>` and `<li>` tags to clearly convey this structure to screen reader users. In fact, W3C examples also show navigation elements structured as `nav > ul > li > a`.

Of course, this is more of a best practice to consider rather than a mandatory standard. Therefore, I would like to ask for your thoughts on this topic.

[Radix UI](https://www.radix-ui.com/primitives/docs/components/navigation-menu)

- It adds the aria-current="page" attribute via an active prop.
- It is structured as `nav > ul > li > a`.

[Hero UI](https://www.heroui.com/docs/components/navbar)

- It adds the aria-current="page" attribute via an isActive prop.
- It is structured as `nav > ul > li > a`.

atomiks:
Fixes #2525

- `NavigationMenu.List` is `<ul>`
- `NavigationMenu.Item` is `<li>`
- `NavigationMenu.Link` receives `active` prop to specify `aria-current="page"`


Raghuboi (approved):

flaviendelangle:
Should we mark this PR as breaking since people can have visual regressions if they don't update there CSS with `list-style: none;`?

atomiks:
@flaviendelangle yeah, seems fair

packages/react/src/navigation-menu/link/NavigationMenuLink.tsx
```diff
@@ -19,13 +19,21 @@ export const NavigationMenuLink = React.forwardRef(function NavigationMenuLink(
...
   forwardedRef: React.ForwardedRef<HTMLAnchorElement>,
 ) {
-  const { className, render, ...elementProps } = componentProps;
+  const { className, render, active = false, ...elementProps } = componentProps;
 
   const { setValue, popupElement, positionerElement, rootRef } = useNavigationMenuRootContext();
   const nodeId = useNavigationMenuTreeContext();
   const tree = useFloatingTree();
 
+  const state: NavigationMenuLink.State = React.useMemo(
```
flaviendelangle, line 28: Do we need to have a `NavigationMenuLinkDataAttributes.ts` file with `data-active` or is the state of  `<CompositeItem />` not passed to the DOM as data attrs?

atomiks: Yep we do, nice catch

flaviendelangle (approved):
Nothing more to add 😆

PR #2527, opened 2025-08-17
[useRender] Add `defaultTag` parameter

atomiks:
Adds a `defaultTag: typeof React.JSX.IntrinsicElements` property to the `useRender` options, instead of needing to set a default to `useRender`, which uses `React.cloneElement` and is a slower default for perf sensitive primitives

### DIFF
--- a/packages/react/src/use-render/useRender.ts
+++ b/packages/react/src/use-render/useRender.ts
@@ -20,7 +20,7 @@ export function useRender<
   };
   renderParams.disableStyleHooks = true;
 
-  return useRenderElement(undefined, renderParams, renderParams);
+  return useRenderElement(renderParams.defaultTag, renderParams, renderParams);
 }
 
 export namespace useRender {
@@ -55,7 +55,7 @@ export namespace useRender {
     /**
      * The React element or a function that returns one to override the default element.
      */
-    render: RenderProp<State>;
+    render?: RenderProp<State>;
     /**
      * The ref to apply to the rendered element.
      */
@@ -77,6 +77,11 @@ export namespace useRender {
      * @default true
      */
     enabled?: Enabled;
+    /**
+     * The default tag to use for the rendered element.
+     * @default 'div'
+     */
+    defaultTag?: keyof React.JSX.IntrinsicElements;
   }
 
   export type ReturnValue<Enabled extends boolean | undefined> = Enabled extends false
--- a/docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/utils/use-render/demos/render-callback/css-modules/index.tsx
@@ -11,7 +11,7 @@ interface CounterState {
 interface CounterProps extends useRender.ComponentProps<'button', CounterState> {}
 
 function Counter(props: CounterProps) {
-  const { render = <button />, ...otherProps } = props;
+  const { render, ...otherProps } = props;
 
   const [count, setCount] = React.useState(0);
   const odd = count % 2 === 1;
@@ -32,6 +32,7 @@ function Counter(props: CounterProps) {
   };
 
   const element = useRender({
+    defaultTag: 'button',
     render,
     state,
     props: mergeProps<'button'>(defaultProps, otherProps),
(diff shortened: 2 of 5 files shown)


packages/react/src/use-render/useRender.ts
```diff
@@ -20,7 +20,7 @@ export function useRender<
   };
   renderParams.disableStyleHooks = true;
 
-  return useRenderElement(undefined, renderParams, renderParams);
+  return useRenderElement(renderParams.defaultTag, renderParams, renderParams);
```
oliviertassinari: The terminology feels wrong. "tag" does not seem part of the React language. 

 If we follow https://legacy.reactjs.org/docs/implementation-notes.html#mounting-host-elements:

```suggestion
  return useRenderElement(renderParams.defaultHostElement, renderParams, renderParams);
```

If we follow: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/64de0973102a5a89ac3c5c4f9913459fd93cb273/types/react/index.d.ts#L4051

```suggestion
  return useRenderElement(renderParams.defaultIntrinsicElement, renderParams, renderParams);
```

oliviertassinari: But I wonder, why only support host elements:
```suggestion
  return useRenderElement(renderParams.defaultComponent, renderParams, renderParams);
```

atomiks: Yeah I don't know the best word for this 😬  

I find the word "host" to be pretty abstract, and sort of implies knowledge of how React's internals work, while HTML "tag" is common knowledge. I  also see the word "element" representing the object value when writing JSX, not a string. `tag` matches `ref.current.tagName`, except in lowercase

So, maybe `defaultTagName`.

> But I wonder, why only support host elements

It's a shortcut for rendering the JSX element as a function and doesn't need to support components. The `render` prop itself is for that purpose

romgrk: "tag" sounds fine to me, if that prop only supports HTML tag names.

> and doesn't need to support components. The render prop itself is for that purpose

`render: <Something />` is more expensive, it forces more memory allocations and a `React.cloneElement()` call. Being able to pass `Something` the same way we can pass `"div"` would also allow better perf in the cases where it matters. And it could solve the naming topic by forcing this param to be named `defaultComponent`.

atomiks: The `useRender` hook is specifically for building primitive components where you aren't rendering another component (yet). The callsite of the component that has `useRender` is where they do the replacing with `render`

romgrk: Makes sense. I'm ok with `defaultTag` in that case. The other terms, "intrinsic elements" and "host elements", aren't used enough outside of react's codebase for them to be easily understandeable. Anyway the typings on `defaultTag` will make it very clear what goes in there.

oliviertassinari: > The useRender hook is specifically for building primitive components where you aren't rendering another component (yet)

I wonder if we won't see cases where we need to expand the need to support any components. From what I understand of https://github.com/radix-ui/primitives/blob/36d954d3c1b41c96b1d2e875b93fc9362c8c09e6/packages/react/slot/src/slot.test.tsx, this is supported.

> not a string. tag matches ref.current.tagName, except in lowercase

Right, ok, there is a way to justify the name, I framed this too strongly with "wrong" 😁.

romgrk: > I wonder if we won't see cases where we need to expand the need to support any components.

And if we pick `defaultComponent` for this param name now, we could later expand the typings without a breaking change.

mnajdova: I would use `defaultTag` as it's the only things we support now. If in the future we want to extend the behavior, we can add a new option `defaultComponent` and deprecate the `defaultTag` in favor of it. It would feel wrong naming it `defaultComponent` if it does not support custom components.

mnajdova (approved):

packages/react/src/use-render/useRender.ts
colmtuite: > It would feel wrong naming it defaultComponent if it does not support custom components.

Agree

oliviertassinari: > If in the future we want to extend the behavior

@mnajdova Right, I was mostly wondering about either there is a use case or not for any components.

I have tried to push the reflection on it, but I can't find a solid argument for extending support. The rationale I land on: this utile is slow, so we only want to allow React tree leaves to be swapped with it. For an element in the middle of the tree, instead, we can use a slot or composition.

PR #2528, opened 2025-08-17
[number field] Export formatNumber utilities

### LINKED ISSUE #2277: [utils][number field] Export formatNumber utilities
# Feature request

## Summary

`formatNumber` utilities should be exported. Whilst the user can write their own format function, exporting these utilities ensure that number formatting works exactly like `NumberInput` and users benefit from some niceties such as formatter caching.

N/A

## Motivation

This is useful for when users need to display number input values in a separate "review" section.

Raghuboi:

Closes #2277

Exports `formatNumber`, `formatNumberMaxPrecision`, and `getFormatter` utilities.


michaldudak:
`packages/react/src/utils` contain internal utilities and we don't want to export them.
The number formatting utility can, however, be useful in other projects as well, so we can move it to the `@base-ui-components/utils` package.

mj12albert:
I think `formatNumberMaxPrecision` should be kept internal since the use-case is very specific to number field: https://github.com/mui/base-ui/issues/1817

and it's same as `formatNumber` except with `maximumFractionDigits: 20` hardcoded

Raghuboi:
> I think `formatNumberMaxPrecision` should be kept internal since the use-case is very specific to number field: #1817
> 
> and it's same as `formatNumber` except with `maximumFractionDigits: 20` hardcoded

what about `getFormatter`?

oliviertassinari:
To what I understand, #2277 is about allowing developers to render the numbers the same way inside the NumberInput component and outside. So shouldn't the solution be about allowing developers to provide their custom formating function to the NumberInput? Simpler (less mental gymnastic) and covers more use cases than allowing developers to use the NumberInput formater outside.

In the current shape of the PR, it seems that we are not fixing #2277, but we are building the default number formating utils that we will use later on to solve number formating for locales across multiple components https://github.com/mui/material-ui/issues/24495. For example, @siriwatknp looked at a similar function https://github.com/mui/mui-x/issues/13222#issuecomment-3183062461

Raghuboi:
> To what I understand, #2277 is about allowing developers to render the numbers the same way inside the NumberInput component and outside. So shouldn't the solution be about allowing developers to provide their custom formating function to the NumberInput? Simpler (less mental gymnastic) and covers more use cases than allowing developers to use the NumberInput formater outside.
> 
> 
> 
> In the current shape of the PR, it seems that we are not fixing #2277, but we are building the default number formating utils that we will use later on to solve number formating for locales across multiple components https://github.com/mui/material-ui/issues/24495. For example, @siriwatknp looked at a similar function https://github.com/mui/mui-x/issues/13222#issuecomment-3183062461

i see, inverting control and allowing the user to plug their own formatter will cover more use cases

oliviertassinari:
@Raghuboi I think you can hold on a bit on the PR changes. I think best to align on what's the problem we want to solve, and then on how to best solve it. From what I can observe, it feel like we are not at this stage yet (see my comments as only ways to poke holes).

PR #2536, opened 2025-08-19
[popups] Support `initialFocus` and `finalFocus` functions

atomiks:
@vladmoroz this should solve your concerns with managing focus.

```js
// Do not move focus at all
<Dialog.Popup initialFocus={null} />
<Dialog.Popup finalFocus={null} />

// Move focus to a custom element at focus time
<Dialog.Popup initialFocus={(openType) => myElement} />
<Dialog.Popup finalFocus={(closeType) => myElement} />

// When a function *returns* null, it uses default behavior, which allows
// you to fallback to the default when your condition isn't satisfied

// When a function *returns* undefined, it does nothing at all.
// Use custom options, e.g. `preventScroll: true`
<Dialog.Popup initialFocus={() => {
  myElement.focus({ preventScroll: true });
}} />
```

Breaking: if you specify a function, return the element directly instead of a ref

### DIFF
--- a/packages/react/src/dialog/popup/DialogPopup.test.tsx
+++ b/packages/react/src/dialog/popup/DialogPopup.test.tsx
@@ -454,7 +454,7 @@ describe('<Dialog.Popup />', () => {
       await user.click(getByText('Open'));
 
       await waitFor(() => {
-        expect(getByText('Close')).toBeVisible();
+        expect(getByTestId('initial-outside')).toHaveFocus();
       });
 
       await user.click(getByText('Close'));
@@ -490,7 +490,7 @@ describe('<Dialog.Popup />', () => {
       await user.click(getByText('Open'));
 
       await waitFor(() => {
-        expect(getByText('Close')).toBeVisible();
+        expect(getByTestId('initial-outside')).toHaveFocus();
       });
 
       await user.click(getByText('Close'));
--- a/packages/react/src/dialog/popup/DialogPopup.tsx
+++ b/packages/react/src/dialog/popup/DialogPopup.tsx
@@ -149,6 +149,10 @@ export namespace DialogPopup {
     /**
      * Determines the element to focus when the dialog is opened.
      * By default, the first focusable element is focused.
+     *
+     * - `null`: Do not focus any element.
+     * - `RefObject`: Focus the ref element. Falls back to default behavior when `null`.
+     * - `function`: Return the element to focus. Called with the interaction type (`mouse`, `touch`, `pen`, or `keyboard`) that caused the open. Falls back to default behavior when `null` is returned, or does nothing when `void` is returned.
      */
     initialFocus?:
       | null
@@ -157,6 +161,10 @@ export namespace DialogPopup {
     /**
      * Determines the element to focus when the dialog is closed.
      * By default, focus returns to the trigger.
+     *
+     * - `null`: Do not focus any element.
+     * - `RefObject`: Focus the ref element. Falls back to default behavior when `null`.
+     * - `function`: Return the element to focus. Called with the interaction type (`mouse`, `touch`, `pen`, or `keyboard`) that caused the close. Falls back to default behavior when `null` is returned, or does nothing when `void` is returned.
      */
     finalFocus?:
       | null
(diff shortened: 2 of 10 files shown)


packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
```diff
@@ -133,7 +144,11 @@ export interface FloatingFocusManagerProps {
    * It can be also set to a ref to explicitly control the element to return focus to.
    * @default true
    */
-  returnFocus?: boolean | React.RefObject<HTMLElement | null>;
+  returnFocus?:
+    | boolean
+    | React.RefObject<HTMLElement | null>
+    | null
+    | (() => boolean | React.RefObject<HTMLElement | null> | HTMLElement | null | void);
```
mnajdova: Why we are not propagating here the reason too?

atomiks: I don't think it's useful in `finalFocus` to know what opened it. You'd instead probably need `CloseInteractionType`, but we can add that later if it's requested

mnajdova: Yeah, I meant a reason for closing. Happy to add it later, I thought that was one of the requirements we heard about.

atomiks: Added `closeType`, based on Vlad's feedback

packages/react/src/dialog/popup/DialogPopup.test.tsx
```diff
@@ -213,6 +303,116 @@ describe('<Dialog.Popup />', () => {
...
+      const trigger = getByText('Open');
+      await user.click(trigger);
+      const closeButton = getByText('Close');
+      await user.click(closeButton);
+      await waitFor(() => {
+        expect(getByTestId('input-to-focus')).toHaveFocus();
+      });
+    });
+
+    it('should support function returning element for finalFocus when closed', async () => {
```
mnajdova: This test seems to test the same as the previous test. Can we also add test for the reason type param?

packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
```diff
@@ -34,6 +36,29 @@ import { usePortalContext } from './FloatingPortal';
 import { useFloatingTree } from './FloatingTree';
 import { CLICK_TRIGGER_IDENTIFIER } from '../../utils/constants';
 
+function getEventType(event: Event, lastPointerType?: InteractionType): InteractionType {
```
mnajdova: I guess it's good we have this locally :))

mnajdova (commented):
Logic looks good 👌

packages/react/src/dialog/popup/DialogPopup.test.tsx
atomiks: @mnajdova updated!

packages/react/src/dialog/popup/DialogPopup.tsx
```diff
@@ -83,6 +82,22 @@ export const DialogPopup = React.forwardRef(function DialogPopup(
     titleElementId,
   });
 
+  // Default initial focus logic:
+  // If opened by touch, focus the popup element to prevent the virtual keyboard from opening
+  // (this is required for Android specifically as iOS handles this automatically).
+  const defaultInitialFocus = React.useCallback(
```
LukasTy: Nitpick: could this be moved to a constant or inside of `useDialogPopup`, since now it looks duplicated in several places. 🤔

atomiks: It's used in Popover too, and I think it's small enough that duplication isn't too much of a worry. I've changed to useEventCallback to remove the dependency array at least.

atomiks:
> WDYT, would it make sense to mention this new capability in any of the examples or add an additional example? 🤔

I think the types are good enough for this feature as it's somewhat of an advanced use case

packages/react/src/dialog/popup/DialogPopup.tsx
LukasTy: Oh, sure, not a big worry, just the added risk of possibly forgetting to update an instance if there is ever a need to change something in this function. 👌

Issue #615, opened 2024-09-08
[code-infra] Clean-up warnOnce duplication


oliviertassinari:
### Problem

The `warnOnce` logic is duplicated all over the place in the MUI codebase:

- MUI X: there are a couple of identical helpers https://github.com/search?q=org%3Amui+%22function+warnOnce%28%22&type=code
- MUI: duplicate the logic too: https://github.com/mui/mui-x/blob/13a2ff112d2a7dec41625b06db04db860f590f58/packages/x-data-grid/src/components/GridPagination.tsx#L132
- Material UI: duplicate the logic: https://github.com/mui/material-ui/blob/9e18971ee00c222c1ae317d673555187c9982ca2/packages/mui-material/src/Tabs/Tabs.js#L419
- Base UI: seems to do nothing yet, but we need to create something to improve the DX.

### Context

I initially worked on this because of https://github.com/mui/mui-x/pull/13911. This looks like a perfect helper for `@base-ui/internals` per https://www.notion.so/mui-org/engineering-mui-utils-purpose-9a9fc9da3a004864b6c4e1f4d1f24f95.

oliviertassinari:
Oh, I had completely forgotten about https://github.com/oliviertassinari/babel-plugin-transform-dev-warning. Maybe it would make sense to rewrite that for [Oxc](https://oxc.rs/).

PR #2537, opened 2025-08-19
[slider] Fix a11y issues with `Slider.Thumb`

### LINKED ISSUE #2512: [slider] Focus styling for slider thumb and the hidden `<input />`
# Get help

Is it intended that slider thumb is two focusable elements? The wrapper gets focused when tabbing into the slider via keyboard, but the hidden input in the thumb gets focused when moving the slider via mouse.

https://github.com/user-attachments/assets/13d2ec4e-1601-4dfe-b8ef-f12bb1c05987

https://base-ui.com/react/components/slider

On the video, I'm interacting with the pointer first and then continuing with the keyboard, which fails to show the `:focus-visible` styles because under the hood, the hidden input is focused.

I'm confused if it's a mistake or if I'm supposed to style both elements.

### LINKED ISSUE #2575: [slider] Broken TalkBack support
## Current behavior

https://github.com/user-attachments/assets/39b17b25-4a0f-4290-bd83-9fc1c4648602

I don't know if there is a way to write a test that would assert this, I can't find any. It actually looks like the only way to fix this is to use `<input type="range">`. When I dive into the source of Chrome, It seems that the issue is https://issues.chromium.org/issues/40816094.

> I started digging the issue and I discovered that Talkback *does not* use increase and decrease actions. Instead, it calls the "set value" action on the slider, which is supported because it's an <input> element.

There was some exploration to fix with evolving the platform e.g.  https://github.com/WICG/aom/issues/183, but it was abandoned.

## Expected behavior

Can increase/decrease with TalkBack.

## Reproducible example

Open https://master--base-ui.netlify.app/react/components/slider

## Base UI version

For example v1.0.0-beta.2

Chrome

## Which OS are you using?

Android

## Additional context

Same bug as:

- https://github.com/mui/material-ui/issues/23506. It's a regression compared to MUI Base. It works perfectly on https://v6.mui.com/base-ui/react-slider/.
- https://github.com/radix-ui/primitives/issues/2850

React Aria does this correctly.

mj12albert:
This PR fixes focus on the thumb when mixing keyboard and pointer modality:

https://github.com/user-attachments/assets/ab5bbef7-0ed7-4e95-ad79-cde5f7a0eb1e

#### Breaking changes
- `Slider.Thumb` renders `<div role="slider">` without the nested `<input type="range">`, the `render` prop no longer contains the third `inputProps` argument.
- An `index` prop is required on the thumb for range sliders to improve SSR.

#### Demos
- hero: https://deploy-preview-2537--base-ui.netlify.app/react/components/slider
- vertical & range sliders: https://deploy-preview-2537--base-ui.netlify.app/experiments/slider/slider 
- range slider SSR with `index` prop (disable JS in devtools): https://deploy-preview-2537--base-ui.netlify.app/react/components/slider#range-slider

Fixes #2512
Fixes #2575


Dakkers:
Is it possible for a Thumb to use `0` as the default value for index, and then to throw an error at runtime if multiple Thumbs use the same index value?

atomiks:
The `:focus-visible` ring shows when pressing the thumb after reload now; not the case before

packages/react/src/slider/root/SliderRoot.test.tsx
```diff
@@ -382,22 +363,22 @@ describe.skipIf(typeof Touch === 'undefined')('<Slider.Root />', () => {
...
+        <TestSlider value={0.2} min={-100} max={100} step={0.00000001} />,
       );
       const slider = getByRole('slider');
 
       await act(async () => {
         slider.focus();
       });
 
-      fireEvent.change(slider, { target: { value: '51.1' } });
+      await setProps({ value: 51.1 });
```
LukasTy: Is there any particular reason to change the testing approach?
Ideally majority of tests should be testing behaviors as closely mimicking user interactions. 🤔

LukasTy (commented):
Nice improvement. 👍 

However, am I missing something, or are there currently issues with using range slider in a form and this change only changes how that issue is reflected?
I wasn't able to parse the data from a range slider with `FormData`.
I was using the form experimen.

oliviertassinari:
> The :focus-visible ring shows when pressing the thumb after reload now; not the case before

But `:focus-visible` is broken now for regular interactions: 

https://github.com/user-attachments/assets/adcebc82-106c-4fa9-af6e-10f215ff050d

https://deploy-preview-2537--base-ui.netlify.app/react/components/slider

mj12albert:
> > The :focus-visible ring shows when pressing the thumb after reload now; not the case before
> 
> But `:focus-visible` is broken now for regular interactions

Showing the focus-visible ring when clicking/pressing the thumb (also upon dragging) matches Radix's behavior:


https://github.com/user-attachments/assets/558b7fd6-9814-4532-8fca-5f70af552628



Although [their docs](https://www.radix-ui.com/primitives/docs/components/slider) use `:focus` on the thumb, the behavior is the same when using `:focus-visible`: https://codesandbox.io/p/devbox/radix-ui-slider-forked-dc2k59

@atomiks The reloading issue seems to be that on first load, if the thumb is clicked/dragged too soon (before everything is hydrated?), the focus visible ring doesn't show on clicks/drags anymore until the page is reloaded or if you tab in/out of the thumb 🤔

mj12albert:
> Is it possible for a Thumb to use `0` as the default value for index, and then to throw an error at runtime if multiple Thumbs use the same index value?

It is possible, but there would be a bit more overhead to centralize all the individual `index` props somewhere (in a context) in order to check for duplicates

packages/react/src/slider/root/SliderRoot.test.tsx
mj12albert: Yeah a lot of the tests that use `fireEvent.change()` on the `input type="range"` were carried over from MUI Base as-is, and it didn't reflect an actual user interaction with a slider accurately to begin with; and with this PR the `div role="slider"` doesn't have a change handler anymore

Could probably update these tests further to not use `setProps`

atomiks:
> The reloading issue seems to be that on first load, if the thumb is clicked/dragged too soon (before everything is hydrated?), the focus visible ring doesn't show on clicks/drags anymore until the page is reloaded or if you tab in/out of the thumb 🤔

@mj12albert I had this issue with Menu/Select when calling `event.preventDefault()` in `onMouseDown`. The browser needs an interaction with the pointer on a focusable element to occur, and calling `.preventDefault()` stops that check, so it thinks you're using a keyboard when focusing with a pointer.

If you interact with anything else first, then the issue doesn't occur.

I fixed the issue by letting the trigger take focus then opening in `requestAnimationFrame` instead: https://github.com/mui/base-ui/pull/1846

> Showing the focus-visible ring when clicking/pressing the thumb (also upon dragging) matches Radix's behavior:

This is wrong though, the ring should only show when using a keyboard

mj12albert:
> This is wrong though, the ring should only show when using a keyboard

@atomiks Fixed the focus state when mixing pointer/keyboard input:

https://github.com/user-attachments/assets/ab5bbef7-0ed7-4e95-ad79-cde5f7a0eb1e

This matches RA's behavior now: https://react-spectrum.adobe.com/react-aria/Slider.html#example

CC @vladmoroz

packages/react/src/slider/root/SliderRoot.test.tsx
LukasTy: > and with this PR the `div role="slider"` doesn't have a change handler anymore

Isn't this a problem? 🤔 
How do we expect our users to test this component entry?

mj12albert: Generally by focusing `role="slider"` then using Mouse/Pointer/Key(e.g. Arrows) events, the newer non-legacy tests in this file use this, e.g. `fireEvent.keyDown`, `user.keyboard`

An ARIA widget without `input type="range"` doesn't involve change events: https://www.w3.org/WAI/ARIA/apg/patterns/slider/examples/slider-color-viewer/

packages/react/src/slider/value/SliderValue.tsx
```diff
@@ -29,9 +29,10 @@ export const SliderValue = React.forwardRef(function SliderValue(
 
   const outputFor = React.useMemo(() => {
     let htmlFor = '';
-    for (const thumbMetadata of thumbMap.values()) {
-      if (thumbMetadata?.inputId) {
-        htmlFor += `${thumbMetadata.inputId} `;
+    for (const thumbElement of thumbMap.keys()) {
+      if (thumbElement instanceof Element) {
```
atomiks: `isElement` is a local util

If there are other `instanceof`s, they should be updated as well since I don't think this check works in iframes/different doc contexts

oliviertassinari:
> Fixed the focus state when mixing pointer/keyboard input:

@mj12albert 👍 thanks for restoring this behavior. The native range slider also looks like it handles this correctly. See https://drafts.csswg.org/selectors-4/#example-58f4807d for why. `:focus-visible` is about allowing people to quickly find where the focused element is, but **only when the user might lose where the focus went** (`:focus` always shows it, which is annoying when it's obvious, e.g; with a slider: https://mui-org.slack.com/archives/C089BDL6XST/p1754572250519519). 

I'm also linking this to the bug in Radix's slider https://github.com/radix-ui/primitives/issues/2397, so the next Radix users who are frustrated about this can discover that we fixed this in Base UI. *For example, Leboncoin, who had to patch the bug: https://github.com/leboncoin/spark-web/pull/1617 and is now migrating: https://github.com/leboncoin/spark-web/pull/2711.*

Actually, this makes me think about https://youtu.be/vHFZFXtjKNE?si=CsgDtctJilFHnP34&t=1076, where people could be afraid to get new bugs in Base UI, which Radix doesn't have but are unclear about bugs Base UI fixed that Radix didn't.
One random idea, we could imagine using GitHub labels to keep track of those: `radix: solved`, `radix: unsolved`, `react-aria: solved`, `react-aria: unsolved` cc @colmtuite.

packages/react/src/slider/thumb/SliderThumb.test.tsx
```diff
@@ -42,6 +42,55 @@ describe('<Slider.Thumb />', () => {
     refInstanceof: window.HTMLDivElement,
   }));
 
+  describe.skipIf(isJSDOM)('server-side rendering', () => {
+    it('single thumb', async () => {
+      await renderToString(
```
oliviertassinari: This is a synchronous function, so?

```diff
diff --git a/packages/react/src/slider/thumb/SliderThumb.test.tsx b/packages/react/src/slider/thumb/SliderThumb.test.tsx
index c2cbad07..5993b26b 100644
--- a/packages/react/src/slider/thumb/SliderThumb.test.tsx
+++ b/packages/react/src/slider/thumb/SliderThumb.test.tsx
@@ -43,8 +43,8 @@ describe('<Slider.Thumb />', () => {
   }));

   describe.skipIf(isJSDOM)('server-side rendering', () => {
-    it('single thumb', async () => {
-      await renderToString(
+    it('single thumb', () => {
+      renderToString(
         <Slider.Root
           defaultValue={30}
           style={{
@@ -66,8 +66,8 @@ describe('<Slider.Thumb />', () => {
       );
     });

-    it('multiple thumbs', async () => {
-      const { container } = await renderToString(
+    it('multiple thumbs', () => {
+      renderToString(
         <Slider.Root
           defaultValue={[30, 40]}
           style={{
@@ -84,7 +84,7 @@ describe('<Slider.Thumb />', () => {
         </Slider.Root>,
       );

-      const [thumb0, thumb1] = Array.from(container.querySelectorAll('[data-testid="thumb"]'));
+      const [thumb0, thumb1] = Array.from(document.querySelectorAll('[data-testid="thumb"]'));

       expect(getComputedStyle(thumb0).getPropertyValue('left')).to.equal('30px');
       expect(getComputedStyle(thumb1).getPropertyValue('left')).to.equal('40px');
```

packages/react/src/slider/thumb/SliderThumb.tsx
```diff
@@ -136,9 +118,10 @@ export const SliderThumb = React.forwardRef(function SliderThumb(
     values: sliderValues,
   } = useSliderRootContext();
 
-  let renderPropRef = null;
-  if (typeof render !== 'function') {
-    renderPropRef = isReactVersionAtLeast(19) ? (render.props as any).ref : render.ref;
+  if (typeof document === 'undefined' && indexProp === undefined && sliderValues.length > 1) {
```
oliviertassinari: https://github.com/mui/mui-public/issues/207

```suggestion
  if (process.env.NODE_ENV !== 'production') {
    if (typeof document === 'undefined' && indexProp === undefined && sliderValues.length > 1) {
```

Also, the function name `warn()` feels confusing. I find`warnOnce()` clearer. Related to https://github.com/mui/base-ui/issues/615.

packages/react/src/slider/value/SliderValue.tsx
```diff
@@ -1,5 +1,6 @@
 'use client';
 import * as React from 'react';
+import { isElement } from '@floating-ui/utils/dom';
```
mj12albert, line 3: Should we copy this to `floating-ui-react/utils/element`? @atomiks

atomiks: Leave it external, as it's used across the `@floating-ui/dom` package and shared with other libraries

mj12albert:
> I wasn't able to parse the data from a range slider with `FormData`.
> I was using the form experimen.

@LukasTy Fixed in https://github.com/mui/base-ui/pull/2537/commits/d18411d55695b9c123c2da405cc8a02e6cb83641 the issue was that the experiment didn't use formData methods to get the slider values (is broken in main as well), the tests do it the right way: https://github.com/mui/base-ui/blob/master/packages/react/src/slider/root/SliderRoot.test.tsx#L1873

packages/react/src/slider/root/SliderRoot.test.tsx
LukasTy: But native `<input type="range" />` supports updating value on `change`. 🤔 
And it also works currently on the `Slider` component.

Is there a blocker for no longer supporting it?
Or is it a conscious decision? 🤔

mj12albert:
> This change breaks TalkBack
> It actually looks like the only way to fix this is to use `<input type="range">`

This answers @LukasTy's question in https://github.com/mui/base-ui/pull/2537#discussion_r2293812603 😓 

Changed to directly use `input type="range"` as the thumb in https://github.com/mui/base-ui/pull/2537/commits/54f7b760690a807418743a0c88584a69e441afcc which mostly works, but it's more annoying to reset the styles and in Firefox it seems impossible to do so:\
Screenshot 2025-08-25 at 12 22 20 AM
(in Firefox there's also a weird issue with the thumb jerking around on pointer down that doesn't occur when Quicktime is screen recording 🤯 )

We'd need to try this JSX:
```jsx
<>
  {thumbDiv}
  <input type="range" />
</>
```
and try to sync up the focus correctly in order to do all of:
- require only styling 1 element (the thumb div) 
- have correct `:focus-visible` when mixing keyboard/pointer 
- not having to special-case the `render` prop for the thumb part

mj12albert:
FWIW this change didn't fix the issue for Firefox yet

And while RA has the right behavior, they use a custom polyfill for focus visible instead of native `:focus-visible`, so it's possible there could be browser issues with transferring/syncing focus between a `div` and an `input` 🤔

oliviertassinari:
>  so it's possible there could be browser issues with transferring/syncing focus between a div and an input

@mj12albert To note that React Aria, Material UI, stick to have the same `<input>` focused for all modalities (touch, mouse, keyboard), and radix sticks to have the same `<span>`. Maybe this can solve the problem.

> Fixes https://github.com/mui/base-ui/issues/2545

The PR is likely in a draft mode right now; we didn't wire the aria with the new input strategy yet, so can't fully give feedback.
But from what I can see of the code in the demo: https://deploy-preview-2537--base-ui.netlify.app/react/components/slider#range-slider, we only have "Start", "End", we miss context.

SCR-20250824-qxzb

I would either expect it to behave like in https://react-spectrum.adobe.com/react-aria/Slider.html where there is uses `aria-labelledby` with the range label + the thumb label or behaves like in https://www.w3.org/WAI/ARIA/apg/patterns/slider-multithumb/examples/slider-multithumb/ (which seems simpler so better) where each thumb has an `aria-label` describing the whole value.

I also don't hear any of those two demos announce "Group" but we do, and this feels like noise, so we shouldn't.

mj12albert:
@oliviertassinari Does https://github.com/mui/base-ui/pull/2537/commits/87171bd5bf7036d0b577f62eecc704291d1d423d fix Talkback? I added back the range input as a sibling so at least it avoids hacking the render prop. Looks like Radix also has this issue: https://github.com/radix-ui/primitives/issues/2850

> To note that React Aria, Material UI, stick to have the same `<input>` focused for all modalities (touch, mouse, keyboard), and radix sticks to have the same `<span>`

If the input element can receive focus, then styling the thumb properly would require styling both the `div class="Thumb"` and also the "hidden" input (for the focus states) which is one of the points in Vlad's issue; RA and Material UI are polyfilling focus-visible to avoid this

With native `:focus-visible` it works in browsers except for Firefox where it's a bug on their side, using the keyboard on a already focused (by pointer) element doesn't apply `:focus-visible`: https://bugzilla.mozilla.org/show_bug.cgi?id=1688539

From the linked [web platform test](https://wpt.live/css/selectors/focus-visible-007.html), it seems that if the thumb has `:focus-visible`, using the mouse without moving focus (i.e. continue to drag the thumb) should not cancel it:

> Using keyboard while element is focused should trigger :focus-visible; using mouse without moving focus should not cancel it; moving focus using mouse should cancel it

So according to this RA's behavior is wrong as well and it should be like this:

https://github.com/user-attachments/assets/38503f81-e82b-48fd-9d18-56516f7afff5

CC @atomiks @vladmoroz

oliviertassinari:
> Does https://github.com/mui/base-ui/commit/87171bd5bf7036d0b577f62eecc704291d1d423d fix Talkback? I added back the range input as a sibling so at least it avoids hacking the render prop.

@mj12albert No, it doesn't work. <kbd>Tab</kbd> needs to move to an `<input type="range">` for  TalkBack to work.

> Looks like Radix also has this issue: https://github.com/radix-ui/primitives/issues/2850

I can reproduce, the Radix Primitive Slider is broken, but this makes sense, the tab focus a span, it can't work.

> If the input element can receive focus, then styling the thumb properly would require styling both the div class="Thumb" and also the "hidden" input (for the focus states), which is one of the points in Vlad's issue

I don't understand the issue; why can't we reproduce Material UI Slider or React Aria? There is no real styling overhead, to what I understand.

> With native :focus-visible it works in browsers except for Firefox where it's a bug on their side, using the keyboard on a already focused (by pointer) element doesn't apply :focus-visible: https://bugzilla.mozilla.org/show_bug.cgi?id=1688539

Oh, interesting. I think we could sleep on this one: a. it's not that important of a problem and b. it doesn't impact that many people (Firefox only).

mj12albert:
> Tab needs to move to an `<input type="range">` for TalkBack to work.

🤔 then is it not even working in [main](https://master--base-ui.netlify.app/react/components/slider)? Considering the input had `tabIndex={-1}` this whole time @oliviertassinari 

> There is no real styling overhead, to what I understand.

If the `input` was the only focusable element, the styling overhead would be having to do this:
```css
.Thumb {
 /* normal styles */

 & input[type="range"]:focus-visible {
   /* focus styles */
 }
}
```
The input would otherwise be completely un-exposed to developers

oliviertassinari:
> 🤔 then is it not even working in [main](https://master--base-ui.netlify.app/react/components/slider)? Considering the input had tabIndex={-1} this whole time

@mj12albert Right, so on HEAD, with VoiceOver and with TalkBack, when I navigate with them, I first move to the `<div tabindex="0">` then to the `<input type="range">`, so both are accessible to screen readers. On VoiceOver, both element responds to increase/decrease events. On TalkBack, only `<input type="range">` responds to increase/decrease events. So to answer your question, it's pretty broken, but there is kind of a way to use this as an end-user.

Issue created: #2575.

> If the input was the only focusable element, the styling overhead would be having to do this:

This seems fine: https://play.tailwindcss.com/vd5828nXRH

```html
<div class="m-4">
  <div class="has-[:focus-visible]:border-2 size-4 rounded-full bg-white outline outline-1 outline-gray-300 select-none focus-visible:outline focus-visible:outline-2 focus-visible:outline-blue-800">
   // Hidden inside Base UI 
   <input
      type="range"
      style="clip: rect(0 0 0 0); overflow: hidden; white-space: nowrap; position: fixed; top: 0; left: 0; border: 0; padding: 0; width: 100%; height: 100%; margin: -1px;"
    />
  </div>
</div>
```

mj12albert:
Split this into:
- https://github.com/mui/base-ui/pull/2578
- https://github.com/mui/base-ui/pull/2584

Issue #2533, opened 2025-08-18
[test] Argos CI doesn't run anymore


oliviertassinari:
## Current behavior

Last build was 5-6 week ago https://app.argos-ci.com/mui/base-ui. 

## Expected behavior

Argos CI runs, like we work toward with #1030. 

## Additional context

This is actually showing 4 bugs:

- [x] Nobody noticed? Issue created, solved
- [x] Argos CI should be a required step by the ruleset. Added, now, users with the maintainers role need to tick a checkbox to merge
- [x] The process crash, but exit 0. https://app.circleci.com/pipelines/github/mui/base-ui/12908/workflows/20a9ac28-f0f7-46a6-b599-db035684bb58/jobs/130822. It should exit with 1 https://github.com/mui/mui-public/pull/557
- [x] The process pass. #2538

LukasTy:
> * [ ]  The process crash, but exit 0. https://app.circleci.com/pipelines/github/mui/base-ui/12908/workflows/20a9ac28-f0f7-46a6-b599-db035684bb58/jobs/130822. It should exit with 1

It looks like Vitest might be to blame here.
I couldn't force tests to keep the browser open, by removing:
https://github.com/mui/base-ui/blob/caa205021e9cdc58d6e8e509a3adf381469fa40a/test/regressions/index.test.ts#L84-L86

It could be that Vitest is force-closing the browser and triggering the error. 🤔 
@JCQuintas do you know what could be to blame here based on your explorations?

JCQuintas:
> > * [ ]   The process crash, but exit 0. https://app.circleci.com/pipelines/github/mui/base-ui/12908/workflows/20a9ac28-f0f7-46a6-b599-db035684bb58/jobs/130822. It should exit with 1
> 
> It looks like Vitest might be to blame here. I couldn't force tests to keep the browser open, by removing:
> 
> [base-ui/test/regressions/index.test.ts](https://github.com/mui/base-ui/blob/caa205021e9cdc58d6e8e509a3adf381469fa40a/test/regressions/index.test.ts#L84-L86)
> 
> Lines 84 to 86 in [caa2050](/mui/base-ui/commit/caa205021e9cdc58d6e8e509a3adf381469fa40a)
> 
>  afterAll(async () => { 
>    await browser.close(); 
>  }); 
> It could be that Vitest is force-closing the browser and triggering the error. 🤔 [@JCQuintas](https://github.com/JCQuintas) do you know what could be to blame here based on your explorations?

[...]

PR #2538, opened 2025-08-19
[code-infra] Use correct command for Argos Push

LukasTy:
Attempt to fix https://github.com/mui/base-ui/issues/2533.

`push-argos` does not exist, it's `argos-push`:

Sidenote: this approach is slightly brittle. 🙈

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "test:webkit": "cross-env VITEST_ENV=webkit pnpm test:_unit",
     "test:webkit:ui": "pnpm test:webkit --browser.headless=false",
     "test:browsers": "cross-env VITEST_ENV=all-browsers pnpm test:_unit",
-    "test:argos": "code-infra push-argos --folder test/regressions/screenshots/chrome",
+    "test:argos": "code-infra argos-push --folder test/regressions/screenshots/chrome",
     "typescript": "tsc -b tsconfig.json",
     "validate-declarations": "tsx scripts/validateTypescriptDeclarations.mts",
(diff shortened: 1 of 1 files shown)


Janpot:
👀 Uhm, the cli should fail if a non-existing command is called.

_Edit: opening https://github.com/mui/mui-public/pull/557_

LukasTy:
Bump: I have also removed `test_regressions` from the React 18 suite.
If you need more info, check the description.
⚠️  We need `ci/circleci: Regression tests (React 18)` removed from required steps.

Janpot:
> Also, remove test_regressions from the React 18 suite to avoid race conditions and possible strange behaviors.
We have only one Argos comparison against master, it should be run on the same React version to avoid unforeseen issues.

I believe we should be able to [split builds](https://argos-ci.com/docs/build-splitting#leveraging-build-splitting-in-monorepos) with `buildName`. I can add an option to the cli if you want. It would instantly double our screenshots though.

LukasTy:
> I believe we should be able to [split builds](https://argos-ci.com/docs/build-splitting#leveraging-build-splitting-in-monorepos) with `buildName`. I can add an option to the cli if you want. It would instantly double our screenshots though.

I don't think it's worth it until we see clear issues with visual differences between React majors. 🤔 
Maybe for Base it's justifiable, given a significantly smaller screenshot footprint, but for other repos it's not justifiable. 🙈 

WDYT @michaldudak?

LukasTy:
Bump @michaldudak @Janpot.
We kind of need this merged to unblock pipelines. 🤔

Janpot (approved):

Issue #1530, opened 2025-03-04
[slider] Thumb alignment for inset slider designs


raclee7:
# Feature request

Is it possible to have the "handle/thumb" align with the track when it is at it's min & max?
Right now the handle is going beyond the bounds of the component, it will likely not feel as aligned once used.
I noticed the native html slider does this:
https://www.w3schools.com/howto/howto_js_rangeslider.asp

Native html slider
![Image](https://github.com/user-attachments/assets/68783230-23e4-4fd0-be8e-9d0278242ad0)

My use case below where thumb goes beyond the bound
![Image](https://github.com/user-attachments/assets/8110c368-8e4f-46fe-b78b-6592b5180074)

mj12albert:
Vlad proposed a `position: "inset" | "centered"` prop for the Thumb in https://github.com/mui/base-ui/issues/981

> [@colmtuite](https://github.com/colmtuite) I'd actually argue that inset thumb has to be the default
> 
> > the thumb position usually doesn't visually communicate the value accurately
> 
> Slider isn't a component for entering precise values anyway, so this doesn't seem like the most important constraint here. Sometimes it's important, sure—I'm well aware of [radix-ui/primitives#1966](https://github.com/radix-ui/primitives/issues/1966), but precise mapping is more of an edge case for slider in my opinion. It's fine as long as it's going to be an option.
> 
> Otherwise, we can't ignore a class of slider designs, like this one:
> 
[...]

Issue #981, opened 2024-12-06
[slider] Incorrect pointer tracking when padding is applied to the "poles" of the Control


vladmoroz:
- We need a prop or a CSS variable to make a slider to work like Radix/macOS slider—Radix and macOS sliders position the Thumb so that its edges are contained within the track (and we always align the Thumb's center to the track edge instead, so it ends up poking out of the Track at the edges)
- `minStepsBetweenValues` name wasn't too intuitive—will think if there's something better
- Cursor movement is tracked incorrectly when the Control part has padding

https://github.com/user-attachments/assets/16a1bf2e-2ead-4555-8b65-167653920706

Not sure what Control sizing has to do anything with the cursor movement; I'd expect it all to be done relative to the Track

### Done

- ✅ ~I expected divs instead of spans, would have saved me a couple situations when my width/height styles wouldn't work right away~ https://github.com/mui/base-ui/pull/1017
- ✅ ~Indicator has position absolute, which makes it harder to style. I don't understand how this is helping me. See [this PR](https://github.com/mui/base-ui/pull/980/files#diff-330f4fceb57c701e55a3f3f00a53cdad1f2505a40c351e5e49263c810cb30b25) for an example.~ https://github.com/mui/base-ui/pull/1017
  - ✅ ~Note: when closing this issue, please make sure that [this](https://github.com/mui/base-ui/blob/master/docs/src/app/(public)/(content)/react/components/slider/demos/hero/tailwind/index.tsx#L9-L10) and [this](https://github.com/mui/base-ui/blob/master/docs/src/app/(public)/(content)/react/components/slider/demos/hero/css-modules/index.module.css#L19-L20) kludge in the demos are removed.~  https://github.com/mui/base-ui/pull/1054
- ✅ ~Thumb has position absolute, which makes sense. However, if it's positioned absolutely, we need to specify relative to what, otherwise it's a gotcha for the dev to figure out and find the Thumb in a random position on the page first. We need to set `position: relative` on the Track.~ https://github.com/mui/base-ui/pull/1017

**Search keywords**:

mj12albert:
> I'd expect it all to be done relative to the Track

Movements are actually relative to the `Control` which is the "tracked area", and the `Track` is just a dumb component purely for styling the track, because without it you'd have to use a pseudo-element on the `Control` to style a "visible track" (or insert your own span)... naming issue?

> Cursor movement is tracked incorrectly when the Control part has padding

This is probably still a bug, I'll check it out

[...]

PR #2540, opened 2025-08-20
[slider] Add `thumbAlignment` prop

mj12albert:
Docs preview: https://deploy-preview-2540--base-ui.netlify.app/react/components/slider#inset-slider
Extra demos: https://deploy-preview-2540--base-ui.netlify.app/experiments/slider/inset

- Adds `thumbAlignment: 'center' | 'edge' | 'edge-client-only'` prop to enable inset positioning
- Can opt-out of pre-rendering with `'edge-client-only'` which reduces the bundle size but needs to wait for React hydration to render the thumbs and indicator
- The docs demo is the same as the hero demo except with the track height increased to equal thumb height(radius)
- Added a `--watch`/`-w` option to `pnpm inline-scripts`

Closes https://github.com/mui/base-ui/issues/981
Closes https://github.com/mui/base-ui/issues/1530


mj12albert:
API-wise it could be improved: `<Slider.Root inset renderBeforeHydration />`

Alternatively it could be `thumbPosition: 'centered' | 'inset' | 'inset-prerendered'`?
(Or `thumbAlignment` like in one of the Radix PRs)

docs/reference/generated/slider-root.json
```diff
@@ -36,6 +42,12 @@
       "type": "RefObject<HTMLDivElement>",
       "detailedType": "React.RefObject<HTMLDivElement> | undefined"
     },
+    "renderBeforeHydration": {
+      "type": "boolean",
+      "default": "false`",
+      "description": "Whether to render inset sliders before React hydration.\nThis minimizes the time that the thumb and indicator aren’t visible after server-side rendering.",
```
atomiks: Wording should probably be changed here: instead of "minimize", it "ensures" the thumb is also visible before JS loads, since the script is executed simultaneously as the thumb HTML gets parsed. Unless there's a tiny fraction of time where the track is visible without a thumb while parsing? 

Edit: I see Tabs already says that, but could be tweaked too

mj12albert: Yeah with this and tabs there is a very brief moment when it's not visible that isn't too hard to spot, so "minimize" is accurate

However `'inset-client-only'` explains it backward now: https://deploy-preview-2540--base-ui.netlify.app/react/components/slider#SliderRoot-thumbAlignment

atomiks:
I wonder if a different term than "inset" is better to pair with `centered`. 

Also `center` is more common in CSS, so that might be better as well.

`thumbAlignment?: 'center' | 'edge' | 'edge-client-only'`

Though `edge` is sort of ambiguous as it refers to the "outer" edge (of the thumb) not the "inner" edge 🤔  But are there any sliders that ever align to the inner edge? If so, the ambiguity may not be so relevant. Could also go explicit with `outer-edge`.

docs/reference/generated/slider-root.json
atomiks: I tested the demo page on Performance: 20x slowdown with Network: Slow 3G (+ disable cache; basically the worst case scenarios), and there were 0 frames where the thumb wasn't visible with the track - they paint simultaneously. Though maybe depending on how the CSS is specified in the app, it may be different? 

I feel like the word "minimize" is underselling how effectively it eliminates the flash.

> However 'inset-client-only' explains it backward now:

But yeah, I suppose this is no longer relevant anyway

mj12albert:
> I wonder if a different term than "inset" is better to pair with `centered`.

The values suggested in https://github.com/radix-ui/primitives/pull/2806 are pretty good and is closer to CSS vocabulary as well: ~`thumbAlignment: 'center' | 'contain' | 'contain-client-only'`~ (also avoids mentioning edges)

What do you think? CC @colmtuite 

Update: better pairs, either
- `center` + `edge`, or
- `overflow` + `contain`

aarongarciah:
@mj12albert `center` + `edge` sounds better because they describe the thumb position, whereas `overflow` + `contain` describe the behavior of the thumb based on the position.

atomiks:
I also think `center` + `edge` are fine for `thumbAlignment`. 

I don't think `edge` is ambiguous in the context of slider thumb alignment. 

Also `contain` sounds ambiguous in the sense that the thumb can _vertically_ overflow its container, it just can't horizontally overflow (for the regular horizontal slider).

mj12albert:
OK let's go with `center + edge`; I guess "inset slider" isn't really a term anyway (search for it and there is nothing related) so it's just as ambiguous

Should the [docs section](https://deploy-preview-2540--base-ui.netlify.app/react/components/slider#inset-slider) just be titled "Thumb alignment" instead?

aarongarciah:
> Should the [docs section](https://deploy-preview-2540--base-ui.netlify.app/react/components/slider#inset-slider) just be titled "Thumb alignment" instead?

+1

docs/reference/generated/slider-root.json
```diff
@@ -36,6 +36,12 @@
       "type": "RefObject<HTMLDivElement>",
       "detailedType": "React.RefObject<HTMLDivElement> | undefined"
     },
+    "thumbAlignment": {
+      "type": "'center' | 'edge' | 'edge-client-only'",
+      "default": "'center'",
+      "description": "How the thumb(s) are aligned relative to `Slider.Control` when the value is at `min` or `max`\n- `center`: The center of the thumb is aligned with the control edges\n- `edge`: The thumb(s) are inset within the control such that thumb edges are aligned with the control edges\n- `edge-client-only`: Same as `edge` but renders after React hydration on the client, gaining a smaller bundle size in return",
```
atomiks: "gaining" something smaller reads slightly weird

Maybe, "reducing bundle size in return"

atomiks (approved):

raclee7:
Thanks all!

colmtuite:
Is there anything to be said for `overshoot` and `contain`?

I think it becomes clear pretty quickly that we're talking about horizontal alignment, and once you realise that, `contain` makes sense.

I think `center` feels ambiguous to me, my brain thinks about the thumb being horizontally center aligned in the track. Whereas `overshoot` is clear to me.

`edge` doesn't do much for me, it seems confusing. Especially when I see it combined with "client", I start thinking about bleeding-edge, serverless, edge clients for network access and stuff like that.

Anyone else?

@

DiegoAndai:
> Anyone else?

My two cents:

I prefer `contain` and `overflow` of all the options. Hijacking Aaron's [argument](https://github.com/mui/base-ui/pull/2540#issuecomment-3326704485), I think we're describing behavior here, not position, so I prefer them over `center` or `edge`.

I don't like `overshoot`, sounds like an accident, like it's inadvertently going over

aarongarciah:
> I think we're describing behavior here, not position

Maybe I'm wrong, but I think most users would think about the position of the thumb, not the behavior. Subtle difference, but it's not the same. The prop is called "thumbAlignment" and the prop values descriptions talk about edges, alignment, inset, etc. so unless we change the language I think it's natural many users would think about it as something that controls the position (aka alignment) of the thumb.

<details>
  <summary>Current prop values descriptions</summary>

  > - `center`: The center of the thumb is aligned with the control edges
  > - `edge`: The thumb(s) are inset within the control such that thumb edges are aligned with the control edges
</details>

---

> overshoot

I think this term won't be very familiar to many non-english devs. At least I haven't heard it much. I'd prefer `overflow`.

I'm fine with `center` and `edge`, but ultimately the most important part is the docs doing a good job explaining it. So no strong preference.

colmtuite:
giphy

PR #2541, opened 2025-08-20
[website] Hide the Design Engineer role

mnajdova:
We are closing this role, I kept the page in case we want to re-open the role again.


packages/react/src/combobox/root/utils/index.ts
```diff
@@ -0,0 +1,110 @@
+import { serializeValue } from '../../../utils/serializeValue';
+import { useFilter } from './useFilter';
+
+export interface ComboboxGroup<Item = any> {
+  /**
+   * A label or value that identifies this group when required by the consumer.
```
atomiks: This seems to be a mistake, from reviewing Combobox?

mnajdova: Ooops, removing :D

Issue #2530, opened 2025-08-18
[field][input] Unable to set uncontrolled `Input` value via ref when wrapped in `Field`


mathias-falkenberg:
# Bug report

## Current behavior

I am trying to reset the value of an uncontrolled input by setting setting `value= ""` on the associated `ref`. This approach works as expected for a plain HTML `input` or stand-alone `Input`, even if a `defaultValue` is specified. If I wrap the Input in a Field, the next time the input gets focus, it reverts back to the `defaultValue`, if set.

## Expected behavior

I would expect similar behaviour to a plain HTML input where the above works.

## Reproducible example

[Example in CodeSandbox](https://codesandbox.io/p/sandbox/baseui-test-62ckvk)

The example provides three different inputs, all with `defaultValue` applied. After clicking the 'clear' button, focusing the `Input` wrapped in the `Field` (top one) resets the value to the `defaultValue`. The other two inputs do not exhibit the same behaviour.

## Base UI version

v1.0.0-beta.2

## Which browser are you using?

Chrome/Edge

## Which OS are you using?

Windows 11

mj12albert:
It reproduces when changing the value from `defaultValue` to anything else e.g. `inputRef.current.value="abc"`, not only `""`

ingokpp:
If it's ok for you, i would love to take a look at this.

mathias-falkenberg:
While the `value` of the wrapped `Input` can now be reset through the `ref` (after updating to 1.0.0-beta.3), it appears the state of the `data-` attributes is not updating, If setting `inputRef.value=""`, the `data-filled` attribute is still being applied, unlike if I manually delete the contents.

atomiks:
@mathias-falkenberg this is more so of a React issue where it doesn't fire `onChange` when mutating the ref imperatively. That's where the logic is to set attributes is.

Dispatching an event simultaneously when mutating would be necessary to get it to run:

```jsx
// Haven't tested, but would be something like this
// onChange in React is wired to `oninput`
ref.current.dispatchEvent(new Event('input', { bubbles: true }));
```

mathias-falkenberg:
Thanks for the explanation @atomiks . unfortunately, dispatching the event doesn't seem to be enough

PR #2543, opened 2025-08-20
[field][input] Prevent defaultValue reset on focus for uncontrolled inputs

ingokpp:
This will solve [#2530](https://github.com/mui/base-ui/issues/2530)

**Problem:**
Inputs wrapped in Field.Root were incorrectly resetting to their defaultValue when focused after being programmatically changed via ref (e.g., inputRef.current.value = ""). This behavior differed from plain HTML inputs and unwrapped Input components.

**Cause:**
The FieldControl component was always passing a value prop to the input element, making React treat it as controlled even when only defaultValue was provided. When the DOM value was changed directly, React's internal state wasn't updated, causing re-renders to reset the input.

 **My Solution:**
Modified FieldControl to conditionally pass either value (for controlled) or defaultValue (for uncontrolled) props to the input element, matching standard behavior.

Note: Before implementing the fix, I created some failing tests that reproduced the behaviour mentioned in the issue.

### DIFF
--- a/packages/react/src/field/control/FieldControl.tsx
+++ b/packages/react/src/field/control/FieldControl.tsx
@@ -90,6 +90,8 @@ export const FieldControl = React.forwardRef(function FieldControl(
     state: 'value',
   });
 
+  const isControlled = valueProp !== undefined;
+
   const setValue = useEventCallback((nextValue: string, event: Event) => {
     setValueUnwrapped(nextValue);
     onValueChange?.(nextValue, event);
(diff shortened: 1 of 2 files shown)


packages/react/src/field/control/FieldControl.tsx
```diff
@@ -114,9 +116,9 @@ export const FieldControl = React.forwardRef(function FieldControl(
         name,
         ref: inputRef,
         'aria-labelledby': labelId,
-        value,
+        ...(isControlled ? { value } : { defaultValue }),
         onChange(event) {
-          if (value != null) {
+          if (isControlled) {
```
atomiks: Do the tests still pass without changing this? The `useControlled` hook should fire `onValueChange` for both `defaultValue` and `value` when it gets changed (not via ref mutation though of course)

```js
const [value, setValueUnwrapped] = useControlled({
  controlled: valueProp,
  default: defaultValue,
  name: 'FieldControl',
  state: 'value',
});
```

ingokpp: They do! Adjusted

atomiks (approved):

PR #2562, opened 2025-08-22
[code-infra] Configure infra packages renovate groups

Janpot:
Reconfigure existing [groups](https://github.com/mui/mui-public/blob/a6f0bf1d38a4303edfc0b3b0a94e57807a95c399/renovate/default.json#L30-L39)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -24,13 +24,11 @@
     },
     {
       "groupName": "MUI public packages",
-      "matchPackageNames": ["@mui/*", "!@mui/internal-*", "!@mui/docs"],
       "allowedVersions": "!/-dev/"
     },
     {
-      "matchDepTypes": ["devDependencies"],
+      "groupName": "MUI infra packages",
       "matchUpdateTypes": ["minor", "patch", "pin", "digest"],
-      "matchPackageNames": ["@mui/*"],
       "automerge": true,
       "minimumReleaseAge": "1 minute"
     },


renovate.json
```diff
@@ -24,13 +24,11 @@
     },
     {
       "groupName": "MUI public packages",
```
michaldudak, line 26: This doesn't apply just to MUI public packages anymore

Janpot: I believe in the preset

```json
    {
      "groupName": "MUI public packages",
      "matchPackageNames": ["@mui/*"]
    },
```
creates the group, and
```json
    {
      "groupName": "MUI infra packages",
      "matchPackageNames": ["@mui/internal-*", "@mui/docs"],
      "followTag": "canary",
      "schedule": null
    },
```
filters out the infra packages into its own group

As far as I understand, a package can only belong to one group at a time. 

_edit: Will update the group, just to be sure_

LukasTy: Do we still need this group if it is coming from `mui-public`? 🤔

renovate.json
```diff
@@ -24,13 +24,11 @@
...
-      "matchPackageNames": ["@mui/*", "!@mui/internal-*", "!@mui/docs"],
       "allowedVersions": "!/-dev/"
     },
     {
-      "matchDepTypes": ["devDependencies"],
```
LukasTy, line 31: My hunch is also the same - that this group is breaking the `playwright` bump. 🙈

renovate.json
Janpot: This just configures it with `"allowedVersions": "!/-dev/"`. not sure whether that is still relevant

LukasTy: Don't we release packages as `canary` now? 🤔 
https://www.npmjs.com/package/@mui/internal-code-infra?activeTab=versions
Wasn't this only relevant to avoid bumps when we were releasing on each push to a branch?
cc @michaldudak

Janpot: There are still some packages published from the core repo. However I would discourage using the canary publish script. I think by now all latest versions are higher than the dev versions

renovate.json
Janpot: The group is created by `matchDepTypes` and `matchPackageNames`, I'm not sure it includes playwright.

_edit: ah, but [this one](https://github.com/mui/base-ui/blob/c70b61e4b340d0a8801303b94101056045b08f9f/renovate.json#L9-L14) may, yes!_

_edit 2: opening https://github.com/mui/mui-public/pull/583_

PR #2565, opened 2025-08-23
[select] Fix `IndexGuessBehavior` in React 18

atomiks:
I noticed my Combobox tests were failing on React 18 after some recent refactors and found this bug in React 18 in Select.

`IndexGuessBehavior` causes incorrect indices in React 18 Strict Mode:

React 18: https://codesandbox.io/p/sandbox/r5lzl3?file=%2Fsrc%2FApp.tsx%3A30%2C1
React 19: https://codesandbox.io/p/sandbox/72n5h6?file=%2Fsrc%2FApp.tsx%3A30%2C1

I'm not sure how bad the guess behavior is for concurrent feature support in general, but I've simply disabled it in dev mode here. 

To be specific, in dev mode in React 18, conditionally rendering different select items causes the index logic to break.

### DIFF
--- a/packages/react/src/select/root/SelectRoot.test.tsx
+++ b/packages/react/src/select/root/SelectRoot.test.tsx
@@ -1375,15 +1375,19 @@ describe('<Select.Root />', () => {
 
       const { user } = await render(<App />);
 
-      const renderExtra = screen.getByTestId('render-extra');
-      await user.click(renderExtra);
-
       const trigger = screen.getByRole('combobox');
       await user.click(trigger);
 
       await waitFor(() => {
         expect(screen.queryByRole('listbox')).not.to.equal(null);
       });
+
+      await user.keyboard('{Escape}');
+
+      const renderExtra = screen.getByTestId('render-extra');
+      await user.click(renderExtra);
+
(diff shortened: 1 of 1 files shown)


atomiks:
Ok, somehow it's working now in React 18 🙃  

I was testing in a Vite app to debug Combobox and now it's working there with latest pkg.pr.new  from master.

Select manages to correct itself in spite of this difference, since it does a second pass with new indices, but I changed how Combobox item registration works to be more understandable and it doesn't self-correct like Select

Screenshot 2025-08-23 at 11 52 17 pm

atomiks:
To close: `IndexGuessBehavior` is self-correcting by default in cases where it will break (i.e. React 18 dev mode). 

The index was not made reactive in the latest Combobox changes, which was the key issue for the React 18 tests

romgrk:
So all is good iiuc? I did implement the feature to work even in strict mode with double rendering.

atomiks:
@romgrk yep all good - took me a second to debug what was going on 😬

romgrk:
As usual, strict mode consumes more time to debug than it saves by avoiding bugs.

PR #2578, opened 2025-08-25
[slider] Fix interacting with TalkBack

mj12albert:
Demo: https://deploy-preview-2578--base-ui.netlify.app/react/components/slider

### Breaking changes:
- The `input type="range"` element receives focus and the thumb `div` is no longer focusable, focus styles targeting the thumb e.g. `.Thumb:focus-visible` should use `.Thumb:has(:focus-visible)` instead
- The `tabIndex` prop is moved from `Root` to `Thumb` where it gets forwarded to the `input`
- The thumb's `render` prop no longer contains the third `inputProps` argument, the `input` element is instead merged with `children`

Fixes https://github.com/mui/base-ui/issues/2575

### DIFF
--- a/packages/react/src/slider/root/SliderRoot.tsx
+++ b/packages/react/src/slider/root/SliderRoot.tsx
@@ -219,15 +219,11 @@ export const SliderRoot = React.forwardRef(function SliderRoot<
     thumbRefs.current?.[0]?.focus();
   });
 
-  useIsoLayoutEffect(() => {
-    if (valueProp === undefined || dragging) {
-      return;
-    }
-
+  if (process.env.NODE_ENV !== 'production') {
     if (min >= max) {
-      warn('Slider `max` must be greater than `min`');
+      warn('Slider `max` must be greater than `min`.');
     }
-  }, [dragging, min, max, valueProp]);
+  }
 
   useIsoLayoutEffect(() => {
     const activeEl = activeElement(ownerDocument(sliderRef.current));


oliviertassinari:
This work on my end with TalkBack 👍

oliviertassinari:
To avoid future regressions, maybe we can add a test after https://github.com/mui/base-ui/blob/475158265da79c9a9050a89c8f5678bd28c096c0/test/e2e/index.test.ts#L115 where tabbing should land on an input type range, and explain why it's done this way (not sure we actually need the e2e infrastructure for this test, maybe a simple unit test would support it).

packages/react/src/slider/thumb/SliderThumb.test.tsx
```diff
@@ -42,6 +17,239 @@ describe('<Slider.Thumb />', () => {
...
+  });
+
+  // AT (e.g. Android Talkback) may use increase/decrease actions to interact
+  // with the slider which works on `input type="range"` via change events, but
+  // not pure ARIA implementations using `div role="slider"`. The `input`
+  // element(s) must be the only focusable element(s).
+  // See:
+  // - https://issues.chromium.org/issues/40816094
+  // - https://github.com/mui/material-ui/issues/23506
+  describe('events', () => {
```
mj12albert, line 46: > where tabbing should land on an input type range, and explain why it's done this way (not sure we actually need the e2e infrastructure for this test, maybe a simple unit test would support it)

I've added a bunch of tests here around this @oliviertassinari

atomiks:
When pressing the thumb with the pointer on first load, I see the `:focus-visible` outline, or is this fixed in a different PR?

mj12albert:
> When pressing the thumb with the pointer on first load, I see the `:focus-visible` outline, or is this fixed in a different PR?

Yeah I made a different branch to tweak this so the Talkback regression can get merged sooner ~

packages/react/src/slider/thumb/SliderThumb.test.tsx
```diff
@@ -42,6 +17,239 @@ describe('<Slider.Thumb />', () => {
...
+              <Slider.Thumb onFocus={focusAndBlurSpy} onBlur={focusAndBlurSpy} />
+            </Slider.Control>
+          </Slider.Root>,
+        );
+        expect(document.body).toHaveFocus();
+        const input = screen.getByRole('slider');
+        expect(input).to.equal(container.querySelector<HTMLInputElement>('input[type="range"]'));
+
+        await user.keyboard('[Tab]');
+        expect(input).toHaveFocus();
```
oliviertassinari, line 65: We can probably add context for the next person.

```suggestion
        // We assert above that the tabbable elements of the slider are
        // input[type="range"] because TalkBack doesn't simulate keyboard events for increments
        // or decrements (proof: https://issues.chromium.org/issues/40816094). Instead, it triggers change events on those native slider inputs.
        expect(input).toHaveFocus();
```

packages/react/src/slider/thumb/SliderThumb.test.tsx
oliviertassinari: Yeah, ok, makes sense to me :+1:

oliviertassinari (commented):
> When pressing the thumb with the pointer on first load, I see the :focus-visible outline, or is this fixed in a different PR? 

I guess it's https://github.com/w3c/csswg-drafts/issues/5885. Is this even solvable? It seems to have been done on purpose by Chromium and Firefox.

mj12albert:
> I guess it's https://github.com/w3c/csswg-drafts/issues/5885. Is this even solvable?

It's related and indeed intentional on the browser's part, if we give up Firefox support for now (it's a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1775199) on their side) it seems to be solvable: https://github.com/mui/base-ui/pull/2584

packages/react/src/slider/thumb/SliderThumb.tsx
```diff
@@ -415,17 +372,16 @@ export namespace SliderThumb {
...
-     * Accepts a `ReactElement` or a function that returns the element to render.
+     * A blur handler forwarded to the `input`.
+     */
+    onBlur?: React.FocusEventHandler<HTMLInputElement>;
+    /**
+     * A focus handler forwarded to the `input`.
+     */
+    onFocus?: React.FocusEventHandler<HTMLInputElement>;
+    /**
+     * Optional tab index attribute forwarded to the `input`
```
LukasTy: Nitpick:
```suggestion
     * Optional tab index attribute forwarded to the `input`.
```

oliviertassinari:
> It's related and indeed intentional on the browser's part, if we give up Firefox support for now (it's a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1775199) on their side) it seems to be solvable: https://github.com/mui/base-ui/pull/2584

@mj12albert Right, ok, because https://github.com/mui/base-ui/blob/f56eea5c22c2c369b762660f94d1089f811266a3/packages/react/src/slider/control/SliderControl.tsx#L227-L229

But then I don't understand https://github.com/mui/base-ui/pull/2584, there are a lot of lines of code (on top of this PR). Why can't we fix https://github.com/mui/base-ui/pull/2578#issuecomment-3224523294 with:

```diff
diff --git a/packages/react/src/slider/control/SliderControl.tsx b/packages/react/src/slider/control/SliderControl.tsx
index caa2a0be..0f0ab451 100644
--- a/packages/react/src/slider/control/SliderControl.tsx
+++ b/packages/react/src/slider/control/SliderControl.tsx
@@ -226,7 +226,7 @@ export const SliderControl = React.forwardRef(function SliderControl(
       setActive(thumbIndex);
       thumbRefs.current?.[thumbIndex]
         ?.querySelector<HTMLInputElement>('input[type="range"]')
-        ?.focus();
+        ?.focus({ focusVisible: false });
     }
   });
```

If the interaction is a keyboard, it will focus the right element, so focus() won't be called.

packages/react/src/slider/root/SliderRoot.test.tsx
```diff
@@ -514,30 +455,24 @@ describe.skipIf(typeof Touch === 'undefined')('<Slider.Root />', () => {
...
-    const MAX = 750;
-
     it('should set the max and aria-valuemax on the input', async () => {
-      const { getByRole } = await render(
-        <TestSlider defaultValue={150} step={100} max={MAX} min={150} />,
-      );
-      const slider = getByRole('slider');
-
-      expect(slider).to.have.attribute('aria-valuemax', String(MAX));
-      expect(slider).to.have.attribute('max', String(MAX));
```
LukasTy, line 526: Shouldn't we keep the assertion that both `aria-valuemax` and `max` are set when `max` is provided? 🤔

packages/react/src/slider/root/SliderRoot.test.tsx
```diff
@@ -590,49 +523,41 @@ describe.skipIf(typeof Touch === 'undefined')('<Slider.Root />', () => {
...
-    const MIN = 150;
-
     it('should set the min and aria-valuemin on the input', async () => {
-      const { getByRole } = await render(
-        <TestSlider defaultValue={150} step={100} max={750} min={MIN} />,
-      );
-      const slider = getByRole('slider');
-
-      expect(slider).to.have.attribute('aria-valuemin', String(MIN));
-      expect(slider).to.have.attribute('min', String(MIN));
```
LukasTy, line 602: Same comment here as for the `max` prop

packages/react/src/slider/root/SliderRoot.test.tsx
```diff
@@ -641,28 +566,27 @@ describe.skipIf(typeof Touch === 'undefined')('<Slider.Root />', () => {
...
-      const thumbTwo = getByTestId('thumb-1');
+      const [input1, input2] = screen.getAllByRole('slider');
 
       await act(async () => {
-        thumbOne.focus();
+        input1.focus();
       });
 
-      fireEvent.keyDown(thumbOne, { key: 'ArrowUp' });
+      fireEvent.keyDown(input1, { key: ARROW_UP });
```
LukasTy: Nitpick: could be refactored to use `user.keyboard` as in other tests. 🙈

mj12albert:
> Why can't we fix [#2578 (comment)](https://github.com/mui/base-ui/pull/2578#issuecomment-3224523294) with:

The `focusVisible` option doesn't have good browser support ([caniuse](https://caniuse.com/mdn-api_htmlelement_focus_options_focusvisible_parameter)), and didn't seem to work very consistently in the supported ones

LukasTy (approved):
Looks great, nice work! 👍 
The component now works with screen readers (the current released version is completely broken on NVDA (with Firefox)  - kind of an industry standard)

I take it the focusing is handled in the separate PR, since that's the only regression I can see. 🤔

oliviertassinari:
> Option doesn't have good browser support ([caniuse](https://caniuse.com/mdn-api_htmlelement_focus_options_focusvisible_parameter))

@mj12albert Right, it only works with Safari and Firefox. 

> didn't seem to work very consistently in the supported ones

What are those issues? I couldn't notice problems.

Two thoughts:

1. We could sleep on this. Add `focusVisible: false` and wait. The problem itself is not that important. What's the likelihood that someone's first focus interaction on a page is a slider? As a user, do you want the extra bundle size & complexity to fix this? Maybe not.
2. If we don't do 1. let's have a clear step to remove all the unnecessary logic once supported. I have upvoted https://issues.chromium.org/issues/40834435.

mj12albert:
> What are those issues?

I meant that it didn't work when applied to this scenario, something like:
```jsx
<div
  id="thumb"
  onPointerDown={() => {
    requestAnimationFrame(() => {
      inputRef.current?.focus({ focusVisible: false });
    });
  }}
>
  <input
    ref={inputRef}
    style={visuallyHidden}
    type="range"
    onKeyDown={() => {
      requestAnimationFrame(() => {
        inputRef.current?.focus({ focusVisible: true });
      });
    }}
  />
</div>
```
Though if it had wider support it could indeed be used for a simpler fix here

oliviertassinari:
> I meant that it didn't work when applied to this scenario, something like:

@mj12albert Ok, let's wait then. Issue created #2592 for the future.

PR #2580, opened 2025-08-25
[website] Update Google Analytics tracking ids

michaldudak:
I set up two properties in GA ("base-ui-staging" and "base-ui-production") and updated the site with their tracking ids.

Work on https://github.com/mui/mui-public/issues/269.

### DIFF
--- a/docs/src/components/GoogleAnalytics.tsx
+++ b/docs/src/components/GoogleAnalytics.tsx
@@ -8,7 +8,7 @@ import { GoogleTagManager } from '../blocks/GoogleTagManager';
 
 const PRODUCTION_GA =
   process.env.DEPLOY_ENV === 'production' || process.env.DEPLOY_ENV === 'staging';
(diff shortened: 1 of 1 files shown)


LukasTy (approved):

oliviertassinari:
1. We have separate properties, but I wonder whether those should be only one: https://www.reddit.com/r/GoogleAnalytics/comments/11gw8rm/one_ga4_for_multiple_websites/
2. We will likely have a good chunk of the traffic that goes between domains. I'm on Base UI, I want an easy way to use this, Material UI (e.g. ton of people using shadcn UI). I'm on Sadcn UI I want to understand in detail what I can do with that component. So it feels like we should set cross-domain tracking (following the user journey across domains): https://support.google.com/analytics/answer/10071811?hl=en. To do this setup, it seems that it requires 1. But maybe we should sleep on this, unclear.

PR #2584, opened 2025-08-26
[slider] Fix thumb `:focus-visible` with mixed keyboard and pointer modality

mj12albert:
### Preview
Single thumb: https://deploy-preview-2584--base-ui.netlify.app/react/components/slider

https://github.com/user-attachments/assets/70cff343-46da-4ffa-9ee3-afda8b0598a0

Multiple thumbs: https://deploy-preview-2584--base-ui.netlify.app/react/components/slider#range-slider

https://github.com/user-attachments/assets/82c8eed1-e124-46e9-bc38-95e42e518900

Differences from other libraries:
- Using the pointer (only) on the thumb doesn't trigger `:focus-visible` which is a bug in Radix https://github.com/radix-ui/primitives/issues/2397
- Using the pointer on a thumb that already shows `:focus-visible` without moving focus (e.g. starting a drag) does not cancel `:focus-visible`, there's a [web platform test](https://wpt.live/css/selectors/focus-visible-007.html) related to this. This is [different from RA](https://react-spectrum.adobe.com/react-aria/Slider.html#example) where dragging the thumb that already has `:focus-visible` will cancel it which seems incorrect.

Doesn't work in Firefox due to a [browser bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1775199).

Fixes https://github.com/mui/base-ui/issues/2512

### DIFF
--- a/packages/react/src/slider/control/SliderControl.tsx
+++ b/packages/react/src/slider/control/SliderControl.tsx
@@ -3,7 +3,8 @@ import * as React from 'react';
 import { ownerDocument } from '@base-ui-components/utils/owner';
 import { useAnimationFrame } from '@base-ui-components/utils/useAnimationFrame';
 import { useEventCallback } from '@base-ui-components/utils/useEventCallback';
-import { activeElement } from '../../floating-ui-react/utils';
+import { activeElement, contains } from '../../floating-ui-react/utils';
+import type { Coords } from '../../floating-ui-react/types';
 import { clamp } from '../../utils/clamp';
 import type { BaseUIComponentProps, Orientation } from '../../utils/types';
 import { createBaseUIEventDetails } from '../../utils/createBaseUIEventDetails';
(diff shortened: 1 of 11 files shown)


mj12albert:
@atomiks It's not fully fixed in Firefox due to their bug, pointerdown on the thumb will not trigger the focus-visible ring (correctly), but a subsequent keydown fails to trigger the ring (it does in other browsers): https://codesandbox.io/p/sandbox/base-ui-slider-focus-visible-puzzle-7h9v55

atomiks:
@mj12albert that's a pretty minor detail to begin, with since it's only in FF, it doesn't matter - the fix here looks good

packages/react/src/slider/control/SliderControl.tsx
```diff
@@ -391,7 +379,18 @@ export const SliderControl = React.forwardRef(function SliderControl(
               return;
             }
 
-            focusThumb(finger.thumbIndex);
+            const pressedOnFocusedThumb = thumbRefs.current[finger.thumbIndex]?.contains(
+              activeElement(ownerDocument(control)),
+            );
```
atomiks: Use the `contains` util, which supports shadow DOM

atomiks (approved):

Issue #2586, opened 2025-08-26
[select] Support arbitrary object value types


atomiks:
# Bug report

## Current behavior

Only primitive `string`/`number` types are properly supported despite `value` being typed `any`. The `Select.Root` `items` prop and form value can't resolve the value or label from object value items.

Combobox has `itemToLabel/String` and `itemToValue` props to resolve the display and form submission value from the object value

## Expected behavior

`<Select.Item value={object}>` should be properly supported

This reduces or prevents the need to map data to the conforming `value`/`label` format before rendering it

```js
const items = [
  { id: 'test', name: 'Test' }
];

<Select.Root 
  defaultValue={items[0]} 
  items={items}
  itemToLabel={(item) => item.name} {/* <Select.Value> */}
  itemToValue={(item) => item.id} {/* hidden input value */}
>
[...] <Select.Item value={item}>
```

## Base UI version

1.0.0-beta.2

ingokpp:
@atomiks, Should this replace the current behavior (breaking change) or fallback to it if itemToValue is not provided?

atomiks:
@ingokpp it should be supplementary without breaking changes ideally, so fallbacks should be fine

oliviertassinari:
I guess the equivalent issue is https://github.com/mui/material-ui/issues/10845 (could be useful to see the challenges with fixing this in the past, the follow ups that we didn't think of initially).

Issue #516, opened 2024-08-02
[NumberField] Unit test broken when using non-English locale


michaldudak:
Running `pnpm test:unit` or `pnpm test:karma` on a machine with non-English regional settings results in failing tests. For example, when ran on Windows with Polish locale, 9 errors are reported:

<details>
  <summary>Test output</summary>
  
  ```
  1) <NumberField.Input />
       commits formatted value only on blur:

      AssertionError: expected input#:r1n2:[inputmode="numeric"][type="text"][autocomplete="off"][autocorrect="off"][spellcheck="false"][aria-roledescription="Number field"][value="1234"] to have value '1,234', but the value was '1234'
      + expected - actual

      -1234
      +1,234

      at Context.<anonymous> (packages\mui-base\src\NumberField\Input\/NumberFieldInput.test.tsx:136:27)
      at processImmediate (node:internal/timers:478:21)

  2) <NumberField />
       prop: smallStep
         should increment the value by the default `smallStep` prop of 0.1 while holding the alt key:

      AssertionError: expected input#:r1po:[inputmode="numeric"][type="text"][autocomplete="off"][autocorrect="off"][spellcheck="false"][aria-roledescription="Number field"][value="5,1"] to have value '5.1', but the value was '5,1'
      + expected - actual

      -5,1
      +5.1

      at Context.<anonymous> (packages\mui-base\src\NumberField\Root\/NumberFieldRoot.test.tsx:376:29)
      at processImmediate (node:internal/timers:478:21)

  3) <NumberField />
       prop: smallStep
         should decrement the value by the default `smallStep` prop of 0.1 while holding the alt key:

      AssertionError: expected input#:r1pq:[inputmode="numeric"][type="text"][autocomplete="off"][autocorrect="off"][spellcheck="false"][aria-roledescription="Number field"][value="5,9"] to have value '5.9', but the value was '5,9'
      + expected - actual

      -5,9
      +5.9

      at Context.<anonymous> (packages\mui-base\src\NumberField\Root\/NumberFieldRoot.test.tsx:384:29)
      at processImmediate (node:internal/timers:478:21)

  4) <NumberField />
       prop: smallStep
         should use explicit `smallStep` value if provided while holding the alt key:

      AssertionError: expected input#:r1ps:[inputmode="numeric"][type="text"][autocomplete="off"][autocorrect="off"][spellcheck="false"][aria-roledescription="Number field"][value="5,5"] to have value '5.5', but the value was '5,5'
[...]

PR #2590, opened 2025-08-28
[test] Add global locale mock to ensure consistent number formatting …

Powerplex:
…(#2586)

Related to [#516](https://github.com/mui/base-ui/issues/516)

I forked Base UI to investigate a bug on `NumberField` but noticed many tests failing on my machine on this component. This is because my locale was `fr-FR` and so `Intl` number formating was outputting `4,7` instead of `4.7` for example.

What if instead of adjusting the formatting in individual test files you added a global mock of the locale for `vitest` ?

`NumberField` is not the only component with tests failing because of that:
- `src/number-field/input/NumberFieldInput.test.tsx`
- `src/number-field/decrement/NumberFieldDecrement.test.tsx`
- `src/number-field/increment/NumberFieldIncrement.test.ts`
- `src/number-field/root/NumberFieldRoot.test.tsx`
- `src/progress/root/ProgressRoot.test.tsx`
- `src/meter/value/MeterValue.test.tsx`

after
before


atomiks:
Maybe it's just that the tests shouldn't hardcode the formatted assertion values and use `Intl.NumberFormat` in the `expect(...)` calls 🤔

Powerplex:
Ok I will update to that strategy, it was already what was done [here](https://github.com/mui/base-ui/pull/524/files), but as I saw it required updating 6 files, I wondered if a global mock would be preferred.

atomiks:
@Powerplex yeah I think what was done there is better as it asserts the real behavior without mocking

Powerplex:
@atomiks I updated with a test-by-test approach. WDYT ?

atomiks (approved):
Nice, this approach looks good!  Thanks for contributing

PR #2593, opened 2025-08-28
[slider] Add `index` prop to `Slider.Thumb`

mj12albert:
Improves SSR for range sliders by allowing the thumbs to be positioned in CSS without waiting for the DOM


docs/src/app/(public)/(content)/react/components/slider/page.mdx
```diff
@@ -34,6 +34,9 @@ import { Slider } from '@base-ui-components/react/slider';
 
 ### Range slider
 
-To create a range slider, provide an array of values with corresponding thumbs:
+To create a range slider:
+
+1. Pass an array of values and place `Slider.Thumb`s for each value in the array
+2. Specify a numeric `index` for each thumb that corresponds to the index of its value in the value array
```
oliviertassinari: Adding the `index` seems to be a progressive enhancement; only needed if someone needs SSR, so the wording in this line feels odd.

packages/react/src/slider/thumb/SliderThumb.tsx
```diff
@@ -123,6 +125,14 @@ export const SliderThumb = React.forwardRef(function SliderThumb(
     values: sliderValues,
   } = useSliderRootContext();
 
+  if (process.env.NODE_ENV !== 'production') {
+    if (typeof document === 'undefined' && indexProp === undefined && sliderValues.length > 1) {
+      warn(
+        'A `Slider.Thumb` was rendered on the server without an `index` prop, it must be specified for full SSR support.',
+      );
+    }
+  }
```
atomiks: I think it's fine not to have a warning if the demos specify it (and since it's progressive enhancement as mentioned)

docs/src/app/(public)/(content)/react/components/slider/page.mdx
mj12albert: Updated `2.` to state that it's for SSR

packages/react/src/slider/thumb/SliderThumb.tsx
oliviertassinari: But then, how would people, like Joy UI know they need to update their implementation? https://joy-ui.com/slider/#range-slider:

SCR-20250901-kvwr

---

Actually, on the need to fix Joy UI and Base UI, who is responsible for it? Per https://www.notion.so/mui-org/Introduction-5933bc8bccf544c7aca78d0033ded787?source=copy_link#a65e9ad76dd047cdaf62447c278e0cc2, this would be more for @mj12albert to handle? cc @mnajdova for alignment on how we split responsibilities.

mj12albert: > how would people, like Joy UI know they need to update their implementation?

The changelog entry should be enough?

oliviertassinari: Maybe, the warning seemed likr a solid one, it didn't seem to have downsides.

PR #2596, opened 2025-08-29
[select] Add `Select.List` component

### LINKED ISSUE #2046: [select] Allow `ScrollArrow` parts to be rendered when `alignItemWithTrigger={false}`

### LINKED ISSUE #2516: Select scroll arrows not visible when alignItemWithTrigger={false}

atomiks:
`<Select.List>` allows parts like `<Select.Arrow>` and `<Select.ScrollArrow>` to escape the `overflow: scroll` context

https://deploy-preview-2596--base-ui.netlify.app/experiments/long-select

Fixes #2516
Fixes #2046

## Breaking changes

It is now possible for `Select.ScrollArrow` to show when in fallback (`alignItemWithTrigger` deactivated). As a result, the styles should be updated like so to match before:

```css
.ScrollArrow {
  display: none; /* instead of display: flex */

  &[data-side="none"] { 
    display: flex;
  }
}
```

Otherwise, the demos have been updated to use `Select.List`, where the scroll arrows are also visible in fallback mode. If you're using `alignItemWithTrigger={false}` and don't want scroll arrows, you can omit `Select.List` (unless using `Select.Arrow`, which is rare) and `Select.ScrollArrow`s from the JSX entirely.


packages/react/src/select/positioner/SelectPositionerContext.ts
```diff
@@ -7,6 +7,8 @@ export interface SelectPositionerContext extends Omit<useAnchorPositioning.Retur
   setControlledAlignItemWithTrigger: React.Dispatch<React.SetStateAction<boolean>>;
   scrollUpArrowRef: React.RefObject<HTMLDivElement | null>;
   scrollDownArrowRef: React.RefObject<HTMLDivElement | null>;
+  hasScrollArrows: boolean;
+  setHasScrollArrows: React.Dispatch<React.SetStateAction<boolean>>;
```
romgrk: Why not use the store?

packages/react/src/select/store.ts
```diff
@@ -27,6 +27,7 @@ export type State = {
   triggerProps: HTMLProps;
   triggerElement: HTMLElement | null;
   positionerElement: HTMLElement | null;
+  scrollList: HTMLDivElement | null;
```
romgrk: The naming of every other element field ends with `-Element`. It would be more consistent to keep that convention. Considering that this field is assigned to a variable named `popupElement`, I would suggest to also name it `popupElement`.

packages/react/src/select/scroll-arrow/SelectScrollArrow.tsx
```diff
@@ -35,6 +36,14 @@ export const SelectScrollArrow = React.forwardRef(function SelectScrollArrow(
 
   const { mounted, transitionStatus, setMounted } = useTransitionStatus(visible);
 
+  useIsoLayoutEffect(() => {
+    setHasScrollArrows(true);
+    return () => {
+      // Assume that both scroll arrows (up/down) are rendered together.
+      setHasScrollArrows(false);
+    };
+  }, [setHasScrollArrows]);
```
romgrk: This approach forces a re-render right after mount, it's really not ideal.

packages/react/src/select/positioner/SelectPositionerContext.ts
flaviendelangle: In the store, it would only re-render `Select.List` on update, not `Select.Root` (even if the double render on mount is not optimal, not sure how we can avoid that one. We could of course add a prop to `Select.List` to let the user enable the behavior instead of detecting it, but it's not great)

packages/react/src/select/scroll-arrow/SelectScrollArrow.tsx
atomiks: Yeah, unfortunately, but it's the only way to avoid a prop

flaviendelangle: Nitpick but, could we have bugs if you have 2 arrows and one unmounts?

atomiks: Yeah though that's why I wrote

> // Assume that both scroll arrows (up/down) are rendered together.

I assume both will be specified in the JSX together + conditionally rendering them is likely to be rare

romgrk: > Yeah, unfortunately, but it's the only way to avoid a prop

Do we really want to avoid a prop? BUI is forcing re-renders all over the place, there's no way we can build a performant JUI with this approach being used all over the place. Couldn't the prop also be somewhat interesting to have on its own, just to disable the scrollbar?

> Nitpick but, could we have bugs if you have 2 arrows and one unmounts?

This could be solved by having that field be a count rather than a boolean flag.

atomiks: > BUI is forcing re-renders all over the place, there's no way we can build a performant JUI with this approach being used all over the place

Well it depends what the perf degradation is when measured. The degradation seems to be nonexistent when measured.

This is with 10 different runs (of each) measured on `long-select` demo with 20x slowdown in dev mode

With extra re-render:
> Average: 1513.6 ms
> 95% CI: [1472.2, 1555.0] ms

1. 1448ms
2. 1512ms
3. 1544ms
4. 1560ms
5. 1464ms
6. 1600ms
7. 1440ms
8. 1520ms
9. 1464ms
10. 1584ms

Without extra re-render (this one had a random outlier result of over 1700ms)
> Average: 1524.0 ms
> 95% CI: [1463.1, 1584.9] ms

1. 1472ms
2. 1512ms
3. 1568ms
4. 1456ms
5. 1504ms
6. 1744ms
7. 1456ms
8. 1544ms
9. 1496ms
10. 1488ms

> With the 6th value removed from both datasets (outlier cleanup):
> With avg: 1504.0 ms
> Without avg: 1499.6 ms

So, in this case, I'd prefer the auto-detection as it's a better API.

michaldudak: It doesn't really cause a re-render, as the state update is batched with others (we already trigger a rerender when certain components such as Positioner mount). Until React comes with a better way to communicate between component parts, this is not avoidable with the current API.

michaldudak:
Highlighting the last partially visible item causes this:

image

atomiks:
@michaldudak which demo? The `List` element has `overflow: hidden auto` so a highlighted shouldn't be able to escape the bottom bounds of the container

I also can't repro it on the docs hero:
Screenshot 2025-09-22 at 3 24 40 pm

edit: ah wait, you're probably checking the CodeSandbox demo, which means there's probably a `box-sizing: border-box` missing somewhere

packages/react/src/select/scroll-down-arrow/SelectScrollDownArrow.tsx
```diff
@@ -4,7 +4,7 @@ import { SelectScrollArrow } from '../scroll-arrow/SelectScrollArrow';
 import type { BaseUIComponentProps } from '../../utils/types';
 
 /**
- * An element that scrolls the select menu down when hovered.
+ * An element that scrolls the select menu down when hovered. Does not render when using touch input.
```
michaldudak, line 7: I'm almost certain someone will complain at some point about this ;)

atomiks: Maybe, but it doesn't work properly on touch to begin with, so we'd need to add support for pressing it not just hovering. 

But the pattern is poor UX, and I think we can convince people a scrollbar is better on touch devices (and to use custom visual indicators for touch if necessary, where rendering `ScrollArrow` in the JSX is unnecessary.)

PR #2599, opened 2025-09-01
[popups] Use boolean values for fallback `initialFocus`/`finalFocus` behavior

### LINKED ISSUE #1679: [Popover] Could we switch off `initialFocus`?
# Feature request

## Summary

I've noticed that [`Popover` has an `initialFocus` prop](https://github.com/mui/base-ui/blob/eefc4856a09b78311f58c793c31ac1bc22328f52/packages/react/src/popover/popup/usePopoverPopup.ts#L65-L68) that can take a ref or function returning a ref. By default it focuses the popup.

I'd like to be able to switch off initial focus.

[`FloatingFocusManager` supports a negative number to ignore the `initialFocus`](https://floating-ui.com/docs/floatingfocusmanager#negative-number).

Therefore, I am actually able to pass in `-1` into `Popover` and it'll do as expected. However I will get a type error.

## Motivation

I sometimes use this component to display transient flyouts that display information (and only gain focus if you interact with them). I need another element on the page to maintain its focus because it reacts to keyboard shortcuts.

atomiks:
Based on feedback from users, so `initialFocus={false|true}` and `initialFocus={() => false|true}`  are consistent.

Returning `null` in a function is the same as `true`, so that ref object or query selector nulls safely fall back to the default behavior.

> Marked internal since it's a post-PR refactor not yet in a release

Note - not necessarily a breaking change but should probably be mentioned: some users were using `initialFocus={-1}` to block initial focus even though it was an invalid type; now it's `initialFocus={false}` (closes #1679)

### DIFF
--- a/docs/reference/generated/alert-dialog-popup.json
+++ b/docs/reference/generated/alert-dialog-popup.json
@@ -3,14 +3,14 @@
   "description": "A container for the alert dialog contents.\nRenders a `<div>` element.",
   "props": {
     "initialFocus": {
-      "type": "RefObject<HTMLElement | null> | ((openType: InteractionType) => void | HTMLElement | null) | null",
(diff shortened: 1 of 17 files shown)


LukasTy (approved):
LGTM. 👍 

> Note - not necessarily a breaking change but should probably be mentioned: some users were using initialFocus={-1} to block initial focus even though it an invalid type; now it's initialFocus={false}. Closes https://github.com/mui/base-ui/issues/1679.

It doesn't look like it was listen in the public API, hence, yes, we could not treat it as a BC per se.
But if typing allowed a number before, I think it is definitely worth a mention.

PR #2601, opened 2025-09-01
[context menu] Fix default offsets when `align=center`

### LINKED ISSUE #2595: [context menu] Popup closes on mouseup with `align` set to `center`
# Bug report

## Current behavior

When the `align` prop is set to `center`, the popup closes by itself on mouseup.

## Expected behavior

The popup stays open so that I can select an option from the menu.

## Reproducible example

https://codesandbox.io/p/sandbox/thirsty-bouman-xysk5y

https://github.com/user-attachments/assets/d32dff5a-d5fc-4094-b5c4-116b12ad927c

## Base UI version

1.0.0-beta.2

Chrome Version 139.0.7258.139,

## Which OS are you using?

Mac OS 15.6.1

atomiks:
Fixes #2595 

If `align` is `center`, the `sideOffset` should remain unchanged from `0` so the item doesn't appear underneath the cursor


michaldudak (approved):
This change is OK, but I don't think it solves the issue entirely (see https://github.com/mui/base-ui/issues/2595#issuecomment-3248280443)

atomiks:
@michaldudak yeah, indeed - something like Vlad's suggestion about the 'hit area' (5px square) around the initial right click point is possibly needed to solve this properly

PR #2604, opened 2025-09-01
[useListNavigation] Ignore disabled item on initial focusing

mnajdova:
Same as https://github.com/mui/base-ui/pull/2579, but for components that use `useListNavigation`. Namely, initially the first non-disabled item is highlighted when opening the list - it's still possible to focus disabled items afterwards. 

In order for this to work, we still propagate empty `disabledIndices` as an option, but the hook skips the disabled items on initial focus.


atomiks (commented):
Change looks solid, though I wonder what happens with Select's positioning if the disabled item is currently selected given it overlaps the trigger?

mnajdova:
> Change looks solid, though I wonder what happens with Select's positioning if the disabled item is currently selected given it overlaps the trigger?

@atomiks if the selected item is first and disabled it focuses it. Otherwise it finds the first non-disabled item if no item is preselected. I think this behavior makes sense. Here is a sandbox to test the behavior: https://codesandbox.io/p/devbox/wizardly-shannon-dq8svg?file=%2Fsrc%2FApp.tsx&workspaceId=ws_HE9JRzFW36Y2kpBoSwdvZ5

atomiks (approved):

PR #2613, opened 2025-09-02
[utils] Add `timeout.startIfEmpty()`

romgrk:
This new method will allow me to get rid of [`throttle.ts`](https://github.com/mui/mui-x/blob/1d8854d9902cbabf034ddb469793c563918dd344/packages/x-internals/src/throttle/throttle.ts) once we migrate to BUI utils in X.

### DIFF
--- a/packages/utils/src/useTimeout.ts
+++ b/packages/utils/src/useTimeout.ts
@@ -24,6 +24,19 @@ export class Timeout {
     }, delay) as unknown as number; /* Node.js types are enabled in development */
   }
 
+  /**
+   * Executes `fn` after `delay` but only if no other timeout is ongoing.
+   */
+  startIfEmpty(delay: number, fn: Function) {
+    if (this.currentId !== EMPTY) {
+      return;
+    }
+    this.currentId = setTimeout(() => {
+      this.currentId = EMPTY;
+      fn();
+    }, delay) as unknown as number; /* Node.js types are enabled in development */
+  }
+
   isStarted() {
     return this.currentId !== EMPTY;
   }


packages/utils/src/useTimeout.ts
```diff
@@ -24,6 +24,19 @@ export class Timeout {
     }, delay) as unknown as number; /* Node.js types are enabled in development */
   }
 
+  /**
+   * Executes `fn` after `delay` but only if no other timeout is ongoing.
+   */
+  startIfEmpty(delay: number, fn: Function) {
```
michaldudak, line 30: `startIfNotPending`?

romgrk: Imo, doesn't sound better than `.startIfEmpty()`, although I'm not very satisfied with either name. I also thought about naming it `.throttle()`, although I've always been confused by "debounce" and "throttle" so I picked a name that describes directly what the function does.

We also have the `.isStarted()` method, it would be nice to tie it all together into two states ("empty" and "...."?) rather than use different terms ("started" and "pending"). I guess I'd be more fine with "pending" if we renamed that method (and that woud be a breaking change for the utils package - which I don't mind).

Open to more suggestions/feedback.

michaldudak: "pending" is a term already used in web animation APIs (along with "started", "idle", etc.) and it's pretty clear. "empty" is not intuitive for me at all (I'm not "filling" anything when setting a timeout).

romgrk: The `Timeout` instance can only hold a single `setTimeout` instance at once, that's what's being filled. The base method is also named `.start()` instead of `.set()` because it's not just a `setTimeout`, it also clears the pending `setTimeout` if it already holds one.

I guess I'm ok with "pending", but then I will rename `isStarted` to `isPending`, and this will be a breaking change (but we are in v0 so we can do anything anyway).

michaldudak: :+1: this works for me

romgrk: I've had a change of mind, `startIfNotPending` is more complex and verbose than `startIfEmpty` (due to the negation) and I don't like it. I wouldn't mind using "empty" and "pending" for the 2 states a `Timeout` instance can be in (and do the `isPending` rename), but I'd rather use the name for the empty state rather than a negation. I'd also be fine with another name for the empty state if we can figure out one that we agree on, although I don't see a better one.

I think it's important that people understand the `Timeout` class as a container (with a single slot) for a `setTimeout` call, so the "empty" state feels very natural to me. But I've often seen people call `.clear()` before calling `.start()` so I guess it's not clear for the rest of the team what a `Timeout` instance is, maybe we could improve on that.

michaldudak: How about `startIfIdle` or `startIfClear` (consistent with `clearTimeout` terminology). Both are short and without negation.

mj12albert: IMO "startIfIdle" is much less ambiguous than "clear"

michaldudak: @romgrk, any thoughts on the last two comments?

romgrk: I don't like them too much, "clear" isn't really correct because it's also the initial state before `start()` or `clear()` have ever been called. "idle" is less problematic. Here's my preference ranking for the suggested state pairs:
 1. empty/started (consistent with the terminology of the `.start()` method)
 1. idle/pending
 1. clear/pending

michaldudak: Can we have "idle" as a compromise we all seem to like (or at least not hate), then?

PR #2615, opened 2025-09-03
[internal] Fix the changelog generation script

michaldudak:
Quick fix of the changelog generation script after label changes

### DIFF
--- a/scripts/releaseChangelog.mts
+++ b/scripts/releaseChangelog.mts
@@ -239,12 +239,19 @@ function getComponentsFromLabels(labels: string[]): string[] {
     return [GENERAL_CHANGES_HEADER];
   }
 
+  const excludedScopeLabels = ['scope: support-infra', 'scope: code-infra', 'scope: docs-infra'];
+
   const components = labels
     .filter((label) => {
-      return label.startsWith('component:') || label.startsWith('hook:');
+      return (
+        (label.startsWith('component:') ||
+          label.startsWith('hook:') ||
+          label.startsWith('scope:')) &&
+        !excludedScopeLabels.includes(label)
+      );
     })
     .map((label) => {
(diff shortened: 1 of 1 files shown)


scripts/releaseChangelog.mts
```diff
@@ -239,12 +239,19 @@ function getComponentsFromLabels(labels: string[]): string[] {
     return [GENERAL_CHANGES_HEADER];
   }
 
+  const excludedScopeLabels = ['scope: support-infra', 'scope: code-infra', 'scope: docs-infra'];
```
michelengelen, line 242: Is this list exhaustive?

michaldudak: Not even close. There are currently [100 "scope: " labels](https://github.com/mui/base-ui/labels?q=%22scope%3A+%22), but we don't use most of them, so it's good enough for now.

michaldudak: I hope we can go back to using "component: " labels so this won't be necessary.

oliviertassinari: > Is this list exhaustive?

Related to https://github.com/mui/mui-x/blob/7681f3147bb1915222d39e941e53a736c64f7313/scripts/changelogUtils.mjs#L339-L342

```jsx
        case 'internal':
        case 'support-infra':
        case 'code-infra':
        case 'docs-infra':
          internalCommits.push(commitItem);
          break;
```

> There are currently 100 "scope: " labels, but we don't use most of them

We have this issue https://github.com/mui/mui-public/issues/150 about allowing us to select which repository a label gets synced to. It would avoid having this many scope labels.

In practice, I suspect that even though there are so many labels, it's relatively hard to use the wrong one, but to double-check.

Issue #2608, opened 2025-09-02
[select] Typeahead remembers the previous item after resetting the selected value


michaldudak:
Repro: https://codesandbox.io/p/sandbox/bold-kowalevski-pkk6wt?file=%2Fsrc%2FApp.tsx%3A11%2C1

1. Focus the Select
2. Press <kbd>A</kbd>
3. "A1" is correctly selected
4. Press Reset
5. Focus the Select
6. Press <kbd>A</kbd>
7. "A2" is now selected instead of "A1"

atomiks:
I can't reproduce in the CodeSandbox with beta.3

Edit: never mind - this applies to typeahead while closed. It works while open.

PR #2618, opened 2025-09-03
[useTypeahead] reset typeahead on external blur and start new session…

antonfrolovsky:
Fixes issue #2608

Updated **useTypeahead** to handle `onBlur` so the typeahead session resets only when focus leaves the reference ↔ floating pair.

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useTypeahead.ts
+++ b/packages/react/src/floating-ui-react/hooks/useTypeahead.ts
@@ -63,7 +63,7 @@ export interface UseTypeaheadProps {
  * @see https://floating-ui.com/docs/useTypeahead
  */
 export function useTypeahead(context: FloatingRootContext, props: UseTypeaheadProps): ElementProps {
-  const { open, dataRef } = context;
+  const { open, dataRef, elements } = context;
   const {
     listRef,
     activeIndex,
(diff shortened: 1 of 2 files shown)


mj12albert (approved):
Works great, thanks for the contribution @antonfrolovsky

packages/react/src/floating-ui-react/hooks/useTypeahead.ts
```diff
@@ -194,7 +200,27 @@ export function useTypeahead(context: FloatingRootContext, props: UseTypeaheadPr
     }
   });
 
-  const reference: ElementProps['reference'] = React.useMemo(() => ({ onKeyDown }), [onKeyDown]);
+  const onBlur = useEventCallback((event: React.FocusEvent) => {
+    const next = event.relatedTarget as Node | null;
+    const withinReference = next ? elements.domReference?.contains(next) === true : false;
+    const withinFloating = next ? elements.floating?.contains(next) === true : false;
```
atomiks: These should use shadow DOM `contains` util

antonfrolovsky: Okay, I'll check it out as soon as possible

atomiks:
The solution here might be simpler given it works when it re-opens:  https://github.com/mui/base-ui/issues/2608#issuecomment-3352377973

Maybe the resetting just needs to happen when `selectedIndex === null` as well as when it re-opens.

```js
  useIsoLayoutEffect(() => {
    if (open) {
      timeout.clear();
      matchIndexRef.current = null;
      stringRef.current = '';
    }
  }, [open, timeout]);
```

antonfrolovsky:
@atomiks Used the contains util in blur handling and added the `selectedIndex === null` check so typeahead resets when the selection clears.

michaldudak:
I found one other issue with `prevIndexRef` not being initialized properly at the start of new typing session and I fixed it as it was a one-liner.
@atomiks, do you have any other remarks or can we merge it?

michaldudak:
Thanks for your contribution, @antonfrolovsky!

PR #2626, opened 2025-09-04
[docs][combobox] Generate clear data attributes

atomiks:
Transitions are used for animating the button's visibility state


packages/react/src/combobox/clear/ComboboxClear.tsx
```diff
@@ -82,7 +86,6 @@ export const ComboboxClear = React.forwardRef(function ComboboxClear(
 
   const element = useRenderElement('button', componentProps, {
     state,
-    enabled: visible,
```
atomiks, line 85: Bug fix: removing this is necessary to ensure the `ending` transition can play

colmtuite (approved):

PR #2630, opened 2025-09-04
[experiments] Add priority combobox experiment

mnajdova:

### DIFF
--- a/docs/src/app/(private)/experiments/combobox/priority-combobox.tsx
+++ b/docs/src/app/(private)/experiments/combobox/priority-combobox.tsx
@@ -1,0 +1,182 @@
+'use client';
+import * as React from 'react';
+import { Combobox } from '@base-ui-components/react/combobox';
+
+export default function ExamplePopoverCombobox() {
+  return (
+    <Combobox.Root items={priorities} defaultValue={priorities[0]}>
+      <Combobox.Trigger className="flex bg-[canvas] h-10 min-w-[8rem] items-center justify-between gap-3 rounded-md border border-gray-200 pr-3 pl-3.5 text-base text-gray-900 select-none hover:bg-gray-100 focus-visible:outline focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-blue-800 data-[popup-open]:bg-gray-100 cursor-default">
+        <Combobox.Value>
+          {(priority: Priority) => (
+            <div className="flex items-center gap-2">
+              {priority.icon}
+              <span>{priority.label ?? priority.value}</span>
+            </div>
+          )}
+        </Combobox.Value>
+      </Combobox.Trigger>
+      <Combobox.Portal>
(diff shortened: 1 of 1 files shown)


colmtuite:
@mnajdova 

Very good 👌 Here's a first visual pass:

- Set font sizes on options, input, and trigger to 14px computed.
- Set the numbers font size to 12px computed.
- Make the numbers a lighter gray.
- Input should be 32px computed and have no space above or below it.
- Set the space between the popup edge and the menu items edge to 4px computed.
- Set trigger height to 32px computed.
- Set trigger horizontal padding to 12px computed.
- Remove `min-width` from trigger.
- Trigger and popup `border-radius` should both be `8px`, if that's a thing in TW? If not, they should be the same, whatever they're set to.
- Add `box-shadow: 0 1px 2px rgba(0,0,0,.025), 0 1px 3px rgba(0,0,0,.025)` to the trigger. Or try to get a similar shadow from TW. But I think it's possible to set custom opacity to shadows with TW?
- Add `background-clip: padding-box` to the trigger.
- Add `background-clip: padding-box` to the popup.
- Make the trigger hover background color way lighter, it should be barely noticeable.

atomiks:
`aria-label="Select country"` also needs to be changed.

To match Linear's, we need to add an `autoHighlight` prop for Combobox, not just Autocomplete. I made an issue: https://github.com/mui/base-ui/issues/2633

mnajdova:
> Trigger and popup border-radius should both be 8px, if that's a thing in TW? If not, they should be the same, whatever they're set to.

Yep, TW has this, done. What should be the radiuses of the higlighted items and the P badge?

atomiks:
The popup shifts itself (to remain anchored to the trigger) while the exit transition runs but this doesn't look great. Couple of options:

- Change trigger sizing to what you had before (trigger is as wide as the largest item text)
- Left-align the trigger button (but still centered in the experiment) so the popup's left edge doesn't shift
- Set `trackAnchor={false}` on `Combobox.Positioner`

mnajdova:
> The popup shifts itself (to remain anchored to the trigger) while the exit transition runs but this doesn't look great. 

I added the `trackAnchor={false}` prop to the `Combobox.Positioner`, but in general I don't like when elements change width, but I am not a design expert by any mean so I would follow what Colm would suggest :D

docs/src/app/(private)/experiments/combobox/priority-combobox.tsx
```diff
@@ -0,0 +1,191 @@
+'use client';
+import * as React from 'react';
+import { Combobox } from '@base-ui-components/react/combobox';
+
+function CustomCombobox(props: { items: Priority[] }) {
+  return (
+    <Combobox.Root items={props.items} defaultValue={props.items?.[0]}>
```
mnajdova:
```suggestion
    <Combobox.Root items={props.items} defaultValue={props.items?.[0]} autoHighlight>
```

PR #2631, opened 2025-09-04
[slider] Remove redundant hidden inputs

mj12albert:
Breaking: The `inputRef` prop is moved from `Root` to `Thumb`

Since the range input elements are focusable now (previously had `tabindex="-1"`) the hidden inputs are no longer needed to integrate with forms; also makes more sense for `inputRef` to be on the thumb since the input element isn't hidden to developers anymore

Follow up to https://github.com/mui/base-ui/issues/2575#issuecomment-3246992929

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Set up pnpm
         uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
       - name: Use Node.js 22.x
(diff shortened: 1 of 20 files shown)


packages/react/src/slider/thumb/SliderThumb.test.tsx
```diff
@@ -203,7 +203,7 @@ describe('<Slider.Thumb />', () => {
     });
   });
 
-  describe('prop: children', async () => {
+  describe('prop: children', () => {
```
oliviertassinari: Moving to #2632

```suggestion
  describe('prop: children', async () => {
```

LukasTy (approved):
LGTM. 👌 
Nice direction towards more simplicity and clarity. 👍

PR #2648, opened 2025-09-05
[docs] Eliminate errors in the Toolbar demos

chuganzy:
This is a very minor PR aims to resolve errors in the Toolbar's text editor sample.

### DIFF
--- a/docs/src/app/(private)/experiments/toolbar/text-editor.tsx
+++ b/docs/src/app/(private)/experiments/toolbar/text-editor.tsx
@@ -136,6 +136,7 @@ export default function App() {
               <Toolbar.Button
                 disabled={SELECT_DISABLED}
                 render={<Select.Trigger />}
+                nativeButton={false}
                 className={styles.select.Select}
               >
                 <Select.Value />
--- a/docs/src/app/(private)/experiments/toolbar/text-editor.tsx
+++ b/docs/src/app/(private)/experiments/toolbar/text-editor.tsx
@@ -331,7 +331,7 @@ export default function App() {
                     <Menu.Item className={styles.menu.Item}>Keyboard Shortcuts</Menu.Item>
                     <Menu.Item className={styles.menu.Item}>Release Notes</Menu.Item>
                     <Menu.Separator className={styles.menu.Separator} />
-                    <Menu.Root>
+                    <Menu.SubmenuRoot>
                       <Menu.SubmenuTrigger className={styles.menu.SubmenuTrigger}>
                         Debug
                         <ChevronRightIcon />
@@ -352,7 +352,7 @@ export default function App() {
                           </Menu.Popup>
                         </Menu.Positioner>
                       </Menu.Portal>
-                    </Menu.Root>
+                    </Menu.SubmenuRoot>
                     <Menu.Separator className={styles.menu.Separator} />
                     <Menu.Item className={styles.menu.Item}>Log Out</Menu.Item>
                   </Menu.Popup>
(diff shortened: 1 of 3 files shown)


docs/src/app/(private)/experiments/toolbar/text-editor.tsx
```diff
@@ -136,6 +136,7 @@ export default function App() {
               <Toolbar.Button
                 disabled={SELECT_DISABLED}
                 render={<Select.Trigger />}
+                nativeButton={false}
```
mj12albert, line 139: @chuganzy Did you encounter issues without this? SelectTrigger internally defaults this to `false` already: https://github.com/mui/base-ui/blob/master/packages/react/src/select/trigger/SelectTrigger.tsx#L46

chuganzy: Yes, without this it shows the following error:

```
A component that acts as a button was not rendered as a native <button>, which does not match the default. Ensure that the element passed to the `render` prop of the component is a real <button>, or set the `nativeButton` prop on the component to `false`.
```

This is because `Select.Trigger` is a div element, and `Toolbar.Button` also has `useButton` that checks if element is a button. I am not sure if that is valid to use non-button element as a `Toolbar.Button` though.

mj12albert: Ah I forgot the error only shows in dev mode 👍 

The [hero demos](https://github.com/mui/base-ui/tree/master/docs/src/app/(public)/(content)/react/components/toolbar/demos/hero) also need this then, would you mind updating them as well @chuganzy

chuganzy: Addressed in https://github.com/mui/base-ui/pull/2648/commits/56c26bc9fb848e6953700ba0f29c2cf2f0d5023d

docs/src/app/(public)/(content)/react/components/toolbar/demos/hero/tailwind/index.tsx
```diff
@@ -42,12 +42,12 @@ export default function ExampleToolbar() {
       </Toolbar.Group>
       <Toolbar.Separator className="m-1 h-4 w-px bg-gray-300" />
       <Select.Root defaultValue="Helvetica">
-        <Select.Trigger className="flex min-w-[8rem] h-8 text-sm font-medium items-center justify-between gap-3 rounded-md pr-3 pl-3.5 text-gray-600 select-none hover:bg-gray-100 focus-visible:outline focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-blue-800 data-[popup-open]:bg-gray-100 cursor-default">
+        <Toolbar.Button render={<Select.Trigger />} nativeButton={false}  className="flex min-w-[8rem] h-8 text-sm font-medium items-center justify-between gap-3 rounded-md pr-3 pl-3.5 text-gray-600 select-none hover:bg-gray-100 focus-visible:outline focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-blue-800 data-[popup-open]:bg-gray-100 cursor-default">
```
chuganzy: Made the Tailwind sample use `Select.Trigger` as `Toolbar.Button`, matching the CSS Modules sample.

mj12albert (approved):
Looks good ~ appreciate the PR @chuganzy

chuganzy:
Thank you for reviewing and handling these that quickly!

PR #2663, opened 2025-09-05
[experiments] Combobox creatable tags example

michaldudak:
Created a variation of the Creatable demo: new items are created without a dialog, and the comma key can be used to accept them.

The experiment also shows a reusable Combobox component.

https://deploy-preview-2663--base-ui.netlify.app/experiments/combobox/creatable-tags

### DIFF
--- a/docs/src/app/(private)/experiments/combobox/creatable-tags.tsx
+++ b/docs/src/app/(private)/experiments/combobox/creatable-tags.tsx
@@ -116,7 +116,7 @@ function Combobox(props: ComboboxProps) {
     }
 
     const normalized = trimmedValue.toLocaleLowerCase();
-    const existing = items.find((item) => item.trim().toLocaleLowerCase() === normalized);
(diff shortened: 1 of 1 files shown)


atomiks:
Is it intentional that the initial/default tags (after you've created some custom ones) appear before the custom ones when adding them?

e.g. "red, aqua, orange" -> orange appears before aqua

michaldudak:
Right, they were just appended to the list. I made the list sorted alphabetically now.

atomiks:
The sorting feels even weirder to me. I would expect the selected tags to just appear as the most recent one, no matter if they were created or were already defined in the list?

michaldudak:
I fixed the selection bug and reordered the items so that selected ones always appear at the top.

michaldudak:
It's actually a feature :) When using a comma, you'll likely want to select more items, so I kept the popup open.

PR #2664, opened 2025-09-05
[accordion] Use `useId` instead of composite index as fallback value

mj12albert:
Breaking: Accordion items must have an explicit `value` in order to be initially open.

In other words this is no longer supported:
```jsx
// ❌
<Accordion.Root defaultValue={[0, 1]}>
  <Accordion.Item>
  <Accordion.Item>
</Accordion.Root>
```

Alternative to https://github.com/mui/base-ui/issues/1880

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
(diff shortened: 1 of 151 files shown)


mj12albert:
This change means that `value` is still required if you want any of the panels to be initially open, since the default item value can't be predicted anymore (and is breaking)
```jsx
// no longer works
<Accordion.Root defaultValue={[0, 1]}>
  <Accordion.Item>
  <Accordion.Item>
</Accordion.Root>
```
It seems like an indirect way to require the `value` 😅 (for tabs as well) what do you think? @michaldudak @atomiks

PR #2665, opened 2025-09-05
[combobox] Fix initial closed typeahead

atomiks:
This handler belongs in `onFocus` to mount the list for first typeahead, same as with Select.


LukasTy (approved):
Nice, LGTM. 👌

PR #2666, opened 2025-09-05
[experiments] Remove @mui/system usage

mnajdova:
Removes the @mui/system as a dependency from the docs, by migrating all experiments that use styled() to use CSS modules.

TODO:
- [ ] do the transformation in the rest of the experiments that use styled()


mnajdova:
Closing in favour of https://github.com/mui/base-ui/pull/2667

PR #2670, opened 2025-09-05
[code-infra] Migrate error code extraction to code-infra

brijeshb42:

https://deploy-preview-2670--base-ui.netlify.app/production-error?code=1

### DIFF
--- a/docs/src/error-codes.json
+++ b/docs/src/error-codes.json
@@ -70,5 +70,9 @@
   "69": "Base UI: ToolbarRootContext is missing. Toolbar parts must be placed within <Toolbar.Root>.",
   "70": "Base UI: <Tooltip.Portal> is missing.",
   "71": "Base UI: TooltipPositionerContext is missing. TooltipPositioner parts must be placed within <Tooltip.Positioner>.",
-  "72": "Base UI: TooltipRootContext is missing. Tooltip parts must be placed within <Tooltip.Root>."
+  "72": "Base UI: TooltipRootContext is missing. Tooltip parts must be placed within <Tooltip.Root>.",
+  "73": "Base UI: useToastManager must be used within <Toast.Provider>.",
+  "74": "Base UI: PopoverTrigger must be either used within a PopoverRoot component or have the `handle` prop set.",
+  "75": "Base UI: PopoverTrigger must have an `id` prop specified.",
+  "76": "Base UI: selectors are required to call useState."
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12736,7 +12736,7 @@ snapshots:
       eslint-config-prettier: 10.1.8(eslint@9.37.0(jiti@2.6.1))
       eslint-import-resolver-typescript: 4.4.4(eslint-plugin-import@2.32.0(eslint@9.37.0(jiti@2.6.1)))(eslint@9.37.0(jiti@2.6.1))
       eslint-module-utils: 2.12.1(eslint-import-resolver-node@0.3.9)(eslint-import-resolver-typescript@4.4.4(eslint-plugin-import@2.32.0(eslint@9.37.0(jiti@2.6.1)))(eslint@9.37.0(jiti@2.6.1)))(eslint@9.37.0(jiti@2.6.1))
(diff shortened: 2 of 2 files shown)


packages/utils/src/store/createSelector.ts
```diff
@@ -125,7 +125,7 @@ export const createSelector = ((
   } else if (a) {
     selector = a;
   } else {
-    throw new Error('Missing arguments');
+    throw /* minify-error-disabled */ new Error('Missing arguments');
```
brijeshb42, line 128: Doesn't really save that much.

packages/utils/src/store/createSelector.ts
```diff
@@ -206,7 +206,7 @@ export const createSelectorMemoized: CreateSelectorFunction = (...selectors: any
       case 2: return fn(state, a1, a2);
       case 3: return fn(state, a1, a2, a3);
       default:
-        throw new Error('unreachable');
+        throw /* minify-error-disabled */ new Error('unreachable');
```
brijeshb42, line 209: Same.

packages/utils/src/formatErrorMessage.ts
```diff
@@ -0,0 +1,14 @@
...
+ * WARNING: Don't import this directly. It's imported by the code generated by
+ * `@mui/interal-babel-plugin-minify-errors`. Make sure to always use string literals in `Error`
+ * constructors to ensure the plugin works as expected. Supported patterns include:
+ *   throw new Error('My message');
+ *   throw new Error(`My message: ${foo}`);
+ *   throw new Error(`My message: ${foo}` + 'another string');
+ *   ...
+ */
+export default function formatErrorMessage(code: number, ...args: string[]): string {
+  const url = new URL(`https://base-ui.com/production-error/${code}`);
```
brijeshb42, line 11: Implemented here https://github.com/mui/base-ui/pull/1463

packages/utils/src/useEventCallback.ts
```diff
@@ -45,7 +45,5 @@ function createStableCallback() {
 }
 
 function assertNotCalled() {
-  if (process.env.NODE_ENV !== 'production') {
-    throw new Error('Base UI: Cannot call an event handler while rendering.');
-  }
+  throw new Error('Base UI: Cannot call an event handler while rendering.');
```
brijeshb42: NODE_ENV check not needed now with extraction enabled.

packages/utils/src/formatErrorMessage.ts
brijeshb42: It can be this or `https://base-ui.com/production-error?code=${code}`

docs/src/app/(public)/(content)/production-error/ErrorDisplay.tsx
```diff
@@ -0,0 +1,36 @@
+'use client';
+import * as React from 'react';
+import { useSearchParams } from 'next/navigation';
+import codes from 'docs/src/error-codes.json';
```
brijeshb42, line 4: We can have basic inline markdown formatting rendering in a followup.

packages/utils/src/useEventCallback.ts
```diff
@@ -46,6 +46,8 @@ function createStableCallback() {
 
 function assertNotCalled() {
   if (process.env.NODE_ENV !== 'production') {
```
Janpot, line 48: 🤔 That's a case we may be able to detect in the babel plugin

brijeshb42:
@mui/base-ui Can I get a review/approval ?

docs/README.md
```diff
@@ -16,3 +16,20 @@ Package managers other than pnpm (like npm or Yarn) are not supported and will n
...
+```bash
+pnpm extract-error-codes
+```
+
+This updates the `./src/error-codes.json` file with the newly extracted errors.
+
+Important: If you just altered the text of an error, you are allowed to update the existing error code with the new text in `./src/error-codes.json`, but only under the following conditions:
+
+1. There hasn't been an update to the semantic meaning of the error message. Error codes need to outlive Base UI versions, so the same code must mean the same thing across versions.
+2. There hasn't been a change in parameters, no added and no removed.
```
LukasTy, line 33: Won't this be "hardish" to maintain without automation as time goes by? 🤔

LukasTy (approved):
LGTM overall.
Great work. 👍

PR #2675, opened 2025-09-07
[composite] Fix CompositeList not updating item order on reordering

chuganzy:
This PR fixes a bug in CompositeList where the internal item order was not updated when the list items were reordered ~in React versions earlier than 19~ (this was also the case in React 19 production build) and as a result, arrow key navigation could behave incorrectly.

After doing some tests, it turned out React 19 remounts items accordingly and this is not an issue.

Sample project to reproduce that issue: https://codesandbox.io/p/sandbox/qfsgqw

This PR fixes the issue by making use of MutationObserver ~if the React version is <19~.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Set up pnpm
         uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
       - name: Use Node.js 22.x
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
         with:
           node-version: '22.18'
           cache: 'pnpm' # https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies
--- a/.github/workflows/ensure-triage-label.yml
+++ b/.github/workflows/ensure-triage-label.yml
@@ -16,7 +16,7 @@ jobs:
     permissions:
       issues: write
     steps:
-      - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
+      - uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
         with:
           script: |
             const { data: labels } = await github.rest.issues.listLabelsOnIssue({
--- a/.github/workflows/fixed-issue.yml
+++ b/.github/workflows/fixed-issue.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - name: Get associated PRs
         id: get-prs
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
         with:
           script: |
             const issue_number = context.payload.issue.number;
@@ -109,7 +109,7 @@ jobs:
 
       - name: Add comment to issue
         if: steps.get-prs.outputs.has_pr == 'true'
-        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
         with:
           script: |
             // Get PR number from the previous step's output
--- a/docs/package.json
+++ b/docs/package.json
@@ -21,7 +21,6 @@
     "@emotion/styled": "^11.14.1",
     "@mdx-js/loader": "^3.1.1",
     "@mdx-js/react": "^3.1.1",
-    "@mui/system": "7.3.2",
     "@next/mdx": "^15.5.2",
     "@react-spring/web": "^10.0.1",
     "@stefanprobst/rehype-extract-toc": "^3.0.0",
--- a/docs/reference/generated/autocomplete-root.json
+++ b/docs/reference/generated/autocomplete-root.json
@@ -46,7 +46,7 @@
     "autoHighlight": {
       "type": "boolean",
       "default": "false",
-      "description": "Whether to automatically highlight the first item when the popup opens.",
+      "description": "Whether to automatically highlight the first item while filtering.",
       "detailedType": "boolean | undefined"
     },
     "cols": {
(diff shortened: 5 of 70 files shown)


packages/react/src/composite/root/CompositeRoot.test.tsx
```diff
@@ -100,6 +100,29 @@ describe('Composite', () => {
...
+      const { getByTestId, user, rerender } = render(<App items={['1', '2', '3']} />);
+      rerender(<App items={['1', '3', '2']} />);
+
+      const item1 = getByTestId('1');
+      const item3 = getByTestId('3');
+
+      act(() => item1.focus());
+      await user.keyboard('{ArrowDown}');
+      expect(item3).toHaveFocus();
+    });
```
chuganzy, line 124: This fails on the main branch when running the test after doing `node scripts/useReactVersion.mjs ^18`

packages/react/src/composite/list/CompositeList.tsx
```diff
@@ -27,18 +28,18 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
   // information for list navigation.
 
   const map = useRefWithInit(createMap<Metadata>).current;
-  const [mapTick, setMapTick] = React.useState(0);
+  const [mapTick, setMapTick] = React.useState({});
```
chuganzy: Update to use just empty object instead because there is a (very low) risk of number overflow.

packages/react/src/composite/list/CompositeList.tsx
```diff
@@ -66,10 +88,11 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
       if (labelsRef && labelsRef.current.length !== sortedMap.size) {
         labelsRef.current.length = sortedMap.size;
       }
+      nextIndexRef.current = sortedMap.size;
```
chuganzy, line 99: Found nextIndexRef only keeps incrementing.

packages/react/src/composite/list/CompositeList.tsx
mj12albert: There's a static empty obj that could be used: https://github.com/mui/base-ui/blob/master/packages/react/src/utils/constants.ts#L4

mj12albert:
Tested the fix and it works: https://codesandbox.io/p/devbox/2675-forked-68z63n

packages/react/src/composite/list/CompositeList.tsx
chuganzy: We cannot use a static value here because we have to *change* the state to trigger the render.

atomiks:
> After doing some tests, it turned out React 19 remounts items accordingly and this is not an issue.

Curious if this only happens in React 18 dev mode? See: https://github.com/mui/base-ui/pull/2565 

We also want to make sure performance doesn't degrade for composite list components if possible:

- https://deploy-preview-2675--base-ui.netlify.app/experiments/select-perf
- https://deploy-preview-2675--base-ui.netlify.app/experiments/combobox-perf

chuganzy:
@atomiks 

> Curious if this only happens in React 18 dev mode?

Actually, this was also the case in React 19..
React 19 dev mode (StrictMode) is the only one works fine, but production mode failed..

Added https://github.com/mui/base-ui/pull/2675/commits/d04751af16389dcd36b62b7a68324b0da7218524 to remove the version check..

Regarding the performance, I will look into it deeper later!

chuganzy:
Quick summary of performance check:

- In the _"Without manually specified index"_ cases of Select and Combobox, when elements change (on open / on filter), there is additional O(n) operation to start the MutationObserver. In dev, this takes less than 300μs even with 1,000 items, which is negligible.
- With [this change](https://github.com/mui/base-ui/pull/2675/commits/e6806daef0fa8f43a4d5bc8f61b31fe116a809a6), even though the MutationObserver fires on open / close, it does not require the element order to be re-calculated in the above cases.

Given the above, in my opinion these results are acceptable, but what do you think?

packages/react/src/composite/list/CompositeList.tsx
atomiks: @chuganzy I found this line causes a bug in Autocomplete + autoHighlight where the scroll on the page jumps on the second+ open.

Did this actually cause a bug or can it be removed?

chuganzy: @atomiks My bad.. Yes, this line can be isolated and reverted! 🙏

atomiks: Nevermind, it just uncovered a bug actually. https://github.com/mui/base-ui/pull/2815

packages/react/src/composite/list/CompositeList.tsx
romgrk: This should be reverted, there is no realistic risk of number overflow. `Number.MAX_SAFE_INTEGER` is the upper limit for integers. Even with 1,000 updates per second, it would take 285 years before it causes problems.

chuganzy: Thanks for the feedback!  
You're right that the risk of number overflow is practically negligible. I only meant to highlight that "negligible" isn't quite the same as "impossible", but my main motivation here was consistency with other parts of the codebase that use the object pattern, such as the one used in `useForceRerendering`.

That said, I don't feel strongly about it, and I'm happy to revert this change if the project prefers to keep the counter-based approach here when I have a chance.
Would you prefer I update this line back, or align the other occurrences instead? Thanks!

romgrk: I would switch this line back to a number. And tbh number overflow is not possible with JS numbers, the issue would be hitting the double precision limit which would render the `+ 1` operation void. Number overflow would actually not be an issue, it would be a good thing. We could wrap on the `int32` range by casting via `>>> 0` to get number overflow.

chuganzy: Sorry, I was not very clear above. I'm aware of that behavior and understand that +1 can become ineffective (void) in JS. My point was that, this could cause `Object.is` in React to return true, preventing re-renders and memo updates.

That said, as you mentioned, this risk is extremely low, and I'm cool with reverting this change. When we do that, I'd like to leave a comment to prevent the same change from happening again since this change has passed the review, and I'll also need to make sure to mention it in the PR. To clarify, when explaining why the counter approach is used, is it sufficient to mention it for memory efficiency (GC)? Thanks again!

romgrk: We can use `setMapTick(t => (t + 1) >>> 0)` to wrap around if we're really worried about hitting the precision limit, and yes we can mention memory efficiency.

PR #2676, opened 2025-09-07
[FocusGuard] Set tabindex to -1 for non-Safari browsers

seongminn:

https://dequeuniversity.com/rules/axe/4.10/aria-hidden-focus?application=axeAPI

Elements with `aria-hidden="true"` should not be focusable. Therefore, tabIndex is set to -1 when role is undefined.

### DIFF
--- a/packages/react/src/utils/FocusGuard.tsx
+++ b/packages/react/src/utils/FocusGuard.tsx
@@ -24,10 +24,10 @@ export const FocusGuard = React.forwardRef(function FocusGuard(
 
   const restProps = {
     ref,
-    tabIndex: 0,
     // Role is only for VoiceOver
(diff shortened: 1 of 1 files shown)


atomiks:
> Elements with aria-hidden="true" should not be focusable. Therefore, tabIndex is set to -1 when role is undefined.

aXe doesn't know what focus guards are for. See: https://github.com/floating-ui/floating-ui/discussions/3364#discussioncomment-13835221

seongminn:
@atomiks 

Thank you for your response! So FocusGuard doesn't actually receive the focus, it only moves the focus. I fully understand that this isn't the issue. I'll close this PR. Thank you!

PR #2680, opened 2025-09-08
[combobox] Set default input value based on `value` prop

### LINKED ISSUE #2672: [combobox] Automatically set default `inputValue` based on `defaultValue`/`value`
## Current behavior

I made my combobox controlled using a state, and the state has the default item. When I reload the page, the value/label is not rendered/shown initially in the input, and I want to click on the input and close the popup to show the label in the input.

This is my usage:

```tsx
import { Combobox } from '@/components/combobox';

...

const [value, setValue] = useState({ label: 'Banana', value: 'banana' });

...

<Combobox
     value={value}
     onValueChange={setValue}
     items={[
       { label: 'Apple', value: 'apple' },
       { label: 'Banana', value: 'banana' },
       { label: 'Cherry', value: 'cherry' },
     ]}
/>
```

(Open the reproduction link to see what is inside `@/components/combobox`)

And, This is what is happening:

https://github.com/user-attachments/assets/ad99bbc6-487f-4dd3-9b29-ad3f7e9c1ccb

## Expected behavior

The controlled combobox should show the initial value correctly in the input field on first render.

## Reproducible example

Here is a reproduction: https://stackblitz.com/edit/base-ui-combobox-initial-value-not-working?file=app%2Fpage.tsx

## Base UI version

`^1.0.0-beta.3`

Chrome

## Which OS are you using?

Mac OS

atomiks:
Closes #2672
Avoids the need to manually align the input value with a default value for single selection mode

### DIFF
--- a/packages/react/src/combobox/root/ComboboxRoot.test.tsx
+++ b/packages/react/src/combobox/root/ComboboxRoot.test.tsx
@@ -1067,6 +1067,73 @@ describe('<Combobox.Root />', () => {
     });
   });
 
+  describe('initial input value derivation', () => {
+    it('derives input from defaultValue on first mount when unspecified', async () => {
+      await render(
+        <Combobox.Root defaultValue="apple">
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      expect(screen.getByRole('combobox')).to.have.value('apple');
+    });
+
+    it('derives input from defaultValue on first mount with items prop', async () => {
+      const items = [{ value: 'apple', label: 'Apple' }];
+      await render(
+        <Combobox.Root items={items} defaultValue={items[0]}>
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      expect(screen.getByRole('combobox')).to.have.value('Apple');
+    });
+
+    it('derives input from controlled value on first mount when unspecified', async () => {
+      await render(
+        <Combobox.Root value="banana">
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      expect(screen.getByRole('combobox')).to.have.value('banana');
+    });
+
+    it('defaultInputValue overrides derivation when provided', async () => {
+      await render(
+        <Combobox.Root defaultValue="apple" defaultInputValue="x">
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      expect(screen.getByRole('combobox')).to.have.value('x');
+    });
+
+    it('inputValue overrides derivation when provided', async () => {
+      await render(
+        <Combobox.Root value="apple" inputValue="x">
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      expect(screen.getByRole('combobox')).to.have.value('x');
+    });
+
+    it('multiple mode initial input remains empty', async () => {
+      const items = [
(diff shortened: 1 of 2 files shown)


packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -1067,6 +1067,73 @@ describe('<Combobox.Root />', () => {
...
+      await render(
+        <Combobox.Root value="banana">
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      expect(screen.getByRole('combobox')).to.have.value('banana');
+    });
+
+    it('defaultInputValue overrides derivation when provided', async () => {
```
mnajdova, line 1102: What's the usecase of supporting this? Or just covering edge case if both are specified?

packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -1067,6 +1067,73 @@ describe('<Combobox.Root />', () => {
...
+      await render(
+        <Combobox.Root value="apple" inputValue="x">
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      expect(screen.getByRole('combobox')).to.have.value('x');
+    });
+
+    it('multiple mode initial input remains empty', async () => {
```
mnajdova, line 1122: Did you mean here to test if `value` or `defaultValue` is specified?

packages/react/src/combobox/root/ComboboxRoot.test.tsx
atomiks: The `defaultInputValue` should be respected since it's acting as controlled/external logic. Use case I'm not 100% sure but there'll probably be some reason.

packages/react/src/combobox/root/ComboboxRoot.test.tsx
atomiks: `multiple` inputs should remain empty regardless (since it uses chips instead); setting a default input value applies only to single select mode

mdpx:
Hey thanks for the work on convenience, but it seems to break [this "no input" example](https://codesandbox.io/p/sandbox/clever-monad-z6n8n7) based on the [Input inside popup](https://base-ui.com/react/components/combobox#input-inside-popup).
it works on 1.0.0-beta.3.

atomiks:
@mdpx isn't this just a Select? The Combobox pattern (and all our tests) assume there's an input somewhere. I don't think Combobox needs to be used if an input is missing; for instance, there's no keyboard text typeahead while open like Select, along with some other UX affordances

mdpx:
@atomiks
1. in our team we have a full-featured combobox (with async data, loading/empty state...) and oftentimes it's used as a Select with remote options, by simply adding `showSearch={false}`.
2. [Select-Only Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) is a legitimate pattern, according to w3c.

So you're right I probably shouldn't say it "breaks" anything as there's no related example or test for such use case, but let's wait for the team's decision on whether this should be supported or not.

For now adding `defaultInputValue=""` or `inputValue=""` explicitly would help, but this seems familar...

PR #2682, opened 2025-09-08
[combobox] Fix duplicate `onOpenChange` calls and pass correct DOM `event`

atomiks:
This state set call is redundant as it already happens inside `Combobox.Item`.


packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -37,6 +37,85 @@ describe('<Combobox.Root />', () => {
 
   describe('selection behavior', () => {
     describe('single', () => {
+      it('fires onOpenChange once with reason item-press on mouse click', async () => {
+        const items = ['apple', 'banana'];
+        const onOpenChange = spy();
+
+        const { user } = await render(
+          <Combobox.Root items={items} onOpenChange={onOpenChange}>
```
oliviertassinari, line 45: I wonder if we shouldn't create a test abstraction for this. All the render setups of the combobox are close to identical.

packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -37,6 +37,85 @@ describe('<Combobox.Root />', () => {
...
+                  </Combobox.List>
+                </Combobox.Popup>
+              </Combobox.Positioner>
+            </Combobox.Portal>
+          </Combobox.Root>,
+        );
+
+        const input = screen.getByRole('combobox');
+        await user.click(input);
+        await waitFor(() => expect(screen.getByRole('listbox')).not.to.equal(null));
```
oliviertassinari: We could have this:

```suggestion
        expect(screen.getByRole('listbox')).not.to.equal(null);
```

mj12albert:
https://codesandbox.io/p/sandbox/sweet-sound-5dhpys

I found that when clicking the empty input, `onOpenChange` runs twice with `reason: 'trigger-press'`

atomiks:
@mj12albert this is likely because it uses the `.click()` method  on the item on Enter selection (it uses the `Combobox.Item`'s `onClick` handler instead of handling it again inside `Input`/`List`) - I guess for correctness it should pass the input's event instead

PR #2683, opened 2025-09-08
[combobox][autocomplete] Refactor `cols` to `grid` prop

atomiks:
The  composite logic was updated midway through development to read column sizes from DOM rows (to handle uneven / filtered rows). This prop to manually specify column size then became redundant and can just be a bool, as `Combobox.Row` is required when building a grid list.


packages/react/src/combobox/root/ComboboxRootInternal.tsx
```diff
@@ -824,8 +824,8 @@ export function ComboboxRootInternal<Value = any, Mode extends SelectionMode = '
     loop: true,
     allowEscape: !autoHighlight,
     focusItemOnOpen: queryChangedAfterOpen || selectionMode === 'none' ? false : 'auto',
-    cols,
-    orientation: cols > 1 ? 'horizontal' : undefined,
+    cols: grid ? 2 : 1,
```
flaviendelangle, line 864: Why can we set the amount of cols the `2` whenever it's a grid?
Can"t the grid have more than 2 columns?
Would be worth a small code comment explaining why, if it's the expected value 👍

atomiks: `cols >= 2` turns the behavior into a grid internally

A bunch of Floating UI tested components rely on a number (without row components), so I kept the change to the minimum here

mj12albert:
Toolbar also exposes the `cols` prop directly but I'm starting to think it's a mistake, and that it should be a `role="grid"` widget if it has grid-like keyboard nav 🤔

atomiks:
@mj12albert yeah possibly; makes sense

I think it's ok to leave it different for the time being, but that is something we need to look at

PR #2685, opened 2025-09-08
[field] Refactor `useFieldControlValidation` access

atomiks:
- Colocates field logic
- Reduces bundle size if Field is unused


mj12albert:
@atomiks Actually this PR already starts to make fieldControlValidation optional

To make it an optional part, looks like it's just moving `const fieldControlValidation = useFieldControlValidationFrom()` from FieldRoot to a new part?

atomiks:
@mj12albert yes it's probably something like that
This PR was broken to begin with though, and I have way too many open PRs already, so I'll close for someone else to pick up

PR #2688, opened 2025-09-08
[slider] Fix pointer tracking bugs

mj12albert:
Added e2e tests for this because it seems more straightforward than doing so with testing library and faking all the `getBoundingClientRect`s etc; also the e2e test infra is very under-used now.

Fixes:

1. When RTL the thumb jumps when clicking on it:

https://github.com/user-attachments/assets/b66eeb4f-7f87-45d9-bddf-bacab5694e44

2. The wrong thumb can be dragged when overlapping:

https://github.com/user-attachments/assets/d06cbc90-2c71-466e-a219-fb4d81ab9729


atomiks (approved):
Nice - works well
Side note: on the Direction Provider page, it's missing the focus style updates

test/e2e/index.test.ts
```diff
@@ -129,4 +129,39 @@ describe('e2e', () => {
...
+
+      // mouse down at the center of the lower thumb but the upper thumb
+      // is moved due to overlap
+      await page.mouse.move(25, 10);
+      await page.mouse.down();
+      await page.mouse.move(100, 10);
+      await page.mouse.up();
+
+      // eslint-disable-next-line testing-library/no-await-sync-queries
+      const output = await screen.getByRole('status');
```
LukasTy: I find this strange and confusing. 🤯 
cc @oliviertassinari
Was there a specific reason to opt for using non-standard Playwright queries? 🤔 

I would propose dropping the custom solution and using Playwright API for e2e tests, WDYT?


```diff
diff --git a/eslint.config.mjs b/eslint.config.mjs
index 7e48fbce..dce75236 100644
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -152,4 +152,10 @@ export default defineConfig(
       'testing-library/render-result-naming-convention': 'off', // inconsequential in regression tests
     },
   },
+  {
+    files: [`test/e2e/*.${EXTENSION_TS}`],
+    rules: {
+      'testing-library/prefer-screen-queries': 'off', // Enable usage of playwright queries
+    },
+  },
 );
diff --git a/test/e2e/index.test.ts b/test/e2e/index.test.ts
index 77c55128..cfaa9fcc 100644
--- a/test/e2e/index.test.ts
+++ b/test/e2e/index.test.ts
@@ -1,14 +1,5 @@
-import { expect } from 'chai';
-import { chromium, ElementHandle, Page, Browser } from '@playwright/test';
+import { chromium, Page, Browser, expect } from '@playwright/test';
 import { describe, it, beforeAll } from 'vitest';
-import type {
-  ByRoleMatcher,
-  ByRoleOptions,
-  Matcher,
-  MatcherOptions,
-  SelectorMatcherOptions,
-} from '@testing-library/dom';
-import '@mui/internal-test-utils/initMatchers';
 import '@mui/internal-test-utils/initPlaywrightMatchers';
 
 const BASE_URL = 'http://localhost:5173';
@@ -21,19 +12,6 @@ function sleep(duration: number): Promise<void> {
   });
 }
 
-interface PlaywrightScreen {
-  getByLabelText: (
-    labelText: Matcher,
-    options?: SelectorMatcherOptions,
-  ) => Promise<ElementHandle<HTMLElement>>;
-  getByRole: (role: ByRoleMatcher, options?: ByRoleOptions) => Promise<ElementHandle<HTMLElement>>;
-  getByTestId: (testId: string, options?: MatcherOptions) => Promise<ElementHandle<HTMLElement>>;
-  getByText: (
-    text: Matcher,
-    options?: SelectorMatcherOptions,
-  ) => Promise<ElementHandle<HTMLElement>>;
-}
-
 /**
  * Attempts page.goto with retries
  *
@@ -63,32 +41,6 @@ async function attemptGoto(page: Page, url: string): Promise<boolean> {
 describe('e2e', () => {
   let browser: Browser;
   let page: Page;
-  const screen: PlaywrightScreen = {
-    getByLabelText: (...inputArgs) => {
-      return page.evaluateHandle(
-        (args) => window.DomTestingLibrary.getByLabelText(document.body, ...args),
-        inputArgs,
-      );
-    },
-    getByRole: (...inputArgs) => {
-      return page.evaluateHandle(
-        (args) => window.DomTestingLibrary.getByRole(document.body, ...args),
-        inputArgs,
-      );
-    },
-    getByText: (...inputArgs) => {
-      return page.evaluateHandle(
-        (args) => window.DomTestingLibrary.getByText(document.body, ...args),
-        inputArgs,
-      );
-    },
-    getByTestId: (...inputArgs) => {
-      return page.evaluateHandle(
-        (args) => window.DomTestingLibrary.getByTestId(document.body, ...args),
-        inputArgs,
-      );
-    },
-  };
 
   async function renderFixture(fixturePath: string) {
     await page.goto(`${BASE_URL}/e2e-fixtures/${fixturePath}#no-dev`);
@@ -117,16 +69,16 @@ describe('e2e', () => {
       await renderFixture('Radio');
 
       await page.keyboard.press('Tab');
-      await expect(screen.getByTestId('one')).toHaveFocus();
+      await expect(page.getByTestId('one')).toBeFocused();
 
       await page.keyboard.press('ArrowRight');
-      await expect(screen.getByTestId('two')).toHaveFocus();
+      await expect(page.getByTestId('two')).toBeFocused();
 
       await page.keyboard.press('ArrowLeft');
-      await expect(screen.getByTestId('one')).toHaveFocus();
+      await expect(page.getByTestId('one')).toBeFocused();
 
       await page.keyboard.press('ArrowLeft');
-      await expect(screen.getByTestId('three')).toHaveFocus();
+      await expect(page.getByTestId('three')).toBeFocused();
     });
   });
 });
diff --git a/test/e2e/main.tsx b/test/e2e/main.tsx
index be7a22d3..e06851a7 100644
--- a/test/e2e/main.tsx
+++ b/test/e2e/main.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import * as ReactDOMClient from 'react-dom/client';
 import { BrowserRouter as Router, Routes, Route, Link } from 'react-router';
-import * as DomTestingLibrary from '@testing-library/dom';
 import TestViewer from './TestViewer';
 import 'docs/src/styles.css';
 
@@ -118,17 +117,3 @@ if (container != null) {
   const reactRoot = ReactDOMClient.createRoot(container);
   reactRoot.render(children);
 }
-
-window.DomTestingLibrary = DomTestingLibrary;
-window.elementToString = function elementToString(element) {
-  if (
-    element != null &&
-    (element.nodeType === element.ELEMENT_NODE || element.nodeType === element.DOCUMENT_NODE)
-  ) {
-    return window.DomTestingLibrary.prettyDOM(element as Element, undefined, {
-      highlight: true,
-      maxDepth: 1,
-    });
-  }
-  return String(element);
-};
```

packages/react/src/slider/control/SliderControl.tsx
```diff
@@ -224,9 +224,9 @@ export const SliderControl = React.forwardRef(function SliderControl(
   });
 
   const handleTouchMove = useEventCallback((nativeEvent: TouchEvent | PointerEvent) => {
-    const fingerPosition = getFingerPosition(nativeEvent, touchIdRef);
+    const fingerCoords = getFingerCoords(nativeEvent, touchIdRef);
```
LukasTy, line 227: Nitpick: prefer full variable names
```suggestion
    const fingerCoordinates = getFingerCoordinates(nativeEvent, touchIdRef);
```

packages/react/src/slider/control/SliderControl.tsx
```diff
@@ -1,68 +1,49 @@
...
+  const end = !vertical ? 'InlineEnd' : 'Bottom';
 
   return {
     start: parseFloat(styles[`border${start}Width`]) + parseFloat(styles[`padding${start}`]),
     end: parseFloat(styles[`border${end}Width`]) + parseFloat(styles[`padding${end}`]),
   };
 }
 
-function getFingerPosition(
+function getFingerCoords(
```
LukasTy, line 41:
```suggestion
function getFingerCoordinates(
```

packages/react/src/slider/control/SliderControl.tsx
mj12albert: I was trying to match the type named `Coords` from Floating UI 😅 

https://github.com/mui/base-ui/blob/master/packages/react/src/floating-ui-react/types.ts#L37

LukasTy: Ohh, I didn't see that. 🙈 
@atomiks Was there a specific reason for using the shorthand for the type? 🤔

test/e2e/index.test.ts
```diff
@@ -117,16 +69,49 @@ describe('e2e', () => {
...
+      // mouse down at the center of the lower thumb but the upper thumb
+      // is moved due to overlap
+      await page.mouse.move(25, 10);
+      await page.mouse.down();
+      await page.mouse.move(100, 10);
+      await page.mouse.up();
+
+      const output = page.getByRole('status');
+      const outputText = await output.evaluate((el) => el.textContent);
+      expect(outputText).toBe('25 – 100');
```
LukasTy: Nitpick: WDYT about writing it this way? 🤔 
The idea: avoid dropping to JS level if possible.
```suggestion
      expect(page.getByRole('status', { name: '25 – 100' })).toBeVisible();
```

packages/react/src/slider/control/SliderControl.tsx
atomiks: It's shorter and communicates the same thing 🤷

test/e2e/index.test.ts
mj12albert: Good point ~ updated in https://github.com/mui/base-ui/pull/2688/commits/65fda05eb05bea7cf4f320716aca827ca65d23d0

Found the original reasoning in https://github.com/mui/material-ui/pull/25860

>  so that people can potentially just copy&past mocha tests to playwright with minimal usage of playwright API

and 

> using playwright API for querying and asserting. With this API the smallest matchers/queries are most of the time the less expressive ones. Expressive assertions need to provide information when they fail. Oftentimes developers only think about the happy path when writing tests which is not the primary concern of tests

Considering we were practically not running any e2e tests at all, I think we should keep it as simple as possible and just use plain Playwright for now

LukasTy: Gotcha. Thanks for digging up the details. 🙏 

I agree, the way we use Playwright is already highly custom.
IMHO, we should stick to the tool API if it works well. 🤔 
On `mui-x` there are numerous e2e tests and I didn't see issues using Playwright API. 🤷

test/e2e/index.test.ts
mj12albert: It seemed `evaluate` was the only way to avoid timeouts but `await expect(page.getByRole('status')).toHaveText('25 – 100')` works too (`toBeVisible` also times out here I dunno why)

LukasTy: Cool, that's even better/cleaner. 👍 
For some reason `toHaveText` was throwing errors to me locally. 🙈

PR #2698, opened 2025-09-09
[all components] Refine event details

atomiks:
Feedback from JUI team.

- Components that have their own `onValueChange` export `ChangeEventDetails` on their namespace. It was confusing that you could reuse the root's type.
- Accordion and Collapsible previously didn't emit an event; added `eventDetails` for them as well. Also refactored part of Tabs internally to be consistent.
- Reason string unions are narrowed down to the minimum necessary, without extraneous strings that don't apply to a given component.
- Adds specific reasons for certain emitted events that were `'none'` before but can be useful to know.

### DIFF
--- a/docs/reference/generated/menu-submenu-root.json
+++ b/docs/reference/generated/menu-submenu-root.json
@@ -14,9 +14,9 @@
       "detailedType": "boolean | undefined"
     },
     "onOpenChange": {
-      "type": "((open: boolean, eventDetails: Menu.Root.ChangeEventDetails) => void)",
+      "type": "((open: boolean, eventDetails: Menu.SubmenuRoot.ChangeEventDetails) => void)",
       "description": "Event handler called when the menu is opened or closed.",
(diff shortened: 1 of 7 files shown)


docs/reference/generated/accordion-item.json
```diff
@@ -6,9 +6,9 @@
       "type": "AccordionItemValue"
     },
     "onOpenChange": {
-      "type": "((open: boolean) => void)",
+      "type": "((open: boolean, eventDetails: Collapsible.Root.ChangeEventDetails) => void)",
```
michaldudak: It's a bit weird that Accordion uses Collapsible's ChangeEventDetails. Let's reexport the type as Accordion.Root.ChangeEventDetails.

atomiks: Yeah, that's wrong🤦 it's the same issue that was present for `Menu.CheckboxItem`/`RadioGroup` and dot point 1 in the PR description

michaldudak (approved):
LGTM!

PR #2700, opened 2025-09-09
[autocomplete][combobox] Add `alwaysSubmitOnEnter` prop and allow form submission on `Enter` if no item is highlighted by default

### LINKED ISSUE #2686: [autocomplete] Allow submitting arbitrary value in Autocomplete with a single Enter key press
# Feature request

## Summary

Currently, when using the Autocomplete component and entering a custom value that is not in the suggestion list, the user must press Enter twice to submit the form: once to close the popup and a second time to actually submit. This occurs whether or not the Empty component is rendered. It would improve UX if arbitrary values could be submitted with a single Enter key press, even when the suggestions popup is open and no suggestion is selected.

## Motivation

This change would streamline the user experience for forms using Autocomplete, especially in scenarios where arbitrary values are common. It removes unnecessary interaction steps and makes the component feel more intuitive for end users.

## Sandbox

https://codesandbox.io/p/sandbox/mtkp2r

atomiks:
Fixes #2686

- **Bug fix**: allows form submission if no item is highlighted.
- **Feature**: only for Autocomplete, allows form submission even if an item is highlighted via `alwaysSubmitOnEnter` prop (useful for single-field form autocompletes like a search widget). With this, you only need to use one central `onSubmit` handler and read the input value there


docs/reference/generated/autocomplete-root.json
```diff
@@ -108,6 +108,12 @@
       "description": "Whether the popup opens when clicking the input.",
       "detailedType": "boolean | undefined"
     },
+    "submitOnEnter": {
+      "type": "boolean",
+      "default": "false",
+      "description": "Whether pressing Enter in the input should submit the surrounding form even if an item is highlighted.",
+      "detailedType": "boolean | undefined"
+    },
```
atomiks: Other possible names:

- `forceSubmitOnEnter`
- `alwaysSubmitOnEnter`
- `submitOnEnter="always|never|no-highlight"` (could be useful for Combobox specifically where pressing Enter in the input should always disallow submission?)

packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
```diff
@@ -552,6 +552,124 @@ describe('<Autocomplete.Root />', () => {
...
+          <button type="submit">Submit</button>
+        </Form>,
+      );
+
+      const input = screen.getByRole<HTMLInputElement>('combobox');
+      await user.type(input, 'al');
+      // With autoHighlight, first match is highlighted. Enter should submit, not select-only.
+      await user.keyboard('{Enter}');
+
+      expect(submitValue).to.equal('al');
```
mnajdova: Shouldn't the `submitValue` be "alpha", as we use `autoHighligt`?

docs/reference/generated/autocomplete-root.json
mnajdova: I would use `forceSubmitOnEnter` or `alwaysSubmitOnEnter` (I think the second one explains better the behavior), I don't see a real use-case for `submitOnEnter="never"`, so there are only two options really.

packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
atomiks: Yes, 👍 the `flushSync` needs to just surround `handleEnterSelection` - updated

Issue #2312, opened 2025-07-17
[dialog] OnClick event from wrapper is fired when clicking inside Dialog


marcojacobsNL:
# Bug report

## Current behavior

In our current app we have a timeline message that is wrapped in a div element that contains an onClick to open our message dialog. Within this timeline message, we also have buttons that can open other dialogs such as 1 to see the people that liked the message. With this setup though, were encountering an issue that when you click on one of the buttons to open let's say the Likes dialog after clicking within that dialog, the onClick from the wrapping element is fired and it opens the message Dialog.
It seems to also happen when we wrap the timeline message in an `<a>` tag, and the link is opened after clicking inside of the Likes dialog. Using the `ESC` seems to work fine, it's just clicks

## Expected behavior

The `onClick` from the wrapping element should not be fired anymore.

## Reproducible example

https://codesandbox.io/p/sandbox/twilight-worker-fyr52x

## Base UI version

Latest version

## Which browser are you using?

FireFox

## Which OS are you using?

Mac OS

## Additional context

Since Base-UI uses Floating-UI, I also opened a issue on their Github
https://github.com/floating-ui/floating-ui/issues/3363

atomiks:
I think this is a duplicate of https://github.com/mui/base-ui/issues/2195?

`<Dialog.Popup onClick={(e) => e.stopPropagation()}>` should work as a workaround?

I'm just curious if there are other events that should prevent bubbling/propagation inside dialogs, popovers, and others. `keydown` for instance intentionally does this so pressing `Escape` on nested popups doesn't close them all simultaneously.

ericmatthys:
> I'm just curious if there are other events that should prevent bubbling/propagation inside dialogs, popovers, and others. keydown for instance intentionally does this so pressing Escape on nested popups doesn't close them all simultaneously.

I'll be migrating components from Radix to Base UI, and it seems like I'll to have to keep something like this in place for the `Dialog`:

```tsx
  const onInteractionEvent = useCallback(
    (event) => {
      // React portals allow events to bubble up through the component tree,
      // so a click event can bubble up to the button that opened the portal.
      // This can lead to unexpected behavior, so stop propagation of
      // interaction events up from the portal.
      event.stopPropagation();
    },
    [],
  );

...
        onClick={onInteractionEvent}
        onMouseDown={onInteractionEvent}
        onMouseUp={onInteractionEvent}
        onPointerDown={onInteractionEvent}
        onPointerUp={onInteractionEvent}
[...]

PR #2702, opened 2025-09-09
[composite] Fix CompositeRoot keyboard navigation when it is nested

chuganzy:
This PR aims to fix an issue keyboard navigation breaks when CompositeRoot [stopEventPropagation: false(default)] is nested.

Code to reproduce: https://codesandbox.io/p/sandbox/pensive-dream-vy57fh

The code above puts Toolbar in NavigationMenu, but basically when CompositeRoot is nested in that way, keyboard navigation breaks. This PR fixes it by only handling the keyboard event of its own items. I took this approach because to move the focus continuously, the next element also has to be in the scope but once it leaves the focus, it cannot go back with arrow keys, and I do not think it should be considered to be part of focus roving

This usage might be niche and something debatable, so please feel free to let me know!

**Alternative solutions**
- Check if the event is already _defaultPrevented_. 
  - pros: Can support keyboard navigation to be triggered from elements not in the list
  - cons: There might be a case users call `preventDefault` on their own items
- Set `stopEventPropagation: true` when its nested.
  - pros: Same as above
  - cons: There might be a case users want the event in the ancestor nodes. This is just my two cents but I prefer UI libraries to avoid calling `stopPropagation` as far as possible to enable developers to listen to the events consistently in wider scope for many other purpose such as analytics.


packages/react/src/composite/root/useCompositeRoot.ts
```diff
@@ -140,6 +143,10 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
         event.target.setSelectionRange(0, event.target.value.length ?? 0);
       },
       onKeyDown(event) {
+        if (!mapRef.current.has(event.target as Element)) {
```
chuganzy, line 150: Using map here for the better performacne.

packages/react/src/composite/root/useCompositeRoot.ts
```diff
@@ -350,6 +362,10 @@ export function useCompositeRoot(params: UseCompositeRootParameters) {
   );
 }
 
+function createMap() {
+  return new Map<Element, CompositeMetadata<any>>();
```
chuganzy, line 366: 🤔 wondering if we should use loose type (`EventTarget`) here to avoid `as` usage in each handlers

atomiks:
Toolbar is missing the `stopEventPropagation` prop on its `CompositeRoot` which should be the cause of this issue, although maybe this was intentional and your technique here is better overall 
 @mj12albert 

> There might be a case users want the event in the ancestor nodes. This is just my two cents but I prefer UI libraries avoid calling stopPropagation as far as possible to enable developers to listen to the events consistently in wider scope for many other purpose such as analytics

We've generally been using stop event propagation everywhere as arrow keys pretty much should never be propagated. This experiment, particularly the popover with sliders, was used to test this functionality: https://deploy-preview-2702--base-ui.netlify.app/experiments/toolbar/triggers

chuganzy:
I agree, especially the combination of Portal + event bubbling is always tricky..
That said, this PR hopefully can replace most cases of `stopPropagation()` as well as the `stopEventPropagation` option in CompositeRoot.

In fact, [in the case you shared](https://deploy-preview-2702--base-ui.netlify.app/experiments/toolbar/triggers), since the Slider and its subcomponents are not managed by the Composite by default, the Composite still works correctly even without `stopEventPropagation()` inside SliderThumb [[ref](https://github.com/mui/base-ui/blob/b9b4c63503bcb3bda92d602af2c130f5c86ea0f0/packages/react/src/slider/thumb/SliderThumb.tsx#L249-L251)], and this means SliderThumb no longer needs to mind the Composite at all.

mj12albert:
> Toolbar is missing the `stopEventPropagation` prop on its `CompositeRoot` which should be the cause of this issue, although maybe this was intentional

I think it's just missing 🤔 testing it now I found that both with/without `stopEventPropagation` the slider works fine, but when the NumberFieldInput is focused, press left or right arrow a few times and Toolbar will pick up the arrow key (thus closing the popup)

> this PR hopefully can replace most cases of `stopPropagation()` as well as the `stopEventPropagation` option in CompositeRoot

I agree this is overall better 👍

atomiks:
But without `stopEventPropagation()`, arrow keydowns will fire event listeners on parent components unintentionally while using the composite. That's usually why stopping the propagation to fix this bug is also good in that it fixes unwanted firing of unrelated parent handlers too (i.e. ones not related to composite but possibly global listeners for an app)

Somewhat related: https://github.com/mui/base-ui/issues/2312

atomiks:
We're going to go with #2870, and there's also another reason why it's potentially better: interoperability. If you're using another library's composite in a parent, it will (unless they do what this is doing) also move focus. `event.stopPropagation()` is more universal/compatible

But this can use more investigation if needed

chuganzy:
Apologies for being inactive on this, and sounds good to me! Thanks for sharing!

PR #2704, opened 2025-09-09
[select][combobox] Support object values with `isItemEqualToValue` prop

### LINKED ISSUE #2586: [select] Support arbitrary object value types

### LINKED ISSUE #2694: [select] <Select.Value> render prop could also receive the full item object
# Feature request

## Summary

Right now, the render prop of `<Select.Value>` only receives the value.
This works fine if you just want to render the formatted label, but in more complex UIs it's often necessary to show additional data (for example, a color indicator, a flag, or an icon).

At the moment, the only way to achieve this is to manually look up the selected item in the items array again inside the render function, this works, but it feels a bit redundant since the select component already knows which item is selected.

It could be useful if `<Select.Value>` also passed the entire item object to the render prop.

React-Select: exposes a `formatOptionLabel` prop that receives the full option object, so you can render both the dropdown options and the selected value with additional data like icons or colors => [https://codesandbox.io/p/sandbox/reactselect-formatoptionlabel-bde1q?file=%2Findex.js%3A22%2C14](https://codesandbox.io/p/sandbox/reactselect-formatoptionlabel-bde1q?file=%2Findex.js%3A22%2C14)

## Motivation
Here's a small reproduction of our use case in the Scheduler. We’re displaying resources, and each resource has an associated color shown with a small dot => [https://codesandbox.io/p/sandbox/t8k85t](https://codesandbox.io/p/sandbox/t8k85t)

```jsx
import { ChevronDown } from "lucide-react";
import { Select } from "@base-ui-components/react/select";
import "./styles.css";

const resourcesOptions = [
  { value: null, label: "No resource", color: "lightgray" },
  { value: "resource-1", label: "Resource 1", color: "tomato" },
  { value: "resource-2", label: "Resource 2", color: "skyblue" },
  { value: "resource-3", label: "Resource 3", color: "limegreen" },
];

  return (
    <Select.Root
      defaultValue={resourcesOptions[0].value}
      items={resourcesOptions}
    >
      <Select.Trigger className="SelectTrigger">
        <Select.Value>
          {(value) => {
            const selected = resourcesOptions.find(
              (opt) => opt.value === value
            );
            return (
              <div className="SelectItemTitleWrapper">
                <span
                  style={{
                    display: "inline-block",
                    width: 10,
                    height: 10,
                    borderRadius: "50%",
                    backgroundColor: selected?.color ?? "lightgray",
                  }}
                />
                <span>{selected?.label ?? "No resource"}</span>
              </div>
            );
          }}
        </Select.Value>
        <Select.Icon className="SelectIcon">
          <ChevronDown size={14} />
        </Select.Icon>
      </Select.Trigger>
      <Select.Portal>
        <Select.Positioner align="start">
          <Select.Popup className="SelectPopup">
            {resourcesOptions.map((resource) => (
              <Select.Item
                key={resource.value ?? "none"}
                value={resource.value}
              >
                <div className="SelectItemTitleWrapper">
                  <span
                    style={{
                      display: "inline-block",
                      width: 10,
                      height: 10,
                      borderRadius: "50%",
                      backgroundColor: resource.color,
                    }}
                  />
                  <Select.ItemText>{resource.label}</Select.ItemText>
                </div>
              </Select.Item>
            ))}
          </Select.Popup>
        </Select.Positioner>
      </Select.Portal>
    </Select.Root>
  );
}
```

### LINKED ISSUE #2754: [combobox] Allow to check selected items with custom logic
# Feature request

Add an `itemToKey` prop to `Combobox.Root` to improve `isSelected` check when using objects as `Item` value

## Summary

When using objects as combobox item value the `isSelected` compare the full object of the item with the combobox value.

```
  isSelected: createSelector((state: State, index: number, value: any) => {
    if (state.multiple) {
      return Array.isArray(state.value) && state.value.includes(value);
    }
    // `selectedIndex` is only updated after the items mount for the first time,
    // the value check avoids a re-render for the initially selected item.
    return state.selectedIndex === index || state.value === value;
  }),
```

if the object is not the same (for example it can be "cloned" when updating the value state), the `isSelected` check doesn't work.

I think this can be resolved adding a prop to get a key from the value `itemToKey` like for the one for the label `itemToStringLabel` and if provided compare the keys instead of the whole object.

atomiks:
Fixes #2694
Fixes #2586
Closes #2754

Makes Combobox and Select consistent, and fixes handling of object values for autofill.

Adds a new `isItemEqualToValue` prop for equality comparison, similar to [`isOptionEqualToValue`](https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-isOptionEqualToValue) in Material UI

Select object values demo: https://deploy-preview-2704--base-ui.netlify.app/react/components/select#object-values
Many Combobox demos already use object-as-values


packages/react/src/autocomplete/root/AutocompleteRoot.tsx
```diff
@@ -180,7 +181,8 @@ export namespace AutocompleteRoot {
      */
     onValueChange?: (value: string, eventDetails: ChangeEventDetails) => void;
     /**
-     * When items' values are objects, converts its value to a string representation for display in the input.
+     * When the item values are objects (`<Autocomplete.Item value={object}>`), this function converts the object value to a string representation for display in the input.
+     * If the shape of the object is `{ value, label }`, the label will be used automatically without needing to specify this prop.
      */
     itemToStringValue?: (itemValue: ItemValue) => string;
```
mj12albert, line 189: It's a bit confusing that `itemToStringValue` is for the input value here in Autocomplete, but in Combobox `itemToStringValue` is for form submission

Perhaps `itemToInputValue`?

atomiks: In Autocomplete, `value` refers to the input value, while in Combobox, it's the selected value (and it has `inputValue` as a separate prop); so from that perspective it's consistent

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -66,16 +80,24 @@ export namespace ComboboxRoot {
...
-     * When items' values are objects, converts its value to a string value for form submission.
+     * When the item values are objects (`<Combobox.Item value={object}>`), this function converts the object value to a string representation for form submission.
+     * If the shape of the object is `{ value, label }`, the value will be used automatically without needing to specify this prop.
      */
     itemToStringValue?: (itemValue: ItemValue) => string;
+    /**
+     * Custom comparison logic used to determine if a combobox item value matches the current selected value. Useful when item values are objects without matching referentially.
+     * Defaults to `Object.is` comparison.
+     */
+    isItemEqualToValue?: (itemValue: ItemValue, selectedValue: ItemValue) => boolean;
```
mj12albert, line 99:
```suggestion
    isItemEqualToValue?: (itemValue: ItemValue | null, selectedValue: ItemValue | null) => boolean;
```
When using this with an empty initial value or Combobox.Clear, these could be null, could it be reflected here?

atomiks: I need to think about this a bit more, because I think Select also breaks in this case if you set the default value to a string but also render a `null` item in the popup list (which acts as a kind of `Select.Clear`, like `Combobox.Clear`)

I noticed adding `null` can break type inference -  removing `| null` from `value` was necessary here: https://github.com/mui/base-ui/pull/2372, where the user instead specifies `| null` in their generic `useState` call to get TS to know it can accept `null`s

atomiks: The solution should be to instead avoid calling the  provided function if either `itemValue`/`value` are null, and default to `Object.is(...)` internally.

PR #2707, opened 2025-09-09
[combobox][autocomplete] Fix controlled input value updates

### LINKED ISSUE #2703: [combobox][autocomplete] Input cursor jumps to end with async suggestions
# Bug report

## Current behavior

When using the Autocomplete component with asynchronous suggestions (similar to the example in the documentation), editing the input anywhere other than at the end causes the cursor to jump to the end after each keystroke. This prevents editing the text in the middle of the input.

## Expected behavior

The cursor should remain at the expected position (where the user is typing), and should not be automatically moved to the end after each character is entered.

## Reproducible example

This problem can be reproduced using the example provided in the official documentation.

https://base-ui.com/react/components/autocomplete#async-search

## Base UI version

1.0.0-beta.3

Chrome

## Which OS are you using?

Mac OS

atomiks:
Fixes #2703

Previously:

1. `store.state.setInputValue(...)`
2. Return value from useControlled
3. `React.useEffect` + `store.apply({ inputValue })`
4. `inputValue` updates too late - it must be sync for the caret position to be preserved.

```js
function Bug() {
  const [value, setValue] = React.useState('');
  return (
    <input
      value={value}
      onChange={(event) => {
        const value = event.target.value;
        // This is similar to what we're doing; the onChange sets the value from useControlled,
        // which returns a value that is updated asynchronously (with useEffect) to apply to the store,
        // which then informs useStore subscribers. But this happens too late and doesn't work for onChange.
        queueMicrotask(() => {
          setValue(value);
        });
      }}
    />
  );
}
```

Now:

1. `store.state.setInputValue(...)`
2. Return `inputValue` from `useControlled`
3. No synchronizing to the store with an effect; updates are sync through a derived context value

cc: @romgrk maybe you know of a better way to update the store based on derived values from other hooks, or perhaps a new context provider is required just for this special case.

### DIFF
--- a/packages/react/src/combobox/input/ComboboxInput.tsx
+++ b/packages/react/src/combobox/input/ComboboxInput.tsx
@@ -4,7 +4,7 @@ import { useStore } from '@base-ui-components/utils/store';
 import { useEventCallback } from '@base-ui-components/utils/useEventCallback';
 import { BaseUIComponentProps } from '../../utils/types';
 import { useRenderElement } from '../../utils/useRenderElement';
-import { useComboboxRootContext } from '../root/ComboboxRootContext';
+import { useComboboxInputValueContext, useComboboxRootContext } from '../root/ComboboxRootContext';
 import { selectors } from '../store';
 import { pressableTriggerOpenStateMapping } from '../../utils/popupStateMapping';
 import { useFieldRootContext } from '../../field/root/FieldRootContext';
@@ -54,10 +54,10 @@ export const ComboboxInput = React.forwardRef(function ComboboxInput(
   const triggerProps = useStore(store, selectors.triggerProps);
   const open = useStore(store, selectors.open);
   const selectedValue = useStore(store, selectors.selectedValue);
-  const inputValue = useStore(store, selectors.inputValue);
 
-  const [composingValue, setComposingValue] = React.useState<string | null>(null);
-  const isComposingRef = React.useRef(false);
+  // `inputValue` can't be placed in the store.
+  // https://github.com/mui/base-ui/issues/2703
+  const inputValue = useComboboxInputValueContext();
 
   const disabled = fieldDisabled || comboboxDisabled || disabledProp;
 
@@ -145,7 +145,7 @@ export const ComboboxInput = React.forwardRef(function ComboboxInput(
       triggerProps,
       {
         type: 'text',
-        value: componentProps.value ?? composingValue ?? inputValue,
+        value: inputValue,
         'aria-readonly': readOnly || undefined,
         'aria-labelledby': labelId,
         disabled,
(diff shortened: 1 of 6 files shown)


packages/react/src/combobox/input/ComboboxInput.tsx
```diff
@@ -163,62 +163,7 @@ export const ComboboxInput = React.forwardRef(function ComboboxInput(
...
-        },
-        onCompositionEnd(event) {
-          isComposingRef.current = false;
-          const next = event.currentTarget.value;
-          setComposingValue(null);
-          store.state.setInputValue(
-            next,
-            createBaseUIEventDetails('input-change', event.nativeEvent),
-          );
-        },
```
atomiks: cc: @mj12albert to test with composed characters, none of this should be necessary now

atomiks:
@romgrk I think the problem here is that it's the _user's_ external `React.useState()` that is passing the input `value` down to us. The store reacts to updates from their change in `useState` - `useControlled` simply returns the `controlled: inputValue` state

The store can't update at the same time in controlled mode; it has to be derived?

```js
const [value, setValue] = React.useState('');
<Autocomplete.Root value={value} onValueChange={setValue}>
```

I suppose a workaround would be to set the state in the store manually in `setInputValue` and just "bypass" the `useControlled` hook for controlled mode only but that's arguably even messier - plus it would break if they decided to change the value in `onValueChange` under a certain condition in more exotic use cases

romgrk:
Right, that's annoying. Have you tried updating the store in the render phase in reaction to the controlled prop change? React might be fine with it, considering it's an external store. Otherwise I guess the additional context could work. Anyway I'm quite unhappy about having to use `useEffect` for updates (the store itself is built to be completely synchronous, so the async nature of `useEffect` is not what we want), it would be nice to find a way to get rid of it.

atomiks:
Nope, it errors because it tries to update another component as the root is rendering:

> Store.ts:45 Cannot update a component (`ForwardRef(ComboboxInput)`) while rendering a different component

I agree the effect is annoying to synchronize values but I wonder how we can restructure the code to avoid it - it seems difficult

packages/react/src/combobox/input/ComboboxInput.tsx
mj12albert: This still works correctly~

Also noticed in the async search autocomplete demo undo/redo using the keyboard (after typing some stuff) is messed up but works correctly in this PR now

mj12albert:
I found an issue when pressing Enter to pick the first IME suggestion doesn't work:

https://github.com/user-attachments/assets/1d3fa7c7-43e5-4031-bc2e-61bc0392f905

Same in all 3 browsers and also happens in master, using a number key to pick the first option works so it's not that bad

mj12albert:
`onInputValueChange` is firing incorrectly during composition now:

https://github.com/user-attachments/assets/ae704695-a8e9-4dbd-ae4a-6fe157ef8b4c

Here's the same input on master:

https://github.com/user-attachments/assets/4e1fb5d6-021a-494e-b019-041e27978579

atomiks:
@mj12albert ah right, makes sense - the composition handling also fixed the issue Material UI has without the handling added. I've restored it back.

PR #2714, opened 2025-09-10
[select] Add `open` state for `Select.Icon` and fix `ref` type

seongminn:

1. Add the data-popup-open attribute to the Icon component, similar to the NavigationMenu component. (6c96cf8845e79f3332b064513f36874c13fc8efc)
2. Fix the issue where the Icon component's ref is incorrectly typed as HTMLDivElement, contrary to the documentation. (259418049fd69f8d1135d3fdb7f0ba766624156d)
3. ~Add the data-placeholder attribute to Select.Value when a placeholder item with no value is selected. This allows for applying custom styles. (d3eb6387059d25ce2c63081446c298a48a0be465)~ split this out into a https://github.com/mui/base-ui/pull/2737


packages/react/src/select/value/SelectValue.tsx
```diff
@@ -52,6 +52,7 @@ export const SelectValue = React.forwardRef(function SelectValue(
   const state: SelectValue.State = React.useMemo(
     () => ({
       value,
+      placeholder: value === null,
     }),
```
atomiks: One potential issue with this is with item objects as values (https://github.com/mui/base-ui/pull/2704), where the null item is an object. 

Two use cases:

- The `null` item object isn't in the popup list, so this isn't a concern (`null` is the default and will match like a real placeholder; you can't "deselect" to go back to the placeholder via the popup items list)
- The `null` item object is in the popup list, and you need to use `defaultValue={items[0]}` to ensure it's initially selected, since it appears in the list. Then `data-placeholder` won't be added.

seongminn: Thank you for the detailed feedback!

To make sure I understand correctly, are you referring to a use case like the one below, where the placeholder is an item object with an empty string as its value?

```tsx
const items = [
  { country: 'Select Country', code: '' },
  { country: 'United States', code: 'US' },
  { country: 'Canada', code: 'CA' },
  { country: 'Australia', code: 'AU' },
];

<Select.Root
  name="country"
  defaultValue={items[0]}
  itemToStringLabel={(item) => item.country}
  itemToStringValue={(item) => item.code}
>
  <Select.Trigger>
    <Select.Value />
  </Select.Trigger>
  <Select.Portal>
    <Select.Positioner>
      <Select.Popup>
        {items.map((it) => (
          <Select.Item key={it.code} value={it}>
            {it.country}
          </Select.Item>
        ))}
      </Select.Popup>
    </Select.Positioner>
  </Select.Portal>
</Select.Root>
```

If so, what are your thoughts on using the existing serializedValue from SelectRoot to handle this?

```ts
const serializedValue = React.useMemo(() => {
  if (isMultiple && Array.isArray(value) && value.length === 0) {
    return '';
  }
  return stringifyAsValue(value, itemToStringValue);
}, [isMultiple, value, itemToStringValue]);

const state: SelectValue.State = React.useMemo(
  () => ({
    value,
    placeholder: !serializedValue, // An empty string will correctly evaluate to `true`
  }),
  [value, serializedValue],
);
```

atomiks: Yeah, effectively - an empty string could be considered a placeholder as well:

```js
const items = [
  { value: null, label: 'select' },
  { value: 'apple', label: 'Apple' },
]

<Select.Root items={items} defaultValue={items[0]}>
  <Select.Item value={item}> {/* full object when mapping */}
```

seongminn: @atomiks 

Got it, thanks for confirming!

It looks like this changes depends on [#2704](https://github.com/mui/base-ui/pull/2704), so I'm going to revert it for now.

And I'll open a new PR for these changes, using #2704 as the base.

Thank you!

atomiks (approved):
Looks good, thanks for the contribution

`pnpm docs:api` needs to be run though. `SelectIconDataAttributes` file also needs to be created (with that command run) so it displays the `data-popup-open` attr in a table in the docs

seongminn:
@atomiks I added `SelectIconDataAttributes.ts` and updated docs. Thank you for your feedback!

Issue #2715, opened 2025-09-10
[select] `readOnly` Behavior in the `Select` Component


seongminn:
# Feature request

## Summary

When the `Select` component is in a `readOnly` state, it should prevent the popup from being opened by keyboard interactions, such as the `ArrowDown` key. This would align its keyboard behavior with its current mouse-click behavior.

## Motivation

This enhancement would create a more consistent and predictable experience for users.

Currently, the `readOnly` `Select` component responds differently to mouse clicks versus keyboard interactions. By aligning these behaviors, we can make the component more intuitive. This change would also create consistency with other components like `Combobox`, which handles the `readOnly` state uniformly for all interactions.

atomiks:
The `enabled` prop in `useListNavigation` should handle all keyboard navigation but there's also this extra logic here for some reason; removing this part or at least returning early when disabled/readOnly would fix, depending on how it's being tested

https://github.com/mui/base-ui/blob/fb460dd48b566cb2cbbfb2825fa49a4f851d00b9/packages/react/src/select/trigger/SelectTrigger.tsx#L165-L167

PR #2717, opened 2025-09-10
[select] block opening the popup when provided `readOnly`

seongminn:

Fixes https://github.com/mui/base-ui/issues/2715

### DIFF
--- a/packages/react/src/select/trigger/SelectTrigger.tsx
+++ b/packages/react/src/select/trigger/SelectTrigger.tsx
@@ -160,6 +160,10 @@ export const SelectTrigger = React.forwardRef(function SelectTrigger(
         store.set('touchModality', pointerType === 'touch');
       },
       onKeyDown(event) {
+        if (disabled || readOnly) {
+          return;
+        }
+
         keyboardActiveRef.current = true;
 
         if (event.key === 'ArrowDown') {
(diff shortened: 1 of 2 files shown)


packages/react/src/select/trigger/SelectTrigger.tsx
```diff
@@ -160,6 +160,10 @@ export const SelectTrigger = React.forwardRef(function SelectTrigger(
         store.set('touchModality', pointerType === 'touch');
       },
       onKeyDown(event) {
+        if (disabled || readOnly) {
+          return;
+        }
+
         keyboardActiveRef.current = true;
 
         if (event.key === 'ArrowDown') {
```
atomiks: Should this code block actually exist - is there a test covering it? `useListNavigation` should handle the logic already

seongminn: That makes sense. It's clear now that the logic in `onKeyDown` is redundant since it's covered by `useListNavigation`.

I have one follow-up question, though. Am I correct in thinking that the `keyboardActiveRef` is still necessary to track keyboard activity?

atomiks: Yeah; it should just be the `if (...)` block that's redundant.

seongminn: Did it [here](https://github.com/mui/base-ui/pull/2717/commits/016db4e138a384ba9a91b315b4136f7234755631). Thank you for your feedback!

PR #2719, opened 2025-09-10
[docs] Add Customization and TypeScript handbook pages

atomiks:
- [The Customization page](https://deploy-preview-2719--base-ui.netlify.app/react/handbook/customization) details how to customize Base UI event handling, events, and components using external state.
- [The TypeScript page](https://deploy-preview-2719--base-ui.netlify.app/react/handbook/typescript) details how to type wrapping components or functions.


docs/src/app/(public)/(content)/react/handbook/typescript/page.mdx
```diff
@@ -0,0 +1,72 @@
+# TypeScript
+
+<Subtitle>A guide to using TypeScript with Base UI.</Subtitle>
+<Meta name="description" content="A guide to using TypeScript with Base UI." />
+
+## Namespaces
+
+Base UI uses namespaces to organize types. Every component has two core interfaces:
```
michaldudak, line 8: This will be obsolete by the next release. Let's skip this part for now.

atomiks: I'm still not certain that just exporting the portable version of types in addition to namespaces isn't acceptable. 

I understand the concerns with duplicate types, but if VS Code doesn't show them as imports (and it just fixes the portable bug), I think that's fine? The DX of namespaces is really good.

But also: this page should be fairly easy to update anyway with the new pattern if necessary - so I don't thnik it needs to be removed.

michaldudak: > The DX of namespaces is really good.

It is, and I like them as well, but it seems they cause errors when the library is used as intended (to build other libraries) and work just when playing with it or when building DS in the same package as the rest of the application. I'd like to avoid situations where types work for developers when they're evaluating the library and suddenly break when they move the code to a separate package - it would be very frustrating.

mnajdova: It would be a hard no for me to keep both when we know there are issues we can't fix. It just adds maintenance burden for no clear benefit.

atomiks: My understanding was that the portable version of types just let TS work correctly, while developers (and library authors) keep using the namespace types without issues. If that's not true and their components need to use the portable types anyway, then I agree that namespace types should just be removed since they're problematic by default and should be avoided.

docs/src/app/(public)/(content)/react/handbook/customization/page.mdx
```diff
@@ -0,0 +1,124 @@
...
+  Most IDEs will show the possible string values after typing `reason === '`.
+- `event` is the native DOM event that caused the change.
+- `cancel` stops the component from changing its internal state.
+- `allowPropagation` allows the DOM event to propagate in cases where Base UI will stop the propagation.
+- `isCanceled` indicates whether the change event has been canceled.
+- `isPropagationAllowed` indicates whether the DOM event is allowed to propagate.
+
+### Canceling a Base UI event
+
+To leave a component uncontrolled, the event can be canceled with the `cancel()` method:
```
flaviendelangle: If the goal of this section is to say that you can influence the internal state of a component without having to control this state, then I don't think the sentence is super clear.


```suggestion
You can cancel updates without having to control the updates state thanks to the `cancel()` method:
```

I'm not convinced by my proposal either, but I think there is room for improvement here

atomiks: I tweaked the wording a bit

docs/src/app/(public)/(content)/react/handbook/typescript/page.mdx
atomiks: Either way though, this page can be updated in the future very easily when we change the API. It at least reflects the current state of the library so it doesn't really matter what we'll do at this point.

docs/src/app/(public)/(content)/react/handbook/customization/page.mdx
```diff
@@ -0,0 +1,124 @@
...
+onPressedChange: (pressed, eventDetails) => void
+```
+
+The `eventDetails` property is passed as a second argument to Base UI event handlers.
+This enables the event to be customized, and also allows you to conditionally run side effects based on the reason or DOM event that caused the change.
+
+```tsx title="eventDetails object"
+interface BaseUIEventDetails {
+  reason: string;
+  event: Event;
```
oliviertassinari, line 23: Two things that feel a bit off:

- Under React JS terminology, in their code base, they seem to use "native event":

``` suggestion
  nativeEvent: Event;
```

- React has a bunch of logic to normalize event behavior https://github.com/facebook/react/blob/e02c173fa597ce914c6876f71556c39bf20f40dc/packages/react-dom-bindings/src/events/SyntheticEvent.js#L333. Not exposing the synthetic event could deprive users of important behavior. But maybe it's fine.

colmtuite:
I suppose the sidebar links could get the "New" badge?

docs/src/app/(public)/(content)/react/handbook/customization/page.mdx
atomiks: Many times the events are called from a native event, so we used native events everywhere so it's consistent no matter how the event handlers are registered. 

If people request React events, it could be added as a `.reactEvent` property (that's potentially undefined). These change events are different from React event handlers to begin with signature-wise etc., so it's not a big deal I think

oliviertassinari: @atomiks In the end, I was wondering about having:
```suggestion
  event?: React.SyntheticEvent;
  nativeEvent: Event;
```

but yeah, nothing critical, it's more of a v2 kind of potential change, depending on how people respond.

PR #2721, opened 2025-09-10
[docs][select] Add docs about Select positioning for `alignItemWithTrigger` prop

### LINKED ISSUE #2713: [select] Document `alignItemWithTrigger` positioning behavior better
# Docs feedback

There's been a couple of issues (#2712, #2587) wondering why positioning props like `side` do not work at all times, and that's because [`alignItemWithTrigger`](https://base-ui.com/react/components/select#SelectPositioner-alignItemWithTrigger) activates its own special positioning mode that ignores those props unless it enters fallback.

API-wise, we didn't choose `side="none"` (the same as the data-* attribute) because the side props do still activate in "fallback" mode when `alignItemWithTrigger={true}` as well.

The documentation could have a section on this, high up, to inform users of how this works since Radix is the only other library that uses this positioning as a default.

atomiks:
Closes #2713

https://deploy-preview-2721--base-ui.netlify.app/react/components/select#positioning


docs/src/app/(public)/(content)/react/components/select/page.mdx
```diff
@@ -11,6 +11,7 @@
 ## Usage guidelines
 
 - **Prefer Combobox for large lists**: Select is not filterable, aside from basic keyboard typeahead functionality to find items by focusing and highlighting them. Prefer [Combobox](/react/components/combobox) instead of Select when the number of items is sufficiently large to warrant filtering.
+- **Special positioning behavior**: The select menu by default overlaps its trigger so the selected item's text is aligned with the trigger's value text. This behavior [can be disabled or customized](/react/components/select#positioning).
```
mnajdova:
```suggestion
- **Special positioning behavior**: The select popover by default overlaps its trigger so the selected item's text is aligned with the trigger's value text. This behavior [can be disabled or customized](/react/components/select#positioning).
```

Or "options list" or "popup" (as the component is called `Select.Popup`, I am afraid menu is a bit misleading as we also have a `Menu` component.

mnajdova: We should use the same name in other sections too.

docs/src/app/(public)/(content)/react/components/select/page.mdx
```diff
@@ -53,6 +54,18 @@ import { Select } from '@base-ui-components/react/select';
...
 
+## Positioning
+
+`Select.Positioner` has a special prop called `alignItemWithTrigger` which causes the positioning to act differently by default from other `Positioner` components.
+The prop makes the select menu overlap the trigger so the selected item's text is aligned with the trigger's value text.
+
+To prevent the select menu from overlapping its trigger, set the `alignItemWithTrigger` prop to `false`.
+When set to `true` (its default) there are a few important points to note about its behavior:
+
+- **Interaction type dependent**: The `alignItemWithTrigger` positioning mode is only active if `mouse` or `keyboard` was the input modality used to open the menu. It does not apply for `touch` or `pen` input.
```
mnajdova: Can we add a short explanation why?

atomiks: Hard to explain why in a way that isn't long-winded or hard to understand 🙈 . As there's a few reasons, it will just add clutter

docs/src/app/(public)/(content)/react/components/select/page.mdx
atomiks: Yeah, changed

packages/react/src/select/positioner/SelectPositioner.tsx
```diff
@@ -22,7 +22,7 @@ import { findItemIndex, itemIncludes } from '../../utils/itemEquality';
 const FIXED: React.CSSProperties = { position: 'fixed' };
 
 /**
- * Positions the select menu popup.
+ * Positions the select popup popup.
```
mnajdova:
```suggestion
 * Positions the select popup.
```

mnajdova (approved):
Just left one comment about a typo, it looks good 👌

PR #2725, opened 2025-09-11
[number field] Improve parsing logic

### LINKED ISSUE #2735: [number field] `onValueChange` not triggered on input blur (focusout) since v1.0.0-beta.1
_**Bug report**_

**Current behavior**

_After upgrading from_

`"@base-ui-components/react": "1.0.0-beta.2"`

to

`"@base-ui-components/react": "1.0.0-beta.3"`

the onValueChange callback signature has changed:
```

// before
event?: Event

// now
eventDetails: NumberField.Root.ChangeEventDetails
```

This change itself is fine, however I lost access to the focusout event.

When I increment/decrement the value via the NumberField controls, the onValueChange callback is triggered as expected.

But when I type a value directly into the input and then blur/lose focus (focusout), the onValueChange callback is not triggered anymore.

This breaks the previous behavior where changes were detected after editing the input manually and leaving the field.

**Expected behavior**

The onValueChange callback should also be triggered when the user enters a value manually and then blurs the input (focusout), not only when incrementing/decrementing.

**Reproducible example**

A minimal reproduction showing the difference between 1.0.0-beta.2 and 1.0.0-beta.3:
[CodeSandbox/StackBlitz link here if available]

**Base UI version**

@base-ui-components/react: 1.0.0-beta.3

atomiks:
Handles a few more edge cases correctly.

Fixes #2735


packages/react/src/number-field/utils/parse.ts
```diff
@@ -1,12 +1,50 @@
...
+  〇: '0',
+  一: '1',
+  二: '2',
+  三: '3',
+  四: '4',
+  五: '5',
+  六: '6',
+  七: '7',
+  八: '8',
+  九: '9',
```
mj12albert, line 16:
```suggestion
  九: '9',
  十: '10',
```
@atomiks This is supposed to simply convert single digits?

packages/react/src/number-field/utils/parse.test.ts
```diff
@@ -52,5 +52,175 @@ describe('NumberField parse', () => {
...
+    });
+
+    it('returns null for Infinity-like inputs', () => {
+      expect(parseNumber('Infinity')).to.equal(null);
+      expect(parseNumber('-Infinity')).to.equal(null);
+      expect(parseNumber('∞')).to.equal(null);
+    });
+
+    it('collapses extra dots from mixed-locale inputs', () => {
+      // First '.' is decimal; subsequent '.' are removed
```
michaldudak:
```suggestion
      // Last '.' is decimal; previous '.' are removed
```

atomiks: This could be the culprit in `onKeyDown` as it doesn't pass the `locale`, although it uses your runtime locale. Is it in US/English for you when the bug occurs?

```js
const { decimal, currency, percentSign } = getNumberLocaleDetails(
  [],
  formatOptionsRef.current,
);
```

packages/react/src/number-field/utils/parse.ts
atomiks: It should - let me know if that char is necessary though to handle another use case

atomiks: LLM review as this is out of my depth a bit:

> We intentionally only normalize single-glyph Han digits (零/〇…九). Adding 十: '10' would break inputs like 三十 (30), which would become "310" under per-char substitution. If we need full ideographic number parsing, that should be a separate parser; this function is just digit normalization across scripts.

mj12albert: OK it doesn't matter then if it's just converting each individual char into a digit

Though in practice this would only work for 0-9 (and not 10+) since there are characters for 10/100/1000, so the numbering system is similar to roman numerals

But this is indeed an incredibly exotic use-case so it's ok

packages/react/src/number-field/utils/parse.test.ts
michaldudak: Yup, but I see you fixed it. Seems to work OK, but I found another issue 😅 - when on "pl-PL' locale (where we use spaces as thousand separators), I can't type in space at all.

atomiks: @michaldudak I see, fixed

Issue #579, opened 2024-09-03
[NumberField]: onValueChange not being called when input value is changed


foobarvaz:
### Search keywords

onValueChange

### Latest version

- [X] I have tested the latest version

### Steps to reproduce

Link to live example:
https://codesandbox.io/p/sandbox/wpqjmg?file=%2Fsrc%2FApp.tsx%3A10%2C43-10%2C56

Steps:
1. Focus on input and change it's value

### Current behavior

`onValueChange` not being called when input value is changed

You can see that the text below the input will remain `value: 0` until you focus out of the input. The blur leads to `onValueChange` being called.

### Expected behavior

`onValueChange` **being** called when input value is changed

### Context

Thanks for adding support for decimals in the new version of the number input ❤️ 

I'm working on an interactive calculator in which changes to input values should be reflected in calculations that are shown in the UI.

I tried to debug the issue and I found that on each change I reach this part of the code:

https://github.com/mui/base-ui/blob/78d4abd6a72b0e3d783aa7a09bfda8c7a7518b88/packages/mui-base/src/NumberField/Root/useNumberFieldRoot.ts#L607

Because `event.isTrusted === true` in this case, we call only `setInputValue` without calling `setValue` (which internally calls `onValueChange`)

I'm not sure if it's by design or not. So forgive me if I classified this issues as bug 🐛  by mistake.

### Your environment

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

atomiks:
> I'm not sure if it's by design or not. So forgive me if I classified this issues as bug 🐛 by mistake.

Yes it's by design. The reason is partial input may not be parseable as a valid number, so it can't properly be called in realtime/on change.

bengry:
> > I'm not sure if it's by design or not. So forgive me if I classified this issues as bug 🐛 by mistake.
> 
> Yes it's by design. The reason is partial input may not be parseable as a valid number, so it can't properly be called in realtime/on change.

Can this potentially be made available on the hook variant of `NumberField`? Or an optional prop that defaults to `false` (= immediate change behavior disabled).

I understand the concern, but it gives a worse UX (less immediate feedback) because of an edge case, for a lack of a better term - when a user enters invalid values. In the happy flow path, the UX is just worse without any benefit.
[...]

Issue #2102, opened 2025-06-11
[number field] Make START_AUTO_CHANGE_DELAY Configurable for NumberField


michael-land:
# Feature request

## Summary

I'm using the NumberField component as a quantity input in a shopping cart.
To avoid sending too many HTTP requests, I use a debounce of 200ms on value change.

## Motivation

Problem:
The component’s built-in `START_AUTO_CHANGE_DELAY` is fixed at 400ms.
So when a user holds down the increment button:

At 200ms: debounce triggers an HTTP request (from the initial increment)

At 400ms: auto-repeat starts, causing a second value change → another HTTP request

This leads to two requests even though the user only held the button once.

Proposal:
Allow START_AUTO_CHANGE_DELAY to be set via props, so it can be aligned with external debounce timing.

Example API:

```tsx
<NumberField startAutoChangeDelay={200} />
```

Current Workaround:
Set the debounce timeout to >= 400ms, however this introduces a less responsive UX and feels laggy.

atomiks:
In #2726, an `onValueCommitted` callback was added, which fires on pointerup/release of the spin buttons now. It's not the same as configuring the delay, but I wonder if it would solve your use case? It will avoid firing network requests until the user settles their interaction. 

Based on this, it sounds like you still want network requests to fire while pressing before release though:

> Set the debounce timeout to >= 400ms, however this introduces a **less responsive UX and feels laggy**.

PR #2726, opened 2025-09-11
[number field] Align value changes with Slider

atomiks:
Fixes https://github.com/mui/base-ui/issues/579

`onValueChange` now fires on every change to the input provided the value is parseable to some form.

To retain the old behavior if needed (waiting until blur), `onValueCommitted` is provided, which is similar to Slider's in that it's a lazy version of `onValueChange`. 

`onValueCommitted` differs from `onValueChange` in the following ways:
 
- The user is typing; it will wait until blur (like before)
- The user is scrubbing; it will wait until pointerup (similar to Slider)
- The user is pressing & holding the spin buttons; it will wait until pointerup

In the other keyboard interaction cases (wheel scrubbing, arrow keys) they fire at the same time, same as Slider's keyboard interactions.

### DIFF
--- a/packages/react/src/number-field/input/NumberFieldInput.tsx
+++ b/packages/react/src/number-field/input/NumberFieldInput.tsx
@@ -276,7 +276,7 @@ export const NumberFieldInput = React.forwardRef(function NumberFieldInput(
       if (event.isTrusted) {
         setInputValue(targetValue);
         const parsedValue = parseNumber(targetValue, locale, formatOptionsRef.current);
-        if (parsedValue !== null && isValidCharacterString) {
+        if (parsedValue !== null) {
           setValue(parsedValue, event.nativeEvent);
         }
         return;


packages/react/src/number-field/root/NumberFieldRoot.tsx
```diff
@@ -545,6 +558,14 @@ export namespace NumberFieldRoot {
      * Callback fired when the number value changes.
      */
     onValueChange?: (value: number | null, eventDetails: ChangeEventDetails) => void;
+    /**
+     * Callback function that is fired when the value is committed. It runs later than `onValueChange`, when:
+     * - The input is blurred after typing a value.
+     * - The pointer is released after scrubbing or pressing the increment/decrement buttons.
+     *
+     * **Warning**: This is a generic event not a change event.
```
michaldudak, line 620:
```suggestion
     * **Warning**: This is a generic event, not a change event.
```

michaldudak:
Holding the arrow key while the input is focused shouldn't trigger `onValueCommitted`. Only after the key is released, the event should occur (similarly to how scrubbing commits the value only after the mouse button is released).

Also, I'm wondering if it would be beneficial to populate the reason field in the events (to differentiate scrubbing from button clicks and keyboard input) - not necessarily in this PR.

atomiks:
> Holding the arrow key while the input is focused shouldn't trigger onValueCommitted. Only after the key is released, the event should occur (similarly to how scrubbing commits the value only after the mouse button is released).

`Slider` doesn't do this currently, wonder if it's necessary, or if they should be aligned both in a separate PR at least

> Also, I'm wondering if it would be beneficial to populate the reason field in the events (to differentiate scrubbing from button clicks and keyboard input) - not necessarily in this PR.

Yes, this makes sense to me

michaldudak:
> if they should be aligned both in a separate PR at least

Yeah, I'd prefer this. I would be consistent with scrubbing and holding the inc/dec button.

michaldudak (approved):

packages/react/src/number-field/root/NumberFieldRoot.tsx
atomiks: It's using the same comment format as Slider

We also likely need a different param than `eventDetails` for non-change events, which I'll have to look into

oliviertassinari:
> To retain the old behavior if needed (waiting until blur), onValueCommitted is provided, which is similar to Slider's in that it's a lazy version of onValueChange.

A side observation ([#suggestion](https://www.notion.so/mui-org/Olivier-Tassinari-a6d0f1d19f8a4cfbb20469878ded8aa8?pvs=4#ac52aa0d7db5447b832416ef2c790f83)), if the only goal is to retain the previous behavior to avoid painful breaking changes, but this previous behavior was added by mistake, I think we could argue that the best path is not to have the prop. Maybe we can remove the bundle size/maintenance overhead of having it. It could look like this: wait for people to ask for it, and make a strong case to add it back.

atomiks:
@oliviertassinari it's not just for backwards-compat, but for consistency with Slider which has the same callbacks. 

If your UI maps number field to an element's visual transform, you may want to wait for them to finish ("commit") before changing the UI, otherwise it will jump around. There are probably other cases too, like firing network requests lazily. (In fact it might kind of solve this inadvertently: https://github.com/mui/base-ui/issues/2102)

PR #2729, opened 2025-09-11
[code-infra] Remove unused packages

brijeshb42:
* Also migrated fs-extra to use fs/promises
* Migrated code to use globby instead of fast-glob
* Use base markdownlint config from code-infra


package.json
michaldudak, line 1: ❤️

Issue #2727, opened 2025-09-11
[menubar] Menubar `disabled` prop does not disable menus


kenanyusuf:
# Bug report

## Current behavior

Passing the `disabled` prop to the `Menubar` component does not disable the menus.

## Expected behavior

I expected the menus within the menubar to be disabled when this prop is set.

## Reproducible example

https://codesandbox.io/p/sandbox/disabled-menubar-dkg635

## Base UI version

1.0.0-beta.3

PR #2736, opened 2025-09-11
[menubar] Fix Menubar not disabling child Menus

aarongarciah:
Fixes https://github.com/mui/base-ui/issues/2727

Before: https://codesandbox.io/p/sandbox/disabled-menubar-dkg635
After: https://codesandbox.io/p/devbox/menubar-disabled-menubar-forked-ywc2q8

### DIFF
--- a/packages/react/src/menubar/Menubar.test.tsx
+++ b/packages/react/src/menubar/Menubar.test.tsx
@@ -643,6 +643,11 @@ describe('<Menubar />', () => {
       // Trigger should be disabled
       expect(fileTrigger).to.have.attribute('disabled');
 
+      // It should not be reachable via Tab
+      await user.tab();
+      expect(fileTrigger).not.toHaveFocus();
+      expect(document.body).toHaveFocus();
+
       // Clicking should not open the menu
       await user.click(fileTrigger);
       expect(screen.queryByTestId('file-menu')).to.equal(null);


aarongarciah:
~Adding a test would be nice.~

edit: Added.

packages/react/src/menubar/Menubar.test.tsx
```diff
@@ -634,6 +634,21 @@ describe('<Menubar />', () => {
     });
   });
 
+  describe('prop: disabled', () => {
```
aarongarciah, line 637: I wasn't sure if this test should be in the Menubar suite or in the Menu suite. Probably in the Menu suite? Let me know your preference.

packages/react/src/menubar/Menubar.test.tsx
```diff
@@ -634,6 +634,21 @@ describe('<Menubar />', () => {
...
+      const fileTrigger = screen.getByTestId('file-trigger');
+
+      // Trigger should be disabled
+      expect(fileTrigger).to.have.attribute('disabled');
+
+      // Clicking should not open the menu
+      await user.click(fileTrigger);
+      expect(screen.queryByTestId('file-menu')).to.equal(null);
+    });
+  });
```
mj12albert, line 655: Maybe check that it can't be Tab focused as well ~

aarongarciah: Added https://github.com/mui/base-ui/pull/2736/commits/e7ffa9234c43d7c85ec55be794551c2e216ed36c

packages/react/src/menubar/Menubar.test.tsx
michaldudak: :+1: Menubar tests are OK.

PR #2737, opened 2025-09-11
[select] add data-placeholder

seongminn:

based on https://github.com/mui/base-ui/pull/2704

- Add the `data-placeholder` attribute to `Select.Value` when a placeholder item with no value is selected. This allows for applying custom styles. (1d928aac62bc26204ad303a79d33b8def9610cf7)
- I've opened the PR to address [this feedback](https://github.com/mui/base-ui/pull/2714#discussion_r2337314734).


seongminn:
@atomiks The base PR for this pull request has been merged into master. Could you please review this PR?

atomiks:
@seongminn super sorry, this PR got away from me and I forgot to review it!

Some thoughts:

- Should `Select.Trigger` also have this state?
- Can you add tests for this state attribute (direct `null`, object with `value: null`, custom property with `itemToStringValue`)

seongminn:
@atomiks 
Sorry for the late reply 😅
All requirements have been incorporated.

I always appreciate your feedback.

packages/react/src/select/trigger/SelectTrigger.tsx
```diff
@@ -87,6 +89,17 @@ export const SelectTrigger = React.forwardRef(function SelectTrigger(
...
+  const isMultiple = multiple ?? false;
+
+  const itemToStringValue = useStore(store, selectors.itemToStringValue);
+
+  const serializedValue = React.useMemo(() => {
+    if (isMultiple && Array.isArray(value) && value.length === 0) {
+      return '';
+    }
+    return stringifyAsValue(value, itemToStringValue);
+  }, [isMultiple, value, itemToStringValue]);
```
atomiks: I think this also exists in the root, the logic could be reused between the three sites

seongminn: I have included serializedValue in the SelectRootContext! [38c7763](https://github.com/mui/base-ui/pull/2737/commits/38c7763b5e6fc528831a1d5b77ce3dd9db2b5214)

packages/react/src/select/root/SelectRoot.tsx
```diff
@@ -406,6 +413,7 @@ export function SelectRoot<Value, Multiple extends boolean | undefined = false>(
       disabled,
       readOnly,
       multiple,
+      serializedValue,
```
atomiks: This should be placed in the `store` sync effect where `store.update(...)` is called (like `value` is).
Also it seems `serializedValue: createSelector((state: State) => {` should be removed.

seongminn: I have incorporated all the points you mentioned!

atomiks:
Thanks for the contribution!

PR #2742, opened 2025-09-12
[toast] Support variable height stacking

### LINKED ISSUE #1955: [toast] Support stacking toasts of differing heights
# Feature request

## Summary

Currently the `expanded` state is forced to be `true` if the heights of the toasts aren't all the same.

https://github.com/mui/base-ui/blob/5d2d1653852c42e4de6e64432b0b76fdcbe12699/packages/react/src/toast/provider/ToastProvider.tsx#L45-L51

This is because if a larger height toast is stacked beneath a smaller height one, it overflows the bottom of it. However there's a way to make it work: minimize larger toasts' height and fade out their content

https://emilkowal.ski/ui/css-transforms

atomiks:
Closes #1955

https://github.com/user-attachments/assets/4dd21782-0662-48e8-84eb-240a019b3ad1

## New part

Adds a new `<Toast.Content>` part to apply fade-out and `overflow: hidden` effects to, and to allow `ResizeObserver` to only measure content changes now that the `height` of the toast changes and transitions.

## Styles fixes

There's a new `--peek` variable (user-defined in the demos) and changed `transform` logic that ensures the "peek" part of stacked toasts behind the front one is constant and doesn't change based on the height of the toasts. I also noticed this fixed the animation bug that was present when holding down `Enter` on the toast creator button.


packages/react/src/toast/viewport/ToastViewportCssVars.ts
```diff
@@ -0,0 +1,7 @@
+export enum ToastViewportCssVars {
+  /**
+   * Indicates the height of the frontmost toast at index 0 in the list.
+   * @type {number}
+   */
+  index0Height = '--toast-index-0-height',
```
aarongarciah: Would `frontmostHeight` or `foremostHeight` be a better name? My brain instinctively thought there were also `index-1`, `index-2`, ... vars.

```suggestion
  frontmostHeight = '--toast-frontmost-height',
```

docs/reference/generated/toast-viewport.json
```diff
@@ -18,5 +18,10 @@
       "type": "boolean"
     }
   },
-  "cssVariables": {}
+  "cssVariables": {
+    "--toast-frontmost-height": {
+      "description": "Indicates the height of the frontmost toast at index 0 in the list.",
```
aarongarciah: nit: "index 0" sounds redundant with frontmost


```suggestion
      "description": "Indicates the height of the frontmost toast in the list.",
```

or

```suggestion
      "description": "Indicates the height of the frontmost toast.",
```

docs/src/app/(public)/(content)/react/components/toast/demos/position/css-modules/index.module.css
```diff
@@ -122,6 +127,19 @@
   }
 }
 
+.Content {
+  overflow: hidden;
+  transition: opacity 0.25s;
+
+  &[data-behind] {
+    opacity: 0;
+  }
```
aarongarciah, line 136: Would this have the same effect?


```diff
-  &[data-behind] {
     opacity: 0;
-  }
```

Am I right at assuming that a toast is always in the "behind" or "expanded" state, and only in one state at a time?

packages/react/src/toast/content/ToastContent.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const contentRef = React.useRef<HTMLDivElement | null>(null);
+
+  useIsoLayoutEffect(() => {
+    const node = contentRef.current;
+    if (!node) {
+      return undefined;
+    }
+
+    function updateHeight() {
```
michaldudak: No need to define this function.

packages/react/src/toast/content/ToastContent.tsx
```diff
@@ -0,0 +1,77 @@
...
+    };
+  }, [recalculateHeight]);
+
+  const state: ToastContent.State = React.useMemo(
+    () => ({
+      expanded,
+      behind: index > 0,
+    }),
+    [expanded, index],
+  );
```
michaldudak, line 51: nit: calculating `behind` outside of useMemo could prevent some unnecessary object allocations as dependencies will change less often

```suggestion
  const behind: index > 0;
  const state: ToastContent.State = React.useMemo(
    () => ({
      expanded,
      behind,
    }),
    [expanded, behind],
  );
```

PR #2732, opened 2025-09-11
[docs][accordion][toolbar] Fix ToolbarButton, and ToolbarInput to use proper state types

chuganzy:
Related: #2697

This is a minor PR to fix AccordionPanel, ToolbarButton, and ToolbarInput to use proper state types. 

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #2751, opened 2025-09-12
[code-infra] Run label check on pull_request synchronize

chuganzy:
Related: https://github.com/mui/base-ui/pull/2732

It currently does not check when new commits are pushed to a PR, and too easy to overlook.
This PR adds `synchronize` to trigger it on new commits as well.


.github/workflows/check-if-pr-has-label.yml
```diff
@@ -2,7 +2,7 @@ name: Check if PR has label
 
 on:
   pull_request:
-    types: [opened, reopened, labeled, unlabeled]
+    types: [opened, unlabeled, synchronize]
```
chuganzy: * ~`labeled` is removed, since the minimum label count is 1.~ this was needed to update failed -> success state.
* `reopened` shouldn't be needed.

LukasTy: Why do you think `synchronize` is needed here? 🤔 
Other repositories are successfully running with these exact trigger types...

I don't see how this could pass after a synchronization if a label hasn't been added... 🤷‍♂️

chuganzy: Oops, my bad, `labeled` was actually needed. Added it back in https://github.com/mui/base-ui/pull/2751/commits/b4fef771d8657ccc054d4a7a5b9afd8b68b15f74.

`synchronize` is needed because, as happened in #2732, when there are changes to the PR branch, checks are not run against the latest commit, making it appear as though everything has passed and easy to miss.

In #2732, `test-label-applied` state was added to the latest commit _after_ the labels are added, which is _after_ merging it to the main branch.

LukasTy: But shouldn't the `opened` be enough? 🤔 
Cases I see:
1. Community member opens a PR (without labels due to lack of access) -> triggers an action error
2. Team member opens a PR -> triggers an action error if no labels were added
3. Any further commits on this particular PR shouldn't change anything to make this particular GH workflow pass, unless labels are edited.

Am I missing something? 🤷 
cc @michelengelen for possible extra context. 🤔

chuganzy: Here is the PR to show what was happening before this change: https://github.com/mui/base-ui/pull/2753 - it looks as if all checks passed, even though no labels are set.

There might be new changes pushed to a PR before maintainers start reviewing. If they always manually verify the status of all commits before merging, this change wouldn’t be necessary. The point, however, is that it’s too easy to miss.

LukasTy: Gotcha.
Thanks for clarifying. 🙏 
It looks like the issue might lie elsewhere - the CI is not properly triggered by community pushes. 🙈 
Maybe @michelengelen knows more context about this? 🤔

michelengelen: Yes, the reason is pretty simple.
The CI suite runs on all commits and since the label check is necessary for a green status it needs to run as well. The problem is that that the workflow is not getting triggered when adding a new commit. That leaves the CI "hanging" ... blocking the PR.
It is a Github workflow limitation to not "cache" the result when there are not meaningful changes, so `synchronize` is necessary.

LukasTy: Gotcha, thanks for confirming. Well, it won't hurt and can help avoid some edge cases. 👌
Just wanted to confirm if it's not a redundant step, since other repos do not have it listed. 🤔

.github/workflows/check-if-pr-has-label.yml
```diff
@@ -2,7 +2,7 @@ name: Check if PR has label
 
 on:
   pull_request:
-    types: [opened, reopened, labeled, unlabeled]
```
LukasTy, line 5: Was there a clear reason for removing the `reopened` state? 🤔 
An edge case, but I think this process would cause an unexpected state:
1. Open PR
2. Add a label <-- CI is green
3. Close PR
4. Remove a label (clear all labels)
5. Reopen PR <-- CI is green, but technically incorrect

chuganzy: Yes should be ok. At the step 4, it re-triggers the check, so the status turns into red.

chuganzy: Also, JFYI, I found that the action README also uses the same trigger in its example: https://github.com/mheap/github-action-required-labels?tab=readme-ov-file#complete-example

LukasTy: Thanks for the exploration. 🙏 
I would have assumed that closed PRs wouldn't trigger actions. 🤔

PR #2183, opened 2025-06-27
[field] New `Field.Item` part

mj12albert:
**💥 Breaking:** When integrating `Field` with `CheckboxGroup` or `RadioGroup`, `Field.Item` must be wrapped around each individual radio/checkbox in the group

Preview: https://deploy-preview-2183--base-ui.netlify.app/experiments/form

Option 1 from https://github.com/mui/base-ui/issues/2172

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #2756, opened 2025-09-13
[useButton] Update disabled prop of the button when ref changes

chuganzy:
As put in the title, this PR updates `useButton` hook to update `disabled` property also when element itself changes. Without this, it becomes disabled when ref changes and `focusableWhenDisabled` stops working properly.

Ref can change especially when using `render` prop and custom components: https://codesandbox.io/p/sandbox/2756-l3j4xk

To solve it, this PR updates `buttonRef` to a callback and update the option manually. We can use `useState` as the ref, but it forces re-render which I think is not necessary at the moment.

### DIFF
--- a/packages/react/src/use-button/useButton.ts
+++ b/packages/react/src/use-button/useButton.ts
@@ -17,12 +17,12 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
     native: isNativeButton = true,
   } = parameters;
 
-  const buttonRef = React.useRef<HTMLButtonElement | HTMLAnchorElement | HTMLElement | null>(null);
+  const elementRef = React.useRef<HTMLElement | null>(null);
 
   const isCompositeItem = useCompositeRootContext(true) !== undefined;
 
   const isValidLink = useEventCallback(() => {
-    const element = buttonRef.current;
+    const element = elementRef.current;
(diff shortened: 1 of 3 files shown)


atomiks (approved):
Nice, thanks for the contribution!

packages/react/src/use-button/useButton.ts
```diff
@@ -173,6 +175,11 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
     [disabled, focusableWhenDisabledProps, isNativeButton, isValidLink],
   );
 
+  const buttonRef = useEventCallback((element: HTMLElement | null) => {
+    elementRef.current = element;
+    updateDisabled();
+  });
```
mj12albert, line 181: A small annoyance with this is that I can't use the returned `buttonRef.current` in an effect easily anymore because it's always this function, example: https://github.com/mui/base-ui/pull/2183/commits/2bd52b483ea98e6e4d7c6a7285f03e63e1935221

It was easier to add another ref than to make buttonRef work 😓

chuganzy: Would you please elaborate this when you have a chance?

I am not sure if I understand correctly, but if that means exporting two refs, we have to update all the places use `useButton`, and merge them everywhere, which I think we probably want to avoid 🤔

Issue #2759, opened 2025-09-14
[combobox] An example for Custom value to be rendered inside Combobox


sarthak-simpplr:
### Support needed
Input inside trigger example, something like React Select.

## Use Case
I want to use Combobox to build a component where consumers can render custom content as the selected value (e.g., a person's avatar with their name, product images with titles, etc.). This is similar to how React Select allows custom `SingleValue` components.

Currently, the default examples in the docs don't address this use case for single-select scenarios.

## Current Challenge
When trying to implement this pattern, I'm running into architectural issues with how to structure the trigger and input elements:

### Approach 1: Input inside Trigger (causes event conflicts)
```jsx
<Combobox.Trigger>
  <Combobox.Input /> {/* Input clicks bubble to trigger, causing open/close flickering */}
  <CustomSelectedValue />
</Combobox.Trigger>
```

### Approach 2: Multiple separate triggers (workaround)
```jsx
<div className="combobox-container">
  <Combobox.Input className={selectedValue ? 'hidden' : 'visible'} />
  <Combobox.Trigger className={selectedValue ? 'visible' : 'hidden'}>
    <CustomSelectedValue />
  </Combobox.Trigger>
</div>
```

### Expected behavior:
- No selection: Show input with placeholder, clicking anywhere opens popup
- With selection: Show custom selected value, clicking anywhere opens popup and focuses input
- While typing: Hide selected value, show input for filtering
- The input does not retain the value of the selected item.

### Questions for the Team

- What's the recommended pattern for implementing custom selected value displays with React select like behavior?

atomiks:
> What's the recommended pattern for implementing custom selected value displays with React select like behavior?

Does `openOnInputClick={false}` on `Combobox.Root` solve this particular structure?

```jsx
<Combobox.Trigger>
  <Combobox.Input /> {/* Input clicks bubble to trigger, causing open/close flickering */}
  <CustomSelectedValue />
</Combobox.Trigger>
```

That should let the trigger handle the click alone without the flicker because of the input handling `mousedown`.

@sarthak-simpplr also, are you able to share a CodeSandbox of what you're making? It would be useful to add a new demo on the docs for this pattern where the input is placed inside the trigger

sarthak-simpplr:
Thanks for your response @atomiks.

> Does `openOnInputClick={false}` on `Combobox.Root` solve this particular structure?

[...]

PR #2760, opened 2025-09-15
[combobox][docs] Decorated input example

atomiks:
Part of #2759

https://deploy-preview-2760--base-ui.netlify.app/react/components/combobox#decorated-input


docs/src/app/(public)/(content)/react/components/combobox/demos/decorated-input/tailwind/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+              className="relative group flex w-[16rem] min-[500px]:w-[20rem]"
+              aria-label={person ? person.name : undefined}
+            >
+              <Combobox.Input
+                id={id}
+                placeholder={person ? '' : 'Select person'}
+                className="h-10 w-full rounded-md font-normal border border-gray-200 pl-9 pr-16 text-base text-gray-900 bg-[canvas] group-hover:bg-gray-100 group-focus-within:group-hover:bg-[canvas] focus:outline focus:outline-2 focus:-outline-offset-1 focus:outline-blue-800 cursor-default focus:cursor-text"
+                onBlur={() => {
+                  setInputValue('');
+                }}
```
atomiks, line 42: This pattern requires controlling the input value, with this special handler

docs/src/app/(public)/(content)/react/components/combobox/demos/decorated-input/css-modules/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+                  <Combobox.Icon
+                    className={`${styles.Avatar} ${!person ? styles.AvatarPlaceholder : ''}`}
+                  >
+                    {person ? initials(person.name) : <PersonIcon className={styles.AvatarIcon} />}
+                  </Combobox.Icon>
+                  {person && !isTyping && <span className={styles.Text}>{person.name}</span>}
+                </div>
+
+                <div className={styles.ActionButtons}>
+                  <Combobox.Clear className={styles.Clear} aria-label="Clear selection">
```
sarthak-simpplr, line 53: This should probably have a tabIndex of 0, otherwise keyboard user won't be able to clear the value.

docs/src/app/(public)/(content)/react/components/combobox/demos/decorated-input/css-modules/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <label htmlFor={id}>Assignee</label>
+        <div className={styles.Control}>
+          <Combobox.Value>
+            {(person: Person | null) => (
+              <div className={styles.InputContainer} aria-label={person ? person.name : undefined}>
+                <Combobox.Input
+                  id={id}
+                  placeholder={person ? '' : 'Select person'}
+                  className={styles.Input}
+                  onBlur={() => {
```
sarthak-simpplr, line 39: We probably don't need this.

docs/src/app/(public)/(content)/react/components/combobox/demos/decorated-input/tailwind/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+              No people found.
+            </Combobox.Empty>
+            <Combobox.List>
+              {(person: Person) => (
+                <Combobox.Item
+                  key={person.id}
+                  value={person}
+                  className="box-border grid cursor-default select-none grid-cols-[1.25rem_1fr_0.75rem] items-center gap-2 py-2 pr-5 pl-4 text-base leading-4 outline-none data-[highlighted]:relative data-[highlighted]:z-0 data-[highlighted]:before:absolute data-[highlighted]:before:inset-x-2 data-[highlighted]:before:inset-y-0 data-[highlighted]:before:z-[-1] data-[highlighted]:before:rounded-sm data-[highlighted]:before:bg-gray-100"
+                >
+                  <div className="col-start-1 inline-flex size-5 items-center justify-center rounded-full bg-blue-800 text-[0.6rem] font-semibold text-white">
```
sarthak-simpplr: aria-hidden just like the module css example

sarthak-simpplr:
Thank you for the useful example ❤️

docs/src/app/(public)/(content)/react/components/combobox/demos/decorated-input/css-modules/index.tsx
atomiks: It has a default `tabIndex` of `-1` as it usually shouldn't be part of the tab order (namely for forms). For keyboards, <kbd>Esc</kbd> clears the value, while screen reader virtual cursors can access also the button to clear it as with pointers

docs/src/app/(public)/(content)/react/components/combobox/demos/decorated-input/css-modules/index.tsx
atomiks: You mean the `onBlur` handler? If you type a value and tab out/blur the input, the input stays as the custom text instead of the selected value text/placeholder without it

sarthak-simpplr: With the `onBlur` added I found out that clicking anywhere in the popup except on an item (for eg. `Combobox.Empty`)  causes the input to reset, which is a bit unexpected from a user's POV, but isn't too bad now that I think about.

docs/src/app/(public)/(content)/react/components/combobox/demos/decorated-input/tailwind/index.tsx
mj12albert: It would also need to set it to the real text in `onFocus`, or when tabbing into the filled component it would appear like arrowing through the text doesn't work:

https://github.com/user-attachments/assets/272321ae-74e1-4f5a-af38-8d125abae361

atomiks: This is just how this pattern works I think 😅  

Although Vercel makes it look like a placeholder when focused. That's probably the best UX.

https://github.com/user-attachments/assets/51e79e55-403d-4fc6-860b-10cf0a78f6b7

atomiks: The main problem with that pattern is that selecting an item keeps focus on the input, so it shows the "placeholder" after selection too, which looks odd

Vercel just blurs the input and focus gets lost which is not accessible...

colmtuite: > arrowing through the text doesn't work

Doesn't this negate the example? Seems like something we shouldn't present in our docs?

How would it work with `multiple`? Isn't this a use case for Chips? It strikes me as similar Gmail's "To:" UI, where you can add multiple emails and they are listed as chips with avatars.

sarthak-simpplr:
@atomiks This example breaks on `v1.0.0-beta.4`. `onInputValueChange` is called twice when you press an item, once with `item-press`, and again with reason `none`. This seems unexpected to me as on `v1.0.0-beta.3`, its called twice with the reason `item-press`.

Do you know if this is tracked in an issue or already fixed? Thank you!

atomiks:
I think the example should be updated to check for `"none"` too. It was intentionally changed here: https://github.com/mui/base-ui/pull/2830

PR #2768, opened 2025-09-16
[context menu][menubar] Ensure submenus close when parents close

atomiks:
Fixes these scenarios (and possibly more given it's general):

- Context Menu: Right click (without a pointerup release), move to a submenu, release on an empty area; root closes but submenu doesn't, so it lingers around until the root unmounts without fading out correctly (only visible when there are exit transitions, but also doesn't work with `keepMounted` prop). In JUI it happens while releasing over an item as well 
- Menubar: <kbd>ArrowRight</kbd> inside a submenu to land on a new menuitem trigger in the menubar; same issue as above

### DIFF
--- a/packages/react/src/context-menu/root/ContextMenuRoot.test.tsx
+++ b/packages/react/src/context-menu/root/ContextMenuRoot.test.tsx
@@ -1,0 +1,80 @@
+import * as React from 'react';
+import { expect } from 'chai';
+import { fireEvent, screen, waitFor } from '@mui/internal-test-utils';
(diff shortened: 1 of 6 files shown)


mnajdova:
I can't reproduce this on master. Can you share a reproduction and maybe a recording of what's happening?

atomiks:
@mnajdova Menubar needs exit transitions on the menus: https://github.com/mui/joy-ui/pull/319#issuecomment-3282554478

Context Menu one is reproducible on JUI Context Menu docs

PR #2770, opened 2025-09-16
[toast] Allow collapsed state on mouseleave

atomiks:
Changes the `data-expanded` state for pointers to make the toast expanded state feel less sticky


atomiks:
@mnajdova  The library shadcn uses (sonner) also has this behavior

PR #2771, opened 2025-09-16
[combobox] Fix unexpected close with multiple selection and input inside popup

atomiks:
The popup should only close in this scenario if the input isn't inside the popup.


mnajdova (approved):
Work as expected 👌

Issue #2767, opened 2025-09-15
[collapsible] Height is miscalculated when Panel uses flex or grid


DiegoAndai:
# Bug report

## Current behavior

When the collapsible (or accordion) panel has flex or grid and uses properties that modify the height, the`--collapsible-panel-height` variable is incorrectly calculated.

## Expected behavior

The variable should be calculated correctly.

## Reproducible example

- Collapsible using flex and gap: https://codesandbox.io/p/sandbox/vigorous-sun-7j2t7t?file=%2Fsrc%2Findex.module.css%3A8%2C1
- Accordion using grid and subgrid for alignment: https://codesandbox.io/p/sandbox/clever-sun-9glwyh?file=%2Fsrc%2FApp.tsx

## Base UI version

v1.0.0-beta.3

## Which browser are you using?

Chrome

## Which OS are you using?

Mac OS

## Cause

The issue is that we use `display: block important!` when opening in these two places:

- https://github.com/mui/base-ui/blob/master/packages/react/src/collapsible/panel/useCollapsiblePanel.ts#L120
- https://github.com/mui/base-ui/blob/master/packages/react/src/collapsible/panel/useCollapsiblePanel.ts#L177

Which overrides the panel's display and makes the height read is incorrect.

Removing `important` fixes the issue.

Using flex or grid on the content instead of the panel is not always possible. For example, in the accordion reproduction, we need grid and subgrid to be able to share the subgrid with the content for alignment.

mj12albert:
> Removing important fixes the issue.

@DiegoAndai Maybe it could work if we read the actual display property, e.g. in your use-case where you need the panel to be grid, internally it should apply `element.style.setProperty('display', 'grid', 'important')` instead of `'block'`

Copying the sandbox where the panel is `display: flex`, setting `element.style.setProperty('display', 'flex', 'important')` internally seems to work:

https://github.com/user-attachments/assets/ec60c5cc-1b10-4299-bed6-de2350bda8b1

DiegoAndai:
> Maybe it could work if we read the actual display property

Yes! I think that works as well

PR #2772, opened 2025-09-16
[collapsible] Respect user's CSS `display` property on panel

mj12albert:
We need to force the `display` property to ensure correct measurement but shouldn't assume the value is `'block'`

Fixes https://github.com/mui/base-ui/issues/2767

### DIFF
--- a/packages/react/src/collapsible/panel/useCollapsiblePanel.ts
+++ b/packages/react/src/collapsible/panel/useCollapsiblePanel.ts
@@ -36,6 +36,7 @@ export function useCollapsiblePanel(
     width,
   } = parameters;
 
+  const computedStylesRef = React.useRef<CSSStyleDeclaration>(null);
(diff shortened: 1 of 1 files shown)


packages/react/src/collapsible/panel/useCollapsiblePanel.ts
```diff
@@ -174,7 +180,11 @@ export function useCollapsiblePanel(
 
     if (open) {
       /* opening */
-      panel.style.setProperty('display', 'block', 'important');
+      panel.style.setProperty(
+        'display',
+        computedStylesRef.current?.display ?? 'block',
```
mj12albert, line 185: Read again in case it changed since initialization e.g. from a media query

DiegoAndai (approved):
I confirm this fixes the issue on my side ✅

PR #2776, opened 2025-09-18
[scroll area] Fix RTL horizontal scrollbar on Safari

atomiks:
The `getOffset(...)` value when calculating `margin` offset on the scrollbar is reported differently in Safari compared to Chrome/Firefox. Switching to logical properties (as we also switched to use use logical functional CSS a while back) fixes the issue

https://deploy-preview-2776--base-ui.netlify.app/experiments/scroll-area/scroll-area-inset


atomiks:
@LukasTy welp, the tests were failing with just `marginInline` so I changed it to `marginInlineStart` + `marginInlineEnd` and it has the exact same issue it seems 😅

I think for RTL we have to assume the margin is symmetric and avoid `end`

PR #2778, opened 2025-09-18
[scroll area] Fix thumb size flicker

atomiks:
This is mainly visible in Safari with inset scrollbars due to the updates happening in a microtask.

### DIFF
--- a/packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
+++ b/packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
@@ -1,5 +1,6 @@
 'use client';
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 import { useEventCallback } from '@base-ui-components/utils/useEventCallback';
 import { useIsoLayoutEffect } from '@base-ui-components/utils/useIsoLayoutEffect';
 import { useTimeout } from '@base-ui-components/utils/useTimeout';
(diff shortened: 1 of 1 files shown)


LukasTy (commented):
Does it happen only on macOS 26?
Or am I looking at a wrong place?
Could you share a video of the problem? 🤔

atomiks:
On #2776's deploy preview, it can sometimes even not update until scroll. But usually it just flickers (for a frame) into place and is correct before scrolling:

https://github.com/user-attachments/assets/43f8f4a1-78d0-4a60-8b9c-98ade2ee19d1


And yes, this is macOS 26

LukasTy:
> On #2776's deploy preview, it can sometimes even not update until scroll. But usually it just flickers (for a frame) into place and is correct before scrolling:
> And yes, this is macOS 26

Gotcha. Could we first merge the mentioned PR and then re-check?
As it currently stands, there is a clear issue that is interfering with the reproduction of the issue. 🙈

oliviertassinari:
There is a downside to this approach when there are multiple ScrollArea on screen (e.g., our docs pages), we force React to update the DOM to render each, so it means that each component needs to read the DOM. We have a cascade of forced reflows:

SCR-20251213-msts

https://693ae9e6a41ca30008471aaf--base-ui.netlify.app/react/components/select, from #3531

Intuitively, I wouldn't expect flushSync to be needed in this context; it feels like this fix hides a deeper problem.

---

Idea moved to #3534.

PR #2779, opened 2025-09-18
[navigation menu] Fix focus returning to trigger without animations

### LINKED ISSUE #2775: NavigationMenu doesn’t return focus to trigger, if there are no transition styles
# Bug report

## Current behavior

If there are no `transition` style defined for `Container[data-ending-style]`, then focus does not return to the trigger after menu is closed.

## Expected behavior

Focus going back to the trigger should not rely on styles.

## Reproducible example

https://codesandbox.io/p/sandbox/muddy-browser-fnwq7t

## Base UI version

1.0.0-beta.3

Google Chrome	140.0.7339.127 (Official Build) (64-bit) 

## Which OS are you using?

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 25.04
Release:        25.04
Codename:       plucky
```

_None._

atomiks:
Fixes #2775

Focus can be on the `<body>` when this handler is called with no transitions/animations specified. Wasn't able to write a failing test in either Chromium env


packages/react/src/navigation-menu/root/NavigationMenuRoot.tsx
```diff
@@ -114,10 +116,14 @@ export const NavigationMenuRoot = React.forwardRef(function NavigationMenuRoot(
...
+    const isReturnFocusBlocked = closeReasonRef.current
+      ? blockedReturnFocusReasons.has(closeReasonRef.current)
+      : false;
+
     if (
-      closeReasonRef.current !== 'trigger-hover' &&
+      !isReturnFocusBlocked &&
       isHTMLElement(prevTriggerElementRef.current) &&
-      contains(popupElement, activeEl) &&
+      (activeEl === ownerDocument(popupElement).body || contains(popupElement, activeEl)) &&
```
mnajdova, line 126: Do we know why the focus ends up on the body? Looks like the focus now moves from the body and only then to the trigger element, no? Is this the reason why you can't write the test, or something unrelated?

packages/react/src/navigation-menu/root/NavigationMenuRoot.tsx
```diff
@@ -114,10 +116,14 @@ export const NavigationMenuRoot = React.forwardRef(function NavigationMenuRoot(
     const doc = ownerDocument(rootRef.current);
     const activeEl = activeElement(doc);
 
+    const isReturnFocusBlocked = closeReasonRef.current
```
mnajdova, line 119: Should we maybe cover this change with test (the different closing reasons that should be ignored).

packages/react/src/navigation-menu/root/NavigationMenuRoot.tsx
atomiks: When focus is inside the popup on unmount, the browser moves it to the body. It happens to still be inside the popup when there are animations probably because of a small delay with the animation handling. 

The test written for the chromium env wasn't failing before the change either

mnajdova: Got ya, thanks for the explanation.

michaldudak (approved):

PR #2797, opened 2025-09-19
[code-infra] Set up stylelint

Janpot:


stylelint.config.mjs
```diff
@@ -1,23 +1,12 @@
-import baseline from '@mui/monorepo/stylelint.config.mjs';
-
 export default {
-  ...baseline,
+  extends: '@mui/internal-code-infra/stylelint',
```
oliviertassinari: The previous approach has its pros: we could easily introspect what the current values are to debug them or extend the logic. Not sure if we gain by moving to a custom `extends` implementation.

package.json
```diff
@@ -24,7 +24,7 @@
     "jsonlint": "code-infra jsonlint",
     "eslint": "eslint . --cache --report-unused-disable-directives --ext .js,.ts,.tsx,.mts --max-warnings 0 && pnpm -r lint",
     "eslint:ci": "eslint . --report-unused-disable-directives --ext .js,.ts,.tsx,.mts --max-warnings 0 && pnpm -r lint",
-    "stylelint": "stylelint --reportInvalidScopeDisables --reportNeedlessDisables \"docs/**/*.{js,ts,tsx}\" \"**/*.css\" --ignore-path .gitignore",
+    "stylelint": "stylelint --reportInvalidScopeDisables --reportNeedlessDisables \"docs/**/*.{js,ts,tsx}\" \"**/*.css\" --ignore-path .lintignore",
```
oliviertassinari, line 27: While we are at it, how about we add [` --max-warnings `](https://stylelint.io/user-guide/options/#maxwarnings)? I imagine we don't want to fix the warnings here, but we can enforce that it can't grow.

stylelint.config.mjs
Janpot: For introspection I'd recommend `--print-config`, since the resolved config very much depends on the file that is being linted:

```tsx
pnpm exec stylelint --print-config docs/src/components/DescriptionList.css
```

The advantages of `extends` is:
a. it handles merging correctly for e.g. the `rules` object, you can't accidentally overwrite it. 
b. it handles plugin dependencies correctly.

Nothing stops you from printing it manually should you have a real need for this:

```tsx
import base from '@mui/internal-code-infra/stylelint'

console.log(base)
```

We should also be able to do

```tsx
import base from '@mui/internal-code-infra/stylelint';

export default {
  extends: base,
}

```

oliviertassinari: > We should also be able to do

This could be the best of the two worlds

package.json
Janpot: I'd prefer if we fix the warnings that are straightforward and add ignore logic for the rest. that way fixing an issue doesn't allow you to introduce a new one

PR #2801, opened 2025-09-20
[code-infra]: Migrate 'lodash' to 'es-toolkit'

sukvvon:
Migrated from lodash-es to [es-toolkit](https://es-toolkit.dev/) to remove lodash as a dependency, resolving performance and maintenance concerns.

Replaced all lodash methods with equivalent methods provided by es-toolkit, which is actively maintained and optimized for modern JS usage.

This change was motivated by:

Performance improvements (es-toolkit is faster in [benchmarks](https://es-toolkit.dev/performance.html) compared to lodash).
Ongoing maintenance and active community support.
The [es-toolkit/compat](https://es-toolkit.dev/compatibility.html) module fully covers the existing lodash test coverage, ensuring no side effects or regressions.
Lodash has not had a major release since v4 on January 13, 2016 ([lodash releases](https://github.com/lodash/lodash)).
Lodash is no longer actively maintained, evidenced by unanswered user questions regarding a new version (https://github.com/lodash/lodash/issues/5989).
The latest commit to the planned v5 of Lodash was 8 months ago ([v5-wip branch](https://github.com/lodash/lodash/tree/v5-wip)).
@types/lodash contains numerous any types, negatively impacting overall application type safety.
Lodash lacks proper tree-shaking, thus requiring the separate lodash-es package.
No additional breaking changes were introduced.


docs/package.json
```diff
@@ -29,12 +29,12 @@
     "clipboard-copy": "^4.0.1",
     "clsx": "^2.1.1",
     "cross-env": "^10.0.0",
+    "es-toolkit": "^1.39.10",
```
oliviertassinari, line 32: I don't understand why lodash is defined twice. If it's already in the root package, it seems to duplicate here.

```suggestion
```

oliviertassinari (commented):
+1 on my end for this direction:

lodash - [3.57 kB (gzip)](https://bundlejs.com/?q=lodash%2FcamelCase%2Clodash%2FupperFirst&treeshake=%5B*%5D%2C%5B*%5D&text=%22export+%7B+camelCase%2C+upperFirst+%7D%22)
es-toolik - [1.02 kB (gzip)](https://bundlejs.com/?q=es-toolkit%2Fcompat%2FcamelCase%2Ces-toolkit%2Fcompat%2FupperFirst&treeshake=%5B*%5D%2C%5B*%5D&text=%22export+%7B+camelCase%2C+upperFirst+%7D%22)

docs/package.json
michaldudak: It's because packages should explicitly define their dependencies. Ideally, the root package doesn't contain any dependencies (even pnpm warns when adding a dependency to the workspace root). I'd prefer to keep it this way. Since we use Renovate to update dependencies, there's no additional effort to keep everything in sync.

scripts/api-docs-builder/src/formatter.ts
```diff
@@ -1,9 +1,9 @@
 /* eslint-disable no-await-in-loop */
-import sortBy from 'lodash/sortBy';
+import sortBy from 'es-toolkit/compat/sortBy';
 import * as tae from 'typescript-api-extractor';
 import fs from 'fs';
 import path from 'path';
-import _ from 'lodash';
+import * as _ from 'es-toolkit/compat';
```
LukasTy: Nitpick: Only `uniq` is used in this file. Maybe we could use explicit import instead?

scripts/releaseChangelog.mts
```diff
@@ -5,7 +5,7 @@ import { Octokit } from '@octokit/rest';
 import chalk from 'chalk';
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
-import _ from 'lodash';
+import * as _ from 'es-toolkit/compat';
```
LukasTy: Same nitpick: only `startCase` is used... 🤷 🙈

docs/package.json
oliviertassinari: > It's because packages should explicitly define their dependencies.

@michaldudak Ah, but it's what I don't understand. What are the advantages of this?
The values that I could think of for having it twice are: a. it allows us to have different versions, but we use the same versions in this context? b. it would allow us to move the code around, but we practically never do it?

@sukvvon Since my comment is off-topic, we should stick to what HEAD already does (have the dependency listed in docs too), we can resolve this independently.

scripts/api-docs-builder/src/formatter.ts
sukvvon: @LukasTy I reflected your opinion 👍

scripts/releaseChangelog.mts
sukvvon: @LukasTy I reflected your opinion 👍

docs/package.json
sukvvon: @oliviertassinari @michaldudak  Understood. I added `es-toolkit` to the `docs` repo again.

michaldudak: It makes the package.json of a package a source of truth about dependencies. I don't need to look at various package.json files to understand what dependencies are used and what are their versions. It also makes it easy to extract the project and run it independently from everything else in the repo.

PR #2807, opened 2025-09-22
[backdrops] Add `--body-client-height CSS` variable

### LINKED ISSUE #2799: [backdrops] Add `document.body.clientHeight` as a CSS variable
# Bug report

## Current behavior

`position: fixed` elements don't extend past the UI chrome in Safari on iOS 26. It leaves an area that remains undimmed and uncovered by the backdrop.

`position: absolute` + `height: document.body.clientHeight` works correctly, but accessing that variable in a style tag errors during SSR and is hackish

Image

Image

---

## Expected behavior

```css
.Backdrop {
  position: absolute;
  height: var(--body-client-height);
}
```

This works correctly (scrolled or not) for the different tab layout options in Safari, collapsed chrome or not.

Image
Image
Image

atomiks:
Closes #2799

### DIFF
--- a/packages/react/src/alert-dialog/backdrop/AlertDialogBackdropCssVars.ts
+++ b/packages/react/src/alert-dialog/backdrop/AlertDialogBackdropCssVars.ts
@@ -1,0 +1,7 @@
+export enum AlertDialogBackdropCssVars {
+  /**
+   * Indicates the client height of the document body.
+   * @type {number}
+   */
+  bodyClientHeight = '--body-client-height',
+}
--- a/docs/reference/generated/alert-dialog-backdrop.json
+++ b/docs/reference/generated/alert-dialog-backdrop.json
@@ -33,5 +33,10 @@
       "description": "Present when the dialog is animating out."
     }
   },
-  "cssVariables": {}
+  "cssVariables": {
+    "--body-client-height": {
+      "description": "Indicates the client height of the document body.",
+      "type": "number"
+    }
+  }
 }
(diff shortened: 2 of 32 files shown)


packages/react/src/alert-dialog/backdrop/AlertDialogBackdropCssVars.ts
```diff
@@ -0,0 +1,7 @@
+export enum AlertDialogBackdropCssVars {
+  /**
+   * Indicates the client height of the document body.
```
michaldudak: Let's explain why this is useful. I'd intuitively just use `position: fixed` on backdrops and ignore this.

atomiks: Our docs don't support long comments here (it becomes horizontally scrollable). All the examples at least have updated to use it, so maybe people will realize why it's necessary when looking at their components on iOS 26 otherwise.

Seems unfortunate if we need to add a section on the docs about it. I think people will update over time and learn that `position: fixed` is no longer viable for mobile Safari

atomiks:
It seems that for horizontally scrollable websites (though not exactly common), we probably need `--body-client-width` as well otherwise it gets cut off after scrolling

welp, seems `document.body.clientWidth` doesn't consider horizontal scrolling unlike the `height` axis...

atomiks:
Simpler solution: https://github.com/mui/base-ui/pull/2848

PR #2808, opened 2025-09-22
[docs][dialog] Scrollable examples

atomiks:
Adds "outer" + "inner" scrollable layout examples

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/dialog/demos/inside-scroll/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/dialog/demos/inside-scroll/css-modules/index.module.css
@@ -1,0 +1,192 @@
+.Button {
+  box-sizing: border-box;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 2.5rem;
+  padding: 0 0.875rem;
+  margin: 0;
+  outline: 0;
+  border: 1px solid var(--color-gray-200);
+  border-radius: 0.375rem;
+  background-color: var(--color-gray-50);
+  font-family: inherit;
+  font-size: 1rem;
+  font-weight: 500;
+  line-height: 1.5rem;
+  color: var(--color-gray-900);
+  user-select: none;
+
+  @media (hover: hover) {
+    &:hover {
+      background-color: var(--color-gray-100);
+    }
+  }
+
+  &:active {
+    background-color: var(--color-gray-100);
+  }
+
+  &:focus-visible {
+    outline: 2px solid var(--color-blue);
+    outline-offset: -1px;
+  }
+}
+
+.Backdrop {
+  position: fixed;
+  inset: 0;
+  background-color: black;
+  opacity: 0.2;
+  transition: opacity 150ms cubic-bezier(0.45, 1.005, 0, 1.005);
+
+  @media (prefers-color-scheme: dark) {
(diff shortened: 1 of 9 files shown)


packages/react/src/dialog/root/useDialogRoot.ts
```diff
@@ -140,7 +140,9 @@ export function useDialogRoot(params: useDialogRoot.Parameters): useDialogRoot.R
         // https://github.com/mui/base-ui/issues/1320
         if (modal) {
           return internalBackdropRef.current || backdropRef.current
-            ? internalBackdropRef.current === eventTarget || backdropRef.current === eventTarget
+            ? internalBackdropRef.current === eventTarget ||
+                backdropRef.current === eventTarget ||
+                contains(eventTarget, popupElement)
```
atomiks, line 146: The outside scroll example won't allow scrolling while hovering outside the popup (aka directly on the viewport wrapper) unless we change the location of the internal backdrop, and add this new `contains` check to prevent breaking outside presses.

```jsx
<div className="Viewport">
  <Dialog.Popup />
</div>
```

atomiks: Though the `MobileNav` on our docs does something different. It makes the `Popup` be the entire viewport and creates an "inner" part instead, which is the actual white popup area. Then it uses `Dialog.Close` as the "outside press" area rather than relying on the internal checker. That's probably just because it didn't work correctly to begin with:

```jsx
{/* We need the area behind the panel to close on tap but also to scroll the viewport. */}
<Dialog.Close
  className="MobileNavBackdropTapArea"
  tabIndex={-1}
  nativeButton={false}
  render={<div />}
/>
```

Screenshot 2025-09-22 at 2 34 43 pm

atomiks:
This is probably worth merging for the bug fixes alone; the dual space with inset scrollbars isn't _too_ terrible. They could use `ScrollArea` instead as well so it looks better. @colmtuite 

I think we may need a `Dialog.Viewport` (or similar) part that informs the scroll lock whether or not it should set `scrollbar-gutter: stable` (where it will only set it if `viewport.scrollHeight` isn't `> viewport.clientHeight`), which will only create one scrollbar space, but this should be done in a different PR.

colmtuite:
@atomiks 

1. Use ScrollArea for both the outset and inset examples?
2. Remove the h4 headings and have two h3 headings instead?
3. These new scrollable examples should probably at the bottom of the examples section, rather than at the top.

atomiks:
A couple of `ScrollArea` issues with the outside-scroll example. Both have workarounds but they aren't great to display as examples:

- There are no animation attributes on the scrollbar, so it abruptly unmounts instead of fading out when the dialog starts closing. To be fair this also happens with the native scrollbars - we'd need `Dialog.Viewport` as a way to contain content that can fade in or out. This would also solve the "dual" space scrollbar issue with inset scrollbars, as we can communicate with the scroll lock in that case. Alternatively, we could have an animation component that could be applied to anything. As a workaround, you need to control the dialog and set `data-closed={!open ? "" : undefined}` as an attribute on the scrollbar.
- `ScrollArea.Root` has a `position: relative` functional style but this needs to be `position: fixed`, so `style={{ position: undefined (or 'fixed') }}` needs to be used to workaround it.

mnajdova (approved):

PR #2813, opened 2025-09-23
[all components] Include ref in `BaseUIComponentProps`

atomiks:
This makes React 17-18 lie, but shouldn't cause type errors.

This prioritises React 19 where `ref`s are included in `props` to make the types easier.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  code-infra: https://raw.githubusercontent.com/mui/mui-public/9cc15ae47e8be0c5f9fe8c09553d5aeb238b9d8a/.circleci/orbs/code-infra.yml
+  code-infra: https://raw.githubusercontent.com/mui/mui-public/2f944b985f17ebcd9280e00974035e38e496cbcf/.circleci/orbs/code-infra.yml
 
 parameters:
   browserstack-force:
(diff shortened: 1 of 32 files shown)


docs/reference/generated/tabs-tab.json
```diff
@@ -3,7 +3,7 @@
   "description": "An individual interactive tab button that toggles the corresponding panel.\nRenders a `<button>` element.",
   "props": {
     "value": {
-      "type": "Tabs.Tab.Value",
```
atomiks, line 6: Possible bug with API extractor when `ref` is included? cc @michaldudak

michaldudak: This is fixed in `typescript-api-extractor@1.0.0-alpha.10`

packages/react/src/menu/positioner/MenuPositioner.test.tsx
```diff
@@ -8,7 +8,7 @@ import { describeConformance, createRenderer, isJSDOM } from '#test-utils';
 
 const Trigger = React.forwardRef(function Trigger(
   props: Menu.Trigger.Props,
-  ref: React.ForwardedRef<HTMLDivElement>,
+  ref: React.ForwardedRef<any>,
```
michaldudak, line 11: Why is this needed? Can we use `Element` instead?

atomiks: There's an error at `ref` otherwise; `HTMLElement`/`Element` both don't work as it expects `HTMLButtonElement`. 

React 19 works fine though without `forwardRef`

michaldudak (approved):

PR #2814, opened 2025-09-23
[collapsible] Fix opening transition regression

atomiks:
The change to respect the `display` property has some quirks. If `justify-content: end` is being used (which is necessary so that the `border-radius` of the bottom edge doesn't get clipped while transitioning), the `.scrollHeight` is `0` initially. We need to manually unset the `justify-content`/`align-items` values. 

Also the computed styles object is "live" so it should be recalculated on every access. It was empty `""` after the first open.

### DIFF
--- a/packages/react/src/collapsible/panel/useCollapsiblePanel.ts
+++ b/packages/react/src/collapsible/panel/useCollapsiblePanel.ts
@@ -38,7 +38,6 @@ export function useCollapsiblePanel(
     width,
   } = parameters;
 
-  const computedStylesRef = React.useRef<CSSStyleDeclaration>(null);
   const isBeforeMatchRef = React.useRef(false);
   const latestAnimationNameRef = React.useRef<string>(null);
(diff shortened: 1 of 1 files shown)


packages/react/src/collapsible/panel/useCollapsiblePanel.ts
```diff
@@ -181,12 +175,15 @@ export function useCollapsiblePanel(
     }
 
     if (open) {
+      const originalStyles = {
+        'justify-content': panel.style.justifyContent,
+        'align-items': panel.style.alignItems,
```
mj12albert, line 180: `align-content` is needed here as well, I found that changing the hero styles to grid without clipping the border radius like this also causes the bug:
```css
display: grid;
align-content: end;
```
(Maybe `justify-items` too for horizontal?)

atomiks: Yeah makes sense, there's quite a few properties 😬

PR #2817, opened 2025-09-23
[docs][autocomplete] Command palette / inside dialog demo

atomiks:
This takes the demo from the initial `FilterableMenu` implementation and uses `Autocomplete` instead to implement it.

https://deploy-preview-2817--base-ui.netlify.app/react/components/autocomplete#command-palette  

One difference with `cmdk` is that we allow the highlight to be removed on hover out, while it does not. I'm not 100% sure why command palettes do this.

I also noticed macOS 26's Spotlight widget doesn't have hover styles at all and just uses focus styles with keyboard


colmtuite:
@atomiks Compose ScrollArea here too?

colmtuite:
@atomiks The dialog should be much taller on mobile?

IMG_9738

michaldudak:
@atomiks, could you merge in the latest master and redeploy the site? The current deploy preview is no longer there.

PR #2819, opened 2025-09-24
[select][combobox][autocomplete] Use `ReadonlyArray` type for items

atomiks:
Unlike `Array` it supports both `ReadonlyArray` + regular `Array`

### DIFF
--- a/packages/react/src/autocomplete/root/AutocompleteRoot.tsx
+++ b/packages/react/src/autocomplete/root/AutocompleteRoot.tsx
@@ -6,8 +6,6 @@ import { useCoreFilter } from '../../combobox/root/utils/useFilter';
 import { stringifyAsLabel } from '../../utils/resolveValueLabel';
 import type { Group } from '../../utils/resolveValueLabel';
 
-const DEFAULT_FILTER_OPTIONS = { sensitivity: 'base' } as const;
-
 /**
  * Groups all parts of the autocomplete.
  * Doesn't render its own HTML element.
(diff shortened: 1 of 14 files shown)


packages/react/src/autocomplete/root/AutocompleteRoot.tsx
```diff
@@ -6,19 +6,17 @@ import { useCoreFilter } from '../../combobox/root/utils/useFilter';
 import { stringifyAsLabel } from '../../utils/resolveValueLabel';
 import type { Group } from '../../utils/resolveValueLabel';
 
-const DEFAULT_FILTER_OPTIONS = { sensitivity: 'base' } as const;
-
```
atomiks, line 10: Removed this as  it's redundant since the defaults internally use this for Combobox.

packages/react/src/combobox/root/ComboboxRootInternal.tsx
```diff
@@ -415,7 +415,7 @@ export function ComboboxRootInternal<Value = any, Mode extends SelectionMode = '
   useIsoLayoutEffect(() => {
     if (items) {
       valuesRef.current = flatFilteredItems;
-      allValuesRef.current = flatItems;
+      allValuesRef.current = flatItems as Value[];
```
flaviendelangle: Can we update the type of `allValuesRef` to be `readonly Array<any>` and avoid the casting?

atomiks: It gets mutated in `Combobox.Item`, so it can't be readonly

flaviendelangle: Don't we have a problem then?

`allValuesRef.current` is set to equal `flatItems`
And `flatItems` can equal `items` (when there is not grouping.

So we are mutating the `items` prop?
Shouldn't we make sure we pass a shallow copy of the items to the ref?

```diff
   const flatItems = React.useMemo(() => {
     if (!items) {
       return [] as any[];
     }

     if (isGrouped) {
       return items.flatMap((group) => group.items);
     }

-    return items;
+    return [...items]
  }, [items, isGrouped]);
```

atomiks: The mutation doesn't actually happen if an `items` prop is specified (it early-returns). The `allValuesRef` is only used when `items` is missing.

Looking at the rest of the code, it could be that this line is redundant/unnecessary

packages/react/src/combobox/root/ComboboxRootInternal.tsx
```diff
@@ -191,9 +191,9 @@ export function ComboboxRootInternal<Value = any, Mode extends SelectionMode = '
   const query = closeQuery ?? (inputValue === '' ? '' : String(inputValue).trim());
   const isGrouped = isGroupedItems(items);
 
-  const flatItems: Value[] = React.useMemo(() => {
+  const flatItems = React.useMemo(() => {
     if (!items) {
-      return [];
+      return [] as readonly any[];
```
michaldudak: As these arrays are readonly, we can reuse a shared instance instead of creating new ones (here and in more places below)

PR #2822, opened 2025-09-24
[POC] Support controlled props in store

michaldudak:
Adds support to controlled values in the Store, eliminating the need to call `useControlled` and handle updates manually.


packages/utils/src/store/ControllableStore.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+
+      if (isControlled && this.state[key] !== controlled) {
+        // Set the internal state to match the controlled value
+        super.set(key, controlled as State[typeof key]);
+      }
+    }, [key, controlled, defaultValue, isControlled]);
+  }
+
+  public set<T>(key: keyof State, value: T, ...additionalArgs: any[]): void {
```
michaldudak: Overriding `set` is quite straightforward, but I'm not sure how to handle `apply` and `update`. Both of them would have to receive additional parameters for change callbacks (or skip calling them), and `update` wouldn't directly replace the internal state with provided object as it would have to filter out controlled props.

One solution to the callback problem might be not calling them at all automatically by the store.

docs/src/app/(private)/experiments/storeWithControlledValues.tsx
```diff
@@ -0,0 +1,112 @@
...
+  value?: number;
+}
+
+function ControllableComponent(props: Props) {
+  const store = useRefWithInit(
+    () => new ControllableStore<State>({ open: props.defaultOpen ?? false, value: 0 }),
+  ).current;
+
+  store.useControlledProp('open', props.open, props.defaultOpen ?? false, props.onOpenChange);
+  store.useProp('value', props.value ?? 0);
```
atomiks: So this gets rid of the need to manually use an effect to synchronize props to the store?

michaldudak: Yup.

packages/utils/src/store/ReactStore.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+    }, [key, value]);
+  }
+
+  useProps(props: Partial<State>) {
+    // False positive - ESLint thinks we're calling a hook from a class component.
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useIsoLayoutEffect(() => {
+      this.apply(props);
+    }, [props]);
```
atomiks, line 83: So to avoid calling `.apply()` too frequently within the root component, `props` must be a memo'd object

michaldudak: Not really. `apply` itself does change detection so if no properties change, the store won't be updated. Calling `useMemo` would just move this change detection elsewhere (and introduce an unnecessary allocation).

packages/utils/src/store/ControllableStore.ts
```diff
@@ -0,0 +1,82 @@
+import { useIsoLayoutEffect } from '../useIsoLayoutEffect';
+import { Store } from './Store';
+
+export class ControllableStore<State> extends Store<State> {
+  /**
+   * Keeps track of which properties are controlled.
+   */
+  #controlledValues: Map<keyof State, boolean> = new Map();
+
+  #changeCallbacks: Map<keyof State, (value: any, ...rest: any) => void> = new Map();
```
romgrk: Personal preference but I don't like ES private properties, I generally prefer using TS `private`.

packages/utils/src/store/ControllableStore.ts
michaldudak: Update: I removed the automatic calling of event handlers. We sometimes need to call them before the state is actually updated, so this gives us flexibility.

packages/utils/src/store/ControllableStore.ts
michaldudak: No preference here - I'm OK with `private` as well.

docs/src/app/(private)/experiments/storeWithControlledValues.tsx
```diff
@@ -0,0 +1,118 @@
+'use client';
+import * as React from 'react';
```
atomiks, line 2: I think this feature should be validated by changing Select + Combobox's actual source as well to use it

michaldudak: I'm working on updating the Dialog to use it right now, as I need it for detached triggers.

michaldudak: https://github.com/mui/base-ui/pull/2834 - seems like it works well

packages/utils/src/store/Store.ts
```diff
@@ -5,6 +5,10 @@ type Listener<T> = (state: T) => void;
  * It uses an observer pattern to notify subscribers when the state changes.
  */
 export class Store<State> {
+  /**
+   * The current state of the store.
+   * This proprerty shouldn't be used directly. Use `useState` to subscribe to state changes.
+   */
```
atomiks, line 15: Why shouldn't it be used directly if inside an event handler or effect? also, `property` has a typo

michaldudak: It's not really a strict requirement. More of a way to ensure consistency with useState - if you set ordinary React state, it won't be available immediately during this render. Same with our store's state returned with `useState`, but the value in `store.state` is updated immediately.
I'll update the docs to mention this.

atomiks: Shouldn't it mention it's fine for non-reactive reads but not writes then? And that it should be preferred over `useState` to avoid a subscription if the value isn't needed during render because it's not reactive

PR #2823, opened 2025-09-24
[code-infra] Update changelog script to use utils from code-infra

brijeshb42:

### DIFF
--- a/scripts/releaseChangelog.mts
+++ b/scripts/releaseChangelog.mts
@@ -1,34 +1,25 @@
-/* eslint-disable no-console */
-import childProcess from 'node:child_process';
-import { promisify } from 'node:util';
+import {
+  fetchCommitsBetweenRefs,
+  findLatestTaggedVersion,
+  type FetchedCommitDetails,
+} from '@mui/internal-code-infra/changelog';
 import { Octokit } from '@octokit/rest';
 import chalk from 'chalk';
+import readline from 'node:readline/promises';
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
 import startCase from 'es-toolkit/compat/startCase';
 import packageJson from '../package.json' with { type: 'json' };
 
-const exec = promisify(childProcess.exec);
-
 const GENERAL_CHANGES_HEADER = 'general changes';
 
 async function main(parameters: CommandParameters) {
-  const { githubToken, lastRelease: previousReleaseParam, release, format } = parameters;
+  const { lastRelease: previousReleaseParam, release, format } = parameters;
 
-  if (!githubToken) {
-    throw new TypeError(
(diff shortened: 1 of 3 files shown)


scripts/releaseChangelog.mts
```diff
@@ -37,8 +26,15 @@ async function main(parameters: CommandParameters) {
     );
   }
 
-  const commits = await findCommits(octokit, previousRelease, release);
-  const changelogEntries = await getChangelogEntries(commits, octokit);
+  const commits = await fetchCommitsBetweenRefs({
+    lastRelease: previousRelease,
+    release,
+    repo: 'base-ui',
+    // If present, use env token. Otherwise, device based token will be generated
```
Janpot: Would it make sense to deprecate this? To discourage users from having a high privileged long living token in their environment and just use the new auth strategy?

scripts/releaseChangelog.mts
```diff
@@ -1,34 +1,23 @@
...
-      'Unable to authenticate. Make sure you either call the script with `--githubToken $token` or set the `GITHUB_TOKEN` environment variable. The token needs `public_repo` permissions.',
-    );
-  }
-
-  const octokit = new Octokit({
-    auth: githubToken,
-    request: {
-      fetch,
-    },
+  const latestTaggedVersion = await findLatestTaggedVersion({
```
Janpot, line 19: Didn't realize before, but is there sufficient overlap between [`findLatestTaggedVersion`](https://github.com/mui/mui-public/blob/67aee4d1566ff1bdb001664a94aff573072f7818/packages/code-infra/src/utils/changelog.mjs#L37) and [`getLastCanaryTag`](https://github.com/mui/mui-public/blob/67aee4d1566ff1bdb001664a94aff573072f7818/packages/code-infra/src/cli/cmdPublishCanary.mjs#L35) to justify a ahared abstraction?

scripts/releaseChangelog.mts
brijeshb42: Yeah. We can warn users about it. But I am not sure if anyone will read it given its a changelog generation script and it would get buried before the changelog content.
Or we can also warn after the changelog is generated.

Janpot: Or we could warn and ask for confirmation before continuing the script?

scripts/releaseChangelog.mts
brijeshb42: Somewhat similar in that both fetches tags. But `getLastCanaryTag` fetches just that one tag whereas the otherone needs to fetch all to determine that latest one.
I don't see a reason to abstract them just yet.

scripts/releaseChangelog.mts
brijeshb42: Added confirmation.

PR #2827, opened 2025-09-25
[docs][scroll area] Fix hijacking scroll when using touch

### LINKED ISSUE #2825: [scroll area] Horizontal scrollbar block scroll on mobile
*Context: I noticed this while reading #2719.*

## Current behavior

The component is hard to to use on mobile for horizontal scroll use cases.

## Expected behavior

Behaves smoothly.

## Reproducible example

1. Open https://master--base-ui.netlify.app/react/handbook/typescript on mobile iOS or Android, same stuff
2. Try to scroll the page vertically

https://github.com/user-attachments/assets/491ef1aa-58cd-47ef-8efc-8932e9f4dd59

## Base UI version

v1.0.0-beta.3

## Additional context

So, there are a bunch of problems here:

1. https://base-ui.com/react/components/scroll-area has not demo of it. It looks like we need a demo there or on the experiments: https://master--base-ui.netlify.app/experiments/scroll-area/scroll-area-inset.
2. We shouldn't block scrolling in the opposite direction. I have tried to play with

```diff
diff --git a/packages/react/src/scroll-area/scrollbar/ScrollAreaScrollbar.tsx b/packages/react/src/scroll-area/scrollbar/ScrollAreaScrollbar.tsx
index 8d4a415f..571e006d 100644
--- a/packages/react/src/scroll-area/scrollbar/ScrollAreaScrollbar.tsx
+++ b/packages/react/src/scroll-area/scrollbar/ScrollAreaScrollbar.tsx
@@ -190,7 +190,7 @@ export const ScrollAreaScrollbar = React.forwardRef(function ScrollAreaScrollbar
     onPointerUp: handlePointerUp,
     style: {
       position: 'absolute',
-      touchAction: 'none',
+      touchAction: orientation === 'vertical' ? 'pan-x' : 'pan-y',
       WebkitUserSelect: 'none',
       userSelect: 'none',
       ...(orientation === 'vertical' && {
```

Browser support seems good enough now: https://caniuse.com/css-touch-action. Compared to before: https://github.com/mui/material-ui/issues/20990. I guess we don't want to allow another scroll container to start to scroll when we use a touch pointer to move the scroll bar. If we do want t allow it (which I dont exepect us to), the only solution I could think of is either what Radix does, hide the scrollbar on mobile, or do a smart scroll detection: https://github.com/mui/material-ui/blob/3538543794d9c33295ad5e4d015ad9f1dacb0615/packages/mui-material/src/SwipeableDrawer/SwipeableDrawer.js#L374.

3. With a touch pointer, the start event should define the starting point we drag from, not the final value (IMHO). Changing this behavior would match iOS and would solve the hijacking of the scroll position in the scroll container
4. For RTL, I believe we can delete

https://github.com/mui/base-ui/blob/83b60d3814b3d4b0ca8fde46ca9613d29f13b9af/packages/react/src/scroll-area/scrollbar/ScrollAreaScrollbar.tsx#L178-L179

per https://github.com/mui/material-ui/pull/42283/files#r1617830270

https://www.notion.so/mui-org/core-Scroll-Area-278cbfe7b66080e2a862ec8a1dd7a480

atomiks:
Fixes #2825

This prevents pointer interactions from taking effect unless you start scrolling on touch, at which point you can grab the thumb like on iOS.

Preview: https://deploy-preview-2827--base-ui.netlify.app/react/components/scroll-area

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/scroll-area/demos/hero/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/scroll-area/demos/hero/css-modules/index.module.css
@@ -42,12 +42,14 @@
   margin: 0.5rem;
   opacity: 0;
   transition: opacity 150ms 300ms;
+  pointer-events: none;
 
   &[data-hovering],
   &[data-scrolling] {
     opacity: 1;
     transition-duration: 75ms;
     transition-delay: 0ms;
+    pointer-events: auto;
   }
 
   &::before {
--- a/docs/src/app/(public)/(content)/react/components/scroll-area/demos/hero/tailwind/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/scroll-area/demos/hero/tailwind/index.tsx
@@ -27,7 +27,7 @@ export default function ExampleScrollArea() {
           </p>
         </div>
       </ScrollArea.Viewport>
-      <ScrollArea.Scrollbar className="m-2 flex w-1 justify-center rounded bg-gray-200 opacity-0 transition-opacity delay-300 data-[hovering]:opacity-100 data-[hovering]:delay-0 data-[hovering]:duration-75 data-[scrolling]:opacity-100 data-[scrolling]:delay-0 data-[scrolling]:duration-75">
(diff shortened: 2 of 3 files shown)


oliviertassinari:
Thoughts:

- The diff seems to fix the issue
- https://joy-ui.com/scroll-area/ is broken too
- this is the kind of style that feels like it should be built-in
- should we add an experiment demo for this?

atomiks:
> this is the kind of style that feels like it should be built-in

Yes, maybe, though the user might not want these styles with "inset" scrollbars that are always visible in some layouts (e.g. if the whole app is a single container with scrollbars around it) - I don't know of the use case on touch devices, but, we're kind of trying to avoid using functional styles where possible since it's harder to override them if necessary

> should we add an experiment demo for this?

I don't see it as necessary as the docs have examples of both vertical + horizontal already to test it with.

michaldudak (approved):

PR #2829, opened 2025-09-25
[combobox][autocomplete] Fix stale `onItemHighlighted` data when filtering with `autoHighlight`

atomiks:
`setIndices({ activeIndex: 0 })` is called on `input-change` but the list hasn't filtered yet, so it grabs a stale value at the given index to call `onItemHighlighted` with.  An effect is needed to call `onItemHighlighted` when the filtered items are determined on the following render.


mnajdova (approved):

michaldudak (changes requested):
In Autocomplete, when I type in a value that's not in the suggestion list, two itemHighlighted events are fired: one with index = 0 and another with index = -1


https://github.com/user-attachments/assets/9445fade-79c9-450d-9691-fdb56c8a7850

michaldudak:
Combobox issue:
1. Focus the combobox with `autoHighlight` (I tried with the modified hero demo)
2. Type in 'ap'
3. Press <kbd>Enter</kbd> to select Apple
4. 3 highlight events are fired:
   Screenshot 2025-09-30 at 12 50 48

michaldudak:
The `onItemHighlight` doesn't fire when the Autocomplete popup opens with `autoHighlight` (but fires when the next character is typed). So it seems that internally the selected item is -1 right after opening?

PR #2831, opened 2025-09-25
[combobox][autocomplete] Improve JSDoc

atomiks:
A few tweaks to match other components. 

The `onClick` handler is also documented for `Menu.Item`, which indicates it's used for click events (and not e.g. an alternative API like `onPress` or `onSelect`)

### DIFF
--- a/packages/react/src/combobox/item/ComboboxItem.tsx
+++ b/packages/react/src/combobox/item/ComboboxItem.tsx
@@ -208,6 +208,11 @@ export namespace ComboboxItem {
       Omit<BaseUIComponentProps<'div', State>, 'id'> {
     children?: React.ReactNode;
     /**
+     * An optional click handler for the item when selected.
+     * It fires when clicking the item with the pointer, as well as when pressing `Enter` with the keyboard if the `Input` or `List` element has focus.
+     */
+    onClick?: React.MouseEventHandler<HTMLElement>;
+    /**
      * The index of the item in the list. Improves performance when specified by avoiding the need to calculate the index automatically from the DOM.
      */
     index?: number;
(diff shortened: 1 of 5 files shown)


packages/react/src/combobox/item/ComboboxItem.tsx
```diff
@@ -207,6 +207,11 @@ export namespace ComboboxItem {
     extends NonNativeButtonProps,
       Omit<BaseUIComponentProps<'div', State>, 'id'> {
     children?: React.ReactNode;
+    /**
+     * An optional click handler for the item when selected.
+     * It fires when clicking the item with the pointer, as well as when pressing `Enter` with the keyboard if the `Input` or `List` element has focus.
```
michaldudak:
```suggestion
     * It fires when clicking the item with the pointer, as well as when pressing `Enter` with the keyboard if the item is highlighted and the `Input` or `List` element has focus.
```

Not sure about the "`Input` or `List` element has focus" - can we have a highlighted item without Input or List focused?

atomiks: It listens for `onKeyDown` on those two components, so even if it's highlighted in some other way, it won't fire the handler.

scripts/api-docs-builder/src/componentHandler.ts
```diff
@@ -85,7 +85,126 @@ function rewriteTypeValue(value: string, componentGroup: string): string {
     next = next.replaceAll(/\bCombobox\./g, 'Autocomplete.');
   }
 
-  return next;
+  return dedupeUnionMembers(next);
```
michaldudak, line 88: Adding this to my TODO list for TS API extractor.

michaldudak (approved):

PR #2832, opened 2025-09-25
[slider] Allow thumbs to swap for pointers

### LINKED ISSUE #1220: Swap thumbs on range slider when thumbs are overlapping
Coming from `mui-base`, `base-ui-components` seems to lack the correct thumb placement (will be fixed with https://github.com/mui/base-ui/pull/1175) and the ability to swap thumbs.  While swapping values does not follow the [Slider WAI-ARIA design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/slider-multithumb) (see [relevant RadixUI discussion about that](https://github.com/radix-ui/primitives/issues/2247#issuecomment-2299363019)), swapping might provide better UX, specially when thumbs are overlapping.

Where is a comparison with`base-ui-components`  vs  `mui-base`. In the base ui example, since the clicks happen on the left side of the thumb, the left thumb is selected, even thought there is an intent to slide to the right.

https://github.com/user-attachments/assets/7ae03807-237f-4d37-95a0-4122566ae882

https://github.com/user-attachments/assets/3751dec9-bd28-4dc6-9633-8b7080dcab54

So, it could be great to have the ability to enable always thumb swaping like in `mui-base`, and would be amazing if this library manages to include a threshold of when it should overlap. I couldn't find a library that does this last behavior 🤔

atomiks:
This allows pointers to swap min/max thumbs, while keyboard usage cannot (primary screen reader usage) as a trade-off

iOS VoiceOver already automatically changes between thumbs when swiping up to move the thumb near the other one, even without this change.

Closes https://github.com/mui/base-ui/issues/1220

Preview: https://deploy-preview-2832--base-ui.netlify.app/react/components/slider#range-slider
Many thumbs: https://codesandbox.io/p/devbox/sleepy-zeh-999qf4


mnajdova (approved):
> Many thumbs: https://codesandbox.io/p/devbox/sleepy-zeh-999qf4

This is broken, or not shared publicly. Works as expected, I've tested with Voice over too. I wonder if it makes sense to include these information somewhere in the docs page, maybe under Accessibility section to explain why the feature works only with pointer. It may be surprising for developers.

packages/react/src/slider/control/SliderControl.tsx
```diff
@@ -119,6 +119,23 @@ export const SliderControl = React.forwardRef(function SliderControl(
...
+
+    const thumbElement = thumbRefs.current[nextIndex];
+
+    if (!thumbElement) {
+      pressedThumbCenterOffsetRef.current = null;
+      pressedInputRef.current = null;
+      return;
+    }
+
+    pressedThumbCenterOffsetRef.current = 0;
```
mj12albert: Does it work if this isn't reset to 0? 
Currently the pointer will shift to the center of the new thumb after the thumbs swap

atomiks: Yeah, it still seems to work and fixes the bug when the pointer isn't dragging from the center

atomiks:
We need a prop to disable this (or go the other way, to enable it), per https://github.com/radix-ui/primitives/issues/2247

atomiks:
Replacing with https://github.com/mui/base-ui/pull/2856

PR #2834, opened 2025-09-25
[dialog, alert dialog] Keep state in a store

michaldudak:
Updates the internal implementation of Dialog and AlertDialog to use the Store instance. This is required for detached triggers.

This is created on top of https://github.com/mui/base-ui/pull/2822.


packages/react/src/dialog/store.ts
```diff
@@ -0,0 +1,118 @@
...
+    return new DialogStore(initialState);
+  }
+
+  public elements: ElementRefs = {
+    popupRef: { current: null },
+    backdropRef: { current: null },
+    internalBackdropRef: { current: null },
+  };
+
+  public useState<K extends keyof Selectors>(key: K): ReturnType<Selectors[K]> {
```
michaldudak: `const open = store.useState('open')` feels much better to me than `const open = useStore(store, selectors.open)` - fewer imports and shorter code.

cc @romgrk

romgrk: I'm happy to refactor the API to make things more ergonomic. But I think I'd go more for something like this:

```js
const open = store.use(selectors.open)
```

Reasons:
 - Selector support is required anyway, and I prefer to expose one right way to do things rather than multiple APIs (reduces the decision load for the developper).
 - I try to make commonly used low-level primitives allocation-less on render, which pre-declared selectors make possible.

packages/react/src/dialog/root/useDialogRoot.ts
```diff
@@ -1,10 +1,8 @@
 'use client';
 import * as React from 'react';
-import { useControlled } from '@base-ui-components/utils/useControlled';
+// store.useState provides typed slices; no direct useStore here
```
atomiks: stale

packages/react/src/dialog/root/useDialogRoot.ts
```diff
@@ -111,14 +95,30 @@ export function useDialogRoot(params: useDialogRoot.Parameters): useDialogRoot.R
 
   floatingEvents = context.events;
 
+  React.useLayoutEffect(() => {
```
atomiks:
```suggestion
  React.useEffect(() => {
```

packages/react/src/dialog/store.ts
michaldudak: Selectors are still predefined; they are just not exposed directly.
Why I think my approach works better here:

- It has similar syntax to the `.set` method (`store.set('open', true)`).
- It prevents developers from making mistakes such as `store.use(state => state.open)` and not using the preallocated selector.

I'm not going to die on this hill, though - the main improvement is not having to import `useStore` anyway.

michaldudak:
~~I introduced strongly typed event handlers and cleared the context even further.~~

EDIT: events are no longer used.

packages/react/src/dialog/root/useDialogRoot.ts
michaldudak: I created a hook for event subscriptions (that uses `useEffect`).

packages/react/src/dialog/popup/DialogPopup.tsx
```diff
@@ -39,44 +39,45 @@ export const DialogPopup = React.forwardRef(function DialogPopup(
...
   useDialogPortalContext();
 
   useOpenChangeComplete({
     open,
-    ref: popupRef,
+    ref: store.elements.popupRef,
     onComplete() {
       if (open) {
-        onOpenChangeComplete?.(true);
+        store.events.emit('openChangeComplete', true);
```
atomiks: Slightly confused about the `events` part. Why is this used instead of `store.state.onOpenChangeComplete?.(true)`

michaldudak: It could be that as well. It feels a bit off to me to keep event handlers in state, as they're not a piece of state. They also might change on every render, which would trigger unnecessary store updates.

michaldudak: Alternatively, we could have `store.eventHandlers.onOpenChangeComplete` registered with a custom hook that wraps the provided prop in `useEventCallback` - this would also separate these handlers from state and prevent unnecessary updates. Plus, remove this `events.emit` indirection.

```ts
// Root
const { onOpenChangeComplete, ... } = props;
store.useEventHandler('onOpenChangeComplete', onOpenChangeComplete); // wraps in useEventCallback, ensures it's not null eliminating the need to call conditionally

// Popup
store.eventHandlers.onOpenChangeComplete(true);
```

atomiks: I'm not really a fan of having random new properties that increases the API surface to understand/learn

I'd still consider it "state" in the sense that it's derived from props and represents a given value at a given time, even if it lacks an associated setter

docs/src/app/(private)/experiments/storeWithControlledValues.tsx
```diff
@@ -0,0 +1,118 @@
...
+function ControllableComponent(props: Props) {
+  const store = useRefWithInit(
+    () => new ControllableStore<State>({ open: props.defaultOpen ?? false, value: 0 }),
+  ).current;
+
+  store.useControlledProp('open', props.open, props.defaultOpen ?? false);
+  store.useSyncedValue('value', props.value ?? 0);
+
+  const open = useStore(store, (state) => state.open);
+  const value = useStore(store, (state) => state.value);
```
atomiks: `store.useState()`?

packages/react/src/dialog/backdrop/DialogBackdrop.tsx
```diff
@@ -36,7 +41,7 @@ export const DialogBackdrop = React.forwardRef(function DialogBackdrop(
 
   return useRenderElement('div', componentProps, {
     state,
-    ref: [backdropRef, forwardedRef],
+    ref: [store.elements.backdropRef, forwardedRef],
```
atomiks: Same here: I'd rather everything be contained in `.state.`, less to learn & understand why there are differing properties

michaldudak: The main difference between state items and others like refs and event handlers is that you should always access state with `store.useState` and never do so for everything else (as it's pointless).

atomiks: I'd be fine differentiating just two things with `props` and `state` in that case. I'm not sure about `store.events`, though

docs/src/app/(private)/experiments/storeWithControlledValues.tsx
michaldudak: Currently `useState` is implemented only in DialogStore, but if it's OK, I'll move it to the base Store.

atomiks: I imagine every other component will also use it, so that makes sense

packages/react/src/dialog/backdrop/DialogBackdrop.tsx
michaldudak: Things like `backdropRef` aren't necessarily props, though. As for `events` - this is meant to be an improved implementation of Floating UI's event emitter (strongly typed and not tied to the floating context).

atomiks: I'd consider them props from the view of the consuming  child components. Unless `context` instead of `props` is better.

michaldudak: How about `refs`?

atomiks: I'd want `store.context.onOpenChangeComplete?.(true)` and `store.context.backdropRef`. And state is whatever can be used with `store.useState()` I'd like to avoid arbitrary namespaces on the object

michaldudak: > I'd consider them props from the view of the consuming child components.

"Props" have a precise meaning in React. These aren't props.

michaldudak: > I'd want store.context.onOpenChangeComplete?.(true) and store.context.backdropRef. And state is whatever can be used with store.useState() I'd like to avoid arbitrary namespaces on the object

OK, this could be a good compromise. So `context` would be everything that's not reactive.

michaldudak: I changed the implementation and I can't say I'm fully happy with it. I realized that we now have a context (DialogRootContext) that contains a `store` which in turn contains another `context` - this can be confusing (plus, we have yet another context in state, but that's at least called `floatingContext`).

packages/react/src/dialog/store.ts
romgrk: Tbh I'm not super happy with `.set()` to start with, I added it as a temporary ergonomic API but I didn't end up having enough time to complete the vision I had - I was originally hoping to go for full functional lenses (a lense is a read/write object, while a selector is a read-only object) to get an API like `store.use(lens.open)` and `store.set(lens.open, true)` (if you know jotai, lenses are pretty much the thing they call [atoms](https://jotai.org/docs/core/atom)). I'm still somewhat hoping to complete that project if I can find some free time.

So anyway, for that reason I'd prefer to limit the API surface for the moment, but I also see how convenient `.useState()` is immediately, so I'll also not die on this hill.

packages/react/src/alert-dialog/trigger/AlertDialogTrigger.tsx
```diff
@@ -39,6 +41,13 @@ export const AlertDialogTrigger = React.forwardRef(function AlertDialogTrigger(
     native: nativeButton,
   });
 
+  const setTriggerElement = React.useCallback(
+    (node: HTMLButtonElement | null) => {
+      store.set('triggerElement', node);
+    },
+    [store],
+  );
+
```
romgrk: This seems to be a common pattern, we might want to add it as a `Store` method at some point.

packages/react/src/dialog/description/DialogDescription.tsx
```diff
@@ -17,16 +17,16 @@ export const DialogDescription = React.forwardRef(function DialogDescription(
...
 
   useIsoLayoutEffect(() => {
-    setDescriptionElementId(id);
+    store.set('descriptionElementId', id);
     return () => {
-      setDescriptionElementId(undefined);
+      store.set('descriptionElementId', undefined);
     };
-  }, [id, setDescriptionElementId]);
+  }, [id, store]);
```
romgrk: Another common pattern. The semantics are close but slightly different than `.useSyncedValue()`.

packages/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -14,53 +17,94 @@ import { BaseUIEventDetails } from '../../utils/createBaseUIEventDetails';
...
+    descriptionElementId: undefined,
+    titleElementId: undefined,
+    openMethod: null,
+    mounted: false,
+    transitionStatus: 'idle',
+    nestedOpenDialogCount: 0,
+    triggerProps: EMPTY_OBJECT,
+    popupProps: EMPTY_OBJECT,
+    floatingRootContext: getEmptyContext(),
+  }).current;
```
romgrk: Tbh avoiding allocating the inline init function forces allocating this big params object on every render, so it's not a win. We might as well just call `useRefWithInit(() => new DialogStore({ ... }))`.

packages/utils/src/store/Store.ts
```diff
@@ -1,17 +1,46 @@
...
+  /**
+   * Non-reactive values such as refs, callbacks, etc.
+   * Unlike `state`, this property can be accessed directly.
+   */
+  public readonly context: Context;
+
   private listeners: Set<Listener<State>>;
 
-  constructor(state: State) {
+  private selectors?: Selectors;
```
romgrk:
```suggestion
  private selectors: Selectors | undefined;
```

nitpick: The `?` notation implies the possibility of the field being absent, while `| undefined` implies the field is always there but possibly contains an `undefined` value. The latter is more correct.

packages/utils/src/store/Store.ts
romgrk, line 1: Conceptually speaking, I think we could keep the base `Store` clean of any React and DOM specific methods - so it's pure - and put all the React related methods in the descendant class which we could rename `ReactStore` instead of `ControllableStore`.

packages/utils/src/store/Store.ts
```diff
@@ -71,4 +100,41 @@ export class Store<State> {
...
+  }
+
+  /**
+   * Wraps a function with `useEventCallback` to ensure it has a stable reference
+   * and assigns it to the context.
+   *
+   * @param key Key of the event callback. Must be a function in the context.
+   * @param fn Function to assign.
+   */
+  public useEventCallback<Key extends ContextFunctionKeys<Context>>(
```
romgrk: The naming makes it feel like it should return a function like a `useEventCallback()` call would do. Maybe I'd rework the name a bit - `store.useContextCallback()`?.

packages/react/src/dialog/root/DialogRoot.tsx
michaldudak: This can be extracted to a const outside of any functions, so it won't be reallocated.

michaldudak:
One other improvement - I moved `setOpen` to the store, so all parts can access it without having to set up refs and event callbacks.

michaldudak:
@romgrk, I addressed your comments and updated #2822. I'd like this PR to focus just on Dialog changes.

packages/react/src/dialog/store.ts
michaldudak: I'm happy to revisit this once a better implementation is in place. For now, I'd like to stick with `useState`.

packages/react/src/alert-dialog/popup/AlertDialogPopup.tsx
```diff
@@ -39,49 +38,42 @@ export const AlertDialogPopup = React.forwardRef(function AlertDialogPopup(
...
       if (open) {
-        onOpenChangeComplete?.(true);
+        store.context.openChangeComplete?.(true);
       }
     },
   });
 
-  const mergedRef = useMergedRefs(forwardedRef, popupRef);
-
-  const { popupProps } = useDialogPopup({
```
michaldudak, line 75: useDialogPopup was basically just preparing props, so I removed it and moved its logic in DialogPopup and AlertDialogPopup.

PR #2835, opened 2025-09-25
[slider] Fix regression of 3+ thumb sliders

atomiks:
Trying to drag the third thumb would drag the first (post-release regression)


mj12albert (approved):
👍

PR #2847, opened 2025-09-26
[portals] Ensure `container` is reactive from `undefined` transitions

### LINKED ISSUE #2780: [toast] Portal doesn't respect provided `container`
# Bug report

## Current behavior

We're providing a `container` to `Toast.Portal` but it's still using the default portal.

## Expected behavior

`Toast.Portal` honours `container` prop.

## Reproducible example

https://codesandbox.io/p/devbox/kind-shadow-y4prt7?workspaceId=ws_4qKVDKGuRrrvd8oLfoRWSz

## Base UI version

For example v1.0.0-beta.3

Chrome

## Which OS are you using?

Mac OS

Voiceover/JAWS etc.

## Additional context

atomiks:
Fixes #2780

Also removed support for `id` as we don't need it.

https://codesandbox.io/p/sandbox/kind-moon-fstqzf

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingPortal.test.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingPortal.test.tsx
@@ -1,13 +1,14 @@
-import { fireEvent, flushMicrotasks, render, screen } from '@mui/internal-test-utils';
 import * as React from 'react';
-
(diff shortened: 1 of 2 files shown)


atomiks:
Replaced with #2889

PR #2848, opened 2025-09-26
[docs] Fix backdrops on iOS 26

### LINKED ISSUE #2799: [backdrops] Add `document.body.clientHeight` as a CSS variable

atomiks:
Fixes #2799

Alternative to #2807

### DIFF
--- a/docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
+++ b/docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
@@ -13,7 +13,9 @@ npm i @base-ui-components/react
 
 All components are included in a single package. Base UI is tree-shakeable, so your app bundle will contain only the components that you actually use.
 
-## Set up portals
+## Set up
+
+### Portals
 
 Base UI uses portals for components that render popups, such as Dialog and Popover.
 To make portalled components always appear on top of the entire page, add the following style to your application layout root:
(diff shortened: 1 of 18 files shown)


docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
```diff
@@ -36,6 +38,16 @@ To make portalled components always appear on top of the entire page, add the fo
 This style creates a separate stacking context for your application's `.root` element.
 This way, popups will always appear above the page contents, and any `z-index` property in your styles won't interfere with them.
 
+### iOS Safari
+
+Backdrops must use `position: absolute` instead of `position: fixed` in order to cover the entire visible viewport on iOS 26. For this to work after scrolling, `position: relative` must be set on the `<body>`.
+
+```css title="styles.css"
+body {
+  position: relative;
```
michaldudak, line 47: Will it still work if body is shorter than the viewport?

atomiks: Hmm, nope it doesn't. And none of the viewport units will cover the whole screen on iOS in that case 🫠  so `height: 100vh` on `.Backdrop` won't work. 

https://pbs.twimg.com/media/G1buu_1W0AAzcZh?format=jpg&name=large

`html, body { height: 100% }` is yet _another_ requirement (edit: that doesn't really work either. I don't think it can ever extend past that if you're at the bottom of the page already.)

atomiks: `100dvh` essentially matches `document.body.clientHeight` when it's shorter than the viewport and `<body>` has `position: relative`

atomiks:
Even with the backdrop covering the bottom of the screen, "drawers" that emerge from the bottom look pretty bad still:


Screenshot 2025-09-26 at 9 48 38 pm


The backdrop needs to be the same color at the bottom as the popup element itself instead of transparent. This is a different topic to consider, though.

PR #2849, opened 2025-09-26
[context menu] Fix initial double click

### LINKED ISSUE #2828: ContextMenu double click
After opening the context menu, the first click on an element triggers a double click. Therefore, clicking a checkbox toggles its state twice.

atomiks:
Fixes #2828

@sadetskayfu since I can't reproduce it, can you let me know if this fixes it? It's a hunch that it's the mouseup/drag-release behavior.  https://codesandbox.io/p/devbox/runtime-http-t7t2pc (using this PR's release version)

For context menu, this should only apply if `even.button === 2` in `onMouseUp`, though this could still not work in the "right click pointerdown (without release) -> drag -> mouseup on item" scenario


sadetskayfu:
Yes, it helped me. Thank you

michaldudak (approved):

Issue #2634, opened 2025-09-04
[docs] Prop defaults in docs


vladmoroz:
# Docs feedback

## How can we improve Base UI documentation?

First of all, appreciate the table update with more details behind the accordions.

Would love to see the prop defaults presented in a cleaner way.

Current:

<img width="1668" height="1946" alt="Image" src="https://github.com/user-attachments/assets/195ab67c-62d7-4737-8b1e-ae5c59d1dc1f" />

***

Suggested:

<img width="1668" height="1946" alt="Image" src="https://github.com/user-attachments/assets/978b402f-5431-454a-bdda-e8c487478cf4" />

***

In the current version I find it difficult to scan both the type defs and the defaults given they are joined in a single column:
- The parenthesis and the repeating “defaults” word are noisy
- It's hard to scan which props are initialized with a default value at all
- It's hard to scan props based on their type. Even if I know that I am looking for a boolean, the type is just hard to pick it out scanning the table.

oliviertassinari:
We would need to account for default value that could be really long, e.g. `reduceAnimations` in https://github.com/mui/material-ui/issues/34085#issuecomment-1232786535. I'm not sure how it's handled today, but each column we add reduces the space available, so this proposal involves this consideration.

mj12albert:
> We would need to account for default value that could be really long

Each "cell" in both the original/current design accounts for this, e.g. see the default for `collisionBoundary` in the old design: https://deploy-preview-2359--base-ui.netlify.app/react/components/tooltip#positioner

PR #2851, opened 2025-09-26
[docs] Separate prop defaults column

mj12albert:
Preview: https://deploy-preview-2851--base-ui.netlify.app/react/components/checkbox#root

The layout matches the [original](https://deploy-preview-2359--base-ui.netlify.app/react/components/checkbox#root) `<table>` design:

Screenshot 2025-09-26 at 7 02 20 PM

Fixes https://github.com/mui/base-ui/issues/2634


docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
```diff
@@ -279,12 +278,11 @@ const TRIGGER_GRID_LAYOUT =
   'xs:grid ' +
   'xs:grid-cols-[theme(spacing.48)_1fr_theme(spacing.10)] ' +
   'sm:grid-cols-[theme(spacing.56)_1fr_theme(spacing.10)] ' +
-  'md:grid-cols-[1fr_2fr_theme(spacing.10)] ';
+  'md:grid-cols-[5fr_7fr_3fr_theme(spacing.10)] ';
```
atomiks: Let's try to optimize for there to be no overflowing defaults. I think `clipping-ancestors` is our longest default name, and 3.5fr or 4fr could handle it

atomiks:
I feel like `undefined` could just be `-` as it's quite distracting compared to before

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
LukasTy: Good idea. 👍 
However, the overflow will still be present depending on the viewport width. 🙈

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
```diff
@@ -139,10 +141,13 @@ export async function PropsReferenceAccordion({
           useMDXComponents: () => ({ code: TableCode }),
         });
 
-        const PropDefault = await createMdxComponent(`\`${prop.required ? '—' : prop.default}\``, {
-          rehypePlugins: rehypeSyntaxHighlighting,
-          useMDXComponents: () => ({ code: TableCode }),
-        });
+        const PropDefault = await createMdxComponent(
+          `\`${prop.required || prop.default === undefined ? '—' : prop.default}\``,
```
LukasTy: What do you think about leaving it empty instead of showing a dash in case there is no explicit default? 🤔

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
atomiks: Yeah, but we're optimizing for large desktop screens here

mj12albert:
> I assume this is something we'd like to avoid

oops fixed

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
mj12albert: I think having the dash (or something) helps to make the empty space (there's usually no default) look less empty and look "more aligned"

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
mj12albert: I changed it to `4.5fr` and it seems good, though it still clips when you are close to the breakpoints where sidebars appear (`1024px` and `1344px`)

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
vladmoroz: Yeah it's better with the dash to keep the column rhythm, but it needs to use a much lighter color (check `var(--syntax-nullish)`)

mj12albert: Updated the color in https://github.com/mui/base-ui/pull/2851/commits/9b4eeb68356de9de0cf3d489a1fc0b3a430b286a

atomiks (approved):
Looks great!

docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
LukasTy: Ok, it looks solid with a less pronounced color.
Thank you for sharing your opinions. 👍

LukasTy (approved):

PR #2856, opened 2025-09-29
[slider] Add `thumbCollisionBehavior` prop

### LINKED ISSUE #1220: Swap thumbs on range slider when thumbs are overlapping

### LINKED ISSUE #1730: [Slider] Change default `minStepsBetweenValues` from 0 to 1
Consider changing the default to `1`, range sliders by default shouldn't allow values like `[10, 10]`

Also an opportunity to rename this prop if we can come up with a better name, originally suggested in https://github.com/mui/base-ui/issues/981

atomiks:
Alternative to  https://github.com/mui/base-ui/pull/2832

Closes #1220
Closes #1730

```js
thumbCollisionBehavior="push|swap|none"
```

Preview: https://deploy-preview-2856--base-ui.netlify.app/react/components/slider#range-slider

- `none`: seems to work ok if `minStepsBetweenValues` was specified to a sufficiently large visual gap. The current non-configurable behavior.
- `swap`: seems to work best with 3+ thumb sliders with no min steps between thumbs, or gradient/color pickers. 
- `push` (default): works well with 2 thumb range sliders with or without `minStepsBetweenValues`; allows the ability to configure the min/max thumbs simultaneously with a single pointer interaction for certain ranges


colmtuite:
Do we need to handle the SR experience here somehow? When you push a thumb by dragging the other thumb, only the dragged thumb value gets announced, even though both have been updated.

atomiks:
@colmtuite I think for users using a pointer, their sight is good enough to not rely on a screen reader update as they can see the other thumb moved. If they're non-sighted they'll be using the keyboard where thumb collision behavior doesn't apply. On iOS VoiceOver, the thumbs already swap though when swiping up as focus moves automatically (edit it's not reliable though, sometimes it doesn't; maybe it depends where the thumb is swiping and it catches it)

michaldudak:
"swap" makes the slider glitch when used with `minStepsBetweenValues`:

https://github.com/user-attachments/assets/c9f9d0ff-df57-406c-9936-8317881dfccb

atomiks:
@michaldudak fixed this, though the UX for swap is pretty bad if `minStepsBetweenValues` is used because it creates a dead zone where nothing happens while dragging, so it probably should be avoided anyway

michaldudak:
Still, if we allow this, I'm pretty sure it'll have its users. I found two more glitches (likely related):

1. Swapping updates the previously dragged trigger's value (so during swap both thumbs actually move):
   
   https://github.com/user-attachments/assets/e63ecbce-2d4c-4a21-bab9-a3e36ab988f4

2. Moving the pointer quickly positions the thumbs all over the track:
   
   https://github.com/user-attachments/assets/ef356815-061e-4f16-989a-64dd07113444

mj12albert:
Push seems a good default, certainly better than the current overlapping behavior; this PR could close this issue as well: https://github.com/mui/base-ui/issues/1730

atomiks:
@mj12albert added to close in PR description

docs/reference/generated/slider-root.json
```diff
@@ -42,6 +42,12 @@
       "description": "How the thumb(s) are aligned relative to `Slider.Control` when the value is at `min` or `max`:\n- `center`: The center of the thumb is aligned with the control edge\n- `edge`: The thumb is inset within the control such that its edge is aligned with the control edge\n- `edge-client-only`: Same as `edge` but renders after React hydration on the client, reducing bundle size in return",
       "detailedType": "'center' | 'edge' | 'edge-client-only' | undefined"
     },
+    "thumbCollisionBehavior": {
+      "type": "'none' | 'push' | 'swap'",
+      "default": "'push'",
```
LukasTy, line 47: Shouldn't we treat this as a BC?
Current "default" seems to be `none`. 🤔

docs/src/app/(public)/(content)/react/components/slider/page.mdx
```diff
@@ -37,6 +37,8 @@ To create a range slider:
 1. Pass an array of values and place a `<Slider.Thumb>` for each value in the array
 2. Additionally for server-side rendering, specify a numeric `index` for each thumb that corresponds to the index of its value in the value array
 
+Thumbs can be configured to behave differently when they collide during pointer interactions using the `thumbCollisionBehavior` prop on `<Slider.Root>`.
```
LukasTy, line 40: Would it be worth adding another sentence further clarifying that this will not apply to keyboard usage? 🤔

mj12albert (approved):
Looks good 👍

mj12albert:
New `push` default is breaking right?

mj12albert:
> Wasn't the plan to bump `minStepsBetweenValues` default to `1`?

That was way before we started to reconsider thumb-swapping, since the new default is `push` it makes more sense for `minSteps` to remain 0 so it only starts to push when the thumbs overlap at the same value

docs/src/app/(public)/(content)/react/components/slider/page.mdx
mj12albert: "during pointer interactions" seems clear enough, but it may be helpful to explain why it doesn't apply to keyboard

PR #2860, opened 2025-09-30
[combobox] Clear single value on input clear

atomiks:
Retaining the value in a single-select combobox when clearing the input is somewhat unexpected behavior

Preview: https://deploy-preview-2860--base-ui.netlify.app/react/components/combobox

### DIFF
--- a/packages/react/src/combobox/input/ComboboxInput.test.tsx
+++ b/packages/react/src/combobox/input/ComboboxInput.test.tsx
@@ -174,6 +174,43 @@ describe('<Combobox.Input />', () => {
   });
 
   describe('interaction behavior', () => {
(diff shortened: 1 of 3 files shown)


mnajdova (changes requested):
The clearing behavior works, but there is regression, if you tab outside of the combobox it doesn't close.

Edit: Here is a video as it is not 100% reproducible (it does not close the third time):

https://github.com/user-attachments/assets/56853123-0c08-4b04-b18b-f02869ace266

mnajdova:
Ok, it's not a regression it happens on base-ui.com, if you clear and leave only one or two characters, it doesn't close the popup.

mnajdova (approved):
Approving this fix as it's not regression. We are handling it in another PR.

PR #2864, opened 2025-09-30
[combobox] Fix `focusout` of input not closing popup under certain conditions

atomiks:
Fixes https://github.com/mui/base-ui/pull/2860#pullrequestreview-3283149669

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
@@ -1585,6 +1585,63 @@ describe.skipIf(!isJSDOM)('FloatingFocusManager', () => {
     expect(screen.getByTestId('after')).toHaveFocus();
   });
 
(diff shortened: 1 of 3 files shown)


packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
```diff
@@ -1585,6 +1585,63 @@ describe.skipIf(!isJSDOM)('FloatingFocusManager', () => {
     expect(screen.getByTestId('after')).toHaveFocus();
   });
 
+  test('untrapped typeable combobox closes on second tab sequence (click -> tab -> click -> tab)', async () => {
```
mnajdova, line 1588: The test passes on master too, it's likely not testing the right thing.

atomiks: It seems like JSDOM has different behavior than the browser. 

This entire component is wrapped inside a skip so it only runs in JSDOM, probably because it was flaky, so we'll probably have to skip testing this in the browser here.

mnajdova (approved):

PR #2865, opened 2025-09-30
[docs] Use `pointer` cursor for Codesandbox external "link"

LukasTy:
Given that we discussed that only external links should use the `pointer` cursor, I've added this style to the `CodeSandbox` "link".


atomiks:
A bit of an ambiguous case since you can't right click on it to reveal native `<a>` related options 🤔

LukasTy:
> A bit of an ambiguous case since you can't right click on it to reveal native `<a>` related options 🤔

Yeah. That's true.
We can keep it as is, If there is an agreement that this is a blocker for signalling this as a link.
But on the other hand, it has an anchor icon in the text. 🤔 🤷

michaldudak:
I think it's misleading. You can't middle-click to open it in a new tab, for example. I'd leave the default cursor.

PR #2868, opened 2025-09-30
Store refactor

romgrk:
Import changes from the X version:
 - Support recursive state updates efficiently
 - Rename `.update()` to `.setState()`: this naming seems more explicit about what happens
 - Rename `.apply()` to `.update()`: `.update(partialState)` seems more readable than `.apply(partialState)`, "apply" is already used by `Function#apply` for a different purpose

### DIFF
--- a/packages/utils/src/store/ReactStore.ts
+++ b/packages/utils/src/store/ReactStore.ts
@@ -12,7 +12,7 @@ import { NOOP } from '../empty';
  *
  * - Keys registered through {@link useControlledProp} become controlled when a non-undefined
  *   value is provided. Controlled keys mirror the incoming value and ignore local writes
- *   (via {@link set}, {@link apply}, or {@link update}).
+ *   (via {@link set}, {@link update}, or {@link setState}).
  * - When a key is uncontrolled, an optional default value is written once on first render.
  * - Use {@link useSyncedValue} and {@link useSyncedValues} to synchronize external values/props into the
  *   store during a layout phase using {@link useIsoLayoutEffect}.
@@ -79,7 +79,7 @@ export class ReactStore<
    */
   public useSyncedValues(props: Partial<State>) {
     useIsoLayoutEffect(() => {
-      this.apply(props);
+      this.update(props);
     }, [props]);
   }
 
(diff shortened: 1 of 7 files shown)


packages/utils/src/store/Store.ts
```diff
@@ -47,10 +51,22 @@ export class Store<State> {
    *
    * @param newState The new state to set for the store.
    */
-  public update(newState: State) {
-    if (this.state !== newState) {
-      this.state = newState;
-      this.listeners.forEach((l) => l(newState));
+  public setState(newState: State) {
```
michaldudak, line 54: Having both `set` and `setState` isn't ideal, but I can't think of a better name. `replaceState` describes what's happening, but I'm not sure I like it.

We could also rename `set` to `setField`, but since it's commonly used, it's better to have it short.

LukasTy (approved):
Looks good. 👍

packages/utils/src/store/Store.ts
flaviendelangle: >  but since it's commonly used, it's better to have it short.

Yeah, I would rather have a super ugly name for the full state update than compromising the quality of the DX for the `set` method :+1:

packages/utils/src/store/ReactStore.ts
```diff
@@ -113,14 +113,14 @@ export class ReactStore<
       this.controlledValues.set(key, isControlled);
 
       if (!isControlled && !Object.is(this.state[key], defaultValue)) {
-        super.update({ ...(this.state as State), [key]: defaultValue } as State);
+        super.setState({ ...(this.state as State), [key]: defaultValue } as State);
```
flaviendelangle, line 116: Why are we using `setState` here instead of:


```suggestion
        super.set(key, defaultValue);
```

Same for controlled below

romgrk: I didn't write that code but `setState` is more direct, so I can see how it makes sense to use it.

packages/utils/src/store/Store.ts
romgrk: I agree that the full update will be uncommonly used so I'm ok with an ugly name, any suggestion? It could be something like `.updateFull(state)`, by analogy with `.update(partialState)`. I'd also be fine with having both `set` and `setState` - `setState` is just really accurate considering the existing React idioms like `const [state, setState] = useState({})`.

packages/utils/src/store/ReactStore.ts
michaldudak: `super.set` (`Store.set`) calls `this.setState`, which, as it's virtual, ends up being the ReactStore's implementation (that includes filtering), which is not what we want.

flaviendelangle: OK, maybe a JSDoc line making that explicit would be good :+1:

packages/utils/src/store/Store.ts
romgrk: Any more feedback for the naming? In the absence of it, I'd go ahead with `.set` and `.setState`. We can still rename later if we figure out something better.

PR #2870, opened 2025-10-01
[composite] Always stop event propagation for composite keys

atomiks:

This is an alternative to #2702. Given we already stop propagation on some but not all, it should just be universally applied. I also have not seen the  need to track composite keys in parent components while using a composite.


mnajdova (approved):
The change makes sense to me. This PR basically changes the behavior of the Toolbar, Menubar and TabList as those were the only components that did not set the `stopEventPropagation` prop. I manually tested them and everything seems to work as expected.

PR #2872, opened 2025-10-01
[code-infra] Revert `no-descending-specificity` fix

Janpot:
Issue was fixed too eagerly in https://github.com/mui/base-ui/pull/2797/files, some cascades depend on ordering.

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/number-field/demos/hero/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/number-field/demos/hero/css-modules/index.module.css
@@ -54,16 +54,6 @@
   }
 }
 
-.Decrement {
-  border-top-right-radius: 0;
-  border-bottom-right-radius: 0;
-}
-
-.Increment {
-  border-top-left-radius: 0;
-  border-bottom-left-radius: 0;
-}
-
 .Decrement,
 .Increment {
   box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


docs/src/app/(public)/(content)/react/components/select/demos/object-values/css-modules/index.module.css
```diff
@@ -195,6 +189,12 @@
   line-height: 1.5rem;
 }
 
+.ItemDescription {
```
LukasTy, line 192: Just noting that this particular change does not seem impacted by the declaration order.

docs/src/app/(public)/(content)/react/components/number-field/demos/hero/css-modules/index.module.css
```diff
@@ -92,3 +82,13 @@
     background-color: var(--color-gray-100);
   }
 }
+
+.Decrement {
```
LukasTy, line 86: Do you know why Argos did not catch this?
It seems that in Argos this demo looks as expected, whereas on actual page it renders with issues... 🤷 

Image

Janpot: Is argos rendering the css modules variant or only the tailwind one? it's only broken in the css modules variant.

LukasTy: Both. 🤷 
Screenshot 2025-10-01 at 12 41 55

Janpot: Then that must be broken, argos should have caught this if we're rendering the css modules variant.

Janpot: Investigating in https://github.com/mui/base-ui/pull/2875 and https://github.com/argos-ci/argos/issues/1839

PR #2873, opened 2025-10-01
[docs] Fix edge-aligned slider demo styles

mj12albert:
Fixes the thumb's outline causing the thumb to overflow the track

Screenshot 2025-10-01 at 5 25 15 PM


atomiks:
`calc(1rem - 2px)` is probably better as otherwise it creates the muddy border issue

Screenshot 2025-10-01 at 7 34 59 pm

Edit: although it doesn't look good when at the min/max

I think the current styles are actually fine visually, but we could sidestep this entirely if the demo used the same styles as the hero so it becomes more clear `thumbAlignment` only affects the alignment

LukasTy (approved):
Fixes the visual problem. 👌 
As for changing the whole demo - maybe that could be left for after the release? 🤔

docs/src/app/(public)/(content)/react/components/slider/demos/edge-alignment/css-modules/index.module.css
```diff
@@ -11,14 +11,14 @@
...
-  box-shadow: inset 0 0 0 1px var(--color-gray-200);
+  border: 1px solid var(--color-gray-200);
   border-radius: 9999px;
   user-select: none;
 }
 
 .Indicator {
   border-radius: 9999px 0 0 9999px;
-  background-color: var(--color-gray-600);
+  background-color: white;
```
atomiks, line 21: Dark mode looks best with a light gray instead of pure white

PR #2875, opened 2025-10-01
[code-infra] Test argos threshold

Janpot:
Trying out the effect of sensitivity threshold on detection of a [recent regression](https://github.com/mui/base-ui/pull/2872)

Even at 0, it doesn't seem to detect the change


michaldudak:
@Janpot, do you still need this PR open?

PR #2876, opened 2025-10-01
[docs] Update edge-aligned slider demo

mj12albert:
Changed to use the same design as hero

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/slider/page.mdx
+++ b/docs/src/app/(public)/(content)/react/components/slider/page.mdx
@@ -43,7 +43,7 @@ To create a range slider:
 
 ### Thumb alignment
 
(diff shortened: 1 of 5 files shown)


docs/src/app/(public)/(content)/react/components/slider/page.mdx
```diff
@@ -45,6 +45,6 @@ To create a range slider:
 
 Set `thumbAlignment="edge"` to inset the thumb(s) such that their edges align with the edges of the control when the value is at `min` or `max`, without overflowing the control like the default `"center"` alignment.
```
LukasTy: Complete nitpick: WDYT about avoiding the mix of singular and plural?
I feel like it should be somewhat evident that it would apply to several thumbs as well. 🤔 
```suggestion
Set `thumbAlignment="edge"` to inset the thumb such that its edge aligns with the edge of the control when the value is at `min` or `max`, without overflowing the control like the default `"center"` alignment.
```

PR #2883, opened 2025-10-02
[autocomplete][combobox] Move `CompositeList` to `List` component to make `Input` work with composites

chuganzy:
This PR just moves the `CompositeList` from the Root element to the List to make Input element work with Composites, such as Toolbar.

CodeSandbox: https://codesandbox.io/p/sandbox/quizzical-swanson-7q3yqh

It can move the focus from the input successfully because the keyboard event is handled by the CompositeRoot, however it cannot move back with arrow keys because the ComboboxInput is unavailable there but registered to the context in the ComboboxRoot(Internal).

---

As a side note, I am not sure if it is intended to enable focus roving on the input elements (or focusable elements in general) that is not rendered as CompositeItem. Somewhat relates to https://github.com/mui/base-ui/pull/2702.


atomiks:
Works with the PR release preview: https://codesandbox.io/p/sandbox/2883-forked-38q7wf
But this breaks React 18 tests

packages/react/src/composite/list/CompositeList.tsx
```diff
@@ -102,6 +102,19 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
...
+    if (labelsRef) {
+      labelsRef.current.length = 0;
+    }
+  });
+
+  useIsoLayoutEffect(() => {
+    return () => {
+      cleanupRefs();
+    };
+  }, [cleanupRefs]);
```
chuganzy, line 116: The changes in CompoboxList revealed a bug where refs are not cleaned up on unmount, keeps holding all references to the elements. 
This caused the test fail in React 18 Strict Mode, because `listRef` is not empty on the second+ time on open and `waitForListPopulated` in `useListNavigation` assumes it is ready to navigate as soon as it opens.

This was not an issue in React 19 because `IndexGuessBehavior.GuessFromOrder` works more accurate.

packages/react/src/composite/list/CompositeList.tsx
```diff
@@ -102,6 +102,20 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
...
+    };
+  }, [elementsRef]);
+
+  useIsoLayoutEffect(() => {
+    return () => {
+      if (labelsRef) {
+        labelsRef.current = [];
+      }
+    };
+  }, [labelsRef]);
```
chuganzy, line 117: The changes in CompoboxList revealed an issue where refs are not cleaned up on unmount, keeps holding all references to the elements. It caused the test fail in React 18 Strict Mode, because listRef is not empty on the second+ time on open and waitForListPopulated in useListNavigation assumes it is ready to navigate as soon as it opens.

This was not an issue in React 19 Strict Mode, since IndexGuessBehavior.GuessFromOrder behaved more accurately.

atomiks (approved):
Nice, thank you for another fix

PR #2885, opened 2025-10-02
[docs][combobox] Fix popup height style for inside-popup demo

atomiks:
This is a really minor, but weird, issue in the docs style for the `input-inside-popup` demo.

Even if there was enough space on the bottom to position the popup there, it would choose to flip to the top if there was **more** space there. Kind of like `autoPlacement()` in Floating UI instead of `flip()` (even though it's not being used.)

`var(--available-height)` is already handled by `List`, so there aren't any visual changes.

### Fix

With the fix, it correctly chooses the bottom if it can. Previously it was choose the top (shown with the arrow) in this particular scenario. The viewport height and scroll position need to be specific enough for the issue to be reproduced.

Screenshot 2025-10-02 at 4 21 33 pm

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/combobox/demos/input-inside-popup/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/combobox/demos/input-inside-popup/css-modules/index.module.css
@@ -95,7 +95,7 @@
     transform 150ms,
     opacity 150ms;
   max-width: var(--available-width);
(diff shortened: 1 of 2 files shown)


LukasTy (approved):
LGTM. 👍 
With this update, it opens to the bottom with less space left at the bottom (eagerly cuts off more content), whereas on master, it would flip to the top sooner if opening to the bottom would cut off content.

Issue #1893, opened 2025-05-08
[dialog] Lock scroll on iOS if navbar is collapsed


atomiks:
In #1890 the scroll lock for iOS was changed to a basic technique (`overflow: hidden` on `<html>`) that only works if the navbar is expanded.

Numerous bugs and issues arise on iOS when trying to robustly lock the scroll. A solution must avoid all of these bugs to be implemented:

- Textboxes must slide into view when focused, nor cause a glitchy scroll animation
- The navbar must not force itself into view and cause layout shift
- Scroll containers must not flicker upon closing a popup when it has an exit animation

atomiks:
## Possible solution

https://github.com/mui/base-ui/pull/3183

## Workaround

This can be solved externally in the meantime using a special component:

```jsx
const scrollerStyles: React.CSSProperties = {
  width: '100%',
  height: '100%',
  overflowY: 'auto',
  overscrollBehavior: 'none',
  scrollbarWidth: 'none',
  // For old Safari, a class with `::-webkit-scrollbar{display:none}` is needed
};

const overflowWrapperStyles: React.CSSProperties = {
  height: 'calc(100% + 1px)',
};

const stickyWrapperStyles: React.CSSProperties = {
  position: 'sticky' as const,
  top: 0,
  bottom: 0,
  height: 'calc(100% - 1px)',
};

function ScrollLock({ children }) {
  return (
    <div style={scrollerStyles}>
      <div style={overflowWrapperStyles}>
        <div style={stickyWrapperStyles}>{children}</div>
      </div>
    </div>
  );
}
```

### Usage

`<Dialog.Viewport>` was released after `beta.4`, but you can use a `<div>`:

```jsx
<Dialog.Portal>
  <Dialog.Viewport
    render={(props, state) => (
      <div
        {...props}
        style={{
          position: 'fixed',
          inset: 0,
          // Just so long exit animations don't block scroll
          pointerEvents: state.open ? 'auto' : 'none',
          ...props.style,
        }}
      />
    )}
  >
    <ScrollLock>
      <Dialog.Popup />
    </ScrollLock>
  </Dialog.Viewport>
</Dialog.Portal>
```

rikbrown:
Thanks very much for this. It's crazy the hacks we've had to go through to make a usable dialog box on iOS, thanks Safari!. Hoping Base UI merge this or a similar fix soon.

[...]

PR #2886, opened 2025-10-02
[docs] Fix iOS 26+ backdrops

atomiks:

Third attempt at fixing this...

This one instead only targets iOS Safari. Caveat: if other mobile browsers start allowing content to be seen beneath the viewport, it'll fail for them. 

### Second alternative

To avoid the caveat, we'd need to target the `data-base-ui-scroll-locked` attr placed on the `<html>` tag, which is **only** applied if there are inset scrollbars. The scroll lock for inset scrollbars is special and breaks `position: absolute` backdrops because it sets a fixed height on the body.

This one avoids breaking the inset scrollbar scroll lock, while allowing non-iOS mobile browser UIs to also work if they change. Presumably, browsers with inset scrollbars will never face this issue given it's a mobile/overlay-scrollbar type of UI paradigm 

```css
.Backdrop {
  position: absolute;
  min-height: 100dvh;
  inset: 0;

  [data-base-ui-scroll-locked] & {
    position: fixed;
  }
}
```

This could be renamed to something different, or add an attribute on the `.Backdrop` element itself to avoid the parent > child selector (which is quite awful in Tailwind)

### DIFF
--- a/docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
+++ b/docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
@@ -13,7 +13,9 @@ npm i @base-ui-components/react
 
 All components are included in a single package. Base UI is tree-shakeable, so your app bundle will contain only the components that you actually use.
 
-## Set up portals
+## Set up
+
+### Portals
 
 Base UI uses portals for components that render popups, such as Dialog and Popover.
 To make portalled components always appear on top of the entire page, add the following style to your application layout root:
@@ -35,6 +37,16 @@ To make portalled components always appear on top of the entire page, add the fo
 
 This style creates a separate stacking context for your application's `.root` element.
 This way, popups will always appear above the page contents, and any `z-index` property in your styles won't interfere with them.
+
+### iOS 26+ Safari
+
+Starting in iOS 26, Safari allows content beneath the UI chrome to be visible. Backdrops such as those used by dialogs must use `position: absolute` instead of `position: fixed` to cover the entire visual viewport. For this to work after the page was scrolled, the following style must be added to your global styles:
+
+```css title="styles.css"
+body {
+  position: relative;
+}
+```
 
 ## Assemble a component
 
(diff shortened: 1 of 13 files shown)


docs/src/app/(public)/(content)/react/overview/quick-start/page.mdx
```diff
@@ -36,6 +38,16 @@ To make portalled components always appear on top of the entire page, add the fo
 This style creates a separate stacking context for your application's `.root` element.
 This way, popups will always appear above the page contents, and any `z-index` property in your styles won't interfere with them.
 
+### iOS 26+ Safari
+
+Starting in iOS 26, Safari allows content beneath the UI chrome to be visible. Backdrops such as those used by dialogs must use `position: absolute` instead of `position: fixed` to cover the entire visual viewport. For this to work after the page was scrolled, the following style must be added to your global styles:
```
LukasTy:
```suggestion
Starting with iOS 26, Safari allows content beneath the UI chrome to be visible. Backdrops such as those used by dialogs must use `position: absolute` instead of `position: fixed` to cover the entire visual viewport. For this to work after the page was scrolled, the following style must be added to your global styles:
```

LukasTy (approved):
Looks great.
Tested on simulator:
Image

This solution strikes a great balance of not impacting anyone to whom this change is irrelevant (at least for now).
Tested on iOS 18, doesn't seem to visually change anything as well. 👌

chrisneven:
When is this going to be released?

@atomiks

atomiks:
@chrisneven this PR wasn't a Base UI source code change, it's a docs-only change and it was already released in beta.4. Your code is responsible for the styles here

chrisneven:
Ahhh thank you for your respone @atomiks

Seems there's still something off with our app (on beta.4) as well as the base ui documentation website screen recording:
https://github.com/user-attachments/assets/a9199065-76a9-4a85-9dd1-a181c60b9cfa

Do you want me to create an issue for it? Thanks in advance

atomiks:
By "off" do you mean the background scrolling? If so, separate issue: https://github.com/mui/base-ui/issues/1893

chrisneven:
@atomiks, yes this is what I meant. Thanks, will keep an eye on it

PR #2889, opened 2025-10-02
[portals] Create portal elements inside React

### LINKED ISSUE #2780: [toast] Portal doesn't respect provided `container`

### LINKED ISSUE #2887: Access the Portal part node
# Feature request

I need to coordinate how the positioning of multiple portalled elements works, but there is no direct access to the `[data-base-ui-portal]` nodes to even set a CSS class name on them.

## Summary

Forward props and expose a ref to the portal node like with other parts

- Radix: doesn't render a separate portal node so N/A
- Ariakit: can pass your own element to `[portalElement](https://ariakit.org/reference/dialog#portalelement)`

atomiks:
Built off #2847

This makes the portal element (with the internal `id`) a regular element rendered inside React. So you can pass props to it, access it with a ref, etc. The `container` prop is still used as a location to inject it.

Closes #2887
Fixes #2780 (https://codesandbox.io/p/sandbox/kind-moon-fstqzf)


docs/src/components/Demo/DemoLoader.tsx
```diff
@@ -254,6 +254,21 @@ function resolveExtensionlessFile(filePath: string, preferTs: boolean): string {
     ? ['.tsx', '.ts', '.jsx', '.js', '.json']
     : ['.jsx', '.js', '.tsx', '.ts', '.json'];
 
+  if (existsSync(filePath)) {
```
michaldudak, line 257: This doesn't seem related to the rest of the PR.

atomiks: It broke without this change due to the way imports became structured, can't really remember why exactly, but `useAnchorPositioning` experiments route in particular was broken

PR #2890, opened 2025-10-02
[code-infra] Cleanup unused dev dependencies

brijeshb42:
Also fixed the prettier config to include tailwindcss plugin which got removed during this code-infra [migration](https://github.com/mui/base-ui/pull/2054).


Janpot (approved):

michaldudak:
Honestly, I prefer it without the Tailwind plugin. It forces all the classes to be on a single line and disallows to group them and arrange as I'd like to, making the code unreadable (see https://github.com/mui/base-ui/pull/2336/commits/f7a7d26a8f973004ac28d83ff7775b8898026b74#diff-96601bb7d2ee2db985a331d8c709158d96e9fce66fb92b32e5f869c881a42daa).

EDIT: I see it's possible to configure whitespace preservation - IMO we should: https://github.com/tailwindlabs/prettier-plugin-tailwindcss?tab=readme-ov-file#preserving-whitespace

brijeshb42:
If everyone's onboard, we can just remove the plugin. I tried formatting the file (that you referred) with the option to preserve whitespace and it didn't work. I think maybe because its for whitespace and not for newlines.
As for the changes in the PR itself, none of the changed files had the classes added with newlines so it didn't raise any major flags.

brijeshb42:
Read this [comment](https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/39) from the plugin author and they won't support multiline classes.

So we can just remove the plugin and revert the code back to before this PR for the class name changes.

PR #2893, opened 2025-10-02
[combobox] Take into account `isItemEqualToValue` when selecting an option in multiple mode

### LINKED ISSUE #2892: [combobox] Selection does not take into account `isItemEqualToValue` and `multiple` is `true`
# Bug report

## Current behavior

Selecting the same option in a Combobox with `multiple:  true` appends said option to the current selected values, thus creating duplicates when you use objects which do not maintain referential integrity (e.g `Proxy` values through `valtio`) which I think is okay, but when using the `isItemEqualToValue` property, the behaviour stays the same.

## Expected behavior

Selecting the same option should deselect it if the `isItemEqualToValue` result is evaluated to `true` for that option.

## Reproducible example

[CodeSandbox repro](https://codesandbox.io/p/devbox/cold-fast-y96r3g?file=%2Fsrc%2FApp.tsx&workspaceId=ws_RRJobaGRBhfDBJFVKHJUux)

## Base UI version

v1.0.0-beta.4

Chrome

## Which OS are you using?

Mac OS

N/A.

## Additional context

Going through the code, I suspect [this](https://github.com/mui/base-ui/blob/67811d07c56cbdfb722de431a538ae36e1804cc8/packages/react/src/combobox/root/ComboboxRootInternal.tsx#L717) may be the line which should take into account the function defined through the `isItemEqualToValue` property (currently uses `includes`).

I can open a PR with this change if there isn't any more context that is affected by this.

Thank you!

epr3:
Fixes #2892.

### DIFF
--- a/packages/react/src/combobox/root/ComboboxRoot.test.tsx
+++ b/packages/react/src/combobox/root/ComboboxRoot.test.tsx
@@ -3433,6 +3433,54 @@ describe('<Combobox.Root />', () => {
       );
     });
 
+    it('properly deselects object values using the provided comparator', async () => {
+      const users = [
+        { id: 1, name: 'Alice' },
+        { id: 2, name: 'Bob' },
+      ];
+
+      await render(
+        <Combobox.Root
+          items={users}
+          defaultValue={[{ id: 2, name: 'Bob' }]}
+          itemToStringLabel={(item) => item.name}
+          itemToStringValue={(item) => String(item.id)}
+          isItemEqualToValue={(item, value) => item.id === value.id}
+          defaultOpen
+          multiple
+        >
+          <Combobox.Input data-testid="input" />
+          <span data-testid="value">
+            <Combobox.Value />
+          </span>
+          <Combobox.Portal>
+            <Combobox.Positioner>
+              <Combobox.Popup>
+                <Combobox.List>
+                  {(item) => (
+                    <Combobox.Item key={item.id} value={item}>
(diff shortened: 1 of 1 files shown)


atomiks:
Thanks for the contribution! There's a test suite with `describe('prop: isItemEqualToValue'` for `ComboboxRoot` to test the behavior

It looks like Select correctly handles this in multiple mode, but adding a similar test for it would be good too

packages/react/src/combobox/root/ComboboxRootInternal.tsx
```diff
@@ -714,9 +714,13 @@ export function ComboboxRootInternal<Value = any, Mode extends SelectionMode = '
...
         const currentSelectedValue = Array.isArray(selectedValue) ? selectedValue : [];
-        const isCurrentlySelected = currentSelectedValue.includes(value);
+        const isCurrentlySelected = itemIncludes(
+          currentSelectedValue,
+          value,
+          store.state.isItemEqualToValue,
+        );
         const nextValue = isCurrentlySelected
-          ? currentSelectedValue.filter((v) => v !== value)
+          ? currentSelectedValue.filter((v) => !store.state.isItemEqualToValue(v, value))
```
atomiks: The `removeItem` util is used by Select here

epr3: Fixed

PR #2896, opened 2025-10-02
[checkbox group] Fix parent checkbox ignoring custom `id` prop

### LINKED ISSUE #2691: [checkbox][CheckboxGroup] Checkbox.Root `id` not working in "parent checkbox" use case
# Bug report

## Current behavior

In the "parent checkbox" use case of `CheckboxGroup`, the consumer-specified `id` on `Checkbox.Root` gets ignored. This makes it impossible to associate a label to the checkbox using `htmlFor`.

## Expected behavior

`Checkbox.Root` should honor the `id` passed by the consumer.

## Reproducible example

This is a slightly altered version of the official example, with the "Fuji" checkbox changed to associate the label using `htmlFor`.

https://codesandbox.io/p/sandbox/base-ui-checkbox-nested-bug-yz9lhv

https://github.com/user-attachments/assets/1ae57c08-c473-4f8a-9f3c-34010bd52a02

## Base UI version

1.0.0-beta.3

Chrome

## Which OS are you using?

Mac OS

## Additional context

This only happens in the "parent checkbox" use case. It doesn't happen as soon as I remove the `allValues` prop from `CheckboxGroup`.

edmolima:
Fixes https://github.com/mui/base-ui/issues/2691

## Context

In the "parent checkbox" use case of CheckboxGroup, the consumer-specified `id` on `Checkbox.Root` was being ignored. This made it impossible to associate a label to the checkbox using `htmlFor`.

The issue occurred because `useCheckboxGroupParent` was returning its own generated `id` in `getParentProps()`, which overrode the user-provided `id` prop.

## Summary

- Modified `getParentProps()` to accept an optional `parentId` parameter
- `CheckboxRoot` now passes its`id` to `getParentProps()` for parent checkboxes
- This ensures custom `id` props are honored and `aria-controls` uses the correct id

## Test plan
- Added test to verify custom `id` is respected on parent checkbox
- Added test to verify label association via `htmlFor` works correctly
- All existing tests pass"

### DIFF
--- a/packages/react/src/checkbox-group/useCheckboxGroupParent.ts
+++ b/packages/react/src/checkbox-group/useCheckboxGroupParent.ts
@@ -23,11 +23,10 @@ export function useCheckboxGroupParent(
   const onValueChange = useEventCallback(onValueChangeProp);
 
   const getParentProps: useCheckboxGroupParent.ReturnValue['getParentProps'] = React.useCallback(
-    () => ({
-      id,
+    (parentId?: string) => ({
       indeterminate,
       checked,
-      'aria-controls': allValues.map((v) => `${id}-${v}`).join(' '),
+      'aria-controls': allValues.map((v) => `${parentId ?? id}-${v}`).join(' '),
       onCheckedChange(_, eventDetails) {
         const uncontrolledState = uncontrolledStateRef.current;
 
@@ -114,8 +113,7 @@ export namespace useCheckboxGroupParent {
     id: string | undefined;
     indeterminate: boolean;
     disabledStatesRef: React.RefObject<Map<string, boolean>>;
-    getParentProps: () => {
-      id: string | undefined;
+    getParentProps: (parentId?: string) => {
       indeterminate: boolean;
       checked: boolean;
       'aria-controls': string;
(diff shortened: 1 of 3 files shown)


mj12albert:
@edmolima I tested this PR using the original repro: https://codesandbox.io/p/sandbox/base-ui-checkbox-nested-bug-forked-zkhqtl

A custom `id` on non-parent checkboxes doesn't work still (which the OP's sandbox demonstrates), as clicking the label doesn't toggle the checkbox

edmolima:
Hey @mj12albert! 

Thanks for testing this and catching the issue!

I added a new test, and right now everything looks good.

The playground is here to easily the test again:
https://codesandbox.io/p/sandbox/base-ui-checkbox-nested-bug-forked-4gzlc6

packages/react/src/checkbox-group/useCheckboxGroupParent.ts
```diff
@@ -23,11 +23,10 @@ export function useCheckboxGroupParent(
   const onValueChange = useEventCallback(onValueChangeProp);
 
   const getParentProps: useCheckboxGroupParent.ReturnValue['getParentProps'] = React.useCallback(
-    () => ({
-      id,
+    (parentId?: string) => ({
       indeterminate,
       checked,
-      'aria-controls': allValues.map((v) => `${id}-${v}`).join(' '),
+      'aria-controls': allValues.map((v) => `${parentId ?? id}-${v}`).join(' '),
```
mj12albert, line 29: `aria-controls` here also needs to reference custom child `id`s when present

This is where it gets complicated, I think for this to work all the children need to register their `id`s with the parent, currently the parent is just deriving them from `allValues`

PR #2897, opened 2025-10-03
[combobox][select] Fix `onValueChange` type inference when `value` is unspecified

### LINKED ISSUE #2891: [combobox][typescript] onValueChange value is nullable
# Bug report

## Current behavior

`value` is not correctly typed as `| null`

## Expected behavior

Since `value` is nullable, it should be expressed as such in Typescript

## Base UI version

1.0.0-beta.4

atomiks:

This includes `| null` in the `onValueChange` type if `value` isn't specified as a prop. If `value` is specified it adheres to the generic used.

Fixes #2891 

- Autocomplete: inference comes from `items`: the `Autocomplete.Item` `value` must reference the same shape of an individual item for the inference to be correct in `itemToStringValue`.
- Combobox: inference comes from `value`/`defaultValue`, like Select: the `Combobox.Item` `value` must reference the same shape as the root props for inference to be correct in `itemToStringValue`/`itemToStringLabel`. It's not possible(?) to fallback to `items` for inference in case one of those is missing, so explicit typing is required.

Also cleans up JSDoc types and wrapping logic + renamed `ComboboxRootInternal` to `AriaCombobox` (that is, `role=combobox`) for `Autocomplete` wrapping to be less confusing

### DIFF
--- a/packages/react/src/autocomplete/root/AutocompleteRoot.tsx
+++ b/packages/react/src/autocomplete/root/AutocompleteRoot.tsx
@@ -182,7 +182,7 @@ export namespace AutocompleteRoot {
      */
     value?: AriaCombobox.Props<React.ComponentProps<'input'>['value'], 'none'>['inputValue'];
     /**
-     * Callback fired when the input value of the autocomplete changes.
+     * Event handler called when the input value of the autocomplete changes.
      */
     onValueChange?: (value: string, eventDetails: ChangeEventDetails) => void;
     /**
--- a/packages/react/src/select/root/SelectRoot.tsx
+++ b/packages/react/src/select/root/SelectRoot.tsx
@@ -204,11 +204,11 @@ interface SelectRootProps<Value> {
    */
   multiple?: boolean;
   /**
-   * The value of the select.
+   * The value of the select. Use when controlled.
    */
   value?: Value;
   /**
-   * Callback fired when the value of the select changes. Use when controlled.
+   * Event handler called when the value of the select changes.
    */
   onValueChange?: (value: Value, eventDetails: SelectRoot.ChangeEventDetails) => void;
   /**
@@ -314,7 +314,7 @@ type SelectRootControlledProps<Value, Multiple extends boolean | undefined> = Se
    */
   value: SelectValueType<Value, Multiple>;
   /**
-   * Callback fired when the value of the select changes.
+   * Event handler called when the value of the select changes.
    */
   onValueChange?: (
     value: SelectValueType<Value, Multiple>,
@@ -331,7 +331,7 @@ type SelectRootUncontrolledProps<Value, Multiple extends boolean | undefined> = 
    */
   value?: undefined;
   /**
-   * Callback fired when the value of the select changes.
+   * Event handler called when the value of the select changes.
    */
   onValueChange?: (
     value: SelectValueType<Value, Multiple> | (Multiple extends true ? never : null),
--- a/docs/reference/generated/autocomplete-root.json
+++ b/docs/reference/generated/autocomplete-root.json
@@ -19,7 +19,7 @@
     },
     "onValueChange": {
       "type": "((value: string, eventDetails: Autocomplete.Root.ChangeEventDetails) => void)",
-      "description": "Callback fired when the input value of the autocomplete changes.",
+      "description": "Event handler called when the input value of the autocomplete changes.",
       "detailedType": "| ((\n    value: string,\n    eventDetails: Autocomplete.Root.ChangeEventDetails,\n  ) => void)\n| undefined"
     },
     "defaultOpen": {
--- a/docs/reference/generated/combobox-root.json
+++ b/docs/reference/generated/combobox-root.json
@@ -19,7 +19,7 @@
     },
     "onValueChange": {
       "type": "((value: Value[] | Value, eventDetails: Combobox.Root.ChangeEventDetails) => void) | ((value: Value[] | Value | any | null, eventDetails: Combobox.Root.ChangeEventDetails) => void)",
-      "description": "Callback fired when the selected value of the combobox changes.",
+      "description": "Event handler called when the selected value of the combobox changes.",
       "detailedType": "| ((\n    value: Value[] | Value,\n    eventDetails: Combobox.Root.ChangeEventDetails,\n  ) => void)\n| ((\n    value: Value[] | Value | any | null,\n    eventDetails: Combobox.Root.ChangeEventDetails,\n  ) => void)\n| undefined"
     },
     "defaultOpen": {
--- a/docs/reference/generated/select-root.json
+++ b/docs/reference/generated/select-root.json
@@ -20,7 +20,7 @@
     },
     "onValueChange": {
       "type": "((value: Value[] | Value, eventDetails: Select.Root.ChangeEventDetails) => void) | ((value: Value[] | Value | any | null, eventDetails: Select.Root.ChangeEventDetails) => void)",
-      "description": "Callback fired when the value of the select changes.",
+      "description": "Event handler called when the value of the select changes.",
       "detailedType": "| ((\n    value: Value[] | Value,\n    eventDetails: Select.Root.ChangeEventDetails,\n  ) => void)\n| ((\n    value: Value[] | Value | any | null,\n    eventDetails: Select.Root.ChangeEventDetails,\n  ) => void)\n| undefined"
     },
     "defaultOpen": {
--- a/packages/react/src/combobox/root/ComboboxRoot.tsx
+++ b/packages/react/src/combobox/root/ComboboxRoot.tsx
@@ -121,7 +121,7 @@ type ComboboxRootControlledProps<
    */
   value: ComboboxValueType<Value, Multiple>;
   /**
-   * Callback fired when the selected value of the combobox changes.
+   * Event handler called when the selected value of the combobox changes.
    */
   onValueChange?: (
     value: ComboboxValueType<Value, Multiple>,
@@ -138,8 +138,7 @@ type ComboboxRootUncontrolledProps<
    */
   value?: undefined;
   /**
-   * Callback fired when the selected value of the combobox changes.
-   * Receives `null` when the selection is cleared.
+   * Event handler called when the selected value of the combobox changes.
    */
   onValueChange?: (
     value: ComboboxValueType<Value, Multiple> | (Multiple extends true ? never : null),


packages/react/src/select/root/SelectRoot.tsx
```diff
@@ -17,6 +17,12 @@ import { stringifyAsValue } from '../../utils/resolveValueLabel';
  *
  * Documentation: [Base UI Select](https://base-ui.com/react/components/select)
  */
+export function SelectRoot<Value, Multiple extends boolean | undefined = false>(
+  props: SelectRootControlledProps<Value, Multiple>,
+): React.JSX.Element;
+export function SelectRoot<Value, Multiple extends boolean | undefined = false>(
+  props: SelectRootUncontrolledProps<Value, Multiple>,
+): React.JSX.Element;
```
atomiks, line 25: Wrappers also would need to create overloads to have correct inference for the consumers, but these types seem overly complex to specify on the namespace. Maybe they need to just recreate them locally.

packages/react/src/autocomplete/root/AutocompleteRoot.tsx
```diff
@@ -70,30 +79,29 @@ export function AutocompleteRoot<Value>(props: AutocompleteRoot.Props<Value>): R
...
 
-  // In "both", wrap filtering to use only the typed value, ignoring overlay.
-  let effectiveFilter: typeof other.filter;
-  if (mode !== 'both') {
-    effectiveFilter = staticItems ? null : other.filter;
-  } else {
-    effectiveFilter = (item: any, _query: string, toString?: (item: any) => string) => {
-      return baseFilter(item, query, toString);
+  // In "both", wrap filtering to use only the typed value, ignoring the inline value.
+  const resolvedFilter: typeof other.filter = React.useMemo(() => {
```
LukasTy, line 94: Is `React.memo` worth it here, given that `resolvedQuery` will force its recalculation very often? 🤔

packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1270,17 +1271,16 @@ interface ComboboxRootProps<ItemValue> {
...
    * Filter function used to match items vs input query.
-   * The `itemToStringLabel` function is provided to help convert items to strings for comparison.
    */
   filter?:
     | null
     | ((
         itemValue: ItemValue,
         query: string,
-        itemToStringLabel?: (itemValue: ItemValue) => string,
+        itemToString?: (itemValue: ItemValue) => string,
```
LukasTy, line 1277: Isn't this name change a BC? 🤔

packages/react/src/combobox/root/ComboboxRoot.spec.tsx
```diff
@@ -13,10 +13,54 @@ const objectItemsReadonly = [
...
+    return item.label;
+  }}
+/>;
+
 <Combobox.Root
   items={objectItems}
   defaultValue="a"
   onValueChange={(value) => {
+    // @ts-expect-error
     value.startsWith('a');
```
LukasTy: Maybe we could maintain the previous check? 🤔 
```suggestion
    // @ts-expect-error - possibly null
    value.startsWith('a');
    value?.startsWith('a');
```

packages/react/src/combobox/root/ComboboxRoot.spec.tsx
```diff
@@ -25,6 +69,7 @@ const objectItemsReadonly = [
   items={objectItemsReadonly}
   defaultValue="a"
   onValueChange={(value) => {
+    // @ts-expect-error
     value.startsWith('a');
```
LukasTy:
```suggestion
    // @ts-expect-error - possibly null
    value.startsWith('a');
    value?.startsWith('a');
```

packages/react/src/select/root/SelectRoot.tsx
LukasTy: This is a slight inconvenience. 🙈 
Have you considered alternatives where it wouldn't require consumer intervention? 🤔 
As it stands, we are complicating the code as well as usage only for TS correctness in one case .

LukasTy (commented):
Very nice improvements, nice work. 👍 
However, I have some reservations about the approach and the resulting types/docs. 🤔

packages/react/src/combobox/root/AriaCombobox.tsx
atomiks: Nope, it's just the parameter name that shows up in types. It was confusing because Autocomplete doesn't have this same prop name

packages/react/src/autocomplete/root/AutocompleteRoot.tsx
atomiks: Probably depends on the overall component logic outside, but I think it's worth it here overall to reduce additional subscription re-renders. Root hook calls are less expensive than item ones

packages/react/src/select/root/SelectRoot.tsx
atomiks: To be fair, Combobox has the same overloads already. The overloads seemed necessary to make the type inference work correctly in all the `spec` cases I made, maybe a skill issue on my behalf, or TS limitations. It's a tricky topic 🤔

packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1270,17 +1271,16 @@ interface ComboboxRootProps<ItemValue> {
    * The items to be displayed in the list.
    * Can be either a flat array of items or an array of groups with items.
    */
-  items?: readonly ItemValue[] | readonly Group<ItemValue>[];
+  items?: readonly any[] | readonly Group<any>[];
```
LukasTy, line 1268: Could `any` be avoided here? 🤔 
In general, I feel like the current `"ItemValue[] | Group<ItemValue>[] | undefined"` in docs types is slightly more clear than the suggested `"any[] | Group[] | undefined"`. 🙈

atomiks: `ItemValue` is misleading for `Combobox` since it might not represent the item value. `any[]` is more accurate here

packages/react/src/combobox/root/AriaCombobox.tsx
LukasTy: Oh, I missed that it's not an object parameter. 👌 
As for the name, it looks like it is the same for both `Combobox` and `Autocomplete` on master. 🤔

docs/src/app/(public)/(content)/react/components/autocomplete/page.mdx
```diff
@@ -64,6 +64,10 @@ import { Autocomplete } from '@base-ui-components/react/autocomplete';
   parts="Input, Trigger, Icon, Clear, List, Portal, Backdrop, Positioner, Popup, Arrow, Status, Empty, Collection, Row, Item, Group, GroupLabel, Separator"
 />
 
+## TypeScript inference
+
+Autocomplete infers the item type from the `items` prop passed to `Autocomplete.Root`. If using `itemToStringLabel`, `Autocomplete.Item`'s `value` prop must match the type of an item in the `items` array.
```
LukasTy:
```suggestion
Autocomplete infers the item type from the `items` prop passed to `<Autocomplete.Root>`.
If using `itemToStringLabel`, `<Autocomplete.Item>`'s `value` prop must match the type of an item in the `items` array.
```

docs/src/app/(public)/(content)/react/components/combobox/page.mdx
```diff
@@ -117,6 +117,10 @@ Accepts all `Intl.CollatorOptions`, plus the following options:
   }}
 />
 
+## TypeScript inference
+
+Combobox infers the item type from the `defaultValue` or `value` props passed to `Combobox.Root`. When using `itemToStringLabel` or `itemToStringValue`, `Combobox.Item`'s `value` prop must match the type passed to `defaultValue` or `value`. When not specifying value props, the parameter's type must be explicitly specified.
```
LukasTy, line 122:
```suggestion
Combobox infers the item type from the `defaultValue` or `value` props passed to `<Combobox.Root>`.
When using `itemToStringLabel` or `itemToStringValue`, `<Combobox.Item>`'s `value` prop must match the type passed to `defaultValue` or `value`.
When value props are not specified, the parameter's type must be explicitly specified.
```

packages/react/src/autocomplete/root/AutocompleteRoot.tsx
LukasTy: I've added a console.log in this memo and tried searching/using Autocomplete examples.
It looks like basically every keystroke results in rerunning this memo.
Given Rom's position, he would push towards replacing this with a simple function if it reads nicer with one, because such inefficient memos are worse for performance. 🤔

packages/react/src/combobox/root/AriaCombobox.tsx
LukasTy: Ok, makes sense.
But maybe the docs could resolve it more precisely?
Not the `<any>` is excluded from the docs. 🙈

packages/react/src/select/root/SelectRoot.tsx
LukasTy: Ok.
Then maybe it could make sense to add documentation about this fact if majority of our users are going to wrap these components? 🤔

docs/src/app/(public)/(content)/react/components/autocomplete/page.mdx
atomiks: I agree this is better, but I think we should do a mass-update in a different PR as most of our pages aren't using this format yet

packages/react/src/combobox/root/AriaCombobox.tsx
atomiks: Since it's reused across both, it has to be less descriptive so it can suit both components

packages/react/src/combobox/root/AriaCombobox.tsx
atomiks: That's an issue for the API docs generator, not much we can do here atm

packages/react/src/autocomplete/root/AutocompleteRoot.tsx
atomiks: I was more so thinking about how often the parent re-renders `Autocomplete.Root` since in that case it also passes an unmemoized function down. For keystrokes it doesn't matter yeah. I feel like the `filter` function should always be memoized where possible; I think this one is low-impact

packages/react/src/select/root/SelectRoot.tsx
atomiks: Yes, possibly. A few of our other components do this, so I think it's probably something better for a new PR. Maybe our TypeScript page needs an update.

docs/src/app/(public)/(content)/react/components/autocomplete/page.mdx
LukasTy: Sure, makes sense. 👌

packages/react/src/autocomplete/root/AutocompleteRoot.tsx
LukasTy: Oh, gotcha. 👌 
We can keep the memo to improve that case. 👍

LukasTy (approved):
LGTM. 👍 
Some improvements can be done as follow-ups and are not directly related to this PR. 👌

packages/react/src/select/root/SelectRoot.tsx
songkeys: I bumped my head into this while upgrading from beta.4 to beta.5. This pr is not in the breaking changes. And no docs about this. Might be confusing?

SCR-20251118-gngv

atomiks: @songkeys yeah definitely, it's unfortunately a complex problem (from what I see) - I opened #3252

atomiks: Update: we may remove the overloads to simplify this use case, with a small inconvenience for controlled values: https://github.com/mui/base-ui/pull/3254

PR #2912, opened 2025-10-03
[all components] Fix type portability

### LINKED ISSUE #2629: Autocomplete.useFilter bundle dts generation error
when re-exporting useFilter from Autocomplete component for a ui package, the bundling(tsup/tsdown) process emits this error

this is the line that produces the error: export const {useFilter} = Autocomplete;

error TS2742: The inferred type of 'us
eFilter' cannot be named without a reference to '../../../../node_modules/@base
-ui-components/react/esm/combobox/root/utils/useFilter'. This is likely not por
table.

### LINKED ISSUE #2881: Types are not portable
# Bug report

## Current behavior

When building Storybook documentation for components that wrap Base UI Popover primitives, TypeScript reports that certain Base UI types are not portable. This prevents Storybook from properly generating documentation and causes build failures.

Specifically, when extracting prop types from Base UI components using indexed access (e.g., `BasePopover.Positioner.Props["sideOffset"]`), the TypeScript compiler reports these types as non-portable during Storybook's build process.

Error example:
```
Type 'BasePopover.Positioner.Props["sideOffset"]' is not portable
```

This occurs in wrapper components that attempt to re-export or extend Base UI component props:

```typescript
export interface PopoverContentProps extends React.ComponentProps<typeof BasePopover.Popup> {
  sideOffset?: BasePopover.Positioner.Props["sideOffset"];  // ❌ Causes "not portable" error
}
```

## Expected behavior

Base UI component prop types should be fully portable and serializable so they can be:
1. Extracted using TypeScript utility types like `ComponentProps<typeof Component>`
2. Accessed using indexed types like `Component.Props["propName"]`
3. Re-exported in wrapper component interfaces
4. Properly processed by documentation tools like Storybook with autodocs

The types should integrate seamlessly with TypeScript's type system without triggering portability warnings during build processes.

## Reproducible example

https://github.com/makeplane/plane/tree/preview/packages/propel/src/popover

See `root.tsx` line 8 where we attempt to extract the `sideOffset` type from Base UI's Positioner component.

## Base UI version

v1.0.0-beta.2

N/A - This is a TypeScript/tooling issue that occurs at build time

## Which OS are you using?

Linux (though this is a TypeScript type system issue affecting all platforms)

N/A

## Additional context

This issue significantly impacts the developer experience when building component libraries that wrap Base UI primitives. Storybook is an industry-standard tool for component documentation, and non-portable types break the autodocs feature.

**Current workaround:** Manually redefining all prop types instead of extracting them from Base UI:
```typescript
sideOffset?: number;  // Manual definition instead of extracting from Base UI
```

**Problems with workaround:**
- Creates maintenance overhead (types must be manually kept in sync)
- Loses type safety benefits from Base UI's source of truth
- Risk of type drift over time as Base UI updates

**Impact:** This affects any team building a design system or component library on top of Base UI that uses Storybook for documentation.

**Related tools:**
- @storybook/react-vite
- TypeScript 5.x
- Vite bundler

Would appreciate guidance on whether this is an issue with how Base UI exports its types, or if there's a recommended pattern for extracting and re-exporting Base UI types that maintains portability.

atomiks:
Fixes #2881
Fixes #2629

@Janpot, I created a few test files in `test/public-types` to ensure the types are portable. Maybe it should run in CI?

With this change, types are visible in Intellisense:

Screenshot 2025-10-07 at 11 12 32 pm

The main import always comes first so pressing <kbd>Enter</kbd> will still be correct, at which point you can ignore the top-level types and just use namespaces

Screenshot 2025-10-07 at 11 13 29 pm

### DIFF
--- a/test/public-types/index.tsx
+++ b/test/public-types/index.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Menu } from '@base-ui-components/react/menu';
 import { Toast } from '@base-ui-components/react/toast';
 import type { CheckboxProps } from './checkbox';
+import type { SeparatorProps } from './separator';
 import type { SimpleAutocompleteProps } from './autocomplete';
 import { AutocompleteHarness } from './autocomplete';
 import type { SimpleComboboxProps } from './combobox';
@@ -23,6 +24,10 @@ import type {
 
 export const MyCheckbox = React.forwardRef<HTMLDivElement, CheckboxProps>(() => {
   return <div />;
+});
+
+export const MySeparator = React.forwardRef<HTMLHRElement, SeparatorProps>(() => {
+  return <hr />;
 });
 
 export const SimpleCombobox = React.forwardRef<HTMLInputElement, SimpleComboboxProps>(
--- a/test/public-types/separator.tsx
+++ b/test/public-types/separator.tsx
@@ -1,0 +1,6 @@
+import { Separator } from '@base-ui-components/react/separator';
+
+export type SeparatorProps = Omit<Separator.Props, 'children'> & {
+  // Add an optional className to mirror original example's augmentation pattern.
+  className?: string;
+};


test/public-types/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+  ToastManagerAddOptions,
+  ToastManagerPromiseOptions,
+  ToastManagerReturn,
+  ToastManagerUpdateOptions,
+  ToastProviderProps,
+} from './toast';
+
+export const MyCheckbox = React.forwardRef<HTMLDivElement, CheckboxProps>(() => {
+  return <div />;
+});
```
atomiks, line 27: The main public test for type portability

packages/react/src/dialog/index.ts
```diff
@@ -1 +1,10 @@
 export * as Dialog from './index.parts';
+
+export type * from './root/DialogRoot';
+export type * from './trigger/DialogTrigger';
+export type * from './portal/DialogPortal';
+export type * from './popup/DialogPopup';
+export type * from './backdrop/DialogBackdrop';
+export type * from './title/DialogTitle';
+export type * from './description/DialogDescription';
+export type * from './close/DialogClose';
```
atomiks, line 10: Makes types portable (re-exports top-level types). Namespaces themselves re-export those top-level types so there are no breaking changes

Janpot:
> Maybe it should run in CI?

Yes absolutely. That's the reason I named the script `release:test` (as opposed to test) so it could be set up as a command that runs on all workspaces that runs after `release:build` has run. the `test` command doesn't necessarily depend on production build.

.github/workflows/ci.yml
```diff
@@ -36,6 +36,7 @@ jobs:
       - run: pnpm install
       - run: pnpm release:build
       - run: pnpm validate-declarations
+      - run: pnpm -F ./test/public-types release:test
```
Janpot: 🤔 This and `validate-declarations` kind of feels more at place in [circleci](https://github.com/mui/base-ui/blob/83b8292485e5ee6864a677ce96a6a374ffdb563c/.circleci/config.yml#L298-L319).

Not that it matters that much

.circleci/config.yml
```diff
@@ -304,6 +304,12 @@ jobs:
       - run:
           name: Build packages
           command: pnpm release:build
+      - run:
+          name: Validate type declarations
+          command: pnpm validate-declarations
+      - run:
+          name: Check public types
+          command: pnpm -F ./test/public-types release:test
```
Janpot: Suggestion: We could give control over orchestrating the task to pnpm and fold more workspaces into this as we see fit. (Thinking tests in different workspaces with different module resolution settings)

```suggestion
          command: pnpm -r run release:test
```

Janpot (commented):
Only sampled a few of the library files for review, but seems solid.

packages/react/src/accordion/item/AccordionItem.tsx
```diff
@@ -87,8 +87,9 @@ export const AccordionItem = React.forwardRef(function AccordionItem(
       open: collapsible.open,
       disabled: collapsible.disabled,
       hidden: !collapsible.mounted,
+      transitionStatus: collapsible.transitionStatus,
```
mnajdova, line 90: Can we extract this in a dedicated PR, is not related to the types changes.

packages/react/src/accordion/item/AccordionItem.tsx
```diff
@@ -139,37 +140,42 @@ export const AccordionItem = React.forwardRef(function AccordionItem(
...
+   * open state.
+   * @example
+   * ```tsx
+   * <Accordion.Root value={['a']}>
+   *   <Accordion.Item value="a" /> // initially open
+   *   <Accordion.Item value="b" /> // initially closed
+   * </Accordion.Root>
+   * ```
+   */
+  value?: any;
```
mnajdova, line 163: Can we extract the removal of the AccordionItemValue type in a dedicated PR, it will be hard to track back the change to this PR. Let's scope the PR only to moving types.

packages/react/src/accordion/root/AccordionRoot.tsx
```diff
@@ -161,72 +161,79 @@ export const AccordionRoot = React.forwardRef(function AccordionRoot(
...
+   * Whether to loop keyboard focus back to the first item
+   * when the end of the list is reached while using the arrow keys.
+   * @default true
+   */
+  loop?: boolean;
+  /**
+   * Event handler called when an accordion item is expanded or collapsed.
+   * Provides the new value as an argument.
+   */
+  onValueChange?: (value: AccordionValue, eventDetails: AccordionRoot.ChangeEventDetails) => void;
```
mnajdova:
```suggestion
  onValueChange?: (value: AccordionValue, eventDetails: AccordionRootChangeEventDetails) => void;
```

I am wondering why not using this?

packages/react/src/autocomplete/index.ts
```diff
@@ -1 +1,58 @@
...
+  ComboboxListProps as AutocompleteListProps,
+  ComboboxListState as AutocompleteListState,
+} from '../combobox/list/ComboboxList';
+export type {
+  ComboboxItemProps as AutocompleteItemProps,
+  ComboboxItemState as AutocompleteItemState,
+} from '../combobox/item/ComboboxItem';
+export type {
+  ComboboxValueProps as AutocompleteComboboxValueProps,
+  ComboboxValueState as AutocompleteComboboxValueState,
```
mnajdova:
```suggestion
  ComboboxValueProps as AutocompleteValueProps,
  ComboboxValueState as AutocompleteValueState,
```

mnajdova: I also noticed there is no State exported in the corresponding namespace: https://github.com/mui/base-ui/blob/6d09caa4e7404090e62a9293c6139c672e1fbad0/packages/react/src/autocomplete/value/AutocompleteValue.tsx#L32

packages/react/src/autocomplete/index.ts
```diff
@@ -1 +1,58 @@
...
+} from '../combobox/group-label/ComboboxGroupLabel';
+export type {
+  ComboboxEmptyProps as AutocompleteEmptyProps,
+  ComboboxEmptyState as AutocompleteEmptyState,
+} from '../combobox/empty/ComboboxEmpty';
+export type {
+  ComboboxStatusProps as AutocompleteStatusProps,
+  ComboboxStatusState as AutocompleteStatusState,
+} from '../combobox/status/ComboboxStatus';
+export type { ComboboxCollectionProps as AutocompleteCollectionProps } from '../combobox/collection/ComboboxCollection';
```
mnajdova, line 55: We are missing some parts namespaces types, e.g. Separator, don't we need them here too?

packages/react/src/checkbox-group/CheckboxGroup.tsx
```diff
@@ -141,43 +141,49 @@ export const CheckboxGroup = React.forwardRef(function CheckboxGroup(
...
+   * Names of the checkboxes in the group that should be initially ticked.
+   *
+   * To render a controlled checkbox group, use the `value` prop instead.
+   */
+  defaultValue?: string[];
+  /**
+   * Event handler called when a checkbox in the group is ticked or unticked.
+   * Provides the new value as an argument.
+   */
+  onValueChange?: (value: string[], eventDetails: CheckboxGroup.ChangeEventDetails) => void;
```
mnajdova:
```suggestion
  onValueChange?: (value: string[], eventDetails: CheckboxGroupChangeEventDetails) => void;
```

packages/react/src/checkbox/root/CheckboxRoot.tsx
```diff
@@ -320,97 +320,103 @@ export const CheckboxRoot = React.forwardRef(function CheckboxRoot(
...
+  defaultChecked?: boolean;
+  /**
+   * Whether the component should ignore user interaction.
+   * @default false
+   */
+  disabled?: boolean;
+  /**
+   * Event handler called when the checkbox is ticked or unticked.
+   */
+  onCheckedChange?: (checked: boolean, eventDetails: CheckboxRoot.ChangeEventDetails) => void;
```
mnajdova:
```suggestion
  onCheckedChange?: (checked: boolean, eventDetails: CheckboxRootChangeEventDetails) => void;
```

packages/react/src/collapsible/root/CollapsibleRoot.test.tsx
```diff
@@ -166,7 +166,7 @@ describe('<Collapsible.Root />', () => {
   describe('prop: render', () => {
     it('does not render a root element when `null`', async () => {
       const { getByRole, container } = await render(
-        <Collapsible.Root defaultOpen render={null}>
+        <Collapsible.Root defaultOpen render={null as any}>
```
mnajdova: Why did this fail? Is there a breaking changes somewhere with this PR?

packages/react/src/collapsible/root/CollapsibleRoot.tsx
```diff
@@ -41,9 +41,10 @@ export const CollapsibleRoot = React.forwardRef(function CollapsibleRoot(
     () => ({
       open: collapsible.open,
       disabled: collapsible.disabled,
+      hidden: !collapsible.mounted,
```
mnajdova: Let's extract runtime changes if possible.

packages/react/src/collapsible/root/CollapsibleRoot.tsx
```diff
@@ -78,35 +83,46 @@ export const CollapsibleRoot = React.forwardRef(function CollapsibleRoot(
...
+   * Whether the collapsible panel is initially open.
+   *
+   * To render a controlled collapsible, use the `open` prop instead.
+   * @default false
+   */
+  defaultOpen?: boolean;
+  /**
+   * Event handler called when the panel is opened or closed.
+   */
+  onOpenChange?: (open: boolean, eventDetails: CollapsibleRoot.ChangeEventDetails) => void;
```
mnajdova:
```suggestion
  onOpenChange?: (open: boolean, eventDetails: CollapsibleRootChangeEventDetails) => void;
```

packages/react/src/collapsible/root/CollapsibleRoot.tsx
```diff
@@ -78,35 +83,46 @@ export const CollapsibleRoot = React.forwardRef(function CollapsibleRoot(
...
+   * Whether the component should ignore user interaction.
+   * @default false
+   */
+  disabled?: boolean;
+  render?: BaseUIComponentProps<'div', CollapsibleRoot.State>['render'];
+}
+
+export type CollapsibleRootChangeEventReason = 'trigger-press' | 'none';
+export type CollapsibleRootChangeEventDetails =
+  BaseUIChangeEventDetails<CollapsibleRoot.ChangeEventReason>;
```
mnajdova:
```suggestion
  BaseUIChangeEventDetails<CollapsibleRootChangeEventReason>;
```

packages/react/src/accordion/item/AccordionItem.tsx
atomiks: I had to change this as it was interfering with the docs API generation. It doesn't seem like it's necessary as there's no inference used.

packages/react/src/popover/index.ts
```diff
@@ -1 +1,12 @@
...
+export type * from './root/PopoverRoot';
+export type * from './trigger/PopoverTrigger';
+export type * from './portal/PopoverPortal';
+export type * from './positioner/PopoverPositioner';
+export type * from './popup/PopoverPopup';
+export type * from './arrow/PopoverArrow';
+export type * from './backdrop/PopoverBackdrop';
+export type * from './title/PopoverTitle';
+export type * from './description/PopoverDescription';
+export type * from './close/PopoverClose';
```
mnajdova, line 12: PopoverViewport types are missing

packages/react/src/autocomplete/index.ts
atomiks: Separator is already making its types portable separately, so I don't think so

PR #2917, opened 2025-10-04
[internal] Remove unneeded `useEventCallback` usage

atomiks:
After React 19.2 introduced [`useEffectEvent`](https://react.dev/reference/react/useEffectEvent) I decided to investigate why we're using `useEventCallback` (which is a broader version of it without linting restrictions) to see how we could migrate to it or why the React team chose to keep it restricted.

Firstly, I removed most instances where it's truly unnecessary as no memoization is occurring.

Secondly, I found various use cases where it directly matches `useEffectEvent`'s intended purpose — an event handler that's invoked inside an effect. Or in one special case, used to stabilize a function used as a dependency of an effect.

The use cases I found that are broader than the hook's intended purpose:

1. Using a callback ref where specifying the dependencies is unnecessary, especially for complex ref callbacks that don't need reactivity
2. Reusable event handlers that are forwarded through contexts or memoized groups of props where memoization is a must - nothing that's called during render and needs reactivity
3. A tiny DX improvement where it just turns an `undefined` value into a NOOP function

We also have `useLatestRef`, which is a non-function version of making an "untracked" helper. Most other libraries provide this built-in: https://x.com/rickhanlonii/status/1974118493185540588. 

If the React team weren't wary of wholesale turning off reactivity for something (other libraries provide this already), `useEffectEvent` could've been called `useUntracked` to have the broadest set of use cases.

We could create a `useUntracked` helper that handles all non-reactive values simultaneously, merging `useLatestRef` and `useEventCallback` together @romgrk 

```jsx
'use client';
import { useEventCallback } from './useEventCallback';
import { useLatestRef } from './useLatestRef';

type Callback = (...args: any[]) => any;

/**
 * Non-reactive storage for any value.
 * - If `value` is a function, returns a stable callable that always uses the latest implementation.
 * - Otherwise, returns a `RefObject` object whose `.current` is kept in sync with the latest value.
 */
export function useUntracked<T extends Callback>(value: T | undefined | null): T;
export function useUntracked<T>(value: T): { current: T };
export function useUntracked(value: unknown): unknown {
  const callable = useEventCallback(
    (typeof value === 'function' ? value : undefined) as Callback | undefined,
  );
  const latestRef = useLatestRef(value);
  return typeof value === 'function' ? callable : latestRef;
}
```


packages/react/src/composite/list/CompositeList.tsx
```diff
@@ -13,7 +13,9 @@ export type CompositeMetadata<CustomMetadata> = { index?: number | null } & Cust
  * @internal
  */
 export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
-  const { children, elementsRef, labelsRef, onMapChange } = props;
+  const { children, elementsRef, labelsRef, onMapChange: onMapChangeProp } = props;
+
+  const onMapChange = useEventCallback(onMapChangeProp);
```
romgrk, line 18: Why is this change beneficial?

packages/react/src/checkbox-group/CheckboxGroup.tsx
```diff
@@ -75,11 +75,12 @@ export const CheckboxGroup = React.forwardRef(function CheckboxGroup(
...
 
   const controlRef = React.useRef<HTMLButtonElement>(null);
-  const registerControlRef = useEventCallback((element: HTMLButtonElement | null) => {
+
+  const registerControlRef = React.useCallback((element: HTMLButtonElement | null) => {
     if (controlRef.current == null && element != null && !element.hasAttribute(PARENT_CHECKBOX)) {
       controlRef.current = element;
     }
-  });
+  }, []);
```
romgrk, line 83: Why is this changed? `useEventCallback` makes the intent more clear and is more performant.

packages/react/src/utils/useOpenInteractionType.ts
```diff
@@ -22,9 +22,9 @@ export function useOpenInteractionType(open: boolean) {
     },
   );
 
-  const reset = useEventCallback(() => {
+  const reset = React.useCallback(() => {
     setOpenMethod(null);
-  });
+  }, []);
```
romgrk, line 27: Idem, why move to `useCallback`?

romgrk:
> to see how we could migrate to it or why the React team chose to keep it restricted

The React team often makes decisions that align mostly with Meta's internal needs, and they've stopped putting much effort & thought into the external DX (which the react-hooks eslint package is part of). I wouldn't read too much into why they took that decision.

> A useUntracked helper that handles all non-reactive values simultaneously, merging useLatestRef and useEventCallback together

The code example you listed does runtime type detection and calls both `useEventCallback` and `useLatestRef` each time it's used - this is unefficient. Calling react hooks is not a free operation, there is a runtime cost to be paid. And usually, runtime type detection is a code smell. There's two different use-cases, there is no need to merge them in a single hook. At most, we could rename them to highlight their similarities in behavior, but merging them muddles their type signature and return valule (refs & callbacks are fairly different types of values) and affects performance.

packages/react/src/checkbox-group/CheckboxGroup.tsx
atomiks: @michaldudak wrote that `useCallback` can be more performant

https://github.com/mui/base-ui/pull/2336#discussion_r2354908410

packages/react/src/composite/list/CompositeList.tsx
atomiks: It aligns with what the `useEffectEvent` linter wants

atomiks:
@romgrk yeah the overload isn't great, which is why I wrote about `useEffectEvent` being able to return values.

What do you think about renaming `useEventCallback` to `useUntracked` to better indicate it can return values and be used for that case over `useLatestRef`? It will still only return functions, but you can call them inside effects (similar to accessing refs)

```js
const callback = useUntracked(() => {});
const getValue = useUntracked(() => prop);
```

packages/react/src/checkbox-group/CheckboxGroup.tsx
romgrk: It makes sense now that I re-read the source code, `useEventCallback` has to deal with init/render cycles. We could add something like `useUntrackedCallback` (name tbd) as a version of `React.useCallback` for functions without tracked dependencies like this one.

romgrk:
> What do you think about renaming `useEventCallback` to `useUntracked`

I think that name is confusing. The "callback" token is fairly important semantically. I would instead suggest `useLatestCallback`, though I'm not sure renaming it is worth it.

atomiks:
@romgrk it's the term used to turn off reactivity in every other library (Svelte, Solid, Preact, Angular, Vue, ...). 
Here's a big thread with them all: https://x.com/rickhanlonii/status/1974118493185540588

It's just that returning values from `useEffectEvent` (or `useEventCallback`) seems weird as they seem like they should just fire side-effects alone. I was never a big fan of the term "latest", though.

> The "callback" token is fairly important semantically

Somewhat, but `useEffectEvent` returns a function even though an "event" is typically an object, so I guess it's not critical. I could live with `useUntrackedCallback` but I like shorter names (and since it only ever accepts a function as a param, it seems unnecessary as the return value matches the input. It's intuitive that calling `useUntracked` will untrack the function passed).

packages/react/src/switch/root/SwitchRoot.tsx
```diff
@@ -124,51 +124,37 @@ export const SwitchRoot = React.forwardRef(function SwitchRoot(
...
-      'aria-labelledby': labelId,
-      onFocus() {
-        setFocused(true);
-      },
-      onBlur() {
-        const element = inputRef.current;
-        if (!element) {
-          return;
-        }
+  const rootProps: React.ComponentPropsWithRef<'button'> = {
```
mnajdova, line 127: Why remove the memoization here? Btw, we have the same pattern in the `RadioRoot` component, and we keep the memoization there.

packages/react/src/slider/control/SliderControl.tsx
```diff
@@ -259,7 +258,7 @@ export const SliderControl = React.forwardRef(function SliderControl(
     }
   });
 
-  const handleTouchEnd = useEventCallback((nativeEvent: TouchEvent | PointerEvent) => {
+  function handleTouchEnd(nativeEvent: TouchEvent | PointerEvent) {
```
mnajdova, line 261: Why is this changed?

mnajdova (commented):
Can we document the `useEffectEvent` and explain when it should be used? To have at least some kind of guidelines for other contirbutors.

atomiks:
> Can we document the useEffectEvent and explain when it should be used? To have at least some kind of guidelines for other contirbutors.

I was going to do this in a new PR. This one doesn't create any new hooks/utils yet, just cleans up the code.

packages/react/src/slider/control/SliderControl.tsx
atomiks: It's not necessary, it's not passed as a dependency anywhere.

packages/react/src/switch/root/SwitchRoot.tsx
atomiks: There's no memoization needed, it's not used as a dependency anywhere

> Btw, we have the same pattern in the RadioRoot component

Good spot, that one can be changed as well

mnajdova (approved):
Changes make sense 👌 I haven't check the whole codebase to find other occurences, but if there is anything missed we can potentially fix those afterwards.

PR #2925, opened 2025-10-07
[select][combobox] Fix required form submission with multiple values

### LINKED ISSUE #2913: [combobox] cant submit a form with required and multiple prop
# Bug report

## Current behavior

If used in a form and when using both `required` and `multiple`: `<Combobox.Root required name="language" items={langs} multiple>` it is not possible to submit a form. 

It seems that the required validation is checking wether the input has a value and not if the Combobox itself has a value.

## Expected behavior

Should be able to submit a form when there are selected values in the combobox.

## Reproducible example

https://codesandbox.io/p/sandbox/keen-boyd-5h8x69

## Base UI version

1.0.0-beta.4

Chrome

## Which OS are you using?

Mac OS 15.5

atomiks:
Since `multiple` modes on each use separate hidden inputs, the main hidden input needs to remove `required` when a value exists

Fixes #2913 (https://codesandbox.io/p/sandbox/ecstatic-fire-nzxvy9)

### DIFF
--- a/packages/react/src/combobox/root/ComboboxRoot.test.tsx
+++ b/packages/react/src/combobox/root/ComboboxRoot.test.tsx
@@ -1031,6 +1031,32 @@ describe('<Combobox.Root />', () => {
     });
   });
 
+  describe('prop: required', () => {
+    it('does not mark the hidden input as required when selection exists in multiple mode', async () => {
+      const { container } = await render(
+        <Combobox.Root multiple required name="languages" value={['a']}>
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      const hiddenInput = container.querySelector<HTMLInputElement>('input[aria-hidden="true"]');
+      expect(hiddenInput).not.to.equal(null);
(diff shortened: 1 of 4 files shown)


packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -1031,6 +1031,32 @@ describe('<Combobox.Root />', () => {
...
 
+  describe('prop: required', () => {
+    it('does not mark the hidden input as required when selection exists in multiple mode', async () => {
+      const { container } = await render(
+        <Combobox.Root multiple required name="languages" value={['a']}>
+          <Combobox.Input />
+        </Combobox.Root>,
+      );
+
+      const hiddenInput = container.querySelector<HTMLInputElement>('input[aria-hidden="true"]');
```
LukasTy: Nitpick: Have you tried `screen.getByRole('textbox', { hidden: true })` instead? 🤔

LukasTy (approved):
LGTM. 👌

PR #2926, opened 2025-10-07
[combobox][autocomplete] Add empty and side styling attributes on `Input`

### LINKED ISSUE #2921: [autocomplete][combobox] Expose  `[data-popup-side]` attribute on `Input`
# Feature request

## Summary

 Currently, the Combobox input does not expose any data-[side] property.
Because of this, there's no straightforward way to style the input based on the popup's position (top, bottom, etc.).

If Base UI could expose a property like data-[popup-side]="top" on the input element, it would allow developers to conditionally style the input depending on where the popup opens.
## Motivation

I’m trying to create a seamless visual merge between the input and the popup — with no visible gap.
To achieve this, I remove the input’s bottom border and bottom border-radius when the popup opens below.

However, when the popup opens above the input, this styling breaks.
With data-[popup-side], I could easily handle both cases:

atomiks:
Closes #2921


michaldudak:
In the async search demo, the new attributes are present on the input even after the popup closes.

packages/react/src/combobox/store.ts
```diff
@@ -36,6 +37,7 @@ export type State = {
   listElement: HTMLElement | null;
   triggerElement: HTMLElement | null;
   inputElement: HTMLInputElement | null;
+  popupSide: Side | '';
```
romgrk: The empty value for all other fields is `null`, including for other `string` derived types like `openMethod` right below. It would be more consistent to use the same empty value here.

atomiks: That's a private util though, I'm not sure if there's a public type that uses the same `null`. 

I thought keeping the type the same as  string is slightly more convenient, though this is rarely used to begin with as the data attr is more typical

`keepMounted` popups actually default to `"bottom"` when they aren't open/not positioned yet, so it's sort of consistent from that perspective

romgrk: Should it have a default instead of being nullable then? But otherwise the inconsistency is still annoying, imho. If another dev comes in and reads this code, having all of those as `X | null` is simple and consistent - if I'd come into this as another dev, I'd wonder why the empty string is part of that type.

atomiks: It can't have a default since if the popup isn't mounted, it doesn't make sense for it to have a side. I've changed to `null` after thinking about it more given it's linked to a null element.

PR #2929, opened 2025-10-07
[toast] Allow `React.ReactNode` for `title`/`description` properties

### LINKED ISSUE #2888: [toast] Allow `title` and `description` to receive `ReactNode` instead of just `string`
# Feature request

## Summary

Currently these properties only allow `string`. If you want to render something like a link inside the title or description then you have to use [custom data](https://base-ui.com/react/components/toast#custom). I don't see a reason why these properties couldn't receive `ReactNode`?

As [discussed on Discord](https://discord.com/channels/1287292451308048406/1423037732212441148).

- https://sonner.emilkowal.ski/toast#api-reference

atomiks:
Closes #2888; already works at runtime

### DIFF
--- a/docs/reference/generated/toast-root.json
+++ b/docs/reference/generated/toast-root.json
@@ -12,7 +12,7 @@
       "type": "Toast.Root.ToastObject<any>",
       "required": true,
       "description": "The toast to render.",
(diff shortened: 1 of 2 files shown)


LukasTy (commented):
WDYT about adding an example using `ReactNode` instead of a `string` to showcase this option?

atomiks:
@LukasTy I don't think that's really necessary, it should be obvious to devs once they need it

LukasTy (approved):

PR #2932, opened 2025-10-07
[docs-infra] Migrate to @mui/internal-docs-infra Type Generation

dav-is:
> [!IMPORTANT]
> This PR is still a work in progress.

Migrates this repo to use `@mui/internal-docs-infra` [`useTypes`](https://github.com/mui/mui-public/blob/davis/add-types-generation/docs/app/docs-infra/functions/load-precomputed-types-meta/page.mdx) (being worked on in https://github.com/mui/mui-public/pull/489). Maintains the existing design and theme.

Preview:

## New Features

- Supports Hot Reloading Type Meta
- Build time caching for type meta per component
- Reduce single page typedoc generation from 2s to 200ms (*90%* decrease)
- Improved Developer Experience for (Cmd/Ctrl)+Clicking imports in VS Code

## Notes

For data attributes and CSS variable names to be detected, they need to be imported somewhere in the actual component file. For example, in `accordion/item/AccordionItem.tsx`, we add the side-effect import statement: `import './AccordionItemDataAttributes';`. The component should be using this import anyway to increase type safety.


mnajdova:
https://deploy-preview-2932--base-ui.netlify.app/react/components/accordion#root.changeeventdetails should open the accordion, not just link to the accordion title.

docs/package.json
```diff
@@ -20,7 +20,7 @@
     "@base-ui/utils": "workspace:*",
     "@mdx-js/loader": "^3.1.1",
     "@mdx-js/react": "^3.1.1",
-    "@mui/internal-docs-infra": "0.4.1-canary.7",
+    "@mui/internal-docs-infra": "https://pkg.pr.new/mui/mui-public/@mui/internal-docs-infra@dcc912a",
```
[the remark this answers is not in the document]
dav-is: This is only temporary

docs/src/app/(docs)/react/components/accordion/page.mdx
```diff
@@ -39,7 +39,27 @@ import { DemoAccordionMultiple } from './demos/multiple';
 
 ## API reference
 
-<Reference component="Accordion" parts="Root, Item, Header, Trigger, Panel" />
+import { TypesAccordion } from './types';
+
+### Root
```
michaldudak, line 44: I see that the types.md file already has full content. Why do we need to add each section separately here?

dav-is: This structure also allows overriding the description (see the [merge props docs page](https://github.com/dav-is/base-ui/blob/9ac821e14138ea2ed1c23821e358c0fcedc45655/docs/src/app/(docs)/react/utils/merge-props/page.mdx?plain=1#L71-L85)) and allows introducing the parts in any section of the page. The heading structure of the page.mdx mirrors that of the rendered page for more predictable linking between pages (e.g. if alert dialog adds a link `[Dialog.Root](../dialog/page.mdx#root)`), the links work in markdown when editing and on the website. It also works naturally with the table of contents generator without an additional rehype plugin.

Eventually, I'd like to add a lint script that automatically adds a link to the types.md file for better DX:

```mdx
## API Reference

import { TypesCheckbox } from './types'

### Root

A custom description.

<TypesCheckbox.Root hideDescription />

[See Types](./types.md#root)
```

The See Types link is removed during build so it doesn't appear on the website.

It also mirrors a similar API of Base UI: to get the types for `<Checkbox.Root />`, you just use `<TypesCheckbox.Root />`

Each part is also typed, so you can't use `<TypesCheckbox.Unknown />` without an error

mnajdova:
@aarongarciah can you check the visual updates here, especially the linking between the types?

michaldudak:
@codex review this

michaldudak:
What's the new workflow? When I create a new component and a new docs page, what files do I have to create manually and what are autogenerated (and how)?

Having this information in the README and AGENTS.md could be helpful for other humans and agents.

> To manually update all the types.md without the cache you can run pnpm `docs:validate --types`

Why does the "validate" command update anything?

dav-is:
> What's the new workflow? When I create a new component and a new docs page, what files do I have to create manually and what are autogenerated (and how)?

Simply create a `types.ts` file next to the respective `page.mdx` file and import it into the page. `types.md` is an autogenerated file. To create a new component:

1. Create the component's page. e.g. `docs/src/app/(docs)/react/components/button/page.mdx`
2. Create the component's types component. e.g.  `docs/src/app/(docs)/react/components/button/types.ts`
```ts
import { Button } from '@base-ui/react/button';
import { createTypes } from 'docs/src/utils/createTypes';

export const TypesButton = createTypes(import.meta.url, Button);
```
3. Import the `TypesButton` component to the page.
```mdx
# Button

This is a button.

## API Reference

import { TypesButton } from './types'

<TypesButton />
```
4. Start the dev server `pnpm docs:dev` and load the new component's page: `http://localhost:3005/react/component/button`. The `types.md` is automatically generated. If you don't want to view the component in the browser before committing, just run `pnpm docs:build` and the file will be created.

If the component has multiple parts, use `createMultipleTypes()`:
```ts
import { Checkbox } from '@base-ui/react/checkbox';
import { createMultipleTypes } from 'docs/src/utils/createTypes';

const { types, AdditionalTypes } = createMultipleTypes(import.meta.url, Checkbox);

export const TypesCheckbox = types;
export const TypesCheckboxAdditional = AdditionalTypes;
```
and add multiple sections to the page:
```mdx
# Checkbox

This is a checkbox.

## API reference

import { TypesCheckbox } from './types';

### Root

<TypesCheckbox.Root />

### Indicator

<TypesCheckbox.Indicator />
```

> Having this information in the README and AGENTS.md could be helpful for other humans and agents.

Good idea. I added some details to both [AGENTS.md](https://github.com/dav-is/base-ui/blob/davis/improved-typegen/AGENTS.md#documentation) and [CONTRIBUTING.md](https://github.com/dav-is/base-ui/blob/davis/improved-typegen/CONTRIBUTING.md#documentation)

> Why does the "validate" command update anything?

The validate command is primarily intended to run in CI to detect and fail whenever a committed file is out of sync. When something is out of sync, the CI job will give you a specific command to run to quickly update only the out of date file. `pnpm docs:validate "(docs)/react/components/button"` will run very quickly. During normal development, it is best to run `pnpm docs:dev` or `pnpm docs:build`, as they benefit from the cache. Today, the build takes a bit of time, but eventually it will be faster than the validate command. The naming is meant to nudge the user towards running a build.

Janpot:
I get 
```
[SocketServer] Server error: Error: listen EINVAL: invalid argument /Users/janpotoms/Projects/base-ui/docs/.next/cache/docs-infra/types-meta-worker/mui-types-meta-worker.sock
    at Server.setupListenHandle [as _listen2] (node:net:1918:21)
    at listenInCluster (node:net:1997:12)
    at Server.listen (node:net:2119:5)
    at file:///Users/janpotoms/Projects/base-ui/node_modules/.pnpm/@mui+internal-docs-infra@https+++pkg.pr.new+@mui+internal-docs-infra@7dff01a_@types+rea_afdbb90227fcc721947b6d9099affdab/node_modules/@mui/internal-docs-infra/esm/pipeline/loadServerTypesMeta/socketServer.js:73:19
    at new Promise (<anonymous>)
    at SocketServer.start (file:///Users/janpotoms/Projects/base-ui/node_modules/.pnpm/@mui+internal-docs-infra@https+++pkg.pr.new+@mui+internal-docs-infra@7dff01a_@types+rea_afdbb90227fcc721947b6d9099affdab/node_modules/@mui/internal-docs-infra/esm/pipeline/loadServerTypesMeta/socketServer.js:72:12)
    at initSocket (file:///Users/janpotoms/Projects/base-ui/node_modules/.pnpm/@mui+internal-docs-infra@https+++pkg.pr.new+@mui+internal-docs-infra@7dff01a_@types+rea_afdbb90227fcc721947b6d9099affdab/node_modules/@mui/internal-docs-infra/esm/pipeline/loadServerTypesMeta/worker.js:24:24)
    at async MessagePort.<anonymous> (file:///Users/janpotoms/Projects/base-ui/node_modules/.pnpm/@mui+internal-docs-infra@https+++pkg.pr.new+@mui+internal-docs-infra@7dff01a_@types+rea_afdbb90227fcc721947b6d9099affdab/node_modules/@mui/internal-docs-infra/esm/pipeline/loadServerTypesMeta/worker.js:53:5) {
  code: 'EINVAL',
  errno: -22,
  syscall: 'listen',
  address: '/Users/janpotoms/Projects/base-ui/docs/.next/cache/docs-infra/types-meta-worker/mui-types-meta-worker.sock',
  port: -1
}
[TypesMetaWorker] Worker error: Error: listen EINVAL: invalid argument /Users/janpotoms/Projects/base-ui/docs/.next/cache/docs-infra/types-meta-worker/mui-types-meta-worker.sock
    at new Promise (<anonymous>) {
  code: 'EINVAL',
  errno: -22,
  syscall: 'listen',
  address: '/Users/janpotoms/Projects/base-ui/docs/.next/cache/docs-infra/types-meta-worker/mui-types-meta-worker.sock',
  port: -1
}
[TypesMetaWorker] Worker stopped with exit code 1
```
running `docs:build` locally. Am I forgetting something?

dav-is:
@Janpot it seems MacOS has a shorter limit for socket paths length (104 bytes) than Linux. We could place the socket in the global tmp directory, or we could try to reduce the length of the path within `.next/`. Any preference to which we do?

We could place it in `.next/di/s/t.sock`(di = docs-infra, s = sockets, t = types) or `.next/docs-infra/types.sock` or `/tmp/{hashOfNextjsProjectDir}`

Edit: I went with `.next/docs-infra/types.sock` as it's 40 bytes shorter than before and that seems to keep us well below the 104 byte limit (in your case 66 bytes instead of 106).

dav-is:
@Janpot ok, the bug should be fixed. Can you `git pull && pnpm i && pnpm docs:build` again?

Janpot:
It builds now locally

packages/react/src/field/control/FieldControl.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-irregular-whitespace */
```
aarongarciah: This ESLint rule has a [`skipComments`](https://eslint.org/docs/latest/rules/no-irregular-whitespace#options) option, maybe we should enable it?

docs/src/components/TypeRef/TypeRef.tsx
```diff
@@ -0,0 +1,164 @@
...
+                              <path
+                                d="M10.3333 3.34539L5.47654 7.71648C4.55842 8.54279 3.36693 9 2.13172 9H0V8H2.13172C3.11989 8 4.07308 7.63423 4.80758 6.97318L9.66437 2.60207C10.0447 2.25979 10.622 2.2598 11.0023 2.60207L15.8591 6.97318C16.5936 7.63423 17.5468 8 18.5349 8H20V9H18.5349C17.2998 9 16.1083 8.54278 15.1901 7.71648L10.3333 3.34539Z"
+                                className="TypeRefInfoArrowStrokeDark"
+                              />
+                            </svg>
+                          </Popover.Arrow>
+                          <span>
+                            Both <code className="TypeRefInfoCode">{typeData.meta.name}</code> and{' '}
+                            <code className="TypeRefInfoCode">{aliases[0]}</code> refer to the same
+                            type. Use the dotted form when{' '}
```
aarongarciah: Is this better? "dotted" made me think about the type names with dotted underline in the UI 😅 

```suggestion
                            type. Use the namespaced form when{' '}
```

docs/src/components/TypeRef/TypeRef.tsx
```diff
@@ -0,0 +1,164 @@
...
+                              />
+                              <path
+                                d="M8.99542 1.85876C9.75604 1.17425 10.9106 1.17422 11.6713 1.85878L16.5281 6.22989C17.0789 6.72568 17.7938 7.00001 18.5349 7.00001L15.89 7L11.0023 2.60207C10.622 2.2598 10.0447 2.2598 9.66436 2.60207L4.77734 7L2.13171 7.00001C2.87284 7.00001 3.58774 6.72568 4.13861 6.22989L8.99542 1.85876Z"
+                                className="TypeRefInfoArrowStroke"
+                              />
+                              <path
+                                d="M10.3333 3.34539L5.47654 7.71648C4.55842 8.54279 3.36693 9 2.13172 9H0V8H2.13172C3.11989 8 4.07308 7.63423 4.80758 6.97318L9.66437 2.60207C10.0447 2.25979 10.622 2.2598 11.0023 2.60207L15.8591 6.97318C16.5936 7.63423 17.5468 8 18.5349 8H20V9H18.5349C17.2998 9 16.1083 8.54278 15.1901 7.71648L10.3333 3.34539Z"
+                                className="TypeRefInfoArrowStrokeDark"
+                              />
+                            </svg>
```
aarongarciah: Looks like we're recreating the arrow in all popup components from the docs (not demos). Could we reuse the same arrow in this file at least? Although they use different class names, they seem to look the same. Can we extract it to a component in this file?

docs/src/components/TypeRef/TypeRef.tsx
```diff
@@ -0,0 +1,164 @@
...
+                          fontSize="10"
+                          fontWeight="500"
+                          fontFamily="inherit"
+                        >
+                          ?
+                        </text>
+                      </svg>
+                    </Popover.Trigger>
+                    <Popover.Portal>
+                      <Popover.Positioner side="right" sideOffset={6}>
```
aarongarciah: The arrow is too close to the edge of the icon trigger. Adding a couple of pixels makes it look better.

```suggestion
                      <Popover.Positioner side="right" sideOffset={8}>
```

before

Image

after

Image

docs/src/components/CodeBlockContent.tsx
```diff
@@ -0,0 +1,16 @@
...
+import * as React from 'react';
+import type { ContentProps } from '@mui/internal-docs-infra/CodeHighlighter/types';
+import { useCode } from '@mui/internal-docs-infra/useCode';
+import * as CodeBlock from './CodeBlock';
+
+export function CodeBlockContent(props: ContentProps<object>) {
+  const code = useCode(props);
+
+  return (
+    <div className="w-full">
```
aarongarciah: We are not using Tailwind anymore outside Tailwind demos. You could add a `bui-w-full` utility class to the `/docs/src/css/utilities/width.css` file.

docs/src/components/TypeRef/TypeRef.tsx
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  return (
+    <Popover.Root open={open} onOpenChange={setOpen}>
+      <Popover.Trigger className={`${className ?? ''} TypeRefTrigger`.trim()}>
+        {children}
+      </Popover.Trigger>
+      <Popover.Portal>
+        <Popover.Positioner sideOffset={8}>
+          <Popover.Popup render={<Popup className="TypeRefPopup" />}>
```
aarongarciah, line 63: We should make the paddings match: top is too small, bottom is too large (apparently caused by the scrollbar container height?). Also, making the scrollbar wider to match the content size would look better.

Image

docs/src/components/TypeRef/TypeRef.tsx
```diff
@@ -0,0 +1,164 @@
...
+              </span>
+              {typeData.meta.type !== 'raw' && (
+                <a href={href} className="TypeRefDocsLink">
+                  Full docs
+                </a>
+              )}
+            </div>
+            <ActiveTypeRefContext.Provider value={nextActiveTypeNames}>
+              {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */}
+              <div className="TypeRefContent" onClick={handleContentClick}>
```
aarongarciah: Is the `onClick` handler still needed? Do you have a type def where it's needed?

aarongarciah:
@dav-is the docs validation step is failing on CI but passing locally, do you know why? https://app.circleci.com/pipelines/github/mui/base-ui/22230/workflows/a87bc974-7e20-4a41-8e1e-46dd72422d0a/jobs/246090

dav-is:
@aarongarciah looks like it is working now, I'll take a look though

docs/src/components/CodeBlockContent.tsx
aarongarciah: removed

packages/react/src/field/control/FieldControl.tsx
aarongarciah: only three occurrences for now, let's wait

docs/src/utils/createTypes.tsx
```diff
@@ -22,13 +22,9 @@ const components: MDXComponents = {
...
 };
 
-const inlineComponents: MDXComponents = {
-  ...mdxComponentsInline,
-  code: (props) => (
-    <Code {...props} data-table-code="" className={clsx(props.className, 'TableCode')} />
-  ),
-  pre: CodeBlock.PreInline,
-};
+function TableCode(props: React.ComponentProps<typeof Code>) {
```
aarongarciah: Looks like this component is the same as `/docs/src/components/TableCode.tsx`?

dav-is: Right, we should import that instead

aarongarciah: Removed

aarongarciah:
@dav-is the `pnpm docs:validate` script and Netlify build are failing on CI, but work locally. The Netlify build looks like a cache issue?

michaldudak:
Can we make the old `pnpm docs:api` an alias for `pnpm docs:validate --types`? It's shorter to write and we're already used to it.
I know that we can run `pnpm run docs:build` instead, but it takes ages to complete, and I'd prefer using a command that just updates all the docs.

dav-is:
> Can we make the old pnpm docs:api an alias for pnpm docs:validate --types?

Ok, done. It also takes roughly the same amount of time to run as the previous `docs:api`

> the pnpm docs:validate script and Netlify build are failing on CI, but work locally. The Netlify build looks like a cache issue?

Should be fixed now

---

I also decreased the time needed to run `pnpm docs:validate --types` by ~35% and as much as 52% decrease in CI

dav-is:
> iirc we decided to revert the Prettier width.

Right, fixed

colmtuite:
@dav-is @aarongarciah We agreed to revert to the horizontal scroll inside the Popover, right?

dav-is:
@colmtuite during the call, I think we had decided that horizontal scroll is useful for this case:

Before:
image

After:
Screenshot From 2026-03-17 10-34-50

Most types do not need horizontal scroll

docs/src/app/(docs)/react/utils/merge-props/page.mdx
```diff
@@ -70,16 +70,16 @@ const merged = mergeProps(
 
 ## API reference
 
-### mergeProps
+import { TypesMergeProps, TypesMergePropsN } from './types';
```
michaldudak, line 73: We're missing the header for `mergeProps`

michaldudak:
The Parameters table for useRender has incorrectly aligned chevrons
image

dav-is:
@michaldudak ok both issues are fixed

michaldudak:
## Codex Review

### Findings

#### 1. 🟡 [Non-blocking] Typedoc migration drops the inline type-detail affordances that master added
Impact: Readers lose the inline way to inspect richer type details in API tables. Non-raw type references now fall back to plain links, and property references no longer open any inline detail UI, which is a noticeable regression from the richer API-reference UX already on master.

Evidence: [docs/src/components/TypeRef/TypeRef.tsx#L67-L73](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/docs/src/components/TypeRef/TypeRef.tsx#L67-L73) returns a plain <a> for every non-raw type, making the ReferenceTable branch at [#L90-L94](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/docs/src/components/TypeRef/TypeRef.tsx#L90-L94) unreachable for component/hook/class types. [docs/src/components/TypePropRef/TypePropRef.tsx#L17-L31](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/docs/src/components/TypePropRef/TypePropRef.tsx#L17-L31) now renders property references as bare anchors only. On master, complex prop/type cells still expose inline detail affordances via [docs/src/components/ReferenceTable/ReferenceAccordion.tsx](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/docs/src/components/ReferenceTable/ReferenceAccordion.tsx) and the removed [docs/src/components/ReferenceTable/ReferenceTableTooltip.tsx](https://github.com/mui/base-ui/blob/master/docs/src/components/ReferenceTable/ReferenceTableTooltip.tsx).

Recommendation: Restore the previous inline popover/tooltip behavior, or at minimum remove the non-raw early return so known component/hook/class types still open the new ReferenceTable popover instead of degrading to plain links.

#### 2. 🟡 [Non-blocking] The new callable reference-table paths have already drifted
Impact: Hook and function API docs are implemented twice and already disagree on how to render a simple return type with detailedType. That makes follow-up fixes easy to miss and increases the chance that hook/function docs quietly diverge again.

Evidence: The hook path in [docs/src/components/ReferenceTable/ReferenceTable.tsx#L61-L145](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/docs/src/components/ReferenceTable/ReferenceTable.tsx#L61-L145) and the function path in [#L150-L229](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/docs/src/components/ReferenceTable/ReferenceTable.tsx#L150-L229) are nearly identical. They have already drifted at the simple return-value block: hooks wrap detailedType in CodeBlock.Root at [#L133-L139](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/docs/src/components/ReferenceTable/ReferenceTable.tsx#L133-L139), while functions render detailedType raw at [#L225-L229](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/docs/src/components/ReferenceTable/ReferenceTable.tsx#L225-L229).

Recommendation: Extract the shared hook/function rendering into one helper, or at least centralize the return-value rendering so the two paths cannot diverge on formatting/behavior.

#### 3. ℹ️ [Informational] The new select test reintroduces React.useLayoutEffect despite repo guidance
Impact: This is not a correctness bug, but it reintroduces a pattern the repo explicitly avoids. Leaving new examples behind makes future patches more likely to keep using the wrong hook.

Evidence: [packages/react/src/select/root/SelectRoot.test.tsx#L1008-L1020](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/packages/react/src/select/root/SelectRoot.test.tsx#L1008-L1020) uses React.useLayoutEffect(...) directly. The repo guidance in [AGENTS.md](https://github.com/mui/base-ui/blob/6de04ee5bfb732cafe9f050b47a54ce1017a5ab1/AGENTS.md) says to use useIsoLayoutEffect instead of React.useLayoutEffect.

Recommendation: Swap that test helper to useIsoLayoutEffect so new code stays aligned with the repo’s required hook convention.

### Confidence: 3/5
Moderate confidence. I reviewed the full branch diff with targeted passes over the handwritten runtime/docs changes, plus independent AGENTS/history/simplification sweeps, but the PR is very large and contains extensive generated-doc churn alongside unrelated commits.

### Notes
The branch carries several commits beyond the docs-infra migration itself, so the findings above focus on issues that are still visible in the full master...head review rather than on generated-file noise.

docs/package.json
```diff
@@ -68,6 +68,7 @@
   "devDependencies": {
     "@mdx-js/mdx": "3.1.1",
     "@tailwindcss/postcss": "4.2.1",
+    "@types/chai": "5.2.3",
```
LukasTy: https://github.com/mui/base-ui/pull/4331 removed this and the `chai` package.
Do we still need them for something?
Or is this a merge conflict mishap?

packages/react/src/calendar/store/SharedCalendarStore.ts
```diff
@@ -258,7 +258,7 @@ export interface SharedCalendarStoreParameters<TValue extends TemporalSupportedV
   readOnly?: boolean | undefined;
   /**
    * Whether the calendar is forcefully marked as invalid.
-   * A calendar can be invalid when the selected date fails validation (i.e., is outside of the allowed `minDate` and `maxDate` range).
+   * A calendar can be invalid when the selected date fails validation (that is, is outside of the allowed `minDate` and `maxDate` range).
```
LukasTy, line 261: Nitpick: Is there a problem with `i.e.,` usage?
Or was this a result of some new linting plugin?

packages/react/src/alert-dialog/index.ts
```diff
@@ -1,3 +1,36 @@
 export * as AlertDialog from './index.parts';
 
 export type * from './root/AlertDialogRoot';
+
+export type {
+  DialogBackdropProps as AlertDialogBackdropProps,
+  DialogBackdropState as AlertDialogBackdropState,
+} from '../dialog/backdrop/DialogBackdrop';
```
LukasTy, line 8: I take it we'll need to ensure such export consistency going forward manually?
WDYT, would adding them in the `index.parts` beside the aliased parts export be cleaner and would help prevent forgetting some export?

packages/react/src/direction-provider/DirectionProvider.tsx
```diff
@@ -1,11 +1,13 @@
+/* eslint-disable no-irregular-whitespace */
+/* TODO: Replace `Base UI` with a non-breaking space version in the types.md */
```
LukasTy: Is this a stale TODO?

packages/react/src/field/control/FieldControl.tsx
LukasTy: I reckon we need it, since every component has the root comment linking to the docs. 🙈 
It's 210 instances. 😂

CONTRIBUTING.md
```diff
@@ -11,3 +11,119 @@ There are, however, a few differences:
 
 - The default GitHub branch might be different.
 - The local documentation site opens on a different port: 3005 instead of 3000.
+
+## Documentation
```
LukasTy, line 15: Nitpick: This introduces duplication that can accidentally stray away.
Have you considered adding this only here and then in `AGENTS.MD` say "Read `CONTRIBUTING.md` before proceeding"?

packages/react/src/utils/createBaseUIEventDetails.ts
```diff
@@ -88,7 +88,7 @@ type BaseUIChangeEventDetail<Reason extends string, CustomProperties extends obj
 export type BaseUIChangeEventDetails<
   Reason extends string,
   CustomProperties extends object = {},
-> = Reason extends string ? BaseUIChangeEventDetail<Reason, CustomProperties> : never;
+> = Reason extends string ? BaseUIChangeEventDetail<Reason, CustomProperties> & {} : never;
```
LukasTy, line 91: Was this addition necessary?
I'd assume we want to have the option to control the "other" props via `CustomProperties`. 🤔

packages/react/src/calendar/store/SharedCalendarStore.ts
aarongarciah: There's a Vale linting rule that prevents using "i.e."

image

LukasTy: Ok, thanks for noting. 🙏 
Since those are not enforced, I completely missed that. 🙈

aarongarciah: iirc CI fails when the `valelint` script fails

LukasTy: Haven't seen it fail.
Ran locally on master - produced no errors... 🙈
There must be something wrong with it, or I'm misunderstanding how and when it should flag errors. 🤔

aarongarciah: ok, looks like it only runs on md/mdx files?

LukasTy: Ah, yes, I missed that part... 🙈
Not sure if it makes sense, I reckon it should run on the source code as well as it is transformed into docs. 🤔

aarongarciah: Wait, now that JSDoc comments gets converted to types.md, `valelint` will also parse JSDoc comments.

LukasTy: Ok, I played around a bit locally with this:
- I tried reverting the changes to the `Base UI` spacing in `DirectionProvider`
-  then running `prettier`, `eslint`, `docs:validate`, and `valelint`
    - Only the last step triggers this issue after transforming JSDoc into markdown.

Could this process be improved, and maybe even outside of this PR, to reduce the overhead of this change?
As it stands, the process is somewhat brittle and annoying. 🙈

Besides, side discovery - everyone needs to purge `.next` folder locally for this PR to work properly, otherwise the results are not great.
Just a heads up that after merging, everyone needs to be notified of this necessity. 👌

packages/react/src/direction-provider/DirectionProvider.tsx
dav-is: Addressed the TODO by adding config option `descriptionReplacements` in `next.config.mjs`

packages/react/src/calendar/store/SharedCalendarStore.ts
dav-is: > I reckon it should run on the source code as well

I think Vale is specifically a markdown linter, so I'm not sure that it can run directly on the source.

It looks like a great advantage of Vale for typedocs is it can catch misspellings and hold us to the same writing standard as the rest of the docs.

> Could this process be improved

If you run `pnpm docs:dev`, open `http://localhost:3005/react/utils/direction-provider`, make the change to `DirectionProvdier`, it will update `types.md` and it will hot-reload your fix live. If `valelint` has a watch mode, it could detect an issue right away.

> then running prettier, eslint

Just to add a note, types.md have prettier applied to it before writing. Prettier for source code can run on-save in your IDE. Eslint rules can be detected in the IDE.

> Besides, side discovery - everyone needs to purge .next folder locally for this PR to work properly, otherwise the results are not great.
Just a heads up that after merging, everyone needs to be notified of this necessity. 👌

After merging I'll put a note in slack along with instructions on resolving any merge conflicts on PRs

packages/react/src/alert-dialog/index.ts
dav-is: When this re-export is missing, the `types.md` show `DialogBackdropState` as the type, so it should be spotted in review.

If a public type like `CheckboxRootChangeEventDetails` isn't re-exported, it won't be able to map the source's type to the namespaced alias `Checkbox.Root.ChangeEventDetails`

In the future, I'd like to add some config to be stricter about expecting namespaced types to be re-exported un-namepsaced.

As for revising the export conventions, I think that is reasonable, but I'd say outside of the scope of this PR.

I also added a note in [`CONTRIBUTING.md` about this](https://github.com/dav-is/base-ui/blob/davis/improved-typegen/CONTRIBUTING.md#namespaces-in-typesmd)

docs/package.json
dav-is: I don't think we  need it, good catch

packages/react/src/utils/createBaseUIEventDetails.ts
dav-is: Yes, this ensures the change event detail always forms a distinct type. 
This is how the majority of types already appear in JSDoc. The only types this fix affects is `CalendarRootValueChangeEventDetails`, `NumberFieldRootChangeEventDetails`, `SliderRootChangeEventDetails`.

Without this change this is what we get in JSDoc when hovering and in the docs:
```tsx
type SliderRootChangeEventDetails =
  | BaseUIChangeEventDetail<'none', SliderRootChangeEventCustomProperties>
  | BaseUIChangeEventDetail<'input-change', SliderRootChangeEventCustomProperties>
  | BaseUIChangeEventDetail<'track-press', SliderRootChangeEventCustomProperties>
  | BaseUIChangeEventDetail<'drag', SliderRootChangeEventCustomProperties>
  | BaseUIChangeEventDetail<'keyboard', SliderRootChangeEventCustomProperties>;
```

With this change the docs will show:
```tsx
type SliderRootChangeEventDetails = (
  | { reason: 'none'; event: Event }
  | { reason: 'input-change'; event: Event | InputEvent }
  | { reason: 'track-press'; event: PointerEvent | MouseEvent | TouchEvent }
  | { reason: 'drag'; event: PointerEvent | TouchEvent }
  | { reason: 'keyboard'; event: KeyboardEvent }
) & {
  /** Cancels Base UI from handling the event. */
  cancel: () => void;
  /** Allows the event to propagate in cases where Base UI will stop the propagation. */
  allowPropagation: () => void;
  /** Indicates whether the event has been canceled. */
  isCanceled: boolean;
  /** Indicates whether the event is allowed to propagate. */
  isPropagationAllowed: boolean;
  /** The element that triggered the event, if applicable. */
  trigger: Element | undefined;
  /** The index of the active thumb at the time of the change. */
  activeThumbIndex: number;
};
```

And JSDoc will show:
```tsx
type SliderRootChangeEventDetails = ({
    reason: "none";
    event: Event;
    cancel: () => void;
    allowPropagation: () => void;
    isCanceled: boolean;
    isPropagationAllowed: boolean;
    trigger: Element | undefined;
} & SliderRootChangeEventCustomProperties) | ({
    reason: "track-press";
    event: PointerEvent | MouseEvent | TouchEvent;
    cancel: () => void;
    allowPropagation: () => void;
    isCanceled: boolean;
    isPropagationAllowed: boolean;
    trigger: Element | undefined;
} & SliderRootChangeEventCustomProperties) | ({
    ...;
} & SliderRootChangeEventCustomProperties) | ({
    ...;
} & SliderRootChangeEventCustomProperties) | ({
    ...;
} & SliderRootChangeEventCustomProperties)
```

dav-is:
> Nitpick: there are some newly added files in docs/reference/generated which should be removed before merging.

Done

> This pins to a PR commit hash from mui/mui-public, not a published version.

I plan on merging the upstream `mui-public` PR when this PR is ready to merge

> This regex is quite broad — it'll match any types.ts file under any app/ directory. If someone creates a types.ts utility file outside the expected pattern, the loader would unexpectedly process it. The turbopack config is more targeted ('./src/app/**/types.ts'), so consider tightening the webpack regex to match

Done

> The --useVisibleDescription flag was removed from the validate command but is still kept in the transformMarkdownMetadata config (useVisibleDescription: true).

`pnpm docs:validate` now uses config from `next.config.mjs`

> just needs clear .gitignore or CI rules to ensure regeneration stays in sync.

`pnpm docs:validate` runs in CI to ensure the files stay in sync.

> The socketDir: '.next/cache/docs-infra/types-meta-worker' path is hardcoded in both webpack and turbopack configs — consider extracting to a constant.

Fixed.

> The ordering import from typeOrder.mjs is only used in the new configs but the file doesn't exist on master yet — confirm it's properly added in the PR.

Migrates the existing ordering.json file.

> Reduce duplication between AGENTS.md and CONTRIBUTING.md documentation sections

Fixed

packages/react/src/utils/createBaseUIEventDetails.ts
LukasTy: Thanks for clarifying. 🙏

packages/react/src/alert-dialog/index.ts
LukasTy: Makes sense. 👍 
The re-exporting can be handled separately and doesn't impact this PR. 👌

packages/react/src/calendar/store/SharedCalendarStore.ts
LukasTy: Ok, the arguments sound logical. 👍 
Let's move forward with this and monitor if it causes any problems.

docs/package.json
```diff
@@ -76,6 +76,7 @@
     "@types/unist": "3.0.3",
     "@wooorm/starry-night": "3.9.0",
     "baseline-browser-mapping": "2.10.0",
+    "chai": "6.2.2",
```
LukasTy: This should also be removed for the same reason as `@types/chai`

LukasTy (approved):
Looks really great. 👍 

A few final Claude Opus follow-up remarks

### Blockers (must fix before merge)

| # | Issue | Status |
|---|-------|--------|
| 1 | **`chai` devDependency re-addition** — Your [prior review comment](https://github.com/mui/base-ui/pull/2932#discussion_r2958668946) flagged that PR #4331 removed both `@types/chai` and `chai`. On this branch, `@types/chai` is removed but `chai: 6.2.2` persists. This is likely a rebase artifact. Note: both currently still exist on master, so PR #4331 may not have been merged yet. Either way, this needs reconciliation. | **OPEN** |

### Issues (should fix)

| # | Issue | Detail |
|---|-------|--------|
| 2 | **Webpack regex too broad** — `test: /[/\\\\]src[/\\\\]app[/\\\\].*[/\\\\]types\.ts$/` matches *any* `types.ts` under `src/app/`. The turbopack config is correctly scoped to `'./src/app/**/types.ts'` (which only runs in the docs context), but the webpack regex could hit unrelated files. Should tighten to include `(docs)` in the path pattern. | Flagged by Copilot review, **not addressed** |
| 3 | **`splice(3)` vs `slice(3)`** — In typedocProcessor.mjs, line 49 uses `.splice(3)` which mutates the source array. While the mutation is harmless here (temp array), `.slice(3)` is more idiomatic and intention-revealing. | Copilot review comment, **partially addressed** (file header fixed, splice not changed) |

### Observations (minor / informational)

| # | Item | Notes |
|---|------|-------|
| 4 | **No AGENTS.md / CONTRIBUTING.md duplication** | Good news: the CONTRIBUTING.md gets new detailed Documentation section, and AGENTS.md merely links to it with "Before any changes, review CONTRIBUTING.md". Your review comment about this was addressed — no content duplication. |
| 5 | **Source file changes are safe** | 609 source files changed in src — these are: (a) JSDoc additions for type extraction (e.g. adding `/** The current value. */` to state interface fields), (b) type re-exports in `index.ts` files for the typedoc pipeline, (c) Chai-to-Vitest assertion migration. No runtime behavior changes detected. |
| 6 | **Old generated fully removed** | All ~245 JSON files deleted, directory empty on PR branch. |
| 7 | **api-docs-builder removed** | `pnpm docs:api` is repurposed as an alias for `docs-infra validate --types`. |
| 8 | **`typeOrder.mjs` and `createTypes.tsx` exist** | Both utility files are properly added and look well-implemented. |
| 9 | **Branch is fully up-to-date** with `upstream/master` (merge base = HEAD of master). No rebase needed. |

---

### Merge Confidence Score: **55/100** (Not Ready)

**Rationale:**

- **Architecture: Excellent.** The migration from JSON-based type generation to colocated `types.ts` + TS Language Service API is a significant DX improvement. Well-designed.
- **Quality: Good.** Test migration is thorough, source changes are non-behavioral, documentation is solid.
- **Completeness: Incomplete.** The `pkg.pr.new` dependency is a hard blocker — this PR cannot ship to CI/production with a transient URL dependency. Until a published `@mui/internal-docs-infra` version is available, the PR is structurally unmerge-able.
- **Risk: Medium.** The sheer size (1,486 files, 449 commits) makes it difficult to review exhaustively. The webpack regex scope issue is a time bomb. The assertion migration, while mechanical, touches thousands of lines.
- **Review Status: Partially approved.** aarongarciah gave conditional approval ("revert Prettier width, then LGTM"). Your review is still pending. michaldudak raised questions that appear addressed.

**To reach merge-ready (~85+):** resolve the `pkg.pr.new` dependency (blocker #1), tighten the webpack regex (#3), and confirm the `chai` dependency situation with respect to PR #4331 (#2).

aarongarciah:
> UX nitpick: personally, I feel like something is missing on these popup triggers. Maybe cursor: help? Or something else other than underline indicating they are interactive?

@LukasTy we might rethink this and open type def popovers on hover, but for now I think it's enough. We'll be reworking the UI in the upcoming weeks so anything is subject to change.

aarongarciah:
> 1. 🟡 [Non-blocking] Typedoc migration drops the inline type-detail affordances that master added

@michaldudak Nothing to act on here. @dav-is said: "I don't think we should remove the early return, that would break the broken links checker".

> 2. 🟡 [Non-blocking] The new callable reference-table paths have already drifted

I pushed a [refactor](https://github.com/mui/base-ui/pull/2932/changes/075f0faafb58573a2453a2a887634bce3ab44fc1) (made by Codex). @dav-is may I ask you to take a look? I don't wanna break any edge cases.

dav-is:
> I pushed a refactor (made by Codex). @dav-is may I ask you to take a look? I don't wanna break any edge cases.

I think the refactor looks correct 👍 

Edit: I cleaned up the function/hook types a bit

colmtuite:
@dav-is We have a late request to switch the popover back to `openOnHover`. Same styles etc., think we just need to add that prop.

aarongarciah:
@dav-is the way `actionsRef`'s type from the `Combobox.Root` was displayed before seems better in a single line and without `|` at the start of each line:

before
Screenshot 2026-03-23 at 12 21 10

after
Screenshot 2026-03-23 at 12 21 01

dav-is:
> switch the popover back to openOnHover

Done

>  the Combobox.Root was displayed before seems better in a single line and without | at the start of each line

Done

PR #2935, opened 2025-10-08
[combobox][autocomplete] Allow `Esc` to bubble if `Empty` is not used

atomiks:
If the `Empty` component is not used, then you're hiding the positioner/popup with CSS (while the popup is open). In this case, pressing <kbd>Esc</kbd> shouldn't block the event bubbling since the popup doesn't appear to be open and a single press should close outer popups.


michaldudak:
IMO the behavior will be less surprising if we close the popup instead of requiring them to hide it with CSS

atomiks:
@michaldudak this can't be done because the input value resets once it closes, which breaks the filtering. I've updated the code to check if `emptyRef.current` exists instead, which is simpler.

PR #2937, opened 2025-10-08
[experiments] Add combobox to react-hook-form demo

mj12albert:


atomiks:
I don't see any options:

Screenshot 2025-10-08 at 10 14 55 pm

Edit: never mind 😅  it wasn't clear it was linked to the field directly above it

docs/src/app/(private)/experiments/forms/rhf.tsx
```diff
@@ -284,18 +299,26 @@ export default function ExampleForm() {
         />
 
         <Controller
-          name="selectCountry"
+          name="country"
```
atomiks, line 302: Can we add Select + Combobox side-by-side? It wasn't clear (even with all the messaging that you need to interact with the other field first)

I guess that doesn't really work on narrow viewports,  but still

docs/src/app/(private)/experiments/forms/rhf.tsx
```diff
@@ -49,23 +51,36 @@ export const settingsMetadata: SettingsMetadata<Settings> = {
...
+        numberField: 5,
+        country: '',
+        states: [],
+        radioGroup: 'auto',
+        checkboxGroup: [],
+      },
+      mode: settings.validationMode,
+    });
+
+  const watchedCountry = watch('country');
```
gilsmt, line 76: `watch()` breaks when used with React compiler https://github.com/react-hook-form/react-hook-form/issues/11910 giving stale values often. `useWatch` is the alternative

mj12albert:
Gonna redo this whole example later with Tailwind to avoid adding to the overhead of the css file here

PR #2941, opened 2025-10-08
[code-infra] Enable test-library rules for the repo

brijeshb42:
For `testing-library/no-container`, I have disabled it selectively since at those places, it was testing internal dom elements. So it made sense. But in general, I have enabled the rule for the majority of the rest of the cases. Open to suggestions.

Relates to https://github.com/mui/mui-public/issues/173

### DIFF
--- a/packages/react/src/checkbox/indicator/CheckboxIndicator.test.tsx
+++ b/packages/react/src/checkbox/indicator/CheckboxIndicator.test.tsx
@@ -34,69 +34,67 @@ describe('<Checkbox.Indicator />', () => {
     },
   }));
 
-  /* eslint-disable testing-library/no-container */
   it('should not render indicator by default', async () => {
-    const { container } = await render(
+    await render(
       <Checkbox.Root>
-        <Checkbox.Indicator />
+        <Checkbox.Indicator data-testid="indicator" />
       </Checkbox.Root>,
     );
-    const indicator = container.querySelector('span');
+    const indicator = screen.queryByTestId('indicator');
     expect(indicator).to.equal(null);
   });
 
   it('should render indicator when checked', async () => {
-    const { container } = await render(
+    await render(
       <Checkbox.Root checked>
-        <Checkbox.Indicator />
+        <Checkbox.Indicator data-testid="indicator" />
       </Checkbox.Root>,
     );
-    const indicator = container.querySelector('span');
+    const indicator = screen.getByTestId('indicator');
     expect(indicator).not.to.equal(null);
   });
 
   it('should spread extra props', async () => {
-    const { container } = await render(
+    await render(
       <Checkbox.Root defaultChecked>
-        <Checkbox.Indicator data-extra-prop="Lorem ipsum" />
+        <Checkbox.Indicator data-testid="indicator" data-extra-prop="Lorem ipsum" />
       </Checkbox.Root>,
     );
-    const indicator = container.querySelector('span');
+    const indicator = screen.getByTestId('indicator');
     expect(indicator).to.have.attribute('data-extra-prop', 'Lorem ipsum');
   });
 
   describe('keepMounted prop', () => {
     it('should keep indicator mounted when unchecked', async () => {
-      const { container } = await render(
+      await render(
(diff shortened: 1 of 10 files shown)


packages/react/src/checkbox/indicator/CheckboxIndicator.test.tsx
```diff
@@ -34,6 +34,7 @@ describe('<Checkbox.Indicator />', () => {
     },
   }));
 
+  /* eslint-disable testing-library/no-container */
```
LukasTy: Nitpick: The fact that we need to rely on `container` could signal a lack of a11y for those certain elements or a need to change the way we test, i.e.: 
```js
expect(screen.getByRole('checkbox', { checked: false }).firstChild).to.equal(null);
```

packages/react/src/checkbox/root/CheckboxRoot.test.tsx
```diff
@@ -39,8 +37,9 @@ describe('<Checkbox.Root />', () => {
   });
 
   it('should change its state when clicked', async () => {
-    const { getAllByRole, container } = await render(<Checkbox.Root />);
-    const [checkbox] = getAllByRole('checkbox');
+    const { container } = await render(<Checkbox.Root />);
+    const [checkbox] = screen.getAllByRole('checkbox');
+    // eslint-disable-next-line testing-library/no-container
     const input = container.querySelector('input[type=checkbox]') as HTMLInputElement;
```
LukasTy: Nitpick: Doesn't it signal that there might be something lacking to elegantly access the relevant elements? 🤔 
The following could be used in the meantime:
```ts
await render(<Checkbox.Root />);
const [checkbox, checkboxInput] = screen.getAllByRole('checkbox');
const input = checkboxInput as HTMLInputElement;
```

packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -351,6 +351,7 @@ describe('<Combobox.Root />', () => {
 
         expect(screen.getByRole('listbox')).not.to.equal(null);
 
+        // eslint-disable-next-line testing-library/no-container
```
LukasTy: Almost all (except 3) use cases seem replaceable with:
```diff
-const hiddenInput = container.querySelector('[name="test"]');
+const hiddenInput = screen.getByRole('textbox', { hidden: true });
```

packages/react/src/select/root/SelectRoot.test.tsx
```diff
@@ -224,6 +224,7 @@ describe('<Select.Root />', () => {
         </Select.Root>,
       );
 
+      // eslint-disable-next-line testing-library/no-container -- No appropriate method on screen since it's a hidden input without any type
       const hiddenInput = container.querySelector('input[name="country"]');
```
LukasTy: Majority of use cases are replaceable with:
```diff
-const hiddenInput = container.querySelector('input[name="country"]');
+const hiddenInput = screen.getByRole('textbox', { hidden: true });
```

packages/react/src/select/value/SelectValue.test.tsx
```diff
@@ -386,6 +386,7 @@ describe('<Select.Value />', () => {
       );
 
       expect(screen.getByTestId('value')).to.have.text('Canada');
+      // eslint-disable-next-line testing-library/no-container -- No appropriate method on screen since it's a hidden input without any type
       const hiddenInput = container.querySelector('input[name="country"]');
```
LukasTy:
```suggestion
const hiddenInput = screen.getByRole('textbox', { hidden: true });
```

brijeshb42:
Took a 2nd pass and I was able to remove a bunch more eslint ignores. The only cases being ignored now are where we are querying `<input type="hidden" value="" />`, since this isn't a `textbox` role as well as is hidden. I couldn't find another way to query it.

brijeshb42:
Merging since Argos has taken up `docs-v1` as the base branch and not `master` and there isn't a way to change that in Argos.

PR #2947, opened 2025-10-09
[docs][combobox] Async search examples

atomiks:
This pattern is used when searching for selectable items without loading them all upfront. In this case, the selected value  needs to be retained in the `items` list at all times if present, as the internal logic removes selections if it disappears from the items list.

Preview: https://deploy-preview-2947--base-ui.netlify.app/react/components/combobox#async-search-single

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/combobox/demos/async-single/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/combobox/demos/async-single/css-modules/index.tsx
@@ -16,6 +16,8 @@ export default function ExampleAsyncSingleCombobox() {
 
   const abortControllerRef = React.useRef<AbortController | null>(null);
 
+  const trimmedSearchValue = searchValue.trim();
+
   const items = React.useMemo(() => {
     if (!selectedValue || searchResults.some((user) => user.id === selectedValue.id)) {
       return searchResults;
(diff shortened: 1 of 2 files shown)


docs/src/app/(public)/(content)/react/components/combobox/page.mdx
```diff
@@ -98,6 +98,22 @@ import { DemoComboboxGrouped } from './demos/grouped';
 
 <DemoComboboxGrouped compact />
 
+### Async search (single)
+
+Load items from a remote source by fetching on input changes. Keep the selected item in the `items` list so it remains available while new results stream in. This pattern avoids needing to load items upfront.
```
mj12albert, line 103: > Keep the selected item in the items list so it remains available while new results stream in

@atomiks This makes sense for multiple but for single selection it doesn't seem that useful? Once a (single) selection is made, you won't get new results without clearing/re-typing a new input value:

https://github.com/user-attachments/assets/2fddbacb-76eb-4b34-9e4e-74dad44fffda

atomiks: It also needs to stay present in the list since otherwise the selection gets removed, and typically that only happens if the entire input is cleared from text (and matches non-async demos)

mj12albert: Ok that makes sense, I guess it's a bit confusing since the description is very similar to the async multiple one

mj12albert (approved):

PR #2948, opened 2025-10-09
[experiments] Add Tanstack Form integration

mj12albert:
https://deploy-preview-2948--base-ui.netlify.app/experiments/forms/tanstack

The validation can get a bit weird since the API and docs are huge and a bit confusing

Some findings:
- TSF doesn't directly support focusing errors: https://tanstack.com/form/latest/docs/framework/react/guides/focus-management
- The definition of "dirty" is different from RHF: https://tanstack.com/form/latest/docs/framework/react/guides/basic-concepts#understanding-isdirty-in-different-libraries

### DIFF
--- a/docs/src/app/(private)/experiments/forms/tanstack.tsx
+++ b/docs/src/app/(private)/experiments/forms/tanstack.tsx
@@ -1,0 +1,125 @@
+/* eslint-disable react/no-children-prop */
+'use client';
+import * as React from 'react';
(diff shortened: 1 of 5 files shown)


docs/src/app/(private)/experiments/forms/tanstack.tsx
```diff
@@ -0,0 +1,1081 @@
+/* eslint-disable react/no-children-prop */
```
mj12albert, line 1: Just following their convention, no idea why they do this

PR #2950, opened 2025-10-10
[field] Add `dirty` and `touched` props

mj12albert:
Allows these states to be controlled by external libraries, both react-hook-form and Tanstack form provide these two states in their API

This is useful as Tanstack form treats "dirty" differently from RHF and other libraries: https://tanstack.com/form/latest/docs/framework/react/guides/basic-concepts#understanding-isdirty-in-different-libraries

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/popover/page.mdx
+++ b/docs/src/app/(public)/(content)/react/components/popover/page.mdx
@@ -159,4 +159,3 @@ Inside the `Viewport`, the content is further wrapped in `div`s with data attrib
 You can use these attributes to style the enter and exit animations.
 
 <Demo path="./demos/detached-triggers-full" />
-```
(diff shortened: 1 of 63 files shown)


packages/react/src/field/root/FieldRoot.tsx
```diff
@@ -42,19 +44,39 @@ export const FieldRoot = React.forwardRef(function FieldRoot(
...
+
+  const setTouched: typeof setTouchedUnwrapped = React.useCallback(
+    (value) => {
+      if (touchedProp !== undefined) {
+        return;
+      }
+      setTouchedUnwrapped(value);
+    },
+    [touchedProp],
+  );
```
atomiks: `useEventCallback` seems like it can be used here

PR #2966, opened 2025-10-10
[internal] Refactor to `useUntracked`

atomiks:
- Renames `useEventCallback` to `useUntracked`
- Removes usage of `useLatestRef` in favor of `useUntracked`'s return value
- Removes incorrect usage of `useLatestRef`
- `useLatestRef` has a special use case remaining for Number Field as the ref is actually mutated as well, changed it to a plain layout effect synchronizer to still remove the utility, as it should be readonly to begin with


packages/utils/src/useUntracked.ts
romgrk, line 1: I strongly disagree with the naming change. This hook does more than just providing an untracked stable callback, it also does the switch in an effect, which is a necessary implementation detail for event callbacks (e.g. `onClick={someEventCallback}`), but not strictly required for other types of callbacks. Can we clarify the use-cases for which we want to use this, and figure out a more appropriate name?

packages/utils/src/useLatestRef.ts
```diff
@@ -1,25 +0,0 @@
-'use client';
```
romgrk, line 1: Removing this hook in favor of `useUntracked` creates more inline allocations, I somewhat disagree with this change. Can you explain more clearly why this is beneficial?

packages/utils/src/useUntracked.ts
atomiks: > This hook does more than just providing an untracked stable callback, it also does the switch in an effect

Can you explain this more?

The hook itself is the same as `useEffectEvent` but without the linting restrictions which is annoying in practice (and is also a better name for non-function untracked values)

> but not strictly required for other types of callbacks

It's also necessary for effect event callbacks, and in general to avoid writing/reading refs in render which is how untracked polyfills work

packages/utils/src/useLatestRef.ts
atomiks: I really just wanted one hook to untrack all values, and not worrying about whether it's a function or not. Conceptually it's much simpler to me that it's called this and it's less confusing than the current `useEventCallback`/`useLatestRef` which are semantically confusing names

If the inline allocations are a problem (seems like a stretch), the function could return an object so it's only invoked once in the component body for many untracked values

packages/utils/src/useUntracked.ts
romgrk: > and is also a better name for non-function untracked values

But this hook always returns a function, even if the use-case is as a replacement of `useLatestRef`.

Can you do a quick poll on slack to check the consensus for the naming of this hook? The utils package isn't just for BUI anymore, it impacts all the other teams, so it's important that the naming be very clear for everyone. Maybe we can also get other naming suggestions.

atomiks: > But this hook always returns a function, even if the use-case is as a replacement of useLatestRef.

Yep, but I figured calling it to retrieve the value makes sense with the `useUntracked` name as way to avoid having two hooks. `useEffectEvent` implies it doesn't return anything as event handlers don't return values, they just fire side effects. Plus the name (and linting guard) doesn't cover the use case where it just memoizes the function for use in a context value or memo dep to be forwarded to a sub-component's props. It's unexpected that `useEffectEvent` (or `useEventCallback`) can be used as a way to untrack non-functions by calling them and retrieving the return value, when compared with `useUntracked`.

Also `useLatestRef` doesn't throw (since there's nothing to call during render), where I found a few mistakes in the codebase (that I've fixed in this PR) where it was reading a ref in render. So throwing is a bit of a DX guard there.

> Can you do a quick poll on slack [...] The utils package isn't just for BUI anymore, it impacts all the other teams

Yeah sure, this makes sense

PR #2974, opened 2025-10-13
[dialog][alert dialog] Support detached triggers

### LINKED ISSUE #2069: [RFC] Root components state can be accessible via context / children render function
# Feature request

Good day! First of all I'm a huge fan of this project. I've build and used many UI libraries in the past and I'm impressed about the consistency and general effort put into BaseUI. That being said, I'm missing a (in my opinion) quite important feature which I think would elevate the quality and DX of BaseUI even more.

Every component has its `Root` which manages and holds the internal state of the whole component. I'll use the `Dialog` component as the main example going forward.

Currently the basic example for the dialog component is as follows:

```ts
<Dialog.Root>
  <Dialog.Trigger>
    Open Dialog
  </Dialog.Trigger>
  // ...
</Dialog.Root>
```

This markup is nice and clear, but it limits the developer to use the `Dialog.Trigger` as the element which opens the dialog. This limitation can be quite annoying in real world applications, because this limitation seems to have no benefit, because one could also do this:

```ts
const [open, setOpen] = useState(false);

<Dialog.Root open={open} onOpenChange={setOpen}>
  <button onClick={() => setOpen(true)}>
    Open Dialog
  </button>
  // ...
</Dialog.Root>
```

Image in this case the `button` element to be a custom element which should open the dialog. As you can see, for this approach you need quite a bit of additional code because in order to achieve that you have to make the dialog root controlled, so you can manage the state yourself. In this simple case this is quite annoying because the controlled state is essentially doing the exact same thing what the uncontrolled state is doing.

The idea is now to introduce a possibility to make this need of making this controlled state obsolete:

```ts
<Dialog.Root>
  {({ open, setOpen }) => <>
    <button onClick={() => setOpen(true)}>
      Open Dialog
    </button>
    // ...
  </>}
</Dialog.Root>
```

The `Dialog.Root` components children prop can be a function, which receives `childrenProps`. These children props contain internal state information and their setters. That children props object can be thought of a "public state interface" which represents the whole state of the component + its setters. It could (and should) not only be exposed via children props but also via a context which would even more increase DX in the case when components are nested over multiple layers.

Now you might think that all of this is pointless, because why not just use the `render` prop of the `Dialog.Trigger` component, and there are multiple reasons:

1. In real world application a single button can do different things depending on the application state. For example if the user is logged in the button should do actionA while when I'm logged out the button should do actionB, where any action could be a modal to be opened or something entierly different.
2. BaseUI is an "abstraction" of a UI library and each adpoter can descide how to use the building blocks how every he / she sees fit. The addition of this feature would open up new patterns to be used (e.g. would improve DX).

## Summary

- All `Root` components define a `public state interface` which consists of one or multiple state values + their setters.
- The `children` prop accepts not only `ReactNode` as its value but also a function with the interface `(publicStateInterface) => ReactNode`, so the state can be accessed directly in the children.
- In nested components the `public state interface` can be accessed via a hook which behaves like `useContext`, this can be either something like `const { open, setOpen } = Dialog.Root.useContext()` or a separate export like `const { open, setOpen } = useDialogRoot()`.

I'm using this pattern in my own UI library, [headlessUI](https://headlessui.com/react/popover#using-render-props) and [react-aria](https://react-spectrum.adobe.com/react-aria/Modal.html#props), are using this pattern as well (its a little bit inconsitent in headlessUI).

## Motivation

In my opinion not having this feature is limiting DX greatly and BaseUI adopters are bound to restrictions which dont serve any benefit (please correct me if I'm wrong in this aspect).

If this is something BaseUI wants to implement, I would be ready to help with the implementation & make a PR.

### LINKED ISSUE #2802: Feature request: dialog manager
shouldn't there be a dialog manager the same way there is a toast manager?

simple dialogs like prompt, confirm, or alert are usually called in the same way that the corresponding built in browser functionality is, through an imperative API

they are also very similar to toasts in functionality - a part of the UI that lives separately from the rest of the app

michaldudak:
Implemented detached / multiple trigger support in Dialog and AlertDialog, similarly to https://github.com/mui/base-ui/pull/2336

It is now possible to define triggers outside of Dialog.Root (or AlertDialog.Root) using the `handle` prop. Multiple triggers per dialog with a dynamic payload are also implemented.

Couple of additions not present in the Popover:

1. Reactive payload - any changes to the active trigger's `payload` will be reflected in the Dialog.Root's child render function's parameter. This has been also implemented in Popover in this PR.
2. Imperative methods on handles - handles are now separate classes (not just Store instances as before) and expose public `open`, `openWithPayload`, and `close` methods. They can be used to interact with the dialog imperatively, even when in uncontrolled mode. They offer the ability to open a dialog and associate it with a given trigger (setting its ARIA attributes properly), opening a standalone dialog, and optionally providing a custom payload. This should make the controlled mode almost unnecessary and will make it much more convenient to implement scenarios like those described in #2802 and #2069. This will be ported to Popover in a separate PR.

Docs:
https://deploy-preview-2974--base-ui.netlify.app/react/components/dialog#detached-triggers
https://deploy-preview-2974--base-ui.netlify.app/react/components/alert-dialog#detached-triggers
https://deploy-preview-2974--base-ui.netlify.app/experiments/dialog/dialogs

Closes #2802
Closes #2069

### DIFF
--- a/docs/reference/generated/combobox-input.json
+++ b/docs/reference/generated/combobox-input.json
@@ -56,6 +56,13 @@
     },
     "data-focused": {
       "description": "Present when the input is focused (when wrapped in Field.Root)."
+    },
+    "data-list-empty": {
+      "description": "Present when the corresponding items list is empty."
+    },
+    "data-popup-side": {
+      "description": "Indicates which side the corresponding popup is positioned relative to its anchor.",
+      "type": "'top' | 'bottom' | 'left' | 'right' | 'inline-end' | 'inline-start' | null"
     }
   },
   "cssVariables": {}
--- a/docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
+++ b/docs/src/components/ReferenceTable/PropsReferenceAccordion.tsx
@@ -56,6 +56,10 @@ function getShortPropType(name: string, type: string | undefined) {
     return { type: 'string | function', detailedType: true };
   }
 
+  if (name === 'style') {
+    return { type: 'React.CSSProperties | function', detailedType: true };
+  }
+
   if (name === 'render') {
     return { type: 'ReactElement | function', detailedType: true };
   }
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
   "packageManager": "pnpm@10.18.2",
   "engines": {
     "pnpm": "10.18.2",
-    "node": ">=22.11"
+    "node": ">=22.18.0"
   },
   "resolutions": {
     "@babel/core": "^7.28.4",
--- a/packages/react/src/combobox/input/ComboboxInput.test.tsx
+++ b/packages/react/src/combobox/input/ComboboxInput.test.tsx
@@ -515,4 +515,62 @@ describe('<Combobox.Input />', () => {
       expect(screen.queryByRole('listbox')).to.equal(null);
     });
   });
+
+  describe('data state attributes', () => {
+    it.skipIf(isJSDOM)('sets data-popup-side to the current popup side', async () => {
+      const { user } = await render(
+        <Combobox.Root items={['apple']}>
+          <Combobox.Input />
+          <Combobox.Portal>
+            <Combobox.Positioner side="right">
+              <Combobox.Popup>
+                <Combobox.List>
+                  {(item: string) => (
+                    <Combobox.Item key={item} value={item}>
+                      {item}
+                    </Combobox.Item>
+                  )}
+                </Combobox.List>
+              </Combobox.Popup>
+            </Combobox.Positioner>
+          </Combobox.Portal>
+        </Combobox.Root>,
+      );
+
+      const input = screen.getByRole('combobox');
+      expect(input).not.to.have.attribute('data-popup-side');
+
+      await user.click(input);
+
+      await waitFor(() => expect(screen.queryByRole('listbox')).not.to.equal(null));
+      expect(input).to.have.attribute('data-popup-side', 'right');
+
+      await user.click(document.body);
+
+      await waitFor(() => expect(screen.queryByRole('listbox')).to.equal(null));
+      expect(input).not.to.have.attribute('data-popup-side');
+    });
+
+    it('toggles data-empty when the filtered list is empty', async () => {
+      const { user } = await render(
+        <Combobox.Root items={[]}>
+          <Combobox.Input />
+          <Combobox.Portal>
+            <Combobox.Positioner>
+              <Combobox.Popup>
+                <Combobox.List />
+              </Combobox.Popup>
+            </Combobox.Positioner>
+          </Combobox.Portal>
+        </Combobox.Root>,
+      );
+
+      const input = screen.getByRole('combobox');
+
+      await user.click(input);
+
+      await waitFor(() => expect(screen.getByRole('listbox')).not.to.equal(null));
+      expect(input).to.have.attribute('data-list-empty');
+    });
+  });
 });
--- a/packages/react/src/combobox/input/ComboboxInput.tsx
+++ b/packages/react/src/combobox/input/ComboboxInput.tsx
@@ -7,7 +7,11 @@ import { isAndroid, isFirefox } from '@base-ui-components/utils/detectBrowser';
 import { BaseUIComponentProps } from '../../utils/types';
 import { useBaseUiId } from '../../utils/useBaseUiId';
 import { useRenderElement } from '../../utils/useRenderElement';
-import { useComboboxInputValueContext, useComboboxRootContext } from '../root/ComboboxRootContext';
+import {
+  useComboboxDerivedItemsContext,
+  useComboboxInputValueContext,
+  useComboboxRootContext,
+} from '../root/ComboboxRootContext';
 import { selectors } from '../store';
 import { pressableTriggerOpenStateMapping } from '../../utils/popupStateMapping';
 import type { FieldRoot } from '../../field/root/FieldRoot';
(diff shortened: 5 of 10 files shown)


docs/reference/generated/dialog-root.json
```diff
@@ -23,12 +23,22 @@
       "description": "A ref to imperative actions.\n- `unmount`: When specified, the dialog will not be unmounted when closed.\nInstead, the `unmount` function must be called to unmount the dialog manually.\nUseful when the dialog's animation is controlled by an external library.",
       "detailedType": "React.RefObject<Dialog.Root.Actions> | undefined"
     },
+    "defaultTriggerId": {
+      "type": "string | null",
+      "description": "ID of the trigger that the dialog is associated with.\nThis is useful in conjuntion with the `defaultOpen` prop to create an initially open dialog.",
```
atomiks: `conjunction`

atomiks:
"Controlled, detached triggers" in the experiment seems to have an issue(?) with overlapping dialogs

michaldudak:
It was a bug in the experiment - too much copy-pasta.

docs/src/app/(public)/(content)/react/components/alert-dialog/page.mdx
```diff
@@ -87,3 +87,91 @@ To implement this, both dialogs should be controlled. The confirmation dialog ma
...
+
+```jsx title="Detached triggers"
+const demoAlertDialog = AlertDialog.createHandle();
+
+<AlertDialog.Trigger handle={demoAlertDialog}>Open</AlertDialog.Trigger>
+
+<AlertDialog.Root handle={demoAlertDialog}>
+  ...
+</AlertDialog.Root>
+```
```
atomiks: Using character syntax highlighting (`"demoAlertDialog"` in the code block fence frontmatter) would be helpful

docs/src/app/(public)/(content)/react/components/alert-dialog/page.mdx
```diff
@@ -87,3 +87,91 @@ To implement this, both dialogs should be controlled. The confirmation dialog ma
...
+
+```jsx title="Multiple detached triggers"
+const demoAlertDialog = AlertDialog.createHandle();
+
+<AlertDialog.Trigger handle={demoAlertDialog}>Trigger 1</AlertDialog.Trigger>
+
+<AlertDialog.Trigger handle={demoAlertDialog}>Trigger 2</AlertDialog.Trigger>
+
+<AlertDialog.Root handle={demoAlertDialog}>
```
atomiks, line 129: I don't think there should be line white space here, it makes it harder to read

packages/react/src/alert-dialog/trigger/AlertDialogTrigger.tsx
```diff
@@ -41,17 +63,56 @@ export const AlertDialogTrigger = React.forwardRef(function AlertDialogTrigger(
...
   });
 
+  const id = useBaseUiId(idProp);
+  const registerTrigger = useTriggerRegistration(id, store);
+
+  useIsoLayoutEffect(() => {
+    if (isTriggerActive) {
+      store.set('payload', payload);
+    }
+  }, [isTriggerActive, payload, store]);
```
atomiks, line 73: What happens if a trigger gets unmounted? I'm sort of wary of effects not returning cleanup functions as it's caused some bugs in these scenarios 🙈

packages/react/src/utils/popupStoreUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Returns a callback ref that registers/unregisters the trigger element in the store.
+ *
+ * @param id ID of the trigger. This must not be undefined by the time a ref is assigned.
+ * @param store The Store instance where the trigger should be registered.
+ */
+export function useTriggerRegistration<State extends { triggers: PopupTriggerMap }>(
+  id: string | undefined,
+  store: Store<State>,
+) {
+  const registeredId = React.useRef<string>(null);
```
atomiks: `registeredIdRef`

docs/src/app/(public)/(content)/react/components/alert-dialog/demos/detached-triggers-controlled/css-modules/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  return (
+    <React.Fragment>
+      <div className={styles.Container}>
+        <AlertDialog.Trigger
+          className={`${styles.Button} ${styles.DangerButton}`}
+          handle={demoAlertDialog}
+          id="alert-trigger-1"
+          payload={{ message: 'Discard draft?' }}
```
atomiks, line 25: Assuming there isn't one, it may be good to add a test of a function inside the payload that closes over a state variable to ensure the stale closure bug isn't present with this API

atomiks (approved):
Some minor remarks, but looks solid

docs/src/app/(public)/(content)/react/components/alert-dialog/demos/detached-triggers-controlled/css-modules/index.tsx
michaldudak: Is this what you mean? https://github.com/mui/base-ui/pull/2974/files#diff-20f068878440913459443b577cebb90419834b289d4f43eddbaecd8ed1ef8671R1147-R1203

packages/react/src/alert-dialog/trigger/AlertDialogTrigger.tsx
michaldudak: We still keep the payload in this case. Only when a new trigger becomes active, the payload changes. This is to avoid clearing the popup contents when close transition is running.

PR #2976, opened 2025-10-13
[autocomplete][combobox] `autoHighlight: "always"`, `keepHighlight`, `highlightItemOnHover` props

### LINKED ISSUE #2731: [combobox] No way for me to differentiate between hover and focus state on Combobox Items
# Feature request

Thank you for the combobox component!

I want to know how should I be styling the Combobox items to have different hover and focus styles?

As a consumer I would have thought that data-highlighted is applied on a `Combobox.Item` only when the options are made active through a keyboard.

### LINKED ISSUE #2922: [autocomplete][combobox] Controllable highlight prop
# Feature request

## Summary

[`autoHighlight`](https://base-ui.com/react/components/autocomplete#AutocompleteRoot-autoHighlight) is a bit limiting for some use cases:

> Whether to automatically highlight the first item while filtering.

## Motivation

Sometimes you:

### 1. Want an item to always be highlighted, not just when filtering

VS Code for instance does this, along with some Command Palette patterns:

Image

### 2. Want to prevent the highlight from disappearing when hovering out

If the highlight is linked to a side-panel appearing, then hovering out causes the side panel to disappear and become empty. Sometimes you want to retain the highlight while hovering out so something is always present. Additionally, it can allow sloppy pointer usage to quickly highlight an item and press <kbd>Enter</kbd> even if the pointer left, which may be useful.

`cmd+k` retains the highlight when hovering out

### 3. Want to highlight something other than the first item initially

Haven't yet seen this use case yet, but it allows for this.

[Ariakit](https://ariakit.org/reference/combobox#autoselect) allows the highlight to always be present with `"always" | boolean`, so it's a tri-state prop. It also has [`getAutoSelectId`](https://ariakit.org/reference/combobox#getautoselectid). 

### API

One solution is allowing `autoHighlight` to accept an object on top of a boolean instead of just a string

```js
<Combobox.Root
  autoHighlight={{
    item: ItemValue,
    behavior: "always" | "never" | "input-filter" 
   // also possibly "input-change" to keep the highlight when deleting all text
  }}
/>
```

The other is making it a real controllable prop (and removing `autoHighlight` + `onItemHighlighted`), but this forces you to make `inputValue` controlled if you want to replicate what `autoHighlight: true` does

- `defaultHighlightedItem={items[0]}` replicates `autoHighlight: "always"`
- `highlightedItem` (controlled) needs you to control `inputValue` to be able to tell whether filtering is occurring so you can set it to a nil value or not, which is less convenient
- `onHighlightedItemChange` can prevent a hover-out from removing the highlight by calling `eventDetails.cancel()` when `highlightedItem` is `undefined` (and replaces `onItemHighlighted`), but this is less nice than a dedicated API

atomiks:
- `autoHighlight: "always"` for Autocomplete ensures the highlight is always visible even if not filtering. Useful for command palettes
- `keepHighlight: boolean` retains the highlight when hovering out, useful if the highlight is linked to a side panel that should never disappear
- `highlightItemOnHover` disables the highlight moving on hover if the focus and hover states should be differentiated (closes #2731) - https://codesandbox.io/p/sandbox/keen-mirzakhani-6tk996

Closes #2922

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/autocomplete/page.mdx
+++ b/docs/src/app/(public)/(content)/react/components/autocomplete/page.mdx
@@ -141,9 +141,9 @@ Limit the number of visible items using the `limit` prop and guide users to refi
 
 ### Auto highlight
 
-Automatically highlight the first matching option as the user types by adding the `autoHighlight` prop.
+The first matching item can be automatically highlighted as the user types by specifying the `autoHighlight` prop on `<Autocomplete.Root>`. Set the prop's value to `"always"` if the highlight should always be present, such as when the list is rendered inline within a dialog.
 
-Set to `"always"` if the highlight should always be present, such as when the list is rendered inline within a dialog, and combine with the `keepHighlight` and `highlightItemOnHover` props to configure how the highlight behaves during mouse interactions.
+The prop can be combined with the `keepHighlight` and `highlightItemOnHover` props to configure how the highlight behaves during mouse interactions.
 
 <Demo path="./demos/auto-highlight" compact />
 
--- a/packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
+++ b/packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
@@ -294,16 +294,17 @@ describe('<Autocomplete.Root />', () => {
 
       await user.click(input);
       await user.type(input, 'ban');
+
       await waitFor(() => expect(screen.getByRole('listbox')).not.to.equal(null));
-      await waitFor(() => expect(input).to.have.attribute('aria-activedescendant'));
+      expect(input).to.have.attribute('aria-activedescendant');
+
       const highlightedBefore = input.getAttribute('aria-activedescendant');
       expect(highlightedBefore).to.not.equal(null);
 
       await user.clear(input);
+
       await waitFor(() => expect(screen.getByRole('listbox')).not.to.equal(null));
-      await waitFor(() =>
-        expect(input.getAttribute('aria-activedescendant')).to.equal(highlightedBefore),
-      );
+      expect(input.getAttribute('aria-activedescendant')).to.equal(highlightedBefore);
     });
 
     it('highlights the first item immediately when behavior is "always"', async () => {
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -1328,7 +1328,7 @@ interface ComboboxRootProps<ItemValue> {
    */
   openOnInputClick?: boolean;
   /**
-   * Controls whether the first matching item is highlighted automatically.
+   * Whether the first matching item is highlighted automatically.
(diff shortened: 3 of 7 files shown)


docs/src/app/(public)/(content)/react/components/autocomplete/page.mdx
```diff
@@ -141,7 +141,9 @@ Limit the number of visible items using the `limit` prop and guide users to refi
 
 ### Auto highlight
 
-Automatically highlight the first matching option as the user types using the `autoHighlight` prop.
+Automatically highlight the first matching option as the user types by adding the `autoHighlight` prop.
+
+Set to `"always"` if the highlight should always be present, such as when the list is rendered inline within a dialog, and combine with the `keepHighlight` and `highlightItemOnHover` props to configure how the highlight behaves during mouse interactions.
```
LukasTy:
```suggestion
Set to `"always"` if the highlight should always be present, such as when the list is rendered inline within a dialog.
Combine it with the `keepHighlight` and `highlightItemOnHover` props to configure how the highlight behaves during mouse interactions.
```

packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
```diff
@@ -269,6 +269,69 @@ describe('<Autocomplete.Root />', () => {
...
+          </Autocomplete.Portal>
+        </Autocomplete.Root>,
+      );
+
+      const input = screen.getByRole<HTMLInputElement>('combobox');
+
+      await user.click(input);
+      await user.type(input, 'ban');
+      await waitFor(() => expect(screen.getByRole('listbox')).not.to.equal(null));
+      await waitFor(() => expect(input).to.have.attribute('aria-activedescendant'));
```
LukasTy: It should be enough to `waitFor` the first assertion only. 🤔 
Same would apply to all the cases.

```suggestion
      expect(input).to.have.attribute('aria-activedescendant');
```

packages/react/src/combobox/root/ComboboxRoot.tsx
```diff
@@ -62,6 +65,21 @@ type ComboboxRootBaseProps<Value, Multiple extends boolean | undefined> = Omit<
    * @default false
    */
   multiple?: Multiple;
+  /**
+   * Controls whether the first matching item is highlighted automatically while filtering.
```
LukasTy: To align with other boolean fields.
```suggestion
   * Whether the first matching item is highlighted automatically while filtering.
```

packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1291,10 +1328,23 @@ interface ComboboxRootProps<ItemValue> {
    */
   openOnInputClick?: boolean;
   /**
-   * Whether to automatically highlight the first item while filtering.
+   * Controls whether the first matching item is highlighted automatically.
```
LukasTy:
```suggestion
   * Whether the first matching item is highlighted automatically.
```

Issue #2809, opened 2025-09-22
[toast] Toast.Description does not render when a custom component is passed to render


t6adev:
# Bug report

## Current behavior

There’s an issue with Toast.Description when providing a custom component via the render prop.
Using children as in [the docs](https://base-ui.com/react/components/toast#custom) works, but switching to render={<div>...</div>} makes the description disappear.

Like below (does not show):
```tsx
<Toast.Description
  render={<div>{'`data.userId` is '}{toast.data.userId}</div>}
/>
```

If you add `description: "foo"` to the add()'s params, it shows `data.userId` is XXX.
But it's not expected behavior.

## Expected behavior

Toast.Description should render the provided custom content when passed through the render prop without doing like add({ description: "foo" }).

## Reproducible example

https://codesandbox.io/p/sandbox/7slnx9

## Base UI version

v1.0.0-beta.3

mj12albert:
The issue is that this [`shouldRender`](https://github.com/mui/base-ui/blob/master/packages/react/src/toast/description/ToastDescription.tsx#L26) is false when children are passed via the render prop

This seems to fix it, even when there's no children within the render prop (`render={<div />}`)

```diff
diff --git a/packages/react/src/toast/description/ToastDescription.tsx b/packages/react/src/toast/description/ToastDescription.tsx
index bfe74eeba7..f5866c9504 100644
--- a/packages/react/src/toast/description/ToastDescription.tsx
+++ b/packages/react/src/toast/description/ToastDescription.tsx
@@ -23,7 +23,7 @@ export const ToastDescription = React.forwardRef(function ToastDescription(
 
   const children = childrenProp ?? toast.description;
 
-  const shouldRender = Boolean(children);
+  const shouldRender = Boolean(children) || render != undefined;
 
   const id = useId(idProp);
 
```

Another inelegant workaround is this:
```jsx
<Toast.Description
  render={(props) => {
    return (
      <div {...props}>
        this will be rendered
      </div>
    );
  }}
>
  fake children to force shouldRender to true that won't be rendered
</Toast.Description>
```

mj12albert:
So the preferred way for children is to NOT pass it through the render prop, unless you need to conditionally render using state e.g. https://master--base-ui.netlify.app/react/components/toggle

PR #2977, opened 2025-10-14
[toast] Fix Toast.Description not rendering when a custom component is passed

sai6855:
closes https://github.com/mui/base-ui/issues/2809

### Before
https://codesandbox.io/p/sandbox/compassionate-sid-vz7mmz?file=%2Fsrc%2FApp.tsx%3A59%2C36

### After
https://codesandbox.io/p/devbox/elegant-turing-r2xd6n?file=%2Fsrc%2FApp.tsx%3A62%2C12&workspaceId=ws_6M5xWXbU7jpjfBKgg78swk


atomiks:
We should add a test for this

sai6855:
> We should add a test for this

added 👍

packages/react/src/toast/description/ToastDescription.test.tsx
```diff
@@ -90,4 +90,34 @@ describe('<Toast.Description />', () => {
     expect(titleElement).not.to.equal(null);
     expect(titleElement.textContent).to.equal('description');
   });
+
+  it('renders the description when passed to render prop', async () => {
+    function CustomList() {
```
atomiks, line 94: I think the `Title` component has the same issue

atomiks:
This will render the element if you do `<Toast.Description render={<span />}` (with no children) and `description` is missing, though... not sure if there's a way to detect it

sai6855:
> This will render the element if you do `<Toast.Description render={<span />}` (with no children) and `description` is missing, though... not sure if there's a way to detect it

Got it, as changes in this PR are improved version compared to master branch and considering there isn't elegant way to detect your edge case, should we merge this or keep current changes as is and think of better solution.

michaldudak:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

**Approach validation:** The fix `Boolean(children) || render !== undefined` is the correct approach. When a user provides a render prop, they are explicitly opting to control the rendering, so the component should render even without immediate children content. The historical context confirms this was an oversight in the original implementation, not intentional behavior.

michaldudak:
> Got it, as changes in this PR are improved version compared to master branch and considering there isn't elegant way to detect your edge case, should we merge this or keep current changes as is and think of better solution.

I don't think it's an edge case - this is a default way of styling the component using CSS in JS. I'd be in favor of documenting the current behavior better (especially underlining that children shouldn't be passed in a render prop).
Additionally, we can consider add an `alwaysRender` prop that will force rendering of empty description if this helps.

Issue #2758, opened 2025-09-14
[form] Documentation on the `onClearErrors` prop for the form component is insufficient


marcospgp:
> Event handler called when the errors object is cleared.

it doesn't say when the errors object is cleared or why. thanks!

https://base-ui.com/react/components/form

mj12albert:
@marcospgp Are you already using it? Or is it not working as you expected

I guess the name `onClearErrors` is kind of misleading, it's called whenever one error is cleared e.g. when the user corrects a wrong value (and not errors plural as the name may imply), and the `errors` arg are the remaining errors in the form

Maybe it would be helpful to provide the `name` of the field that was cleared of an error in the callback as well

marcospgp:
@mj12albert personally I'm generally not finding form error handling useful, I just handle validation independently for each field because managing a dictionary of errors keyed by input field name seems like it would be way too much trouble.

I just validate fields independently. And as long as they are validated when the form is submitted, then it works fine.

Then if the server returns an error that the client-side validation did not catch, I just assume that that is a bug and display a generic error dialog, similarly to whenever any other unhandled exception would occur.

mj12albert:
`onClearErrors` will be dropped in https://github.com/mui/base-ui/issues/2758

Instead external `errors` are always cleared on change as they cannot be revalidated internally by `<Form>`, will also add more docs on using external/server-side errors in https://github.com/mui/base-ui/pull/2989

PR #2983, opened 2025-10-15
[docs][form] Add an example with `React.useActionState`

mj12albert:
Preview: https://deploy-preview-2983--base-ui.netlify.app/react/components/form#submit-with-a-server-function

TODO
- [ ] Figure out automatic form reset: https://github.com/facebook/react/issues/29034
- [ ] Test progressive enhancement permalink https://react.dev/reference/rsc/server-functions#progressive-enhancement-with-useactionstate

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/form/demos/form-action/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/form/demos/form-action/css-modules/index.tsx
@@ -1,3 +1,4 @@
+'use client';
 import * as React from 'react';
 import { Field } from '@base-ui-components/react/field';
 import { Form } from '@base-ui-components/react/form';
@@ -14,7 +15,7 @@ export default function ActionStateForm() {
   });
 
   return (
-    <Form className={styles.Form} action={formAction}>
+    <Form className={styles.Form} errors={state.serverErrors} action={formAction}>
       <Field.Root name="username" className={styles.Field}>
         <Field.Label className={styles.Label}>Username</Field.Label>
         <Field.Control
@@ -25,10 +26,6 @@ export default function ActionStateForm() {
           className={styles.Input}
         />
         <Field.Error className={styles.Error} />
-
-        <Field.Error match={!!state.serverErrors?.username}>
-          {state.serverErrors?.username}
-        </Field.Error>
       </Field.Root>
       <button disabled={loading} type="submit" className={styles.Button}>
         Submit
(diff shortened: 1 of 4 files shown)


docs/src/app/(public)/(content)/react/components/form/demos/form-action/css-modules/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          required
+          defaultValue="admin"
+          placeholder="e.g. alice132"
+          className={styles.Input}
+        />
+        <Field.Error className={styles.Error} />
+
+        <Field.Error match={!!state.serverErrors.username}>
+          {state.serverErrors.username}
+        </Field.Error>
```
mj12albert: When returning errors with `useActionState`, passing them to the errors prop forces the validity state to be invalid: https://github.com/mui/base-ui/blob/master/packages/react/src/field/utils/getCombinedFieldValidityData.ts#L15

This blocks re-submitting the form for revalidation on the server (in the server fn) so the error gets stuck

docs/src/app/(public)/(content)/react/components/form/demos/form-action/tailwind/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export default function ActionStateForm() {
+  const [state, formAction, loading] = React.useActionState<FormState, FormData>(submitForm, {
+    success: false,
+    serverErrors: {},
+  });
+
+  return (
+    <Form className="flex w-full max-w-64 flex-col gap-4" action={formAction}>
```
atomiks: `onSubmit` (or the internal effect from `errors` prop + `submittedRef.current`) handles focusing the first control with an error, which seems to be missing here. Does something else need to be done?

mj12albert: Oh it's because I'm not passing the errors (assumed to be returned by a server/externally) to the errors prop https://github.com/mui/base-ui/pull/2983#discussion_r2432107546

Maybe here: https://github.com/mui/base-ui/blob/master/packages/react/src/form/Form.tsx#L64
...could we assume that external errors via the `errors` prop are expected to be cleared externally (server-side) and shouldn't block submission? 🤔 

Somewhat related - I'm not sure where `onClearErrors` should fit into all of this https://github.com/mui/base-ui/issues/2758

atomiks: > could we assume that external errors via the errors prop are expected to be cleared externally (server-side) and shouldn't block submission? 🤔

I think this relates to `Field.Root`'s `validationMode` prop - that makes sense for `onSubmit` but not `onBlur` and `onChange`? Maybe `Form` should be able to override all of their props with its own `validationMode` prop that you can set to `onSubmit`, in which case that behavior would make sense, because then if there are any errors it shouldn't block submission as the validation happens on submit

PR #2985, opened 2025-10-15
[collapsible] Fix `starting-style` state

### LINKED ISSUE #2984: [collapsible] Starting styles do not apply when using Tailwind
# Bug report

## Current behavior

Using `data-[starting-style]` with tailwind to customize the starting transition for the collapsible panel stopped working after the beta 4 release. Ending style works as expected. Tested beta 3 with the same code and works normally in that version so it seems to be specific to the beta 4 release. 

## Expected behavior

When something like `data-[starting-style]:opacity-0` is used, the panel does not animate the opacity from 0. It starts with the opacity at 100. However, `data-[ending-style]:opacity-0` does work as expected. 

## Reproducible example

[Example here based on docs example](https://codesandbox.io/p/sandbox/epic-wave-y5hshk)

## Base UI version

v1.0.0-beta.4

Chrome/Firefox

## Which OS are you using?

Windows 11

Voiceover/JAWS etc.

## Additional context

atomiks:
Fixes #2984

CSS Modules: https://codesandbox.io/p/sandbox/amazing-ben-86krv3
Tailwind: https://codesandbox.io/p/sandbox/fancy-leftpad-xnsxsg

https://deploy-preview-2985--base-ui.netlify.app/experiments/collapsible/transitions

Weird behaviors in CodeSandbox (default browser behavior) not on our docs (due to custom styles): with `keepMounted`, the transition plays on mount. CSS reset of `[hidden]` removes this behavior, but this is a separate issue to the one linked. The animation correctly plays with `keepMounted` when the `[hidden]` issue is dealt with by adding:

```css
&[hidden]:not([hidden='until-found']) {
  display: none;
}
```

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/collapsible/demos/hero/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/collapsible/demos/hero/css-modules/index.module.css
@@ -58,6 +58,10 @@
   line-height: 1.25rem;
   transition: all 150ms ease-out;
 
+  &[hidden]:not([hidden='until-found']) {
+    display: none;
+  }
+
   &[data-starting-style],
   &[data-ending-style] {
     height: 0;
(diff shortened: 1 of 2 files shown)


mj12albert (approved):
It works well, tested the subgrid case as well

The Tailwind + defaultOpen + opacity transition issue also happens in main 😓

PR #3005, opened 2025-10-17
[utils] Make sure Store state is not mutable outside of the class

michaldudak:
As stores are accessible to developers with the `handle` prop on several components, I made it harder to accidentally update the state directly by making it read-only.

### DIFF
--- a/packages/utils/src/store/ReactStore.ts
+++ b/packages/utils/src/store/ReactStore.ts
@@ -8,14 +8,7 @@ import { useIsoLayoutEffect } from '../useIsoLayoutEffect';
 import { NOOP } from '../empty';
 
 /**
- * A Store that supports controlled state keys.
- *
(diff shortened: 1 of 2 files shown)


atomiks:
> I would prefer if we didn't merge this change, and we aimed to keep Store instances as an implementation detail - not accessible externally.

I agree with this - the exposed store instance it should be changed to only export a subset of methods/properties, to avoid breaking changes

michaldudak:
We can't make it properly inaccessible from the outside, as `store` is a field in "handles" - data structures several components use to detach triggers.
I suppose it should be enough to mark the field as `@internal` in JSDOM and perhaps call it `__store` or similar instead to indicate it's not something to peek into.

romgrk:
Marking the field as internal/private is enough imo.

PR #3006, opened 2025-10-17
[utils] Store inspector

michaldudak:
Vibe-coded a utility to visualize the content of the Store. I found it helpful, so perhaps it'll be handy for others as well.

Example: https://codesandbox.io/p/sandbox/quiet-fire-pprjnf

### DIFF
--- a/packages/utils/src/store/StoreInspector.tsx
+++ b/packages/utils/src/store/StoreInspector.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import { Store } from './Store';
-import { ReactStore } from './ReactStore';
 import { useForcedRerendering } from '../useForcedRerendering';
 import { useStableCallback } from '../useStableCallback';
 import { useAnimationFrame } from '../useAnimationFrame';


michaldudak:
I found it helpful to have a glance at how the components' state changes in real time and be able to inspect deeply nested objects. This is also possible with console.log (and in some cases more convenient), but I was able to pinpoint several issues with this inspector already, so it's definitely useful for me. I don't mind if it doesn't get merged - I can always have it in a Gist and copy when needed, but I thought it may be helpful for others as well.

packages/utils/src/store/StoreInspector.tsx
```diff
@@ -0,0 +1,567 @@
...
+
+    return unsubscribe;
+  }, [store, rerender, rerenderTimeout]);
+
+  const logToConsole = useStableCallback(() => {
+    const data: any = {
+      state: store.state,
+    };
+
+    if (store instanceof ReactStore && Object.keys((store as any).context ?? {}).length > 0) {
```
Janpot: @romgrk Bit of an edge-case maybe, but this `instanceof` could be broken if multiple instances of the utils package are installed, right?

mj12albert (commented):
As I'm not too familiar with the store I think this is pretty useful (probably the more familiar you are with it the less useful this would be)

I wonder if this can be integrated with the experiments settings panel somehow

packages/utils/src/store/StoreInspector.tsx
romgrk: Yes. That check could be removed.

PR #3011, opened 2025-10-18
[popover] Fix swiping or scrolling on nested popup dismissing popover on touch

### LINKED ISSUE #3008: [popover] Modal inside Popover is closed unexpectedly on mobile when swiping
# Bug report

## Current behavior

The setup is simple: Place a modal inside a popover and open the modal by clicking on the modal trigger.

On Desktop everything behaves as expected, the popover can't be closed unless the modal is being closed.
On mobile however the popover is being closed by a users swipe and thus also the modal is being closed.
 

## Expected behavior

Like on desktop, on mobile the open modal should prevent the popover to be closed.

## Reproducible example

https://codesandbox.io/p/devbox/musing-leakey-67pv74?workspaceId=ws_9E7g8meGhxx9KyReSrgWiy

Here is also an additional video of triggering the bug via the Firefox dev tools (mobile touch events simulation):

https://github.com/user-attachments/assets/abdd90e8-26e1-408a-b1a5-22f7ca13e65b

## Base UI version

v1.0.0.beta-4

Chrome Mobile, Firefox Mobile

## Additional context

https://github.com/mui/base-ui/issues/2390 might be related

atomiks:
Fixes #3008

This ensures pressing on nested children (in the React tree) doesn't dismiss on `pointermove` (swiping/scrolling) by using the capture event trick, which is already done for the other event handlers

Also refactored from pointer to touch events because (at least on non-iOS/on Chrome DevTools), `pointermove` stops firing once scroll starts, while `touchmove` always fires which ensures the popover always dismisses while scrolling away

Note: https://github.com/mui/base-ui/pull/2978 refactors away the timeout for the `handleCaptureInside` logic - I will rebase one of the PRs depending on which is merged first

### DIFF
--- a/docs/src/app/(private)/experiments/navigation-menu-popups.tsx
+++ b/docs/src/app/(private)/experiments/navigation-menu-popups.tsx
@@ -1,0 +1,243 @@
+'use client';
+import * as React from 'react';
+import { NavigationMenu } from '@base-ui-components/react/navigation-menu';
+import { Dialog } from '@base-ui-components/react/dialog';
+import { Popover } from '@base-ui-components/react/popover';
+import { AlertDialog } from '@base-ui-components/react/alert-dialog';
+
+export default function ExampleNavigationMenu() {
+  return (
+    <React.Fragment>
+      <NavigationMenu.Root className="min-w-max rounded-lg bg-gray-50 p-1 text-gray-900">
+        <NavigationMenu.List className="relative flex">
+          <NavigationMenu.Item>
+            <NavigationMenu.Trigger className={triggerClassName}>
+              Dialog Test
+            </NavigationMenu.Trigger>
+
+            <NavigationMenu.Content className={contentClassName}>
+              <Dialog.Root>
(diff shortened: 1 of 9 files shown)


packages/react/src/floating-ui-react/hooks/useDismiss.ts
```diff
@@ -113,10 +113,6 @@ export interface UseDismissProps {
    * floating elements.
    */
   bubbles?: boolean | { escapeKey?: boolean; outsidePress?: boolean };
-  /**
-   * Determines whether to use capture phase event listeners.
-   */
-  capture?: boolean | { escapeKey?: boolean; outsidePress?: boolean };
```
mnajdova, line 119: Don't we still need this? Why was it removed?

mnajdova (commented):
Csb with the fix: https://codesandbox.io/p/devbox/brave-butterfly-rm8zjs?workspaceId=ws_HE9JRzFW36Y2kpBoSwdvZ5

Can you elaborate why we are removing the way to customize the capture of the keyboard event using escape? What's the correlation to the bug?

packages/react/src/floating-ui-react/hooks/useDismiss.ts
atomiks: No, it's an option from Floating UI config, but we don't use it anywhere in our components, so it can be removed.

Issue #2142, opened 2025-06-20
[form] Support `validationMode: onSubmit`


mj12albert:
Example: https://deploy-preview-2137--base-ui.netlify.app/experiments/forms/rhf (change `validationMode` in the settings panel)

Using `onSubmit` validation, errors are not displayed until the first (unsuccessful) submission. After that RHF lets you configure `revalidateMode`, but we could possibly come up with a good opinionated behavior.

This would provide an option that's even less noisy than `onBlur` validation and closer to a native form submission: https://adamsilver.io/blog/the-problem-with-live-validation-and-what-to-do-instead/

PR #3013, opened 2025-10-19
[form][field] Add `'onSubmit'` validation mode

mj12albert:
**Breaking:** `<Field>`'s default `validationMode` is now `'onSubmit'`. 

- `validationMode` can be set on `<Form>` to configure all child `<Field>`s
- `validationMode: 'onSubmit'` runs validations when the `<Form>` is first submitted, afterwards fields are re-validated on change.

This mode matches native form submission behavior and the defaults of RA, Ariakit and react-hook-form.

Can be tested in the [forms experiment](https://deploy-preview-3013--base-ui.netlify.app/experiments/forms/form) though it's mostly `required` validations

Closes https://github.com/mui/base-ui/issues/2142

### DIFF
--- a/packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
+++ b/packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
@@ -14,6 +14,58 @@ describe('<Autocomplete.Root />', () => {
 
   const { render } = createRenderer();
 
+  describe('keyboard interactions', () => {
+    it('closes popup on Tab after selecting with Enter and typing again', async () => {
+      const { user } = await render(
+        <Autocomplete.Root items={['alpha', 'alpine', 'beta']} autoHighlight>
+          <Autocomplete.Input />
+          <Autocomplete.Portal>
+            <Autocomplete.Positioner>
+              <Autocomplete.Popup>
+                <Autocomplete.List>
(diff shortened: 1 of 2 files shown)


packages/react/src/field/root/FieldRoot.tsx
```diff
@@ -195,19 +196,24 @@ export interface FieldRootProps extends BaseUIComponentProps<'div', FieldRoot.St
   /**
    * A function for custom validation. Return a string or an array of strings with
    * the error message(s) if the value is invalid, or `null` if the value is valid.
+   * Asynchronous functions are supported, but they do not prevent form submission
+   * when using `validationMode="onSubmit"`.
```
mj12albert, line 208: Not sure if it's a good idea to try to handle it, but its different from the native behavior anyway so I think we could just ignore and only document it for now, what do you think? @atomiks

atomiks: I don't think it's possible to support it given `event.preventDefault()` needs to be sync, it makes sense to document it here

mj12albert:
Needs https://github.com/mui/base-ui/pull/3048

atomiks (approved):
Nice~ matching RHF is a good default

PR #3018, opened 2025-10-21
[autocomplete] Refactor `alwaysSubmitOnEnter` to `submitOnItemClick` prop

atomiks:
This was an oversight in the original PR since the `onClick` behavior was temporarily changed to not fire when pressing <kbd>Enter</kbd>, but that was reverted before the beta.4 release

The `alwaysSubmitOnEnter` prop is limiting since the submission should occur on item click (both pointer press AND <kbd>Enter</kbd> press), as happens with search autocomplete widgets

- `submitOnItemClick={false}`: does not submit the owning form when pressing <kbd>Enter</kbd> in `Input` by stopping the event, desirable with multi-field forms
- `submitOnItemClick={true}`: submits the owning form when pressing <kbd>Enter</kbd> (does not stop the event) along with calling `.requestSubmit()` when pressing with the pointer, for single-field forms. 

Note that a submit button must be present inside the form because of the hidden input. Native HTML behavior doesn't cause the form to submit if there are two inputs without a submit button (it only submits if there's one single input when there's no button)

### DIFF
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -1465,8 +1465,7 @@ interface ComboboxRootProps<ItemValue> {
    */
   locale?: Intl.LocalesArgument;
   /**
-   * Whether clicking an item should submit the autocomplete's owning form.
-   * By default, clicking an item via a pointer or <kbd>Enter</kbd> key does not submit the owning form.
+   * Whether clicking an item should submit the owning form.
    * @default false
    */
   submitOnItemClick?: boolean;


packages/react/src/autocomplete/root/AutocompleteRoot.tsx
```diff
@@ -196,6 +196,12 @@ export interface AutocompleteRootProps<ItemValue>
    * Event handler called when the input value of the autocomplete changes.
    */
   onValueChange?: (value: string, eventDetails: AutocompleteRootChangeEventDetails) => void;
+  /**
+   * Whether clicking an item should submit the autocomplete's owning form.
+   * By default, clicking an item via a pointer or Enter key does not submit the owning form.
```
LukasTy: WDYT, would adding a note about this prop being mostly useful for Autocomplete widgets improve clarity? 🤔 

```suggestion
   * By default, clicking an item via a pointer or <kbd>Enter</kbd> key does not submit the owning form.
   * Mostly useful when creating custom widgets.
```

LukasTy (approved):
Nice work. 👍

PR #3022, opened 2025-10-21
[menu] Keep state in a store 

michaldudak:
Updates the internal implementation of Menu to use the ReactStore instance. This is required for detached triggers.

Built on top of #3037
Part of #3031

### DIFF
--- a/packages/react/src/menu/store/MenuStore.ts
+++ b/packages/react/src/menu/store/MenuStore.ts
@@ -98,7 +98,7 @@ export class MenuStore extends ReactStore<State, Context, typeof selectors> {
     );
 
     // Sync `allowMouseEnter` with parent menu if applicable.
-    this.observeSelector(
+    this.observe(
       createSelector((state) => state.allowMouseEnter),
       (allowMouseEnter) => {
         if (this.state.parent.type === 'menu') {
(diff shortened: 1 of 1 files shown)


packages/react/src/menu/store/MenuStore.ts
```diff
@@ -0,0 +1,164 @@
...
+    return value;
+  },
+};
+
+function registerParent(store: ReactStore<any, Context, any>, parent: MenuParent) {
+  if (parent.type === 'menu') {
+    parent.store.subscribe(() => {
+      // Propagate changes from parent menu
+      store.notifyAll();
+    });
```
romgrk: The dispose callback returned by `subscribe()` should always be used.

michaldudak: After we resolve #3037, I'll update the implementation here.

atomiks:
Nested menu is glitching

https://github.com/user-attachments/assets/82201c29-7b61-4187-9230-f353d70e4298

atomiks:
Trying to hover a nested menu (via the triangle area movement, not rectangular bridge) now closes it: https://deploy-preview-3022--base-ui.netlify.app/experiments/menu/menu-fully-featured

PR #3037, opened 2025-10-22
[utils] Add infrastructure to support nested stores

michaldudak:
A couple of changes to Store and ReactStore:

- **BREAKING CHANGE:** renamed ReactStore's `getElementSetter` to `useStateSetter` as it calls `React.useCallback` internally.
- Added a `notifyAll` method to the Store. It's useful to notify listeners of a store when a nested store is updated. See the "supports nested stores as state values" test for usage example.
- Added a `select` method to get a non-reactive store value. It's a shortcut to calling a selector manually.
- Added an `observeSelector` method to be notified about changes to a particular slice of the state.
- Added support for parametrized selectors in `useState` and `select` methods.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -57,4 +57,4 @@ jobs:
             ARGS="$ARGS --tag ${{ inputs.dist-tag }}"
           fi
 
-          pnpm code-infra publish $ARGS
+          pnpm code-infra publish --ci $ARGS
--- a/docs/reference/generated/toolbar-button.json
+++ b/docs/reference/generated/toolbar-button.json
@@ -36,9 +36,6 @@
     }
   },
   "dataAttributes": {
-    "data-highlighted": {
-      "description": "Present when the button is the active item in the toolbar."
-    },
     "data-orientation": {
       "description": "Indicates the orientation of the toolbar.",
       "type": "'horizontal' | 'vertical'"
--- a/docs/reference/generated/toolbar-input.json
+++ b/docs/reference/generated/toolbar-input.json
@@ -34,9 +34,6 @@
     }
   },
   "dataAttributes": {
-    "data-highlighted": {
-      "description": "Present when the input is the active item in the toolbar."
-    },
     "data-orientation": {
       "description": "Indicates the orientation of the toolbar.",
       "type": "'horizontal' | 'vertical'"
--- a/docs/reference/generated/toolbar-link.json
+++ b/docs/reference/generated/toolbar-link.json
@@ -18,9 +18,6 @@
     }
   },
   "dataAttributes": {
-    "data-highlighted": {
-      "description": "Present when the link is the active item in the toolbar."
-    },
     "data-orientation": {
       "description": "Indicates the orientation of the toolbar.",
       "type": "'horizontal' | 'vertical'"
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
     "release:version": "lerna version --no-changelog --no-push --no-git-tag-version --no-private",
     "release:build": "lerna run --concurrency 8 --no-private build --skip-nx-cache",
     "release:changelog": "tsx scripts/releaseChangelog.mts",
-    "release:publish": "pnpm publish --recursive --tag latest",
-    "release:publish:dry-run": "pnpm publish --recursive --tag latest --registry=\"http://localhost:4873/\"",
+    "release:publish": "code-infra publish --github-release",
+    "release:publish:dry-run": "code-infra publish --github-release --dry-run ",
     "docs:api": "pnpm --filter api-docs-builder start",
     "docs:build": "pnpm --filter docs generate-llms && pnpm --filter docs build",
     "docs:deploy": "pnpm --filter docs run deploy",
@@ -20,8 +20,8 @@
     "docs:link-check": "pnpm --filter docs link-check",
     "docs:generate-llms": "pnpm --filter docs run generate-llms",
     "extract-error-codes": "code-infra extract-error-codes --errorCodesPath docs/src/error-codes.json --detection opt-out",
-    "eslint": "eslint . --cache --report-unused-disable-directives --ext .js,.ts,.tsx,.mts --max-warnings 0 && pnpm -r lint",
-    "eslint:ci": "eslint . --report-unused-disable-directives --ext .js,.ts,.tsx,.mts --max-warnings 0 && pnpm -r lint",
+    "eslint": "eslint . --cache --report-unused-disable-directives --max-warnings 0 && pnpm -r lint",
+    "eslint:ci": "eslint . --report-unused-disable-directives --max-warnings 0 && pnpm -r lint",
     "stylelint": "stylelint --reportInvalidScopeDisables --reportNeedlessDisables \"docs/**/*.{js,ts,tsx}\" \"**/*.css\" --ignore-path .lintignore",
     "markdownlint": "markdownlint-cli2 \"**/*.{md,mdx}\"",
     "valelint": "pnpm vale sync && git ls-files | grep -h \".md$\" | xargs pnpm vale --filter='.Level==\"error\"'",
@@ -53,7 +53,8 @@
     "test:browsers": "cross-env VITEST_ENV=all-browsers pnpm test:_unit",
     "test:argos": "code-infra argos-push --folder test/regressions/screenshots/chrome",
     "typescript": "tsc -b tsconfig.json",
-    "inline-scripts": "tsx ./scripts/inlineScripts.mts"
+    "inline-scripts": "tsx ./scripts/inlineScripts.mts",
+    "use-react-18": "code-infra set-version-overrides --pkg react@18"
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.18.2",
--- a/packages/react/src/popover/root/PopoverRoot.test.tsx
+++ b/packages/react/src/popover/root/PopoverRoot.test.tsx
@@ -580,12 +580,18 @@ describe('<Popover.Root />', () => {
 
           await user.tab({ shift: true });
 
-          expect(screen.getByRole('button')).toHaveFocus();
+          await waitFor(() => {
+            expect(screen.getByRole('button')).toHaveFocus();
+          });
 
-          expect(screen.queryByTestId('popup')).to.toBeVisible();
+          await waitFor(() => {
+            expect(screen.queryByTestId('popup')).toBeVisible();
+          });
 
           await user.tab();
-          expect(screen.getByTestId('input-inside')).toHaveFocus();
+          await waitFor(() => {
+            expect(screen.getByTestId('input-inside')).toHaveFocus();
+          });
         },
       );
     });
--- a/packages/react/src/toolbar/button/ToolbarButtonDataAttributes.ts
+++ b/packages/react/src/toolbar/button/ToolbarButtonDataAttributes.ts
@@ -12,8 +12,4 @@ export enum ToolbarButtonDataAttributes {
    * Present when the button remains focusable when disabled.
    */
   focusable = 'data-focusable',
-  /**
-   * Present when the button is the active item in the toolbar.
-   */
-  highlighted = 'data-highlighted',
 }
--- a/packages/react/src/toolbar/input/ToolbarInputDataAttributes.ts
+++ b/packages/react/src/toolbar/input/ToolbarInputDataAttributes.ts
@@ -12,8 +12,4 @@ export enum ToolbarInputDataAttributes {
    * Present when the input remains focusable when disabled.
    */
   focusable = 'data-focusable',
-  /**
-   * Present when the input is the active item in the toolbar.
-   */
-  highlighted = 'data-highlighted',
 }
--- a/packages/react/src/toolbar/link/ToolbarLinkDataAttributes.ts
+++ b/packages/react/src/toolbar/link/ToolbarLinkDataAttributes.ts
@@ -4,8 +4,4 @@ export enum ToolbarLinkDataAttributes {
    * @type {'horizontal' | 'vertical'}
    */
   orientation = 'data-orientation',
-  /**
-   * Present when the link is the active item in the toolbar.
-   */
-  highlighted = 'data-highlighted',
 }
--- a/packages/react/src/tooltip/root/TooltipRoot.test.tsx
+++ b/packages/react/src/tooltip/root/TooltipRoot.test.tsx
@@ -1187,19 +1187,31 @@ describe('<Tooltip.Root />', () => {
       expect(screen.queryByTestId(popupId)).to.equal(null);
 
       await user.hover(trigger1);
-      expect(screen.queryByTestId(popupId)).toBeVisible();
+      await waitFor(() => {
+        expect(screen.queryByTestId(popupId)).toBeVisible();
+      });
       await user.unhover(trigger1);
-      expect(screen.queryByTestId(popupId)).to.equal(null);
+      await waitFor(() => {
+        expect(screen.queryByTestId(popupId)).to.equal(null);
+      });
 
       await user.hover(trigger2);
-      expect(screen.queryByTestId(popupId)).toBeVisible();
+      await waitFor(() => {
+        expect(screen.queryByTestId(popupId)).toBeVisible();
+      });
       await user.unhover(trigger2);
-      expect(screen.queryByTestId(popupId)).to.equal(null);
+      await waitFor(() => {
+        expect(screen.queryByTestId(popupId)).to.equal(null);
+      });
 
       await user.hover(trigger3);
-      expect(screen.queryByTestId(popupId)).toBeVisible();
+      await waitFor(() => {
+        expect(screen.queryByTestId(popupId)).toBeVisible();
+      });
       await user.unhover(trigger3);
-      expect(screen.queryByTestId(popupId)).to.equal(null);
+      await waitFor(() => {
+        expect(screen.queryByTestId(popupId)).to.equal(null);
+      });
     });
 
     it('should open the tooltip with any trigger on focus', async () => {
--- a/packages/react/src/utils/useScrollLock.ts
+++ b/packages/react/src/utils/useScrollLock.ts
@@ -3,6 +3,7 @@ import { ownerDocument, ownerWindow } from '@base-ui-components/utils/owner';
 import { useIsoLayoutEffect } from '@base-ui-components/utils/useIsoLayoutEffect';
 import { Timeout } from '@base-ui-components/utils/useTimeout';
 import { AnimationFrame } from '@base-ui-components/utils/useAnimationFrame';
+import { isOverflowElement } from '@floating-ui/utils/dom';
 import { NOOP } from './noop';
 
 /* eslint-disable lines-between-class-members */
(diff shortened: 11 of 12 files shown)


packages/utils/src/store/Store.ts
```diff
@@ -76,7 +76,18 @@ export class Store<State> {
...
+  /**
+   * Gives the state a new reference and updates all registered listeners.
+   */
+  public notifyAll() {
+    if (typeof this.state === 'object' && this.state != null) {
+      this.state = { ...(this.state as Record<string, unknown>) } as State;
+    }
+
+    this.listeners.forEach((l) => l(this.state));
+  }
```
romgrk, line 105: Can we implement this by calling an existing update method?

packages/utils/src/store/ReactStore.ts
```diff
@@ -134,6 +170,16 @@ export class ReactStore<
...
   public set<T>(key: keyof State, value: T): void {
+    const interceptor = this.writeInterceptors.get(key);
+    if (interceptor) {
+      const updatedValue = interceptor(value, this);
+      if (!this.controlledValues.get(key) === true) {
+        super.set(key, updatedValue);
+      }
+
+      return;
+    }
```
romgrk: This implementation makes a hashmap check for every write even if no `writeInterceptors` were provided. It would be better for performance to keep `writeInterceptors` empty and avoid the `Map.get()` call.

This also applies for the other write methods.

packages/utils/src/store/ReactStore.ts
```diff
@@ -151,9 +197,20 @@ export class ReactStore<
   public apply(values: Partial<State>): void {
     const newValues = { ...values };
     for (const key in newValues) {
+      if (!Object.prototype.hasOwnProperty.call(newValues, key)) {
```
romgrk: `Object.hasOwn` should be fine to use now.

packages/utils/src/store/Store.ts
```diff
@@ -62,7 +62,7 @@ export class Store<State> {
   public apply(changes: Partial<State>) {
     for (const key in changes) {
       if (!Object.is(this.state[key], changes[key])) {
-        this.update({ ...this.state, ...changes });
+        Store.prototype.update.call(this, { ...this.state, ...changes });
```
romgrk: Why is this needed?

packages/utils/src/store/ReactStore.ts
romgrk, line 1: > Added writeInterceptors to ReactStore. They fire whenever a value is about to be set and let users update the value or trigger any side effects (like calling an onChange event). This is also useful to register to a nested store's updates (required in [Menu] Keep state in a store  #3022).

Could you expand on why this logic is required for the menu store?

packages/utils/src/store/Store.ts
michaldudak: Due to dynamic dispatch, if `this` is a ReactStore, `this.update` called from Store will call ReactStore's `update` - and this is not what we want here (as it will run interceptors again).

This ensures that the base class implementation is called.

packages/utils/src/store/ReactStore.ts
michaldudak: Menus can be nested within other menus, menubars, etc. Each level of the menu has its own store. In some cases, when a menu is nested, the root menu's state should take precedence over the local state (for example, if the menubar is disabled, the menu within must be disabled as well).

This code allows me to subscribe to the parent store's changes as if they were local state updates (no matter how many levels deep the menu is). Thanks to interceptors, I can also set the state in the parent menus (as in https://github.com/mui/base-ui/pull/3022/files#diff-b9e964ca92468413a6386a3abbb38d86ccc231ff12fa1ca3c3919de0d2fdc443R80-R89)

packages/utils/src/store/ReactStore.ts
michaldudak: It required updating tsconfig to use ES2022, but IMO it's safe to expect people will use browsers that support it (cc @mui/infra)

Janpot: If the browserslist configuration covers the API then it should be good to go. We're not polyfilling anything during the build. Opening https://github.com/mui/mui-public/issues/831. Please verify this API against the browserslist yourself for now.

michaldudak: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#browser_compatibility and https://caniuse.com/mdn-javascript_builtins_object_hasown, the only problematic browser is the KaiOS one. I'm not sure why we even support it - it has 0.02% global usage share (2.5 only) and it relies on Gecko 48 (from 2016), so I'm quite sure it might have issues with other APIs we use.

packages/utils/src/store/ReactStore.ts
romgrk: I don't like this new concept because it introduces complexity in a piece of code that I'd like to keep as simple as possible. Isn't there another way to solve that problem? Could a React context be used to store the first parent menu's store?

If a value needs to depend on either a parent or local store, I'd rather we subscribe to both stores (possibly with a custom hook to make it ergonomic) rather than add `notifyAll()`, which feels a bit hacky to me.

packages/utils/src/store/ReactStore.ts
Janpot: Just for future reference:
- We're not polyfilling APIs, only downleveling syntax
- Browserlist is our source of truth for supported platforms
- Tsconfig setting doesn't affect compilation, only type checking, AFAIK there is no easy way to populate it from browserslist to protect us from using unsupported APIs. But I'd expect 2022 to be mostly covered.

packages/utils/src/store/ReactStore.ts
michaldudak: I can't use React context here, as with detached triggers, the root element can be outside of the parent's React tree:
```tsx
<Menubar.Root>
  <Menu.Trigger handle={myMenu} />
</Menubar.Root>
<Menu.Root handle={myMenu} />
```

The Store instance (within the `myMenu` handle) is the only point of communication between the Trigger and Root.

I can create a Base UI-specific subclass of ReactStore that will handle these cases instead if you don't want it in the shared code. But the interceptors I introduced here can also be used to trigger change callbacks automatically (which could be useful for controlled props), so perhaps it's worth having this in the shared code. For raw performance without any bells and whistles, pure Store can always be used.

romgrk: Could we solve that use case with existing store effects instead of write interceptors? From the use-case in your link above, essentially what I see is "if this value changes, do that", which is already covered by existing API. And for values that depend on both local and parent store, because the parent store is also in the state, you can write selectors like this:

```js
const disabledSelector = createSelector(
  state => state.parent.store?.state.disabled ?? state.disabled
)
```

And we could add just an API to link stores' reactivity together, [something like this](https://github.com/mui/base-ui/pull/3022/files#diff-b9e964ca92468413a6386a3abbb38d86ccc231ff12fa1ca3c3919de0d2fdc443R103-R106), but I think `notifyAll()` doesn't need to change its state ref, it just needs to call the listeners. With selectors like the one above, it would work.

michaldudak: This was my initial implementation, but the tests proved it to be incorrect. React relies on the identity of the snapshot used by `useSyncExternalStore`. If it doesn't change, it assumes there are no changes within it.

romgrk: Alright for the state change then, but that's just a minor part of the comment, let me know what you think about the rest of it.

michaldudak: What would it look like? For example, the use case from https://github.com/mui/base-ui/pull/3022/files#diff-b9e964ca92468413a6386a3abbb38d86ccc231ff12fa1ca3c3919de0d2fdc443R85-R93 - updating the parent store's value if it exists? If I subscribe to `allowMouseEnter` changes and update the parent store after it changes, wouldn't it trigger another render? If there's a way of doing this efficiently and conveniently, I won't insist on my solution.

One other thing interceptors could allow us to do is to wait for the external callback's result before setting the value (so not just "if this value changes, do that", but "before this value changes, do that"):

```tsx
open: (value: boolean, store) => {
  const event = { ... };
  onOpenChange(event);
  if (event.baseUIDefaultPrevented) {
    return store.state.open;
  }
  return value;
}
```

But it's purely theoretical at this point - we don't need it now, so we can ignore it.

romgrk: >  If I subscribe to allowMouseEnter changes and update the parent store after it changes, wouldn't it trigger another render

I don't think it would. The store's update notifies all its listeners, one of which would update the parent store, which would notify its listeners. Both of those may schedule a re-render in React, but that doesn't happen synchronously (afaiu), both of those updates happen within a root `Store.setState()` call that doesn't yield to React. The Store is currently designed to run listeners (& effects) synchronously.

So yes I would just add a subscription to update the parent when `allowMouseEnter` changes.

michaldudak: If I understand it correctly, there's no way currently to be notified if just a specific part of the state changes, right? I'd have to keep the previous value of the `parent` myself and compare it with the new state coming from `subscribe` callback.

We could create something like this method:
```tsx
observe(selector, onChange) {
  let prevValue = selector(this.state);

  return store.subscribe((nextState) => {
    const nextValue = selector(nextState);
    if (!Object.is(prevValue, nextValue)) {
      prevValue = nextValue;
      onChange(nextValue);
    }
  });
}
```

michaldudak: Actually, this won't work like that - if we observe the result of selectors and the selector gets a value from a nested store, we won't ever fire the listener function as only the local state gets updated:

```tsx
const selectors = {
  count: (state: ChildState) => state.parent?.state.count ?? state.count,
  parent: (state: ChildState) => state.parent,
};

// ...

store.observe('count', (newCount) => {
  store.parent.set('count', newCount);
});

// ...

store.set('count', 42) // -> this won't cause the store.parent.set to execute
```

We'd have to have another function that observes the values of state fields, not selector return functions (https://github.com/mui/base-ui/blob/09904b55fddef4655f0923b733253a84d0d28d99/packages/utils/src/store/ReactStore.ts).

Alternatively, we can keep observing selectors' return values and set up two-way data syncing, as in https://github.com/mui/base-ui/blob/a713ccc9ff558864b7aaecb854c7da238d46ac26/packages/utils/src/store/ReactStore.ts and https://github.com/mui/base-ui/blob/a713ccc9ff558864b7aaecb854c7da238d46ac26/packages/utils/src/store/ReactStore.test.tsx#L253-L274. I like it less, though, as the setup is less straightforward and you have to make sure selectors don't read from nested stores.

romgrk: > We'd have to have another function that observes the values of state fields, not selector return functions

That's just because of the name collision in your selectors. The one you named `count` does more than what its name implies. You could add another selector, e.g. `localCount: state => state.count`, and then observing that one would make that snippet of code work properly.

I have some hesitations about calling selectors via strings, because that syntax may cause confusion with other APIs like `.set()` where the string refers to an actual field rather than a selector. In this part of your example, `'count'` refers to different things:

```js
store.observe('count', (newCount) => {
  store.parent.set('count', newCount);
});
```

michaldudak: It's because when `parent` is set, I want to completely ignore the local property (that's the case in Base UI). The caller doesn't really need to know if the value comes from the local state or another store. The `localCount` selector would exist only to set up a listener, and it would be invalid to use it from application code.

> I have some hesitations about calling selectors via strings, because that syntax may cause confusion with other APIs like .set()

I can see that, but on the other hand, it's more convenient to call `store.useState('foo')` than `store.useState(selectors.foo)` (plus it also prevents developers from creating ad-hoc selectors such as `store.useState((state) => state.foo)` as I mentioned in another discussion).

One way to make it clearer would be to rename methods that operate on selectors to contain the "selector" word: `useState -> useSelector`, `observe -> observeSelector`

packages/utils/src/store/ReactStore.ts
```diff
@@ -186,14 +215,21 @@ export class ReactStore<
    *
    * @param key Key of the selector to use.
    */
-  public useState<Key extends keyof Selectors>(key: Key): ReturnType<Selectors[Key]> {
+  public useState<Key extends keyof Selectors>(
+    key: Key,
+    ...args: SelectorArgs<Selectors[Key]>
+  ): ReturnType<Selectors[Key]> {
```
romgrk: Feels unsatisfying to put so much effort into avoiding `...args` with the `a1, a2, a3` stuff in `useStore()` just to end up using it again here :/

packages/utils/src/store/ReactStore.ts
```diff
@@ -126,6 +133,18 @@ export class ReactStore<
...
   }
 
+  /** Gets the current value from the store using a selector with the provided key. */
+  public select<Key extends keyof Selectors>(
+    key: Key,
+    ...args: SelectorArgs<Selectors[Key]>
+  ): ReturnType<Selectors[Key]> {
+    if (!this.selectors || !this.selectors[key]) {
+      throw new Error(`Base UI: Selector for key "${String(key)}" is not defined.`);
+    }
```
romgrk: The typings are already guarding this case, and this is an internal API for which we shouldn't need to baby-proof everything, so I would skip the runtime errors like this one.

packages/utils/src/store/Store.ts
```diff
@@ -78,7 +78,7 @@ export class Store<State> {
   public update(changes: Partial<State>) {
     for (const key in changes) {
       if (!Object.is(this.state[key], changes[key])) {
-        this.setState({ ...this.state, ...changes });
+        Store.prototype.setState.call(this, { ...this.state, ...changes });
```
romgrk, line 81: Is the prototype call still necessary?

packages/utils/src/store/ReactStore.ts
```diff
@@ -214,16 +250,73 @@ export class ReactStore<
...
     );
   }
+
+  /**
+   * Observes changes in the state properties and calls the listener when the value changes.
+   *
+   * @param key Key of the state property to observe.
+   * @param listener Listener function called when the value changes.
+   */
+  public observeState<Key extends keyof State>(
```
romgrk: I just realized that if we expose APIs that act on state keys as `*State`, the `setState` method becomes confusing. Tbh I'm not very enthusiast about having methods like this, I'd rather we keep one way of doing things: selectors. I would also prefer we invoke them without passing through strings - but if I have to pick one I'd rather have "_selectors as function or string_" over "_selectors and state keys as strings_". The "_selectors as function or string_" also lets you have selectors that aren't exposed in the `.selectors` interface, e.g. `.observe(internalSelectors.count)`.

packages/utils/src/store/ReactStore.ts
```diff
@@ -214,16 +250,73 @@ export class ReactStore<
...
+      store: this,
+    ) => void,
+  ) {
+    if (!this.selectors || !Object.hasOwn(this.selectors, key)) {
+      throw new Error(`Base UI: Selector for key "${key as string}" is not defined.`);
+    }
+
+    let prevValue = this.selectors[key]?.(this.state) as ReturnType<Selectors[Key]>;
+
+    listener(prevValue, prevValue, this);
```
romgrk, line 284: In the X version we also have [`useStoreEffect`](https://github.com/mui/mui-x/blob/master/packages/x-internals/src/store/useStoreEffect.ts) which does something similar but doesn't make an initial call. I wonder if we could match the behaviors so that we either always or never make an initial call. OTOH, this behavior matches `ResizeObserver` which has a similar name. No strong opinion.

packages/utils/src/store/Store.ts
michaldudak: Yes - it prevents an unnecessary call when `ReactStore.update` or `ReactStore.set` is called.
Without it the control flow looks like this (for `update`):
`ReactStore.update -> Store.update -> ReactStore.setState -> Store.setState`. 
Introducing it eliminates the `ReactStore.setState` call.

packages/utils/src/store/ReactStore.ts
michaldudak: OK, having both functions and keys seems like a good compromise

packages/utils/src/store/ReactStore.ts
michaldudak: I'd stick with this behavior as it's easier to filter out the initial call if you don't need it (check if `prevValue !== newValue`). And yes, this matches ResizeObserver and also React effects as they also fire on mount.

packages/utils/src/store/ReactStore.ts
michaldudak: Fair point, fixed. It required a bit of type gymnastics to get right, but should be pretty solid.

packages/utils/src/store/Store.ts
romgrk: Let's leave it but the controlled props thingy will need a refactoring/performance pass. The check for controlled props is made 2 times when calling `ReactStore.update`. Originally I aimed to make `.setState` the only method that mutates `.state`, the other methods being built on top of it. Also I've been wondering, is it really necessary to add a defensive check for controlled props? Would it be possible to add that check only in dev?

michaldudak: > The check for controlled props is made 2 times when calling ReactStore.update|

How so? `ReactStore.update` calls `Store.update` that in turn calls `Store.setState`.
`Store.setState` is the only place directly mutating state. All the other methods call it.

> Also I've been wondering, is it really necessary to add a defensive check for controlled props? Would it be possible to add that check only in dev?

It wouldn't work when `setState` is called from the outside. But perhaps we can make it protected? Do you have a need to completely replace the state?

Side note - this doesn't touch this PR directly. I'd appreciate it if we focused on what's being changed here so I can proceed with other PRs and continue the discussion in parallel.

packages/utils/src/store/ReactStore.ts
```diff
@@ -214,16 +235,63 @@ export class ReactStore<
...
+      store: this,
+    ) => void,
+  ): () => void;
+
+  public observeSelector<Selector extends ObserveSelector<State>>(
+    selector: Selector,
+    listener: (newValue: ReturnType<Selector>, oldValue: ReturnType<Selector>, store: this) => void,
+  ): () => void;
+
+  public observeSelector(
```
romgrk: If we agree to go with selectors as the only way to read the state, then we can rename this to just `.observe()`.

packages/utils/src/store/Store.ts
romgrk: > How so? 

[Here](https://github.com/mui/base-ui/blob/c2384f946b8976c55b27f5296bdc86e6a701b05d/packages/utils/src/store/ReactStore.ts#L153-L159) and [here](https://github.com/mui/base-ui/blob/c2384f946b8976c55b27f5296bdc86e6a701b05d/packages/utils/src/store/ReactStore.ts#L171-L177). But I guess with this change the 2 checks will be reduced to just one in `.setState()`.

> It wouldn't work when setState is called from the outside.

But the check is only there in case someones inadvertedly tries to update a controlled value right? Could we consider moving that check to typescript by declaring controlled props and removing them from all methods' typings except `useControlledProp`?

michaldudak: > Could we consider moving that check to typescript by declaring controlled props and removing them from all methods' typings except useControlledProp?

Not really, as we don't know which props are really controlled. We know which are "controllable," but it's impossible to know at the type level whether callers used a controlled or uncontrolled variant.

Issue #2798, opened 2025-09-19
[all components] Support `style` prop as a function


mrmntte:
# Feature request
Add `style` prop as a function support, like className={state => ...} but make it possible to return undefined.

## Examples in other libraries
**react-aria-components** has the following type signatures on their two props.
`className: string | (values: T & { defaultClassName: string | undefined }) => string`
`style: CSSProperties | (values: T & { defaultStyle: CSSProperties }) => CSSProperties | undefined`

## Motivation
I wanted to apply a dynamic value to the "style" property based on the state (like open) of an Autocomplete.Input.
I think making `style` accept an optional function would be more convenient and consistent with className.

PR #3038, opened 2025-10-22
[all components] Accepts style prop as function

mnajdova:
Closes https://github.com/mui/base-ui/issues/2798. Adds support for a style prop as function of the state. Recommend review commit per commit.

Added the new API in the styling docs as well:
https://deploy-preview-3038--base-ui.netlify.app/react/handbook/styling#style-prop


packages/react/src/utils/useRenderElement.tsx
```diff
@@ -198,6 +204,11 @@ export interface UseRenderElementComponentProps<State> {
     | undefined
     | ComponentRenderFn<React.HTMLAttributes<any>, State>
     | React.ReactElement<Record<string, unknown>>;
+  /**
+   * The style to apply to the rendered element.
+   * Can be a CSS propreties object or a function that accepts the state and returns a CSS properties object.
```
aarongarciah:
```suggestion
   * Can be a CSS properties object or a function that accepts the state and returns a CSS properties object.
```

docs/src/app/(public)/(content)/react/handbook/styling/page.mdx
```diff
@@ -47,6 +47,20 @@ Components expose CSS variables to aid in styling, often containing dynamic nume
...
 
+### Style prop
+
+Components that render an HTML element accept a `style` prop to style the element with a CSS object.
+
+```tsx title="switch.tsx"
+<Switch.Thumb style={{ height: '100px' }} />
+```
+
+The prop can also be passed a function that takes the component's state as an argument.
```
michaldudak: This reads better IMO:
```suggestion
The prop also accepts a function that takes the component's state as an argument.
```

packages/react/src/utils/types.ts
```diff
@@ -69,6 +69,11 @@ export type BaseUIComponentProps<
   render?:
     | ComponentRenderFn<RenderFunctionProps, State>
     | React.ReactElement<Record<string, unknown>>;
+  /**
+   * Style applied to the element, or a function that
+   * returns a style object based on the component’s state.
```
aarongarciah, line 74: This description uses the term "style object" but the `useRendereElement.tsx` file uses the "CSS properties object" term. Can we have a single term for it?

packages/react/src/utils/useRenderElement.tsx
mnajdova:
```suggestion
   * Can be a style object or a function that accepts the state and returns a style object.
```

mnajdova: Let's use style object, it's clear what it is already, no need to explain it.

aarongarciah:
@mnajdova not directly related to this PR but, do you have thoughts about this feedback from [Discord](https://discord.com/channels/1287292451308048406/1287295214804799499/1418679342296334336)?

> Out of curiosity, is there a particular reason why the function passed to className has to always return a string—even an empty one when I don’t want to apply anything?
Returning undefined feels more natural, since it clearly means “no class”, whereas "" is just an empty string.

The same applies to the `style` object.

mnajdova:
> @mnajdova do you have thoughts about this feedback from [Discord](https://discord.com/channels/1287292451308048406/1287295214804799499/1418679342296334336)?
> 
> > Out of curiosity, is there a particular reason why the function passed to className has to always return a string—even an empty one when I don’t want to apply anything?
Returning undefined feels more natural, since it clearly means “no class”, whereas "" is just an empty string.
> 
> The same applies to the style object.

I don't think there is a valid reason, it may be just type issue, let me check that quickly.

mnajdova:
@aarongarciah done in https://github.com/mui/base-ui/pull/3038/commits/c5c6504306b72460975f69af3f072078ebec1393, it was only about updating the types, but I added also tests to make sure we don't accidently break this behavior in the future.

aarongarciah:
@mnajdova thanks!

Issue #3007, opened 2025-10-17
Dialog: onOpenChange not firing on open when dialog is controlled


organdonor47:
# Bug report

## Current behavior

I am creating a `<Dialog />` component that can be optionally controlled. when the component is uncontrolled (i.e has a trigger button) `onOpenChange` gets called on open & close, but with a controlled Dialog, this only gets called when the dialog closes. 

(`onOpenChangeComplete`  _does_ get called correctly).

## Expected behavior

in a controlled dialog, `onOpenChange` should be called on every state change, not just when it closes

## Reproducible example

https://codesandbox.io/p/sandbox/zw3p2d

## Base UI version

1.0.0-beta.4

## Which browser are you using?

Chrome / Safari / Firefox

## Which OS are you using?

Mac OS

atomiks:
`onOpenChange` is only fired if the component changes the state internally for you. Since you're setting it to `open` externally, it doesn't get called (and doesn't need to be)

organdonor47:
> `onOpenChange` is only fired if the component changes the state internally for you. Since you're setting it to `open` externally, it doesn't get called (and doesn't need to be)

ok i see. just wondering why it fires on close then, but not on open? (thanks for the quick response)

atomiks:
In the example, the `true` state change is changed externally:

```js
<button type="button" onClick={() => setDialogOpen(true)}>
```

`onOpenChange` has no reason to fire here because you are setting the state yourself. It fires on close because the outside press/dismiss/close logic is an internal event update that gives your setter control of it, unlike the one above, since you're already in control of the change update

organdonor47:
right, got it. reading the docs it doesnt say anything about this behaviour, just that `onOpenChange` fires on open / close. I was intending to use it to handle overflow & other relevant app states myself, but i will just go the `useEffect` route instead then :)

atomiks:
I thought it'd be intuitive given controlled inputs work the same:

```jsx
<input value={value} onChange={(e) => setValue(e.target.value)} />
```

`onChange` on the input isn't called if you call `setValue()` externally. Though the native DOM attribute types have no description. It might be that every change handler's JSDoc description needs an update to clarify that it's not called if the value was changed externally, but closing this for now since it's not a bug

PR #3039, opened 2025-10-22
[dialog] Improve `onOpenChange` prop description

sai6855:
related to https://github.com/mui/base-ui/issues/3007

The current description of `onOpenChange` doesn’t clearly specify that it triggers only in `uncontrolled mode`. Updated the description to explicitly mention that it runs only in `uncontrolled mode`.

### DIFF
--- a/packages/react/src/alert-dialog/root/AlertDialogRoot.tsx
+++ b/packages/react/src/alert-dialog/root/AlertDialogRoot.tsx
@@ -69,7 +69,7 @@ export interface AlertDialogRootProps
   extends Omit<DialogRoot.Props, 'modal' | 'dismissible' | 'onOpenChange' | 'actionsRef'> {
   /**
    * Event handler called when the dialog's open state changes internally.
-   * Only fires in uncontrolled mode (when using `defaultOpen`).
+   * Only fires in uncontrolled mode (when the `open` prop is not provided).
    */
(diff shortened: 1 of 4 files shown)


packages/react/src/alert-dialog/root/AlertDialogRoot.tsx
```diff
@@ -68,7 +68,8 @@ export const AlertDialogRoot: React.FC<AlertDialogRoot.Props> = function AlertDi
 export interface AlertDialogRootProps
   extends Omit<DialogRoot.Props, 'modal' | 'dismissible' | 'onOpenChange' | 'actionsRef'> {
   /**
-   * Event handler called when the dialog is opened or closed.
+   * Event handler called when the dialog's open state changes internally.
+   * Only fires in uncontrolled mode (when the `open` prop is not provided).
```
atomiks, line 72: This isn't correct though. It fires whenever the state changes internally _but_ calls their setter as a way to give them control of the internal update. It just doesn't fire when changing the value externally, because there's no reason to.

I don't see why this needs to be documented because for example, controlled inputs work the same:

```js
<input value={value} onChange={(e) => setValue(e.target.value)} />
```

`onChange` on the input isn't called if you call `setValue()` externally.

Plus, this applies to every controlled handler, not only Dialog's `onOpenChange`

PR #3040, opened 2025-10-22
[tooltip] Keep state in a store

michaldudak:
Updates the internal implementation of Tooltip to use the ReactStore instance. This is required for detached triggers.

Part of https://github.com/mui/base-ui/issues/3032

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/tooltip/demos/hero/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/tooltip/demos/hero/css-modules/index.module.css
@@ -69,7 +69,7 @@
   }
 
   &[data-instant] {
-    transition-duration: 0ms;
+    transition: none;
   }
 
   @media (prefers-color-scheme: light) {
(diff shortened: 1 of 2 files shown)


atomiks:
`data-instant` animation sometimes glitches and is instant when it shouldn't be

First part of the recording is this PR, second part is main branch. It's kind of hard to catch so I wiggle the cursor when it happens for each (first is wrong, second is correct). It shouldn't be instant after pressing down on a trigger, and moving to a new one after 400ms (the provider timeout) has elapsed.


https://github.com/user-attachments/assets/f09aaf5d-8956-46cf-a054-52a7a0e941e2

atomiks:
This issue is back: https://github.com/mui/base-ui/pull/2962

atomiks:
There's an inconsistency with `[data-instant]` when pressing the trigger:

https://github.com/user-attachments/assets/6f53894b-8586-4959-94da-bb0497d906ac

michaldudak:
Approved implicitly with https://github.com/mui/base-ui/pull/3071

PR #3063, opened 2025-10-24
[accordion][collapsible] Add data-hidden attribute to collapsible and accordion when they are not open

BrandonGoren:

Adds a data-hidden attribute to collapsible and accordion when they are not open to match the behavior in tabs (which sets this attribute). 

Attached is a screenshot from an accordion component where data-hidden is now applied!

image

### DIFF
--- a/packages/react/src/collapsible/panel/CollapsiblePanel.tsx
+++ b/packages/react/src/collapsible/panel/CollapsiblePanel.tsx
@@ -124,9 +124,10 @@ export const CollapsiblePanel = React.forwardRef(function CollapsiblePanel(
   const panelState: CollapsiblePanel.State = React.useMemo(
     () => ({
       ...state,
+      hidden: !open,
       transitionStatus,
     }),
-    [state, transitionStatus],
+    [state, transitionStatus, open],
   );
 
   const element = useRenderElement('div', componentProps, {
@@ -158,6 +159,7 @@ export const CollapsiblePanel = React.forwardRef(function CollapsiblePanel(
 
 export interface CollapsiblePanelState extends CollapsibleRoot.State {
   transitionStatus: TransitionStatus;
+  hidden: boolean;
 }
 
 export interface CollapsiblePanelProps extends BaseUIComponentProps<'div', CollapsiblePanel.State> {
--- a/packages/react/src/collapsible/panel/CollapsiblePanel.tsx
+++ b/packages/react/src/collapsible/panel/CollapsiblePanel.tsx
@@ -6,12 +6,18 @@ import { BaseUIComponentProps } from '../../utils/types';
 import { useRenderElement } from '../../utils/useRenderElement';
 import { useCollapsibleRootContext } from '../root/CollapsibleRootContext';
 import type { CollapsibleRoot } from '../root/CollapsibleRoot';
-import { collapsibleStateAttributesMapping } from '../root/stateAttributesMapping';
+import { transitionStatusMapping } from '../../utils/stateAttributesMapping';
+import { panelOpenStateMapping } from '../../utils/collapsibleOpenStateMapping';
 import { useCollapsiblePanel } from './useCollapsiblePanel';
 import { CollapsiblePanelCssVars } from './CollapsiblePanelCssVars';
 import { useOpenChangeComplete } from '../../utils/useOpenChangeComplete';
 import type { TransitionStatus } from '../../utils/useTransitionStatus';
+import type { StateAttributesMapping } from '../../utils/getStateAttributesProps';
 
+const stateAttributesMapping: StateAttributesMapping<CollapsiblePanel.State> = {
+  ...panelOpenStateMapping,
+  ...transitionStatusMapping,
+};
 /**
  * A panel with the collapsible contents.
  * Renders a `<div>` element.
@@ -124,10 +130,9 @@ export const CollapsiblePanel = React.forwardRef(function CollapsiblePanel(
   const panelState: CollapsiblePanel.State = React.useMemo(
     () => ({
       ...state,
-      hidden: !open,
       transitionStatus,
     }),
-    [state, transitionStatus, open],
+    [state, transitionStatus],
   );
 
   const element = useRenderElement('div', componentProps, {
@@ -145,7 +150,7 @@ export const CollapsiblePanel = React.forwardRef(function CollapsiblePanel(
       },
       elementProps,
     ],
-    stateAttributesMapping: collapsibleStateAttributesMapping,
+    stateAttributesMapping,
   });
 
   const shouldRender = keepMounted || hiddenUntilFound || (!keepMounted && mounted);
@@ -159,7 +164,6 @@ export const CollapsiblePanel = React.forwardRef(function CollapsiblePanel(
 
 export interface CollapsiblePanelState extends CollapsibleRoot.State {
   transitionStatus: TransitionStatus;
-  hidden: boolean;
 }
 
 export interface CollapsiblePanelProps extends BaseUIComponentProps<'div', CollapsiblePanel.State> {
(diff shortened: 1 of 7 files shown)


mj12albert:
Maybe `data-hidden` is redundant? The `hidden` attr can be styled directly and we are already doing so: https://github.com/mui/base-ui/blob/e9f98ed126597668ace576d45904b31bad13b17c/docs/src/app/(public)/(content)/react/components/tabs/demos/hero/css-modules/index.module.css#L89-L91

BrandonGoren:
> Maybe `data-hidden` is redundant? The `hidden` attr can be styled directly and we are already doing so:
> 
> https://github.com/mui/base-ui/blob/e9f98ed126597668ace576d45904b31bad13b17c/docs/src/app/(public)/(content)/react/components/tabs/demos/hero/css-modules/index.module.css#L89-L91

When using Tailwind, passing a display class (like block or flex) to the collapsible body will override the built-in hidden attribute, since Tailwind’s utility classes take precedence in the final CSS.

By adding a data-hidden attribute, we can increase CSS specificity and ensure the “hidden” state is correctly applied, even when other display utilities are present.

BrandonGoren:
You can technically still style based on the hidden attribute in tailwind (and that's what I'm doing now), but it's a lot less intuitive:

```
[&[hidden]]:hidden
```
vs. 
```
data-[hidden]:hidden
```

mj12albert:
> in tailwind (and that's what I'm doing now), but it's a lot less intuitive

👍 makes sense ~ we have [`[&[hidden]:not([hidden='until-found'])]:hidden`](https://github.com/mui/base-ui/blob/e9f98ed126597668ace576d45904b31bad13b17c/docs/src/app/(public)/(content)/react/components/collapsible/demos/hero/tailwind/index.tsx#L11C42-L11C88) in our demo as well

packages/react/src/collapsible/panel/CollapsiblePanel.tsx
```diff
@@ -124,9 +124,10 @@ export const CollapsiblePanel = React.forwardRef(function CollapsiblePanel(
   const panelState: CollapsiblePanel.State = React.useMemo(
     () => ({
       ...state,
+      hidden: !open,
       transitionStatus,
     }),
-    [state, transitionStatus],
+    [state, transitionStatus, open],
```
mj12albert: `open` is already in the `...state`, we don't need to modify the state to add a new data attr based on it, instead modify the state<>data attr mapping here: https://github.com/mui/base-ui/blob/e9f98ed126597668ace576d45904b31bad13b17c/packages/react/src/utils/collapsibleOpenStateMapping.ts#L26-L35

BrandonGoren:
@mj12albert I updated the PR to update the data attribute states in the common functions. 

There are some different patterns in the repo for assembling these state attributes. I tried to match each collapsible/accordion to other parts of their "component" (e.g. the triggers and the roots), but it still felt a little awkward to merge these with the "parent" data-attributes in this way - let me know what you think!

packages/react/src/utils/collapsibleOpenStateMapping.ts
```diff
@@ -33,3 +33,21 @@ export const collapsibleOpenStateMapping = {
...
+
+export const panelOpenStateMapping = {
+  open: (open) => {
+    const panelHidden = !open;
+    const collapsibleStateMapping = collapsibleOpenStateMapping.open(open);
+
+    if (panelHidden) {
+      return {
+        ...collapsibleStateMapping,
+        [CollapsiblePanelDataAttributes.hidden]: '',
```
mj12albert, line 45: I think updating `PANEL_CLOSED_HOOK` is enough since 'closed' and 'hidden' always come together:
```js
const PANEL_CLOSED_HOOK = {
  [CollapsiblePanelDataAttributes.closed]: '',
  [CollapsiblePanelDataAttributes.hidden]: '',
};
```

packages/react/src/collapsible/panel/CollapsiblePanel.tsx
```diff
@@ -6,12 +6,18 @@ import { BaseUIComponentProps } from '../../utils/types';
...
+import { panelOpenStateMapping } from '../../utils/collapsibleOpenStateMapping';
 import { useCollapsiblePanel } from './useCollapsiblePanel';
 import { CollapsiblePanelCssVars } from './CollapsiblePanelCssVars';
 import { useOpenChangeComplete } from '../../utils/useOpenChangeComplete';
 import type { TransitionStatus } from '../../utils/useTransitionStatus';
+import type { StateAttributesMapping } from '../../utils/getStateAttributesProps';
 
+const stateAttributesMapping: StateAttributesMapping<CollapsiblePanel.State> = {
+  ...panelOpenStateMapping,
+  ...transitionStatusMapping,
```
mj12albert, line 19: The `transitionStatusMapping` can be added directly to `collapsibleOpenStateMapping` since it's the same for both panel components

mj12albert (commented):
@BrandonGoren I just remembered that internally `hidden` does not 100% correspond to `open`: https://github.com/mui/base-ui/blob/2b4e9c2818428ca9f2d6c4173de6e353d6b53885/packages/react/src/collapsible/panel/useCollapsiblePanel.ts#L50-L56

Instead I think `props.hidden` that `useCollapsiblePanel` returns needs to be added to the state in both panel components

michaldudak:
# PR review

There is one merge-blocking correctness issue: `data-hidden` is derived from `open`, not from the actual DOM hidden state, so it appears while animated panels are still visible. I also found one missing coverage gap and one formatting check failure. Focused jsdom tests pass, but the narrow Prettier check fails.

## Bugs (1)

### 1. 🔴 `data-hidden` is set before the panel is actually hidden

**Location:** `packages/react/src/utils/collapsibleOpenStateMapping.ts:42`

```ts
if (panelHidden) {
  return {
    ...collapsibleStateMapping,
    [CollapsiblePanelDataAttributes.hidden]: '',
  };
}
```

This maps `data-hidden` from `!open`. During CSS exit transitions, `open` becomes `false` immediately, but `useCollapsiblePanel` intentionally keeps the DOM `hidden` prop false until the exit animation finishes. The same issue exists in `AccordionPanel.tsx`, where the custom mapping also uses `!open`.

**Failure scenario:** A user styles `[data-hidden] { display: none; }` or otherwise treats it as “actually hidden”. Closing a Collapsible or Accordion with an exit transition applies `data-hidden` immediately, so the exit animation is skipped even though the panel is still visible and not yet `hidden`.

**Fix:** Derive `data-hidden` from the actual computed `hidden` value used by `useCollapsiblePanel`, not from `open`. Add `hidden` to the panel state or return it from the hook, then map that state key to `data-hidden`.

## Tests (1)

### 1. 🟡 New `data-hidden` behavior is not covered

**Location:** `packages/react/src/collapsible/panel/CollapsiblePanel.test.tsx:39`

```tsx
expect(screen.queryByText(PANEL_CONTENT)).to.have.attribute('data-closed');
```

The PR adds a public data attribute for both Collapsible and Accordion panels, but no test asserts when it is present or absent. Existing tests still pass even with the transition-state bug above.

**Failure scenario:** `data-hidden` can be applied while a panel is still visible, or omitted when it is actually hidden, and the current suite will not catch it.

**Fix:** Add panel tests for closed/open states, plus a transition case where `data-hidden` stays absent during the exit animation and appears only once the DOM `hidden` prop is applied.

## Simplifications (1)

### 1. 🟡 Run Prettier on the new Collapsible enum member

**Location:** `packages/react/src/collapsible/panel/CollapsiblePanelDataAttributes.ts:23`

```ts
hidden = 'data-hidden'
```

`pnpm prettier --check ...` fails on this file because the enum member needs the repo’s formatting.

**Failure scenario:** The PR fails the required formatting check before merge.

**Fix:** Run Prettier or add the trailing comma.

## Docs (0)

No findings.

## Verdict

**Request changes** — `data-hidden` currently ships a broken public contract for animated panels.

PR #3066, opened 2025-10-26
[composite] Avoid skipping missing elements in `listRef`

atomiks:
Tweak to https://github.com/mui/base-ui/pull/2604

Because of virtualization, missing elements in the `listRef` array shouldn't be skipped when determining the active index. In the virtualized demo, pressing <kbd>ArrowUp</kbd> lands on the 31st item (last item in the virtual window) instead of the 10,000th item.

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useListNavigation.test.tsx
+++ b/packages/react/src/floating-ui-react/hooks/useListNavigation.test.tsx
@@ -978,16 +978,27 @@ describe('useListNavigation', () => {
 
     await act(async () => {});
 
-    expect(screen.getByRole('textbox')).toHaveFocus();
+    const input = screen.getByRole('textbox');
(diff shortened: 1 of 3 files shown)


atomiks:
@LukasTy you sure? It happens for me on master

https://github.com/user-attachments/assets/e029c267-7c1b-44fe-b1cb-be35e77ca85b

LukasTy:
@atomiks Yes, on Chrome running MacOS 18 it works like this:

https://github.com/user-attachments/assets/9661a817-04da-47fe-b4d4-0d0c48180386

Also tested Safari and Firefox... 🤷

atomiks:
Right, the bug occurs if you focus the input without clicking it, so it doesn't open yet

LukasTy:
> Right, the bug occurs if you focus the input without clicking it, so it doesn't open yet

Whoops. My bad, I skipped that part. 🙈 
Yup, looks great then. 👍

PR #3067, opened 2025-10-26
[combobox] Fix input value derivation on value updates

### LINKED ISSUE #3064: [combobox] Combobox value does not appear in Combobox.Input when options arrive late
# Bug report

## Current behavior

1. When I initially render a `Combobox.Root`, provide a value, and empty items array, noting is shown in `Combobox.Input` value.
if i render: `<Combobox.Root value="Banana" items={[]}>` the `Combobox.Input` value will be: `""`
This is seems correct.

2. When I then re-render the Combobox, keep the same value, and provide items array with one of the items matching, the `Combobox.Input` value will not show, and the item will not be highlighted in the `Combobox.List` items.
if i re-render: `<Combobox.Root value="Banana" items={["Apple", "Banana", "Bread"]}>` the `Combobox.Input` value will be: `""`
This is incorrect.
When the Input is focused and blurred, the value will correctly update. This should not be required.

This is a real scenario better described here: https://discord.com/channels/1287292451308048406/1431125297201418260
TLDR: APIs that return options for Combobox and Select components often provide them late, after they first render. This often results in presenting a Combobox with a spinner or some kind of disabled / loading state. After this state finishes, this bug results in Combobox showing empty to the user.

https://github.com/user-attachments/assets/f365fc33-e8ed-4ea6-8684-4fa926098d80

## Expected behavior

`Combobox.Input` value should **react** to changes in items array.

## Reproducible example

https://codesandbox.io/p/sandbox/r2h8ry

## Base UI version

1.0.0-beta.4

Chrome

## Which OS are you using?

Mac OS

atomiks:

Fixes #3064

The input value should be reactive to external value changes; previously it was determined only once on mount

https://codesandbox.io/p/sandbox/ecstatic-wozniak-forked-nw9znd

Todo:

- [x] React to `items` prop changing, not just `value`

### DIFF
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -551,30 +551,6 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
     },
   );
 
-  const syncSelectedIndex = useStableCallback(() => {
-    if (selectionMode === 'none') {
-      return;
-    }
-
-    const registry = items ? flatItems : allValuesRef.current;
-
-    if (multiple) {
-      const currentValue = Array.isArray(selectedValue) ? selectedValue : [];
-      const lastValue = currentValue[currentValue.length - 1];
-      const lastIndex = findItemIndex(registry, lastValue, isItemEqualToValue);
-      setIndices({ selectedIndex: lastIndex === -1 ? null : lastIndex });
-    } else {
-      const index = findItemIndex(registry, selectedValue, isItemEqualToValue);
-      setIndices({ selectedIndex: index === -1 ? null : index });
-    }
-  });
-
-  useIsoLayoutEffect(() => {
-    if (!open) {
-      syncSelectedIndex();
-    }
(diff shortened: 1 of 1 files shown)


packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -857,6 +689,183 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
...
+    updateValue(selectedValue);
+
+    if (selectionMode === 'single' && !hasInputValue && !inputInsidePopup) {
+      const nextInputValue = stringifyAsLabel(selectedValue, itemToStringLabel);
+      const previousInputValue = stringifyAsLabel(previousSelectedValue, itemToStringLabel);
+
+      if (nextInputValue !== previousInputValue && inputValueRef.current !== nextInputValue) {
+        setInputValue(nextInputValue, createChangeEventDetails('none'));
+      }
+    }
```
atomiks, line 843: Actual diff (+ `items` hook below). I moved the effects down since this needs to access `setInputValue`, and all the similar types internal data are located next to each other intentionally (setters vs. effects, etc)

PR #3068, opened 2025-10-26
[combobox][autocomplete] Add `filteredItems` prop to improve virtualized demos with object items

atomiks:
This improves the `virtualized` demos when the item values are objects. Three issues were occurring before:

1. **Two passes of filtering was occurring**: Added a `filteredItems` prop to the root components, which overrides the internal filtering. Two passes of filtering was occurring (external + internal) and you needed to pass `filter={contains}` to sync the internal and external filtering together. This reduces it to one pass and removes this need to sync the filtering.
2. **Object items were serialized slowly**: This was the biggest cause of slowness - when using objects in a virtual list, `JSON.stringify()` serialization was being performed on every single item, which made filtering slow. Now the `useFilter` hook has the third optional param `itemToString` in its types (previously it only existed at runtime), which should be passed when externally filtering object items.
3. **Filtering work blocks typing**: `React.useDeferredValue()` can be used to improve typing performance while filtering work is ongoing

Now with 10,000 object items, even with 20x slowdown (M4 Pro), it's speedy:  

https://github.com/user-attachments/assets/c4128038-1a58-477c-a266-7f5616170d45


packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -955,9 +960,7 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
     // Floating UI tests don't require `role="row"` wrappers, so retains the number API.
     cols: grid ? 2 : 1,
     orientation: grid ? 'horizontal' : undefined,
-    disabledIndices: virtualized
```
mnajdova, line 1021: Why was this logic changed?

atomiks: It's not actually necessary, but I was debugging an issue with the virtualized demos which I PR'd here: https://github.com/mui/base-ui/pull/3066

mnajdova (approved):

PR #3071, opened 2025-10-27
[tooltip] Support detached triggers

### LINKED ISSUE #3032: [tooltip] Support detached triggers
# Feature request

Similar to https://github.com/mui/base-ui/pull/2974, but for tooltips

michaldudak:
Implemented detached / multiple trigger support in Tooltip, similarly to https://github.com/mui/base-ui/pull/2336

It is now possible to define triggers outside of Tooltip.Root using the handle prop. Multiple triggers per dialog with a dynamic payload are also implemented.

Delay groups created by Tooltip.Provider now apply to triggers (so when using detached triggers make sure the Tooltip.Trigger components are inside the Provider).

### ⚠️ Breaking changes

`delay` and `closeDelay` were moved from Tooltip.Root to Tooltip.Trigger, as they can be different per trigger.

Playground: https://deploy-preview-3071--base-ui.netlify.app/experiments/tooltip/tooltips
Docs: https://deploy-preview-3071--base-ui.netlify.app/react/components/tooltip#detached-triggers

Closes #3032
Built on top of #3040

### DIFF
--- a/packages/react/src/tooltip/root/TooltipRoot.test.tsx
+++ b/packages/react/src/tooltip/root/TooltipRoot.test.tsx
@@ -1157,6 +1157,7 @@ describe('<Tooltip.Root />', () => {
 
     it('should open the tooltip with any trigger on hover', async () => {
       const testTooltip = Tooltip.createHandle();
+      const popupId = randomStringValue();
       const { user } = await render(
         <div>
           <Tooltip.Trigger handle={testTooltip} delay={0}>
@@ -1172,7 +1173,7 @@ describe('<Tooltip.Root />', () => {
           <Tooltip.Root handle={testTooltip}>
             <Tooltip.Portal>
               <Tooltip.Positioner>
-                <Tooltip.Popup>Tooltip Content</Tooltip.Popup>
+                <Tooltip.Popup data-testid={popupId}>Tooltip Content</Tooltip.Popup>
               </Tooltip.Positioner>
             </Tooltip.Portal>
           </Tooltip.Root>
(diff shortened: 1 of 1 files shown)


atomiks:
Super cool 🙌 

Screenshot 2025-10-28 at 10 56 11 pm

Does it look better if there isn't any whitespace padding that clips the content? It seems like the content should have padding instead of the popup?

---

Also imo with the animation demo at the bottom, it'd be good to show varying widths. I remember it looked weird before (with the Popover one) if they were different, but it's more likely in the real world for the widths to vary quite a bit. 

If the transition transform amount needs to be manually adjusted for it to look good, ideally the demo should show that. (At least until we figure out an automatic method, maybe sizing CSS variables for the animation is needed)

---

When using the keyboard, the content animations play, but not the other popup animations

michaldudak:
You're right. I updated the animations demo styles.
The middle tooltip is now much shorter, and I think the content transition still looks good.
I also added `data-instant` to Viewport and Arrow so they don't transition unnecessarily.

docs/reference/generated/tooltip-root.json
```diff
@@ -23,11 +23,26 @@
       "description": "A ref to imperative actions.\n- `unmount`: When specified, the tooltip will not be unmounted when closed.\nInstead, the `unmount` function must be called to unmount the tooltip manually.\nUseful when the tooltip's animation is controlled by an external library.",
       "detailedType": "React.RefObject<Tooltip.Root.Actions> | undefined"
     },
+    "defaultTriggerId": {
+      "type": "string | null",
+      "description": "ID of the trigger that the dialog is associated with.\nThis is useful in conjunction with the `defaultOpen` prop to create an initially open dialog.",
```
atomiks: `tooltip`

atomiks (approved):
Wasn't able to find any issues aside from some JSDoc tweaks - animation example looks perfect

Issue #809, opened 2024-11-06
[core] Adopt react compiler?


oliviertassinari:
### Search keywords

-

### Latest version

- [X] I have tested the latest version

### Summary

Per https://youtu.be/qd5yk2gxbtg?t=1061&si=ZueAd7hJ3rvYQBod we might want to explore adopting the react compiler. 

The main question is does this even make sense? I suspect that we need to cherry-pick the optimizations that help and opt out of the harmful ones (extra bundle size, wasted CPU cycles, extra memory), using some sort of test app. Because we are almost exclusively leaf components, maybe this could be mostly harmful, not having much in the render tree that is pruned.

Another blocker is the overall community interest in this, which seems low: https://npm-stat.com/charts.html?package=babel-plugin-react-compiler&package=eslint-plugin-react-compiler for now.

The action plan would involve:

- [ ] Lint the codebase with eslint-plugin-react-compiler.
- [ ] Run all the tests on the output of the compiler. I believe we already do this with the Babel optimizations that we have, so won't be too hard. It's important to guarantee behavior.
- [ ] Publish the source with the output of the compiler.

### Examples

- https://github.com/TanStack/table/issues/5567
- https://github.com/TanStack/virtual/issues/736

### Motivation

Performance.

U-OK:
@oliviertassinari Hello! Is there any work planned on this issue?

LukasTy:
@U-OK There is a pending PR: https://github.com/mui/base-ui/pull/3076.

PR #3076, opened 2025-10-28
[WIP][code-infra] Setup react-compiler

brijeshb42:
Related to https://github.com/mui/base-ui/issues/809.

Enabled for both linting and building packages

The idea is first to explore the changes required at the repo level and then abstract it out as required to be part of `code-infra` package.

The bundle size increases significantly with the compiler output. This is something that we'll need to discuss.

### DIFF
--- a/babel.config.mjs
+++ b/babel.config.mjs
@@ -19,11 +19,21 @@ export default function getBabelConfig(api) {
         errorCodesPath,
       },
     ],
+    [
+      'babel-plugin-react-compiler',
+      /** @type {import('babel-plugin-react-compiler').PluginOptions} */ ({
+        compilationMode: 'infer',
+        // When targeting react v17 or v18, the packages also need to have
+        // "react-compiler-runtime" as a dependency.
+        target: '17',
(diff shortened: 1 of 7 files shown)


flaviendelangle:
Some of the bundle increase can probably be compensated by removing all the redundant `useMemo` / `useCallback` that are no longer needed.
But I doubt it will enough, 40% increase is massive :grimacing:

Janpot:
FYI: you can explore the output when following the links to the diff pages on https://frontend-public.mui.com/repository/mui/base-ui/prs/3076

> Some of the bundle increase can probably be compensated by removing all the redundant useMemo / useCallback that are no longer needed.

Compiler removes them, but maybe some code can be simplified if memoization is not a concern anymore?

atomiks:
This also seems to break list navigation entirely (`useListNavigation` hook) used in Menu, Select, Combobox, etc

LukasTy:
Maybe we should start with enabling this only for linting?
Looks like realistically it's far from reality.

PR #3081, opened 2025-10-28
[docs-infra] Upgrade Next.js to v16

dav-is:
Closes https://github.com/mui/base-ui/pull/3119.

Upgrade to Next.js version 16.

Preview deploy: https://deploy-preview-3081--base-ui.netlify.app/

We can't use turbopack yet because we [pass a function within a loader's options](https://github.com/mui/base-ui/blob/ad0190c7ead28f4d65bdc57c703c04d1c63681c5/docs/src/syntax-highlighting/index.mjs#L343).

We should be able to remove this when we migrate inline code to use the new code highlighter component in docs-infra.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -64,7 +64,7 @@
   },
   "devDependencies": {
     "@mdx-js/mdx": "^3.1.1",
-    "@mui/internal-docs-infra": "^0.2.3-canary.10",
+    "@mui/internal-docs-infra": "^0.2.3-canary.12",
     "@mui/internal-test-utils": "^2.0.15",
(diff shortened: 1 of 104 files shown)


docs/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference path="./.next/types/routes.d.ts" />
+import "./.next/dev/types/routes.d.ts";
```
dav-is: Running `next dev` makes this change. Causes a lint error, so I add this file to lint ignore

Issue #3041, opened 2025-10-22
Should `DirectionProvider` set `dir` attribute on portalled elements (e.g. `Tooltip.Popup`)?


benface:
# Feature request

## Summary

So if a part of my app is rendered in a right-to-left language, I should wrap that part with `<DirectionProvider direction="rtl">`, right? I would kind of assume that this handles setting the `dir` attribute of not just the element that `DirectionProvider` renders, but also of all the Base UI components in `DirectionProvider`'s React tree, especially those that get rendered in a portal and likely end up outside that element in the actual DOM (tooltips, popovers, menus, etc.), making them _not actually right-to-left_.

## Examples in other libraries

Not sure.

## Motivation

Not having to manually track the current direction to set `dir` on some components that could be deeply nested would be great.

Raghuboi:
Radix exports a DirectionProvider which takes a dir prop
https://www.radix-ui.com/primitives/docs/utilities/direction-provider

Ant Design exports a ConfigProvider which takes the dir prop
https://ant.design/components/config-provider

Meanwhile in React-aria the direction is embedded into their internationalization system
https://react-spectrum.adobe.com/react-aria/useLocale.html
https://react-spectrum.adobe.com/react-aria/I18nProvider.html

eg.

```
import { I18nProvider, useLocale } from 'react-aria-components';

function App() {
  let { locale, direction } = useLocale();
  
  return (
    <html lang={locale} dir={direction}>
      {/* your app here */}
    </html>
  );
}

<I18nProvider locale="ar-AE">
  <App />
</I18nProvider>
```

Raghuboi:
I could contribute to this

mj12albert:
I did notice this when testing RTL a while back, it is a bit annoying to have to manually set this on every nested positioner level: https://github.com/mui/base-ui/blob/master/docs/src/app/(private)/experiments/rtl.tsx#L32

However, setting the `dir` attribute is all or nothing though, and we decided to not set it to separate the HTML/CSS concerns from configuring the behavior (RA/Ariakit do not set `dir` either)

A slightly less annoying way I guess would be for your wrapping component e.g. `MyMenu.Root` to accept a `dir`/`direction` prop and forward it via a custom context to `MyMenu.Positioner` (or Trigger`), so you're just tracking the roots and not the nested parts 🤔

Raghuboi:
[...]

PR #3082, opened 2025-10-28
[direction provider] Provide `useDirection` hook

mj12albert:
Since we intentionally do not set the `dir` attribute, providing this hook seems the best way to handle the issue with portalled components and makes building custom wrapping components easier

Closes https://github.com/mui/base-ui/issues/3041

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 # IntelliJ IDEA module file
 *.iml
 *.log
-*.tsbuildinfo
 /.eslintcache
 /coverage
 /docs/.env.local
@@ -34,3 +33,7 @@ docs/public/llms-full.txt
 docs/public/react
 .cursor/rules/nx-rules.mdc
 .github/instructions/nx.instructions.md
+
+# typescript
+*.tsbuildinfo
+next-env.d.ts
--- a/docs/next-env.d.ts
+++ b/docs/next-env.d.ts
@@ -1,6 +1,0 @@
-/// <reference types="next" />
-/// <reference types="next/image-types/global" />
-/// <reference path="./.next/types/routes.d.ts" />
-
-// NOTE: This file should not be edited
-// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.


docs/src/app/(private)/experiments/rtl.tsx
```diff
@@ -1,5 +1,6 @@
 'use client';
-import { DirectionProvider } from '@base-ui-components/react/direction-provider';
+import clsx from 'clsx';
+import { DirectionProvider, useDirection } from '@base-ui-components/react/direction-provider';
```
mj12albert, line 3: Is this weird? This is just a single provider so `DirectionProvider.useDirection` can't work @atomiks

atomiks: It deviates from our other components, but given this is a utility hook, I think it's okay... but we can't add hooks for `Form`, or new components on it either without breaking changes (unless we add new exports), so it's a potential problem for v1

docs/src/app/(public)/(content)/react/utils/direction-provider/page.mdx
```diff
@@ -21,4 +21,26 @@ import { DirectionProvider } from '@base-ui-components/react/direction-provider'
...
 
+This only affects component interactions and behavior, you must additionally set `dir="rtl"` to affect HTML and CSS.
+
+## API reference
+
 <Reference component="DirectionProvider" />
+
+## useDirection
+
+Use this hook to read the current text direction. This is useful for wrapping portalled components that may be rendered outside your application root and are un-affected by the `dir` attribute set within.
```
atomiks: `unaffected`

docs/scripts/generateLlmTxt/__snapshots__/mdxToMarkdown.test.mjs.snap
```diff
@@ -504,6 +504,10 @@ import { DirectionProvider } from '@base-ui-components/react/direction-provider'
 </DirectionProvider>
 \`\`\`
 
+This only affects component interactions and behavior, you must additionally set \`dir="rtl"\` to affect HTML and CSS.
```
atomiks: Maybe:
```suggestion
`<DirectionProvider> enables child Base UI components to be aware of RTL text direction, but it does not affect HTML and CSS. The `dir="rtl"` HTML attribute or `direction: rtl` CSS style must be set additionally by your own application code.
```

docs/src/app/(private)/experiments/rtl.tsx
LukasTy: I don't think there is a problem with this approach for "Provider" style components, which have a slim chance of getting additional "parts". 🤔

LukasTy: WDYT about making the RTL a setting in this experiment? 🤔 
It could help with debugging and playing around.
```diff
diff --git a/docs/src/app/(private)/experiments/rtl.tsx b/docs/src/app/(private)/experiments/rtl.tsx
index ed96b1d7a..41a21c3e7 100644
--- a/docs/src/app/(private)/experiments/rtl.tsx
+++ b/docs/src/app/(private)/experiments/rtl.tsx
@@ -1,19 +1,31 @@
 'use client';
 import clsx from 'clsx';
-import { DirectionProvider, useDirection } from '@base-ui-components/react/direction-provider';
+import {
+  DirectionProvider,
+  TextDirection,
+  useDirection,
+} from '@base-ui-components/react/direction-provider';
 import { Menu } from '@base-ui-components/react/menu';
 import { Popover } from '@base-ui-components/react/popover';
 import { PreviewCard } from '@base-ui-components/react/preview-card';
+import {
+  SettingsMetadata,
+  useExperimentSettings,
+} from 'docs/src/components/Experiments/SettingsPanel';
 import c from './rtl.module.css';
 
-const dir = 'rtl';
+interface Settings {
+  rtl: boolean;
+}
 
 export default function RtlNestedMenu() {
+  const { settings } = useExperimentSettings<Settings>();
   const createHandleMenuClick = (menuItem: string) => {
     return () => {
       console.log(`Clicked on ${menuItem}`);
     };
   };
+  const dir: TextDirection = settings.rtl ? 'rtl' : 'ltr';
 
   return (
     <div className={c.rtl} dir={dir}>
@@ -189,3 +201,11 @@ function PreviewCardPositioner({ className, ...props }: PreviewCard.Positioner.P
     />
   );
 }
+
+export const settingsMetadata: SettingsMetadata<Settings> = {
+  rtl: {
+    type: 'boolean',
+    label: 'Right-to-left?',
+    default: true,
+  },
+};
```

LukasTy (approved):
Great work, the change makes sense to me. 👍

mj12albert:
> WDYT about making the RTL a setting in this experiment?

Was thinking to make a global setting in the experiments layout actually 😬

PR #3084, opened 2025-10-29
[docs][dialog] Uncontained example

atomiks:

"Uncontained" dialogs are used for lightboxes e.g. Instagram's image dialog, where elements can appear outside the main popup area (left/right arrows, close button)

Preview: https://deploy-preview-3084--base-ui.netlify.app/react/components/dialog#placing-elements-outside-the-popup

I don't think it's pertinent to create placeholder content here as it's just to demonstrate the necessary layout. Placeholder content would add needless complexity

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/dialog/page.mdx
+++ b/docs/src/app/(public)/(content)/react/components/dialog/page.mdx
@@ -172,6 +172,8 @@ import { DemoDialogInsideScroll } from './demos/inside-scroll';
 
 When adding elements that should appear "outside" the colored popup area, continue to place them inside `<Dialog.Popup>`, but create a child element that has the popup styles. This ensures they are kept in the tab order and announced correctly by screen readers.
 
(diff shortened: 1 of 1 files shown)


colmtuite:
Wouldn't a more literal heading be more intuitive to both humans + robots? Something like "Placing elements outside the popup"?

colmtuite:
Why not center the popup in the viewport?

atomiks:
> Wouldn't a more literal heading be more intuitive to both humans + robots? Something like "Placing elements outside the popup"?

Yeah, probably.

> Why not center the popup in the viewport?


It depends how the "outside" close button should appear on small screens, as there are overlapping issues (white close icon on top of the white popup). I've edited the styles a bit in a new commit, using media queries to position it better and keep the popup centered it in the viewport. 

Although it requires setting `pointer-events: none` on `Dialog.Popup` and then manually unsetting `pointer-events: auto` on the main popup and every "outside" element because otherwise it blocks clicks on the backdrop. Maybe you have a better layout example that avoids needing to do that?

Issue #3028, opened 2025-10-22
[docs] Move examples above API documentation


mnajdova:
# Docs feedback

## How can we improve Base UI documentation?

Move the examples to be above the API documentation, this way they would be more discoverable. The issue is that at this moment, the API information being the first thing means that people need to scroll a lot to get to the examples (especially on mobile). More over, the table of content can get too long on some pages, that you can't even see the examples in the table of content in order to know that they are there and be able to easily access them (even on desktop). There was one [feedback](https://github.com/mui/base-ui/issues/2933#issuecomment-3386116854) about this from the comunity too.

mj12albert:
I think the API reference should always come last, on certain components there are other sections of content like [Toast General usage](https://master--base-ui.netlify.app/react/components/toast#general-usage) which is quite important and should be un-missable

LukasTy:
> I think the API reference should always come last, on certain components there are other sections of content like [Toast General usage](https://master--base-ui.netlify.app/react/components/toast#general-usage) which is quite important and should be un-missable

Oh yeah, even if we kept the API reference "first" (before Examples), any sections explaining general concepts about the component should come before it. 🙈

PR #3088, opened 2025-10-29
[docs] Move API section last on the components page

mnajdova:
Closes https://github.com/mui/base-ui/issues/3028


docs/src/app/(public)/(content)/react/components/combobox/page.mdx
```diff
@@ -63,13 +63,6 @@ import { Combobox } from '@base-ui-components/react/combobox';
...
 ```
 
-## API reference
-
-<Reference
-  component="Combobox"
-  parts="Root, Value, Icon, Input, Clear, Trigger, Chips, Chip, ChipRemove, List, Portal, Backdrop, Positioner, Popup, Arrow, Status, Empty, Collection, Row, Item, ItemIndicator, Group, GroupLabel, Separator"
-/>
-
 ## useFilter
```
mj12albert, line 124: I think the useFilter section should follow the reference to the end as well ~

mnajdova: You mean having this order of the sections?

Hero
Usage guidelines
Anatomy
TypeScript inference
Examples
useFilter
API

Typically on the other pages the explanation and code samples are first, and they are all the examples, this is why I didn't change the order. Happy to revisit this tough. We have a hook documentation on the autocomplete and the toast pages too, and they are all following this order. cc @atomiks as he wrote the documentation on these components.

atomiks: `useFilter` is "part" of the API reference, but just a hook instead of a component. Our docs don't distinguish them properly though. I'd place it at the very end, after the API reference (same with e.g. Toast)

I would also change "Accessibility guidelines" to "Usage guidelines" as well so it's consistent on all pages

colmtuite (approved):

docs/src/app/(public)/(content)/react/components/combobox/page.mdx
mnajdova: > I would also change "Accessibility guidelines" to "Usage guidelines" as well so it's consistent on all pages

Would you add them after the examples? I thought it make sense for them to be before the examples as they explain something about the component.

atomiks: It's kept at the top intentionally as it's important to read. Only the hooks/API reference should go at the end

mj12albert: "[General usage](https://deploy-preview-3088--base-ui.netlify.app/react/components/toast#general-usage)" under Toast should also be changed to "Usage guidelines" as well (it's the only "General usage" heading we have)

atomiks: The second dot point isn't a guideline though, more of just how it can work

mnajdova: Ok, we are on the same page then regarding the order 👌 I agree around renaming "General usage" to "Usage guidelines". Yep, the second dot is an information, but I think it still fit in this section, I wouldn't move it.

mnajdova: I wonder if Global Manager and Stacking and animations shouldn't be subsection of the Usage guidelines.

LukasTy: I agree with keeping the General usage section as one.
I think the second point could be split into a new section that some components could use, namely, `Keyboard navigation`.

LukasTy (approved):
LGTM. 👍 
Should changes also apply to the [useRender](https://base-ui.com/react/utils/use-render) page? 🤔

PR #3092, opened 2025-10-29
[combobox][autocomplete] Keep focus on input when pressing list element

atomiks:

Ensures focus isn't lost from the input when pressing the list element with the pointer, in case it has the padding instead of the popup (as with changes in https://github.com/mui/base-ui/pull/3091)


LukasTy (approved):
Nice fix. 👍 
I've adjusted tests, feel free to change it though, but seems to make sense.

Also tested on Spectrum, it keeps the focus on input when clicking on scrollbar - the state of this PR. 👍

PR #3095, opened 2025-10-30
[avatar] Add `aria-hidden` to `AvatarFallback`

battaglr:

### Rationale

`AvatarFallback` is purely presentational and doesn’t convey meaningful information. Marking it with `aria-hidden="true"` prevents redundant or confusing announcements.

When an avatar is used inside an interactive element (e.g., a button or link), that element (I think) should provide an accessible name using `aria-label` or a similar attribute.

### Concerns / Questions

Are there cases where the avatar fallback is the *only* identifier for an entity (user, group, etc.)?  

If so, should `aria-hidden` be applied, depending on whether an `aria-label` is present?

### DIFF
--- a/packages/react/src/avatar/fallback/AvatarFallback.tsx
+++ b/packages/react/src/avatar/fallback/AvatarFallback.tsx
@@ -40,7 +40,10 @@ export const AvatarFallback = React.forwardRef(function AvatarFallback(
   const element = useRenderElement('span', componentProps, {
     state,
     ref: forwardedRef,
-    props: elementProps,
+    props: {
+      'aria-hidden': true,
+      ...elementProps
+    },
     stateAttributesMapping: avatarStateAttributesMapping,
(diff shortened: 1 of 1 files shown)


michaldudak:
I'm wondering why the fallback should be treated differently from the image? Should non-sighted users be concerned if the provided image didn't load?

@colmtuite, any thoughts?

battaglr:
@michaldudak My reasoning is that images add extra information while fallbacks (generally) don’t.

LukasTy:
I don't think we can apply this change by default globally.
It would make the fallback invisible to screen readers, even though it might convey useful information in case of a failure to load an image.
Also, misuse - in case someone is using only `<Avatar.Fallback>` explicitly for some reason. 🤔 
Just my take, but Radix also doesn't apply this prop. 🤷

battaglr:
That's a fair take!

Feel free to close this if that's the consensus.

Thanks for taking the time to consider this!

PR #3096, opened 2025-10-30
[toast] Add ability to anchor to an element

### LINKED ISSUE #3026: [toast] Add ability to anchor to an element
# Feature request

## Summary

Add an ability to anchor the toast to a certain element, e.g. copy button (shows a toast to indicate something has been copied to clipboard).

atomiks:
Closes #3026

Preview: https://deploy-preview-3096--base-ui.netlify.app/react/components/toast#anchored-toasts

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/toast/demos/anchored/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/toast/demos/anchored/css-modules/index.tsx
@@ -45,9 +45,6 @@ function CopyButton() {
   const buttonRef = React.useRef<HTMLButtonElement | null>(null);
 
   function handleCopy() {
-    // Simulate copying to clipboard
-    navigator.clipboard.writeText('Hello, world!');
-
     setCopied(true);
 
     anchoredToastManager.add({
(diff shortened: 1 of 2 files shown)


mj12albert:
Testing on iPhone - the anchored "copy" toast in the demo only dismisses on down/right swipes, I kind of expected to be more lenient and dismiss on a (outside) press 🤔

atomiks:
@mj12albert oops, the timeout was `10s` instead of `1.5s` in the CSS Modules demo - the brief time preriod for an anchored toast means you don't need a close button or outside tap to dismiss. In fact, that can lead to the toast disappearing when you don't want it to unintentionally. If you have a toast that's visible for awhile (>3-5s), then you should use `<Toast.Close>` which will alleviate that issue

mj12albert:
> the brief time preriod for an anchored toast means you don't need a close button or outside tap to dismiss

OK at 1.5s the behavior makes sense ~ maybe it shouldn't even be swipe-dismissable given the briefness?

packages/react/src/toast/arrow/ToastArrowDataAttributes.ts
```diff
@@ -0,0 +1,22 @@
+import { CommonPopupDataAttributes } from '../../utils/popupStateMapping';
+
+export enum ToastArrowDataAttributes {
+  /**
+   * Present when the anchor is hidden.
+   */
+  anchorHidden = CommonPopupDataAttributes.anchorHidden,
```
LukasTy, line 7: Is this attribute really possible on the Arrow part?
Looking at the code, it looks like only `Positioner` should be able to receive it. 🤷

atomiks: Looks like Tooltip also has this incorrectly documented; will need to update separately

PR #3097, opened 2025-10-30
[code-infra] Fix breaking change for ci publishing

brijeshb42:


scripts/README.md
```diff
@@ -45,6 +47,20 @@ A typical release goes like this:
 4. Refresh the page to see the newly created workflow, and click it.
 5. The next screen shows "@username requested your review to deploy to npm-publish", click "Review deployments" and authorize your workflow run. **Never approve workflow runs you didn't initiaite.**
 
+From terminal:
```
Janpot: Is there any reason why we would keep the manual way explained as well?

brijeshb42: Mainly to give both options since both are doing the same thing at the end.

Janpot: Let's document only one? To avoid confusion and to prevent us from having to maintain double documentation. Unless there is a good reason for when our devs would need to do use both in different occasions, but that need to be documented then.

PR #3100, opened 2025-10-30
[popups] Lock full-width touch popups

### LINKED ISSUE #3098: [Popover] Modal mode doesn't lock scroll on "touch" events
# Bug report

## Current behavior

When Popover component is used with "Modal" mode, it prevents a page from scrolling correctly for desktops.
But on mobile scrolling is still possible

<details><summary>Video 📽️</summary>
<p>

https://github.com/user-attachments/assets/5b9892ba-9a66-4d39-9037-ce966a79bb9d

</p>
</details> 

## Expected behavior

There should be a mode, that will lock scroll regardless device

## Reproducible example

[Sandbox](https://codesandbox.io/p/devbox/elegant-bassi-7n75hl?file=%2Fsrc%2FApp.tsx&workspaceId=ws_Ddd5rrbjzugd76Ymez2gb7)

## Base UI version

1.0.0-beta.4

Chrome 141

## Which OS are you using?

Windows 11

None

## Additional context

[Discord discussion](https://discord.com/channels/1287292451308048406/1287292451308048409/1433217154785869835)

atomiks:
Fixes #3098

Touch-opened popups intentionally skip scroll lock so swipe-outside dismissal still works, but that breaks modal popups that effectively fill the viewport. This change only re-enables scroll lock for the touch-open path when the positioned popup is effectively full-width, so smaller anchored popups keep the existing outside-swipe behavior.

## Changes

- Re-enable scroll lock for touch-opened Menu, Popover, Select, and Combobox popups only when the positioner width is within 20px of the full viewport width.
- Run the touch scroll-lock check from the positioners instead of the roots for internal consistency (+mount perf)

Preview: https://deploy-preview-3100--base-ui.netlify.app/react/components/select


atomiks:
Issue: on iOS, when scroll is activated (navbar expanded), this prevents scroll when swiping to dismiss

For the Popover use case, it seems like the structure in https://github.com/mui/base-ui/pull/3183 would work better (across all platforms/browsers)

atomiks (commented):
## Codex Review (GPT-5.4)

Reviewed against the refreshed full branch range from `upstream/master` to the current PR head. Dominant type: `bug-fix` with secondary `tests` and `react-ui`; the earlier validation gap is now addressed, and I did not find any remaining actionable issues in the amended branch.

### 1. Bugs / Issues (None)

The previously missing direct regression coverage for the new touch-lock path is now addressed. `PopoverRoot.test.tsx` and `ComboboxRoot.test.tsx` both add explicit Chromium assertions for touch-opened, effectively full-width popups, which closes the only branch-relevant gap I had in the prior pass.

Moving the touch-specific scroll-lock decision into the positioners still looks like the right fix. The width heuristic now lives next to the positioned element that owns `--available-width`, and the shared helper keeps the four updated call sites aligned without changing the non-touch path.

### Test Coverage Assessment

I ran `pnpm test:chromium MenuRoot --no-watch`, `pnpm test:chromium SelectRoot --no-watch`, `pnpm test:chromium PopoverRoot --no-watch`, `pnpm test:chromium ComboboxRoot --no-watch`, and `pnpm typescript`; all passed. With the added Popover and Combobox regressions, the amended branch now directly covers the touch-lock behavior across all four updated positioners.

### Recommendation

Approve ✅

The runtime change remains targeted, the newly added regressions close the earlier coverage hole, and the validation I ran is green on the current head.

flaviendelangle (approved):
# PR #3100 Review: [popups] Lock full-width touch popups

**PR:** https://github.com/mui/base-ui/pull/3100
**Author:** atomiks
**Fixes:** #3098

## Summary

Touch-opened popups intentionally skip scroll lock so swipe-outside dismissal still works, but that breaks modal popups that effectively fill the viewport. This PR re-enables scroll lock for the touch-open path only when the positioner width is within 20px of full viewport width. It also moves scroll lock logic from Root components to Positioner components for Menu and Popover.

## Overall Assessment

Well-structured PR that follows project conventions. The shared `useTouchOpenScrollLock` utility avoids logic duplication across 4 components (per AGENTS.md guidelines). The hook correctly uses `useIsoLayoutEffect`, `ownerDocument`, and the existing `useScrollLock` primitive. Tests are properly gated with `describe.skipIf(isJSDOM)`.

---

## Critical Issues (0)

None found.

---

## Important Issues (3)

### 1. Layout effect timing may read stale positioner dimensions

**File:** `packages/react/src/utils/useTouchOpenScrollLock.ts` (lines 18-32)

`useIsoLayoutEffect` fires synchronously after DOM mutations but before paint. At this point, `positionerElement.offsetWidth` reflects the element's *current* layout. However, Floating UI's positioning logic also runs in layout effects. The dependency array is `[enabled, touchOpen, positionerElement]` -- it does **not** include any signal that positioning has completed.

Race condition scenario:
1. Popup opens, `positionerElement` is set (triggers the effect)
2. Effect reads `offsetWidth`, but Floating UI hasn't applied final position/size yet
3. Width measurement is stale/zero, `touchOpenShouldLockScroll` = `false`
4. Floating UI applies final width, but the effect doesn't re-run (no dependency changed)

In practice, tests use explicit CSS widths (`style={{ width: 'calc(100vw - 10px)' }}`), which are applied before the layout effect, masking this timing issue. For dynamically-sized popups whose width depends on Floating UI computation, this could silently produce the wrong scroll lock decision.

**Suggestion:** Document the assumption that the positioner width must be deterministic at layout-effect time (e.g., set via CSS), or consider depending on a positioning-complete signal.

### 2. Reference element changed from positionerElement to triggerElement

**Files:** `PopoverRoot.tsx`, `MenuRoot.tsx` (removed), `PopoverPositioner.tsx`, `MenuPositioner.tsx` (added)

Previously, `useScrollLock` was called in the Root with `positionerElement` as the reference. Now `useTouchOpenScrollLock` passes `triggerElement` as `referenceElement` to `useScrollLock`. The reference element is used by `ScrollLocker` to resolve `ownerDocument` and `ownerWindow`.

If the trigger and positioner are in different documents (e.g., positioner rendered in a portal to a different iframe), scroll lock would target the wrong document. This is an edge case, but it's a silent behavioral change.

**Suggestion:** Verify this is intentional. Consider adding a brief comment on why `triggerElement` is preferred over `positionerElement`.

### 3. Inconsistent `enabled` conditions across call sites

| Component | `enabled` parameter |
|---|---|
| Combobox | `open && modal` |
| Menu | `open && modal && lastOpenChangeReason !== REASONS.triggerHover` |
| Popover | `open && modal === true && openReason !== REASONS.triggerHover` |
| Select | `(alignItemWithTriggerActive \|\| modal) && open` |

Combobox does not exclude `triggerHover`, while Menu and Popover do. Popover uses strict equality (`modal === true`) while others use truthy checks. These differences likely reflect legitimate behavioral differences, but the inconsistency increases the risk of a wrong condition at one call site.

**Suggestion:** Add a brief comment at each call site explaining why its `enabled` condition differs from the others.

---

## Suggestions (4)

### 1. Accidental blank line removal in PopoverPositioner.tsx

**File:** `packages/react/src/popover/positioner/PopoverPositioner.tsx`

A blank line was removed between the `useMemo` block and the `state` declaration. The other positioners keep this separator. Minor formatting inconsistency.

### 2. JSDoc is incomplete for the new hook

**File:** `packages/react/src/utils/useTouchOpenScrollLock.ts`

The JSDoc says *"Enables scroll lock for touch-opened popups when their positioned width is effectively viewport-sized"*, but the hook also handles the non-touch case by delegating to `useScrollLock(enabled && !touchOpen)`. It's a unified scroll lock entry point, not just a touch-open handler.

**Suggestion:** Expand to: *"Manages scroll lock for popup open states. For non-touch opens, scroll lock is applied when enabled. For touch opens, scroll lock is applied only when the positioner width is effectively viewport-sized."*

### 3. The 20px tolerance constant lacks rationale

**File:** `packages/react/src/utils/useTouchOpenScrollLock.ts`, line 9

`VIEWPORT_WIDTH_TOLERANCE_PX = 20` has no comment explaining why 20px was chosen. Does it account for scrollbar width? Mobile safe areas? Sub-pixel rounding?

**Suggestion:** Add a comment explaining what the tolerance accounts for.

### 4. Module-level comment conflates scroll lock with dismissal

**File:** `packages/react/src/utils/useTouchOpenScrollLock.ts`, lines 7-8

> "Touch-opened popups normally avoid scroll locking so users can still swipe outside to dismiss."

Dismissal is handled by `useDismiss`, not scroll lock. The actual reason is so users can still *scroll the page*.

**Suggestion:** Rewrite to: *"Touch-opened popups normally avoid scroll locking so the user can still scroll the page. This hook re-enables scroll lock only when the popup is effectively full-width."*

---

## Test Coverage

### What's covered
- Full-width touch popup applies scroll lock (all 4 components)
- Narrow touch popup does NOT apply scroll lock (all 4 components)
- Existing tests for mouse-open scroll lock preserved
- Existing tests for touch interaction type tracking preserved (Select)
- All new tests properly gated with `describe.skipIf(isJSDOM)`

### Gaps
- **No boundary test:** No test at exactly the 20px threshold (e.g., `calc(100vw - 20px)` and `calc(100vw - 21px)`) to verify the tolerance precisely
- **No scroll lock cleanup test:** No test verifying scroll lock is removed when the popup closes
- **No non-modal test:** No explicit test verifying non-modal touch popups are unaffected
- **Test constant coupling:** The test value `calc(100vw - 10px)` is coupled to the 20px constant but the relationship is not documented. A future change to the tolerance could silently break tests.

---

## Strengths

- Clean extraction of shared logic into `useTouchOpenScrollLock` -- avoids 4x duplication
- Correct use of project utilities (`useIsoLayoutEffect`, `ownerDocument`)
- Defensive guards (`viewportWidth > 0 && popupWidth > 0`) prevent false positives from unmounted elements
- Moving scroll lock to positioners is architecturally sound -- the positioner has direct access to the element needed for width comparison
- Adding `openMethod` to `MenuStore` is the right approach for making the value available in the positioner

atomiks:
> 1. Layout effect timing may read stale positioner dimensions

Yes, but not exactly relevant here as this is a simple heuristic where it's unlikely for a ~full-width popup to need to wait for measurement

Addressed the others

PR #3123, opened 2025-10-31
[test] Fix flaky tests

michaldudak:
Added `waitFor` to Tooltip and Popover tests

### DIFF
--- a/packages/react/src/popover/root/PopoverRoot.test.tsx
+++ b/packages/react/src/popover/root/PopoverRoot.test.tsx
@@ -580,12 +580,18 @@ describe('<Popover.Root />', () => {
 
           await user.tab({ shift: true });
 
-          expect(screen.getByRole('button')).toHaveFocus();
+          await waitFor(() => {
+            expect(screen.getByRole('button')).toHaveFocus();
+          });
 
-          expect(screen.queryByTestId('popup')).to.toBeVisible();
+          await waitFor(() => {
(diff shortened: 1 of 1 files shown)


packages/react/src/popover/root/PopoverRoot.test.tsx
```diff
@@ -580,12 +580,18 @@ describe('<Popover.Root />', () => {
...
 
-          expect(screen.getByRole('button')).toHaveFocus();
+          await waitFor(() => {
+            expect(screen.getByRole('button')).toHaveFocus();
+          });
 
-          expect(screen.queryByTestId('popup')).to.toBeVisible();
+          await waitFor(() => {
+            expect(screen.queryByTestId('popup')).toBeVisible();
+          });
```
LukasTy, line 589: Nitpick: this shouldn't be necessary. 🤔 
The above assertion already adds the "delay".

```suggestion
          expect(screen.queryByTestId('popup')).toBeVisible();
```

michaldudak: How so? AFAIK `findByTestId` is async, not `queryByTestId`.

LukasTy: I mean that the awaiting for the button (above) to have focus should have been enough to ensure that the popup is also open, since thee are no extra actions performed after the button assertion.
However, I'm not sure about the component specifics, maybe there is merit for this assertion due to how visibility is handled. 🤔

PR #3130, opened 2025-11-03
fix: dynamically measure combobox items

fredericoo:
This PR fixes a mislieading demo for combobox where it would not dynamically measure items and instead use the estimated height in a hardcoded way


docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/tailwind/index.tsx
```diff
@@ -114,7 +116,6 @@ export default function ExampleVirtualizedCombobox() {
                             top: 0,
                             left: 0,
                             width: '100%',
-                            height: `${virtualItem.size}px`,
```
fredericoo, line 117: if we hardcode the height then it can’t be measured correctly

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/tailwind/index.tsx
```diff
@@ -105,6 +105,8 @@ export default function ExampleVirtualizedCombobox() {
                         <Combobox.Item
                           key={virtualItem.key}
                           index={virtualItem.index}
+                          data-index={virtualItem.index}
```
atomiks, line 108: Curious why `data-index` is necessary?

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/tailwind/index.tsx
```diff
@@ -105,6 +105,8 @@ export default function ExampleVirtualizedCombobox() {
...
+													ref={virtualizer.measureElement}
```
atomiks: Formatting is incorrect

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/tailwind/index.tsx
atomiks: @fredericoo thanks for the PR. This should also be applied to the Autocomplete virtualized demo

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/tailwind/index.tsx
fredericoo: tanstack virtual uses it to map the measurements to an internal record of sizes and it uses the index, so this is necessary

omitting it leads to warnings in the console and measurements not updating correctly

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/tailwind/index.tsx
fredericoo: > @fredericoo thanks for the PR. This should also be applied to the Autocomplete virtualized demo

done!

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/tailwind/index.tsx
fredericoo: we _could_ modify base-ui to also have `data-index` added as part of the `index` prop, it would be practical and prevent setup errors, is that something that could be approved?

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/css-modules/index.tsx
```diff
@@ -114,7 +116,6 @@ export default function ExampleVirtualizedCombobox() {
                             left: 0,
                             width: '100%',
                             height: `${virtualItem.size}px`,
-                            transform: `translateY(${virtualItem.start}px)`,
```
atomiks, line 117: `height` should be removed instead

docs/src/app/(public)/(content)/react/components/autocomplete/demos/virtualized/css-modules/index.tsx
```diff
@@ -101,6 +101,8 @@ export default function ExampleVirtualizedAutocomplete() {
                         <Autocomplete.Item
                           key={virtualItem.key}
                           index={virtualItem.index}
+                          data-index={virtualItem.index}
+                          ref={virtualizer.measureElement}
```
atomiks, line 105: Is the `estimateSize` property unnecessary to specify with this?

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/css-modules/index.tsx
fredericoo: LOL i did it so automatically i didnt notice sorry

docs/src/app/(public)/(content)/react/components/combobox/demos/virtualized/tailwind/index.tsx
atomiks: We try to have fast combobox lists without needing virtualization for common use cases, and even tiny DOM operations like setting `data-index` can impact perf (with 1k items). I think manually specifying it for the virtualized demos is ok - plus TanStack isn't the only virtualized library you can use

docs/src/app/(public)/(content)/react/components/autocomplete/demos/virtualized/css-modules/index.tsx
fredericoo: no, it’s still necessary as you need a first frame with some items rendered and they use the estimated size

PR #3131, opened 2025-11-04
[form] Add `onFormSubmit` callback

mj12albert:
Preview: https://deploy-preview-3131--base-ui.netlify.app/react/components/form

Adds `onFormSubmit` that provides form values as a JS object as an argument, and `preventDefault`s the native submit event when used.

A lot of modern apps submit forms to backends like REST APIs, GraphQL, RPC that typically receive JS/JSON payloads instead of FormData, and there isn't a simple/robust way to easily convert them while accounting for numeric and/or array values ([example](https://github.com/mui/base-ui/blob/83d31d33d02331d3d34d787a9a327a9467b4b334/docs/src/app/(private)/experiments/forms/form.tsx#L79-L84)).


packages/react/src/form/Form.tsx
```diff
@@ -112,6 +130,11 @@ export const Form = React.forwardRef(function Form(
   return <FormContext.Provider value={contextValue}>{element}</FormContext.Provider>;
 });
 
+export type FormValues = Record<string, any>;
```
flaviendelangle: Did you consider adding a generic to the component (with `Record<string, any>` as the default)?

mj12albert: Good idea ~ there's the limitation that the `validate` prop on Field (that's called with form values as an arg) can't be aware of this, but doing something with submission is much more common so probably worth it

flaviendelangle: Yeah, if you want the individual field to be linked to the form in the typing, you need to do something like [Tanstack Form](https://tanstack.com/form/latest/docs/overview) but that's totally outside of the scope of this PR and would drastically change the approach :+1:

docs/src/app/(public)/(content)/react/components/form/page.mdx
```diff
@@ -30,6 +30,33 @@ import { Form } from '@base-ui-components/react/form';
 
 ## Examples
 
+### Submit form values as a JavaScript object
```
mj12albert, line 33: On second thought it's probably better to keep the hero demo using FormData so there's at least one example of that; also the form integration doc's demo would use `onFormSubmit` as it involves more data types

LukasTy: Sure, it's better being explicit about all the things we support. 👍

PR #3132, opened 2025-11-04
[slider][number field] Granular change reasons

atomiks:

Previously, all changes/commits had a `'none'` reason - this updates them to be more detailed which matches the popup-like components. Also updates generic events to have event narrowing as with change events, maintaining a `ReasonToEvent` map for all reasons we emit across components


mj12albert:
It's super useful to have this for differentiating pointer vs keyboard value changes when customizing a slider with `swap` 👍

packages/react/src/utils/createBaseUIEventDetails.ts
```diff
@@ -1,81 +1,100 @@
 import { EMPTY_OBJECT } from './constants';
 
+interface ReasonToEventMap {
+  none: Event;
```
mj12albert: I think it's worth constant-izing these strings now that there are so many and they're used quite a bit throughout the package

packages/react/src/utils/reason-parts.ts
```diff
@@ -0,0 +1,37 @@
+export const none = 'none' as const;
+
+export const triggerPress = 'trigger-press' as const;
```
LukasTy, line 3: Nitpick: I personally prefer seeing constants in uppercase,  but that's highly subjective. 👌 

```diff
-export const triggerPress = 'trigger-press' as const;
+export const TRIGGER_PRESS = 'trigger-press' as const;
```

atomiks: I do as well but in this case they're grouped into `REASONS`, so it's `REASONS.triggerPress`, while `REASONS.TRIGGER_PRESS` seems kind of unnecessary

LukasTy: Fair point. 👌

PR #3134, opened 2025-11-05
[code-infra] Include cli config files in typechecking

brijeshb42:
Also bumped to latest code-infra version.


scripts/tsconfig.json
```diff
@@ -2,9 +2,19 @@
...
+  "include": [
+    "**/*.ts",
+    "**/*.mts",
+    "../package.json",
+    "../eslint.config.mjs",
+    "../babel.config.mjs",
+    "../prettier.config.mjs",
+    "../.markdownlint-cli2.mjs",
+    "../stylelint.config.mjs",
+    "../vitest.*.mts"
```
brijeshb42: Added these to existing `scripts` tsconfig. Let me know if its not ideal. I didn't want to have one more tsconfig at the top-level.

scripts/api-docs-builder/src/componentHandler.ts
```diff
@@ -1,8 +1,8 @@
 import fs from 'fs';
 import path from 'path';
 import * as tae from 'typescript-api-extractor';
-import { formatProperties, formatEnum } from './formatter';
-import memberOrder from './order.json';
+import { formatProperties, formatEnum } from './formatter.js';
+import memberOrder from './order.json' with { type: 'json' };
```
brijeshb42: Doesn't work with `node16` and with `nodenext`, needs the `with` attribute.

scripts/api-docs-builder/src/hookHandler.ts
```diff
@@ -1,5 +1,5 @@
 import * as tae from 'typescript-api-extractor';
-import { formatProperties, formatParameters, formatType } from './formatter';
+import { formatProperties, formatParameters, formatType } from './formatter.js';
```
brijeshb42: `nodenext` needs file extensions. I figured there won't be a huge number of files in this package.

scripts/tsconfig.json
```diff
@@ -2,9 +2,19 @@
   "extends": "../tsconfig.base.json",
   "compilerOptions": {
     "lib": ["ES2021"],
-    "module": "preserve",
-    "moduleResolution": "bundler"
+    "module": "nodenext",
+    "moduleResolution": "nodenext"
```
brijeshb42: Since these scripts are executed directly without transpilation, `bundler` didn't make sense.

scripts/tsconfig.json
LukasTy: Shouldn't the previously present `"**/*.mjs"` cover these?
Or was it overly inclusive?

scripts/tsconfig.json
Janpot: I believe they are run with [`tsx`](https://github.com/mui/base-ui/blob/5daf2048d3c12b6564fad3623638f0c6fe1205c5/scripts/api-docs-builder/package.json#L14). Which uses `esbuild` under the hood, which resolves in a way compatible with `Bundler`. `nodenext` is stricter, so it's definitely not wrong, but if we only run this with `tsx`, then `bundler` should be sufficient.

scripts/tsconfig.json
brijeshb42: That would only cover the current directory (`scripts`) and not the parent folder.

scripts/tsconfig.json
brijeshb42: I removed all `tsx` with `node` invocations and everything worked without issues (except I had to replace all `.js` extensions with `.ts` in `scripts/api-docs-builder`. I just had to configure eslint to allow `.ts` imports.
So we can also remove all `tsx` cli calls as well. If others are onboard, I can do that as well.

scripts/tsconfig.json
LukasTy: Good point. What about `"../*.mjs", "../*.mts"` then? 🤔

scripts/tsconfig.json
Janpot: `tsx` comes with the convenience that we can execute code that was also meant for bundlers. e.g. we can import code from the docs or packages that doesn't work in native node (including .tsx files). I know we do this in some places in core.

I don't really see for the use-cases that we need it, what moving to native node.js brings us, it just locks us in a smaller box.

scripts/tsconfig.json
brijeshb42: Added it now. Its more future proof.

scripts/tsconfig.json
brijeshb42: Reverted to bundler. Makes sense to continue with `tsx` to access more features even though they are currently not being used.

brijeshb42:
Merging since the failing job is unrelated to the changes and already being fixed [here](https://github.com/mui/base-ui/pull/3138).

PR #3136, opened 2025-11-05
[form] Drop `onClearErrors` prop and always clear `errors` on change

mj12albert:
**Breaking:** Removed `onClearErrors` prop from `<Form>`

Demo: https://deploy-preview-3136--base-ui.netlify.app/react/components/form#submit-with-a-server-function

For all validation modes, it's a better UX to optimistically clear external (server) errors on change without needing to add `onClearErrors={setErrors}`
- Matches [RA](https://react-spectrum.adobe.com/react-aria/forms.html#server-validation:~:text=These%20are%20displayed%20to%20the%20user%20as%20soon%20as%20the%20validationErrors%20prop%20is%20set%2C%20and%20cleared%20after%20the%20user%20modifies%20each%20field%27s%20value)
- Makes `useActionState` viable, compare with [the preview](https://deploy-preview-2983--base-ui.netlify.app/react/components/form#submit-with-a-server-function) of this [PR](https://github.com/mui/base-ui/pull/2983), it would always require an extra state and effect 

Also frees up the API surface a bit if there's a need to customize revalidation more deeply in the future

Closes https://github.com/mui/base-ui/issues/2758


packages/react/src/form/Form.tsx
```diff
@@ -42,6 +41,14 @@ export const Form = React.forwardRef(function Form(
     }
   });
 
+  const [errors, setErrors] = React.useState(externalErrors);
+
+  React.useEffect(() => {
+    if (externalErrors) {
+      setErrors(externalErrors);
+    }
+  }, [externalErrors]);
```
atomiks: So this means we don't need to give them control of clearing errors when the field control changes, i.e. their errors object stays filled despite them being cleared internally by us

Also seems like the state could be set in render directly:

```jsx
if (externalErrors && errors !== externalErrors) {
  setErrors(externalErrors);
}
```

mj12albert: > Also seems like the state could be set in render directly

Tried and doesn't work since without the effect, when we update the errors internally with `setErrors(nextErrors)`, this will immediately set it back to the (now stale) `externalErrors` @atomiks 

> So this means we don't need to give them control of clearing errors

Yeah, useActionState reminded me of the redux-like flows that don't involve a lot of state, where you receive the "next error state" as a prop after doing the designated action (submit). Otherwise to use `errors` you must additionally create your own errors state plus remember to add `onClearErrors={setErrors}` which isn't really having control over the behavior anyway 😅 

I think this would also be clearer with some more extensive documentation on server errors, probably in the form integration page?

flaviendelangle: > Tried and doesn't work since without the effect, when we update the errors internally with setErrors(nextErrors), this will immediately set it back to the (now stale) externalErrors

You have to also store `props.events` in the state and check that it did change since last time :+1:

mj12albert: It works with `useValueChanged`!

atomiks: > Tried and doesn't work since without the effect, when we update the errors internally with setErrors(nextErrors), this will immediately set it back to the (now stale) externalErrors

Ah right... I wonder if instead of the effect you can derive it and set the internal `errors` state to `null` initially?

```jsx
const errors = internalErrors ?? externalErrors;
```

mj12albert: It didn't work either, since `internalErrors` needs to be `{}` after they are cleared but before re-submit, if it's null-ed then it'll fall through to `externalErrors` again

But when `internalErrors` is `{}` and after re-submit new (or the same) external errors are received, it'll stay on `{}` @atomiks

atomiks: `useValueChanged` seems good 👍

packages/react/src/select/root/SelectRoot.test.tsx
```diff
@@ -1229,11 +1229,11 @@ describe('<Select.Root />', () => {
 
     it('clears errors on change', async () => {
       function App() {
-        const [errors, setErrors] = React.useState<Record<string, string | string[]>>({
+        const [errors] = React.useState<Record<string, string | string[]>>({
           select: 'test',
         });
```
atomiks: Nit: should update these to just be non-stateful values

Issue #3027, opened 2025-10-22
[docs] Add more demos


mnajdova:
# Docs feedback

## How can we improve Base UI documentation?

Add more demos on the components to make it easier to showcase different use-cases.

- dialog
  - [x] dialog opens imperatively (we have a code-block)
  - [ ] move focus to another element after closing a dialog
  - [ ] reset dialog state on open/close
- popover
  - [x] open on hover
- menu
  - [x] menu checkbox items
  - [x] menu radio items
- accordion
  - [x] single
- collapsable
  - [ ] initially open collapsible
  - [ ] semi open state (e.g. codeblock that shows only first few lines)
- toggle
  - [x] multiple https://github.com/mui/base-ui/pull/4343
- tooltip
  - [ ] trigger with custom elements?
- scroll area
  - [ ] hard scrollbar, stable gutter, etc
  - [ ] vertical & horizontal scrollbar
  - [ ] overflow behaviour
- slider
  - [x] vertical https://github.com/mui/base-ui/pull/4346
  - [ ] steps
- field
  - [ ] field custom error
  - [ ] field custom element
  - [ ] input/checkbox/etc in field
- checkbox / checkbox group
  - [x] indeterminate checkbox https://github.com/mui/base-ui/pull/4345
  - [x] parent checkbox
- tabs
  - [x] tabs as links https://github.com/mui/base-ui/pull/4341
  - [ ] scrollable tabs?
  - [ ] animated tabs panels
- other
  - [ ] hidden until found
  - [ ] form with more validation examples
  - [ ] hide when detached
  - [ ] custom positioner

PR #3139, opened 2025-11-05
[dialog][docs] Add demo for custom focus management

mnajdova:
Part of https://github.com/mui/base-ui/issues/3027.

Preview: https://deploy-preview-3139--base-ui.netlify.app/react/components/dialog#custom-focus-management


atomiks:
I think it'd be worth changing this to be a demo that configures both `initialFocus` and `finalFocus`; also Popover might be a good one to add this for as well

mnajdova:
> I think it'd be worth changing this to be a demo that configures both initialFocus and finalFocus; also Popover might be a good one to add this for as well

Good idea, added 👌

atomiks:
This should be moved above the code block:
Screenshot 2025-11-25 at 9 38 26 pm

docs/src/app/(public)/(content)/react/components/dialog/page.mdx
```diff
@@ -152,6 +152,16 @@ import { DemoDialogCloseConfirmation } from './demos/close-confirmation';
...
 
+### Custom focus management
+
+You can control where the focus goes when the dialog opens and close using the `initialFocus` and `finalFocus` props on the `<Dialog.Popup>` component.
+
+import { DemoDialogFocusManagement } from './demos/focus-management';
+
+<DemoDialogFocusManagement compact />
+
+You can also set these props to `false` if you don't want to move the focus anywhere after opening/closing the dialog, or a function for specifying the target focusable element ref dynamically.
```
mnajdova:
```suggestion
You can also set these props to `false` if you don't want to move the focus anywhere after opening/closing the dialog, or a function for specifying the target focusable element ref dynamically.

import { DemoDialogFocusManagement } from './demos/focus-management';

<DemoDialogFocusManagement compact />
```

michaldudak:
### Code review

Found 2 issues:

1. **Outdated package imports** - The PR uses `@base-ui-components/react` but the package was renamed to `@base-ui/react`. All existing dialog demos on master use `@base-ui/react/dialog`.

https://github.com/mui/base-ui/blob/9b5ef259de0b751e161fe830d252fd3d66b596cd/docs/src/app/(public)/(content)/react/components/dialog/demos/focus-management/css-modules/index.tsx#L3-L5

https://github.com/mui/base-ui/blob/9b5ef259de0b751e161fe830d252fd3d66b596cd/docs/src/app/(public)/(content)/react/components/dialog/demos/focus-management/tailwind/index.tsx#L3-L5

2. **Typo in demo text** - "You feedback means a lot to us!" should be "Your feedback means a lot to us!" (appears in both CSS Modules and Tailwind versions)

https://github.com/mui/base-ui/blob/9b5ef259de0b751e161fe830d252fd3d66b596cd/docs/src/app/(public)/(content)/react/components/dialog/demos/focus-management/css-modules/index.tsx#L24-L26

flaviendelangle:
We need to update this demo to use the new doc design, or close it

michaldudak:
I updated the design. @colmtuite, @aarongarciah, would you mind taking a look?

aarongarciah (commented):
@michaldudak the UI looks good.

AFAIK in this use case, removing `initialFocusRef` has the same effect as having it. Maybe it's fine, as it showcases how to do deal with the initial focus, but ideally the demo would showcase a scenario where `initialFocusRef` achieves something that Dialog doesn't do by default.

colmtuite (approved):

flaviendelangle (approved):
# PR review

A clean docs-only PR adding a "Custom focus management" Dialog demo; nothing is merge-blocking (no 🔴) and CI is fully green. The two findings are demo-quality, both in the Docs section: the `initialFocus` half of the demo is invisible on desktop because it targets the element the default already focuses, and one sentence describes the function form imprecisely.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (2)

### 1. 🟡 `initialFocus` target coincides with the default, so it shows nothing on desktop

**Location:** `docs/src/app/(docs)/react/components/dialog/demos/focus-management/css-modules/index.tsx:24` (and `tailwind/index.tsx:24`)

```tsx
<Dialog.Popup initialFocus={initialFocusRef} finalFocus={finalFocusRef}>
  ...
  <Field.Control ref={initialFocusRef} placeholder="Enter your name" .../>  {/* first tabbable element */}
```

Per the `initialFocus` JSDoc on `DialogPopup.tsx`, the default behavior already "moves focus to the first tabbable element inside the popup" (except on touch). In this demo the first tabbable element *is* the "Full name" input — which is exactly what `initialFocusRef` points to. So on mouse/keyboard (i.e. anyone viewing the deploy preview on desktop) `initialFocus` produces no observable difference from the default; it only changes behavior on touch (forcing the input + virtual keyboard instead of focusing the popup). The section is titled "Custom focus management," but only the `finalFocus` half (focus jumps to the external "Final focus" button instead of returning to the trigger) is actually visible.

**Failure scenario:** A reader opens the demo on desktop, removes/toggles `initialFocus`, and sees no change — the prop the section is teaching appears to do nothing.

**Fix:** Point `initialFocus` at a non-first element so the effect is visible everywhere — e.g. attach `initialFocusRef` to the "Feedback" `Field.Control` (the second field) instead of "Full name", so focus visibly skips the first input on open.

### 2. ℹ️ Prose calls the function form a "ref"; it returns an element

**Location:** `docs/src/app/(docs)/react/components/dialog/page.mdx:167`

```mdx
You can also set these props to `false` if you don't want to move the focus anywhere after opening/closing the dialog, or a function for specifying the target focusable element ref dynamically.
```

The function form of `initialFocus`/`finalFocus` is `(interactionType) => boolean | HTMLElement | null | void` — it returns the element to focus (or `true`/`null`/`false`), not a ref. "the target focusable element ref" is slightly off.

**Failure scenario:** A reader writes a function returning a `RefObject` based on this sentence and is surprised it doesn't focus.

**Fix:** Reword to "…or a function that returns the element to focus, called with the interaction type" (or simply drop "ref").

## Verdict

**Approve after nits** — no correctness issues and CI is green; the `initialFocus` target just doesn't visibly demonstrate the prop on desktop, which is worth a small tweak before merge.

---

PR #3140, opened 2025-11-05
[docs][popover] Add `openOnHover` demo

mnajdova:

Preview https://deploy-preview-3140--base-ui.netlify.app/react/components/popover#opening-on-hover


docs/src/app/(public)/(content)/react/components/popover/page.mdx
```diff
@@ -37,6 +37,16 @@ import { Popover } from '@base-ui-components/react/popover';
 
 ## Examples
 
+### Opening on hover
+
+This example shows how you can configure the popover to open on hover using the `openOnHover` prop.
```
flaviendelangle, line 42: I think it's worth adding a callout explaining when this pattern should be used instead of a Tooltip 👀

docs/src/app/(public)/(content)/react/components/popover/page.mdx
```diff
@@ -37,6 +37,16 @@ import { Popover } from '@base-ui-components/react/popover';
...
+
+You can use the `delay` prop to specify how long to wait (in milliseconds) before the popover may be opened on hover.
```
LukasTy:
```suggestion
You can use the `delay` prop to specify how long to wait (in milliseconds) before the popover opens on hover.
```

docs/src/app/(public)/(content)/react/components/popover/page.mdx
mnajdova: We need to add section on each component in my opinion on when to use the component vs others. I will bring this up on Monday on the call, as it is not specific for this demo.

docs/src/app/(public)/(content)/react/components/popover/page.mdx
```diff
@@ -37,6 +37,16 @@ import { Popover } from '@base-ui-components/react/popover';
 
 ## Examples
 
+### Opening on hover
+
+This example shows how you can configure the popover to open on hover using the `openOnHover` prop.
+
+You can use the `delay` prop to specify how long to wait (in milliseconds) before the popover opens on hover.
```
LukasTy, line 44: Have you considered showcasing the `delay` prop as well?

LukasTy (approved):

docs/src/app/(public)/(content)/react/components/popover/page.mdx
mnajdova: I didn't want to change as the default value is kind of optimal when testing demos, I think it's pretty clear how people can use it.

PR #3141, opened 2025-11-05
[accordion] Change `multiple` prop to be false by default and add a demo

mnajdova:
Part of https://github.com/mui/base-ui/issues/3027.

PR #3154, opened 2025-11-06
[docs-infra] Use Docs-Infra for Markdown Demos Output

dav-is:
When generating markdown, we should use docs-infra's `loadCodeVariant()` to load demo files, so their output is the same as the docs site.

Without this we get missing files: https://github.com/mui/base-ui/pull/3141/files#r2495156213

Depends on https://github.com/mui/mui-public/pull/884 (merged)

[Note on async work in visitors](https://github.com/syntax-tree/unist-util-visit-parents/issues/8#issuecomment-1413405543)

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -21,7 +21,7 @@
     "@emotion/styled": "^11.14.1",
     "@mdx-js/loader": "^3.1.1",
     "@mdx-js/react": "^3.1.1",
-    "@mui/internal-docs-infra": "https://pkg.pr.new/mui/mui-public/@mui/internal-docs-infra@849b9fc",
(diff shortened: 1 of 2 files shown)


LukasTy:
Something is wrong since all the pages between `Checkbox` and `Separator` (inclusively) do not work.

dav-is:
@LukasTy what issue are you seeing? I'm not sure what you mean

LukasTy:
It's strange... Maybe I had some caching malfunction... 🤷 
All of those pages used to route to `<component>.txt` page (instead of `<component>`) and showed JS source code. 🤷

dav-is:
@LukasTy I think the *.txt files are RSC payload. The markdown is at *.md

PR #3155, opened 2025-11-06
[docs-infra] Preload Local Fonts

dav-is:
We should preload our [local fonts](https://github.com/mui/base-ui/blob/master/docs/src/fonts/index.css), so they start loading right away instead of after the first css file loads.

Before:

image

After:

image

Screenshot From 2025-11-07 08-40-56

image

Also seems to fix the issue where the page renders without any text:
image


mj12albert (approved):
Nice 👍

PR #3156, opened 2025-11-07
[scroll area] Improve CSS vars performance

atomiks:

This removes the inheritance of the scroll area CSS variables, which are expensive since they continually update on scroll. Instead, children need to opt in by specifying e.g. `--scroll-area-overflow-y-start: inherit` down the tree:

```css
  &::before,
  &::after {
    --scroll-area-overflow-y-start: inherit;
    --scroll-area-overflow-y-end: inherit;
```

The difference in rendering work is clear when scrolling a complex scroll area subtree (I copy-pasted the entire docs container three times inside the scroll area and measured rendering work while scrolling):

Screenshot 2025-11-07 at 1 51 04 am

Screenshot 2025-11-07 at 1 51 53 am

---

Great article about this CSS vars inheritance perf issues: https://motion.dev/blog/web-animation-performance-tier-list

Gradient fade example: https://deploy-preview-3156--base-ui.netlify.app/react/components/scroll-area#gradient-scroll-fade

---

**Breaking changes**:

The CSS variables are now on the `<ScrollArea.Viewport>` part, not `<ScrollArea.Root>`, which is where they'll typically be used to reduce how many `inherit` CSS properties need to be specified (e.g. on pseudo-elements), and  as stated, children no longer inherit the variables.

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/scroll-area/page.mdx
+++ b/docs/src/app/(public)/(content)/react/components/scroll-area/page.mdx
@@ -39,7 +39,45 @@ Use the viewport overflow CSS variables to fade the scroll edges, which graduall
 
 The CSS variables do not inherit by default to improve rendering performance in complex scroll areas with deep subtrees. To use them in child elements (or pseudo-elements on `<ScrollArea.Viewport>`), you must manually set each variable to `inherit`.
 
-<DemoScrollAreaScrollFade />
+```css title="scroll-area.module.css" {15,22}
+.Viewport {
+  &::before,
+  &::after {
+    content: '';
+    display: block;
+    left: 0;
+    width: 100%;
+    position: absolute;
+    pointer-events: none;
+    border-radius: 0.375rem;
+    transition: height 0.1s ease-out;
+  }
+
+  &::before {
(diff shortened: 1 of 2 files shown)


docs/src/app/(public)/(content)/react/components/scroll-area/page.mdx
```diff
@@ -29,6 +29,18 @@ import { ScrollArea } from '@base-ui-components/react/scroll-area';
...
+
+### Gradient scroll fade
+
+import { DemoScrollAreaScrollFade } from './demos/scroll-fade';
+
+Use the viewport overflow CSS variables to fade the scroll edges, which gradually increases in strength as the user scrolls away from the edges.
+
+The CSS variables do not inherit by default to improve rendering performance in complex scroll areas with deep subtrees. To use them in child elements (or pseudo-elements on `<ScrollArea.Viewport>`), you must manually set each variable to `inherit`.
+
+<DemoScrollAreaScrollFade />
```
mj12albert: Maybe this could be a `compact` demo, and add a CSS code block showing usage of the CSS vars, since the base styles are very similar to the hero demo and its easier than trying to find the differences ~

PR #3167, opened 2025-11-07
[combobox] Support drag-to-select

### LINKED ISSUE #3086: [combobox] Combobox with trigger does not support click and drag to select
# Bug report

## Current behavior

If Combobox is used with Combobox.Trigger, it does not support the click and drag to select functionality.
 
## Expected behavior

The Select component supports that a user can click on the trigger button 
and then drag the mouse down to select an item: ([Example](https://base-ui.com/react/components/select#multiple-selection))

I would expect the Combobox to behave the same, but it doesn't: ([Example](https://base-ui.com/react/components/combobox#input-inside-popup))

## Base UI version

1.0.0-beta.4

Chrome

## Which OS are you using?

Mac OS

atomiks:

Closes #3086

### DIFF
--- a/packages/react/src/combobox/trigger/ComboboxTrigger.tsx
+++ b/packages/react/src/combobox/trigger/ComboboxTrigger.tsx
@@ -64,6 +64,7 @@ export const ComboboxTrigger = React.forwardRef(function ComboboxTrigger(
   const readOnly = useStore(store, selectors.readOnly);
   const listElement = useStore(store, selectors.listElement);
   const triggerProps = useStore(store, selectors.triggerProps);
+  const triggerElement = useStore(store, selectors.triggerElement);
   const inputInsidePopup = useStore(store, selectors.inputInsidePopup);
   const open = useStore(store, selectors.open);
   const selectedValue = useStore(store, selectors.selectedValue);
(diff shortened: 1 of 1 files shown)


packages/react/src/combobox/item/ComboboxItem.tsx
```diff
@@ -154,24 +168,33 @@ export const ComboboxItem = React.memo(
...
-          store.state.requestSubmit();
-        } else {
-          selectItem();
+        if (
+          disabled ||
+          readOnly ||
+          event.button !== 0 ||
+          pointerStartedOnItem ||
+          !highlighted ||
+          !store.state.inputInsidePopup
```
mj12albert: Just wondering could this also work for input-outside-popup? If the pointer starts at the smaller trigger, not the input

atomiks: Possibly, though it feels a bit inconsistent for it to work only on the trigger with that layout/pattern

LukasTy: I agree with the proposal.
Seems strange to support this only on the popup mode, whereas regular `Combobox` closely resembles `Select`, but does not allow this behavior.

If we keep this behavior as is, then maybe the trigger drag -> move away behavior should change in regular behavior?
It seems slightly strange now. 🤷 

https://github.com/user-attachments/assets/cef40caf-3f99-4cdd-a4ee-4a42a0d3b308

atomiks: Dragging on the input selects the text (if you had some in the input in that recording) though, so there's a functional difference there. Pretty sure no one is going to notice any of this though.

I removed the `mouseup` drag out behavior to match the input

LukasTy (approved):
Looks great, especially after the tiny fix. 👌

packages/react/src/combobox/item/ComboboxItem.tsx
mj12albert: OK I don't think it's hugely important now, but would be nice to have later ~

RA let's you single-pointer the trigger in this layout: https://react-spectrum.adobe.com/react-aria/ComboBox.html#example

Also maybe because it's styled it more distinctly from the input, seems like this interaction would be more discoverable

atomiks: Hmm fair enough, the styled trigger does make it more clear there. I think we can unconditionally allow drags on both the trigger and input (even if it selects text) for consistency across both, but leave the `cancel-open` exclusive to the `input-inside-popup` layout

LukasTy:
Is it expected that releasing the mouse outside of the Popup closes the Popup in case of https://deploy-preview-3167--base-ui.netlify.app/react/components/combobox#input-inside-popup, but doesn't when used with an input? 🤔

atomiks:
@LukasTy yeah so it's consistent with the input: https://github.com/mui/base-ui/pull/3167#discussion_r2520737844

PR #3170, opened 2025-11-07
[menu] Support detached triggers

### LINKED ISSUE #3031: [menu] Support detached triggers
# Feature request

Similar to https://github.com/mui/base-ui/pull/2974, but for menus

michaldudak:
Implemented detached / multiple trigger support in Menu, similarly to https://github.com/mui/base-ui/pull/2336

It is now possible to define triggers outside of Menu.Root using the handle prop. Multiple triggers per menu with a dynamic payload are also implemented.

Detached triggers also work in Menubar menus.

Content transitions will be implemented in a follow-up PR.

Closes #3031 

### Preview

- https://deploy-preview-3170--base-ui.netlify.app/experiments/menu/triggers
- https://deploy-preview-3170--base-ui.netlify.app/experiments/menu/nested-detached-triggers

### Breaking changes

- `openOnHover`, `delay`, and `closeDelay` props moved from Menu.Root to Menu.Trigger.
- Menus must have at least one trigger.

### DIFF
--- a/docs/reference/generated/context-menu-trigger.json
+++ b/docs/reference/generated/context-menu-trigger.json
@@ -17,6 +17,13 @@
       "detailedType": "| ReactElement\n| ((\n    props: HTMLProps,\n    state: ContextMenu.Trigger.State,\n  ) => ReactElement)"
     }
   },
-  "dataAttributes": {},
+  "dataAttributes": {
+    "data-popup-open": {
+      "description": "Present when the corresponding context menu is open."
+    },
+    "data-pressed": {
+      "description": "Present when the trigger is pressed."
+    }
+  },
   "cssVariables": {}
 }
--- a/docs/reference/generated/select-trigger.json
+++ b/docs/reference/generated/select-trigger.json
@@ -65,6 +65,9 @@
     },
     "data-focused": {
       "description": "Present when the select trigger is focused (when wrapped in Field.Root)."
+    },
+    "data-placeholder": {
+      "description": "Present when the select doesn't have a value."
     }
   },
   "cssVariables": {}
--- a/docs/reference/generated/select-value.json
+++ b/docs/reference/generated/select-value.json
@@ -23,6 +23,10 @@
       "detailedType": "| ReactElement\n| ((\n    props: HTMLProps,\n    state: Select.Value.State,\n  ) => ReactElement)"
     }
   },
-  "dataAttributes": {},
+  "dataAttributes": {
+    "data-placeholder": {
+      "description": "Present when the select doesn't have a value."
+    }
+  },
   "cssVariables": {}
 }
--- a/docs/src/app/(public)/(content)/react/components/toast/page.mdx
+++ b/docs/src/app/(public)/(content)/react/components/toast/page.mdx
@@ -53,6 +53,7 @@ import { Toast } from '@base-ui-components/react/toast';
 - `<Toast.Provider>` can be wrapped around your entire app, ensuring all toasts are rendered in the same viewport.
 - <kbd>F6</kbd> lets users jump into the toast viewport landmark region to navigate toasts with
   keyboard focus.
+- The `data-swipe-ignore` attribute can be manually added to elements inside of a toast to prevent swipe-to-dismiss gestures on them. Interactive elements are automatically prevented.
 
 ## Global manager
 
@@ -298,11 +299,11 @@ const toastManager = Toast.useToastManager();
 <PropsReferenceTable
   data={{
     title: {
-      type: 'string',
+      type: 'React.ReactNode',
       description: 'The title of the toast.',
     },
     description: {
-      type: 'string',
+      type: 'React.ReactNode',
       description: 'The description of the toast.',
     },
     type: {
--- a/packages/react/src/combobox/empty/ComboboxEmpty.tsx
+++ b/packages/react/src/combobox/empty/ComboboxEmpty.tsx
@@ -2,7 +2,10 @@
 import * as React from 'react';
 import { BaseUIComponentProps } from '../../utils/types';
 import { useRenderElement } from '../../utils/useRenderElement';
-import { useComboboxDerivedItemsContext } from '../root/ComboboxRootContext';
+import {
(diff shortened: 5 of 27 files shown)


atomiks:
This is also on main, but not in beta.4, perhaps it should be fixed here though:

On https://deploy-preview-3170--base-ui.netlify.app/experiments/popups/popups-in-popups, click-to-drag on the menu to a submenu item and releasing closes the dialog unexpectedly

atomiks:
On https://deploy-preview-3170--base-ui.netlify.app/experiments/menu/nested-detached-triggers,

- When a submenu is open, trying to hover a new menubar trigger doesn't work. (Works fine on the Menubar docs hero demo, though)
- When clicking the first trigger, then hovering to the last trigger, hovering over and out of the menu, it closes unexpectedly. Afterwards, clicking a trigger causes it to close on pointerup.

packages/react/src/menu/trigger/MenuTrigger.tsx
```diff
@@ -33,35 +54,110 @@ export const MenuTrigger = React.forwardRef(function MenuTrigger(
...
+    );
+  }
+
+  const contextMenuContext = useContextMenuRootContext(true);
+  const parentContext = useMenuRootContext(true);
+  const menubarContext = useMenubarContext(true);
+  const compositeRootContext = useCompositeRootContext(true);
+
+  const parent: MenuParent = React.useMemo(() => {
+    if (menubarContext) {
```
atomiks, line 91: Even though `React.useMemo()` is useful as a neat "IIFE" to avoid using `let`, we need to avoid this since each memo has a cost~ calculate it inline in the component body with `let`

packages/react/src/menu/trigger/MenuTrigger.tsx
```diff
@@ -199,6 +367,29 @@ export type MenuTriggerState = {
...
 
 export namespace MenuTrigger {
-  export type Props = MenuTriggerProps;
+  export type Props<Payload = unknown> = MenuTriggerProps<Payload>;
   export type State = MenuTriggerState;
 }
+
+/**
+ * Determines whether to ignore clicks after a hover-open.
+ */
```
atomiks, line 461: Popover also uses this, maybe it could be extracted as a common util

atomiks:
- Tabbing out of the menu in the controlled section of the experiment skips over the other triggers
- Pressing <kbd>Esc</kbd> instantly returns focus instead of waiting for unmount

packages/react/src/dialog/root/useDialogRoot.ts
```diff
@@ -144,7 +144,8 @@ export function useDialogRoot(params: useDialogRoot.Parameters): useDialogRoot.R
           return store.context.internalBackdropRef.current || store.context.backdropRef.current
             ? store.context.internalBackdropRef.current === eventTarget ||
                 store.context.backdropRef.current === eventTarget ||
-                contains(eventTarget, popupElement)
+                (contains(eventTarget, popupElement) &&
```
mnajdova, line 147: Is this change intentional as part of this PR?

packages/react/src/floating-ui-react/hooks/useFocus.ts
```diff
@@ -194,5 +194,8 @@ export function useFocus(context: FloatingRootContext, props: UseFocusProps = {}
     [dataRef, elements.domReference, elements.triggers, onOpenChange, visibleOnly, timeout],
   );
 
-  return React.useMemo(() => (enabled ? { reference } : {}), [enabled, reference]);
+  return React.useMemo(
+    () => (enabled ? { reference, trigger: reference } : {}),
```
mnajdova, line 198: What's the value of adding the trigger as key, if it's always the same as the reference value?

docs/src/app/(private)/experiments/menu/nested-detached-triggers.tsx
```diff
@@ -0,0 +1,268 @@
...
+          ))}
+        </div>
+
+        <ReusableMenu handle={menu1Handle} />
+      </div>
+    </div>
+  );
+}
+
+function ReusableMenu(props: { handle: Menu.Handle<ContentKey> }) {
```
mnajdova: I really like the API, it's very clean 👌

packages/react/src/dialog/root/useDialogRoot.ts
michaldudak: Yes, it's a fix for https://github.com/mui/base-ui/pull/3170#issuecomment-3515044473

packages/react/src/floating-ui-react/hooks/useFocus.ts
michaldudak: `useInteractions` returns methods like `getReferenceProps` and `getTriggerProps` that use these keys. MenuTrigger calls these methods depending on whether it's an active trigger. Previously, nothing would be returned in the `getTriggerProps` call.

packages/react/src/menu/trigger/MenuTrigger.tsx
michaldudak: It's not just for convenience. `parent` is synchronized to the store and needs to be a stable reference.

packages/react/src/menu/trigger/MenuTrigger.tsx
michaldudak: I'm planning a follow-up PR to create common utils, split other Floating UI interaction hooks and generally clean up the code.

atomiks:
In the Menubar experiment, the safe polygon logic broke on the 'Add to playlist' submenu

mnajdova:
Another bug I noticed in the Menubar is that if I open a submenu, the other menus no longer open on hover. There is also a flicker if I try to reopen the same submenu, see the recording attached.


https://github.com/user-attachments/assets/2a40e6c7-085f-49d3-a9a3-a64bd8303c5b

atomiks (approved):
Haven't been able to find regressions in regular menus - in the docs, some of the components need to be updated to use the tag format `<Menu.Root>` (not `Menu.Root`)

PR #3173, opened 2025-11-07
[toggle-group][toggle] Improve type safety and inference

### LINKED ISSUE #3142: [ToggleGroup] Better type safety
# Feature request

## Summary

Currently `ToggleGroup`'s props are filled with `any` instead of just being generic.
on top of that, the `value` prop and `onValueChange` don't depend on if the `multiple` is true or not causing the component to always expect and array (even if multiple select is not expected).

What i suggest is the behavior to be the same as the way that `Combobox` component handles the `value` and `multiple` props.

## Motivation

Feels more consistent with the rest of the library.

michaelhazan:
Closes #3142.
Closes https://github.com/mui/base-ui/issues/3756.

> [!NOTE]
> Originally this was meant to only introduce generic types to the `ToggleGroup`, but i realized while working on this that `Toggle` too had to be changed to handle a generic value.

### DIFF
--- a/packages/react/src/toggle-group/ToggleGroup.spec.tsx
+++ b/packages/react/src/toggle-group/ToggleGroup.spec.tsx
@@ -16,11 +16,11 @@ const values = ['a', 'b', 'c'];
   }}
 />;
 
-const readonlyValues = ['a', 'b', 'c'] as const;
-type ReadonlyValue = (typeof readonlyValues)[number];
+const narrowedValues = ['a', 'b', 'c'] as const;
+type ReadonlyValue = (typeof narrowedValues)[number];
 
 <ToggleGroup
-  value={readonlyValues}
+  value={narrowedValues}
   onValueChange={(value) => {
     expectType<ReadonlyValue[], typeof value>(value);
   }}


packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -24,32 +24,34 @@ const stateAttributesMapping = {
...
     disabled: disabledProp = false,
     loop = true,
     onValueChange,
     orientation = 'horizontal',
-    multiple = false,
+    multiple = false as Multiple,
     value: valueProp,
     className,
     render,
+    rootRef = { current: null },
```
mj12albert: We must forwardRef for a consistent API, to get around the issue you can cast the result of forwardRef like this: https://github.com/mui/base-ui/blob/master/packages/react/src/slider/root/SliderRoot.tsx#L380-L387

michaelhazan:
it seems like this is failing over something i have not affected...

would appreciate help

mj12albert:
@michaelhazan My bad, I remembered we actually do this intentionally:
> the value prop and onValueChange don't depend on if the multiple is true or not causing the component to always expect and array

Accordion is the same and the value is always an array

michaelhazan:
> @michaelhazan My bad, I remembered we actually do this intentionally:
> > the value prop and onValueChange don't depend on if the multiple is true or not causing the component to always expect and array
> 
> Accordion is the same and the value is always an array
> 

@mj12albert lmao, no worries, I'll work on keeping it an array while having it still be generic

michaelhazan:
@mj12albert sorry it took me so long, work had me backed up 😅 

it should be good now, because `Toggle` now extends `BaseUIComponentProps`, i had to make the value an extendable of `number | string`, should nonetheless be an improvement to use with `Enum`s and `Constant Object`s :D

LukasTy (commented):
@michaldudak, is this still relevant after merging https://github.com/mui/base-ui/pull/3770?

michaldudak:
@LukasTy, yes, I haven't noticed this PR. It's basically an equivalent of what I've recently done in #3788.

This approach has a small issue we discussed internally:

When Toggle's `value` is undefined, we generate it using `useBaseUiId`, which returns a string. So we can't really guarantee that `ToggleGroup<T>` with `Toggle<T>` components inside is indeed type-safe. Options to solve it could be:
- Enforce a Toggle's value when it's used within a ToggleGroup. It actually doesn't make much sense otherwise, but technically is a breaking change.
- Ignore it. Since the contract between Toggle and ToggleGroup is weak anyway (we can't enforce them to use the same type of value), it will be up to developers to make sure the correct values are provided. I think it should be possible to make it a bit more obvious by requiring value on a type level whenever a type parameter is set explicitly (so `<Toggle<'left' | 'right'>` would error in TS without a value)


Also, I'm afraid we can't use `| number` at this point, as it would break all the codebases that assume it's a string.

LukasTy:
@michaldudak Thanks for sharing more details. 🙏 
Do you think it makes sense to keep this PR open for v2 as a breaking change, depending on which route we decide to go with (enforce a value or consider supporting `| number` as well if there is enough need for it)?

michaldudak:
We discussed this with the team and agreed that we can proceed with this implementation ignoring the potentially incorrect types. To further prevent surprises, we can add a dev-time warning when a Toggle within a ToggleGroup doesn't have an explicit value.

@michaelhazan, would you like to continue working on this?

michaelhazan:
> @michaelhazan, would you like to continue working on this?

Sure!
I will merge it with main and add the warning 👍

michaelhazan:
@michaldudak Hey! should be good now.. 
if you have anymore suggestions you want me to implement let me know!

packages/react/src/toggle/Toggle.tsx
```diff
@@ -115,7 +128,15 @@ export const Toggle = React.forwardRef(function Toggle(
...
-});
+}) as {
+  <Value extends ToggleValue>(
+    props: Toggle.Props<Value> & {
+      ref?: React.RefObject<HTMLButtonElement> | undefined;
+    },
+  ): React.JSX.Element;
+};
+
+export type ToggleValue = string | number;
```
michaldudak: As noted [here](https://github.com/mui/base-ui/pull/3173#issuecomment-3777360147), we can't really use a number at this point. Let's make it just a string (and we can use it without the ToggleValue indirection)

michaelhazan: fixed!

packages/react/src/toggle/Toggle.tsx
```diff
@@ -47,6 +49,17 @@ export const Toggle = React.forwardRef(function Toggle(
 
   const disabled = (disabledProp || groupContext?.disabled) ?? false;
 
+  if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useIsoLayoutEffect(() => {
+      if (groupContext && valueProp === undefined) {
+        warn(
+          'Value prop in `Toggle` component is `undefined` while its in a `ToggleGroup` component. This can cause type issues between the `ToggleGroup` values and the `Toggle` value, Provide the `Toggle` with a value prop matching the `ToggleGroup` values prop type.',
```
LukasTy: WDYT about this wording change?
```suggestion
          'A `<Toggle>` component rendered in a `<ToggleGroup>` has no explicit `value` prop.
          This can cause type issues between the Toggle Group and Toggle values.
          Provide the `<Toggle>` with a `value` prop matching the `<ToggleGroup>` values prop type.',
```

packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -197,7 +207,7 @@ export type ToggleGroupChangeEventDetails = BaseUIChangeEventDetails<ToggleGroup
 
 export namespace ToggleGroup {
   export type State = ToggleGroupState;
-  export type Props = ToggleGroupProps;
+  export type Props<Value extends string> = ToggleGroupProps<Value>;
```
LukasTy: Correct me if I'm wrong, but is there a use case for why we still need the generic here? 🤔 

As far as I can see, it only helps with type inference, like if you have the following:
```tsx
type Alignment = 'horizontal' | 'vertical';

const alignments: Alignment[] = ['horizontal', 'vertical'];

<ToggleGroup
  multiple
  defaultValue={alignments}
  // this is going to know that the type is `Alignment` instead of plain string as it is now
  value={}
>
```
But, the current types don't seem to block the above usage as well.

If we go with this, then we should probably adjust other components (i.e., RadioGroup, CheckboxGroup) accordingly.

WDYT @michaldudak?

packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -148,24 +155,27 @@ export interface ToggleGroupState {
...
    */
-  value?: readonly string[] | undefined;
+  value?: readonly Value[] | undefined;
   /**
    * The open state of the toggle group represented by an array of
    * the values of all pressed toggle buttons.
    * This is the uncontrolled counterpart of `value`.
    */
-  defaultValue?: readonly string[] | undefined;
+  defaultValue?: Value[] | undefined;
```
LukasTy: This has to remain readonly.
Please add a `.Spec` file asserting the usage to discover the issue that now exists and the spec file would prevent introducing it later down the line.

```suggestion
  defaultValue?: readonly Value[] | undefined;
```

packages/react/src/toggle-group/ToggleGroup.tsx
mj12albert: I think this is useful and addresses https://github.com/mui/base-ui/issues/3756

(it should also be able to infer the value arg in `onValueChange` I assume)

LukasTy: Gotcha. Makes sense, especially given we already have others noticing this shortcoming. 👍 
I think we need to do a follow-up and address this issue for all relevant components.

packages/react/src/toggle/Toggle.tsx
michaelhazan: makes sense 😅 will change!

packages/react/src/toggle-group/ToggleGroup.tsx
michaelhazan: @LukasTy would you like me to fix it in this pr? or leave it for now?

LukasTy: I think we can keep this one isolated to these specific components.
WDYT?

michaelhazan: for sure

michaelhazan:
@LukasTy fixed!
was uncertain if you wanted me to implement something from [your comment here](https://github.com/mui/base-ui/pull/3173#discussion_r2731316507), from what i got, you want to make a general change in the library following what i've done here and #3756 , and therefore no change in necessary in this pr.

did i get that right? 😅 

_please note_ i've also set the props's generic type to equal `string` by default after i saw similar thing in `SliderRootProps`.
[here](https://github.com/mui/base-ui/pull/3173/changes/cf9b46e316682179a0c0cafaf74be4f43f8f3c55#diff-8384198ce367f24ce232b74009d5f3c0a1827d9d956e8e733b7a6e1560bfda85R210) and [here](https://github.com/mui/base-ui/pull/3173/changes/cf9b46e316682179a0c0cafaf74be4f43f8f3c55#diff-3d9001b9d92df79342d89b626bf30220612b209eabe45c906523055e69af89a2R187)

packages/react/src/toggle/Toggle.tsx
```diff
@@ -47,6 +49,17 @@ export const Toggle = React.forwardRef(function Toggle(
 
   const disabled = (disabledProp || groupContext?.disabled) ?? false;
 
+  if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useIsoLayoutEffect(() => {
+      if (groupContext && valueProp === undefined) {
+        warn(
+          'A `<Toggle>` component rendered in a `<ToggleGroup>` has no explicit `value` prop. This can cause type issues between the Toggle Group and Toggle values. Provide the `<Toggle>` with a `value` prop matching the `<ToggleGroup>` values prop type.',
```
mj12albert: What type issue does this cause?
This seems like a clumsy but legit way to use toggle group without values:
```jsx
<ToggleGroup>
  <Toggle onPressedChange={setState1} />
  <Toggle onPressedChange={setState2} />
</ToggleGroup>
```

LukasTy: I think Michal raised this point that in case the `<ToggleGroup>` `defaultValue` would have a type of `Alignment[]`, then you'd end up in a slight type discrepancy, because internally a Toggle without a `value` would have generated an internal value type of `string`.
However, this is a very edge case condition. 🤔 
Would it make more sense if the warning would be shown only in your case (no value on the Toggle and no default and no direct value on the Toggle Group)?

michaelhazan: i guess it should be more like 
```ts
if(groupContext && 
   groupContext.valueProp /* somehow the `Toggle` needs to know the prop given to the `ToggleGroup` */ && 
   valueProp === undefined) 
```

but this is problematic obviously... adding a value to the context simply for debugging doesnt seem right...

and there isn't really a way to check for it otherwise..

LukasTy: There is value in the context already, or am I missing something?

I played around with the PR, I hope you don't mind that I hijacked it a bit and pushed a few changes. 🙈

Indeed, I don't see the "typing" as an issue, the real issue is the value problem.

Try this example, for instance:
```tsx
'use client';
import { ToggleGroup } from '@base-ui/react/toggle-group';
import { Toggle } from '@base-ui/react/toggle';

const alignmentOptions = ['start'] as const;

export default function ToggleGroupExperiment() {
  return (
    <React.Fragment>
      <ToggleGroup
        multiple
        defaultValue={alignmentOptions}
        onValueChange={(value, eventDetails) => console.info(value, eventDetails)}
      >
        <Toggle>Start</Toggle>
        <Toggle>Center</Toggle>
        <Toggle>End</Toggle>
      </ToggleGroup>
    </React.Fragment>
  );
}
```

The `onValueChange` will always have a ghost `start` value, that can not be unchecked ("unpressed").
It looks like the bigger problem is a case, where the Group has a `value` or `defaultValue` (both map to `groupContext.value`).
I feel like we should better warn about these cases instead.

WDYT @michaldudak, @mj12albert?
I leave the approval to anyone of you, given that I modified the approach.

packages/react/src/toggle/Toggle.tsx
```diff
@@ -52,9 +52,11 @@ export const Toggle = React.forwardRef(function Toggle<Value extends string>(
   if (process.env.NODE_ENV !== 'production') {
     // eslint-disable-next-line react-hooks/rules-of-hooks
     useIsoLayoutEffect(() => {
-      if (groupContext && valueProp === undefined) {
+      if (groupContext && valueProp === undefined && groupContext?.value.length > 0) {
```
michaelhazan: but wont this be triggered if the `ToggleGroup` simply has `multiple` prop?
so 
```tsx
<ToggleGroup multiple>
    <Toggle>One</Toggle>
    <Toggle>Two</Toggle>
    <Toggle>Three</Toggle>
</ToggleGroup>
```

will cause an warning upon multiple toggles being selected even though its valid.

what i meant in my comment is that we need to check if the `ToggleGroup` i controlled or not, that is where the issue stems from.

if the `ToggleGroup` is controlled but the `Toggle` has no value prop, it *could* cause an issue.

LukasTy: Yes, sorry, my bad, this will log if there is no explicit `value` or `defaultValue` and you select something. 🙈

packages/react/src/toggle/Toggle.tsx
```diff
@@ -47,6 +49,19 @@ export const Toggle = React.forwardRef(function Toggle(
 
   const disabled = (disabledProp || groupContext?.disabled) ?? false;
 
+  if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useIsoLayoutEffect(() => {
+      if (groupContext && valueProp === undefined && groupContext.isValueInitialized) {
```
michaelhazan, line 55: feels weird having a context value only for debugging, but it is more accurate now.

maybe write a comment on it in the context?

nonetheless LGTM 👍

packages/react/src/toggle/Toggle.tsx
LukasTy: @michaelhazan I've gone for storing the value in the group context to ensure correctness instead of using a `React.useRef` inside of `Toggle`.
I don't think it's a big problem to store this flag in the context
Can you think of problems with this?

packages/react/src/toggle/Toggle.tsx
LukasTy: IMHO, we can even consider "upgrading" this message to an `error`, given that it can/will cause actual usage problems.
WDYT?

packages/react/src/toggle/Toggle.tsx
michaelhazan: wrote you a comment about it, but generally dont have an issue with it, does make sense to have some sort of understanding in the context of whether or not it is controlled 👍
awesome stuff 😤

packages/react/src/toggle/Toggle.tsx
michaelhazan: sounds good!

packages/react/src/toggle-group/ToggleGroup.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+const values = ['a', 'b', 'c'];
+
+<ToggleGroup
+  value={values}
+  onValueChange={(value) => {
+    expectType<string[], typeof value>(value);
+  }}
+/>;
+
+const readonlyValues = ['a', 'b', 'c'] as const;
```
michaldudak: nit: technically, it's as readonly as `values`. `narrowedValues` would be more correct.

packages/react/src/toggle/Toggle.tsx
michaldudak: I'm wondering if we even need to check if groupContext's value is initialized. I can't think of a use case where users would want to have value-less Toggles within a ToggleGroup

packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -131,15 +137,22 @@ export const ToggleGroup = React.forwardRef(function ToggleGroup(
       )}
     </ToggleGroupContext.Provider>
   );
-});
+}) as {
+  <Value extends string>(
+    props: ToggleGroup.Props<Value> & { ref?: React.RefObject<HTMLDivElement> | undefined },
```
michaldudak: This forbids passing in function refs.

```suggestion
    props: ToggleGroup.Props<Value> & React.RefAttributes<HTMLDivElement>,
```

packages/react/src/toggle/Toggle.tsx
```diff
@@ -115,7 +130,13 @@ export const Toggle = React.forwardRef(function Toggle(
   }
 
   return element;
-});
+}) as {
+  <Value extends string>(
+    props: Toggle.Props<Value> & {
+      ref?: React.RefObject<HTMLButtonElement> | undefined;
+    },
```
michaldudak: This forbids passing in function refs.


```suggestion
    props: Toggle.Props<Value> & React.RefAttributes<HTMLButtonElement>,
```

packages/react/src/toggle/Toggle.tsx
michaelhazan: @mj12albert raised a fair point [here](https://github.com/mui/base-ui/pull/3173#discussion_r2735488423)

michaldudak: Right, OK.

PR #3175, opened 2025-11-09
[docs] Set explicit height in virtualized demos, avoid deferred value flickers

atomiks:
- `height` still needs to be specified for each virtualized item (which was removed in https://github.com/mui/base-ui/pull/3130); when selecting e.g. the 10th item, the next open doesn't scroll to the item correctly otherwise
- Fixed flickers related to `React.useDeferredValue` by using the non-deferred value when either is an empty string (added in https://github.com/mui/base-ui/pull/3068):
  - Selecting an item, selecting all text and pressing <kbd>Backspace</kbd> would briefly show the selected item alone in the list before revealing all options again
  - Tabbing into input and typing <kbd>f</kbd> to show the empty state would briefly show all items before the empty state is shown


mj12albert:
Found a filter related bug while testing this https://github.com/mui/base-ui/issues/3192

mj12albert (approved):

Issue #2006, opened 2025-05-30
[tabs] Respect Pointer Cancellation accessibility guidelines


davidm-m:
# Feature request

## Summary

Current behaviour for tabs is to activate as soon as the pointer is pressed. In order to meet [WCAG Level A 2.2](https://www.w3.org/WAI/WCAG22/Understanding/pointer-cancellation.html) this should instead be changed to meet one of the following:

1. No Down-Event
    a. The [down-event](https://www.w3.org/WAI/WCAG22/Understanding/pointer-cancellation.html#dfn-down-event) of the pointer is not used to execute any part of the function;
2. Abort or Undo
    a. Completion of the function is on the [up-event](https://www.w3.org/WAI/WCAG22/Understanding/pointer-cancellation.html#dfn-up-event), and a [mechanism](https://www.w3.org/WAI/WCAG22/Understanding/pointer-cancellation.html#dfn-mechanism) is available to abort the function before completion or to undo the function after completion;
3. Up Reversal
    a. The up-event reverses any outcome of the preceding down-event;
4. Essential
    a. Completing the function on the down-event is [essential](https://www.w3.org/WAI/WCAG22/Understanding/pointer-cancellation.html#dfn-essential).

My recommendation is to go for 1. as other examples do.

## Examples in other libraries

- [Material UI](https://mui.com/material-ui/react-tabs/)'s implementation of tabs already follows this.
- Radix Primitives https://github.com/radix-ui/primitives/issues/1879
- React Aria makes it configurable https://github.com/adobe/react-spectrum/pull/4342

Additionally, WCAG provides an [example implementation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/). Observe that in this implementation the only pointer event listened for is `onClick`, which only fires when the pointer is released.

## Motivation

Changing this improves consistency with other components that meet accessibility standards and allows users to recover from accidental clicks.

mj12albert:
This was discussed in depth in https://github.com/mui/base-ui/issues/1323

I think we should make this configurable by the user, even if we keep the current default

mj12albert:
We decided to change the default behavior to activate on click instead of pointerdown in order to pass the success criterion

Additionally it will not be made configurable at the moment, but we'll consider adding it back in the future

benface:
> the default behavior

> it will not be made configurable

😅

mnajdova:
[...]

PR #3176, opened 2025-11-10
[tabs] Change `activateOnFocus` to false 

mnajdova:
This PR changes the activateOnFocus to be false in the TabsList component and also makes the pointer activating behavior follow the prop (by default it activates the tabs on click, but if the prop is applied it activates on pointer down).

Closes https://github.com/mui/base-ui/issues/2006

### DIFF
--- a/packages/react/src/tabs/root/TabsRoot.test.tsx
+++ b/packages/react/src/tabs/root/TabsRoot.test.tsx
@@ -174,12 +174,12 @@ describe('<Tabs.Root />', () => {
   });
 
   describe('prop: onValueChange', () => {
-    it('should call onValueChange on pointerdown', async () => {
+    it('when `activateOnFocus = true` should call onValueChange on pointerdown', async () => {
       const handleChange = spy();
       const handlePointerDown = spy();
       const { user } = await render(
         <Tabs.Root value={0} onValueChange={handleChange}>
-          <Tabs.List>
+          <Tabs.List activateOnFocus>
             <Tabs.Tab value={0} />
             <Tabs.Tab value={1} onPointerDown={handlePointerDown} />
           </Tabs.List>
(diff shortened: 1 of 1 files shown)


packages/react/src/tabs/tab/TabsTab.tsx
```diff
@@ -140,8 +140,9 @@ export const TabsTab = React.forwardRef(function TabsTab(
     }
 
     if (
-      (activateOnFocus && !isPressingRef.current) || // keyboard or touch focus
-      (isPressingRef.current && isMainButtonRef.current) // mouse focus
+      activateOnFocus &&
```
mnajdova, line 143: I changed the mouse down behavior to work only if `activateOnFocus` is set, was that what you had in mind @colmtuite? By default it activates on click.

mj12albert: I thought activateOn(Keyboard)Focus is separate from the pointer activation event (down vs up/click)?

mnajdova: Ok, good that I asked. I thought I remember that we said activateOnFocus is the same as pointer down, while click is the same as (activateOnFocus={false}). Let's get second confirmation :)

colmtuite: No I did suggest that, but you said the connection is not that intuitive, and might surprise people, which makes sense.

So I was thinking we just change the interaction to click, and set `activateOnFocus` to false by default.

mnajdova: That means we never give people option to support mouse down, are we ok with it?

colmtuite: I'm ok with it since it matches other components that aren't popups.

mnajdova: For future reference, we decided to go with using mouse down if `activeOnFocus` is set, click is still the default behaviour.

PR #3177, opened 2025-11-10
[select] Make the trigger native button by default

mnajdova:
There was a Safari bug that prevented us to use button by default, but it seems to be working now.


packages/react/src/select/trigger/SelectTrigger.tsx
```diff
@@ -45,7 +45,7 @@ export const SelectTrigger = React.forwardRef(function SelectTrigger(
     render,
     className,
     disabled: disabledProp = false,
-    nativeButton = false,
+    nativeButton = true,
```
atomiks, line 48: `useRenderElement('div', ...` needs to be changed to `button` for the actual element to change

mj12albert:
Here's the bug: https://bugs.webkit.org/show_bug.cgi?id=285047

It was reported on Safari 18.x but it's no longer the current version

atomiks:
In Safari 26, it weirdly says "Mispelt" before saying the trigger text, even when it's a `<div>`. The truncation issue doesn't happen though, and this also matches `<Combobox.Trigger>`. 

The styles of the demos need updating though; the trigger background color is incorrect on them.

docs/src/app/(public)/(content)/react/components/select/demos/object-values/css-modules/index.module.css
```diff
@@ -12,11 +12,11 @@
   outline: 0;
   border: 1px solid var(--color-gray-200);
   border-radius: 0.375rem;
+  background-color: white;
```
atomiks:
```suggestion
  background-color: canvas;
```
(and update all other ones)

atomiks (approved):

Issue #2598, opened 2025-08-31
expose `useScrollLock` utility


pondorasti:
# Feature request

## Summary

I was snooping around the source code and I noticed that `base-ui` has an internal utility for scroll locking, `useScrollLock`.

https://base-ui.com/experiments/scroll-lock

## Examples in other libraries

There's 2 relatively popular libraries that solve this problem, both of them unmaintained for years at this point. Would love to have `base-ui` as a replacement for them.

* https://github.com/jossmac/react-scrolllock
* https://github.com/willmcpo/body-scroll-lock

atomiks:
The hook has an awkward API for public use because of this effect to fix a bug

It requires a special `mounted` value

```js
  // https://github.com/mui/base-ui/issues/1135
  useIsoLayoutEffect(() => {
    if (enabled && isWebKit && mounted && !open) {
      const doc = ownerDocument(referenceElement);
      const originalUserSelect = doc.body.style.userSelect;
      const originalWebkitUserSelect = doc.body.style.webkitUserSelect;
      doc.body.style.userSelect = 'none';
      doc.body.style.webkitUserSelect = 'none';
      return () => {
        doc.body.style.userSelect = originalUserSelect;
        doc.body.style.webkitUserSelect = originalWebkitUserSelect;
      };
    }
    return undefined;
  }, [enabled, mounted, open, referenceElement]);
```

pondorasti:
Did this issue get inadvertently closed? As far as I can tell, the `useScrollLock` utility has yet to be introduced as a public API.

LukasTy:
> Did this issue get inadvertently closed? As far as I can tell, the `useScrollLock` utility has yet to be introduced as a public API.

It is exported from [`@base-ui-components/utils`](https://www.npmjs.com/package/@base-ui-components/utils/v/0.2.0?activeTab=code) starting with `v0.2.0`.

<img width="797" height="493" alt="Image" src="https://github.com/user-attachments/assets/c3a57513-15e5-44bd-a04a-6a573e5a8d17" />

pondorasti:
Completely missed that, thank you so much!

Issue #2126, opened 2025-06-17
[menu] Scroll lock prevents `MenuItem render={<a>}` from navigating within same page


mj12albert:
Repro: https://codesandbox.io/p/sandbox/frosty-rui-r7lp6w

If the menu item is a link to a different part of the same page as the menu, when clicking the link, the scroll lock seems to be restoring the scroll position to the trigger/menu

mj12albert:
Discovered this when checking if any components were still using `useButton` to render links, and `MenuItem` kind of indirectly allows it

Maybe we should have a `MenuItemLink` component? Otherwise it's possible to have disabled links which isn't good

michaldudak:
I prefer having a separate component than supporting `MenuItem render={<a>}`

PR #3187, opened 2025-11-11
[useScrollLock] Simplify behavior when `scrollbar-gutter` is supported

### LINKED ISSUE #3180: Menu has wrong position after document height changes
# Bug report

## Current behavior

Steps to reproduce:

1. On macOS, make sure scroll bar behaviour is set to "Always"
1. In the [reduced test case](https://codesandbox.io/p/devbox/zen-sky-fpcd7k?file=%2Fsrc%2FApp.tsx&workspaceId=ws_4qKVDKGuRrrvd8oLfoRWSz), scroll down, open the menu, then click the menu item.

Behaviour: the menu is repositioned to the bottom of the screen.

https://github.com/user-attachments/assets/48392727-33c1-4e88-a698-2d797fc4a0ae

## Expected behavior

The menu should remain anchored to the trigger.

## Reproducible example

https://codesandbox.io/p/devbox/zen-sky-fpcd7k?file=%2Fsrc%2FApp.tsx&workspaceId=ws_4qKVDKGuRrrvd8oLfoRWSz

## Base UI version

1.0.0-beta.4

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

N/A

atomiks:

Fixes #3180
Fixes https://github.com/mui/base-ui/issues/2126 (https://codesandbox.io/p/sandbox/kind-tu-yp2qtf)
Fixes https://github.com/mui/base-ui/issues/3189
Closes https://github.com/mui/base-ui/issues/2598 (simplified signature for `/utils` package which people can use)

This removes the special `height` handling set on the body which breaks custom scroll updates. Only old Safari is likely to be an issue here (before 18.2) which is why I've kept the old logic still. We can probably remove the logic entirely in a year or so when Safari 27 is released, given a "last two-three versions" support guide

## Breaking change

The `useScrollLock` utility signature was simplified to `useScrollLock(enabled: boolean, referenceElement?: Element | null)`

### DIFF
--- a/packages/react/src/utils/useScrollLock.ts
+++ b/packages/react/src/utils/useScrollLock.ts
@@ -234,31 +234,9 @@ const SCROLL_LOCKER = new ScrollLocker();
  * Locks the scroll of the document when enabled.
  *
  * @param enabled - Whether to enable the scroll lock.
+ * @param referenceElement - Element to use as a reference for lock calculations.
  */
-export function useScrollLock(params: {
-  enabled: boolean;
-  mounted: boolean;
-  open: boolean;
-  referenceElement?: Element | null;
-}) {
-  const { enabled = true, mounted, open, referenceElement = null } = params;
-
-  // https://github.com/mui/base-ui/issues/1135
-  useIsoLayoutEffect(() => {
-    if (enabled && isWebKit && mounted && !open) {
-      const doc = ownerDocument(referenceElement);
-      const originalUserSelect = doc.body.style.userSelect;
-      const originalWebkitUserSelect = doc.body.style.webkitUserSelect;
-      doc.body.style.userSelect = 'none';
-      doc.body.style.webkitUserSelect = 'none';
-      return () => {
-        doc.body.style.userSelect = originalUserSelect;
-        doc.body.style.webkitUserSelect = originalWebkitUserSelect;
-      };
-    }
-    return undefined;
-  }, [enabled, mounted, open, referenceElement]);
-
+export function useScrollLock(enabled: boolean = true, referenceElement: Element | null = null) {
   useIsoLayoutEffect(() => {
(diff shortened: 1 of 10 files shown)


packages/react/src/utils/useScrollLock.ts
```diff
@@ -98,47 +98,58 @@ function preventScrollStandard(referenceElement: Element | null) {
...
-      overflow: 'hidden',
-      scrollBehavior: 'unset',
-    });
-
-    body.scrollTop = scrollTop;
-    body.scrollLeft = scrollLeft;
-    html.setAttribute('data-base-ui-scroll-locked', '');
-    html.style.scrollBehavior = 'unset';
+    if (supportsStableScrollbarGutter) {
+      html.style.scrollbarGutter = 'stable';
```
mj12albert: https://github.com/mui/base-ui/issues/3189

Perhaps this is an easy fix to include since computed styles are already available here

packages/react/src/utils/useScrollLock.ts
```diff
@@ -219,31 +234,9 @@ const SCROLL_LOCKER = new ScrollLocker();
...
-export function useScrollLock(params: {
-  enabled: boolean;
-  mounted: boolean;
-  open: boolean;
-  referenceElement?: Element | null;
-}) {
-  const { enabled = true, mounted, open, referenceElement = null } = params;
-
-  // https://github.com/mui/base-ui/issues/1135
-  useIsoLayoutEffect(() => {
```
atomiks, line 232: This hook was no longer executing after we changed `enabled` to be `false` immediately on close (without waiting for exit animations). And the bug it was fixing isn't present now due to that, so it can be removed with the `useScrollLock` signature simplified. That also solves: https://github.com/mui/base-ui/issues/2598

Issue #2144, opened 2025-06-20
Docs search


thelulzy:
# Docs feedback

## How can we improve Base UI documentation?
Would be great to have a native search feature in the Base UI Docs similar to the search within the MUI docs.

thelulzy:
Meta discussion:
Should the "Docs feedback" Github issue template apply a Docs tag to the issue?

michaldudak:
We'll likely add it at some point before the stable release. No ETA, though.

PR #3193, opened 2025-11-12
[docs-infra] Add Search Functionality from docs-infra

dav-is:
> [!NOTE]
> This PR depends on https://github.com/mui/mui-public/pull/901 (merged)

Adds a search index and serach bar to the docs site.

Also populates navigation from the index. Whenever a new item is added to the index, it is automatically marked `[New]`. Try creating a new component page in development mode and see it automatically appear in the nav bar and search. Update the title and see it hot reload.

[Preview Deployment](https://deploy-preview-3193--base-ui.netlify.app/react/overview/quick-start)

[Components Index](https://github.com/dav-is/base-ui/blob/davis/add-search-index/docs/src/app/(docs)/react/components/page.mdx)

All the data shown in the above markdown file is included in the search index and can be used as results.

## Screenshots

Screenshot From 2025-12-09 10-12-02

Screenshot From 2025-12-09 10-11-50

Screenshot From 2025-12-09 10-11-57

image

## Performance

Index size: `15.6KB` (`75KB` uncompressed)
Index loads into memory in 15ms-20ms
Searches in less than 1ms

With very quick benchmarking, there is a small increase in the uncached build time, as would be expected to generate the index. But the cached build sees no increase in build time.
Build Time Before: `50s` / `30s` cached
Build Time After: `56s` / `30s` cached

To ensure the files are not edited manually or become out of sync, a CI test should run in parallel with the build to run `pnpm docs:validate`. if out of sync, it will instruct you what command to run.

<details>

<summary>Experiments with Vector Search (not included in this PR)</summary>

To increase matches for indirect words, like `profile` -> `avatar`, `person` -> `avatar`, etc, we could add vectors that measure similarity.

Index size grows to `500KB` / `226KB` gzip (uncompressed 470KB larger, compressed 220KB larger) 
Build Time: `75s` / `30s` cached (15 seconds slower on first build, afterwards only pages that have changed)
Requires `28MB` model to be loaded on client

This might be too much of an increase to be worth it. Without this, we can still optimize searches manually using `keywords`, which should also improve SEO.

</details>

Closes https://github.com/mui/mui-public/issues/419
Closes https://github.com/mui/base-ui/issues/2144


mj12albert:
The search functionality works pretty good, have some initial comments on the UI:

The focus ring in the search field has the wrong color:

https://github.com/user-attachments/assets/271e4e0f-85b5-4e7d-9c35-704b368f3551

As you type, the results in the list (and even the group labels) shift around very visibly (more noticeable on mobile):

https://github.com/user-attachments/assets/3d0b5507-4b06-44bb-9b61-50e8fe1b33a9

The popup opening animation feels a bit slow? It's more noticeable on desktop, it might be better for it to just appear in the center like a command picker, example: https://deploy-preview-2817--base-ui.netlify.app/react/components/autocomplete#command-palette

The popup could maintain a minimum height when there are few results, otherwise blurred content under the backdrop could be distracting:

Screenshot 2025-11-27 at 1 26 29 PM

dav-is:
@mj12albert 
> The focus ring in the search field has the wrong color

Should be fixed now

> the results in the list (and even the group labels) shift around very visibly

This was intentional, using view transitions, but it seems the consensus is that it is too distracting, so I removed it.

> The popup opening animation feels a bit slow?

James suggested using a spring easing for the popup animation. Does it feel snappier with this applied?

> The popup could maintain a minimum height when there are few results

What do you think about `min-height: 80vh`, or maybe more? Do you think we need this for desktop too?

Screenshot From 2025-11-27 13-45-06

michaldudak:
It would be nice if the search box showed <kbd>ctrl</kbd> <kbd>K</kbd> instead of <kbd>⌘</kbd> <kbd>K</kbd> on Windows

dav-is:
@michaldudak How does this look?

image

Along with this change, we can't show the search bar until hydration when we can check if it is mac or windows.

docs/src/app/(docs)/react/components/accordion/page.mdx
```diff
@@ -40,3 +40,18 @@ import { DemoAccordionMultiple } from './demos/multiple';
...
+    'React Accordion',
+    'Accordion Component',
+    'Collapsible Content',
+    'Expandable Panel',
+    'Accessible Accordion',
+    'Multi-Panel UI',
+    'Disclosure Widget',
+    'FAQ Component',
+    'Headless React Components',
+    'Base UI',
```
LukasTy, line 55: Does the `nbsp` symbol make sense here? 🤔 
If search engines will use these, maybe it would be safer to have plain space?

Image

docs/src/app/(docs)/react/components/tabs/page.mdx
```diff
@@ -30,3 +30,22 @@ import { Tabs } from '@base-ui/react/tabs';
...
+    'React Tabs',
+    'Tabbed Interface Component',
+    'Tab Panel',
+    'Tab Control',
+    'Tabbed Navigation',
+    'Tabbed UI',
+    'Page Switcher',
+    'Content Switcher',
+    'View Switcher',
+    'Carousel Alternative',
```
LukasTy, line 45: Was this suggested by AI? 🤔 
IMHO, seems like a slightly strange keyword. 🤷

colmtuite (approved):

docs/src/app/(docs)/react/components/tooltip/page.mdx
```diff
@@ -177,3 +177,23 @@ import { DemoTooltipDetachedTriggersFull } from './demos/detached-triggers-full'
...
+    'React Tooltip',
+    'Tooltip Component',
+    'Detached Trigger Tooltip',
+    'Hint',
+    'Help Text',
+    'Hover Text',
+    'Hover Hint',
+    'Info Popup',
+    'Infotip',
+    'Flyout',
```
LukasTy, line 192: Isn't flyout usually used in conjunction with `menu`? 🤔 
I'd say that this keyword feels slightly strange for this component. 🙈

docs/src/app/(docs)/(content)/layout.tsx
```diff
@@ -2,29 +2,47 @@ import * as React from 'react';
...
-        ))}
+        {sitemap &&
+          Object.entries(
+            sitemap.data as Record<
+              string,
+              {
+                title?: string;
+                prefix?: string;
+                pages: { title: string; tags?: string[]; path: string }[];
+              }
```
LukasTy: Nitpick: Consider defining outside of the JSX to improve readability.

docs/src/app/(docs)/react/components/tabs/page.mdx
dav-is: Yes, I asked it to come up with as many synonyms as it could

LukasTy: Gotcha. Not a blocker, and if no one else complained, maybe it's good to go. 🤷

mnajdova (approved):
One small nit is that I would focus the input on mobile when opening the dialog, so people don't have to click on it before typing.

From the behaviour side I couldn't spot any issues.

dav-is:
Merging now, we'll address this feedback in a followup

PR #3201, opened 2025-11-13
[combobox] Fix stuck filtering with differing stringifiers

### LINKED ISSUE #3192: [combobox] Filtered items stuck after making a selection when filtering by multiple obj properties
https://stackblitz.com/edit/ve6npfxd?file=src%2FApp.tsx

A fn that's different from `itemToStringValue` is passed to the `contains` filter in order to filter by 2 different object properties.

After making a selection, closing the popup, then reopening the popup, the previous filtered items appear instead of all items 

e.g. Items are movie objects with different properties for the title in diff languages, I want it to be filterable by any language

atomiks:

Fixes #3192 

This normalizes the query used for item filtering so that when a single selection combobox reopens and the input still shows the selected label, the query is treated as empty and show the full list again (even if a custom filter uses a different stringifier that makes the combobox thinks it's currently filtering given there isn't a match with the selected value's label). This compares the visible label against the trimmed query with the collator and falls back to unfiltered data before invoking user filters

### DIFF
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -155,6 +155,7 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
   const disabled = fieldDisabled || disabledProp;
   const name = fieldName ?? nameProp;
   const multiple = selectionMode === 'multiple';
+  const single = selectionMode === 'single';
   const hasInputValue = inputValueProp !== undefined || defaultInputValueProp !== undefined;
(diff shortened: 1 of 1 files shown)


mj12albert (approved):
Works nicely now 👍
https://stackblitz.com/edit/ve6npfxd-f1hdlm7u?file=src%2FApp.tsx

This is somewhat common when the dataset can be searchable by the name/label and also some kind of abbreviation/acronyn/code
```
{
    name: 'Haneda Airport',
    city: 'Tokyo',
    country: 'Japan',
    iso2: 'JP',
    code: 'HND',
  }
```

Issue #3143, opened 2025-11-05
[radio group] Allow controlling the render of the hidden `input` as well, not just the `button`


benface:
# Feature request

## Summary

When using `Radio.Root`'s `render` prop, whatever we return only replaces the `button` that `Radio.Root` normally renders, but there's no way to control where/how to render the hidden `input` that it renders beside it. I would expect that level of control from a headless component.

## Examples in other libraries

Not sure.

## Motivation

I would like to wrap the `button` and `input` that `Radio.Root` renders in a `div` or something so that Tailwind classes starting with `last:` or `only:` work as expected (they don't because the `button` which receives the class is _never_ the last or only child, due to the `input` that gets rendered beside it). Of course, I could wrap `Radio.Root` itself and set the `last:`/`only:` classes on the wrapper `div`, but in my case, I actually want to set the classes conditionally on `state` that is only available in `Radio.Root`'s `render` (namely `checked`)...

Thank you for considering this. 🙏

mj12albert:
@benface Are you talking about the `input` that's a sibling of the `button role="radio"` or the one that's a sibling of the `div role="radiogroup"`? You just reminded me that the sibling-of-button is redundant, as we don't intend to support a standalone group-less `<Radio>` 😅

benface:
The sibling of the `button role="radio"`! Wow sorry @mj12albert I just noticed that I called it `RadioGroup` 5 times in the original post, just changed to `Radio.Root`. 😅

mj12albert:
For Radio, removing the redundant input should solve the issue

But I think Checkbox has the same issue and it's sibling-to-button input isn't redundant, in which case the only "workaround" is to use `Field.Item` or an implicit label and set `first/last:` on that 🤔

benface:
Yes, please consider all components that use this pattern; I guess I haven't been bitten by `Checkbox` yet but it's probably a matter of time.

Does `Field.Item` have access the `checked` state of the checkbox? If not, it's not a working workaround for the problem I described with `Radio`.

mj12albert:
> wrap the `button` and `input` that `Radio.Root` renders in a `div` or something so that Tailwind classes starting with `last:` or `only:` work as expected

Is this what you're trying to do?https://stackblitz.com/edit/vm4oh1kt?file=src%2FApp.tsx

[...]

PR #3203, opened 2025-11-13
[docs] Fix incorrect usage of `label` and `Field.Label` in `Checkbox`, `Radio`, and `Switch`

battaglr:

### Description

The current examples and guidelines show wrapping components in a `<label>`, which is invalid for components that render both a `<button>` and an `<input>` (as Checkbox, Radio, and Switch do).

> [!NOTE]
> The HTML spec for [`label`](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) defines its content model as:
> > Phrasing content, but with no descendant labelable elements unless it is the element's labeled control, and no descendant label elements.
>
> That's why the example below fails the [W3C Markup Validator](https://validator.w3.org/):
>
> ```html
> <label>
>   <button>Text</button>
>   <input type="checkbox" />
> </label>
> ```

This PR includes:
- Adding a usage-guideline bullet clarifying that these components must not be wrapped in a `<label>`.
- Updating the recent `Form` integrations page to remove references to wrapper labels.
- Updating examples in `Checkbox`, `Radio`, and `Form` to use independent labels.
- Adding a new demo for `Radio` + `Form`.

> [!CAUTION]
> These changes make the bug described in #2691 more obvious, since labels associated with `<Checkbox parent />` don’t work. The good news is that PR #2896 appears to address that issue.

### DIFF
--- a/docs/src/app/(public)/(content)/react/handbook/forms/demos/components/field.tsx
+++ b/docs/src/app/(public)/(content)/react/handbook/forms/demos/components/field.tsx
@@ -6,15 +6,21 @@ export function Root({ className, ...props }: Field.Root.Props) {
   return <Field.Root className={clsx('flex flex-col items-start gap-1', className)} {...props} />;
 }
 
-export function Label({ className, ...props }: Field.Label.Props) {
+export function Item({ className, ...props }: Field.Item.Props) {
   return (
-    <Field.Label
+    <Field.Item
       className={clsx(
-        'text-sm font-medium text-gray-900 has-[[role="checkbox"]]:flex has-[[role="checkbox"]]:items-center has-[[role="checkbox"]]:gap-2 has-[[role="radio"]]:flex has-[[role="radio"]]:items-center has-[[role="radio"]]:gap-2 has-[[role="switch"]]:flex has-[[role="switch"]]:items-center has-[[role="radio"]]:font-normal',
+        'has-[[role="checkbox"]]:flex has-[[role="checkbox"]]:items-center has-[[role="checkbox"]]:gap-2 has-[[role="radio"]]:flex has-[[role="radio"]]:items-center has-[[role="radio"]]:gap-2 has-[[role="switch"]]:flex has-[[role="switch"]]:items-center has-[[role="radio"]]:font-normal',
         className,
       )}
       {...props}
     />
+  );
+}
+
+export function Label({ className, ...props }: Field.Label.Props) {
+  return (
+    <Field.Label className={clsx('text-sm font-medium text-gray-900', className)} {...props} />
   );
 }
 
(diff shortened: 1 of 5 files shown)


atomiks:
I wonder if these components should render `<span>` by default instead; the implicit labeling was an intentional feature we added support for since it works neatly without Field.

mj12albert:
> I wonder if these components should render `<span>` by default instead

If we did this it could render
```jsx
<span role="checkbox">
  <input aria-hidden="true" />
</span>
```
Which would also solve: https://github.com/mui/base-ui/issues/3143

And the only breaking change would be button->span; however when explicitly labelled and the label doesn't wrap `span role="checkbox"`, the ability to trigger the control by clicking the label is lost

mj12albert:
> * Adding a usage-guideline bullet clarifying that these components must not be wrapped in a `<label>`

And since this is [natively supported](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label#associating_a_label_with_a_form_control) I think we can't explicitly forbid it whether `<label>` or `<Field.Label>` is involved @battaglr

battaglr:
@atomiks 

> I wonder if these components should render `<span>` by default instead

Absolutely! That would address the root issue.

@mj12albert 
> And since this is [natively supported](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label#associating_a_label_with_a_form_control) I think we can't explicitly forbid it whether `<label>` or `<Field.Label>` is involved

Totally agree!

@mj12albert 
> If we did this it could render: [span>input]

I like that. A 1:1 output for each component is a more predictable API and gives implementers better control.

---

My intention with this PR was to avoid invalid HTML based on the current implementation, but updating the implementation itself would definitely be the ideal fix.

Would you prefer I close this and open an issue outlining the problem?

battaglr:
@mj12albert 

> however when explicitly labelled and the label doesn't wrap span `role="checkbox"`, the ability to trigger the control by clicking the label is lost

I missed that. I don’t think we should settle for losing either implicit or explicit label behavior.

Hmm… could we forward the user-defined `id` from any of these components to the underlying `input` and somehow capture whatever focus state it receives? 🤔

battaglr:
Closing in favor of https://github.com/mui/base-ui/pull/3205.

Dakkers:
> however when explicitly labelled and the label doesn't wrap span role="checkbox", the ability to trigger the control by clicking the label is lost

is there a workaround for this as it stands? my markup looks like this (some css/style omitted for brevity)

Screenshot 2025-12-05 at 12 35 03

---

markup for just the radio button:

```
<label for="_r_8_" id="_r_8_-label" style="">
  <span data-unchecked="" role="radio" tabindex="-1" aria-checked="false"
    aria-labelledby="_r_8_-label" id="base-ui-_r_9_" style="">
    <div style="">
      <div style=""></div>
    </div>
  </span>
  <input id="_r_8_" tabindex="-1" aria-hidden="true" type="radio" style="">
  <div style="">Item 3</div>
</label>
```

mj12albert:
> when explicitly labelled and the label doesn't wrap span role="checkbox"

This still works after the change, there's a test here: https://github.com/mui/base-ui/blob/82b925e8bc8c5a588437c271c3d6060372b5d698/packages/react/src/radio-group/RadioGroup.test.tsx#L481-L511

Your markup looks more or less the same as the example in https://base-ui.com/react/handbook/forms

Maybe you're using `Field.*` components and left out [`Field.Item`](https://base-ui.com/react/handbook/forms#:~:text=Optionally%20use%20%3CField.Item%3E%20in%20checkbox%20or%20radio%20groups%20to%20individually%20label%20each%20control%20when%20not%20implicitly%C2%A0labeled%3A)? @Dakkers

PR #3205, opened 2025-11-13
[checkbox][radio][switch] Render root as `span` instead of `button`

mj12albert:
**Breaking:** `Checkbox.Root`, `Radio.Root` and `Switch.Root` renders `<span>` instead of `<button>`

Labels can't wrap both a `button` and the hidden `input` since it's [invalid HTML](https://validator.w3.org/nu/?doc=https%3A%2F%2Fbase-ui.com%2Freact%2Fcomponents%2Fradio)

Screenshot 2025-11-13 at 4 16 13 PM

WAVE also reports errors on the `button`:
Screenshot 2025-11-17 at 3 03 02 PM

- ~Clicking the label to trigger the radio/checkbox/switch doesn't work if it's a `span`, but this is added back via `Field.Label`~ actually works
- HTML validator flags `aria-required` on `role="checkbox|switch"` as invalid, but VO announces the required state correctly so I think it should be kept for checkbox and switch.

Demos:
- https://deploy-preview-3205--base-ui.netlify.app/experiments/forms/button-controls
- HTML validator: https://validator.w3.org/nu/?doc=https%3A%2F%2Fdeploy-preview-3205--base-ui.netlify.app%2Freact%2Fcomponents%2Fradio

Alternative to https://github.com/mui/base-ui/pull/3203


atomiks:
> Clicking the label to trigger the radio doesn't work if it's a span, but this can be added back via FieldItem

It "works" for me in the example, but the focus state doesn't sync properly. `onFocus()` on the hidden input to focus the `<span>` seems like it'd be a workaround (similar to `<Select.Trigger>` behavior)

mj12albert:
This is still invalid nesting of interactive elements:
```
<span role="radio">
  <input type="radio" aria-hidden />
</span>
```
Need to try to get `input type="hidden"` to work

atomiks:
On https://deploy-preview-3205--base-ui.netlify.app/experiments/forms/button-controls,
Clicking the label with the mouse doesn't set focus on the radio `<span>` control, so using <kbd>ArrowUp</kbd>/<kbd>ArrowDown</kbd> doesn't work. (There's also a mismatch in the focus style when tabbing into it and using the controls)

mj12albert:
There's also an issue where the `label`'s `for` must match the `input`'s `id` instead of the `id` of the root element, in which case it should be labelled with `aria-labelledby="(label id)"`

Screenshot 2025-11-14 at 6 45 42 PM

atomiks:
@mj12albert yeah, `Select` forwards the `id` to the hidden input. Clicking the label focuses the hidden input, so `onFocus` is used to immediately relocate it to the trigger. I think that's the correct solution when using non-`<button>` triggers (although Select was recently changed to be a native `<button>`)

mj12albert:
> `Select` forwards the `id` to the hidden input.

OK doign the same here but a gotcha with this is that `<Select.Trigger id="MyTrigger" />` id's the input and not the actual trigger, though I suppose manual ids are an edge case in 2025...

packages/react/src/switch/root/SwitchRoot.tsx
```diff
@@ -144,13 +147,23 @@ export const SwitchRoot = React.forwardRef(function SwitchRoot(
...
       }
 
       event.preventDefault();
+
       inputRef?.current?.click();
     },
+    onKeyDown(event) {
+      if (nativeButton || event.defaultPrevented || readOnly || disabled) {
+        return;
+      }
```
atomiks: What's the reason for this handler? Shouldn't `useButton` already call `onClick`?

mj12albert: useButton prevents default in onKeyDown + Enter key: https://github.com/mui/base-ui/blob/56dee07c3e40c396a30e72752f630069a233634d/packages/react/src/use-button/useButton.ts#L130-L132

So the click in Switch doesn't fire because it returns if default prevented: https://github.com/mui/base-ui/blob/56dee07c3e40c396a30e72752f630069a233634d/packages/react/src/switch/root/SwitchRoot.tsx#L146-L149

It's the same for checkbox, both of these are necessary though I think

atomiks: > It's the same for checkbox, both of these are necessary though I think

I think these were added before `.preventBaseUIHandler` was a thing. I think it's generally better to not check `event.preventDefault` as much as possible (at least we seemed to rarely do it in later components)

atomiks:
> Clicking a plain sibling <label> (even with for/id set correctly) is unable to toggle span role="checkbox"

It seems to work with a minimal repro:

```jsx
export default function Checkbox() {
  const ref = React.useRef<HTMLSpanElement>(null);
  const [checked, setChecked] = React.useState(false);
  return (
    <div>
      <label htmlFor="checkbox">Checkbox</label>
      <span
        ref={ref}
        role="checkbox"
        aria-checked={checked}
        style={{
          display: 'inline-block',
          width: 20,
          height: 20,
          background: checked ? 'blue' : 'gray',
        }}
        onClick={() => setChecked(!checked)}
        tabIndex={0}
      />
      <input
        id="checkbox"
        onFocus={() => {
          ref.current?.focus();
        }}
        onClick={() => {
          ref.current?.click();
        }}
        tabIndex={-1}
        aria-hidden
      />
    </div>
  );
}
```

mj12albert:
> > Clicking a plain sibling  (even with for/id set correctly) is unable to toggle span role="checkbox"
> 
> It seems to work with a minimal repro:

Oh yeah it works, I didn't update this comment yet 😅 
Originally I thought it wasn't working because there is a weird behavior where an identical test for Switch fails for Checkbox because it's flaky when using `user.click`

docs/src/app/(private)/experiments/forms/button-controls.tsx
```diff
@@ -90,17 +90,23 @@ function StickersRadioGroup() {
       </Field.Item>
 
       <Field.Item className={styles.FieldItem}>
```
atomiks, line 92: Maybe it's worth having four separate sections here (unless already covered by unit tests):

- `<Field.Label>` sibling
- `<Field.Label>` enclosing
- `<label>` sibling
- `<label>` enclosing

mj12albert: Yeah I added all the the variations in unit tests to avoid future surprises 😬

mj12albert:
Features all seem intact, the only issue now is `<label><span role="checkbox" /></label>` like in the demos messes up the accName:
Screenshot 2025-11-17 at 2 59 15 PM

Manually setting `htmlFor`/`id` doesn't help because the `id` is forwarded to the `input`, so it doesn't link the span, only `aria-label` on the span works... what do you think? @atomiks

docs/src/app/(public)/(content)/react/components/checkbox-group/page.mdx
```diff
@@ -47,40 +47,6 @@ import { DemoCheckboxGroupNested } from './demos/nested';
 
 <DemoCheckboxGroupNested compact />
 
-### Form integration
```
mj12albert, line 50: The Forms page has this now

atomiks:
> in the demos messes up the accName:

Is this more so an axe false positive? Without `aria-label`, the enclosing label announces its text when focus is on the `<span>` checkbox. If so, I think it's okay since there are workarounds to satisfy it if necessary

mj12albert:
Alright, with VO the differences seem negligible anyway

atomiks (approved):

hsimpson270:
I have updated my project to use beta 5, but I am seeing a conflict because of this change. `CheckboxRootProps` was updated to extend `BaseUIComponentProps<'span', CheckboxRoot.State>,` but `CheckboxRoot` was not updated as well, so it still typed as `React.ForwardRefExoticComponent<CheckboxRootProps & React.RefAttributes<HTMLButtonElement>>`. This causes an issue because `CheckboxRoot` is expected to extend a button, but its props are expected to extend a span.

In my implementation, updating my custom props to extend either a button or a span both result in errors when passed into `CheckboxRoot`. I was using button previously, which now generates a "Types of property 'onCopy' are incompatible" error (I'm assuming because that's the first native prop that conflicts), but if I update to use span instead I get a "Types of property 'ref' are incompatible." error. This error completely breaks the current implementation, so the only work around I can think of for now would be to not extend a native HTML element in my props at all.

PR #3217, opened 2025-11-14
[docs] Fix detached triggers demos layout

michaldudak:
Added `flex-wrap` to detached trigger demos (Dialog, AlertDialog, Tooltip, and Popover) so they look better on narrow screens.

Before:
image

After:
image

Preview:
- https://deploy-preview-3217--base-ui.netlify.app/react/components/alert-dialog#controlled-mode-with-multiple-triggers
- https://deploy-preview-3217--base-ui.netlify.app/react/components/dialog#controlled-mode-with-multiple-triggers
- https://deploy-preview-3217--base-ui.netlify.app/react/components/popover#controlled-mode-with-multiple-triggers
- https://deploy-preview-3217--base-ui.netlify.app/react/components/tooltip#controlled-mode-with-multiple-triggers


docs/src/app/(public)/(content)/react/components/tooltip/demos/index.module.css
```diff
@@ -1,13 +1,14 @@
 .Container {
   display: flex;
   gap: 0.5rem;
+  flex-wrap: wrap;
```
LukasTy, line 4: WDYT about also adding a `justify-content: center;` to keep the center alignment when wrapped? 🤔

michaldudak: Good point!

PR #3221, opened 2025-11-14
[test] Make the tests more reliable

michaldudak:
- Added a couple of `wait`s
- Added a new way to check if the menubar menu is open (data attribute). It might be useful outside tests as well.
- Fixed the `key` prop spread warnings

Looks like the main culprit was that keyboard events were dispatched slightly too early, before handlers were set.


packages/react/src/menubar/Menubar.test.tsx
```diff
@@ -8,6 +8,12 @@ import { Menubar } from '@base-ui-components/react/menubar';
 import { Menu } from '@base-ui-components/react/menu';
 import { useRefWithInit } from '@base-ui-components/utils/useRefWithInit';
 
+async function wait(ms: number) {
```
LukasTy, line 11: Would it make sense to move this to `test-utils`? 🤔

michaldudak: I'm not keen on this because of how ugly this hack is :/

michaldudak: But yeah, it seems inevitable for now. I added it in https://github.com/mui/base-ui/pull/3226

PR #3226, opened 2025-11-14
[test] Make the tests more reliable, part 2

michaldudak:
Added short delays to TooltipViewport tests


packages/react/src/tooltip/viewport/TooltipViewport.test.tsx
```diff
@@ -207,9 +207,13 @@ describe('<Tooltip.Viewport />', () => {
       const trigger2 = screen.getByTestId('trigger2');
       const trigger3 = screen.getByTestId('trigger3');
 
+      await waitSingleFrame();
```
LukasTy, line 210: These are not pretty.  🙈 
Have you tried adding assertions to check if whatever needs to render after a focus change has rendered instead of a direct timeout? 🤔

michaldudak: I don't think it's about rendering - it seems that event handlers are not set up by the time events are fired in tests.

PR #3231, opened 2025-11-16
[all components] Fix crash in Next.js 16 when accessing `render.props.ref`

### LINKED ISSUE #3144: getChildRef function of useRenderElement is causing unexpected crashes in React 19 + Next 16
# Bug report

## Current behavior

This function errors when executing render.props.ref. It only happens in built production instance of Next JS 16 with cache components enabled. `render.props` is undefined causing the crash. Likely because render getChildRef is applied on a server component passed through to client side render method. 

```tsx
function getChildRef(render) {
  if (render && typeof render !== 'function') {
    return isReactVersionAtLeast(19) ? render.props.ref : render.ref;
  }
  return null;
}
```

## Expected behavior

Should not crash

## Reproducible example

## Base UI version

beta-4

Chrome

## Which OS are you using?

Macos

atomiks:

Fixes #3144

### DIFF
--- a/packages/utils/src/useMergedRefs.test.tsx
+++ b/packages/utils/src/useMergedRefs.test.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { expect } from 'chai';
 import { createRenderer, MuiRenderResult, screen } from '@mui/internal-test-utils';
 import { spy } from 'sinon';
+import { getReactElementRef } from './getReactElementRef';
 import { useMergedRefs } from './useMergedRefs';
 
 describe('useMergedRefs', () => {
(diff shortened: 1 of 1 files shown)


packages/utils/src/getReactElementRef.ts
michaldudak, line 1: Why not fix the utils function instead? It can be used in other projects, so removing it is a breaking change.

atomiks: We weren't using it internally, and I don't think other projects would've been either. If it's requested, we can export it, but I suspect it won't be used given it's a low level util. `utils` already has a breaking change this release, so it's fine I think

michaldudak: Material UI has its own copy of it, so it's useful for sure (they don't use `@base-ui-components/utils` yet). Our utils are generally low-level, and everything not Base UI-specific that can come in handy in other MUI projects should be placed there. Can you see any advantage of removing it from the shared package?

atomiks: Not really, we can keep it and make the internal module import from it now.

PR #3237, opened 2025-11-17
[alert dialog] Re-export Dialog parts

michaldudak:
Re-exporting Dialog parts instead of defining them for AlertDialog specifically to save bundle size.

### DIFF
--- a/packages/react/src/alert-dialog/index.ts
+++ b/packages/react/src/alert-dialog/index.ts
@@ -1,11 +1,3 @@
 export * as AlertDialog from './index.parts';
 
 export type * from './root/AlertDialogRoot';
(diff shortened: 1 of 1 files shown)


michaldudak:
No, there are no public-facing changes here. The proposal in #2265 is to add the `alert` prop to the Dialog.Root. This PR still keeps the AlertDialog.Root and doesn't change the API at all.

LukasTy:
> No, there are no public-facing changes here. The proposal in #2265 is to add the `alert` prop to the Dialog.Root. This PR still keeps the AlertDialog.Root and doesn't change the API at all.

Gotcha. Well, I was referring to the fact that the underlying result is the optimisation of code.

PR #3238, opened 2025-11-17
[code-infra] Build packages with react compiler enabled

brijeshb42:
Added tests with React compiler enabled solely on CI to make sure building with react compiler doesn't break any component.

The two new jobs (`test_unit_compiler` and `test_browser_compiler`) perform the build with full react compiler transform and then run the tests without aliasing to the source files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,19 @@ default-context: &default-context
 #   restore_cache:
 #     key: v1-repo-{{ .Branch }}-{{ .Revision }}
 
+commands:
+  build-with-compiler:
+    description: 'Build packages with React Compiler'
+    steps:
+      - run:
+          name: Add compiler package
+          command: pnpm -F "./packages/*" add react-compiler-runtime
+      - run:
+          name: Build with compiler
+          command: pnpm -F "./packages/*" build --enableReactCompiler
+          environment:
+            MUI_REACT_COMPILER_MODE: 'infer'
+
 jobs:
   test_unit:
     <<: *default-job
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -226,6 +267,12 @@ workflows:
       - test_package:
           <<: *default-context
           name: 'Package verification'
+      - test_unit_compiler:
+          <<: *default-context
+          name: 'Unit tests (React Compiler)'
```
LukasTy:
```suggestion
          name: 'JSDOM tests (React Compiler)'
```

test/vite.shared.config.mjs
```diff
@@ -2,13 +2,19 @@ import * as path from 'path';
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
 
+const isReactCompilerTest = Boolean(process.env.MUI_REACT_COMPILER_TEST);
```
Janpot: nitpick, but better to name the flag after what it enables/disables rather than the purpose it's currently used for. e.g.

```suggestion
const isReactCompilerTest = Boolean(process.env.MUI_DISABLE_WORKSPACE_ALIASES);
```

scripts/addCompilerPackage.mts
```diff
@@ -0,0 +1,16 @@
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+
+async function addCompilerPackage() {
```
Janpot: To keep it simpler, would it work if we just install it top-level? Or provide it in vitest config?

brijeshb42: I've already kept it in top-level package.json to avoid installing packages a 2nd time. But we still need it for the build tool to not fail since that throws if the dependency is not declared.

.circleci/config.yml
```diff
@@ -197,6 +197,47 @@ jobs:
...
+    <<: *default-job
+    executor:
+      name: code-infra/mui-node-browser
+      playwright-img-version: v1.56.1-noble
+    steps:
+      - checkout
+      - code-infra/install-deps
+      - run:
+          name: Add compiler package
+          command: pnpm -r add-compiler-package
```
Janpot: Wouldn't this suffice?
```suggestion
          command: pnpm -F "./packages/*" add react-compiler-runtime
```

scripts/addCompilerPackage.mts
Janpot: 👍 yeah, just realized after reviewing your other PR

.circleci/config.yml
brijeshb42: I was hopeful for a bit but was disappointed when I tried. This is same as doing `pnpm -r add-compiler-package && pnpm install` since it actually downloads the new dep.

Janpot: shouldn't it install it?

brijeshb42: It's not needed now specifically for the packages since its already installed at the repo level.

Janpot: If we can't get around having these dependency in package.json, then I would remove it from root and just install it. The few seconds we lose are worth not working with a project where the actually installed dependencies don't correspond with the specified ones.

Issue #3043, opened 2025-10-23
[tabs] NextJS 16 `cacheComponents` error when `TabsIndicator` generates id


mj12albert:
Repro: https://codesandbox.io/p/github/joncoronel/next-16-baseui-tabs-error/master

Tabs indicator throws this error in NextJS 16 with the new [`cacheComponents`](https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheComponents) feature

Possibly `useId` can be used instead to generate the instance id: https://github.com/mui/base-ui/blob/3e3914cc393d78b17bbc76be936f615538315d28/packages/utils/src/generateId.ts

```sh
Error: Route "/" used `Math.random()` inside a Client Component without a Suspense boundary above it. See more info here: https://nextjs.org/docs/messages/next-prerender-random-client
    at Home (app/page.tsx:29:11)
  27 |             Account
  28 |           </Tabs.Tab>
> 29 |           <Tabs.Indicator className="absolute top-1/2 left-0 z-[-1] h-6 w-[var(--active-tab-width)] translate-x-[var(--active-tab-left)] -translate-y-1/2 rounded-sm bg-gray-100 transition-all duration-200 ease-in-out" />
     |           ^
  30 |         </Tabs.List>
```

Originally reported [in Discord](https://discord.com/channels/1287292451308048406/1287295829887025224/1430767532008275968)

PR #3241, opened 2025-11-17
[alert dialog][dialog][menu][popover][tooltip] Do not use client-side code in stores

michaldudak:
The Store classes created for Dialog, AlertDialog, Tooltip, Popover, and Menu used `getEmptyContext` defined in useFloatingRootContext and `FloatingTreeStore` from `FloatingTree`. These modules are client-only as they use context APIs, and this made the stores (and, by extension, the root components' exports) fail on the server.

This fix moved `FloatingTree` and `getEmptyContext` to separate modules (+ renamed to `useEmptyRootContext`). It also adds missing Handle types to exports of the relevant components.

---

To test, create a Next.js app with app router and use any of the affected components on a server route.

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingTreeStore.ts
+++ b/packages/react/src/floating-ui-react/components/FloatingTreeStore.ts
@@ -10,8 +10,6 @@ export class FloatingTreeStore<RT extends ReferenceType = ReferenceType> {
 
   public readonly events: FloatingEvents = createEventEmitter();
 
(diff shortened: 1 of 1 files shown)


dmitri-gb:
@michaldudak `FloatingTreeStore` also calls `Math.random()` (though seemingly the result is unused), which makes Next.js 16 with `cacheComponents` enabled think that all those root components should be rendered inside `Suspense`. (Same issue that was reported in https://github.com/mui/base-ui/issues/3043). Could you tackle this issue here as well?

PR #3242, opened 2025-11-17
[select] Remove `type="button"` from trigger

mj12albert:
From https://github.com/mui/base-ui/pull/2718#issuecomment-3543197727, a `button` with `role="combobox"` doesn't need the `type` attribute and causes invalid HTML

HTML validator results: https://validator.w3.org/nu/?doc=https%3A%2F%2Fbase-ui.com%2Freact%2Fcomponents%2Fselect

Screenshot 2025-11-18 at 2 11 57 AM

With `type` removed: https://validator.w3.org/nu/?doc=https%3A%2F%2Fdeploy-preview-3242--base-ui.netlify.app%2Freact%2Fcomponents%2Fselect

Screenshot 2025-11-18 at 2 16 08 AM

### DIFF
--- a/packages/react/src/select/trigger/SelectTrigger.tsx
+++ b/packages/react/src/select/trigger/SelectTrigger.tsx
@@ -1,5 +1,7 @@
 'use client';
 import * as React from 'react';
+import { isWebKit } from '@base-ui-components/utils/detectBrowser';
+import { EMPTY_OBJECT } from '@base-ui-components/utils/empty';
 import { ownerDocument } from '@base-ui-components/utils/owner';
 import { useTimeout } from '@base-ui-components/utils/useTimeout';
 import { useStableCallback } from '@base-ui-components/utils/useStableCallback';
@@ -228,8 +230,9 @@ export const SelectTrigger = React.forwardRef(function SelectTrigger(
       // Set the role last to ensure nested useButton does not overwrite it
       // e.g. <Toolbar.Button render={<Select.Trigger />} />
       role: 'combobox',
-      // combobox role doesn't need the type="button" attribute
-      type: undefined,
+      // In HTML `button role="combobox"` doesn't need `type="button"`, but Safari
+      // will still consider missing `type` to be an implicit submit button
+      ...(isWebKit ? EMPTY_OBJECT : { type: undefined }),
     },
   );
 


packages/react/src/select/trigger/SelectTrigger.tsx
```diff
@@ -225,12 +225,15 @@ export const SelectTrigger = React.forwardRef(function SelectTrigger(
     validation.getValidationProps,
     elementProps,
     getButtonProps,
+    {
+      // ensure nested useButton does not overwrite the combobox role:
+      // <Toolbar.Button render={<Select.Trigger />} />
+      role: 'combobox',
```
michaldudak, line 232: `role` is already specified above, so it can be removed from there.

mj12albert: Updated ~

packages/react/src/select/trigger/SelectTrigger.tsx
```diff
@@ -225,12 +224,15 @@ export const SelectTrigger = React.forwardRef(function SelectTrigger(
     validation.getValidationProps,
     elementProps,
     getButtonProps,
+    {
+      // Set the role last to ensure nested useButton does not overwrite it
+      // e.g. <Toolbar.Button render={<Select.Trigger />} />
+      role: 'combobox',
+      // combobox box doesn't need the type="button" attribute
+      type: undefined,
```
atomiks: Typo here. So browsers read `role="combobox"` and don't make it a submit button implicitly?

mj12albert: Good you pointed this out, forgot to check Safari at first and apparently only Safari (iOS and MacOS) submits the form 🥲

packages/react/src/select/trigger/SelectTrigger.tsx
```diff
@@ -228,8 +230,9 @@ export const SelectTrigger = React.forwardRef(function SelectTrigger(
       // Set the role last to ensure nested useButton does not overwrite it
       // e.g. <Toolbar.Button render={<Select.Trigger />} />
       role: 'combobox',
-      // combobox role doesn't need the type="button" attribute
-      type: undefined,
+      // In HTML `button role="combobox"` doesn't need `type="button"`, but Safari
+      // will still consider missing `type` to be an implicit submit button
+      ...(isWebKit ? EMPTY_OBJECT : { type: undefined }),
```
atomiks, line 235: Won't this cause a hydration error in Safari? It may be best to just leave the HTML validator error after all

mj12albert: Yeah it's better for this to be done in userland

PR #3246, opened 2025-11-17
[checkbox][switch][radio] Correct the type of forwarded refs

michaldudak:
Fixes https://github.com/mui/base-ui/pull/3205#issuecomment-3543275783

### DIFF
--- a/packages/react/src/select/positioner/SelectPositioner.test.tsx
+++ b/packages/react/src/select/positioner/SelectPositioner.test.tsx
@@ -6,9 +6,9 @@ import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
 
 const Trigger = React.forwardRef(function Trigger(
   props: Select.Trigger.Props,
-  ref: React.ForwardedRef<HTMLDivElement>,
+  ref: React.ForwardedRef<HTMLButtonElement>,
 ) {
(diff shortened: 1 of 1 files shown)


packages/react/src/checkbox/root/CheckboxRoot.tsx
```diff
@@ -352,7 +352,7 @@ export interface CheckboxRootState extends FieldRoot.State {
 }
 
 export interface CheckboxRootProps
-  extends NativeButtonProps,
+  extends NonNativeButtonProps,
```
atomiks, line 355: RadioRoot also needs this update (its ref is correct)

packages/react/src/switch/root/SwitchRoot.tsx
```diff
@@ -31,7 +31,7 @@ import { useValueChanged } from '../../utils/useValueChanged';
  */
 export const SwitchRoot = React.forwardRef(function SwitchRoot(
   componentProps: SwitchRoot.Props,
-  forwardedRef: React.ForwardedRef<HTMLButtonElement>,
+  forwardedRef: React.ForwardedRef<HTMLSpanElement>,
```
atomiks: `HTMLElement` alone may be better to allow changing to `<button>`, which `<Button>` itself does

PR #3247, opened 2025-11-17
[combobox][select] Fix value type error with mergeProps

### LINKED ISSUE #3240: mergeProps can't infer value with Combobox.Root
```ts
import { Combobox } from '@base-ui-components/react/combobox';
import { mergeProps } from '@base-ui-components/react/merge-props';

mergeProps<typeof Combobox.Root<any>>({ value: null }, {});
```
I get:
```ts
Type 'null' is not assignable to type 'undefined'.
```

## Base UI version

Beta 5

atomiks:

Fix #3240

### DIFF
--- a/docs/reference/generated/combobox-root.json
+++ b/docs/reference/generated/combobox-root.json
@@ -13,9 +13,9 @@
       "detailedType": "Value[] | Value | null | undefined"
     },
     "value": {
-      "type": "Value[] | Value",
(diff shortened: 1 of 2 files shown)


atomiks:
@mnajdova  the overloads seemed necessary for `onValueChange` to have correct inference in all the cases the `spec.tsx` file has, but if we find an alt solution that still lets them all pass, that'd be better for sure

PR #3250, opened 2025-11-18
[popups] Prefer non-adaptive anchoring position and fix `autoFocus` scroll jumps

### LINKED ISSUE #1525: [all components] The `autoFocus` prop should be controlled by each component
# Bug report

## Current behavior

The native `autofocus` DOM prop doesn't wait for anchored popups to be positioned. As a result, the page scrolls to the top when an anchored popup is positioned.

## Expected behavior

Avoid forwarding the prop to the native elements and use an effect to manage the prop inside the component

## Reproducible example

https://codesandbox.io/p/sandbox/7j567w?file=%2Fsrc%2FApp.tsx%3A4%2C1

The `initialFocus` prop is preferred to choose which element to focus in a Popover (which avoids this issue), but the `autoFocus` prop can be used on Base UI components which causes this problem. This won't work on native elements though – the `initialFocus` prop is preferred. Not sure if this should be handled as a result.

## Base UI version

1.0.0-alpha.6

## Which OS are you using?

Mac OS/Windows/Other etc.

Voiceover/JAWS etc.

atomiks:
`adaptiveOrigin` has some drawbacks when the height of the document changes (for `top` position). This PR changes it so it's only applied to shared animated tooltips/popovers (with detached triggers) by using a simple heuristic checking if `Positioner` has a transition applied.

Example: on the docs, hover over an API reference row's type and tap to dismiss once the tooltip shows. It incorrectly shifts downward by the collapsible expansion amount before unmounting (which creates a visible flicker), which isn't the case when it's anchored to the top.

This also fixes a bug where the styles would lack `top: 0, left: 0` in the initial styles, which can cause issues with positioning (it had `undefined: '0px'` incorrectly)

There is also now support for  `autoFocus` without scroll jumps by defaulting to `position: fixed`

Fixes #1525

### DIFF
--- a/packages/react/src/utils/useAnchorPositioning.ts
+++ b/packages/react/src/utils/useAnchorPositioning.ts
@@ -423,7 +423,7 @@ export function useAnchorPositioning(
   const floatingStyles = React.useMemo<React.CSSProperties>(
     () =>
       adaptiveOrigin
-        ? { position: resolvedPosition, [sideX]: `${x}px`, [sideY]: `${y}px` }
+        ? { position: resolvedPosition, [sideX]: x, [sideY]: y }
         : { position: resolvedPosition, ...originalFloatingStyles },
     [adaptiveOrigin, resolvedPosition, sideX, x, sideY, y, originalFloatingStyles],
   );


packages/react/src/utils/adaptiveOriginMiddleware.ts
```diff
@@ -15,7 +20,18 @@ export const adaptiveOrigin: Middleware = {
       placement,
     } = state;
 
-    const win = floating.ownerDocument.defaultView;
+    const win = ownerWindow(floating);
+    const styles = win.getComputedStyle(floating);
+    const hasTransition = styles.transitionDuration !== '0s' && styles.transitionDuration !== '';
```
mnajdova, line 25: Shouldn't we also check for `0ms` here too? Just to be safe

packages/react/src/utils/useAnchorPositioning.ts
```diff
@@ -413,14 +414,19 @@ export function useAnchorPositioning(
     externalTree,
   });
 
-  const { sideX, sideY } = middlewareData.adaptiveOrigin || {};
+  const { sideX, sideY } = middlewareData.adaptiveOrigin || DEFAULT_SIDES;
+
+  // Default to `fixed` when not positioned to prevent `autoFocus` scroll jumps.
+  // This ensures the popup is inside the viewport initially before it gets
+  // positioned.
```
mnajdova:
```suggestion
  // Default to `fixed` when not positioned to prevent `autoFocus` scroll jumps.
  // This ensures the popup is inside the viewport initially before it gets positioned.
```

mnajdova:
I tried to test the failing sandbox, but when I tried the package from this PR it fails with "TypeError: el.focus is not a function", can you check it: https://codesandbox.io/p/devbox/base-ui-anchored-popups-scroll-jump-forked-fwhstl?file=%2Fpackage.json%3A26%2C2&workspaceId=ws_HE9JRzFW36Y2kpBoSwdvZ5

packages/react/src/utils/adaptiveOriginMiddleware.ts
atomiks: The computed styles resolve to `s`, this check is also used for Collapsible

atomiks:
@mnajdova the Sandbox is passing an invalid ref type into the `initialFocus` prop which is why it's breaking. It's cast to `as any` but it's not related to the changes here - it's just invalid code

mnajdova:
> @mnajdova the Sandbox is passing an invalid ref type into the initialFocus prop which is why it's breaking. It's cast to as any but it's not related to the changes here - it's just invalid code

Ah, I didn't noticed, I just forked the original issue reproduction, works as expected 👌

PR #3252, opened 2025-11-18
[combobox][select] Docs for typed wrappers

atomiks:

The type setup required to satisfy all the type spec tests for [Select](https://github.com/mui/base-ui/blob/eb7e80b12c3bbd3e4db922753f57b65941235880/packages/react/src/select/root/SelectRoot.spec.tsx) and [Combobox](https://github.com/mui/base-ui/blob/eb7e80b12c3bbd3e4db922753f57b65941235880/packages/react/src/combobox/root/ComboboxRoot.spec.tsx) is now quite complex, so creating wrappers that inherit the root's props is not straightforward. These demos display copy-pasteable starting points for wrappers to use

- We could export `UncontrolledProps` & `ControlledProps` on the root to further simplify types
- There may be a way to avoid overloads while letting all the type tests pass, but I didn't find a way how to do it (would be ideal)

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/combobox/demos/typed/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/combobox/demos/typed/css-modules/index.tsx
@@ -1,0 +1,56 @@
+import * as React from 'react';
+import { Combobox } from '@base-ui-components/react/combobox';
+
+export default function App() {
+  return <MyCombobox />;
+}
+
+export function MyCombobox<Value, Multiple extends boolean | undefined = false>(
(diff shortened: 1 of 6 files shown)


matthewdavi:
I think exporting the ControlledProps and UncontrolledProps would be very useful here, it's very painful to type the generics manually

atomiks:
@matthewdavi In #3254, we've decided to remove the overloads to simplify the typed wrappers use case - there's only one small inconvenience for the controlled use case but the tradeoff is overall worth it

atomiks:
Replaced with #3254

PR #3253, opened 2025-11-18
[combobox][select] Export controlled/uncontrolled prop types

atomiks:
- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

This exports the overload types to simplify building wrappers, and adds a code block demo of how to create a wrapper with correct types (alt: #3252)

Combobox: https://deploy-preview-3253--base-ui.netlify.app/react/components/combobox#typed-wrapper
Select: https://deploy-preview-3253--base-ui.netlify.app/react/components/select#typed-wrapper

PR #3254, opened 2025-11-18
[combobox][select] Non-overload types

atomiks:

Last variant of #3252, #3253 - non-overload types sacrifice a minor useability quirk with generic types of a controlled `value` if you explicitly don't have it be nullable (more common in selects) - but it can be worth it for convenience over the 3 overloads

https://deploy-preview-3254--base-ui.netlify.app/react/components/select#typescript
https://deploy-preview-3254--base-ui.netlify.app/react/components/combobox#typescript

### DIFF
--- a/packages/react/src/combobox/root/ComboboxRoot.spec.tsx
+++ b/packages/react/src/combobox/root/ComboboxRoot.spec.tsx
@@ -262,3 +262,9 @@ mergeProps<typeof Combobox.Root<any>>(
   },
   {},
 );
+
+export function Wrapper<Value, Multiple extends boolean | undefined = false>(
+  props: Combobox.Root.Props<Value, Multiple>,
+) {
+  return <Combobox.Root {...props} />;
+}
--- a/docs/src/app/(public)/(content)/react/components/combobox/page.mdx
+++ b/docs/src/app/(public)/(content)/react/components/combobox/page.mdx
@@ -66,17 +66,13 @@ import { Combobox } from '@base-ui-components/react/combobox';
 </Combobox.Root>;
 ```
 
-## TypeScript inference
+## TypeScript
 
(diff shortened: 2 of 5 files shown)


packages/react/src/combobox/root/ComboboxRoot.spec.tsx
```diff
@@ -217,7 +219,11 @@ function App2() {
     <Combobox.Root
       value={value}
       onValueChange={(newValue) => {
+        // @ts-expect-error
         newValue.length;
+        // @ts-expect-error - user is forced to type useState with null
+        // even if they don't want to allow null
+        setValue(newValue);
       }}
```
atomiks, line 240: This is pretty much the only quirk/downside

docs/src/components/Demo/Demo.tsx
```diff
@@ -94,7 +94,7 @@ export function Demo({
                   onVariantChange={demo.expand}
                   variants={demo.variants}
                   selectedVariant={demo.selectedVariant}
-                  selectVariant={demo.selectVariant}
+                  selectVariant={demo.selectVariant as any}
```
atomiks, line 97: Somewhat annoying: setters don't always expect `| null` (which three overloads will solve) since the API might intentionally be non-nullable

packages/react/src/select/root/SelectRoot.test.tsx
```diff
@@ -2052,7 +2052,7 @@ describe('<Select.Root />', () => {
     it('falls back to null when both selected and initial default are removed (controlled)', async () => {
       function TestControlled() {
         const [items, setItems] = React.useState(['a', 'b', 'c']);
-        const [value, setValue] = React.useState('c');
+        const [value, setValue] = React.useState<string | null>('c');
```
atomiks, line 2055: Without overloads, you're now forced to type `<string | null>` in state hook calls for controlled usage

PR #3255, opened 2025-11-18
[menu][popover] Do not pass `key` to the rendered element

### LINKED ISSUE #3248: [menu][popover] `key` prop in render prop callback causes React warning
# Bug report

## Current behavior

When using the `<Menu.Trigger/>` render prop, simply forwarding all props to a custom component causes a React error:
> Warning: button: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://reactjs.org/link/special-props)

## Expected behavior

I should be able to do a render prop like `(props)=><button {...props}/>` without any errors

## Reproducible example

https://stackblitz.com/edit/vitejs-vite-nbqxy7hw?file=src%2FApp.tsx

## Base UI version

For example v1.0.0-beta.5 (also seeing the issue with https://pkg.pr.new/mui/base-ui/@base-ui-components/react@3246 )

This was not an issue in 1.0.0-beta.4

Chrome and Safari

## Which OS are you using?

MacOS

N/A

## Additional context

I can work around the issue by splitting out `key` from the rest of the props:
```
<Menu.Trigger render={({key, ...props})=><button key={key} {...props}/>}/>
```

michaldudak:
Avoids adding the `key` prop to the rendered element as it causes problems with function form of the render prop. Because `key` is in `props`, spreading the props on the rendered element triggers a React warning.

Fixes #3248

### DIFF
--- a/packages/react/test/conformanceTests/propForwarding.tsx
+++ b/packages/react/test/conformanceTests/propForwarding.tsx
@@ -45,8 +45,7 @@ export function testPropForwarding(
       await render(
         React.cloneElement(element, {
           render: (props: any) => {
-            const { key, ...propsWithoutKey } = props;
-            return <Element key={key} {...propsWithoutKey} data-testid="custom-root" />;
+            return <Element {...props} data-testid="custom-root" />;
           },
           ...otherProps,
         }),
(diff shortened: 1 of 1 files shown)


atomiks:
IIRC you changed the tests just to silence the warning earlier, let's revert that so it also confirms there's no warning for public use

PR #3266, opened 2025-11-19
[docs-infra] Skip adding markdown link in prod error page

brijeshb42:
The link doesn't make sense here.

Screenshot 2025-11-19 at 9 39 48 AM

### DIFF
--- a/docs/src/app/(public)/(content)/production-error/page.mdx
+++ b/docs/src/app/(public)/(content)/production-error/page.mdx
@@ -3,7 +3,7 @@ import ErrorDisplay from './ErrorDisplay';
 
 # Production error #<ErrorCode />
 
-<Subtitle>Explanation for minified production error message.</Subtitle>
+<Subtitle skipMarkdownLink>Explanation for minified production error message.</Subtitle>
 <Meta
   name="description"
   content="In the production build, error messages are minified to keep your application lightweight."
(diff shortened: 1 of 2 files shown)


docs/src/app/(public)/(content)/production-error/page.mdx
```diff
@@ -3,7 +3,7 @@ import ErrorDisplay from './ErrorDisplay';
 
 # Production error #<ErrorCode />
 
-<Subtitle>Explanation for minified production error message.</Subtitle>
+<Subtitle skipMarkdownLink>Explanation for minified production error message.</Subtitle>
```
mj12albert, line 6:
```suggestion
<Subtitle skipMarkdownLink>A minified Base UI error occurred in the production build of React.</Subtitle>
```

Not directly related but I think the text in the `<Subtitle>` is kinda redundant.

Would also be good to replace "A minified Base UI...." below with the text in the meta description "In the production build...."

brijeshb42: Open to suggestions for the copy. Or could be done separately.

LukasTy: Nitpick. The current variant also works.
I'm not sure if it's worth spending too much time on perfecting this page, since it should not be discovered by users. 🤔 
```suggestion
<Subtitle skipMarkdownLink>Explanation of a minified error message corresponding to a particular error code.</Subtitle>
```

Issue #3260, opened 2025-11-18
[fieldset] `Fieldset.Description` and `Fieldset.Error`


brianle1301:
# Get help

## Ask a question

I was wondering if it's possible to have description and errors on `Fieldset`? Or is this already possible with `Field.Description` and `Field.Error` somehow? My current snippet:

```
<Field.Root>
  <Fieldset.Root render={<CheckboxGroup />}>
    <Fieldset.Legend>Fruits</Fieldset.Legend>
    <Field.Description>This is some field description</Field.Description> {/* <-- Not possible */}
    <Field.Item>
      <Checkbox.Root />
      <Field.Label>Apple</Field.Label>
    </Field.Item>
    <Field.Item>
      <Checkbox.Root />
      <Field.Label>Banana</Field.Label>
    </Field.Item>
    <Field.Item>
      <Checkbox.Root />
      <Field.Label>Orange</Field.Label>
    </Field.Item>
  </Fieldset.Root>
</Field.Root>
```

mj12albert:
Adding `Description` makes sense, but how would `Error` work? e.g. in the Fieldset + CheckboxGroup example, `Field.Error` would already hold the errors for the checkbox group

brianle1301:
`Field.Error` would already add the id of the error part to the individual checkbox/radio inputs (I just confirmed via the tests)
`Field.Description` seems to work on `RadioGroup` (it adds `aria-describedby` to the div with `role="radiogroup"`) but doesn't work on `CheckboxGroup` weirdly @mj12albert

mj12albert:
> `Field.Description` seems to work on `RadioGroup` (it adds `aria-describedby` to the div with `role="radiogroup"`) but doesn't work on `CheckboxGroup` weirdly

Oh it's actually a bug that it doesn't work on checkbox group; however in radio group where it's "working", the ARIA attrs are correct but the group description is repeated by VO when moving between radios which seems wrong

Though it still makes sense to have `Fieldset.Description` to reduce ambiguity

PR #3269, opened 2025-11-19
[field][checkbox group] Fix `aria-describedby` on checkbox group

mj12albert:
From https://github.com/mui/base-ui/issues/3260#issuecomment-3547738245

Aligns the behavior with radio group; also cleaned up an unused ref.


docs/src/app/(public)/(content)/react/handbook/forms/page.mdx
```diff
@@ -44,8 +44,8 @@ import { Switch } from '@base-ui-components/react/switch';
   <Field.Label>
     <Switch.Root />
     Developer mode
-    <Field.Description>Enables extra tools for web developers</Field.Description>
   </Field.Label>
+  <Field.Description>Enables extra tools for web developers</Field.Description>
```
mj12albert, line 48: `<label>` can't contain a `<p>`: https://github.com/mui/base-ui/blob/d78ce4ca8d816c6ef6b46d6bf957f2cd06276748/docs/src/app/(private)/experiments/forms/button-controls.tsx#L101-L104

atomiks: Is this just so clicking the description text also activates the radio control? I was confused why this one was inside the label when others aren't (and don't activate the control)

mj12albert: Yeah in the experiment I was just trying out different arrangements

mj12albert:
> Like - Adding relationship only with the group, if `Field.Description` is added as a child of `Field.Root` instead of `Field.Item`?

Adding the description to the group's `aria-describedby` doesn't cause it to be read as part of the group, it's that the `<p>` itself will be read, it's probably better to leave it to the user to `aria-hidden` it if they wish; same for repeating the description for each option, it's a bit contextual

PR #3277, opened 2025-11-20
[popups] Improve performance when detached triggers are used

### LINKED ISSUE #3300: [popups] useTriggerRegistration throws error when react version < 18
# Bug report

## Current behavior

Base UI: Trigger must have an `id` prop specified.

Image

## Expected behavior

support react 17

## Reproducible example

[StackBlitz](https://stackblitz.com/edit/f2kf9dtw?file=package.json,src%2Findex.tsx,src%2FApp.tsx)

## Base UI version

1.0.0-beta.6

## Additional context

```ts
export function useTriggerRegistration(id: string | undefined) {
  const [element, setElement] = React.useState<HTMLElement | null>(null);

  useIsoLayoutEffect(() => {
    if (id == null) {
      if (element != null) {
        throw new Error('Base UI: Trigger must have an `id` prop specified.');
      }
      return;
    }
}
```

```ts
function useGlobalId(idOverride?: string, prefix: string = 'mui'): string | undefined {
  const [defaultId, setDefaultId] = React.useState(idOverride);
  const id = idOverride || defaultId;
  React.useEffect(() => {
    if (defaultId == null) {
      // Fallback to this default id when possible.
      // Use the incrementing value for client-side rendering only.
      // We can't use it server-side.
      // If you want to use random values please consider the Birthday Problem: https://en.wikipedia.org/wiki/Birthday_problem
      globalId += 1;
      setDefaultId(`${prefix}-${globalId}`);
    }
  }, [defaultId, prefix]);
  return id;
}
```

### LINKED ISSUE #3304: [tooltip] Tooltip sometimes does not close on mouse-leave
# Bug report

## Current behavior

When the specific conditions are met, the tooltip will open on hover, but will not leave when you move the mouse away.

This only seems to be an issue when:
* On Base UI 1.0.0-beta.5 or later
* Using ReactDOM.render (createRoot works as expected)
* Using Tooltip as a "controlled" component (open and onOpenChange props)
* Using the render prop on the Tooltip.Trigger, with an `id` on the element. (`render={<button id="blah"/>}`)

## Expected behavior

Moving the mouse away should close the tooltip

## Reproducible example

https://stackblitz.com/edit/vitejs-vite-vxvvxp9u?file=package.json,src%2Fmain.tsx,src%2FApp.tsx

## Base UI version

1.0.0-beta.6 (not present in 1.0.0-beta.4)

Chrome and Safari

## Which OS are you using?

macOS

## Additional context

We ran into this because we have a component that uses a Menu and a Tooltip on the same trigger, and Menu adds an ID. ~I'm investigating whether ignoring that ID is a valid workaround for the issues~ a better workaround is to add the ID to Tooltip.Trigger instead of the element passing into `render`.

michaldudak:
Fixes #3300
Fixes #3304

Refactored the FloatingRootContext to be based on the Store, allowing for selective subscription to changes in its state.

Previously, FloatingRootContext was recreated whenever the open state changed. As triggers were subscribed to it, they all rerendered whenever an associated popup opened or closed. 

This improvement applies to all "detachable" components: Menu, Tooltip, Popover, and (Alert)Dialog.

Additionally, in the Menu, due to unnecessary subscriptions, triggers also rerendered when the active item index within a menu was updated (so hovering over menu items caused re-rendering of all related menu triggers).

Before (Menu example):

https://github.com/user-attachments/assets/c25b37a2-c102-482b-9641-3f3ab7d1938b

After:

https://github.com/user-attachments/assets/97b0eb0f-6c5b-4cf2-af7f-109c7e3421d6

---

## Technical notes

The biggest change is the rewrite of FloatingRootContext. It now acts as a backing data store that `useFloating` uses. Instead of being recreated on every render or dependency change, it keeps a stable reference for the lifetime of a component. A store implementation allows hooks to selectively depend on pieces of state. Also, thanks to Store's `select` method, interaction hooks can use non-reactive values in event handlers and effects.

The FloatingRootStore is created based on the component store with the new hook: `UseSyncedFloatingRootContext`. It makes sure the pieces of state in these two stores are kept in sync (see also point 1 in Next steps below).

To simplify types, I removed the type parameter from `useFloating` (previously `RT extends ReferenceType`). We never made use of it, so I used `ReferenceType` directly where `RT` was used before.

I moved most of the logic out of triggers, reducing the initial number of renders to 2 (the minimum possible with composable API). I also unified their implementation, extracting common parts.

## Next steps

1. We can further improve the implementation by deduplicating pieces of state such as `open`, `floatingElement`, etc. from both stores (main component store and FloatingRootStore). We can consider making the Floating UI primitives to work with our PopupStore (or the other way around - use FloatingRootStore to keep component data).

2. Make triggers lazy - skip all effects until the rendered element is interacted with. This requires further investigation.


atomiks:
- The `[data-popup-open]` state (on both Menu/Tooltip) seems to be incorrectly linked to `mounted` now instead of `open` so it lingers as popups close
- The `[data-ending-style]` transition of an instant tooltip issue returned (#2962)

packages/utils/src/store/ReactStore.ts
```diff
@@ -86,11 +86,28 @@ export class ReactStore<
...
+    }
+
+    const dependencies = Object.values(statePart);
+
     useIsoLayoutEffect(() => {
-      this.update(props);
-    }, [props]);
+      this.update(statePart);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, dependencies);
```
romgrk, line 110: Tbh considering how we've implemented `Store#update()`, this should also work:

```suggestion
useIsoLayoutEffect(() => {
      this.update(statePart);
    });
```

michaldudak: For some reason, it breaks React 18 tests

atomiks:
Bugs:

- Tooltip: Pressing a trigger with the pointer before the tooltip shows no longer cancels it from showing
- Combobox: Pressing a trigger to toggle the popup (closed once it's open) no longer works when the input is outside the popup

Let's add regression tests for the issues encountered here as well

OliverJAsh:
If I'm not mistaken, this seems to have introduced a regression: https://github.com/mui/base-ui/issues/3363

PR #3302, opened 2025-11-21
[code-infra] Enable undefined addition to optional properties

### LINKED ISSUE #1916: Improve ergonomics when using `exactOptionalPropertyTypes`
# Bug report

## Current behavior

In projects where [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/exactOptionalPropertyTypes.html) is enabled, we frequently encounter errors like this:

```tsx
import { Menu } from '@base-ui-components/react';
import type { FC, ReactNode } from 'react';
import * as styles from './MenuItem.module.css';

export const MenuItem: FC<{ disabled?: boolean; children: ReactNode }> = ({
  disabled,
  children,
}) => (
  // Type 'undefined' is not assignable to type 'boolean'.
  <Menu.Item disabled={disabled} className={styles.item}>
    {children}
  </Menu.Item>
);

```

```tsx
import { Tooltip } from '@base-ui-components/react';
import type { FC, ReactNode } from 'react';

export const MyTooltip: FC<{
  anchor?: HTMLElement;
  children: ReactNode;
}> = ({ anchor, children }) => (
  <Tooltip.Portal>
    {/* Type 'undefined' is not assignable to type 'Element | VirtualElement | RefObject<Element | null> | (() => Element | VirtualElement | null) | null'. */}
    <Tooltip.Positioner anchor={anchor}>
      <Tooltip.Popup>{children}</Tooltip.Popup>
    </Tooltip.Positioner>
  </Tooltip.Portal>
);

```

It's possible to workaround these errors using spread syntax, but it's very ugly, and we forgo excess property checking.

## Expected behavior

The examples above should not error.

Optional properties should also allow explicit `undefined`, as is the case for props like `className` on built-in elements such as `div`.

## Reproducible example

https://codesandbox.io/p/sandbox/condescending-glade-ryyx5d

## Base UI version

1.0.0-alpha.8

N/A

## Which OS are you using?

N/A

N/A

## Additional context

N/A

brijeshb42:
Configured a new rule to error only in packages files (no docs or tests). It has autofix to add the `| undefined` type to the relevant optional types. Currently, it checks for all the optional types and there is not filtering (ie, only properties of exported types, etc). The main change is in the [first](https://github.com/mui/base-ui/pull/3302/changes/181ab4b8ade8a61cf7b64788d9946b34ca490d83) commit. [2nd](https://github.com/mui/base-ui/pull/3302/changes/bb9d3a23433a77433e66027f0b6209c6701586f2) commit is the autofix.

Fixes #1916 
PR for the eslint rule - https://github.com/mui/mui-public/pull/1032

### DIFF
--- a/docs/reference/generated/tabs-root.json
+++ b/docs/reference/generated/tabs-root.json
@@ -5,11 +5,13 @@
     "defaultValue": {
       "type": "Tabs.Tab.Value",
       "default": "0",
-      "description": "The default value. Use when the component is not controlled.\nWhen the value is `null`, no Tab will be active."
+      "description": "The default value. Use when the component is not controlled.\nWhen the value is `null`, no Tab will be active.",
+      "detailedType": "Tabs.Tab.Value | undefined"
     },
     "value": {
       "type": "Tabs.Tab.Value",
-      "description": "The value of the currently active `Tab`. Use when the component is controlled.\nWhen the value is `null`, no Tab will be active."
+      "description": "The value of the currently active `Tab`. Use when the component is controlled.\nWhen the value is `null`, no Tab will be active.",
+      "detailedType": "Tabs.Tab.Value | undefined"
     },
     "onValueChange": {
       "type": "((value: Tabs.Tab.Value, eventDetails: Tabs.Root.ChangeEventDetails<'none', { activationDirection: Tabs.Tab.ActivationDirection }>) => void)",
--- a/docs/reference/generated/toast-positioner.json
+++ b/docs/reference/generated/toast-positioner.json
@@ -12,7 +12,7 @@
       "type": "ToastObject<any>",
       "required": true,
       "description": "The toast object associated with the positioner.",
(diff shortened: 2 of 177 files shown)


docs/reference/generated/tabs-panel.json
```diff
@@ -3,7 +3,7 @@
   "description": "A panel displayed when the corresponding tab is active.\nRenders a `<div>` element.",
   "props": {
     "value": {
-      "type": "Tabs.Tab.Value",
+      "type": "any",
```
michaldudak: We'll have to update the API Extractor before this is merged - looks like there's a bug.

michaldudak: `typescript-api-extractor@1.0.0-alpha.11` should work better. We can continue with this PR.

brijeshb42: This is done now. The output is as expected.

brijeshb42:
Need to get the upstream PR reviewed/merged first.
As for conflicts, it'd be easy to fix in the PR since it just needs to run eslint autofix, prettier and docs:api.

brijeshb42:
@michaldudak Are these flaky tests (failing in the CI) ?

Issue #3308, opened 2025-11-23
[navigation menu] In Safari, positionerElement's width is set to 0 on hover


EmilNordling:
# Bug report

## Current behavior

The positionerElement has an issue on Safari, where it may enter a state of having no width. The bug occurs in a sequence of hovering an item that has content into one that has no content and back to the first item that was hovered, in a quite rapid succession.  It's best shown in the video below:

https://github.com/user-attachments/assets/e9753c56-2281-48ef-86b7-0ca74d1e7ba5

<img width="1036" height="726" alt="Image" src="https://github.com/user-attachments/assets/a23cf3a7-ade7-4a1d-b8ed-6321bbb69245" />

## Expected behavior

positionerElement's width should be the same as in Firefox or any Chromium browser.

## Reproducible example

Going to the https://base-ui.com/react/components/navigation-menu and follow the mouse movement in the video.

## Base UI version

Was in prior but also in 1.0.0-beta.6

## Which browser are you using?

Safari 18.1.1 (20619.2.8.11.12) 

## Which OS are you using?

macOS 15.1.1 (24B2091)

## Which assistive tech are you using (if applicable)?

-

## Additional context

-

PR #3309, opened 2025-11-23
[navigation menu] "Fix" Safari issue where positionerElement's width is set to 0 on hover

EmilNordling:
Possible fix for: https://github.com/mui/base-ui/issues/3308

Writing possible because:
1. The test might be valid, if it's actually is running Safari during testing.
2. Moving the timing might not be the correct approach of solving the bug. But it does end up in the expected behavior, see the video below:

https://github.com/user-attachments/assets/5aef2d76-1e51-489b-bd66-6461db64521e

3. There seems to be some fishy memory leak? Because if `getCssDimensions` is invoked before `sizeFrame1.request`, it'll get into the buggy state again.\
\
Here's an recording that shows how it's behaving on master, then on this PR's change and lastly adding a `getCssDimensions` call before `sizeFrame1.request` executes:

https://github.com/user-attachments/assets/367ec470-4ee2-4376-a8d9-11cb9ee344ec

This is really surprised me. Something with invoking `getCssDimensions` causes it to behave differently.


atomiks:
For some reason, I can't reproduce the original issue on Safari at all (26.1)

> Moving the timing might not be the correct approach of solving the bug

It ends up breaking the transitions of the size of the popup, so doesn't seem so.

`Popover.Viewport`/`Tooltip.Viewport` have a different technique of solving size transitions, it's possible they don't have the bug, but I can't reproduce it it to begin with

---

It's possible `handleOpenEvent` needs reworking. From what I see locally it always "self-corrects" via the layout effect where `handleValueChange(0, 0)` is called (it seems not for you?), but `handleOpenEvent` gets called with `0` width/height initially.

Falling back may help:

```jsx
  const handleValueChange = useStableCallback((currentWidth: number, currentHeight: number) => {
    if (!popupElement || !positionerElement) {
      return;
    }

    popupElement.style.removeProperty(NavigationMenuPopupCssVars.popupWidth);
    popupElement.style.removeProperty(NavigationMenuPopupCssVars.popupHeight);
    positionerElement.style.removeProperty(NavigationMenuPositionerCssVars.positionerWidth);
    positionerElement.style.removeProperty(NavigationMenuPositionerCssVars.positionerHeight);

    const { width, height } = getCssDimensions(popupElement);
    const measuredWidth = width || prevSizeRef.current.width;
    const measuredHeight = height || prevSizeRef.current.height;

    if (currentHeight === 0 || currentWidth === 0) {
      currentWidth = measuredWidth;
      currentHeight = measuredHeight;
    }

    popupElement.style.setProperty(NavigationMenuPopupCssVars.popupWidth, `${currentWidth}px`);
    popupElement.style.setProperty(NavigationMenuPopupCssVars.popupHeight, `${currentHeight}px`);
    positionerElement.style.setProperty(
      NavigationMenuPositionerCssVars.positionerWidth,
      `${measuredWidth}px`,
    );
    positionerElement.style.setProperty(
      NavigationMenuPositionerCssVars.positionerHeight,
      `${measuredHeight}px`,
    );

    sizeFrame1.request(() => {
      popupElement.style.setProperty(NavigationMenuPopupCssVars.popupWidth, `${measuredWidth}px`);
      popupElement.style.setProperty(
        NavigationMenuPopupCssVars.popupHeight,
        `${measuredHeight}px`,
      );

      sizeFrame2.request(() => {
        animationAbortControllerRef.current = new AbortController();
        runOnceAnimationsFinish(setAutoSizes, animationAbortControllerRef.current.signal);
      });
    });
  });
```

Or figuring out why the self-correction via the layout effect doesn't occur in your case

EmilNordling:
>For some reason, I can't reproduce the original issue on Safari at all (26.1)

Sorry I forgot to add which version I'm experiencing the behavior at. I'm running the version before macOS 26. Seems to only  be an issue on ~18.1.1 (20619.2.8.11.12) in that case!

---

>Falling back may help

Tried to apply the above, it'll not fallback to anything with a size unfortunately. In the scope the `width`, `currentWidth` and `prevSizeRef.current.width` will all be set to zero in this Safari version.

Though, and if it makes sense to change, it'll self-correct if it wouldn't reset the `prevSizeRef` to `DEFAULT_SIZE`.

```diff
React.useEffect(() => {
    if (!open) {
      stickIfOpenTimeout.clear();
      sizeFrame1.cancel();
      sizeFrame2.cancel();
-     prevSizeRef.current = DEFAULT_SIZE;
    }
  }, [stickIfOpenTimeout, open, sizeFrame1, sizeFrame2]);
```

https://github.com/user-attachments/assets/5c6f0c6c-aec4-423c-93ed-6b999bf7fc13

---

Seems like figuring out why `popupElement` yields a faulty width value in the version ~18 of Safari is wanted. I'm going to see if I can fish something out!

atomiks:
Try changing `prevSizeRef` to be set to the default on `mounted=false` instead of `open=false`? If it fixes it, that seems fine.

EmilNordling:
>... to be set to the default on mounted=false

That does work!

https://github.com/user-attachments/assets/62ce685d-c268-4086-83d0-246a35cfb378

atomiks:
Alright, nice! The other side effects still need to be done with `open=false` though, so an independent effect seems necessary.

EmilNordling:
Added it to a separate effect in the latest commit 🤝

PR #3315, opened 2025-11-24
[triggers] Fix React 17 support

### LINKED ISSUE #3300: [popups] useTriggerRegistration throws error when react version < 18

atomiks:

Fixes #3300

The `id` should always exist from the `useBaseUiId` util, so this throw doesn't need to happen following the switch to the effect instead of callback ref.

### DIFF
--- a/packages/react/src/utils/popupStoreUtils.ts
+++ b/packages/react/src/utils/popupStoreUtils.ts
@@ -17,9 +17,6 @@ export function useTriggerRegistration<State extends { triggers: PopupTriggerMap
 
   useIsoLayoutEffect(() => {
     if (id == null) {
-      if (element != null) {
-        throw new Error('Base UI: Trigger must have an `id` prop specified.');
-      }
       return;
     }
 


michaldudak:
https://github.com/mui/base-ui/pull/3277 changes how trigger registration works and removes the undefined ID check. This PR can be closed.

PR #3316, opened 2025-11-24
[internal] Perf: remove useless code

romgrk:
Various small changes to remove useless or unefficient code.


packages/react/src/tooltip/store/TooltipStore.ts
```diff
@@ -161,3 +161,7 @@ function createInitialState<Payload>(): State<Payload> {
     positionerElement: null,
   };
 }
+
+type Mutable<T> = {
+  -readonly [P in keyof T]: T[P];
+};
```
romgrk:
```suggestion
```

Issue #3225, opened 2025-11-14
[docs] Scroll position should follow "Hide code" button when closing code blocks


mj12albert:
When collapsing a demo code block, the scroll position doesn't move and it leaves you stranded somewhere in the document. It would be better if the scroll position followed the "Hide/Show code" button (as long as it isn't disorienting in another way)

https://github.com/user-attachments/assets/1b6169f5-34df-4c5f-b71e-7bf16cecc51a

(The longer the code block and the page, the more annoying it is)

References
- https://samding.dev/crafts/upward-accordion/
- https://nitter.net/kaishan_ding/status/1992468624884613394

PR #3321, opened 2025-11-25
[docs] Adjust scroll position when closing expanded demos

mj12albert:
After closing an expanded demo, scroll to the trigger if it's outside the viewport after close. 

~Uses a custom function because the built-in `behavior` options didn't seem good - `'instant'` can be disorientating because the trigger doesn't move but all the surrounding UI suddenly changes; `'smooth'` is too static, sometimes it feels slow even for short distances.~
Changed to use `'instant'`

Preview:
- short demo: https://deploy-preview-3321--base-ui.netlify.app/react/components/avatar
- long demo: https://deploy-preview-3321--base-ui.netlify.app/react/components/autocomplete#fuzzy-matching

Closes https://github.com/mui/base-ui/issues/3225


atomiks:
Another thing worth adding that was discussed a while ago is `position: sticky; bottom: -1px` to `.DemoCollapseButton`. That way you don't need to scroll for so long to collapse it.

> 'instant' can be disorientating because the trigger doesn't move but all the surrounding UI suddenly changes

Given the collapsible doesn't have a transition, I think instant would seem less disorienting? Currently the UI changes underneath while it's scrolling back

mj12albert:
@atomiks This is how instant looks, you think it's ok? My first impressions is that it felt a bit too fast/sudden

https://github.com/user-attachments/assets/3946f6d1-76c0-4bbe-a30a-e2d95e9e821b

atomiks:
@mj12albert imo yes, with the smooth scroll I can't actually remember which demo I was looking at once it finishes scrolling. When it's instant it's clearer which one you collapsed. (The animation would work if the scroll tracked with the button simultaneously, but there's no animation on the panel)

mj12albert:
OK made the button sticky-bottom and changed the scroll back to instant, it was actually quite weird if this is the only animated collapsible change in the docs

docs/src/components/Demo/Demo.tsx
```diff
@@ -67,6 +70,27 @@ export function Demo({
     </GhostButton>
   );
 
+  React.useEffect(() => {
+    if (
+      !demo.expanded &&
```
atomiks: Does using `onOpenChange` on `Collapsible.Root` itself work? When does `useDemo` change the expanded state otherwise? 

Otherwise, it would be good if the demo hook had an event handler to avoid needing to use an effect

mj12albert: Yeah though it needs one RAF to get `rect.top` before/after it closes

`onOpenChangeComplete` should work? But collapsible doesn't have it

atomiks: Try `ReactDOM.flushSync` to avoid a rAF?

atomiks:
- The `:active` state of the collapse button has the wrong z-index when it's stuck
- When collapsing while it's stuck, the scroll position restores such that the collapse button is still at the bottom of the screen. Maybe it should restore it to the scroll position when first opening it? Having multiple open simultaneously then closing out-of-order may complicate this

docs/src/components/Demo/Demo.tsx
mj12albert: flushSync works 👍

mj12albert:
> - The `:active` state of the collapse button has the wrong z-index when it's stuck

Fixed, it was an extra position:relative

> - When collapsing while it's stuck, the scroll position restores such that the collapse button is still at the bottom of the screen. Maybe it should restore it to the scroll position when first opening it?

Hmm I think it's ok and better that the trigger stays in place so the cursor is still on the thing that was clicked

atomiks:
> Fixed, it was an extra position:relative

Still seeing it:

https://github.com/user-attachments/assets/d1cfebd1-0b65-4f7b-9d9a-17d29ee2ba1a

mj12albert:
Fixed the `:active` color:
```css
&:active {
  background-color: color-mix(in oklch, var(--color-gray-75), var(--color-gray-600) 5%);
}
```
Tricky since all the grays have some transparency except one

mj12albert:
Added a container query to remove the border radius on the trigger when it's stuck (only Firefox unsupported)

https://github.com/user-attachments/assets/9b0bcfaf-6fba-4da5-a613-cad1b271fe91

atomiks:
Styles seem broken now:
Screenshot 2025-11-26 at 10 29 22 pm

mj12albert:
Worked in dev mode 🥲 
Screenshot 2025-11-26 at 9 14 22 PM

docs/src/components/Demo/Demo.css
```diff
@@ -208,7 +208,14 @@
     }
   }
 
-  .DemoCollapseButton {
+  .DemoCollapseButtonStickyContainer {
+    container-type: scroll-state;
```
LukasTy: Regardless of how nice it looks, IMHO, we shouldn't use experimental features. 🤔 
https://caniuse.com/mdn-css_properties_container-type_scroll-state
It doesn't work on the latest Safari, Safari iOS, and Firefox.
Depending on the continent, this could mean half the user base. 🙈

mj12albert: I think this data is wrong? My [recording](https://github.com/mui/base-ui/pull/3321#issuecomment-3580441328) was done in Safari

It's a nice to have detail for supported browsers, so I think the current support level is fine; but it's not worth trying to fight the CSS parser for it so forget about it 😩

LukasTy: Oh, that's interesting... 🤷 
Maybe the support was added in the latest version, and they have not yet updated it...

LukasTy (approved):
LGTM. 👍 
Super useful improvement. 💯 

It might be useful to avoid sticking at a certain point (i.e., the minimal height of open code examples), but I'm not sure it's worth the extra complication. 👍

PR #3334, opened 2025-11-26
[number field] Fix literal space handling with symbols

### LINKED ISSUE #3333: [number field] Cant change number when formatting as currency and german locale
# Bug report

## Current behavior

After the initial input of a number in a number input field with formatting as currency and the locale set to german. The value is displayed correctly but one can not change it anymore after it is originally shown. I guess this is happening because the symbol is at the end!

Image

## Expected behavior

Should sill be able to change. 

## Reproducible example

[Link to a CodeSandbox](https://codesandbox.io/p/sandbox/kt4hyq)

## Base UI version

1.0.0-beta.6

Chrome

## Which OS are you using?

Mac OS

atomiks:

Fixes #3333

https://codesandbox.io/p/sandbox/bug-report-numberinput-baseui-forked-7x5y7r

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/dialog/demos/outside-scroll/css-modules/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/dialog/demos/outside-scroll/css-modules/index.module.css
@@ -252,8 +252,3 @@
   line-height: 1.5rem;
   color: var(--color-gray-600);
 }
-
-.Actions {
-  display: flex;
-  justify-content: flex-end;
-}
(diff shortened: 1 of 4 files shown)


mj12albert:
Is it a separate issue that a text selection in the input can't be replaced with a `,`? It works with a number or `.`

https://github.com/user-attachments/assets/6e8727d4-31df-4fc7-a265-d2ca04addb94

atomiks:
@mj12albert yeah, that's a separate issue with how some symbols that should only have one unique instance present in the input are handled, depending on the text selection. Removing all text works, or only that char alone, but it could/should be more permissive

mj12albert (approved):

PR #3342, opened 2025-11-27
[experiments] Align Performance experiments with using a shared benchmark harness

LukasTy:
Merge the per-variant pages from `experiments/perf/` and the original component pages from this PR into a single **Performance** section, driven by a shared variants-aware benchmark harness. Each page picks a variant, or all of them, from a single dropdown, and presents `Last`, `Samples`, `Avg`, `Std dev`, `Min`, and `Max` in one on-page results table.

### Pages

- [`/experiments/performance/tooltip-perf`](https://deploy-preview-3342--base-ui.netlify.app/experiments/performance/tooltip-perf) -- 2000 instances. Variants: Plain buttons, Base UI, Base UI with detached triggers, Radix.
- [`/experiments/performance/triggers-perf`](https://deploy-preview-3342--base-ui.netlify.app/experiments/performance/triggers-perf) -- 500 rows x Dialog/Menu/Popover/Preview Card/Tooltip. Variants: Base UI contained triggers, Base UI detached triggers, Radix (Hover Card stands in for Preview Card). Uses the sidebar `SettingsPanel` to toggle which components render across all variants; changing a toggle resets the results.
- [`/experiments/performance/combobox-perf`](https://deploy-preview-3342--base-ui.netlify.app/experiments/performance/combobox-perf) -- 1000 items, open at mount so the list is part of the measurement. Variants: Base UI with index, Base UI without index, Base UI open API.
- [`/experiments/performance/select-perf`](https://deploy-preview-3342--base-ui.netlify.app/experiments/performance/select-perf) -- 1000 items, open at mount so the list is part of the measurement. Variants: Base UI Select, Radix Select.
- [`/experiments/performance/menu-open-perf`](https://deploy-preview-3342--base-ui.netlify.app/experiments/performance/menu-open-perf) -- times opening an already mounted Menu with 50 items, through a `Menu.createHandle()` handle. Ported from master's `perf/menu-open.tsx` onto the shared harness.

Combobox and Select use `defaultOpen` on the Root because closed popups are lazily mounted -- without it the harness only measures the trigger, not the 1000-item list. The Base UI Select is `modal={false}` and the Combobox lists use `collisionAvoidance={{ side: 'none' }}`, so the open popups neither trap focus nor cover the harness controls.

### Shared harness

The harness in `performance/utils/benchmark.tsx` keeps the existing MutationObserver + 32 ms quiet-window measurement and IQR-based outlier removal, but now:

- Accepts a `variants: { key, label, render, interact? }[]` prop and tracks per-variant samples.
- Provides `Re-render`, `Run 10`, `Run 20`, `Run 50`, `Remove outliers`, and `Reset` controls. The "All variants" option runs each variant in turn and fills the whole table. The picker is hidden when a page has a single variant.
- Renders nothing until `Re-render` or a `Run` button is pressed. Selecting a variant only clears the one on screen.
- Replaces the controls with a "Running <variant> <n> times" status while measuring. The controls keep their space, so the table does not move.
- For variants with `interact`, mounts the variant untimed, waits for the DOM to settle, then times the interaction until the DOM settles again.
- Observes the whole document, since variants portal their popups to `body`, ignoring its own controls and the Next.js dev overlay. Runs cancelled by unmounting, a workload change, or a 10 s timeout record nothing.
- Surfaces stats in an on-page table (one row per variant) and still logs raw arrays to the console for tooling.

Covered by jsdom tests in `docs/src/utils/benchmark.test.tsx`.

### Removed (folded into the above)

- `experiments/perf/{contained,detached,radix}-triggers.tsx` -> covered by `triggers-perf` variants.
- `experiments/perf/menu-open.tsx` -> moved to `performance/menu-open-perf.tsx`.
- `experiments/perf/utils/benchmark.tsx` -> moved under `performance/utils/`. `wait.ts` is deleted; nothing uses it any more.
- `experiments/menu/perf-{contained,detached}.tsx` -> covered by `triggers-perf` with the Menu-only toggle.
- The per-page CSS modules -> merged into a single `performance.module.css`.

Not carried over: the "Measure unmount" toggle of the old trigger pages.


michaldudak:
I'm wondering if we could merge this with experiments within the "perf" directory. Yours has a better UI and presents results in a better way, while the others test more components and re-render components multiple times.

LukasTy:
> I'm wondering if we could merge this with experiments within the "perf" directory. Yours has a better UI and presents results in a better way, while the others test more components and re-render components multiple times.

@michaldudak Sorry, I missed your comment.
By merging do you mean taking the best parts from each implementation?
Or just moving this experiment under `perf` subpath?

What about the other experiments?
Screenshot 2025-12-11 at 13 27 53

Maybe they all should be under the `performance` subpath? 🤔

michaldudak:
I meant taking the best parts from each. And yes, they should be under Performance.

flaviendelangle:
What do we want to do with this one?
I can merge the conflicts but now that we have the new perf benchmark in the infra, is this experiment still relevant @michaldudak?

michaldudak:
IDK, it may still be worth adding some ideas from this PR into the experiments (like running all the variants together and presenting results in a single table), but it's not a must-have for sure.

As for the perf tests in infra, they're good for monitoring long-term trends, but definitely not for determining perf issues in a single PR.

LukasTy:
@michaldudak @flaviendelangle, I've updated the PR.
Let me know what do you think about the consolidated performance pages.

LukasTy:
BUMP.
@michaldudak, @flaviendelangle, the PR has been updated.
I cross-reviewed it with Claude and Codex and addressed the findings both found.

michaldudak:
Hey @LukasTy, thanks for pulling this together. After reviewing it, I pushed a few commits on top of your branch:

- New "All variants" option runs every variant in one go.
- `menu-open-perf` uses the shared harness through a new `interact` callback. It now measures until the DOM settles, so the numbers aren't comparable with the old ones.
- Preview Card is back in `triggers-perf`.
- Dropped the beta.4 tooltip variant and its dependency.

michaldudak (approved):

LukasTy:
Thanks for the nice improvements, @michaldudak. 🙏 
The `All` variant is a very neat idea. 💯

PR #3352, opened 2025-11-27
[internal] Add render performance benchmark to experiments

michaldudak:
Create render performance benchmark for popup components.

https://deploy-preview-3352--base-ui.netlify.app/experiments/perf/contained-triggers
https://deploy-preview-3352--base-ui.netlify.app/experiments/perf/detached-triggers
https://deploy-preview-3352--base-ui.netlify.app/experiments/perf/radix-triggers


michaldudak:
Approved implicitly with #3385

PR #3360, opened 2025-11-28
[internal] Performance: use raw `useSyncExternalStore`

romgrk:
See https://github.com/mui/mui-x/pull/20447

### DIFF
--- a/packages/utils/src/store/useStore.ts
+++ b/packages/utils/src/store/useStore.ts
@@ -3,62 +3,72 @@ import * as React from 'react';
  * More info: https://github.com/mui/mui-x/issues/18303#issuecomment-2958392341 */
 import { useSyncExternalStore } from 'use-sync-external-store/shim';
 import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/shim/with-selector';
-import type { Store } from './Store';
+import { isReactVersionAtLeast } from '../reactVersion';
+import type { ReadonlyStore } from './Store';
 
-const majorVersion = parseInt(React.version, 10);
-const canUseRawUseSyncExternalStore = majorVersion >= 19;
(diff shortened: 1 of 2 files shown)


packages/utils/src/store/useStore.ts
```diff
@@ -2,8 +2,12 @@ import * as React from 'react';
 /* We need to import the shim because React 17 does not support the `useSyncExternalStore` API.
  * More info: https://github.com/mui/mui-x/issues/18303#issuecomment-2958392341 */
 import { useSyncExternalStore } from 'use-sync-external-store/shim';
+import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/shim/with-selector';
 import type { Store } from './Store';
 
+const majorVersion = parseInt(React.version, 10);
```
michaldudak: We have a `reactVersion` module with `isReactVersionAtLeast` you can use here.

packages/utils/src/store/useStore.ts
```diff
@@ -43,10 +47,18 @@ export function useStore(
...
-
-  return useSyncExternalStore(store.subscribe, getSelection, getSelection);
+  if (canUseRawUseSyncExternalStore) {
+    const getSelection = React.useMemo(
+      () => () => selector(store.getSnapshot(), a1, a2, a3),
+      [store, selector, a1, a2, a3],
+    );
+    return useSyncExternalStore(store.subscribe, getSelection, getSelection);
+  } else {
+    return useSyncExternalStoreWithSelector(
```
michaldudak: Have you found out what's the actual issue with React 18?

romgrk: No. I compared the implementation of `useSyncExternalStore` in R18 and R19 but couldn't find relevant changes, and I don't want to invest too much time on this. I might give it another try in a later iteration, I have more work to do on the store anyway.

PR #3361, opened 2025-11-28
[number field] Fix parsing numbers with Swiss locale

michaldudak:
I had a failing NumberField test locally. Turns out some environments use `'` and others `’` for Swiss group separators.


packages/react/src/number-field/utils/parse.ts
```diff
@@ -147,9 +147,19 @@ export function parseNumber(
 
   let groupRegex: RegExp | null = null;
   if (group) {
+    const isSpaceGroup = /\p{Zs}/u.test(group);
+    const isApostropheGroup = group === "'" || group === '’';
```
atomiks, line 151: Add a regression test where this fails otherwise?

michaldudak: It depends on the browser / env settings, I think. It failed consistently on my laptop.

PR #3362, opened 2025-11-28
[code-infra] Run `material-ui/disallow-react-api-in-server-components`

Janpot:
Trying out https://github.com/mui/mui-public/pull/919

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -88,7 +88,7 @@ export default defineConfig(
   },
   {
     files: [`packages/*/src/**/*${EXTENSION_TS}`],
-    ignores: [`**/*${EXTENSION_TEST_FILE}`, `**/*.spec${EXTENSION_TS}`,`test/**/*${EXTENSION_TS}`],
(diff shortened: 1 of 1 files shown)


Janpot:
Test: @greptile

Janpot:
@greptile I think you're a bit confused.

1. We only consider our package.json exports as code that potentially can be imported by a RSC. But they may also be imported by client components. Perfectly valid to use useMemo here. It just will be a no-op on the server.
2. Internally it'd be perfectly valid to omit `'use client'` for modules using these client only APIs if, and only if, the module is imported only by (descendants of) modules that are a client boundary

Janpot:
@greptile

> The rule needs more nuance to understand the module graph and import contexts rather than flagging all client API usage uniformly.

Correct, but this is good enough. It's impossible to make this 100% sound without checking this at runtime. e.g. the react module may be passed as a variable or something which can't deterministically be tracked.

Janpot:
@greptile It's perfectly fine to use components that are defined with `React.forwardRef` in RSC, you just can't create a ref to pass to it. But there is no problem if the ref is optional.

Janpot:
> I'm just curious why the list rule is prefixed with "material-ui/".

Historically this plugin was ported from the core repo and nobody ever bothered to change the name as it's work with no clear benefit.

PR #3364, opened 2025-11-28
[popover] Fix popover glitching when flipped

### LINKED ISSUE #3363: [popover] Placement regression
After upgrading to 1.0.0-beta.7, we noticed a regression in our popovers.

I've narrowed this down to 7a3eb2a402d529a1d4c79ad3346b536a090513fd.

This popover popup has `max-height: calc(var(--available-height) - 24px);`. Previously it would detect a conflict and switch sides, whereas now it seems to just reduce `--available-height`. Towards the end of the video there seems to be another issue where the page jumps and the popover is detached from its trigger.

I'll try and create a reduced test case.

### After

https://github.com/user-attachments/assets/da07f35e-85c0-49c5-984f-ff412f38dac5

### Before

https://github.com/user-attachments/assets/9890b8e8-db95-4c66-bca1-c3f4204b5ab7

michaldudak:
Fixes #3363

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/tooltip/demos/detached-triggers-controlled/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/tooltip/demos/detached-triggers-controlled/css-modules/index.tsx
@@ -50,7 +50,7 @@ export default function TooltipDetachedTriggersControlledDemo() {
         triggerId={triggerId}
       >
         <Tooltip.Portal>
-          <Tooltip.Positioner sideOffset={10}>
+          <Tooltip.Positioner sideOffset={10} className={styles.Positioner}>
             <Tooltip.Popup className={styles.Popup}>
               <Tooltip.Arrow className={styles.Arrow}>
                 <ArrowSvg />
--- a/docs/src/app/(public)/(content)/react/components/tooltip/demos/detached-triggers-controlled/tailwind/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/tooltip/demos/detached-triggers-controlled/tailwind/index.tsx
@@ -98,7 +98,14 @@ export default function TooltipDetachedTriggersControlledDemo() {
         triggerId={triggerId}
       >
         <Tooltip.Portal>
-          <Tooltip.Positioner sideOffset={10}>
+          <Tooltip.Positioner
+            className="
+              h-(--positioner-height)
+              w-(--positioner-width)
+              max-w-(--available-width)
+            "
+            sideOffset={10}
+          >
             <Tooltip.Popup
               className="
                 px-2 py-1
--- a/docs/src/app/(public)/(content)/react/components/tooltip/demos/index.module.css
+++ b/docs/src/app/(public)/(content)/react/components/tooltip/demos/index.module.css
@@ -65,6 +65,12 @@
   height: 1.25rem;
 }
 
+.Positioner {
+  height: var(--positioner-height);
+  width: var(--positioner-width);
+  max-width: var(--available-width);
+}
+
 .Popup {
   box-sizing: border-box;
   font-size: 0.875rem;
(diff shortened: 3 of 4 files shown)


packages/react/src/popover/popup/PopoverPopup.tsx
```diff
@@ -126,7 +125,10 @@ export const PopoverPopup = React.forwardRef(function PopoverPopup(
 
   // If there's just one trigger, we can skip the auto-resize logic as
   // the popover will always be anchored to the same position.
-  const autoresizeEnabled = () => store.context.triggerElements.size > 1;
+  const autoresizeEnabled = React.useCallback(
+    () => store.context.triggerElements.size > 1,
+    [store],
+  );
```
atomiks, line 131: Maybe it should check if animations are enabled on the positioner? Or if it has fixed size variables? e.g. if detached triggers are being used without transitions/fixed sizes, does this still work?

packages/react/src/utils/usePopupAutoResize.ts
```diff
@@ -44,6 +69,10 @@ export function usePopupAutoResize(parameters: UsePopupAutoResizeParameters) {
       return undefined;
     }
 
+    Object.entries(anchoringStyles).forEach(([key, value]) => {
+      popupElement.style.setProperty(key, value as string);
+    });
```
atomiks, line 74: Actually, thinking about how this works, I think this is fine given the other mutations happening on `popupElement`. I don't think these `style` props needs to be rendered by React.

packages/react/src/popover/popup/PopoverPopup.tsx
michaldudak: Yeah, detecting transitions/animations makes sense. I don't think we can detect whether a `--popup-width` variable is used in the Popup styles.

packages/react/src/utils/usePopupAutoResize.ts
```diff
@@ -44,6 +69,16 @@ export function usePopupAutoResize(parameters: UsePopupAutoResizeParameters) {
       return undefined;
     }
 
+    const popupStyles = getComputedStyle(popupElement);
+    if (popupStyles.transitionDuration === '0s' && popupStyles.animationDuration === '0s') {
+      // If there are no transitions or animations, we can skip the auto-resize logic.
+      return undefined;
+    }
```
atomiks: This should check `positionerElement`, not `popupElement` (since that is typically animated to begin with). In `adaptiveOrigin`, it checks only `transitionDuration` (`'0ms'` and `''` values which matches Collapsible) as I animations on the positioner are not useful

michaldudak: It's the Popup that gets resized (see https://deploy-preview-3364--base-ui.netlify.app/react/components/popover#animating-the-popover). Positioner can have its position transitioned. As for animations - they will certainly be less common, but I made the first iteration of our demos with them and they worked fine, so perhaps some devs would prefer them.

atomiks: Yeah, though I figured that when you have auto-resize enabled, you also have transitions on the positioner specified (to transition between triggers smoothly). Unless you mean auto resizing based purely on new content being inserted without changing triggers — though when I mutate content in the popover, it doesn't transition already?

Also `adaptiveOrigin` will need tweaking to support that use case properly. So imo, we can just check the `positioner` (matching `adaptiveOrigin`) and look into content insertion transitions later, as this fix is more critical to do properly.

michaldudak: When you change the payload dynamically, the popup will resize without changing position. There's actually a bug that prevents this content transition when there's just one trigger

atomiks: Alright, let's go with this in the meantime, then.

michaldudak: Actually, this is not reliable :/ When the popup first appears, it has `data-instant`, which disables transitions, so the first time the size changes, it'll be instant. I have to revert this.

michaldudak: I don't want to block the release, so I'm reverting the attempt to fix the content transitions when triggers don't change. I'll open a separate PR for it.

PR #3370, opened 2025-11-30
[useButton] Prevent `Space` default on valid anchor keydown

kyukyu-dev:

This PR fixes an issue where pressing the `Space` key on a valid anchor (`<a href="…">`) did not call `preventDefault()`, causing the page to scroll.
The logic is updated to ensure the `Space` key behavior on valid links matches expected button-like interaction and prevents unintended scrolling.

### DIFF
--- a/packages/react/src/use-button/useButton.ts
+++ b/packages/react/src/use-button/useButton.ts
@@ -119,21 +119,22 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
             }
 
             const shouldClick =
-              event.target === event.currentTarget &&
-              !isNativeButton &&
-              !isValidLink() &&
-              !disabled;
+              event.target === event.currentTarget && !isNativeButton && !disabled;
             const isEnterKey = event.key === 'Enter';
             const isSpaceKey = event.key === ' ';
 
             // Keyboard accessibility for non interactive elements
             if (shouldClick) {
-              if (isSpaceKey || isEnterKey) {
-                event.preventDefault();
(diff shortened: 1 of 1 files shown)


mj12albert:
@kyukyu-dev Are you using this like `<Button render={<a />} />`?
We don't really intend on supporting this, are you aiming to make a link look like a button?

kyukyu-dev:
@mj12albert 
Yes, exactly. Since I think button styled links are used quite frequently, I wanted to suggest that base-ui consider supporting them.

mj12albert:
@kyukyu-dev It's intentional that we don't preventDefault on the space here, native `<button>`s and `<a>`s respond to the space key differently, so mixing up their interactions could harm the experience for keyboard or assistive tech users

We generally don't recommend `<Button render={<a />} />` and will show a warning in dev mode in the future; for sharing styles, we recommend making separate Link and Button components that use the same styles

Alternatively it's also possible (but not a recommendation for the above reason 😬) to preventDefault on space in your own component:
```tsx
function LinkButton() {
  const buttonRef = React.useRef<HTMLAnchorElement>(null);
  return (
    <Button
      ref={buttonRef}
      onKeyDown={(event) => {
        if (event.key === ' ') {
          event.preventDefault();
          if (buttonRef.current?.href) {
            buttonRef.current?.click();
          }
        }
      }}
      nativeButton={false}
      render={<a href="https://github.com/mui/base-ui/issues">Issues</a>}
    />
  );
}
```

PR #3372, opened 2025-11-30
[tabs] Fix Panel `keepMounted` behavior

atomiks:

Fixes https://github.com/mui/base-ui/issues/1872

Ensures tab panels are unmounted like other components when `keepMounted={false}` - requires fixes to registration of panels and composite list

**Breaking change:** make `value` required on `Tabs.Tab` and `Tabs.Panel`

### DIFF
--- a/packages/react/src/tabs/panel/TabsPanel.tsx
+++ b/packages/react/src/tabs/panel/TabsPanel.tsx
@@ -81,7 +81,7 @@ export const TabsPanel = React.forwardRef(function TabPanel(
       return undefined;
     }
 
-    if (value == null || id == null) {
+    if (id == null) {
       return undefined;
     }
 
@@ -101,7 +101,7 @@ export const TabsPanel = React.forwardRef(function TabPanel(
 
 export interface TabsPanelMetadata {
   id?: string;
-  value: TabsTab.Value | undefined;
+  value: TabsTab.Value;
 }
 
 export interface TabsPanelState extends TabsRoot.State {


packages/react/src/composite/list/CompositeList.tsx
```diff
@@ -50,7 +50,11 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
     disableEslintWarning(mapTick);
 
     const newMap = new Map<Element, CompositeMetadata<Metadata>>();
-    const sortedNodes = Array.from(map.keys()).sort(sortByDocumentPosition);
+    // Filter out disconnected elements before sorting to avoid inconsistent
```
mnajdova, line 53:
```suggestion
    // Filter out disconnected elements (elements not added in the document yet) before sorting to avoid inconsistent
```

To be more clear what disconnected means.

mnajdova (approved):
Looks good, I test this together with https://github.com/mui/base-ui/pull/3373, works nicely.

packages/react/src/composite/list/CompositeList.tsx
atomiks: I think `elements are detached from the DOM` should be clear enough (on the second line?) Disconnected elements may have been in the document but aren't anymore, so that comment is not quite accurate

packages/react/src/tabs/panel/TabsPanel.tsx
```diff
@@ -84,17 +72,36 @@ export const TabsPanel = React.forwardRef(function TabPanel(
...
     ],
     stateAttributesMapping: tabsStateAttributesMapping,
   });
 
+  useIsoLayoutEffect(() => {
+    if (hidden && !keepMounted) {
+      return undefined;
+    }
+
+    if (value == null || id == null) {
```
mnajdova: When would value be null/undefined?

mnajdova:
Please update the PR title

atomiks:
> Are we ever throwing errors in similar cases (expected item not found -> would result in a "broken" UI)?

For required props, we don't do runtime validation unless it may be confusing even type-wise. e.g. `Toast.Root` requires the `toast` prop and breaks otherwise - I consider this to be similar

> Please update the PR title

The fix is still the same even with the breaking change though - in the release, the breaking change will be described along with the fix.

PR #3375, opened 2025-11-30
[number field] Fix fractional step snapping

### LINKED ISSUE #3369: [number field] `snapOnStep` with decimal `step` causes increment/decrement to get stuck
# Bug report

## Current behavior

When the `step` prop on `NumberField.Root` is set to a decimal value and the `snapOnStep` prop is set to `true`, the increment/decrement actions get stuck at some value.

## Expected behavior

The increment/decrement actions should not get stuck.

## Reproducible example

https://stackblitz.com/edit/vitejs-vite-qqug74ny?file=src%2FApp.tsx

## Base UI version

1.0.0-beta.7

Chromium

## Which OS are you using?

Mac OS

None

## Additional context

In the StackBlitz there are 3 inputs with different decimal `step` props set on them. You may use arrow up/down and see that they all get stuck at some point.

atomiks:

Fixes #3369. Reworks logic to avoid floating point precision bugs with fractional step snapping

https://stackblitz.com/edit/vitejs-vite-fydiznkt

### DIFF
--- a/packages/react/src/number-field/utils/validate.test.ts
+++ b/packages/react/src/number-field/utils/validate.test.ts
@@ -216,5 +216,27 @@ describe('NumberField validate', () => {
         }),
       ).to.equal(3.4);
     });
+
+    it('rounds to the nearest value when using small step', () => {
+      expect(
+        toValidatedNumber(0.15, {
+          ...defaultOptions,
+          step: 0.1,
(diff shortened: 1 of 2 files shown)


atomiks:
@LukasTy that's essentially what's being done here, but generalized and handles arbitrary steps/formatting precision

PR #3377, opened 2025-11-30
[combobox][select][menu] Sync highlight props

### LINKED ISSUE #3279: [Select] Different hover and focus state for Select.Item
# Get help
I am trying to get different focus and hover states for `Select.Item`. Very similar to https://github.com/mui/base-ui/issues/2731

I am using `data-highlighted` and a combination of `:hover` state. Now when I select an option and the select menu closes, and when I open it again `data-highlighted` is applied on the selected option (or the last selected option for a multi select).

https://github.com/user-attachments/assets/b0b5cf3e-11e9-4b2c-9e08-b7eebbe3bb85

I don't want the marker style on the selected item when I open the select again.

Here is a reproduction link - https://stackblitz.com/edit/19hu8jg4-tkjc2mzr?file=src%2FApp.tsx

Is there a better way to do this?

atomiks:

For Menu and Select, because they use DOM focus, pressing an item initiates the `[data-highlighted]` state: https://stackblitz.com/edit/zb6utusy. This is unlike Combobox/Autocomplete given their focus is virtual.

Closes #3279

## Features

Add `highlightItemOnHover` prop to `Menu` and `Select`

## Breaking changes

Remove `keepHighlight` from Combobox (kept for Autocomplete), given the common use cases for it are often command palette related rather than Select related. If requested, it could be returned, but it may need to be added for Select, too.

### DIFF
--- a/docs/reference/generated/select-root.json
+++ b/docs/reference/generated/select-root.json
@@ -41,7 +41,7 @@
     "highlightItemOnHover": {
       "type": "boolean",
       "default": "true",
-      "description": "Whether moving the pointer over items should highlight them.\nDisabling this prop allows CSS `:hover`  to be differentiated from the `:focus` (`data-highlighted`) state.",
+      "description": "Whether moving the pointer over items should highlight them.\nDisabling this prop allows CSS `:hover` to be differentiated from the `:focus` (`data-highlighted`) state.",
       "detailedType": "boolean | undefined"
     },
     "actionsRef": {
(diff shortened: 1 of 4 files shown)


atomiks:
@LukasTy I think we should keep the examples to showcase headless-like functionality mostly exclusively, as this is a prop for styling customization. Plus, the default is recommended over the alternative to match OS behavior unless your design system has a lighter colored hover state

michaldudak:
In Autocomplete with `keepHighlight`, when I move the mouse outside and use the arrow keys, the highlight jumps to the top.

https://github.com/user-attachments/assets/d9aa38de-1deb-4554-b165-7817760fee15

EDIT: It also occurs on master, so we can fix it in a separate PR.

LukasTy (approved):
LGTM. 👍 
Thanks for taking care of this and aligning the behavior. 👍

PR #3379, opened 2025-12-01
[dialog] Fix closing after pointer lock exit in Firefox

### LINKED ISSUE #3306: [NumberInput + Dialog] scrolling `ScrubArea` with mouse closes the Dialog on mouse release
# Bug report

## Current behavior

If the `NumberInput` component is in a `Dialog`, changing the value by scrolling with `ScrubArea` will also closes the dialog when i stop holding the mouse, rendering the component momentarily unusable

## Expected behavior

releasing mouse (and potentially thumb on mobile devices) should not propagate events to dialogs

## Reproducible example

https://stackblitz.com/edit/awgfv3sl?file=src%2FApp.tsx

## Base UI version

v1.0.0-beta.6

Firefox

## Which OS are you using?

Linux

None

## Additional context

None

atomiks:

After pointer lock exit, Firefox fires a click event on `<body>` (unlike Chrome), so it's an "outside press" (incorrectly)

`NumberField.ScrubArea` calls `.preventDefault()` in `onPointerDown` (to avoid stealing focus from the input) but this blocks the `handlePressedInside` call which prevents the outside click of popovers and dialogs from firing if the press started inside but ended outside.

Fixes #3306 - https://stackblitz.com/edit/awgfv3sl-g3rdti11

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -43,6 +43,7 @@
     "postcss": "^8.5.6",
     "postcss-custom-media": "^11.0.6",
     "postcss-import": "^16.1.1",
+    "radix-ui": "^1.4.3",
     "react": "^19.2.0",
     "react-dom": "^19.2.0",
     "react-error-boundary": "^6.0.0",
--- a/docs/src/app/(private)/experiments/[...slug]/page.tsx
+++ b/docs/src/app/(private)/experiments/[...slug]/page.tsx
@@ -52,11 +52,13 @@ export async function generateStaticParams() {
     cwd: experimentsRootDirectory,
   });
 
-  return files.map((file) => {
-    return {
-      slug: file.replace(/\.tsx$/, '').split('/'),
-    };
-  });
+  return files
+    .filter((file) => file.split('/').length <= 2)
+    .map((file) => {
+      return {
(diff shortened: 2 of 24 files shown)


packages/react/src/floating-ui-react/hooks/useDismiss.ts
```diff
@@ -691,8 +691,10 @@ export function useDismiss(
       onKeyDown: closeOnEscapeKeyDown,
       onMouseDown: handlePressedInside,
       onMouseUp: handlePressedInside,
-      onPointerDownCapture: markInsideReactTree,
-      onMouseDownCapture: markInsideReactTree,
+      onPointerDownCapture(event) {
```
mnajdova: Please add a comment explaining why this is added, it's something that can easily be broken in the future, especially as it is one browser edge case.

atomiks: The change itself is not necessarily Firefox-only, but it happens to fix a bug on Firefox. I adjusted the code and added a more general comment as to why `onPointerDown: handlePressedInside` should be added.

PR #3384, opened 2025-12-01
[internal] Refactor the store

romgrk:
Simplify controlled values and small changes to match the Store in X.

### DIFF
--- a/packages/utils/src/store/ReactStore.ts
+++ b/packages/utils/src/store/ReactStore.ts
@@ -118,8 +118,6 @@ export class ReactStore<
     controlled: Value | undefined,
   ): void {
     React.useDebugValue(key);
-    // eslint-disable-next-line consistent-this
-    const store = this;
     const isControlled = controlled !== undefined;
 
     useIsoLayoutEffect(() => {


packages/utils/src/store/Store.ts
```diff
@@ -13,7 +15,7 @@ export class Store<State> {
    *
    * Do not modify properties in state directly. Instead, use the provided methods to ensure proper state management and listener notification.
    */
-  public state: State;
+  state: State;
```
romgrk, line 18: I want this file to be a copy-paste of the one in X, I've sync'ed the two and removed the `public` modifiers (which are the default) to match the convention with classes in the rest of the codebase.

packages/utils/src/store/ReactStore.ts
```diff
@@ -232,12 +161,15 @@ export class ReactStore<
    *
    * @param key Key of the selector to use.
    */
-  public useState = ((key: keyof Selectors, a1?: unknown, a2?: unknown, a3?: unknown) => {
```
romgrk, line 241: I've avoided the `method = () => { ... }` syntax when possible, because that's +1 memory allocation for each instance, and with stores being widespread it make sense to keep them lean.

packages/react/src/menu/root/MenuRoot.tsx
romgrk, line 1: This change complexifies a bit the open-by-default feature of menu/tooltips, but it keeps the Store simple, and I'd rather have the core of the codebase be simple.

We could make the controlled props more ergonomic, I also thought about joining the two fields in some standardized way, e.g. call them `open` and `open:controlled-prop`, then we could derive selectors & effects automatically, but I've kept this simple for now.

packages/utils/src/store/ReactStore.ts
```diff
@@ -117,113 +109,50 @@ export class ReactStore<
    * - If `controlled` is undefined, the key is marked as uncontrolled. The store's state
    *   is initialized to `defaultValue` on first render and can be updated with local writes.
```
michaldudak: The JSDoc is no longer accurate

packages/utils/src/store/Store.ts
michaldudak: Why doesn't X import this one?

packages/react/src/popover/root/PopoverRoot.tsx
```diff
@@ -28,7 +29,7 @@ function PopoverRootComponent<Payload>({ props }: { props: PopoverRoot.Props<Pay
   const {
     children,
     open: openProp,
-    defaultOpen: defaultOpenProp = false,
+    defaultOpen: defaultOpen = false,
```
michaldudak:
```suggestion
    defaultOpen = false,
```

packages/utils/src/store/Store.ts
```diff
@@ -90,19 +92,32 @@ export class Store<State> {
...
    * Gives the state a new reference and updates all registered listeners.
    */
-  public notifyAll() {
+  notifyAll() {
     const newState = { ...this.state };
-    Store.prototype.setState.call(this, newState);
+    this.setState(newState);
+  }
+
+  use<F extends (...args: any) => any>(selector: F, ...args: SelectorArgs<F>): ReturnType<F>;
```
michaldudak, line 109: Wasn't Store supposed to be library-agnostic with React-specific methods in ReactStore?

romgrk: I did say that but it's not a good split. We're only doing React, it doesn't make sense to consider anything else, and some of these methods are too ergonomic to skip them. What I wanted to avoid was mainly the overhead from BUI features like the controlled values, which was the main reason to not merge the store classes in a single one. Now that I've removed controlled values, I'm considering merging everything, but I'm still hesitating a bit due to the `.selectors` logic — the `.context` logic is pretty useful to have for everyone though.

packages/utils/src/store/Store.ts
romgrk: By "this one", do you mean why isn't X importing this file directly?

It's because if we have Store tweaks to do, managing releases with BUI is too painful. I tried to do a `utils`-only release one time, but some people were opposed to it. So anyway, because we can't easily do tweaks on BUI utils, we have them duplicated in the X codebase at [multiple](https://github.com/mui/mui-x/tree/master/packages/x-scheduler-headless/src/base-ui-copy) [places](https://github.com/mui/mui-x/tree/master/packages/x-internals). Yay.

docs/src/app/(private)/experiments/storeWithControlledValues.tsx
```diff
@@ -72,10 +72,19 @@ interface Props {
...
+          openProp: props.open,
+          value: 0,
+        },
+        {},
+        selectors,
+      ),
   ).current;
 
-  store.useControlledProp('open', props.open, props.defaultOpen ?? false);
+  store.useControlledProp('openProp', props.open);
```
michaldudak, line 87: Do we actually still need `useControlledProp` now? If it's just to display a warning in dev, IMO we can just drop it and use `useSyncedValue`

romgrk: The check for `isControlled` is still there, and it feels a bit more declarative. I'd keep it for now.

Issue #1615, opened 2025-03-30
[docs-infra] SEO issues on https://base-ui.com/


oliviertassinari:
This issue is a subset of #588, focused on keeping track of the SEO issues that we need to fix. I will continue to add items as I notice them.

The list:

- [ ] 1. We are missing the canonical links on all the pages. This is important so Google knows to not duplicate the pages

<img width="556" alt="SCR-20240905-ulxz" src="https://github.com/user-attachments/assets/7c3cac43-07ac-478a-ab38-1166b4b86433">

https://search.google.com/search-console/index/drilldown?resource_id=sc-domain%3Abase-ui.com&item_key=CAMYDyAC&hl=en

<img width="545" alt="Image" src="https://github.com/user-attachments/assets/7a43fc2f-9f19-42a7-b68b-d927be2baab7" />

- [ ] 2. We use https://base-ui.com/react/components/slider for the URL. The "components" in the URL doesn't seem to serve a purpose; worse, it make the URL longer. Also, it prevents having a continuous keyword hit for "react slider", see

<img width="713" alt="Image" src="https://github.com/user-attachments/assets/649476a8-fb27-40b7-a76f-4946a0ff287e" />

https://analytics.moz.com/pro/research/page-grader?keyword=react%20slider&url=http%3A%2F%2Fbase-ui.com

pondorasti:
> * 2. We use [base-ui.com/react/components/slider](https://base-ui.com/react/components/slider) for the URL. The "components" in the URL doesn't seem to serve a purpose; worse, it make the URL longer. Also, it prevents having a continuous keyword hit for "react slider", see

@oliviertassinari still plans to tackle this? Happy to take this on!

Should be fairly easy to move all the components to a new [route group ](https://nextjs.org/docs/app/api-reference/file-conventions/route-groups), `/(public)/(content)/react/(components)/...` and have SEO friendly paths.

And then for backwards compatibility, we could scaffold routes on the old paths with [permanentRedirect](https://nextjs.org/docs/app/api-reference/functions/permanentRedirect) so that Google catches on to the changes.

LukasTy:
@mui/base-ui What are your thoughts on dropping the `components/` part from the URL?
It makes sense to me, but I'm curious to hear others' thoughts. Maybe there was a specific reason to include it? 🤔

mj12albert:
I guess because we have paths that are `/react/utils/use-render`?

LukasTy:
[...]

PR #3388, opened 2025-12-01
[docs] Add `canonical` link tag

pondorasti:

See #1615

> 1. We are missing the canonical links on all the pages. This is important so Google knows to not duplicate the pages

This PR follows up on an older SEO issues and adds canonical link tags for all public pages. 

This should fix all, if not most, google search console errors for duplicate canonical. So any URL with a hash or query param (https://base-ui.com/react/overview/quick-start#install-the-library, https://base-ui.com/react/overview/quick-start#portals, etc) will be indexed only once under the base url of the given page (https://base-ui.com/react/overview/quick-start).

CleanShot 2025-12-01 at 22 05 39@2x

### DIFF
--- a/docs/src/app/(public)/layout.tsx
+++ b/docs/src/app/(public)/layout.tsx
@@ -21,6 +21,9 @@ export default function Layout({ children }: React.PropsWithChildren) {
 
 export const metadata: Metadata = {
   metadataBase: new URL('https://base-ui.com'),
+  alternates: {
+    canonical: './',
+  },
 };
 
 export const viewport: Viewport = {


docs/src/app/(public)/layout.tsx
```diff
@@ -21,6 +21,9 @@ export default function Layout({ children }: React.PropsWithChildren) {
 
 export const metadata: Metadata = {
   metadataBase: new URL('https://base-ui.com'),
+  alternates: {
+    canonical: './',
+  },
 };
```
pondorasti, line 27: see https://nextjs.org/docs/app/api-reference/functions/generate-metadata#url-composition

dav-is (approved):

PR #3391, opened 2025-12-02
[popups] Refactor `useDismiss` to use an internal class instance

michaldudak:


packages/react/src/floating-ui-react/hooks/useDismiss.ts
```diff
@@ -482,227 +522,286 @@ export function useDismiss(
...
+  }, Object.values(controllerParameters));
+
+  controller.onPropsChange = () => {
+    setReferenceProps(controller.getReferenceProps());
+    setFloatingProps(controller.getFloatingProps());
+  };
+
+  controller.useSetup();
+
+  React.useEffect(() => {
```
flaviendelangle: Nitpick, but here you can save one function allocation:


```suggestion
  React.useEffect(controller.disposeEffect);
```

Where `disposeEffect` is:

```ts
  public disposeEffect() {
    return () => {
      this.cancelDismissOnEndTimeout.clear();
      this.clearInsideReactTreeTimeout.clear();
    }
  }
```

romgrk: `disposeEffect = ...` for the bound `this` though. The [`useTimeout` hook](https://github.com/mui/base-ui/blob/master/packages/utils/src/useTimeout.ts) is a good model for class-based hooks (and it saves another allocation by using `useOnMount` instead of `useEffect(..., [])`). I've been thinking we might want to create a base class for class hooks to standardize and/or manage automatic resource cleanup.

michaldudak:
After #4661, there's no more noticeable improvement in the mount time of this PR.

Issue #3389, opened 2025-12-02
[toast] Setting `swipeDirection` to both direction on the same axis on `Toast.Root` doesn't work


RaenonX:
# Get help

## Ask a question

I am not sure if this is a bug, and not finding any documentations mention about this.

If I set `swipeDirection={['right', 'left']}` on `<Toast.Root/>`, only one direction (either right or left) will work. The other direction will "bounce back."

mj12albert:
I can reproduce this: `<Toast.Root swipeDirection={['right', 'left']} />` only dismisses on left swipes, right doesn't work

PR #3392, opened 2025-12-02
[toast] Fix multiple swipe directions on same axis

mj12albert:
Fixes https://github.com/mui/base-ui/issues/3389

Preview: https://stackblitz.com/edit/qaq73qg2?file=src%2FApp.tsx


atomiks:
This was not really supported intentionally as trying to swipe in the opposite direction is just strange when there's a large distance to cover on the screen. Arguably, the types should disallow the same axis within the array

e.g. top-center shouldn't allow a bottom swipe

mj12albert:
Edge case for sure but on a phone it could be useful to dismiss full width toasts by swiping left or right, esp for one handed use and swiping with thumb, it's easier to swipe one way over the other

atomiks (approved):

PR #3394, opened 2025-12-02
[form] More permissive submit handler types

mj12albert:
[`no-misused-promises`](https://typescript-eslint.io/rules/no-misused-promises/#options) can be triggered when passing async functions to submit handlers

Stackblitz doesn't show the red squiggly for eslint issues and needs `npm run lint` in their terminal manually

Before: https://stackblitz.com/edit/vitejs-vite-mzas28cq?file=src%2FApp.tsx

After: https://stackblitz.com/edit/vitejs-vite-spmas6pv?file=src%2FApp.tsx

Also reproducible in the vite-css example following the "Linting with Type Information" [setup](https://typescript-eslint.io/getting-started/typed-linting/)

### DIFF
--- a/packages/react/src/checkbox/root/CheckboxRoot.test.tsx
+++ b/packages/react/src/checkbox/root/CheckboxRoot.test.tsx
@@ -358,7 +358,7 @@ describe('<Checkbox.Root />', () => {
       });
 
       await render(
-        <Form onSubmit={submitSpy}>
+        <Form onSubmit={submitSpy as unknown as Form.Props['onSubmit']}>
           <Field.Root name="test-checkbox">
             <Checkbox.Root />
           </Field.Root>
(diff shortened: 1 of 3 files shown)


packages/react/src/checkbox/root/CheckboxRoot.test.tsx
```diff
@@ -358,7 +358,7 @@ describe('<Checkbox.Root />', () => {
       });
 
       await render(
-        <Form onSubmit={submitSpy}>
+        <Form onSubmit={submitSpy as unknown as Form.Props['onSubmit']}>
```
atomiks, line 361: Curious why this errors?

michaldudak:
I'm wondering if this is actually a good idea. Native event handlers don't accept async functions. Also, if we're doing this, why limit ourselves to just the submit handler?

Accepting a promise-returning function might improperly suggest we're awaiting it.

There is a syntax that solves the issue quite elegantly: `<MyForm onSubmit={void onSubmit}>`. It requires an explicit `void`, so you know that the promise result will be discarded, not awaited.

Issue #3323, opened 2025-11-25
How to trigger initial validation in Field.Root?


Vinayaka-Hebbar:
I need help understanding how to trigger **initial validation** for a field created with `Field.Root`.

In my form, the field is marked as **required**, but `Field.Root` does **not** validate on mount.  
Because of this, the field does not show its error state until the user interacts with it.

### What I need help with

- How can I trigger validation on mount for a `Field.Root` field?
- Is there an API or recommended pattern to manually call validation (like `validate()` or `triggerValidation()` )?
- What is the correct approach to show initial error UI for required fields?

### Context

- Validation must run immediately when the component is rendered.
- Error state should be visible on initial load without user interaction.

mj12albert:
> * Validation must run immediately when the component is rendered

I don't think we would support this directly (e.g. as a `validationMode`) since it's not an ideal UX

> Is there an API or recommended pattern to manually call validation (like validate() or triggerValidation() )

There currently isn't, but I agree that this could be useful for other use-cases, we could possibly add it with an `actionsRef` prop ([example](https://master--base-ui.netlify.app/react/components/dialog#DialogRoot-actionsRef))

mj12albert:
```tsx
const formRef = React.useRef()

React.useEffect(() => {
  formRef.current?.requestSubmit();
}, [])

<Form ref={formRef} />
```
Tried this as a workaround but flushSync prevents it from working: https://github.com/mui/base-ui/blob/b185aafc58b6099985812ec75fcaccf90cae7b92/packages/react/src/field/useField.ts#L48-L49

Vinayaka-Hebbar:
I was able to implement a workaround on my side with my custom `FormStore` and `Form` wrapper (built on top of `react-hook-form`).

<details>
  <summary>FormStore.ts</summary>

```tsx
import {
  createFormControl,
  DefaultValues,
  ErrorOption,
  FieldErrors,
  FieldPath,
  FieldPathValue,
  FieldValues,
  FormState,
  UseFormProps,
  UseFormReturn,
} from 'react-hook-form';

import { createSelector, ReactStore } from '@base-ui-components/utils/store';

export interface FormStoreState<TValues extends FieldValues = FieldValues> {
  formState: FormState<TValues>;
  values: TValues;
}

export type FormStoreOptions<TValues extends FieldValues = FieldValues> = UseFormProps<TValues>;

function createFormSelectors<TValues extends FieldValues>() {
  return {
[...]

PR #3395, opened 2025-12-02
[form][field] Add `actionsRef`

mj12albert:
Adds a `actionsRef.validate` method to Form and Field so validation can be called imperatively (and arbitrarily).

Form:
```tsx
const actionsRef = React.useRef<Form.Actions>(null);

<Form actionsRef={actionsRef} />>

// validate all fields
actionsRef.current?.validate();
// validate one field by name
actionsRef.current?.validate('username');
```
Field:
```tsx
const actionsRef = React.useRef<Field.Root.Actions>(null);

<Field.Root actionsRef={actionsRef} />>

// validate the field
actionsRef.current?.validate();
```

Closes https://github.com/mui/base-ui/issues/3323


packages/react/src/form/Form.tsx
```diff
@@ -69,6 +70,25 @@ export const Form = React.forwardRef(function Form<
...
+  const handleImperativeValidate = React.useCallback((fieldName?: string | undefined) => {
+    const values = Array.from(formRef.current.fields.values());
+
+    if (fieldName) {
+      const namedField = values.find((field) => field.name === fieldName);
+      if (namedField) {
+        namedField.validate(false);
+      }
+    } else {
+      values.forEach((field) => {
```
LukasTy, line 82: Looks slightly more efficient.

```suggestion
    if (fieldName) {
      const namedField = formRef.current.fields.get(fieldName);
      if (namedField) {
        namedField.validate(false);
      }
    } else {
      Array.from(formRef.current.fields.values()).forEach((field) => {
```

atomiks:
> const actionsRef = React.useRef<Form.Actions>(null);

With `Form.Actions` accepting `null`, we should probably also adapt other `actionsRef` props to accept `null`. Right now you need to fill it in with the object with a `noop` `unmount()` function

packages/react/src/form/Form.tsx
mj12albert: Gotta do it like this because the [map key is the id](https://github.com/mui/base-ui/blob/49cfd5ec722a92a182f193cdc1f3f76e032785fb/packages/react/src/field/useField.ts#L36) and not the field name 😬

LukasTy: Gotcha.
Well, then maybe move the L74 to be inline in L82 to avoid unnecessary calculations in case they are not needed? 🤔

mj12albert: We need the array in the `if` as well to do a `.find()` on it for the field name ~

LukasTy: Oh yes, silly me. Sorry for causing commotion here. 🙈 
Regardless, it would be nice to have access to the field ID here to improve the performance and simplicity of the code.

PR #3406, opened 2025-12-04
[checkbox][switch] Match native off state in form submission

### LINKED ISSUE #3239: Checkbox off state doesn't behave in accordance to the corresponding html inputs in forms
Here's the repro: https://codesandbox.io/p/sandbox/base-ui-beta-4-checkbox-inconsistency-th4xrk

The issue is simple:

```jsx
<form action={console.log}>
  <input type="checkbox" name="vanila" value="42" defaultChecked />
  <input type="checkbox" name="vanila" value="69" />

  <Checkbox.Root name="base-ui" value="42" defaultChecked />
  <Checkbox.Root name="base-ui" value="69" />
</form>
```

Submitting this form will result with values: `[["vanila", "42"], ["base-ui", "42"], ["base-ui", "off"]]`. This is inconsistent with the HTML spec, where unchecked checkboxes submit no values. 

Moreover, react 19's form action pattern will reset the form to their default values. Submitting the same form for a second time will yield the desired outcome of `[["vanila", "42"], ["base-ui", "42"]]`.

<hr>

I would desire consistency with the html checkboxes, so that unchecked values do not appear in form submission.

atomiks:

Fixes #3239

**Breaking:** In form submission data, Checkbox and Switch no longer have an `"off"` value when unchecked to match native behavior. 

A new prop, `uncheckedValue`, has been added [if you want a value submitted when they're unchecked](https://stackoverflow.com/questions/1809494/post-unchecked-html-checkboxes).

### DIFF
--- a/packages/react/src/switch/root/SwitchRoot.test.tsx
+++ b/packages/react/src/switch/root/SwitchRoot.test.tsx
@@ -354,6 +354,77 @@ describe('<Switch.Root />', () => {
       expect(customSubmitSpy.lastCall.returnValue.get).to.equal('on');
     });
 
+    it.skipIf(isJSDOM)(
+      'should submit uncheckedValue when switch is off and uncheckedValue is specified',
+      async () => {
+        const submitSpy = spy((event) => {
+          event.preventDefault();
+          const formData = new FormData(event.currentTarget);
+          return formData.get('test-switch');
+        });
+
+        const { user } = await render(
+          <Form onSubmit={submitSpy}>
+            <Field.Root name="test-switch">
+              <Switch.Root uncheckedValue="off" />
+            </Field.Root>
+            <button type="submit">Submit</button>
+          </Form>,
+        );
+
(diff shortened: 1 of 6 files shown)


packages/react/src/switch/root/SwitchRoot.test.tsx
```diff
@@ -262,7 +262,139 @@ describe('<Switch.Root />', () => {
 
   describe('Form', () => {
     // FormData is not available in JSDOM
-    it.skipIf(isJSDOM)('should include the switch value in form submission', async () => {
+    it.skipIf(isJSDOM)(
+      'should include the switch value in form submission, matching native checkbox behavior',
```
michaldudak, line 266: This doesn't actually check if the `value` is submitted

atomiks: `Switch` only supports `"on"`

PR #3415, opened 2025-12-04
[docs-infra] Add `og:url` tag

pondorasti:

see https://github.com/mui/base-ui/issues/1615, closes https://github.com/mui/mui-public/issues/561

Continuation of https://github.com/mui/base-ui/pull/3388:

- configured meta `og:url` tags for all pages
- collocated `metadataBase` to the root layout

CleanShot 2025-12-04 at 11 58 04@2x

### DIFF
--- a/docs/src/app/(public)/layout.tsx
+++ b/docs/src/app/(public)/layout.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import type { Metadata, Viewport } from 'next/types';
+import type { Viewport } from 'next';
 import { GoogleAnalytics } from 'docs/src/components/GoogleAnalytics';
 import { DocsProviders } from 'docs/src/components/DocsProviders';
 import 'docs/src/styles.css';
@@ -18,13 +18,6 @@ export default function Layout({ children }: React.PropsWithChildren) {
     </DocsProviders>
   );
 }
-
-export const metadata: Metadata = {
-  metadataBase: new URL('https://base-ui.com'),
-  alternates: {
-    canonical: './',
-  },
-};
 
 export const viewport: Viewport = {
   themeColor: [
--- a/docs/src/app/layout.tsx
+++ b/docs/src/app/layout.tsx
@@ -39,6 +39,7 @@ export const viewport: Viewport = {
 };
 
 export const metadata: Metadata = {
+  metadataBase: new URL('https://base-ui.com'),
   title: {
     template: '%s · Base UI',
     default: 'Base UI',
@@ -47,7 +48,11 @@ export const metadata: Metadata = {
     site: '@base_ui',
     card: 'summary_large_image',
   },
+  alternates: {
+    canonical: './',
+  },
   openGraph: {
+    url: './',
     type: 'website',
     locale: 'en_US',
     title: {


docs/src/app/(public)/layout.tsx
```diff
@@ -19,13 +19,6 @@ export default function Layout({ children }: React.PropsWithChildren) {
   );
 }
 
-export const metadata: Metadata = {
```
mj12albert, line 22: Why move this? @pondorasti

LukasTy: So those settings would apply to both public and private (experiments) pages.

mj12albert: Does it matter if it's private? https://github.com/mui/base-ui/pull/3416

LukasTy: It doesn't, but it already has the other `og:` attributes.
Do you see any problem with the colocation of the config?

mj12albert: Nope just wondering ~

pondorasti: > Why move this? @pondorasti

Just cleaning up the code a bit. imo, it's much easier to follow things around when metadata configuration is collocated in one file and there's no _layout inheritance_ at play.

I would only keep the metadata from `/(public)/layout` only if it has a functional role.

pondorasti:
> Would using a current domain be possible and make more sense?

It's likely possible, but I'm not so sure what some of the benefits would be. 

I guess it comes down to meta tag testing, i.e. if you want to test opengraph previews and make them point back to the netlify preview url.

docs/src/app/layout.tsx
```diff
@@ -39,6 +39,7 @@ export const viewport: Viewport = {
 };
 
 export const metadata: Metadata = {
+  metadataBase: new URL('https://base-ui.com'),
```
brijeshb42, line 42: We do have `process.env. SITE_DEPLOY_URL` that'll have the correct value based on the environment. Don't need to hard code.

Janpot: I don't think `SITE_DEPLOY_URL` would suffice as I believe it contains https://master--base-ui.netlify.app/ on production deploys. I think the desired behavior would be
1. production: `base-ui.com` = `URL`
2. PRs: `deploy-preview-<pr-number>--base-ui.netlify.app` = `DEPLOY_PRIME_URL `
3. legacy: `v1.base-ui.com` = ? (currently not necessary yet for base, but it is for core and X)

I believe we should encode this in `withDeploymentConfig()` function in infra. We can base it off of the method used by [llms.txt generation](https://github.com/mui/base-ui/blob/1b23b4b3cbb48112928dbfed8789b58e4b1c088c/docs/scripts/generateLlmTxt/index.mjs#L22-L23):

```tsx
const NETLIFY_DEPLOYMENT_URL =
  process.env.PULL_REQUEST === 'true' ? process.env.DEPLOY_PRIME_URL : process.env.URL;
```

We should just add an optional variable in the mix for legacy deployments, e.g.

```tsx
// whatever we know from Netlify
const NETLIFY_DEPLOYMENT_URL =
  process.env.PULL_REQUEST === 'true' ? process.env.DEPLOY_PRIME_URL : process.env.URL;
// optional override from DEPLOYMENT_URL variable
const DEPLOYMENT_URL = process.env.DEPLOYMENT_URL || NETLIFY_DEPLOYMENT_URL;
```

LukasTy: Thanks, Jan, I think this makes the most sense. 👍 
If we ever need subdomains, it should also cover this use case the best. 🤔 
@pondorasti, could you update the solution to use the suggested URL resolution method?

michaldudak:
This has already been implemented in #4385.

PR #3421, opened 2025-12-04
[popover] Run all tests with detached triggers

michaldudak:
Runs the whole existing Popover test suite using both contained and detached triggers.

I recommend reviewing through https://github.dev/mui/base-ui/pull/3421 - seems much more manageable.


packages/react/src/popover/root/PopoverRoot.detached-triggers.test.tsx
michaldudak, line 1: These tests were extracted without any changes to keep the file smaller.

LukasTy (approved):

PR #3423, opened 2025-12-04
[button] Warn in dev mode when rendering a native link

mj12albert:


packages/react/src/use-button/useButton.ts
```diff
@@ -53,6 +54,10 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
         error(
           'A component that acts as a button was rendered as a native <button>, which does not match the default. Ensure that the element passed to the `render` prop of the component is not a real <button>, or set the `nativeButton` prop on the component to `true`.',
         );
+      } else if (elementRef.current.tagName === 'A') {
+        warn(
```
LukasTy: WDYT, would `error` work here as well? 🤔

LukasTy (approved):
Makes sense to me. 👍

packages/react/src/use-button/useButton.ts
atomiks: I wonder if this should just apply to the `<Button>` component for now? I'm not really certain on replacing the `<Menu.Item>` with `<Menu.Link>` given `Link` is a generic term yet it's an "item link"

mj12albert: Or we could do this post-v1?

atomiks:
It might be too late to add this warning given shadcn's integration:

Screenshot 2026-01-28 at 1 35 55 pm

---

Might be better to have a bullet point under "Usage guidelines" and a doc section about not using `<Button>` to render links (and get shadcn to update on it)

mj12albert:
Maybe we should warn against passing disabled if rendering a link instead? That's more of an actual a11y mistake than simply replacing the element with a link 🤔

atomiks:
> Maybe we should warn against passing disabled if rendering a link instead? That's more of an actual a11y mistake than simply replacing the element with a link 🤔

Another issue is that `role="button"` is set but shouldn't be if it's an `<a>` (but it still should be if it's a `<div>`). So `nativeButton` doesn't cover this during SSR

atomiks:
My current thoughts on this:

- Coordinate with shadcn to remove the invalid usage. A quick patch would be to get shadcn to reuse the Radix button logic for Base UI (given Radix doesn't have a button component anyway). @colmtuite 
- In the meantime (interim, 1.2.0 release), detect `<a>` in an effect and remove `role="button"`. No warning.
- In 1.3.0 or some later release eventually, add a warning for this usage and recommend what to do instead.

packages/react/src/use-button/useButton.ts
```diff
@@ -53,6 +54,10 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
...
+          'A component that acts as a button was rendered as an <a> tag, which could cause usability issues for keyboard and assistive tech users. Prefer using `<a>` directly.',
```
atomiks: shadcn updated, so we can likely warn now.

- This should say "causes" not "could cause"
- Mention `Menu.LinkItem` for the old Menu case, or possibly silence/ignore the warning in that case?

mj12albert: shadcn isn't using links in their regular `<Menu>`, it's dev mode and just a warning (not error) so maybe it's not worth making a special case for it? @atomiks

atomiks: I think a lot of users who were using the render pattern might not know of `Menu.LinkItem` yet, so silencing might be better or at least mentioning that component name in the warning message to let them know to migrate

michaldudak:
We're postponing it for v2.

PR #3441, opened 2025-12-05
[combobox][select] Respect `itemToStringValue` for `onFormSubmit`

atomiks:

Fixes object values with `<Form onFormSubmit>` passing the plain object instead of the stringified form value

### DIFF
--- a/packages/react/src/toast/content/ToastContent.tsx
+++ b/packages/react/src/toast/content/ToastContent.tsx
@@ -33,8 +33,8 @@ export const ToastContent = React.forwardRef(function ToastContent(
       return undefined;
     }
 
-    const resizeObserver = new ResizeObserver(recalculateHeight);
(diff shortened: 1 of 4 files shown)


packages/react/src/select/root/SelectRoot.tsx
```diff
@@ -168,7 +175,7 @@ export function SelectRoot<Value, Multiple extends boolean | undefined = false>(
     value,
     controlRef,
     name,
-    getValue: () => value,
+    getValue: () => fieldStringValue,
```
flaviendelangle, line 178: Since we only use `fieldStringValue` inside this callback, does it make sense to compute it lazily? 
Or is it used frequently enough to always compute it in the render?

atomiks: `value`/`getValue` are distinct for the `useField` hook

PR #3443, opened 2025-12-05
[toast] Fix `flushSync` dev error when toast is added

### LINKED ISSUE #3442: [Toast] flushSync error in v1.0.0-rc
# Bug report

https://github.com/user-attachments/assets/e070e478-daf2-4b9e-9d1b-89c87ef20688

## Current behavior

When adding an anchored toast using Toast.Positioner with an anchor prop, React throws the following error:

`flushSync was called from inside a lifecycle method. React cannot flush when React is already rendering. Consider moving this call to a scheduler task or micro task.`

The error occurs at Toast.Root inside the toast rendering loop. This happens even when deferring the toastManager.add() call using setTimeout, queueMicrotask, requestAnimationFrame

## Expected behavior

Adding an anchored toast should not trigger a flushSync error.

## Reproducible example

[Link](https://stackblitz.com/edit/psxau6po?file=src%2FApp.tsx,package.json)

## Base UI version

v1.0.0-rc.0

Chrome

## Which OS are you using?

macOS

atomiks:

Fixes #3442. `flushSync` should only be called in observer callbacks to prevent flickers but not inside the layout effect directly.

### DIFF
--- a/packages/react/src/toast/root/ToastRoot.tsx
+++ b/packages/react/src/toast/root/ToastRoot.tsx
@@ -146,7 +146,12 @@ export const ToastRoot = React.forwardRef(function ToastRoot(
     },
   });
 
-  const recalculateHeight = useStableCallback(() => {
+  /**
+   * Recalculates the natural height of the toast and updates it in the toast manager.
+   * @param flushSync Whether to flush the update synchronously. Use in observer
+   * callbacks to avoid visual flickers.
+   */
+  const recalculateHeight = useStableCallback((flushSync: boolean = false) => {
     const element = rootRef.current;
     if (!element) {
       return;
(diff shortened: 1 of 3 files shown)


packages/react/src/toast/root/ToastRoot.tsx
```diff
@@ -146,7 +146,7 @@ export const ToastRoot = React.forwardRef(function ToastRoot(
     },
   });
 
-  const recalculateHeight = useStableCallback(() => {
+  const recalculateHeight = useStableCallback((flushSync = false) => {
```
mnajdova: Please add a comment explaining when flashSync should be set to true.

PR #3445, opened 2025-12-05
[internal] Performance: fast `useStore` implementation

romgrk:
### Problem

The `useStore()` hook can be called many times per component. The naive implementation defers to React's `useSyncExternalStore()`, which creates one store subcription per call, which means one component ends up with multiple subscriptions to the store. This is unefficient because the component re-renders as an unit, and only requires 1 subscription.

### Solution

This PR introduces a wrapper around components to establish an `instance` context object per component instance, which we can use to call `useSyncExternalStore()` at most once per component. Subsequent calls just accumulate the selectors in an array, without calling `useSyncExternalStore()` again.

The one downside of this solution is that we need to wrap components to make store accesses faster, e.g.:

```tsx
// before
const TooltipTrigger = (props) => { /* ... */ }

// after
const TooltipTrigger = fastHooks.create((props) => { /* ... */ })
```

### Results

This change results in a 15-35% decrease in mount time for the contained tooltip triggers benchmark.

TBD: re-run benchmarks

### DIFF
--- a/packages/react/src/menu/root/MenuRoot.tsx
+++ b/packages/react/src/menu/root/MenuRoot.tsx
@@ -594,55 +594,55 @@ export interface MenuRootProps<Payload = unknown> {
    * To render a controlled menu, use the `open` prop instead.
    * @default false
    */
-  defaultOpen?: boolean;
+  defaultOpen?: boolean | undefined;
   /**
    * Whether to loop keyboard focus back to the first item
    * when the end of the list is reached while using the arrow keys.
    * @default true
    */
-  loopFocus?: boolean;
+  loopFocus?: boolean | undefined;
   /**
    * Whether moving the pointer over items should highlight them.
    * Disabling this prop allows CSS `:hover` to be differentiated from the `:focus` (`data-highlighted`) state.
    * @default true
    */
-  highlightItemOnHover?: boolean;
+  highlightItemOnHover?: boolean | undefined;
   /**
    * Determines if the menu enters a modal state when open.
    * - `true`: user interaction is limited to the menu: document page scroll is locked and pointer interactions on outside elements are disabled.
    * - `false`: user interaction with the rest of the document is allowed.
    * @default true
    */
-  modal?: boolean;
+  modal?: boolean | undefined;
   /**
    * Event handler called when the menu is opened or closed.
    */
-  onOpenChange?: (open: boolean, eventDetails: MenuRoot.ChangeEventDetails) => void;
+  onOpenChange?: ((open: boolean, eventDetails: MenuRoot.ChangeEventDetails) => void) | undefined;
   /**
    * Event handler called after any animations complete when the menu is closed.
    */
-  onOpenChangeComplete?: (open: boolean) => void;
+  onOpenChangeComplete?: ((open: boolean) => void) | undefined;
   /**
    * Whether the menu is currently open.
    */
-  open?: boolean;
+  open?: boolean | undefined;
   /**
    * The visual orientation of the menu.
    * Controls whether roving focus uses up/down or left/right arrow keys.
    * @default 'vertical'
    */
-  orientation?: MenuRoot.Orientation;
+  orientation?: MenuRoot.Orientation | undefined;
   /**
    * Whether the component should ignore user interaction.
    * @default false
    */
-  disabled?: boolean;
+  disabled?: boolean | undefined;
   /**
    * When in a submenu, determines whether pressing the Escape key
    * closes the entire menu, or only the current child menu.
    * @default false
    */
-  closeParentOnEsc?: boolean;
+  closeParentOnEsc?: boolean | undefined;
   /**
    * A ref to imperative actions.
    * - `unmount`: When specified, the menu will not be unmounted when closed.
@@ -650,23 +650,23 @@ export interface MenuRootProps<Payload = unknown> {
    *   Useful when the menu's animation is controlled by an external library.
    * - `close`: When specified, the menu can be closed imperatively.
    */
-  actionsRef?: React.RefObject<MenuRoot.Actions | null>;
+  actionsRef?: React.RefObject<MenuRoot.Actions | null> | undefined;
   /**
    * ID of the trigger that the popover is associated with.
    * This is useful in conjunction with the `open` prop to create a controlled popover.
    * There's no need to specify this prop when the popover is uncontrolled (i.e. when the `open` prop is not set).
    */
-  triggerId?: string | null;
+  triggerId?: (string | null) | undefined;
   /**
    * ID of the trigger that the popover is associated with.
    * This is useful in conjunction with the `defaultOpen` prop to create an initially open popover.
    */
-  defaultTriggerId?: string | null;
+  defaultTriggerId?: (string | null) | undefined;
   /**
    * A handle to associate the menu with a trigger.
    * If specified, allows external triggers to control the menu's open state.
    */
-  handle?: MenuHandle<Payload>;
+  handle?: MenuHandle<Payload> | undefined;
   /**
    * The content of the popover.
    * This can be a regular React node or a render function that receives the `payload` of the active trigger.
--- a/docs/reference/generated/tabs-root.json
+++ b/docs/reference/generated/tabs-root.json
@@ -5,11 +5,13 @@
     "defaultValue": {
       "type": "Tabs.Tab.Value",
       "default": "0",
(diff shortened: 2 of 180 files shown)


packages/utils/src/fastHooks.ts
```diff
@@ -0,0 +1,68 @@
...
+    return result;
+  };
+  FastComponent.displayName = (fn as any).displayName || fn.name;
+  return FastComponent;
+}
+
+export function createRef<P extends object, E extends HTMLElement>(
+  fn: (props: P, forwardedRef: React.Ref<E>) => React.ReactNode,
+): (props: P, forwardedRef: React.Ref<E>) => React.ReactNode {
+  return React.forwardRef<E, P>(create(fn as any)) as any;
```
romgrk: todo: pass ref

packages/utils/src/fastHooks.ts
```diff
@@ -0,0 +1,72 @@
...
+export type Instance = {
+  didInitialize: boolean;
+};
+
+type HookType = {
+  before: (instance: any) => void;
+  after: (instance: any) => void;
+};
+
+const hooks: HookType[] = [];
```
cherniavskii, line 13: Singleton in the utils package here, but I guess it's not harmful, even if utils package gets duplicated, right?

packages/react/src/menu/root/MenuRoot.tsx
cherniavskii, line 1: Did you consider a more explicit approach, where a single `store.useState` accepts multiple selectors and composes them together?

romgrk: Not really possible, some store calls might be spread across multiple files & hooks. The "popup" logic is shared by many components (tooltip, menu, popover) with hooks like `useFocus` or `useTriggerRegistration` accessing the common state interface from outside the component.

packages/utils/src/fastHooks.ts
romgrk: I've been considering de-duplicating the global state at runtime (`const hooks = globalThis.fastHooks_v1 ??= []`), but it's indeed not harmful. There is no logical bug from doing so, only lesser gains.

packages/react/src/menu/root/MenuRoot.tsx
cherniavskii: Gotcha, so in this example of MenuRoot we want to batch all `store.useState` calls, while some of them might be made in custom hooks, correct?

romgrk: Yes, that's correct.

packages/utils/src/fastHooks.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export function fastComponent<P extends object, E extends HTMLElement, R extends React.ReactNode>(
+  fn: (props: P) => R,
+): typeof fn {
+  const FastComponent = (props: P, forwardedRef: React.Ref<E>): R => {
+    const instance = useRefWithInit(createInstance).current;
+
+    let result;
+    try {
+      currentInstance = instance;
```
michaldudak, line 37: WDYT about attaching the current instance to the component function instead of keeping it global to solve the concurrent rendering problem?

romgrk: What do you mean by attaching to the component function? And which concurrent rendering problem?

romgrk: > which concurrent rendering problem

To expand on that, AFAIU concurrent rendering can interrupt rendering a React commit, but it does so in between component renders — it doesn't stack component render functions on top of another. In other words, concurrent renderering is not parallel rendering. So as long as one component render function exits/enters without another component render function being entered, then this implementation is sound.

Even if React was entering multiple component functions, we could use this implementation to push/pop the current instance on the stack, but I think it's safe to assume that React doesn't and won't ever do so — it would complexify their codebase substantially.

```js
const previousInstance = currentInstance
currentInstance = instance
const result = render(...args)
currentInstance = previousInstance
```

So my analysis is that concurrent rendering is not an issue, despite what copilot commented above. Server rendering is no different as we're always in a single-threaded context in javascript.

babel.config.mjs
```diff
@@ -21,6 +21,15 @@ export default function getBabelConfig(api) {
     ],
   ];
 
+  const displayNamePlugin = baseConfig.plugins.find(
+    (p) => p[2] === '@mui/internal-babel-plugin-display-name',
+  );
+  displayNamePlugin[1].allowedCallees ??= {};
+  displayNamePlugin[1].allowedCallees['@base-ui-components/utils/fastHooks'] = [
```
michaldudak:
```suggestion
  displayNamePlugin[1].allowedCallees['@base-ui/utils/fastHooks'] = [
```

packages/utils/src/fastHooks.ts
michaldudak: > What do you mean by attaching to the component function?

I meant having the `instance` and `hooks` as properties on the component function. But this won't help if React ever interleaves render functions of two instances of the same component function.

My biggest worry is that this breaks the assuptions React has about user code - it makes renders unpure. While all the tests pass today, we can't be sure React won't change anything in their internal implementation that will make it introduce subtle bugs. Also we don't test our implementation with suspended components. Concurrent rendering is poorly documented, so I'm not sure what's exactly going on there - perhaps that's the main reason I'm cautious.

But perhaps someone with better knowledge of React internals could chime in here. @eps1lon, can I pick your brains (or is there someone who can know this better)?  Is the solution proposed here safe enough?

romgrk: > My biggest worry is that this breaks the assuptions React has about user code - it makes renders unpure

React hooks are impure to start with — they're stateful functions that depend on hidden global state, and they even have weird rules that don't exists anywhere else. If React hooks are impure, then render functions that use hooks are also impure, by definition. This is just to say that complete purity has never been an option, no language can ever be completely pure, not even Haskell, but the absence of purity also doesn't mean that we can't make a function idempotent and well-behaved. The trick is just to make our impurity dependent on stuff that comes out of React hooks, which `currentInstance` does as it's dependent on a ref.

> But this won't help if React ever interleaves render functions of two instances of the same component function.

I haven't been able to imagine an issue with that. This hook checks if the current render differs from the previous one. If there is interleaving of different React commits, from this hook's point of view it's the same thing as if React was rendering the commits in full (aka without concurrent rendering): render functions are called one after the other. The only problem I could imagine is if React tries to render an older commit after a newer commit — but that would break a lot of user code (think about the hacks we sometimes put in effects & refs), so I feel that assumption is safe to make.

> Also we don't test our implementation with suspended components

I have considered that case. Essentially, suspense `throws` stuff around, that's why I'm wrapping with `try / finally` — and `finally` is guaranteed by the language spec to be executed before the stack exits the function. I also placed the `instance.didInitialize = true` inside the `try` block (and not the `finally` block) precisely because I have considered the scenario where Suspense interrupts a component's first render(s) mid-way.

packages/utils/src/fastHooks.ts
romgrk: Actually there wouldn't be less performance gains, there would only be more bundle size. Even if another library were to take BUI components and add wrappers around it with a different fast hooks version, those wrappers would still be different components that would use their own version of fast hooks, without interfering with this one. Using a `globalThis` state might actually introduce more bugs and break stuff, and using a module-global variable like here is more sound.

michaldudak:
I tested it quite extensively today and couldn't find any issues.

Add comments to the code explaining the implementation and rationale to reduce the number of raised eyebrows of external contributors and future us.

We're going to have a 1.0.1 release next week, and I'd like to have this merged in after the release, so it lives on master for another month, just in case we find something unexpected.

romgrk:
This is ready for merging on my side, please approve if there's nothing blocking.

michaldudak (approved):

packages/utils/src/fastHooks.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function setInstance(instance: Instance | undefined): void {
+  currentInstance = instance;
+}
+
+export function register(hook: HookType): void {
+  hooks.push(hook);
+}
+
+export function fastComponent<P extends object, E extends HTMLElement, R extends React.ReactNode>(
```
oliviertassinari, line 29: Should we add documentation for the next person to explain when to use this and its limits?

Beyond contributor value, there also seems to be user value: in the eye of someone who is used to canonical React code, this is advanced, so it looks like an opportunity to teach them something, so to gain more trust from them in us.

PR #3448, opened 2025-12-06
[accordion][docs] Fix incorrect Accordion multiple prop default

### LINKED ISSUE #3447: [Docs] Accordion API table incorrectly lists `multiple` default as `true` (contradicts v1.0.0-beta.5 release)
### Description
I noticed a discrepancy in the Accordion documentation regarding the `multiple` prop on the `Root` component.

In the **API Reference** table, the default value for `multiple` is listed as:
> **multiple** | `boolean` | `true`

However, the actual default is `false`. This was introduced as a breaking change in **v1.0.0-beta.5**, but the API table seems to be outdated.

### Proof
The [release notes for v1.0.0-beta.5](https://base-ui.com/react/overview/releases#v1.0.0-beta.5-accordion) explicitly state:
> "Breaking change: Change multiple prop to be false by default and add a demo (#3141)"

Additionally, the "Open multiple panels" demo in the current docs explicitly passes the `multiple` prop, further implying the default is single-open.

### Expected Behavior
The API Reference table should be updated to list the default value for `multiple` as `false`.

mj12albert:
Fixes #3447, we forgot to update the docs in #3141.


oliviertassinari:
> #3447

Oh, interesting. At the root, this is a regression of the code-infra (e.g. IntelliSense is wrong) in Base UI. It's correct in Material UI. 

In Material UI:
- `pnpm proptypes` converts the TypeScript props into prop-types. Then
- `pnpm docs:api` converts prop-types into .json. When it does so, it parses the AST to verify that the function component default prop and prop-types default docs are the same. If different, it crashes:

SCR-20251207-sean

It was solved in https://github.com/mui/material-ui/pull/22447 "We verify that actual default values are kept in sync with default values from the default annotation in yarn docs:api."

In Base UI, it looks like it does this:

- `pnpm docs:api` converts the TypeScript props into .json files. No checks are run. 

If we get more issues like #3447, we will start to want to enforce this constraint back to avoid having to go back to old PRs to finish the work we started. Issue created: https://github.com/mui/mui-public/issues/943 to track this opportunity.

Issue #3444, opened 2025-12-05
Render performance of BaseUI primitives (tooltip specifically)


misha-erm:
# Feature request

## Summary

Hello, 
I was comparing render performance of Radix vs BaseUI and noticed that BaseUI renders the same amount of components 2x slower. Here is the demo page https://radix-base-tooltip-performance.vercel.app/ and the repo https://github.com/misha-erm/radix-base-tooltip-performance.

In my "stress-test" I'm rendering 2k items wrapped in Tooltip. Also I tried the approach with detached triggers and lazy activation.

Wdyt, Is it something that's worth of investigation? Maybe there is a problem with my benchmark? Thanks in advance🙇🏻‍♂️

## Motivation
we at https://fibery.io currently use Radix at the moment but I closely follow base-ui's progress and thinking about want to migrate to it at one day. However our UI is pretty dense, so at the same time there might be a lot of tooltip/context-menu wrappers at the screen. We've had performance problems with wrappers already so 2x slowdown can be quite noticeable for us. Wonder if base-ui team want to look into this direction or it's sort of expected performance level

colmtuite:
@misha-erm If you update to `v1-0-0-rc-0`, you should notice a significant improvement.

misha-erm:
Damn, forgot to update before creating the issue.

Can confirm that it's faster now, impressive work. Without CPU throttling, the difference is barely noticeable now

Just for reference, here are the numbers I see now on my MBP with 4x throttling:

- radix: 254ms, 249ms, 241ms
- base: 271ms, 256ms, 403ms
- base with detached handlers and lazy activation: 125ms, 111ms, 109ms

Well, I guess the issue can be closed then.

romgrk:
And we still have more performance optimizations coming ([1](https://github.com/mui/base-ui/pull/3391), [2](https://github.com/mui/base-ui/pull/3445)). We should be noticeably faster than Radix soon.

oliviertassinari:
This is what I get with 4x CPU throttling, Chrome private mode, Apple M1 Pro:

- Radix Primitives: 249ms, 303ms, 235ms
- Base UI v1.0.0-rc.0: 391ms, 446ms, 385ms
- Base UI with [PR 3445](https://github.com/mui/base-ui/pull/3445): 333ms, 389ms, 351ms
- Material UI v7.3.6: 385ms, 425ms, 408ms (https://github.com/oliviertassinari/radix-base-tooltip-performance)
[...]

PR #3451, opened 2025-12-07
[internal] Refactor `useHover` hooks as classes

romgrk:
Counterpart of #3391

Refactor the `useHover` hooks into classes.

#### Results

Screenshot From 2025-12-07 21-38-48
Screenshot From 2025-12-07 21-38-52


packages/react/src/floating-ui-react/hooks/useHoverInteractionSharedState.ts
```diff
@@ -26,53 +27,46 @@ export interface HoverInteractionSharedState {
...
-  const unbindMouseMoveRef = React.useRef<() => void>(() => {});
-  const restTimeoutPendingRef = React.useRef(false);
-  const openChangeTimeout = useTimeout();
-  const restTimeout = useTimeout();
-  const handleCloseOptionsRef = React.useRef<SafePolygonOptions | undefined>(undefined);
+class HoverInteraction {
+  pointerTypeRef: string | undefined;
+  interactedInsideRef: boolean;
+  handlerRef: ((event: MouseEvent) => void) | undefined;
+  blockMouseMoveRef: boolean;
```
flaviendelangle: I don't think we should keep the `Ref` suffix for all those elements.

Other than that, the new implementation is both cleaner and probably more  performant 🥳

romgrk: I don't plan to keep the suffix, I just like to do small incremental commits, it lets me see easily when/where a regression occurs.

atomiks:
I don't see a difference with master running the tooltip benchmark here

romgrk:
This might be missing commits from master or my baseline might not have been up-to-date, but I'll wait before resuming the work here.

packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts
cherniavskii, line 1: Where do performance gains come from?

flaviendelangle: `useStableCallback` runs an effect on every render to make sure the function you call is the one created in the last render (and which has access to up-to-date variables around it).
This goes away with a store class that is initialized once on mount.

Same with `React.useCallback`, on every render React has to compare the old and new deps.

From what I understand, it's death by a thousand cuts. A lot of very cheap operations on a hook used in lots of components accross the page.
I would not expect similar gains when applying the same logic on a bigger component (but I still think the code is more readable on the Tree View post migration).

flaviendelangle: (maybe I'm missing other perf gains in this PR ofc)

For example, I think having lots of `useRef` in a component can have a small cost that adds up when compared to a class field. But I did not check React implementation of this hook.

flaviendelangle: @cherniavskii you can have a look at https://github.com/mui/base-ui/issues/3444 for some additional context

cherniavskii: Is the rule of thumb "extract everything stable into a class" correct?
And for effects or callbacks with dependencies, keep them in the hook (the one that would initialize the class), or in the `Hook.use` method?

flaviendelangle: Performancewise, maybe it's the optimal solution
I would keep internal DX in mind if this comes at the cost of our productivity :eyes: 

But I suggest that we let the Base UI team continue the exploration for some time.
Then see where we would have the biggest gain on X (mostly for the new version of the components, I don't suggest to spend time on something that would not benefit the new versions) and start from there.

I don't think it's worth converting a hook / component that has 2 `React.useCallback` and handles a big amount of data to a class, because the gain will be marginal (the `React.useCallback` is not the bulk of the work).
But I honestly don't have a clear vision of where this pattern should be used.

romgrk: > From what I understand, it's death by a thousand cuts.

This is it.

> For example, I think having lots of useRef in a component can have a small cost that adds up when compared to a class field. But I did not check React implementation of this hook.

For sure, instead of having on single big object in a contiguous memory region, you have many small objects spread in memory, so the [cache locality is bad](https://romgrk.com/posts/optimizing-javascript/#5-avoid-cache-misses). React compounds the issue by using a linked list instead of an arrya to store hook state.

```
const hookObject = {
  fields: {
    value: 1,
    next: {
      value: 2,
      next: {
        value: 3,
        next: null, } } }
}

const classObject = {
  fieldA: 1,
  fieldB: 2,
  fieldC: 3,
}
```

> But I suggest that we let the Base UI team continue the exploration for some time.

Yes, there are ergonomy problems with mixing hooks with classes due to typescript limitations, and we still need to think about sync'ing props at the right time in the render pipeline to avoid hard to debug issues. I want to test this out slowly with a few hooks.

romgrk:
Stale, I might redo this on top of master.

PR #3453, opened 2025-12-07
[tooltip][popover] Fix skipped viewport transitions

atomiks:

https://master--base-ui.netlify.app/react/components/tooltip#animating-the-tooltip

Sometimes the viewport (content) transition would be skipped.
In the recording, the first part is master, while the second is this branch.

https://github.com/user-attachments/assets/a57f2019-eb8a-42ca-bfa2-6a999078a1e7


packages/react/src/popover/popup/PopoverPopup.tsx
```diff
@@ -125,10 +126,7 @@ export const PopoverPopup = React.forwardRef(function PopoverPopup(
 
   // If there's just one trigger, we can skip the auto-resize logic as
   // the popover will always be anchored to the same position.
-  const autoresizeEnabled = React.useCallback(
-    () => store.context.triggerElements.size > 1,
-    [store],
-  );
+  const autoresizeEnabled = useStableCallback(() => store.context.triggerElements.size > 1);
```
michaldudak: `store` is stable, so `useCallback` should be slightly faster.

PR #3455, opened 2025-12-07
[menu] Avoid disabling modality on click after hover-open

atomiks:

Fixes a bug when a menu is opened on hover, and is then clicked to "stick": the `modal` prop would incorrectly be ignored


michaldudak:
Opening on hover doesn't work reliably now with `modal=true`. Also, the flashing sidebar when hovering over triggers is quite annoying.

https://github.com/user-attachments/assets/908115cf-05ba-409a-9e7b-11414ab1a370

PR #3458, opened 2025-12-07
[menu][docs] Fix `closeParentOnEsc` default

atomiks:


mnajdova:
@LukasTy this prop shouldn't even exists in the props table for menu, it makes sense only for submenus. We are missing an infra support for ignoring props, I let Connor know about this. For now, just updating the value and description should be sufficient.

PR #3462, opened 2025-12-08
[core] Rename packages to use the @base-ui org

mnajdova:
## BREAKING CHANGES

Renamed the packages:
- `@base-ui-components/react` to `@base-ui/react`
- `@base-ui-components/utils` to `@base-ui/utils`

The changes are done using find replace, the only reverted change is the one listed below in the TODOs.

TODOs (after publishing):
- [x] Update examples/vite-css to use the new org once the package is published, currently it fails as the latest tag for @base-ui/react does not exists on npm
- [x] Deprecate https://www.npmjs.com/package/@base-ui-components/react and point to https://www.npmjs.com/package/@base-ui/react

### DIFF
--- a/packages/react/src/popover/root/PopoverRoot.detached-triggers.test.tsx
+++ b/packages/react/src/popover/root/PopoverRoot.detached-triggers.test.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { createRenderer, isJSDOM } from '#test-utils';
 import { act, screen, waitFor } from '@mui/internal-test-utils';
-import { Popover } from '@base-ui-components/react/popover';
+import { Popover } from '@base-ui/react/popover';
 
 describe('<Popover.Root />', () => {
   beforeEach(() => {


flaviendelangle:
Not strictly related to this PR, but how do we push people to migrate from the old name to the new one if they don't see a new release?
Do you remember how we handled it for `@materia-ui/core`? Was the doc and other communication channels enough? Or did we log some warning when installing the latest version of the old package?

atomiks:
> Not strictly related to this PR, but how do we push people to migrate from the old name to the new one if they don't see a new release?

We can deprecate `@base-ui-components/react` and point them to `@base-ui/react` in the deprecation message. I've done this for some packages.

mnajdova:
> We can deprecate @base-ui-components/react and point them to @base-ui/react in the deprecation message. I've done this for some packages.

Sure, we can do this after the release @base-ui/react 👌 Adding it in the TODOs

PR #3463, opened 2025-12-08
[useRender] Fix element return type

### LINKED ISSUE #3461: Can't pass `useRender` return value into another component's `render` prop
# Bug report

## Current behavior

I'm trying to pass the return value of `useRender` into the `render` prop of another component, but Typescript won't let me without casting it to e.g. `any`.

The problem is that `useRender.render` accepts `React.ReactElement<Record<string, unknown>>` while the `useRender` return value is `React.ReactElement` which resolves to `React.ReactElement<unknown>`

## Expected behavior

Passing the result of a `useRender` call into a `render` prop should give no Typescript errors.

## Reproducible example

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgVwM4FMBK6B2ATdKOAXzgDMoIQ4ByAAQCMBDDAWmWFYGMrIdcYqAPRR0TLjCFp0rUfkI0AUKEiw42cfApVaozUsU8cqeAHkYAC0JwAvHAAUSOQSgAaOADovYSmFQkALhQMbHkoDwBhXgh+HBgABV9UAB4AIlQwJhxUgD4ASlscxEU4OFEYZCgcYKxcF0c4AjImZAAbGAAVJgBzADkmEHQg9Mzs9x8IP3dna2I8xWJFQxiTOAAxaBAACXRWsEIO9AAPeDsGmbdPbyTAmtCXSOjYhKS0sFyCmyKEErgjVd26EGcVsdzqhAaTRa7S6fQGQzgqXe4yS03BRDmS1K5Uq1WS5isRAuNgQgOBMFICC8HgmflIQhyC0UQA

## Base UI version

1.0.0-rc.0

atomiks:

Fixes #3461

Updates `React.ReactElement<Record<string, unknown>>` instances to just `React.ReactElement`

### DIFF
--- a/packages/react/src/use-render/useRender.spec.tsx
+++ b/packages/react/src/use-render/useRender.spec.tsx
@@ -53,7 +53,7 @@ const element7 = useRender({
 
 expectType<React.ReactElement, typeof element7>(element7);
 
-function MyComponent() {
+function App() {
   const element = useRender({ defaultTagName: 'div' });
   return <Button render={element} />;
 }


packages/react/src/use-render/useRender.ts
romgrk, line 1: Why do we require `React.ReactElement<Record<...>>` in the input parameters?

atomiks: Good point. It seems needless, and we can leave `React.ReactElement` to have no type param to loosen it.

michaldudak (approved):

PR #3464, opened 2025-12-08
[toast] Introduce a store

### LINKED ISSUE #3922: [toast] Critical toast properties calculations are based on object referential equality
# Bug report

## Current behavior

Currently, some toast properties critical calculations ([such as the `domIndex` calculation](https://github.com/mui/base-ui/blob/3d8adbdb1bb1db51e778a75aa972fcb668dd5d54/packages/react/src/toast/root/ToastRoot.tsx#L130)) are based on referential equality between the toast object passed to the toast root component and the ones in the internal toast context.

This makes it impossible to, among other use cases, enforce type-safety on the `data` field. This is a snippet from my `useToast` hook which wraps `Toast.useToastManager`:

```ts
export const useToast = () => {
    const toastManager = ToastPrimitive.useToastManager()

    [...]

    const list = toastManager.toasts.flatMap(toast => {
        const parsed = toastSchema.safeParse(toast.data)
        if (!parsed.success) {
            return []
        }

        // creating a new object breaks all internal toast calculations relying on referential equality
        return {
            ...toast,
            data: parsed.data,
        }
    })
}
```

## Expected behavior

The toast properties calculations should never be based on referential equality between the toast object passed as prop to the toast root component and the ones in the internal context, instead they should be based on the toast's unique identifier, which is passable to user-created objects matching the `ToastRootToastObject<unknown>` type

## Reproducible example

https://stackblitz.com/edit/vitejs-vite-twkafu6v?file=src%2FApp.tsx

## Base UI version

For example v1.1.0

Chrome

## Which OS are you using?

Mac OS

None

## Additional context

In the StackBlitz when the button is clicked and the toast is added its ref function fires an alert with its toast index variable, it should be zero but it's minus one (i.e. not found return value of the `indexOf` array function). If the `flatMap` at line 15 is changed to not spread the toast object into a new object but just pass it on, then the added toasts exhibit the correct toast index variable

flaviendelangle:
Fixes #3922

This should not be part of the stable release it can be reviewed after :+1: 

- [x] Introduce a store in `ToastProvider`
- [x] Use a memoized selectors to compute the index, visible index and offset of each toast (instead of looping on the list of each item)
- [x] Move all the actions to a store class to reduce the usage of `useStableCallback`

### DIFF
--- a/packages/react/src/toast/close/ToastClose.tsx
+++ b/packages/react/src/toast/close/ToastClose.tsx
@@ -2,7 +2,7 @@
 import * as React from 'react';
 import type { BaseUIComponentProps, NativeButtonProps } from '../../utils/types';
 import { useToastRootContext } from '../root/ToastRootContext';
-import { useToastContext } from '../provider/ToastProviderContext';
+import { useToastProviderContext } from '../provider/ToastProviderContext';
 import { useButton } from '../../use-button/useButton';
 import { useRenderElement } from '../../utils/useRenderElement';
 
@@ -18,7 +18,7 @@ export const ToastClose = React.forwardRef(function ToastClose(
 ) {
   const { render, className, disabled, nativeButton = true, ...elementProps } = componentProps;
 
-  const store = useToastContext();
+  const store = useToastProviderContext();
   const { toast } = useToastRootContext();
   const expanded = store.useState('expanded');
 
--- a/packages/react/src/toast/positioner/ToastPositioner.tsx
+++ b/packages/react/src/toast/positioner/ToastPositioner.tsx
@@ -12,7 +12,7 @@ import { useFloatingRootContext } from '../../floating-ui-react';
 import { NOOP } from '../../utils/noop';
 import type { ToastObject } from '../useToastManager';
 import { ToastRootCssVars } from '../root/ToastRootCssVars';
-import { useToastContext } from '../provider/ToastProviderContext';
+import { useToastProviderContext } from '../provider/ToastProviderContext';
 
 /**
  * Positions the toast against the anchor.
@@ -26,7 +26,7 @@ export const ToastPositioner = React.forwardRef(function ToastPositioner(
 ) {
   const { toast, ...props } = componentProps;
 
-  const store = useToastContext();
+  const store = useToastProviderContext();
 
   const positionerProps = (toast.positionerProps ?? EMPTY_OBJECT) as NonNullable<
     typeof toast.positionerProps
--- a/packages/react/src/toast/provider/ToastProviderContext.ts
+++ b/packages/react/src/toast/provider/ToastProviderContext.ts
@@ -6,7 +6,7 @@ export type ToastContext = ToastStore;
 
 export const ToastContext = React.createContext<ToastContext | undefined>(undefined);
 
-export function useToastContext() {
+export function useToastProviderContext() {
   const context = React.useContext(ToastContext);
   if (!context) {
     throw new Error('Base UI: useToastManager must be used within <Toast.Provider>.');
--- a/packages/react/src/toast/viewport/ToastViewport.tsx
+++ b/packages/react/src/toast/viewport/ToastViewport.tsx
@@ -6,7 +6,7 @@ import { useTimeout } from '@base-ui/utils/useTimeout';
 import { activeElement, contains, getTarget } from '../../floating-ui-react/utils';
 import { FocusGuard } from '../../utils/FocusGuard';
 import type { BaseUIComponentProps, HTMLProps } from '../../utils/types';
-import { useToastContext } from '../provider/ToastProviderContext';
+import { useToastProviderContext } from '../provider/ToastProviderContext';
 import { useRenderElement } from '../../utils/useRenderElement';
 import { isFocusVisible } from '../utils/focusVisible';
 import { ToastViewportCssVars } from './ToastViewportCssVars';
@@ -23,7 +23,7 @@ export const ToastViewport = React.forwardRef(function ToastViewport(
 ) {
   const { render, className, children, ...elementProps } = componentProps;
 
-  const store = useToastContext();
+  const store = useToastProviderContext();
   const windowFocusTimeout = useTimeout();
 
   const handlingFocusGuardRef = React.useRef(false);
--- a/packages/react/src/toast/root/ToastRoot.tsx
+++ b/packages/react/src/toast/root/ToastRoot.tsx
@@ -11,7 +11,7 @@ import type { ToastObject as ToastObjectType } from '../useToastManager';
 import { ToastRootContext } from './ToastRootContext';
 import { transitionStatusMapping } from '../../utils/stateAttributesMapping';
 import type { TransitionStatus } from '../../utils/useTransitionStatus';
-import { useToastContext } from '../provider/ToastProviderContext';
+import { useToastProviderContext } from '../provider/ToastProviderContext';
 import { StateAttributesMapping } from '../../utils/getStateAttributesProps';
 import { useRenderElement } from '../../utils/useRenderElement';
 import { useOpenChangeComplete } from '../../utils/useOpenChangeComplete';
@@ -99,7 +99,7 @@ export const ToastRoot = React.forwardRef(function ToastRoot(
 
   const swipeEnabled = swipeDirections.length > 0;
 
-  const store = useToastContext();
+  const store = useToastProviderContext();
 
   const [currentSwipeDirection, setCurrentSwipeDirection] = React.useState<
     'up' | 'down' | 'left' | 'right' | undefined


packages/react/src/toast/useToastManager.ts
```diff
@@ -121,7 +124,7 @@ export interface ToastManagerAddOptions<Data extends object>
 }
 
 export interface ToastManagerUpdateOptions<Data extends object>
-  extends Partial<ToastManagerAddOptions<Data>> {}
+  extends Partial<Omit<ToastObject<Data>, 'id'>> {}
```
flaviendelangle: This allows to use `store.updateToast` instead of `store.set('toasts, ....)` in `Toast.Root`.
If you want to keep some properties private (even though their is no runtime safe guard, it's only typing), I can create a `store.updateToastPublic` method that keeps the previous typing and expose it in `Toast.useToastManager` instead of `store.updateToast`

packages/react/src/toast/viewport/ToastViewport.tsx
```diff
@@ -289,25 +277,23 @@ export const ToastViewport = React.forwardRef(function ToastViewport(
...
-
   const highPriorityToasts = React.useMemo(
     () => toasts.filter((toast) => toast.priority === 'high'),
     [toasts],
   );
 
   return (
-    <ToastViewportContext.Provider value={contextValue}>
-      {numToasts > 0 && prevFocusElement && <FocusGuard onFocus={handleFocusGuard} />}
+    <React.Fragment>
```
flaviendelangle, line 274: This context was never used.
But if we need `viewportRef`, we can now use `useToastContext` since it's stable accross render 👌

packages/react/src/toast/viewport/ToastViewport.tsx
```diff
@@ -123,74 +111,71 @@ export const ToastViewport = React.forwardRef(function ToastViewport(
       win.removeEventListener('focus', handleWindowFocus, true);
     };
   }, [
-    pauseTimers,
-    resumeTimers,
-    viewportRef,
-    windowFocusedRef,
-    setFocused,
```
flaviendelangle, line 130: For me the logic still works here but I would like a double check 👀

michaldudak:
I'd rather not include this in v1. Introducing a Store to other components usually came with a couple of subtle bugs, so let's test it without rushing after the release.

flaviendelangle:
> I'd rather not include this in v1.

Clearly, I'm not in favor of rushing anything before the stable, unless the gains are huge :+1:

packages/react/src/toast/viewport/ToastViewport.tsx
```diff
@@ -123,74 +111,55 @@ export const ToastViewport = React.forwardRef(function ToastViewport(
...
-      }
-
-      resumeTimers();
-      setHovering(false);
-      setFocused(false);
-    }
-
-    doc.addEventListener('pointerdown', handlePointerDown, true);
+    const doc = ownerDocument(viewport);
+    doc.addEventListener('pointerdown', store.handleDocumentPointerDown, true);
```
michaldudak, line 130: `handleDocumentPointerDown` is not an arrow function, so by calling it like this, you'll lose `this` binding. Either make it an arrow function (preferred imo), or bind `this` manually with `.bind()`.

packages/react/src/toast/provider/ToastProviderContext.ts
```diff
@@ -1,39 +1,10 @@
 import * as React from 'react';
-import type {
-  ToastObject,
-  ToastManagerAddOptions,
-  ToastManagerPromiseOptions,
-  ToastManagerUpdateOptions,
-} from '../useToastManager';
+import type {} from '../useToastManager';
```
michaldudak: To remove

packages/react/src/toast/viewport/ToastViewport.tsx
flaviendelangle: Nice catch :+1:

flaviendelangle:
@greptileai

flaviendelangle:
@greptileai

flaviendelangle:
@michaldudak this PR is ready for another round of review :pray:

packages/react/src/toast/close/ToastClose.tsx
```diff
@@ -18,8 +20,9 @@ export const ToastClose = React.forwardRef(function ToastClose(
 ) {
   const { render, className, disabled, nativeButton = true, ...elementProps } = componentProps;
 
-  const { close, expanded } = useToastContext();
+  const store = useToastContext();
   const { toast } = useToastRootContext();
+  const expanded = useStore(store, selectors.expanded);
```
michaldudak: Other components (except Select, I think, as it was created first) use the `store.useState('expanded')` pattern.

packages/react/src/toast/provider/ToastProvider.tsx
```diff
@@ -336,61 +41,26 @@ export const ToastProvider: React.FC<ToastProvider.Props> = function ToastProvid
...
-      prevFocusElement,
-      promise,
-      remove,
-      resumeTimers,
-      scheduleTimer,
-      toasts,
-      update,
-    ],
-  ) as ToastContext<any>;
+  useIsoLayoutEffect(() => {
```
michaldudak: There's a hook for this: `store.useSyncedValues`

packages/react/src/toast/close/ToastClose.tsx
flaviendelangle: I migrated to ReactStore to use this hook :+1:

michaldudak:
Review by Codex:

The new `updateToast` implementation only mutates the toast list and no longer adjusts timers when `update()` changes timeout or type. This breaks existing behaviors where updating a toast from `timeout: 0` to a positive value should start auto-dismiss, or updating a loading toast to a non-loading type should schedule a timer; conversely, setting `timeout: 0` won’t cancel an already running timer. This is a regression from the previous provider logic and will cause updated toasts to never dismiss or to dismiss unexpectedly based on stale timers.

flaviendelangle:
@michaldudak I think the did the review before this morning's fix.
Here is Claude Code answer (we really live in a wonderful world :laughing: )




Analysis
The Original Problem
The commit 4da0b6914 ("Move actions to the store") had a simple updateToast implementation that only mutated the toast list:


updateToast<Data extends object>(id: string, updates: ToastManagerUpdateOptions<Data>) {
  this.set('toasts', this.state.toasts.map((toast) => 
    toast.id === id ? { ...toast, ...updates } : toast
  ));
}
This was a regression because the original ToastProvider code had timer adjustment logic in the update function.

The Fix (Already Applied)
The commit b39a21855 ("Fix tests + review: Michal") from today added the timer management logic back to updateToast at [store.ts:174-218](vscode-webview://1aoteh6lal273uckis57s80rjrog16doesr8knactmjj6ja2hghm/packages/react/src/toast/store.ts#L174-L218). The fix:

Cancels timers when timeout: 0 is set or type changes to loading
Schedules new timers when:
Timeout changes from 0 to a positive value
Type changes from loading to another type
Timeout value is explicitly updated
Pauses timers appropriately when viewport is expanded/out of focus
The promiseToast method was also simplified to delegate timer handling to updateToast by passing timeout in the update options.

Verdict
No action required - the issue identified by Codex has been addressed in the latest commit. The current implementation in [store.ts:174-218](vscode-webview://1aoteh6lal273uckis57s80rjrog16doesr8knactmjj6ja2hghm/packages/react/src/toast/store.ts#L174-L218) properly handles all the timer scenarios:

timeout: 0 → positive value: starts auto-dismiss timer
type: 'loading' → 'success'/'error': schedules timer
timeout: X → timeout: 0: cancels running timer

michaldudak:
Yup, Codex agrees, it's fixed :)

packages/react/src/toast/provider/ToastProviderContext.ts
```diff
@@ -1,39 +1,9 @@
...
-  scheduleTimer: (id: string, delay: number, callback: () => void) => void;
-}
-
-export type ToastContext<Data extends object> = ToastContextValue<Data>;
+export type ToastContext = ToastStore;
 
-export const ToastContext = React.createContext<ToastContext<any> | undefined>(undefined);
+export const ToastContext = React.createContext<ToastContext | undefined>(undefined);
 
 export function useToastContext() {
```
michaldudak: Perhaps we can call it `useToastStore` to differentiate it from `useToastRootContext`?

packages/react/src/toast/positioner/ToastPositioner.tsx
```diff
@@ -51,11 +51,8 @@ export const ToastPositioner = React.forwardRef(function ToastPositioner(
 
   const [positionerElement, setPositionerElement] = React.useState<HTMLDivElement | null>(null);
 
-  const domIndex = React.useMemo(() => toasts.indexOf(toast), [toast, toasts]);
-  const visibleIndex = React.useMemo(
-    () => toasts.filter((t) => t.transitionStatus !== 'ending').indexOf(toast),
-    [toast, toasts],
-  );
+  const domIndex = store.useState('toastIndex', toast.id);
+  const visibleIndex = store.useState('toastVisibleIndex', toast.id);
```
atomiks, line 56: Looks like this will also fix #3922

packages/react/src/toast/useToastManager.ts
flaviendelangle: > ToastManagerUpdateOptions now allows internal-only fields (ref, height, transitionStatus, limited) because it’s Partial<Omit<ToastObject, 'id'>>. That expands the public API and lets consumers put the toast into invalid states. Consider keeping the public type narrow (as before) and using an internal update type/method for ToastRoot height updates.


@atomiks Codex review is about this point
I don't have the full context here, but I'm fine going in either direction

atomiks: I'd keep the public types narrow as they are now, as the internal fields are intentionally private atm

flaviendelangle: OK, I'll split the update method in a public one and a private one then :+1:

packages/react/src/toast/positioner/ToastPositioner.tsx
flaviendelangle: Seems to also be fixed indeed, I added some tests

packages/react/src/toast/provider/ToastProviderContext.ts
flaviendelangle: I went with `useToastProviderContext` to remain consistent with the other part's contexts.

PR #3470, opened 2025-12-08
[menu] Allow side axis fallback for submenus by default

### LINKED ISSUE #3349: [menu] sub popup clips the viewport
# Bug report

## Current behavior

I noticed that on small enough viewports (i.e. mobile), nested sub-menus overflow the available space.

## Expected behavior

I would expect the positioner to automatically adapt to the constraints and render the sub menu (partially) on top of the root menu.

## Reproducible example

The bug is reproducible on the base ui website itself, see [Nested Menu](https://base-ui.com/react/components/menu#nested-menu) section.

Image

## Base UI version

v1.0.0-beta.7

Chrome Version 142.0.7444.176 (Official Build) (arm64)

## Which OS are you using?

macOS 26

N/A

## Additional context

N/A

atomiks:

Fixes #3349 

This uses the same collision avoidance config that Popover + Tooltip use for submenus. That is, it allows the submenus to flip to the bottom or top if `left`/`right` don't fit by default, which doesn't apply to the top-level root menu (since it's undesirable to have that flip from bottom/top to left/right)

### DIFF
--- a/docs/src/app/(public)/(content)/react/components/menu/demos/submenu/css-modules/index.tsx
+++ b/docs/src/app/(public)/(content)/react/components/menu/demos/submenu/css-modules/index.tsx
@@ -1,3 +1,4 @@
+'use client';
 import * as React from 'react';
 import { Menu } from '@base-ui/react/menu';
 import styles from './index.module.css';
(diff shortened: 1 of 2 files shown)


packages/react/src/menu/positioner/MenuPositioner.tsx
```diff
@@ -80,6 +81,7 @@ export const MenuPositioner = React.forwardRef(function MenuPositioner(
   if (parent.type === 'menu') {
     computedSide = computedSide ?? 'inline-end';
     computedAlign = computedAlign ?? 'start';
+    collisionAvoidance = POPUP_COLLISION_AVOIDANCE;
```
michaldudak: It might be confusing if we silently overwrite the prop.

```suggestion
    collisionAvoidance = componentProps.collisionAvoidance ?? POPUP_COLLISION_AVOIDANCE;
```

PR #3472, opened 2025-12-09
[internal] Use combined window listeners

romgrk:
Some components, amongst which tooltip triggers, add native event listeners on the `window`. Those native calls are slow, so we replace them with a combined listener that only attaches once.


romgrk:
Stale, could be re-considered if profiling shows those calls to be impactful.

PR #3475, opened 2025-12-09
[tabs] Do not initially select a disabled tab

### LINKED ISSUE #2098: [tabs] Disabled tabs should not be automatically picked as the initially selected tab
# Bug report

## Current behavior

If a tab is disabled, it still gets automatically selected.

## Expected behavior

- When no explicit selection is made by the consumer of the component (ie. via `value` or `defaultValue` props), disabled tabs should not be picked as the initially selected tab. The first non-disabled tab should be picked instead
- A disabled tab should still be selected if:
  - The consumer explicitly sets it as selected (via `defaultValue` or `value` props)
  - The tab was selected while enabled, and later disabled

## Reproducible example

## Base UI version

`1.0.0-beta.0`

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

I understand this may be an opinionated take — in any case, whatever is decided, it shoud be applied to all components whenever a value changes for reasons outside of the consumer's control (and not just tabs).

michaldudak:
This is a solution inspired by #2579 adapted to the current state where `value` is mandatory on Tab and TabPanel parts.

Due to how tabs are registered, this doesn't work with SSR - we don't know which tabs are disabled on the server. To overcome this, explicitly provide a `defaultValue` pointing to the first non-disabled tab.

Fixes #2098

### DIFF
--- a/packages/react/src/tabs/tab/TabsTab.tsx
+++ b/packages/react/src/tabs/tab/TabsTab.tsx
@@ -240,11 +240,11 @@ export interface TabsTabProps
   /**
    * Whether the Tab is disabled.
    *
-   * If a first Tab on a TabList is disabled, it won't initially be selected.
+   * If a first Tab on a `<Tabs.List>` is disabled, it won't initially be selected.
    * Instead, the next enabled Tab will be selected.
-   * However, this does not work this way during Server-Side Rendering, as it is not known
+   * However, it does not work like this during server-side rendering, as it is not known
    * during pre-rendering which Tabs are disabled.
-   * To work around this, ensure that `defaultValue` or `value` on `Tabs.Root` is set to an enabled Tab's value.
+   * To work around it, ensure that `defaultValue` or `value` on `<Tabs.Root>` is set to an enabled Tab's value.
    */
   disabled?: boolean;
 }
--- a/docs/reference/generated/tabs-tab.json
+++ b/docs/reference/generated/tabs-tab.json
@@ -15,7 +15,7 @@
     },
     "disabled": {
       "type": "boolean",
-      "description": "Whether the Tab is disabled.\n\nIf a first Tab on a TabList is disabled, it won't initially be selected.\nInstead, the next enabled Tab will be selected.\nHowever, this does not work this way during Server-Side Rendering, as it is not known\nduring pre-rendering which Tabs are disabled.\nTo work around this, ensure that `defaultValue` or `value` on `Tabs.Root` is set to an enabled Tab's value.",
+      "description": "Whether the Tab is disabled.\n\nIf a first Tab on a `<Tabs.List>` is disabled, it won't initially be selected.\nInstead, the next enabled Tab will be selected.\nHowever, it does not work like this during server-side rendering, as it is not known\nduring pre-rendering which Tabs are disabled.\nTo work around it, ensure that `defaultValue` or `value` on `<Tabs.Root>` is set to an enabled Tab's value.",
       "detailedType": "boolean | undefined"
     },
     "className": {


packages/react/src/tabs/indicator/TabsIndicator.test.tsx
```diff
@@ -202,5 +202,23 @@ describe('<Tabs.Indicator />', () => {
...
+            </Tabs.Tab>
+            <Tabs.Tab value={1}>Enabled Tab</Tabs.Tab>
+            <Tabs.Indicator data-testid="bubble" />
+          </Tabs.List>
+        </Tabs.Root>,
+      );
+
+      // Indicator should be hidden when selected tab is disabled
+      const bubble = screen.getByTestId('bubble');
+      expect(bubble).to.have.attribute('hidden');
```
LukasTy: Personally, I prefer the current behavior, where a tab is marked as selected regardless of the `disabled` state.
I feel like it conveys the data a bit more correctly. 🤔 
What is the reason for your choice of hiding the state in this case?

michaldudak: This causes an issue when the first tab is disabled and we're using SSR. On the server, the first tab is selected and that's what the indicator shows, but when React hydrates, the indicator transitions to the next non-disabled tab, which looks weird.

atomiks: I think it's fine to require the user to handle this manually (explicitly specify a non-disabled `defaultValue`) given it doesn't work on SSR either.

LukasTy: Am I missing something? But I meant this case, which changes in behavior.
Screenshot 2025-12-10 at 12 30 27
Currently, it shows the active indicator even on a disabled tab when it is selected (i.e. `defaultValue` like in the example), while in this PR the indicator is missing in the same case.

michaldudak: > I think it's fine to require the user to handle this manually (explicitly specify a non-disabled defaultValue) given it doesn't work on SSR either.

Fair enough. I reverted the TabsIndicator change.

packages/react/src/tabs/tab/TabsTab.tsx
```diff
@@ -232,6 +237,16 @@ export interface TabsTabProps
    * The value of the Tab.
    */
   value: TabsTab.Value;
+  /**
+   * Whether the Tab is disabled.
+   *
+   * If a first Tab on a TabList is disabled, it won't initially be selected.
```
LukasTy:
```suggestion
   * If a first Tab on a `<Tabs.List>` is disabled, it won't initially be selected.
```

packages/react/src/tabs/tab/TabsTab.tsx
```diff
@@ -232,6 +237,16 @@ export interface TabsTabProps
...
+   * Instead, the next enabled Tab will be selected.
+   * However, this does not work this way during Server-Side Rendering, as it is not known
```
LukasTy:
```suggestion
   * However, it does not work like this during Server-Side Rendering, as it is not known
```

packages/react/src/tabs/tab/TabsTab.tsx
```diff
@@ -232,6 +237,16 @@ export interface TabsTabProps
...
+   * during pre-rendering which Tabs are disabled.
+   * To work around this, ensure that `defaultValue` or `value` on `Tabs.Root` is set to an enabled Tab's value.
```
LukasTy:
```suggestion
   * To work around it, ensure that `defaultValue` or `value` on `<Tabs.Root>` has the value of an enabled Tab.
```

LukasTy (approved):
LGTM, nice work. 👍

Issue #3465, opened 2025-12-08
[toggle] Internal `pressed` check only happens if the `value` is truthy


michaelhazan:
# Bug report

## Current behavior

`Toggle` checks if the provided prop value is truthy, and only if so, it will check if the `groupContext` contains it.

*This behavior can be seen here*
https://github.com/mui/base-ui/blob/aec3cd5d8ee6aa1a1b00a424cd64a8396db334ce/packages/react/src/toggle/Toggle.tsx#L51

## Expected behavior

The `Toggle` should always check if the group contains it, regardless of if it is truthy or not.

## Reproducible example

https://stackblitz.com/edit/kiarpqg4?file=src%2FApp.tsx

## Base UI version

v1.0.0-rc.0

## Which browser are you using?

Firefox

## Which OS are you using?

Mac OS

## Additional context

I suggest initially checking if the `valueProp` is actually undefined rather than falsey, and only then ignoring the `groupContext` check.

mj12albert:
Didn't originally intend for it to work without a truthy value but I think it's OK to be more permissive

PR #3477, opened 2025-12-09
[toggle group] More permissive towards falsy toggle values

mj12albert:
Generate a default value for toggles inside a toggle group if the `value` is missing or falsy so it's still usable.

Closes https://github.com/mui/base-ui/issues/3465


packages/react/src/toggle/Toggle.tsx
```diff
@@ -37,26 +38,26 @@ export const Toggle = React.forwardRef(function Toggle(
     ...elementProps
   } = componentProps;
 
-  const value = valueProp ?? '';
-
+  const value = useBaseUiId(valueProp || undefined);
```
atomiks, line 42: Is the fallback necessary?

LukasTy: Yes, it's needed to handle cases, where value is an empty string: `""`
```suggestion
  // `|| undefined` handles cases, where value is falsy (i.e. "")
  const value = useBaseUiId(valueProp || undefined);
```

PR #3482, opened 2025-12-09
[menu] Fix focus not returning to trigger on <kbd>Esc</kbd> while pointer rests on popup

atomiks:

The `mouseleave` `onScrollMouseLeave` handler fires after pressing <kbd>Esc</kbd> due to it closing with `pointer-events: none`. This ends up causing the `FloatingFocusManager` to write to `preventReturnFocusRef` since the `trigger-hover` reason gets emitted after the `escape-key` reason, which blocks focus returning to the trigger.

The fix here is to check the open state (`store.select('open')`) before running the handler logic so it doesn't try to fire a new event. 

NB: regression test doesn't work without animations on the popup (this issue doesn't occur there) and would likely be flaky if it was attempted to be added


LukasTy (approved):
Nice and clean fix. 👍

PR #3487, opened 2025-12-10
[docs] Fix mobile header z-index

LukasTy:
Fixes certain elements (with explicit `z-index: 1`) covering the mobile navigation header.
Considered removing the explicit `z-index: 1` in those particular cases, but I'm not sure if I'm missing certain cases why they were added for, so, I thought that this is the less "evil" solution. 🤷 

Issues:

https://github.com/user-attachments/assets/38147a73-b5f8-4e15-8086-53ba177a565d


mnajdova (approved):
Nice catch 👌

Issue #3484, opened 2025-12-10
[Combobox] value and defaultValue Root Props type mismatch (null union)


Yonava:
# Bug report

## Current behavior

`Combobox.Root.Props` adds a union of `null` to the type of `defaultValue` but not on `value`. This means `value` isn't compatible with `defaultValue`.

## Expected behavior

`value` and `defaultValue` are the same type.

Assuming to control `Combobox` while beginning with no value, we would want to set `value` to `null` as opposed to leaving it `undefined`.

## Base UI version

v1.0.0-rc.0

PR #3488, opened 2025-12-10
[combobox][select] Add `null` as an option for the value prop

mnajdova:
Fixes https://github.com/mui/base-ui/issues/3484

The functionality was there, not sure why the types were not updated.


atomiks:
This also applies to Select

I remember this used to cause type issues with the previous implementation, but it looks like those issues don't apply now  👍

mnajdova:
> This also applies to Select

Updated too, nice catch 👌

PR #3499, opened 2025-12-11
[docs-infra] Fix Code Highlighting on Longer Demos

dav-is:
Upstream PR: https://github.com/mui/mui-public/pull/949 (merged)

We aren't properly highlighting longer chunks of code:

image

Can be seen here: https://base-ui.com/react/handbook/forms#hero:tailwind:index.tsx

And fixed in this PR: https://deploy-preview-3499--base-ui.netlify.app/react/handbook/forms#hero:tailwind:index.tsx


pnpm-lock.yaml
```diff
@@ -12316,7 +12322,7 @@ snapshots:
       kebab-case: 2.0.2
       lz-string: 1.5.0
       path-module: 0.1.2
-      prettier: 3.7.4
+      prettier: 3.6.2
```
LukasTy: This is probably here, because you branched off from a stale master copy? 🤔

LukasTy (approved):
Nice and quick fix, great work. 👍 
Shall we wait for this before the release, or it's a bit too much time to propagate such non-critical thing?
cc @michaldudak @mnajdova

pnpm-lock.yaml
dav-is: I updated prettier in the upstream PR, and now this change went away

LukasTy: Nice, looks great. 👍

PR #3501, opened 2025-12-11
[code-infra] Remove vale from critical dependencies

Janpot:
Download only when needed. The linting job will fail, but at least everything else runs in CI


dav-is (approved):

LukasTy (commented):
Nice workaround. 👍 
Makes sense, given we don't have any programmatic usage.

PR #3513, opened 2025-12-12
[code-infra] Reproduction of flakyness conditions

Janpot:
Forcing reliable conditions for a flaky test in https://github.com/mui/base-ui/pull/3121. The race condition is dependent on the responsivity of `requestAnimationFrame`. Under some conditions it fires after the click and resets the focus to the initial element before the expectation runs. The race condition can be made to reproduce reliably by throttling RAF a little and adding some idle time between the click and assertion.


Janpot:
@netlify, you ok bro?

Janpot:
Tests pass after rebasing.

PR #3520, opened 2025-12-12
[code-infra] Store test results circleci

Janpot:
Allows us to gain insights in the overall health and performance of the test suite over time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,8 @@ jobs:
       - run:
           name: Run tests on JSDOM
           command: pnpm test:jsdom:coverage
+      - store_test_results:
+          path: test-results
       - code-infra/upload-coverage:
           key: << parameters.react-version >>-jsdom
   test_lint:
@@ -160,6 +162,8 @@ jobs:
       - run:
           name: Run tests on headless Chromium
           command: pnpm test:chromium --coverage
+      - store_test_results:
+          path: test-results
       - code-infra/upload-coverage:
           key: << parameters.react-version >>-browser
   test_regressions:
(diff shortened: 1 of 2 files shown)


michaldudak (approved):

michaldudak:
@LukasTy see https://app.circleci.com/pipelines/github/mui/base-ui/18136/workflows/ac396a8a-d60e-4b47-bf0f-085447acac51/jobs/189827/tests and https://app.circleci.com/insights/github/mui/base-ui/workflows/pipeline/tests?branch=pull/3520

LukasTy:
> LukasTy see https://app.circleci.com/pipelines/github/mui/base-ui/18136/workflows/ac396a8a-d60e-4b47-bf0f-085447acac51/jobs/189827/tests and https://app.circleci.com/insights/github/mui/base-ui/workflows/pipeline/tests?branch=pull/3520

Damn, how did I miss those. 🙈 
~100 kB doesn't sound too bad, but still, that's ~0.5 MB currently for each push/CI run.

WDYT if we would only do this on CI run on `master` branch (i.e., after every PR merge)?

Janpot:
We've been doing this for a while in X (~170kB/run browser + 200kB/run jsdom) and it doesn't seem to account for much (last billing period)

We're currently at 663.3 MB of 2 GB used

LukasTy:
> We're currently at 663.3 MB of 2 GB used

Gotcha. So, the strategy would be to purge manually when we exceed the limit?
That works. 👌 
But wouldn't running only on master be sufficient and require even less purging? 🤔

Janpot:
> So, the strategy would be to purge manually when we exceed the limit?

It resets every month. Actually, just checked, we were already exceeding this last month. 813 credits vs. 2,757,296 in compute seems manageable. X is the large user though with 1.5 GB, followed by private

PR #3528, opened 2025-12-12
[docs-infra] Add site wide og:image

Janpot:

### DIFF
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -3,7 +3,6 @@ import * as path from 'path';
 import * as url from 'url';
 import * as fs from 'fs';
 import { withDeploymentConfig } from '@mui/internal-docs-infra/withDocsInfra';
-import transformMarkdownMetadata from '@mui/internal-docs-infra/pipeline/transformMarkdownMetadata';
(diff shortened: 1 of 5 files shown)


docs/next.config.mjs
```diff
@@ -22,29 +21,7 @@ const workspaceRoot = path.resolve(currentDirectory, '../');
 
 const withMdx = nextMdx({
   options: {
-    remarkPlugins: [
```
Janpot, line 25: @dav-is The plugin is breaking `og:image` on subpages

LukasTy: Is it still relevant? 🤔 
It seems to work on subpages: https://www.opengraph.xyz/url/https%3A%2F%2Fmaster--base-ui.netlify.app%2Freact%2Fcomponents%2Faccordion

What I think is missing is the title of the page in the OG title.
I.e. in the above example, I would expect to AT LEAST see `Base UI - Accordion`, or the same title as used for the page.
Maybe if the explicit `og:title` is omitted, the actual page title would be used? 🤔

Janpot: > Is it still relevant? 🤔

Nope, it has been fixed in the meantime, this PR was demonstrating the bug.

The metadata cans till be a bit reorganized though I believe

PR #3529, opened 2025-12-12
[docs-infra] Fix OpenGraph Images for Docs Pages

dav-is:
The landing page [has the opengraph image applied](https://orcascan.com/tools/open-graph-validator?url=https%3A%2F%2Fbase-ui.com%2F), but it [doesn't seem to be applying to children pages](https://orcascan.com/tools/open-graph-validator?url=https%3A%2F%2Fbase-ui.com%2Freact%2Fcomponents%2Faccordion). Fixed by upstream PR: https://github.com/mui/mui-public/pull/953

[Preview Deploy Landing Page](https://orcascan.com/tools/open-graph-validator?url=https%3A%2F%2Fdeploy-preview-3529--base-ui.netlify.app%2F)
[Preview Deploy Docs Page](https://orcascan.com/tools/open-graph-validator?url=https%3A%2F%2Fdeploy-preview-3529--base-ui.netlify.app%2Freact%2Fcomponents%2Faccordion)

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -42,7 +42,7 @@
     "lucide-react": "^0.556.0",
     "lz-string": "^1.5.0",
     "match-sorter": "^8.2.0",
-    "next": "16.0.9",
+    "next": "16.0.10",
     "postcss": "^8.5.6",
     "postcss-custom-media": "^11.0.6",
     "postcss-import": "^16.1.1",
--- a/docs/reference/generated/combobox-trigger.json
+++ b/docs/reference/generated/combobox-trigger.json
@@ -33,6 +33,13 @@
     "data-popup-open": {
       "description": "Present when the corresponding popup is open."
     },
+    "data-popup-side": {
+      "description": "Indicates which side the corresponding popup is positioned relative to its anchor.",
+      "type": "'top' | 'bottom' | 'left' | 'right' | 'inline-end' | 'inline-start' | null"
+    },
+    "data-list-empty": {
+      "description": "Present when the corresponding items list is empty."
+    },
     "data-pressed": {
       "description": "Present when the trigger is pressed."
     },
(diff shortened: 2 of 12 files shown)


Janpot:
when I checked this before it seemed like we can hoist more of the metadata and images to the top level /app folder
https://github.com/mui/base-ui/pull/3528

Janpot:
I was wondering, to avoid the plugin altogether, could it be an idea to invert the single source of truth for title and description?

```tsx
export const metadata = {
  title: 'Accordion',
  description:
    'A high-quality, unstyled React accordion component that displays a set of collapsible panels with headings.',
};

# {metadata.title}

<Subtitle>A set of collapsible panels with headings.</Subtitle>
<Meta name="description" content={metadata.description} />
```

I could see this have a few other advantages:
- Easier for maintainers to debug issues like the opengraph images, no underlying mechanisms that magic the title out of mdx and into metadata
- Easier to reuse the titles, within the page and on other page, e.g. import the title in a navigation component pr something
- Or just extract the whole page header
    ```tsx
    export const metadata = {
      title: 'Accordion',
      description:
        'A high-quality, unstyled React accordion component that displays a set of collapsible panels with headings.',
    };

    <PageHeader metadata={metadata} subtitle='A set of collapsible panels with headings.' />
    ```

dav-is:
> to avoid the plugin altogether

The plugin is necessary to extract metadata for the index, especially extracting each heading

> invert the single source of truth for title and description

I don't think this is necessary. The title is the first h1, the description is the first paragraph. It's a simple convention that reads well and is normal markdown.

The selection is also being committed to the index file, so if there's any mistake it should be present in the diff

> import the title in a navigation component

If you imported the page's metadata, it would require the page to be compiled before it could be used. If a navigation component imports all pages, then the first render in dev mode would require compiling all pages. On the other hand, the sitemap does not depend on page compilation.

PR #3532, opened 2025-12-13
[select] Fix support for transform animations when `alignItemWithTrigger` is active

atomiks:
Currently transform animations cause a misalignment with the positioning of `alignItemWithTrigger` (more noticeable with smaller initial scales or bigger lists). We disable this on our docs, but  [shadcn/create](https://ui.shadcn.com/create?base=base) is using this as a default. Plus Radix seems to handle it correctly.

Exaggerated for effect: https://stackblitz.com/edit/y6zpn7cf

https://deploy-preview-3532--base-ui.netlify.app/react/handbook/animation#animating-select-component-with-motion

### DIFF
--- a/docs/reference/generated/autocomplete-root.json
+++ b/docs/reference/generated/autocomplete-root.json
@@ -111,10 +111,10 @@
       "detailedType": "boolean | undefined"
     },
     "mode": {
-      "type": "'none' | 'list' | 'both' | 'inline'",
+      "type": "'none' | 'list' | 'inline' | 'both'",
       "default": "'list'",
       "description": "Controls how the autocomplete behaves with respect to list filtering and inline autocompletion.\n- `list` (default): items are dynamically filtered based on the input value. The input value does not change based on the active item.\n- `both`: items are dynamically filtered based on the input value, which will temporarily change based on the active item (inline autocompletion).\n- `inline`: items are static (not filtered), and the input value will temporarily change based on the active item (inline autocompletion).\n- `none`: items are static (not filtered), and the input value will not change based on the active item.",
-      "detailedType": "'none' | 'list' | 'both' | 'inline' | undefined"
+      "detailedType": "'none' | 'list' | 'inline' | 'both' | undefined"
     },
     "onItemHighlighted": {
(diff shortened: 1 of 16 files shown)


packages/react/src/select/popup/SelectPopup.tsx
```diff
@@ -216,131 +217,165 @@ export const SelectPopup = React.forwardRef(function SelectPopup(
...
+          const popupTop = positionerRect.top;
+          const popupHeight = positionerRect.height;
+          const textCenterY = textRect.top + textRect.height / 2;
+
+          const transformOriginY =
+            popupHeight > 0 ? ((textCenterY - popupTop) / popupHeight) * 100 : 50;
+
+          const clampedY = clamp(transformOriginY, 0, 100);
+
+          popupElement.style.setProperty('--transform-origin', `50% ${clampedY}%`);
```
atomiks: Since the addition of `try {} finally {}` breaks the diff:

This part is new: it maintains a correct transform origin based on the location of the item.

packages/react/src/select/popup/SelectPopup.tsx
```diff
@@ -216,131 +217,165 @@ export const SelectPopup = React.forwardRef(function SelectPopup(
...
-
-      const idealHeight = availableSpaceBeneathTrigger + offsetY + marginBottom + borderBottom;
-      let height = Math.min(viewportHeight, idealHeight);
-      const maxHeight = viewportHeight - marginTop - marginBottom;
-      const scrollTop = idealHeight - height;
+      // Ensure we remove any transforms that can affect the location of the popup
+      // and therefore the calculations.
+      const originalTransform = popupElement.style.transform;
+      popupElement.style.transform = 'none';
+      popupElement.style.removeProperty('--transform-origin');
```
atomiks, line 248: The key solution is to remove any transforms when doing the initial calculations (which is the "final" state of the popup location) then restore it back afterwards.

atomiks:
> P.S. Best to review by adding ?w=1 to the URL (hide whitespace).

TIL!

Issue #3534, opened 2025-12-13
[scroll area] Forced reflows / Layout trashing


oliviertassinari:
## Reproducible example

Since #2778, when there are multiple ScrollArea on screen (e.g., our docs pages), we force React to update the DOM to render each, so it means that each component needs to read the DOM. We have a cascade of forced reflows:

<img width="531" height="84" alt="SCR-20251213-msts" src="https://github.com/user-attachments/assets/a5daa2f9-ede1-4d76-8c8f-48809d163b88" />

https://693ae9e6a41ca30008471aaf--base-ui.netlify.app/react/components/select, from the before example of #3531

## Expected behavior

Intuitively, I wouldn't expect flushSync to be needed in this context; it feels like this fix hides a deeper problem.

## Base UI version

v1.0.0

## Which browser are you using?

Chrome

oliviertassinari:
@atomiks Nice clean-up of the logic in #3536 👍. Love to see us fixing stuff at the root.

As for #3534, as far as I can read the performance timeline on https://deploy-preview-3536--base-ui.netlify.app/react/components/select. We, in a loop, read, then write, then read, then write in the DOM. Proof:

<img width="699" height="272" alt="SCR-20260105-besj" src="https://github.com/user-attachments/assets/6e28ff0f-5128-4052-a8f3-5122e10cc5fb" />

9 times. So the PR doesn't solve the issue. It's coherent with what the code has. We read the DOM with stuff like `viewportRef.current`, and write with stuff like `thumbYEl.style.transform`.

Considering that under a light page, each read/write cycle in the DOM takes 1ms, a page with 10 instances of the scroll view component would slow the page rendering by 20ms. It's not that bad. But a human eye can perceive motion changes starting from [5ms differences](https://www.mdpi.com/2411-5150/3/1/5). Looking at the [alternatives](https://www.notion.so/mui-org/core-Scroll-Area-278cbfe7b66080e2a862ec8a1dd7a480):

- https://github.com/KingSora/OverlayScrollbars: they gave up https://github.com/KingSora/OverlayScrollbars/issues/703, while it's really bad on https://kingsora.github.io/OverlayScrollbars/, 100ms spent in layout trashing 🫠
- https://github.com/sakhnyuk/rc-scrollbars: not even aware
[...]

PR #3536, opened 2025-12-13
[scroll area] Perf improvements

atomiks:

Help with #3534 @oliviertassinari 

The key issue causing the "flicker" from what I found is that the `<Corner>` element's size wasn't being taken into account during the first calculation of the thumb position. Once the corner has its size rendered, then the thumb calculation was correct. Since multiple calculations were happening on mount, this caused a visible flicker that was masked with a sync state update.

- Prevented `computeThumbPosition` being called unnecessarily on mount (even though state updates were typically discarded) by blocking initial updates in observer callbacks and removing redundant effects
- Made some variable names smaller for bundle size
- Removed some memory allocations
- Removed refs inserted into the context memo dep array unnecessarily

Also cleaned up the code a bit to remove duplication and grouped some hook calls more precisely

Preview: https://deploy-preview-3536--base-ui.netlify.app/react/components/select

### DIFF
--- a/packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
+++ b/packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
@@ -14,21 +14,15 @@ import { useBaseUiId } from '../../utils/useBaseUiId';
 import { scrollAreaStateAttributesMapping } from './stateAttributes';
 import { contains } from '../../floating-ui-react/utils';
 
-interface Size {
-  width: number;
-  height: number;
-}
+const DEFAULT_COORDS = { x: 0, y: 0 };
+const DEFAULT_SIZE = { width: 0, height: 0 };
+const DEFAULT_OVERFLOW_EDGES = { xStart: false, xEnd: false, yStart: false, yEnd: false };
+const DEFAULT_HIDDEN_STATE = { x: false, y: false, corner: false };
 
-const DEFAULT_SIZE = {
-  width: 0,
-  height: 0,
-};
-const DEFAULT_OVERFLOW_EDGES = {
-  xStart: false,
-  xEnd: false,
-  yStart: false,
-  yEnd: false,
-};
+export type HiddenState = typeof DEFAULT_HIDDEN_STATE;
+export type OverflowEdges = typeof DEFAULT_OVERFLOW_EDGES;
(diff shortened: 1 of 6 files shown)


packages/react/src/scroll-area/root/ScrollAreaRoot.tsx
```diff
@@ -47,15 +37,21 @@ export const ScrollAreaRoot = React.forwardRef(function ScrollAreaRoot(
...
+
+  const scrollYTimeout = useTimeout();
+  const scrollXTimeout = useTimeout();
+
   const [hovering, setHovering] = React.useState(false);
   const [scrollingX, setScrollingX] = React.useState(false);
   const [scrollingY, setScrollingY] = React.useState(false);
-  const [cornerSize, setCornerSize] = React.useState<Size>(DEFAULT_SIZE);
-  const [thumbSize, setThumbSize] = React.useState<Size>(DEFAULT_SIZE);
+  const [cornerSize, setCornerSize] = React.useState<Dimensions>(DEFAULT_SIZE);
```
flaviendelangle: Nitpick but if we rename `Size` into `Dimensions`, I would also rename the constants and variables to be consistent 👍

atomiks: Probably better just to rename the import to `Size`

packages/react/src/scroll-area/root/ScrollAreaRootContext.ts
```diff
@@ -25,18 +25,8 @@ export interface ScrollAreaRootContext {
...
-  };
-  setHiddenState: React.Dispatch<
-    React.SetStateAction<{
-      scrollbarYHidden: boolean;
-      scrollbarXHidden: boolean;
-      cornerHidden: boolean;
-    }>
-  >;
+  hiddenState: { y: boolean; x: boolean; corner: boolean };
+  setHiddenState: React.Dispatch<React.SetStateAction<{ y: boolean; x: boolean; corner: boolean }>>;
```
LukasTy: Nitpick: Have you considered extracting this to a common `HiddenState` type? 🤔

LukasTy (approved):
Looks and works nicely. 👍 
Couldn't find issues, and Olivier's observed issue looks resolved. 👌

PR #3541, opened 2025-12-14
[popups] Fix touch `openMethod` when tapping outside element bounds on Safari

atomiks:
- For Dialog, Menu, Popover, Combobox, and Autocomplete, this fixes `openType` not being `"touch"` in `initialFocus` when tapping outside an element's bounds but inside the hitslop area
- For Select, this fixes `alignItemWithTrigger` not being deactivated on touch when tapping outside the trigger's bounds but inside the hitslop area

### DIFF
--- a/packages/react/src/utils/useOpenInteractionType.ts
+++ b/packages/react/src/utils/useOpenInteractionType.ts
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { InteractionType, useEnhancedClickHandler } from '@base-ui/utils/useEnhancedClickHandler';
+import { isIOS } from '@base-ui/utils/detectBrowser';
 
 /**
  * Determines the interaction type (keyboard, mouse, touch, etc.) that opened the component.
@@ -19,7 +20,7 @@ export function useOpenInteractionType(open: boolean) {
             // On iOS Safari, the hitslop around touch targets means tapping outside an element's
             // bounds does not fire `pointerdown` but does fire `mousedown`. The `interactionType`
             // will be "" in that case.
-            'touch',
+            (isIOS ? 'touch' : ''),
         );
       }
     },


packages/react/src/utils/useOpenInteractionType.ts
```diff
@@ -2,25 +2,44 @@
 import * as React from 'react';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { InteractionType, useEnhancedClickHandler } from '@base-ui/utils/useEnhancedClickHandler';
+import { ownerDocument } from '@base-ui/utils/owner';
 
 /**
  * Determines the interaction type (keyboard, mouse, touch, etc.) that opened the component.
  *
  * @param open The open state of the component.
```
LukasTy, line 10: Nitpick: consider adding JSDoc for the new param.

```suggestion
 * @param open The open state of the component.
 * @param referenceElement The element used to get the document of.
```

atomiks:
> Would this ever resolve to anything other than touch in the fallback case?
Could we sneak in a plain || 'touch' fallback? 🥷 🙈 😆

It is indeed more expensive than I'd like. This seems like a decent solution, though the interaction type will never be `""` in that case. Another solution is to use `isIOS()` but I am not sure if Android also has hitslop issues (Chrome touch DevTools doesn't seem to have it.)

packages/react/src/utils/useOpenInteractionType.ts
atomiks: Not applicable with your fallback suggestion

LukasTy (approved):
Nice final solution. 👍 
If we ever notice problems with this, a more robust solution could be explored.

I checked on Android, and at least on Chrome, it doesn't exhibit the same problem with empty `openMethod` when clicked in the "slop" area.

LukasTy:
> This seems like a decent solution, though the interaction type will never be `""` in that case.

Should we limit the type value to account for this? 🤔 
I don't think this is a BC. 🤷

atomiks:
> Should we limit the type value to account for this? 🤔

Since it doesn't happen on Android, we can limit this to iOS

PR #3553, opened 2025-12-15
[csp provider] Add `CSPProvider`

### LINKED ISSUE #2619: styleDisableScrollbar unexpectedly injected into NodeView rendering (Tiptap + React)
# Bug report

## Current behavior

I’m using base-ui inside a [Tiptap React NodeView](https://tiptap.dev/docs/editor/extensions/custom-extensions/node-views/react). When I use components like Popover or Select (which mount their popup elements to document.body via Floating UI), I encounter an unexpected error.

When the trigger is clicked, Tiptap throws an error because it sees the first rendered child without the data-node-view-wrapper attribute.

After inspecting, I noticed that base-ui temporarily injects a <style> element like this:

```html
<style data-href="base-ui-disable-scrollbar" data-precedence="base-ui:low">
  .base-ui-disable-scrollbar {
    scrollbar-width: none;
  }
  .base-ui-disable-scrollbar::-webkit-scrollbar {
    display: none;
  }
</style>
```

[This <style> tag](https://github.com/mui/base-ui/blob/79840aff4816e97718e9cbbe2cb51f8b3b8c01b2/packages/react/src/select/popup/SelectPopup.tsx#L390) appears under the React-rendered element (inside the NodeView renderer), but disappears almost instantly. It’s extremely hard to catch, but logging confirmed that it’s being inserted and removed.

This seems to confuse Tiptap’s NodeView renderer, which then fails because the unexpected <style> node is present in the DOM tree.

- With Radix UI, this issue does not happen.

- With base-ui, it consistently happens when using popup components (Popover/Select).

Steps to reproduce:

1. Use Tiptap’s ReactNodeViewRenderer.
2. Inside the NodeView, render a base-ui Popover or Select component.
3. Click the trigger to open the popup.
4. Observe that Tiptap throws an error (Please use the NodeViewWrapper component for your node view.).

## Expected behavior

Popup components should not cause transient <style> elements to appear inside the NodeView renderer.

## Reproducible example

## Base UI version

For example v1.0.2

## Which OS are you using?

Mac OS/Windows/Other etc.

Voiceover/JAWS etc.

## Additional context

### LINKED ISSUE #3550: Incompatibility with strict CSPs
# Bug report

## Current behavior
Scroll Area does not work with strict CSPs as there's inline styles.

## Steps to reproduce the bug:
Use any CSP with the style-src directive set to anything other than unsafe-inline.

## Base UI version

V1.0.0

atomiks:

This allows adding a `nonce` prop to every component to allow inline `<script>`/`<style>` elements with CSP using `CSPProvider`: https://deploy-preview-3553--base-ui.netlify.app/react/utils/csp-provider.

Affected components:

- `<ScrollArea.Root>`
- `<Select.Popup>`
- `<Tabs.Indicator>`
- `<Slider.Thumb>`

Fixes #3550
Fixes #2619 (via `disableStyleElements`)

---

I verified with a custom app (client-only though). 

From my findings:

- React inline styles `style={...}` don't trigger `style-src-attr 'none'` in a client-rendered app. The app needs to be server pre-rendered (`style="..."` present in the HTML while parsing) to trigger a violation. So popup components don't violate it, but other ones like Slider will.
- Passing `nonce` correctly fixes the issue. Since the `<style>` tag is hoisted, the first component to be created with a `nonce` means other ones don't need it, though in practice all of them should have the `nonce` still
- `<script>` violations apply only to SSR apps

## Alternatives

For inline `<style>` tags to remove scrollbars, there could be an option to avoid rendering them entirely so the user has to specify in the CSS manually. The only reason `<style>` needs to exist is for [scrollbar-width](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scrollbar-width) which is not yet Baseline Widely Available under our browser support guide as `::-webkit-scrollbar{display:none}` must exist for older browsers.

### DIFF
--- a/docs/src/app/(docs)/react/utils/csp-provider/page.mdx
+++ b/docs/src/app/(docs)/react/utils/csp-provider/page.mdx
@@ -56,7 +56,20 @@ This will ensure that all inline `<style>` and `<script>` tags rendered by BaseÂ
 
 ## Disable inline `<style>` elements
 
-You can avoid supplying a `nonce` if you disable inline `<style>` elements entirely and rely on external stylesheets only. Specify `disableStyleElements`:
+You can avoid supplying a `nonce` if you disable inline `<style>` elements entirely and rely on external stylesheets only. The relevant components are `<ScrollArea.Viewport>` and `<Select.Popup>` or `<Select.List>` when `alignItemWithTrigger` is enabled, which inject a style tag to disable native scrollbars.
+
+```html
+<style>
+  .base-ui-disable-scrollbar {
+    scrollbar-width: none;
+  }
+  .base-ui-disable-scrollbar::-webkit-scrollbar {
+    display: none;
+  }
+</style>
+```
+
+Specify `disableStyleElements` to remove these tags:
 
 ```jsx title="Disabling style elements"
 <CSPProvider disableStyleElements>{/* ... */}</CSPProvider>


Janpot:
Few observations that pop in my mind:
- It feels a bit weird to me to have this as a prop as I don't see a use case where you want to supply it on a case-by-case basis. It's really just all or nothing-at-all. A nonce context could solve that. It would also have the benefit that you can add style tags to components without it being a breaking change.
- Looking at this, it feels super inviting to just supply a static string to shut up CSP warnings, completely defeating its purpose.

atomiks:
`NonceProvider` would indeed be much more ergonomic

Janpot:
As an API for passing nonces, I think this looks good. (Though my suggestion for an implementation would be to just create an `InlineStyle` component that wraps the `style` element but is nonce aware.)

I was wondering, would it make sense to create a single provider for multiple global settings. e.g. `<BaseUiprovider disableStyleElements >` or `<BaseUiprovider nonce={nonce} >`

atomiks:
> (Though my suggestion for an implementation would be to just create an InlineStyle component that wraps the style element but is nonce aware.)

Since we have `<script>` tags as well, it'd need `InlineScript` too. So imo the `useNonce` hook seems fine to use everywhere.

> I was wondering, would it make sense to create a single provider for multiple global settings. e.g. <BaseUiprovider disableStyleElements > or <BaseUiprovider nonce={nonce} >

`DirectionProvider` and `Tooltip.Provider` already set the precedent for different providers, but this is possible. Although both of those don't necessarily have to be app-level configuration.

docs/src/app/(docs)/react/utils/csp-provider/page.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+`style-src` covers both style elements and attributes, but if you want to block only style elements while allowing style attributes, you can replace `style-src` with the `style-src-elem` directive in your CSP.
+
+`style-src-attr` will only trigger a violation while parsing pre-rendered (SSR) components in your HTML. If your CSP blocks inline style attributes, you have a few options:
+
+1. Relax your CSP by adding `'unsafe-inline'` to the `style-src-attr` directive (or using only `style-src-elem` instead of `style-src`). Style attributes specifically pose a less severe security risk than style elements, but this approach may not be acceptable in high-security environments.
+2. Render the affected components only on the client, so that no inline styles are present in the initial HTML.
+3. Manually unset inline styles and specify them in your CSS instead. Any component can have its inline styles unset, such as `<ScrollArea.Viewport style={{ overflow: undefined  }}>`. Note that you'll need to ensure you vet upgrades for any new inline styles added by Base UI components.
```
atomiks, line 90: Per @Janpot's suggestion. The Provider could be extended (or named differently).

`disableScriptElements` isn't necessary because it's not a default and must be turned on manually. 

`<CspProvider nonce="..." disableStyleElements disableInlineStyles>`

Disabling inline styles (that are present in the initial HTML) is probably going to take a bunch of work but is theoretically possible to solve `style-src-attr`

michaldudak:
> I was wondering, would it make sense to create a single provider for multiple global settings. 

I'd prefer not introducing it now and instead rely on multiple more granular providers. We can always introduce a helper wrapper that uses the NonceProvider, but let's not do this until there's a clear need for it.

michaldudak (approved):

atomiks:
@greptileai

docs/src/app/(docs)/react/utils/csp-provider/page.mdx
```diff
@@ -0,0 +1,92 @@
...
+function App({ nonce }) {
+  return <CSPProvider nonce={nonce}>{/* ... */}</CSPProvider>;
+}
+```
+
+This will ensure that all inline `<style>` and `<script>` tags rendered by Base UI components include the correct nonce attribute, allowing them to function under your CSP.
+
+## Disable inline `<style>` elements
+
+You can avoid supplying a `nonce` if you disable inline `<style>` elements entirely and rely on external stylesheets only. Specify `disableStyleElements`:
```
michaldudak: I'd expand on this a bit - users likely don't know where we use style tags.

PR #3554, opened 2025-12-15
[docs] Improve JavaScript animation handbook wording

### LINKED ISSUE #3548: [docs] The Javascript animation looks like their is a missing part
In https://base-ui.com/react/handbook/animation#javascript-animations

> Base UI relies on [element.getAnimations()](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAnimations) to detect if animations have finished on an element.
> When using Motion, the opacity property lets this detection work easily, so always animating opacity to a new value for exit animations will work. If it shouldn’t be animated, you can use a value close to 1, such as opacity: 0.9999.

I don't get the link between Motion and the opacity. Is it a reference to the examples in the previous sections?

I feels like a sentence is missing between the first and the second one

atomiks:

https://deploy-preview-3554--base-ui.netlify.app/react/handbook/animation#javascript-animations

Adjusts wording to be clearer and adds code demos, along with a new 'Select' section as it's special compared to other components. Note that https://github.com/mui/base-ui/pull/3532 will fix the select positioning for the demo.

Closes #3548

### DIFF
--- a/docs/src/app/(docs)/react/handbook/page.mdx
+++ b/docs/src/app/(docs)/react/handbook/page.mdx
@@ -48,9 +48,9 @@ A guide to animating Base UI components.
   - CSS transitions
   - CSS animations
   - JavaScript animations
-    - Elements removed from the DOM when closed
-    - Elements kept in the DOM when closed
-    - Select
+    - Animating components unmounted from DOM when closed with Motion
+    - Animating components kept in DOM when closed with Motion
+    - Animating Select component with Motion
     - Manual unmounting
 
 </details>


docs/src/app/(docs)/react/handbook/animation/page.mdx
```diff
@@ -76,18 +76,20 @@ Use the following Base UI attributes for creating CSS animations when a compone
 
 JavaScript animation libraries such as [Motion](https://motion.dev) require control of the mounting and unmounting lifecycle of components in order for exit animations to play.
 
-Base UI relies on [`element.getAnimations()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAnimations) to detect if animations have finished on an element.
-When using Motion, the `opacity` property lets this detection work easily, so always animating `opacity` to a new value for exit animations will work.
-If it shouldn't be animated, you can use a value close to `1`, such as `opacity: 0.9999`.
+Base UI relies on [`element.getAnimations()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAnimations) to detect if animations have finished on an element. When using Motion, `opacity` animations are reflected in `element.getAnimations()`, so Base UI automatically waits for the animation finish before unmounting the component. If `opacity` shouldn't be animated (such as in a translating drawer component), you can use a value close to `1`, such as `opacity: 0.9999`.
```
mj12albert:
```suggestion
Base UI relies on [`element.getAnimations()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAnimations) to detect if animations have finished on an element. When using Motion, `opacity` animations are reflected in `element.getAnimations()`, so Base UI automatically waits for the animation finish before unmounting the component. If `opacity` isn't part of your animations (such as in a translating drawer component), you can include it with a value close to `1`, such as `opacity: 0.9999` to ensure this detection is accurate.
```

mj12albert: Maybe more explicit?

docs/src/app/(docs)/react/handbook/animation/page.mdx
```diff
@@ -121,44 +123,51 @@ function App() {
...
-    </Select.Root>
+          </Popover.Popup>
+        </Popover.Positioner>
+      </Popover.Portal>
+    </Popover.Root>
   );
 }
 ```
 
+### Select
```
mj12albert: I wonder if people will try to look for this on the Select page 🤔 

With the docs search, this animations page is the last result when I search for `select motion` or `motion select`

docs/src/app/(docs)/react/handbook/animation/page.mdx
atomiks: Changed to:
> If `opacity` isn't part of your animation (such as in a translating drawer component), you should still animate it using a value close to `1` (such as `opacity: 0.9999`), so that Base UI can detect the animation.

docs/src/app/(docs)/react/handbook/animation/page.mdx
LukasTy: Maybe this would work best?
But I guess we would need a bit more complex search algorithm, which would be able to use relations between heading levels to discover such items and place them at the top. 🤔 

```suggestion
### Animate Select
```

docs/src/app/(docs)/react/handbook/animation/page.mdx
```diff
@@ -121,44 +123,51 @@ function App() {
...
+
+The Select component is initially unmounted before staying mounted after being interacted with. To animate it with Motion, a mix of the two previous approaches is needed.
```
LukasTy:
```suggestion
The Select component is initially unmounted but remains mounted after interaction. To animate it with Motion, a mix of the two previous approaches is needed.
```

docs/src/app/(docs)/react/handbook/animation/page.mdx
atomiks: Might be worth calling out that it lacks `keepMounted` and "mixes" the two together

PR #3557, opened 2025-12-16
[code-infra] Migrate to new build

brijeshb42:
Parent PR - https://github.com/mui/mui-public/pull/625

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -20,7 +20,7 @@
     "@base-ui/utils": "workspace:*",
     "@mdx-js/loader": "^3.1.1",
     "@mdx-js/react": "^3.1.1",
-    "@mui/internal-docs-infra": "0.3.1-canary.9",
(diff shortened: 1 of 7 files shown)


michaldudak:
What's the rationale behind the new `exports` field format?

```json
  "exports": {
    ".": {
      "require": {
        "types": "./index.d.ts",
        "default": "./index.js"
      },
      "default": {
        "types": "./index.d.mts",
        "default": "./index.mjs"
      }
    },
```

Since we have `"type": "commonjs"`, shouldn't "default" refer to CJS? I suppose it works this way, but IMO it would be slightly less confusing if we had either "require" and "import" or "import" and "default".

brijeshb42:
> What's the rationale behind the new exports field format?

Nothing in particular. Either combination should work. This is already what's added to the currently [published](https://app.unpkg.com/@base-ui/react@1.0.0/files/package.json) packages. So no changes here.

PR #3564, opened 2025-12-17
[toast] Fix timers not being rescheduled when updated

### LINKED ISSUE #2990: Ability to override `toast.loading' with toast.success or toast.error manually without using toast.promise
# Feature request

Currently, even if you update the toast.loading toast with success or error, it does not use the timeout of both, which keeps the toast lingering indefinitely as it doesn't call `scheduleTimer`. Would be great if I can manually override `loading`, without needing to use `toast.promise`.

## Summary

Same as above 

## Motivation

Same as above

atomiks:

Fixes #2990


atomiks:
@greptileai

PR #3567, opened 2025-12-17
[docs][select] Add placeholder styles

aarongarciah:
Adjusts the color of the value when the `data-placeholder` attribute is present. Since there are no specific docs around placeholders, I think it's a good way of documenting how to deal with placeholder styles without adding a specific demo.

|before|after|
|---|---|
|Screenshot 2025-12-17 at 12 53 02|Screenshot 2025-12-17 at 12 52 47|


aarongarciah:
Closing since we don't want to promote placeholders in Selects.

PR #3573, opened 2025-12-18
[select] Fix support for `max-height` popup style when `alignItemWithTrigger` is active

### LINKED ISSUE #3396: [select] Glitchy `alignItemWithTrigger` behavior when Popup has max-height
# Bug report

## Current behavior

The `alignItemWithTrigger` behavior of the `Select.Positioner` can be glitchy when there's a `max-height` on the `Select.Popup` and the list is overflowing. It can look as if the popover is scrolling away from the trigger.

## Expected behavior

Scrolling is contained within the overflowing list.

## Reproducible example

https://stackblitz.com/edit/ypr9pa86?file=src%2Findex.module.css

https://github.com/user-attachments/assets/31e33394-550c-4db8-a8b6-2dc7ff444632

## Base UI version

1.0.0-beta.7

Chrome

## Which OS are you using?

Mac OS

atomiks:

Built off #3532. Specifying `max-height` on `.Popup` is now respected without the popup moving while scrolling.

https://stackblitz.com/edit/dzf17jwn

Fixes #3396


packages/react/src/select/popup/SelectPopup.tsx
```diff
@@ -108,14 +108,15 @@ export const SelectPopup = React.forwardRef(function SelectPopup(
     const positionerStyles = getComputedStyle(positionerElement);
     const marginTop = parseFloat(positionerStyles.marginTop);
     const marginBottom = parseFloat(positionerStyles.marginBottom);
+    const maxPopupHeight = getMaxPopupHeight(popupRef.current);
```
michaldudak, line 114: This calls `getComputedStyle` every frame when scrolling. How about calling this when scroll starts, caching the value and clearing the cache `onScrollEnd`?

atomiks: That was already being done with `positionerStyles` though. Also Floating UI does the same when scrolling - I don't think it's a concern

michaldudak: It's not really a concern, just an opportunity to save some CPU cycles.

atomiks: Alright - this can be done in a separate PR as it was already the case in the code previously for the positioner element

PR #3575, opened 2025-12-18
[useScrollLock] Check if stable scrollbar gutter works more precisely to support custom scrollbars

chuganzy:
This PR updates `useScrollLock` to check if `scrollbar-gutter: stable` is actually functional.

On macOS, when scrollbar behavior is set to "Automatically based on mouse or trackpad" in System Settings, and there is custom scrollbar (a.k.a. `::-webkit-scrollbar`), `scrollbar-gutter: stable` does not work on `html` and `body` tags.

CodeSandbox:
https://codesandbox.io/p/sandbox/2883-forked-kyl4p8


packages/utils/src/useScrollLock.ts
```diff
@@ -95,8 +117,8 @@ function preventScrollInsetScrollbars(referenceElement: Element | null) {
       htmlStyles.overflowX === 'scroll' || bodyStyles.overflowX === 'scroll';
 
     // Values can be negative in Firefox
-    const scrollbarWidth = Math.max(0, win.innerWidth - html.clientWidth);
-    const scrollbarHeight = Math.max(0, win.innerHeight - html.clientHeight);
+    const scrollbarWidth = Math.max(0, win.innerWidth - body.clientWidth);
+    const scrollbarHeight = Math.max(0, win.innerHeight - body.clientHeight);
```
chuganzy, line 121: It disables the body scroll at the end, so scrollbar width is technically window.width - body.width, otherwise the layout shifts.

chuganzy:
This change was not enough for Safari.. Fixing.

chuganzy:
just a friendly reminder: this is not a `docs` update, but `package: utils` might be more suitable.

packages/utils/src/useScrollLock.ts
```diff
@@ -109,13 +133,15 @@ function preventScrollInsetScrollbars(referenceElement: Element | null) {
      * Do not read the DOM past this point!
      */
 
-    if (supportsStableScrollbarGutter) {
+    if (supportsStableScrollbarGutter(referenceElement)) {
+      updatedGutterOnly = true;
```
atomiks: Doesn't this function break this warning and cause layout thrashing?

> Do not read the DOM past this point!

chuganzy: No, as far as I can see.
https://deploy-preview-3575--base-ui.netlify.app/experiments/scroll-lock

My understanding is that the issue is when something affects DOM values that are later restored below that point. Since `supportsStableScrollbarGutter` does not change the DOM values themselves, it does not seem to cause a problem in that sense.

That said, you are absolutely right that it does touch the DOM, and it indeed is more accurate to move it earlier. I have addressed that in https://github.com/mui/base-ui/pull/3575/commits/1a84fe848f94ce61c874da44db76ea29bd232103.

atomiks:
This seems okay to me. When body has default margins, the behavior is a bit weird, though that is mostly a corner case

atomiks (approved):

PR #3580, opened 2025-12-18
[docs][select] Remove placeholders

aarongarciah:
Select placeholders don't add value and are not required in the vast majority of use cases. This PR removes all placeholders from Select demos so we don't promote its usage.

- In single selection demos, options with `null` value have been removed and a `defaultValue` has been added instead.
- In the multiple selection demo, the placeholder has been removed altogether because it already had a `defaultValue`.

Updated demos:
- Select
  - [Hero](https://deploy-preview-3580--base-ui.netlify.app/react/components/select)
  - [Multiple selection](https://deploy-preview-3580--base-ui.netlify.app/react/components/select#multiple-selection)
  - [Formatting the value](https://deploy-preview-3580--base-ui.netlify.app/react/components/select#formatting-the-value) (code snippet)
- Forms (handbook)
  - [Hero](https://deploy-preview-3580--base-ui.netlify.app/react/handbook/forms)
  - [React Hook Form](https://deploy-preview-3580--base-ui.netlify.app/react/handbook/forms#react-hook-form)
  - [TanStack Form](https://deploy-preview-3580--base-ui.netlify.app/react/handbook/forms#tanstack-form)
- [Animation](https://deploy-preview-3580--base-ui.netlify.app/react/handbook/animation#animating-select-component-with-motion) (handbook)

Note: Ideally all Selects should have a proper label, but I assume we don't want to add noise when showcasing individual components demos.

### DIFF
--- a/docs/src/app/(docs)/react/components/select/demos/hero/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/select/demos/hero/css-modules/index.tsx
@@ -3,7 +3,6 @@ import { Select } from '@base-ui/react/select';
 import styles from './index.module.css';
 
 const fonts = [
-  { label: 'Select font', value: null },
   { label: 'Sans-serif', value: 'sans' },
   { label: 'Serif', value: 'serif' },
   { label: 'Monospace', value: 'mono' },
@@ -12,7 +11,7 @@ const fonts = [
 
 export default function ExampleSelect() {
   return (
-    <Select.Root items={fonts}>
+    <Select.Root items={fonts} defaultValue="sans">
       <Select.Trigger className={styles.Select}>
         <Select.Value />
         <Select.Icon className={styles.SelectIcon}>
--- a/docs/src/app/(docs)/react/components/select/demos/hero/tailwind/index.tsx
+++ b/docs/src/app/(docs)/react/components/select/demos/hero/tailwind/index.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 import { Select } from '@base-ui/react/select';
 
 const fonts = [
-  { label: 'Select font', value: null },
   { label: 'Sans-serif', value: 'sans' },
   { label: 'Serif', value: 'serif' },
   { label: 'Monospace', value: 'mono' },
@@ -11,7 +10,7 @@ const fonts = [
 
 export default function ExampleSelect() {
   return (
-    <Select.Root items={fonts}>
+    <Select.Root items={fonts} defaultValue="sans">
       <Select.Trigger className="flex h-10 min-w-36 items-center justify-between gap-3 rounded-md border border-gray-200 pr-3 pl-3.5 text-base bg-[canvas] text-gray-900 select-none hover:bg-gray-100 focus-visible:outline focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-blue-800 data-[popup-open]:bg-gray-100">
         <Select.Value />
         <Select.Icon className="flex">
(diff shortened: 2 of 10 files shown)


atomiks:
> Select placeholders don't add value and are not required in the vast majority of use cases.

Two arguments for "placeholders" when there's an empty state (e.g. country select which can't have a default value, unlike the font example)

1. Empty state recognition

In a long form, users often scan for empty fields to see what is left to fill in. If a select already has a value filled in (the default option), it might look finished. A placeholder with a muted text color (like gray) clearly signals "This field is incomplete"

2. Consistent interaction model

Text inputs typically start empty with a placeholder, but select inputs that start pre-filled break this consistency, forcing the user to process two different types of "unfilled" states

---

I agree that in a typical web app UI, all selects are prefilled with some value, but in a traditional "form", selects start off empty.

For example, how would you model a "Select country" select widget without a placeholder? Which btw the Combobox has as a demo, and would need updating here too.

I would say at the very least, we need some kind of section documenting placeholders and recommended guidelines for them.

---

> Ideally all Selects should have a proper label, but I assume we don't want to add noise when showcasing individual components demos.

We should update the demos to use them because the demos contradict the recommendation in Usage guidelines and every other demo has a `<label>` as well

aarongarciah:
The idea is to not promote placeholders from the get go. The thing is, when a Select is properly labelled, there aren't many use cases for placeholders.

> 1. Empty state recognition
> In a long form, users often scan for empty fields to see what is left to fill in. If a select already has a value filled in (the default option), it might look finished. A placeholder with a muted text color (like gray) clearly signals "This field is incomplete"

Yeah, I see this pretty often. But again, if the Select is properly labelled, what value is the placeholder adding if the user is picking from a set of predefined values? Isn't an empty Select equally signaling "This field is incomplete"?

> 2. Consistent interaction model
> Text inputs typically start empty with a placeholder, but select inputs that start pre-filled break this consistency, forcing the user to process two different types of "unfilled" states

But a Select can start empty, right? No need for it to be prefilled. It's true that sometimes seeing placeholders in all inputs (text, select, etc.) it's visually appealing, and in rare cases they can help. For example, the month and year placeholders in this screenshot from shadcn helps know the format you need to type if you're using typeahead:

image

> I agree that in a typical web app UI, all selects are prefilled with some value, but in a traditional "form", selects start off empty.

I didn't mean all Selects should be prefilled. The only reason to prefill the demos in this PR was because there's no labels. They could be empty initially.

> For example, how would you model a "Select country" select widget without a placeholder? Which btw the Combobox has as a demo, and would need updating here too.

If that [demo](https://base-ui.com/react/components/combobox#input-inside-popup) had a label (as it should), what would be the added value of a placeholder? If the intention is to be able to "empty" the value, shouldn't a specific option be added? e.g. "None", instead of using the placeholder as label? If the field was required, the "None" option could be obviated, of course.

> I would say at the very least, we need some kind of section documenting placeholders and recommended guidelines for them.

A section on how to deal with placeholders would be nice e.g. recommending not using placeholders as a substitute for labels, etc.

I'm pretty sure users will keep using placeholders, and there are valid use cases, but maybe not using them in all demos helps set a better tone.

> We should update the demos to use them because the demos contradict the recommendation in Usage guidelines and every other demo has a <label> as well

👍

atomiks:
> Yeah, I see this pretty often. But again, if the Select is properly labelled, what value is the placeholder adding if the user is picking from a set of predefined values? Isn't an empty Select equally signaling "This field is incomplete"?

Colm had a thread about this - an empty button with no label looks weird when it's not an input (at least imo 😅). I haven't seen that anywhere on the web and it looks confusing. I think all buttons need to have something inside of them, regardless if they're labeled explicitly or not. https://x.com/colmtuite/status/1942903716602102018

> I'm pretty sure users will keep using placeholders, and there are valid use cases, but maybe not using them in all demos helps set a better tone.

The main issue is a `null` placeholder item may not be obvious how to use. A user might wonder "how do I have a placeholder item initially", as it may not be clear the default value of Select is `null`. They might also want to omit the item from the select list, but this breaks clearing expectations (a user cannot go back to `null` value if the item isn't in the list), so showing how to do it on the docs offers more guidance.

atomiks:
This will be replaced by #3604

PR #3581, opened 2025-12-19
[number field] Fix hidden input focus on submit

### LINKED ISSUE #3574: [number field] Input with `aria-hidden` focus prevents form submission
# Bug report

## Current behavior

In some scenarios when submitting a form with `Number Field` component in it, the following console warning appears and **form submit event is not triggered**:

```
Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: <input>
Ancestor with aria-hidden: <input> 
```

Component is used with following format settings:

```
format={{
    style: 'currency',
    currency: 'EUR',
    currencySign: 'standard',
    currencyDisplay: 'symbol',
    minimumFractionDigits: 2,
    maximumFractionDigits: 2,
}}
```

## A way to reproduce

1. Set up a basic HTML `<form>` with `Number Field` (with `Field` component wrapper in my case) in it and `<button type="submit">`.
2. Input some random number, e.g. `12`.
3. Click somewhere outside the input. 
4. The input gets formatted to `12,00 €`.
5. Comeback to input and edit cents, e.g. `12,33 €`.
6. Try to submit form, submit event is not triggered and above warning appears in browser console.

## Expected behavior

Editing `Number Field` component value should not block form submission

## Possible solution

Use `inert` instead of `aria-hidden="true"` on "mirror" input.

## Base UI version

v1.0.0

Google Chrome 143.0.7499.110

## Which OS are you using?

Mac OS Sequoia 15.7.2

atomiks:

Fixes #3574

The main problem in the issue is that `stepMismatch` was sometimes occurring, and the hidden `<input type="number">` was being focused. This emitted an `aria-hidden` warning and appeared to "block submission" but the real issue is that the submission was (correctly) blocked from `stepMismatch`, just that focus wasn't being relocated to the right input.

Native `<input type="number">` supports `step="any"` to turn off step validation (while internal controls still assume `step={1}`), which I have added here for parity. If users want a custom step _without_ step validation, this would require another prop, which can likely wait until an issue is created.

Screenshot 2025-12-28 at 5 44 10 pm

## Related

- `stepMismatch` validation is broken unless you add an explicit `min` like `min={0}`.  It's some kind of esoteric React bug with controlled number inputs? https://github.com/facebook/react/issues/12334. Syncing the `value` with `.setAttribute('value', ...)` doesn't seem to work in an effect. I've added this to the documentation for `step`.

### DIFF
--- a/packages/react/src/number-field/root/NumberFieldRoot.tsx
+++ b/packages/react/src/number-field/root/NumberFieldRoot.tsx
@@ -56,7 +56,7 @@ export const NumberFieldRoot = React.forwardRef(function NumberFieldRoot(
     min,
     max,
     smallStep = 0.1,
-    step = 1,
+    step: stepProp = 1,
     largeStep = 10,
     required = false,
     disabled: disabledProp = false,
@@ -90,6 +90,7 @@ export const NumberFieldRoot = React.forwardRef(function NumberFieldRoot(
 
   const disabled = fieldDisabled || disabledProp;
   const name = fieldName ?? nameProp;
+  const step = stepProp === 'any' ? 1 : stepProp;
 
   const [isScrubbing, setIsScrubbing] = React.useState(false);
 
@@ -547,7 +548,7 @@ export const NumberFieldRoot = React.forwardRef(function NumberFieldRoot(
         max={max}
         // stepMismatch validation is broken unless an explicit `min` is added.
         // See https://github.com/facebook/react/issues/12334.
-        step={step}
+        step={stepProp}
         disabled={disabled}
         required={required}
         aria-hidden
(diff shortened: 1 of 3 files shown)


packages/react/src/number-field/root/NumberFieldRoot.tsx
```diff
@@ -542,7 +545,10 @@ export const NumberFieldRoot = React.forwardRef(function NumberFieldRoot(
         value={value ?? ''}
         min={min}
         max={max}
-        step={step}
+        // stepMismatch validation is broken unless an explicit `min` is added.
```
mnajdova, line 549: Let's add a note in the props descriptions that min needs to be provided if the step prop is used.

mnajdova (approved):
Logic looks good, let's just update the props description before merging.

PR #3586, opened 2025-12-20
[navigation menu] Add support for opening menu on hover during hydration

### LINKED ISSUE #3570: [Navigation Menu]: Hover events that happen before hydration don't get replayed by React
# Bug report

## Current behavior

Render a NavigationMenu.Root with one or more NavigationMenu.Trigger + NavigationMenu.Content items on the server (SSR).
Load the page and immediately move the mouse over a dropdown trigger (before hydration completes).
Observe that the dropdown does not open on that first hover.
Hover away and hover the trigger again (after hydration) — now it opens normally.

The initial hover interaction is effectively “lost” during hydration, so users must hover again to open the menu.

## Expected behavior

If the pointer is already hovering a trigger when hydration completes (or if the first hover happens immediately after load), the corresponding menu should open without requiring a second hover.

## Reproducible example

https://base-ui.com/react/components/navigation-menu

## Base UI version

1.0.0

Chrome

## Which OS are you using?

Mac OS

theskillwithin:
Fixes: https://github.com/mui/base-ui/issues/3570

### DIFF
--- a/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
+++ b/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
@@ -328,24 +328,7 @@ describe('<NavigationMenu.Root />', () => {
     });
 
     it('opens when the trigger is already hovered when it mounts (hydration)', async () => {
-      const originalMatchMedia = window.matchMedia;
       const originalMatches = window.HTMLElement.prototype.matches;
-
-      window.matchMedia = ((query: string) => {
-        const matches = query === '(any-hover: hover)' || query === '(any-pointer: fine)';
-        return {
-          matches,
-          media: query,
-          onchange: null,
-          addListener() {},
-          removeListener() {},
-          addEventListener() {},
-          removeEventListener() {},
-          dispatchEvent() {
-            return false;
-          },
-        };
-      }) as any;
 
       window.HTMLElement.prototype.matches = function matches(selector: string) {
         if (selector === ':hover' && this.getAttribute?.('data-testid') === 'trigger-1') {
@@ -365,7 +348,6 @@ describe('<NavigationMenu.Root />', () => {
           expect(trigger).to.have.attribute('aria-expanded', 'true');
         });
       } finally {
-        window.matchMedia = originalMatchMedia;
         window.HTMLElement.prototype.matches = originalMatches;
       }
     });
--- a/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
+++ b/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
@@ -327,6 +327,49 @@ describe('<NavigationMenu.Root />', () => {
       expect(trigger).to.have.attribute('aria-expanded', 'true');
     });
 
+    it('opens when the trigger is already hovered when it mounts (hydration)', async () => {
+      const originalMatchMedia = window.matchMedia;
+      const originalMatches = window.HTMLElement.prototype.matches;
+
+      window.matchMedia = ((query: string) => {
+        const matches = query === '(any-hover: hover)' || query === '(any-pointer: fine)';
+        return {
+          matches,
+          media: query,
+          onchange: null,
+          addListener() {},
+          removeListener() {},
+          addEventListener() {},
+          removeEventListener() {},
+          dispatchEvent() {
+            return false;
+          },
+        };
+      }) as any;
+
+      window.HTMLElement.prototype.matches = function matches(selector: string) {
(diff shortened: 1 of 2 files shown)


packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
```diff
@@ -286,6 +288,56 @@ export const NavigationMenuTrigger = React.forwardRef(function NavigationMenuTri
...
+        return;
+      }
+
+      // Avoid `flushSync` during effects (React warns about this during hydration).
+      // This mirrors the `isHover` branch of `handleOpenChange` without forcing sync flushing.
+      setStickIfOpen(true);
+      stickIfOpenTimeout.clear();
+      stickIfOpenTimeout.start(PATIENT_CLICK_THRESHOLD, () => {
+        setStickIfOpen(false);
+      });
```
theskillwithin: this is probably not needed here?

theskillwithin:
Screenshot 2025-12-19 at 23 18 08

packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
```diff
@@ -327,6 +327,49 @@ describe('<NavigationMenu.Root />', () => {
       expect(trigger).to.have.attribute('aria-expanded', 'true');
     });
 
+    it('opens when the trigger is already hovered when it mounts (hydration)', async () => {
+      const originalMatchMedia = window.matchMedia;
+      const originalMatches = window.HTMLElement.prototype.matches;
+
+      window.matchMedia = ((query: string) => {
```
oliviertassinari: Material UI uses `css-mediaquery` to mock this during the test, to consider.

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
```diff
@@ -286,6 +288,56 @@ export const NavigationMenuTrigger = React.forwardRef(function NavigationMenuTri
     },
   });
 
+  const replayHoverOnHydration = useStableCallback(() => {
```
oliviertassinari: There is no need for `useStableCallback()` here, no? We can move the logic inside the `useIsoLayoutEffect` and ignore missing dependencies. leading to a faster runtime, not calling useStableCallback at each render?

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
```diff
@@ -286,6 +288,56 @@ export const NavigationMenuTrigger = React.forwardRef(function NavigationMenuTri
...
+    const win = ownerWindow(triggerElement);
+    const canHover =
+      typeof win.matchMedia === 'function' ? win.matchMedia('(any-hover: hover)').matches : true;
+    const hasFinePointer =
+      typeof win.matchMedia === 'function' ? win.matchMedia('(any-pointer: fine)').matches : true;
+
+    if (!canHover || !hasFinePointer) {
+      didReplayHoverOnHydrationRef.current = true;
+      return;
+    }
```
oliviertassinari: Do we need the bundle size cost of all this code?

Also `.matchMedia('(any-hover: hover)'` and `.matchMedia('(any-pointer: fine)')` tends to be unusable: https://github.com/mui/mui-x/issues/10039#issuecomment-1690755359 (not sure it's still true, but).

packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx
atomiks: `ScrollAreaViewport` has this:

```jsx
useIsoLayoutEffect(() => {
  // `onMouseEnter` doesn't fire upon load, so we need to check if the viewport is already
  // being hovered.
  if (viewportRef.current?.matches(':hover')) {
    setHovering(true);
  }
}, [viewportRef, setHovering]);
```

Could you refactor Navigation Menu to just call `handleOpenEvent(event)`, but changing the code to allow an undefined event?

```jsx
useIsoLayoutEffect(() => {
  if (triggerRef.current?.matches(':hover')) {
    handleOpenEvent(); // allow undefined `event`
  }
}, []);
```

theskillwithin:
very frustrating, i still can't account for the layout shift this causes when the pre hydration hover fires (see gif https://github.com/mui/base-ui/pull/3586#issuecomment-3677461056) (it looks like the panel opens about 5px further to the right than it should be, then it immediately corrects itself)

it's almost like it works too fast, initiating the panel before some calculations are done? the workaround fix in https://github.com/mui/base-ui/issues/3570 didn't have this problem somehow

PR #3592, opened 2025-12-21
[combobox][autocomplete] Add `loopFocus` prop

### LINKED ISSUE #3456: [combobox] Focus does not loop correctly — requires two Down Arrow presses for focus to wrap from last item to first
## Current Behavior
When using the Combobox component with loop focus enabled (which is the default, as it allows looping through items), pressing the down arrow key from the last item does not immediately focus the first item. Instead, the focus seems to disappear or go elsewhere (I'm not sure where), and I need to press the down arrow a second time to reach the first item. This disrupts smooth keyboard navigation.

## Expected Behavior
Pressing the down arrow from the last item should seamlessly loop back and focus the first item on the first press, without any skip or lost focus. For comparison, in similar libraries like Ark UI (https://ark-ui.com/docs/components/combobox), the combobox handles this looping correctly without requiring an extra key press.

## Steps To Reproduce
1. Go to the official Base UI Combobox documentation: https://base-ui.com/react/components/combobox.
2. Open the combobox popup (on any example).
3. **Use the down arrow to navigate to the last item.**
4. **Press down arrow again —focus doesn't go to the first item; press once more to get there.**

## Environment
- Base UI version: 1.0.0-rc.0
- Browser: Brave 1.85.111 (Official Build) (64-bit), Chromium: 143.0.7499.40
- OS: Windows 10

<br>

_I noticed that the example of combobox from [Floating UI](https://floating-ui.com/) home page has the same problem._

atomiks:

Closes #3456

Adds `loopFocus` prop (default: `true`) to `Combobox` + `Autocomplete`. The input is **always** included in the focus loop (per ARIA guidelines), but you can disable it optionally (and the input can't be / isn't part of the focus loop in that case).

NB: A `data-highlighted` state attribute (for when the `input` part has virtual focus) will be added in a separate PR

### DIFF
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -5,7 +5,7 @@ import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
 import { useOnFirstRender } from '@base-ui/utils/useOnFirstRender';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { useMergedRefs } from '@base-ui/utils/useMergedRefs';
-import { visuallyHidden } from '@base-ui/utils/visuallyHidden';
+import { visuallyHiddenInput } from '@base-ui/utils/visuallyHidden';
 import { useRefWithInit } from '@base-ui/utils/useRefWithInit';
 import { Store, useStore } from '@base-ui/utils/store';
 import { useValueAsRef } from '@base-ui/utils/useValueAsRef';
@@ -563,7 +563,7 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
   );
 
   const setSelectedValue = useStableCallback(
-    (nextValue: Value | Value[], eventDetails: AriaCombobox.ChangeEventDetails) => {
+    (nextValue: Value | Value[] | null, eventDetails: AriaCombobox.ChangeEventDetails) => {
       // Cast to `any` due to conditional value type (single vs. multiple).
       // The runtime implementation already ensures the correct value shape.
       onSelectedValueChange?.(nextValue as any, eventDetails);
@@ -852,15 +852,14 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
       const current = Array.isArray(selectedValue) ? selectedValue : EMPTY_ARRAY;
       const next = current.filter((v) => itemIncludes(registry, v, store.state.isItemEqualToValue));
       if (next.length !== current.length) {
-        setSelectedValueUnwrapped(next);
+        setSelectedValue(next, createChangeEventDetails(REASONS.none));
       }
       return;
     }
 
     const isStillPresent =
-      selectedValue == null
-        ? true
-        : itemIncludes(registry, selectedValue, store.state.isItemEqualToValue);
+      selectedValue == null ||
+      itemIncludes(registry, selectedValue, store.state.isItemEqualToValue);
     if (isStillPresent) {
       return;
     }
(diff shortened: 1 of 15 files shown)


oliviertassinari:
I was curious, maybe it would break https://deploy-preview-3592--base-ui.netlify.app/react/components/combobox#creatable, but no, it still seems OK.

docs/reference/generated/autocomplete-root.json
```diff
@@ -104,6 +104,12 @@
       "description": "The locale to use for string comparison.\nDefaults to the user's runtime locale.",
       "detailedType": "Intl.LocalesArgument | undefined"
     },
+    "loopFocus": {
+      "type": "boolean",
+      "default": "true",
+      "description": "Whether to loop keyboard focus back to the input when the end of the list is reached while using the arrow keys. The first item can then be reached by pressing <kbd>ArrowDown</kbd> again from the input, or the last item can be reached by pressing <kbd>ArrowUp</kbd> from the input.\nThe input is always included in the focus loop per [ARIA Authoring Practices](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/).",
```
mnajdova: Would be great to explain that putting false as a value, means that the arrow down does nothing on the last element.

packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1056,8 +1057,8 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
     activeIndex,
     selectedIndex,
     virtual: true,
-    loopFocus: true,
-    allowEscape: !autoHighlightMode,
+    loopFocus,
+    allowEscape: loopFocus,
```
mnajdova: Why was this changed? How does escape gets effected whether the focus is looping or not?

packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -680,6 +680,57 @@ describe('<Combobox.Root />', () => {
       });
     });
 
+    it('loops ArrowDown from last to first by default', async () => {
```
mnajdova: The test is duplicated, no? It's already included in the tests regarding the loopFocus prop.

packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -4654,6 +4705,190 @@ describe('<Combobox.Root />', () => {
...
+
+      // Then to last item
+      await user.keyboard('{ArrowUp}');
+
+      await waitFor(() => {
+        expect(input).to.have.attribute('aria-activedescendant', options[2].id);
+      });
+    });
+
+    it('does not loop focus when loopFocus={false}', async () => {
```
mnajdova:
```suggestion
    it('does not loop focus from last to first with ArrowDown when loopFocus={false}, async () => {
```

packages/react/src/combobox/root/AriaCombobox.tsx
atomiks: `allowEscape` (poor term from Floating UI) means focus can escape the list (aka return to the input) but this requires `loopFocus` to be `true` - just a limitation of the implementation. I noticed we should keep the `!autoHighlightMode` check as well to avoid breaking previous behavior.

mnajdova: Keeping the previous behaviour makes sense 👌 . For some reason initially I thought this is related to the <kbd>Escape</kbd> key press, but now reading the code more carefully is related for allowing the focus to move outside of the list, maybe we should rename it to be more explicit, but we can do this outside of this PR.

mnajdova: Btw, we were anyway throwing a warning here https://github.com/mui/base-ui/blob/master/packages/react/src/floating-ui-react/hooks/useListNavigation.ts#L283 if loopFocus was false, so this doesn't change the behaviour in any way 👌

mnajdova (approved):
Looks good 👌

Issue #3559, opened 2025-12-16
[select] [combobox] Allow to use Suspense for async data


KingSora:
# Feature request
Since React19 now officially supports async operations via the `use` hook I would suggest a better API for selectable components such as `combobox` and `select`.

The main idea & change is that every prop that has impact on the items (e.g. `items` array) is not applied to the `Root` component, instead those things are managed by one or more child components.

Here is a StachBlitz with the API I have envisioned: https://stackblitz.com/run?file=src%2FApp.tsx,src%2FuseLoadUsers.ts

```js
export default function ExampleAsyncSingleCombobox() {
  const id = React.useId();
  const { usersPromise, getSearchSuggestions } = useLoadUsers();
  return (
    <Combobox.Root
      onInputValueChange={(nextSearchValue) => {
        getSearchSuggestions(nextSearchValue);
      }}
    >
      <div className={styles.Label}>
          <Combobox.Input
            id={id}
            placeholder="e.g. Michael"
            className={styles.Input}
          />
          <div className={styles.ActionButtons}>
            <Combobox.Clear
              className={styles.Clear}
            >
              <ClearIcon className={styles.ClearIcon} />
            </Combobox.Clear>
            <Combobox.Trigger
              className={styles.Trigger}
            >
              <ChevronDownIcon className={styles.TriggerIcon} />
            </Combobox.Trigger>
          </div>
        </div>
      </div>

      <Combobox.Portal>
        <Combobox.Positioner className={styles.Positioner} sideOffset={4}>
          <Combobox.Popup className={styles.Popup}>
            <Suspense fallback={
                  <Combobox.Status className={styles.Status}>Loading</Combobox.Status>
               }>
              <SuspendedUsers usersPromise={usersPromise} />
            </Suspense>
          </Combobox.Popup>
        </Combobox.Positioner>
      </Combobox.Portal>
    </Combobox.Root>
  );
}

function SuspendedUsers({
  usersPromise,
}) {
  const users = usersPromise ? React.use(usersPromise) : [];

  return (
    <Combobox.List>
      {users.length === 0 && (
        <Combobox.Empty className={styles.Empty}>Empty</Combobox.Empty>
      )}
      {users.map((user) => (
        <Combobox.Item key={user.id} className={styles.Item} value={user}>
          <Combobox.ItemIndicator className={styles.ItemIndicator}>
            <CheckIcon className={styles.ItemIndicatorIcon} />
          </Combobox.ItemIndicator>
[...]

PR #3594, opened 2025-12-21
[autocomplete][combobox] Support external items mapping

atomiks:

This adds support for the cmdk-style composition pattern without requiring `items` on `Root`, and without forcing the filtering/rendering to happen through a function child:

```jsx
<Combobox.Root>
  <Combobox.Input />
  <Combobox.Portal>
    <Combobox.Positioner>
      <Combobox.Popup>
        <Combobox.List>
          {fruits.map((item) => (
            <Combobox.Item key={item} value={item}>
              <Combobox.ItemIndicator>
                <CheckIcon />
              </Combobox.ItemIndicator>
              <div>{item}</div>
            </Combobox.Item>
          ))}
        </Combobox.List>
      </Combobox.Popup>
    </Combobox.Positioner>
  </Combobox.Portal>
</Combobox.Root>
```

Before this, `Autocomplete` could support this shape reasonably well, but filtering had to happen externally. `Combobox` only had partial support and did not keep selection/highlight/empty-state behavior in sync.

## Implementation

- `Combobox.Item` filters itself and returns `null` when it should not be shown.
- `Combobox.Group` uses `hidden` instead of unmounting its children so item registration effects can still run.
- The rebased version simplifies the bookkeeping a bit by reusing `CompositeList` metadata for item values instead of maintaining a separate DOM-node-to-value map.

## Pros

- Works with a more composition-oriented API, including server-component-friendly usage where a function child would otherwise need to be hoisted into a client component.
- TS inference stays natural because the item type comes from the surrounding array.
- Rendering/order are fully controlled by user composition.

### Real

- This is still more internal complexity than the `items`-driven path.
- For selection modes with no `items`, Combobox still needs to force-mount the popup on interaction so labels and selected indices can be registered from the rendered tree.
- Virtualization still requires `items` or `filteredItems`, because unrendered rows are otherwise unknowable.

### Not really new / overstated

- Strings already work fine. Objects are only needed when the value/label are distinct, which is the same existing `itemToStringLabel` story rather than a new requirement introduced by this approach.
- Some of the implementation overhead was incidental rather than fundamental; the rebased version trims part of it by removing extra value-mapping plumbing.

## Validation

- `pnpm test:jsdom ComboboxRoot --no-watch`
- `pnpm test:jsdom AutocompleteRoot --no-watch`
- `pnpm typescript`
- `pnpm eslint docs/reference/generated/combobox-empty.json packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx packages/react/src/combobox/empty/ComboboxEmpty.tsx packages/react/src/combobox/group/ComboboxGroup.tsx packages/react/src/combobox/group/ComboboxGroupContext.ts packages/react/src/combobox/input/ComboboxInput.tsx packages/react/src/combobox/item/ComboboxItem.tsx packages/react/src/combobox/list/ComboboxList.tsx packages/react/src/combobox/popup/ComboboxPopup.tsx packages/react/src/combobox/positioner/ComboboxPositioner.tsx packages/react/src/combobox/root/AriaCombobox.tsx packages/react/src/combobox/root/ComboboxRoot.test.tsx packages/react/src/combobox/root/ComboboxRootContext.tsx packages/react/src/combobox/store.ts packages/react/src/combobox/trigger/ComboboxTrigger.tsx packages/react/src/composite/list/useCompositeListItem.ts`

### DIFF
--- a/docs/reference/generated/combobox-empty.json
+++ b/docs/reference/generated/combobox-empty.json
@@ -1,6 +1,6 @@
 {
   "name": "ComboboxEmpty",
-  "description": "Renders its children only when the list is empty.\nRequires the `items` prop on the root component.\nAnnounces changes politely to screen readers.\nRenders a `<div>` element.",
+  "description": "Renders its children only when the list is empty.\nAnnounces changes politely to screen readers.\nRenders a `<div>` element.",
   "props": {
     "className": {
       "type": "string | ((state: Combobox.Empty.State) => string | undefined)",
--- a/packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
+++ b/packages/react/src/autocomplete/root/AutocompleteRoot.test.tsx
@@ -574,6 +574,85 @@ describe('<Autocomplete.Root />', () => {
       expect(screen.getAllByRole('option')).toHaveLength(2);
     });
 
+    it('mode="list": filters when rendering items as children (no `items` prop)', async () => {
+      const tags = [
+        { id: 'apple', value: 'apple' },
+        { id: 'banana', value: 'banana' },
+        { id: 'cherry', value: 'cherry' },
+      ];
+
+      const { user } = await render(
+        <Autocomplete.Root mode="list">
+          <Autocomplete.Input data-testid="input" />
+          <Autocomplete.Portal>
+            <Autocomplete.Positioner>
+              <Autocomplete.Popup>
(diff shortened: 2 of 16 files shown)


KingSora:
@atomiks do you think this would also cover the usecase described here: https://github.com/mui/base-ui/issues/3559 ?

PR #3597, opened 2025-12-22
[internal] Add Claude local settings to gitignore

michaldudak:


michaldudak:
We tend to mostly use Codex, so we haven't prepared the project settings yet. I sometimes switch to Claude so I added this to gitignore not to accidentally commit this file.

PR #3604, opened 2025-12-24
[combobox][select] Add `placeholder` prop to `Value` part

### LINKED ISSUE #3720: <Select.Value /> placeholder
# Feature request

I would like an example in the docs showing how to add a placeholder to the select. Radix would have a `placeholder` prop on the <Select.Value /> element, but I could not find it in base ui.

## Summary

Add placeholder prop or show how to do it the base ui way

https://www.radix-ui.com/primitives/docs/components/select
-> placeholder property

## Motivation

Improve DX for most common use cases

atomiks:

When a `null` item isn't rendered in the list, `placeholder` acts as a typical placeholder, though it cannot be cleared.

Issues when trying to use a native `<label>`, which prompted the demos to use `Field`:

- Clicking a `<label>` opens the popup, but this doesn't match native `<select>` behavior. Trying to workaround this is quite expensive and heavy-handed. On the other hand, `<Field.Label>` links to the hidden input and only gives focus to the trigger and matches native behavior.
- Hovering the label activates `.Trigger:hover`, which unlike Checkbox/Radio, doesn't really look or feel right. `<Field.Label>` also avoids this.

Select: https://deploy-preview-3604--base-ui.netlify.app/react/components/select
Combobox: https://deploy-preview-3604--base-ui.netlify.app/react/components/combobox#input-inside-popup

Closes #3720

### DIFF
--- a/packages/react/src/select/item/SelectItem.test.tsx
+++ b/packages/react/src/select/item/SelectItem.test.tsx
@@ -12,13 +12,6 @@ import { expect } from 'chai';
 
 describe('<Select.Item />', () => {
   const { render } = createRenderer();
-  const { render: renderFakeTimers, clock } = createRenderer({
-    clockOptions: {
-      shouldAdvanceTime: true,
-    },
-  });
-
-  clock.withFakeTimers();
 
   describeConformance(<Select.Item value="" />, () => ({
     refInstanceof: window.HTMLDivElement,
@@ -208,9 +201,17 @@ describe('<Select.Item />', () => {
     });
   });
 
-  it.skipIf(!isJSDOM)(
-    'should not select an item on quick mouseup when showing a placeholder (no null item)',
-    async () => {
+  describe.skipIf(!isJSDOM)('quick selection', () => {
+    const { render: renderFakeTimers, clock } = createRenderer({
+      clockOptions: {
+        shouldAdvanceTime: true,
+      },
+    });
+
+    clock.withFakeTimers();
+
+    it('should not select an item on quick mouseup when showing a placeholder (no null item)', async () => {
+      ignoreActWarnings();
       const fonts = [
         { label: 'Sans-serif', value: 'sans' },
         { label: 'Serif', value: 'serif' },
@@ -253,14 +254,14 @@ describe('<Select.Item />', () => {
       fireEvent.mouseMove(option);
 
       // Release quickly over an unselected option.
-      clock.tick(200);
+      await clock.tickAsync(250);
       fireEvent.mouseUp(option);
 
       await waitFor(() => {
         expect(value.textContent).to.equal('Select font');
       });
-    },
-  );
+    });
+  });
 
   describe.skipIf(!isJSDOM)('style hooks', () => {
     it('should apply data-highlighted attribute when item is highlighted', async () => {
--- a/docs/src/app/(docs)/react/components/combobox/page.mdx
+++ b/docs/src/app/(docs)/react/components/combobox/page.mdx
@@ -72,6 +72,10 @@ import { Combobox } from '@base-ui/react/combobox';
 Combobox infers the item type from the `defaultValue` or `value` props passed to `<Combobox.Root>`.
 The type of items held in the `items` array must also match the `value` prop type passed to `<Combobox.Item>`.
 
+## Examples
+
+### Typed wrapper component
+
 The following example shows a typed wrapper around the Combobox component with correct type inference and type safety:
 
 ```tsx title="Specifying generic type parameters"
@@ -84,8 +88,6 @@ export function MyCombobox<Value, Multiple extends boolean | undefined = false>(
   return <Combobox.Root {...props}>{/* ... */}</Combobox.Root>;
 }
 ```
-
-## Examples
 
 ### Multiple select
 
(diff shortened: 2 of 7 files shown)


packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
```diff
@@ -607,5 +607,71 @@ describe('<Combobox.Trigger />', () => {
...
+            </Combobox.Positioner>
+          </Combobox.Portal>
+        </Combobox.Root>,
+      );
+
+      const trigger = screen.getByTestId('trigger');
+      expect(trigger).not.to.have.attribute('data-placeholder');
+    });
+
+    it('has data-placeholder when multiple mode has empty array', async () => {
```
mnajdova, line 678: Let's we add test that there is no data-placeholder when multiple has some default value.

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
```diff
@@ -277,6 +279,10 @@ export interface ComboboxTriggerState extends FieldRoot.State {
    * Present when the corresponding items list is empty.
    */
   listEmpty: boolean;
+  /**
+   * Whether the combobox doesn't have a value.
```
mnajdova, line 284:
```suggestion
   * Indicates that combobox doesn't have a value.
```

To read better.

mnajdova: I would rename all state booleans to use this format :D

packages/react/src/combobox/value/ComboboxValue.test.tsx
```diff
@@ -618,4 +618,188 @@ describe('<Combobox.Value />', () => {
...
+      await render(
+        <Combobox.Root items={items}>
+          <Combobox.Trigger data-testid="value">
+            <Combobox.Value placeholder="Select an option" />
+          </Combobox.Trigger>
+          <Combobox.Portal>
+            <Combobox.Positioner>
+              <Combobox.Popup>
+                <Combobox.List>
+                  <Combobox.Item value={null}>None</Combobox.Item>
```
mnajdova, line 846: Is this even valid? I mean having items array that difference than the Combobox Items rendered?

packages/react/src/select/item/SelectItem.test.tsx
```diff
@@ -193,6 +200,60 @@ describe('<Select.Item />', () => {
     });
   });
 
+  it.skipIf(!isJSDOM)(
+    'should not select an item on quick mouseup when showing a placeholder (no null item)',
```
mnajdova: Nice touch 👌

packages/react/src/select/value/SelectValue.tsx
```diff
@@ -22,35 +22,52 @@ export const SelectValue = React.forwardRef(function SelectValue(
...
+    ...elementProps
+  } = componentProps;
 
   const { store, valueRef } = useSelectRootContext();
 
   const value = useStore(store, selectors.value);
   const items = useStore(store, selectors.items);
   const itemToStringLabel = useStore(store, selectors.itemToStringLabel);
-  const serializedValue = useStore(store, selectors.serializedValue);
+  const hasSelectedValue = useStore(store, selectors.hasSelectedValue);
```
mnajdova, line 38: There is a lot of repetition here and in the combobox, does it makes sense to extract this in a usePlaceholder hook, or is it too much of an abstraction considering we use different stores? 😕

packages/react/src/utils/resolveValueLabel.tsx
```diff
@@ -27,10 +27,38 @@ export function isGroupedItems(
...
+ */
+export function hasNullItemLabel(items: ItemsInput): boolean {
+  if (!Array.isArray(items)) {
+    return items != null && !('null' in items);
+  }
+
+  if (isGroupedItems(items)) {
+    for (const group of items) {
+      for (const item of group.items) {
+        if (item && item.value == null && item.label != null) {
```
mnajdova, line 45: Would be good to add some tests with group items.

mnajdova (commented):
Logic looks good, left few comments/questions to be resolved before merging.

packages/react/src/combobox/value/ComboboxValue.test.tsx
atomiks: Technically yes, this pattern is currently half-broken but is specified throughout the tests. This PR gives it full support: https://github.com/mui/base-ui/pull/3594

packages/react/src/select/value/SelectValue.tsx
atomiks: I think "rule of three" for DRY works here and this is only 2 separate places

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
atomiks: The CSS vars use "Present when..." so I've changed it to match that

atomiks: Other `boolean` state attrs use `Whether` so I think that's correct for consistency. `Indicates` on the other hand is used when there is a value alongside the attribute. The CSS vars use "Present when..." for boolean attributes, but that's only because the attribute doesn't exist when `false`

PR #3606, opened 2025-12-24
[form controls] Fix visually hidden input styles

atomiks:

The `visuallyHidden` util isn't suitable for hidden inputs in form controls because native validation popups won't point to them (since the hidden input is at the top-left of the screen currently). They instead should be positioned where the control is. 

Affects:

- `NumberField`
- `Radio`
- `Checkbox`
- `Switch`
- `Select`
- `Combobox`

`Slider` already handled this correctly

---

The main issue when not adding `top`/`left` is stray overflow (since the visually hidden input takes up layout size) however this should not be a concern for form inputs specifically

### DIFF
--- a/packages/react/src/radio/root/RadioRoot.tsx
+++ b/packages/react/src/radio/root/RadioRoot.tsx
@@ -2,7 +2,7 @@
 import * as React from 'react';
 import { useMergedRefs } from '@base-ui/utils/useMergedRefs';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
-import { visuallyHidden } from '@base-ui/utils/visuallyHidden';
+import { visuallyHiddenInput } from '@base-ui/utils/visuallyHidden';
 import type { BaseUIComponentProps, NonNativeButtonProps } from '../../utils/types';
 import { createChangeEventDetails } from '../../utils/createBaseUIEventDetails';
 import { REASONS } from '../../utils/reasons';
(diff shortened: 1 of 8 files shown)


packages/react/src/radio/root/RadioRoot.tsx
```diff
@@ -131,58 +132,42 @@ export const RadioRoot = React.forwardRef(function RadioRoot(
...
-      disabled,
-      checked,
-      required,
-      readOnly,
-      onChange(event) {
-        // Workaround for https://github.com/facebook/react/issues/9023
-        if (event.nativeEvent.defaultPrevented) {
-          return;
-        }
+  const inputProps: React.ComponentPropsWithRef<'input'> = {
```
mnajdova, line 135: Good catch for removing the useMemo 👍

mnajdova (approved):
Example of fixed component (RadioGroup): https://stackblitz.com/edit/nrb7shnu?file=src%2FApp.tsx

Broken on master: https://stackblitz.com/edit/ub6e7bxf?file=src%2FApp.tsx

I noticed that the number field was already working, I am not sure why, e.g. check: https://stackblitz.com/edit/snm2yjeu?file=package.json, but in general using the new visually hidden styles for inputs makes sense.

atomiks:
@mnajdova ah it's because `NumberField` adds the `required` form validation prop on the visible input `<NumberField.Input>` as well. But, `min`/`max`/`step` validation popups wouldn't point correctly, since the visible input isn't `type="number"` and the validation props are only on the hidden input in that case.

Screenshot 2025-12-24 at 8 17 53 pm

mnajdova:
> @mnajdova ah it's because NumberField adds the required form validation prop on the visible input <NumberField.Input> as well. But, min/max/step validation popups wouldn't point correctly, since the visible input isn't type="number" and the validation props are only on the hidden input in that case.

Ah ok, makes sense 👌

PR #3607, opened 2025-12-24
[combobox] Fix controlled value prop when items change

atomiks:

The `value` gets cleared if the items change and the selected value is no longer in it; in controlled mode, this wasn't synchronized properly since it bypassed the `onValueChange` call

### DIFF
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -562,7 +562,7 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
   );
 
   const setSelectedValue = useStableCallback(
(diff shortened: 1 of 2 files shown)


packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -871,26 +870,17 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
     ) {
       fallback = defaultSelectedValue;
     }
-    setSelectedValueUnwrapped(fallback);
 
-    // Keep the input text in sync when the input is rendered outside the popup.
-    if (!store.state.inputInsidePopup) {
```
mnajdova, line 877: Note to myself, this logic for updating the input is already inside the `setSelectedValue`.

mnajdova (approved):

PR #3608, opened 2025-12-24
[select] Fix filled state in `multiple` mode

atomiks:

With `multiple`, the empty state is an empty array, not `null`.

Fixes https://github.com/mui/base-ui/issues/3610

### DIFF
--- a/packages/react/src/select/root/SelectRoot.test.tsx
+++ b/packages/react/src/select/root/SelectRoot.test.tsx
@@ -1526,6 +1526,44 @@ describe('<Select.Root />', () => {
 
         expect(trigger).not.to.have.attribute('data-filled');
       });
+
+      it('does not add [data-filled] attribute when multiple defaultValue is empty array', async () => {
+        const { user } = await renderFakeTimers(
+          <Field.Root>
+            <Select.Root multiple defaultValue={[]}>
+              <Select.Trigger data-testid="trigger" />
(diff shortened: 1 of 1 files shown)


packages/react/src/select/root/SelectRoot.test.tsx
```diff
@@ -1488,6 +1488,44 @@ describe('<Select.Root />', () => {
 
         expect(trigger).to.have.attribute('data-filled');
       });
+
+      it('does not add [data-filled] attribute when multiple value is empty', async () => {
+        const { user } = await renderFakeTimers(
+          <Field.Root>
+            <Select.Root multiple>
```
mnajdova, line 1495: Should we add test also when `defaultValue={[]}`?

packages/react/src/select/root/SelectRoot.tsx
```diff
@@ -194,8 +194,8 @@ export function SelectRoot<Value, Multiple extends boolean | undefined = false>(
   }, [store, value]);
 
   useIsoLayoutEffect(() => {
-    setFilled(value !== null);
-  }, [value, setFilled]);
+    setFilled(multiple ? Array.isArray(value) && value.length > 0 : value != null);
+  }, [multiple, value, setFilled]);
```
oliviertassinari: +1 for the move (we have added `undefined` as also an empty value).

PR #3609, opened 2025-12-24
[select][combobox] Do not trigger Field `onBlur` handlers when opening popup

atomiks:

`focused`/`touched` states/`onBlur` validation shouldn't run when e.g. clicking the trigger and focus moves into the popup. In that case `onBlur` runs, but it should really only run when blur leaves the trigger not as the result of opening the popup.

This also addresses `act()` warnings, since `enqueueFocus` in `FloatingFocusManager` is (intentionally) async: clicking the trigger causes focus to move into the popup async, which _then_ fires `onBlur`, causing the state updates to happen outside `act(...)`


atomiks:
@greptileai

Issue #3595, opened 2025-12-22
[context menu] Context menu closes when right-clicking a menu item on Windows


aeonzz:
# Bug report

## Current behavior

When a context menu is open, right-clicking on an item inside the context menu immediately closes the menu instead of allowing the item to handle the right-click interaction.

The menu closes as if an outside click occurred.

## Expected behavior

Right-clicking on a context menu item should not close the parent context menu by default.

The context menu should remain open and allow the item to handle the right-click event (e.g. opening a nested menu or triggering a custom action).

## Reproducible example

- Open a context menu

- Right-click on one of the context menu items

- The context menu closes immediately

## Base UI version

v1.0.0

## Which browser are you using?

Chrome

## Which OS are you using?

Windows

None

Voiceover/JAWS etc.

## Additional context

None

atomiks:
Does it reproduce on the docs website with Windows? 

It's not reproducible on macOS at least:

https://github.com/user-attachments/assets/c822a1e5-8208-40c3-8802-e0332d4d9a6f

---

There's also an experiment here with complex nested context menus where you can open a nested menu by right clicking on an item too (demo no. 3): https://base-ui.com/experiments/context-menu, which doesn't present the issue on Mac

aeonzz:
on the docs using windows

https://github.com/user-attachments/assets/41b1e535-bf0d-4d7b-baec-7f2e314bf752

mnajdova:
I debugged this on my Windows machine, the issue is that https://github.com/mui/base-ui/blob/master/packages/react/src/menu/item/useMenuItem.ts#L75 fires on Windows, but not on MacOS. Apperantly on MacOS the oeprating system intercepts the event to show the context menu.

I added a check at the beginning of the handler to ignore the code if it comes from right click, this way we can have coherent behavior between Windows and other OSs.

I opened https://github.com/mui/base-ui/pull/3612, let me just verify that this wouldn't break something else.

Issue #3224, opened 2025-11-14
[context menu] Popup closes on mouseup when `alignOffset` is less than 2


aarongarciah:
# Bug report

## Current behavior

The popup closes by itself on mouseup when `alignOffset` is less than `2`. Looks like https://github.com/mui/base-ui/issues/2595 wasn't fully solved.

## Expected behavior

The popup stays open so that I can select an option from the menu.

## Reproducible example

https://codesandbox.io/p/devbox/hungry-cherry-qvhghg?workspaceId=ws_TYgwg9BhnaH6MHvmXJ1vH9

## Base UI version

1.0.0-beta-4

## Which browser are you using?

Chrome v142, Firefox 145

## Which OS are you using?

Mac OS 15.6.1

PR #3612, opened 2025-12-25
[context-menu] Right click on menu item closes context menu on Windows

mnajdova:
Fixes https://github.com/mui/base-ui/issues/3595

The issue is that https://github.com/mui/base-ui/blob/master/packages/react/src/menu/item/useMenuItem.ts#L75 fires on Windows, but not on MacOS. Apperantly on MacOS the operating system intercepts the event to show the context menu.

I added a check at the beginning of the handler to ignore the code if it comes from right-click, this way we can have coherent behavior between Windows and other OSs.

----

There is another difference that I noticed between Windows and MacOS, on Windows the context menu opens on mouse up, while on MacOs it opens on mouse down.

### DIFF
--- a/packages/react/src/menu/item/useMenuItem.ts
+++ b/packages/react/src/menu/item/useMenuItem.ts
@@ -72,6 +72,13 @@ export function useMenuItem(params: useMenuItem.Parameters): useMenuItem.ReturnV
             }
           },
           onMouseUp(event) {
+            // On MacOS right click does not invoke onMouseUp, however on Windows it does
+            // This checks makes sure we are consistent behavior and ignore the logic below
+            // on right click
+            if (event.button === 2) {
+              return;
+            }
+
             if (contextMenuContext) {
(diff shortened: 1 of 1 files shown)


mnajdova:
@atomiks can you provide some reproduction for what the failing test is doing? See https://app.circleci.com/pipelines/github/mui/base-ui/18503/workflows/13724f2e-ee8b-435d-94da-e7b422ebdde0/jobs/194806 

I am asking because on MacOS, the mouse up does not even fire, while on Windows the context menu opens on mouse up, so I don't understand when will this code run. I can't reproduce https://github.com/mui/base-ui/issues/3224 on this PR or master, so I am not sure what should the test capture.

atomiks:
@mnajdova it's for right-click down -> drag -> right-click up (release), which macOS's native context menu supports

atomiks:
Replaced with https://github.com/mui/base-ui/pull/3944

PR #3634, opened 2025-12-27
[menu] Add `pressMode` option to the MenuTrigger

chuganzy:
This PR adds a `pressMode` option to `MenuTrigger`.

Currently, `Menu` opens on `mousedown`, which allows users to drag directly to select a menu item. This is a nice and intentional interaction design, however, there are cases where this behavior is not suitable.

https://codesandbox.io/p/sandbox/z2jwsl

For example, when the element wrapping the trigger is draggable, it can be necessary to distinguish between a drag gesture and a menu interaction. In such cases, opening the menu on `click` rather than `mousedown` provides more appropriate behavior.

There is a related discussion in Radix:
https://github.com/radix-ui/primitives/issues/2867

At the moment, this behavior can be implemented using `onOpenChange` / `preventBaseUIHandler`, but I thought it would be beneficial to support this use case directly in Base UI via an explicit option.

~Please note that this PR will conflict with #3631. I will address it as needed.~

### DIFF
--- a/docs/reference/generated/menu-trigger.json
+++ b/docs/reference/generated/menu-trigger.json
@@ -18,6 +18,11 @@
       "description": "A payload to pass to the menu when it is opened.",
       "detailedType": "Payload | undefined"
     },
+    "pressMode": {
+      "type": "'click' | 'mousedown'",
+      "description": "Determines how the menu is opened when the trigger is pressed.",
+      "detailedType": "'click' | 'mousedown' | undefined"
+    },
     "disabled": {
       "type": "boolean",
       "default": "false",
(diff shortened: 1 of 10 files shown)


mnajdova:
Can we for starter create a demo for this? To see how it looks in user-land and then decide if we want to have it as a feature directly in the component.

chuganzy:
@mnajdova Thanks for taking a look! Sure, I can add a demo later. In the meantime, the CodeSandbox I shared already includes this. Although it is very rough, please refer to the "AFTER" section to see how it looks.

https://codesandbox.io/p/sandbox/z2jwsl

chuganzy:
@mnajdova Added a demo here: https://github.com/mui/base-ui/pull/3634/commits/917626d8174e5a6b372b883e45ca41c04327c87c

https://deploy-preview-3634--base-ui.netlify.app/react/components/menu#press-mode

chuganzy:
Apparently I accidentally deleted the remote branch but did not mean to close this PR! Restored the branch. 
There are conflicts but leaving them for now.

PR #3638, opened 2025-12-28
[all components] Fix forwarded ref types

atomiks:

This loosens or restricts the type of the `ref` for some components

Some rules:
- Non-button `<div>` elements should not be polymorphic (as in rendered as another tag, not replaced with a same-tag component), so these have strict `HTMLDivElement`
- Button `<div>` elements (e.g. composite items) can be polymorphic (e.g. rendered as an anchor tag), so use `HTMLElement`
- `<Tooltip.Trigger>` could be a non-HTML element (e.g. using SVG elements), so I've replaced instances of `Element` with `HTMLElement` where applicable, except for `<Tooltip.Trigger>`, where I made `any` stricter

---

This doesn't solve `event.currentTarget` typesafety in event handlers, yet. Also, some button components which are unlikely to be polymorphic (but technically allow it via `nativeButton={false}`) still use strict `HTMLButtonElement` instead of `HTMLElement`. Changing them to be looser could be ok, if worth it.

### DIFF
--- a/packages/react/src/accordion/trigger/AccordionTrigger.tsx
+++ b/packages/react/src/accordion/trigger/AccordionTrigger.tsx
@@ -49,7 +49,7 @@ function getActiveTriggers(accordionItemRefs: { current: (HTMLElement | null)[] 
 
 export const AccordionTrigger = React.forwardRef(function AccordionTrigger(
   componentProps: AccordionTrigger.Props,
-  forwardedRef: React.ForwardedRef<Element>,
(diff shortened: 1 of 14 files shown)


michaldudak:
I'm wondering if we can/should enable non-HTML triggers for more components (think SVG buttons) as a follow-up (or if there's community demand). This would require adding a couple of conditions in focus management logic (as `focus` is not available in SVG elements).

PR #3640, opened 2025-12-29
[form controls] Fix hidden input `id` and `required` props

### LINKED ISSUE #3263: Radio hover css behaves differently in Beta 5/6
# Bug report

## Current behavior

If you do a css selector like `.radio:hover`, beta 4 and below would apply the styles when you hover the `label` associated with the radio element. In beta 6, this is not the case.

## Expected behavior

It would be nice if the beta 4 behavior could be brought back.

## Reproducible example
Shows the two versions side-by-side. A workaround is added to App.css and commented out
https://stackblitz.com/edit/vitejs-vite-bwvxcdt3?file=src%2FApp.tsx,src%2FApp.css

## Base UI version

1.0.0-beta.6

Chrome and Safari

## Which OS are you using?

macOS

## Additional context

This also applies to `Checkbox`, and other css states like `:active`.

I suspect this might just be an expected outcome of https://github.com/mui/base-ui/pull/3205. 

My workaround is using a css selector like `.radio:hover, .radio:has(~ label:hover)`

atomiks:

1. In Switch, Checkbox, and Radio, when rendering a `<button>` element, the `id` will point to the `<button>` control instead of the hidden `<input>`, matching Radix and native behavior. Enclosing labels in this format creates invalid HTML, but sibling labels will work correctly by default. Closes #3263

2. In Autocomplete, the `required` prop is also forwarded to the visible `Input` part, matching Number Field (but does not apply to Combobox). Also adds `aria-required` for missing controls (as is currently applied to Checkbox/Radio)

### DIFF
--- a/packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
+++ b/packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
@@ -466,6 +466,29 @@ describe('<Combobox.Trigger />', () => {
   });
 
   describe('aria attributes', () => {
+    it('sets aria-required attribute when required (input inside popup)', async () => {
+      await render(
+        <Combobox.Root required>
+          <Combobox.Trigger data-testid="trigger" />
+        </Combobox.Root>,
+      );
+
+      const trigger = screen.getByTestId('trigger');
+      expect(trigger).to.have.attribute('aria-required', 'true');
+    });
+
+    it('does not set aria-required attribute when the input is outside the popup', async () => {
+      await render(
+        <Combobox.Root required>
+          <Combobox.Input />
+          <Combobox.Trigger data-testid="trigger" />
+        </Combobox.Root>,
+      );
+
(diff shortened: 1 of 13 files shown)


packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
```diff
@@ -466,6 +466,29 @@ describe('<Combobox.Trigger />', () => {
...
+        <Combobox.Root required>
+          <Combobox.Trigger data-testid="trigger" />
+        </Combobox.Root>,
+      );
+
+      const trigger = screen.getByTestId('trigger');
+      expect(trigger).to.have.attribute('aria-required', 'true');
+    });
+
+    it('does not set aria-required attribute when not the combobox element', async () => {
```
mj12albert:
```suggestion
    it('does not set aria-required attribute when the input is not inside the popup', async () => {
```
Is this what it was supposed to mean?

PR #3642, opened 2025-12-29
[mergeProps] Make `mergeProps` public

michaldudak:
`mergeProps` was already exported from the package, but not documented. As it's useful (shadcn components use it, for example), I added the docs page for it. 
The docs were generated using an updated infrastructure, so the API definitions are taken from JSDocs (as in components) and not typed by hand.

Docs: https://deploy-preview-3642--base-ui.netlify.app/react/utils/merge-props


docs/src/app/(docs)/react/utils/merge-props/page.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+This function accepts up to 5 arguments, each being either a props object or a function that returns a props object.
+If you need to merge more than 5 sets of props, use `mergePropsN` instead.
+
+<Reference name="mergeProps" />
+
+### mergePropsN
+
+This function accepts an array of props objects or functions that return props objects.
+It is slightly less efficient than `mergeProps` so only use it when you need to merge more than 5 sets of props.
```
LukasTy:
```suggestion
It is slightly less efficient than `mergeProps`, so only use it when you need to merge more than 5 sets of props.
```

docs/src/app/(docs)/react/utils/merge-props/page.mdx
```diff
@@ -0,0 +1,57 @@
+# mergeProps
+
+<Subtitle>A utility to merge multiple sets of React props.</Subtitle>
+<Meta
+  name="description"
+  content="A utility to merge multiple sets of React props, handling event handlers, className, and style props intelligently."
+/>
+
+`mergeProps` helps you combine multiple prop objects (for example internal props + user props) into a single set of props you can spread onto an element.
```
LukasTy:
```suggestion
`mergeProps` helps you combine multiple prop objects (for example, internal props + user props) into a single set of props you can spread onto an element.
```

docs/src/app/(docs)/react/utils/merge-props/page.mdx
```diff
@@ -0,0 +1,57 @@
...
+It behaves like `Object.assign` (rightmost wins) with a few special cases so common React patterns work as expected.
```
LukasTy:
```suggestion
It behaves like `Object.assign` (rightmost wins) with a few special cases, so common React patterns work as expected.
```

LukasTy (approved):
LGTM. 👍 
Thanks for addressing this. 🙏

docs/src/components/ReferenceTable/ParametersReferenceTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export async function ParametersReferenceTable({
+  data,
+  name,
+  renameFrom,
+  renameTo,
+  ...props
+}: ParametersReferenceTableProps) {
+  return (
+    <PropsReferenceAccordion
```
dav-is: Would it make sense to abstract this into a `<ReferenceAccordion/>` component which both the parameters and props table would use?

docs/src/app/(docs)/react/utils/merge-props/page.mdx
```diff
@@ -0,0 +1,57 @@
...
+`mergeProps` helps you combine multiple prop objects (for example, internal props + user props) into a single set of props you can spread onto an element.
+It behaves like `Object.assign` (rightmost wins) with a few special cases, so common React patterns work as expected.
+
+## How merging works
+
+- For most keys (everything except `className`, `style`, and event handlers), the value from the rightmost object wins:
+  ```ts
+  mergeProps({ id: 'a', dir: 'ltr' }, { id: 'b' }); // returns { id: 'b', dir: 'ltr' }
+  ```
+  Note that this includes refs.
```
flaviendelangle: I don't think this is the behavior most people would expect without diving deeper into why we don't merge the refs.
So I think it's worth being extra-explicit here and add a bullet point specifically for the `ref` 👍

docs/src/app/(docs)/react/utils/merge-props/page.mdx
```diff
@@ -0,0 +1,57 @@
...
+    This does not call `preventDefault()` or `stopPropagation()`.
+    It only tells Base UI to skip earlier merged handlers or internal logic that checks the flag.
+  - For non-synthetic events (custom events with primitive/object values), this mechanism isn't available and all handlers always execute.
+
+## Passing a function instead of an object
+
+Each argument can be a props object or a function that receives the merged props up to that point (left to right) and returns a props object.
+This is useful when you need to compute the next props from whatever has already been merged.
+
+When you return event handlers from a function argument, `preventBaseUIHandler()` is not applied automatically to those handlers.
```
flaviendelangle: A couple of code snippets on how to use the `baseUIHandlerPrevented` and `preventBaseUIHandler()` could be nice (with an object and with a function)

docs/src/app/(docs)/react/utils/merge-props/page.mdx
```diff
@@ -0,0 +1,57 @@
...
+- Event handlers are merged and executed right-to-left (rightmost first):
+
+  ```ts
+  mergeProps({ onClick: a }, { onClick: b }); // b runs before a
+  ```
+
+  - For React synthetic events, Base UI adds `event.preventBaseUIHandler()`. Calling it sets `event.baseUIHandlerPrevented = true`.
+    This does not call `preventDefault()` or `stopPropagation()`.
+    It only tells Base UI to skip earlier merged handlers or internal logic that checks the flag.
+  - For non-synthetic events (custom events with primitive/object values), this mechanism isn't available and all handlers always execute.
```
atomiks: There's extra indentation here

docs/src/app/(docs)/react/utils/merge-props/page.mdx
```diff
@@ -0,0 +1,57 @@
...
+- `className` values are concatenated right-to-left (rightmost first):
+  ```ts
+  mergeProps({ className: 'a' }, { className: 'b' }); // className is 'b a'
+  ```
+- `style` objects are merged, with keys from the rightmost style overwriting earlier ones.
+- Event handlers are merged and executed right-to-left (rightmost first):
+
+  ```ts
+  mergeProps({ onClick: a }, { onClick: b }); // b runs before a
+  ```
```
atomiks: The code blocks are missing a title, maybe the comments could be moved to the title

michaldudak:
I addressed all the comments.

flaviendelangle (approved):

PR #3643, opened 2025-12-30
[button] Remove discriminated props union

### LINKED ISSUE #3641: Cannot set `Button` `type` prop
# Bug report

## Current behavior

I cannot set `type` prop for a `Button` component. 

```tsx
<Button type="submit" {...props}>
     {children}
</Button>
```

```cli
error TS2322: Type '{ children: string | number | bigint | boolean | ReactElement<unknown, string | JSXElementConstructor<any>> | Iterable<ReactNode> | ReactPortal | Promise<...> | null; ... 288 more ...; ref?: Ref<...> | undefined; } | { ...; }' is not assignable to type 'IntrinsicAttributes & (ButtonProps & RefAttributes<HTMLElement>)'.
  Type '{ children: string | number | bigint | boolean | ReactElement<unknown, string | JSXElementConstructor<any>> | Iterable<ReactNode> | ReactPortal | Promise<...> | null; ... 280 more ...; type: "submit"; }' is not assignable to type 'IntrinsicAttributes & (ButtonProps & RefAttributes<HTMLElement>)'.
    Property 'type' does not exist on type 'IntrinsicAttributes & ButtonNonNativeProps & RefAttributes<HTMLElement>'.

37       <Button type="submit" {...props}>
                 ~~~~
``` 

## Expected behavior

I should be able to set it:

Image

## Reproducible example

I am not sure how to provide an example.

## Base UI version

v1.0.0

N/A (it's happening during type checking)

## Which OS are you using?

Mac OS

None

## Additional context

I was trying the Shadcn create, but I get the same error when I am using the "native" BaseUI `Button`.

atomiks:

Trying to use `Omit`, `& {...}`, `extends` etc. breaks the `Button.Props` type. All other button components (e.g. `Popover.Trigger`) do not yet handle this either. It can be solved with overloads, but this also requires wrappers to handle overloads as well to retain the typesafety. The types for all button components can be changed in a separate PR.

Fixes #3641


docs/reference/generated/button.json
```diff
@@ -10,11 +10,8 @@
     },
     "nativeButton": {
       "type": "boolean",
-      "detailedType": "boolean | undefined"
-    },
-    "disabled": {
```
michaldudak, line 15: Why is this removed?

packages/react/src/button/Button.tsx
michaldudak, line 1: We could add dev-time warnings if native button props are used alongside `nativeButton={false}`

docs/reference/generated/button.json
atomiks: The `disabled` prop is already a native prop part of `HTMLButtonProps`, so it's redundant here (at least with the current solution)

packages/react/src/button/Button.tsx
atomiks: Runtime warnings are better than nothing sure, but this applies to every button component, so it's probably better done in a separate PR to make it cohesive across the library

atomiks:
@LukasTy that definitely looks promising~ I think it's better to change this for all `<button>` components simultaneously though, in a separate PR

LukasTy:
Sounds nice. 👍 
I take it if the separate PR works out, this one won't be needed anymore?

atomiks:
I think it's worth merging this in the meantime to fix the linked issue (and make it consistent with other components' props currently)

PR #3644, opened 2025-12-30
[docs-infra] Apply Docs Specific Prettier Rules

dav-is:
I noticed that the docs specific prettier overrides are not being applied.

Today, this just means setting `printWidth: 85` (the rest of the repo has `printWidth: 100`): 
https://github.com/mui/mui-public/blob/dbef881bae7f3202ea876354b83af83f8970559a/packages/code-infra/src/prettier.mjs#L8-L22

### DIFF
--- a/docs/src/app/(private)/experiments/dialog/nested.tsx
+++ b/docs/src/app/(private)/experiments/dialog/nested.tsx
@@ -142,7 +142,7 @@ function ReactSpringDialogDemo({
   keepMounted,
   modal,
   disablePointerDismissal,
-  // @ts-expect-error To be used later]
+  // @ts-expect-error To be used later
 }: DemoProps) {
   const [open, setOpen] = React.useState(false);
 


prettier.config.mjs
```diff
@@ -1,3 +1,3 @@
-import { createBaseConfig } from '@mui/internal-code-infra/prettier';
+import { createBaseConfig, docsOverrides } from '@mui/internal-code-infra/prettier';
 
-export default createBaseConfig();
+export default createBaseConfig({ overrides: [...docsOverrides] });
```
dav-is, line 3: This is the change to the config

colmtuite:
I'm not sure what I should check here? All I notice is that in the Accordion demo, text wraps too early, which is a downgrade from prod.

Screenshot 2025-12-30 at 11 46 24

Screenshot 2025-12-30 at 11 46 11

dav-is:
> This setting is probably only relevant to the Material docs with their styles

Ok, no need to merge this then

PR #3647, opened 2025-12-31
[combobox] Add `toolbar` role to `<Combobox.Chips>` part

### LINKED ISSUE #3629: [combobox] Combobox keyboard navigation not working while screen reader is active
# Bug report

## Current behavior

When using a multiple select combobox with chips, keyboard navigation fails when a screen reader is active:
- Pressing the **left arrow key** while focused on the input moves focus to the last chip
- Pressing the **right arrow key** from a chip keeps focus on the chip instead of returning to the input field

This issue only occurs with screen readers enabled.

## Expected behavior

Arrow keys should provide consistent bidirectional navigation between chips and the input field, allowing users to:
- Move left from the input to navigate through chips
- Move right from chips to return focus to the input
- Navigate seamlessly without getting stuck on any element

This navigation pattern works correctly when screen readers are disabled.

## Reproducible example

[Combobox example](https://stackblitz.com/edit/vitejs-vite-hc1ch6qf)
Copied from https://base-ui.com/react/components/combobox#multiple-select

## Base UI version

"@base-ui/react": "^1.0.0"

Chrome

## Which OS are you using?

Windows

NVDA

## Additional context

The issue appears to be related to screen reader mode switching. NVDA may be conflicting between browse mode and focus mode when navigating between the interactive elements, causing the unexpected keyboard behavior.

atomiks:

This prevents NVDA from entering "browse" mode when navigating chips with arrow keys when it should stay in "focus" mode (NVDA otherwise swallows arrow keys and the user gets stuck, needing to press <kbd>Tab</kbd>). It is also more semantically correct as `<Combobox.Chips>` acts as a toolbar role wise when at least 1 chip is present.

I tested manually on Windows + NVDA

The ARIA guide says toolbars should only contain 3+ items (that means `Input` + at least two chips) however this is likely not a strict/hard rule, because it doesn't break behavior

Fixes #3629

### DIFF
--- a/packages/react/src/combobox/chips/ComboboxChips.tsx
+++ b/packages/react/src/combobox/chips/ComboboxChips.tsx
@@ -7,6 +7,7 @@ import { ComboboxChipsContext } from './ComboboxChipsContext';
 import { CompositeList } from '../../composite/list/CompositeList';
 import { useComboboxRootContext } from '../root/ComboboxRootContext';
 import { selectors } from '../store';
+import { EMPTY_OBJECT } from '../../utils/constants';
 
 /**
  * A container for the chips in a multiselectable input.
@@ -21,6 +22,7 @@ export const ComboboxChips = React.forwardRef(function ComboboxChips(
   const store = useComboboxRootContext();
 
   const open = useStore(store, selectors.open);
+  const hasSelectionChips = useStore(store, selectors.hasSelectionChips);
 
   const [highlightedChipIndex, setHighlightedChipIndex] = React.useState<number | undefined>(
     undefined,
@@ -34,7 +36,9 @@ export const ComboboxChips = React.forwardRef(function ComboboxChips(
 
   const element = useRenderElement('div', componentProps, {
     ref: [forwardedRef, store.state.chipsContainerRef],
-    props: elementProps,
+    // NVDA enters browse mode instead of staying in focus mode when navigating with
+    // arrow keys inside a container unless it has a toolbar role.
+    props: [hasSelectionChips ? { role: 'toolbar' } : EMPTY_OBJECT, elementProps],
   });
 
   const contextValue: ComboboxChipsContext = React.useMemo(
--- a/packages/react/src/combobox/chips/ComboboxChips.test.tsx
+++ b/packages/react/src/combobox/chips/ComboboxChips.test.tsx
@@ -1,5 +1,7 @@
 import { Combobox } from '@base-ui/react/combobox';
+import { screen } from '@mui/internal-test-utils';
 import { createRenderer, describeConformance } from '#test-utils';
+import { expect } from 'chai';
 
 describe('<Combobox.Chips />', () => {
   const { render } = createRenderer();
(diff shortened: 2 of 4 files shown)


packages/react/src/combobox/chips/ComboboxChips.tsx
```diff
@@ -21,6 +21,9 @@ export const ComboboxChips = React.forwardRef(function ComboboxChips(
   const store = useComboboxRootContext();
 
   const open = useStore(store, selectors.open);
+  const selectedValue = useStore(store, selectors.selectedValue);
```
flaviendelangle: Nitpick but since `selectedValue` is never used outside of `hasSelectionChips`, you could create a new selector that returns `hasSelectionChips` directly and do not re-render when selecting 1, then 2, then 3 items etc...

colmtuite:
> a collection of commonly used function buttons or controls
> A toolbar is a collection of commonly used controls, such as buttons or checkboxes
> The toolbar is commonly a subset of functions found in a menubar as a way to reduce user effort.

This strikes me as misuse of the toolbar role.

atomiks:
> This strikes me as misuse of the toolbar role.

Can you explain? I don't see how `toolbar` doesn't describe a row of chip buttons/controls that can be navigated via arrow keys? The toolbar includes the `Input` in the bar too, so it's not only one single type of control.

> a collection of commonly used function buttons or controls

It's a collection of controls

> A toolbar is a collection of commonly used controls, such as buttons or checkboxes

This is a bespoke pattern that isn't shown anywhere in ARIA guides afaik, but they can be thought of as buttons

> The toolbar is commonly a subset of functions found in a menubar as a way to reduce user effort.

I think by user effort, they're referring to creating one single tab stop which is exactly what this does

---

The other option is `role=application` but this is more controversial than `role=toolbar` to some degree but it could work here (scoped tightly, all keyboard interactions handled)

colmtuite:
Isn't `role="group"` an option?

colmtuite:
My understanding (and this is all just interpretation, since the guides suck, as usual) is that the toolbar role is primarily intended for:

- UIs where the toolbar is kinda the top-level widget. Like a rich-text editor toolbar, or a toolbar at the top of a screen, or a toolbar at the top of a table/grid.
- Peer controls, where each control performs actions on the same thing. Like rich text editor toolbar controls perform actions on the text area content. This combobox toolbar sometimes performs actions on items inside the toolbar, and sometimes on items outside the toolbar. So they're not peer controls.
- Controls that perform actions on external elements, elements that are not its own value. Like the chip button performs an action on itself. Imo, toolbar controls are supposed to perform actions of other things.

My understanding is that toolbar is almost like a landmark. Obviously, I'm expanding on the ARIA guides a huge amount here, but that is my interpretation.

atomiks:
> Isn't role="group" an option?

NVDA still enters "browse" mode with that role (I tried)

> UIs where the toolbar is kinda the top-level widget. Like a rich-text editor toolbar, or a toolbar at the top of a screen, or a toolbar at the top of a table/grid.

This feels overly restrictive to me. A toolbar is any bar of tools, and it shouldn't matter where it is?

> Peer controls, where each control performs actions on the same thing. [...] Controls that perform actions on external elements,

Each tool does remove itself, though it also performs actions (conceptually) on the combobox listbox to remove them from the list. So each chip is performing actions on the same "external thing" (the listbox options), it just happens that the chips also remove themselves upon doing so - which is probably the biggest question mark... if a toolbar control can never remove itself, `role=application` might be the only sound role for this widget.

PR #3650, opened 2025-12-31
[interactive elements] remove duplicated disabled prop

### LINKED ISSUE #3649: `disabled` attribute applied inappropriately.
# Bug report

## Current behavior

When applying the ₩disabled₩ attribute to elements used as buttons, such as `XXX.Trigger` or `XXX.Close`, both the `disabled` attribute and the `data-disabled` attribute are applied in the HTML.

These elements can be rendered as non-button elements using prop like `render={<span />}`. However, the disabled attribute remains in this case. According to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled#overview), the disabled attribute is not applied to span tags. Therefore, the aria-disabled attribute should be applied instead.

## Expected behavior

- When an element is rendered as a `button` tag, apply the `data-disabled` and `disabled` attributes.
- When an element is rendered as a tag other than a `button`, apply the `data-disabled` and `aria-disabled` attributes.

## Reproducible example

https://stackblitz.com/edit/kn1mbuqo?file=src%2FApp.tsx

## Base UI version

`1.0.0`

Chrome

## Which OS are you using?

Mac OS

x

## Additional context

x

seongminn:

`getButtonProps` selects and applies either `disabled` or `aria-disabled` appropriately. But the `disabled` attribute was always being added separately, meaning it was always present regardless.

Fixes #3649

### DIFF
--- a/packages/react/src/number-field/increment/NumberFieldIncrement.tsx
+++ b/packages/react/src/number-field/increment/NumberFieldIncrement.tsx
@@ -104,8 +104,7 @@ export const NumberFieldIncrement = React.forwardRef(function NumberFieldIncreme
 export interface NumberFieldIncrementState extends NumberFieldRoot.State {}
 
 export interface NumberFieldIncrementProps
-  extends NativeButtonProps,
-    BaseUIComponentProps<'button', NumberFieldIncrement.State> {}
+  extends NativeButtonProps, BaseUIComponentProps<'button', NumberFieldIncrement.State> {}
 
 export namespace NumberFieldIncrement {
   export type State = NumberFieldIncrementState;
--- a/packages/react/src/number-field/decrement/NumberFieldDecrement.tsx
+++ b/packages/react/src/number-field/decrement/NumberFieldDecrement.tsx
@@ -104,8 +104,7 @@ export const NumberFieldDecrement = React.forwardRef(function NumberFieldDecreme
 export interface NumberFieldDecrementState extends NumberFieldRoot.State {}
 
 export interface NumberFieldDecrementProps
-  extends NativeButtonProps,
-    BaseUIComponentProps<'button', NumberFieldDecrement.State> {}
+  extends NativeButtonProps, BaseUIComponentProps<'button', NumberFieldDecrement.State> {}
 
 export namespace NumberFieldDecrement {
   export type State = NumberFieldDecrementState;


seongminn:
When `focusableWhenDisabled` is true, `aria-disabled` should be applied instead of `disabled`. Accordingly, I modified the test code for `ComboboxChipRemove`.

Additionally, when the `disabled` attribute was added to `ComboboxChipRemove`, it was not composing with the `disabled` attribute passed to `ComboboxRoot`. So I fixed that as well.

packages/react/src/number-field/increment/NumberFieldIncrement.tsx
```diff
@@ -63,7 +63,7 @@ export const NumberFieldIncrement = React.forwardRef(function NumberFieldIncreme
     maxWithDefault,
     value,
     inputValue,
-    disabled,
+    disabled: composedDisabled,
```
atomiks: Thanks for the fixes - what is this particular file (and Decrement) changed for?

seongminn: As with other components, it appropriately merges the `disabled` passed from the Root component and the `disabled` passed from the Increment component.

And unlike other components, NumberField recalculates `disabled` within `useNumberFieldButton`. Therefore, it separately creates `composedDisabled`. If declaring this variable is unnecessary, directly assigning it as shown below is also valid.

```ts
const { disabled = false, ...props } = useNumberFieldButton({
    // ....
    disabled: disabledProp || contextDisabled,
});
```

atomiks: Ah I see that you're checking the result of the internal calculation:

```jsx
disabled: disabled || (isIncrement ? isMax : isMin),
```

I would do this calculation before passing to `useNumberFieldButton`, that way it's composed of all the states and `NumberFieldIncrement.State` has the correct `disabled` state, without needing the destructure props collection?

Furthermore, I realized that the Number Field buttons should use `focusableWhenDisabled: true` in `useButton` so focus isn't lost when it reaches the min/max value.

seongminn: Before passing it to the `useNumberFieldButton` hook, completing the calculation of the disabled property would look like this:

```ts
const composedDisabled = disabledProp || contextDisabled;
const isMax = value != null && value >= maxWithDefault;

```ts
const { disabled = false, ...props } = useNumberFieldButton({
    // ...
    disabled: composedDisabled || isMax,
});
```

However, doing this might allow the `isDisabled` value to change when calculated inside `useNumberFieldButton`.

```ts
// current
const isDisabled = disabled || readOnly || (isIncrement ? isMax : isMin);

// changed
const isDisabled = disabled || readOnly;
```

But I don't think this difference is particularly significant. So, I think it would be okay to change the internal logic of the `useNumberFieldButton` hook. What do you think?

atomiks: Yes, it's better as it reduces the repetition in the event handlers

It should look like this:
```jsx
const isMax = value != null && value >= maxWithDefault;
const disabled = disabledProp || contextDisabled || isMax;

const props = useNumberFieldButton({
    disabled,
});
```

Then internally, 

```jsx
const isDisabled = disabled || readOnly;
```

Since `disabled` has the min/max condition

seongminn: Thank you for your feedback. I did it [here](https://github.com/mui/base-ui/pull/3650/commits/33a3bf8a322c5f55f99df42b7960615ccdab0c1f)

seongminn:
I did it [here](https://github.com/mui/base-ui/pull/3650/commits/0edf41ebf98e322395d3063e32ea10ec2fe7c068)!

PR #3652, opened 2025-12-31
[tooltip][popover] Refactor popup auto resize logic

atomiks:
- Moves `usePopupAutoResize()` into `Viewport` part for treeshaking when it's not being used.
- Ensures `translate()` positioning is used instead of `top`/`left` if the `Viewport` part is missing. This should solve things like this https://x.com/sajadevo_/status/2004063902636945842 (I have no way to verify though) because it rounds to the device's subpixel grid instead of being totally unrounded. This also ensures you can use detached triggers without needing to specify `--popup-width/--positioner-width` etc. variables as it's explicitly opt-in.

### DIFF
--- a/packages/react/src/popover/viewport/PopoverViewport.tsx
+++ b/packages/react/src/popover/viewport/PopoverViewport.tsx
@@ -6,12 +6,15 @@ import { usePreviousValue } from '@base-ui/utils/usePreviousValue';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { usePopoverRootContext } from '../root/PopoverRootContext';
+import { usePopoverPositionerContext } from '../positioner/PopoverPositionerContext';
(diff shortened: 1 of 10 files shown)


packages/react/src/popover/viewport/PopoverViewport.tsx
```diff
@@ -90,31 +105,18 @@ export const PopoverViewport = React.forwardRef(function PopoverViewport(
...
-    nextDimensions: Dimensions;
-  };
-
-  const handleMeasureLayoutComplete = useStableCallback((data: MeasureLayoutCompleteData) => {
-    currentContainerRef.current?.style.removeProperty('animation');
-    currentContainerRef.current?.style.removeProperty('transition');
-
-    previousContainerRef.current?.style.removeProperty('display');
+  const handleMeasureLayoutComplete = useStableCallback(
+    (previousDimensions: Dimensions | null, _nextDimensions: Dimensions) => {
```
michaldudak: _nextDimensions seems unused

PR #3654, opened 2026-01-01
[menu][popover][combobox] Fix focus guard handling

atomiks:

Fixes a few issues with the focus guards:

- Tabbing out of a Popover gets stuck on the focus guard in Safari  (or moving virtual cursor out of a Popover on iOS)
- Tabbing out of a Menu inside a Popover incorrectly restores focus to the popover container instead of the next tabbable
- Tabbing out of a Combobox input whose popover is open should close the popup when focus lands on the Popover trigger (when it's the first tabbable element or the last one, when tabbing forward)

https://deploy-preview-3654--base-ui.netlify.app/experiments/popup-tabbing

### DIFF
--- a/packages/react/src/menu/trigger/MenuTrigger.tsx
+++ b/packages/react/src/menu/trigger/MenuTrigger.tsx
@@ -207,7 +207,7 @@ export const MenuTrigger = React.forwardRef(function MenuTrigger(
   });
 
   const focus = useFocus(floatingRootContext, {
-    enabled: !disabled && ((!isInMenubar && isOpenedByThisTrigger) || parentMenubarHasSubmenuOpen),
+    enabled: !disabled && parentMenubarHasSubmenuOpen,
   });
 
   const mixedToggleHandlers = useMixedToggleClickHandler({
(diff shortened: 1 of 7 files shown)


packages/react/src/menu/trigger/MenuTrigger.tsx
```diff
@@ -292,7 +292,9 @@ export const MenuTrigger = React.forwardRef(function MenuTrigger(
         );
       });
 
-      let nextTabbable = getTabbableAfterElement(triggerElementRef.current);
+      let nextTabbable = getTabbableAfterElement(
+        store.context.triggerFocusTargetRef.current || triggerElementRef.current,
+      );
```
atomiks, line 297: In Safari, `aria-hidden={false}` on the focus guard, so it should look for next tabbable starting from (after) the focus guard

packages/react/src/menu/trigger/MenuTrigger.tsx
```diff
@@ -207,7 +207,7 @@ export const MenuTrigger = React.forwardRef(function MenuTrigger(
   });
 
   const focus = useFocus(floatingRootContext, {
-    enabled: !disabled && ((!isInMenubar && isOpenedByThisTrigger) || parentMenubarHasSubmenuOpen),
+    enabled: !disabled && parentMenubarHasSubmenuOpen,
```
atomiks, line 210: This otherwise causes  <kbd>Shift + Tab</kbd> not to close the popup when focus lands back on the trigger in Menu; seems to still pass all Menubar tests

atomiks:
@LukasTy can't reproduce anymore after rebasing with main 🤔 

I [noticed a bug with React hydration](https://github.com/vercel/next.js/issues/84029#issuecomment-3565082492) would break the behavior in dev mode though, when the hydration error appeared in the console

LukasTy:
@atomiks I can confirm that it works as expected in Firefox on https://master--base-ui.netlify.app/experiments/popup-tabbing.
However, Safari's behavior still seems somewhat strange, but I always found its focus management especially strange. 🙈

PR #3675, opened 2026-01-04
Update docv1 by softwaredev509

softwaredev509:
## Description
This PR improves the documentation for the `Popover` component to make it more compatible with LLMs (Large Language Models) like Claude or ChatGPT, as well as more informative for human users.

## Problem
Currently, the documentation relies heavily on visual demos, which leads to generated content (like `llms.txt`) that lacks semantic context. This causes LLMs to hallucinate imports or usage patterns (e.g., confusing `base-ui` with `base-ui-components`) because they don't "see" the implementation details hidden in demos.

## Changes
- Added a clear **Description** and **Use Cases** section to [popover/page.mdx]
- Added an **Anatomy** section with text descriptions for each sub-component (`Root`, `Trigger`, etc.).
- Added an **Accessibility** section detailing keyboard interactions and ARIA roles.

These changes will automatically populate the generated `llms.txt` with rich context, explaining the component's purpose and structure to AI assistants.

### DIFF
--- a/docs/src/app/(docs)/react/components/popover/page.mdx
+++ b/docs/src/app/(docs)/react/components/popover/page.mdx
@@ -18,9 +18,9 @@ Unlike a [Tooltip](/react/components/tooltip), which is used for brief helper te
 
 ## Use cases
 
--   **User Profiles**: Displaying user details and actions in a dropdown.
--   **Filters**: Showing complex filter options for a data table.
--   **Feature Tours**: Providing detailed context about a new feature.
+- **User Profiles**: Displaying user details and actions in a dropdown.
+- **Filters**: Showing complex filter options for a data table.
+- **Feature Tours**: Providing detailed context about a new feature.
 
 ## Anatomy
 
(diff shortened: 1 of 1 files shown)


colmtuite:
Hey, thanks for attempting to improve the docs. There are very many ways we could improve the docs, so I appreciate the efforts. However, I'm closing this PR for a few reasons:

1. The parts are already described under each heading in the API Reference. Some of this is word-for-word duplicate content.
2. We don't document markup or keyboard interactions, since that is handled by Base UI for the user, and they don't need to implement anything there. Therefore they don't need implementation guidance.
3. If we were going to add new docs content similar to this, we would do it across all components simultaneously, for a cohesive experience.

> confusing base-ui with base-ui-components

None of these changes will help with this. I recommend you point llms to the[ /releases page](https://base-ui.com/react/overview/releases#v1-0-0), for new projects, and ask it to use the latest conventions and packages.

> Use Cases section

This is one is a good idea imo. However, I think images are probably necessary. And I think the content here is more useful if it compares the component to other similar components, and explains when to use one over the other.

PR #3680, opened 2026-01-05
[drawer] Create new Drawer / Sheet component

### LINKED ISSUE #38: [drawer] Create the unstyled Drawer / Sheet component
This is the place to gather ideas, measure interest and discuss the API and implementation details of the drawer component.

## User requests

- https://www.reddit.com/r/reactjs/comments/1pk18v9/comment/ntqbuf0/
  > I beg you please copy Vaul, make your own swipe-able drawer with handle and snap points and make sure it's smooth on mobile because Vaul is practically deprecated
- Not a direct request, but people are looking for a way to move forward with Vaul:
  - https://github.com/shadcn-ui/ui/issues/9191
  - https://github.com/shadcn-ui/ui/issues/8455
  - https://github.com/emilkowalski/vaul/issues/630

## Benchmarks

- View only: https://mui-org.notion.site/core-Drawer-2ebcbfe7b66080a68150cdc9e0b0ac20?source=copy_link
- Edit: https://www.notion.so/mui-org/core-Drawer-2ebcbfe7b66080a68150cdc9e0b0ac20

atomiks:
Closes #38

- [x] Nested drawer demo/logic/styling
- [x] Background indent provider
- [x] Handle text selection with swiping interaction
- [x] Fix momentum when swipe dismissing
- [x] Snap points

Preview: https://deploy-preview-3680--base-ui.netlify.app/react/components/drawer


piehouserat:
any idea when this will ship? I need a drawer component fairly soon but I can always use vaul in the meantime

atomiks:
@piehouserat probably in v1.2.0 release around mid-February

philwolstenholme:
A neat feature would be to use a [CloseWatcher](https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher) to allow closing the drawer with the Android 'back' swipe gesture. This can make drawers and modals feel much more native. The browser support is Chromium only at the moment (I think?) but I usually feature detect it and fall back to an `Esc` key listener instead on unsupported browsers.

packages/react/src/drawer/popup/DrawerPopup.tsx
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    if (nestedDrawerOpen) {
+      return;
+    }
+
+    swipePointerProps.onPointerDown?.(event);
+  });
+
+  const handleTouchStart = useStableCallback((event: React.TouchEvent<HTMLDivElement>) => {
```
oliviertassinari: In Material UI, we ignore the events that have `event.defaultPrevented` to avoid event handling conflict:

1. A slider is inside the drawer, and it also try to intercept the event
2. A drawer from another library/version that overlay on top of ours.

I have tested 1. out of curiosity, and it didn't miss: it's broken in our case:

https://github.com/user-attachments/assets/f911ddc1-7361-427a-ba82-978229f2bc39

I don't know if Material UI's approach is the best possible solution, but it looks like we need to fix this.

oliviertassinari (commented):
@atomiks Really cool!

I don't know if you already need feedback, maybe not, but since I'm the only one? on Android, I thought I could have a quick play on my phone, and compare with Material UI:

1. It's impossible to close the drawer, starting outside of the Popup:

https://github.com/user-attachments/assets/2971c5be-a09d-4e51-9f69-32c965fde225

Vaul has the same problem. But on Android, all the native apps, e.g., Google Drive supports it. It works "correctly" with Material UI. I'm used to it at this point. The value for end-users is not having to put any mental energy in think where to place your finger to start the dismissal interaction.

2. It's possible to extend the drawer to infinity, it shouldn't:

https://github.com/user-attachments/assets/e8b792f2-4f53-43db-bb05-7880332d7902

I guess it's an Android-only issue, that we are good on iOS, we only have this issue because we want to have [resistance](https://github.com/user-attachments/assets/d75349cc-2d67-4dee-9204-ea3fcaa4cdcc) when going in the wrong direction, and it's a bit trickier in Android to make it work.

3. The close interaction feels too sensitive. I felt I had "cancelled" my interaction, I barely retracted the drawer, but still, it closed it:

https://github.com/user-attachments/assets/fef42e6d-fd6a-454a-8b94-8df7b7fadc9a

We could have short interactions like this to close the drawer, but then it should take into account the velocity of the movement; this is the approach used in Material UI.

Otherwise:

- I created a start of a benchmark page https://www.notion.so/mui-org/core-Drawer-Sheet-2ebcbfe7b66080a68150cdc9e0b0ac20, to help copy the best stuff that is out there. 
- In #38, linked to the relevant Shadcn issues: helps us understand a bit what those folks need (the key bugs from Vaul) and help folks find our solution.

atomiks:
@oliviertassinari thanks for all those points, very helpful - they should be addressed in the latest update

oliviertassinari:
@atomiks Point 1 solved. Point 2 solved, this is going to be even better than Vaul 👍. 

On Point 3, I feel similar struggles with the drawer closing or not closing based on my effort/intention.
When I looked at this in the past, I could only find a solution with two parameters https://github.com/mui/material-ui/blob/master/packages/mui-material/src/SwipeableDrawer/SwipeableDrawer.js#L703-L715:

- `hysteresis`: Something that defines how far in % the sheet needs to be retracted to be considered as a closed intent. Right now, it seems to be a fixed px value, and hardcoded: 80px. It feels like it should be in %, the value depends on the size of the sheet. Does it need to be controlled by designers? I don't know, keeping private would actually force them to open a GitHub issue, so it would allow us to learn if our default is great. Vaul has this notion too https://github.com/emilkowalski/vaul/blob/3e97aac6a38e4481bade71d7233ed6002e80f9b0/src/index.tsx#L60
- `minFlingVelocity`: Something that defines how much speed in the close interaction is considered a close intent. We don't have this today, which means that I have to move my finger a long distance to close. Doing a burst movement is faster and expresses a non-ambiguous intent. Vaul has this notion too https://github.com/emilkowalski/vaul/blob/3e97aac6a38e4481bade71d7233ed6002e80f9b0/src/index.tsx#L645.

oliviertassinari:
I was curious, had some fun with Design Arena, asked "a drawer component" with a couple of AI models: https://www.designarena.ai/vote/874afe70-a1de-450c-90e4-ed4cb073d5bf. AI is so far to be able to one-shot this component 😁

mnajdova:
Few things I noticed:
1. On iOS + Safari I can scroll the page while the drawer is open (we likely need to block the scrolling interactions when the drawer starts to open).
https://github.com/user-attachments/assets/c60f1f3c-3911-495c-bd56-bd1441aae8e7
2. +1 on the `hysteresis`, it can't be the same value for all drawer sizes, better to do it in % (or at least combine it with a min px value)

Inputs are handled nicely in combination with the drawer, well done 👌

atomiks:
I added velocity-based detection, so fast swipes dismiss with short distances. If you're dragging slowly, you need to cover 50% of the element's size for it to dismiss. (Not sure of the best values). It probably doesn't need customization to begin with, though

@mnajdova ~will be fixed by https://github.com/mui/base-ui/pull/3744, part of Dialog which Drawer extends~ I actually tested it out and it breaks a couple of things:

On iOS 26.1+, the transparent navbar fills in with the drawer's color, but the wrapper nodes break that. In addition, swipe gestures break because the scroll container blocks them. I think `touch-action: none;` is the only viable solution (Vaul adds this). Scroll containers inside the drawer can unset `touch-action` and use `overscroll-behavior: contain` to keep the scroll locked.

zobweyt:
This is really cool! Base UI is going to be the best headless UI library I have ever seen!

I was just wondering how to make this component responsive and simple in the code at the same time? I really want to use one component for large screens and the same one for small ones. It is already possible to implement this using dialog but the UX is not as convenient as drawer on mobile. The shadcn documentation recommends simply using a media query and rendering different components based on the its value. I don't really like this, since there's code duplication and they are essentially very similar components.

It would be great if you added an example to the documentation on how to make this responsive for large screens using the same component if possible.

Good luck with the implementation, it’s already awesome!

colmtuite:
### Swipe from outside to dismiss

> It's impossible to close the drawer, starting outside of the Popup
> The value for end-users is not having to put any mental energy in think where to place your finger to start the dismissal interaction.

This one is controversial. There's value in not having it dismiss too, so the backdrop is a safe zone for swiping (cleaning dust off your screen or whatever). But the biggest issue is that the Drawer should feel native, and iOS does not work like this. So we should either revert this, or detect the OS and have different behaviour for each.

### Swipe area to open
On YT, you can swipe a specific area of the viewport to open the Drawer, which is a nice touch. We should add this to be feature-complete.

https://github.com/user-attachments/assets/bac23901-83e8-4771-86dd-1560087acc58

oliviertassinari:
> that the Drawer should feel native, and iOS does not work like this.

Interesting, I was coming from the Android side, pushing on the "feel native" direction + upside of the approach.
+1 for a branch based on the OS then, even though I don't understand the upside of the safe zone.

atomiks:
> Swipe from outside to dismiss

@colmtuite turns out iOS isn't consistent here. Yesterday I was using Maps and it has a Drawer component that does exactly what we're doing in this PR. Honestly, it feels the best/most natural to me and allows you to sloppily swipe from above, which feels the most intuitive. 

Maybe it's intentionally context-dependent for some reason?

IMG_1794

colmtuite:
@atomiks 

> turns out iOS isn't consistent here

Interesting. I can't seem to find another iOS drawer that works like this. But yeah ok, I guess the question now is do we make it configurable (and only apply the option to iOS)? Have you looked at how this works in Swift, and how to configure it?

`Drawer.SwipeArea` is nice. It feels too aggressive to me though. Looking at Youtube, the threshold for opening after releasing while the pointer is static, seems to be if the drawer is open to >50% of its width/height (depending on direction/position).

atomiks:
> Interesting. I can't seem to find another iOS drawer that works like this. But yeah ok, I guess the question now is do we make it configurable (and only apply the option to iOS)? Have you looked at how this works in Swift, and how to configure it?

I would not make it configurable unless requested. It doesn't seem like it's configurable and Maps uses a "custom" sheet:

> Very likely it’s not using the stock sheet gesture recognizer behavior in those cases (Maps historically used custom bottom sheets well before UISheetPresentationController existed).

I would say it's not controversial though, it's probably something 99% of people don't even notice in iOS to begin with

philwolstenholme:
This is looking fantastic, just been playing around with it on Android Chrome and it's looking and feeling great. Thank you for adding the CloseWatcher support!

oliviertassinari:
The CloseWatcher is so good. Tweet material right there.

Having it supported in the drawer makes me feel like its missing everywhere now. The dialogs, the docs search, the docs nav
SCR-20260130-ulev

But it will take me a long time before trusting UIs that I can use the back button on a webpage and it won't go to the previous URL. A website having it would actually probably mess me up (train me to bad habits, **I would want to close a dialog to the action, and instead load the previous page, and lose what I was typing in a form, nooo**), it would make me want to expect all websites to have it, or have it disabled in the few that have it. Maybe because `<dialog>` supports it by default, people will get more used to it, and it's fine. Unclear #3905.

packages/react/src/drawer/root/DrawerRoot.tsx
```diff
@@ -353,5 +366,37 @@ function DrawerProviderReporter() {
     providerContext?.setDrawerOpen(drawerId, open);
   }, [drawerId, open, providerContext]);
 
+  React.useEffect(() => {
+    // CloseWatcher enables the Android back swipe gesture (Chromium-only).
```
oliviertassinari: It depends on how people configure it. It's a back button or a back swipe.

```suggestion
    // CloseWatcher enables the Android back gesture (Chromium-only).
```

https://eu.community.samsung.com/t5/galaxy-s24-series/buttons-vs-swipe-gestures-what-do-you-all-use-why/td-p/12769482

philwolstenholme:
> The CloseWatcher is so good. Tweet material right there.
> 
> Having it supported in the drawer makes me feel like its missing everywhere now.
> 
> But it will take me a long time before trusting UIs that I can use the back button on a webpage and it won't go to the previous URL. A website having it would actually probably mess me up (train me to bad habits), it would make me want to expect all websites to have it, or have it disabled in the few that have it.

I think it's my favourite recent web API addition, but I know what you mean about it being a bit confusing. It's being adopted slowly and there's no way for users to tell if a site supports it. I've added it to the site I look after at work as our users are 85% mobile and I imagine the Android users are used to being able to swipe back.

docs/src/app/(docs)/react/components/drawer/demos/snap-points/css-modules/index.tsx
michaldudak, line 1: Once the drawer is expanded, I can't make it compact again - it always closes when swiped down.

michaldudak:
The Drawer's contents are unusable with a stylus. Dragging works, but anything that requires clicks (buttons, inputs) don't work unless a really precise tap is performed (with no movement)

docs/src/app/(docs)/react/components/drawer/demos/snap-points/css-modules/index.tsx
atomiks: It's velocity-based, so if you do it slowly enough it won't skip the snap point. The `snapToSequentialPoints` prop disables this, if necessary

michaldudak:
When the Drawer content overflows, it's not possible to close it by swiping anymore (using touch, the other pointer types work ok)

michaldudak:
The new swipe reason is emitted by Drawer, but ReasonToEventMap has no swipe entry. That widens eventDetails.event to Event for swipe cases, losing pointer/touch typing for consumers handling onOpenChange.

michaldudak:
Right-clicking during dragging leaves the drawer in a weird state where it's draggable without a pressed mouse button.

atomiks:
> When the Drawer content overflows, it's not possible to close it by swiping anymore (using touch, the other pointer types work ok)

Can you elaborate on this? It's intentional that scrolling is allowed (and swiping doesn't work) unless `.scrollTop === 0` (for bottom sheets)

michaldudak:
Here's the issue:

https://github.com/user-attachments/assets/8a6f8ae2-4ace-432c-9ded-afe125a3a542

michaldudak:
Sometimes when I swipe to open, the drawer doesn't open fully:

https://github.com/user-attachments/assets/6b5a97d8-0af9-4437-9e8e-efaeec67ab4f

colmtuite:
@atomiks Do you think we need anything new to handle multiple panels in a single drawer? Or would it just be that the drawer itself is transparent, and you add your own divs and buttons inside to achieve the effect?

iPhone 17 - 1 (1)

atomiks:
> Do you think we need anything new to handle multiple panels in a single drawer? Or would it just be that the drawer itself is transparent, and you add your own divs and buttons inside to achieve the effect?

Yep, I would say this is the same as this pattern for Dialog: https://base-ui.com/react/components/dialog#placing-elements-outside-the-popup (popup is transparent, inner elements are colored)

colmtuite:
@atomiks An example might be useful, if you have time

green-mint:
Been following this for sometime, just wanted to point out that the drawer does not close if you try to flick to close inside a scrollable container. 

On examples:

https://deploy-preview-3680--base-ui.netlify.app/react/components/drawer#snap-points

https://deploy-preview-3680--base-ui.netlify.app/react/components/drawer#mobile-navigation

ios 26 safari and chrome

atomiks:
@green-mint thanks, swipe behavior changed after addressing the recent reviews, this is fixed with regression tests in place

docs/src/app/(docs)/react/components/drawer/demos/uncontained/css-modules/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+            <Drawer.Content className={styles.Surface}>
+              <Drawer.Title className={styles.VisuallyHidden}>Profile actions</Drawer.Title>
+              <Drawer.Description className={styles.VisuallyHidden}>
+                Choose an action for this user.
+              </Drawer.Description>
+
+              <ul className={styles.Actions} aria-label="Profile actions">
+                {ACTIONS.map((action) => (
+                  <li key={action} className={styles.Action}>
+                    <button type="button" className={styles.ActionButton}>
```
michaldudak: This can also show how to close the Drawer programmatically without the need for the Close part.

michaldudak (approved):
I'm seeing the behavior I reported in https://github.com/mui/base-ui/pull/3680#issuecomment-3884349254 in other demos as well (mobile navigation, indent), but we can fix it after the preview version is released. I haven't found other issues.

colmtuite (approved):

colmtuite:
@atomiks When I drag a bottom drawer to very close to the bottom of the viewport, then stop dragging completely, then release, it closes.

On Instagram iOS, it seems to work like this too. On Instagram, it requires at least some velocity in order to close (although I think Instagram use RN?)

oliviertassinari:
Just to be clear, this is not getting to be released today? If it's, let's hold a bit on the decision about the plan. (Or maybe it's simply about renaming `DrawerPreview` to [`Unstable_Drawer`](https://www.notion.so/mui-org/React-releases-7490ef9581b4447ebdbf86b13164272d?source=copy_link#13ccbfe7b660807d91a5e6e0c213c27f)).

PR #3682, opened 2026-01-05
[all components] Allow `actionsRef` to be `null`

mj12albert:
Allows initializing actionsRef with `null` instead of a noop (https://github.com/mui/base-ui/pull/3395#issuecomment-3694512603)


michaldudak:
What's the benefit here? If they don't use `actionsRef`, they can simply leave it `undefined`, no?

atomiks:
@michaldudak this is about the ref type: it doesn't accept `null` in the ref type itself (`useRef(null)`) currently so you need to match the shape of `Component.Root.Actions` which is cumbersome/needless with the ease of optional chaining `?.`

PR #3683, opened 2026-01-05
[internal] Add Claude instructions file

michaldudak:
Added CLAUDE.md referencing AGENTS.md

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md


CLAUDE.md
```diff
@@ -0,0 +1 @@
+@AGENTS.md
```
Janpot, line 1: Is it enough to merely import the file or should we add a little bit of context?
```suggestion
See @AGENTS.md for instructions
```

michaldudak: It seems to work as it is - Claude was able to correctly answer my questions that required knowledge from AGENTS.md

Janpot: Yeah, I meant it just in case we need to add further claude-only instructions to this file.

PR #3684, opened 2026-01-05
[accordion] Fix keyboard navigation with non-interactive trigger elements

### LINKED ISSUE #3632: [accordion] `Accordion.Trigger` custom `render` + `nativeButton={false}` breaks semantics and keyboard nav
# Bug report

## Current behavior

When using `Accordion.Trigger` with the `render` prop and `nativeButton={false}`, the documentation suggests rendering a non-button element (e.g. `<div>`):

> Set to false if the rendered element is not a button (e.g. div).

https://base-ui.com/react/components/accordion#trigger

However, `Accordion.Trigger` is expected to be used inside `Accordion.Header`, which renders an `<h3>` element.

According to HTML specifications, any heading element may only contain phrasing content and cannot contain a `<div>`. This leads to invalid HTML when following the documented guidance.

Additionally, when rendering a non-button valid element (e.g. `<span>`) with `nativeButton={false}`, keyboard navigation (Arrow up down keys, Home, End) does not work out of the box due to querySelector only selecting active triggers with `type="button"`. It should also select with `role="button"`.

## Expected behavior

- The docs should not mention `div` inside `h3` when using `render` + `nativeButton={false}`. 
- Keyboard navigation should work if a custom trigger element is used for e.g. `span`.

## Reproducible example

Link to docs: https://base-ui.com/react/components/accordion#trigger

Link to StackBlitz: https://stackblitz.com/edit/t6iymq5o?file=src%2FApp.tsx

Screen recording for navigation:

https://github.com/user-attachments/assets/cd36857a-5934-4f43-841a-73b469622943

## Base UI version

`latest`

Chrome

## Which OS are you using?

Windows

ZeeshanTamboli:
Fixes #3632

---

Before: https://stackblitz.com/edit/t6iymq5o?file=src%2FApp.tsx
After: https://stackblitz.com/edit/faqtsf3b?file=src%2FApp.tsx

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  code-infra: https://raw.githubusercontent.com/mui/mui-public/dbef881bae7f3202ea876354b83af83f8970559a/.circleci/orbs/code-infra.yml
+  code-infra: https://raw.githubusercontent.com/mui/mui-public/7f5c7019e5ec5cc1413a5d6d22137941bba633c8/.circleci/orbs/code-infra.yml
 
 parameters:
   browserstack-force:
--- a/docs/package.json
+++ b/docs/package.json
@@ -20,7 +20,7 @@
     "@base-ui/utils": "workspace:*",
     "@mdx-js/loader": "^3.1.1",
     "@mdx-js/react": "^3.1.1",
-    "@mui/internal-docs-infra": "0.3.1-canary.4",
+    "@mui/internal-docs-infra": "0.3.1-canary.5",
     "@next/mdx": "^16.1.1",
     "@react-spring/web": "^10.0.3",
     "@stefanprobst/rehype-extract-toc": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@arethetypeswrong/cli": "0.18.2",
     "@babel/plugin-transform-react-constant-elements": "7.27.1",
     "@base-ui/monorepo-tests": "workspace:*",
-    "@mui/internal-code-infra": "0.0.3-canary.64",
+    "@mui/internal-code-infra": "0.0.3-canary.65",
     "@mui/internal-netlify-cache": "0.0.2-canary.1",
     "@mui/internal-test-utils": "2.0.18-canary.0",
     "@next/eslint-plugin-next": "16.1.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -31,8 +31,8 @@ importers:
         specifier: workspace:*
         version: link:test
       '@mui/internal-code-infra':
-        specifier: 0.0.3-canary.64
-        version: 0.0.3-canary.64(@next/eslint-plugin-next@16.1.1)(@types/node@22.18.13)(eslint@9.39.2(jiti@2.6.1))(postcss@8.5.6)(prettier@3.7.4)(stylelint@16.26.1(typescript@5.9.3))(typescript@5.9.3)(vitest@4.0.16)
+        specifier: 0.0.3-canary.65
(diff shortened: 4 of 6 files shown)


packages/react/src/accordion/root/AccordionRoot.test.tsx
```diff
@@ -316,6 +316,47 @@ describe('<Accordion.Root />', () => {
       expect(trigger1).toHaveFocus();
     });
 
+    it('ArrowUp and ArrowDown moves focus between custom rendered triggers when nativeButton={false} and loops by default', async () => {
```
mj12albert: I think it's better to loop these new tests within `describe('keyboard interactions')` similar [to here](https://github.com/mui/base-ui/blob/master/packages/react/src/tabs/root/TabsRoot.test.tsx#L622-L627) to avoid having near-identical looking tests

packages/react/src/accordion/trigger/AccordionTrigger.tsx
```diff
@@ -22,17 +22,15 @@ import { useRenderElement } from '../../utils/useRenderElement';
...
   const { current: accordionItemElements } = accordionItemRefs;
 
-  const output: HTMLButtonElement[] = [];
+  const output: HTMLElement[] = [];
 
   for (let i = 0; i < accordionItemElements.length; i += 1) {
     const section = accordionItemElements[i];
     if (!isElementDisabled(section)) {
-      const trigger = section?.querySelector('[type="button"]') as HTMLButtonElement;
+      const trigger = section?.querySelector('[type="button"], [role="button"]') as HTMLElement;
```
mj12albert:
```suggestion
      const trigger = section?.querySelector<HTMLElement>('[type="button"], [role="button"]');
```

packages/react/src/accordion/root/AccordionRoot.test.tsx
ZeeshanTamboli: I didn't understand how we can apply it here. The JSX is different with `render` and `nativeButton` props.

packages/react/src/accordion/trigger/AccordionTrigger.tsx
```diff
@@ -22,19 +22,17 @@ import { useRenderElement } from '../../utils/useRenderElement';
...
+  const output: HTMLElement[] = [];
 
   for (let i = 0; i < accordionItemElements.length; i += 1) {
     const section = accordionItemElements[i];
     if (!isElementDisabled(section)) {
-      const trigger = section?.querySelector('[type="button"]') as HTMLButtonElement;
-      if (!isElementDisabled(trigger)) {
-        output.push(trigger);
+      const trigger = section?.querySelector<HTMLElement>('[type="button"], [role="button"]');
+      if (!isElementDisabled(trigger!)) {
```
LukasTy: WDYT about avoiding the nasty `!`?

```suggestion
      if (trigger && !isElementDisabled(trigger)) {
```

packages/react/src/accordion/trigger/AccordionTrigger.tsx
```diff
@@ -22,19 +22,17 @@ import { useRenderElement } from '../../utils/useRenderElement';
...
+        output.push(trigger!);
```
LukasTy:
```suggestion
        output.push(trigger);
```

packages/react/src/accordion/root/AccordionRoot.test.tsx
mj12albert: Added in https://github.com/mui/base-ui/pull/3684/commits/f46de67ec22d71d592a1efded9e0e850b0558072
This worked ~
```jsx
<Accordion.Trigger
  nativeButton={isNativeButton}
  render={isNativeButton ? undefined : <span />}
>
  Trigger
</Accordion.Trigger>
```

packages/react/src/accordion/trigger/AccordionTrigger.tsx
ZeeshanTamboli: Makes sense.

Issue #3685, opened 2026-01-05
[docs-infra][docs] Custom 404 page doesn't work since v1.0.0


oliviertassinari:
## Reproducible example

1. Open https://base-ui.com/drgdrgdr
2. Open https://base-ui.com/react/overview/quick-startseff

## Current behavior

1. Default Next.js 404 error template
2. Default Next.js 404 error template

## Expected behavior

1. Custom 404 error template that match **website** layout
2. Custom 404 error template that match **docs** layout

## Base UI version

v1.0.0

## Additional context

This is a regression introduced in #3493. For example, v1.0.0-rc.2 works https://693aa73c0eca2100084ef620--base-ui.netlify.app/

<img width="456" height="286" alt="Image" src="https://github.com/user-attachments/assets/74290b5a-fdab-4739-8088-cc6d1da20084" />

It's also related to https://github.com/mui/material-ui/issues/41042.

michaldudak:
I'm not sure if having separate pages per route segment is supported. The current setup clearly doesn't work, and the Next.js docs suggest that for websites with multiple root layouts, `global-not-found` is the way to go (https://nextjs.org/docs/app/api-reference/file-conventions/not-found#global-not-foundjs-experimental). @dav-is, perhaps you know any tricks we can use here?

If there's nothing we can do, #3689 should close this issue.

dav-is:
In this case, it would be worth looking into removing the multiple root layouts. The root layout can be thin. The only imposed limitation is that the `(website)/` route group cannot control the `<html>`, `<head>`, and `<body>` tags of the page.

michaldudak:
One reason to use separate root layouts was to ensure full-page navigation between the docs and marketing website to avoid global styles leaking between them. We can likely achieve the same by replacing all `Link`s between subsites to plain `<a>` tags, but it's less reliable, and it'll be harder to ensure no new `Link`s are introduced.

dav-is:
> by replacing all Links between subsites to plain <a> tags

I think we should avoid this as well.

Couldn't we avoid using global css altogether like this?

```css
.docs {
  a { }
  p { }
}
```

```html
<html>
  <body>
    <div class="docs">
```

michaldudak:
@colmtuite ☝️

colmtuite:
We can avoid global styles yes, and should. But it requires rewriting the docs, which I don't have time for right now. Ultimately, I want the docs to use the same little component set I started for the website.

michaldudak:
[...]

PR #3689, opened 2026-01-05
[docs] Fix custom 404 page

michaldudak:
Preview: https://deploy-preview-3689--base-ui.netlify.app/nonexistent-page

Part of #3685


LukasTy (approved):
LGTM. 👍 
In any case, it looks like a very nice improvement over the current broken behavior. 🙈 
Nitpick: WDYT, would using the same font-family as for the rest of the text be better?
https://github.com/mui/base-ui/blob/2fa5f69f7b6b1290f5cfb69ac0d2b23e8fcd7f55/docs/src/app/(website)/css/components/Text.css#L3-L10
Now it's pretty evident that the error message is in different font. 🙈

dav-is (approved):

PR #3687, opened 2026-01-05
[typescript] Fix typechecking with `typescript@next`

michaldudak:
Our `test_types_next` CI job fails with the following error: https://app.circleci.com/pipelines/github/mui/base-ui/18889/workflows/b0e9a22e-d393-4a8d-9b4f-2d81cc01beb5/jobs/200063/parallel-runs/0/steps/0-106

This PR removes the deprecated setting.

PR #3692, opened 2026-01-06
[typescript] Fix CSS module resolution in TS@next

michaldudak:
In https://github.com/mui/base-ui/pull/3687 I missed one error - in TypeScript 6, CSS modules are not resolved correctly without an explicit module definition. TS 5 seems to pick up the module definition from Next.js, while 6.0, for some reason, doesn't. I copied the module definition from Next.js to our code.

To test it, I ran:

```sh
pnpm dlx @mui/internal-code-infra@canary set-version-overrides --pkg typescript@next
pnpm typescript
```

No errors were reported.

### DIFF
--- a/docs/types.d.ts
+++ b/docs/types.d.ts
@@ -7,9 +7,4 @@ declare module '*.mdx' {
   export default MDXComponent;
 }
 
-declare module '*.module.css' {
-  const classes: { readonly [key: string]: string };
-  export default classes;
-}
-
 declare module '*.css';
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -13,6 +13,7 @@
     "allowJs": true,
     "skipLibCheck": true,
     "allowImportingTsExtensions": true,
+    "noUncheckedSideEffectImports": true,
     "paths": {
       "@base-ui/react": ["./packages/react/src"],
       "@base-ui/react/*": ["./packages/react/src/*"],


docs/types.d.ts
```diff
@@ -6,3 +6,10 @@ declare module '*.mdx' {
   const MDXComponent: (props) => JSX.Element;
   export default MDXComponent;
 }
+
+declare module '*.module.css' {
+  const classes: { readonly [key: string]: string };
+  export default classes;
```
Janpot: This seems to break for me when I turn off `skipLibCheck`, it looks like it's enough to only do the `declare module '*.css'` line

Janpot:
The issue is that in typescript 7 [`noUncheckedSideEffectImports`](https://www.typescriptlang.org/tsconfig/#noUncheckedSideEffectImports) will default to `true`, this is why imports like 

```tsx
import './some/css/file.css'
```

start breaking. It's enough to only retain the ambient module declaration for css files. The module.css declaration causes conflicts with the next.js ones (issue that is hidden by `skipLibCheck`). My suggestion would also be to already set `noUncheckedSideEffectImports` to `true` in tsconfig, when I do this I see the same error in current typescript.

(See https://github.com/microsoft/TypeScript/issues/54500, and https://github.com/vercel/next.js/discussions/70165)

Janpot (changes requested):
See comment

michaldudak:
Fixed as suggested. Both current and next versions of TS still pass.

PR #3700, opened 2026-01-07
[dialog] Fix `Maximum update depth exceeded` error with Suspense

### LINKED ISSUE #3695: [dialog] Suspense across portal causes "Maximum update depth exceeded." error
# Bug report

## Current behavior

When a component within `Dialog.Portal` triggers a Suspense boundary outside of it, the component seems to be crashing with the following.

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

```
<Suspense fallback="Loading...">
  <Dialog.Root open={true}>
    <Dialog.Portal>
      <Dialog.Popup>
        {/* Triggers Suspense */}
        <Message />
      </Dialog.Popup>
    </Dialog.Portal>
  </Dialog.Root>
</Suspense>
```

## Expected behavior

No error is thrown.

## Reproducible example

https://codesandbox.io/p/sandbox/base-ui-dialog-suspense-m7h2l3

## Base UI version

1.0.0

Chrome

## Which OS are you using?

Mac OS

## Additional context

* If the Suspense boundary itself is inside the Portal, the error is not present.
* This scenario is working in Radix

michaldudak:
Adjusted the portal node ref handling in FloatingPortal to avoid state updates during ref cleanup, which prevents the Suspense hide/show loop while still clearing the portal node via container changes.

Also added a missing `open` state getter in DialogPortal.

Fixes #3695

### DIFF
--- a/packages/react/src/dialog/portal/DialogPortal.test.tsx
+++ b/packages/react/src/dialog/portal/DialogPortal.test.tsx
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import { Dialog } from '@base-ui/react/dialog';
 import { createRenderer, describeConformance } from '#test-utils';
+import { expect } from 'vitest';
+import { screen } from '@mui/internal-test-utils';
 
 describe('<Dialog.Portal />', () => {
   const { render } = createRenderer();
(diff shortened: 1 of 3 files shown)


mj12albert (approved):
It fixes the issue in the repro: https://codesandbox.io/p/sandbox/base-ui-dialog-suspense-forked-4gnszf

But the new tests need to be skipped for React 18:
Screenshot 2026-01-08 at 3 06 36 PM

michaldudak:
I actually updated the test to use `React.lazy` instead of `React.use` as it also suspends and is available in React 18.

Tiedye:
Thank you so much, this was killing me

PR #3702, opened 2026-01-08
[FieldItem] Add missing type export

DiegoAndai:
The `FieldItem` type export was missing. Discovered while working with the components.


DiegoAndai:
@michaldudak not sure what caused [the failing check](https://app.circleci.com/jobs/github/mui/base-ui/202340) 🤔

PR #3710, opened 2026-01-09
[docs] Fix Next.js warnings in dev mode

michaldudak:
Fixed two warnings reported when running the docs app in dev mode:
- added the `baseline-browser-mapping` package
- ~~removed the unsupported `eslint` config key (cc @mui/infra, to verify if this can be removed from the shared config)~~ fixed in #3701 in the meantime


docs/next.config.mjs
```diff
@@ -121,7 +121,8 @@ const nextConfig = {
 };
 
 const mergedConfig = withMdx(withDeploymentConfig(nextConfig));
-
+// @ts-expect-error - eslint config is no longer supported in Next.js 16
+delete mergedConfig.eslint;
```
brijeshb42: Bumping `@mui/internal-docs-infra` would fix this. There's already the PR - https://github.com/mui/base-ui/pull/3701

brijeshb42 (approved):

PR #3723, opened 2026-01-12
[field] Add `nativeLabel` prop to `Label` component

atomiks:

This lets users replace `<Field.Label>` with a `<div>` to prevent inheriting some unexpected `<label>` behaviors, especially for `<button>` form controls where you want clicks to only focus a control instead of triggering `.click()`, along with avoiding `:hover` on the `<button>` being triggered when hovering the label.

See https://github.com/mui/base-ui/pull/3722

### DIFF
--- a/packages/react/src/field/label/FieldLabel.tsx
+++ b/packages/react/src/field/label/FieldLabel.tsx
@@ -1,5 +1,6 @@
 'use client';
 import * as React from 'react';
+import { error } from '@base-ui/utils/error';
 import { isHTMLElement } from '@floating-ui/utils/dom';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
 import { ownerDocument } from '@base-ui/utils/owner';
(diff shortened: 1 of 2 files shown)


docs/src/app/(docs)/react/components/select/page.mdx
```diff
@@ -128,9 +128,11 @@ To avoid lookup, [object values](#object-values) for each item can also be used.
 
 A select should have a visible label. Use the [Field](/react/components/field) component to provide a label for the select trigger:
```
mj12albert, line 129: I think it could be helpful to add a sentence explaining why `nativeLabel={false}` is used here (and the combobox page too)

packages/react/src/field/label/FieldLabel.tsx
```diff
@@ -69,9 +108,23 @@ export const FieldLabel = React.forwardRef(function FieldLabel(
 
 export type FieldLabelState = FieldRoot.State;
 
-export interface FieldLabelProps extends BaseUIComponentProps<'label', FieldLabel.State> {}
+export interface FieldLabelProps extends BaseUIComponentProps<'label', FieldLabel.State> {
+  /**
+   * Whether the component renders a native `<label>` element when replacing it via the `render` prop.
+   * Set to `false` if the rendered element is not a `<label>`.
```
mj12albert: Perhaps even worth mentioning here that this is useful for select/combobox?

PR #3724, opened 2026-01-12
[popups] Fix detached trigger remounting

### LINKED ISSUE #3681: [popover] Incorrect positioning with multiple detached triggers
# Bug report

### Issue 1: Incorrect positioning during flip with multiple detached triggers

When multiple detached triggers are present, the popover positions incorrectly after flipping.

Image

https://github.com/user-attachments/assets/e652ed5c-918f-4fc7-91f1-c42f58777e21

### Issue 2: Position error when toggling `showErrorDemo`

When toggling `showErrorDemo` from `true` to `false`, the popover displays at the wrong position:

Image

**Note:** If the initial value of `showErrorDemo` is `false`, the position renders correctly.

## Expected behavior

The popover should maintain correct positioning regardless of:
- The number of detached triggers
- The flip state
- The initial or toggled state of conditional renders

## Reproducible example

[StackBlitz](https://stackblitz.com/edit/tzqusig4?file=src%2FApp.tsx)

## Base UI version

v1.0.0

Chrome

## Which OS are you using?

Mac OS

atomiks:
Fixes #3681

https://stackblitz.com/edit/tzqusig4-qsqweuem?file=src%2FApp.tsx

1. Fixed detached trigger remount positioning by keeping `FloatingRootStore.positionReference` synced with the latest referenceElement (when it hasn't been overridden), so Floating UI doesn't remain anchored to a stale trigger
2. Hardened trigger registration against keyed remount ordering (avoid unregistering a newer element under the same id) and refresh `activeTriggerElement` when the active trigger re-registers
3. Fixed hot reloading and general behavior when the trigger gets reparented. `Popover.createHandle()` is recreated: previously `useRefWithInit(() => externalStore ?? ...)` would keep the old store forever, so the Root kept syncing listeners/floatingRootContext into the old store while the Trigger was using the new one. Additionally, the store needs to be reactive to the `store` handle itself, not just its values

Review: @greptileai


michaldudak (approved):
Nice one!

Issue #3669, opened 2026-01-02
[combobox] Access filtered items without having to control it entirely (e.g. to implement virtualized)


guisehn:
# Feature request

## Summary

I'm currently implementing a virtualized combobox, [following the docs](https://base-ui.com/react/components/combobox#virtualized).

To implement a virtualized list, we need to pass a `children` to `Combobox.List` with a  `ReactNode` containing our virtualized list.

When passing a `ReactNode` to `Combobox.List`, we opt-out of the automatic filtering, and need to handle everything around filtering manually, which considerably increases the complexity of the consumer code, e.g. the example in the docs creates a `searchValue` state, define a `deferredSearchValue`, call `Combobox.useFilter`, then do the filtering inside a `useMemo`.

It would be nice if we could implement a virtualized list without having to give up on the simplicity that base-ui provides out of the box for filtering.

## Motivation

Implementing a virtualized list properly is already a complex thing on its own. If we could remove the complexity of filtering which is currently necessary in this case, it would make things much easier for consumers of the library.

## API ideas

Some ideas on API, in order of personal preference, without doing a breaking change:

- New hook `Combobox.useFilteredItems` that can be called anywhere inside `<Combobox.Root>`

- `Combobox.List.State` received as 2nd arg of `render` could expose the filtered items.

- A new prop on `Combobox.List` that changes the way that `children` works when passing a function (currently, this function is evaluated for every item, we could have a function that is called once and receives the filtered items)

PR #3732, opened 2026-01-13
[combobox][autocomplete] Add `useFilteredItems` hook, allowing for simpler implementation of virtualized comboboxes

guisehn:
This PR adds a `Combobox.useFilteredItems` and a `Autocomplete.useFilteredItems` hook, allowing for simpler implementation of virtualized comboboxes and autocompletes. Solves https://github.com/mui/base-ui/issues/3669.


docs/src/app/(docs)/react/components/combobox/demos/virtualized/css-modules/index.tsx
```diff
@@ -44,98 +72,77 @@ export default function ExampleVirtualizedCombobox() {
...
+      const shouldScroll = reason === 'none' || (reason === 'keyboard' && (isStart || isEnd));
+
+      if (shouldScroll) {
+        queueMicrotask(() => {
+          virtualizer.scrollToIndex(index, { align: isEnd ? 'start' : 'end' });
+        });
+      }
+    },
+    [filteredItems, virtualizer],
+  );
```
guisehn: I'm not super happy with this approach of putting this callback in a ref. But it's the solution I found to make it work given that the parent component `ExampleVirtualizedCombobox` doesn't have access to `virtualizer` or `filteredItems`.

Another alternative I could think of would be to expose `filteredItems` in the event details of `onItemHighlighted`, and expose virtualizer via a ref. If we did that, we could move the callback back up to `ExampleVirtualizedCombobox`. But since it involves change a lot of extra code, I didn't want to touch it for now before getting some feedback.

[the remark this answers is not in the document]
guisehn: Adjusted on https://github.com/mui/base-ui/pull/3732/commits/17fb6dcbcbd0edf14498c170090c8eb3dbceeffe

atomiks: Yeah, this seems to be a handle pattern ([`React.useImperativeHandle`](https://react.dev/reference/react/useImperativeHandle)) to access code from lower in the tree from higher in the tree. The main issue here (at least DX wise) is that you need to call a function lower in the tree via the ref and sync it with `React.useEffect` 

I think the handle pattern is okay, but it should probably be a handle to the virtualizer object, not the callback fn. Then you can still do the `onItemHighlighted` stuff from the root.

So `React.useImperativeHandle(virtualizerRef, () => virtualizer, [virtualizer])`

Then you have access to `virtualizerRef.current` from the root inside `onItemHighlighted`

guisehn: @atomiks thanks, I didn't know about `useImperativeHandle`.

I adjusted it to use it on https://github.com/mui/base-ui/pull/3732/commits/90770d17209a77455fd5de5f40019ddfaf805d80, and the code looks way better now.

Turns out I can access the amount of filtered items via `virtualizer.options.count`, so only the virtualizer ref is enough.

Also [updated the Tailwind example](https://github.com/mui/base-ui/pull/3732/commits/61a658d8d42ad8efd451849d3aa6591174f6a261), and [added `useFilteredItems` to the docs](https://github.com/mui/base-ui/pull/3732/commits/116aef9f6c0187a2c56705c8888521a1263088c6).

packages/react/src/combobox/root/utils/useFilteredItems.ts
```diff
@@ -0,0 +1,9 @@
+import { useComboboxDerivedItemsContext } from '../ComboboxRootContext';
+
+/**
+ * Return filtered items from the combobox.
+ */
```
atomiks, line 5: This seems fine to me - unless the list is really huge, I think `React.useDeferredValue()` can be skipped. 

Can you update the `Autocomplete` demos/API as well?

guisehn: > Can you update the Autocomplete demos/API as well?

Done! https://github.com/mui/base-ui/pull/3732/commits/d5e8c73d3fd68490126d0bce11efc9573cc83f18 and https://github.com/mui/base-ui/pull/3732/commits/6246b3a9306ff10c5bad8a724b41588d779c3da7

atomiks:
@greptileai

docs/src/app/(docs)/react/components/combobox/page.mdx
```diff
@@ -255,3 +255,19 @@ export const metadata = {
     'Base UI',
   ],
 };
+
+## useFilteredItems
+
+Returns the filtered items.
```
atomiks:
```suggestion
Returns the internally filtered items when called inside `<Combobox.Root>`.
```

docs/src/app/(docs)/react/components/autocomplete/page.mdx
```diff
@@ -216,3 +216,19 @@ export const metadata = {
     'Base UI',
   ],
 };
+
+## useFilteredItems
+
+Returns the filtered items.
```
atomiks:
```suggestion
Returns the internally filtered items when called inside `<Autocomplete.Root>`.
```

packages/react/src/combobox/root/utils/useFilteredItems.ts
atomiks, line 5:
```suggestion
/**
 * Returns the internally filtered items.
 */
```

packages/react/src/autocomplete/index.parts.ts
```diff
@@ -22,3 +22,4 @@ export { ComboboxEmpty as Empty } from '../combobox/empty/ComboboxEmpty';
 export { Separator } from '../separator/Separator';
 
 export { useCoreFilter as useFilter } from '../combobox/root/utils/useFilter';
+export { useFilteredItems as useFilteredItems } from '../combobox/root/utils/useFilteredItems';
```
atomiks:
```suggestion
export { useFilteredItems } from '../combobox/root/utils/useFilteredItems';
```

PR #3735, opened 2026-01-14
[docs] Update `handle` prop description in docs

ZeeshanTamboli:
Noticed the descriptions says "popover" for all the relevant components that has the `handle` prop.

~Not sure why is the docs not updated: https://deploy-preview-3735--base-ui.netlify.app/react/components/alert-dialog#root 🤔~
Fixed: https://deploy-preview-3735--base-ui.netlify.app/react/components/alert-dialog#AlertDialogRoot-handle


LukasTy:
I'm not sure why the docs deploy preview does not reflect the change. 🤷 
I've seen problems locally with this data not updating until a development server is restarted, but here... 🤷‍♂️ Logs don't show any problems... 🤔 

Regarding the change itself, I'm not sure if this place is problematic. Popover is a PART that each of those components has.
What do others think?

michaldudak:
This is now correct, thanks. We could make it more generic by writing "popup" instead of the actual component name, but the way it is now is even better IMO.

> Popover is a PART that each of those components has.
What do others think?

A **popup** is a part of each component, not a **popover**

LukasTy:
> A **popup** is a part of each component, not a **popover**

Sorry for that mistake. 🙈 
If you think it's better, let's go with it. 👍

ZeeshanTamboli:
But why is the preview doc not updated?

LukasTy:
> But why is the preview doc not updated?

Does it work locally?
Could be something related to caching, this PR doesn't change any data in any of the MD files...
If it works locally, try pushing a dummy change to any of the docs mdx pages to see if that forces a rebuild.

michaldudak:
Yeah, we're seeing caching issues from time to time (cc @mui/infra)

dav-is:
The caching issue should be resolved with https://github.com/mui/base-ui/pull/2932 which is very close to being ready for review

LukasTy:
Confirming the assertion.
When I pushed a change to mdx, the updates were visible.
https://6967b1303fdb1800082afff9--base-ui.netlify.app/react/components/alert-dialog#AlertDialogRoot-handle
After reverting the change, it's again the cached version. 🙈

LukasTy:
@mui/infra It looks like we also have an issue with `dedupe`.
I changed nothing, just pushed a random commit, and for some reason, dedupe is acting up.
If I run `pnpm dedupe` locally, it produces the same result as in CI.
After rerunning it once more, the change is reverted. 🙈 🤯 🤦

PR #3738, opened 2026-01-14
[preview card] Fix `PreviewCardTrigger` updates delay refs on every render

ZeeshanTamboli:
Noticed in #3182. Because there's no dependency array passed, this runs every render. This is just an improvement.

### DIFF
--- a/packages/react/src/preview-card/trigger/PreviewCardTrigger.tsx
+++ b/packages/react/src/preview-card/trigger/PreviewCardTrigger.tsx
@@ -22,7 +22,7 @@ export const PreviewCardTrigger = React.forwardRef(function PreviewCardTrigger(
 
   useIsoLayoutEffect(() => {
     writeDelayRefs({ delay, closeDelay });
-  });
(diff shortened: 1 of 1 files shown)


packages/react/src/preview-card/trigger/PreviewCardTrigger.tsx
```diff
@@ -22,7 +22,7 @@ export const PreviewCardTrigger = React.forwardRef(function PreviewCardTrigger(
 
   useIsoLayoutEffect(() => {
     writeDelayRefs({ delay, closeDelay });
-  });
+  }, [writeDelayRefs, delay, closeDelay]);
```
atomiks, line 25: There's no need for a dep array here. `writeDelayRefs` only sets `.current = ...` on the ref objects (no state updates, etc.)

ZeeshanTamboli: You are right, closing this PR.

PR #3744, opened 2026-01-15
[dialog] Add iOS scroll lock elements to `Viewport` by default

### LINKED ISSUE #1893: [dialog] Lock scroll on iOS if navbar is collapsed

atomiks:

Fixes #1893 

- [x] Update all dialog demos to use `<Dialog.Viewport>`

### Downsides

- Users cannot edit these wrapper nodes (easily). One solution to that is to make this opt-out or opt-in with a prop.
- The wrapper nodes may cause layout issues in certain scenarios. However, I didn't see any issues on the docs for the existing dialogs that used the Viewport part
- Some users want to lock document scroll for `Popover` as well when it isn't scrollable (#3098), but this only applies to Dialog

### Alternatives

The alternative is to document scroll lock on iOS (which also lets users know how to do it for Popover, though I haven't tried to do it for that component yet to see if it's viable). However, not adding it by default means it will probably be ignored in many cases.

### DIFF
--- a/docs/reference/generated/dialog-viewport.json
+++ b/docs/reference/generated/dialog-viewport.json
@@ -1,6 +1,6 @@
 {
   "name": "DialogViewport",
-  "description": "A positioning container for the dialog popup that can be made scrollable.\nRenders a `<div>` element.",
+  "description": "A positioning container for the dialog popup that can be made scrollable.\nThis part also robustly locks the scroll on iOS devices.\nRenders a `<div>` element.",
   "props": {
     "className": {
       "type": "string | ((state: Dialog.Viewport.State) => string | undefined)",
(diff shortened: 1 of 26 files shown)


atomiks:
Replacing with https://github.com/mui/base-ui/pull/3860

PR #3745, opened 2026-01-15
[combobox][autocomplete] Add `InputGroup` part

### LINKED ISSUE #3943: Combobox.Positioner aligns to Input element instead of parent container when using render prop
## Current behavior

When using `Combobox.Input` with a custom `render` prop that wraps the input in a container (like an InputGroup with addon buttons), the `Combobox.Positioner` aligns to the actual `<input>` element rather than the outer container element.

This causes the popup to misalign when using `align="end"` - it aligns to the end of the input field, not the end of the full container that includes addon buttons.

**Structure:**
```
InputGroup (container - expected anchor)
├── input element (actual anchor)
└── InputGroupAddon with buttons (not included in anchor width)
```

## Expected behavior

The `Combobox.Positioner` should use the outermost element returned by the `render` prop as the anchor, not the internal input element. This would allow the popup to align to the full container width including any addon elements.

## Reproducible example

https://stackblitz.com/~/github.com/mengxi-ream/base-ui-combobox-repro

Image

The key code pattern:

```tsx
function ComboboxInput({
  className,
  children,
  disabled = false,
  showTrigger = true,
  showClear = false,
  ...props
}: ComboboxPrimitive.Input.Props & {
  showTrigger?: boolean
  showClear?: boolean
}) {
  return (
    <InputGroup className={cn('w-auto', className)}>
      <ComboboxPrimitive.Input
        render={<InputGroupInput disabled={disabled} />}
        {...props}
      />
      <InputGroupAddon align="inline-end">
        {showTrigger && (
          <InputGroupButton
            size="icon-xs"
            variant="ghost"
            render={<ComboboxTrigger />}
            data-slot="input-group-button"
            className="group-has-data-[slot=combobox-clear]/input-group:hidden data-pressed:bg-transparent"
            disabled={disabled}
          />
        )}
        {showClear && <ComboboxClear disabled={disabled} />}
      </InputGroupAddon>
      {children}
    </InputGroup>
  )
}

// Usage with align="end" - popup aligns to input, not InputGroup
<ComboboxContent align="end">
```

this is the code that how shadcn library design the combobox input, I think it might not follow the best practice? 

**May need help to know what's the best practice to design Input box with icon on the right to be the Input of the Combobox**

## Base UI version

v1.1.0

Chrome

## Which OS are you using?

Mac OS

## Additional context

A workaround exists using the `anchor` prop on `Combobox.Positioner` with a ref to the container, but this requires manual ref management and breaks the clean component abstraction.

atomiks:

The `InputGroup` part is added to make the styled input for multiple and single selection the same, instead of styling the `Chips` part or `Input` part separately depending on the selection mode.

- ~Adds a `[data-highlighted]` state to `InputGroup` part to indicate when virtual focus is on it~ (deferring to a new PR)
- Automatically sets `anchor` on `Positioner` to the `InputGroup` part when it exists, preventing the need to manually set `anchor` to `<Combobox.Chips>` (closes #3943)

### DIFF
--- a/packages/react/src/combobox/positioner/ComboboxPositioner.test.tsx
+++ b/packages/react/src/combobox/positioner/ComboboxPositioner.test.tsx
@@ -19,6 +19,39 @@ describe('<Combobox.Positioner />', () => {
   }));
 
   describe.skipIf(isJSDOM)('default anchor', () => {
+    it('uses the input when input group is absent', async () => {
+      const inputWidth = 120;
+      const triggerWidth = 240;
+      let anchorWidth = 0;
+      const inputRef = React.createRef<HTMLInputElement>();
+
+      await render(
+        <Combobox.Root open>
+          <Combobox.Input ref={inputRef} style={{ width: inputWidth }} />
+          <Combobox.Trigger style={{ width: triggerWidth }}>Open</Combobox.Trigger>
+          <Combobox.Portal>
+            <Combobox.Positioner
+              sideOffset={(data) => {
+                anchorWidth = data.anchor.width;
+                return 0;
+              }}
+            >
+              <Combobox.Popup>
+                <Combobox.List>
+                  <Combobox.Item value="One">One</Combobox.Item>
+                </Combobox.List>
+              </Combobox.Popup>
+            </Combobox.Positioner>
+          </Combobox.Portal>
+        </Combobox.Root>,
+      );
+
+      await waitFor(() => {
+        expect(anchorWidth).toBeCloseTo(inputRef.current!.getBoundingClientRect().width, 0);
+        expect(anchorWidth).not.toBeCloseTo(triggerWidth, 0);
+      });
+    });
+
     it('uses the input group when present', async () => {
       const inputGroupWidth = 240;
       const inputWidth = 120;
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review

### Overview
This patch adds `InputGroup` for Combobox and Autocomplete, updates popup anchoring to use that wrapper automatically, and refreshes the docs demos around the new part. The merge resolution looks correct, and the changed Combobox demos I spot-checked still align and open correctly.

### Findings (None)

### Confidence: 4/5
High confidence from a full diff review, focused browser spot-checks of the updated Combobox docs demos, and passing targeted tests for `ComboboxInputGroup`, `ComboboxRoot`, and `ComboboxPositioner`.

### Notes
- The merge fix correctly keeps the newer `Combobox.Label` guidance for the input-inside-popup pattern while preserving the new `InputGroup` docs and API surface.

atomiks:
> The implementation looks OK. I'd just add an explanation in the docs that this is an optional part that aims to help style the control.

It's not purely cosmetic as it groups the input and adjacent controls with `role="group"`, and it also affects popup anchoring and outside-press handling, which makes it useful when those parts should behave as one control

But yeah, the old pattern doesn't break with this still, but I'd recommend using it going forward since it simplifies things

PR #3726, opened 2026-01-12
[docs] Improve DemoTabsList when there are many tabs 

sai6855:
closes https://github.com/mui/base-ui/issues/3197
### Before
image

### After
image

PR #3748, opened 2026-01-15
[docs] Avoid overflowing Demo Tabs List on narrow viewport

LukasTy:
Follow up on https://github.com/mui/base-ui/pull/3726.
Avoid cases where only actions occupy most of the viewport.
Reverting to the previous approach of regularly displaying actions.

Avoids: 
Screenshot 2026-01-15 at 7 56 10 pm

### DIFF
--- a/docs/src/components/Demo/Demo.css
+++ b/docs/src/components/Demo/Demo.css
@@ -79,8 +79,8 @@
   }
 
   .DemoToolbarActions {
-    /* Apply sticky only on hover-capable devices to avoid actions taking up most of viewport on touch/small devices */
-    @media (hover: hover) {
+    /* Apply sticky only on wider viewports to avoid actions taking up most of it */
+    @media (min-width: 768px) {
       position: sticky;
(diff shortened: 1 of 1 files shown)


docs/src/components/Demo/Demo.css
```diff
@@ -79,13 +79,16 @@
   }
 
   .DemoToolbarActions {
-    position: sticky;
+    /* Apply sticky only on hover-capable devices to avoid actions taking up most of viewport on touch/small devices */
+    @media (hover: hover) {
```
LukasTy: I wouldn't mind relying on width here instead, but if you think touch capability is a good enough trigger, let's go with it. 👌

atomiks: `width` is definitely more robust here since this can occur on any sized screen technically

LukasTy: Updated to apply styles starting with 768px. 👍 
Other common number: 500 felt a bit too narrow.

PR #3751, opened 2026-01-15
[popover][dialog] Select whole text in initially focused input

### LINKED ISSUE #1523: [popups] Select inputs on focus
Migrating Paper from Radix to Base UI, I expected that input contents would be selected when popups are opened:

Radix Popover:

https://github.com/user-attachments/assets/e1477c50-dab4-4c8f-9d66-22d3fdebbb43

Base UI Popover:

https://github.com/user-attachments/assets/a81d40a3-4e1c-4d47-bf14-4cc7b171d6df

***

As another example, native dialogs do the same:

https://github.com/user-attachments/assets/73cdf1d7-b82d-47fa-880f-4dd3c0e2c34b

michaldudak:
When a Popover or Dialog contains a text input, the input is now correctly focused with all text selected upon opening, matching native popover behavior.

### Changes

- Added `isSelectableInput()` utility to detect text inputs/textareas that support selection
- Enhanced `enqueueFocus()` to accept an `onFocused` callback that runs immediately after focus
- Updated `FloatingFocusManager` to automatically select text in inputs when focused

### Implementation Details

Text selection happens in the same `requestAnimationFrame` as the focus operation, ensuring no visual flicker and matching native platform behavior. The changes are backward compatible and only affect text inputs/textareas (excludes checkboxes, radios, etc.).

Fixes #1523


packages/react/src/floating-ui-react/utils/element.ts
```diff
@@ -82,6 +82,46 @@ export function isTypeableCombobox(element: Element | null) {
...
+/**
+ * Determines if an element is an input or textarea that supports text selection.
+ * @param element Element to check
+ */
+export function isSelectableInput(
+  element: Element | null,
+): element is HTMLInputElement | HTMLTextAreaElement {
+  if (!element || !isHTMLElement(element)) {
+    return false;
+  }
```
atomiks: This seems unnecessary. `Form` just checks if the `tagName === 'INPUT'` because if it's `type="checkbox"` etc., it doesn't actually do anything (and doesn't error either)

I also don't think `<textarea>`s should be selected (which Form also doesn't do)

michaldudak: :+1: Makes sense to have it consistent.

atomiks:
I'm kind of wondering if this is even actually a good default that Radix does?

It feels better not to select the whole text in the first example to me.

Radix: https://ui.shadcn.com/docs/components/radix/dialog
Base: https://ui.shadcn.com/docs/components/base/dialog

However, Radix's is better in the share example since you actually want the text selected explicitly: 
https://ui.shadcn.com/docs/components/radix/dialog#custom-close-button

michaldudak:
In the wild, there's a mix of both approaches. For example, macOS dialogs seem to select the text initially, but the native HTML `<dialog>` does not. Tabbing between fields selects the content, though.

atomiks:
Maybe as a compromise, we can delay the `initialFocus` function call into the rAF/enqueueFocus (to avoid scroll issues), so users can just do `initialFocus={() => inputRef.current?.select()}`

michaldudak:
This could be a bit hard to use, as they don't know which input will be initially focused. A better option, DX-wise, would be to have an `onInitiallyFocused(element: HTMLElement)` prop

atomiks:
> This could be a bit hard to use, as they don't know which input will be initially focused. A better option, DX-wise, would be to have an onInitiallyFocused(element: HTMLElement) prop

But it's always the first one by default, so it should be clear which ref to attach? Also, `.select()` both focuses and selects, so focus will be taken from whichever other input was chosen

I think the explicit `.select()` for copy inputs as in the shadcn example, where you need to attach a ref, is fine/explicit

michaldudak:
If you have a reusable Dialog component, you won't have a ref to what's inside. Also requiring a ref moves the responsibility to call this to the styled component user, whereas `onInitiallyFocused` allows handling it on the design system component level.

michaldudak:
I introduced the prop I mentioned. LMK how you feel about it.

jjenzz (commented):
# DX review

The capability is valid, but the proposed lifecycle name and contract need revision before release. The new focus API should also converge with the polymorphic `initialFocus` prop in v2.

## Findings (4)

### 1. 🔴 `onInitiallyFocused` breaks the established event naming pattern

**Source:** [DialogPopup.tsx:135](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/packages/react/src/dialog/popup/DialogPopup.tsx#L135), [SliderThumb.tsx:546](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/packages/react/src/slider/thumb/SliderThumb.tsx#L546), [PopoverRoot.tsx:144](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/packages/react/src/popover/root/PopoverRoot.tsx#L144)

**Problem:**

Base UI and React name the focus event `onFocus`. The adverb-plus-past-participle form `onInitiallyFocused` gives the same focus concept a new grammatical convention and does not pair naturally with `initialFocus`.

**Requested changes:**

> Could we rename this to `onInitialFocus` across Dialog, Popover, Alert Dialog, the internal focus manager, tests, generated API reference, and the experiment? It pairs directly with `initialFocus` and follows the established `onFocus` event form.
>
> Having both `onFocus` and `onInitiallyFocused` makes consumers memorize two naming conventions for focus handlers without a domain requirement for the difference. Because this prop is unreleased, `onInitialFocus` can become the final name now and remain unchanged in v2 without a compatibility alias.

### 2. 🔴 Two imperative APIs can control the initial-focus transition

**Source:** [DialogPopup.tsx:126](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/packages/react/src/dialog/popup/DialogPopup.tsx#L126), [FloatingFocusManager.tsx:690](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx#L690)

**Problem:**

The function form of `initialFocus` can choose the target, allow the default, suppress focus by returning `false` or `undefined`, and perform arbitrary imperative work. The newly named `onInitialFocus` can also perform arbitrary work and replace or repeat focus. This splits control of one transition across two callbacks without a long-term ownership model, defined precedence, or convergent v2 contract.

**User scenario:**

A design-system wrapper needs interaction-dependent focus and target-dependent text selection. It must coordinate `initialFocus={(openType) => ...}` with `onInitialFocus={(element) => ...}`, relying on call order and deciding which callback should move focus when their actions conflict.

**Requested changes:**

> This seems to introduce a second imperative API for the same initial-focus transition. I recommend making `onInitialFocus` the single long-term intervention point and narrowing `initialFocus` back to stable configuration.
>
> `onInitialFocus` should receive both the resolved candidate and opening interaction details, and should support observing, canceling, or replacing Base UI’s default action. `initialFocus` can retain its boolean and ref forms for stable target configuration. Please deprecate its function member on Dialog Popup, Popover Popup, and the inherited Alert Dialog Popup now, then remove that member in v2. The replacement lifecycle event should ship with its final contract now and remain unchanged in v2.
>
> During compatibility, resolve `initialFocus` first. If it resolves to `false` or `undefined`, stop processing; otherwise invoke `onInitialFocus` with the resolved candidate, and let its cancellation win. Because the `initialFocus` function is currently an inline union member, reshape the public declarations so function-valued `initialFocus` usage receives an editor-visible `@deprecated` signal, recommending `onInitialFocus` and the v2 removal. Verify that signal in generated declarations, JSX selection, `React.ComponentProps`, ref typing, autocomplete, and API generation.

**Downstream consequences of this recommendation:**

- If this recommendation is adopted, the `initialFocus` function and `onInitialFocus` will temporarily overlap. Address it with the precedence above, an editor-visible deprecation on the exact function-valued usage, public-contract tests, and removal of the function member in v2.
- If this recommendation is adopted, function users will need equivalent access to the opening interaction and resolved candidate. Address it by including both in the lifecycle event contract and documenting equivalent before/after examples.

### 3. 🔴 Initial focus cannot be canceled or replaced cleanly

**Source:** [DialogPopup.tsx:135](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/packages/react/src/dialog/popup/DialogPopup.tsx#L135), [enqueueFocus.ts:19](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/packages/react/src/floating-ui-react/utils/enqueueFocus.ts#L19), [createBaseUIEventDetails.ts:56](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/packages/react/src/internals/createBaseUIEventDetails.ts#L56)

**Problem:**

`onInitialFocus` currently runs after `element.focus()` and receives no cancelable event details. Consumers can move focus again, but they cannot prevent Base UI’s first focus event, scroll effect, or assistive-technology announcement.

**User scenario:**

A design-system wrapper inspects the resolved candidate and decides that another element should receive focus. The current callback must bounce focus from the original candidate to the replacement, exposing both transitions to focus handlers and assistive technology.

**Requested changes:**

> Could we invoke `onInitialFocus` in the queued animation frame after resolving the candidate and positioning the popup, but before calling `.focus()`? It should receive the candidate and Base UI-style event details; `eventDetails.cancel()` should skip the default focus action.
>
> This increases the overlap with the polymorphic `initialFocus` function because both APIs would then be able to suppress or replace focus. Please use the ownership and migration described above: resolve `initialFocus` first during compatibility, let lifecycle cancellation win after a candidate exists, deprecate the function member with an editor-visible signal, remove it in v2, and keep `onInitialFocus` unchanged.
>
> Update the selection example to call `eventDetails.cancel()` before `element.select()`. Please cover default focus, ref and function candidates, cancellation, replacement focus, both props supplied together, `initialFocus={false}`, and focus already being inside the popup.

**Downstream consequences of this recommendation:**

- If this recommendation is adopted, cancellation will increase the temporary overlap with the `initialFocus` function. Address it with defined precedence, tooling-visible deprecation, coexistence tests, and v2 removal of the function member.
- If this recommendation is adopted, selection becomes a replacement action rather than a post-focus follow-up. Address it with a copyable `eventDetails.cancel(); element.select()` example and a test confirming that Base UI does not focus again afterward.

### 4. 🟠 The public docs do not explain the focus-management decision boundary

**Source:** [dialog page.mdx:126](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/docs/src/app/%28docs%29/react/components/dialog/page.mdx#L126), [on-initially-focused.tsx:7](https://github.com/mui/base-ui/blob/b24b890dec065e0490cd833996b838ff79e561ec/docs/src/app/%28private%29/experiments/on-initially-focused.tsx#L7)

**Problem:**

The useful selection and design-system examples exist only in a private experiment. The public Dialog focus-management section covers only `initialFocus` and `finalFocus`, and Popover has no equivalent guidance.

**User scenario:**

A consumer trying to select a share-link input finds no public example and may continue using a ref plus their own animation frame, missing the wrapper-friendly lifecycle API.

**Suggested change:**

> Could we promote a small, copyable selection example into the public Dialog and Popover docs? Please explain when to use declarative `initialFocus`, cancelable `onInitialFocus`, and ordinary DOM `onFocus`; show the cancel-then-select pattern. Alert Dialog can point to the same contract because it reuses Dialog Popup.

## What is already sound

- Text selection remains opt-in, avoiding accidental replacement of editable text.
- A resolved-target lifecycle is useful for reusable design-system wrappers whose consumers provide arbitrary children.
- Running the work after positioning avoids the page jump that motivated the original issue.
- The stable callback plumbing avoids requiring consumers to memoize handlers for correctness.
- The tests already cover the resolved default target, a custom ref target, selection, and disabled initial focus.

## Verdict

**Request changes** - The proposed lifecycle uses an inconsistent public name, overlaps with the polymorphic `initialFocus` function, and cannot cancel the default focus action. CC: @jjenzz

---

🤖 Review generated with Codex by

PR #3753, opened 2026-01-15
[form controls] Fix hidden input stray overflow when not in a form

atomiks:

This prevents stray overflow unless inside a form, where native validation popups need to point to the correct area.
This is just a heuristic by checking if `name` is present, and if so, not anchoring to `top`/`left`

### DIFF
--- a/docs/src/app/(docs)/react/handbook/forms/page.mdx
+++ b/docs/src/app/(docs)/react/handbook/forms/page.mdx
@@ -184,6 +184,21 @@ import { Field } from '@base-ui/react/field';
 </Field.Root>;
 ```
 
+Base UI form components use a hidden input to participate in native form submission and validation.
+To anchor the hidden input near a control so the native validation bubble points to the correct area, ensure the component has been given a `name`, and wrap controls in a relatively positioned container for best results.
+
+```tsx title="Positioning hidden inputs" {4,6}
+import { Field } from '@base-ui/react/field';
+import { Select } from '@base-ui/react/select';
+
+<Field.Root name="apple">
+  <Field.Label>Apple</Field.Label>
+  <div className="relative">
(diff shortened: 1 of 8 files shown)


docs/src/app/(docs)/react/handbook/forms/page.mdx
```diff
@@ -184,6 +184,20 @@ import { Field } from '@base-ui/react/field';
 </Field.Root>;
 ```
 
+Base UI form components use a hidden input to participate in native form submission and validation. To anchor the hidden input near a control so the native validation bubble points to the correct area, ensure the component has been given a `name`, and wrap controls in a relatively positioned container for best results.
```
LukasTy: Nitpick to split to a newline only for source code readability.
```suggestion
Base UI form components use a hidden input to participate in native form submission and validation.
To anchor the hidden input near a control so the native validation bubble points to the correct area, ensure the component has been given a `name`, and wrap controls in a relatively positioned container for best results.
```

LukasTy: On another note, this wrapping is needed for some components.
Input, Autocomplete, Combobox don't need it. 🤔 
Should we somehow mention it?

atomiks: Those work with `required` by default since the visible input gets it, but technically Number Field does need it in some cases too. Probably best just to make it consistent everywhere...

LukasTy: Sure, it won't hurt much. 👌 
However, "optimisationist" in me would like to rephrase it like: if you see native validation bubbles positioned incorrectly, please wrap your control in a relatively positioned element. 😆

PR #3758, opened 2026-01-15
[tabs] Fire `onValueChange` for automatic tab selection

### LINKED ISSUE #2097: [tabs] `onValueChange` does not fire when the initially selected tab is picked by the `Tabs` component
# Bug report

## Current behavior

When this happens, the `onValueChange` callback does not fire, even if the selected tab changed, caused by an external factor (from the POV of the consumer).

## Expected behavior

I'd expect the `onValueChange` callback to fire when the component internally picks a selected tab.

## Reproducible example

## Base UI version

`1.0.0-beta.0`

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

I understand this may be an opinionated take — in any case, whatever is decided, it shoud be applied to all components whenever a value changes for reasons out of the consumer's control (and not just tabs).

michaldudak:
Fixes a bug where the `onValueChange` callback was not invoked when the selected tab changed automatically due to external factors.

Previously, onValueChange only fired for user-initiated tab changes (clicks, keyboard navigation). It did not fire when:

- A selected tab became disabled
- A selected tab was removed from the DOM
- All tabs became disabled

Now, `onValueChange` fires in all these scenarios with appropriate reason values.

It also fires during initial render when no `value` or `defaultValue` is provided, and the component automatically selects the first enabled tab.

---

New reason values in onValueChange:

- `'initial'`: First automatic selection on mount (no value/defaultValue provided)
- `'disabled'`: Selected tab became disabled after initial render
- `'missing'`: Selected tab was removed from the DOM

The existing `'none'` reason is used for user-initiated changes.

--

New experiment: https://deploy-preview-3758--base-ui.netlify.app/experiments/tabs/tabs-onValueChange

Fixes https://github.com/mui/base-ui/issues/2097


packages/react/src/tabs/root/TabsRoot.tsx
```diff
@@ -192,44 +195,91 @@ export const TabsRoot = React.forwardRef(function TabsRoot(
...
+  const initialDisabledStateRef = React.useRef<boolean | undefined>(undefined);
+
   useIsoLayoutEffect(() => {
     if (isControlled || tabMap.size === 0) {
       return;
     }
 
     const selectionIsDisabled = selectedTabMetadata?.disabled;
     const selectionIsMissing = selectedTabMetadata == null && value !== null;
+    const hasNoSelection = value == null; // Catches both null and undefined
```
atomiks: It seems uncontrolled `defaultValue={null}` can no longer mean "no active tab" as `hasNoSelection` treats `null` as a signal to auto-select, so the effect will always pick the first enabled tab (and fire onValueChange) even when `null` is explicitly provided

michaldudak: I restored the null value behavior: https://codesandbox.io/p/sandbox/hero---base-ui-example-forked-j4cmf6

atomiks:
## Codex Review

### Overview
This patch makes uncontrolled `Tabs.Root` report `onValueChange` when it changes selection automatically, adds explicit reason strings for those fallback cases, and updates the docs and experiments around that behavior. The fallback state machine itself looks much stronger now, but the new `Tabs.Panel` visibility guard regresses the server-rendered output.

### Findings
#### 1. 🔴 [Blocking] Selected panel is hidden until the client builds the tab map

**Impact**: Server-rendered tabs no longer include the active panel in the initial HTML, so a normal `Tabs.Root` starts with every `tabpanel` hidden until hydration runs. That is a user-visible regression for SSR consumers and it weakens the accessibility story because the selected panel content is missing from the prerendered markup.

**Evidence**: `packages/react/src/tabs/panel/TabsPanel.tsx:59-60` now requires `getTabIdByPanelValue(value)` before a panel can be considered open. That tab id is only registered from `packages/react/src/composite/list/useCompositeListItem.ts:73-86`, which runs in `useIsoLayoutEffect`, so the map is empty during server rendering. I verified this by running `renderToStaticMarkup` on the PR head for a basic `<Tabs.Root defaultValue={0}>…</Tabs.Root>` example: both rendered `tabpanel` elements came out with `hidden=""`, so no panel content was present in the server HTML.

**Recommendation**: Decouple panel openness from the runtime tab-id lookup. The selected panel should stay open whenever `value === selectedValue`, even if `aria-labelledby` is still unavailable until after mount. Then add an SSR-focused regression test that renders tabs to static markup and asserts the selected panel is visible in the initial HTML.

### Confidence: 4/5
High confidence based on a full-pass review of the current `master...head` diff, targeted `TabsRoot` test runs in both JSDOM and Chromium, and a direct server-render repro of the panel visibility change.

### Notes
- `pnpm test:jsdom TabsRoot --no-watch` and `pnpm test:chromium TabsRoot --no-watch` both pass on this branch.
- The latest commits appear to fix the earlier `null`-fallback regression from the previous review; the remaining issue is on the SSR/initial-markup path instead.

michaldudak:
## Codex Review

### Overview
This patch makes uncontrolled `Tabs.Root` report `onValueChange` for automatic selection changes, adds explicit reason values for those fallback cases, and documents the updated callback contract. The latest revisions also tighten the uncontrolled fallback state machine around `null` and initially disabled defaults.

### Findings (None)

### Confidence: 4/5
High confidence based on a full-pass review of the current `master...head` diff, the added edge-case coverage in `packages/react/src/tabs/root/TabsRoot.test.tsx`, and the green CI signal on the PR.

### Notes
- The latest changes address the earlier `null` fallback regression and add coverage for the disabled-default transition.

atomiks:
## Codex Review

### Overview
This patch makes uncontrolled `Tabs.Root` report `onValueChange` when it changes selection automatically, adds explicit reason strings for those fallback cases, and updates the docs and experiments around that behavior. The fallback state machine itself looks much stronger now, but the new `Tabs.Panel` visibility guard regresses the server-rendered output.

### Findings
#### 1. 🔴 [Blocking] Selected panel is hidden until the client builds the tab map

**Impact**: Server-rendered tabs no longer include the active panel in the initial HTML, so a normal `Tabs.Root` starts with every `tabpanel` hidden until hydration runs. That is a user-visible regression for SSR consumers and it weakens the accessibility story because the selected panel content is missing from the prerendered markup.

**Evidence**: `packages/react/src/tabs/panel/TabsPanel.tsx:59-60` now requires `getTabIdByPanelValue(value)` before a panel can be considered open. That tab id is only registered from `packages/react/src/composite/list/useCompositeListItem.ts:73-86`, which runs in `useIsoLayoutEffect`, so the map is empty during server rendering. I verified this by running `renderToStaticMarkup` on the PR head for a basic `<Tabs.Root defaultValue={0}>…</Tabs.Root>` example: both rendered `tabpanel` elements came out with `hidden=""`, so no panel content was present in the server HTML.

**Recommendation**: Decouple panel openness from the runtime tab-id lookup. The selected panel should stay open whenever `value === selectedValue`, even if `aria-labelledby` is still unavailable until after mount. Then add an SSR-focused regression test that renders tabs to static markup and asserts the selected panel is visible in the initial HTML.

### Confidence: 4/5
High confidence based on a full-pass review of the current `master...head` diff, targeted `TabsRoot` test runs in both JSDOM and Chromium, and a direct server-render repro of the panel visibility change.

### Notes
- `pnpm test:jsdom TabsRoot --no-watch` and `pnpm test:chromium TabsRoot --no-watch` both pass on this branch.
- The latest commits appear to fix the earlier `null`-fallback regression from the previous review; the remaining issue is on the SSR/initial-markup path instead.

ciampo:
Hello there! Any chances this PR will be merged soon? I'm also happy to take over if necessary

atomiks:
## Codex Review (GPT-5.5)
Reviewed the full PR branch against freshly fetched `upstream/master`. This is a `bug-fix` PR with secondary public API/docs and regression-test coverage.

### 1. Bugs / Issues

#### 1. 🔴 Controlled panels are hidden when their tab is absent (blocking)

`Tabs.Root` skips automatic fallback in controlled mode, but `Tabs.Panel` now hides the selected panel after tab registration settles unless a corresponding tab is currently registered:

`packages/react/src/tabs/panel/TabsPanel.tsx`

```tsx
const correspondingTabId = getTabIdByPanelValue(value);
const open = value === selectedValue && (!tabRegistrationSettled || correspondingTabId !== null);
```

That changes controlled ownership. A parent can intentionally keep `value` pointed at a panel while the matching tab is temporarily absent, and `upstream/master` keeps that panel open because panel visibility is based only on `value === selectedValue`. This PR makes the selected controlled panel hidden/inert once `tabRegistrationSettled` flips.

**Fix:** Apply the matching-tab guard only for uncontrolled roots where the automatic missing-tab fallback owns the value, or expose controlled state through context so `Tabs.Panel` preserves controlled visibility.

#### 2. 🟡 Remove stale generated JSON artifact (non-blocking)

`docs/reference/generated/tabs-root.json` is introduced as a lone generated artifact, but `upstream/master` has no `docs/reference/generated` entries and this file is not referenced elsewhere in the PR.

```json
{
  "name": "TabsRoot",
  "description": "Groups the tabs and the corresponding panels.\nRenders a `<div>` element.",
  "props": {
```

**Fix:** Remove `docs/reference/generated/tabs-root.json` unless the generated-reference directory is intentionally being restored.

#### 3. ℹ️ Rework the generated prop description so the docs remain readable (informational)

The new `onValueChange` JSDoc bullet list gets flattened inside the generated prop table, causing adjacent reason entries to run together:

`docs/src/app/(docs)/react/components/tabs/types.md`

```md
`'none'`: User-initiated change (click, keyboard navigation).`'initial'`: Automatic selection on mount
```

**Fix:** Rewrite the source JSDoc as table-safe prose or semicolon-separated text, then rerun `pnpm docs:api`.

### Test Coverage Assessment

The added tests cover the intended uncontrolled fallback paths well: initial automatic selection, disabled selected tabs, missing tabs, all-disabled fallback to `null`, explicit `defaultValue`, `defaultValue={undefined}`, cancellation behavior, controlled mode, and panel remount/visibility cases.

Validation performed on the PR head:
- `pnpm test:jsdom TabsRoot --no-watch` passed.
- `pnpm test:chromium TabsRoot --no-watch` passed on rerun.
- `pnpm typescript` passed.

### Recommendation

Request changes 🔴

The uncontrolled fallback behavior and tests are in good shape, but the controlled-mode panel visibility regression should be fixed before merge. The remaining docs/artifact items are small cleanup issues.

packages/react/src/tabs/root/TabsRoot.tsx
```diff
@@ -212,65 +238,126 @@ export const TabsRoot = React.forwardRef(function TabsRoot(
...
+    const eventDetails = {
+      reason,
+      event: new Event('base-ui'),
+      cancel: NOOP,
+      allowPropagation: NOOP,
+      isCanceled: false,
+      isPropagationAllowed: false,
+      trigger: undefined,
+      activationDirection: 'none',
+    } as TabsRoot.ChangeEventDetails;
```
atomiks: Using `createChangeEventDetails` and mutating properties to be NOOP might be cleaner?

packages/react/src/tabs/root/TabsRoot.tsx
```diff
@@ -212,65 +238,126 @@ export const TabsRoot = React.forwardRef(function TabsRoot(
...
+      return;
+    }
+
+    let reason: typeof REASONS.disabled | typeof REASONS.initial | typeof REASONS.missing =
+      REASONS.missing;
+    if (isAutomaticDefault) {
+      reason = REASONS.initial;
+    } else if (selectionIsDisabled) {
+      reason = REASONS.disabled;
+    }
```
atomiks, line 350: Is it worth having these custom reason strings?

michaldudak: I’d keep them as they make it clearer what's going on and allow to differentiate user-initiated changes from automatic ones.

michaldudak:
Replaced by #4704

Issue #3756, opened 2026-01-15
[Toggle Group] `any[]` value type breaks type safety


valerii15298:
# Bug report

## Current behavior

Toggle Group accepts `any[]` type for `value` and `defaultValue` while `Toggle` accepts only `string` as a value.
https://github.com/mui/base-ui/blob/master/packages/react/src/toggle-group/ToggleGroup.tsx#L153

Using `any` types is quite discouraging especially for library public API, this behavior allows passing plain objects or  just anything to the `value` and `defaultValue` props.

## Expected behavior

Type of `value`, `defaultValue` and `onValueChange` arg `groupValue`  should be generic, something like `<T extends string>`. So that consumers of the library can have great DX and type safety when working with *ToggleGroup*.

Or if there are reasons why not do it like so, at least make it a `string[]` type.

## Reproducible example

https://base-ui.com/react/components/toggle-group#ToggleGroup-value

## Base UI version

**v1.0.0**

## Additional context
If there are reasons for it, can you please explain why such decisions with types was made?

markocupic024:
Seems like Toggle values expect string as well, so I think this would be correct to type it as string

ZeeshanTamboli:
Fixed in #3770

valerii15298:
@ZeeshanTamboli I would not necessary agree that it is a best solution. While it is much better, the DX is still not great.
It is quite a common case that `ToggleGroup` value would be not just a string but enum like: `"option1" | "option2" | "option3"` or array of enums. In such case consumer will need to do extra checks and write manual typescript which is not the best DX.

Adding a simple generic like `<T extends string>` would be better. Internally the value still can be treated as `string[]` but for consumer when doing something like:
```tsx
type Option = "left" | "right" | "down"
const value: Option[] = [];

<ToggleGroup.Root
  value={value}
  onValueChange={(newValue: Option[]) => {
    // when using generic the type of `newValue` will be Option
    // when hardcoded string we will have type error here
  }} 
/>
```

markocupic024:
This might also require a change to Toggle and other internal code as well, but can probably be done

valerii15298:
On the second look I think this issue is a bit wider.

If you look at `Select.Root` it uses generic(the same I suggested* to do here) and `Select.Item` has `any` type: https://base-ui.com/react/components/select#SelectItem-value which is also not great I guess should be string also.

[...]

PR #3770, opened 2026-01-16
[toggle group] Type value as string to match Toggle

markocupic024:
Solves https://github.com/mui/base-ui/issues/3756

Toggle only accepts string as type for its values.


michaldudak:
Could you please run `pnpm prettier` on it?

---

It's worth noting that while _technically_ it's a breaking change, any working code should still work, as it's not possible to provide anything other than a string to the Toggle.

PR #3776, opened 2026-01-16
[docs] Support Cmd/Alt + Enter in search

aarongarciah:
Adding support for opening search result links in new tab/new split using `Cmd + Enter` and `Alt + Enter` respectively so there's no need to grab the mouse to do so.

I'm not sure the implementation is the most straightforward one. My first idea was to do something in the `onClick` event in `<Autocomplete.Item render={<Link onNavigate={} onClick={} />}>` (Next.js prevents `onNavigate` when any modifier key is pressed), but the `onClick` event never runs when doing `Cmd/Alt + Enter`. AFAIK this is the piece of code responsible for `onClick` not running: `ComboboxList` component stops all `onKeyDown` events with `Enter` ([source](https://github.com/mui/base-ui/blob/v1.1.0/packages/react/src/combobox/list/ComboboxList.tsx#L85-L109)).

### DIFF
--- a/docs/src/components/Search/SearchBar.tsx
+++ b/docs/src/components/Search/SearchBar.tsx
@@ -189,15 +189,9 @@ export function SearchBar({
     [search],
   );
 
-  const modifierPressedRef = React.useRef(false);
   const highlightedResultRef = React.useRef<SearchResult | undefined>(undefined);
 
   const handleItemClick = React.useCallback(() => {
-    // Don't close dialog if modifier was pressed
-    if (modifierPressedRef.current) {
-      modifierPressedRef.current = false;
-      return;
-    }
     handleCloseDialog(false);
   }, [handleCloseDialog]);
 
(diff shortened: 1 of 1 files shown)


colmtuite:
I'm happy with the UX. It needs a code review.

docs/src/components/Search/SearchBar.tsx
```diff
@@ -189,10 +189,47 @@ export function SearchBar({
     [search],
   );
 
+  const modifierPressedRef = React.useRef(false);
+  const highlightedResultRef = React.useRef<SearchResult | undefined>(undefined);
+
   const handleItemClick = React.useCallback(() => {
+    // Don't close dialog if modifier was pressed
+    if (modifierPressedRef.current) {
```
michaldudak: Is this necessary? If a modifier key is used, `keyDown`'s default will be prevented, which should also prevent the click event from firing.

aarongarciah: Looks like it's not necessary. I removed it.

PR #3783, opened 2026-01-18
[menu] Fix submenu stuck glitch

### LINKED ISSUE #3774: [menu] Performance issues and bugs in sub-menus
Slow and buggy sub-menus

## Current behavior

1. Hovering over multiple sub menu triggers is very sluggish under certain conditions (higher cpu load)
2. Sometimes when hovering across sub menu triggers one of them opens, even though the cursor is already over a different sub trigger. When that happens hovering over any item in the same menu doesn't close the sub menu, because all items have `pointer-events: none;`

Example from our app with no CPU throttling:

https://github.com/user-attachments/assets/82586ff0-4c17-4135-a1de-655f004a1dae

Example in `/experiments/menu/` with 6x CPU throttling:

https://github.com/user-attachments/assets/221939b7-0158-4480-aeff-63633afaa65a

## Expected behavior

1. Better performance when hovering across many sub menu triggers
2. When sub menu trigger opens its submenu, moving the cursor to another item or sub-menu trigger should close that sub menu.

## Reproducible example

- create a Menu with 10 sub-menus, each with 20 items inside
- set CPU throttling in Devtools to 6x
- move your cursor around the menu up and down at different speeds. you should notice sluggishness with the hover/highlight effect and at some point one of the sub menus will open even though the cursor is no longer on its trigger

## Base UI version
`1.0.0`

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

1. I think there's lots of `getBoundingClientRect` action going on on mouse move with `flushSync` and bunch of `timeouts` causing big CPU spikes, forced layouts and style recalcs.

2. Some race condition which causes wrong sub menu to open and prevent hover on other menu items

atomiks:
The `allowMouseEnter` condition is part of the `useHoverReferenceInteraction` `enabled` state in `<Menu.Trigger>`. This can undesirably switch to `false` while hovering submenu triggers, which causes the event listener registration effect to cleanup then re-execute. During this phase (high CPU load/many submenu triggers makes it more obvious), `mouseleave` may not fire since the event listener is temporarily not registered - this prevents the `mousemove` listener for `safePolygon()` from registering, leaving `pointer-events: none;` stuck since it gets cleaned up/removed by that module.

Fixes #3774. The performance issue seems ~related to this but performance is somewhat subjective here as to how it should feel, at least without comparison with other libraries

https://deploy-preview-3783--base-ui.netlify.app/experiments/menu/menu-submenus

### DIFF
--- a/packages/react/src/menu/root/MenuRoot.test.tsx
+++ b/packages/react/src/menu/root/MenuRoot.test.tsx
@@ -1149,10 +1149,11 @@ describe('<Menu.Root />', () => {
 
         const submenuTrigger = screen.getByTestId('submenu-trigger');
 
-        fireEvent.mouseEnter(submenuTrigger);
-        fireEvent.mouseMove(submenuTrigger);
+        await userEvent.hover(submenuTrigger);
 
-        expect(screen.queryByTestId('submenu')).not.to.equal(null);
+        await waitFor(() => {
+          expect(screen.queryByTestId('submenu')).not.to.equal(null);
+        });
       });
 
       it('should not close when submenu is hovered after root menu is hovered', async () => {
--- a/docs/src/app/(private)/experiments/menu/menu-submenus.tsx
+++ b/docs/src/app/(private)/experiments/menu/menu-submenus.tsx
@@ -1,0 +1,162 @@
+'use client';
+import * as React from 'react';
+import { Menu } from '@base-ui/react/menu';
+import { useStableCallback } from '@base-ui/utils/useStableCallback';
+import {
+  SettingsMetadata,
+  useExperimentSettings,
+} from '../../../../components/Experiments/SettingsPanel';
+import '../../../../demo-theme.css';
+import classes from './menu.module.css';
+
+interface Settings {
+  customAnchor: boolean;
+  modal: boolean;
+  openOnHover: boolean;
+  disabled: boolean;
+  customTriggerElement: boolean;
+  side: Menu.Positioner.Props['side'];
+  align: Menu.Positioner.Props['align'];
+}
+
+export default function MenuSubmenus() {
+  const { settings } = useExperimentSettings<Settings>();
+
+  const anchorRef = React.useRef<HTMLDivElement>(null);
+
+  const triggerRender = React.useMemo(
+    () => (settings.customTriggerElement ? <span /> : undefined),
+    [settings.customTriggerElement],
+  );
+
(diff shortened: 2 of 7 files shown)


packages/react/src/menu/root/MenuRoot.test.tsx
```diff
@@ -18,6 +18,15 @@ import { createRenderer, isJSDOM, popupConformanceTests, wait } from '#test-util
...
 import { PATIENT_CLICK_THRESHOLD } from '../../utils/constants';
 
+type HoverUser = {
+  hover: (element: Element) => Promise<void>;
+};
+
+async function hoverWithMouseMove(element: Element, user: HoverUser = userEvent) {
+  fireEvent.mouseMove(element);
+  await user.hover(element);
+}
```
atomiks: The popup intentionally needs to have a `mousemove` fired first before submenu triggers are allowed to fire hover events. This is so, if you're resting the cursor over where the submenu will appear and open it with a keyboard, it won't open the submenu if the cursor happens to be resting over it at the time. As far as I can tell, we need an extra `mouseMove` here so it can update the `allowMouseEnter` state before `user.hover(...)` fires its events. While `user.hover(...)` also fires `mousemove`, the `allowMouseEnter` state updates after the event sequence and so it's too late, while the explicit `mouseMove` updates it beforehand.

This does make testing it harder though - `process.env.NODE_ENV === 'test'` could be used (as elsewhere already), but is pretty hacky

michaldudak:
I updated the playground to have an additional level of menu nesting and was able to break it a few times. I can't reliably reproduce it, though. I was randomly hovering in and out, and a couple of times the menu didn't close.

https://github.com/user-attachments/assets/d6adb5ac-c03a-4e92-af25-51867a0c6419

I also noticed that sometimes an error appears:
`flushSync was called from inside a lifecycle method. React cannot flush when React is already rendering. Consider moving this call to a scheduler task or micro task.`
I was able to reproduce it on master as well, though.

atomiks:
@michaldudak I was able to reproduce the `flushSync` error (fixed), but I couldn't get a stuck submenu stuck at all, updating to 3 levels of nesting. Previously, it was happening quite frequently

If you're able to reproduce & debug, that'd be helpful, otherwise this at least fixes the vast majority of the issue in the common cases

michaldudak:
I can't reproduce it reliably, unfortunately. It happens quite rarely. I'm OK with leaving it as is in this PR.

Issue #3752, opened 2026-01-15
[Select] Missing `finalFocus` prop


vladmoroz:
# Feature request

Select component is missing the `finalFocus` prop, which is present in `Menu` and other modal components

In our case we want to use it in Paper to move focus from the select to canvas after a value is set via pointer

PR #3785, opened 2026-01-18
[select] Add finalFocus prop

markocupic024:
Closes https://github.com/mui/base-ui/issues/3752

Followed existing patterns in code


atomiks (approved):
Thanks for the contribution!

PR #3791, opened 2026-01-19
[internal] Refactor: use class for hover hook state

romgrk:
New attempt of #3451

Refactor to use a class for hover hook state.


packages/react/src/floating-ui-react/hooks/useHoverInteractionSharedState.ts
```diff
@@ -12,67 +12,47 @@ export function isInteractiveElement(element: Element | null) {
...
+  constructor() {
+    this.pointerType = undefined;
+    this.interactedInside = false;
+    this.handler = undefined;
+    this.blockMouseMove = true;
+    this.performedPointerEventsMutation = false;
+    this.unbindMouseMove = () => {};
+    this.restTimeoutPending = false;
+    this.openChangeTimeout = new Timeout();
+    this.restTimeout = new Timeout();
```
michaldudak, line 38: Are these timeouts cleared anywhere when components unmount?

packages/react/src/floating-ui-react/hooks/useHoverInteractionSharedState.ts
```diff
@@ -54,5 +64,7 @@ export function useHoverInteractionSharedState(store: FloatingRootContext): Hove
     data.hoverInteractionState = instance;
   }
 
+  useOnMount(data.hoverInteractionState.disposeEffect);
```
michaldudak, line 68: I'm not sure this is correct. Since this runs on all components using `useHoverInteractionSharedState`, unmounting a single Trigger will reset the  timers. The more correct way of handling this would be triggering the cleanup when the Root part unmounts, but this would require having another hook for the Root.

romgrk: Doesn't that issue also apply to the already existing hook?

michaldudak: It does, indeed. Let's tackle it in a separate PR, then.

PR #3794, opened 2026-01-20
[navigation menu] Add `keepMounted` prop to `Content` part

### LINKED ISSUE #3755: [navigation menu] support for SSR / forceMount
# Feature request

## Summary

Make sure all navigation menu content is rendered to the HTML sent to the browser.

Radix UI made this possible by using `forceMount=true` on the `Content` component: https://www.radix-ui.com/primitives/docs/components/navigation-menu

## Motivation

For SEO purposes, I would like my entire navigation structure be present in raw HTML of my website. The way NavigationMenu is implemented now, this does not seem possible (the "drop down areas" / Content, are not visible). This means unfortunately for now I can't migrate from Radix to Base UI

atomiks:

Closes #3755

`<NavigationMenu.Content keepMounted>` will keep the content mounted in the DOM with a `hidden` attribute during SSR and after hydration - this ensures the content is crawlable. Whenever it should be inserted into the popup, the component moves in there permanently, and stays mounted inside. `<NavigationMenu.Portal keepMounted>` needs to be used for it to "stay" on the DOM after it gets moved.

### DIFF
--- a/packages/react/src/navigation-menu/content/NavigationMenuContent.tsx
+++ b/packages/react/src/navigation-menu/content/NavigationMenuContent.tsx
@@ -2,7 +2,6 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import { inertValue } from '@base-ui/utils/inertValue';
-import { useOnMount } from '@base-ui/utils/useOnMount';
 import { FloatingNode } from '../../floating-ui-react';
 import { contains, getTarget } from '../../floating-ui-react/utils';
 import type { BaseUIComponentProps, HTMLProps } from '../../utils/types';
@@ -17,6 +16,7 @@ import { transitionStatusMapping } from '../../utils/stateAttributesMapping';
 import { StateAttributesMapping } from '../../utils/getStateAttributesProps';
 import { CompositeRoot } from '../../composite/root/CompositeRoot';
 import { popupStateMapping } from '../../utils/popupStateMapping';
+import { EMPTY_OBJECT } from '../../utils/constants';
 
 const stateAttributesMapping: StateAttributesMapping<NavigationMenuContent.State> = {
   ...popupStateMapping,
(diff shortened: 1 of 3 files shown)


michaldudak:
I'd make this opt-in to avoid potential performance hits. With it enabled by default, the initial payload received from the server is larger. The browser also adds these elements to the DOM just to remove them a split-second later. It doesn't have to lay out or paint, so it's not a huge issue, but memory will be allocated, increasing the GC pressure.

atomiks:
@michaldudak yeah good point, I figured navigation menus should always be mounted for SEO, but we can keep it opt-in with `keepMounted` specified on `Root` and possibly on `Content` as well to let them opt-out for specific menus if they're expensive and not critical for SEO

Alternatively (and this actually is probably necessary), we could keep the mounted content in the DOM at all times, since Google might re-evaluate contents after hydration finishes.

packages/react/src/navigation-menu/content/NavigationMenuContent.tsx
```diff
@@ -59,9 +60,11 @@ export const NavigationMenuContent = React.forwardRef(function NavigationMenuCon
   const ref = React.useRef<HTMLDivElement | null>(null);
 
   const [focusInside, setFocusInside] = React.useState(false);
-
+  const [selfMounted, setSelfMounted] = React.useState(false);
   const { mounted, setMounted, transitionStatus } = useTransitionStatus(open);
 
+  useOnMount(() => setSelfMounted(true));
```
michaldudak: There's no need to set this and trigger another render if `keepMounted === false`

packages/react/src/navigation-menu/content/NavigationMenuContent.tsx
```diff
@@ -137,6 +148,17 @@ export const NavigationMenuContent = React.forwardRef(function NavigationMenuCon
     </FloatingNode>,
     portalContainer,
   );
+
+  if (keepMounted) {
```
michaldudak: A comment explaining what's going on here could be useful for future readers.

michaldudak (approved):

packages/react/src/navigation-menu/content/NavigationMenuContent.tsx
atomiks: I removed the clone approach to be simpler

atomiks:
@greptileai

PR #3795, opened 2026-01-20
[select] Fix scroll height loop

### LINKED ISSUE #3789: Select popover growth feedback loop
# Bug report

## Current behavior

Given:
- The list of items in a select exceeds the viewport height
- Positioner is using `alignItemWithTrigger` positioning
- The popover that appears has an item selected so that the popover needs to grow when scrolling down

When I scroll down, the popover continues to grow 1px at a time in a loop until it reaches the max height.

https://github.com/user-attachments/assets/54fc2352-388a-48c7-9592-e2dc973f216b

## Expected behavior

Popover grows when scrolling down and stops growing when scroll input stops.

## Reproducible example

https://stackblitz.com/edit/43a3dxj7?file=src%2FApp.tsx,src%2Findex.module.css,demo.css,theme.css

This bug is very sensitive to spacing and is dependent on the viewport height. I can reliably reproduce this bug with these styles on my 15" mac book with Chrome maximized with the above example. That's a viewport height of 857.

## Base UI version

For example v1.1.0

Chrome 144.0.7559.60

## Which OS are you using?

Mac OS

N/A

## Additional context

Claude summary of issue exploring the code and debugging:

> When the list is scrolled, the Positioner recalculates its position based on the new scroll state, which in turn triggers another scroll adjustment to maintain alignment - creating an infinite feedback loop.

atomiks:

Can't reproduce this at all on our docs (or a simple Vite demo) but in the Stackblitz it's reproducible.

Chrome; browser window height 857px (various smaller values work too)

Master: https://stackblitz.com/edit/43a3dxj7
This PR: https://stackblitz.com/edit/43a3dxj7-tfa2kpiz

Fixes #3789

### DIFF
--- a/packages/react/src/select/popup/SelectPopup.tsx
+++ b/packages/react/src/select/popup/SelectPopup.tsx
@@ -106,10 +106,6 @@ export const SelectPopup = React.forwardRef(function SelectPopup(
       return;
     }
 
-    const setHeight = (height: number) => {
-      positionerElement.style.height = `${height}px`;
-    };
-
     const isTopPositioned = positionerElement.style.top === '0px';
     const isBottomPositioned = positionerElement.style.bottom === '0px';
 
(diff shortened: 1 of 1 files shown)


tcrognon:
This is a more widespread bug. The `Element` `scrollHeight`, `scrollWidth`, `clientHeight`, `clientWidth` properties are rounded integers, but the  `scrollTop`, `scrollLeft` properties are not rounded. The math in this repo mixes the two kinds causing results that can be inaccurate by up to one pixel.

The MDN article (https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled) even calls this out and says that the only way to determine if an element is completely scrolled is
```
Math.abs(element.scrollHeight - element.clientHeight - element.scrollTop) <= 1;
```
(although I'd use a 2px threshold because I've scrolled to the bottom and had `scrollHeight - clientHeight - scrollTop` be slightly greater than 1px before, maybe due to browser rounding implementation)

There is at least one other bug caused by this also - `data-overflow-x-end`/`data-overflow-y-end` attribute not being removed when completely scrolled and the viewport width/height is not an integer.

So it would be nice to have a more general solution.

atomiks:
@tcrognon can you make an issue for Scroll Area? The Select popup scroll logic is self contained and doesn't affect other components

mitchlloyd:
Thanks for fixing this @atomiks! I'm excited to go back to the OSX style select drop downs in my app. One thing I missed drawing attention to was the `font-size: 87.5%;` CSS property in the [StackBlitz](https://stackblitz.com/edit/43a3dxj7?file=src%2FApp.tsx,src%2Findex.module.css,demo.css,theme.css) that likely helped expose pixel rounding issues.

PR #3796, opened 2026-01-20
[combobox] Fix the type of the ref of the `Icon` part

### LINKED ISSUE #3790: ComboboxIcon has inconsistent types
# Bug report

## Current behavior

The Combobox type refers to an HTMLDivElement, but the props type refers to a span

## Expected behavior

Presumably both align on `span` as per the comment above?

## Reproducible example

```typescript
import type { BaseUIComponentProps } from "../../utils/types.js";
/**
 * An icon that indicates that the trigger button opens the popup.
 * Renders a `<span>` element.
 */
export declare const ComboboxIcon: React.ForwardRefExoticComponent<Omit<ComboboxIconProps, "ref"> & React.RefAttributes<HTMLDivElement>>;
export interface ComboboxIconState {}
export interface ComboboxIconProps extends BaseUIComponentProps<'span', ComboboxIcon.State> {}
```

## Base UI version

1.1.0

NA

## Which OS are you using?

NA

None

flaviendelangle:
Fixes #3790

We have a some components the use `HTMLElement` instead of the exact typing, which most of them use the exact typing (`HTMLDivElement`, `HTMLSpanElement`, etc...)

Is there a reason to have this distinction? 

| File | Current Ref Type | Rendered Element | Should Be |
|------|------------------|------------------|-----------|
| `packages/react/src/button/Button.tsx` | `HTMLElement` | `button` | `HTMLButtonElement` |
| `packages/react/src/checkbox/root/CheckboxRoot.tsx` | `HTMLElement` | `span` | `HTMLSpanElement` |
| `packages/react/src/radio/root/RadioRoot.tsx` | `HTMLElement` | `span` | `HTMLSpanElement` |
| `packages/react/src/switch/root/SwitchRoot.tsx` | `HTMLElement` | `span` | `HTMLSpanElement` |
| `packages/react/src/tabs/tab/TabsTab.tsx` | `HTMLElement` | `button` | `HTMLButtonElement` |
| `packages/react/src/accordion/trigger/AccordionTrigger.tsx` | `HTMLElement` | `button` | `HTMLButtonElement` |
| `packages/react/src/popover/trigger/PopoverTrigger.tsx` | `HTMLElement` | `button` | `HTMLButtonElement` |
| `packages/react/src/menu/item/MenuItem.tsx` | `HTMLElement` | `div` | `HTMLDivElement` |
| `packages/react/src/menu/checkbox-item/MenuCheckboxItem.tsx` | `HTMLElement` | `div` | `HTMLDivElement` |
| `packages/react/src/menu/radio-item/MenuRadioItem.tsx` | `HTMLElement` | `div` | `HTMLDivElement` |
| `packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx` | `HTMLElement` | `div` | `HTMLDivElement` |
| `packages/react/src/select/item/SelectItem.tsx` | `HTMLElement` | `div` | `HTMLDivElement` |

### DIFF
--- a/docs/reference/generated/menu-checkbox-item-indicator.json
+++ b/docs/reference/generated/menu-checkbox-item-indicator.json
@@ -1,6 +1,6 @@
 {
   "name": "MenuCheckboxItemIndicator",
-  "description": "Indicates whether the checkbox item is ticked.\nRenders a `<div>` element.",
+  "description": "Indicates whether the checkbox item is ticked.\nRenders a `<span>` element.",
   "props": {
     "className": {
       "type": "string | ((state: Menu.CheckboxItemIndicator.State) => string | undefined)",
--- a/docs/reference/generated/menu-radio-item-indicator.json
+++ b/docs/reference/generated/menu-radio-item-indicator.json
@@ -1,6 +1,6 @@
 {
   "name": "MenuRadioItemIndicator",
-  "description": "Indicates whether the radio item is selected.\nRenders a `<div>` element.",
+  "description": "Indicates whether the radio item is selected.\nRenders a `<span>` element.",
   "props": {
     "className": {
       "type": "string | ((state: Menu.RadioItemIndicator.State) => string | undefined)",


packages/react/src/menu/checkbox-item-indicator/MenuCheckboxItemIndicator.tsx
```diff
@@ -9,7 +9,7 @@ import { useOpenChangeComplete } from '../../utils/useOpenChangeComplete';
 
 /**
  * Indicates whether the checkbox item is ticked.
- * Renders a `<div>` element.
+ * Renders a `<span>` element.
```
flaviendelangle, line 12: I noticed those two other inconsistencies that only impact the JSDoc

atomiks:
> We have a some components the use HTMLElement instead of the exact typing, which most of them use the exact typing (HTMLDivElement, HTMLSpanElement, etc...)

> Is there a reason to have this distinction?

Some of my reasoning in this PR https://github.com/mui/base-ui/pull/3638. Depends if it's likely to be polymorphic or not. Though for typesafety, the ref can be attached to the rendered element itself anyway: `<Button ref={buttonRef} render={<div ref={divRef} />} />`

flaviendelangle:
@atomiks thats what I thought, thanks for the confirmation :pray:

PR #3797, opened 2026-01-20
[popups] Improve performance by batching ref access

michaldudak:
Created the `useStableCallbacks` hook to batch individual `useStableCallback` calls.
Used the new hook in `useDismiss`. After the change, components using `useDismiss` have 34 fewer hooks (for example, DialogRoot, 150 -> 116).

### Performance benchmark results

_Edge 144 @ M3 Pro, no throttling, all components (Dialog, Menu, Popover, Tooltip)_

#### Contained triggers

https://deploy-preview-3797--base-ui.netlify.app/experiments/perf/contained-triggers

this PR: 150.5ms
master: 167ms

#### Detached triggers

https://deploy-preview-3797--base-ui.netlify.app/experiments/perf/detached-triggers

this PR: 72.3ms
master: 75ms


michaldudak:
@greptileai review this

packages/utils/src/useStableCallbacks.ts
romgrk, line 1: In case you're interested, I'm pretty sure it would be possible to re-use the same `fastComponent()` logic from https://github.com/mui/base-ui/pull/3445 to reduce the cost of multiple `useStableCallback()` calls to 2 hooks (down from `O(n)`), just like we did for `useStore()`. The advantage is that the ergonomy of the API doesn't change.

This PR's API is probably slightly more efficient though, because there is a single object declared once, whereas with multiple calls we'd need to accumulate the stable callbacks in an array incrementally, so a bit more bookkeeping.

michaldudak: This solution is also more straightforward and IMO easier to understand, as it's just another hook call, so I prefer it this way.

romgrk: But it does requires code to be re-organized in a way where the callbacks are declared all at once, and it wouldn't work if the hook was modularized into multiple variants (e.g. see the `useHover...` hooks). The strength of the fast hooks approach is that it doesn't require the code using it to be aware of it or be organized in a specific way. It also works for all the other interaction hooks at once (e.g. `useHover...`) without requiring more refactoring.

atomiks:
Is it feasible to also combine `React.useRef()`/`useValueAsRef` calls into a single one (object ref), which could also help reduce hook calls?

packages/utils/src/useStableCallbacks.ts
michaldudak: I'm OK with requiring callbacks to be declared together via `useStableCallbacks`. It’s more explicit and doesn't introduce more "magic".
I’d rather keep this PR’s scope focused and avoid introducing fastComponent here unless we decide to standardize on that approach across the board.

michaldudak:
> Is it feasible to also combine React.useRef()/useValueAsRef calls into a single one (object ref), which could also help reduce hook calls?

I created a `useValuesAsRef` hook.
3 more hook calls saved in DialogRoot, but I don't see any improvements in mount times - perhaps the gain is too small.

michaldudak:
This adds significant weight to the bundle while improving the performance only marginally. It's not worth it. I'll explore the fasthooks implementation instead.

PR #3798, opened 2026-01-20
[popups] Fix nested hoverable popups

### LINKED ISSUE #3673: [preview card] nested preview cards don't work
# Bug report

1. open the repro below
2. hover over the link
3. hover over the link inside the preview card
4. hover over the second preview card

## Current behavior

https://github.com/user-attachments/assets/ac2be947-2835-4ba8-bfc1-cd679947418b

## Expected behavior

the nested preview card is interactive

## Reproducible example

https://stackblitz.com/edit/l5ifddn6

## Base UI version

1.0.0

chrome 142

## Which OS are you using?

macos

n/a

## Additional context

this is definitely not the best pattern in terms of ui/ux, but i think it should still be supported

flaviendelangle:
Atempt to fix #3673 

The hardest part was to now close the root Preview Card when hovering through its children Preview Card

Preview: https://deploy-preview-3798--base-ui.netlify.app/experiments/preview-card/nested


packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
```diff
@@ -200,6 +202,20 @@ export function useHoverFloatingInteraction(
     // Ensure the floating element closes after scrolling even if the pointer
     // did not move.
     // https://github.com/floating-ui/floating-ui/discussions/1692
+    // Check if the target is inside any child floating element in the tree
+    // This handles portaled nested popups that aren't DOM children
+    function isTargetInsideChildFloating(target: Element | null): boolean {
```
atomiks: Popover uses FloatingTree to support nested `openOnHover` already but doesn't (shouldn't) need any changes to the code, unless there was a regression. `safePolygon` logic checks for tree nodes/nesting

flaviendelangle: If you remove this new code, you can see that the root preview card closes when hovering the children.
Not sure why though

atomiks: Seems like Popover regressed: https://stackblitz.com/edit/9wmm9ajo, possibly from the detached triggers refactoring

atomiks: [Navigation Menu still works](https://base-ui.com/react/components/navigation-menu#nested-submenus), which uses the old `useHover` hook instead of `useHoverReferenceInteraction`/`useHoverFloatingInteraction`. I'd recommend seeing why the split caused a regression, instead of patching in new code since it's not fixing the root cause

flaviendelangle: I'll have a look :+1:

packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
```diff
@@ -222,8 +229,11 @@ export function useHoverFloatingInteraction(
     function onFloatingMouseEnter(event: MouseEvent) {
       openChangeTimeout.clear();
       clearPointerEvents();
+      // Call the safePolygon handler if it exists, but DON'T clear it.
+      // The handler needs to remain active to detect when mouse leaves to
+      // a nested portaled popup. The safePolygon logic checks for open children
+      // in the floating tree and prevents closing when appropriate.
       handlerRef.current?.(event);
```
flaviendelangle: @atomiks I think this fixes the problem.
If we clean the mouse move events when opening the nested popover / preview card, then `safePolygon` can't properly do its job.

Tbh I don't know the codebase enough yet to be fully confident of the fix.
My experiment of both the nested Popover AND the nested Preview Card now work at least.
And without introducing additional logic that `safePolygon` should already handle.

packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
flaviendelangle: Discussing the new proposal in https://github.com/mui/base-ui/pull/3798#discussion_r2712495718

flaviendelangle:
@greptileai

packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
atomiks: Yeah that looks to be a difference between the two hooks.

`useHover`

```jsx
    function onFloatingMouseEnter() {
      timeout.clear();
      clearPointerEvents();
    }
```

`useHoverFloatingInteraction`

```jsx
    function onFloatingMouseEnter(event: MouseEvent) {
      openChangeTimeout.clear();
      clearPointerEvents();
      handlerRef.current?.(event);
      cleanupMouseMoveHandler();
    }
```

Though the old one still doesn't call `handlerRef.current`

flaviendelangle: I don't get any failing test and the nested popover and preview cards still behavior correctly if I comment the `handlerRef.current?.(event);` line.

But that's a change that could easily not be covered by any test.
cc @michaldudak since you did the initial `useHoverFloatingInteraction` implementation

CiscoFran10 (commented):
@flaviendelangle I found a couple of issues while testing nested previews:

---

### 1. Parent trigger does not regain focus when closing with `Esc`

When pressing `Esc`, only the **child preview** returns focus to its trigger.
The **parent trigger** loses focus and does not regain it.

[https://github.com/user-attachments/assets/2cc4b01a-123c-42a8-8939-f9cae0259174](https://github.com/user-attachments/assets/2cc4b01a-123c-42a8-8939-f9cae0259174)

---

### 2. Mouse-leave close logic is scoped per preview instead of globally

Each preview popup appears to compute its own “mouse outside” state.
This means the user must move the mouse outside of *each* preview to close them, even if the cursor is already outside all previews.

[https://github.com/user-attachments/assets/b9d1c5bf-a456-4e40-974d-54364a124bd0](https://github.com/user-attachments/assets/b9d1c5bf-a456-4e40-974d-54364a124bd0)

---

### 3. Unclear but reproducible odd behavior

**Steps to reproduce**

1. Hover the parent trigger to open the parent preview.
2. Hover the child trigger to open the child preview.
3. Move the mouse outside of *all* previews.
4. Before the parent preview finishes closing (there seems to be a small close delay), move the mouse back into the parent preview.
5. Hover the child trigger again.

**Actual result**
Both parent and child previews close.

**Expected result**
The parent preview should remain open, and the child preview should open again normally.

This feels like a race condition between delayed close timers and hover-open logic, but I’m not fully sure.

[https://github.com/user-attachments/assets/eaace2f9-011a-4e2c-857e-6b2ae4ea1057](https://github.com/user-attachments/assets/eaace2f9-011a-4e2c-857e-6b2ae4ea1057)

atomiks:
@CiscoFran10 number 1 is because preview cards don't have focus management (same as tooltip), I would say this is not a concern in practice. For number 2, this is a known issue that Navigation Menu with submenus also has - definitely needs investigating but technically a separate issue. For number 3, not sure on this one, might be the difference between `useHover`

flaviendelangle:
I'll have a look at the 3rd issue when I have some time then :+1:

packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
flaviendelangle: Looks like the point 3 in Cisco's comment is caused by this `handlerRef.current` call
Are we fine removing it?

The diff is here: https://github.com/mui/base-ui/pull/3798/changes/7a9660126ebc38a4694bb4bd0df25c425a840ba3

atomiks:
Doesn't seem like [7a96601](https://github.com/mui/base-ui/pull/3798/commits/7a9660126ebc38a4694bb4bd0df25c425a840ba3) addressed the third point @CiscoFran10 pointed out as I can still reproduce it in the experiments link. (Edit: actually, it made it worse as the "hover-out" step is no longer needed).

Separate issue as well: traversing through the triangle area closes it unexpectedly once you land on the popup.

https://github.com/user-attachments/assets/56fa9873-d089-4e89-a5ed-87c3286615eb

atomiks (approved):
While not perfect this solves the `openOnHover` regression for Popover. 

@CiscoFran10's second and third points still apply, but both of them have a decent workaround by using `closeDelay={0}` on the triggers, since mousemove events are typically fired for both root and child when moving outside, since the child closes instantly and more than 1 mousemove event is usually fired. 

I had a prototype working to solve this robustly, but it adds a lot of code and it's not clear how worth it is over using a 0 close delay.

Issue #3688, opened 2026-01-05
[performance] Bundle size vs. Radix


oliviertassinari:
## Reproducible example

1. Go to https://ui.shadcn.com/create
2. Export with Radix and Base UI
3. Use the new Next.js bundle size analyzer https://nextjs.org/docs/app/guides/package-bundling#nextjs-bundle-analyzer-experimental to compare both

## Current behavior

You will get:

Base UI 463.57 KB

<img width="1471" height="733" alt="Image" src="https://github.com/user-attachments/assets/7066095f-bea8-4dc5-a7cf-31284ebf18cd" />

Radix 443.64 KB

<img width="1429" height="730" alt="Image" src="https://github.com/user-attachments/assets/451520d0-5384-4290-833f-303511b86119" />

So Base UI is 20KB gzipped larger, while it benefits from utils deduplication relative to Radix since Radix imports Base UI Combobox.

## Expected behavior

Base UI is smaller.

## Base UI version

v1.0.0

## Additional context

Solving this would help convince people to migrate. I noticed this trying to compare the two https://www.notion.so/mui-org/2026-Shadcn-Radix-vs-Base-UI-demo-2dfcbfe7b6608090b851c5f6e1f60e95.

atomiks:
I dug into this a bit and the main cost is the "detached triggers" feature / stores we have, which add around 10KB to popup related components, comparing `canary` to Base UI `beta.3`.

The other ones are a mix of two things mainly:

1. Better correctness, a11y, and UX — which is an unavoidable sacrifice for the most part
2. More features generally (detached triggers are the biggest cost)

No doubt we can improve a bit here (e.g. https://github.com/mui/base-ui/pull/4336), but there's no chance we'll go as low as Radix. Making many features optional would likely be the biggest improvement to the bundle size of the library, but it will harm DX.

oliviertassinari:
I did another dive with Ark UI vs. Base UI Combobox this time, from https://github.com/shadcn-ui/ui/issues/9393#issuecomment-4021334880

> I'm uncomfortable with adding Base UI to **my bundle** just for a combobox.
I need the trigger to be an input, and using the 'command' component just doesn't feel right.
So I switched to Ark UI for the combobox, since **I already use** Ark UI as a complement to shadcn.

The result:

<img width="727" height="120" alt="Image" src="https://github.com/user-attachments/assets/89159cb6-6e55-4228-bdeb-91f5c7776378" />

[...]

PR #3805, opened 2026-01-21
[progress][meter] De-duplicate `formatValue` function

sai6855:
I was going through this issue https://github.com/mui/base-ui/issues/3688 and this comment https://github.com/mui/base-ui/pull/3709#issuecomment-3733708193, changes in this PR reduced size by 68B without hurting readability. Felt it as quick win

image


packages/react/src/utils/formatNumber.ts
```diff
@@ -35,3 +35,18 @@ export function formatNumberMaxPrecision(
     maximumFractionDigits: 20,
   });
 }
+
+export function formatValue(
```
LukasTy: Nitpick: Given that this is now a more global util, I feel like `formatNumberValue` might be a more precise name for it. 🤔 
WDYT?

sai6855: Yes, makes sense. I'll update and merge

PR #3806, opened 2026-01-21
[context menu] Fix `disabled` prop not working

arturbien:
## Summary

The `disabled` prop on `ContextMenu.Root` was not preventing the context menu from opening. 
When opened with `disabled` prop the context menu would completely block the page and could not be closed with clicks outside or ESC key.

Here's a recording of the issue:

https://github.com/user-attachments/assets/db7447f9-f8cd-4122-b389-cca7df1a8666

This PR fixes the issue by checking the `disabled` state in `ContextMenuTrigger` before opening the menu:

https://github.com/user-attachments/assets/743ece62-b35e-4d43-9ba8-d0b0ab8213dc

### Changes

- **Fix**: Check for `disabled` state in `ContextMenuTrigger` before opening on right-click (`handleContextMenu`) and long press (`handleTouchStart`)
- **Tests**: Added tests for the `disabled` prop covering both right-click and long press (touch) interactions
- **Demo**: Added a "Disabled Context Menu" demo to the experiments page

## Test plan

- [x] Verify right-clicking the trigger does not open the menu when `disabled={true}`
- [x] Verify long press on touch devices does not open the menu when `disabled={true}`
- [x] Verify `onOpenChange` is not called when the menu is disabled
- [x] Verify the native context menu opens when disabled context-menu component is clicked
- [x] Existing tests still pass (`pnpm test:jsdom ContextMenu`)


packages/react/src/context-menu/trigger/ContextMenuTrigger.test.tsx
```diff
@@ -135,6 +136,30 @@ describe('<ContextMenu.Trigger />', () => {
     expect(onOpenChange.lastCall.args[0]).to.equal(false);
   });
 
+  describe('prop: disabled', () => {
+    it('does not open on right-click when disabled', async () => {
+      const onOpenChange = vi.fn();
```
mj12albert: Should use `spy` and the chai `expect` here for consistency

docs/src/app/(private)/experiments/context-menu.tsx
```diff
@@ -1,12 +1,15 @@
 'use client';
+import * as React from 'react';
 import { ContextMenu } from '@base-ui/react/context-menu';
 import { Menu } from '@base-ui/react/menu';
 import { Popover } from '@base-ui/react/popover';
 
 export default function ContextMenuExperiment() {
+  const [disabled, setDisabled] = React.useState(false);
```
mj12albert, line 8: For this it would be good to use
```js
import {
  SettingsMetadata,
  useExperimentSettings,
} from '../../../components/Experiments/SettingsPanel'
```
example: https://github.com/mui/base-ui/blob/b782bdd15253d29634a25a152f94b041dbe3497e/docs/src/app/(private)/experiments/menubar.tsx#L28

mj12albert (commented):
Thanks for the PR ~ I can reproduce the bug and left some comments

packages/react/src/context-menu/trigger/ContextMenuTrigger.test.tsx
arturbien: done

docs/src/app/(private)/experiments/context-menu.tsx
arturbien: Not sure if its the right approach in this case. There's bunch of complex examples of the context-menu on this page like multi level nested context menus, and its not clear where the disabled prop should be applied (the inner, outer, nested, all of them?)

in the menubar case, there's only one instance of the menubar on the page so its clear where the settings props go

Issue #3807, opened 2026-01-21
CheckboxGroup + Fieldset/Field causes duplicate input id (and duplicate label[for]) when using Field.Root name + Fieldset.Root render={<CheckboxGroup />}


imadroshan:
# Bug report

## Current behavior

When composing `CheckboxGroup` with `Field.Root` and `Fieldset.Root` (using render={<CheckboxGroup ... />} as recommended by docs), the hidden `<input type="checkbox">` elements created by `Checkbox.Root` end up with the same id across multiple checkboxes.

This also leads to `label[for]` collisions when using `Field.Label` (same `for` and same label `id` reused across multiple items).

This violates HTML id uniqueness rules and breaks label → control associations (and can confuse assistive tech and testing tools).

## Reproducible example

```tsx
import * as React from "react";
import { Field } from "@base-ui/react/field";
import { Fieldset } from "@base-ui/react/fieldset";
import { CheckboxGroup } from "@base-ui/react/checkbox-group";
import { Checkbox } from "@base-ui/react/checkbox";

export default function Repro() {
  return (
    <Field.Root name="allowedNetworkProtocols">
      <Fieldset.Root render={<CheckboxGroup defaultValue={["https"]} />}>
        <Fieldset.Legend>Allowed network protocols</Fieldset.Legend>

        {/* Version A: native label */}
        <label>
          <Checkbox.Root value="http" />
          HTTP
        </label>

        <label>
          <Checkbox.Root value="https" />
          HTTPS
        </label>

        <label>
          <Checkbox.Root value="ssh" />
          SSH
        </label>

        {/* Version B: Field.Label (also shows duplicate ids/for) */}
        {/*
        <Field.Label>
          <Checkbox.Root value="http" />
          HTTP
        </Field.Label>
        <Field.Label>
          <Checkbox.Root value="https" />
          HTTPS
        </Field.Label>
        <Field.Label>
          <Checkbox.Root value="ssh" />
          SSH
        </Field.Label>
        */}
      </Fieldset.Root>
    </Field.Root>
  );
}
```

## Actual Results
Inspecting the DOM shows:

### Version A (native <label>)

All hidden inputs share the same id, e.g.:
```html
<input id="base-ui-_r_1or_" ... name="allowedNetworkProtocols" />
<input id="base-ui-_r_1or_" ... name="allowedNetworkProtocols" />
<input id="base-ui-_r_1or_" ... name="allowedNetworkProtocols" />
```
[...]

PR #3809, opened 2026-01-21
[field][docs] Update `Field.Item` related docs

mj12albert:
Closes https://github.com/mui/base-ui/issues/3807


docs/src/app/(docs)/react/components/checkbox-group/page.mdx
```diff
@@ -50,24 +50,30 @@ For individual checkboxes, use an enclosing `<label>` element that wraps each `<
 The [Field](/react/components/field) and [Fieldset](/react/components/fieldset) components eliminate the need to track `id` or `aria-labelledby` associations manually. They support both enclosing and sibling labeling patterns, along with a UX improvement to prevent double clicks from selecting the label text.
 
 1. **Group label**: Since the group contains multiple controls, it can be labeled as a fieldset. Use [Fieldset](/react/components/fieldset) by replacing the `<Fieldset.Root>` element with `<CheckboxGroup>` via the `render` prop. `<Fieldset.Legend>` provides the visible label for the group.
-2. **Individual label**: Label an individual checkbox using an enclosing `<Field.Label>`. Enclosing labels ensure that clicking on any gaps between the label and checkbox still toggles it.
+2. **Individual label**: Place each individual checkbox in `<Field.Item>`, then label it using an enclosing `<Field.Label>`. Enclosing labels ensure that clicking on any gaps between the label and checkbox still toggles it.
```
atomiks, line 54: Curious why (implementation wise) this is necessary? Is there a way to avoid it for the enclosing case?

All the other component docs about labeling would also need to be updated here as they omit `<Field.Item>`

mj12albert: IIRC it's because we changed from `<button>` to spans so even for enclosing labels `htmlFor` needs to be set explicitly

> Is there a way to avoid it for the enclosing case?

Could possibly make `Field.Label` render `Field.Item` internally when necessary? 🤔

> All the other component docs about labeling would also need to be updated here

This is only relevant for checkbox group and radio group though

atomiks: > IIRC it's because we changed from <button> to spans so even for enclosing labels htmlFor needs to be set explicitly

What does the `htmlFor` do though when enclosed? IIRC the input becomes the label focus target when enclosed, which redirects focus to the `<span>`, so it works anyway. And when focus is on the `<span>`, the label text is still announced to screen readers (as it works with native `<label>` as well)

mj12albert: It seems intentional since all the tests were updated to have FieldItem in all cases, perhaps it can be omitted for enclosing FieldLabels as long as it doesn't trigger any chrome errors

Update: it works but does indeed cause `duplicate id` errors in Chrome https://github.com/mui/base-ui/pull/3815#issue-3839131635 @atomiks

atomiks: Alright, let's update the relevant docs to use `<Field.Item>` then

PR #3812, opened 2026-01-21
[all components] Do not memoize the state when not needed

### LINKED ISSUE #3495: [all components] Do not memoize the `state` object before passing it to `useRenderElement()`
We currently memoize the `state` object but never use this memoization  (the data attrs are generated on every render and the render prop is called on every render).

I think we can remove the `useMemo` to slightly improve the performance (skip the deps comparison on every render).

flaviendelangle:
Closes #3495


flaviendelangle:
@greptileai

atomiks (approved):
The only case I can think of memo being used is:

```jsx
<Component 
  render={
    (props, state) => <MemoComp {...props} state={state} />
  } 
/>
```

Though individual props are unmemoized already (notably, event handlers), so I don't think this was relevant to begin with?

flaviendelangle:
> The only case I can think of memo being used is:

Yes, that's the only actual difference in behavior I could find, but I think it's not one we should optimize for

PR #3813, opened 2026-01-21
[docs] Align the tooltip content with trigger aria-label

silviuaavram:

- provided the same content in both tooltip trigger aria-label and tooltip content
- move aria-label from the trigger to the child in the main example
- add role="img' to the svg elements

### DIFF
--- a/docs/src/app/(docs)/react/components/tooltip/demos/detached-triggers-controlled/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/tooltip/demos/detached-triggers-controlled/css-modules/index.tsx
@@ -19,15 +19,15 @@ export default function TooltipDetachedTriggersControlledDemo() {
       <div className={styles.Container}>
         <div className={styles.ButtonGroup}>
           <Tooltip.Trigger className={styles.IconButton} handle={demoTooltip} id="trigger-1">
-            <InfoIcon aria-label="Information 1" className={styles.Icon} />
+            <InfoIcon aria-label="Controlled tooltip" className={styles.Icon} />
           </Tooltip.Trigger>
 
           <Tooltip.Trigger className={styles.IconButton} handle={demoTooltip} id="trigger-2">
-            <InfoIcon aria-label="Information 2" className={styles.Icon} />
+            <InfoIcon aria-label="Controlled tooltip" className={styles.Icon} />
           </Tooltip.Trigger>
 
           <Tooltip.Trigger className={styles.IconButton} handle={demoTooltip} id="trigger-3">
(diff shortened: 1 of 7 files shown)


colmtuite:
We should close this PR as per the discussion earlier.

silviuaavram:
@colmtuite the PR implements the changes as per the latest discussion

docs/src/app/(docs)/react/components/tooltip/demos/detached-triggers-controlled/css-modules/index.tsx
```diff
@@ -19,15 +19,15 @@ export default function TooltipDetachedTriggersControlledDemo() {
       <div className={styles.Container}>
         <div className={styles.ButtonGroup}>
           <Tooltip.Trigger className={styles.IconButton} handle={demoTooltip} id="trigger-1">
-            <InfoIcon aria-label="Information 1" className={styles.Icon} />
+            <InfoIcon role="img" aria-label="Controlled tooltip" className={styles.Icon} />
```
atomiks: I don't think `role="img"` is necessary here? 

I know the AI said it but I don't think it's right:

> The ARIA img role can be used to identify **multiple elements inside page content that should be considered as a single image**. These elements could be images, code snippets, text, emojis, or other content that can be combined to deliver information in a visual manner.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/img_role

silviuaavram: I saw it used in other icon buttons as well, but indeed, it's not a widely used approach. Will remove it.

atomiks: Actually the page has a special section for SVG, which I didn't scroll down to see:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/img_role#svg_and_roleimg

> If you are using embedded SVG images in your page, it is a good idea to set role="img" on the outer [<svg>](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/svg) element and give it a label. This will cause screen readers to just consider it as a single entity and describe it using the label, rather than trying to read out all the child nodes

I didn't notice any issue with VoiceOver without `role="img"`, but other SRs might need it 🤷  - maybe because all our SVGs only have a single node `<path>` so it's not a problem

Though every other demo with an icon button (e.g. Popover) would also need this updated

PR #3815, opened 2026-01-21
Omit field item when implicitly labeling

mj12albert:
https://deploy-preview-3815--base-ui.netlify.app/experiments/forms/button-controls

It works but will trigger Chrome's `Duplicate form field id in the same form` error since the default id needs to be unset in an effect


atomiks:
> It works but will trigger Chrome's Duplicate form field id in the same form error since the default id needs to be unset in an effect

`<Field.Item>` is probably better here then given the IDs will be different during SSR

PR #3817, opened 2026-01-21
[docs] Forms guide visual improvements

DiegoAndai:
The Forms guide could use several visual improvements to make a better impression on users with higher design expectations.

This is a proposal of some things we could improve:

([Preview](https://deploy-preview-3817--base-ui.netlify.app/react/handbook/forms))

- Add required field indicators
- Add error
- Improve descriptions
- Improve layout and sizes, use full-width on input-like fields, use a smaller text size.
- Add error messages for native validation. Without this, the browser shows the default message, which might be in the user's language, but the labels and descriptions will still be in English. This could be confusing, so I think it's better to have the same approach for all text.

| Before | After |
|--------|-------|
| Forms guide without PR changes | Forms guide after PR changes |
| Errors in forms guide without PR changes | Errors in forms guide after PR changes |


DiegoAndai:
Closing as I won't have bandwidth to see this through at this moment. Pinging @aarongarciah in case you want to pick some things up in the future 😊.

PR #3821, opened 2026-01-21
[menu] Fix race conditions

atomiks:
Found a couple of these working on  #3783

1. Sometimes the highlight would remain stuck on an item despite the pointer leaving. `onFocus` gets called after `onPointerLeave` is called with `null` incorrectly.
2. Sometimes a submenu would incorrectly close on mouseleave despite being hovered over already  (where it should remain open).

Both are hard to reproduce; more frequent with CPU slowdown

### DIFF
--- a/packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx
+++ b/packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx
@@ -92,7 +92,9 @@ export const MenuSubmenuTrigger = React.forwardRef(function SubmenuTriggerCompon
   const itemMetadata = React.useMemo(
     () => ({
       type: 'submenu-trigger' as const,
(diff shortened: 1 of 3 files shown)


packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx
```diff
@@ -92,7 +92,9 @@ export const MenuSubmenuTrigger = React.forwardRef(function SubmenuTriggerCompon
   const itemMetadata = React.useMemo(
     () => ({
       type: 'submenu-trigger' as const,
-      setActive: () => parentMenuStore.set('activeIndex', listItem.index),
+      setActive() {
```
michaldudak, line 95: Any particular reason for this change?

atomiks: No, was just testing things. I left it because I prefer the formatting

PR #3823, opened 2026-01-22
[scroll area] Add `data-scrolling` state attribute to `Root` and `Viewport` parts

arturbien:
## Summary

This PR exposes the `data-scrolling` attribute on `ScrollArea.Root` and `ScrollArea.Viewport` components, in addition to the existing implementation on `ScrollArea.Scrollbar`.

## Motivation

Previously, `data-scrolling` was only available on the Scrollbar component. However, many common use cases require detecting scroll state on the Root or Viewport to apply styles or disable interactions on ScrollArea children during scroll via CSS.

#### 1. Disable pointer events during scroll to prevent overfetching

When you have a list of items that prefetch data on hover, scrolling through the list can trigger many unnecessary prefetch requests. Disabling pointer events during scroll solves this:

```css
.Viewport[data-scrolling] .ListItem {
  pointer-events: none;
}
```
```jsx
<ScrollArea.Root>
  <ScrollArea.Viewport className="Viewport">
    {items.map((item) => (
      <div 
        key={item.id} 
        className="ListItem"
        onMouseEnter={() => prefetch(item.id)}
      >
        {item.name}
      </div>
    ))}
  </ScrollArea.Viewport>
</ScrollArea.Root>
```

https://github.com/user-attachments/assets/c9295b93-5820-43da-b7d9-91da5bce5420

#### 2. Hiding Tooltips on scroll:

https://github.com/user-attachments/assets/971cdbab-84e8-4404-b2fa-59c5ee05ce28

### Changes
- Added scrolling to ScrollAreaRootState and ScrollAreaViewportState
- Root and Viewport now compute scrolling: scrollingX || scrollingY
- Added data-scrolling to ScrollAreaRootDataAttributes
- Updated documentation reference files
- Added tests for the new attribute

### Test Plan
- [x] pnpm typescript passes
- [x] pnpm eslint passes
- [x] pnpm prettier passes
- [x] Added tests in ScrollAreaRoot.test.tsx for Root's data-scrolling
- [x] Added tests in ScrollAreaViewport.test.tsx for Viewport's data-scrolling
- [x] Existing ScrollArea tests pass

### DIFF
--- a/docs/src/app/(docs)/react/components/page.mdx
+++ b/docs/src/app/(docs)/react/components/page.mdx
@@ -1356,6 +1356,10 @@ A popup that appears when an element is hovered or focused, showing a hint for s
 - Sections:
   - Usage guidelines
   - Anatomy
+  - Alternatives to tooltips
+    - Infotips
+    - Description text
+    - Contextual feedback messages
   - Examples
     - Detached triggers
     - Multiple triggers
--- a/docs/src/app/(docs)/react/components/tooltip/page.mdx
+++ b/docs/src/app/(docs)/react/components/tooltip/page.mdx
@@ -14,12 +14,8 @@ import { DemoTooltipHero } from './demos/hero';
 
 ## Usage guidelines
 
-To ensure that tooltips are accessible and helpful, follow these guidelines:
-
-- **Provide an accessible name for the trigger**: The tooltip's trigger must have a meaningful label. This can be its visible text or an `aria-label`/`aria-labelledby` attribute. The label should closely match the tooltip's content to ensure consistency for screen reader users.
-- **Avoid tooltips for critical information**: Tooltips work well for enhancing UI clarity (like labeling icon buttons) but should not be the sole means of conveying important information. Since tooltips do not appear on touch devices, consider using a [Popover](/react/components/popover) for essential content.
-- **Avoid tooltips for "infotips"**: If your tooltip is attached to an "info icon" button whose only purpose is to show the tooltip, opt for [Popover](/react/components/popover) and add the `openOnHover` prop instead. Tooltips should describe an element that performs an action separate from opening the tooltip itself.
-- **Avoid for transient information**: Use the Toast component's [anchoring ability](/react/components/toast#anchored-toasts) instead of Tooltip for transient notifications, like a "Copied" popup that appears above a copy-to-clipboard button. This way, the message is announced to screen readers when it appears. Toasts can also contain more complex content, such as links or actions.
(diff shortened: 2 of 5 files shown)


atomiks:
I wonder if this CSS feels too aggressive in terms of blocking pointers after a scroll with the default debounce (`500ms`)

```css
.Viewport[data-scrolling] .ListItem {
  pointer-events: none;
}
```

If you scroll, abruptly stop, then move your pointer, it's temporarily (admittedly a short period) going to prevent the UI from being interactive. It's somewhat minor, but for fast interactions it could be an annoyance. Have you found that in your testing?

arturbien:
> I wonder if this CSS feels too aggressive in terms of blocking pointers after a scroll with the default debounce (`500ms`)
> 
> ```css
> .Viewport[data-scrolling] .ListItem {
>   pointer-events: none;
> }
> ```
> 
> If you scroll, abruptly stop, then move your pointer, it's temporarily (admittedly a short period) going to prevent the UI from being interactive. It's somewhat minor, but for fast interactions it could be an annoyance. Have you found that in your testing?

@atomiks  Not really, we're using this approach in multiple places in our production app with lots of users and never got any complaints. 

But there's different way of doing this via MutationObserver like:
- observe ScrollArea.Root `data-scrolling` attribute and update the scrolling state changes in `React.useRef`
- `onMouseEnter={() => isScrollingRef.current ? undefined : prefetch}`

This way interactions will not be blocked at all.

Additionally I have another PR coming that implements the `scrollend` event in ScrollArea in supported browsers. With this there will be no 500ms delay

But then again, its just one of the use cases but there's more like styling the scroll area based on the scrolling state. In which case having the attribute directly on the Root and Viewport is better than doing complex selectors like  `.Root:has(.Scrollbar[data-scrolling]) {...}` 

Another use case is disabling expensive hover effects in list items like animating box shadow when scroll is in progress etc

arturbien:
@atomiks please lmk if we can roll with it 🫡

packages/react/src/scroll-area/root/ScrollAreaRoot.test.tsx
```diff
@@ -4,20 +4,59 @@ import { createRenderer, isJSDOM } from '#test-utils';
...
 const VIEWPORT_SIZE = 200;
 const SCROLLABLE_CONTENT_SIZE = 1000;
 const SCROLLBAR_WIDTH = 10;
 const SCROLLBAR_HEIGHT = 10;
 
 describe('<ScrollArea.Root />', () => {
-  const { render } = createRenderer();
+  const { render, clock } = createRenderer();
+
+  clock.withFakeTimers();
```
atomiks: This breaks some browser tests it seems; you can nest a describe block to create a new renderer with fake timeouts, I believe

arturbien: should be fixed now, but circleci jobs are not running, they're stuck at `Waiting for status to be reported`

PR #3824, opened 2026-01-22
[combobox] Avoid clearing selected value if item is not present in items array

### LINKED ISSUE #3502: [combobox] `Combobox.Input` value reset on re-renders
# Bug report

## Current behavior

At `Combobox.Root` with non-memoized `items` and `value` set to empty string, `Combobox.Input` value is getting reset during re-renders.

https://github.com/user-attachments/assets/8d7e3945-2fd8-4e2e-b431-5b2bfbb8ddcb

## Expected behavior

Input value should remain untouched?

## Reproducible example

[StackBlitz](https://stackblitz.com/edit/uwnkipbe)

## Base UI version

1.0.0-rc.0

## Additional context

Hard to tell exactly what happens and under what circumstances, maybe something related to [input value sync](https://github.com/mui/base-ui/blob/def18afb2f1776444837f8b23d5e363a84e1fb62/packages/react/src/combobox/root/AriaCombobox.tsx#L880)?

### LINKED ISSUE #3818: [combobox] Value gets cleared when items change
# Bug report

## Current behavior

I believe https://github.com/mui/base-ui/pull/3607 has caused a regression for a component I have built, but this should affect other people as well.

I have a `Combobox` where `items` are fed by an API. When you choose a value, and then search for something else, the value gets cleared (`onValueChange` is called with `null`) and `onInputValueChange` is called if the items do not contain selected value.

This will also fail, if you were coming back to editing a form after saving and refreshing and your data is fed by a paginated API, where your selected item might be on the last page. The value will be cleared because the selected item will not be present in `items`.

https://github.com/user-attachments/assets/30e3f565-4fe2-48df-b631-3d06ff9d0005

This issue is not reproducible in v1.0.0.

I believe this is a valid use case and should be supported.

## Expected behavior

The value should not be cleared. 

## Reproducible example

I have rebuilt this example from https://github.com/mui/base-ui/pull/2760

https://codesandbox.io/p/sandbox/gallant-meadow-x2r7d2

## Base UI version

v1.1.0

atomiks:

This avoids surprises with `value` getting cleared (to `null`) when the selected item isn't present in `items`, particularly in async combobox implementations. Native `<select>`s [retain the previous](https://codesandbox.io/p/sandbox/rkh4fm) `value` in state when the `<option>` gets removed from the list, with the exception that it automatically selects a fallback value which is a bit confusing here

Fixes #3818
Fixes #3502


michaldudak (approved):

atomiks:
@sarthakmalik0810 async items when the selected value wasn't in the `items` list was already broken for uncontrolled comboboxes, the logic previously was just to sync controlled ones afaict

We're sticking to monthly releases except in critical cases, you can use 1.0.0 or the [canary release](https://github.com/mui/base-ui/pull/3824#issuecomment-3783036739) in the meantime

PR #3846, opened 2026-01-23
[dialog][alert dialog][popover] Fix focus management for children with positive tabindex

### LINKED ISSUE #3844: [Dialogs] focus trap breaks when dialog contains an element with tabIndex greater than 0
Visit [StackBlitz](https://stackblitz.com/edit/otqpkvu3?file=src%2FApp.tsx) to see the reproduction. Open the dialog and press 'Tab' twice; you should find yourself outside the focus boundaries. this happens when the dialog contains an element with tabIndex greater than 0.

I tried to take advantage of this behavior to have a better workaround for [this](https://github.com/mui/base-ui/issues/3750) issue, with no luck. Perhaps both can be resolved at the same time?

michaldudak:
Fixed modal focus trapping when Dialogs or Popovers contain positive tabIndex elements by adding manual <kbd>Tab</kbd> cycling in FloatingFocusManager.

Fixes #3844

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
@@ -359,13 +359,37 @@ export function FloatingFocusManager(props: FloatingFocusManagerProps): React.JS
 
     function onKeyDown(event: KeyboardEvent) {
       if (event.key === 'Tab') {
+        const activeEl = activeElement(getDocument(floatingFocusElement));
         // The focus guards have nothing to focus, so we need to stop the event.
         if (
-          contains(floatingFocusElement, activeElement(getDocument(floatingFocusElement))) &&
+          contains(floatingFocusElement, activeEl) &&
           getTabbableContent().length === 0 &&
           !isUntrappedTypeableCombobox
         ) {
           stopEvent(event);
+        }
+
+        const tabbableContent = getTabbableContent() as Array<FocusableElement>;
+        if (
+          contains(floatingFocusElement, activeEl) &&
+          !isUntrappedTypeableCombobox &&
+          tabbableContent.some((element) => element.tabIndex > 0)
+        ) {
+          // Positive tabIndex content can jump ahead of the focus guards in the browser's
+          // global tab order, so manually cycle within the floating element to preserve
+          // the focus trap.
+          stopEvent(event);
+          const activeIndex = tabbableContent.indexOf(activeEl as FocusableElement);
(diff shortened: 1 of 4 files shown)


packages/react/src/popover/root/PopoverRoot.test.tsx
```diff
@@ -990,6 +990,51 @@ describe('<Popover.Root />', () => {
...
       });
 
+      it('keeps focus trapped when content includes a positive tabIndex', async () => {
+        function Test() {
+          const priorityRef = React.useRef<HTMLButtonElement>(null);
+
+          return (
+            <div>
+              <TestPopover
+                rootProps={{ modal: 'trap-focus' }}
```
mj12albert, line 1000: Is it expected that for popover this doesn't work with `modal={true}` like dialog? https://stackblitz.com/edit/nht123vz?file=src%2FApp.tsx

They are documented the same: `true: user interaction is limited to the popover`

packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
```diff
@@ -359,14 +359,38 @@ export function FloatingFocusManager(props: FloatingFocusManagerProps): React.JS
...
+
+        const tabbableContent = getTabbableContent() as Array<FocusableElement>;
+        if (
+          contains(floatingFocusElement, activeEl) &&
+          !isUntrappedTypeableCombobox &&
+          tabbableContent.some((element) => element.tabIndex > 0)
+        ) {
+          // Positive tabIndex content can jump ahead of the focus guards in the browser's
+          // global tab order, so manually cycle within the floating element to preserve
+          // the focus trap.
```
atomiks, line 380: Does Radix handle this? imo if it doesn't, let's leave it off. It adds extra cost to handle an edge case  that's bad to begin with (you should never use positive tab indices)

packages/react/src/popover/root/PopoverRoot.test.tsx
michaldudak: Yes, it's deliberate.

packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
michaldudak: No, they explicitly ignore it:
https://github.com/radix-ui/primitives/blob/main/packages/react/focus-scope/src/focus-scope.tsx#L259
I suppose you're right, the use case is quite exotic anyway.

PR #3852, opened 2026-01-24
add tooltip role and aria-describedby to Tooltip for screenreaders

bremmdev:
- [x ] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

Used the built-in `useRole` hookt to add role of 'tooltip' to the Tooltip Content so the contents of the popup are announced by the screenreader when focusing the TooltipTrigger using aria-describedby

### DIFF
--- a/packages/react/src/tooltip/root/TooltipRoot.test.tsx
+++ b/packages/react/src/tooltip/root/TooltipRoot.test.tsx
@@ -41,6 +41,38 @@ describe('<Tooltip.Root />', () => {
     { name: 'detached triggers', Component: DetachedTriggerTooltip },
     { name: 'multiple detached triggers', Component: MultipleDetachedTriggersTooltip },
   ])('when using $name', ({ Component: TestTooltip }) => {
+    describe('ARIA attributes', () => {
+      clock.withFakeTimers();
+
+      it('should have role="tooltip" on the popup', async () => {
+        await render(<TestTooltip rootProps={{ open: true }} />);
+
+        expect(screen.getByRole('tooltip')).not.to.equal(null);
+      });
+
+      it('should have aria-describedby on the trigger pointing to the popup', async () => {
+        await render(
+          <TestTooltip
+            rootProps={{ open: true, triggerId: 'test-trigger' }}
+            triggerProps={{ id: 'test-trigger' }}
+          />,
+        );
+
+        const trigger = screen.getByRole('button', { name: 'Toggle' });
+        const popup = screen.getByRole('tooltip');
+
+        expect(trigger.getAttribute('aria-describedby')).to.equal(popup.id);
+      });
+
+      it('should not have aria-describedby on the trigger when closed', async () => {
+        await render(<TestTooltip />);
+
+        const trigger = screen.getByRole('button', { name: 'Toggle' });
+
+        expect(trigger.getAttribute('aria-describedby')).to.equal(null);
+      });
+    });
+
     describe('uncontrolled open', () => {
       clock.withFakeTimers();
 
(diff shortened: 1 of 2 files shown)


atomiks:
Hi @bremmdev, we intentionally made tooltips visual-only (inaccessible to screen readers). From our [usage guidelines](https://deploy-preview-3852--base-ui.netlify.app/react/components/tooltip#usage-guidelines):

> Provide an accessible name for the trigger: **The tooltip’s trigger must have a meaningful label. This can be its visible text or an aria-label/aria-labelledby attribute.** The label should closely match the tooltip’s content to ensure consistency for screen reader users.

In other words, you must ensure the trigger's `aria-label` matches the tooltip content. 

If your tooltip is trying to be a "description" that aims to further describe the trigger/label, that is a design mistake (if the info is critical/important) and should be an infotip (`Popover` + `openOnHover`), because tooltips aren't accessible to touch users.

We treat tooltips as visual-only enhancements for sighted users that can safely be omitted at any point. If touch users cannot see them, screen reader users on desktop must also be okay with not seeing them.

---

If you want, you can change this PR to be about amending the docs with examples or independent sections on the docs describing this philosophy to users to be clearer

bremmdev:
> Hi @bremmdev, we intentionally made tooltips visual-only (inaccessible to screen readers). From our [usage guidelines](https://deploy-preview-3852--base-ui.netlify.app/react/components/tooltip#usage-guidelines):
> 
> > Provide an accessible name for the trigger: **The tooltip’s trigger must have a meaningful label. This can be its visible text or an aria-label/aria-labelledby attribute.** The label should closely match the tooltip’s content to ensure consistency for screen reader users.
> 
> In other words, you must ensure the trigger's `aria-label` matches the tooltip content.
> 
> If your tooltip is trying to be a "description" that aims to further describe the trigger/label, that is a design mistake (if the info is critical/important) and should be an infotip (`Popover` + `openOnHover`), because tooltips aren't accessible to touch users.
> 
> We treat tooltips as visual-only enhancements for sighted users that can safely be omitted at any point. If touch users cannot see them, screen reader users on desktop must also be okay with not seeing them.
> 
> If you want, you can change this PR to be about amending the docs with examples or independent sections on the docs describing this philosophy to users to be clearer

In your solution, you let the developer take care of matching the aria-label with the tooltip content. With the solution I propose, this is guaranteed by the role and the aria-describedby. Aria-describedby is meant as a direct description for the trigger, not meant as '"description" that aims to further describe the trigger/label'. This also follows this W3C pattern: https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/

This is also automatically more inclusive in my optinion.

atomiks:
> Aria-describedby is meant as a direct description for the trigger, not meant as '"description" that aims to further describe the trigger/label'

Tooltips are visual labels not descriptions. `aria-describedby` is not guaranteed to be announced by screen readers in certain contexts, and is delayed compared to `aria-label`. `aria-labelledby` also doesn't work consistently across screen readers if the element is conditionally rendered (which tooltips are by default, and for perf reasons), so `aria-label` is the only viable solution.

Our recommendation is that your wrapper component around `Tooltip` synchronizes the content of the tooltip and the `aria-label` attribute on the trigger:

```jsx
<Tooltip label="My tooltip">
  <button />
</Tooltip>
```

Produces this React tree:

```jsx
<Tooltip.Root>
  <Tooltip.Trigger aria-label="My tooltip" />
  <Tooltip.Portal>
     <etc>
       My tooltip
     </etc>
  </Tooltip.Portal>
</Tooltip.Root>
```

The APG page admits it's still in progress without consensus:

> NOTE: This design pattern is work in progress; it does not yet have task force consensus. Progress and discussions are captured in [issue 128.](https://github.com/w3c/aria-practices/issues/128)

We thought a lot about what tooltips are in UIs, how they should be used etc., and we came to this conclusion. We didn't follow the ARIA guide because it's not correct in our view

PR #3854, opened 2026-01-24
[useScrollLock] Reduce style recalculation with classic scrollbars

mdm317:

## Current
When using classic (inset) scrollbars with scroll locking, full document recalc can occur.

### [Chromium 143]
 3 full document recalc 
Screenshot 2026-01-24 at 11 08 39 PM

### [Chromium 144] 
1 full document recalc
Screenshot 2026-01-24 at 11 06 22 PM

## Cause
Directly changing `<html>` styles causes a full document style recalculation.

## Solution
When checking scrollbar gutter stable, use body by default, except in cases where overflow-y: scroll is set on `<html>`.

## Test
- Verified scroll locking works correctly.
  - [Preview link](https://deploy-preview-3854--base-ui.netlify.app/experiments/scroll-lock)
    - should check both classic and overlay scrollbars.

### [Chromium 143]
1 full document recalc (Because we shoud set scrollbar-gutter in `html`)

### [Chromium 144] 
0 full document recalc
Screenshot 2026-01-25 at 11 14 11 AM

## Limitation
Improving this behavior is not feasible for users who set overflow-y: scroll on `<html>`.


packages/utils/src/useScrollLock.ts
```diff
@@ -30,20 +30,24 @@ function supportsStableScrollbarGutter(referenceElement: Element | null) {
...
+  const scrollContainer = isOverflowElement(html) ? html : body;
 
   const originalStyles = {
-    scrollbarGutter: html.style.scrollbarGutter,
-    overflowY: html.style.overflowY,
+    scrollbarGutter: scrollContainer.style.scrollbarGutter,
+    overflowY: scrollContainer.style.overflowY,
   };
 
   html.style.scrollbarGutter = 'stable';
```
mdm317, line 40: `scrollbar-gutter` works as expected on the html element.
check [blog](https://www.matuzo.at/blog/2022/100daysof-day25)  [css-overflow-3-spec](https://www.w3.org/TR/css-overflow-3)

docs/src/app/(private)/experiments/scroll-lock.tsx
```diff
@@ -6,15 +6,20 @@ import { isWebKit } from '@base-ui/utils/detectBrowser';
 export default function ScrollLock() {
   const [enabled, setEnabled] = React.useState(false);
   const [bodyScrollY, setBodyScrollY] = React.useState(false);
+  const [htmlScrollY, setHtmlScrollY] = React.useState(false);
```
mdm317, line 9: Add `htmlScrollY` to ensure scroll blocking works correctly when `overflow-y: scroll` is set on <html>.

atomiks:
@greptileai

packages/utils/src/useScrollLock.ts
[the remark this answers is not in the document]
mdm317: I've fixed it.

atomiks:
In the demo when `Long content` is unchecked, the scrollbar disappears (differing from master) - enabling scroll lock then shifts the content

mdm317:
I missed that case — thanks for pointing it out. 😔

--- 

> In the demo when Long content is unchecked, the scrollbar disappears

This difference comes from how the demo is configured:
 - In the demo(master), `overflow-y: scroll` is applied via a class on the `html` element.
 - In the demo(current), `overflow-y: scroll` is instead added as a class on the `body`.
 - Also, for testing purposes, when neither body nor html has an overflow property, i temporarily set `overflow-y: visible`  inline style.  [diff](https://github.com/mui/base-ui/pull/3854/changes#diff-44eb40aabe6e1e54cbc7315f753c8b23ac59446236922dd1211f5a5af282c539R16)

This happens in the demo(master) as well if you manually disable `overflow-y: scroll` on `html`.
So when both `body` and `html` have no `overflow: scroll` and the `Long content` is unchecked, there is no scrollbar.

--- 

> enabling scroll lock then shifts the content

The layout shift issue occurs when `preventScrollOverlayScrollbars` is used.
When scroll lock is released, the previously set inline `overflow-y: visible` is not restored correctly, which results shift.

I fixed this issue in this [commit](https://github.com/mui/base-ui/pull/3854/changes/13a5f27f897d013efff6ddcdece7daa07d670191%E3%85%81).

So in the current demo, when`Long content` is unchecked and neither body nor `html` has `overflow: scroll` set, the scrollbar disappears.
Unlocking scroll lock also keeps the scrollbar hidden, so the content doesn’t shift.

atomiks (approved):
Nice, wasn't able to find any new issues with the latest implementation

atomiks:
@greptileai

PR #2988, opened 2025-10-16
[docs][slider] Fix a11y issues in demos

sai6855:
Closes #2545

Preview: https://deploy-preview-2988--base-ui.netlify.app/react/components/slider

### Before
image

### After
image

PR #3857, opened 2026-01-25
[docs][slider] Add visible labels and value display to demos for a11y best practices 

### LINKED ISSUE #2545: [slider] Demo should show a11y best practice
## Current behavior

Slider demo doesn't respect a11y best practices. 

## Expected behavior

It does.

## Reproducible example

1. Open https://base-ui.com/react/components/slider
2. See 

Image

## Additional context

For comparison, none of those alternatives have this issue:

- https://mui.com/material-ui/react-slider/#accessibility
- https://radix-ui.com/docs/primitives/components/slider
- https://react-spectrum.adobe.com/react-aria/Slider.html

CiscoFran10:
Addresses accessibility best practices in Slider demos by adding visible labels and value displays, aligning with patterns used by peer libraries (MUI, Radix UI, React Spectrum).

Preview: https://deploy-preview-3857--base-ui.netlify.app/react/components/slider

**Changes:**
- Add visible `<label>` elements with proper `aria-labelledby` association to all Slider demos
- Add `Slider.Value` component to display current value(s)
- Use `React.useId()` for SSR-safe unique ID generation
- Add `aria-label` to range slider thumbs for screen reader distinction ("Minimum price", "Maximum price")
- Update CSS to use consistent grid layout matching Progress/Meter demo patterns

## Accessibility

Lighthouse audit confirms full accessibility compliance after changes:

Captura de tela 2026-01-25 005455

## Reviewer Notes

Following @mj12albert guidance, I aligned this PR with the preferred pattern of using visible labels and `Slider.Value` rather than relying on `aria-label` as the primary accessible name mechanism, consistent with the Combobox demos and the documented accName guidelines.

**Open questions for discussion:**

1. Should visible labels use `<label>` elements or neutral elements (e.g., `<span>`) paired with `aria-labelledby`, given that `Slider.Root` is not a native form control?

2. Should `Slider.Value` be marked `aria-hidden="true"` to avoid potential duplicate announcements?

## Related

Closes #2545
Related #2988

## References

- [MUI Slider Accessibility](https://mui.com/material-ui/react-slider/#accessibility)
- [Radix Slider](https://radix-ui.com/docs/primitives/components/slider)
- [React Aria Slider](https://react-spectrum.adobe.com/react-aria/Slider.html)
- [WAI-ARIA Slider Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/slider-multithumb/)

### DIFF
--- a/packages/react/src/drawer/root/DrawerRoot.tsx
+++ b/packages/react/src/drawer/root/DrawerRoot.tsx
@@ -4,6 +4,7 @@ import { useControlled } from '@base-ui/utils/useControlled';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { ownerWindow } from '@base-ui/utils/owner';
 import { isAndroid } from '@base-ui/utils/detectBrowser';
+import { useId } from '@base-ui/utils/useId';
 import {
   DrawerRootContext,
   type DrawerNestedSwipeProgressStore,
@@ -414,7 +415,7 @@ function createNestedSwipeProgressStore(): NestedSwipeProgressStore {
 }
 
 function DrawerProviderReporter() {
-  const drawerId = React.useId();
+  const drawerId = useId();
 
   const providerContext = useDrawerProviderContext(true);
   const dialogRootContext = useDialogRootContext(false);
@@ -426,7 +427,7 @@ function DrawerProviderReporter() {
   const isTopmost = nestedOpenDialogCount === 0;
 
   React.useEffect(() => {
-    if (!providerContext) {
+    if (!providerContext || drawerId == null) {
       return undefined;
     }
 
@@ -436,6 +437,10 @@ function DrawerProviderReporter() {
   }, [drawerId, providerContext]);
 
   React.useEffect(() => {
+    if (drawerId == null) {
+      return;
+    }
+
     providerContext?.setDrawerOpen(drawerId, open);
   }, [drawerId, open, providerContext]);
 


michaldudak:
This seems to be a duplicate of #2988

@mj12albert, could you please verify?

CiscoFran10:
> This seems to be a duplicate of #2988
> 
> @mj12albert, could you please verify?

My PR follows a different approach suggested by @mj12albert in that thread. Since there were no updates from the original author after his comment in Nov 2025, I opened a new PR implementing the solution he proposed there.

> I think `aria-label` is the least-preferred way for setting accNames; the best way would probably to add `Slider.Value` to the demos, and use a plain `<label>` like the Combobox demos, what do you think @colmtuite
> 
> (There's already a `Usage guideline` now about accNames and a link to a fully labelled example)

mj12albert:
Closed https://github.com/mui/base-ui/pull/3857 as it's a bit stale

docs/src/app/(docs)/react/components/slider/demos/hero/css-modules/index.tsx
```diff
@@ -1,13 +1,20 @@
...
+import * as React from 'react';
 import { Slider } from '@base-ui/react/slider';
 import styles from './index.module.css';
 
 export default function ExampleSlider() {
+  const id = React.useId();
   return (
-    <Slider.Root defaultValue={25}>
+    <Slider.Root defaultValue={25} className={styles.Root} aria-labelledby={id}>
+      <label id={id} className={styles.Label}>
```
mj12albert: We don't recommend a plain `<label>` per the doc as it causes a `A <label> isn't associated with a form field` error in Chrome: https://github.com/mui/base-ui/blob/ad6f7fd8c2fe784a55b827388c1555cceaef4485/docs/src/app/(docs)/react/components/slider/page.mdx?plain=1#L75

Would be good to use `Field.Label` here because we don't show it anywhere else, it's able to focus the thumb when clicked and also avoids text selection on double click

CiscoFran10: @mj12albert  is this pattern acceptable for the hero demos ??

```ts
import { Slider } from '@base-ui/react/slider';
import { Field } from '@base-ui/react/field';
import styles from './index.module.css';

export default function ExampleSlider() {
  return (
    <Field.Root render={<Slider.Root defaultValue={25} className={styles.Root} />}>
      <Field.Label className={styles.Label}>Volume</Field.Label>
      <Slider.Value className={styles.Value} />

      <Slider.Control className={styles.Control}>
        <Slider.Track className={styles.Track}>
          <Slider.Indicator className={styles.Indicator} />
          <Slider.Thumb className={styles.Thumb} />
        </Slider.Track>
      </Slider.Control>
    </Field.Root>
  );
}
```

mj12albert: No need to use the render prop ~ FieldRoot should enclose SliderRoot

CiscoFran10: @mj12albert I was trying to avoid the additional wrapper `<div>` that gets rendered when `Field.Root` encloses `Slider.Root`. We end up with an extra DOM node:

```html
<div>
  <div data-orientation="horizontal" role="group" ...>
    ...
  </div>
</div>
```

My concern was just keeping the markup as minimal as possible for the hero demos.
That said, if you’re okay with this structure and the extra wrapper is expected here, I’m good with that as well.

mj12albert:
Apologies for the wait ~ we decided to add `Slider.Label` in https://github.com/mui/base-ui/pull/4167 and avoid using Field in the non-field/form demos

CiscoFran10:
> Apologies for the wait ~ we decided to add `Slider.Label` in #4167 and avoid using Field in the non-field/form demos

No problem, I’ll close this in favor of #4167

Issue #3850, opened 2026-01-23
[menu] `Menu.SubmenuTrigger` that renders a disabled button is still enabled


benface:
# Bug report

## Current behavior

There is an inconsistency between `Menu.Trigger` and `Menu.SubmenuTrigger`:

- `<Menu.Trigger render={<button type="button" disabled>Open menu</button> />` doesn't open the menu when clicked, thanks to the `disabled` attribute on the rendered `<button>`.
- `<Menu.SubmenuTrigger nativeButton={true} render={<button type="button" disabled>Open submenu</button> />` still opens the submenu when clicked, despite the `disabled` attribute.

## Expected behavior

A `Menu.SubmenuTrigger` that renders a `disabled` button should not open the submenu when clicked.

## Reproducible example

https://stackblitz.com/edit/aktrg8j2?file=src%2FApp.tsx

## Base UI version

v1.1.0

## Which browser are you using?

Chrome/Safari/Firefox

## Which OS are you using?

macOS

## Which assistive tech are you using (if applicable)?

None

## Additional context

Related to #1976

CiscoFran10:
There is currently a workaround for this case: instead of relying on the `disabled` attribute on the rendered `<button>`, you can pass `disabled` directly to `Menu.SubmenuTrigger` itself, and then style the trigger using the `data-disabled` attribute that Base UI applies.

Example:

Demo: https://stackblitz.com/edit/aktrg8j2-zzhknf6y?file=src%2FApp.tsx

```tsx
<Menu.SubmenuTrigger
  nativeButton
  disabled
  render={<button type="button" />}
>
  Open submenu
</Menu.SubmenuTrigger>
```

Then style based on `data-[disabled]`:

```css

&[data-disabled][data-highlighted]::before {
  background-color: var(--color-gray-300);
}

&[data-disabled] {
  color: var(--color-gray-400);
}
```



That said, I’ve been exploring a possible fix as a contributor so that `Menu.SubmenuTrigger` can also derive `disabled` from the rendered element.

I’ll share a PR if it gets approved internally, until then, using the `disabled` prop on `Menu.SubmenuTrigger` directly is the recommended approach.

PR #3858, opened 2026-01-25
[menu] Fix SubmenuTrigger not respecting disabled state from render prop

CiscoFran10:
Fixes an issue where `Menu.SubmenuTrigger` did not respect the `disabled` attribute when a disabled button was provided via the `render` prop. Also adds visual disabled styles to Menu demos to match existing patterns in experiments components.

## Changes

- Fixed `Menu.SubmenuTrigger` not detecting the `disabled` state when a disabled button is provided via the `render` prop (e.g., `render={<button disabled />}`). The component now correctly derives the disabled state from the rendered element and applies it to submenu trigger behavior.
- Added disabled styles to all Menu demo components (CSS Modules and Tailwind CSS) to provide visual feedback for disabled menu items, following the existing pattern used in the experiments components.

## Problem

When using `Menu.SubmenuTrigger` with a render prop that contains a disabled button:

```tsx
<Menu.SubmenuTrigger
  nativeButton
  render={<button type="button" disabled />}
>
  Open submenu
</Menu.SubmenuTrigger>
````

The disabled state was not detected, causing the submenu to still open via mouse hover, even though the underlying button was disabled.

Additionally, Menu demos lacked disabled styling. There was no visual distinction between enabled and disabled menu items when highlighted.

**Before**

Before - no visual difference for disabled items

**After**

After - clear visual indication for disabled items

Added `data-[disabled]` styles to all Menu demos.

**CSS Modules**

```css
&[data-disabled][data-highlighted]::before {
  background-color: var(--color-gray-300);
}

&[data-disabled] {
  color: var(--color-gray-400);
}
```

**Tailwind CSS**

```txt
data-[disabled]:text-gray-400
data-[disabled]:data-[highlighted]:before:bg-gray-300
```

## Reviewer Notes

This follows the existing pattern used by experiments components where disabled state is expressed via `data-[disabled]` and styling is driven from that attribute.

Although there is a possible workaround today (passing `disabled` directly to `Menu.SubmenuTrigger`), this change ensures the render-prop pattern behaves consistently with developer expectations.

## Tests

`packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.test.tsx`

* Renders a disabled submenu trigger with correct attributes (`data-disabled`, `aria-disabled`)
* Does not open submenu on click when the render prop contains a disabled button
* Does not open submenu via keyboard (ArrowRight / ArrowLeft) when disabled, in both LTR and RTL directions

## Related

Experiment Path: docs/src/app/(private)/experiments/menu/menu-disabled-submenu-trigger.tsx
Demo with Error: https://stackblitz.com/edit/aktrg8j2?file=src%2FApp.tsx

Fixes https://github.com/mui/base-ui/issues/3850


packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx
```diff
@@ -40,6 +40,12 @@ export const MenuSubmenuTrigger = React.forwardRef(function SubmenuTriggerCompon
     ...elementProps
   } = componentProps;
 
+  const renderElementDisabled = Boolean(
+    render &&
+    React.isValidElement<{ disabled?: boolean | undefined }>(render) &&
+    render.props.disabled,
+  );
```
oliviertassinari: This should probably be done the other way around: if the trigger is DOM disabled, have a warning logic in dev mode to detect it and ask developers to move the logic to the parent.

I don't see how we could do React tree introspection, the Base UI component wouldn't even know if that "disabled" prop defined user land maps to an actual disabled behavior, or something else that the user has defined.

atomiks: This also doesn't support the function form of `render`. It seems like the event handlers in `useMenuItem` could inspect the disabled state of the ref to prevent themselves from executing instead, as we can't determine it in the render body of the component

CiscoFran10: Thanks for the feedback! I've implemented the approach of checking the disabled state via refs in event handlers, and also a warning when the trigger resolves to a DOM-disabled element, suggesting that the disabled state be moved to the Base UI component rather than being applied only at the rendered element level.

The core issue (submenu opening on hover while disabled) is addressed by the guard in `useHoverReferenceInteraction.ts`. I also added disabled checks inside `useMenuItem.ts`, which seemed conceptually reasonable, but I’m not fully confident they are strictly necessary.

Do you consider the additional guards in `useMenuItem.ts` valuable, or should they be removed to keep the fix minimal and scoped solely to preventing submenu activation?

michaldudak: I'm in favor of removing all checks in production mode. I'm also not very keen on development-time warnings. Disabling a Base UI component's logic should be separate from rendering a custom component (so we can, for example, render Tooltips on disabled buttons).

CiscoFran10: @michaldudak To make sure I’m aligning the PR with this direction, could you clarify what you’d like to see as the concrete outcome?

Based on your comment, should I:

1. Remove **all** DOM-disabled checks from event handlers (including the one in `useHoverReferenceInteraction`) and rely solely on the component-level `disabled` state?
2. Remove the dev-time warning in `MenuSubmenuTrigger` entirely?
3. Keep *any* minimal guard to prevent submenu opening when a trigger resolves to a disabled DOM element, or is that behavior intentionally unsupported?


Regarding
> Disabling a Base UI component's logic should be separate from rendering a custom component (so we can, for example, render Tooltips on disabled buttons).

From an accessibility standpoint, most guidance suggests that tooltips should only be attached to focusable/hoverable elements; disabled buttons and plain text are not focusable, which makes tooltip content unreachable for keyboard and screen reader users:

https://github.com/adobe/react-spectrum/discussions/9232

https://github.com/adobe/react-spectrum/discussions/7038

https://react-spectrum.adobe.com/v3/Tooltip.html#usage-on-disabled-or-non-interactive-elements

https://mui.com/material-ui/react-tooltip/#disabled-elements

A common workaround is to render a focusable wrapper but i also found this approach to work:

```ts
<Tooltip.Trigger
  render={(props) => (
    <Button focusableWhenDisabled {...props} disabled>
      <BoldIcon aria-label="Bold" />
    </Button>
  )}
/>
```

(StackBlitz: https://stackblitz.com/edit/ljb99a3s?file=src%2FApp.tsx)

michaldudak: My point is that I'm pretty fine with the current state of things. We *may* show a warning in dev for submenu triggers, but not for other components.

> From an accessibility standpoint, most guidance suggests that tooltips should only be attached to focusable/hoverable elements; disabled buttons and plain text are not focusable, which makes tooltip content unreachable for keyboard and screen reader users.

We believe tooltips are inherently inaccessible and should be visual-only components.
BTW, you can make a disabled Button focusable using the [`focusableWhenDisabled`](https://base-ui.com/react/components/button#Button-focusableWhenDisabled) prop.

CiscoFran10: Thnks, I reverted the submenu changes. But I kept the dev warning, the disabled state styles in the demos and some of the tests for the disabled submenu trigger.

packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx
```diff
@@ -77,6 +80,19 @@ export const MenuSubmenuTrigger = React.forwardRef(function SubmenuTriggerCompon
     [store],
   );
 
+  if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useIsoLayoutEffect(() => {
+      const element = triggerElementRef.current;
+      if (element && isElementDisabled(element) && !disabledProp) {
+        warn(
```
michaldudak, line 90: It would be great to include the component stack trace, as in #3861

CiscoFran10: Added stack trace to warn message:

```ts
    useIsoLayoutEffect(() => {
      const element = triggerElementRef.current;
      if (element && isElementDisabled(element) && !disabledProp) {
        const ownerStackMessage = SafeReact.captureOwnerStack?.() || '';
        warn(
          `A disabled element was detected on <Menu.SubmenuTrigger>. To properly disable the trigger, use the \`disabled\` prop on the component instead of setting it on the rendered element.${ownerStackMessage}`,
        );
      }
    });
  }
```

mj12albert (commented):
The demos don't actually contain disabled items, are the additional styles necessary?

CiscoFran10:
@mj12albert 
> The demos don't actually contain disabled items, are the additional styles necessary?

I think it’s a good enhancement for the sandboxes, even if the demos don’t include disabled items.

docs/src/app/(docs)/react/components/menu/demos/checkbox-items/tailwind/index.tsx
```diff
@@ -21,7 +21,7 @@ export default function ExampleMenu() {
             <Menu.CheckboxItem
               checked={showMinimap}
               onCheckedChange={setShowMinimap}
-              className="grid cursor-default grid-cols-[0.75rem_1fr] items-center gap-2 py-2 pr-8 pl-2.5 text-sm leading-4 outline-hidden select-none data-[highlighted]:relative data-[highlighted]:z-0 data-[highlighted]:text-gray-50 data-[highlighted]:before:absolute data-[highlighted]:before:inset-x-1 data-[highlighted]:before:inset-y-0 data-[highlighted]:before:z-[-1] data-[highlighted]:before:rounded-xs data-[highlighted]:before:bg-gray-900"
+              className="grid cursor-default grid-cols-[0.75rem_1fr] items-center gap-2 py-2 pr-8 pl-2.5 text-sm leading-4 outline-hidden select-none data-[highlighted]:relative data-[highlighted]:z-0 data-[highlighted]:text-gray-50 data-[highlighted]:before:absolute data-[highlighted]:before:inset-x-1 data-[highlighted]:before:inset-y-0 data-[highlighted]:before:z-[-1] data-[highlighted]:before:rounded-xs data-[highlighted]:before:bg-gray-900 data-[disabled]:text-gray-400 data-[disabled]:data-[highlighted]:before:bg-gray-300"
```
CiscoFran10, line 24: > data-[highlighted]:before:rounded-xs

Note: rounded-xs doesnt exist in tailwind v3.

michaldudak: We're moving away from Tailwind v3 and showcase just v4 in the docs.

michaldudak:
Thanks for your contribution!

PR #3859, opened 2026-01-25
[combobox][autocomplete] Fix popup closing on iOS VoiceOver

atomiks:

This change was added as part of https://github.com/mui/base-ui/pull/3654, but it's not necessary following the changes in `FloatingFocusManager` in later commits of the PR.

This fixes the popup closing (input outside popup pattern) when swiping onto an option

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,3 +1,5 @@
+<!-- markdownlint-disable MD038 -->
+
 # Repository Guidelines
 
 This repository contains the source code and documentation for Base UI: a headless, unstyled React component library.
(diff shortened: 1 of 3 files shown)


atomiks:
> This seems to regress the Popover closing when navigating forward with Tab on Firefox.

That's not really possible because it's not touching anything to do with Popover.  I tested and it's sometimes broken but is flaky. After refreshing several times it doesn't happen anymore.

I think that's the flake issue with React's hydration `useId()` [error I mentioned in that PR](https://github.com/mui/base-ui/pull/3654#issuecomment-3727509257), which breaks the focus guards. Because the race condition affects a bunch of other React behaviors -- React has [released a fix for it in](https://github.com/vercel/next.js/issues/84029#issuecomment-3765779599) latest canaries.

You can confirm by looking in the console on the `popup-tabbing` experiment on localhost:

Screenshot 2026-01-26 at 11 16 56 pm

The focus guards are **only** broken when that error appears in the console, but you can't see it in the deploy preview.


> Besides, given that this has regressed, do you think it's possible to write a test cover this behavior?

It's pretty tightly related to real screen reader operation with how it skips over `aria-hidden` and applies focus and isn't really testable with automation. We need to make sure changes to focus guards manually test iOS VoiceOver each time

PR #3860, opened 2026-01-25
[useScrollLock] Improve iOS scroll locking

### LINKED ISSUE #1893: [dialog] Lock scroll on iOS if navbar is collapsed

atomiks:

Fixes https://github.com/mui/base-ui/issues/1893 (mostly). Avoids breaking pinch-zoom and text selection in scroll locked dialogs.

Some concessions:
- Scrollable containers inside dialogs must set `overscroll-behavior: contain`, fixable with a hack if not set, but would rather avoid
- The page scrolls a bit after focusing an input and tapping away (native/current behavior) — can be looked into later, prone to breaking behaviors
- Text selection can select stuff in the background and scroll the page, not really fixable in any sense afaik
- Scrolling with two fingers is still possible
- When using an Apple Pencil, scroll is not locked when navbar is collapsed


packages/utils/src/useScrollLock.ts
```diff
@@ -7,6 +7,101 @@ import { Timeout } from './useTimeout';
...
+      event.preventDefault();
+    }
+  }
+
+  const touchOptions = { passive: false, capture: true };
+  doc.addEventListener('touchstart', onTouchStart, touchOptions);
+  doc.addEventListener('touchmove', onTouchMove, touchOptions);
+
+  return () => {
+    doc.removeEventListener('touchstart', onTouchStart, touchOptions);
```
michaldudak, line 113: Does the comment from line 290 apply here? It seems that ScrollLocker handles this asynchronously

michaldudak (commented):
This makes a pen unusable - tested on iPad with Apple Pencil. When Dialog is open, nothing on the page responds to pen taps.

packages/utils/src/useScrollLock.ts
atomiks: Likely not, since `isIOS` won't be `true` in testing environments, except in extremely niche scenarios

dippas:
is this planned to be fixed ? right now the new drawer has this bug with `modal` set to `true` @atomiks

atomiks:
@dippas I haven't encountered the issue in Drawer on a real device except in iOS Simulator on Mac, which doesn't seem to emulate touch correctly. What version of iOS/device are you using?

Also note if there is text selected inside the drawer, it may be possible to scroll the background.

But this PR likely wouldn't fix those issues, because the logic acts similarly to what drawer is doing.

dippas:
@atomiks we've checked in these OS/devices

- iOS 26.3.1 / iPhone 15 pro
- iOS 26.3.1 / iPhone 13 pro
- iOS 26.2 / iPhone 13

atomiks:
@dippas do any of the [drawer demos](https://base-ui.com/react/components/drawer) reproduce for you on the Base UI docs? I tested iPhone 12 + 16 on iOS 26.3.1 and the background never scrolls on any modal demo when the navbar is collapsed. 

Are you facing the issue with your own drawer? Do you have a Stackblitz reproduction where it happens?

(Note: in the Dialog docs on base-ui.com, the scroll is locked for the demos near the top of the screen for some reason I haven't worked out yet. The Dialog demos lower on the page don't lock the scroll. So I figured a reproduction would require interacting with one of the Drawer demos lower on the page, but none have the issue.)

dippas:
yes the demos are working fine, so it should be something on our end, I've tried to strip at most in order to create a reproduction snippet, and here it is https://stackblitz.com/edit/drawer-ios-scrollock, please let me know if you need anything else

dippas:
@atomiks forgot to tag you in the previous comment 🤦‍♂️

atomiks:
@dippas thanks for the repro. It's because you aren't using `Drawer.Viewport` which is what has the scroll locking logic. Actually, Drawer is meant to require this part, but for some reason it's not erroring 🤦

dippas:
@atomiks it worked, but yes having that erroring would have been helpful for us, thank you for the support

atomiks:
## Codex Review (GPT-5.4)

This remains a `bug-fix` PR with `tests`, `public-api`, and `refactor` as secondary tags. I re-reviewed the updated branch against a freshly fetched `upstream/master`, reran targeted validation, and the current version lands in a cleaner place: the iOS scroll-lock fix remains intact, and the shared touch helpers are deduplicated again.

I do not see any remaining branch-relevant correctness or packaging blockers in the latest pushed branch.

| Area | Assessment | Notes |
|------|:----------:|-------|
| iOS background scroll lock | Improved | `useScrollLock` now handles range inputs, text selection, stylus touches, and nested scrollers more robustly |
| Nested scroll boundary handling | Fixed | The branch prevents scroll chaining at top/bottom boundaries instead of allowing leakage whenever a scrollable ancestor exists |
| Drawer parity | Improved | Shared low-level touch helpers align behavior between Drawer and `useScrollLock` without pulling in Drawer-specific gesture policy |

| Area | Status | Notes |
|------|:------:|-------|
| Shared selection/range logic | Consistent | Drawer and `useScrollLock` reuse the same low-level touch helpers again |
| Shared scrollable helpers | Consistent | The React-side internal `scrollable` module forwards to the shared utils implementation |
| Component-specific policy | Good | Drawer keeps its gesture-specific edge logic local; `useScrollLock` only takes the boundary guard it needs |

**Strengths:**
- Removes the duplicate range-input and text-selection helpers from `DrawerViewport`.
- Restores deduplication for the shared touch/scrollable primitives after the temporary internalization detour.
- Keeps the shared layer at the “touch primitives” level rather than introducing a broad gesture abstraction.

| Pinch zoom | Handled | `useScrollLock` still bails for two-finger gestures |
| Range inputs | Handled | Covered in targeted `useScrollLock` tests |
| Text selection handles | Handled | Covered in targeted `useScrollLock` tests |
| Scrollable container overscroll | Handled | Top/bottom boundary regression coverage is included |
| Package exports | Acceptable | The shared helpers remain in `@base-ui/utils`; given the repo’s current package shape, that is the cleanest way to avoid duplication |

I ran:

- `pnpm test:jsdom useScrollLock --no-watch`
- `pnpm test:jsdom DrawerViewport --no-watch`
- `pnpm typescript`
- `pnpm --filter @base-ui/utils test:package`

**Strengths:**
- `useScrollLock` has focused iOS tests for no-scroll-container blocking, allowed nested scrolling, top/bottom boundary blocking, range inputs, and selection handles.
- The existing `DrawerViewport` suite still passes after moving the shared helpers back into `@base-ui/utils`.
- The `@base-ui/utils` package-level validation passes with the shared modules present.

|----------------|:------:|-------|
| Nested scroll boundary leakage | Resolved | The branch now prevents scroll chaining at scroll boundaries |
| Accidental export-surface concern | Resolved by decision | The shared helpers intentionally remain in `@base-ui/utils` to preserve deduplication |

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Correctness / regression risk | 4/5 | The main runtime blocker is fixed and covered by targeted tests |
| Pattern consistency | 4/5 | Shared behavior is aligned without introducing a broader abstraction than needed |
| Public API / packaging risk | 4/5 | The shared helper exports are now an intentional tradeoff rather than an accidental leak |

Overall merge confidence is 4/5. I would be comfortable merging this version of the PR.

atomiks:
## Codex Review (GPT-5.4)

Overview: Re-reviewed the full PR after the touch-axis follow-up in `e0591f89e`, with the earlier concern about eager axis selection as the main focus. I don't see any remaining blocking findings in the current branch state.

Type: `bug-fix`
Secondary tags: `tests`, `refactor`, `react-ui`

Findings: No blocking findings.

What changed since the last pass:
- `useScrollLock` now defers axis commitment through the shared `getTouchMoveAxis()` helper in `@base-ui/utils/touch`, instead of deciding immediately from raw `abs(dx) > abs(dy)`.
- Drawer reuses that same helper for its cross-axis preservation path, so the gesture tolerance lives in one place instead of drifting again.
- `useScrollLock` now has explicit regression coverage for ambiguous diagonal movement inside a scrollable container.

Validation:
- `pnpm test:jsdom useScrollLock --no-watch`
- `pnpm test:jsdom DrawerViewport --no-watch`
- `pnpm typescript`
- `pnpm eslint packages/utils/src/touch.ts packages/utils/src/useScrollLock.ts packages/react/src/drawer/viewport/DrawerViewport.tsx packages/utils/src/useScrollLock.test.tsx`
- `pnpm prettier --check packages/utils/src/touch.ts packages/utils/src/useScrollLock.ts packages/react/src/drawer/viewport/DrawerViewport.tsx packages/utils/src/useScrollLock.test.tsx`

Merge confidence: 4.5/5

packages/utils/src/useScrollLock.ts
```diff
@@ -2,11 +2,118 @@
...
+type EventWithPointerType = { pointerType: PointerEvent['pointerType'] };
+
+function isStylusTouch(event: TouchEvent) {
+  const touch = event.touches[0] ?? event.changedTouches[0];
+  const touchType = (touch as TouchWithTouchType | undefined)?.touchType;
+  const pointerType = (event as unknown as EventWithPointerType).pointerType;
+  return touchType === 'stylus' || touchType === 'pen' || pointerType === 'pen';
+}
+
+export function preventScrollIOS(referenceElement: Element | null = null) {
```
romgrk: Does it need to be exported?

packages/utils/src/useScrollLock.ts
```diff
@@ -257,15 +364,17 @@ class ScrollLocker {
...
 
-    const hasOverlayScrollbars = isIOS || !hasInsetScrollbars(referenceElement);
+    if (isIOS) {
+      const restoreOverlayScrollbars = preventScrollOverlayScrollbars(referenceElement);
+      const restoreIOS = preventScrollIOS(referenceElement);
+      this.restore = () => {
+        restoreIOS();
+        restoreOverlayScrollbars();
+      };
+      return;
```
romgrk: Could this be moved to inside `preventScrollIOS`? Keeps this section more clean.

packages/utils/src/useScrollLock.ts
atomiks: Eh, probably not

atomiks: Made `preventScrollIOS` private in `ccda3f9b4`; it is only used inside `useScrollLock` now.

packages/utils/src/useScrollLock.ts
atomiks: Moved the combined iOS cleanup into `preventScrollIOS` in `ccda3f9b4`, so the `ScrollLocker.lock()` branch is back to a single `this.restore = preventScrollIOS(referenceElement)` assignment.

atomiks:
[iOS 26.4 finally fixes this](https://webkit.org/blog/17862/webkit-features-for-safari-26-4/) — mostly. In some rare cases I was still able to get it to scroll, but it works a lot better than before when testing Dialog. Probably can't rely on it for Drawer still, but for Dialog and other scroll-locked components it's likely not worth adding this new logic in.

mririgoyen:
This is a major problem for our use of `Menu`. `modal` does not work on touch devices. Would really love to see this, or another similar fix, go in soon.

atomiks:
@mririgoyen for Menu specifically it's not really related to this PR.  Menu intentionally doesn't lock scroll on touch devices so you can scroll outside of it to dismiss. Is your use case something like this: https://github.com/mui/base-ui/issues/3098

mririgoyen:
> @mririgoyen for Menu specifically it's not really related to this PR. Menu intentionally doesn't lock scroll on touch devices so you can scroll outside of it to dismiss. Is your use case something like this: #3098

Yes. There is a Menu on desktop that, when opened on mobile, takes up the full screen. Users can scroll the content behind it, even when `modal` is `true`. `modal` shouldn't be ignored just because its a mobile device. Watch the DOM, the scrollbar gutter and the overflow styles never get applied to the html and body tags respectively.

atomiks:
@mririgoyen I'll try to revive https://github.com/mui/base-ui/pull/3100

The goal with `modal={true}` not locking scroll on touch was to match native anchored popups on iOS like `<select>`, which allow you to scroll outside of them to dismiss them (making the page feel less dead/improving UX), but still block click activation when tapping outside — meaning partially modal. 

But `modal={true}` on touch should still lock scroll if you're swiping within the popup still, I just didn't find a good solution for that yet. That would fix the reported issue and yours. Though it would only fully work in iOS 26.4, and work on previous versions only if the navbar isn't collapsed. I consider that more of a UX papercut than a serious issue, though. (Also, if the popup is scrollable, `overscroll-behavior: contain` already works.)

PR #3861, opened 2026-01-25
[useButton] Capture component stack for `nativeButton` error message

### LINKED ISSUE #3779: Improve error message for `render` prop
## Summary

Currently, there is no easy way to find out which component is causing the following error:
> Base UI: A component that acts as a button was not rendered as a native <button>, which does not match the default. Ensure that the element passed to the `render` prop of the component is a real <button>, or set the `nativeButton` prop on the component to `false`.

The source of the error is `error.js` which is not helpful.

Usually, the error would arise from the component file itself or at least contain a reference of the file or code causing it.

## Motivation

It is near impossible to debug once you have a lot of components in a page. There might be an existing solution that I am not aware of, if so, please do share it.

atomiks:

Fixes #3779, at least for React 19.1+ (when `React.captureOwnerStack()` was released). React 18 is supposedly better at printing the component stack trace inside effects already, though.

Before (+ expanded):
Screenshot 2026-01-26 at 9 13 40 am

After (collapsed):
Screenshot 2026-01-26 at 9 12 51 am

⚠️ You need to refresh the page to see the stack from first load(?)

### DIFF
--- a/packages/react/src/use-button/useButton.ts
+++ b/packages/react/src/use-button/useButton.ts
@@ -43,15 +43,16 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
       }
 
       const isButtonTag = elementRef.current.tagName === 'BUTTON';
-      const ownerStackMessage = SafeReact.captureOwnerStack?.() || '';
 
       if (isNativeButton) {
         if (!isButtonTag) {
+          const ownerStackMessage = SafeReact.captureOwnerStack?.() || '';
           error(
             `A component that acts as a button was not rendered as a native <button>, which does not match the default. Ensure that the element passed to the \`render\` prop of the component is a real <button>, or set the \`nativeButton\` prop on the component to \`false\`.${ownerStackMessage}`,
           );
         }
       } else if (isButtonTag) {
+        const ownerStackMessage = SafeReact.captureOwnerStack?.() || '';
         error(
           `A component that acts as a button was rendered as a native <button>, which does not match the default. Ensure that the element passed to the \`render\` prop of the component is not a real <button>, or set the \`nativeButton\` prop on the component to \`true\`.${ownerStackMessage}`,
         );


packages/react/src/use-button/useButton.ts
```diff
@@ -45,13 +46,15 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
 
       if (isNativeButton) {
         if (!isButtonTag) {
+          const ownerStackMessage = SafeReact.captureOwnerStack?.() || '';
           error(
-            'A component that acts as a button was not rendered as a native <button>, which does not match the default. Ensure that the element passed to the `render` prop of the component is a real <button>, or set the `nativeButton` prop on the component to `false`.',
+            `A component that acts as a button was not rendered as a native <button>, which does not match the default. Ensure that the element passed to the \`render\` prop of the component is a real <button>, or set the \`nativeButton\` prop on the component to \`false\`.${ownerStackMessage}`,
```
oliviertassinari, line 51: Would this make more sense? I'm struggling to understand what "which does not match the default" mean.

```suggestion
            `A component that acts as a button was not rendered as a native <button>. For the component to behave correctly, it needs the `nativeButton` boolean prop to match the actual DOM node rendered. To fix this, ensure that the element passed to the \`render\` prop of the component is a real <button>, or set the \`nativeButton\` prop on the component to \`false\`.${ownerStackMessage}`,
```

matching:

1. Say what’s the problem
2. Say why it's a problem
3. Point out how to solve the problem

oliviertassinari:
Nice 👍

packages/react/src/use-button/useButton.ts
atomiks: "Default" means the "default rendered element tag", so e.g. a component's JSDoc says "Renders a `<button>` element."

Yeah, the error message is missing the second dot point (it's needed for SSR), but handles the first and third.

PR #3866, opened 2026-01-26
[input] Update ref type to `HTMLElement`

mj12albert:
This allows `<Field.Control render={<textarea />} />` without type errors when passing a ref to a `HTMLTextAreaElement`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,3 +1,5 @@
+<!-- markdownlint-disable MD038 -->
+
 # Repository Guidelines
 
 This repository contains the source code and documentation for Base UI: a headless, unstyled React component library.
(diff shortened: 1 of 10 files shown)


packages/react/src/field/control/FieldControl.tsx
```diff
@@ -26,7 +26,7 @@ import type { BaseUIChangeEventDetails } from '../../utils/createBaseUIEventDeta
  */
 export const FieldControl = React.forwardRef(function FieldControl(
   componentProps: FieldControl.Props,
-  forwardedRef: React.ForwardedRef<HTMLInputElement>,
+  forwardedRef: React.ForwardedRef<HTMLElement>,
```
LukasTy, line 29: Wouldn't this be more correct?
```suggestion
  forwardedRef: React.ForwardedRef<HTMLInputElement | HTMLTextAreaElement>,
```
Could also update specs with testing both versions.

mj12albert: I think it could actually support `<div contentEditable>`

> The contentEditable example in FieldControl.spec.tsx:9-11 will cause runtime errors because the validation system expects elements with .validity

Regarding this comment from the bot, I don't think it really matters because TS would reject validation props passed to a div anyway?

LukasTy: Yeah, good point, we can keep the added flexibility.
In any case, the user needs to define the actual type that they are passing to the component.

One minor gripe is that the default `ref` type (without any explicit `render`) is going to become more generic, even hough we know that it renders an `<input>`

Issue #3172, opened 2025-11-07
[field] Focus state inaccurate on initial render when `Field.Control` has `autoFocus`


kenanyusuf:
# Bug report

## Current behavior

I am using the Field component and passing the `autoFocus` prop to the `Field.Control`. `state.focused` is `false` and `data-focused` is not present on the initial render.

## Expected behavior

I expected `state.focused` to be `true` and `data-focused` to be present on the field elements.

## Reproducible example

https://codesandbox.io/p/devbox/ts8md6

Steps to reproduce:
- Open the preview in a new tab https://ts8md6-3000.csb.app/
- Notice the `input` is focused, but the focus outline is not present

How it looks:

<img width="274" height="88" alt="Image" src="https://github.com/user-attachments/assets/6ea01005-b839-40ad-85b8-0bb7c19fe0d7" />

How it should look:

<img width="277" height="90" alt="Image" src="https://github.com/user-attachments/assets/de314bf5-2827-4caf-950f-b7512f41f368" />

## Base UI version

v1.0.0-beta.4

## Which browser are you using?

Chrome

## Which OS are you using?

Mac OS

## Additional context

I suspect this may be related to server-side rendering. The reproduction of the issue is using SSR and Next.js, whereas [this version](https://codesandbox.io/p/sandbox/field-autofocus-issue-working-without-ssr-fyq3z3) is client-rendered and working as expected.

mj12albert:
> I suspect this may be related to server-side rendering

I think you're right, related issue: https://github.com/mui/base-ui/issues/1525

mj12albert:
RA has the same issue:
- Next.js https://codesandbox.io/p/devbox/6r2mw9
- plain Vite https://stackblitz.com/edit/vitejs-vite-wb3yzuyc?file=src%2FApp.tsx

PR #3871, opened 2026-01-27
[field] Fix autofocus in SSR environments

mj12albert:
Demo: https://ytf734-3000.csb.app

Fixes https://github.com/mui/base-ui/issues/3172


packages/react/src/field/control/FieldControl.tsx
```diff
@@ -65,6 +77,18 @@ export const FieldControl = React.forwardRef(function FieldControl(
     }
   }, [validation.inputRef, setFilled, valueProp]);
 
+  const inputRef = React.useRef(null);
+
+  useIsoLayoutEffect(() => {
+    if (
+      autoFocus &&
+      inputRef.current === activeElement(ownerDocument(inputRef.current)) &&
+      !fieldState.focused
```
atomiks: Since React bails/won't re-render if already `true`, we can probably skip this check

packages/react/src/field/control/FieldControl.tsx
```diff
@@ -65,6 +77,18 @@ export const FieldControl = React.forwardRef(function FieldControl(
     }
   }, [validation.inputRef, setFilled, valueProp]);
 
+  const inputRef = React.useRef(null);
```
atomiks: nit: type missing

PR #3874, opened 2026-01-27
[experiments] Create Vite playground

michaldudak:
Created a Vite-based playground for testing components without Next.js overhead.
I set up sourcemap generation to make it easier to debug issues and set up a production build with profiling (opt-in) so it's clearer to find perf issues.

I also published the playground as a subpath of the docs. It's available at https://deploy-preview-3874--base-ui.netlify.app/vite-playground/


albertdugba:
This is really nice

I observed that in dark mode, the popup backgrounds (menu, tooltip, popover) stay white so I went ahead to fix it.
Here is the [PR](https://github.com/michaldudak/base-ui/pull/3)

before
Screenshot 2026-01-28 at 12 03 51 AM

after
Screenshot 2026-01-28 at 12 04 07 AM

PR #3877, opened 2026-01-27
[context menu] Fix explicit `collisionAvoidance` with `side: 'flip'` not working

### LINKED ISSUE #3876: [context menu] `collisionAvoidance { side: 'flip' }` shifts menu instead of flipping sides
# Bug report

## Current behavior

When setting `collisionAvoidance={{ side: 'flip' }}` on a `ContextMenu` anchored to an element, the menu does not flip to the opposite side of the anchor when there isn’t enough space. Instead, it appears to shift position while remaining on the same side.

See the demo linked below.

## Expected behavior

My understanding of collisionAvoidance={{ side: 'flip' }} is that when the menu does not fit on its preferred side, it should move to the opposite side of the anchor. I would expect a full side flip rather than a positional shift. In this case, I'd expect it to appear above the anchor, not overlapping it:

Image

## Reproducible example

https://stackblitz.com/edit/jb8x2jfn?file=src%2FApp.tsx

## Base UI version

Latest (v1.1.0)

Chrome

## Which OS are you using?

macOS

None

obeattie:

When setting `collisionAvoidance={{ side: 'flip' }}` on a `ContextMenu`, the menu was not flipping to the opposite side when there wasn't enough space. Instead, it would shift position while remaining on the same side.

Fixes #3876

## Root cause

The `shiftCrossAxis` parameter was unconditionally set to `true` for context menus, which disabled the main axis flip in `useAnchorPositioning` due to this condition:t

https://github.com/mui/base-ui/blob/50e164e86fb6e78c1d82c710187922b28a792874/packages/react/src/utils/useAnchorPositioning.ts#L256

## Fix

Only enable `shiftCrossAxis` for context menus when the user hasn't explicitly requested flip behavior.

This preserves the default shift behavior for context menus (which makes sense when positioned against a mouse click point), while allowing flip to work when explicitly configured.

## Test plan

- Added test for `collisionAvoidance={{ side: 'flip' }}` on `ContextMenu` (previously failing, now passes)
- Existing tests continue to pass

### DIFF
--- a/packages/react/src/menu/positioner/MenuPositioner.tsx
+++ b/packages/react/src/menu/positioner/MenuPositioner.tsx
@@ -106,7 +106,8 @@ export const MenuPositioner = React.forwardRef(function MenuPositioner(
     keepMounted,
     disableAnchorTracking,
     collisionAvoidance,
-    shiftCrossAxis: contextMenu && !('side' in collisionAvoidance && collisionAvoidance.side === 'flip'),
+    shiftCrossAxis:
+      contextMenu && !('side' in collisionAvoidance && collisionAvoidance.side === 'flip'),
     externalTree: floatingTreeRoot,
   });
 


packages/react/src/menu/positioner/MenuPositioner.tsx
```diff
@@ -106,7 +106,7 @@ export const MenuPositioner = React.forwardRef(function MenuPositioner(
     keepMounted,
     disableAnchorTracking,
     collisionAvoidance,
-    shiftCrossAxis: contextMenu,
+    shiftCrossAxis: contextMenu && !('side' in collisionAvoidance && collisionAvoidance.side === 'flip'),
```
atomiks:
```suggestion
    shiftCrossAxis:
      contextMenu && !('side' in collisionAvoidance && collisionAvoidance.side === 'flip'),
```

PR #3882, opened 2026-01-28
[toast] Make `useToastManager` and `createToastManager` generic functions

solastley:
## Overview
This PR makes the hook `useToastManager` a generic function so that the returned `toasts`, `add`, `update`, and `promise` functions all inherit the type argument provided to `useToastManager`.

I thought this might be worth a discussion. For my use case (and I imagine this is true for most folks, but perhaps not), I have a single type which represents a discriminated union of all possible types that can be used as the type of `data` in toast objects. When I access the toast manager with `useToastManager`, it would be very convenient to specify this type at that time and not need to type cast when using the `toasts` on the returned toast manager.

## Example

For example, today I need to do something like the following:

```tsx
enum ToastType {
  A = 'A',
  B = 'B',
}

type CustomToastA = {
  type: ToastType.A;
  someCustomField: boolean;
  ...
};

type CustomToastB = {
  type: ToastType.B;
  someOtherField: string;
  ...
};

type CustomToast = CustomToastA | CustomToastB | ...

const MyComponent = () => {
  const { toasts } = Toast.useToastManager();
  
  return (
    <Toast.Viewport>
      {toasts.map((toast) => {
        // Need to type cast here
        const typedToastData = toast.data as CustomToast;
        ...
      })}
    </Toast.Viewport>
  );
};
```

With this change, if the type `CustomToast` were provided to `useToastManager` as a type argument in the above example, then `toast.data` would be automatically typed as `CustomToast`.

## Concerns

This change modifies `add`, `update`, and `promise` to no longer be generic. So I believe this would constitute a breaking change for those methods if consumers are currently providing type arguments to them.

### DIFF
--- a/packages/react/src/toast/createToastManager.spec.tsx
+++ b/packages/react/src/toast/createToastManager.spec.tsx
@@ -17,6 +17,23 @@ const typedAddId = typedManager.add({
 });
 expectType<string, typeof typedAddId>(typedAddId);
 
+typedManager.add({
+  title: 'wrong-shape',
+  data: {
+    id: 'test',
+    // @ts-expect-error - message is not a valid property
+    message: 'not a number',
+  },
+});
+
+typedManager.add({
+  title: 'wrong-shape',
+  // @ts-expect-error - count is a missing property
+  data: {
+    id: 'test',
+  },
+});
+
 typedManager.update('typed', {
   data: {
     id: 'typed-update',
--- a/packages/react/src/toast/useToastManager.spec.tsx
+++ b/packages/react/src/toast/useToastManager.spec.tsx
@@ -21,6 +21,23 @@ const typedAddId = typedManager.add({
 });
 expectType<string, typeof typedAddId>(typedAddId);
 
+typedManager.add({
+  title: 'wrong-shape',
+  data: {
+    id: 'test',
+    // @ts-expect-error - message is not a valid property
+    message: 'not a number',
+  },
+});
+
+typedManager.add({
+  title: 'wrong-shape',
+  // @ts-expect-error - count is a missing property
+  data: {
+    id: 'test',
+  },
+});
+
 typedManager.update('typed', {
   data: {
     id: 'typed-update',


atomiks:
You should be able to make it backward‑compatible while still enabling the new typed manager by defaulting the method generic to the hook's `Data` type

```jsx
export function useToastManager<Data extends object = any>(): UseToastManagerReturnValue<Data> {

...

export interface UseToastManagerReturnValue<Data extends object = any> {
  toasts: ToastContext<Data>['toasts'];
  add: <T extends object = Data>(options: ToastManagerAddOptions<T>) => string;
  close: (toastId: string) => void;
  update: <T extends object = Data>(toastId: string, options: ToastManagerUpdateOptions<T>) => void;
  promise: <Value, T extends object = Data>(
    promise: Promise<Value>,
    options: ToastManagerPromiseOptions<Value, T>,
  ) => Promise<Value>;
}
```

--- 

Spec file

```tsx
/* eslint-disable react-hooks/rules-of-hooks */
import { expectType } from '#test-utils';
import { useToastManager } from './useToastManager';

type ToastPayload = {
  id: string;
  count: number;
};

const typedManager = useToastManager<ToastPayload>();

const typedToastData = typedManager.toasts[0]?.data;
expectType<ToastPayload | undefined, typeof typedToastData>(typedToastData);

const typedAddId = typedManager.add({
  title: 'typed',
  data: {
    id: 'typed',
    count: 1,
  },
});
expectType<string, typeof typedAddId>(typedAddId);

typedManager.update('typed', {
  data: {
    id: 'typed-update',
    count: 2,
  },
});

typedManager.promise(Promise.resolve(2), {
  loading: 'loading',
  success: (value) => ({
    title: `${value}`,
    data: {
      id: 'typed-success',
      count: value,
    },
  }),
  error: 'error',
});

const legacyManager = useToastManager();

const legacyAddId = legacyManager.add<ToastPayload>({
  title: 'legacy',
  data: {
    id: 'legacy',
    count: 3,
  },
});
expectType<string, typeof legacyAddId>(legacyAddId);

legacyManager.update<ToastPayload>('legacy', {
  data: {
    id: 'legacy-update',
    count: 4,
  },
});

legacyManager.promise<number, ToastPayload>(Promise.resolve(5), {
  loading: 'loading',
  success: (value) => ({
    title: `${value}`,
    data: {
      id: 'legacy-success',
      count: value,
    },
  }),
  error: 'error',
});
```

solastley:
@atomiks nice, just updated the PR with your suggested changes. Did you want me to add that spec as well, or were you just sharing to demonstrate that it works as expected?

If you'd like me to add the spec, let me know.

atomiks:
@solastley yeah, should add it

solastley:
@atomiks hm, might be missing something, but I don't think these tests work since `useToastManager` can only run within a React component which is rendered within a `Toast.Provider`.

atomiks:
The spec file is only for types, it doesn't test runtime behavior. We have other spec files (filename.spec.tsx) for this purpose. `/* eslint-disable react-hooks/rules-of-hooks */` is used to disable the linting.

solastley:
@atomiks done

atomiks:
@solastley thanks! I think this also applies to `createToastManager`?

solastley:
@atomiks yeah good call. Modified `createToastManager` to be generic as well and added the relevant spec.

packages/react/src/toast/useToastManager.ts
```diff
@@ -106,14 +106,14 @@ export interface ToastManagerPositionerProps extends Omit<
   anchor?: (Element | null) | undefined;
 }
 
-export interface UseToastManagerReturnValue {
-  toasts: ToastContext<any>['toasts'];
-  add: <Data extends object>(options: ToastManagerAddOptions<Data>) => string;
+export interface UseToastManagerReturnValue<Data extends object = any> {
+  toasts: ToastContext<Data>['toasts'];
+  add: <T extends object = Data>(options: ToastManagerAddOptions<T>) => string;
```
atomiks: These need to be `T extends Data = Data` to catch wrong shaped payloads:

```jsx
typedManager.add({
  title: 'wrong-shape',
  data: {
    id: 'test',
    // @ts-expect-error - message is not a valid property
    message: 'not a number',
  },
});

typedManager.add({
  title: 'wrong-shape',
  // @ts-expect-error - count is a missing property
  data: {
    id: 'test',
  },
});
```

solastley: @atomiks fixed

atomiks: Can you add those calls I made to the spec file?

solastley: @atomiks done 🫡

atomiks:
@greptileai

Issue #3873, opened 2026-01-27
[tabs] `data-activation-direction` not updated during programmatic value changes


LovelessCodes:
# Bug report

## Current behavior

When using a controlled `Tabs` component, the `data-activation-direction` attribute is only updated when clicking a `Tabs.Tab`. If the `value` is updated programmatically (e.g., via a "Next" button changing the value), the attribute does not update or reflect the direction of the change. This causes CSS transitions relying on this attribute to fail or behave inconsistently.

## Expected behavior

The `data-activation-direction` attribute should be calculated and updated whenever the `value` changes, whether the change originated from a user click on a tab trigger or a programmatic state update in a controlled setup.

## Reproducible example

[Tabs Test - Animation](https://stackblitz.com/edit/vitejs-vite-ab7bnqjz?file=src%2FApp.tsx&terminal=dev)

## Base UI version

v1.1.0

## Which browser are you using?

Helium (Chromium) & Firefox

## Which OS are you using?

Mac OS

## Additional context

In the provided StackBlitz, clicking the "Next" button updates the active tab correctly, but the animation direction (driven by `data-activation-direction`) remains static or incorrect compared to direct tab clicks. This suggests the internal direction logic is tied to the click event handler rather than the value change effect.

PR #3883, opened 2026-01-29
[tabs] Fix activation direction not updating on programmatic value changes

LukasTy:
`data-activation-direction` attribute only updated when clicking tabs directly. Programmatic value changes in controlled mode left the attribute stale, breaking CSS animations that depend on direction.

### Changes

- **TabsRoot.tsx** — New `computeActivationDirection()` function determines direction by comparing tab indices (instead of DOM pixel positions). Direction is computed both during render (for external value changes) and in `onValueChange` (for click/keyboard). A `previousValueRef` + layout effect tracks the prior value.
- **TabsList.tsx** — Removed the old `useActivationDirectionDetector` hook and getInset helper (~89 lines) that relied on `getBoundingClientRect` and only worked for click-triggered changes.
- **TabsRoot.test.tsx** — Added tests for programmatic value changes in horizontal and vertical orientations, verifying both `data-activation-direction` attribute and render callback state. Vertical test is restricted to Chromium (`it.skipIf(isJSDOM)`).

### Before/After

```tsx
// Programmatic value change
const [value, setValue] = useState(0);

<Tabs.Root value={value}>
  <Tabs.List>...</Tabs.List>
</Tabs.Root>

<button onClick={() => setValue(1)}>Next</button>
```

Before: `data-activation-direction` stays `"none"`  
After: `data-activation-direction` updates to `"right"` (or `"left"`, `"up"`, `"down"` based on direction)

Fixed StackBlitz example: https://stackblitz.com/edit/vitejs-vite-htowv2m8?file=src%2FApp.tsx

___

## Note

Initially tried playing with Copilot to fix this, but ended up finishing locally with manual changes, a bit of Copilot and mostly Claude fine-tuning it.

> [!WARNING]
>
> <details>
> <summary>Firewall rules blocked me from connecting to one or more addresses (expand for details)</summary>
>
> #### I tried to connect to the following addresses, but was blocked by firewall rules:
>
> - `https://storage.googleapis.com/chrome-for-testing-public/143.0.7499.4/linux64/chrome-headless-shell-linux64.zip`
>   - Triggering command: `/home/REDACTED/.nvm/versions/node/v22.22.0/bin/node /home/REDACTED/.nvm/versions/node/v22.22.0/bin/node /home/REDACTED/work/base-ui/base-ui/node_modules/.pnpm/playwright-core@1.57.0/node_modules/playwright-core/lib/server/registry/oopDownloadBrowserMain.js` (http block)
> - `https://storage.googleapis.com/chrome-for-testing-public/143.0.7499.4/linux64/chrome-linux64.zip`
>   - Triggering command: `/home/REDACTED/.nvm/versions/node/v22.22.0/bin/node /home/REDACTED/.nvm/versions/node/v22.22.0/bin/node /home/REDACTED/work/base-ui/base-ui/node_modules/.pnpm/playwright-core@1.57.0/node_modules/playwright-core/lib/server/registry/oopDownloadBrowserMain.js` (http block)
> - `iojs.org`
>   - Triggering command: `/usr/bin/curl curl -q --fail --compressed -L -s REDACTED -o -` (dns block)
>
> If you need me to access, download, or install something from one of these locations, you can either:
>
> - Configure [Actions setup steps](https://gh.io/copilot/actions-setup-steps) to set up my environment, which run before the firewall is enabled
> - Add the appropriate URLs or hosts to the custom allowlist in this repository's [Copilot coding agent settings](https://github.com/LukasTy/base-ui/settings/copilot/coding_agent) (admins only)
>
> </details>

<details>

<summary>Original prompt</summary>

> Solve https://github.com/mui/base-ui/issues/3873 by creating a branch in this repo, working on a solution and opening a draft PR against the mui/base-ui repository.
> The issue is clear, programatic value updates indeed do not update the `activationDirection`.
> Create tests asserting the behavior is fixed.
> You can start by creating the tests, which would fail before working on a solution - TDD approach.

</details>

---

💡 You can make Copilot smarter by setting up custom instructions, customizing its development environment and configuring Model Context Protocol (MCP) servers. Learn more [Copilot coding agent tips](https://gh.io/copilot-coding-agent-tips) in the docs.

### DIFF
--- a/packages/react/src/tabs/root/TabsRoot.test.tsx
+++ b/packages/react/src/tabs/root/TabsRoot.test.tsx
@@ -425,7 +425,7 @@ describe('<Tabs.Root />', () => {
       expect(handlePointerDown.mock.calls.length).toBe(1);
     });
 
-    it.skipIf(isJSDOM)('should call onValueChange when clicking', async () => {
+    it('should call onValueChange when clicking', async () => {
       const handleChange = vi.fn();
 
       await render(
@@ -620,7 +620,7 @@ describe('<Tabs.Root />', () => {
     ].forEach((entry) => {
       const [orientation, direction, previousItemKey, nextItemKey] = entry;
 
-      describe.skipIf(isJSDOM && direction === 'rtl')(
+      describe.skipIf(direction === 'rtl')(
         `when focus is on a tab element in a ${orientation} ${direction ?? ''} tablist`,
         () => {
           describe(previousItemKey ?? '', () => {
@@ -1285,7 +1285,7 @@ describe('<Tabs.Root />', () => {
     });
   });
 
-  describe.skipIf(isJSDOM)('activation direction', () => {
+  describe('activation direction', () => {
     it('should set the `data-activation-direction` attribute on the tabs root with orientation=horizontal', async () => {
       await render(
         <Tabs.Root data-testid="root">
@@ -1405,31 +1405,34 @@ describe('<Tabs.Root />', () => {
         expect(spyFn.mock.calls[0][0]).toBe('left');
       });
 
-      it('should set `data-activation-direction` when value is changed programmatically with orientation=vertical', async () => {
-        const spyFn = vi.fn();
-        const { user } = await render(
-          <ControlledTabs orientation="vertical" onRenderCallback={spyFn} />,
-        );
+      it.skipIf(isJSDOM)(
+        'should set `data-activation-direction` when value is changed programmatically with orientation=vertical',
+        async () => {
+          const spyFn = vi.fn();
+          const { user } = await render(
+            <ControlledTabs orientation="vertical" onRenderCallback={spyFn} />,
+          );
 
-        // reset initial render calls
-        spyFn.mockClear();
+          // reset initial render calls
+          spyFn.mockClear();
 
-        const root = screen.getByTestId('root');
-        expect(root).toHaveAttribute('data-activation-direction', 'none');
+          const root = screen.getByTestId('root');
+          expect(root).toHaveAttribute('data-activation-direction', 'none');
 
-        await user.click(screen.getByText('Set 1'));
-        expect(root).toHaveAttribute('data-activation-direction', 'down');
-        expect(screen.getByRole('tabpanel')).toHaveTextContent('down');
-        expect(spyFn.mock.calls[0][0]).toBe('down');
+          await user.click(screen.getByText('Set 1'));
+          expect(root).toHaveAttribute('data-activation-direction', 'down');
+          expect(screen.getByRole('tabpanel')).toHaveTextContent('down');
+          expect(spyFn.mock.calls[0][0]).toBe('down');
 
-        // reset before new render
-        spyFn.mockClear();
+          // reset before new render
+          spyFn.mockClear();
 
-        await user.click(screen.getByText('Set 0'));
-        expect(root).toHaveAttribute('data-activation-direction', 'up');
-        expect(screen.getByRole('tabpanel')).toHaveTextContent('up');
-        expect(spyFn.mock.calls[0][0]).toBe('up');
-      });
+          await user.click(screen.getByText('Set 0'));
+          expect(root).toHaveAttribute('data-activation-direction', 'up');
+          expect(screen.getByRole('tabpanel')).toHaveTextContent('up');
+          expect(spyFn.mock.calls[0][0]).toBe('up');
+        },
+      );
     });
   });
 


LukasTy:
@copilot, I have updated the tests to fail in certain condition.
They do pass if the click is executed on the tab, instead of click handler updating the state.

Update the solution so it behaves the same regardless of the tab change method.

benface:
@LukasTy – Any idea when you'll be able to return to this PR? 🙏

atomiks:
## Codex Review

### Overview
This patch moves activation-direction handling into `Tabs.Root` so controlled/programmatic value changes update `data-activation-direction`, not just direct tab clicks. The overall direction makes sense, but the branch still misses one of the render-prop cases it is trying to fix and it introduces a second direction algorithm that adds size and can drift from the existing click path.

### Findings
#### 1. 🔴 [Blocking] Horizontal controlled updates still expose `tabActivationDirection="none"` on the first render

**Impact**: Consumers that animate from Tabs render props can still see stale direction state after a programmatic horizontal change, so the controlled path does not yet behave the same as direct tab activation.

**Evidence**: Running `pnpm test:chromium TabsRoot --no-watch` on the PR head fails in `packages/react/src/tabs/root/TabsRoot.test.tsx:1383-1405`. After clicking `Set 1`, the root attribute and panel text are both `right`, but the first render-prop callback still receives `none` (`spyFn.mock.calls[0][0]`). The relevant implementation is in `packages/react/src/tabs/root/TabsRoot.tsx:87-150`, where the external direction is derived during render and only persisted afterward in a layout effect.

**Recommendation**: Make the controlled-update path provide the final activation direction on the first render seen by render props, not only after the follow-up effect/state sync. Re-running the Chromium `TabsRoot` suite should make the new horizontal test pass without depending on a second render.

#### 2. 🟡 [Non-blocking] Direction logic is now duplicated, which increases size and leaves the two paths free to disagree

**Impact**: Direct clicks and programmatic changes no longer share one source of truth. That makes the behavior easier to drift over time, and it already comes with a measurable bundle-size regression.

**Evidence**: `packages/react/src/tabs/list/TabsList.tsx:47-95` calculates direction from DOM geometry via `getBoundingClientRect()`, while `packages/react/src/tabs/root/TabsRoot.tsx:87-131` recalculates it from `tabMap` indices plus manual RTL inversion. Those two approaches are not equivalent once visual order differs from composite order, and the PR bundle-size report already shows the extra surface area:

| Bundle | Parsed size | Gzip size |
|:--|--:|--:|
| `@base-ui/react` | `+348B (+0.07%)` | `+99B (+0.07%)` |

**Recommendation**: Collapse both paths onto one shared direction-calculation helper instead of keeping a DOM-based click path and a separate index-based controlled path. That should simplify the patch, reduce drift risk, and likely recover most or all of the added bytes. If the root-side branch stays, it also needs explicit RTL/visual-order regression coverage.

### Confidence: 4/5
I reviewed the full branch diff, did the required second-pass sweep across correctness/accessibility/API/maintainability/perf/test coverage, checked the PR’s bundle-size report, and ran the targeted `TabsRoot` suites in both JSDOM and Chromium. The remaining uncertainty is around edge-case layouts that are not yet covered by automated tests.

### Notes
The JSDOM `TabsRoot` suite passes on this branch; the Chromium `TabsRoot` suite is the one that exposes the blocking regression above.

LukasTy:
@atomiks I've addressed the changes.
Claude suggested aligning the solution to a single one as well.
It might lack some edge-case exotic layout handling, but in general, it feels like the solution is lighter and fixes the issue of only updating the `navigationDirection` on direct tab click.

If we want to keep the DOM-based solution for clicks, then I'm not sure if there is a clean way to handle the programmatic updates without a separate solution path (bigger bundle).

atomiks:
## Codex Review

### Overview
This revision simplifies the fix by moving click and programmatic updates onto one shared activation-direction calculation, and it also reduces bundle size versus `master`. The earlier render-prop mismatch looks fixed, but the new shared path changes click behavior for visually reordered tab layouts and the updated tests now skip RTL keyboard coverage entirely.

### Findings
#### 1. 🔴 [Blocking] Click direction is now wrong when tabs are visually reordered with CSS

**Impact**: Base UI is headless, so consumers can legitimately reorder tabs with CSS (`row-reverse`, `order`, grid placement, etc.). In those layouts, click-driven animations now move in the wrong direction because the patch derives direction from logical index order instead of the rendered position of the tabs.

**Evidence**: `packages/react/src/tabs/root/TabsRoot.tsx:19-56` and `packages/react/src/tabs/root/TabsRoot.tsx:145-153` now compute activation direction from `tabMap` indices for every path. On `master`, click handling measured DOM geometry in `packages/react/src/tabs/list/TabsList.tsx:167-219`, so visual order and activation direction stayed aligned. I verified the regression with the same one-off Chromium check against both refs: a `Tabs.List` rendered with `flexDirection: 'row-reverse'` reports `data-activation-direction="left"` on `master`, but `"right"` on this PR after clicking the visually left tab.

**Recommendation**: Keep the shared programmatic path if you want the smaller patch, but do not replace the click path with an index-based calculation. Either preserve the previous DOM-position-based logic for user clicks or feed the shared helper with rendered positions rather than composite indices.

#### 2. 🟡 [Non-blocking] The test update drops RTL keyboard coverage in Chromium

**Impact**: The branch still contains RTL-specific behavior, but the updated tests no longer execute the RTL keyboard block in any environment. That makes the new direction logic easier to regress without the suite noticing.

**Evidence**: `packages/react/src/tabs/root/TabsRoot.test.tsx:623` now uses `describe.skipIf(direction === 'rtl')`, which skips the entire RTL block unconditionally. On `master`, the same block used `describe.skipIf(isJSDOM && direction === 'rtl')`, so those cases still ran in the browser projects.

**Recommendation**: Restore the old conditional (or another browser-capable equivalent) so the RTL keyboard cases still run in Chromium. Since this patch now centralizes direction logic, it would also be worth adding at least one controlled-value activation-direction assertion under `DirectionProvider direction="rtl"`.

### Confidence: 4/5
I reviewed the full `master...head` diff, did the second-pass sweep across correctness/accessibility/API/maintainability/perf/test coverage, reran the targeted `TabsRoot` suites in JSDOM and Chromium, and compared the CSS-reordered click behavior against `master` with a focused Chromium repro. The remaining uncertainty is limited to other styling edge cases that are not directly covered today.

### Notes
The earlier render-prop issue looks resolved on the current head: both `pnpm test:jsdom TabsRoot --no-watch` and `pnpm test:chromium TabsRoot --no-watch` pass now. The bundle-size bot also flipped from a regression to an improvement on this revision (`@base-ui/react`: `-282B` parsed, `-103B` gzip).

LukasTy:
Closing in favor of: https://github.com/mui/base-ui/pull/4347.

PR #3885, opened 2026-01-29
[select] Fix `items` type definition for groups

aarongarciah:
While adding a grouped demo in the Select docs (https://github.com/mui/base-ui/pull/3884), I realized the `items` prop didn't accept an array of groups.

This change expands the `items` type to also accept an array of groups with these shapes:

```ts
const groups1 = [
  {
    value: 'Fruits',
    items: ['Apple', 'Banana']
  },
  {
    value: 'Vegetables',
    items: ['Broccoli', 'Carrot']
  }
];

const groups2 = [
  {
    value: 'Fruits',
    items: [
      { value: 'apple', label: 'Apple' },
      { value: 'banana', label: 'Banana' }
    ]
  },
  {
    value: 'Vegetables',
    items: [
      { value: 'broccoli', label: 'Broccoli' },
      { value: 'carrot', label: 'Carrot' }
    ]
  }
];
```

### DIFF
--- a/packages/react/src/utils/resolveValueLabel.tsx
+++ b/packages/react/src/utils/resolveValueLabel.tsx
@@ -6,6 +6,7 @@ type ItemRecord = Record<string, React.ReactNode>;
 type ItemsInput =
   | ItemRecord
   | ReadonlyArray<LabeledItem>
+  | ReadonlyArray<Group<any>>
   | ReadonlyArray<Group<LabeledItem>>
   | undefined;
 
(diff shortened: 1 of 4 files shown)


packages/react/src/utils/resolveValueLabel.tsx
```diff
@@ -6,6 +6,7 @@ type ItemRecord = Record<string, React.ReactNode>;
 type ItemsInput =
   | ItemRecord
   | ReadonlyArray<LabeledItem>
+  | ReadonlyArray<Group<any>>
```
aarongarciah, line 9: - Should this be `Group<string>` instead? Same in the other two files. I think it shouldn't, but asking just in case.
- Ideally the types in these three files should have a single source of truth.

aarongarciah:
Supersede by https://github.com/mui/base-ui/pull/3884

PR #3886, opened 2026-01-29
[popups] Reuse Viewport logic across components

michaldudak:
Extracted the Viewport part logic to a hook to reduce duplication


michaldudak:
image

@oliviertassinari, do we participate in bundle size reduction bounties? 😅

oliviertassinari:
@michaldudak Nice win 🪶. This made me iterate a bit more on the idea of this program:

- Added a condition to not have more than 10% increase in submodules. Why in the docs (it seems ok here) https://www.notion.so/mui-org/eng-Issue-bounty-149cbfe7b660802695d8ec7eb97696cb
- Clarified it's for external contributors as a way to delegate work to the community, to get leverage.
- I wonder if the incentive formula is correct. I get $972 for this one ((400 * 1.5 + (1050 - 400) ^(1/1.16) * 1.5)). I guess it's ok-ish. For some new, it could take a bunch of hours.
- Some ops stuff to make a program like this work in practice.

But speaking of bounties, I do think that we should think of something for conference talks https://www.notion.so/devex-Conference-talk-incentivizing-program-28ccbfe7b660807fb32ac5f5fdfc52d5

atomiks (approved):

oliviertassinari:
@Janpot it looks like we have some problems to solve on this one: https://github.com/mui/base-ui/pull/3886#issuecomment-3817644436

1. The comment report doesn't make it clear that we increased the size of the submodules / entry point
2. The detail URL crashes (https://frontend-public.mui.com/size-comparison/mui/base-ui/diff?prNumber=3886&baseRef=master&baseCommit=1a6ce1b15b2e3182b101e4178c64e419e391649d&headCommit=858284cb2efd0b722802c6f4c6638fde4609d0e2)
3. I wonder if we have a problem with the error boundary, I would expect an exception with the error in the console.

Janpot:
1. We used to track these in the PR comment at some point, but it was considered too noisy. It's more likely in end-user programs that multiple subentries are imported simultaneously than only a single one, so it's better to focus on deduplicating code rather than optimizing individual subpaths. The main export, while not perfect, is more representative for end-user bundles than individual paths.
2. [Reverted](https://github.com/mui/mui-public/pull/1091) Next.js, there seems to be a regression in 16.1.2. To be investigated still.
3. I see an error in the console
    ```
    SvgIcon.js:81 Uncaught ReferenceError: _theme_vars is not defined
        at SvgIcon.js:81:22
        at styleMemoized (memoTheme.js:22:32)
        at processStyle (createStyled.js:45:55)
        at styleFunctionProcessor (createStyled.js:162:18)
        at handleInterpolation (emotion-serialize.development.esm.
        ...
    ```

PR #3888, opened 2026-01-29
[scroll area] Fix overflow edge rounding

atomiks:

Fixes https://github.com/mui/base-ui/pull/3795#issuecomment-3797121043


michaldudak:
Review by Codex (looks minor):

Normalize end offsets for tiny scroll ranges — `packages/react/src/scroll-area/utils/scrollEdges.ts:12-17` When the scroll range is ≤ 1px (e.g., content is only 1px larger than the viewport), `normalizeScrollOffset` always returns 0 because the start-edge check runs before the end-edge check. That means even at `scrollLeft === max`, the normalized value stays 0 and `scrollLeftFromEnd` stays max, so `data-overflow-*-end` never clears. This shows up whenever the scrollable range is 1px or less. Consider handling overlapping tolerances (e.g., compare distance to each edge or special-case `max <= tolerance`) so the end state can be reached.

michaldudak (approved):

PR #3889, opened 2026-01-30
[menu] Allow `onKeyDown` handlers to be called at `Group` level during typeahead

atomiks:

Listeners that execute on keydown around Group level should allow to be called and not be blocked by typeahead, as in Radix.


atomiks:
Closing since placing it on `Popup` is usually fine enough and can be migrated simply.

Issue #3906, opened 2026-01-31
[combobox][autocomplete] `aria-readonly` on `Clear` and `Popup` components when `readOnly`


ZeeshanTamboli:
# Bug report

## Current behavior

`aria-readonly` is added on `Clear` and `Popup` components for Autocomplete and Combobox when `readOnly` is used on `Root`.

## Expected behavior

`aria-readonly` seems unnecessary on `Clear` and `Popup` components for Autocomplete and Combobox. They are rendered as buttons.

## Reproducible example

StackBlitz: https://stackblitz.com/edit/3hx98nwo?file=src%2FApp.tsx

## Base UI version

v1.1.0

## Which browser are you using?

Chrome

## Which OS are you using?

Windows

## Which assistive tech are you using (if applicable)?

NVDA

## Additional context

Nothing is announced on NVDA screen-reader when `aria-readonly="true"`. They are rendered as buttons and `aria-readonly` is not applicable to them, unless I am missing something. 

`readOnly` docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly
`aria-readonly` docs: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-readonly

PR #3907, opened 2026-01-31
[combobox][autocomplete] Remove aria-readonly prop from Clear and Popup components when readOnly

markocupic024:
Closes https://github.com/mui/base-ui/issues/3906

readOnly has no effect these buttons


atomiks:
Failing test:

>  FAIL   @base-ui/react  src/combobox/root/ComboboxRoot.test.tsx > <Combobox.Root /> > selection behavior > multiple > should handle readOnly state with chips

atomiks:
Thanks for the fix @markocupic024!

PR #3911, opened 2026-02-01
[menu] Fix submenu position jumping during parent animation

seongminn:

When `SubmenuRoot` is set with the `defaultOpen={true}` option, there is an issue where the submenu's position is determined while the parent `Menu.Root` is undergoing a transition during mounting. This causes the submenu to render at a point where the parent menu's position or size is still changing, leading to a visual bug where it appears in an unintended position before 'teleporting' to the correct location after the animation ends.

https://github.com/user-attachments/assets/a1bf3c03-9dd1-451b-a7b3-d2bb5bbd4ee4

repro: https://stackblitz.com/edit/wngbgd7i?file=src%2FApp.tsx

---

The logic has been improved to prevent the submenu from calculating its position or mounting when the parent element is of type `menu` and an animation is currently in progress (`transitionStatus` is active).

* Added an `isParentAnimating` flag in `MenuPositioner` to check the `transitionStatus` of the parent menu.
* Combined the `mounted` condition of `useAnchorPositioning` with `!isParentAnimating` to ensure the submenu determines its position and mounts only after the parent's animation is completely finished.

https://github.com/user-attachments/assets/79a48ef9-10c2-412c-ad46-dc80944a5478


packages/react/src/menu/positioner/MenuPositioner.tsx
```diff
@@ -62,6 +62,10 @@ export const MenuPositioner = React.forwardRef(function MenuPositioner(
   const floatingNodeId = store.useState('floatingNodeId');
   const floatingParentNodeId = store.useState('floatingParentNodeId');
 
+  // Get parent transition status to prevent position jumping during parent animations
+  const parentType = parent.type;
+  const isParentAnimating = parentType === 'menu' && parent.store.useState('transitionStatus');
```
atomiks: Looks like an invalid conditional hook call here

seongminn: When `parent.type` is `menu`, I set `animationFrame` to `true` to implement a method that tracks the position of the anchor element.

However, this approach can consume significant CPU resources, so I'm cautious about it. If there are any other methods I could consider, I would appreciate your suggestions.

atomiks: `animationFrame: true` would be fine only while the parent is animating (for the brief period) then switch to `false`, if possible

seongminn: I agree. I've reflected that in this commit. [a45dff4](https://github.com/mui/base-ui/pull/3911/commits/a45dff4abb061a599f5f68359f6ecc8d95bb3151)

seongminn: The reason for separating it from disableAnchorTracking is that I observed the following flickering.

https://github.com/user-attachments/assets/aaba0403-008b-418c-8ed7-04d2dab47a4d

packages/react/src/menu/positioner/MenuPositioner.tsx
```diff
@@ -89,6 +89,9 @@ export const MenuPositioner = React.forwardRef(function MenuPositioner(
 
   const contextMenu = parent.type === 'context-menu';
 
+  const parentTransitionStatus = parent.type === 'menu' && parent.store.select('transitionStatus');
+  const parentAnimationFinished = parentTransitionStatus === undefined;
```
atomiks: I'm guessing this happens to work because the parent doesn't re-render until it finishes animating? As `store.select(...)` is reading a non-reactive property, and `transitionStatus === undefined` gets set immediately after mount of the parent (from `starting` -> `undefined` transition status)

seongminn: I added `parentAnimationFinished` to the store so that changes to the `transitionStatus` value are reflected immediately.

[fe7013b](https://github.com/mui/base-ui/pull/3911/commits/fe7013bcd64622980a31ce8284e3cc2d9c32d35c)

atomiks: Unfortunately this doesn't appear right. 

It happens to work due to point 2, but that's a perf regression here and can't be used.

1. The `animationFrame` option is disabled too early, as it treats parent animation as “finished” when `transitionStatus === undefined`. But `transitionStatus` is initialized/returned to undefined very quickly for open state, so this does not represent the real CSS animation duration.

2. The runtime switch of animationFrame is not reliably applied in the default (`keepMounted=false`) path. `useAnchorPositioning` configures autoUpdate through `whileElementsMounted`, but that setup is mount-time in Floating UI, so changing `disabledAnimationFrame` won't update (so the first `animationFrame` value sticks after mount)

seongminn:
I'm sorry. I'll need to think about this issue more deeply to resolve it. For now, I'll close the PR and raise it as an issue. Thank you.

PR #3914, opened 2026-02-01
[scroll area] Fix hidden state during SSR

### LINKED ISSUE #3829: [scroll-area] has overflow data attributes applied on first render before measure
# Bug report

## Current behavior

When the scroll-area component is first rendered it applies the data- attributes for has overflow x and y prior to the measure where it can determine if it actually has overflow or not.

Anyone using these attributes to apply additional styling, like padding right, when there is overflow y (for example) see a UI layout shift that just shouldn't happen.

## Expected behavior

On first render all data- attributes indicating there is overflow x or y should not be applied. These attributes are applied after the first measure and it actually knows that there is and overflow.

## Reproducible example

For a easy and visual example, simply use a scroll-area where the content does not cause any overflow. Add a class of `data-has-overflow-y:pr-4` to the scroll-area root element.

When loading the page it will apply the padding right due to the attribute being present on first render and then immediately remove the padding after the first message and it removes those overflow attributes.

In this Gif you can see the layout shift occurring, the `scroll-area` is wrapping the todo items in the body of the card.

Image

## Base UI version

Currently using version 1.0..0 (latest at time of writing).

Brave (Chromium based), Chrome and Edge.

## Which OS are you using?

Microsoft Windows 11
Version 25H2 (OS Build 26200.7623)

Not applicable.

## Additional context

The issue is that within the `ScrollAreaRoot.tsx` the "has overflow" state is derived from `hiddenState` where it's default state is set to `{ x: false, y: false, corner: false }`. `hasOverflowY` is computed as `!hiddenState.y`.

So on the very **first render**, `hiddenState.y === false` => `hasOverflowY === true`, which means any the `data-has-overflow-y` state attribute will be present immediately - before anything has been measured.

I wonder if the easiest and most correct fix would be to initialise `hiddenState` to "hidden until proven otherwise" (i.e. assume **no overflow** until the viewport/content measurement logic updates it).

Using this one-line semantic change to default hidden state as `true`.
```ts
const DEFAULT_HIDDEN_STATE = { x: true, y: true, corner: true };
```

This should work as the first render, `hasOverflowY = !true = false` => **no** `data-has-overflow-y` yet. After the viewport/content observers run and call `setHiddenState(...)`, the flags become accurate and the attribute appears only when it's actually overflowed.

Until it's measured, saying it's overflowed feel inaccurate and leads to issues like this one; hopefully this tweak doesn't have any knock effect and fixes all of the overflow attributes as they are all derived from that `hiddenState`. I also believe this will help to avoid hydration/SSR awkwardness where the server render would otherwise claim the overflow before the client can measure.

atomiks:

Fixes #3829

### DIFF
--- a/packages/react/src/scroll-area/corner/ScrollAreaCorner.test.tsx
+++ b/packages/react/src/scroll-area/corner/ScrollAreaCorner.test.tsx
@@ -1,8 +1,11 @@
-import { ScrollArea } from '@base-ui/react/scroll-area';
 import { expect } from 'chai';
 import { screen } from '@mui/internal-test-utils';
 import { createRenderer, isJSDOM } from '#test-utils';
+import { ScrollArea } from '@base-ui/react/scroll-area';
 import { describeConformance } from '../../../test/describeConformance';
+
+const VIEWPORT_SIZE = 100;
+const CONTENT_SIZE = 200;
 
 describe('<ScrollArea.Corner />', () => {
   const { render } = createRenderer();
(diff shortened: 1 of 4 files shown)


atomiks:
Replaced with https://github.com/mui/base-ui/pull/4107. This PR reintroduces a flicker which I'm fixing in that one instead.

PR #3917, opened 2026-02-01
[number field] Fix pen pointer handling

### LINKED ISSUE #1528: [NumberField] Increment/decrement buttons toggle virtual keyboard when pressed with stylus
Tested on iPad with Apple Pencil:
Pressing the increment/decrement buttons with a stylus causes the virtual keyboard to appear and disappear. It seems that not every press makes this happen, but usually it's the case.

The problem doesn't occur using touch or a mouse.

https://github.com/user-attachments/assets/007564cb-aa53-49f9-ac85-23a30c5ecbc6

atomiks:

Fixes #1528. Linux Chrome has (had?) a bug where `"pen"` was emitted for mouse usage


packages/react/src/number-field/root/useNumberFieldButton.ts
```diff
@@ -53,6 +54,11 @@ export function useNumberFieldButton(params: useNumberFieldButton.Parameters) {
     ? 'increment-press'
     : 'decrement-press';
 
+  // "pen" maps to mouse-like behavior on Linux Chrome, so only treat it as touch on mobile.
+  function isTouchPointerType(pointerType: string) {
```
mnajdova: Could this create issues with devices like Windows Surface when actually using a pen? Is neither ios nor android and it should return the pen pointer type as far as I know. From what I understand the intent is not to exclude the desktop stylus devices, right?
I don't have a device to test it, but it could be a common use-case.

Some better check if the above intent is correct is to specifically check for Linux + Chrome desktop:

```js
const ua = navigator.userAgent;

const isChrome =
  /Chrome/.test(ua) && !/Edg|OPR/.test(ua);

const isDesktopLinux =
  /Linux/.test(ua) && !/Android/.test(ua);

const isChromeOnLinux = isChrome && isDesktopLinux;
```

atomiks: The main logic to avoid here is focusing the input when pressing a stepper button. The problem is related to the software keyboard opening, which is why iOS/Android are directly targeted. Desktops don't have the software keyboard issue

mnajdova: > Desktops don't have the software keyboard issue

Windows Surface can be used as a tablet (without keyboard), in that use-case it has the virtual keyboard (the screen can be detached from the hardware keyboard).

mnajdova: In general, if we know Linux + Chrome is the issue, let's be specific and add a comment with the known issue so we track what exactly was broken.

atomiks: I adjusted the approach here:

```js
// Treat pen as touch-like to avoid forcing the software keyboard on stylus taps.
// Linux Chrome may emit "pen" historically for mouse usage due to a bug, but the touch path
// still works with minor behavioral differences.
function isTouchLikePointerType(pointerType: string) {
  return pointerType === 'touch' || pointerType === 'pen';
}
```

So `pen` and `touch` are considered the same path, and the input won't receive focus and open any virtual keyboards, no matter the device

PR #3918, opened 2026-02-01
[select] Avoid instant close due to resize event on open

### LINKED ISSUE #3546: [select] Firefox open the Select will close immediately
# Bug report

## Current behavior

In Firefox browser extension (add-ons) environment open the Select will close immediately. 

**This makes the `Select` totally unusable in Firefox extension.**

https://github.com/user-attachments/assets/014c4a08-dfe8-41b2-9038-df26905456a5

## Expected behavior

Provide a clear and concise description of the expected behavior.

should open the Select and stay there

## Reproducible example

1. clone this: https://github.com/mengxi-ream/read-frog/tree/bug/base-ui
2. run `pnpm i`, run `pnpm build:firefox`
3. go to firefox about:debugging#/runtime/this-firefox
4. load temporary add-ons and choose `firefox-mv3` in `.output` folder of the repo
5. click the frog icon in the firefox toolbar and play `Select`

## Base UI version

v1.0.0

Firefox

## Which OS are you using?

Mac OS

## Additional context

This is similar bug like here: 
https://github.com/radix-ui/primitives/issues/2565
https://github.com/radix-ui/primitives/issues/2754

atomiks:

Fixes #3546

### DIFF
--- a/packages/react/src/select/popup/SelectPopup.tsx
+++ b/packages/react/src/select/popup/SelectPopup.tsx
@@ -33,7 +33,7 @@ import { clamp } from '../../utils/clamp';
 import { useCSPContext } from '../../csp-provider/CSPContext';
 
 const SCROLL_EPS_PX = 1;
-const WINDOW_RESIZE_DISMISSAL_DELAY_MS = 300;
+const WINDOW_RESIZE_DISMISSAL_DELAY_MS = 200;
 
 const stateAttributesMapping: StateAttributesMapping<SelectPopup.State> = {
   ...popupStateMapping,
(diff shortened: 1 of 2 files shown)


atomiks:
> Logic looks good, would be good if someone else can also validate that the 200ms delay, I would increase it too 300-400ms.

When I was testing the resize event occurred ~instantly, so probably shorter can work but this seems safe enough to simplify testing and keep them quick

mnajdova:
> When I was testing the resize event occurred ~instantly, so probably shorter can work but this seems safe enough to simplify testing and keep them quick

In that case, let's go with 200ms.

atomiks:
See https://github.com/mui/base-ui/issues/3546#issuecomment-3883164674

PR #3925, opened 2026-02-02
[popups] Fix page scroll jump when input has focus on unmount in Safari

atomiks:

Not really sure why this occurs, but:

https://deploy-preview-3925--base-ui.netlify.app/react/components/combobox#input-inside-popup

Press trigger with pointer then press <kbd>Esc</kbd> in Safari. Opening with a pointer then selecting an item with the keyboard also has the bug.

If you open with a keyboard instead of a pointer, it doesn't happen. Or typing and then selecting with a keyboard after opening with a pointer.


atomiks:
> I couldn't reproduce this though, do you mean opening with the pointer, then highlighting an item with the pointer then Enter? Also tried opening with pointer, arrow down then enter ;you mention typing after opening with pointer is fine

Yeah, opening with pointer, then pressing arrow down and Enter to select an item had the issue. If you typed something in the input first (just one char) it doesn't happen

mj12albert:
> Yeah, opening with pointer, then pressing arrow down and Enter to select an item

Weird, it doesn't happen to me on Sequoia + Safari 26.2 🙃

atomiks:
> Weird, it doesn't happen to me on Sequoia + Safari 26.2 🙃

You need to have something selected first (doesn't happen initially), then re-open with a pointer — then it happens

PR #3939, opened 2026-02-03
[field][avatar] Add transition attributes

### LINKED ISSUE #3921: [field][avatar] Add transition attributes
# Feature request

## Summary

All applicable components that unmount/mount conditionally should be animatable using `[data-starting-style]`/`[data-ending-style]`.

These seem to be main missing ones (aside from `<Tabs.Panel>` which is being addressed separately)

- `<Field.Error>` (appears/disappears based on validation state)
- `<Avatar.Image>` and `<Avatar.Fallback>` (swap based on image loading state / fallback delay)

The other conditional components missing the attrs should avoid animations to begin with

## Motivation

As mentioned in https://github.com/mui/base-ui/issues/3870

atomiks:

Closes #3921

Also fixes a bug where `<Checkbox.Indicator>` and `<Radio.Indicator>` didn't allow exit animations when `keepMounted=false`

https://deploy-preview-3939--base-ui.netlify.app/experiments/transition-attrs


atomiks:
> Transition support would be useful for Field.Validity as well (e.g. animating a validity status indicator/svg) **though it's a bit different**

Yeah, the issue is `<Field.Validity>` doesn't render its own element to add the attributes onto

> also an experiments nit: could maybe just add transitions to existing experiments (e.g. for checkbox, field error) instead of making an all new one ~

I intentionally left off accordion, collapsible, and popups for this demo since they have their own, but checkbox/radio/avatar etc. don't so I just made a single demo for them all together

mj12albert:
Does it make sense to add transitionStatus to the state called by the children render fn?

atomiks:
@mj12albert yeah it's possible to add as part of the argument passed so children can use it (added demo with it)

The main issue is you can't unmount the children since there's no control over `setMounted(false)` or the ref to detect animations on. As a result `transitionStatus` stays as `'ending'` and the children are kept mounted. I suppose that's okay for simple trees, though

mnajdova:
Please update the PR title and labels to include all components that have updates so the changelog can be generated correctly.

atomiks:
>  Please update the PR title and labels to include all components that have updates so the changelog can be generated correctly.

Yeah the issue is that checkbox/radio were drive-by changes here, they'll need to be split into "bug fixes" when writing release notes @flaviendelangle

ilyinvp:
There seems to be an issue with displaying avatar fallback after these changes (on v1.2.0). If I understood correctly, AvatarImage now shows while the image is loading [this line](https://github.com/mui/base-ui/pull/3939/changes#diff-d1ec612ce20e951f7b823825774e69090f05b4726217e8b3fd7b53273942f031R46), but AvatarFallback can also be shown at the same time in low network conditions when `delayPassed` is `true`: [this line](https://github.com/mui/base-ui/pull/3939/changes#diff-c77a782d5be83415943a6ca6138ab5525887c8675e9e1006b3fc4b8dd4475003R35). This results in both being displayed at the same time causing visual glitches when overflow is not set to hidden on Root.

atomiks:
@wishyoudie thanks for the report, can you make a new issue with a reproduction?

ilyinvp:
@atomiks Thank you for quick reply. Filed #4109.

Issue #3848, opened 2026-01-23
[popover] Allow for a Popover that triggers on hover and can lock the scroll


voishu:
# Feature request

## Summary

Hi! 👋

As far as I understand, it's not possible to have a Popover that triggers on hover and locks the scroll at the same time.
If we combine `Popover.Root[modal=true]` and `Popover.Trigger[openOnHover=true]`, what we get is that the trigger, when clicked, opens the popover and locks the scroll, but when hovered, it opens the popover but doesn't lock the scroll.

It looks like the exact cause is here:
https://github.com/mui/base-ui/blob/master/packages/react/src/popover/root/PopoverRoot.tsx#L83
`openReason !== REASONS.triggerHover`

## Motivation

I think it's a reasonable that we should be able to achieve this sort of behaviour with Popover.
It's also something that I would love for my use case.

Perhaps what would be best is that we could get a manual way of specifying if we would like to lock the scroll?

atomiks:
@voishu can you explain more about your use case for this?

voishu:
> [@voishu](https://github.com/voishu) can you explain more about your use case for this?

Hi, sorry for the late reply.
My use case is very simple. In the web app that I work on, we want to lock the scroll of the page when we show a popover that triggers on hover. This is a design requirement, which looking at it now might not actually make a lot of sense and is not really necessary for anything, just a design choice.

PR #3940, opened 2026-02-03
[popover] Add lockScrollOnHover for a Popover that triggers on hover

markocupic024:
Solves https://github.com/mui/base-ui/issues/3848

Tested in a custom playground, works as expected.


michaldudak:
Thanks for your PR, but it seems to me that this isn't actually the best idea.
I haven't seen valid use cases for it, and we don't want to introduce new props without a good reason. If we receive more upvotes and good justification, we can revive this PR.

PR #3944, opened 2026-02-03
[context menu] Ignore context menu mouseup on non-Mac platforms

### LINKED ISSUE #3595: [context menu] Context menu closes when right-clicking a menu item on Windows

atomiks:

Fixes #3595

On Mac, context menus open on mousedown so they support the down -> move -> release up pattern for selection. On Windows it opens on mouseup, so that's irrelevant. I don't know about Linux, but I've just made it match Windows for simplicity.


packages/react/src/context-menu/root/ContextMenuRoot.test.tsx
```diff
@@ -8,10 +8,22 @@ import {
   waitFor,
 } from '@mui/internal-test-utils';
 import { spy } from 'sinon';
+import { vi } from 'vitest';
 import { ContextMenu } from '@base-ui/react/context-menu';
 import { createRenderer, isJSDOM } from '#test-utils';
 import { REASONS } from '../../utils/reasons';
 
+vi.mock('@base-ui/utils/detectBrowser', async () => {
```
mnajdova, line 16: Let's add a test for non-Mac environment, just to be safe.

packages/react/src/menu/item/useMenuItemCommonProps.ts
```diff
@@ -83,6 +84,10 @@ export function useMenuItemCommonProps(params: UseMenuItemCommonPropsParameters)
           ) {
             return;
           }
+
+          if (isContextMenu && !isMac && event.button === 2) {
```
mnajdova, line 90: Please add a comment explaining the line and why it was added, for people that don't have context it will be hard to track back why we have it.

LukasTy (approved):
LGTM. 👍 
Maybe @flaviendelangle could check the behavior on Linux?

flaviendelangle:
> When a context menu is open, right-clicking on an item inside the context menu immediately closes the menu instead of allowing the item to handle the right-click interaction.

If this is the bug, then I can't reproduce it on `master` and on this PR (Linux Mint - Chromium).

LukasTy:
> > When a context menu is open, right-clicking on an item inside the context menu immediately closes the menu instead of allowing the item to handle the right-click interaction.
> 
> If this is the bug, then I can't reproduce it on `master` and on this PR (Linux Mint - Chromium).

Thank you for checking. 🙏 
Yes, and it looks like it was/is only reproducible on Windows. 👌

PR #3945, opened 2026-02-03
[autocomplete] Remove unnecessary double stringification of item in filtering logic

ZeeshanTamboli:
`contains` method already stringify the item as label. Avoid double doing it.


packages/react/src/autocomplete/root/AutocompleteRoot.tsx
```diff
@@ -85,7 +85,7 @@ export function AutocompleteRoot<ItemValue>(
       return other.filter;
     }
     return (item, query, toString) => {
-      return collator.contains(stringifyAsLabel(item, toString), query);
+      return collator.contains(item, query, toString);
     };
```
atomiks:
```suggestion
    return collator.contains;
```

Issue #3937, opened 2026-02-02
[typescript] Type portability for components extended with react-redux


cgatian:
Repro: https://stackblitz.com/edit/bf5bx4pd-epzkeps4?file=src%2Fbutton.tsx

## Ask a question

> [!NOTE]
> Appologies in advance, I'm sure the topic of TypeScript portable types is messy, but I was hoping since I've seen the team address this in other issues they might have ideas what could be going wrong.

I created a `CustomButton` component that extends the Base UI `Button`.

```tsx
import { Button as BaseButton } from '@base-ui/react';

export const CustomButton: React.FC<CustomButton.Props> = (props) => {
    return <BaseButton {...props}/>
}

export declare namespace CustomButton {
   export interface Props extends BaseButton.Props;
}
```

When consumers of `CustomButton` attempt to extend with their own implementation they receive a portablity issue.

```tsx
interface ProductButtonProps extends Omit<CustomButton.Props, 'onClick'> {
  product: string;
  onClick: (product: string) => void;
}

const ProductButton = ({ product, onClick, ...props }: ProductButtonProps) => {
  const handleOnClick = () => {
    onClick(product);
  };

  return (
      <CustomButton onClick={handleOnClick} {...props}>
        {product}
      </CustomButton>
  );
};

const mapDispatchToProps = {
  onClick: (product: string) => actions.toggleProductTracking({ product }),
};

// ⚠️ TS Error here
export default connect(null, mapDispatchToProps)(TrackingButtonWithLabel);
```

> The inferred type of 'default' cannot be named without a reference to '../../../../../../../node_modules/@base-ui/react/esm/utils/types'. This is likely not portable. A type annotation is necessary.ts(2742)

Any ideas would be welcomed. Thank you 😄 

## Edit

I think because `onClick` and other event handlers use `WithPreventBaseUIHandler` this augments the type of event handlers. Which makes them not portable for consumers not directly importing Base UI. I think my options are to:

1. Undo the Base UI wrapped event handlers using only the Native React event handler types
2. Explicitly type the component with `React.ComponentPropsWithoutRef<'button'>` and only expose Base UI props I want consumers to know about.

CiscoFran10:
@cgatian Could you share a minimal reproducible demo (e.g., StackBlitz or CodeSandbox)? It would really help to investigate the issue.

cgatian:
@CiscoFran10 yeah I'll try. It would need to be a monorepo setup which takes some time.

[...]

PR #3946, opened 2026-02-03
[typescript] Fix type references to internal types

mj12albert:
https://stackblitz.com/edit/bf5bx4pd-byucm7h5?file=src%2FApp.tsx

Not sure if there are better ways but this seems to make `import("@base-ui/react/utils/types").BaseUIEvent` work without exposing all the internal types

Options that didn't work:
- Re-exporting from `src/types/index`: `export type { BaseUIEvent } from '../utils/types'`
- Exporting a duplicate `BaseUIEvent` from `src/types/index` directly, TS throws with an error about duplicate exports

Fixes https://github.com/mui/base-ui/issues/3937

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -68,7 +68,7 @@
     "./types": "./src/types/index.ts",
     "./unstable-use-media-query": "./src/unstable-use-media-query/index.ts",
     "./use-render": "./src/use-render/index.ts",
-    "./utils/types": "./src/utils/types.ts"
+    "./utils/types": "./src/types/types.ts"
   },
   "imports": {
     "#test-utils": "./test/index.ts",
--- a/packages/react/src/types/types.ts
+++ b/packages/react/src/types/types.ts
@@ -1,0 +1,1 @@
+export type { HTMLProps, BaseUIEvent, ComponentRenderFn } from '../utils/types';


packages/react/package.json
```diff
@@ -67,7 +67,8 @@
     "./tooltip": "./src/tooltip/index.ts",
     "./types": "./src/types/index.ts",
     "./unstable-use-media-query": "./src/unstable-use-media-query/index.ts",
-    "./use-render": "./src/use-render/index.ts"
+    "./use-render": "./src/use-render/index.ts",
+    "./utils/types": "./src/types/types.ts"
```
atomiks, line 71: maybe public-types.ts is clearer?

cgatian:
@mj12albert I just tried `762406f` and I'm still getting the error. 

⚠️ I'm not completely confident, but I think it might have something to do with how Button imports the types and how they are exported on the package.json. Specifically this path:

`'../utils/types'`

https://github.com/mui/base-ui/blob/3881e612cb3dbfc67a6f8471b820449d3e405325/packages/react/src/button/Button.tsx#L5

And the export is:
```
"./utils/types": "./src/types/types.ts"
```

Button is using an internal import path, when TypeScript then attempts to locate types at `"../utils/types.js";` and but notices that file is not exported.

If I manually change the emitted `package.json` to this, it's now portable.
```
"./utils/types": {
      "require": {
        "types": "./utils/types.d.ts",
        "default": "./utils/types.js"
      },
      "default": {
        "types": "./esm/utils/types.d.ts",
        "default": "./esm/utils/types.js"
      }
    },
```
Obviously, though this change makes all the types in utils public, and is not desired. Perhaps if the portable types were placed in a file `public-types.ts` and then imports on components (Button) were changed to use this file, TypeScript would be able to resolve them because the import in the d.ts would match the export path on the package.json.

### Edit

My efforts might be better spent in making a monorepo example than giving bad advice. 😸

mj12albert:
@cgatian Maybe it depends on your package layout as well, is yours all in one monorepo like this?
```
packages/a/package.json (Base UI dependency)
packages/a/src/MyButton

packages/b/package.json (package a dependency)
packages/b/src/components/ProductButton
packages/b/src/AppThatUsesProductButton
```
Also you are getting output like `onFocus?: ((event: import("@base-ui/react/utils/types").BaseUIEvent<React.FocusEvent<HTMLButtonElement, Element>>) => void) | undefined` in your `.d.ts` file yea?

cgatian:
The package structure is:
```
internal/ui/package.json (Base UI dependency)
internal/ui/src/MyButton

packages/app/package.json (package a dependency)
packages/app/src/shared/ProductButton
packages/app/src/routes/Product
```

I setup an example repository, but it doesnt seem to be an issue 😖 
https://github.com/cgatian/baseui-portable-repro

Edit: 
@mj12albert Ive spent all day trying to get this to repro in the repository above. I dont understand why this repository works, yet my local fails. Im a little worried about making this change without a better repro. (maybe you have ideas).

cgatian:
@mj12albert a breakthrough! I was able to get this to finally reproduce using [my reproduction repo](https://github.com/cgatian/baseui-portable-repro).
 
image
 
ℹ️  This repository is updated with the latest build in the commit, and the portablity issue is still present.

> Also, I _finally_ figured out what was causing the issue with reproduction. I was using `react-redux: 8.1.3` and the issue is only in the latest version `9.1.0` 😭

packages/react/package.json
LukasTy: Would it make sense to add those exports to the existing `src/types/index.ts` exports instead of creating a new one? 🤔

PR #3947, opened 2026-02-03
[combobox][test] Check for popup close when chip receives focus

atomiks:

Adds a test to verify the popup closes when a chip receives focus (`focusout` of popup)

---

(Edit: this no longer reproduces on master, so changes have been removed from this PR)

With changes to focus management, the popup should explicitly close when a chip receives focus. This fixes the ability to escape the popup on iOS VoiceOver (`aria-hidden` on "outside" elements gets removed). This is a post-v1.1.0 regression


mj12albert:
Is the issue only reproducible with iOS voiceover? @atomiks

atomiks:
@mj12albert nah, I can't reproduce it on master anymore for some reason. I'll probably change this PR to keep the regression test without altering the logic for now.

mj12albert (approved):

PR #3952, opened 2026-02-03
[toast] Support type extension via module augmentation

### LINKED ISSUE #3437: [toast] Is it possible to augment/extend existing type?
Not sure if this is a bug or a misconfiguration on my part so please move this issue to discussion if the behavior is as expected

I would like to constraint the `type` in Toast actions to only have a few string literals so I have the following type declaration in my `base-ui.d.ts`
```ts
/** biome-ignore-all lint/correctness/noUnusedVariables: safe, only augmenting types */
import '@base-ui-components/react'

declare module '@base-ui-components/react/toast' {
  interface ToastObject<Data extends object> {
    type?: 'default' | 'info' | 'error' | 'success'
  }
}
```
But the hints from `toast.add({ type: /**/ })` is still showing as the default `string | undefined`, even though when press the "Go to definition" key from the IDE, the editor shows both the library code and my `d.ts` file, which means the LSP properly picked up my types file

Image

mdm317:

fixes #3437

How about allowing users to define their own `toast` types via `module augmentation`?

I implemented a rough version of it. If you think this approach looks good, I can continue refining it.

### DIFF
--- a/packages/react/src/toast/action/ToastAction.tsx
+++ b/packages/react/src/toast/action/ToastAction.tsx
@@ -4,6 +4,7 @@ import type { BaseUIComponentProps, NativeButtonProps } from '../../utils/types'
 import { useToastRootContext } from '../root/ToastRootContext';
 import { useButton } from '../../use-button/useButton';
 import { useRenderElement } from '../../utils/useRenderElement';
+import { ToastType } from '../useToastManager';
 
 /**
  * Performs an action when clicked.
@@ -55,7 +56,7 @@ export interface ToastActionState {
   /**
    * The type of the toast.
    */
-  type: string | undefined;
+  type: ToastType
 }
 
 export interface ToastActionProps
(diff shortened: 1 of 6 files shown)


michaldudak:
Thanks for looking into this. I agree that a need for custom types is real: `type` is a public styling/rendering field, so design systems may want literal completion and validation for it.

I don’t think module augmentation through `UserToastObject` is right, though. It makes toast types global, and it doesn’t account for existing built-in `promise()` behavior, which writes `loading`, `success`, and `error` to `type` and treats `loading` specially.

I’d prefer a typed-manager approach instead, probably a second generic on `useToastManager`, `createToastManager`, `ToastObject`, and the add/update/promise option types, backed by type tests. Let’s not merge this PR in its current form.

CC @atomiks

mdm317:
Thanks for the detailed review and for taking the time to look into this.

This was the wrong approach on my side. I appreciate the feedback and the direction here.

I’ll close this PR in its current form.

Issue #3951, opened 2026-02-03
[combobox] Type safety and type inference


CRC32EX:
# Get help

## Ask a question
Why `items` is defined `any`  type?
I need TypeScript type safety.

https://github.com/mui/base-ui/blob/3881e612cb3dbfc67a6f8471b820449d3e405325/packages/react/src/combobox/root/AriaCombobox.tsx#L1406

# Related isuses
- https://github.com/mui/base-ui/issues/3756

mj12albert:
You can use a generic, then it can for example infer the type of the value in `onValueChange`:
```tsx
interface Country {
  code: string;
  value: string;
  continent: string;
  label: string;
}

const countries: Country[] = [/* array of countries */]

<Combobox.Root<Country> items={countries} />
```

Additionally here's the doc about applying this to a wrapper component: https://base-ui.com/react/components/combobox#typed-wrapper-component

Is this what you're looking for? @CRC32EX

CRC32EX:
@mj12albert 
https://stackblitz.com/edit/l1dajftb-4vyx7zwa?file=src%2FApp.tsx

I'm tried `<Combobox.Root<ProgrammingLanguage>` but `language` is `any` type

<img width="652" height="318" alt="Image" src="https://github.com/user-attachments/assets/b4aae782-564a-4fb1-b170-5ec1bd6351ce" />

mj12albert:
Actually the root should be able to infer from value or defaultValue without the generic; but with the generic, it can't be passed through React context to `Combobox.List` so the simplest and recommended way is to define it inline like the demos

atomiks:
@CRC32EX with compound components, it's not possible to infer the type from the function there. You can make a wrapper around Combobox that has strict typesafety though if the function `children` is a child of the root `Combobox` of your wrapper as it'll be part of the same props interface/generic. 

`items` can be anything (arbitrary objects/primitives), but in #3954 I've made changes to `items` to be stricter regarding what's allowed with `value`/`defaultValue`

flaviendelangle:
It might be worth considering the following pattern (as opt in) for people who really want type safety:

```tsx
import { Combobox as BaseCombobox } from '@base-ui/react/combobox';

const Combobox = BaseCombobox.createTypeSafeParts<MyValue>();

<Combobox.Root>
  <Combobox.List />
</Combobox.Root>
```

Without that, as @atomiks said, it's a limitation of compound components.

Issue #3865, opened 2026-01-26
[combobox] value gets cleared if shape of value differs from items


freshgiammi:
# Bug report

## Current behavior

Currently, there seems to be an issue where the Combobox's value get cleared everytime the user performs a selection if the shape of `items` does not match the one of `value`.

A problematic usage looks like this:

```
type User = { id: string; name: string; email: string }

const users: User[] = [
  { id: "u1", name: "Alice Johnson", email: "alice.johnson@example.com" },
  { id: "u2", name: "Bob Smith", email: "bob.smith@example.com" },
  { id: "u3", name: "Charlie Lee", email: "charlie.lee@example.com" }
]

return (
  <Combobox.Root items={users}>
    {/* when rendering items: */}
    {(item: User) => <Combobox.Item value={item.id}>{item.name}</Combobox.Item>}
  </Combobox.Root>
)


```

Selecting items that set the value as something different that the full item cause the value to be updated, but removed immediately after - the origin of this issues seems to be [this effect](https://github.com/mui/base-ui/blob/50e164e86fb6e78c1d82c710187922b28a792874/packages/react/src/combobox/root/AriaCombobox.tsx#L861).

Additionally, the effect invokes `isItemEqualToValue` with a possibly invalid first argument (at least, considering [this shape](https://github.com/mui/base-ui/blob/master/packages/react/src/combobox/root/ComboboxRoot.tsx#L87)) since `Value` might be different to the items value provided to `Combobox.Root`. 

#3824 fixes this by removing the effect, but I'm opening an issue anyway since this appears to be a different issue to the two that the PR fixes.

## Expected behavior

The Combobox shouldn't clear valid values only because the `items` and `value` objects differ (or if it should, the signature for `isItemEqualToValue` should be less restricted)

## Reproducible example

https://stackblitz.com/edit/yichkwiw?file=src%2FApp.tsx

## Base UI version

v1.1.0

## Which browser are you using?

Chrome

## Which OS are you using?

MacOS

## Which assistive tech are you using (if applicable)?

Voiceover/JAWS etc.

## Additional context

Provide any additional context that might help us identify the problem and find a solution.

atomiks:
> Additionally, the effect invokes isItemEqualToValue with a possibly invalid first argument (at least, considering [this shape](https://github.com/mui/base-ui/blob/master/packages/react/src/combobox/root/ComboboxRoot.tsx#L87)) since Value might be different to the items value provided to Combobox.Root.

[...]

PR #3954, opened 2026-02-03
[combobox] Support primitive value keys with object items

### LINKED ISSUE #3853: [docs][autocomplete] Confusing TypeScript inference section
# Docs feedback

The documentation at
currently states:

atomiks:

Fixes https://github.com/mui/base-ui/issues/3865
Fixes https://github.com/mui/base-ui/issues/3951
Closes #3853

If the  items in the `items` array are objects shaped as `{value, label}`, this supports a `string`/`number` `defaultValue`/`value`, as with Select. However, an explicit `valueMode="value"` must be specified on the root, as there are a number of edge cases when trying to "infer" the value mode automatically when `items` contains objects, and primitives are passed to `<Combobox.Item value={...}>`.

This improves type inference, so Combobox will error if `items` + `defaultValue`/`value` are of an unsupported shape. This didn't error before but runtime behavior was already broken.

To support arbitrary object properties, there'd need to be a new prop/getter but I think this can be omitted (for now) unless considered necessary by users

New value mode docs: http://deploy-preview-3954--base-ui.netlify.app/react/components/combobox#value-modes

### DIFF
--- a/packages/react/src/combobox/item/ComboboxItem.tsx
+++ b/packages/react/src/combobox/item/ComboboxItem.tsx
@@ -54,7 +54,6 @@ export const ComboboxItem = React.memo(
     const { flatFilteredItems, flatFilteredValues, valueMode, hasItems } =
       useComboboxDerivedItemsContext();
 
-    const items = useStore(store, selectors.items);
     const open = useStore(store, selectors.open);
     const selectionMode = useStore(store, selectors.selectionMode);
     const readOnly = useStore(store, selectors.readOnly);


docs/reference/generated/combobox-root.json
```diff
@@ -76,9 +76,9 @@
       "detailedType": "| ((\n    itemValue: Value,\n    query: string,\n    itemToString:\n      | ((itemValue: Value) => string)\n      | undefined,\n  ) => boolean)\n| null\n| undefined"
     },
     "filteredItems": {
-      "type": "any[] | Group[]",
+      "type": "ComboboxItems<Value>",
       "description": "Filtered items to display in the list.\nWhen provided, the list will use these items instead of filtering the `items` prop internally.\nUse when you want to control filtering logic externally with the `useFilter()` hook.",
-      "detailedType": "any[] | Group[] | undefined"
+      "detailedType": "| NoInfer<\n    | unknown\n    | ((args: any[]) => any)\n    | (any[] & Value)\n    | any\n    | {}\n    | string\n    | number\n    | boolean\n    | bigint\n    | Value\n    | any\n  >[]\n| Combobox.Group<\n    NoInfer<\n      | any\n    >\n  >[]\n| LabeledValue<\n    NoInfer<\n      | (string & unknown)\n      | (number & unknown)\n| Combobox.Group<\n    LabeledValue<\n      NoInfer<\n        | any\n      >\n    >\n  >[]\n| undefined"
```
michaldudak: This looks awful in the docs, but I guess it's better than not having it at all.

packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -5382,4 +5384,550 @@ describe('<Combobox.Root />', () => {
...
+      const items: FruitItem[] = [
+        { value: 'apple', label: 'Apple' },
+        { value: 'banana', label: 'Banana' },
+      ];
+      const onItemHighlighted = spy();
+
+      const { user } = await render(
+        <Combobox.Root
+          items={items}
+          defaultValue="apple"
```
michaldudak, line 5981: Could you also add tests that verify the behavior of the primitive `value` prop?

atomiks:
> There's a potential bug Claude found if ever autoHighlight="always" is exposed from the Combobox API:

Yeah this is okay, it will never support it (it's an Autocomplete-only thing). The effect is needed for several reasons and can't be derived from `items` (I tried that).

docs/reference/generated/combobox-root.json
atomiks: Yep :\

packages/react/src/combobox/root/ComboboxRoot.test.tsx
atomiks: Added

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
```diff
@@ -106,19 +109,6 @@ export const ComboboxTrigger = React.forwardRef(function ComboboxTrigger(
     }
   }, [triggerElement, domReference, floatingRootContext, inputInsidePopup]);
 
-  const { reference: triggerTypeaheadProps } = useTypeahead(floatingRootContext, {
-    enabled: !open && !readOnly && !comboboxDisabled && selectionMode === 'single',
```
atomiks, line 110: Removed this because of an ambiguity that cannot be resolved:

If `selectedValue == null` and the popup hasn't mounted, there's no way to tell what the "item value mode" is (`<Combobox.Item value={item}` or `<Combobox.Item value={item.value}`) since both are supported with items shaped `{label, value}`.

Instead, it now opens the popup and fills the search input with the letter, which is also better UX-wise for comboboxes in my view. The ARIA pattern also suggests this should be done for listboxes (open the popup), which differs from native (macOS) `<select>` behavior

michaldudak:
I'm not sure I like `valueMode="value"`. How about `itemShape: 'valueAndLabel' | 'any'`?

Also, a couple of findings from Codex:

[P1] `valueMode="value"` can silently degrade to `'item'` for valid grouped data with an empty first group.
In resolveValueLabel.tsx (line 46), `getFirstFlatItem()` returns `items[0]?.items[0]` for grouped items.
In AriaCombobox.tsx (line 261), `itemValueMode` is inferred only from that single item. If the first group is empty but later groups contain { value, label } items, mode falls back to `'item'`, and keyboard selection/highlight paths use object items instead of primitive values (wrong `onItemHighlighted`/`onValueChange` shape and selected matching).

[P2] `filteredItems` is ignored for mode inference when items is an empty array.
In AriaCombobox.tsx (line 261), inference uses items ?? filteredItemsProp. An empty array is non-nullish, so filteredItemsProp is skipped. In async/external filtering setups (`items=[]` initially, `filteredItems` populated), this causes the same fallback-to-'item' behavior and mismatched emitted value shape.

atomiks:
> I'm not sure I like valueMode="value". How about itemShape: 'valueAndLabel' | 'any'?

`valueAndLabel | any` sounds more confusing to me 🤔.

I think the value mode docs describe it well: it really is about what "mode" the Root's `value` or `defaultValue` prop is based on its `items` array, and it's linked to whether the full "item" or the item's "value" key is being passed to `Combobox.Item`'s `value` prop:

- `<Combobox.Item value={item}>` -> items are type `item[]` (can be primitive, object, anything)
- `<Combobox.Item value={item.value}>` -> items are type `{value, label}[]` (value is a primitive, items are objects of the normative shape). A getter like `getItemValue` could be introduced to support custom object shapes, if requested.

Maybe this could be `valueShape`, though, but a benefit of "mode" is that it also avoids implying that all item objects must be a strict shape.

PR #3955, opened 2026-02-03
[popups] Apply `data-base-ui-inert` to highest level node

### LINKED ISSUE #3950: Adding data-base-ui-inert to DOM elements when floating elements open makes component usage restrictive and unpredictable
# Get help

## Ask a question

I have been using base-ui with tiptap/prosemirror library. It tracks the content editable area for mutations and sync or recalculate it's state if DOM changes.

If I am using nodeView or reactNodeView to render anything inside this area then I need to be very careful of where I place my Popover/Menu trigger or portal as they will add data-base-ui-inert to all the sibling which will trigger the resync with prosemirror

Having these components change the DOM unexpectedly makes it volatile to use as it's a side-effect which may result in unexpected behaviours especially when DOM is being observed by me or any third party library.

There must be a better way to do this, according to my understanding this approach is used to -
1. Trap focus inside Popup and avoid keyboard navigation to select elements outside of Popup area
2. Enable click on an element outside popover area and not register it as outside click if it doesn't have the inert attribute in it's ancestor

Outside click have been solved by hundreds of component libraries, can't those be considered, just want to understand the approach and if there's any fundamental issue that I am missing

Thank you for support!

atomiks:

Fixes #3950

This should avoid interfering with nodes inside editors while continuing to support 3rd party extension elements (as it's assumed they're portaled to `<body>`), and improves perf as well since a smaller number of elements are mutated (with the common portaled popup case)

### DIFF
--- a/packages/react/src/floating-ui-react/utils/markOthers.test.ts
+++ b/packages/react/src/floating-ui-react/utils/markOthers.test.ts
@@ -1,0 +1,401 @@
+import { markOthers } from './markOthers';
+
+afterEach(() => {
+  document.body.innerHTML = '';
+});
+
+test('single call', () => {
+  const other = document.createElement('div');
+  document.body.appendChild(other);
+  const target = document.createElement('div');
+  document.body.appendChild(target);
+
+  const cleanup = markOthers([target], { ariaHidden: true });
+
+  expect(other.getAttribute('aria-hidden')).toBe('true');
+
+  cleanup();
+
+  expect(other.getAttribute('aria-hidden')).toBe(null);
+});
+
+test('multiple calls', () => {
+  const other = document.createElement('div');
+  document.body.appendChild(other);
+  const target = document.createElement('div');
+  document.body.appendChild(target);
+
+  const cleanup = markOthers([target], { ariaHidden: true });
+
+  expect(other.getAttribute('aria-hidden')).toBe('true');
+
+  const nextTarget = document.createElement('div');
+  document.body.appendChild(nextTarget);
+
+  const nextCleanup = markOthers([nextTarget], { ariaHidden: true });
+
+  expect(target.getAttribute('aria-hidden')).toBe('true');
+  expect(nextTarget.getAttribute('aria-hidden')).toBe(null);
+
+  document.body.removeChild(nextTarget);
+
+  nextCleanup();
+
+  expect(target.getAttribute('aria-hidden')).toBe(null);
+  expect(other.getAttribute('aria-hidden')).toBe('true');
+
+  cleanup();
+
+  expect(other.getAttribute('aria-hidden')).toBe(null);
+
+  document.body.appendChild(nextTarget);
+});
+
+test('out of order cleanup', () => {
+  const other = document.createElement('div');
+  document.body.appendChild(other);
+  const target = document.createElement('div');
+  target.setAttribute('data-testid', '');
+  document.body.appendChild(target);
+
+  const cleanup = markOthers([target], { ariaHidden: true });
+
+  expect(other.getAttribute('aria-hidden')).toBe('true');
+
+  const nextTarget = document.createElement('div');
+  document.body.appendChild(nextTarget);
+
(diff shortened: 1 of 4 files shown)


packages/react/src/floating-ui-react/utils/markOthers.test.ts
```diff
@@ -0,0 +1,185 @@
+import { markOthers } from './markOthers';
+
```
atomiks, line 2: This test is ported from `@floating-ui/react`

mech:
Seems to have this on my AlertDialog (both 1.1.1 and 1.1.2) and Autocomplete (only 1.12) happening.

CleanShot 2026-02-27 at 09 11 20@2x

At a lost what to do.

atomiks:
@mech this is unrelated to this PR. It can happen in some scenarios like this: https://github.com/mui/base-ui/issues/4205. If you have a reproducible example, make an issue so we can look into it and fix 🙏

mech:
Wow, you are right. Once I remove my `<ScrollArea>` inside both my `<AlertDialog>` and `<Autocomplete>`, it is gone. Something's wrong with the `<ScrollArea>` I believe. Will investigate more on my side.

CleanShot 2026-02-27 at 10 49 34@2x

atomiks:
@mech it's an edge casey timing bug with focus trap and the scroll area tabIndex + measurements. It's on our side (see #4220). To fix it, just add `tabIndex={0}` to the Viewport part in the meantime as a workaround.

mech:
Tks I have indeed add the tabIndex as a temp fix 👏

atomiks:
## Codex Review

### Overview
This branch refactors `markOthers` to separate marker behavior from control-attribute behavior, updates `FloatingFocusManager` integration, and adds regression coverage for overlapping cleanup, shadow-root host unwrapping, and marker targeting around reference/focus-guard structures. The intended behavior remains: top-level marker application with preserved `aria-hidden`/`inert` cleanup correctness.

### Findings (None)

### Confidence: 4/5
High confidence in functional correctness from full-branch review and test coverage; moderate uncertainty remains on end-to-end perf due normal benchmark variance.

### Notes
- Latest reviewed commit: `f57808e0d`.
- Validation reruns:
  - `pnpm test:jsdom markOthers FloatingFocusManager --no-watch`
  - `pnpm test:chromium markOthers --no-watch`

Final stress review (`Popover.Root` open+close, `modal=true`, 3000 top-level outside nodes), two benchmark passes averaged:

| Metric | Master (`94474b438`) | Branch (`f57808e0d`) | Delta |
|---|---:|---:|---:|
| marker nodes mean | 3304.0 | 3002.0 | -9.1% |
| aria-hidden nodes mean | 1.0 | 1.0 | 0.0% |
| open+close mean (ms) | 270.6882 | 252.0050 | -6.9% |
| open+close median (ms) | 267.2741 | 248.4421 | -7.0% |
| open+close p95 (ms) | 287.4924 | 276.5797 | -3.8% |

Regular hero-like Popover setup rerun (single docs-style hero popover + surrounding demo DOM), 5 alternating passes per side averaged:

| Metric | Master (`94474b438`) | Branch (`f57808e0d`) | Delta |
|---|---:|---:|---:|
| marker nodes mean | 12.0 | 2.0 | -83.3% |
| aria-hidden nodes mean | 2.0 | 2.0 | 0.0% |
| open+close mean (ms) | 16.6264 | 15.5860 | -6.3% |
| open+close median (ms) | 16.2191 | 15.1946 | -6.3% |
| open+close p95 (ms) | 18.8188 | 17.4449 | -7.3% |

Rerun variance note (hero-like): branch mean stddev `0.42ms` vs master `1.35ms`; this run had larger variance on master.

ziyadshaikh013:
Thank you for adressing this issue, I dropped it last time to make the delivery, but got back to it and after bumping to latest base-ui version with this fix it worked as expected. Base-ui not longer mutates elements inside editor boundary which was causing the document rebuild

PR #3960, opened 2026-02-04
[slider] Fix change event cloning

### LINKED ISSUE #3953: [slider] Component fails in Shadow DOM: Cannot read properties of undefined (reading 'constructor')
The Slider component throws an error when used inside a Shadow DOM context due to event cloning logic that assumes `event.constructor` is always available.

### Error Message:
```
Cannot read properties of undefined (reading 'constructor')
```

### Environment:

Package: @base-ui/react
Component: [Slider](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Context: Application running inside a Shadow Root (Chrome extension)

### Root cause: 

The error occurs in [SliderRoot.tsx](https://github.com/mui/base-ui/blob/3881e612cb3dbfc67a6f8471b820449d3e405325/packages/react/src/slider/root/SliderRoot.tsx#L227) at lines 224-227~:

```
// Clone the event to not override `target` of the original event.
// @ts-expect-error The nativeEvent is function, not object
const clonedEvent = new event.constructor(event.type, event);

Object.defineProperty(clonedEvent, 'target', {
  writable: true,
  value: { value: newValue, name },
});
```

When events cross Shadow DOM boundaries, the `event.constructor` or `event` itself can be undefined, causing the code to fail when attempting `new event.constructor(...)`.

### Reproduction:

Create a Shadow DOM root
Render the example Slider implementation from the [docs](https://base-ui.com/react/components/slider) in a shadowroot, then focus the slider & try to adjust its value using the arrow keys. An error is fired per keyboard event fired

### Expected Behavior:

The Slider component should work correctly in Shadow DOM environments, gracefully handling cases where `event.constructor` or `event` are unavailable.

### Proposed Solution:

Add some form of fallback for this event construction, such as:

```
const EventConstructor = event.constructor || Event;
const clonedEvent = new EventConstructor(event.type, event);
```

with whatever level of error handling that is considered acceptable. Cheers

atomiks:

Fixes #3953

### DIFF
--- a/packages/react/src/slider/root/SliderRoot.test.tsx
+++ b/packages/react/src/slider/root/SliderRoot.test.tsx
@@ -1294,47 +1294,97 @@ describe.skipIf(typeof Touch === 'undefined')('<Slider.Root />', () => {
       });
     });
 
-    it('should not rely on the global event when cloning change events', async () => {
-      const hadGlobalEvent = Object.prototype.hasOwnProperty.call(globalThis, 'event');
-      const previousGlobalEvent = (globalThis as any).event;
-      const globalEventConstructor = class {
-        constructor() {
-          throw new Error('Should not construct global event');
+    it.skipIf(!isJSDOM)(
+      'should not rely on the global event when cloning change events',
+      async () => {
+        const hadGlobalEvent = Object.prototype.hasOwnProperty.call(globalThis, 'event');
+        const previousDescriptor = Object.getOwnPropertyDescriptor(globalThis, 'event');
+        const globalEventConstructor = class {
+          constructor() {
+            throw new Error('Should not construct global event');
+          }
+        };
+        const fakeGlobalEvent = {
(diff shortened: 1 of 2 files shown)


packages/react/src/slider/root/SliderRoot.test.tsx
```diff
@@ -1,6 +1,7 @@
 import { expect } from 'chai';
 import * as React from 'react';
 import { spy, stub } from 'sinon';
+import { expect as expectVitest, vi } from 'vitest';
```
mj12albert, line 4: Oh I just noticed our AGENTS.md says this
https://github.com/mui/base-ui/blob/b59d982cea0bffab2b4122e703e88a3affd5ffe8/AGENTS.md?plain=1#L34

PR #3965, opened 2026-02-04
[collapsible] Fix transitions when inside a parent `<Activity>`

mj12albert:
Noticed that collapsible has issues when inside `React.Activity`, similar to  https://github.com/radix-ui/primitives/issues/3778 though it glitches differently

Preview
- CSS transitions: https://deploy-preview-3965--base-ui.netlify.app/experiments/collapsible/transitions-activity
- CSS animations: https://deploy-preview-3965--base-ui.netlify.app/experiments/collapsible/animations-activity

Note this doesn't solve integrating the collapsible's open state with `<Activity>`, I don't think it's possible since it applies `style="display: none !important"` which interferes with the logic, motion even provides an `AnimatedActivity` component as a replacement: https://motion.dev/docs/react-animate-activity


BAM-hub (approved):

atomiks (approved):
Couldn't find issues in the experiments pages or logic, but the logic itself in this hook is getting increasingly complex and should likely be refactored (in a separate PR). 

Ran it through Codex and it gave suggestions:

1. Guard `transition-duration` cleanup so user styles are preserved  
In `packages/react/src/collapsible/panel/useCollapsiblePanel.ts`, only remove/restore `transition-duration` if Base UI set it.

(This one is a pre-existing issue afaik)

```ts
const forcedTransitionDurationRef = React.useRef<string | null>(null);

function forceZeroDuration(panel: HTMLElement) {
  if (forcedTransitionDurationRef.current !== null) return;
  forcedTransitionDurationRef.current = panel.style.transitionDuration;
  panel.style.setProperty('transition-duration', '0s');
}

function restoreDuration(panel: HTMLElement) {
  const prev = forcedTransitionDurationRef.current;
  if (prev === null) return;
  forcedTransitionDurationRef.current = null;

  if (prev === '') {
    panel.style.removeProperty('transition-duration');
  } else {
    panel.style.setProperty('transition-duration', prev);
  }
}
```

2. Replace multiple refs with one explicit “resumed from Activity” signal  
Track one resume condition and consume it once in open paths.

```ts
const prevStatusRef = React.useRef(transitionStatus);
const resumedFromActivityRef = React.useRef(false);

React.useEffect(() => {
  resumedFromActivityRef.current =
    prevStatusRef.current === 'starting' &&
    transitionStatus === 'idle' &&
    open &&
    mounted;
  prevStatusRef.current = transitionStatus;
}, [transitionStatus, open, mounted]);

if (resumedFromActivityRef.current) {
  resumedFromActivityRef.current = false;
  return; // skip re-measure / re-animation
}
```

3. Split transition and animation logic into separate hooks  
Refactor `useCollapsiblePanel` into focused hooks so each state machine is isolated.

```ts
useTransitionPanelLogic({
  enabled: animationTypeRef.current === 'css-transition',
  // transition-specific params
});

useAnimationPanelLogic({
  enabled: animationTypeRef.current === 'css-animation',
  // animation-specific params
});
```

4. Add Chromium regression tests for Activity reveal/hide cycles  
Add tests in `packages/react/src/collapsible/panel/CollapsiblePanel.test.tsx` to verify reveal does not replay transitions and does not remove user inline duration.

```tsx
describe.skipIf(isJSDOM)('React.Activity', () => {
  it('does not replay transition on reveal', async () => {
    // render Collapsible in <React.Activity /> with defaultOpen
    // hide + reveal activity
    // assert no extra transitionrun and transitionDuration preserved
  });
});
```

atomiks (commented):
## Codex Review (GPT-5.4)

Dominant type: `bug-fix` (secondary: `react-ui`, `docs`). This review is based on the full three-commit branch against a freshly fetched `upstream/master`; the Activity-specific direction looks right overall, but I still see one blocking logic issue in the new transition cleanup path.

### 1. Bugs / Issues

#### 1. 🔴 Activity resume cleanup drops author inline `transition-duration` (blocking)

The new `starting -> idle` cleanup for `React.Activity` resumes always removes `transition-duration` from the panel. If a consumer set an inline duration, that value is lost after the first hide/reveal cycle, so later open/close transitions no longer respect the author’s timing.

`packages/react/src/collapsible/panel/useCollapsiblePanel.ts`
```ts
if (prevStatus !== 'starting' || transitionStatus !== 'idle') {
  return;
}

if (animationTypeRef.current !== 'css-transition') {
  return;
}

const panel = panelRef.current;
if (!panel) {
  return;
}

shouldCancelInitialOpenTransitionRef.current = false;
shouldCancelInitialOpenAnimationRef.current = false;

panel.style.removeProperty('transition-duration');
```

That is a real behavior change on the exact path this PR is adding support for. The temporary `0s` override needs to be restored only when Base UI set it, not unconditionally cleared.

**Fix:** Store the previous inline `transition-duration` before forcing `0s`, then restore that value on cleanup; only call `removeProperty` when the previous value was empty.

The core transition fix makes sense: using `transitionStatus` to detect an Activity reveal and skipping the re-measure path should prevent transition replay. The animation-side early return is also aligned with the reported bug. The remaining problem is just that the cleanup path is too aggressive and leaks into author styling.

### Test Coverage Assessment

I did not see automated regression coverage for the new Activity path in the branch. For this kind of lifecycle fix in a shared hook, a Chromium regression test would materially reduce the risk of reintroducing the issue.

### Recommendation

Request changes 🔴

I’d block on the inline-style cleanup issue above. Once `transition-duration` is restored rather than always removed, this looks much closer to ready; adding a focused Activity regression test would make the change considerably safer as well.

mj12albert:
Continued in https://github.com/mui/base-ui/pull/4565

PR #541, opened 2024-08-15
[select] Create new `Select` component

atomiks:
Preview: https://deploy-preview-541--base-ui.netlify.app/components/react-select/

- Closes https://github.com/mui/base-ui/issues/547
- Closes https://github.com/mui/base-ui/issues/476
- Closes https://github.com/mui/base-ui/issues/327
- Closes https://github.com/mui/base-ui/issues/328
- Closes https://github.com/mui/base-ui/issues/217
- Closes https://github.com/mui/base-ui/issues/8
- Closes https://github.com/mui/base-ui/issues/82
- Closes https://github.com/mui/base-ui/issues/66
- Closes https://github.com/mui/base-ui/issues/65
- Closes https://github.com/mui/base-ui/issues/64
- Closes https://github.com/mui/base-ui/issues/56
- Closes https://github.com/mui/base-ui/issues/54 (solved via `Select.Value` `placeholder` prop)
- Closes https://github.com/mui/base-ui/issues/7 (TODO: provide a warning if `defaultValue` / `value` don't match any of the options)
- Closes https://github.com/mui/base-ui/issues/37 (seem the same as #54)
- Closes https://github.com/mui/base-ui/issues/14 (now invalid)

TODO:

- [x] Integrate with `useField`
- [x] Renamed `alignMethod` prop to `alignOptionToTrigger`

Olivier's edit: the v2 of https://github.com/mui/material-ui/pull/8023 😄

PR #3969, opened 2026-02-04
[select] Fire onClick during drag-to-select

obeattie:

The drag-to-select behavior which was originally introduced in #541 (mousedown on trigger → drag to item → mouseup to select) bypasses the item's `onClick` handler since the browser doesn't fire a `click` event when mousedown and mouseup occur on different elements. This is inconsistent with the other ways to select an item: both regular clicks and keyboard selection both invoke `onClick` already.

This PR calls `onClick` from `onMouseUp` before committing the selection, allowing `preventBaseUIHandler()` to cancel the selection.

Includes tests. (Sorry about #3968 - I closed it by accident.)


packages/react/src/select/item/SelectItem.tsx
```diff
@@ -229,7 +236,12 @@ export const SelectItem = React.memo(
           return;
         }
 
-        commitSelection(event.nativeEvent);
+        makeEventPreventable(event);
+        componentProps.onClick?.(event);
+
+        if (!event.baseUIHandlerPrevented) {
+          commitSelection(event.nativeEvent);
+        }
```
atomiks: In `MenuItem`, we call `itemRef.current.click()` instead, I wonder if that's better for consistency

obeattie:
@atomiks Thanks for taking a look. Agree that doing `itemRef.current.click()` is the right move here. I'd actually tried that initially, but it exposed another bug that I've now gone ahead and fixed here.

Specifically, when a `Select` is nested inside a `Popover`, drag-to-select when implemented like this was unexpectedly dismissing the popover. What was happening is that the “started inside” drag state could be consumed too early. The fix is to keep that state through click capture, check whether the click target is inside _first_, and only use the one-shot suppression on the first real outside click.

This also shows up outside `Select`. For example, see this (ugly) Stackblitz that exhibits it with a `Menu` inside a `Popover`: https://stackblitz.com/edit/gzrjdbd3?file=src%2FApp.tsx

packages/react/src/select/item/SelectItem.tsx
```diff
@@ -208,11 +214,12 @@ export const SelectItem = React.memo(
         pointerTypeRef.current = event.pointerType;
         didPointerDownRef.current = true;
       },
-      onMouseUp(event) {
+      onMouseUp(event: BaseUIEvent<React.MouseEvent<HTMLDivElement>>) {
```
atomiks:
```suggestion
      onMouseUp(event) {
```

atomiks:
> but it exposed another bug that I've now gone ahead and fixed here.

Oh nice, thanks for the fix & explanation

atomiks:
Couple lint issues:

> [ci/circleci: Linting](https://circleci.com/gh/mui/base-ui/222107) — Your tests failed on CircleCI

packages/react/src/floating-ui-react/hooks/useDismiss.test.tsx
```diff
@@ -868,6 +868,18 @@ describe.skipIf(!isJSDOM)('useDismiss', () => {
       fireEvent.click(document.body);
       expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
     });
+
+    test('inside click does not suppress the next outside click when no drag occurred', async () => {
```
atomiks: Codex review found a logic regression here:

> a normal inside click in intentional mode can leave endedOrStartedInsideRef set,
so the next outside click is incorrectly ignored once (even without a drag).

I can't reproduce it this on the Popover demo, though. The tests are only failing in JSDOM but not Chromium.

atomiks: @obeattie this seems like an edge case, but theoretically possible to trigger an issue

In-browser popover usage closes because real browser clicks fire the other mouse events (mousedown -> mouseup -> click). And in useDismiss, non-click events clear `endedOrStartedInsideRef`, so by the time outside click runs, it closes normally. 

So it's an issue only if `.click()` runs without the mouse events (tests, assistive tech, programmatic clicks?)

obeattie: Thanks for the review and good catch. The only sane way I could think of to resolve this was to change the suppression logic to better tie the suppression to actual drag sequences.

Now, two flags are used for this:

* `pressStartedInsideRef`: Tracks whether there's an active press that began inside. Always cleared in `pointerup`/`mouseup`
* `suppressNextOutsideClickRef`: Tracks whether we want to ignore an outside `click` event – set on `pointerup`/`mouseup` (which fire before `click` and won't fire in the synthetic cases like the one we're now firing for drag-to-select).

Hopefully this works better!

atomiks:
Simplified the `useDismiss` logic further to avoid a (minzip) bundle size increase, some existing logic wasn't necessary

atomiks (approved):

atomiks:
@greptileai

packages/react/src/select/item/SelectItem.tsx
```diff
@@ -229,7 +230,7 @@ export const SelectItem = React.memo(
           return;
         }
 
-        commitSelection(event.nativeEvent);
+        itemRef.current?.click();
       },
```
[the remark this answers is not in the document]
atomiks: Appears invalid

obeattie:
Anything more I can do to help here?

atomiks:
@obeattie nope, sorry, I edited your PR and then found various issues caused by my edits, and spent awhile refining it the other day. I _think_ its current state is good but I left it off the last release because it changes `useDismiss` significantly and wanted to manually test it more thoroughly

obeattie:
No worries - thank you for your effort!

PR #3973, opened 2026-02-04
[combobox] Clear highlight on inline blur when inline

### LINKED ISSUE #3967: [combobox] `data-highlighted` from items should be removed when inline and focused out
# Bug report

## Current behavior

Items stay highlighted when you focus out of an inline combobox.

https://github.com/user-attachments/assets/d3443d53-4acf-4a9d-be97-fc5a894f9b26

## Expected behavior

`data-highlighted` should be removed when you focus out?

## Reproducible example

This is reproducible in Base UI [combobox in a dialog example](https://base-ui.com/experiments/combobox/dialog-combobox-multiple).

## Base UI version

v1.1.0

atomiks:

Fixes #3967


atomiks:
Yeah I think that makes sense. 

We were going to build a separate Listbox component at some point, but an inline combobox gets most of the way there already

Issue #3958, opened 2026-02-04
[tooltip] Flash at origin before positioning completes in Preact


OliverSpeir:
Floating elements (Tooltip, Popover, etc.) briefly flash at position `0,0` before moving to their correct
  position. This is visible as a flicker in the top-left corner when the element first appears.

  ### Reproduction

  https://github.com/OliverSpeir/preact-tooltip-repro

  ![preact demo](https://github.com/OliverSpeir/preact-tooltip-repro/raw/main/preact.gif)

  ### Root cause

  The `floatingStyles` in `useAnchorPositioning` returns position values immediately, before `isPositioned`
  becomes `true`. In preact `useLayoutEffect` timing differs slightly from React,
  this causes a visible frame where the element renders at `0,0`.

  ### Proposed fix

  Use the existing `isPositioned` flag to hide the element until positioning completes:

  ```diff
  - const floatingStyles = React.useMemo(() => adaptiveOrigin ? {
  -   position: resolvedPosition,
  -   [sideX]: x,
  -   [sideY]: y
  - } : {
  -   position: resolvedPosition,
  -   ...originalFloatingStyles
  - }, [adaptiveOrigin, resolvedPosition, sideX, x, sideY, y, originalFloatingStyles]);
  + const floatingStyles = React.useMemo(() => {
  +   const base = adaptiveOrigin
  +     ? { position: resolvedPosition, [sideX]: x, [sideY]: y }
  +     : { position: resolvedPosition, ...originalFloatingStyles };
  +   return !isPositioned ? { ...base, opacity: 0 } : base;
  + }, [adaptiveOrigin, resolvedPosition, sideX, x, sideY, y, originalFloatingStyles, isPositioned]);
  ```

atomiks:
My main concern with `opacity` is it might interfere with animations. Maybe `visibility: 'hidden'` works better?

OliverSpeir:
> My main concern with `opacity` is it might interfere with animations. Maybe `visibility: 'hidden'` works better?


good call, confirmed this fixes it still in the reproduction repo I linked

PR #3975, opened 2026-02-05
[popups] Fix flash at origin before positioning completes in Preact

OliverSpeir:
Went with opacity because visibility hidden affects focus, happy to go another route though of course

Fixes https://github.com/mui/base-ui/issues/3958

### DIFF
--- a/packages/react/src/combobox/chip-remove/ComboboxChipRemove.tsx
+++ b/packages/react/src/combobox/chip-remove/ComboboxChipRemove.tsx
@@ -75,7 +75,7 @@ export const ComboboxChipRemove = React.forwardRef(function ComboboxChipRemove(
           if (removedIndex !== -1 && activeIndex === removedIndex) {
             store.state.setIndices({
               activeIndex: null,
-              type: store.state.keyboardActiveRef.current ? 'pointer' : 'keyboard',
+              type: store.state.keyboardActiveRef.current ? 'keyboard' : 'pointer',
             });
           }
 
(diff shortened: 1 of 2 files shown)


atomiks:
I actually had a PR open https://github.com/mui/base-ui/pull/3963 but hadn't added tests yet. We can keep this one.

Can you change the logic to mutate the object as I've done (for perf)

OliverSpeir:
> I actually had a PR open #3963 but hadn't added tests yet. We can keep this one.
> 
> Can you change the logic to mutate the object as I've done (for perf)

ah whoops, thanks for the looking into this I appreciate it

packages/react/src/utils/useAnchorPositioning.ts
```diff
@@ -422,13 +422,15 @@ export function useAnchorPositioning(
...
-        : { position: resolvedPosition, ...originalFloatingStyles },
-    [adaptiveOrigin, resolvedPosition, sideX, x, sideY, y, originalFloatingStyles],
-  );
+  const floatingStyles = React.useMemo<React.CSSProperties>(() => {
+    const base = adaptiveOrigin
+      ? { position: resolvedPosition, [sideX]: x, [sideY]: y }
+      : { position: resolvedPosition, ...originalFloatingStyles };
+    if (!isPositioned) {
+      base.opacity = 0;
+    }
```
atomiks, line 431: In this PR, can you add `getDisabledMountTransitionStyles` to the components below?

Image

 Tooltip, Popover, PreviewCard, and Navigation Menu have it to explicitly disable transitions on the positioner from mount, which prevents transitions from running:

```jsx
props: [defaultProps, getDisabledMountTransitionStyles(transitionStatus), elementProps],
```

(Only Popup is meant to be animated anyway, so risk is pretty minimal)

atomiks (approved):
Thanks for the fix @OliverSpeir

PR #3980, opened 2026-02-05
[combobox] Fix highlight change reason in `ChipRemove`

sai6855:
Fix type handling in ComboboxChipRemove component

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,6 +24,7 @@ This repository contains the source code and documentation for Base UI: a headl
 - Do not randomly cast (for example `as any`) if there are no type errors without doing so. Run `pnpm typescript` to verify types.
 - Ensure your changes pass linting - run `pnpm eslint`.
 - Ensure your changes are formatted correctly - run `pnpm prettier`.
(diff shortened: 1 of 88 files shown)


packages/react/src/combobox/chip-remove/ComboboxChipRemove.tsx
```diff
@@ -75,7 +75,7 @@ export const ComboboxChipRemove = React.forwardRef(function ComboboxChipRemove(
           if (removedIndex !== -1 && activeIndex === removedIndex) {
             store.state.setIndices({
               activeIndex: null,
-              type: store.state.keyboardActiveRef.current ? 'pointer' : 'keyboard',
+              type: store.state.keyboardActiveRef.current ? 'keyboard' : 'pointer',
```
sai6855, line 78: I was going through code of ComboBox, noticed slight descripency, In 
https://github.com/sai6855/base-ui/blob/48c6258422710b995d2b3fcb4e1ef766dec23612/packages/react/src/combobox/input/ComboboxInput.tsx#L246
 `type` is set to `keyboard` if `store.state.keyboardActiveRef.current` is true but here it's set to `pointer`

PR #3981, opened 2026-02-05
[utils] Fix event handling in useEnhancedClickHandler

sai6855:
Fix event handling in useEnhancedClickHandler to support pointerType


packages/utils/src/useEnhancedClickHandler.ts
```diff
@@ -38,9 +38,9 @@ export function useEnhancedClickHandler(
       if ('pointerType' in event) {
         // Chrome and Edge correctly use PointerEvent
         handler(event, event.pointerType);
+      } else {
```
sai6855, line 41: Added this else block after line 39 to prevent the handler from being
  called twice per click.

atomiks (approved):

Issue #3977, opened 2026-02-05
[combobox] Native way to prevent clearing of input after selection


sarthakmalik0810:
# Feature request

## Summary

I think the Combobox should have a native way to prevent resetting of `inputValue` after selection of an item.

The prop could be called - `clearInputAfterSelection`.

Maybe another prop called - `closePopupAfterSelection` to prevent closing the popup when the input is not in popup.

## Motivation

This would be useful in lists where you want to select multiple options with the similar search query, but I have seen this in other comboboxes as well. Example from Github label selector.

https://github.com/user-attachments/assets/0be6657b-853d-4eec-999a-1c706769fe5c

This is how I would do it right now, but I feel this would be fragile, and may cause issues as the library evolves - 

```jsx
    <Combobox.Root
      items={fruits}
      inputValue={controlledInputValue}
      onInputValueChange={(inputValue, eventDetails) => {
        if (eventDetails.reason === 'input-change') {
          setControlledInputValue(inputValue);
        }
      }}
      multiple
    />
```

mj12albert:
> This would be useful in lists where you want to select multiple options with the similar search query

@sarthakmalik0810 This seems quite similar to https://github.com/mui/base-ui/issues/3181

Is this what you're trying to achieve? https://stackblitz.com/edit/bcocshwo-kkp2dff4?file=src%2FApp.tsx

```tsx
<Combobox.Root
  items={langs}
  multiple
  onOpenChange={(nextOpen, eventDetails) => {
    if (!nextOpen && eventDetails.reason === 'item-press') {
      eventDetails.cancel();
    }
  }}
/>
```

sarthakmalik0810:
@mj12albert That works for when you have popups, but the query will still be reset for `inline` comboboxes. Therefore I think we should have a native way to do this (for both `inline` and `non-inline` comboboxes). Not resetting the query after selection is a valid use case for quite a few comboboxes. These custom checks would be difficult to maintain as the component internal reasons, etc can change through various minor versions of the library.

PR #3985, opened 2026-02-05
[combobox] Add props to prevent popup close and input clear when multiple

markocupic024:
Solves https://github.com/mui/base-ui/issues/3977

https://github.com/user-attachments/assets/1d75635c-ecfb-4d5e-88c9-1a76c22a96f8


flaviendelangle:
@atomiks if you can take a look at this one, I'm missing a lot of context

PR #3986, opened 2026-02-05
[utils] Fix potential EMPTY_OBJECT mutation in useRenderElement

jkker:
## Summary

This PR fixes a latent bug in `useRenderElement` where the frozen `EMPTY_OBJECT` could potentially be mutated, causing a `TypeError` in strict mode.

### The Problem

`EMPTY_OBJECT` is defined as `Object.freeze({})` - a frozen, immutable object. In `useRenderElementProps`, there's a fallback path where `outProps` could be assigned `EMPTY_OBJECT`:

```typescript
const outProps = enabled
  ? (mergeObjects(stateProps, ...) ?? EMPTY_OBJECT)  // <- fallback to frozen object
  : EMPTY_OBJECT;
```

Later, the code attempts to mutate `outProps`:

```typescript
outProps.ref = useMergedRefs(...);      // line 86-88
outProps.className = mergeClassNames(...);  // line 97
outProps.style = mergeObjects(...);         // line 101
```

If `mergeObjects()` returns `undefined`, mutations on the frozen `EMPTY_OBJECT` would:
- **Fail silently in non-strict mode**
- **Throw TypeError in strict mode** (which React uses)

### Current Status

The bug is **latent but not currently triggerable** because `getStateAttributesProps()` always returns a new `{}`. However, the code is fragile and any refactoring could surface this issue.

### The Fix

Replace `?? EMPTY_OBJECT` with `?? {}` to ensure `outProps` is always a fresh mutable object when `enabled=true`.

```diff
- ? (mergeObjects(...) ?? EMPTY_OBJECT)
+ ? (mergeObjects(...) ?? {})
```

### Tests Added

- 4 new tests verifying edge cases with minimal props configuration
- All 4938 existing JSDOM tests pass
- All 18 Chromium tests for useRenderElement pass


romgrk (approved):

romgrk:
@Jkker Is there a reason to keep the PR as a draft, or is it ready for merging?

romgrk:
That variable could be renamed to `mutableProps` to make its requirement more explicit.

oliviertassinari:
> This PR fixes a latent bug in useRenderElement where the frozen EMPTY_OBJECT could potentially be mutated, causing a TypeError in strict mode.

Could we update the tests so they fail without the change to the source?

packages/react/src/utils/useRenderElement.test.tsx
```diff
@@ -255,4 +256,49 @@ describe('useRenderElement', () => {
...
+      // This would throw "Cannot add property style, object is not extensible"
+      // if outProps were the frozen EMPTY_OBJECT
+      const { container } = await render(<MinimalComponent style={{ color: 'red' }} />);
+      expect(container.firstElementChild).to.not.equal(null);
+    });
+
+    it('renders correctly with className when using minimal props', async () => {
+      const { container } = await render(<MinimalComponent className="custom-class" />);
+      const element = container.firstElementChild;
+      expect(element).to.have.attribute('class', 'custom-class');
```
LukasTy: Looks like we could add this assertion after L282 (first test) and remove this almost duplicate test.

packages/react/src/utils/useRenderElement.test.tsx
```diff
@@ -255,4 +256,49 @@ describe('useRenderElement', () => {
...
+    });
+
+    it('renders correctly with style when using minimal props', async () => {
+      const { container } = await render(<MinimalComponent style={{ color: 'rgb(255, 0, 0)' }} />);
+      const element = container.firstElementChild as HTMLElement;
+      expect(element.style.color).to.equal('rgb(255, 0, 0)');
```
LukasTy: The same comment applies here. Seems like this can be combined with the 2nd test.

packages/react/src/utils/useRenderElement.test.tsx
jkker: Folded the assertion into the existing regression test in f4f2feb to avoid the duplicate case. Thanks.

packages/react/src/utils/useRenderElement.test.tsx
jkker: Updated this in the same commit and removed the extra test so the minimal-props coverage stays focused. Thanks.

jkker:
Added regression coverage so the minimal-props cases now fail against the old frozen-EMPTY_OBJECT path, and I just folded @LukasTy  follow-up into the existing tests to remove the duplicate cases. Tests are updated pet @oliviertassinari 's suggestion. CI is passing. @romgrk could you please help take a final look and consider merging it if viable? Thanks!

PR #3996, opened 2026-02-06
[form controls] Clear `data-focused` when disabled

### LINKED ISSUE #3987: [field] data-focused attribute persists when data-disabled
# Bug report / Feature request

With native HTML `<input>` and `<button>`, the elements are not focusable when disabled. Arguably, the same behavior should be reflected in the Base UI form components like `Field`, `Input`, `Select`, `Autocomplete`, `Combobox`, etc. This could be achieved by removing `data-focused`, while `data-disabled` is present.

The current workaround when using Tailwind is to use `not-data-disabled:data-focused: ...` modifiers, but I don't think this is something consumers should have to implement. I don't think users expect `data-focused` to persist when `data-disabled`.

I do like that the `Button` component has the `focusableWhenDisabled` prop and it correctly defaults to `false`. This could be added to the form-related components, but the primary issue is simply the persistence of `data-focused`.

Thanks for your help.

mj12albert:
Fixes https://github.com/mui/base-ui/issues/3987

### DIFF
--- a/packages/react/src/combobox/trigger/ComboboxTrigger.tsx
+++ b/packages/react/src/combobox/trigger/ComboboxTrigger.tsx
@@ -125,12 +125,12 @@ export const ComboboxTrigger = React.forwardRef(function ComboboxTrigger(
     event: 'mousedown',
   });
 
-  const { buttonRef, getButtonProps } = useButton({
+  const { buttonRef, getButtonProps, focusableWhenDisabled } = useButton({
     native: nativeButton,
     disabled,
   });
 
-  useClearFocusWhenDisabled(disabled);
+  useClearFocusWhenDisabled(disabled, focusableWhenDisabled);
 
   const state: ComboboxTrigger.State = {
     ...fieldState,
--- a/packages/react/src/combobox/trigger/ComboboxTrigger.tsx
+++ b/packages/react/src/combobox/trigger/ComboboxTrigger.tsx
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import { useStore } from '@base-ui/utils/store';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
+import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
 import { useTimeout } from '@base-ui/utils/useTimeout';
 import { ownerDocument } from '@base-ui/utils/owner';
 import { BaseUIComponentProps, NativeButtonProps } from '../../utils/types';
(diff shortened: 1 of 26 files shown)


packages/react/src/field/root/FieldRoot.tsx
```diff
@@ -96,7 +96,7 @@ const FieldRootInner = React.forwardRef(function FieldRootInner(
       dirty,
       valid,
       filled,
-      focused,
+      focused: disabled ? false : focused,
```
atomiks, line 99: I guess there'd need to be an effect to deal with the control itself being disabled too

packages/react/src/number-field/input/NumberFieldInput.tsx
```diff
@@ -124,6 +127,12 @@ export const NumberFieldInput = React.forwardRef(function NumberFieldInput(
     validation.commit(value, true);
   });
 
+  useIsoLayoutEffect(() => {
+    if (disabled) {
+      setFocused(false);
+    }
+  }, [disabled, setFocused]);
```
atomiks: Maybe this can go in `useField`? As it'll apply to all components automatically

The hook would need to accept a new `disabled` option though, and I'm not sure the location of the `useField()` call makes this simple to adjust

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
```diff
@@ -129,6 +130,12 @@ export const ComboboxTrigger = React.forwardRef(function ComboboxTrigger(
     disabled,
   });
 
+  useIsoLayoutEffect(() => {
+    if (disabled) {
+      setFocused(false);
+    }
+  }, [disabled, setFocused]);
```
atomiks: Another issue with this is `[data-focused]` + `focusableWhenDisabled` (or any control which uses aria-disabled, or a non-native form control) shouldn't remove the attribute

packages/react/src/number-field/input/NumberFieldInput.tsx
mj12albert: Updated in https://github.com/mui/base-ui/pull/3996/commits/1b7790e1b6051a702fa07161317f1801224ea881

packages/react/src/combobox/trigger/ComboboxTrigger.tsx
mj12albert: Fixed, I think the affected controls here can only be focusableWhenDisabled via ToolbarButton

atomiks: My last concern is I wonder if this is even worth handling, because a control should never be `disabled` while focus is still on it. In that case, they should be using `focusableWhenDisabled` anyway 🤔

mj12albert: It's not a common case for sure but IIRC "focused state stuck after becoming disabled" has came up in material UI before a few times

It is worth considering if we should provide focusableWhenDisabled on all form controls though 🤔

https://github.com/microsoft/fluentui/issues/4387

PR #3997, opened 2026-02-06
[popover] Fix trapping focus when modal is true

sai6855:
Update modal handling logic to allow trapping focus when `modal` is set to `true`

### Before
https://stackblitz.com/edit/gchds8cz?file=src%2FApp.tsx

### After
https://stackblitz.com/edit/sehut9n6?file=src%2FApp.tsx

### DIFF
--- a/packages/react/src/popover/popup/PopoverPopup.tsx
+++ b/packages/react/src/popover/popup/PopoverPopup.tsx
@@ -120,7 +120,7 @@ export const PopoverPopup = React.forwardRef(function PopoverPopup(
     <FloatingFocusManager
       context={floatingContext}
       openInteractionType={openMethod}
-      modal={modal === 'trap-focus'}
+      modal={modal !== false}
       disabled={!mounted || openReason === REASONS.triggerHover}
       initialFocus={resolvedInitialFocus}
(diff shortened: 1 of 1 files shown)


packages/react/src/popover/popup/PopoverPopup.tsx
```diff
@@ -120,7 +120,7 @@ export const PopoverPopup = React.forwardRef(function PopoverPopup(
     <FloatingFocusManager
       context={floatingContext}
       openInteractionType={openMethod}
-      modal={modal === 'trap-focus'}
+      modal={modal !== false}
```
sai6855, line 123: When `modal` is set to true, `Popover` isn't trapping focus and thus modal is getting closed. Fixed logic to `trap focus` when `modal` is `true`

https://github.com/user-attachments/assets/7ad1ec01-bdbb-4d03-a07c-d082f050724f

atomiks: This was intentional since in popovers, the close button is largely optional unlike in typical dialogs (and would otherwise leave users trapped when there's no <kbd>Esc</kbd> key). `"trap-focus"` is available if you want it to trap focus

PR #4005, opened 2026-02-06
[select][combobox] Add `autoComplete` prop for explicit browser autofill support

mattrothenberg:
## Summary

Adds an `autoComplete` prop to `Select.Root` that gets passed to the hidden input element, enabling proper browser autofill functionality.

## Problem

The Select component creates a hidden input with the `name` attribute for form submission. However, without the `autocomplete` attribute, browsers don't recognize this input as a candidate for autofill.

The existing autofill test verifies that the component *handles* autofill events when they occur (via `fireEvent.change`), but without `autocomplete`, browsers don't trigger those events in the first place.

It could be considered a regression from Material UI Select.

## Solution

Add an `autoComplete` prop that passes through to the hidden input:

```tsx
<Select.Root name="country" autoComplete="country">
  <Select.Trigger>
    <Select.Value />
  </Select.Trigger>
  ...
</Select.Root>
```

This renders:
```html
<input name="country" autocomplete="country" ... />
```

## Changes

- Add `autoComplete` prop to `SelectRootProps` interface with JSDoc
- Destructure and pass `autoComplete` to the hidden input element
- Add test to verify the attribute is passed through

## References

- [MDN: HTML autocomplete attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)

### DIFF
--- a/docs/reference/generated/combobox-root.json
+++ b/docs/reference/generated/combobox-root.json
@@ -70,6 +70,11 @@
       "description": "A ref to imperative actions.\n- `unmount`: When specified, the combobox will not be unmounted when closed.\nInstead, the `unmount` function must be called to unmount the combobox manually.\nUseful when the combobox's animation is controlled by an external library.",
       "detailedType": "| React.RefObject<Combobox.Root.Actions | null>\n| undefined"
     },
+    "autoComplete": {
+      "type": "string",
+      "description": "Provides a hint to the browser for autofill.",
+      "detailedType": "string | undefined"
+    },
     "filter": {
       "type": "((itemValue: Value, query: string, itemToString: ((itemValue: Value) => string) | undefined) => boolean) | null",
       "description": "Filter function used to match items vs input query.",
--- a/docs/reference/generated/select-root.json
+++ b/docs/reference/generated/select-root.json
@@ -49,6 +49,11 @@
       "description": "A ref to imperative actions.\n- `unmount`: When specified, the select will not be unmounted when closed.\nInstead, the `unmount` function must be called to unmount the select manually.\nUseful when the select's animation is controlled by an external library.",
       "detailedType": "| React.RefObject<Select.Root.Actions | null>\n| undefined"
     },
+    "autoComplete": {
+      "type": "string",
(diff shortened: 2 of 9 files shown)


atomiks:
> The Select component creates a hidden input with the name attribute for form submission. However, without the autocomplete attribute, browsers don't recognize this input as a candidate for autofill.
> 
> The existing autofill test verifies that the component handles autofill events when they occur (via fireEvent.change), but without autocomplete, browsers don't trigger those events in the first place.

Is this only for Safari/Firefox?

Chrome does look for `name`, as the [autofill demo](https://deploy-preview-3924--base-ui.netlify.app/experiments/forms/autofill) works in Chrome already for Select:

https://github.com/user-attachments/assets/4e2bfadc-b9a6-4fa0-ab1b-569e8b5d8b2d

mattrothenberg:
@atomiks Strange, that doesn't work for me in Chrome 144.0.7559.110. Happy to close this out if you think it's unnecessary.

atomiks:
Chrome actually gives a warning when `autocomplete` is missing when autofill recognizes the id/name of the field:

> A form field has an id or name attribute that the browser's autofill recognizes. However, it doesn't have an autocomplete attribute assigned. This might prevent the browser from correctly autofilling the form.

So this does still look correct to add. 

@mattrothenberg this prop should also be added for Combobox as well — `AriaCombobox` has an `autoComplete` prop but only for `Autocomplete`'s public API which calls it `mode`. `Combobox` (public) omits it. Both could still have a (public) prop called `autoComplete` which forwards to the hidden input

mattrothenberg:
@atomiks excellent, will get this fixed ASAP

mattrothenberg:
@atomiks Updated. I named the internal prop `formAutoComplete` to distinguish it from the existing ARIA `autoComplete` prop in AriaCombobox. Feel free to tweak the last two commits if I did it wrong!

atomiks:
Thanks for the contribution!

Separate PR that adds the manual autofill demo — will amend it further: https://github.com/mui/base-ui/pull/3924

PR #4015, opened 2026-02-08
[combobox] Distinguish input-press from trigger-press in onOpenChange reason 

### LINKED ISSUE #4000: [combobox] Separate `trigger-press` and `input-press` events
# Feature request

## Summary

Currently `trigger-press` fires when either the `Combobox.Input` **or** `Combobox.Trigger` is pressed — when I expected it to only fire for the `Combobox.Trigger`

N/A

## Motivation

We're currently tracking the `Input` and `Trigger` press to determine where to animate the `Popup`'s origin, but we're unable to do this via `onOpenChange` as there aren't different events to track for. In the meantime, we're tracking the clicks ourselves manually, but others might run into the same hurdle! 

cc @moeamaya

jijiseong:
closes #4000 

## Summary

`onOpenChange` now receives `input-press` when the combobox popup is opened by the Input.

## Changes

- Add `input-press` to Reasons.
- Add optional `reason` prop to `useClick` (default `triggerPress`).
- Pass `input-press`  into `useClick`’s `reason` prop in AriaCombobox.

## Question

Is it okay to keep this in useClick (configurable `reason`)?

### DIFF
--- a/docs/reference/generated/autocomplete-root.json
+++ b/docs/reference/generated/autocomplete-root.json
@@ -134,7 +134,7 @@
     },
     "openOnInputClick": {
       "type": "boolean",
-      "default": "true",
+      "default": "false",
       "description": "Whether the popup opens when clicking the input.",
       "detailedType": "boolean | undefined"
     },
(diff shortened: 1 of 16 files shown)


packages/react/src/combobox/input/ComboboxInput.test.tsx
```diff
@@ -186,6 +188,38 @@ describe('<Combobox.Input />', () => {
     });
   });
 
+  describe('onOpenChange reason', () => {
+    it('fires with reason input-press when Input is clicked', async () => {
```
atomiks, line 192: Thanks for the PR! Seems like a test for ComboboxTrigger (for `trigger-press`) would also be a good idea

jijiseong: I added test. Thank you

atomiks (approved):
Thanks for the fix @jijiseong!

PR #4017, opened 2026-02-08
[popover][combobox] Use reactive `domReferenceElement` subscriptions

atomiks:

Ensures these variables can't be stale during a render.

### DIFF
--- a/packages/react/src/combobox/trigger/ComboboxTrigger.tsx
+++ b/packages/react/src/combobox/trigger/ComboboxTrigger.tsx
@@ -93,7 +93,7 @@ export const ComboboxTrigger = React.forwardRef(function ComboboxTrigger(
     currentPointerTypeRef.current = event.pointerType;
   }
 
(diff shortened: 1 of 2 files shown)


michaldudak:
Does this solve an existing bug? I generally managed to avoid having triggers reactive to the `referenceElement` as this causes re-rendering of all triggers when the active one changes. While this is not an issue with Combobox today, as it doesn't support detached triggers, it might be better to avoid it (in case we want to introduce them in the future).

atomiks:
For Combobox, it does look like a bug if the reference changes

> Does this solve an existing bug? I generally managed to avoid having triggers reactive to the referenceElement as this causes re-rendering of all triggers when the active one changes.

For Popover, it seems like the `select()` calls belongs in the effect instead of render in this case?

michaldudak:
For Popover, this doesn't introduce any additional renders as it's already reactive to `activeTriggerElement`.
I think we can merge it, but it'll have to be refactored if we ever introduce detached triggers to Combobox.

PR #4033, opened 2026-02-10
[radio] Add generic `Value` typing to `Radio`

atomiks:

Same type updates as #3173


packages/react/src/radio-group/RadioGroup.tsx
```diff
@@ -27,8 +27,8 @@ const MODIFIER_KEYS = [SHIFT];
  *
  * Documentation: [Base UI Radio Group](https://base-ui.com/react/components/radio)
  */
-export const RadioGroup = React.forwardRef(function RadioGroup(
-  componentProps: RadioGroup.Props,
+export const RadioGroup = React.forwardRef(function RadioGroup<Value>(
```
mnajdova, line 30: Should the value extend string? as in the Toggle component?

atomiks: Radio was originally unknown (then any) so it wasn't limited to strings. It supports numbers or objects already, so anything else is breaking

mnajdova: Fair enough 👌

mnajdova (approved):

PR #4037, opened 2026-02-10
[internal][combobox] Remove unnecessary TS Omit and rename prop

ZeeshanTamboli:
Remove unnecessary TS Omit. 

`ComboboxRootProps` does not contain these props. These props are of AriaCombobox.

### DIFF
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -63,17 +63,17 @@ import { INITIAL_LAST_HIGHLIGHT, NO_ACTIVE_VALUE } from './utils/constants';
  * @internal
  */
 export function AriaCombobox<Value, Mode extends SelectionMode = 'none'>(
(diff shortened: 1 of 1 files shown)


packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1499,10 +1499,10 @@ interface ComboboxRootProps<ItemValue> {
...
 }
 
-export type ComboboxRootConditionalProps<Value, Mode extends SelectionMode = 'none'> = Omit<
-  ComboboxRootProps<Value>,
-  'selectionMode' | 'selectedValue' | 'defaultSelectedValue' | 'onSelectedValueChange'
-> & {
+export type ComboboxRootConditionalProps<
+  Value,
+  Mode extends SelectionMode = 'none',
+> = ComboboxRootProps<Value> & {
```
atomiks: `ConditionalProps` is an older name for these, it can just be `AriaComboboxProps` (or such)

atomiks (approved):

PR #4040, opened 2026-02-11
[toast] Prevent dismissed promise toast from reopening on updates

atomiks:

Most noticeable with promise toasts: if you dismiss and while exiting it updates, it comes back into (half) view, depending on the swipe amount


packages/react/src/toast/createToastManager.test.tsx
```diff
@@ -204,6 +204,53 @@ describe.skipIf(!isJSDOM)('createToastManager', () => {
...
+        });
+
+        toastManager.promise(pendingPromise, {
+          loading: 'loading',
+          success: 'success',
+          error: 'error',
+        });
+      }
+
+      function Buttons() {
```
mnajdova:
```suggestion
      function Button() {
```

nit: There is only one button :D

Issue #4009, opened 2026-02-07
[docs] Talking about accessibility, why do you use the blue color for non-links?


binarykitchen:
## How can we improve Base UI documentation?

Great documentation and language. Since you promote accessibility, and it's a top-priority, I wonder why you style some words in blue which isn't clickable?

<img width="810" height="313" alt="Image" src="https://github.com/user-attachments/assets/9d7881ba-55d0-4774-8ca1-4e65797e9190" />

https://base-ui.com/react/utils/merge-props

I don't use a screen reader nor am disabled, although whenever I see a word in blue it makes me believe it's a link, clickable.
Because browser defaults display links in blue.

I've clicked on "**mergeProps** helps you combine" ... multiple times and wondered why it didn't go to the next page? 😅 
My recommendation: when it's not a link, don't use blue but display it differently. Perhaps in bold, or really use links and underline them.

Hope this helps.

oliviertassinari:
I think we can also wonder why the font size is different for code sections (this big).

I find https://mui.com/material-ui/react-alert/#severity clearer:

<img width="836" height="99" alt="Image" src="https://github.com/user-attachments/assets/086e9af0-b807-421b-9376-b12f042f8fc7" />

https://www.radix-ui.com/primitives/docs/components/slider#root also feels clearly better.

<img width="763" height="107" alt="Image" src="https://github.com/user-attachments/assets/427aacc6-2099-4468-aac5-1f3f39c2c27b" />

atomiks:
A padded colored background at the least would help make it look like inline code, rather than a link, even if it stays blue (like the Radix example)

Issue #2766, opened 2025-09-15
[docs](docs-infra] Early feedback v2


oliviertassinari:
## Summary

This is a refresh of #588 with the up-to-date problems/opportunities with the documentation as of v1.0.0-beta.3. The goal is to only cover more table stack expectations that users can have, looking at what could be regressions from Material UI docs, it's not to go above and beyond.

## Motivation

Try to provide great DX.

## List

In no specific order, some should probably be moved to their own issue:

- [ ] 1. Headers text are harder to select than in Material UI

https://github.com/user-attachments/assets/1cab1313-1672-4acb-bf4d-2e5a1370f31a

- [ ] 2. There are no "reset focus" feature. It seems harder to test without. Shouldn't there be one (under a triple dot menu)? 

<img width="349" height="104" alt="Image" src="https://github.com/user-attachments/assets/6a6ed6e5-b76f-4abd-905e-84710005c206" />

- [ ] 3. There are no links to the source. It feels impactful to help the community contribute fixes. Shouldn't there be one (under a triple dot menu)?

<img width="323" height="115" alt="Image" src="https://github.com/user-attachments/assets/5225f61d-6d3a-492c-aed1-c84840e18b80" />

- [ ] 4. There is no light/dark mode toggle. It's not obvious that everyone uses the same mode for their OS and code. For example, I'm personally mostly in light mode. I moved to dark mode for some code-related tasks.

<img width="301" height="91" alt="Image" src="https://github.com/user-attachments/assets/7971a5a7-e8a2-4dfe-b1cb-24febdb438e7" />

- [ ] 5. Demos are not editable. Waiting 5-10 seconds to open a demo in CodeSandbox or StackBlitz is not the best experience. Better would be to be able to test prop combinations instantly.

- [ ] 6. Element that handles `onClick` don't have `cursor: pointer`. Current behavior seems correct for people who are building Electron apps, where on desktop, the expectation is that the UI is so normalized, everything looks the same, that, as a user, you can predict what's clickable, and hence, you don't need to be distracted with cursor changes. But for anyone else building for the web, every website is different; it feels a lot more frustrating to not know if an element can be interacted with or not than to have a cursor change distraction.

  To be clear, this point is not me demanding that we make this change; it's me being confused: I don't understand the logic, why we are not giving end-users what feels better UX (on my end), or what they are used to https://github.com/shadcn-ui/ui/issues/6843. 

[...]

PR #4043, opened 2026-02-11
[docs] Style inline code parameters with background to distinguish from links

CiscoFran10:

### Summary

Inline terms like `mergeProps` were styled in blue without a background, so they looked like links. This change gives inline code that represents parameters a padded, colored background so it reads as code instead of a link, improving clarity and accessibility.

### Changes

- Added `--syntax-parameter-background` and a rule for `[data-inline][data-syntax='parameter']` so inline parameter code gets a background, border-radius, and padding.
- Inline code nodes that contain a parameter token are marked with `data-syntax="parameter"` so the new styles apply.

Captura de tela 2026-02-11 030711
Captura de tela 2026-02-11 030639

Closes: #4009

### DIFF
--- a/docs/src/syntax-highlighting/index.css
+++ b/docs/src/syntax-highlighting/index.css
@@ -4,6 +4,7 @@
   --syntax-constant: var(--color-blue);
   --syntax-entity: var(--color-violet);
   --syntax-parameter: var(--color-navy);
+  --syntax-parameter-background: var(--color-inline-highlight);
   --syntax-tag: var(--color-green);
   --syntax-keyword: var(--color-red);
   --syntax-string: var(--color-navy);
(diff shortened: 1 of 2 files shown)


colmtuite:
Hey @CiscoFran10, thanks for the help with this PR, but we're already working on a redesign + rebuild of the docs, which will include this improvement. So I'm gonna close this as we've got it covered, but thanks for the effort, I appreciate it.

CiscoFran10:
@colmtuite No problem at all, looking forward to the new docs 😁 .

Just to confirm, will all the points mentioned in the early feedback issue #2766 be addressed as part of the rebuild, or are there still some areas where contributions would be helpful? I was planning to continue fixing some of the items I could, such as the theme toggle.

Also, I currently have these two open PRs that were mentioned in #2766:

* #4052
* #4076

Should I expect those to be closed as well due to the redesign, or are they still relevant?

PR #4047, opened 2026-02-11
[internal] Remove duplicate getDocument util

mj12albert:


michaldudak:
Nice catch. It would be even better if we didn't export `getDocument` at all and changed all call sites to use `ownerDocument` for consistency.

michaldudak (approved):

PR #4049, opened 2026-02-11
[tooltip] Fix `disabled` prop on Triggers

### LINKED ISSUE #3468: [tooltip] Allow disabling a `Tooltip.Trigger` when using detached triggers
# Feature request

## Summary

When a `Tooltip.Trigger` is the only trigger inside a `Tooltip.Root`, we can toggle the `disabled` prop on the root to conditionally enable/disable the tooltip. However, that's not possible when a `Tooltip.Root` contains multiple triggers (all sharing the same handle) and we want to conditionally disable a single one. I thought I had found a workaround: conditionally change the handle of the `Tooltip.Trigger` that I want to disable (to e.g. one that does not even have a rendered `Tooltip.Root`, e.g. `disabledHandle`), but that doesn't seem to work reliably. :/

Not sure.

## Motivation

I am building a component library with Base UI and some components, like `Button`, have an inner `Tooltip` wrapping them so that they can accept a simple `tooltip` prop. However, not all buttons have tooltips, so I want to be able to disable the `Tooltip` when the prop is not passed (or is empty). That works at the moment, but I want to switch to detached triggers so I can animate tooltip changes, but my `Button`'s `Tooltip.Trigger` doesn't have access to the "global" `Tooltip.Root` it's associated to, so it can't disable it (and even if it could, it shouldn't disable it, because other triggers associated to that root still need to work – really, what I want to do is disable the trigger).

michaldudak:
This PR fixes tooltip behavior for detached triggers when moving between enabled/disabled triggers, and unifies the logic across hover and focus interactions.

### What changed

- Introduced a shared enabled-trigger check (`isTargetInsideEnabledTrigger`) used by hover and focus logic.
- Updated hover handling so switching to a trigger marked with `data-trigger-disabled` no longer keeps the previous popup open.
- Updated focus handling with the same rule, so focusing a disabled trigger closes an already-open tooltip.
- Kept support for rendered native disabled buttons as valid triggers when the trigger itself is enabled.

### Tests

Added tooltip detached-trigger regressions for:

- hover to disabled trigger closes,
- focus to disabled trigger closes,
- rendered disabled button still switches/opens when trigger hover is enabled.

### Result

Detached trigger behavior is now consistent and predictable:

- Disabled triggers do not participate in hover/focus switching.
- Enabled triggers rendered as disabled buttons still behave correctly.

Closes #3468

### Preview

https://deploy-preview-4049--base-ui.netlify.app/experiments/tooltip/disabled

### DIFF
--- a/packages/react/src/floating-ui-react/utils/element.ts
+++ b/packages/react/src/floating-ui-react/utils/element.ts
@@ -91,10 +91,6 @@ export function isRootElement(element: Element): boolean {
   return element.matches('html,body');
 }
 
-export function getDocument(node: Element | null) {
-  return node?.ownerDocument || document;
-}
-
 export function isTypeableElement(element: unknown): boolean {
   return isHTMLElement(element) && element.matches(TYPEABLE_SELECTOR);
 }
--- a/docs/reference/generated/avatar-fallback.json
+++ b/docs/reference/generated/avatar-fallback.json
@@ -8,18 +8,18 @@
       "detailedType": "number | undefined"
     },
     "className": {
-      "type": "string | ((state: Avatar.Root.State) => string | undefined)",
+      "type": "string | ((state: Avatar.Fallback.State) => string | undefined)",
       "description": "CSS class applied to the element, or a function that\nreturns a class based on the component’s state.",
-      "detailedType": "| string\n| ((state: Avatar.Root.State) => string | undefined)"
+      "detailedType": "| string\n| ((state: Avatar.Fallback.State) => string | undefined)"
     },
     "style": {
-      "type": "CSSProperties | ((state: Avatar.Root.State) => CSSProperties | undefined)",
-      "detailedType": "| React.CSSProperties\n| ((state: Avatar.Root.State) => CSSProperties | undefined)\n| undefined"
+      "type": "CSSProperties | ((state: Avatar.Fallback.State) => CSSProperties | undefined)",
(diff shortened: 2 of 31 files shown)


docs/reference/generated/tooltip-trigger.json
```diff
@@ -12,6 +12,12 @@
       "description": "A payload to pass to the tooltip when it is opened.",
       "detailedType": "Payload | undefined"
     },
+    "disabled": {
+      "type": "boolean",
+      "default": "false",
+      "description": "Whether the trigger is disabled. If `true`, the tooltip will not open when interacting with this trigger.\nNote that this doesn't apply the `disabled` attribute to the trigger element.\nIf you want to disable the trigger element itself, you can pass the `disabled` prop to the trigger element via the `render` prop.",
```
atomiks: "Whether the trigger is disabled" should probably be updated

packages/react/src/floating-ui-react/utils/element.ts
```diff
@@ -39,6 +40,29 @@ export function contains(parent?: Element | null, child?: Element | null) {
...
+
+  if (triggerElements.hasElement(targetElement)) {
+    return !targetElement.hasAttribute('data-trigger-disabled');
+  }
+
+  for (const [, trigger] of triggerElements.entries()) {
+    if (contains(trigger, targetElement)) {
+      return !trigger.hasAttribute('data-trigger-disabled');
+    }
+  }
```
atomiks, line 61: `triggerElements.values()`?

atomiks (approved):

PR #4052, opened 2026-02-12
[docs] Add InstallationBlock with package manager tabs to quick start

CiscoFran10:

## Summary

Adds an `InstallationBlock` component to the docs that shows install commands for multiple package managers (npm, yarn, pnpm, bun) in a tabbed UI. The selected package manager is synced with the existing `PackageManagerSnippetProvider` and persisted in `localStorage`.

image

## Changes

- **New:** `InstallationBlock` component (`InstallationBlock.tsx` + `InstallationBlock.css`) with:
  - Tabs for npm, yarn, pnpm, and bun
  - Correct install command per manager (e.g. `npm i`, `yarn add`, `pnpm add`, `bun add`)
  - Copy button and “Terminal” title for consistency with other code blocks
- **Updated:** Quick start page uses `<InstallationBlock package="@base-ui/react" />` instead of a single install snippet
- **Updated:** `docs/src/styles.css` for the new block

## Motivation

Users can see and copy the install command for their preferred package manager directly in the installation docs, and the choice is remembered across the docs via the existing package manager preference.

> 7. Having a terminal toggle would be nice for lazy people. I'm using pnpm, I would rather have a direct copy & paste that is saved in a cookie.

See for context: #2766

### DIFF
--- a/docs/src/components/InstallationBlock.tsx
+++ b/docs/src/components/InstallationBlock.tsx
@@ -9,6 +9,7 @@ import { usePackageManagerSnippetContext } from 'docs/src/blocks/PackageManagerS
 import { GhostButton } from './GhostButton';
 import { CopyIcon } from '../icons/CopyIcon';
 import { CheckIcon } from '../icons/CheckIcon';
+import './InstallationBlock.css';
 
 const PACKAGE_MANAGERS = [
   { value: 'pnpm', label: 'pnpm', command: 'add' },
--- a/docs/src/css/index.css
+++ b/docs/src/css/index.css
@@ -52,7 +52,6 @@
 @import '../components/Table.css';
 @import '../components/TableCode.css';
 @import '../components/DescriptionList.css';
-@import '../components/InstallationBlock.css';
 @import '../components/ReferenceTable/ReferenceTable.css';
 
 @import '../syntax-highlighting/index.css';


colmtuite:
cc @dav-is

colmtuite:
@CiscoFran10 Delete the "Terminal" title, that's very confusing, and not how these UIs typically work. There should only be tabs, nothing else.

@dav-is Are you happy with how this is implemented?

dav-is:
I think the way this is implemented is ok, after the typedoc generation PR is merged, we will be able to migrate code blocks to the new code highlighting infra used in the demos, which has a different format for how this would be written. We should be able to reuse the design from this and should still be simple to migrate, so we shouldn't need to block this PR

docs/scripts/generateLlmTxt/mdxToMarkdown.mjs
```diff
@@ -185,6 +185,46 @@ function transformJsx() {
...
+            const pkg =
+              node.attributes?.find((attr) => attr.name === 'package')?.value || '@base-ui/react';
+
+            const nodes = [
+              mdx.paragraph(
+                'Use the package manager detected from the project lock file. If no lock file exists, use pnpm.',
+              ),
+
+              {
+                type: 'list',
```
dav-is: Do we really need to be this verbose? I would think LLMs know how to select the correct package manager when given each as an option

CiscoFran10:
> @CiscoFran10 Delete the "Terminal" title, that's very confusing, and not how these UIs typically work. There should only be tabs, nothing else.

removed.

image

docs/scripts/generateLlmTxt/mdxToMarkdown.mjs
CiscoFran10: Simplified the LLM text.

Now its only:

> ## Install the library
> 
> Install Base UI using a package manager.
> 
> ### pnpm
> 
> ```bash
> pnpm add @base-ui/react
> ```
> 
> ### npm
> 
> ```bash
> npm install @base-ui/react
> ```
> 
> ### yarn
> 
> ```bash
> yarn add @base-ui/react
> ```
> 
> ### bun
> 
> ```bash
> bun add @base-ui/react
> ```

dav-is:
@CiscoFran10 It looks like the corners of the codeblock are being cut off

Also, what do you think of the order in the UI matching the order you have in the markdown version: pnpm, npm, yarn, bun

CiscoFran10:
> @CiscoFran10 It looks like the corners of the codeblock are being cut off

corners fixed. good catch 😁. 

> Also, what do you think of the order in the UI matching the order you have in the markdown version: pnpm, npm, yarn, bun

I updated the order, I do prefer pnpm being first since its the one I use, but I figured that it should prob be a maintainer decision.

image

docs/src/components/InstallationBlock.css
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /* Increase the clickable size */
+    &::after {
+      inset: -0.375rem -0.5rem;
+    }
+
+    &[data-active] {
+      color: var(--color-foreground);
+      font-weight: bold;
```
aarongarciah: `font-weight: 500` seems enough, given the active state pill is prominent enough.

|700|500|
|---|---|
|Image|Image|

aarongarciah: Forget about this. I see we're using `700` in the demo code block tabs. We'll tackle this in the redesign.

aarongarciah:
@CiscoFran10 can we avoid this flash upon hydration?

https://github.com/user-attachments/assets/d661fbdd-9dc4-4aa2-a905-97ecc6e49ba0

dav-is:
@aarongarciah I don't think the flash can be avoided, it's the same thing we have for the css modules vs tailwind flash

atomiks:
The docs font is changing, but `500` weight looks better than `700` weight on the selected tab @colmtuite?

colmtuite:
@atomiks We only have 400 and 700 now. All component styles will change soon, so it won't be an issue.

aarongarciah:
Thanks @CiscoFran10! I've pushed a couple of commits:

- Prevent tabs from jumping a bit when changing the active tab.
- Load the component CSS in the .tsx file instead of the root stylesheet.

Issue #1746, opened 2025-04-18
[menu] Should link menu items navigate to their `href` when pressing Space?


benface:
# Feature request

## Summary

[As documented](https://base-ui.com/react/components/menu#navigate-to-another-page), `Menu.Item`s can render a `<a href>` to navigate to a URL when clicked, instead of executing an action on the page. Such a menu item is highlightable in exactly the same ways that a regular menu item is, but it is not _activatable_  in the same ways: pressing the Space key doesn't navigate to the URL. It does, however, close the menu (assuming `closeOnClick` is not changed from its default value of `true`), which seems wrong. I feel like if nothing happens, it should not close the menu, but ideally, IMO it _should_ navigate to the URL, just like when pressing Enter.

[CodeSandbox](https://codesandbox.io/p/sandbox/naughty-mclean-p9zwrl)

## Examples in other libraries

Not sure.

## Motivation

Consistency in keyboard behavior for end users who may not understand the difference between menu items that are links and those that are not.

colmtuite:
Here is a relevant thread https://github.com/w3c/aria/issues/1701

Making `<a>` activate with Space is problematic, because Space scrolls the viewport. Users might expect Space to scroll the viewport, even when an `<a>` is focused. However, we could say this behaviour should be different inside widgets, since users generally don't expect Space to scroll the viewport when focus is inside a widget.

The main blocker is that `:active` doesn't fire when pressing Space. So if we do this, we'll need to commit to handling custom pseudo classes via data attrs. We haven't made a decision there yet.

benface:
Thanks @colmtuite. From the thread you linked to:

> If you design a link like a button (which is not a good idea) you can already make the link activatable with SPACE. **This would not be a violation of any specification** - but also not required according to WCAG 2.1.1, since WCAG does not require a specific keyboard operation, but only any.

(Emphasis mine.) In the Menu's case, a link is styled like a menu item, so I would expect it to work like other menu items. But I hear your concerns and shall await your decision!

mj12albert:
Without accounting for the `:active` state, it can be done with a ref and  `event.preventBaseUIHandler()`: https://codesandbox.io/p/sandbox/blissful-bouman-r2lvxg

```jsx
const MenuItemLink = React.forwardRef(function MenuItemLink(
  props: Menu.Item.Props & { href: string },
  forwardedRef: React.ForwardedRef<HTMLAnchorElement>
) {
  const { href, ...otherProps } = props;

[...]

PR #4053, opened 2026-02-12
[all components] Fire Space activation on `keydown` for composite widgets

### LINKED ISSUE #4051: [composite] Fire Space key activation on `keydown` not `keyup`
# Bug report

## Current behavior

I’ve noticed an inconsistency in keyboard behavior between the **Space** and **Enter/Return** keys when interacting with focusable components (e.g., `ToggleGroup` or `Toolbar`).

When a component is focused:

- Pressing **Enter (Return)** to activate an item, then pressing the **Right Arrow** key to move to the next item, behaves as expected, the activated state is preserved.
- However, when pressing **Space** to activate the same item and then pressing the **Right Arrow** key, the activation does not appear to be applied. The behavior is as if the Space key activation was ignored.

## Expected behavior

Both **Space** and **Enter** should activate the focused item consistently, and the activated state should persist when navigating with arrow keys.

## Reproducible example

Any one of these components, if you focus and then press space and then arrow keys, the space does not get applied

https://base-ui.com/react/components/accordion
https://base-ui.com/react/components/toggle-group
https://base-ui.com/react/components/toolbar
https://base-ui.com/react/components/navigation-menu
https://base-ui.com/react/components/menubar
https://base-ui.com/react/components/menu

## Base UI version

"@base-ui/react": "1.1.0"

Helium https://helium.computer
Chrome
Safari

## Which OS are you using?

Mac OS

atomiks:

Fixes #4051 - matches macOS behavior for composite widgets
Closes https://github.com/mui/base-ui/issues/1746

### DIFF
--- a/packages/react/src/use-button/useButton.ts
+++ b/packages/react/src/use-button/useButton.ts
@@ -135,11 +135,7 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
             const isSpaceKey = event.key === ' ';
             const role = currentTarget.getAttribute('role');
             const isTextNavigationRole =
-              role === 'menuitem' ||
-              role === 'menuitemcheckbox' ||
-              role === 'menuitemradio' ||
-              role === 'option' ||
-              role === 'gridcell';
+              role?.startsWith('menuitem') || role === 'option' || role === 'gridcell';
 
             if (isCurrentTarget && isCompositeItem && isSpaceKey) {
               if (event.defaultPrevented && isTextNavigationRole) {


atomiks:
## Codex Review

### Overview
This branch moves composite-widget Space activation from `keyup` to `keydown` across `useButton` consumers (menu/select/combobox/link paths), and aligns typeahead behavior so active text navigation sessions can include spaces without accidental activation. The latest updates also centralize the menu typeahead Space guard and fix the toolbar switch regression by scoping the `defaultPrevented` suppression to text-navigation roles.

### Findings (None)

### Confidence: 4/5
High confidence based on a full-branch review (`upstream/master...head`), targeted behavior checks of the updated keyboard paths, and focused verification of the prior CI failure mode.

### Notes
- Full branch scope reviewed, not only incremental commits.
- The prior CI failure in `ToolbarButton` was addressed by narrowing composite Space suppression to text-navigation roles (`menuitem*`, `option`, `gridcell`) while preserving `switch` activation semantics.
- Perf spot-check executed on representative old/new keydown decision paths (synthetic Node micro-check, 5,000,000 iterations): `old=16.94ms`, `new=47.57ms` (about `6.13ns` additional work per dispatch in the check).

PR #4054, opened 2026-02-12
[select] Improve `onValueChange` type inference

LukasTy:
Experiment to try addressing: https://github.com/mui/base-ui/issues/2891#issuecomment-3830930088.

The only edge-case is that `value` typed with `| null` ends up dropping the `| null` in the `onValueChange`.
But isn't it expected for Select, where you always select a value?
Or we keep this option open for cases where the items list would have a `null` entry?

### DIFF
--- a/docs/reference/generated/select-root.json
+++ b/docs/reference/generated/select-root.json
@@ -18,9 +18,9 @@
       "detailedType": "Value[] | Value | null | undefined"
     },
     "onValueChange": {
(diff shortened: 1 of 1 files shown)


atomiks:
Yeah, you can have `<Select.Item value={null} />` rendered in the list.

LukasTy:
> Yeah, you can have `<Select.Item value={null} />` rendered in the list.

So, basically, due to our API approach, we can not ensure fully compliant type inference?

PR #4060, opened 2026-02-12
[menu] Implement content transitions with Viewport

michaldudak:
Added the Viewport part to the Menu, allowing for animated content transitions when using detached triggers.

https://deploy-preview-4060--base-ui.netlify.app/react/components/menu#animating-the-menu
https://deploy-preview-4060--base-ui.netlify.app/experiments/menu/triggers


atomiks:
## Codex Review

### Overview
This PR adds `Menu.Viewport` so one detached-trigger menu can animate its content, size, and position while reusing the same popup. It also hooks Menu into the shared popup viewport machinery and adds the positioner/state plumbing needed for those transitions.

### Findings (None)

### Confidence: 4/5
High confidence based on a full pass over the current branch diff, a focused sweep of the new viewport/control-flow paths, targeted JSDOM and Chromium verification of the changed Menu tests, and a small Chromium DOM benchmark for the new transition work. Residual risk is limited to highly app-specific styling or accessibility edge cases around cloned transition content.

### Notes
- I am not re-raising the earlier duplicate-ID concern as a blocker. The current branch forces the live subtree to remount when the active trigger changes, and again if the payload arrives a render later, via `usePopupContentKey` in `packages/react/src/utils/usePopupViewport.tsx:88-90` and `packages/react/src/utils/usePopupViewport.tsx:339-367`. That means the active subtree gets fresh generated IDs instead of reusing the exiting subtree's IDs. I also tried a focused Chromium repro with grouped menu labels and did not reproduce duplicate current/previous ID collisions during the transition window.
- The size tradeoff is still real but bounded. The bundle checker on this PR reports `+1.09KB parsed / +441B gzip` versus `master`, which mostly looks like the cost of exposing `Menu.Viewport` and wiring Menu into the shared viewport path (`packages/react/src/menu/viewport/MenuViewport.tsx`, `packages/react/src/menu/positioner/MenuPositioner.tsx`, `packages/react/src/menu/store/MenuStore.ts`, and `packages/react/src/utils/usePopupViewport.tsx`).
- Focused Chromium DOM micro-check for representative detached-trigger switches (`10` samples, `300` switches per sample):

| `master` baseline | 3.18 | 3.10 | 3.80 | 10 |
| PR head | 4.67 | 4.50 | 4.90 | 10 |
| PR head + cloned-subtree sanitization alternative | 6.67 | 6.40 | 7.00 | 10 |

  On this synthetic switch path, the current implementation adds about `1.49ms / 300` switches relative to `master` (roughly `5µs` per switch), while the reverted cloned-subtree sanitization approach would add roughly another `2.00ms / 300` switches on top of the current path. That lines up with the code tradeoff here: the viewport feature is not free, but the remaining transient clone window is short-lived and the heavier sanitization fix appears more expensive in both bytes and per-switch DOM work.
- Verified with `pnpm test:jsdom MenuViewport --no-watch`, `pnpm test:chromium MenuViewport --no-watch`, and `pnpm test:chromium MenuRoot.detached-triggers --no-watch`; all passed.

michaldudak:
This might be slightly problematic if the contents are styled based on their IDs, so there is a tradeoff.
Traversing all the cloned nodes and modifying their attributes would impact performance, so I'm ignoring it for now.

PR #4082, opened 2026-02-13
[popups] Snap `--anchor-width/height` to device pixel grid

### LINKED ISSUE #4061: [combobox] Input and popup elements width mismatch when inheriting width with decimal places
# Bug report

## Current behavior

If the combobox field's width is dictated by the parent (e.g. a grid layout) and ends up with odd width values containing decimal places, the popup seems to have problems rendering its width the same way as the anchor field.

https://github.com/user-attachments/assets/93455e55-3b69-4734-b912-ec2d02735970

## Expected behavior

The width of anchor field and popup should be rendered the exact same as intended - even in the case of values with decimal places caused by e.g. percentage based width rules.

## Reproducible example

[StackBlitz Example](https://stackblitz.com/edit/vtznv3tr-beem4wal?file=src%2FApp.tsx)

## Base UI version

v1.1.0

Behavior identical in Chrome and Firefox

## Which OS are you using?

Mac OS

flaviendelangle:
Closes #4061

The issue was the small black rectangle that flickers at the top right of the floating container.

[Before](https://stackblitz.com/edit/vtznv3tr-beem4wal?file=src%2FApp.tsx)

https://github.com/user-attachments/assets/882ebb77-349e-4ba9-ba23-df32ae505713

[After](https://stackblitz.com/edit/vtznv3tr-gwgarndb?file=package.json)

https://github.com/user-attachments/assets/d254aea3-1b9c-4b31-ba6a-a0b2651993c6

### DIFF
--- a/docs/reference/generated/drawer-indent-background.json
+++ b/docs/reference/generated/drawer-indent-background.json
@@ -1,6 +1,6 @@
 {
   "name": "DrawerIndentBackground",
-  "description": "An element placed before <Drawer.Indent> to render a background layer\nthat can be styled based on whether any drawer is open.",
+  "description": "An element placed before `<Drawer.Indent>` to render a background layer that can be styled based on whether any drawer is open.\nRenders a `<div>` element.",
   "props": {
     "className": {
       "type": "string | ((state: Drawer.IndentBackground.State) => string | undefined)",
--- a/docs/reference/generated/drawer-indent.json
+++ b/docs/reference/generated/drawer-indent.json
@@ -1,6 +1,6 @@
 {
   "name": "DrawerIndent",
-  "description": "A wrapper element intended to contain your app's main UI.\nApplies `data-active` when any drawer within the nearest <Drawer.Provider> is open.",
+  "description": "A wrapper element intended to contain your app's main UI.\nApplies `data-active` when any drawer within the nearest `<Drawer.Provider>` is open.\nRenders a `<div>` element.",
   "props": {
     "className": {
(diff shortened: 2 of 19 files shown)


packages/react/src/utils/useAnchorPositioning.ts
```diff
@@ -308,12 +308,25 @@ export function useAnchorPositioning(
...
+        // Snap anchor dimensions to device pixels to ensure the popup's visual width matches the anchor's one.
+        const dpr = window.devicePixelRatio || 1;
+        const anchorRect = referenceElement.getBoundingClientRect();
+        const anchorWidth =
+          (Math.round(anchorRect.right * dpr) - Math.round(anchorRect.left * dpr)) / dpr;
+        const anchorHeight =
+          (Math.round(anchorRect.bottom * dpr) - Math.round(anchorRect.top * dpr)) / dpr;
+
+        floatingStyle.setProperty('--anchor-width', `${anchorWidth}px`);
+        floatingStyle.setProperty('--anchor-height', `${anchorHeight}px`);
```
atomiks, line 323: 👍 Same technique as the x/y coords themselves:

https://github.com/floating-ui/floating-ui/blob/2c3bf2a45fa06dc562853bcbcd03ffb36289c366/packages/react-dom/src/useFloating.ts#L164-L191

You shouldn't need this though:

> const anchorRect = referenceElement.getBoundingClientRect();

`rects.reference` contains all the info needed already

flaviendelangle: Done :+1:

atomiks (approved):

PR #4089, opened 2026-02-13
[all components] Only define `preventBaseUIHandler` on event handlers typing that actually contain it

### LINKED ISSUE #4088: event.preventBaseUIHandler is not a function
# Bug report

## Current behavior

Calling `event.preventBaseUIHandler()` inside some event handlers yields an error although the types say `preventBaseUIHandler` exists:

```
event.preventBaseUIHandler is not a function
```

Some examples of event handlers that have wrong types:

- `MenuItem`: `onMouseDown`
- `MenuCheckboxItem`: `onMouseDown`
- `MenuRadioItem`: `onMouseDown`
- `MenuLinkItem`: `onMouseDown`, `onKeyDown`, `onKeyUp`
- `AccordionTrigger`: `onMouseDown`, `onMouseUp`

This is probably a general type problem affecting event handlers in many components.

## Expected behavior

Either update the types to remove `preventBaseUIHandler` from the relevant types or include `preventBaseUIHandler` in such events, but the latter probably doesn't make sense if the event handler in question is not doing anything on the Base UI side.

## Reproducible example

- Menu: https://stackblitz.com/edit/xy2mzl1t?file=src/App.tsx
- Accordion: https://stackblitz.com/edit/gkvdhatu?file=src%2FApp.tsx

## Base UI version

v1.2.0

flaviendelangle:
Fixes #4088

I added support for `event.preventBaseUIHandler()` but we could also consider removing it from the types when it was not supported :+1:

### DIFF
--- a/packages/react/src/accordion/header/AccordionHeader.tsx
+++ b/packages/react/src/accordion/header/AccordionHeader.tsx
@@ -30,7 +30,7 @@ export const AccordionHeader = React.forwardRef(function AccordionHeader(
   return element;
 });
 
-export interface AccordionHeaderProps extends BaseUIComponentProps<'h3', AccordionItem.State> {}
+export interface AccordionHeaderProps extends BaseUIComponentProps<'h3', AccordionItem.State, never> {}
 
 export namespace AccordionHeader {
   export type Props = AccordionHeaderProps;
--- a/packages/react/src/accordion/item/AccordionItem.tsx
+++ b/packages/react/src/accordion/item/AccordionItem.tsx
@@ -147,7 +147,7 @@ export interface AccordionItemState extends AccordionRoot.State {
 
 export interface AccordionItemProps
   extends
-    BaseUIComponentProps<'div', AccordionItem.State>,
+    BaseUIComponentProps<'div', AccordionItem.State, never>,
     Partial<Pick<useCollapsibleRoot.Parameters, 'disabled'>> {
   /**
    * A unique value that identifies this accordion item.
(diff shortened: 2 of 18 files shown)


aarongarciah:
> we could also consider removing it from the types when it was not supported

I think this would make more sense. It'd signals wether an event handler can be prevented (i.e. does something on the Base UI side) or not.

flaviendelangle:
The tricky thing is that we have a single interface `BaseUIComponentProps` that applys the `BaseUIEvent` (which contains the faulty method) on ALL event handlers.

We could update the typing to reflect the truth but it would make the types a lot harder to maintain :grimacing: 

With that being said, it's broken on event handlers we do not define at all (like `Accordion.Trigger` `onMouseDown`) and we can't just define every single event handler...

For example:

image

Another solution is to make the callback nullable on all event handlers, but it's not pretty :grimacing:

flaviendelangle:
I'm exploring hardcoding the preventable events on a few components to see what it would mean in practice.

flaviendelangle:
I migrated a few components, but this feels super fragile.
TS can't catch anything if we forget to mark an event as preventable or if we mark an eventable as preventable by mistake.

atomiks:
Seems like this issue exists because only our internal props are marked with `event.preventBaseUIHandler = () => {...}`, instead of their props too (as in, the props unique to the JSX callsite) 

It seems like if both sets of props are marked, then this issue would be fixed without needing to split types or define every handler upfront

It probably does have a non insignificant runtime cost in some cases though, in which case the type-level fix is better

flaviendelangle:
> It seems like if both sets of props are marked, then this issue would be fixed without needing to split types or define every handler upfront

That's another approach indeed, way better than marking all the event handlers that could potentially exist :laughing: 

I think we can discuss it on today's meeting :+1:

PR #4097, opened 2026-02-16
[combobox] Add `getValueFromItem` prop

### LINKED ISSUE #3853: [docs][autocomplete] Confusing TypeScript inference section

### LINKED ISSUE #3865: [combobox] value gets cleared if shape of value differs from items

### LINKED ISSUE #3951: [combobox] Type safety and type inference

atomiks:

Fixes https://github.com/mui/base-ui/issues/3865
Fixes https://github.com/mui/base-ui/issues/3951
Closes https://github.com/mui/base-ui/issues/3853

Smaller and more flexible alternative to #3954. This explicitly supports object `items` with primitive `value`.

`<Combobox.Root>` needs to know what is passed to `<Combobox.Item value={...}>` — `value` could be a primitive `string` or `number`, or the whole item object (when `items` is an array of objects). This new prop `getValueFromItem` lets it know explicitly.

This can't be inferred automatically except in two cases:

- `{ value, label }` is the shape of the `items` objects
- `defaultValue`/`value` aren't `null`

`getValueFromItem` supports arbitrary object shapes and the initial `null` value.

`<Combobox.Root>` will error if `getValueFromItem` is missing and `value`/`defaultValue` don't match the shape of `items`:

```tsx
// @ts-expect-error - primitive selected values with object items require `getValueFromItem`
<Combobox.Root items={objectItems} defaultValue="a" />;
// @ts-expect-error - primitive selected values with object items require `getValueFromItem`
<Combobox.Root items={objectItems} value="a" />;
// @ts-expect-error - primitive selected values with object items require `getValueFromItem`
<Combobox.Root items={numberObjectItems} defaultValue={1} />;
// @ts-expect-error - primitive selected values with object items require `getValueFromItem`
<Combobox.Root items={numberObjectItems} value={1} />;
```

This does require synchronizing `<Combobox.Item value={...}` and `getValueFromItem` together. I added a wrapper example in the spec file. 

The `getValueFromItem` name is differentiated from the other mappers (`itemToStringValue`, `itemToStringLabel`) as those ones operate on item values, not item objects.

---

Explanation why this is needed:

- `Combobox.Root` owns the selection state and matching logic (`selectedIndex`, highlight, input label), so it must know the **selected value shape** before items are interacted with.
- `items` can be objects, but the selected value can be either:
  - the full object, or
  - a primitive extracted from it (`id`, `value`, etc.).
  Root cannot guess which contract you intend.
- `Combobox.Item value={...}` is not a reliable source for automatic inference at Root level: values can be dynamic, transformed, grouped, virtualized, or rendered in different order.
- In single mode, initial `value/defaultValue` can be `null`, so type/value-shape cannot be inferred from those props at mount time.
- The `{ value, label }` object shape is the only safe convention Base UI can special-case automatically; arbitrary object shapes are not inferable.
- `getValueFromItem` gives Root an explicit mapping from item object -> selected value, so all internal comparisons and label resolution stay consistent.
- It also improves TypeScript correctness: the mapper’s return type defines the selected value type and catches mismatches early.

### DIFF
--- a/docs/reference/generated/combobox-root.json
+++ b/docs/reference/generated/combobox-root.json
@@ -85,6 +85,11 @@
       "description": "Filtered items to display in the list.\nWhen provided, the list will use these items instead of filtering the `items` prop internally.\nUse when you want to control filtering logic externally with the `useFilter()` hook.",
       "detailedType": "any[] | Group[] | undefined"
     },
+    "getValueFromItem": {
+      "type": "GetValueFromItemFn<Value, any[] | Group[] | undefined>",
+      "description": "Maps each item in the `items` array to the selected value shape.\nUseful when `items` are objects while `value` / `defaultValue` are primitive values.\nWhen omitted, the full item is used as the selected value.",
+      "detailedType": "| ((item: GroupItem | Item | any) => Value)\n| undefined"
+    },
     "grid": {
       "type": "boolean",
       "default": "false",
@@ -113,7 +118,7 @@
       "detailedType": "((itemValue: Value) => string) | undefined"
     },
     "items": {
-      "type": "any[] | Group[]",
+      "type": "any[] | Group[] | undefined",
       "description": "The items to be displayed in the list.\nCan be either a flat array of items or an array of groups with items.",
       "detailedType": "any[] | Group[] | undefined"
     },
--- a/docs/reference/generated/combobox-root.json
+++ b/docs/reference/generated/combobox-root.json
@@ -76,9 +76,9 @@
       "detailedType": "string | undefined"
     },
     "filter": {
-      "type": "((itemValue: Value, query: string, itemToString: ((itemValue: Value) => string) | undefined) => boolean) | null",
+      "type": "((itemValue: Value | GroupItem | Item | any, query: string, itemToString: ((itemValue: Value | GroupItem | Item | any) => string) | undefined) => boolean) | null",
       "description": "Filter function used to match items vs input query.",
-      "detailedType": "| ((\n    itemValue: Value,\n    query: string,\n    itemToString:\n      | ((itemValue: Value) => string)\n      | undefined,\n  ) => boolean)\n| null\n| undefined"
+      "detailedType": "| ((\n    itemValue: Value | GroupItem | Item | any,\n    query: string,\n    itemToString:\n      | ((\n          itemValue: Value | GroupItem | Item | any,\n        ) => string)\n      | undefined,\n  ) => boolean)\n| null\n| undefined"
     },
     "filteredItems": {
       "type": "any[] | Group[]",
--- a/docs/reference/generated/combobox-root.json
+++ b/docs/reference/generated/combobox-root.json
@@ -76,9 +76,9 @@
       "detailedType": "string | undefined"
     },
     "filter": {
-      "type": "((itemValue: Value | GroupItem | Item | any, query: string, itemToString: ((itemValue: Value | GroupItem | Item | any) => string) | undefined) => boolean) | null",
+      "type": "((itemValue: Value | Item | any, query: string, itemToString: ((itemValue: Value | Item | any) => string) | undefined) => boolean) | null",
       "description": "Filter function used to match items vs input query.",
-      "detailedType": "| ((\n    itemValue: Value | GroupItem | Item | any,\n    query: string,\n    itemToString:\n      | ((\n          itemValue: Value | GroupItem | Item | any,\n        ) => string)\n      | undefined,\n  ) => boolean)\n| null\n| undefined"
+      "detailedType": "| ((\n    itemValue: Value | Item | any,\n    query: string,\n    itemToString:\n      | ((itemValue: Value | Item | any) => string)\n      | undefined,\n  ) => boolean)\n| null\n| undefined"
     },
     "filteredItems": {
       "type": "any[] | Group[]",
@@ -88,7 +88,7 @@
     "getValueFromItem": {
       "type": "GetValueFromItemFn<Value, any[] | Group[] | undefined>",
       "description": "Maps each item in the `items` array to the selected value shape.\nUseful when `items` are objects while `value` / `defaultValue` are primitive values.\nWhen omitted, the full item is used as the selected value.",
-      "detailedType": "| ((item: GroupItem | Item | any) => Value)\n| undefined"
+      "detailedType": "((item: Item | any) => Value) | undefined"
     },
     "grid": {
       "type": "boolean",
--- a/docs/src/app/(docs)/react/components/page.mdx
+++ b/docs/src/app/(docs)/react/components/page.mdx
@@ -329,7 +329,7 @@ An input combined with a list of predefined items to select.
     - Return value
 - Exports:
   - Combobox - Root
-    - Props: actionsRef, autoComplete, autoHighlight, children, defaultInputValue, defaultOpen, defaultValue, disabled, filter, filteredItems, grid, highlightItemOnHover, id, inline, inputRef, inputValue, isItemEqualToValue, itemToStringLabel, itemToStringValue, items, limit, locale, loopFocus, modal, multiple, name, onInputValueChange, onItemHighlighted, onOpenChange, onOpenChangeComplete, onValueChange, open, openOnInputClick, readOnly, required, value, virtualized
+    - Props: actionsRef, autoComplete, autoHighlight, children, defaultInputValue, defaultOpen, defaultValue, disabled, filter, filteredItems, getValueFromItem, grid, highlightItemOnHover, id, inline, inputRef, inputValue, isItemEqualToValue, itemToStringLabel, itemToStringValue, items, limit, locale, loopFocus, modal, multiple, name, onInputValueChange, onItemHighlighted, onOpenChange, onOpenChangeComplete, onValueChange, open, openOnInputClick, readOnly, required, value, virtualized
   - Combobox - Trigger
     - Props: className, disabled, nativeButton, render, style
     - Data Attributes: data-dirty, data-disabled, data-filled, data-focused, data-invalid, data-list-empty, data-placeholder, data-popup-open, data-popup-side, data-pressed, data-readonly, data-required, data-touched, data-valid
--- a/packages/react/src/autocomplete/root/AutocompleteRoot.tsx
+++ b/packages/react/src/autocomplete/root/AutocompleteRoot.tsx
@@ -159,6 +159,7 @@ export interface AutocompleteRootProps<ItemValue> extends Omit<
   | 'onSelectedValueChange'
   | 'fillInputOnItemPress'
   | 'itemToStringValue'
+  | 'getValueFromItem'
   | 'isItemEqualToValue'
   // Different names
   | 'inputValue' // value
--- a/packages/react/src/combobox/item/ComboboxItem.tsx
+++ b/packages/react/src/combobox/item/ComboboxItem.tsx
@@ -31,7 +31,7 @@ export const ComboboxItem = React.memo(
     const {
       render,
       className,
-      value: itemValue = null,
+      value: itemValueProp = null,
       index: indexProp,
       disabled = false,
       nativeButton = false,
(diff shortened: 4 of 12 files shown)


mnajdova (commented):
First quick pass, should the name of the prop be `itemToValue`, to be closer to existing props lie `itemToStringValue`, `itemToStringLabel` etc?

atomiks:
@mnajdova I explained why it's different:

> The getValueFromItem name is differentiated from the other mappers (itemToStringValue, itemToStringLabel) as those ones operate on item values, not item objects.

So:

- `<Combobox.Item value={ItemObject.Primitive}>` -> `itemToStringValue`, `itemToStringLabel` receive `Primitive`
- `<Combobox.Root items={ItemObject[]}>` -> `getValueFromItem` receives `ItemObject`

`itemToValue` would be incongruent with the other mappers if you pass `items={objects}` but `<Combobox.Item value={primitive}>` since the other mappers receive the primitive. (Those mappers are intended to be used only with objects as values for items.)

mnajdova:
Ok, too late to rename those, maybe add it in the v2 ideas. Restructuring the name to have get, doesn't make it more clear that they are receiving different thing as an argument. Ideally those should be named: `itemValueToXXX` in my opinion.

atomiks:
## Codex Review

### Overview
This patch adds `getValueFromItem` so comboboxes can keep object items while storing a different selected value shape. It also updates filtering, label resolution, selected-index sync, typing, and tests so that mapped-value flow behaves the same way as the existing combobox paths.

### Findings (None)

### Confidence: 4/5
High confidence based on a full branch pass, the expanded regression coverage around mapped values, and a follow-up perf check after the latest simplification. Residual risk is mostly around edge-case tradeoffs in very large lists with custom equality.

### Notes
- Latest change: the extra mapped-item arrays and index map are now only built when `getValueFromItem` is actually provided, and the default item path uses the older direct selected-label lookup again.
- Revalidated with `pnpm test:jsdom ComboboxRoot --no-watch`, `pnpm test:chromium ComboboxRoot --no-watch`, and `pnpm typescript`.
- Re-ran the 10,000-item mount + rerender check against `master` after the latest perf fix. The earlier always-on common-path slowdown is no longer consistent across repeated runs:

| Run | Scenario | `master` mean (ms) | `head` mean (ms) |
| --- | --- | ---: | ---: |
| 1 | Primitive items | 3.326 | 3.498 |
| 1 | Object items | 3.034 | 2.992 |
| 2 | Primitive items | 5.241 | 4.770 |
| 2 | Object items | 3.835 | 4.751 |

michaldudak:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

atomiks:
Had a long discussion with Codex then fed it into GPT-5.4 Pro to analyze. Here are the conclusions.

## GPT-5.4 Pro Overview - Combobox keyed values with object items

## Summary

Combobox APIs are easiest to understand when the following all have the same type:

- `items`
- `value` / `defaultValue`
- `<Combobox.Item value={...}>`

That model works well for:

- primitive options such as `string[]`
- object options where the selected value is the whole object

A common application shape does not fit that model as cleanly:

- `items` is an array of objects
- the controlled selection is a primitive key such as an `id`
- the UI still needs a label from the object

Example:

```tsx
const users = [
  { id: '1', name: 'Ada' },
  { id: '2', name: 'Grace' },
];

<Combobox.Root items={users} defaultValue="2">
  {users.map((user) => (
    <Combobox.Item key={user.id} value={user.id}>
      {user.name}
    </Combobox.Item>
  ))}
</Combobox.Root>;
```

In this shape, the selected value is not the same thing as the source item. A Combobox that wants to support this cleanly needs an explicit way to derive the selected key from an item.

## Why a new prop is necessary

### Real application data is often normalized

Many applications store data like this:

```tsx
const post = { authorId: '2' };
const authors = [
  { id: '1', name: 'Ada' },
  { id: '2', name: 'Grace' },
];
```

The UI wants a Combobox bound to `post.authorId`, but the available items are full author objects.

This is a normal and useful shape because it keeps state small and stable:

- IDs are easy to serialize
- IDs are easy to compare
- IDs avoid referential equality problems
- IDs fit well in forms, URLs, and normalized caches

### Labels and identity are different concerns

When items are objects, Combobox usually needs at least two distinct pieces of information:

- **identity**: which option is selected?
- **label**: what text should be shown in the input, used for filtering, or announced to assistive tech?

If the selected value is the whole object, those concerns can often be resolved from the same value.
If the selected value is only a primitive key, they cannot.

That means a stringification prop is not enough on its own. A keyed mode needs:

- a way to derive the selected key from an item
- a way to derive display text from an item

### Existing label/value props are doing different jobs

A prop that converts a value to a string is not the same as a prop that derives the selected identity from an item.

Those are separate responsibilities:

- converting a selected value into a string for form submission
- converting an item into a label for display/filtering
- converting an item into the selected key

Trying to overload one existing prop to do all three makes the API harder to understand.

## Why Base UI needs an explicit keyed mode

Base UI is more flexible than many Combobox libraries because it supports both:

- a root `items` prop
- explicit `<Combobox.Item value={...}>` parts

That flexibility also makes the relationship between source items and selected values more important to define clearly.

In same-shape mode, the rules are simple:

- `items`, `value`, and `<Combobox.Item value>` all line up

In keyed mode, they do not:

- `items` contains objects
- `value` is a primitive key
- the rendered item label still comes from the object

That relationship needs to be stated explicitly in the API rather than emerging indirectly from a combination of label, equality, and serialization props.

## What other libraries do

Most libraries avoid this exact problem by choosing one primary model.

| Library | Source collection | Controlled selected value | Typical model |
| --- | --- | --- | --- |
| React Aria | object items with keys | key / selected key | key-based selection |
| MUI Autocomplete | `options` collection | usually the full option object | object-first |
| React Select | `options` collection | usually the full option object(s) | object-first |
| Downshift | `items` collection | usually the full selected item | object-first |
| Headless UI Combobox | rendered options | primitive or object value | same value shape as each option |

### React Aria

React Aria is the clearest mainstream example of first-class keyed selection:

- items can be objects
- selection is represented by a key
- labels are still derived from item content

This is the closest precedent for supporting object items with primitive selected values directly.

### MUI Autocomplete

MUI Autocomplete is generally object-first:

- `options` is usually an array of objects
- `value` is usually the selected option object
- `getOptionLabel` handles display
- `isOptionEqualToValue` handles equality when references are unstable

When app state stores only an ID, the common pattern is to do an `id -> object` lookup before passing `value` into the component, then convert back to an ID in `onChange`.

### React Select

React Select also usually keeps the public controlled value as the option object or array of option objects.

It exposes helpers like:

- `getOptionLabel`
- `getOptionValue`

But `getOptionValue` does not generally mean the public `value` prop becomes a primitive key API. It is primarily used for option identity and internal behavior while the controlled value still typically uses option objects.

### Downshift

Downshift is also usually item-first:

- `selectedItem` is the selected item
- `itemToString` converts that item into display text

If application state stores only an ID, users typically resolve the selected item outside the component.

### Headless UI Combobox

Headless UI typically uses the same value shape as each option:

- primitives stay primitives
- objects stay objects

If objects are used, comparison helpers can change how equality works, but the selected value still remains object-shaped.

## Why Base UI is different

Base UI sits in a slightly unusual spot because it combines:

- collection-driven data on the root
- explicit item values in the JSX tree

That makes a keyed mode both more possible and more worth defining clearly.

Other libraries often push the normalization boundary outward:

```tsx
const selectedAuthor = authors.find((author) => author.id === post.authorId) ?? null;

<Autocomplete
  options={authors}
  value={selectedAuthor}
  onChange={(_, author) => setAuthorId(author?.id ?? null)}
/>
```

That approach works, but it means the component API is object-first even when the application state is key-first.

A Base UI keyed mode could support the normalized shape directly instead:

```tsx
<Combobox.Root value={post.authorId} items={authors} />
```

That is a meaningful API distinction.

## Why overloading an existing prop is not enough

It may be tempting to avoid a new prop by changing how an existing prop is interpreted.

For example, one idea is to make an existing stringification prop double as the place where the selected key is inferred.

That has a few problems:

1. **The name becomes misleading.** A prop that sounds like string conversion would now also define identity.
2. **The label problem remains.** Even if the selected key can be inferred, the component still needs a label from the item.
3. **The rules become hard to explain.** The API would quietly switch from same-shape mode to keyed mode based on how a stringification prop is used.
4. **It mixes separate concerns.** Identity, display, filtering, and form serialization are related, but they are not the same thing.

If keyed mode is supported, it is better to make that mode explicit.

## Recommended direction

The smallest clear addition is a dedicated prop for deriving the selected key from an item.

For example:

```tsx
<Combobox.Root
  items={users}
  defaultValue="2"
  itemValue="id"
  itemToStringLabel={(item) => item.name}
>
  {users.map((user) => (
    <Combobox.Item key={user.id} value={user.id}>
      {user.name}
    </Combobox.Item>
  ))}
</Combobox.Root>
```

Or function form:

```tsx
<Combobox.Root
  items={users}
  defaultValue="2"
  itemValue={(item) => item.id}
  itemToStringLabel={(item) => item.name}
>
  {users.map((user) => (
    <Combobox.Item key={user.id} value={user.id}>
      {user.name}
    </Combobox.Item>
  ))}
</Combobox.Root>
```

### Semantics

Without `itemValue`:

- `items`, `value`, and `<Combobox.Item value>` stay in the same shape
- existing behavior stays unchanged

With `itemValue`:

- `items` becomes source data
- `value` becomes the selected key
- `<Combobox.Item value>` uses that selected key shape
- label lookup continues to come from the item object

## Tradeoffs

### Benefits

- supports normalized application state directly
- avoids forcing users to rehydrate full objects before rendering
- avoids duplicated object snapshots in component state
- keeps primitive selected values easy to serialize and compare
- makes the object-items + primitive-value case intentional rather than accidental

### Costs

- adds a new concept to the API
- requires docs to explain two modes clearly
- means Base UI must define how selected-label lookup works in keyed mode
- increases the number of combinations between `items`, `value`, and item rendering

## Alternative: object-first only

A simpler alternative is to keep Combobox fully same-shape and recommend object-first usage:

```tsx
const selectedAuthor = authors.find((author) => author.id === post.authorId) ?? null;

<Combobox.Root
  items={authors}
  value={selectedAuthor}
  itemToStringLabel={(author) => author.name}
  itemToStringValue={(author) => author.id}
  isItemEqualToValue={(item, selected) => item.id === selected.id}
>
  {authors.map((author) => (
    <Combobox.Item key={author.id} value={author}>
      {author.name}
    </Combobox.Item>
  ))}
</Combobox.Root>
```

This is a valid model, and it matches what many libraries encourage.

Its downsides are:

- users must do `id -> object` lookup outside the component
- state may duplicate data already present in `items`
- stale labels become a question if selected objects are partial or snapshotted
- async data loading can get awkward when the selected object is not yet available

## Final recommendation

If Base UI wants to keep the API strictly same-shape, it should document that clearly and treat object-items + primitive-selected-value as out of scope.

If Base UI wants to support normalized application data directly, it should add an explicit keyed mode.

The cleanest minimal version of that is:

- keep the existing same-shape model as the default
- add one prop that derives the selected key from an item, such as `itemValue`
- continue using a label prop for display/filtering
- keep identity and string serialization as separate concepts

That makes the most common normalized-data use case first-class without requiring a larger API redesign.

atomiks:
Replacing with https://github.com/mui/base-ui/pull/4985

PR #4098, opened 2026-02-16
[all components][docs] Document missing transition data attributes

atomiks:

### DIFF
--- a/greptile.json
+++ b/greptile.json
@@ -1,0 +1,3 @@
+{
+  "skipReview": "AUTOMATIC"
+}


packages/react/src/drawer/viewport/DrawerViewportDataAttributes.ts
```diff
@@ -0,0 +1,28 @@
...
+   */
+  endingStyle = CommonPopupDataAttributes.endingStyle,
+  /**
+   * Present when the drawer is nested within another drawer.
+   */
+  nested = 'data-nested',
+  /**
+   * Present when the drawer has other open drawers nested within it.
+   */
+  nestedDialogOpen = 'data-nested-dialog-open',
```
LukasTy: Nitpick: did you go for `dialog` here since there are other cases of this data attributes on other `role="dialog"` components?

LukasTy (approved):

packages/react/src/drawer/viewport/DrawerViewportDataAttributes.ts
atomiks: No, this was incorrect as DrawerViewport doesn't have this variable. I've removed it (and if it did have it, it'd be named as `drawer`)

PR #4100, opened 2026-02-16
[all components][test] Normalize test describe naming and grouping

atomiks:

Scanned the codebase with Codex for inconsistent test names or ordering (likely not exhaustive but fixes several issues I saw manually).

### DIFF
--- a/docs/reference/generated/avatar-fallback.json
+++ b/docs/reference/generated/avatar-fallback.json
@@ -22,6 +22,13 @@
       "detailedType": "| ReactElement\n| ((\n    props: HTMLProps,\n    state: Avatar.Fallback.State,\n  ) => ReactElement)"
     }
   },
-  "dataAttributes": {},
+  "dataAttributes": {
+    "data-starting-style": {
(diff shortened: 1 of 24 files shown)


packages/react/src/floating-ui-react/hooks/useDismiss.test.tsx
```diff
@@ -286,27 +286,27 @@ describe.skipIf(!isJSDOM)('useDismiss', () => {
     });
   });
 
-  describe('false', () => {
-    test('dismisses with escape key', async () => {
+  describe('options set to false', () => {
+    test('does not dismiss with escape key', async () => {
```
michaldudak, line 290: Another improvement could be to standardize on `it()`.

LukasTy: In this case, yes, it feels like `it()` would be better.
But for the above prop tests, the usage of `test()` seems slightly more appropriate. 🤷 
I'm not sure if we should strictly stick to just one variant.

LukasTy (approved):

PR #4102, opened 2026-02-16
[drawer] Restore SwipeArea component

atomiks:

### DIFF
--- a/docs/reference/generated/drawer-swipe-area.json
+++ b/docs/reference/generated/drawer-swipe-area.json
@@ -14,20 +14,37 @@
       "detailedType": "boolean | undefined"
     },
     "className": {
-      "type": "string | ((state: DrawerSwipeAreaState) => string | undefined)",
+      "type": "string | ((state: Drawer.SwipeArea.State) => string | undefined)",
       "description": "CSS class applied to the element, or a function that\nreturns a class based on the component’s state.",
-      "detailedType": "| string\n| ((state: DrawerSwipeAreaState) => string | undefined)"
+      "detailedType": "| string\n| ((state: Drawer.SwipeArea.State) => string | undefined)"
     },
     "style": {
-      "type": "CSSProperties | ((state: DrawerSwipeAreaState) => CSSProperties | undefined)",
-      "detailedType": "| React.CSSProperties\n| ((\n    state: DrawerSwipeAreaState,\n  ) => CSSProperties | undefined)\n| undefined"
+      "type": "CSSProperties | ((state: Drawer.SwipeArea.State) => CSSProperties | undefined)",
+      "detailedType": "| React.CSSProperties\n| ((\n    state: Drawer.SwipeArea.State,\n  ) => CSSProperties | undefined)\n| undefined"
     },
     "render": {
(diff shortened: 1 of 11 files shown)


michaldudak:
Sometimes it's impossible to close the drawer by clicking on the backdrop:

https://github.com/user-attachments/assets/1d70d099-cc7a-48f4-8e85-fe8e39c1198b

I'm also still experiencing the original issue where the drawer opens halfway when dragged quickly (both with mouse and touchpad). In these cases it's also impossible to close it by clicking outside.

https://github.com/user-attachments/assets/2e551ae7-a770-4138-81af-ff51a0cc1302

flaviendelangle:
Claude Code mostly complained about tests:

  Issues Found in Tests

  1. Test "re-enables outside press dismissal after opening by swipe" relies on an unrealistic setup

  The test renders <Drawer.Root><Drawer.SwipeArea /></Drawer.Root> with no Portal/Popup/Viewport. After swipe-opening, it clicks document.body and expects the drawer to close via outside-press dismissal.

  This works because floating-ui's useDismiss considers any click as "outside" when there's no popup element (the popup ref is null). In a real app, you'd always render a Popup when the drawer is open. The test is
   technically testing floating-ui's edge behavior rather than the SwipeArea's enableDismissAfterRelease logic in a realistic scenario.

  A more robust test would render the full drawer structure (Portal > Viewport > Popup) and click outside the popup, verifying the drawer closes after the deferred re-enable.

  2. No test coverage for touch input path

  All tests use pointerType: 'mouse'. The component has distinct code paths for touch vs non-touch input (DrawerSwipeArea.tsx:402-429):
  - Non-touch: handled by swipePointerProps (pointer events)
  - Touch: handled by swipeTouchProps (touch events), pointer events are skipped early

  The touch path is entirely untested.

  3. No test for the data-swiping attribute

  The component maps swiping state to data-swiping (line 43-44), but no test verifies this attribute appears during an active swipe gesture.

  4. The size-based threshold test could be fragile

  In "uses a size-based swipe threshold by default", the waitBeforeRelease callback uses a hardcoded wait(120) to ensure the popup has mounted and laid out. While this is Chromium-only (skipping JSDOM), the fixed
  timing could be fragile under CI load. Consider using a layout-readiness check instead of a fixed delay.

atomiks:
## Codex Review

### Overview
This patch restores `Drawer.SwipeArea` as a public drawer part, brings its docs and demo back, and adds focused coverage for swipe-to-open behavior and public state hooks. The latest follow-up also keeps the demo from trapping page scroll when the cursor is over the example panel.

### Findings (None)

### Confidence: 5/5
High confidence from a full-branch diff review and targeted reruns on the rebased branch, including the touch path, the size-threshold path, and the controlled swipe-dismiss cleanup path.

### Notes
- Re-reviewed after rebasing onto the latest `master`.
- Verified with `pnpm test:jsdom DrawerSwipeArea --no-watch`, `pnpm test:chromium DrawerSwipeArea --no-watch`, and `pnpm test:chromium DrawerRoot --no-watch --testNamePattern "does not dismiss a controlled drawer via swipe when open is always true"`.

PR #4107, opened 2026-02-16
[scroll area] Fix remount thumb size

### LINKED ISSUE #3829: [scroll-area] has overflow data attributes applied on first render before measure

### LINKED ISSUE #4103: [scroll-area] Thumb flashes at wrong size on remount
# Bug report

## Current behavior

 When `ScrollArea` is unmounted and remounted (inside a baseui `Tabs.Panel` in this case), the scrollbar thumb renders at the wrong size before correcting on the next user scroll. This is visible when `ScrollArea.Scrollbar` has`display: flex` or `display: grid` and the thumb has any padding.

The thumb also flashes with an initial height of 0 on page load with Next.js before correcting

https://github.com/user-attachments/assets/b36eae01-3c04-41b1-b315-1509b6ba0dcd

## Expected behavior

The thumb should render at the correct size immediately on mount

## Reproducible example

Stackblitz with repro in next.js: https://stackblitz.com/edit/stackblitz-starters-d9hre5ba?file=app%2Fpage.tsx

## Base UI version

1.2.0

Chrome

## Which OS are you using?

Mac OS

n/a

## Additional context

After grappling with this issue for a while, I was curious if this was a CSS issue on my end or a potential bug in the library. After reducing my implementation down to a bare minimum example, and still getting this issue, I started to suspect that the thumb size was being initialized with an incorrect value and corrected after the first paint.

To debug I went into `node_modules` and checked out `ScrollAreaScrollbar` and `ScrollAreaViewport `and left console logs to see what the `--scroll-area-thumb-height` css variable was being set to, and how often it was changing. 

On mount, it _did_ initialize to 0 before updating, and when switching tabs and then scrolling, the height went through three renders from 0px -> 403px on mount, and then to 464px when I began scrolling, which aligns with what I was visually observing.

I was able to partially work around this issue by passing `keepMounted` to my `Tabs.Panel` component, but the initial page load flash persists.

atomiks:

Fixes #4103
Fixes #3829

Preview: https://deploy-preview-4107--base-ui.netlify.app/experiments/scroll-area/tabs-scroll-area

For inset scrollbars, they are hidden (unless `keepMounted` is specified) before hydration.
This PR SSR: https://deploy-preview-4107--base-ui.netlify.app/experiments/scroll-area/scroll-area-inset
Master SSR: https://master--base-ui.netlify.app/experiments/scroll-area/scroll-area-inset

### DIFF
--- a/packages/react/src/scroll-area/root/ScrollAreaRoot.test.tsx
+++ b/packages/react/src/scroll-area/root/ScrollAreaRoot.test.tsx
@@ -140,7 +140,7 @@ describe('<ScrollArea.Root />', () => {
       });
     });
 
-    it('keeps scrollbars hidden until the first ResizeObserver measurement', async () => {
+    it('shows scrollbars after mount compute before the first ResizeObserver measurement', async () => {
       await withMockResizeObserver(async (notifyResizeObserver) => {
         await render(
           <ScrollArea.Root style={{ width: VIEWPORT_SIZE, height: VIEWPORT_SIZE }}>
@@ -156,7 +156,7 @@ describe('<ScrollArea.Root />', () => {
         const verticalScrollbar = await screen.findByTestId('vertical-scrollbar');
 
         await waitFor(() => {
-          expect(getComputedStyle(verticalScrollbar).visibility).to.equal('hidden');
+          expect(getComputedStyle(verticalScrollbar).visibility).to.equal('visible');
         });
 
         await act(async () => {
@@ -169,7 +169,7 @@ describe('<ScrollArea.Root />', () => {
       });
     });
 
-    it('keeps keepMounted scrollbar track visible while thumb waits for first measurement', async () => {
+    it('shows keepMounted scrollbar track and thumb after mount compute', async () => {
       await withMockResizeObserver(async (notifyResizeObserver) => {
         await render(
           <ScrollArea.Root style={{ width: VIEWPORT_SIZE, height: VIEWPORT_SIZE }}>
@@ -191,7 +191,7 @@ describe('<ScrollArea.Root />', () => {
 
         await waitFor(() => {
           expect(getComputedStyle(verticalScrollbar).visibility).to.equal('visible');
-          expect(getComputedStyle(verticalThumb).visibility).to.equal('hidden');
+          expect(getComputedStyle(verticalThumb).visibility).to.equal('visible');
         });
 
         await act(async () => {
--- a/packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
+++ b/packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
@@ -132,10 +132,15 @@ export const ScrollAreaViewport = React.forwardRef(function ScrollAreaViewport(
     const scrollTop = viewportEl.scrollTop;
     const scrollLeft = viewportEl.scrollLeft;
     const lastMeasuredViewportMetrics = lastMeasuredViewportMetricsRef.current;
+    const isFirstMeasurement = Number.isNaN(lastMeasuredViewportMetrics[0]);
     lastMeasuredViewportMetrics[0] = viewportHeight;
     lastMeasuredViewportMetrics[1] = scrollableContentHeight;
     lastMeasuredViewportMetrics[2] = viewportWidth;
     lastMeasuredViewportMetrics[3] = scrollableContentWidth;
+
+    if (isFirstMeasurement) {
+      setHasMeasuredScrollbar(true);
+    }
 
     if (scrollableContentHeight === 0 || scrollableContentWidth === 0) {
       return;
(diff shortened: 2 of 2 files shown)


atomiks:
@codex

atomiks:
## Codex Review

### Overview
This PR stabilizes `ScrollArea` remount and hydration behavior by consolidating mount-time initialization into the viewport measurement path, while expanding Chromium coverage around overflow, corner recomputation, and remount scenarios.

### Findings
#### 1. 🟡 [Non-blocking] Mount-time CPU is still slower than latest `master` in Chromium stress runs

**Impact**: UIs that mount many `ScrollArea` instances can still see a measurable initial-render CPU increase. This is not a correctness blocker, but it is a runtime regression in the tested mount-heavy scenario.

**Evidence**: Consolidation landed in `packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx` (single-path mount initialization through `computeThumbPosition` and removal of duplicate mount-time readiness updates), and Chromium benchmark was re-run against latest `upstream/master` (`205a9a05a`) using interleaved runs (`master` then `head`) with 1,000 mounted `ScrollArea` instances per sample.

Command/method: `pnpm test:chromium ScrollAreaMountPerf.temp --no-watch`, measuring from render start to two `requestAnimationFrame` ticks.

| Ref | Runs | Mean (ms) | p50 (ms) | p95 (ms) | Sample size |
| --- | ---: | ---: | ---: | ---: | ---: |
| `upstream/master` (`205a9a05a`) | 1-5 (interleaved) | 477.42 | 473.90 | 491.80 | 5 |
| `PR head` (`691d8d27a`) | 1-5 (interleaved) | 521.92 | 521.20 | 534.50 | 5 |
| **Delta (head vs master)** | - | **+44.50 (+9.32%)** | **+47.30** | **+42.70** | - |

**Recommendation**: Keep trimming mount initialization transitions in the source path (for example, avoid separate phase updates where equivalent state can be derived in the same compute cycle), then rerun the same interleaved Chromium table against latest `master`.

#### 2. ℹ️ [Informational] Simplification direction is correct, but readiness state still spans multiple components

**Impact**: The recent consolidation reduced duplicate observer/mount work, but readiness/visibility policy still depends on cross-component coordination, which makes future perf tuning and bundle-size tradeoffs harder to reason about.

**Evidence**: Readiness is produced in `packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx`, then consumed for visibility in `packages/react/src/scroll-area/scrollbar/ScrollAreaScrollbar.tsx` and `packages/react/src/scroll-area/thumb/ScrollAreaThumb.tsx`.

**Recommendation**: Keep the simplified single initialization path and continue minimizing cross-component readiness plumbing so mount-time behavior stays easier to optimize without adding new abstraction weight.

### Confidence: 4/5
High confidence based on a full-branch pass, targeted Scroll Area validation in Chromium/JSDOM, and refreshed interleaved Chromium perf sampling against latest `upstream/master`.

### Notes
- Verified test coverage remains green after the consolidation (`pnpm test:chromium ScrollArea --no-watch`, `pnpm test:jsdom ScrollArea --no-watch`).

PR #4110, opened 2026-02-16
[avatar] Remove fallback transition logic and premature image display

### LINKED ISSUE #4109: [avatar] Incorrect Image/Fallback behaviour
# Bug report

## Current behavior

`<Avatar.Image />` is mounted while the image is still loading. At the same time, `<Avatar.Fallback />` is still in place, causing layout shifts.

## Expected behavior

`<Avatar.Image />` only mounts when image is loaded (similar to v1.1.0).

## Reproducible example

https://stackblitz.com/edit/sb1-fkfxj8ab

## Base UI version

v1.2.0

Chrome/Safari

## Which OS are you using?

Mac OS

atomiks:

Setting visibility of the image to include `imageLoadingStatus === 'loading'` causes the image to show prematurely. Additionally, transitioning the fallback component out is not particularly useful and regresses layout cases where the expectation is that only one is mounted at a time.

Fixes #4109 (https://stackblitz.com/edit/sb1-urlzxstp?file=package.json,src%2FApp.tsx)

Image transition remains the same once it's loaded: https://deploy-preview-4110--base-ui.netlify.app/experiments/transition-attrs. If it hasn't been loaded yet (and not in the cache), it may not appear to play if its duration is short, as the image progressively loads in so it plays just before painting. For speed, this is likely preferable.

### DIFF
--- a/docs/src/app/(docs)/react/components/dialog/page.mdx
+++ b/docs/src/app/(docs)/react/components/dialog/page.mdx
@@ -10,6 +10,10 @@ import { DemoDialogHero } from './demos/hero';
 
 <DemoDialogHero />
 
+## Usage guidelines
+
+- **Dialog doesn't support gestures:** Use [Drawer](/react/components/drawer) when you need gesture support or snap points. A panel that slides in from the edge of the screen and doesn't need gesture support is a positioned Dialog.
+
 ## Anatomy
 
 Import the component and assemble its parts:
--- a/docs/src/app/(docs)/react/components/drawer/page.mdx
+++ b/docs/src/app/(docs)/react/components/drawer/page.mdx
@@ -12,6 +12,10 @@ import { DemoDrawerHero } from './demos/hero';
 
 <DemoDrawerHero />
 
+## Usage guidelines
+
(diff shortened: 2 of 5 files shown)


packages/react/src/avatar/fallback/AvatarFallback.tsx
```diff
@@ -32,51 +23,29 @@ export const AvatarFallback = React.forwardRef(function AvatarFallback(
...
-  if (!mounted) {
-    return null;
-  }
-
   return element;
 });
 
-export interface AvatarFallbackState extends AvatarRoot.State {
-  transitionStatus: TransitionStatus;
-}
```
atomiks, line 79: This was undocumented (in 1.2.0), should be safe to revert.

packages/react/src/avatar/fallback/AvatarFallback.tsx
```diff
@@ -32,51 +23,29 @@ export const AvatarFallback = React.forwardRef(function AvatarFallback(
...
-
   React.useEffect(() => {
     if (delay !== undefined) {
       timeout.start(delay, () => setDelayPassed(true));
     }
     return timeout.clear;
   }, [timeout, delay]);
 
-  const state: AvatarFallback.State = {
+  const state: AvatarRoot.State = {
```
LukasTy:
```suggestion
  const state: AvatarFallbackState = {
```

Issue #4095, opened 2026-02-15
[form] field-level validate skips first change after error added to errors prop


DavidCodina:
# Bug Report

I'm not sure if this is ultimately a bug with `Form` or `Field`.

When `Form`'s `errors={errors}` contains an error associated to a particular field, the first change to that field skips the `Field.Root`'s `validate` handler. Presumably, there's some kind of internal logic that tells the `Field.Root` to disregard the `errors` prop after the first subsequent change. That part makes sense. However, `Field.Root` shouldn't opt out of calling the `validate` handler on that first subsequent change. 

## Expected Behavior

Given an associated error in `Form`'s `errors` prop, the first subsequent change to the field should still run field-level validation logic.

## Reproducible Example

In this example if one were to type in `abcde` then submit, validation would fail within `onFormSubmit`. However, if one then deleted the `e`, _validation would not fail_ in the field-level `validate` (_but it should_) because field-level validation doesn't run. In practice, the `Field.Root` and `Input` are given `data-valid`. If  the `d` was then deleted, validation would fail because now field-level validation is running again.

```
import * as React from 'react'
import { Form } from '@base-ui/react/form'
import { Field } from '@base-ui/react/field'
import { Input } from '@base-ui/react/input'

const FIELD_FOCUS_MIXIN = `
focus-visible:shadow-none
focus-visible:border-blue-500
focus-visible:ring-[3px]
focus-visible:ring-blue-500/40
`

const FIELD_VALID_MIXIN = `
data-valid:not-data-disabled:border-green-500
data-valid:focus-visible:border-green-500
data-valid:focus-visible:ring-green-500/40
`

const FIELD_INVALID_MIXIN = `
data-invalid:not-data-disabled:border-red-500
data-invalid:focus-visible:border-red-500
data-invalid:focus-visible:ring-red-500/40
`

const inputClasses = `
flex bg-card
w-full min-w-0
px-[0.5em] py-[0.25em]
rounded-[0.375em]
border border-neutral-400 outline-none
placeholder:text-muted-foreground
${FIELD_FOCUS_MIXIN}
${FIELD_VALID_MIXIN}
${FIELD_INVALID_MIXIN}
`

/* ======================

====================== */

export const ValidationBugDemo = () => {
  const [resetKey, setResetKey] = React.useState(0)
  const [submitting, setSubmitting] = React.useState(false)
  const [errors, setErrors] = React.useState<Record<string, string>>({})

  return (
    <Form
      noValidate
[...]

PR #4112, opened 2026-02-16
[field] Fix field validation with simultaneous Form errors or invalid prop

mj12albert:
Fixes https://github.com/mui/base-ui/issues/4095

Also handles an edge case when a field simultaneously has an error `Form.errors` and `invalid={true}`


packages/react/src/field/root/FieldRoot.tsx
```diff
@@ -75,9 +76,11 @@ const FieldRootInner = React.forwardRef(function FieldRootInner(
...
   );
 
-  const invalid = Boolean(
-    invalidProp || (name && {}.hasOwnProperty.call(errors, name) && errors[name] !== undefined),
-  );
+  const hasFormError =
+    name !== undefined &&
+    EMPTY_OBJECT.hasOwnProperty.call(errors, name) &&
+    errors[name] !== undefined;
+  const invalid = invalidProp === true || hasFormError;
```
atomiks: We can use `Object.hasOwn(...)` now

packages/react/src/field/root/FieldRoot.tsx
```diff
@@ -75,9 +76,11 @@ const FieldRootInner = React.forwardRef(function FieldRootInner(
       (validationMode === 'onSubmit' && submitAttemptedRef.current),
   );
 
-  const invalid = Boolean(
-    invalidProp || (name && {}.hasOwnProperty.call(errors, name) && errors[name] !== undefined),
-  );
+  const hasFormError =
+    name !== undefined &&
```
atomiks: Codex:
> [P2] Empty-string name is now handled inconsistently across form error paths. [FieldRoot.tsx (line 79)](app://-/index.html#) now includes name="" in hasFormError, but related logic still treats empty string as “no name” in:
[Form.tsx (line 138)](app://-/index.html#),
[FieldError.tsx (line 41)](app://-/index.html#),
and form value collection in [Form.tsx (line 122)](app://-/index.html#).

Leaving it as truthy `name` is probably simplest?

PR #4125, opened 2026-02-18
[menu] Prevent `pointerleave` from stealing dialog focus

atomiks:

When clicking a menu item that's a dialog trigger with a pointer, focus would be taken from the dialog by `onPointerLeave`.


atomiks:
## Codex Review

### Overview
This PR updates pointer-leave handling in list navigation so menu focus is only restored when focus is still inside the menu floating element, which prevents nested alert dialog focus from being stolen. It also adds a Chromium-only regression test covering a nested `AlertDialog` trigger inside `Menu`.

### Findings (None)

### Confidence: 4/5
High confidence from a full branch-diff review (`master...c750efaa`), targeted unit coverage checks, and a focused hot-path perf spot-check.

### Notes
- Full branch scope reviewed (single commit branch).
- Test evidence: `pnpm test:jsdom MenuRoot --no-watch` and `pnpm test:chromium MenuRoot --no-watch` both passed.
- Perf spot-check method: Chromium micro-benchmark via `@playwright/test` comparing unconditional focus vs `activeElement + contains + conditional focus` over 5 runs × 100,000 iterations.
- Perf spot-check result: mean overhead was ~`6.26ms` per `100,000` iterations (~`0.063µs` per event), which is negligible in practice.

PR #4132, opened 2026-02-18
[useButton] Avoid checking disabled twice in `onKeyDown` and `onKeyUp`

flaviendelangle:
Tiny improvements on `useButton`:

- [x] Do not compute `isSpaceKey` and `isEnterKey` if `shouldClick` is false
- [x] Add early return in `onKeyDown` and `onKeyUp` when `disabled` is true since there is no behavior applied anyway
- [x] Remove a `useStableCallback` that was only used in one place and didn't access any thing in the scope except a ref

### DIFF
--- a/packages/react/src/use-button/useButton.ts
+++ b/packages/react/src/use-button/useButton.ts
@@ -22,11 +22,6 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
 
   const isCompositeItem = useCompositeRootContext(true) !== undefined;
 
-  const isValidLink = useStableCallback(() => {
-    const element = elementRef.current;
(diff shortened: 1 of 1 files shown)


packages/react/src/use-button/useButton.ts
```diff
@@ -119,25 +114,24 @@ export function useButton(parameters: useButton.Parameters = {}): useButton.Retu
...
+            makeEventPreventable(event);
+            externalOnKeyDown?.(event);
             if (event.baseUIHandlerPrevented) {
               return;
             }
 
+            // Keyboard accessibility for non interactive elements
             const shouldClick =
               event.target === event.currentTarget &&
               !isNativeButton &&
```
flaviendelangle, line 130: Is this check really needed since we throw an error when `isNativeButton` is true and the element is actually a link. 👀

PR #4133, opened 2026-02-18
[drawer] Prevent dismiss through swipe when the component is controlled

### LINKED ISSUE #4119: [drawer] Dismissible trough swipe in controlled mode
# Bug report

## Current behavior

In controlled mode the drawer is still closed when swiping down, even if the controlled prop is dismissing the closed action. 

## Expected behavior

The drawer should not close. Or at least pop open again. This can be used to abort the swipe down event e.g. when a form is not saved.

## Reproducible example

https://stackblitz.com/edit/kmtjez8e?file=src%2FApp.tsx

## Base UI version

1.2.0

flaviendelangle:
Fixes #4119

[Before](https://stackblitz.com/edit/kmtjez8e?file=src%2FApp.tsx)
[After](https://stackblitz.com/edit/kmtjez8e-dretczn2?file=package.json)

### DIFF
--- a/packages/react/src/drawer/root/DrawerRoot.test.tsx
+++ b/packages/react/src/drawer/root/DrawerRoot.test.tsx
@@ -885,7 +885,7 @@ describe('<Drawer.Root />', () => {
         expect(handleOpenChange).toHaveBeenCalledWith(false, expect.anything());
 
         // Wait for the rAF that reverts the controlled dismiss
-        await new Promise((resolve) => requestAnimationFrame(resolve));
+        await new Promise(requestAnimationFrame);
 
         // The drawer should remain open without data-swipe-dismiss
         expect(popup).not.toHaveAttribute('data-swipe-dismiss');


packages/react/src/dialog/store/DialogStore.ts
```diff
@@ -86,6 +86,14 @@ export class DialogStore<Payload> extends ReactStore<
       return;
     }
 
+    // In controlled mode, don't update internal state.
+    // The parent must update the `open` prop for the state to change.
+    // Cancel the event so callers (e.g. DrawerViewport.onDismiss) clean up properly.
+    if (this.state.openProp !== undefined) {
+      eventDetails.cancel();
+      return;
+    }
```
atomiks: This seems too broad and is likely to break other cases. Indeed that's what Codex found:

<details>
<summary>PR #4133 review findings (2 regressions)</summary>

### Summary
This PR fixes controlled Drawer swipe-dismiss cleanup and adds a test, but introduces two controlled-mode regressions that should be addressed before merge.

### Details
1. **[P1] Controlled `Dialog`/`Drawer` no longer emit `openchange`**
   - `DialogStore.setOpen` now cancels+returns in controlled mode before `openchange` emit.
   - `FloatingFocusManager` uses `openchange` to compute close interaction type and return-focus behavior.
   - Impact: controlled `finalFocus={(type) => ...}` and outside-press focus return behavior can become stale/incorrect.

2. **[P2] Controlled Drawer with snap points can restore previous snap point after swipe-close**
   - Forced cancel in controlled mode makes Drawer swipe dismiss always follow the canceled-cleanup branch.
   - That branch restores the pending snap point, which can conflict with the default snap-point reset on close.
   - Impact: if parent accepts close (`open=false`), reopening may start at the old snap point instead of default.

</details>

flaviendelangle: Aren't we doing to much magic with the controlled value here then ? 
I would love to just have an internal `state.open` and if it's controlled then I don't update it and nothing changes :thinking:

atomiks: In the current architecture `state.open` is still used by internal selectors/consumers beyond rendering.

If we stop updating `state.open` in controlled mode, we regress at least:
1) trigger open state (`data-popup-open` via `isOpenedByTrigger`), and
2) `DialogHandle.isOpen` (it reads `store.state.open`).

For this PR, I would keep the Drawer swipe fix localized in `DrawerViewport`, similar to the "workaround" I posted in the issue

<details>
<summary>PR #4133 updated review findings (2 regressions)</summary>



### Summary
The latest update fixes the earlier `openchange` emission issue, but introduces two new controlled-mode regressions.

### Details

1. **[P1] Controlled `Dialog`/`Drawer` triggers lose open state**
   - `DialogStore.setOpen` now avoids updating internal `state.open` in controlled mode.
   - Trigger-open state is still computed from `state.open` (`isOpenedByTrigger`), and `DialogTrigger` uses that to set trigger `open`/`data-popup-open`.
   - Impact: in standard controlled usage (`open` + `onOpenChange={setOpen}`), popup can be open while trigger appears closed (`data-popup-open` missing).

2. **[P2] `DialogHandle.isOpen` becomes stale in controlled mode**
   - `DialogHandle.isOpen` returns `store.state.open`.
   - With the new controlled-mode behavior, `state.open` may no longer track effective open state.
   - Impact: `handle.isOpen` can report incorrect values (commonly `false` while actually open).

</details>

atomiks:
Codex still found an issue with the latest change, but it resolved the broader issues with the DialogStore

<details>
<summary>PR #4133 latest review findings (2 issues)</summary>

### Summary
The latest update narrows the fix to `DrawerViewport`, but one key regression appears to remain in controlled + snap points swipe-close behavior, and the tests still miss that path.

### Details

1. **[P1] Controlled swipe-close is treated as canceled even when parent accepts close**
   - In `DrawerViewport.onDismiss`, dismiss is force-canceled when `!dismissEventDetails.isCanceled && store.select('open')`.
   - In controlled mode, `store.select('open')` is typically still `true` immediately after `setOpen(false, ...)` (before parent re-render), so this often cancels even accepted closes.
   - Canceled cleanup restores `pendingSwipeCloseSnapPointRef`, which can override the close-reset done in `DrawerRoot`.
   - Impact: controlled drawer with `snapPoints` can reopen at the previous snap point after a swipe close that parent actually accepted.

2. **[P2] Test coverage still misses accepted controlled close with snap points**
   - Added test covers only “controlled and always open” (parent refuses close), with no `snapPoints`.
   - Missing: controlled + `snapPoints` + parent sets `open=false` on swipe.
   - That exact path is where the remaining regression is likely.

</details>

atomiks:
## Codex Review

### Overview
This patch fixes controlled drawer swipe-close behavior so a refused close no longer forces the drawer away, while an accepted close still resets snap points correctly. The latest follow-up also aligns the frame scheduling with the shared utility and removes repeated swipe-test setup.

### Findings (None)

### Confidence: 4/5
High confidence from a full `master...head` pass and the added focused coverage for both rejected and accepted controlled swipe-close paths. Residual risk is limited to less common async controlled-close flows.

### Notes
- Latest changes since the previous review addressed the two earlier non-blocking notes: `DrawerViewport` now uses the shared animation-frame helper, and the swipe-dismiss tests now share a single environment setup helper.

flaviendelangle:
@atomiks done

PR #4137, opened 2026-02-19
[select] Fix SelectValue placeholder not rendered with Record items

### LINKED ISSUE #4136: [select] SelectValue placeholder not rendered when items is a Record without a "null" key
## Bug description

`<Select.Value placeholder="..." />` renders an empty `<span>` when the `items` prop on `<Select.Root>` is a `Record<string, ReactNode>` that doesn't contain a `"null"` key. The placeholder text is completely absent from the DOM.

## Steps to reproduce

```tsx
const items = {
  sans: 'Sans-serif',
  serif: 'Serif',
  mono: 'Monospace',
};

<Select.Root items={items}>
  <Select.Trigger>
    <Select.Value placeholder="Choose a font..." />
  </Select.Trigger>
  <Select.Portal>
    <Select.Positioner>
      <Select.Popup>
        <Select.List>
          {Object.entries(items).map(([value, label]) => (
            <Select.Item key={value} value={value}>
              <Select.ItemText>{label}</Select.ItemText>
            </Select.Item>
          ))}
        </Select.List>
      </Select.Popup>
    </Select.Positioner>
  </Select.Portal>
</Select.Root>
```

**Expected:** The trigger shows "Choose a font..." as placeholder text.
**Actual:** The trigger renders an empty `<span>`.

## Root cause

The logic in `hasNullItemLabel()` in `packages/react/src/utils/resolveValueLabel.tsx` is inverted for the Record branch:

```ts
export function hasNullItemLabel(items: ItemsInput): boolean {
  if (!Array.isArray(items)) {
    return items != null && !('null' in items);  // ← bug: `!` inverts the check
  }
  // ...
}
```

For a Record like `{ sans: "Sans-serif", serif: "Serif" }`:
- `'null' in items` → `false` (no `"null"` key)
- `!('null' in items)` → `true`
- Function returns `true`, meaning "the null value has a label"

This causes `SelectValue` to skip the placeholder (line 52):
```ts
} else if (!hasSelectedValue && placeholder != null && !hasNullLabel) {
  children = placeholder;
}
```

Since `hasNullLabel` is `true`, the condition is `false`, and the placeholder is never rendered.

The fix should be removing the `!`:
```ts
return items != null && ('null' in items);
```

The existing tests in `resolveValueLabel.test.ts` only cover grouped array items — no test covers the Record branch.

## Workaround

Use array format for `items` instead of Record:
```tsx
const items = [
  { value: 'sans', label: 'Sans-serif' },
  { value: 'serif', label: 'Serif' },
  { value: 'mono', label: 'Monospace' },
];
```

The array path in `hasNullItemLabel` has correct logic (`item.value == null` check).

## Environment

- `@base-ui/react` version: 1.2.0 (also present in 1.1.0)
- Browser: all (server-rendered HTML already has empty span)
- Introduced in PR #3604

vcode-sh:

## Summary

`<Select.Value placeholder="..." />` renders an empty `<span>` when `items` is a `Record<string, ReactNode>` without a `"null"` key. This is caused by inverted logic in `hasNullItemLabel()`.

The same function is shared by Combobox, so `<Combobox.Value placeholder="..." />` with Record items is also affected.

**Before (buggy):**
```ts
// packages/react/src/utils/resolveValueLabel.tsx
if (!Array.isArray(items)) {
  return items != null && !('null' in items);
}
```

For `{ sans: "Sans-serif", serif: "Serif" }`:
- `!('null' in items)` → `true` → returns `true`
- `SelectValue` skips the placeholder because it thinks the null value already has a label

**After (fixed):**
```ts
if (!Array.isArray(items)) {
  return items != null && 'null' in items;
}
```

## Changes

- **`resolveValueLabel.tsx`**: Remove the `!` operator in the Record branch of `hasNullItemLabel()`
- **`resolveValueLabel.test.ts`**: Add tests for Record items (with and without `"null"` key), flat array items, and `undefined`

Fixes #4136


packages/react/src/utils/resolveValueLabel.test.ts
```diff
@@ -41,5 +41,47 @@ describe('resolveValueLabel', () => {
 
       expect(hasNullItemLabel(items)).to.equal(false);
     });
+
+    it('returns true when flat items contain a null-valued item with a label', () => {
+      const items = [
+        { value: 'a', label: 'A' },
+        { value: null, label: 'None' },
```
atomiks, line 48: Thanks for the fix @vcode-sh, can you add these to `SelectValue.test.tsx` too:

```
    it('displays placeholder when object items do not have a null key', async () => {
      const items = {
        option1: 'Option 1',
        option2: 'Option 2',
      };

      await render(
        <Select.Root items={items}>
          <Select.Value data-testid="value" placeholder="Select an option" />
        </Select.Root>,
      );

      expect(screen.getByTestId('value')).to.have.text('Select an option');
    });

    it('null key label in object items takes precedence over placeholder', async () => {
      const items = {
        null: 'None',
        option1: 'Option 1',
      };

      await render(
        <Select.Root items={items}>
          <Select.Value data-testid="value" placeholder="Select an option" />
        </Select.Root>,
      );

      expect(screen.getByTestId('value')).to.have.text('None');
    });
```

vcode-sh:
Thanks @atomiks

PR #4138, opened 2026-02-19
[accordion] Add generic `Value` typing 

atomiks:

Same updates as https://github.com/mui/base-ui/pull/3173 and https://github.com/mui/base-ui/pull/4033


packages/react/src/accordion/root/AccordionRoot.tsx
```diff
@@ -235,8 +242,9 @@ export type AccordionRootChangeEventDetails =
   BaseUIChangeEventDetails<AccordionRoot.ChangeEventReason>;
 
 export namespace AccordionRoot {
-  export type State = AccordionRootState;
-  export type Props = AccordionRootProps;
+  export type Value<TValue = any> = AccordionValue<TValue>;
```
LukasTy, line 245: Nitpick: We could align on `Value` or `TValue` everywhere. 😃

packages/react/src/accordion/root/AccordionRoot.tsx
```diff
@@ -158,32 +160,37 @@ export const AccordionRoot = React.forwardRef(function AccordionRoot(
       <CompositeList elementsRef={accordionItemRefs}>{element}</CompositeList>
     </AccordionRootContext.Provider>
   );
-});
+}) as {
+  <Value = any>(props: AccordionRoot.Props<Value>): React.JSX.Element;
```
LukasTy, line 164: Doesn't it also need this part?

```suggestion
  <Value = any>(props: AccordionRoot.Props<Value> & React.RefAttributes<HTMLDivElement>): React.JSX.Element;
```

atomiks: The `Props` interface already includes the `ref` (after a change a couple months ago). It looks like the others should be updated to remove it.

LukasTy: Oh, nice, good spot. 👍

PR #4140, opened 2026-02-19
[tooltip] Add `closeOnClick` prop

### LINKED ISSUE #4113: [tooltip] Add a way to prevent closing a tooltip on click
# Feature request

## Summary

I would expect to be able to do that with Base UI given how flexible it is, and indeed it is possible to control the `open` prop and ignore `onOpenChange` events when the reason for the change is `'trigger-press'`, but it is not perfect* and I ultimately believe it would be better to officially support this use case with a `preventCloseOnClick` boolean prop or something.

*Check it out [here](https://stackblitz.com/edit/vtlgj8hk?file=src%2FApp.tsx), and watch the following video that shows 2 issues with this solution which I haven't figured out how to cleanly fix:

1. When clicking and then moving the cursor away, the tooltip shows _instantly_ when the cursor re-enters the trigger, instead of waiting for the configured `delay`. Note: this is actually a "regression" in version 1.2.0  (potentially introduced in #4049?); it didn't occur in version 1.1.0 but since this use case is not documented, I don't know if it would count as a bug...
2. When clicking the trigger _before_ the tooltip has a chance to appear (before the `delay`), the tooltip just never appears. This can be worked around by setting a timeout that opens the tooltip in the `!newOpen && eventDetails.reason === 'trigger-press'` branch (and clearing it if `onOpenChange` is called again before it runs), but it feels pretty fragile, and still not perfect.

https://github.com/user-attachments/assets/d91f2eaf-9cab-4d37-a704-3618cfdab7aa

Dunno.

## Motivation

The designer in my team wants our `Tooltip` component to remain open even after clicking the button it's attached to. This sort of makes sense to me for many use cases, like buttons that you can click multiple times (toggle or increment/decrement) or buttons that have a tooltip whose content changes on click ("Copy" => "Copied").

atomiks:

Closes #4113

`closeOnClick` is `true` by default, but this is acceptable over `false` (e.g. `preventCloseOnClick`) if it's for consistency with other props, and `closeOnClick` is used elsewhere like `Menu.Item`

Demo: https://stackblitz.com/edit/7dzwxbku?file=src%2FApp.tsx


colmtuite:
I want to discuss this one more. I don't currently think it's useful for any design.

atomiks:
@colmtuite it's partially useful even for anchored toasts

It's somewhat of a micro-detail, but you don't want the tooltip to dismiss in the period between `pointerdown` and `pointerup`, so the toast appears immediately on click:

Screenshot 2026-02-21 at 6 23 54 pm

---

The Slider label use case is indeed questionable though — because it seems like Slider should have its own tooltip-like label part, instead of trying to reuse Tooltip (as Tooltip doesn't support touch, which is yet another config option)

Issue #4021, opened 2026-02-09
[Button] Button applying "role=button" to links


mrmckeb:
# Bug report

## Current behavior

When using a non-button element with `nativeButton={false}`, `role="button"` is always applied.

I realise that this is a helper for accessibility, but I would expect a link to retain it's accessible role.

## Expected behavior

The component already checks the element type, and could perhaps only apply the role to non-links?

## Reproducible example

https://ui.shadcn.com/docs/components/base/button#link-1

## Base UI version

1.1.0

## Which browser are you using?

Chrome (and others)

## Which OS are you using?

macOs

## Which assistive tech are you using (if applicable)?

N/A

## Additional context

See the Radix example:
https://ui.shadcn.com/docs/components/radix/button#link-1

atomiks:
Button is intended to be used for real buttons and always applies button semantics when `nativeButton={false}` (it will set `role="button"` for any non-`<button>` element). The problem is shadcn's Base UI docs currently show Buttons being able to render `<a>`, so we should make the docs guidance explicit and ideally coordinate an update there as well. (I'm not sure how shadcn wants to support `focusableWhenDisabled` or `<div>` buttons, as it could reuse the Radix button logic for Base UI given Radix doesn't have a Button component.)

Here's why it doesn't work with our Button, which we intended to be used for button role-based accessibility logic (`role="button"` elements), rather than "appearance" based elements (e.g. links that look like a button):

We can't reliably skip `role="button"` for `<a>` in SSR because the actual tag is only known after mount. That would require a post-hydration fix and would render incorrect semantics in the server HTML.

During render, the props are computed like so, depending on `nativeButton`:
- `<button type="button" disabled="..." ...>` (`nativeButton=true`)
- `<div role="button" aria-disabled="..." ...>` (`nativeButton=false`)

`<a>`s also probably _shouldn't_ ever be disabled unlike `<button>`s, so some props (like `disabled`) are conflicting to begin with.

If you want “a link that looks like a button”, we recommend a separate `LinkButton` component that renders an `<a>` and shares the same styles/variants as Button, rather than reusing Button itself. (see #3423)

mrmckeb:
Thanks, I understand the rationale and this is something we can certainly do on our side by sharing styles.

Should I create an upstream issue for `shadcn/ui` on this one? They may need to provide different guidance.

atomiks:
[...]

PR #4163, opened 2026-02-22
[docs] Document button and link semantic conflict

CiscoFran10:
## Summary

image

- Added a "Rendering links as buttons" example to the Button docs explaining
  that `role="button"` is applied even with `nativeButton={false}`, and
  recommending applying button styles directly to `<a>` elements instead.
- Added a "Conditional default props" section to the useRender docs covering
  how to handle element-specific defaults (e.g. `type="button"`) when the
  `render` prop changes the underlying element.
- Added a usage guideline to the Button docs about not using Button for links.

## Motivation

A common pattern carried over from Radix UI's `asChild` is rendering a Button
as an `<a>`. This applies `role="button"` to the link, causing screen readers
to announce it as a button instead of a link.

- https://github.com/mui/base-ui/issues/4021

This has come up across multiple projects:

- https://github.com/shadcn-ui/ui/pull/9640

- https://basecn.dev/docs/get-started/migrating-from-radix-ui
- https://ui.shadcn.com/docs/components/base/button#link-1
- https://pure.kam-ui.com/docs/components/button
- https://www.roiui.com/docs/ui/button#external-link

*Coss uses useRender instead of the Button component.*:
- https://coss.com/ui/docs/components/button#link


docs/src/app/(docs)/react/utils/use-render/page.mdx
```diff
@@ -156,6 +156,30 @@ function Button({ render, ...props }) {
...
+import { useRender } from '@base-ui/react/use-render';
+import { mergeProps } from '@base-ui/react/merge-props';
+
+interface ButtonProps extends useRender.ComponentProps<'button'> {}
+
+function Button({ render, ...props }: ButtonProps) {
+  const defaultProps: useRender.ElementProps<'button'> = {
+    className: styles.Button,
+    type: render ? undefined : 'button',
+  };
```
atomiks: `render` may still render a `<button>` though (it can't be known at render time what's going to be rendered, only in an effect) - that's why the `nativeButton` prop exists

docs/src/app/(docs)/react/components/button/page.mdx
```diff
@@ -40,6 +41,26 @@ import { Button } from '@base-ui/react/button';
 </Button>;
 ```
 
+### Rendering links as buttons
+
+The Button component enforces button semantics. Even with `nativeButton={false}`, it applies `role="button"` and keyboard interaction handlers to the rendered element. Rendering it as an `<a>` causes screen readers to announce the link as a button:
```
atomiks: The `even` wording could be changed a bit, maybe say that `nativeButton={false}` signals that the tag isn't a `<button>` but must still be a tag that can have button semantics applied to it still (`role="button"`), which doesn't apply to links/`<a>` 

This PR will warn also: https://github.com/mui/base-ui/pull/3423

docs/src/app/(docs)/react/components/button/page.mdx
```diff
@@ -40,6 +41,26 @@ import { Button } from '@base-ui/react/button';
...
+```
+
+For links that are visually styled as buttons, apply button styles directly to the `<a>` element:
+
+```tsx title="Correct"
+import styles from './button.module.css';
+
+<a href="/contact" className={styles.Button}>
+  Contact
+</a>;
```
atomiks: Maybe mention that a separate component should be encouraged for LinkButtons, like `Menu.LinkItem`

CiscoFran10: Better ?

image

docs/src/app/(docs)/react/components/button/page.mdx
CiscoFran10: Is this better ??

> The Button component enforces button semantics. nativeButton={false} signals that the rendered tag is not a
>`<button>`, but it must still be a tag that can receive button semantics (role="button", keyboard interaction 
> handlers). Links (`<a>`) have their own semantics and should not be rendered as buttons:

docs/src/app/(docs)/react/utils/use-render/page.mdx
CiscoFran10: Yeah, it can still render a `<button>`, but I think when a dev opts into using `render`, they're taking responsibility for which props are appropriate for the rendered element. It's safer to omit `type="button"` when render is used rather than risk passing it to a `<Link>` or `<a>` silently.

My thinking is that devs coming from Radix are so used to the asChild/Slot pattern that their first instinct when migrating will be to reach for `render` to create a polymorphic button. This section is meant as a heads-up to be careful about which default props they're passing when doing that. If the dev doesn't want to deal with these tradeoffs, they can opt out of polymorphism entirely and use separate components (e.g., Button and LinkButton) with correct defaults for each.

Maybe it would make sense to link to the actual Button component docs from here? That might be a better place to explain these tradeoffs in more depth.

PS: Sorry for the essay 😄.

docs/src/app/(docs)/react/components/button/page.mdx
atomiks: Yep, looks good!

docs/src/app/(docs)/react/components/button/page.mdx
atomiks: Yep, looks good as well

docs/src/app/(docs)/react/utils/use-render/page.mdx
atomiks: Yeah I think linking Button is okay — but I also think explaining the `nativeButton` prop here could work instead of the code block example (i.e. props need to valid at render time, during SSR.) So for components that have props only valid on the default tag, another prop needs to be specified to signal at render time what's being rendered.

You could also emphasize that the render prop preferrably renders the _same_ tag component in most cases and that its purpose is to delegate/compose event handlers and behavioral props, not necessarily polymorphism which requires more care

docs/src/app/(docs)/react/components/button/page.mdx
colmtuite: What do you think about changing this example to `MarketingLink` or something?

Part of me thinks this is fine as-is, since this is something people are going to do regardless, and it's better to at least get the semantics/behaviour right.

The other part of me thinks making links look like buttons and vice versa is just a bad idea, outside of some very universal patterns like a marketing site link (that looks like a button). In a web app, when I see something that looks like a button, I expect it to open a dialog or something, and it's just a bad UX when it navigates to a new screen. I always have to look at the bottom left corner of the screen to see if this thing is a link or not. If possible, I want to avoid encouraging this crap in Base UI docs.

docs/src/app/(docs)/react/utils/use-render/page.mdx
CiscoFran10: Not sure if this is the correct wording ?

image

atomiks: Yeah, looks great. Can add "can't know what element `render` will produce **at render time and before hydration**"

CiscoFran10: > You could also emphasize that the render prop preferrably renders the same tag component in most cases and that its purpose is to delegate/compose event handlers and behavioral props, not necessarily polymorphism which requires more care

Doesnt this make the "Migrating from Radix UI" section be misleading to devs, since its showing exactly what we are gonna recommend not to do ??

image

docs/src/app/(docs)/react/components/button/page.mdx
atomiks: My only concern is the `Incorrect` title is not super obvious when skimming. Ideally this would be colored as an error, but there aren't any styles in the docs. So try adding `Incorrect example` as the Button's children text instead of Contact (and `example` as part of the code block `title="..."`) _or_ possibly omit the wrong example as a code block entirely.

docs/src/app/(docs)/react/utils/use-render/page.mdx
CiscoFran10: > Yeah, looks great. Can add "can't know what element `render` will produce **at render time and before hydration**"

done.

atomiks: Ahh yeah, that code block is wrong and needs to be updated 🤦 !

docs/src/app/(docs)/react/components/button/page.mdx
CiscoFran10: Removed the incorrect code block. its cleaner without it.

image

docs/src/app/(docs)/react/components/button/page.mdx
CiscoFran10: Yeah its not great UX but unfortunately it's super common, pretty much every component library out there has some version of a "Button as link" pattern, and devs will actively look for how to do it in Base UI. I'd rather show the correct way to handle it than have them reach for <Button render={<a />}> on their own.

the options we have are:

- **Rename to MarketingLink** : make it obvious that this is a narrow use case and not a general-purpose pattern.

- **Keep LinkButton but add a note**: something like "This pattern is common for marketing CTAs but should be avoided in web apps where users expect buttons and links to behave differently" (kinda of opinionated).

- **Drop the component wrapper entirely**:  just show a plain <a className={styles.Button}> without encouraging a reusable abstraction

CiscoFran10: we could just do this instead, no `LinkButton` example, just a guideline:

image

atomiks: Yeah, that works well

CiscoFran10: done.

docs/src/app/(docs)/react/utils/use-render/page.mdx
CiscoFran10: Updated to show a custom button being rendered instead of `<a>`

image

PR #4165, opened 2026-02-23
[tabs] Re-render indicator position on tab resize

### LINKED ISSUE #4156: [tabs] `Tabs.Indicator` position not updated when sibling tab resizes in a scrollable list
# Bug report

## Current behavior

`Tabs.Indicator` CSS variables (`--active-tab-left`, etc.) are not recalculated when a non-selected sibling tab changes size, if the tab list is scrollable and its container dimensions are fixed.

The indicator stays at its previous position instead of following the selected tab.

## Expected behavior

The indicator should reposition to stay aligned with the selected tab whenever any tab in the list changes size.

## Reproducible example

https://stackblitz.com/edit/nhyjse4x?file=src%2FApp.tsx

https://github.com/user-attachments/assets/4bc6af1f-95a1-4a6c-87c8-4feebf37bee2

## Base UI version

1.2.0

Chrome

## Which OS are you using?

Mac OS

## Additional context

The `ResizeObserver` in `TabsIndicator` only observes `tabsListElement`:

https://github.com/mui/base-ui/blob/0d7791f559724c9b54496f7fd9da227155b53ab9/packages/react/src/tabs/indicator/TabsIndicator.tsx#L51-L53

When the list container has a fixed width (constrained by a parent) and is scrollable, a sibling tab resizing doesn't change the container's dimensions, so the observer never fires and the indicator position is never recalculated.

This does not reproduce when the list can freely grow with its content, because in that case the list element itself resizes and the observer fires as expected.

atomiks:

In fixed layouts, observing the dimensions of the tabs list element isn't enough: each tab needs to also be observed.

Fixes #4156 (https://stackblitz.com/edit/nhyjse4x-irvdzugx?file=src%2FApp.tsx)


mj12albert (approved):
Looks good ~

PR #4167, opened 2026-02-23
[select][combobox][slider] Add `Label` parts

atomiks:

The `Label` part is added for components that render a `<div>` label instead of using `<label>`. This matches Progress and Meter, and removes the need to use `Field`.

Forms: https://deploy-preview-4167--base-ui.netlify.app/react/handbook/forms#naming-form-controls

### DIFF
--- a/docs/src/app/(docs)/react/handbook/forms/page.mdx
+++ b/docs/src/app/(docs)/react/handbook/forms/page.mdx
@@ -13,11 +13,7 @@ import { DemoBaseUIForm } from './demos/hero';
 
 Form controls must have an accessible name in order to be recognized by assistive technologies. Use the label strategy below for each control type.
 
-### Labeling by control
-
-Choose the labeling pattern based on which element acts as the form control.
-
-#### Input controls
+### Input controls
 
 Use `<Field.Label>` or a native `<label>` to label the following controls:
 
@@ -44,14 +40,14 @@ import { Switch } from '@base-ui/react/switch';
 </Field.Root>;
 ```
 
-#### Trigger-based controls
+### Trigger-based controls
 
 - `Combobox` (input inside popup): use `<Combobox.Label>`.
 - `Select`: use `<Select.Label>`.
 - `Slider`: use `<Slider.Label>`. For multi-thumb sliders, also add an `aria-label` on each
   `<Slider.Thumb>` to distinguish the thumbs.
 
-#### Fallback
+### Fallback
 
 If no visible label is rendered, provide `aria-label` on the actual form control.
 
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review

### Overview
This patch adds dedicated `Label` parts for `Select`, `Combobox`, and `Slider`, and updates the shared label plumbing so button-based controls can be labeled without routing everything through `Field.Label`. It also updates the public docs and handbook examples so the guidance matches the final standalone labeling APIs, including the grouped-slider guidance.

### Findings

#### 1. ℹ️ [Informational] Focused SSR render benchmark remains effectively flat

**Impact**: The extra label registration and `aria-labelledby` resolution work in the shared paths does not look like a meaningful render-time regression, so performance is not a blocker here.

**Evidence**: The hot paths touched by this branch are in `packages/react/src/labelable-provider/useLabel.ts`, `packages/react/src/select/trigger/SelectTrigger.tsx`, `packages/react/src/combobox/trigger/ComboboxTrigger.tsx`, and `packages/react/src/slider/root/SliderRoot.tsx`. I reran a focused `renderToString` benchmark on the latest `master` versus the current PR head, rendering 50 each of `Field.Label + Select`, `Field.Label + Combobox`, and `Field.Label + Slider` per sample.

| `master` (`e90b027`) | 13.53 | 13.12 | 16.82 | 30 |
| PR head (`e87fdc4`) | 13.25 | 12.88 | 16.94 | 30 |

**Recommendation**: No perf-specific change is needed.

### Confidence: 4/5
High confidence based on a full `master...HEAD` pass over both runtime code and docs, fresh targeted JSDOM coverage for `FieldLabel`, `ComboboxLabel`, `ComboboxRoot`, `SelectLabel`, `SelectRoot`, `SliderLabel`, and `SliderRoot`, `pnpm docs:validate`, and the focused SSR render benchmark above. Residual risk is mainly in unusual label compositions that are not yet covered by explicit tests.

### Notes
- Docs were included in this pass: the Select, Combobox, and Slider component pages, the forms handbook page, the generated handbook index, and the updated demos/examples were all rechecked against the shipped API.
- Latest delta since the previous review: the forms handbook heading structure was flattened, the implicit-label guidance now sits with the naming guidance instead of the description section, and the generated handbook outline was updated to match.

packages/react/src/combobox/root/ComboboxRoot.test.tsx
```diff
@@ -4738,12 +4755,12 @@ describe('<Combobox.Root />', () => {
         </Field.Root>,
       );
 
-      const label = screen.getByTestId<HTMLLabelElement>('label');
+      const label = screen.getByTestId<HTMLDivElement>('label');
```
michaldudak: Doesn't Field.Label render native label by default?

atomiks: Yes, this test had the wrong type there

PR #4179, opened 2026-02-24
[combobox] Fix combobox attributes during SSR

atomiks:

Ensures the attributes are present before hydration. `<Combobox.Trigger>` has an existing mismatch if it's not inside the popup, but the issue is minor and can be worked around externally


atomiks:
## Codex Review

### Overview
This PR ensures Combobox ARIA attributes are present during SSR and initial hydration when the input ref is not yet available. It also adds targeted server-rendering tests to lock in expected combobox attributes for input-based and trigger-based configurations.

### Findings (None)

### Confidence: 4/5
The behavioral change is small, directly addresses the SSR ref-timing gap, and is backed by focused regression tests on the relevant attribute paths.

### Notes
- `packages/react/src/combobox/root/AriaCombobox.tsx:969` introduces `shouldTreatAsInput` so SSR applies combobox ARIA attributes before the input element ref is populated.
- `packages/react/src/combobox/root/ComboboxRoot.test.tsx:96` adds SSR regression coverage for default input mode and trigger-based mode.

PR #4185, opened 2026-02-24
[docs-infra][test] Remove Tailwind classes from docs shell and tests

aarongarciah:
No visual impact expected. This refactoring is the first step towards a more maintainable CSS architecture.

Main changes included:
- Introduce a top-level `/css` folder where all the styles related to the public website will live.
  - A new `index.css` stylesheet is now used in:
    - Docs shell (`/docs/src/app/(docs)/layout.tsx`)
    - Private (experiments) (`/docs/src/app/(private)/layout.tsx`)
    - e2e tests (`test/e2e/main.tsx`)
    - Regression tests (`test/regressions/main.tsx`)
  - The website part (`/docs/src/app/(website)/layout.tsx`) still uses its own stylesheet, although it shares the same utility classes. In the future, the three parts (docs, private, website) will use the exact same stylesheet, fonts, etc.
- Remove Tailwind classes from:
  - Docs shell
    - `/docs/src/app/(docs)`
    - `/docs/src/components`
    - `/docs/src/mdx-components.tsx`
  - Experiments
    - `/docs/src/app/(private)/experiments`
  - Tests
    - `/test/e2e/fixtures`
    - `/packages/react/test/floating-ui-test`
- Change the Tailwind config to a safelist approach, so only Tailwind classes found in Tailwind demos (`/docs/src/app/(docs)/react/**/demos/**/*.{ts,tsx}`) and the private playground (`/docs/src/app/(private)/playground/**/*.{ts,tsx}`) are generated.
- Move experiment specific components under `/docs/app/(private)/experiments/_components`.

Note: Tailwind CSS vars are still used in "Tailwind-free" folders like `/docs/src/components`, especially for typography. We'll remove them as we go while we introduce visual changes to the docs using our own custom CSS vars.

### DIFF
--- a/docs/src/components/Search/SearchBar.css
+++ b/docs/src/components/Search/SearchBar.css
@@ -216,7 +216,6 @@
     gap: 0.5rem;
     height: 2rem;
     border-radius: var(--radius-md);
-    background-color: var(--color-popup);
     padding-inline: 0.75rem;
   }
 


docs/src/css/index.css
```diff
@@ -0,0 +1,302 @@
...
+/* We defined grays, red, and blue in the theme used for demos. */
+/* Uses gray scale that is roughly compatible with Tailwind's grays. */
+/* This is injected in demos when opening in a sandbox like StackBlitz. */
+@import '../demo-data/theme/css-modules/theme.css' layer(theme);
+
+/* ═══ Docs-only imports ═══ */
+
+@import './fonts/index.css';
+@import './custom-media.css';
+@import './mdx.css';
```
dav-is: Maybe we should import this in `mdx-components.tsx`

docs/src/css/index.css
```diff
@@ -0,0 +1,302 @@
...
+@import '../components/Select.css';
+@import '../components/SideNav.css';
+@import '../components/SkipNav.css';
+@import '../components/Subtitle/Subtitle.css';
+@import '../components/Subtitle/MarkdownLink.css';
+@import '../components/Accordion.css';
+@import '../components/Table.css';
+@import '../components/TableCode.css';
+@import '../components/DescriptionList.css';
+@import '../components/ReferenceTable/ReferenceTable.css';
```
dav-is, line 70: Could we import these directly within the component itself? That way it can code split CSS based on what components are being used. If we want to lazy load a large component, we wouldn't be able to lazy load the CSS if it's in this monolithic CSS file .Since the experiments page doesn't use any of these components, and we might eventually have benchmark pages, I think we should try to minimize the global styles we have.

 It's also helpful to be able to click on the CSS import from within the component .tsx to quickly open the CSS.

Also, curious if you would be opposed to using CSS modules for components, which would help isolate Demos from accidentally colliding with global styles as well as allowing us to create better tooling around missing classnames being referenced

docs/src/css/mdx-components.css
```diff
@@ -0,0 +1,95 @@
+@layer components {
```
dav-is, line 1: Could this file live at `/docs/src/mdx-components.css` right next to `/docs/src/mdx-components.tsx`?

docs/src/app/(private)/layout.css
```diff
@@ -1,26 +1,20 @@
...
 
-html,
-body {
-  padding: 0;
-  margin: 0;
+  html,
+  body {
+    padding: 0;
+    margin: 0;
+  }
```
dav-is, line 19: Ideally, nested layout shouldn't be targeting `html` or `body`. Can we apply these changes in the root layout or apply them to a `.root` wrapper?

This would be a blocker to creating multiple 404 pages that differ by subdirectory.

docs/src/css/index.css
aarongarciah: > Could we import these directly within the component itself?

That's the idea, but not in this PR. I tried it but things started to break since some styles are poorly architected and changing the order in which they're loaded affected the ui.

> Also, curious if you would be opposed to using CSS modules for components,

That's the goal as well, to use CSS modules for all components 👍

I left this out of this PR because this was mostly about "removing" TW. But expect these changes in future PRs.

docs/src/css/index.css
aarongarciah: Updated.

docs/src/css/mdx-components.css
aarongarciah: Updated.

dav-is:
There seems to be a visual regression in the search bar with the background:

image

docs/src/app/(private)/layout.css
aarongarciah: I agree, but this is not a nested layout as each one of `(docs)`, `(website)`, and `(private)` have their own top-level layout. It's also a bit outside of the scope of this PR so I'd prefer to not modify it. In a future PR these base styles will be merged for `(docs)` and `(website)`.

colmtuite:
The main page heading is sized incorrectly now. Looks good aside from that.

aarongarciah:
> There seems to be a visual regression in the search bar with the background

@dav-is fixed

PR #3927, opened 2026-02-02
[docs] Migrate from Tailwind v3 to v4

battaglr:
Fixes #2482

### Renamed v4 utilities

- `outline-none` → `outline-hidden`
- `rounded` → `rounded-sm`
- `rounded-sm` → `rounded-xs`

> In v4, several utilities were renamed to make them more consistent and predictable.
> _See [relevant docs](https://tailwindcss.com/docs/upgrade-guide#renamed-utilities)._

### Removed redundant `outline` class

```diff
- outline outline-1
+ outline-1
```

> In v4, the `outline` utility now sets `outline-width: 1px` by default (in v3 it set `outline-style: solid`). When you specify an explicit width like `outline-1`, the base `outline` class becomes redundant.  
> _See [relevant docs](https://tailwindcss.com/docs/upgrade-guide#renamed-outline-utility)._

### Files changed

- `docs/src/app/(docs)/react/*`
- `docs/src/app/(private)/experiments/*`
- `examples/tanstack-start-tailwind-css/*`
- `test/e2e/fixtures/*`

### Relevant issues

- #1185
- #2435
- #2393

PR #4198, opened 2026-02-26
[navigation menu] Fix nested inline menu support

atomiks:

This was previously a hidden experiment, and the popup auto resize logic failed to handle content insertions after it was refactored to the shared `usePopupAutoResize` hook

Preview: https://deploy-preview-4198--base-ui.netlify.app/react/components/navigation-menu#nested-inline-submenus

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts
+++ b/packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts
@@ -16,7 +16,11 @@ import {
   useHoverInteractionSharedState,
 } from './useHoverInteractionSharedState';
 import type { HandleClose, HandleCloseContextBase } from './useHoverShared';
-import { getDelay, getRestMs, isClickLikeOpenEvent as isClickLikeOpenEventShared } from './useHoverShared';
+import {
+  getDelay,
+  getRestMs,
+  isClickLikeOpenEvent as isClickLikeOpenEventShared,
+} from './useHoverShared';
 import { FloatingUIOpenChangeDetails, HTMLProps } from '../../utils/types';
 
 export interface UseHoverReferenceInteractionProps {
--- a/packages/react/src/floating-ui-react/hooks/useHoverShared.ts
+++ b/packages/react/src/floating-ui-react/hooks/useHoverShared.ts
@@ -61,9 +61,6 @@ export function getRestMs(value: number | (() => number)) {
   return value;
 }
 
-export function isClickLikeOpenEvent(
-  openEventType: string | undefined,
-  interactedInside: boolean,
-) {
+export function isClickLikeOpenEvent(openEventType: string | undefined, interactedInside: boolean) {
   return interactedInside || openEventType === 'click' || openEventType === 'mousedown';
 }
--- a/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
+++ b/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
@@ -601,9 +601,7 @@ function TestDeeplyNestedNavigationMenu() {
   );
 }
 
-function TestNavigationMenuWithNestedPopup(props: {
-  children: React.ReactNode;
-}) {
+function TestNavigationMenuWithNestedPopup(props: { children: React.ReactNode }) {
   const { children } = props;
   return (
     <NavigationMenu.Root>
@@ -1413,7 +1411,7 @@ describe('<NavigationMenu.Root />', () => {
     });
 
     it('keeps parent menu open when hovering inline nested triggers without defaultValue', async () => {
-        await render(<TestInlineNestedNavigationMenu nestedDefaultValue={null} />);
+      await render(<TestInlineNestedNavigationMenu nestedDefaultValue={null} />);
       const trigger1 = screen.getByTestId('trigger-1');
 
       fireEvent.mouseEnter(trigger1);


docs/src/app/(docs)/react/components/navigation-menu/demos/nested-inline/css-modules/index.tsx
```diff
@@ -0,0 +1,261 @@
...
+            Product
+            <NavigationMenu.Icon className={styles.Icon}>
+              <ChevronDownIcon />
+            </NavigationMenu.Icon>
+          </NavigationMenu.Trigger>
+          <NavigationMenu.Content className={`${styles.Content} ${styles.ProductContent}`}>
+            <NavigationMenu.Root
+              className={styles.SubmenuRoot}
+              orientation="vertical"
+              defaultValue="developers"
```
CiscoFran10, line 25: Not passing `defaultValue ` to nested `<NavigationMenu.Root/>` causes the entire menu to close when hovering over the menu item triggers.

Something to do with `safePolygon `on line 294, `useHover` hook.

```ts
 handleClose: safePolygon({ blockPointerEvents: pointerType !== 'touch' }),
```

Maybe check if there is a `positionerElement` before adding the safePolygon ??

```ts
 handleClose: positionerElement
      ? safePolygon({ blockPointerEvents: pointerType !== 'touch' })
      : null,
```

docs/src/app/(docs)/react/components/navigation-menu/demos/nested-inline/tailwind/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+          style={{
+            ['--duration' as string]: '0.35s',
+            ['--easing' as string]: 'cubic-bezier(0.22, 1, 0.36, 1)',
+          }}
+        >
+          <NavigationMenu.Popup className="data-[ending-style]:easing-[ease] relative h-[var(--popup-height)] origin-[var(--transform-origin)] rounded-lg bg-[canvas] text-gray-900 shadow-lg shadow-gray-200 outline-1 outline-gray-200 transition-[opacity,transform,width,height,scale,translate] duration-[var(--duration)] ease-[var(--easing)] data-[ending-style]:scale-90 data-[ending-style]:opacity-0 data-[ending-style]:duration-150 data-[starting-style]:scale-90 data-[starting-style]:opacity-0 w-[var(--popup-width)] dark:shadow-none dark:-outline-offset-1 dark:outline-gray-300">
```
CiscoFran10: Note: Tailwind v4 syntax, won't work on demos. consider upgrading stackblitz demos to v4.
>outline-1 outline-gray-200

docs/src/app/(docs)/react/components/navigation-menu/demos/nested-inline/tailwind/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+const triggerClassName =
+  'box-border flex items-center justify-center gap-1.5 h-10 ' +
+  'px-2 sm:px-3.5 m-0 rounded-md bg-gray-50 text-gray-900 font-medium ' +
+  'text-[0.925rem] sm:text-base leading-6 select-none no-underline ' +
+  'hover:bg-gray-100 active:bg-gray-100 data-[popup-open]:bg-gray-100 ' +
+  'focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-blue-800 focus-visible:relative';
+
+const sharedContentClassName =
+  'h-full w-[calc(100vw_-_40px)] p-5 ' +
```
CiscoFran10: `p-5` class is not being overridden by the `p-0` on `productContentClassName` and `guidesContentClassName`. for some reason the compiled css demo is applying `p-5` last. consider removing since `p-5` is supposed be overridden anyway.

docs/src/app/(docs)/react/components/navigation-menu/demos/nested-inline/tailwind/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+  'data-[ending-style]:data-[activation-direction=left]:translate-x-[2rem] ' +
+  'data-[ending-style]:data-[activation-direction=right]:translate-x-[-2rem]';
+
+const productContentClassName = `${sharedContentClassName} min-[700px]:max-w-[675px] p-0`;
+
+const guidesContentClassName = `${sharedContentClassName} min-[700px]:max-w-[500px] p-0`;
+
+const submenuTriggerClassName =
+  'box-border m-0 flex w-full min-w-[10rem] flex-col items-start gap-0.5 rounded-lg ' +
+  'bg-transparent px-3 py-2.5 text-left text-inherit ' +
```
CiscoFran10, line 198: `bg-transparent` should be `bg-canvas`, same as the css-modules version.

Image

CiscoFran10 (changes requested):

docs/src/app/(docs)/react/components/navigation-menu/demos/nested-inline/tailwind/index.tsx
atomiks: Good point, in #3927 this should've been updated as well.

CiscoFran10:
Also consider updating the docs for `NavigationMenu.Root` so that `delay` and `closeDelay` are explicitly described as controlling the navigation popup, not just navigation menu. Right now it’s easy to assume they affect every item or every hover (e.g. switching between nested inline items).

context: I tried adding a delay to the nested menu so that it would not open immediately, to prevent from accidentallly opening another menu when moving the cursor in the direction of the menu content.

this cursor path sometimes accidentally opens the "Design systems" menu item.

image


ex:
```ts
  /**
   * How long to wait before the navigation popup opens. Specified in milliseconds.
   * @default 50
   */
  delay?: number | undefined;
  /**
   * How long to wait before the navigation popup closes. Specified in milliseconds.
   * @default 50
   */
  closeDelay?: number | undefined;
```

image

CiscoFran10:
> context: I tried adding a delay to the nested menu so that it would not open immediately, to prevent from accidentallly opening another menu when moving the cursor in the direction of the menu content.
> 
> this cursor path sometimes accidentally opens the "Design systems" menu item.

this one is even harder 😫 :

image

atomiks:
> context: I tried adding a delay to the nested menu so that it would not open immediately, to prevent from accidentallly opening another menu when moving the cursor in the direction of the menu content.

That's what `blockPointerEvents` in safePolygon tries to fix, but it's not working in nested menus or breaks something as you pointed out above

atomiks:
## Codex Review

### Overview
This branch adds the nested inline Navigation Menu docs demo, makes inline submenu hover use the same pointer-travel protection as positioned submenus, and keeps Navigation Menu popup sizing in the older trigger-owned flow. It also adds broad regression coverage around nested inline traversal, hover stickiness, delayed nested default content, and popup resizing.

### Findings
#### 1. 🟡 Non-blocking Test fixture duplication in `NavigationMenuRoot.test.tsx`

**Impact**: The coverage is much stronger now, but the test file still carries several large Navigation Menu fixture trees that differ only in a few nested-content details. That makes future behavior changes harder to update consistently and increases the chance of one fixture drifting from the others.

**Evidence**: `packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx:94`, `:151`, `:226`, `:282`, and `:351` each build another top-level Navigation Menu shell with only small changes in nested content, tab-flow setup, or dynamic sizing behavior.

**Recommendation**: Keep the current assertions, but collapse more of the repeated root/list/portal shell into configurable helpers and pass only the nested-content differences as props. The recent dialog/popover fixture cleanup is a good direction to continue.

#### 2. ℹ️ Informational Focused Chromium timing check does not show a hot-path slowdown

**Impact**: I did not find evidence here that the added hover and resize logic makes the tested trigger path slower in a meaningful way.

**Evidence**: I compared the latest `master` tip with this PR head using three Chromium runs of `pnpm test:chromium NavigationMenuTrigger --no-watch`, using Vitest’s reported duration from each run.

| Ref | Mean (s) | p50 (s) | p95 (s) | n |
| `master` | 0.82 | 0.76 | 0.98 | 3 |
| PR head | 1.46 | 1.46 | 1.56 | 3 |

**Recommendation**: No immediate action from this check alone. If maintainers want more confidence, the next step would be a browser-level interaction benchmark on the docs demo rather than test-suite timing.

### Confidence: 4/5
High confidence based on a full `master...HEAD` pass through the hover, safe-polygon, Navigation Menu resize, docs, and regression-test changes. I also reran focused Navigation Menu JSDOM and Chromium suites, TypeScript, and a head-versus-master Chromium timing check, but there is still some residual risk around browser-specific interaction feel that automated coverage cannot fully prove.

### Notes
- Latest changes since the previous review simplify the shared hover click-like predicate and trim some duplicated Navigation Menu test helpers without changing coverage.
- The branch also rebases cleanly onto the current `master` tip.

PR #4199, opened 2026-02-26
[performance] Optimize useHover hooks and safePolygon hot paths

atomiks:

Purely perf-related change for bundle size + runtime cost

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
+++ b/packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
@@ -17,7 +17,7 @@ import {
   isInteractiveElement,
   useHoverInteractionSharedState,
 } from './useHoverInteractionSharedState';
-import { getCloseDelay } from './useHoverShared';
+import { getDelay } from './useHoverShared';
 
 export type UseHoverFloatingInteractionProps = {
   /**
@@ -74,7 +74,7 @@ export function useHoverFloatingInteraction(
 
   const closeWithDelay = React.useCallback(
     (event: MouseEvent) => {
-      const closeDelay = getCloseDelay(closeDelayProp, instance.pointerType);
+      const closeDelay = getDelay(closeDelayProp, 'close', instance.pointerType);
       const close = () => {
         store.setOpen(false, createChangeEventDetails(REASONS.triggerHover, event));
         tree?.events.emit('floating.closed', event);
--- a/packages/react/src/floating-ui-react/hooks/useHoverShared.ts
+++ b/packages/react/src/floating-ui-react/hooks/useHoverShared.ts
@@ -48,13 +48,6 @@ export function getDelay(
   return result?.[prop];
 }
 
-export function getCloseDelay(
-  value: number | (() => number) | undefined,
-  pointerType?: PointerEvent['pointerType'],
-) {
-  return resolveValue(value, pointerType);
-}
-
 export function getRestMs(value: number | (() => number)) {
   if (typeof value === 'function') {
     return value();


packages/react/src/floating-ui-react/safePolygon.ts
```diff
@@ -127,31 +152,48 @@ export function safePolygon(options: SafePolygonOptions = {}) {
...
 
-      // If any nested child is open, abort.
-      if (
-        tree &&
-        getNodeChildren(tree.nodesRef.current, nodeId).some(({ context }) => context?.open)
-      ) {
-        return undefined;
+      let hasCheckedOpenChild = false;
+      let hasOpenChild = false;
+      function hasOpenChildNode() {
```
flaviendelangle, line 168: Claude Code review:


### Potential bug: `hasOpenChildNode` semantic change

The old code checked whether any child had an **open** context:

```ts
// OLD
getNodeChildren(tree.nodesRef.current, nodeId).some(({ context }) => context?.open)
```

The new code only checks whether children **exist at all**:

```ts
// NEW
hasOpenChild = getNodeChildren(tree.nodesRef.current, nodeId).length > 0;
```

This is a semantic change: "has an open child" vs "has any child". If closed children remain in the tree (which is what the test `'closes after intent timeout when no nested child is open'` sets up), the new code would incorrectly prevent closing.

The test passes in CI, which suggests either `FloatingTreeStore.nodesRef` / `getNodeChildren` behaves in a way that makes this equivalent in practice, or the code path doesn't reach `closeIfNoOpenChild()` for the specific test coordinates. Either way, the intent is clearly to check for **open** children, and the implementation should match:

```ts
// Suggested fix
hasOpenChild = getNodeChildren(tree.nodesRef.current, nodeId).some(
  ({ context }) => context?.open,
);
```

atomiks: `getNodeChildren` already defaults to `onlyOpenChildren` (as a third arg), so that .some was redundant

atomiks:
## Codex Review

### Overview
This PR refactors hover and safePolygon hot paths, extracts shared hover delay helpers, and adds targeted tests around wrapper-trigger handling and safePolygon geometry. I reviewed the full `upstream/master...head` diff and the latest commit on this branch.

### Findings

#### 1. ℹ️ [Informational] safePolygon hot path is faster than master in the focused micro-benchmark

**Impact**: The main performance-sensitive path reviewed here shows a measurable win relative to `master`, so the scalar geometry rewrite does not introduce a runtime regression in this check.

**Evidence**: `packages/react/src/floating-ui-react/safePolygon.ts` plus the new safePolygon tests (`packages/react/src/floating-ui-react/safePolygon.test.ts`) were benchmarked with the same harness on both refs (`120` samples, `2500` events/sample, `safePolygon({ requireIntent: false })` hot path).

| Ref | Mean (us/event) | p50 (us/event) | p95 (us/event) | Samples |
|---|---:|---:|---:|---:|
| `upstream/master` (`0824de08c`) | 0.1393 | 0.1136 | 0.2006 | 120 |
| PR head (`87abf70dd`) | 0.0755 | 0.0540 | 0.1600 | 120 |

**Recommendation**: No change required for this path based on current evidence.

#### 2. ℹ️ [Informational] DRY simplification in hover delay utilities is clean

**Impact**: Delay resolution and pointer-type gating are now centralized, which lowers maintenance surface without changing public behavior.

**Evidence**: `packages/react/src/floating-ui-react/hooks/useHoverShared.ts` centralizes resolution in `resolveValue`, and `packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts` now reuses `getDelay(..., 'close', ...)` instead of a separate close-only helper.

**Recommendation**: Keep this structure as-is.

### Confidence: 4/5
High confidence from a full-branch diff review, targeted behavioral tests in touched modules, and a direct head-vs-master micro-benchmark on the main hot path.

### Notes
- Review scope covered all files changed in this branch, not just the latest commit.
- CI bundle delta should be read from the latest bot run on current head (`87abf70dd`).

PR #4206, opened 2026-02-26
[popups] Fix nested hoverable popups

dmitri-gb:
This fixes two issues that were not fixed in #3798.

1. **All popups suddenly closing** when entering a child popup, after having briefly moved outside of the parent popup. Or similarly, all popups closing upon entering the child when the parent is `defaultOpen`.

2. **Popups not closing simultaneously** when moving out of a child popup. 

Videos of both issues can be seen [in a comment on the previous PR](https://github.com/mui/base-ui/pull/3798#pullrequestreview-3737196407).

I removed some code that seemed redundant. There was no reason to have two different mouseleave listeners that did very similar things. And the safePolygon "handler" is no longer disposed of so eagerly. It stays "alive" for as long as the popup is open.

I also made changes so that nested popups would continue to work even when no safePolygon "handler" was available (`useHoverReferenceInteraction` can just as well be called without `handleClose: safePolygon()`). The parent-child relationship check is just ~3 lines of code in `safePolygon.ts`, so there's really no need to exclusively rely on the "handler" for this logic.

Lastly, a small side-note — both of these bugs are inherited from `@floating-ui/react` (exact same behaviour is reproduced with simple floating-ui code), and so the older `useHover` hook (which I didn't touch) has these issues too.


atomiks:
The `onScrollMouseLeave` handler wasn't entirely redundant. Here's what it was trying to fix (people did report this as a real issue in Floating UI):

https://github.com/user-attachments/assets/84508fd7-67b5-46ae-a0b5-5cf02b5695ab

It should close once you stop scrolling (does on master, here it doesn't). It does complicate nested popups though.

dmitri-gb:
I see. I added back an invocation of `instance.handler()` that should take care of this issue.

atomiks:
**Findings (Codex)**
1. **[P1] `mousemove` safePolygon handler can survive after popup closes via non-hover paths.**  
`onOpenChangeLocal` clears timers on close but does not clear `instance.handler` / document listener, so closing via `Escape`, outside click, or controlled `open=false` can leave a stale listener active. That stale callback can later call `onClose()` with old context and emit `floating.closed`, causing unexpected follow-up closes.  
Refs: `packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts:139`, `packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts:115`

2. **[P2] Untracked `setTimeout` in nested-close path can fire after teardown.**  
`onNodeClosed` schedules a raw `setTimeout(..., 0)` with no cancellation on cleanup. If the component closes/unmounts before the tick, the callback can still run and emit `floating.closed` from stale state.  (Try using the `Timeout` util / `useTimeout` util)
Ref: `packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts:211`

dmitri-gb:
Those two should be fixed now.

atomiks (approved):
Nice, thanks for the contribution @dmitri-gb! Looks to have fixed all the issues correctly and I couldn't find any regressions

PR #4219, opened 2026-02-27
[docs] Fix Tailwind StackBlitz class pre-injection

atomiks:

Updates StackBlitz Tailwind class pre-injection logic to include classes referenced via `className={identifier}` string constants, not just inline literals.

I noticed the nested Drawer stack example's enter animation was broken on first mount in Tailwind in Stackblitz. This should also fix issues like this: https://github.com/mui/base-ui/pull/3974

### DIFF
--- a/docs/src/utils/demoExportOptions.ts
+++ b/docs/src/utils/demoExportOptions.ts
@@ -30,14 +30,7 @@ const htmlHeadWithCssModulesTheme: ExportConfig['headTemplate'] = () => themeSty
 // Tailwind CSS Setup
 const tailwindSetup = `
 <!-- Check out the Tailwind CSS' installation guide for setting it up: https://tailwindcss.com/docs/installation/framework-guides -->
-<script src="https://cdn.tailwindcss.com"></script>
-<script>
-  tailwind.config = {
-    theme: {
-      extend: {},
-    },
-  }
-</script>`;
+<script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>`;
 const tailwindNote = `
 
 <!-- Inject classes used so that Tailwind loaded from the CDN can pre-render them. -->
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review

### Overview
This patch hardens Tailwind StackBlitz exports in two places: it now resolves the string-expression patterns used in the docs when collecting class names, and it switches the exported runtime to Tailwind v4 so the generated demos understand the same utility set as the docs examples. The result is that the exported Tailwind demos keep the classes needed for their first render, including popup outline styles.

### Findings (None)

### Confidence: 4/5
High confidence based on a full pass over the current PR diff, a second sweep for correctness, accessibility, API design, maintainability, performance, and test coverage, plus direct verification with `pnpm test:jsdom demoExportOptions --no-watch`, file-level eslint/prettier on the touched docs files, and browser-level inspection of the generated StackBlitz form payload. The browser verification confirmed that the exported `project[files][index.html]` now uses `https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4` instead of the old Play CDN script, and that the injected class list includes popup outline classes such as `outline-1 outline-gray-200`.

### Notes
- Re-reviewed at PR head `158105745`; GitHub's current PR diff is still limited to `docs/src/utils/demoExportOptions.ts` and `docs/src/utils/demoExportOptions.test.ts`.
- The browser verification covered Tailwind demos that exercise the risky paths from this patch, including a navigation-menu export path with popup outline styling and a menu export path with inline template-expression classes.

docs/src/utils/demoExportOptions.ts
```diff
@@ -29,19 +30,137 @@ const htmlHeadWithCssModulesTheme: ExportConfig['headTemplate'] = () => themeSty
...
 <!-- Check out the Tailwind CSS' installation guide for setting it up: https://tailwindcss.com/docs/installation/framework-guides -->
-<script src="https://cdn.tailwindcss.com"></script>
-<script>
-  tailwind.config = {
-    theme: {
-      extend: {},
-    },
-  }
-</script>`;
+<script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>`;
```
atomiks, line 33: `outline` styles were broken after the v3 -> v4 demo updates

PR #4220, opened 2026-02-27
[scroll area] Fix focus trap with non-scrollable viewport

### LINKED ISSUE #4205: [scroll-area] non-scrollable scroll-area breaks focus trapping of the dialog
# Bug report

## Current behavior

If you render non-scrollable scroll-area inside of the dialog content it will break dialogs focus trapping.

## Expected behavior

Since you can potentially have dynamic items be rendered in your scroll-area that wont always overflow , the scroll-area should either be focused to preserve dialog focus trapping or do not affect it at all.

## Reproducible example

https://stackblitz.com/edit/tfve4lsf?file=src%2FApp.tsx

## Base UI version

v1.2.0 

Helium(Official Build, Chromium 145.0.7632.109) (arm64)

## Which OS are you using?

Mac OS

## Additional Context
I see that if scroll-area is no longer scrollable then it removes tabIndex={0} . By manually setting tabIndex={0} on the ScrollArea.Viewport it preserves the tab focus order and fixes the bug. Fyi

Not sure if this is expected behaviour but if it is - would be great to mention it or document it somewhere. Thanks !

atomiks:

Fixes #4205 (https://stackblitz.com/edit/tfve4lsf-gffpyxdf?file=src%2FApp.tsx)

Explanation:

ScrollArea starts with `hiddenState = { x: false, y: false }`, so the viewport initially renders with `tabIndex=0`. (fwiw this will change with https://github.com/mui/base-ui/pull/4107)

- **Old behavior**: non-scrollable viewport removed `tabIndex` entirely. In Chromium, when that happens after focus was on the viewport, focus can fall back to `<body>`, which breaks dialog trap-focus flow.
- **New behavior**: non-scrollable viewport gets `tabIndex={-1}`.
Not in tab order. That stability keeps focus inside the dialog, so trap-focus keeps working.


atomiks:
## Codex Review

### Overview
This PR fixes a trap-focus edge case where a non-scrollable `ScrollArea.Viewport` could still participate in keyboard tab order inside a dialog. It updates viewport tab behavior and adds a focused regression test so tab navigation remains trapped within the dialog.

### Findings (None)

### Confidence: 4/5
The fix is narrowly scoped to `tabIndex` behavior and is reinforced by a targeted dialog regression test covering the reported interaction path.

### Notes
- `packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx:371` now sets `tabIndex` to `-1` when both axes are non-scrollable and `0` otherwise.
- `packages/react/src/dialog/root/DialogRoot.test.tsx:1120` adds a regression test for `Dialog.Root` with `modal="trap-focus"` containing a non-scrollable `ScrollArea.Viewport`.

Issue #2940, opened 2025-10-08
[dialog] Can't apply hover styles on `Backdrop`


marcospgp:
It seems that base UI renders an empty div on top of the backdrop so any hover styling is not applied.

atomiks:
`cursor: pointer` is not necessary to use on a backdrop, so this particular style isn't needed

Still, if other hover styles are necessary, you can increase the `z-index` of `Backdrop` and `Popup` to `1`. And ensure keepMounted elements like `Toast.Viewport` are `2`

marcospgp:
isn't the point of base UI that it is a library of unstyled components that users can customize?

why would defining the opacity of the backdrop be a more valid customization than setting what the cursor looks like when hovering it?

what you are suggesting sounds a bit like stepping into z index hell - especially since other components are affected by the fix, such as the toast component you mention

atomiks:
> what you are suggesting sounds a bit like stepping into z index hell

100%, that's annoying. A better workaround is removing the pointer-events from the internal backdrop when rendering your own. Of course this is just a workaround, not a permanent fix - Base UI shouldn't interfere with your styles here to begin with.

```js
<Dialog.Backdrop
  ref={React.useCallback((node) => {
    const internalBackdrop = node && node.nextElementSibling;
    if (internalBackdrop?.role === 'presentation') {
      internalBackdrop.style.pointerEvents = "none";
    }
  }, [])}
/>
```

The `cursor: pointer` thing is more of a personal gripe I have with web apps overusing the style, but in the case of a backdrop, it seems particularly unnecessary (compared to say buttons). For instance, Instagram and GitHub both don't set it on their modal dialogs. Why do you think users need to see the cursor change when hovering the backdrop if the biggest websites, nor OS modals, don't change it? Backdrops are already assumed to be clickable to dismiss, since it's a common UI paradigm

marcospgp:
@atomiks personally I think in terms of affordance in UI it makes sense to be consistent. If some clickable things have cursor pointer on hover, why not all? 

one could argue some things are obviously clickable and don't need the extra hint, and that any affordance benefits are irrelevant for mobile users who don't have pointers at all

but I believe in consistency so would either go with all clickable items having cursor pointer or none, and prefer to make it all

regarding backdrops specifically, there is no visual distinction between a dismissible and a non-dismissible backdrop, so this information is conveyed by the cursor when hovering it on a desktop device.

> A better workaround is removing the pointer-events from the internal backdrop when rendering your own

[...]

PR #4230, opened 2026-03-02
[popups] Fix useDismiss when component is portaled into shadow DOM

### LINKED ISSUE #3603: [menu] Content doesn't close on interaction outside of popup in ShadowRoot
# Bug report

## Current behavior

https://github.com/user-attachments/assets/2f3e7d2e-5064-4973-9040-c33c8ab0887a

When menu opened (with passed ShadowRoot container to MenuPortal), it doesn't close on interaction outside of popup in ShadowRoot. But when clicking not on the ShadowRoot, it closes. 

## Expected behavior

The menu should detect that the click occurred outside its own bounds inside ShadowRoot and close automatically.

dmitri-gb:
This fixes #3603

`useDismiss` expects the root ancestor of the "click-outside" target to contain a "marker" (an element matching `[data-base-ui-inert]`), otherwise the click will be ignored. For components that have backdrops (modal dialogs, modal menu popups etc) this means that the backdrop must also satisfy this condition. These components include such a marker via the `InternalBackdrop` component. However, `useDismiss` uses `document.querySelectorAll` to find all these markers, which means it will not find any that live inside the shadow DOM. Passing `container={shadowRoot}` to `Dialog.Portal`, `Menu.Portal` etc where `shadowRoot` is attached to a direct child of `<body>` makes this marker invisible to `useDismiss`.

As a result, modal dialogs, menu popups etc that are portaled into a top-level ShadowRoot cannot be dismissed by a "click-outside".

There are multiple ways how this could be fixed. In theory, shadow roots could be nested multiple levels deep etc, but for the backdrop click to work we only need to find the one marker that was rendered by the component that called `useDismiss`. So I went with the smallest possible change — only inspecting the direct shadow root of the click target and appending all markers in it to the existing markers collection. The `contains` check on line 367 already pierces shadow roots, so it will see the added markers.

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useDismiss.ts
+++ b/packages/react/src/floating-ui-react/hooks/useDismiss.ts
@@ -6,6 +6,7 @@ import {
   isElement,
   isHTMLElement,
   isLastTraversableNode,
+  isShadowRoot,
   isWebKit,
 } from '@floating-ui/utils/dom';
 import { Timeout, useTimeout } from '@base-ui/utils/useTimeout';
@@ -328,7 +329,7 @@ export function useDismiss(
         ownerDocument(store.select('floatingElement')).querySelectorAll(inertSelector),
       );
       const targetRoot = isElement(target) ? target.getRootNode() : null;
-      if (targetRoot instanceof ShadowRoot) {
+      if (isShadowRoot(targetRoot)) {
         markers = markers.concat(Array.from(targetRoot.querySelectorAll(inertSelector)));
       }
 


packages/react/src/floating-ui-react/hooks/useDismiss.ts
```diff
@@ -324,9 +324,13 @@ export function useDismiss(
...
       const inertSelector = `[${createAttribute('inert')}]`;
-      const markers = ownerDocument(store.select('floatingElement')).querySelectorAll(
-        inertSelector,
+      let markers = Array.from(
+        ownerDocument(store.select('floatingElement')).querySelectorAll(inertSelector),
       );
+      const targetRoot = isElement(target) ? target.getRootNode() : null;
+      if (targetRoot instanceof ShadowRoot) {
+        markers = markers.concat(Array.from(targetRoot.querySelectorAll(inertSelector)));
+      }
```
atomiks: There's an `isShadowRoot` util

atomiks:
`InternalBackdrop` manually sets `data-base-ui-inert=""` on its JSX, but the user-provided Backdrops don't. And I'm assuming this actually regressed after fixing https://github.com/mui/base-ui/issues/2940 because the user-provided Backdrop now renders on top of the InternalBackdrop.

Given this change, could we remove the `data-base-ui-inert=""` prop from InternalBackdrop?

dmitri-gb:
Hmm, no, I don't think we can remove it. We actually rely on it. This is precisely the marker that is found in `useDismiss`. I don't think it matters which backdrop renders on top, it only matters that the marker (InternalBackdrop) exists in the same shadow root where the user backdrop (the click target) is.

atomiks (approved):
Thanks for the contribution @dmitri-gb!

PR #4231, opened 2026-03-02
[menu] Optimize submenu hover pointer-events

atomiks:

With big DOMs, this shows adding `pointer-events: none;` on the `<body>` is much slower than scoping it to the menu itself when quickly sweeping over submenu triggers. The conditional scope check will be removed before merge.

https://deploy-preview-4231--base-ui.netlify.app/experiments/menu/pointer-events-scope

https://github.com/user-attachments/assets/28e4b559-56a0-41af-96ab-6b458ab02a23

### DIFF
--- a/packages/react/src/number-field/root/NumberFieldRoot.test.tsx
+++ b/packages/react/src/number-field/root/NumberFieldRoot.test.tsx
@@ -658,6 +658,20 @@ describe('<NumberField />', () => {
       expect(onValueCommitted.lastCall.args[0]).to.equal(0);
       expect(input).to.have.value('0');
     });
+
+    it('includes the correct reason for increment and decrement button presses', async () => {
+      const onValueCommitted = spy();
+      await render(<NumberField defaultValue={0} onValueCommitted={onValueCommitted} />);
+
+      const inc = screen.getByLabelText('Increase');
+      const dec = screen.getByLabelText('Decrease');
+
+      fireEvent.click(inc);
+      expect(onValueCommitted.lastCall.args[1].reason).to.equal(REASONS.incrementPress);
+
+      fireEvent.click(dec);
+      expect(onValueCommitted.lastCall.args[1].reason).to.equal(REASONS.decrementPress);
+    });
   });
 
   describe('prop: disabled', () => {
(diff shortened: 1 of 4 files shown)


packages/react/src/menu/store/MenuStore.ts
```diff
@@ -128,7 +128,26 @@ export class MenuStore<Payload> extends ReactStore<
       this.unsubscribeParentListener?.();
 
       if (parent.type === 'menu') {
+        let rootId = parent.store.select('rootId');
+        let floatingTreeRoot = parent.store.select('floatingTreeRoot');
+        let keyboardEventRelay = parent.store.select('keyboardEventRelay');
```
atomiks, line 133: This change to MenuStore specifically mainly improves submenus that have nested triggers:

scoped + items (flatter case)
TaskDuration median: 1.602s -> 1.574s (-1.8%)
ScriptDuration median: 0.324s -> 0.308s (-4.7%)

scoped + triggers (nested submenu-trigger stress case)
TaskDuration median: 2.857s -> 2.030s (-28.9%)
ScriptDuration median: 1.483s -> 0.734s (-50.5%)

atomiks:
## Codex Review

### Overview
This PR changes hover submenus so pointer-event blocking stays local to the menu area instead of relying on a whole-page fallback, and it reduces unnecessary submenu updates from parent changes. The latest update removes the body/scoped toggle and keeps scoped behavior on all the time.

### Findings

#### 1. ℹ️ [Informational] Focused Chromium timing is mostly close, with one noisy outlier on head

**Impact**: The updated scoped-only path does not show a clear functional regression, but the timing samples are noisy and include one slower run on head.

**Evidence**: Command used for both refs: `pnpm test:chromium MenuSubmenuTrigger --no-watch` (5 runs each on latest `master` and PR `HEAD`).

|:--|--:|--:|--:|--:|
| `master` (`7db989f31`) | 1727 | 1663 | 2078 | 5 |
| `HEAD` (`c5a7f46bd`) | 2047 | 1754 | 3261 | 5 |

**Recommendation**: Keep an eye on this in CI perf runs; if the outlier pattern repeats, capture a larger sample size on the submenu-hover path before drawing conclusions.

### Confidence: 4/5
High confidence from a full PR pass, focused jsdom/chromium tests on touched menu paths, and fresh head-vs-master timing checks.

### Notes
- The scoped cleanup behavior remains covered by the `MenuRoot` regression test.
- The experiment now reflects only scoped mode, matching the runtime behavior.

PR #4233, opened 2026-03-02
[toast] Add useToastActions hook to avoid unnecessary re-renders

### LINKED ISSUE #4234: [toast] useToastManager() causes unnecessary re-renders for action-only consumers
## Summary

Components that call `useToastManager()` solely to fire toasts (using `add`, `close`, `update`, `promise`) re-render on **every** toast state change — even though they never read the `toasts` array.

## Reproduction

```tsx
function SaveButton() {
  const { add } = Toast.useToastManager();
  return <button onClick={() => add({ title: 'Saved!' })}>Save</button>;
}
```

This component re-renders every time **any** toast is added, closed, or times out — anywhere in the app.

## Root cause

`useToastManager()` unconditionally calls `store.useState('toasts')`, which registers a `useSyncExternalStore` subscription. Every mutation produces a new array reference, so `Object.is(oldToasts, newToasts)` is always `false` and the component re-renders.

The action methods (`addToast`, `closeToast`, etc.) are stable references — arrow functions on the `ToastStore` class instance. But the subscription is registered before the caller destructures the return value, so destructuring only `{ add }` doesn't help.

## Expected behavior

Components that only need to **dispatch** toast actions should **not** re-render when toast state changes. Only components that **read** `toasts` (to display them) should subscribe and re-render.

## Proposed solution

Add a `useToastActions()` hook that returns only the stable action methods without subscribing to state.

PR: #4233

Yevgeni-Esh:
Fixes #4234

## Problem

`useToastManager()` unconditionally calls `store.useState('toasts')`, which registers a `useSyncExternalStore` subscription. This means **every** component that calls `useToastManager()` re-renders on every toast lifecycle event (add, close, update, timeout) — even if the component only destructures action methods like `{ add }` and never reads the `toasts` array.

```tsx
// This button re-renders on every toast state change, despite only using `add`
function SaveButton() {
  const { add } = Toast.useToastManager();
  return <button onClick={() => add({ title: 'Saved!' })}>Save</button>;
}
```

The action methods (`addToast`, `closeToast`, etc.) are stable references — arrow functions on the `ToastStore` class instance. But the subscription is registered before the caller destructures the return value, so destructuring only `{ add }` doesn't help.

## Solution

Add a new `useToastActions()` hook that returns only the stable action methods (`add`, `close`, `update`, `promise`) without subscribing to state:

- No `store.useState()` call → no `useSyncExternalStore` subscription → zero re-renders from toast state changes
- `ToastContext` holds the `ToastStore` instance (stable ref created via `useRefWithInit`), so `useContext` won't trigger re-renders either
- The action methods are arrow functions on the class instance — their references are stable

### Usage

```tsx
// Components that only FIRE toasts — zero re-renders from toast state
function SaveButton() {
  const { add } = Toast.useToastActions();
  return <button onClick={() => add({ title: 'Saved!' })}>Save</button>;
}

// Components that DISPLAY toasts — subscribe to state (unchanged)
function ToastList() {
  const { toasts } = Toast.useToastManager();
  return toasts.map((toast) => <Toast.Root key={toast.id} toast={toast} />);
}
```

## Changes

- **New hook**: `packages/react/src/toast/useToastActions.ts`
- **Exports**: Added to `index.parts.ts` and `index.ts`
- **Tests**: 5 new tests including a render count assertion that verifies the component does **not** re-render when toasts are added/dismissed
- **Type spec**: Mirrors `useToastManager.spec.tsx` — validates generic `<Data>` inference
- **Docs**: Added `useToastActions` section to toast docs with return value reference and usage guidance
- **Public types**: Updated `test/public-types/toast.tsx` with `UseToastActionsReturnValue`
- **Error codes**: New code 93 for missing provider error

All existing `useToastManager` tests (31/31) continue to pass.

### DIFF
--- a/docs/src/app/(docs)/react/components/toast/page.mdx
+++ b/docs/src/app/(docs)/react/components/toast/page.mdx
@@ -445,6 +445,65 @@ const promise = toastManager.promise(
 );
 ```
 
+## useToastActions
+
+Returns only the action methods without subscribing to toast state.
+Use this hook in components that only need to **fire** toasts (for example, buttons) but don't need to read the `toasts` array.
+Unlike `useToastManager`, this hook will not cause the component to re-render when toasts are added, updated, or removed.
+
+```tsx title="Usage"
+const toastActions = Toast.useToastActions();
+```
+
+### Return value
+
+<PropsReferenceTable
+  type="return"
+  data={{
+    add: {
+      type: '(options: ToastManagerAddOptions) => string',
+      description: 'Add a toast to the toast list.',
+    },
+    close: {
+      type: '(toastId: string) => void',
+      description: 'Closes and removes a toast from the toast list.',
+    },
+    update: {
+      type: '(toastId: string, options: ToastManagerUpdateOptions) => void',
+      description: 'Update a toast in the toast list.',
+    },
+    promise: {
+      type: '<Value>(promise: Promise<Value>, options: ToastManagerPromiseOptions) => Promise<Value>',
+      description:
+        'Create a toast that resolves with a value, with three possible states for the toast: `loading`, `success`, and `error`.',
+    },
+  }}
+/>
+
+### When to use
+
+Use `useToastActions` when the component only needs to **dispatch** toasts:
+
+```jsx title="Example" {2,7-9}
+function SaveButton() {
+  const toastActions = Toast.useToastActions();
+  return (
+    <button
+      type="button"
+      onClick={() => {
+        toastActions.add({
+          description: 'Changes saved!',
+        });
+      }}
+    >
+      Save
+    </button>
+  );
+}
+```
+
+Use `useToastManager` when the component needs to **read** the `toasts` array (for example, to render the toast list).
+
 export const metadata = {
   keywords: [
     'React Toast',
--- a/docs/src/app/(docs)/react/components/page.mdx
+++ b/docs/src/app/(docs)/react/components/page.mdx
@@ -1298,6 +1298,9 @@ Generates toast notifications.
     - `update` method
     - `close` method
     - `promise` method
+  - useToastActions
+    - Return value
+    - When to use
 - Exports:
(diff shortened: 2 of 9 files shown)


docs/src/app/(docs)/react/components/toast/page.mdx
```diff
@@ -445,6 +445,65 @@ const promise = toastManager.promise(
 );
 ```
 
+## useToastActions
+
+Returns only the action methods without subscribing to toast state.
+Use this hook in components that only need to **fire** toasts (for example, buttons) but don't need to read the `toasts` array.
+Unlike `useToastManager`, this hook will not cause the component to re-render when toasts are added, updated, or removed.
```
atomiks, line 452: This looks okay, but you can also use the [global manager](https://base-ui.com/react/components/toast#global-manager) to avoid re-renders, and might be better than adding a new API to the mix here

Yevgeni-Esh: I think using just global manager is not ideal

in a nested page or component causing re-renders... i don't think its best practice to use the ref of the global manager object 

**re-render example**
https://stackblitz.com/edit/9xjrqjfk-mu5uzrrf?file=src%2FApp.tsx
using `const { add } = Toast.useToastManager();`
like it says to use in the documentation.

**no rerender**
https://stackblitz.com/edit/9xjrqjfk?file=src%2FApp.tsx
using `toastManager.add({`

but this means you must make an export of export const toastManager = Toast.createToastManager(); in the main app and import it from a page/component and not use it from the provider 

but i get that this involves new API...

or... maybe just update the docs that doing const { add } = Toast.useToastManager(); may cause re-renders so they should do import toastManager and do toastManager.add? :)

Yevgeni-Esh: @atomiks 
WDYT?

PR #4239, opened 2026-03-02
[docs-infra] Update Index to Filter Private Entries

dav-is:
Update pages index to use new format from https://github.com/mui/mui-public/pull/1109 released in `0.5.1-canary.0`

Adds private Calendar page: https://deploy-preview-4239--base-ui.netlify.app/react/components/calendar (this page will not appear at https://master--base-ui.netlify.app/). Based on this ENV variable: https://github.com/mui/mui-public/blob/df142b55588e09b911db8be848c133b0ebad1cc0/packages/docs-infra/src/withDocsInfra/withDeploymentConfig.ts#L8-L46

Also adds filtering from navigation and search when pages are marked with `audience: private`:

```js
export const metadata =
  /** @type {import('@mui/internal-docs-infra/createSitemap/types').NextMetadata} */ ({
    robots: {
      index: false,
    },
    other: {
      audience: 'private',
    },
  });
```

`metadata.robots.index: false` is how Next.js adds the `no-index` meta: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#robots

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -58,7 +58,7 @@
     "remark-typography": "0.7.3",
     "scroll-into-view-if-needed": "3.1.0",
     "server-only": "^0.0.1",
-    "shiki": "^3.23.0",
+    "shiki": "^4.0.1",
     "to-vfile": "^8.0.0",
     "unist-util-visit-parents": "^6.0.2",
     "vfile-matter": "^5.0.1"
--- a/docs/src/app/(private)/experiments/menu/menu-nested.tsx
+++ b/docs/src/app/(private)/experiments/menu/menu-nested.tsx
@@ -48,7 +48,9 @@ export default function NestedMenu() {
               </Menu.SubmenuRoot>
 
               <Menu.SubmenuRoot>
-                <Menu.SubmenuTrigger className={styles.SubmenuTrigger}>Style</Menu.SubmenuTrigger>
+                <Menu.SubmenuTrigger className={styles.SubmenuTrigger} closeDelay={1000}>
+                  Style
+                </Menu.SubmenuTrigger>
                 <Menu.Portal>
                   <Menu.Positioner align="start" side="right" sideOffset={12}>
                     <Menu.Popup className={styles.MenuPopup}>
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "@arethetypeswrong/cli": "0.18.2",
     "@babel/plugin-transform-react-constant-elements": "7.27.1",
     "@base-ui/monorepo-tests": "workspace:*",
-    "@mui/internal-code-infra": "0.0.4-canary.2",
+    "@mui/internal-code-infra": "0.0.4-canary.4",
     "@mui/internal-netlify-cache": "0.0.3-canary.0",
     "@mui/internal-test-utils": "2.0.18-canary.11",
     "@next/eslint-plugin-next": "16.1.6",
@@ -104,9 +104,9 @@
     "vite": "7.3.1",
     "vitest": "4.0.18"
   },
-  "packageManager": "pnpm@10.30.2",
+  "packageManager": "pnpm@10.30.3",
   "engines": {
-    "pnpm": "10.30.2",
+    "pnpm": "10.30.3",
     "node": ">=22.18.0"
   },
   "resolutions": {
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -86,8 +86,8 @@
   "dependencies": {
     "@babel/runtime": "^7.28.6",
     "@base-ui/utils": "workspace:*",
-    "@floating-ui/react-dom": "^2.1.7",
-    "@floating-ui/utils": "^0.2.10",
+    "@floating-ui/react-dom": "^2.1.8",
+    "@floating-ui/utils": "^0.2.11",
     "tabbable": "^6.4.0",
     "use-sync-external-store": "^1.6.0"
   },
--- a/packages/react/src/accordion/root/AccordionRoot.test.tsx
+++ b/packages/react/src/accordion/root/AccordionRoot.test.tsx
@@ -1,4 +1,4 @@
-import { screen } from '@mui/internal-test-utils';
+import { fireEvent, screen } from '@mui/internal-test-utils';
 import { expect } from 'chai';
 import { spy } from 'sinon';
 import { DirectionProvider } from '@base-ui/react/direction-provider';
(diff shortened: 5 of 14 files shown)


docs/src/app/(docs)/react/components/accordion/page.mdx
```diff
@@ -41,17 +41,18 @@ import { DemoAccordionMultiple } from './demos/multiple';
...
-    'Accessible Accordion',
-    'Multi-Panel UI',
-    'Disclosure Widget',
-    'FAQ Component',
-    'Headless React Components',
-    'Base UI',
-  ],
-};
+export const metadata =
+  /** @type {import('@mui/internal-docs-infra/pipeline/transformMarkdownMetadata/types').Metadata} */ ({
```
mnajdova: Can we add the typeonly on pages that actually use params not defined in the next.js type?

docs/src/app/(docs)/react/handbook/page.mdx
```diff
@@ -61,7 +61,7 @@ A guide to animating Base UI components.
 
 A guide to composing Base UI components with your own React components.
 
-<details>
+<details data-autogenerated-from="./composition">
```
mnajdova: Is this leaking from some other changes?

docs/src/components/Header.tsx
```diff
@@ -14,7 +14,7 @@ export const titleMap: Record<string, string> = {
 
 export const HEADER_HEIGHT = 48;
 
-export function Header() {
+export function Header({ isPublic }: { isPublic: boolean }) {
```
mnajdova: Can we call this `isPrivate`? We use private as a tag, private as a audience etc.

docs/src/app/(docs)/react/components/accordion/page.mdx
dav-is: Right now the metadata doesn't have any JSDocs so even the Next.js types aren't shown when hovering or autocomplete. If we have the type here, then our added metadata options are discoverable. This type extends Next.js' types too.

If we don't put our metadata type here, I think we should at least put the `@type` for Next.js metadata.

Edit: I reverted this for now

docs/src/app/(docs)/react/handbook/page.mdx
dav-is: The mui-public PR updates the page index format to address some shortcomings I've noticed since adding it to base-ui. This change allows us to detect that a page is an index from the html directly and helps AI detect that the section is autogenerated

Edit: I reverted this change and copied it to a separate PR

dav-is:
I'll separate out the unrelated changes into another PR to keep this one minimal

dav-is:
@mnajdova Ok, the unrelated changes are reverted, I renamed to `isPrivate`, and I added an empty calendar page to test: https://deploy-preview-4239--base-ui.netlify.app/react/components/calendar

docs/src/app/(docs)/react/components/accordion/page.mdx
mnajdova: We can do it in a follow-up, just so we don't clatter this PR.

docs/src/app/(docs)/react/overview/page.mdx
```diff
@@ -1,13 +1,13 @@
 # Overview
 
-[//]: # 'This file is autogenerated, but the following list can be modified.'
+[//]: # 'This section is autogenerated, but the following list can be modified except within the parentheses.'
```
mnajdova: What exactly can be modified? Can we be more specific? Can I change the hash? Or the title etc.

docs/src/components/Search/Search.tsx
```diff
@@ -6,15 +6,18 @@ const sitemap = () => import('../../app/sitemap');
 export function Search({
   enableKeyboardShortcut = false,
   containedScroll = false,
+  isPrivate = false,
```
mnajdova: When is the default used?

docs/src/app/(docs)/layout.tsx
```diff
@@ -10,6 +10,8 @@ import { sitemap } from 'docs/src/app/sitemap';
 import 'docs/src/css/index.css';
 import './layout.css';
 
+const isPrivate = process.env.DEPLOY_ENV !== 'production' && process.env.DEPLOY_ENV !== 'staging';
```
mnajdova: Could this be a new env variable that we can set in the deploy preview and the prod build? Or if not, let's rename this to `hidePrivatePages` to be more explicit.

Apologies for the double comment on the same prop, but I feel we should be more explicit.

docs/src/app/(docs)/react/overview/page.mdx
dav-is: Title, tags, and the list order can be modified: `- Button [New]`. The comment asks that you don't edit within the anything parenthesis: `(Private, [Outline](#Button), [Contents](./button/page.mdx))`

docs/src/app/(docs)/layout.tsx
dav-is: As in create `process.env.isPrivate`?

docs/src/components/Search/Search.tsx
dav-is: Removed this prop in favor of an env var

docs/src/app/(docs)/react/overview/page.mdx
dav-is: Updated the message to be more explicit: `This section is autogenerated, but the following list order, title, and [Tag]s can be modified, but nothing within the parentheses.`

docs/src/app/(docs)/layout.tsx
dav-is: At Jan's suggestion, created `process.env.SHOW_PRIVATE_PAGES`

docs/src/app/(docs)/react/components/page.mdx
```diff
@@ -215,6 +216,12 @@ A button component that can be rendered as another tag or focusable when disable
 
 [Read more](./button/page.mdx)
 
+## Calendar
```
mnajdova, line 219: Should we check here for private pages too?

mnajdova: Nevermind it's private page :D

docs/src/app/(docs)/layout.tsx
```diff
@@ -10,6 +10,8 @@ import { sitemap } from 'docs/src/app/sitemap';
 import 'docs/src/css/index.css';
 import './layout.css';
 
+const showPrivatePages = process.env.SHOW_PRIVATE_PAGES === 'true';
```
mnajdova, line 13: Perfect 👌

mnajdova:
It looks good to me 👌 Do you plan to add more things in this PR? I am asking because it's still in draft

dav-is:
> Do you plan to add more things in this PR?

No, I can do it in a followup PR. It's a draft because we need to merge https://github.com/mui/mui-public/pull/1109 before this PR can be merged. That PR should be ready to go whenever it is approved

docs/src/app/(docs)/react/components/calendar/page.mdx
```diff
@@ -0,0 +1,18 @@
...
+<Meta
+  name="description"
+  content="A high-quality, unstyled React calendar that is easy to customize."
+/>
+
+export const metadata =
+  /** @type {import('@mui/internal-docs-infra/createSitemap/types').NextMetadata} */ ({
+    keywords: [],
+    robots: {
+      index: false,
```
brijeshb42, line 13: Will this work for the [`generateMetadata`](https://nextjs.org/docs/app/getting-started/metadata-and-og-images#generated-metadata) function ?

dav-is: `generateMetadata` is not supported by page indexes today because it requires running the build and rendering the page. When we use `export const metadata`, we can statically analyze by parsing the file. I want to support it eventually, but we need to further optimize Base UI's build pipeline so that a cached build takes less than 5 seconds (my target).

docs/src/app/(docs)/react/components/calendar/page.mdx
```diff
@@ -0,0 +1,18 @@
+# Calendar
```
mnajdova, line 1: We can drop this page as there are other pages where we can test the private flag.

docs/src/app/(docs)/layout.tsx
mnajdova, line 13: We should make sure we don't forget to set this flag before the next deployment.

mnajdova (approved):
I left two more comments, I don't see anything blocking. Let's update the infra package and merge this one 👌

docs/src/app/(docs)/layout.tsx
dav-is: This is set within docs-infra code `withDeploymentConfig`

PR #4241, opened 2026-03-03
[docs] Show explicit aria props in demos for i18n

### LINKED ISSUE #2245: [docs] Display `aria-label` for components that should be internationalized

### LINKED ISSUE #4226: [number field] min/max attributes
# Get help

## Ask a question

I expected the min and max props on a `NumberField` to set the `input`'s `min` and `max` attributes for screen readers. A suppose this is due to `type="text"` being used for the input.

I suppose the best way to provide accessiblity would be to provide a aria-description for screen readers? Any insights would help. Thank you!

atomiks:

Closes #2245
Closes #4226

The API table shows this can be customized in special cases, but the demos having it in-built is clearer by default.

ARIA translation works okay (but not fully reliably if the app isn't i18n'd): https://adrianroselli.com/2019/11/aria-label-does-not-translate.html#Update07. So the internal English labels are better than nothing to work with auto-translation in browsers. One issue is in buttons if the icon descendant has the `aria-label` instead of the button, our source code may unexpectedly override that with the English internal label

### Caveat

Range-slider demos rely on Slider’s default internal `aria-valuetext` for multi-thumb text ("start range" / "end range"). So while thumb names (`aria-label`) are explicit, the demos do not currently demonstrate customizing/localizing `aria-valuetext` via `getAriaValueText`. I've omitted it since this otherwise makes the demos awkward and often the valuetext duplicates the label anyway

### DIFF
--- a/docs/src/app/(docs)/react/components/number-field/demos/hero/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/number-field/demos/hero/css-modules/index.tsx
@@ -16,11 +16,11 @@ export default function ExampleNumberField() {
       </NumberField.ScrubArea>
 
       <NumberField.Group className={styles.Group}>
-        <NumberField.Decrement className={styles.Decrement}>
+        <NumberField.Decrement aria-label="Decrease" className={styles.Decrement}>
           <MinusIcon />
         </NumberField.Decrement>
-        <NumberField.Input className={styles.Input} />
-        <NumberField.Increment className={styles.Increment}>
+        <NumberField.Input aria-roledescription="Number field" className={styles.Input} />
+        <NumberField.Increment aria-label="Increase" className={styles.Increment}>
           <PlusIcon />
         </NumberField.Increment>
       </NumberField.Group>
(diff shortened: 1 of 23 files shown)


michaldudak:
### Code review

Found 1 issue:

1. The `anchored` Toast demo renders two separate `Toast.Viewport` components on the same page, and the PR assigns `aria-label="Notifications"` to both. Since `Toast.Viewport` renders as a `role="region"` landmark, two regions with an identical accessible name violate WCAG 2.4.6 — screen reader users navigating by landmarks cannot distinguish between them. The two viewports serve different purposes (anchored vs. stacked) and should carry distinct labels (e.g. `aria-label="Anchored notifications"` and `aria-label="Notifications"`).

https://github.com/mui/base-ui/blob/966056f3aa29659ffe0c861a545603f96004642a/docs/src/app/(docs)/react/components/toast/demos/anchored/css-modules/index.tsx#L92-L116

PR #4242, opened 2026-03-03
[slider] Remove extra TalkBack focus targets

### LINKED ISSUE #2575: [slider] Broken TalkBack support

atomiks:

Fixes #2575

Removing the `tabIndex={-1}` from both wrappers makes TalkBack land on the actual `input[type="range"]` only, which is the element that correctly handles increase/decrease actions

Preview: https://deploy-preview-4242--base-ui.netlify.app/react/components/slider

### DIFF
--- a/packages/react/src/slider/thumb/SliderThumb.tsx
+++ b/packages/react/src/slider/thumb/SliderThumb.tsx
@@ -335,10 +335,10 @@ export const SliderThumb = React.forwardRef(function SliderThumb(
       max,
       min,
       name,
-      onChange(event: React.ChangeEvent<HTMLInputElement>) {
+      onChange(event) {
         handleInputChange(event.currentTarget.valueAsNumber, index, event);
       },
-      onFocus(event: React.FocusEvent<HTMLInputElement>) {
+      onFocus(event) {
         const isRestoringFocusVisible = restoringFocusVisibleRef.current;
         restoringFocusVisibleRef.current = false;
         setActive(index);
@@ -348,7 +348,7 @@ export const SliderThumb = React.forwardRef(function SliderThumb(
           event.stopPropagation();
         }
       },
-      onBlur(event: React.FocusEvent<HTMLInputElement>) {
+      onBlur(event) {
         if (restoringFocusVisibleRef.current) {
           event.stopPropagation();
           return;


mj12albert:
In Firefox:
- This PR: pointer always triggers the focus visible style
- master: pointer never triggers focus visible, but keyboard after pointer (incorrectly) never triggers focus visible

Since Firefox is weird either way it's probably OK if it fixes the issue for TalkBack

atomiks:
> This PR: pointer always triggers the focus visible style

Not always, only if you've never interacted with another focusable element before. (Same as the issue we had some time ago.)

That said, FF supports `focusVisible` in `.focus()` broadly, so we can actually just use that 👍

michaldudak:
### Code review

Found 1 issue:

1. New test code uses `chai` and `sinon` instead of vitest-native equivalents. `SliderControl.test.tsx` introduces a fresh `import { expect } from 'chai'` (previously the file had no chai dependency), and `SliderThumb.test.tsx` uses chai-style assertions in the newly added tests. [AGENTS.md says](https://github.com/mui/base-ui/blob/22da2cbc996460d70cfdd5efab3cf7932ac1031c/AGENTS.md#L39): "The repository is transitioning from `chai` and `sinon`, prefer vitest native functions for all new code." The fix is to use vitest's `expect` (imported from `'vitest'`) and replace `.not.to.have.attribute('tabindex')` with `.not.toHaveAttribute('tabindex')`, etc.

https://github.com/mui/base-ui/blob/22da2cbc996460d70cfdd5efab3cf7932ac1031c/packages/react/src/slider/control/SliderControl.test.tsx#L1-L5

https://github.com/mui/base-ui/blob/22da2cbc996460d70cfdd5efab3cf7932ac1031c/packages/react/src/slider/thumb/SliderThumb.test.tsx#L196-L200

atomiks:
## Codex Review (GPT-5.4)

This remains a narrow slider bug-fix PR, and I reviewed the full rebased branch diff against freshly fetched `upstream/master`. The current head now addresses the earlier follow-up concern as well: the internal focus-visible promotion path is covered by a regression that ensures it does not emit spurious blur/focus side effects.

The runtime shape now looks coherent for the bug this PR is trying to fix. Pointer-driven focusing still suppresses `:focus-visible` from the control-side path, while keyboard-driven slider interaction can promote the already-focused range input back into a `:focus-visible` state through the thumb-side refocus path.

The latest follow-up also tightens that path by suppressing bubbled `onBlur`/`onFocus` noise from the internal blur/refocus cycle, so the implementation detail does not leak through the public thumb event handlers.

### 3. Edge Cases Worth Noting

The non-obvious part of the patch is still the guarded blur/refocus used to restore `:focus-visible`. That branch is now better protected because the thumb test suite covers both the visible focus-ring behavior and the absence of extra public blur/focus events during the internal handoff.

Validation I ran on the current rebased head:

- `pnpm typescript`
- `pnpm test:chromium SliderThumb --no-watch`
- `pnpm test:chromium SliderRoot --no-watch`
- `pnpm test:jsdom SliderThumb --no-watch`
- `pnpm test:jsdom SliderRoot --no-watch`
- `pnpm test:jsdom SliderControl --no-watch`

The key coverage improvement is the new browser regression in `packages/react/src/slider/thumb/SliderThumb.test.tsx`, which exercises the pointer-then-keyboard path and asserts that the internal focus-visible restoration does not generate an extra `onBlur`/`onFocus` pair.

### 5. Previous Review Issues Status

**Resolved:** the earlier review note about newly introduced chai-style assertions was addressed, and the later follow-up concern about missing coverage for the blur-suppression branch is now covered by the new thumb-level regression. The branch also includes the follow-up typecheck fix and has been rebased onto current `upstream/master`.

### Merge Confidence Score

Overall merge confidence is 4/5. I’m highly confident in the current patch: the full rebased diff is still small, the critical browser-specific interaction path has direct Chromium coverage, and I did not find any remaining correctness issue in the branch review. The residual risk is mostly the usual browser nuance around nonstandard `focusVisible` behavior rather than a concrete gap in this implementation.

atomiks:
I manually verified this works in TalkBack now. `:focus-visible` issues are resolved across browsers as well.

PR #4257, opened 2026-03-04
[code-infra] Fix typescript-next CI job

Janpot:
Depends on https://github.com/mui/mui-public/pull/1170

## Summary

Add `types: []` to build tsconfigs. This will be the new [default](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0-beta/#types-now-defaults-to-[]) in v6. Fix the issues with ambient `@types/node` not being pulled in automatically anymore. (Which is a good thing)

This fixes the `typescript-next` CI job and prepares for upcoming v6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "@arethetypeswrong/cli": "0.18.2",
     "@babel/plugin-transform-react-constant-elements": "7.27.1",
     "@base-ui/monorepo-tests": "workspace:*",
-    "@mui/internal-code-infra": "https://pkg.pr.new/mui/mui-public/@mui/internal-code-infra@a48038e",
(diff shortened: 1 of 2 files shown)


packages/react/tsconfig.build.json
```diff
@@ -10,7 +10,8 @@
     "moduleResolution": "bundler",
     "noEmit": false,
     "rootDir": "./src",
-    "outDir": "build/esm"
+    "outDir": "build/esm",
+    "types": []
```
brijeshb42: Can `@base-ui/utils/end.d.ts` be added here to avoid duplication ?

Janpot: It should probably rather come from a `@mui/internal-code-infra/build-env`.

PR #4276, opened 2026-03-06
[navigation-menu] Close parent menus when nested link with closeOnClick is clicked

### LINKED ISSUE #4262: [navigation menu] closeOnClick only closes the nearest NavigationMenu.Root in nested menus
I'm not sure if this is intended behavior or a bug, so this may be more of a usage question than a bug report.

In a **nested submenu implementation**, when `NavigationMenu.Link` is rendered as a **Next.js `Link` component**, `closeOnClick` only closes the **nearest `NavigationMenu.Root`** (the submenu), while the **parent/root menu remains open**.

This results in the submenu closing, focus returning to the parent trigger, and the main menu remaining open.

Ideally, when a navigation link is clicked, **all open menu layers would close**, which is typically the expected behavior for navigation menus.

## Base UI version
1.2.0

## Which OS are you using?
Mac OS

## Expected Behavior

When a `NavigationMenu.Link` is clicked with `closeOnClick={true}`:

- The clicked submenu should close
- Any **ancestor menus** should also close
- Navigation should proceed normally

## Actual Behavior

Only the **closest `NavigationMenu.Root`** closes.  
The parent menu remains open and focus returns to its trigger.

## Implementation Context

- Using **Base UI NavigationMenu**
- Nested menus using multiple `NavigationMenu.Root` components adhering to the example code structure
- Links rendered via **Next.js `Link`** using the `render` prop

Example pattern:

```tsx
<NavigationMenu.Link
  render={<Link href="/example" />}
  closeOnClick
  {...props}
/>

```

## Workaround

I know it's possible to control the menu state manually via value / onValueChange and close the root menu when a link is clicked.

However, doing so causes the initial mount animation to originate from the top-left corner of the screen, which produces a less than ideal transition. This could be user error, but I'm largely working from shadcn's example styles at this stage. Previously (with Radix NavigationMenu), I handled mount/unmount animations with Framer Motion, but I'd prefer not to introduce another animation library just to solve this.

## Question

Is the current behavior intentional, or is there a recommended pattern for ensuring all open menus close when a navigation link is clicked in nested menu setups?

CiscoFran10:
Closes #4262

Previously, when a nested link with `closeOnClick` was clicked, only the nearest `NavigationMenu.Root` closed and the parent menu stayed open. We now bubble that close up so the whole menu closes without needing to control the root with `value` / `onValueChange`.

**Changes**

- **NavigationMenuRoot.tsx:** Nested roots get the parent via `useNavigationMenuRootContext(true)`. In `setValue`, after updating local state, if we're closing with `eventDetails.reason === REASONS.linkPress`, we call `parentRootContext.setValue(null, eventDetails)` so the parent (and its ancestors) close as well.
- **NavigationMenuRoot.test.tsx:** Added two fixtures, one with a single nested level and optional `onValueChange`, one with three nested levels and three tests under "nested menus": parent closes when nested link is clicked, parent `onValueChange` is called with `null`, and all three levels close when the deepest link is clicked.

### DIFF
--- a/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
+++ b/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
@@ -1670,6 +1670,7 @@ describe('<NavigationMenu.Root />', () => {
       expect(screen.queryByTestId('content-1')).to.equal(null);
       expect(trigger1).to.have.attribute('aria-expanded', 'false');
     });
+
     describe('inline nested viewport', () => {
       it('renders viewport content correctly for inline nested menu', async () => {
         await render(<TestInlineNestedNavigationMenu />);


atomiks:
## Codex Review

### Overview
This patch makes `closeOnClick` links inside nested navigation menus close the parent menus as well, and it adds regression coverage for nested and deeply nested cases. The branch now includes the current `master`, so this review reflects the full PR diff on top of the latest base.

### Findings (None)

### Confidence: 4/5
High confidence based on a full pass over `packages/react/src/navigation-menu/root/NavigationMenuRoot.tsx` and `packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx`, plus targeted `NavigationMenuRoot` runs in JSDOM and Chromium after merging the latest `master`.

### Notes
- Verified with `pnpm test:jsdom NavigationMenuRoot --no-watch`.
- Verified with `pnpm test:chromium NavigationMenuRoot --no-watch`.

PR #4279, opened 2026-03-06
[docs] Add accessible label to demo Dialog's textarea

### LINKED ISSUE #4251: [dialog] accessibility: close confirmation example includes textarea using a placeholder as its label
# Bug report

## Current behavior

The "Close confirmation" dialog example contains a textarea that is programmatically labelled by pale placeholder text only. This causes problems for people with cognitive or memory impairments as the label disappears on focus / when typing. 

It can also cause problems for speech input users due to the programmatic and visible label mismatch: The visible label above (the dialog heading) is not the programmatic label for the textarea, even though it looks like it should be the label. 

Additionally, the placeholder (#898B8E) text is low contrast against the textarea background (approximately 3.4:1), meaning people with low vision may struggle to read it and understand that is the label.
 
## Expected behavior

This example should include a `<textarea>` that is programmatically labelled by the persistently visible heading text above it instead of by the placeholder. The placeholder should only be used for supplementary information and not as an input label.

## Reproducible example

Steps to replicate:

1. Navigate to [Dialog: Close confirmation](https://base-ui.com/react/components/dialog#close-confirmation)
2. Activate the dialog trigger button
3. Inspect the textarea with devtools to confirm its accessible name is "What’s on your mind?" and this comes from placeholder text
4. Focus on/start typing into the input
5. Note the placeholder text disappears

Optional additional test:

1. Turn on Voice Control on a Mac (or other speech input assistive tech)
2. Navigate to [Dialog: Close confirmation](https://base-ui.com/react/components/dialog#close-confirmation)
3. Speak "Click New Tweet"
4. Note, nothing happens
5. Speak "Click What’s on your mind?" and note focus moves to the textarea

## Base UI version

v1.2.0

Chrome

## Which OS are you using?

MacOS

Apple Voice Control

## Additional context

The dialog heading is sufficient to act as the permanently visible label for the textarea. You could use `aria-labelledby` on the textarea to reference the heading above, or use `aria-label` or a visually-hidden `<label>` element to achieve this. The important point is to ensure that the textarea has the same accessible name as the heading above.

michaldudak:
Labels the Dialog's textarea in the https://deploy-preview-4279--base-ui.netlify.app/react/components/dialog#close-confirmation demo with the Title element

Closes https://github.com/mui/base-ui/issues/4251

### DIFF
--- a/docs/src/app/(docs)/react/components/dialog/demos/close-confirmation/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/dialog/demos/close-confirmation/css-modules/index.tsx
@@ -8,6 +8,7 @@ export default function ExampleDialog() {
   const [dialogOpen, setDialogOpen] = React.useState(false);
   const [confirmationOpen, setConfirmationOpen] = React.useState(false);
   const [textareaValue, setTextareaValue] = React.useState('');
+  const titleId = React.useId();
 
   return (
     <Dialog.Root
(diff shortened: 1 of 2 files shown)


atomiks (approved):
Aaron & Colm are re-doing docs and demos so there might be conflicts, but it's easy enough for LLMs to fix conflicts now for no effort, so I think we can still fix the a11y issues here.

PR #4280, opened 2026-03-06
[toast] Prevent duplicate `onClose` calls for ending toasts

chuganzy:
This PR fixes an issue where `onClose` could be triggered multiple times when a toast is already in the ending state as pointed out [here](https://github.com/mui/base-ui/pull/3979#discussion_r2788414422).


atomiks:
## Codex Review

### Overview
This patch now updates toast state before running `onClose`, and it extends the regression test so a toast can call `close()` from inside `onClose` without double-firing. The branch stays narrowly focused on the toast close path and the new test coverage matches the bug being fixed.

### Findings (None)

### Confidence: 5/5
High confidence based on the full branch diff, the narrow implementation surface, and focused verification of the shared toast-manager code path.

### Notes
- Re-ran `pnpm test:jsdom createToastManager --no-watch` on the PR head.
- Re-ran `pnpm test:jsdom useToastManager --no-watch` on the PR head.
- The latest commit addresses the earlier re-entrant `onClose` concern by moving the callback after the state update and by strengthening the regression test to cover that sequence.

packages/react/src/toast/store.ts
```diff
@@ -273,14 +270,20 @@ export class ToastStore extends ReactStore<State, {}, typeof selectors> {
...
     }
     this.update(updates);
+
+    toastsToClose.forEach((toast) => {
+      if (toast.transitionStatus !== 'ending') {
+        toast.onClose?.();
+      }
+    });
+
+    this.handleFocusManagement(toastId);
```
chuganzy, line 286: This function is the only one with a side effect, so I moved it below to keep the order unchanged.
The execution order remains `onClose` → `focus update (onFocus / onBlur)`.

Although this depends on the state internally, it only looks up the next (or prev) active toast, so I believe changing the position should not cause any issues.

PR #4284, opened 2026-03-09
[accordion] Activate trigger on Space keydown

atomiks:

Fixes https://github.com/mui/base-ui/issues/4051#issuecomment-4003416516, Accordion doesn't infer CompositeRoot and needs an explicit prop

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  code-infra: https://raw.githubusercontent.com/mui/mui-public/716b18b5736b5dbc740d49e481c99cb85b74e42e/.circleci/orbs/code-infra.yml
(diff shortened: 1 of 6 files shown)


atomiks:
## Codex Review

### Overview
This patch makes Accordion triggers activate on Space keydown so they behave like the library's other composite-style widgets. It also adds regression coverage for both native and non-native trigger rendering, including the focus precondition needed for browser keyboard events.

### Findings (None)

### Confidence: 4/5
High confidence based on the small surface area of the change, the full-branch review against `master`, and direct verification of the affected Accordion suites in both JSDOM and Chromium.

### Notes
- Reviewed the full branch diff against `master`, not just the latest commit.
- I also checked for the same `useButton` and composite-list mismatch elsewhere in the library; Accordion appears to be the only affected component in this codebase.
- Verified with `pnpm test:jsdom Accordion --no-watch` and `pnpm test:chromium Accordion --no-watch`.

Issue #4250, opened 2026-03-03
[combobox] accessibility: Guidance text not announced by screen readers


grace-snow:
# Bug report

## Current behavior

The "Async search" variant reveals instruction text on ArrowDown, which is not announced by NVDA, JAWS or VoiceOver. 

![The combobox input with text that has appeared below saying "Start typing to search people..."](https://github.com/user-attachments/assets/41a7c3b4-57c0-40a7-a676-48f10f4e521d)

The `aria-live` region looks like it is being dynamically rendered instead of being dynamically populated.

## Expected behavior

When instructional text appears for an input, this must be announced by screen readers.

## Reproducible example

Steps to replicate:
1. Turn on a screen reader
2. Navigate to [Combobox: Async search example](https://base-ui.com/react/components/combobox#async-search-single)
3. When focused on the input, use ArrowDown to make the instruction text appear visually
4. Note it is not announced by the screen reader

## Base UI version

v1.2.0

## Which browser are you using?

Chrome

## Which OS are you using?

Windows and MacOS

## Which assistive tech are you using (if applicable)?

NVDA, JAWS, VoiceOver

## Additional context

Using `role="alert"` on the guidance text's container or pre-rendering the aria-live region should fix the announcements.

Adding an aria-describedby to the input may also improve announcements.

Issue #4181, opened 2026-02-24
[autocomplete] accessibility: Limited results status message not announced by screen readers


grace-snow:
# Bug report

## Current behavior

In the [Limit Results demo of the autocomplete component](https://base-ui.com/react/components/autocomplete#limit-results), screen readers don't announce the message "Hiding # results (type a more specific query to narrow results" 

![The open autocomplete with message at the end of the listbox highlighted explaining that the number of visible suggestion is limited](https://github.com/user-attachments/assets/bfe3f3a2-3e3c-4405-97f7-3a28ac794847)

Tested with
* VoiceOver in Chrome and Safari on Mac OS 
* NVDA in Chrome on Windows
* JAWS in Chrome on Windows

## Expected behavior

All screen readers should announce the full message so that screen reader users know suggestions have been limited and how many are hidden.

This is currently an issue under [WCAG 2.2 Success Criterion 4.1.3 Status Messages](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html) and possibly [WCAG 2.2 Success Criterion 3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)

## Reproducible example

Steps to reproduce:
1. Navigate to [the Autocomplete Limit Results demo](https://base-ui.com/react/components/autocomplete#limit-rresults)
2. Turn on your screen reader
3. Type a single letter, such as "a", into the input
4. Note that the message at the bottom of the listbox about hiding results is not announced

## Base UI version

v1.2.0

## Which browser are you using?

Safari and Chrome

## Which OS are you using?

Mac OS and Windows

## Which assistive tech are you using (if applicable)?

VoiceOver, NVDA and JAWS

## Additional context

Adding `aria-describedby` to the listbox referencing the unique `id` of the status message may fix the announcement.

There may be a problem with the way the status message has been coded, which is causing it to not be announced. Potential solutions may be: 
* to add a slight delay before populating the message inside the aria-live region
* using an assertive live announcement (or alert role) 
* removing aria-atomic

atomiks:
> Adding aria-describedby to the listbox referencing the unique id of the status message may fix the announcement.

Will look into this idea.

> There may be a problem with the way the status message has been coded, which is causing it to not be announced. Potential solutions may be:

Yes the issue is aria-live regions don't announce their content when first mounted, they only announce content changes (so the delay idea would work but this is problematic when trying to avoid visible layout shifts, etc.)

PR #4286, opened 2026-03-09
[combobox] Fix initial live region announcements

atomiks:

Fixes https://github.com/mui/base-ui/issues/4181
Fixes https://github.com/mui/base-ui/issues/4250

The key issue is `role=status` doesn't announce its content upon DOM insertion (e.g. popup mount) unlike `role=alert` but the latter is too aggressive and annoying for what Empty and Status do. `aria-describedby` doesn't work for the "input outside popup" (default/common) pattern because it doesn't reflect changes while focus remains on the input.

I found a trick: VoiceOver announces the raw text content changing from `{text}&nbsp;` to `{text}` after a task tick (setTimeout). This 1) prevents visual layout shift, and 2) actually forces it to announce the text even though it was pre-populated.

## Manual testing results

- VoiceOver (macOS Chrome, Firefox, Safari) ✅ (Safari requires 200ms delay. I hardcoded 200ms as all work with that)
- VoiceOver (iOS Safari) ✅ [**already worked, no mutation needed, which otherwise double-announces**]
- JAWS (Windows Chrome, Firefox, Edge) ✅ 
- NVDA (Windows Chrome, Firefox, Edge) ✅ 
- Narrator (Windows Chrome, Firefox, Edge) ✅ 
- TalkBack (Android) ✅ [**already worked, and unlike iOS doesn't double-announce with this change**]

## Demos changes

The demos changes aren't strictly necessary but I changed it because users will tend to prefer `hidden` (which breaks announcements) instead of the weird `:not(:empty)`. Making it an always-mounted parent with a layout-styled child makes this more obvious.

### DIFF
--- a/packages/react/src/combobox/empty/ComboboxEmpty.tsx
+++ b/packages/react/src/combobox/empty/ComboboxEmpty.tsx
@@ -22,7 +22,7 @@ export const ComboboxEmpty = React.forwardRef(function ComboboxEmpty(
   componentProps: ComboboxEmpty.Props,
   forwardedRef: React.ForwardedRef<HTMLDivElement>,
 ) {
-  const { render, className, children: childrenProp, ...elementProps } = componentProps;
+  const { render, className, style, children: childrenProp, ...elementProps } = componentProps;
 
   const { filteredItems } = useComboboxDerivedItemsContext();
   const store = useComboboxRootContext();
--- a/packages/react/src/combobox/status/ComboboxStatus.tsx
+++ b/packages/react/src/combobox/status/ComboboxStatus.tsx
@@ -17,7 +17,7 @@ export const ComboboxStatus = React.forwardRef(function ComboboxStatus(
   componentProps: ComboboxStatus.Props,
   forwardedRef: React.ForwardedRef<HTMLDivElement>,
 ) {
-  const { render, className, children: childrenProp, ...elementProps } = componentProps;
+  const { render, className, style, children: childrenProp, ...elementProps } = componentProps;
 
   const statusRef = useInitialLiveRegionTextMutation<HTMLDivElement>();
 


atomiks:
## Codex Review (GPT-5.4)

This follow-up pass reviews the fully rebased `upstream/master...HEAD` branch at `7e5792e8d`. The Safari follow-up keeps the runtime delta very small: the live-region helper still uses the existing `\u2060` text mutation path and only extends the reset delay to 200ms.

No new blocking issues found in the rebased branch.

| Area | Assessment | Notes |
|------|:----------:|-------|
| Initial polite live-region announcement | **Handled** | `Combobox.Empty` and `Combobox.Status` still share the same mount-time text-node mutation helper. |
| Safari follow-up | **Narrowed** | The only runtime change in the follow-up commit is `timeout.start(200, ...)` in `useInitialLiveRegionTextMutation`. |
| Blast radius | **Low** | No browser detection, no extra DOM node, and no additional state plumbing were added for the Safari workaround. |

| Area | Status | Notes |
|------|:------:|-------|
| `Empty` / `Status` parity | **Consistent** | Both components still document and implement the same always-mounted live-region guidance. |
| Docs/examples | **Consistent** | The docs continue to steer consumers toward `:empty`-style layout handling instead of removing the live region from the accessibility tree. |

**Handled:** the heavier Safari-only suffix experiment was not kept. The final runtime shape is about as small as this approach can get without abandoning the helper or changing behavior again.

**Remaining:** if this still proves insufficient in Safari, the next fallback would be a shared persistent announcer, but that would be a materially larger pattern than the current branch.

| Case | Status | Notes |
|------|:------:|-------|
| Raw `textContent` observers | **Low risk** | The `\u2060` marker now lives for 200ms instead of 0ms, so only code that inspects raw text content would observe the marker longer. |
| Layout stability | **Handled** | Keeping the zero-width marker avoids visible content flashes while still giving VoiceOver a text mutation to notice. |

| `pnpm typescript` | **Passed** | Clean on the rebased branch. |
| `pnpm test:jsdom ComboboxStatus --no-watch` | **Mixed** | Passed on the follow-up change before the rebase; the post-rebase rerun now stops during an unrelated `TemporalAdapterDateFns` transform on `date-fns/addDays`. |
| `pnpm test:jsdom ComboboxEmpty --no-watch` | **Mixed** | Passed on the follow-up change before the rebase; the post-rebase rerun now stops during the same unrelated `TemporalAdapterDateFns` transform. |

| Item | Status | Notes |
|------|:------:|-------|
| Earlier review summary | **Updated** | The branch is now rebased onto fresh `upstream/master` and force-pushed. |
| Bundle-trimming follow-up | **Handled** | The final Safari tweak stays inside the existing helper instead of adding browser-specific branching or new DOM structure. |

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Root-cause coverage | 4/5 | The branch still addresses the original mount-announcement gap with a narrow helper. |
| Regression risk | 4/5 | The follow-up changes one timeout value plus matching test expectations. |
| Validation coverage | 3/5 | TypeScript passed, but the latest JSDOM reruns are currently gated by an unrelated transform-time import resolution failure. |

Overall merge confidence is 4/5. I do not see a new correctness issue in the rebased branch, and the Safari follow-up is now about as small as this mutation-based approach can reasonably be.

ciampo:
what about using a shared aria-live region? changing its contents should cause assistive technology to announce it

atomiks:
> what about using a shared aria-live region? changing its contents should cause assistive technology to announce it

Yeah this is one of the alternatives I tried, but it adds another visually-hidden persistent thing after the hidden input, & requires state updates to sync the children which is weird

ciampo:
Yeah I get it. We currently use that approach — a singleton aria-live region appended to the document.body that is used by all componentry for assistive tech announcements. Conceptually, I guess, not too different from addint toasts via the toast manager ?

atomiks:
@ciampo yep, but `Toast.Viewport` is expected to be always-mounted as part of the API unlike these ones, so it just works naturally. This one is conditionally mounted, so I wish there was an `aria` attr to force `role=status` to announce on mount (same as `alert`), which would avoid this problem, but alas

atomiks:
## Codex Review (GPT-5.4)

This is still a focused `bug-fix` PR for combobox live-region announcements, with secondary `a11y`, `tests`, `docs`, and `react-ui` work. I re-reviewed the full branch after the latest test-structure fix, and I don’t have any remaining actionable findings on the pushed head.

I don’t have any remaining actionable findings on the current branch.

The runtime approach looks coherent end-to-end: `Combobox.Status` and `Combobox.Empty` both use the shared initial live-region mutation helper, iOS skips the workaround entirely, and the temporary marker reset delay is sourced from a shared combobox constant instead of being duplicated between runtime and tests.

The pattern remains internally consistent across runtime and docs. The live-region root stays mounted, visible spacing/layout moves onto child wrappers, and the public guidance for Combobox/Autocomplete now matches the updated examples.

The `packages/react/src/combobox/utils/constants.ts` extraction addresses the main maintenance risk from the earlier pass. With the delay shared between implementation and fake-timer assertions, there isn’t an obvious further simplification I’d push for in this patch.

The iOS bypass coverage is now more robust than before because it lives in a dedicated fixed-platform test file, [ComboboxStatus.iOS.test.tsx](/Users/atomiks/.codex/worktrees/51b4/base-ui/packages/react/src/combobox/status/ComboboxStatus.iOS.test.tsx), instead of relying on a mutable browser mock inside the shared status suite. That avoids the Chromium inconsistency the earlier version hit.

`pnpm test:jsdom ComboboxStatus --no-watch` passed.

`pnpm test:chromium ComboboxStatus --no-watch` passed.

`pnpm test:jsdom ComboboxEmpty --no-watch` passed.

`pnpm typescript` passed.

The earlier failing `clock.tick(0)` assertions are aligned with the shared reset-delay constant, and the iOS early-return path now has stable coverage in both test environments.

The previous blocking issue is fixed: the focused suites are green after syncing the fake-timer assertions with the shared delay constant.

The previous non-blocking gap is also fixed: there is now direct, stable coverage for the iOS early-return path, including Chromium.

### Merge Confidence Score

9.0/10. The runtime patch looks solid, the docs/examples stay aligned with the implementation, and the targeted validation I’d want for this accessibility fix is green in the environments that previously surfaced drift.

PR #4292, opened 2026-03-10
[select] Fix `alignItemWithTrigger` fallback with browser zoom

### LINKED ISSUE #4290: [select] alignItemWithTrigger positioning breaks at non-100% browser zoom
## Description

`alignItemWithTrigger={true}` on `Select.Positioner` produces inconsistent popup positioning at non-100% desktop browser zoom levels (Ctrl+/- or Cmd+/-). The popup sometimes overlays the trigger with the selected item aligned (correct), and sometimes falls back to opening below the trigger — varying between opens without any user-visible reason.

## Steps to reproduce

1. Render any `<Select>` with `alignItemWithTrigger={true}` (the default)
2. Set browser zoom to 90% (also reproducible at 110%, 125%)
3. Open and close the Select repeatedly

**Expected**: Popup consistently overlays the trigger with the selected item text aligned to the trigger's value text.

**Actual**: Intermittent — sometimes the popup overlays correctly, sometimes it opens below the trigger (fallback to standard Floating UI positioning fires inconsistently).

## Environment

- `@base-ui/react` 1.2.0
- Chrome (also reproducible in Firefox)
- macOS
- Desktop browser zoom (not pinch zoom)

## Analysis

The root cause appears to be in `SelectPopup.js` (~lines 267–310), where viewport dimensions from `document.documentElement.clientHeight` (integer) are mixed with `getBoundingClientRect()` values (float). At non-100% zoom:

1. **Integer vs float mismatch**: `clientHeight`/`clientWidth` return integers (floor of actual CSS pixel value), while `getBoundingClientRect()` returns floats. At non-100% zoom, sub-pixel rounding errors accumulate in the positioning math.

2. **Fallback threshold not reliably triggered**: The `triggerCollisionThreshold` (20px) is large enough that sub-pixel errors often don't exceed it, so the fallback (`setControlledAlignItemWithTrigger(false)`) fires inconsistently — sometimes the error is just enough to trigger it, sometimes not.

3. **No desktop zoom guard**: The existing Safari pinch-zoom guard checks `visualViewport.scale`, which only reflects mobile pinch zoom — it remains `1` during desktop browser zoom (Ctrl+/-). There's no `devicePixelRatio`-based detection for desktop zoom.

## Suggested fix

A few possible approaches:

- **Desktop zoom detection**: Use `devicePixelRatio` heuristics to detect non-100% desktop zoom and fall back to standard positioning:
  ```js
  const isNonStandardZoom = Math.abs(
    window.devicePixelRatio / Math.round(window.devicePixelRatio) - 1
  ) > 0.01;
  ```

- **Use `visualViewport.width`/`height`** instead of `clientHeight`/`clientWidth` for viewport measurements — these return float values and are more accurate at all zoom levels.

- **Round consistently**: Apply consistent rounding (e.g., `Math.round`) to all computed positioning values to minimize sub-pixel drift.

atomiks:
Fixes #4290

https://github.com/user-attachments/assets/6b3fb0d8-4c84-448a-8fec-458d87c2bf3a

CSS zoom looks to correctly replicate the rounding issues and I manually verified above in the recording

## Summary

`Select.Positioner` with `alignItemWithTrigger` could intermittently fall back to standard positioning at non-100% browser zoom.

The issue came from mixing zoom-scaled `getBoundingClientRect()` values with layout-pixel measurements used during the initial popup placement logic. For short popups, that mismatch could make the computed aligned height land just below the fallback threshold, so the popup would sometimes open below the trigger instead of staying aligned with it.

This showed up more easily in smaller selects like the docs styling selector, while larger demos were less likely to hit the same rounding edge case.

## Changes

- Normalize popup placement rects using Floating UI's scale calculation before comparing them with layout-pixel measurements.
- Allow a 1px tolerance in the fallback height comparison to avoid flipping due to subpixel rounding.
- Add a regression test covering zoom-scaled DOMRect values for a short popup.

## Verification

Verified on `localhost:3007/react/components/select` with synthetic page zoom (`document.documentElement.style.zoom = '0.9'`):

- Before this change, the docs styling selector fell back to standard positioning (`data-side="bottom"`).
- After this change, it consistently stays in aligned mode (`data-side="none"`).
- The larger select demo continues to behave correctly.


atomiks:
## Codex Review

### Overview
This patch fixes the Select popup zoom bug by keeping the placement math in the same coordinate space and by tolerating the small rounding errors that appear in the short-popup path. It also adds browser regressions for the two edge cases that were still causing the popup to fall back or grow too tall at non-100% zoom.

### Findings (None)

### Confidence: 4/5
High confidence after reviewing the full `upstream/master...HEAD` diff for the changed Select popup files and verifying the updated regression coverage. The remaining uncertainty is mostly browser-specific layout behavior, but the new Chromium regressions now cover both the 1px `maxScrollTop` drift and the fractional-height threshold case that were missing earlier.

### Notes
- Reviewed against `upstream/master...543cb35a0`.
- Verified with `pnpm test:jsdom SelectPopup --no-watch`, `pnpm test:chromium SelectPopup --no-watch`, and `pnpm typescript`.

packages/react/src/select/popup/SelectPopup.tsx
```diff
@@ -580,6 +583,23 @@ function getMaxScrollTop(scroller: HTMLElement) {
   return Math.max(0, scroller.scrollHeight - scroller.clientHeight);
 }
 
+function getScale(element: HTMLElement) {
+  // The platform API is async-capable, but the DOM platform returns a plain scale object.
+  return floatingPlatform.getScale(element) as { x: number; y: number };
+}
```
atomiks, line 589: This returns `1` instead of `0` in unsafe cases so there are no divide-by-zero issues 😅

michaldudak:
It seems to fix the issue on Chrome/Edge on both Windows and macOS, but I still see the problem on Safari (both on macOS and iPad when using a mouse or pen).


https://github.com/user-attachments/assets/d12755d8-92d2-42a2-8a05-3452913ce4ef

PR #4295, opened 2026-03-11
[drawer] Make `data-base-ui-swipe-ignore` explicit for touch interactions

### LINKED ISSUE #4294: [Drawer] Swipe dismiss handler blocks onClick on non-interactive elements inside drawer on touch devices
## Summary

The Drawer's swipe dismiss handler (`useSwipeDismiss`) prevents `onClick` from firing on non-interactive elements (e.g., `<div onClick>`) inside the drawer on touch devices, while native interactive elements like `<button>` work correctly.

## Steps to reproduce

1. Render a left-side Drawer with clickable `<div onClick>` elements and `<button onClick>` elements inside the popup content
2. Open the drawer on a touch device (or mobile emulator)
3. Tap a `<div onClick>` element → **onClick does not fire**
4. Tap a `<button onClick>` element → onClick fires correctly

## Root cause

In `useSwipeDismiss.ts`, `startSwipeAtPosition` checks if the touch target matches `DEFAULT_IGNORE_SELECTOR`:

```ts
const DEFAULT_IGNORE_SELECTOR = 'button,a,input,select,textarea,label,[role="button"]';
```

- `<button>` matches → swipe tracking is skipped → normal click fires
- `<div>` does NOT match → swipe tracking starts (`setSwiping(true)`) → touch events are consumed → `onClick` never synthesizes

### Additional issue: `data-swipe-ignore` doesn't work on touch

The `DrawerViewport` has a `data-swipe-ignore` check in `onPointerDown`, but the handler exits early for `event.pointerType === 'touch'` (line ~682) **before** reaching that check. So the escape hatch exists but doesn't work for the input type where the bug occurs.

### Prior art: Vaul

Vaul (which Base UI's drawer is inspired by) provides `data-vaul-no-drag` — a data attribute that opts elements out of swipe tracking for all input types including touch.

## Suggested fix

One or more of:

1. **Make `data-swipe-ignore` work for touch events** — move the check before the touch early-return in the `onPointerDown` handler, or add it to the `onTouchStart` handler
2. **Expose `ignoreSelector` as a configurable prop** on `Drawer.Root` or `Drawer.Viewport` so users can extend `DEFAULT_IGNORE_SELECTOR`
3. **Add a universal `data-no-swipe` attribute** (like Vaul's `data-vaul-no-drag`) that works for both pointer and touch events

## Current workaround

Adding `role="button"` to clickable `<div>` elements makes them match `[role="button"]` in the ignore selector. This works and improves accessibility, but it's a workaround — users shouldn't need to know about internal selectors to make click handlers work inside a drawer.

## Environment

- `@base-ui/react` version: `1.2.0`
- Browser: Mobile Safari / Chrome on iOS/Android
- Framework: React 19

atomiks:

Fixes #4294

## Summary

This changes swipe-ignore handling so Drawer and Toast use the namespaced `data-base-ui-swipe-ignore` attribute, while Toast still accepts the legacy `data-swipe-ignore` alias for compatibility.

For Drawer specifically:

- `Drawer.Content` no longer auto-adds a public swipe-ignore attribute.
- Touch taps on non-interactive elements inside `Drawer.Content` still fire `onClick` without requiring `data-base-ui-swipe-ignore`.
- `data-base-ui-swipe-ignore` now acts as the explicit opt-out for swipe dismissal, matching the Vaul-style behavior for the full touch gesture.
- Mouse and pen interactions on `Drawer.Content` still avoid swipe handling so text selection keeps working.

The patch also adds focused regression coverage for Drawer and Toast, updates the docs, and includes a small private experiment page for manual touch testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  code-infra: https://raw.githubusercontent.com/mui/mui-public/96a7c608aa3603924622f70e8fe30cf084d59909/.circleci/orbs/code-infra.yml
+  code-infra: https://raw.githubusercontent.com/mui/mui-public/993d672e95892fc2f0e64b730dca5a79644b8d0a/.circleci/orbs/code-infra.yml
 
 parameters:
   browserstack-force:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       - run: pnpm release:build
       - name: Publish packages
         if: matrix.os == 'ubuntu-latest'
-        uses: mui/mui-public/.github/actions/ci-publish@96a7c608aa3603924622f70e8fe30cf084d59909
+        uses: mui/mui-public/.github/actions/ci-publish@993d672e95892fc2f0e64b730dca5a79644b8d0a
         with:
           pr-comment: true
           use-compact-url: true
--- a/.github/workflows/new-issue-triage.yml
+++ b/.github/workflows/new-issue-triage.yml
@@ -9,7 +9,7 @@ permissions: {}
 jobs:
   issue_cleanup:
     name: Clean issue body
-    uses: mui/mui-public/.github/workflows/issues_body-cleanup.yml@96a7c608aa3603924622f70e8fe30cf084d59909
+    uses: mui/mui-public/.github/workflows/issues_body-cleanup.yml@993d672e95892fc2f0e64b730dca5a79644b8d0a
     permissions:
       contents: read
       issues: write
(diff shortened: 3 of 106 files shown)


atomiks:
## Codex Review

### Overview
This patch makes swipe opt-out explicit with `data-base-ui-swipe-ignore`, keeps `Drawer.Content` focused on mouse and pen text selection, and preserves the old toast attribute for compatibility. It also keeps touch taps working on non-interactive drawer content, centralizes the `elementFromPoint` fallback, and adds regression coverage for the updated gesture paths.

### Findings (None)

### Confidence: 5/5
High confidence after a full review of the branch diff against `master`, plus a second sweep across React/DOM behavior, accessibility, API design, maintainability, performance, and test coverage. The exercised tests cover the changed gesture paths in JSDOM and Chromium.

### Notes
- Reviewed the full branch diff for PR #4295, not only the latest commit.
- Latest changes since the previous review are wording-only test updates; the underlying behavior and coverage remain consistent.
- Drawer now only honors the namespaced swipe-ignore attribute, while Toast keeps the legacy alias for compatibility.

doruksahin:
Thanks @atomiks :D 
That was fast

PR #4296, opened 2026-03-11
[combobox] Fix Chips area clicks not focusing input or opening popup

### LINKED ISSUE #4289: [combobox][docs] Multiple selection combobox examples ignores mouse clicks in padding area
# Bug report

## Current behavior

When clicking in the "padding" area of a `<Combobox.Chips>` component, the input does not gain focus.

https://github.com/user-attachments/assets/114540be-c357-4a2e-9b9a-dbda93c049ae

## Expected behavior

The input should gain focus when clicking in the padding area (and probably use a "text" cursor there too). This would match what happens with a normal Input component, and is what one would typically expect given that Chips usually looks the same as Input.

## Reproducible example

Can be reproduced [in the docs](https://base-ui.com/react/components/combobox#multiple-select).

## Base UI version

v1.2.0, but also applicable to the current state in `master`.

Chrome

## Which OS are you using?

Mac OS

CiscoFran10:

Clicking anywhere in the `<Combobox.Chips>` container (padding or a chip) now focuses the input and opens the popup when `openOnInputClick` is true, matching the behavior of clicking the input. Demos that use Chips use `cursor: text` on the container so the area looks and behaves like an input.

Closes #4289.

---

**Current behavior (before fix):**
- Clicking the **padding** of the Chips container did nothing: the input did not receive focus.
- Clicking a **chip** focused the input only because each chip had its own `onMouseDown`; the container had no handler.
- The popup did not open when clicking padding or chip, because Floating UI’s `useClick` is bound to the **reference** (the input). Clicks on Chips never hit that reference.

---

**Changes:**
- **`ComboboxChips.tsx`**  
  - Subscribe to store: `disabled`, `readOnly`, `openOnInputClick`.
  - Add an `onMouseDown` handler on the container that:
    - Returns when `readOnly`.
    - Calls `event.preventDefault()` then returns when `disabled`.
    - Otherwise: focuses the input via `store.state.inputRef.current?.focus()`, then, if `openOnInputClick`, calls `store.state.setOpen(true, createChangeEventDetails(REASONS.inputPress, event.nativeEvent))`.
    - Invokes `elementProps.onMouseDown?.(event)` so consumer handlers still run.

- **`ComboboxChip.tsx`**  
  - Remove the per-chip `onMouseDown` (focus input).  
  - Clicks on a chip bubble to the container; the container’s handler now does focus + open. No duplicated logic.

- **`ComboboxChips.test.tsx`**
  - “focuses the input when clicking anywhere in the Chips area”: `fireEvent.mouseDown(chips)` then `fireEvent.mouseDown(chip)`.

### DIFF
--- a/packages/react/src/combobox/input-group/ComboboxInputGroup.test.tsx
+++ b/packages/react/src/combobox/input-group/ComboboxInputGroup.test.tsx
@@ -1,6 +1,7 @@
 import { expect, vi } from 'vitest';
 import * as React from 'react';
 import { Combobox } from '@base-ui/react/combobox';
+import { Field } from '@base-ui/react/field';
 import { createRenderer, describeConformance } from '#test-utils';
 import { fireEvent, screen } from '@mui/internal-test-utils';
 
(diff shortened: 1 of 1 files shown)


atomiks:
## Codex Review

### Overview
This branch makes combobox chips and input-group padding clicks behave more like direct input clicks, so clicking the styled container focuses the input and can open the popup. It also keeps chip-remove presses from collapsing that flow, reuses the shared interactive-element helper in hover logic, and updates the demos to match the new interaction surface.

### Findings
#### 1. ℹ️ [Informational] Add a hover regression around links after the shared selector change

**Impact**: The hover refactor now routes both hover paths through a single interactive-element selector, so future link-related regressions will be harder to catch until that new behavior is pinned by a focused test.

**Evidence**: `packages/react/src/floating-ui-react/hooks/useHover.ts` now imports the shared `isInteractiveElement`, `packages/react/src/floating-ui-react/hooks/useHoverInteractionSharedState.ts` re-exports that same helper, and `packages/react/src/floating-ui-react/utils/element.ts` defines the new shared selector with `a[href]`. The touched `packages/react/src/floating-ui-react/hooks/useHover.test.tsx` coverage still mounts only a `<button>` reference and does not add a link-specific regression.

**Recommendation**: Add a focused `useHover` regression that exercises an anchor-based interactive target so the new shared selector behavior is explicitly covered.

### Confidence: 4/5
High confidence based on a full `master...HEAD` pass of the current branch, plus parallel category sweeps for React correctness, accessibility, API/state design, maintainability, performance, and test quality. The remaining risk is low and mostly limited to unpinned edge-case coverage rather than a concrete logic defect.

### Notes
- Focused verification run on this branch included `pnpm test:jsdom ComboboxChips --no-watch`, `pnpm test:jsdom ComboboxInputGroup --no-watch`, `pnpm test:jsdom ComboboxChipRemove --no-watch`, `pnpm test:jsdom useHover --no-watch`, `pnpm test:chromium ComboboxChips --no-watch`, `pnpm test:chromium ComboboxInputGroup --no-watch`, and `pnpm test:chromium ComboboxChipRemove --no-watch`.
- I did not run a dedicated benchmark because this diff does not introduce a plausible hot-path performance regression.

atomiks:
With the new `InputGroup` part, the padding styles changed a bit, so clicking the padding area still doesn't work here

PR #4299, opened 2026-03-11
[code-infra] Prevent `combiner` to have default parameters

JCQuintas:
In createSelectorMemoized (packages/x-internals/src/store/createSelector.ts:125), the number of extra arguments a combiner expects is determined by combiner.length:

`const argsLength = Math.max(combiner.length - nSelectors, 0);`

In JavaScript, `Function.length` does not count parameters with default values. So a combiner like `(data, param = 'default') => ...` reports `length = 1` instead of `2`, causing `argsLength = 0`.

This means the memoization logic won't track the argument, breaking memoization silently.

This type forces typescript to show an error when the combiner function contains a default parameter

559263102-e4224087-1868-4298-a8c3-6c79e8c55cf3

Related to https://github.com/mui/mui-x/pull/21707

### DIFF
--- a/packages/react/src/calendar/store/selectors.ts
+++ b/packages/react/src/calendar/store/selectors.ts
@@ -1,11 +1,11 @@
 import { createSelector, createSelectorMemoized } from '@base-ui/utils/store';
 import {
+  TemporalAdapter,
   TemporalSupportedObject,
   TemporalSupportedValue,
-  TemporalAdapter,
 } from '../../types/temporal';
-import { validateDate } from '../../utils/temporal/validateDate';
 import { getInitialReferenceDate } from '../../utils/temporal/getInitialReferenceDate';
+import { validateDate } from '../../utils/temporal/validateDate';
 import { CalendarNavigationDirection, SharedCalendarState as State } from './SharedCalendarState';
 
 const timezoneToRenderSelector = createSelectorMemoized(
@@ -108,7 +108,7 @@ const isSetMonthButtonDisabledSelector = createSelector(
     validationProps,
     isCalendarDisabled,
     targetDate: TemporalSupportedObject,
-    disabledProp?: boolean | undefined,
+    disabledProp: boolean | undefined,
   ) => {
     // short-circuit if the disabled prop is explicitly provided.
     if (disabledProp !== undefined) {


JCQuintas:
@romgrk @michaldudak this change was added on X already, would probably be nice to have it in base-ui too 😆 https://github.com/mui/mui-x/pull/21707

romgrk:
The change makes typechecking fail, one Base UI selector has an optional parameter. As a strategy to workaround the limitation, the selector could be split in two versions (one with the optional argument made non-optional, one with the optional argument).

```
packages/react/src/calendar/store/selectors.ts:106:3 - error TS2345: Argument of type '(adapter: TemporalAdapter, validationProps: { minDate: Date | undefined; maxDate: Date | undefined; }, isCalendarDisabled: boolean, targetDate: Date, disabledProp?: boolean | undefined) => boolean' is not assignable to parameter of type '"Combiner cannot have optional or default parameters because memoization relies on Function.length"'.

106   (
      ~
107     adapter,
    ~~~~~~~~~~~~
... 
111     disabledProp?: boolean | undefined,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
112   ) => {
    ~~~~~~~~
```

JCQuintas:
> The change makes typechecking fail, one Base UI selector has an optional parameter. As a strategy to workaround the limitation, the selector could be split in two versions (one with the optional argument made non-optional, one with the optional argument).
> 
> ```
> packages/react/src/calendar/store/selectors.ts:106:3 - error TS2345: Argument of type '(adapter: TemporalAdapter, validationProps: { minDate: Date | undefined; maxDate: Date | undefined; }, isCalendarDisabled: boolean, targetDate: Date, disabledProp?: boolean | undefined) => boolean' is not assignable to parameter of type '"Combiner cannot have optional or default parameters because memoization relies on Function.length"'.
> 
> 106   (
>       ~
> 107     adapter,
>     ~~~~~~~~~~~~
> ... 
> 111     disabledProp?: boolean | undefined,
>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 112   ) => {
>     ~~~~~~~~
> ```

Everywhere the selector was used it had the "optional" prop passed in, so I just made it required.

PR #4308, opened 2026-03-12
[handbook] Add choosing-components guide and table layout

CiscoFran10:

## Draft / feature concept (DO NOT MERGE)

This PR adds a **Choosing the right component** handbook page to help developers pick the right Base UI component by use case.

https://deploy-preview-4308--base-ui.netlify.app/react/handbook/choosing-components

I've seen this come up a lot: devs opening issues and asking in Discord which component to use (e.g. Select vs Combobox, Dialog vs Popover, when to use Tooltip vs Toast). This is an attempt to address that with a single reference page that groups components by purpose and includes short "key differences" so people can decide without digging through every doc.

image

**Why it's a draft / proposal**

I know the Base UI team is working on a new design for the docs. This is **not** meant to be the final solution, just a concrete proposal you can:

- Use as a starting point for a better "choosing components" experience in the new docs
- Discuss and reshape (e.g. different grouping, placement, or UX)
- Or drop if you prefer to solve it differently in the redesign

I wanted to put the idea and a minimal implementation out there so it can be discussed and, if useful, implemented in a better way as part of the new docs. Happy to iterate or close in favor of a different approach.

### DIFF
--- a/docs/src/app/(docs)/react/handbook/choosing-components/page.mdx
+++ b/docs/src/app/(docs)/react/handbook/choosing-components/page.mdx
@@ -6,8 +6,6 @@
   content="Learn which Base UI component to use for common UI patterns like popups, toggles, form controls, and navigation."
 />
 
-<div className="ChoosingComponentsRoot">
-
 ## Overlay content (dialogs, menus, drawers)
 
 | If you need...                                | Use                                            |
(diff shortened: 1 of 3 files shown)


aarongarciah:
Hey @CiscoFran10, thanks for the suggestion. We won't be adding this page for now, but it's on our todo list improving the guidelines on each component page. Closing for now, but thanks again!

Issue #4305, opened 2026-03-12
[navigation menu] duplicate `aria-orientation` attributes


jesse-mindtrip:
# Bug report

## Current behavior

When `orientation` prop is set on NavigationMenu.Root, both the root + list elements both have `aria-orientation` attributes which is invalid.

## Expected behavior

Only the list element should receive the aria-orientation attribute.

## Base UI version

1.2.0

PR #4309, opened 2026-03-12
[navigation menu] Fix duplicate aria-orientation

atomiks:
## Summary

Fixes [#4305](https://github.com/mui/base-ui/issues/4305) by removing the duplicate `aria-orientation` attribute from `NavigationMenu.Root`.

## What changed

- Removed `aria-orientation` from `NavigationMenu.Root`, so the root element no longer renders an invalid duplicate attribute.
- Applied `aria-orientation` to nested `NavigationMenu.List` instances as well, since nested lists bypass `CompositeRoot` and still need the attribute.

### DIFF
--- a/packages/react/src/navigation-menu/list/NavigationMenuList.tsx
+++ b/packages/react/src/navigation-menu/list/NavigationMenuList.tsx
@@ -66,7 +66,9 @@ export const NavigationMenuList = React.forwardRef(function NavigationMenuList(
   // but we want to block it in this case.
   // When nested, skip this handler so arrow keys can reach the parent CompositeRoot.
   const defaultProps: HTMLProps = nested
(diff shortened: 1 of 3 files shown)


atomiks:
## Codex Review

### Overview
This patch fixes the duplicate `aria-orientation` output in `NavigationMenu` by removing the attribute from the root element and keeping it on the list, including nested list cases. The added tests cover both top-level and nested menus so the regression stays guarded.

### Findings (None)

### Confidence: 4/5
High confidence based on a full review of the branch diff against `master`, an explicit second pass over rendering correctness, accessibility semantics, API/state behavior, maintainability, performance risk, and test coverage, plus a targeted `NavigationMenuRoot` JSDOM test run.

PR #4310, opened 2026-03-12
[navigation menu] Fix delayed trigger switches in Safari

atomiks:

## Summary

This improves Navigation Menu hover performance during trigger-to-trigger switching.

### What changed

- Keep `safePolygon` pointer-events blocking enabled for non-WebKit browsers.
- Disable top-level pointer-events blocking on WebKit, which avoids the Safari hover delay when switching triggers.
- Keep nested Navigation Menu blocking behavior intact.
- Reduce redundant size-sync work during trigger switches by skipping an extra auto-size pass when the trigger event already handled the size transition.
- Add a regression test asserting top-level hover pointer-events mutations stay scoped to the list rather than `document.body`.

### Why

Safari could intermittently delay hover activation when moving directly between top-level triggers. Entering the popup first consistently cleared the issue, which pointed to the top-level pointer-events mutation path rather than `safePolygon` geometry. Disabling that mutation on WebKit fixes the delay while preserving the better non-WebKit behavior.

### DIFF
--- a/packages/react/src/navigation-menu/root/NavigationMenuRoot.webkit.test.tsx
+++ b/packages/react/src/navigation-menu/root/NavigationMenuRoot.webkit.test.tsx
@@ -1,0 +1,139 @@
+import * as React from 'react';
+import { expect } from 'chai';
+import { fireEvent, flushMicrotasks, screen } from '@mui/internal-test-utils';
+import { vi } from 'vitest';
+import { NavigationMenu } from '@base-ui/react/navigation-menu';
+import { createRenderer } from '#test-utils';
+import { OPEN_DELAY } from '../utils/constants';
+
+vi.mock('@base-ui/utils/detectBrowser', async () => {
+  const actual = await vi.importActual<typeof import('@base-ui/utils/detectBrowser')>(
+    '@base-ui/utils/detectBrowser',
+  );
+
+  return {
+    ...actual,
+    isWebKit: true,
+  };
+});
+
+function TestNavigationMenu() {
+  return (
+    <NavigationMenu.Root>
+      <NavigationMenu.List data-testid="top-level-list">
+        <NavigationMenu.Item value="item-1">
+          <NavigationMenu.Trigger data-testid="trigger-1">Item 1</NavigationMenu.Trigger>
+          <NavigationMenu.Content data-testid="popup-1">
+            <NavigationMenu.Link href="#link-1">Link 1</NavigationMenu.Link>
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review

### Overview
This patch narrows the Navigation Menu hover workaround to WebKit so Safari avoids the delayed trigger switch, while nested menus keep the existing scoped pointer-events behavior. It also keeps the trigger-switch size shortcut and adds WebKit-specific regression coverage for both top-level and nested menus.

### Findings

#### 1. ℹ️ [Informational] Chromium timing stayed effectively flat

**Impact**: The change looks targeted to Safari/WebKit responsiveness rather than broad browser throughput, and the focused Chromium check did not show a wider performance regression.

**Evidence**: `packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx:552` scopes the safe-polygon pointer-events path to nested menus on WebKit, while `packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx:648` keeps the trigger-switch size shortcut. WebKit-specific regression coverage was added in `packages/react/src/navigation-menu/root/NavigationMenuRoot.webkit.test.tsx:9`. Focused Chromium benchmark using `pnpm exec cross-env VITEST_ENV=chromium vitest run --project @base-ui/react src/navigation-menu/root/NavigationMenuRoot.test.tsx --no-watch`:

| Ref | Commit | Mean (ms) | p50 (ms) | p95 (ms) | Sample size |
|:--|:--|--:|--:|--:|--:|
| `upstream/master` | `a3190b848` | 8603.67 | 8450.49 | 9035.86 | 3 |
| `PR head` | `cbb4e010e` | 8633.01 | 8393.32 | 9068.31 | 3 |

**Recommendation**: None required. Keep a quick Safari manual check in normal release validation, because the user-facing win here is browser-specific and not expected to show up in Chromium timings.

### Confidence: 4/5
High confidence based on a full-branch diff review, focused JSDOM and Chromium test runs, and a latest-`master` versus PR-head Chromium benchmark. Residual risk is low and mainly limited to Safari-specific pointer-events behavior outside the automated harness.

### Notes
- Latest revision squashes the branch into a single commit and uses `isWebKit` rather than a user-agent-specific Safari branch.
- Focused verification run:
  - `pnpm test:jsdom NavigationMenuRoot --no-watch`
  - `pnpm test:chromium NavigationMenuRoot --no-watch`
  - `pnpm prettier`

Issue #4314, opened 2026-03-13
[render prop] False positive warning for useCallback render functions inside PascalCase components


auduongtuan:
# Bug report

## Current behavior
When passing a valid render callback to Base UI components, Base UI logs an `invalid-render-prop` warning if the callback function is declared inside a PascalCase component and wrapped with `useCallback`.

Example warning:
`Base UI: The `render` prop received a function named `DropdownMenu.useCallback[renderSearchInput]` that starts with an uppercase letter...`

In this case, the callback itself is intentionally lower-case and is not used as `render={Component}`. The warning appears to be triggered by inferred function names that include the parent component name (`DropdownMenu`, `CommandPalette`, etc.).

## Expected behavior
No warning for valid render callbacks that are not React components, even when function name inference includes an uppercase parent component prefix (e.g. `DropdownMenu.useCallback[...]`).

## Reproducible example
Minimal shape:

```tsx
function DropdownMenuExample() {
  const renderSearchInput = useCallback((props) => <input {...props} />, []);

  return <Combobox.Input render={renderSearchInput} />;
}
```

In development mode, warning can still be raised because the inferred function name resembles `DropdownMenuExample.useCallback[renderSearchInput]`.

We also reproduce similarly with `Menu.Trigger` / `Combobox.Trigger` callbacks in React 19 + Next.js dev builds.

## Base UI version
@base-ui/react 1.3.0

## Which browser are you using?
Chrome 134

## Which OS are you using?
macOS 15

## Which assistive tech are you using (if applicable)?
N/A

## Additional context
I understand the intent of this warning (avoid passing React components directly as plain render functions), but this seems like a false positive from name-based heuristics.

Potential improvement ideas:
- only warn for names that clearly match component identifiers (e.g. `^[A-Z][A-Za-z0-9]*$`)
- ignore inferred names containing wrappers like `.useCallback[...]`
- or treat lower-case inner callback identifiers as safe even if parent scope is PascalCase

Thanks for considering this — it causes noisy warnings even when using valid callbacks.

atomiks:
I figured there’d be a false positive from this somewhere (mentioned in the PR description), but I didn’t encounter it anywhere in our codebase as we never use `useCallback` on render functions. It was unfortunately hard to know exactly when this would print without real usage from users...

It shouldn't be necessary to memo the render function in your example btw, but we definitely need to tighten the checks to avoid printing the warning incorrectly like this

PR #4324, opened 2026-03-13
[all components] Tighten render prop warning heuristic

atomiks:
Fixes [#4314](https://github.com/mui/base-ui/issues/4314)

## Summary

Fixes a false positive from the `invalid-render-prop` warning when a valid render callback is wrapped in `useCallback` inside a PascalCase component.

Before this change, Base UI warned for any render function whose inferred `name` started with an uppercase letter. In React 19 / Next.js dev builds, callbacks can get inferred names like `DropdownMenuExample.useCallback[renderSearchInput]`, which caused valid render callbacks to be flagged incorrectly.

This PR tightens the heuristic so the warning only fires for names that look like a plain component identifier, while preserving the existing warning for likely `render={Component}` misuse.

## Changes

- Restrict the render-prop warning heuristic to plain component-like function names
- Avoid warning for inferred names such as `DropdownMenuExample.useCallback[renderSearchInput]`
- Add a regression test covering the `useCallback` naming pattern
- Keep the existing warning behavior for direct uppercase component function


atomiks:
## Codex Review

### Overview
This patch narrows the development-only `render` warning so it stops flagging valid callbacks whose inferred names include wrapper text like `useCallback[...]`. It also adds a regression test for that specific case.

### Findings (None)

### Confidence: 4/5
High confidence for the changed behavior. I reviewed the full branch diff, checked the warning path and test coverage, and reran the focused `useRenderElement` test suite. Residual risk is low and mainly limited to unusual function-name shapes outside the covered callback patterns.

### Notes
- I specifically checked the regex change for robustness. The new pattern is intentionally stricter than the old “starts with uppercase ASCII” check, which is what removes the false positive for names like `DropdownMenuExample.useCallback[renderSearchInput]`.
- The allow-list still covers common component-like names, including digits, `_`, and `$` suffixes, so it should behave well for normal authored and transpiled component identifiers.
- I did not find a concrete correctness, accessibility, API, maintainability, performance, or test-coverage issue in the branch diff.

PR #4128, opened 2026-02-18
[all components] Reset openMethod on close transition

atomiks:
- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

Fixes a race condition due to `useClick`'s `mousedown` being deferred in a rAF and the reset happening on close complete. Concrete bug: tapping the Select trigger quickly repeatedly sometimes had a `null` `openMethod` incorrectly.

PR #4325, opened 2026-03-13
[select] Preserve touch exit animations

atomiks:
## Summary

Fixes a Select regression where touch-triggered close animations could lose their exit styling during the close transition.

The change keeps the last non-null `openMethod` available through the close phase so touch-opened Selects do not switch layout modes mid-exit.

## Changes

- Preserve the previous `openMethod` during close using `usePreviousValue`.
- Clear the Select store's `openMethod` only after the popup fully unmounts.
- Add a Chromium regression test covering touch close transitions and `data-ending-style`.

## Notes

- This doesn't affect the `alignItemWithTrigger={false}` mode when explicitly set
- This also preserves the `null` openMethod bug fix from #4128

### DIFF
--- a/packages/react/src/select/root/SelectRoot.test.tsx
+++ b/packages/react/src/select/root/SelectRoot.test.tsx
@@ -861,6 +861,65 @@ describe('<Select.Root />', () => {
 
       expect(isScrollLocked()).toBe(false);
     });
+
+    it('keeps touch positioning during the close transition', async ({ onTestFinished }) => {
+      globalThis.BASE_UI_ANIMATIONS_DISABLED = false;
+
+      onTestFinished(() => {
+        globalThis.BASE_UI_ANIMATIONS_DISABLED = true;
+      });
+
+      const style = `
+        @keyframes select-close-test {
+          to {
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review

### Overview
This patch fixes the touch-specific Select close-animation regression without reintroducing the quick-reopen `openMethod` race. It keeps the last interaction type alive through the close transition, adds a Chromium regression test for that path, and the rebased version keeps the new test on Vitest-native assertions.

### Findings (None)

### Confidence: 4/5
High confidence based on a full-branch pass, a second sweep over render correctness, accessibility, API and state design, maintainability, performance risk, and coverage, plus fresh focused verification after reinstalling dependencies on the rebased branch. I reran `pnpm test:jsdom SelectRoot --no-watch` and `pnpm test:chromium SelectRoot --no-watch`, and both suites passed with the new touch-close regression coverage intact.

### Notes
- Latest delta since the previous review: after reinstalling dependencies on top of current `master`, the focused SelectRoot suites are passing again.
- The new touch close-transition test now uses Vitest-native matchers only.

PR #4327, opened 2026-03-13
[navigation menu] Fix initial trigger switch size snap

atomiks:
This is semi-hard to spot as it requires switching to a new trigger right as the popup's initial open animation is about to finish.

## Summary

Prevents NavigationMenu from snapping to the next panel size when switching triggers right as the initial open animation finishes. The pending size reset from the first open is now canceled before a new trigger-switch resize transition starts.

## Changes

- Abort pending popup auto-size resets before starting another size transition.
- Add a regression test covering the first trigger switch immediately after the initial open animation.


atomiks:
## Codex Review

### Overview
This patch fixes the first trigger-switch size snap by moving pending popup size-reset ownership up to the navigation menu root, so a stale animation callback from the previous trigger cannot win. It also adds a stronger regression test that keeps overlapping animations visible and wraps the async completion path in `act(...)`.

### Findings

#### 1. 🟡 [Non-blocking] Regression coverage still misses the reported hover path

**Impact**: The new test protects the shared reset logic in general, but it does not exercise the exact interaction path from the original bug report. A future regression that only affects hover timing could slip through while this test still passes.

**Evidence**: The user-facing report was about hovering between triggers, but the regression test opens and switches with `fireEvent.click(...)` in `packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx:2476-2485`. In `packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx:503-540`, hover and click do not take the same state-transition path: hover sets `isHover`, runs `changeState` inside `ReactDOM.flushSync(...)`, and also updates the patient-click bookkeeping before calling `setValue(...)`.

**Recommendation**: Add a hover-driven version of this regression, or switch the current test to deterministic hover events, so the test covers the exact timing path that originally reproduced the snap.

### Confidence: 4/5
High confidence based on a full-branch review across correctness, rendering/effect timing, accessibility/semantics, API/state design, maintainability/simplification, performance risk, and test quality. The latest CI on the current PR head also shows the relevant browser, lint, and regression jobs green; the main remaining risk is the hover-specific coverage gap above.

### Notes
- I used subagents for the category sweeps on React/rendering, API/state design, maintainability, performance, and test quality, then consolidated those results into this branch-wide review.
- I did not find a credible runtime performance regression here, so I did not recommend a benchmark for this patch.

PR #4328, opened 2026-03-13
[navigation menu] Add generic `Value` typing

atomiks:
## Summary

Adds generic value typing to `NavigationMenu.Root` so `value`, `defaultValue`, and `onValueChange` infer the selected item type the same way Accordion already does. This keeps the existing permissive default when no generic is provided while improving type safety for controlled and uncontrolled usage.

## Changes

- Add a `NavigationMenuValue<Value>` alias and thread the generic through `NavigationMenu.Root` props and namespace exports.
- Update the root context typing so the internal setter matches the public generic value shape.
- Add compile-time spec coverage for inference, explicit generics, nullable values, and the backward-compatible `any` default.
- Regenerate the Navigation Menu API reference metadata.


atomiks:
## Codex Review

### Overview
This patch adds generic value typing to `NavigationMenu.Root` so `value`, `defaultValue`, and `onValueChange` infer the selected item type instead of defaulting to `any`. It also adds compile-time coverage and updates the generated API reference so the docs show the simpler `Value | null` form.

### Findings (None)

### Confidence: 4/5
High confidence based on a full review of the branch diff, a second-pass sweep for render correctness, accessibility impact, API and state design, maintainability, performance risk, and test coverage, plus validation with `pnpm typescript`, `pnpm test:jsdom NavigationMenuRoot --no-watch`, `pnpm docs:api`, `pnpm eslint`, `pnpm stylelint`, and `pnpm markdownlint`.

### Notes
- The runtime behavior is unchanged; the patch is limited to public typing, compile-time assertions, internal context typing, and generated API metadata.
- I did not run a benchmark because the change does not alter hot-path runtime logic, layout work, or bundle size.

PR #4333, opened 2026-03-13
[popups] Fix outside press dismissal in shared shadow root

### LINKED ISSUE #4332: ShadowDOM: Popup don't dismiss when clicking other elements inside the same ShadowRoot
# Bug report

## Current behavior

Popups (Combobox, Popover, Menu) do not dismiss when clicking elements inside the same shadow root but outside the popup. However, clicking anywhere on the host page outside shadow root does correctly dismiss it.

From what I can tell, `event.target` at the document level gets retargeted to the shadow host by the browser, which seems to confuse the inside/outside detection. The dismiss just never fires for clicks within the shadow root.

## Expected behavior

Clicking outside a popup but within the same shadow root should dismiss it, same as in a regular document.

## Reproducible example

Any React app mounted inside a shadow root. Most common with browser extension content scripts using tools like WXT which use shadow DOM for style isolation.

In my case it's a WXT extension and the app is mounted inside a shadow root via `createShadowRootUi`, and all Base UI portals use the shadow root container:

```ts
const ui = await createShadowRootUi(ctx, {
  name: 'extension-ui',
  position: 'inline',
  anchor: 'body',
  onMount: (container) => {
    const app = document.createElement('div');
    container.append(app);
    const root = ReactDOM.createRoot(app);
    root.render(
      <PortalContainerProvider container={container}>
        <App />
      </PortalContainerProvider>,
    );
    return root;
  },
});
```

All portals point to `container` inside the shadow root:

```tsx
<Popover.Portal container={container}>
```

This produces the following HTML structure:

```html
<extension-ui>
  #shadow-root (open)
    <html>
      <body>
        <div>  <!-- React root -->
          ...app...
        </div>
        <div data-base-ui-portal>  <!-- portals land here, inside shadow body -->
          ...popup...
        </div>
      </body>
    </html>
</extension-ui>
```

## Proposed fix

Looks like the issue might be somewhere in `useDismiss`. I found a workaround using `event.composedPath()` which returns the real clicked element and is never retargeted by the browser. Maybe it could help for the fix too.

One workaround I found is a `useOutsidePress` hook used per component:

```ts
function useOutsidePress(
  enabled: boolean,
  refs: React.RefObject<HTMLElement | null>[],
  onOutsidePress: () => void,
) {
  const callbackRef = React.useRef(onOutsidePress);
  React.useLayoutEffect(() => {
    callbackRef.current = onOutsidePress;
  });

  React.useEffect(() => {
    if (!enabled) return;

    const handler = (e: PointerEvent) => {
      const path = e.composedPath();
      const isInside = refs.some(r => r.current && path.includes(r.current));
      if (!isInside) callbackRef.current();
    };

    document.addEventListener('pointerdown', handler, true);
    return () => document.removeEventListener('pointerdown', handler, true);
  }, [enabled, refs]);
}

// usage
const triggerRef = React.useRef<HTMLButtonElement>(null);
const popupRef = React.useRef<HTMLDivElement>(null);
const refs = React.useMemo(() => [triggerRef, popupRef], []);

useOutsidePress(open, refs, () => setOpen(false));
```

## Base UI version

v1.3.0

Chrome

## Which OS are you using?

All

## Additional context

Found #4230 but looks like it's not helping here still. Relevant older issue I guess #2420.

atomiks:
Fixes #4332

## Summary

This fixes outside press dismissal for popups rendered inside a shared `ShadowRoot`.

The old logic in `useDismiss` collected `data-base-ui-inert` markers from the owner document, and the previous shadow DOM follow-up also appended markers from the target's shadow root. That left the injected-element guard comparing a mixed set of markers from different roots.

For clicks on ordinary sibling elements inside the same `ShadowRoot`, the target's ancestor chain did not contain the document-level markers, so the click could still be treated as a third-party injected element and ignored. The new logic scopes the marker lookup to the event target's actual root, so same-shadow-root outside clicks are compared against the correct marker set and dismiss normally.

## Changes

- Scope the inert marker lookup in `useDismiss` to the event target's actual root instead of mixing document-level and shadow-root markers.
- Add a regression test that mounts the app inside a `ShadowRoot`, portals the dialog into that root, and verifies clicking a sibling element dismisses it.

### DIFF
--- a/packages/react/src/popover/root/PopoverRoot.test.tsx
+++ b/packages/react/src/popover/root/PopoverRoot.test.tsx
@@ -9,6 +9,7 @@ import { spy } from 'sinon';
 import { createRenderer, isJSDOM, popupConformanceTests, wait } from '#test-utils';
 import { OPEN_DELAY } from '../utils/constants';
 import { PATIENT_CLICK_THRESHOLD } from '../../utils/constants';
+import { REASONS } from '../../utils/reasons';
 
 describe('<Popover.Root />', () => {
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


atomiks:
## Codex Review

### Overview
This change fixes outside press dismissal for popups that are mounted and portaled within the same shadow root. It narrows the inert-marker lookup to the event target's actual root and now covers the shadow-root behavior at the hook level plus dialog and popover integration points.

### Findings (None)

### Confidence: 4/5
High confidence based on a full pass over the branch diff, the shared `useDismiss` code path, and focused regression coverage for same-root clicks, outside-the-root clicks, modal dialog behavior, and non-modal popover behavior.

### Notes
- Reviewed the branch against `master`, including correctness, accessibility, API/state, maintainability, performance risk, and test coverage.
- Verified with focused `useDismiss`, `DialogRoot`, and `PopoverRoot` runs in JSDOM and Chromium.

PR #4336, opened 2026-03-15
[all components] Reduce shared bundle size

atomiks:
## Summary

Reduces shared bundle cost across the popup, focus-management, and selector paths used by many Base UI components.

This PR keeps the measurable wins from the internal `tabbable` helper and the plain-vs-memoized selector split, trims shared utility code used across popup-heavy components, and avoids the broader unverified `@base-ui/utils/store/...` import-path churn so store consumers continue to use the standard barrel entrypoint.

## Changes

- Replace the external `tabbable` dependency with a smaller internal helper and add focused regression coverage for shadow DOM, `details`, hidden controls, radio groups, inert content, and embedded focusable elements.
- Split plain and memoized selector helpers so entrypoints that only need plain selectors do not pull memoized selector overhead.
- Remove redundant non-reactive Combobox store selector subscriptions.
- Simplify shared render, transition, animation, popup, and store utilities used across multiple components.

## Notes

- The latest published bundle bot result on this PR reports `@base-ui/react` down by `4.33 KB` parsed and `1.57 KB` gzip versus `master`.
- Latest targeted perf checks in exported production builds still favor the PR head on `combobox-perf` and `select-perf`, while a Chromium `100`-slider mount check is effectively flat versus `master`.
- Broader popup-route perf has been mixed rather than uniformly better, so this PR should be understood primarily as a shared bundle-size reduction with targeted runtime wins, not as an across-the-board popup perf improvement.

### DIFF
--- a/packages/react/src/floating-ui-react/utils/tabbable.test.ts
+++ b/packages/react/src/floating-ui-react/utils/tabbable.test.ts
@@ -15,6 +15,14 @@ it('includes basic tabbable controls and excludes hidden inputs', () => {
   expect(tabbable(document.body)).toEqual([button, input]);
 });
 
+it('includes embedded focusable elements in the tab order', () => {
+  const iframe = document.createElement('iframe');
+
+  document.body.append(iframe);
+
+  expect(tabbable(document.body)).toContain(iframe);
+});
+
 it('excludes disabled controls from the tab order', () => {
   const enabledButton = document.createElement('button');
   const disabledButton = document.createElement('button');
--- a/packages/react/src/floating-ui-react/utils/tabbable.ts
+++ b/packages/react/src/floating-ui-react/utils/tabbable.ts
@@ -6,7 +6,7 @@ import { isElementVisible } from './composite';
 export type FocusableElement = HTMLElement | SVGElement;
 
 const CANDIDATE_SELECTOR =
-  'a[href],button,input,select,textarea,summary,details,[tabindex],[contenteditable]:not([contenteditable="false"]),audio[controls],video[controls]';
+  'a[href],button,input,select,textarea,summary,details,iframe,object,embed,[tabindex],[contenteditable]:not([contenteditable="false"]),audio[controls],video[controls]';
 
 function getParentElement(element: Element) {
   const assignedSlot = (


atomiks:
## Codex Review

### Overview
This PR still delivers a real shared bundle-size win on the latest rebased head by shrinking the common popup, focus, and selector paths without leaning on speculative import churn. After rechecking the full branch and the follow-up fixes on the current head, I did not find a remaining correctness or runtime regression worth blocking on.

### Findings (None)

### Confidence: 4/5
High confidence from a fresh branch-wide pass plus new bundle snapshots against current `master`, exported-production popup benchmarks, a Chromium `Radio` / `Checkbox` / `Tabs` sanity pass, and focused local verification of the lint/test regressions that were still present on the previous pushed head.

### Notes
- Latest head reviewed: `96abcab71`. The follow-up on top of the earlier rebased branch restores `preventBaseUIHandler()` wrapping for initial prop objects in `mergeProps` and folds the duplicate store imports in `AriaCombobox` / `SelectRoot`, which clears the local `eslint:ci` failure and the `useRenderElement` regression from the prior head.
- The latest rebased head still keeps the measured wins and backs out the unverified `@base-ui/utils/store/*` path splitting, so the remaining size reduction comes from changes that still show up in the current snapshots.
- Latest local bundle-size deltas vs current `master` tip:

| Entry | Parsed delta | Gzip delta |
| --- | ---: | ---: |
| `@base-ui/react` | `-9.56 KB` | `-3.35 KB` |
| `@base-ui/react/combobox` | `-9.11 KB` | `-3.04 KB` |
| `@base-ui/react/menu` | `-8.71 KB` | `-3.01 KB` |
| `@base-ui/react/select` | `-8.61 KB` | `-2.95 KB` |
| `@base-ui/react/popover` | `-8.72 KB` | `-3.08 KB` |
| `@base-ui/react/tooltip` | `-4.91 KB` | `-1.76 KB` |
| `@base-ui/react/dialog` | `-8.61 KB` | `-3.11 KB` |
| `@base-ui/react/drawer` | `-8.60 KB` | `-3.06 KB` |
| `@base-ui/react/navigation-menu` | `-8.32 KB` | `-3.00 KB` |
| `@base-ui/react/tabs` | `-0.39 KB` | `-0.13 KB` |
| `@base-ui/react/slider` | `-0.37 KB` | `-0.11 KB` |
| `@base-ui/react/accordion` | `-0.44 KB` | `-0.15 KB` |
| `@base-ui/react/collapsible` | `-0.44 KB` | `-0.16 KB` |
| `@base-ui/react/toast` | `-5.38 KB` | `-1.86 KB` |

- Exported production docs benchmark, alternating PR head vs current `master`, 30 samples per ref on the contained and detached trigger pages:

| Benchmark | Ref | Mean | p99 | Samples |
| --- | --- | ---: | ---: | ---: |
| Contained triggers | PR head | `115.36 ms` | `150.6 ms` | `30` |
| Contained triggers | `master` | `115.07 ms` | `144.1 ms` | `30` |
| Detached triggers | PR head | `49.61 ms` | `60.1 ms` | `30` |
| Detached triggers | `master` | `48.12 ms` | `53.8 ms` | `30` |

- Supplementary 1,000-item JSDOM mount sanity check for `Radio`, `Slider`, and `Tabs`, 15 runs per ref:

| Benchmark | Ref | Mean | p99 | Samples |
| --- | --- | ---: | ---: | ---: |
| Radio ×1000 | PR head | `592.85 ms` | `701.79 ms` | `15` |
| Radio ×1000 | `master` | `603.05 ms` | `719.28 ms` | `15` |
| Slider ×1000 | PR head | `1298.80 ms` | `1479.82 ms` | `15` |
| Slider ×1000 | `master` | `1323.03 ms` | `1528.16 ms` | `15` |
| Tabs ×1000 | PR head | `396.46 ms` | `552.33 ms` | `15` |
| Tabs ×1000 | `master` | `416.41 ms` | `618.72 ms` | `15` |

- Latest focused verification on the reviewed head:
  - `pnpm --filter @base-ui/utils --filter @base-ui/react build`
  - `pnpm eslint:ci`
  - `pnpm stylelint`
  - `pnpm markdownlint`
  - `pnpm test:jsdom useRenderElement tabbable FloatingFocusManager SelectRoot ComboboxList Toast MenuTrigger PopoverTrigger PopoverRoot --no-watch`
  - `pnpm test:chromium tabbable FloatingFocusManager SelectRoot MenuTrigger PopoverTrigger PopoverRoot --no-watch`

atomiks:
Size snapshot summary from the Code Infra comparison page for this exact head commit:
https://frontend-public.mui.com/size-comparison/mui/base-ui/diff?prNumber=4336&baseRef=master&baseCommit=4fd87cad4c25c32c2e143c51089fe50d030dd442&headCommit=2239847c9cf7ff63c9d358cf2e321d7f474a7b8b

Root package:
- `@base-ui/react`: `-9.27 KB` parsed, `-3.44 KB` gzip, final gzip `148 KB`

Largest component-entrypoint gzip wins:
- `@base-ui/react/navigation-menu`: `-3.31 KB`, final gzip `36.3 KB`
- `@base-ui/react/dialog`: `-3.28 KB`, final gzip `20.6 KB`
- `@base-ui/react/alert-dialog`: `-3.27 KB`, final gzip `20.4 KB`
- `@base-ui/react/drawer`: `-3.26 KB`, final gzip `32.8 KB`
- `@base-ui/react/popover`: `-3.23 KB`, final gzip `37.7 KB`
- `@base-ui/react/menu`: `-3.18 KB`, final gzip `49.7 KB`
- `@base-ui/react/context-menu`: `-3.16 KB`, final gzip `46.1 KB`
- `@base-ui/react/combobox`: `-3.06 KB`, final gzip `46.7 KB`
- `@base-ui/react/autocomplete`: `-3.04 KB`, final gzip `45.4 KB`
- `@base-ui/react/select`: `-3.03 KB`, final gzip `43.6 KB`

Other notable requested components:
- `@base-ui/react/toast`: `-1.92 KB` gzip, final gzip `24.2 KB`
- `@base-ui/react/tooltip`: `-1.74 KB` gzip, final gzip `30.8 KB`
- `@base-ui/react/slider`: `-75 B` gzip, final gzip `12.4 KB`
- `@base-ui/react/accordion`: `-82 B` gzip, final gzip `8.63 KB`
- `@base-ui/react/collapsible`: `-70 B` gzip, final gzip `6.53 KB`
- `@base-ui/react/tabs`: `-62 B` gzip, final gzip `12.2 KB`

So the headline is still that the shared-internals work materially helps the heavy popup/focus components, while the simpler leaf components move by much smaller amounts.

michaldudak:
### Code review

Found 3 issues (scored 75/100 confidence -- below normal posting threshold, posting by request):

1. New test file uses Chai-style assertions (`.to.equal()`) instead of vitest-native matchers (`.toBe()`). Two tests are inconsistent with the rest of the file, which already uses vitest-native style. (AGENTS.md says "The repository is transitioning from `chai` and `sinon`, prefer `vitest` native functions for all new code.")

https://github.com/mui/base-ui/blob/d2e76b45dc0d46913cffa7c71e41ce28f7411c63/packages/react/src/floating-ui-react/utils/tabbable.test.ts#L56-L58
https://github.com/mui/base-ui/blob/d2e76b45dc0d46913cffa7c71e41ce28f7411c63/packages/react/src/floating-ui-react/utils/tabbable.test.ts#L70-L72

2. Toast store selector memoization uses module-level closure variables (`lastToasts`, `lastMap`), creating a single shared cache slot. If multiple `ToastProvider` instances coexist, they will thrash each other's cache on every render, returning new Map objects and breaking React memoization. The previous `createSelectorMemoized` used a per-instance `WeakMap` keyed by `state.__cacheKey__`. This is a performance regression, not a correctness bug.

https://github.com/mui/base-ui/blob/d2e76b45dc0d46913cffa7c71e41ce28f7411c63/packages/react/src/toast/store.ts#L34-L66

3. Removed `// We need to mutably own 'merged'` comments in `mergeProps.ts` and `mergePropsN`. The spread on those lines is critical -- without it, `mutablyMergeInto` would corrupt the shared `EMPTY_PROPS` constant. The comment protected against future maintainers removing the "redundant-looking" spread.

https://github.com/mui/base-ui/blob/d2e76b45dc0d46913cffa7c71e41ce28f7411c63/packages/react/src/merge-props/mergeProps.ts#L63-L65
https://github.com/mui/base-ui/blob/d2e76b45dc0d46913cffa7c71e41ce28f7411c63/packages/react/src/merge-props/mergeProps.ts#L103-L105

atomiks:
## Codex Review

### Overview
This is a correctness-only follow-up on the latest PR head. I re-reviewed the full branch against current `master` with subagents focused on render/focus behavior, accessibility, API/state design, maintainability, and test coverage, and one non-blocking correctness issue still stands out.

### Findings
#### 1. 🟡 [Non-blocking] `mergeProps` now mutates the first props-getter result in place

**Impact**: A props getter used as the first input can no longer safely return a reused object. Later merges now mutate that returned object directly, which can leak merged handlers, `className`, or `style` into shared state across renders and make behavior depend on call order.

**Evidence**: `packages/react/src/merge-props/mergeProps.ts:63-119` now seeds `merged` with `mergeInto({}, a)` / `mergeInto({}, props[0])`. When `a` or `props[0]` is a function, `mergeInto()` returns `inputProps(merged)` directly, and later `mutablyMergeInto()` calls update that same object in place. A direct repro with `mergeProps(() => shared, { className: 'next' })` changes both `merged.className` and `shared.className` to `next base`. The previous implementation shallow-copied the first getter result before any later mutation.

**Recommendation**: Restore a shallow copy for the first getter-returned props object before later merges, or otherwise guarantee that getter results are never mutated in place. I would also add a regression test that reuses the same getter-returned object across calls and proves it stays uncontaminated.

#### 2. ℹ️ [Informational] The toast selector rewrite still lacks multi-provider isolation coverage

**Impact**: The current branch replaces the old memoized selector implementation in the toast store, but the test suite still does not prove that two independent toast providers remain isolated. That leaves the new cache behavior riding on indirect coverage only.

**Evidence**: `packages/react/src/toast/store.ts:34-78` now uses a hand-rolled `WeakMap` cache keyed by `state.__cacheKey__`. The new branch tests are focused on `tabbable` and `FloatingFocusManager`; I did not find a toast regression test that mounts two `Toast.Provider` or manager instances at once and asserts updates in one provider do not perturb the other provider’s derived selector state.

**Recommendation**: Add a regression test around `Toast.Provider` or `useToastManager` that renders two independent providers/managers, updates one, and verifies the other provider’s selector outputs stay stable.

#### 3. ℹ️ [Informational] The `useRenderElement` ref-merging rewrite still lacks rerender coverage

**Impact**: The ref path simplification is now covered for static cases, but not for the risky rerender transitions. That means a future ref-shape or enabled-toggle regression could still slip through even though the branch changes this logic substantially.

**Evidence**: `packages/react/src/utils/useRenderElement.tsx:78-98` rewrites the old conditional `useMergedRefs` / `useMergedRefsN` split to a single `useMergedRefsN(refs)` flow. `packages/react/src/utils/useRenderElement.test.tsx:56-188` covers disabled short-circuiting and event preventability, but I did not find a rerender test that toggles `enabled` across renders or switches between a single ref and a ref array.

**Recommendation**: Add a rerender-driven regression test that exercises `enabled` false→true / true→false transitions and single-ref ↔ ref-array changes, then asserts the merged ref and event wiring remain correct.

### Confidence: 4/5
High confidence for a correctness-focused review. I covered the full branch again with subagent sweeps across the major correctness categories, cross-checked the main hot spots locally, and validated the `mergeProps` issue with a direct repro.

### Notes
- This is a correctness-only follow-up comment on top of the earlier size/perf review, so I intentionally left the previous measurement tables untouched.
- Category sweeps came back clean for the remaining accessibility/focus paths in `tabbable` / `FloatingFocusManager`, the selector split in `packages/utils/src/store`, and the current non-reactive Combobox store reads.

LukasTy:
## Claude Opus 4.6 review

### Previously flagged issues — status check:

| Issue | Status |
|-------|--------|
| **Chai-style assertions** (`.to.equal()`) in tabbable tests | **Fixed** — all tests now use vitest-native matchers (`toBe`, `toContain`, etc.) |
| **Toast selector multi-provider isolation** | **Fixed** — now uses `WeakMap` keyed by `state.__cacheKey__` instead of module-level closure variables. Multiple providers each get their own cache entry |
| **Removed safety comments** in `mergeProps` about owning `merged` | **Fixed** — the comment `// We need to mutably own \`merged\`.` is now present above both [the spread in `mergeProps`](packages/react/src/merge-props/mergeProps.ts#L64) and [in `mergePropsN`](packages/react/src/merge-props/mergeProps.ts#L102) |
| **`mergeProps` first-getter mutation** | **NOT fully fixed** — see Finding #1 below |
| **`useRenderElement` ref-merging rerender coverage** | **Not addressed** — no new rerender-driven tests added, but the structural change (always calling `useMergedRefsN` unconditionally) is actually *safer* than the old conditional hook pattern |

---

### New concerns found in this review:

#### 1. `mergeProps` — getter-as-first-argument still mutates the returned object

In mergeProps.ts, the new `mergeInto` function:

```ts
function mergeInto(merged, inputProps) {
  if (typeof inputProps === 'function') {
    return inputProps(merged);  // no shallow copy
  }
  return mutablyMergeInto(merged, inputProps);
}
```

When the first argument is a getter, `mergeInto({}, getter)` calls `getter({})` and returns the result directly. Subsequent `mergeInto(merged, b)` calls then mutate that result in place via `mutablyMergeInto`. The old `createInitialMergedProps` did `{ ...resolvePropsGetter(inputProps, EMPTY_PROPS) }` — a shallow copy.

**Practical risk: Low.** I audited all call sites. No internal call passes a getter as the first argument to `mergeProps`. The only path where a getter could appear as first-in-array is via `mergePropsN` in `useRenderElement`, and those getters always return new objects. Still, it's a documented behavioral contract regression from the old implementation.

#### 2. `useAnimationsFinished` — missing `signal?.aborted` guard on recursive `exec()`

In useAnimationsFinished.ts, the refactored catch block:

```ts
.catch(() => {
  if (treatAbortedAsFinished) {
    if (!signal?.aborted) { done(); }
    return;
  }
  // OLD: had `!signal?.aborted &&` before this check
  const currentAnimations = resolvedElement.getAnimations();
  if (currentAnimations.length > 0 && currentAnimations.some(...)) {
    exec(); // recurses even when signal is aborted
  }
});
```

The old code had `!signal?.aborted &&` before the recursive `exec()` call. The new code drops this guard, meaning if `treatAbortedAsFinished` is false and the signal is aborted, `exec()` will still recurse uselessly (though `done()` will never fire since the `.then()` handler checks `!signal?.aborted`). Not a correctness bug, but wastes CPU and could loop unnecessarily.

#### 3. `tabbable` — no tabIndex sorting (positive tabIndex elements not sorted before tabIndex=0)

The old `tabbable` library sorted positive-tabIndex elements before zero-tabIndex ones per browser spec behavior. The new implementation returns elements strictly in DOM order, regardless of tabIndex value. 

**Practical risk: Negligible.** I searched the entire codebase and found no positive tabIndex usage — everything uses 0 or -1. However, missing a defensive test means a future regression could slip in.

#### 4. `useStableCallback` — `useRefWithInit` usage with a ref argument

The refactored useStableCallback.ts passes `callbackRef` (from `React.useRef`) into `useRefWithInit` as an init argument. This is correct since `useRefWithInit` only runs the init function once, and the trampoline closes over `callbackRef` (a stable ref object). No behavioral regression — verified the initial-render-phase protection (`assertNotCalled`) still works.

#### 5. `useRenderElement` — unconditional `useMergedRefsN` is an improvement

The old code had a rules-of-hooks violation: `useMergedRefs` / `useMergedRefsN` was called conditionally based on `enabled`, `typeof document`, and whether `ref` was an array. The new code always calls `useMergedRefsN(refs)` unconditionally with an empty array when disabled. This is structurally better and eliminates the rules-of-hooks concern.

#### 6. `useTransitionStatus` — `useMemo` removal is safe

All callers immediately destructure the return value. Object identity doesn't matter since the individual values are primitives/stable setState refs.

#### 7. Selector split — works correctly

`createSelector` is split to its own subpath (`@base-ui/utils/store/createSelector`) and the barrel `@base-ui/utils/store` still re-exports both `createSelector` and `createSelectorMemoized`. The `exports` field in package.json has the new `"./store/createSelector"` entry. Components that only need plain selectors (no memoization) can skip pulling in `reselect`. Toast store correctly does its own WeakMap-based memoization.

---

### Summary

| Finding | Severity | Blocking? |
|---------|----------|-----------|
| `mergeProps` getter-first mutation | Low | No (not hit internally) |
| `useAnimationsFinished` missing abort guard | Low | No (wastes CPU, no visible bug) |
| `tabbable` no tabIndex sorting | Low | No (not used) |
| Chai assertions fixed | - | Resolved |
| Toast multi-provider fixed | - | Resolved |
| mergeProps safety comments restored | - | Resolved |

**Overall assessment:** The previously flagged issues are addressed. No blocking regressions found. The two new concerns (#1 and #2) are low-risk but worth noting — #2 in particular is a simple one-line fix (`if (!signal?.aborted && currentAnimations.length > 0 && ...)`).

packages/utils/src/useStableCallback.ts
```diff
@@ -36,22 +27,16 @@ type Stable<T extends Callback> = {
  * This hook is a more permissive version of React 19.2's `React.useEffectEvent` in that it can be passed through contexts and called in event handler props, not just effects.
  */
 export function useStableCallback<T extends Callback>(callback: T | undefined): T {
```
flaviendelangle, line 38: What is the reasoning behind this change?
We create more objects on every render than before.

atomiks: It was part of a broader sweep of size changes, but yeah it's a 1-3% perf regression from testing, undid it

packages/utils/package.json
```diff
@@ -10,6 +10,7 @@
   "license": "MIT",
   "exports": {
     "./store": "./src/store/index.ts",
+    "./store/createSelector": "./src/store/createSelector.ts",
```
romgrk: Should it also export `createSelectorMemoized`?

packages/react/src/utils/useRenderElement.tsx
```diff
@@ -78,23 +78,23 @@ function useRenderElementProps<
...
+      refs.push(ref);
     }
   }
 
+  const mergedRef = useMergedRefsN(refs);
+
+  if (mergedRef !== null) {
+    outProps.ref = mergedRef;
+  }
+
```
romgrk: I don't love the idea of packaging this as part of a PR named "reduce shared bundle size" considering that it also affects behavior/performance.

Also I don't think this reduces bundle-size (considering this is the only call-site for `useMergedRefsN` and `useMergedRefs` is used throughout the rest of the codebase).

packages/react/src/utils/useRenderElement.tsx
```diff
@@ -174,10 +174,14 @@ function evaluateRenderProp<T extends React.ElementType, S>(
...
-    if (typeof element === 'string') {
-      return renderTag(element, props);
+  if (typeof element === 'string') {
+    if (element === 'button') {
+      return <button type="button" {...props} key={props.key} />;
+    }
+    if (element === 'img') {
+      return <img alt="" {...props} key={props.key} />;
     }
+    return React.createElement(element, props);
```
romgrk: Is this really a clear win? We win a few bytes, but the readability decreases because the function name is lost, and the cyclomatic complexity of this function is increased.

If we really want to save bytes like this, I'd rather we write a babel plugin to inline single-use function automatically rather than affect readability.

packages/react/src/utils/usePopupViewport.tsx
```diff
@@ -85,7 +84,8 @@ export function usePopupViewport(parameters: UsePopupViewportParameters): UsePop
   const popupElement = store.useState('popupElement');
   const positionerElement = store.useState('positionerElement');
 
-  const previousActiveTrigger = usePreviousValue(open ? activeTrigger : null);
```
romgrk, line 88: Why is it inlined here but not at the only other place it's used (`SelectRoot`)? Is this really a win?

packages/react/src/toast/store.ts
```diff
@@ -25,15 +26,36 @@ export type State = {
   prevFocusElement: HTMLElement | null;
 };
 
-const toastMapSelector = createSelectorMemoized(
```
romgrk, line 35: Inlining `createSelectorMemoized` feels wrong?

packages/react/src/select/root/SelectRoot.tsx
```diff
@@ -9,7 +9,7 @@ import { useControlled } from '@base-ui/utils/useControlled';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { useValueAsRef } from '@base-ui/utils/useValueAsRef';
-import { useStore, Store } from '@base-ui/utils/store';
+import { Store, useStore } from '@base-ui/utils/store';
```
romgrk: Seems like PR noise that could be avoided. If it's intentional, it should probably be an eslint rule to automate it.

romgrk (changes requested):
Feedback on the generated code: there are substantial issues with some of the changes here. Bundle-size is optimized at the cost of readability, performance and maintainability.

packages/react/src/toast/store.ts
atomiks: I thought so too but it wins 1.5 KB for toast-only imports:

> So this one change is worth about 1.4-1.5 KB gz, which is pretty meaningful for a single local refactor.
> 
> Why it saves that much:
> 
> putting createSelectorMemoized back in store.ts pulls the heavier reselect path back into the toast bundle the current code only needs a very small per-store WeakMap cache, so the bespoke version is much cheaper

atomiks:
@romgrk addressed the review comments

(Also with the useRenderElement changes I couldn't detect a difference in runtime perf running benchmarks locally with Codex. But I reverted them anyway as bundle size difference is negligible).

packages/react/src/toast/store.ts
romgrk: How common are toast-only imports? Is that a real use-case?

If it's a real use-case, this could instead be an opportunity to create a `createMemoizedSelectorLite` and put that code where it belongs (with the Store code, so if there is an implementation change, it's modified at the same time).

atomiks: I generally think it's good for each component to only have as much code as it really needs to. It makes treeshaking as effective as possible throughout the codebase. And yeah I think it's a real use case given you can use any number of Base UI components you want.

I'll look into the `Lite` idea

atomiks: > I'll look into the Lite idea

I think this is a good follow-up idea but not worth the abstraction for this PR because only Toast uses it. It'd be valid to do so if we need it elsewhere later on.

romgrk: I disagree with how valid the use-case is. You're assuming that some users are importing `Toast` without importing any other component that uses `createSelectorMemoized`, this is not a reasonable assumption to make out of the blue and is not supported by any data.

Internal implementation details of `createSelectorMemoized` need to be known by the `Toast` component, this is a poor choice for maintainability. A developper or agent doing work on `createSelectorMemoized` would have no way to know that they need to update `Toast` as well.

atomiks: > Internal implementation details of createSelectorMemoized need to be known by the Toast component, this is a poor choice for maintainability. A developper or agent doing work on createSelectorMemoized would have no way to know that they need to update Toast as well.

Fair enough point. The `Lite` abstraction would make this clear upfront.

> I disagree with how valid the use-case is

For bigger apps that use a broad range of components, it's not relevant of course, but in that case bundle size is less relevant anyway. If users really only want a few components (like Toast, Slider, Tabs, no popup ones, which seems valid to me) the bundle size win is a reasonable tradeoff here imo. All in all not _critical_ to keep this, and I wouldn't mind reverting it, but if we add the Lite abstraction, the maintainability tradeoff should be gone?

romgrk: The maintainability trade-off is less bad with the lite version, but **this actually makes bundle-size worst**. Now, instead of having just `createSelectorMemoized`'s code in their bundle, most users will import both `createSelectorMemoized` and `createSelectorMemoizedLite`, except in the extremely rare case in which they're only importing `Toast`. You're micro-optimizing one edge-case at the expense of the rest.

atomiks: I would agree if the Lite implementation was bigger but it's small enough that the tradeoff is possibly still worth it imo. We already do this splitting with FloatingPortal (FloatingPortalLite + regular one).

romgrk: You're making the assumption _"a non-negligible portion of our users are importing `Toast` without importing other components"_. Can you back that assumption with a rationale and/or data? Can you explain why you think the tradeoff is worth it?

The main problem is that if your assumption is false, the change negatively affects bundle-size for our users.

mnajdova: > The maintainability trade-off is less bad with the lite version, but this actually makes bundle-size worst. Now, instead of having just createSelectorMemoized's code in their bundle, most users will import both createSelectorMemoized and createSelectorMemoizedLite, except in the extremely rare case in which they're only importing Toast. You're micro-optimizing one edge-case at the expense of the rest.

Agree with this, I don't see the benefit for only optimizing for the single case of importing only Toast component.

atomiks: It's not just about importing Toast alone - it's Toast + any other components that don't use the store/reselect deps. For example, an import setup of Toast + Tabs + Slider + Checkbox + Collapsible + Toolbar wouldn't pull in the extra 1.5 kB because it doesn't have to.

That said I'm not gonna die on this hill 😅 and see that the store-based comps are popular enough to usually be used somewhere in an app anyway. I'll just revert the changes here so we can merge it as the 3+ kB total benefit is clearly worth it regardless of the Toast change.

atomiks: > see that the store-based comps are popular enough to usually be used somewhere in an app anyway

Actually I was mistaken about the import graph here. It's _only_ Toast (and Calendar later) that uses the reselect dep. So it pulls in 1.5 kB which increases the root bundle as well. It seems like reselect is needlessly heavy for our use cases.

With the "Lite" implementation: 

Bundle | Gzip size
-- | --
@base-ui/react | ▼-3.3KB


Without

Bundle | Gzip size
-- | --
@base-ui/react | ▼-1.7KB

flaviendelangle: I think we will have more and more components that use the `createSelectorMemoized`.
So if we can make it lighter, that's awesome :ok_hand:

atomiks: Just experimented with this, and Codex wrote one faster than reselect in some cases but not in others (5 selectors, which Calendar has). So I guess because Calendar needs it, we'll have to leave it in the bundle.

atomiks: Bundle size decreases are still better than 1.7 KB if you don't use Toast (or later Calendar):

https://frontend-public.mui.com/size-comparison/mui/base-ui/diff?prNumber=4336&baseRef=master&baseCommit=9ebce1a886ce9eb69fc49e1b372cec66884b3ba9&headCommit=fd95a435f78269040ff859a19d9c77bd50f3867b

romgrk: It sounds like we can make wins on the reselect front, but it feels like those improvements could be left as a follow-up PR considering they require more examination & investigation.

packages/react/src/dialog/store/DialogStore.ts
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { createSelector, ReactStore } from '@base-ui/utils/store';
+import { ReactStore } from '@base-ui/utils/store';
+import { createSelector } from '@base-ui/utils/store/createSelector';
```
romgrk: This type of change is noise, not improvements. It would be nice to avoid them in future PRs.

packages/react/src/menu/root/MenuRoot.tsx
```diff
@@ -185,14 +185,14 @@ export const MenuRoot = fastComponent(function MenuRoot<Payload>(props: MenuRoot
     }
   }
 
+  const { openMethod, triggerProps: interactionTypeProps } = useOpenInteractionType(open);
```
romgrk: Also PR noise.

packages/react/src/merge-props/mergeProps.ts
```diff
@@ -102,52 +99,32 @@ export function mergePropsN<T extends ElementType>(props: InputProps<T>[]): Prop
...
-    return inputProps(merged);
-  }
-  return mutablyMergeInto(merged, inputProps);
-}
-
 function createInitialMergedProps<T extends ElementType>(inputProps: InputProps<T>) {
-  if (isPropsGetter(inputProps)) {
-    // Getter-returned handlers intentionally keep their existing semantics.
-    return { ...resolvePropsGetter(inputProps, EMPTY_PROPS) };
+  if (typeof inputProps === 'function') {
```
romgrk: This change negatively impacts readability. Before, the concept of "props getter" is made explicit with a function name. After, it's gone.

packages/react/src/merge-props/mergeProps.ts
```diff
@@ -102,52 +99,32 @@ export function mergePropsN<T extends ElementType>(props: InputProps<T>[]): Prop
...
+    return { ...inputProps(EMPTY_PROPS as PropsOf<T>) };
   }
 
-  return copyPropsWithWrappedEventHandlers(inputProps);
+  return mutablyMergeInto({}, inputProps);
 }
 
-function copyPropsWithWrappedEventHandlers<T extends ElementType>(
```
romgrk, line 139: This is deleted because the behavior is now handled inside `mutablyMergeInto`, right?

packages/react/src/toast/store.ts
```diff
@@ -34,6 +34,7 @@ const toastMapSelector = createSelectorMemoized(
     >();
     let visibleIndex = 0;
     let offsetY = 0;
+
```
romgrk: PR noise.

packages/utils/src/detectBrowser.ts
romgrk, line 1: I'm having some doubts about these changes. I wonder if you shouldn't be looking for a more generic solution instead of affecting the readability of our codebase like this. Isn't this kind of transformation something that google's closure compiler could do as a build step?

packages/react/src/merge-props/mergeProps.ts
atomiks: Yep, `copyPropsWithWrappedEventHandlers()` was deleted because that behavior is now covered by `mutablyMergeInto({}, inputProps)`. Starting from `{}` gives us an owned copy, and `mutablyMergeInto()` still wraps event handlers while merging

packages/utils/src/detectBrowser.ts
atomiks: (This was resolved earlier as I've reverted it, not worth the readability cost.)

packages/react/src/floating-ui-react/components/FloatingPortal.tsx
```diff
@@ -214,10 +221,11 @@ export const FloatingPortal = React.forwardRef(function FloatingPortal(
   }, [portalNode, modal]);
 
   React.useEffect(() => {
-    if (!portalNode || open) {
+    if (!portalNode || open !== false) {
```
romgrk, line 224: This doesn't feel like a bundle-size change, if I'm not mistaken. It would be really nice to cleanly split "bundle-size only changes" from behavior changes or bug fixes, focusing a PR on a single issue eases the cognitive load.

atomiks: These changes were from trying to improve perf or keep parity following the changes since the tabbable logic is now quite different. I'll try to run some more tests to see if it's still useful given a lot has changed since the last benchmarks

atomiks:
## Codex Review (GPT-5.4)

This remains a `performance` PR with secondary `tests`, `a11y`, and `react-ui` changes. I re-reviewed the latest pushed head `b101e3fa7` against `upstream/master` `463b2ca4b`, including parallel focus/correctness and perf/maintainability sweeps, and I no longer see a remaining blocking correctness issue on the current branch state.

### 1. Remaining Bugs / Issues
No blocking or non-blocking correctness issues found on the current pushed head.

The earlier embedded-element `tabbable()` regression is now fixed, and I am not treating positive-`tabIndex` ordering as a remaining issue here because that divergence is intentionally unsupported in this PR.

### 2. Validation And Perf Snapshot
Latest focused validation on the current pushed head:

- `pnpm test:jsdom tabbable FloatingFocusManager --no-watch`
- `pnpm typescript`
- `pnpm exec eslint packages/react/src/floating-ui-react/utils/tabbable.ts packages/react/src/floating-ui-react/utils/tabbable.test.ts packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx --max-warnings 0`

Current supporting perf context from this review cycle:

| Check | PR head | `master` | Readout |
| --- | ---: | ---: | --- |
| Bundle bot `@base-ui/react` | `-4.33 KB` parsed / `-1.57 KB` gzip | baseline | Real bot result from this review cycle |
| `combobox-perf` (auto index) | `64.23 ms` mean / `71.39 ms` p99 | `69.90 ms` / `73.59 ms` | Favorable to PR head |
| `select-perf` | `76.95 ms` mean / `79.92 ms` p99 | `78.61 ms` / `82.22 ms` | Slightly favorable to PR head |
| Chromium `100`-slider mount | `88.20 ms` mean / `93.04 ms` p99 | `88.67 ms` / `91.77 ms` | Effectively flat |
| `contained-triggers` | `121.28 ms` mean / `126.20 ms` p99 | `126.40 ms` / `127.30 ms` | Modestly favorable to PR head |
| Chromium giant popover forward-tab through popup and out | `431.64 ms` mean / `471.20 ms` p99 | `1130.86 ms` / `1227.60 ms` | Strongly favorable to PR head |

That last popup-tabbing benchmark used the exported `popup-tabbing` experiment page with an injected giant DOM: `600` added focusables inside the open popover and `800` added focusables outside it. The timer ran in the browser from the first real `Tab` keydown on `Inside before` until focus reached the next outside target, with `30` sequential alternating Chromium samples per ref and no concurrent runs.

I also reran `contained-triggers` on the exact latest pushed head using the exported production page, `10` samples per ref, all PR-head samples first and then all `master` samples, with no concurrent measuring. The broader popup perf story therefore still looks favorable-to-mixed rather than uniformly better, but this latest contained pass did come out modestly favorable to the PR head.

### 3. Previous Review Issues Status
| --- | --- | --- |
| Embedded focusable elements omitted from `tabbable()` | Fixed | `iframe` / `object` / `embed` are back in the candidate selector, with regression coverage. |
| `mergeProps` getter-first mutation | Fixed | The first getter result is owned again, with regression coverage. |
| `useAnimationsFinished` abort retry path | Fixed | The retry path now bails when already aborted. |
| Toast multi-provider isolation coverage | Fixed | Explicit multi-provider coverage was added while keeping `createSelectorMemoized`. |
| `useRenderElement` rerender coverage gap | Fixed | Rerender-driven coverage was added for `enabled` toggles and single-ref/array-ref transitions. |
| `MenuPopup` custom focus regression | Fixed | The `initialFocus` experiment and its stale branch-only test were both reverted. |

### Confidence
`4/5`. The branch still looks directionally good on size and on the targeted `Select` / `Combobox` perf routes, and the highest-confidence accessibility regression I had previously flagged is now fixed. Residual risk is mostly the usual browser-edge maintenance burden that comes with the custom `tabbable` replacement, rather than a specific remaining bug I can point to in the latest head.

ciampo:
flagging that the refactor from `tabbable` to first-party checks that rely `checkVisibility` caused us a regression when updating to base ui 1.4.

I shared more details (and a couple of potential solutions) in https://github.com/mui/base-ui/issues/4622

PR #4338, opened 2026-03-15
[accordion][Collapsible] Simplify shouldRender condition in AccordionPanel and CollapsiblePanel

sai6855:
Simplify shouldRender condition in AccordionPanel and CollapsiblePanel


packages/react/src/accordion/panel/AccordionPanel.tsx
```diff
@@ -156,7 +156,7 @@ export const AccordionPanel = React.forwardRef(function AccordionPanel(
     stateAttributesMapping: accordionStateAttributesMapping,
   });
 
-  const shouldRender = keepMounted || hiddenUntilFound || (!keepMounted && mounted);
+  const shouldRender = keepMounted || hiddenUntilFound || mounted;
```
sai6855, line 159: When `shouldRender` is evaluated, if `keepMounted` were true, the expression would already short-circuit. So `!keepMounted` is always true by the time the third operand is reached, making it redundant.

atomiks (approved):

PR #4343, opened 2026-03-16
[docs][toggle group] Add multiple demo

aarongarciah:

https://deploy-preview-4343--base-ui.netlify.app/react/components/toggle-group#multiple

Apart from the demo being added, a couple of border radius are fixed on the ToggleGroup and Tooltip hero Tailwind demos so the radius of the hover state matches the one in the CSS Modules counterpart:
- https://deploy-preview-4343--base-ui.netlify.app/react/components/toggle-group
- https://deploy-preview-4343--base-ui.netlify.app/react/components/tooltip

The new demo reuses the same as icons the Tooltip hero demo. I tried to simplify the demo by not using icons but our current font doesn't support italics, and a serif font looked tough. In fact, markdown content that uses italics is not rendering italics. We'll fix this with future UI changes.

### DIFF
--- a/docs/src/app/(docs)/react/components/toggle-group/demos/hero/css-modules/index.module.css
+++ b/docs/src/app/(docs)/react/components/toggle-group/demos/hero/css-modules/index.module.css
@@ -29,12 +29,6 @@
     outline-offset: -1px;
   }
 
-  @media (hover: hover) {
-    &:hover {
-      background-color: var(--color-gray-100);
-    }
-  }
-
   &:active {
     background-color: var(--color-gray-200);
   }
--- a/docs/src/app/(docs)/react/components/toggle-group/demos/hero/tailwind/index.tsx
+++ b/docs/src/app/(docs)/react/components/toggle-group/demos/hero/tailwind/index.tsx
@@ -11,21 +11,21 @@ export default function ExampleToggleGroup() {
       <Toggle
         aria-label="Align left"
         value="left"
-        className="flex size-8 items-center justify-center rounded-sm text-gray-600 select-none hover:bg-gray-100 focus-visible:bg-none focus-visible:outline-2 focus-visible:-outline-offset-1 focus-visible:outline-blue-800 active:bg-gray-200 data-[pressed]:bg-gray-100 data-[pressed]:text-gray-900"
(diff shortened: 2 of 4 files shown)


flaviendelangle:
Claude Code review:
The only real complain he has is regarding the `focus-visible:bg-none` but since we use it a lot I guess there is a reason :eyes: 

Issues Found
1. focus-visible:bg-none is not a valid Tailwind class (minor bug)
Both the new Tailwind demo and the existing hero demo use focus-visible:bg-none. bg-none is not a standard Tailwind utility and generates no CSS — so this has no effect. It looks like the intent was to reset the background on focus-visible to override hover styles. This pre-existing bug is being copied into the new file, worth fixing here.

2. rounded-xs → rounded-sm fix is correct
In Tailwind v4, rounded-xs = 2px, rounded-sm = 4px (0.25rem). The CSS Modules version uses border-radius: 0.25rem, so rounded-sm is the right match. Good catch.

3. Section ordering: "Examples" placed before "Anatomy"
In other component docs pages (accordion, dialog, etc.), "Examples" typically comes after "Anatomy" and "API reference". Worth confirming this ordering is intentional or aligns with the new docs structure convention.

4. defaultValue pre-selects two items — undocumented
Using defaultValue={['bold', 'italic']} immediately shows multi-selection, which is great. A short prose note like "Use defaultValue to set initially pressed items" would help readers scanning the page.

5. SVG icon duplication (minor, non-blocking)
BoldIcon, ItalicIcon, UnderlineIcon are now copied across 4 files (tooltip hero × 2, toggle-group multiple × 2). Demo files being self-contained is a reasonable convention for copy-paste friendliness, but worth a team decision if shared icons are preferred.

Overall
Clean, focused docs PR with no API or behavior impact. The focus-visible:bg-none issue is the only real correctness concern. Otherwise ready to merge after confirming section ordering intent.

aarongarciah:
@flaviendelangle fixed the section ordering. Re `focus-visible:bg-none`, `bg-none` is a thing in TW https://tailwindcss.com/docs/background-image#removing-background-images and it does have an effect.

image

flaviendelangle:
Well, Claude Code hallucination then

colmtuite:
@aarongarciah Remove the hover style from your example and the main demo.

aarongarciah:
@colmtuite done. The Tooltip [hero demo](https://deploy-preview-4343--base-ui.netlify.app/react/components/tooltip) still has the hover styles, though. Not sure if you prefer to remove that as well.

colmtuite:
That's fine, will fix separately.

PR #4345, opened 2026-03-16
[docs][checkbox group] Mention indeterminate state

aarongarciah:
https://deploy-preview-4345--base-ui.netlify.app/react/components/checkbox-group#parent-checkbox

### DIFF
--- a/docs/src/app/(docs)/react/components/checkbox-group/page.mdx
+++ b/docs/src/app/(docs)/react/components/checkbox-group/page.mdx
@@ -121,7 +121,7 @@ A checkbox that controls other checkboxes within a `<CheckboxGroup>` can be crea
 2. Pass an array of all the child checkbox values to the `allValues` prop on the `<CheckboxGroup>` component
 3. Add the `parent` boolean prop to the parent `<Checkbox.Root>`
 
-The group controls the parent checkbox's [indeterminate](/react/components/checkbox#indeterminate) state when some, but not all, child checkboxes are checked.
+The group controls the parent checkbox's [indeterminate](/react/components/checkbox#CheckboxRoot-indeterminate) state when some, but not all, child checkboxes are checked.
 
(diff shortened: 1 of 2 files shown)


colmtuite:
@aarongarciah Is this useful on Checkbox, outside of Checkbox Group?

aarongarciah:
@colmtuite it's a bit tricky. It's true that the (only?) use case I can think of for indeterminate state is a parent checkbox, but it's probably good to also make the indeterminate state more discoverable on individual checkboxes. The [parent checkbox](https://base-ui.com/react/components/checkbox-group#parent-checkbox) example on the Checkbox Group page requires users to pass the `allValues` prop, which is probably fine, but I thought that in some scenarios some users would want full control of how the parent checkbox/indeterminate state behaves.

cc @atomiks in case you have some thoughts on wether it'd be a valid use case for users to obviate `allValues` and do their own thing.

colmtuite:
Possibly. If there is any valid use case, it's probably helpful. If there isn't (which I suspect is accurate), then it feels like promoting it to an example is confusing, and just the API reference should be enough.

atomiks:
Yeah I think the state only needs to be shown in the Checkbox Group parent demo, which is already the case

aarongarciah:
Fair enough. I removed the example from the Checkbox page. I think it'd be valuable to mention the indeterminate state in the CheckboxGroup page somehow. It helps with discoverability and search. wdyt? See https://github.com/mui/base-ui/pull/4345/changes#diff-df295b9a914c9c7404236127059dfbbebf9e3ae4800ca4530c8dd0c58a254992

edit: with the addition the heading is a bit of a mouthful

Screenshot 2026-03-20 at 08 49 04

colmtuite:
Can just keep the note and remove it from the heading

aarongarciah:
> Can just keep the note and remove it from the heading

Updated

PR #4346, opened 2026-03-16
[slider][docs] Add vertical example

aarongarciah:

https://deploy-preview-4346--base-ui.netlify.app/react/components/slider#vertical


colmtuite:
@aarongarciah What do you think about using the `data-orientation` attribute here, so it's closer to how people will actually want to do it?

This example should prob be moved above the form integration example.

aarongarciah:
@colmtuite I discarded the idea, because it felt a bit weird styling the slider using `data-orientation` when there's only one orientation and the other demos don't use it, but I think you're right since horizontal sliders are the common scenario and vertical sliders are the special case. Updated. Also moved the example before the form integration one.

colmtuite (approved):

PR #4350, opened 2026-03-16
[popover] Remove stray focus guards around trigger when `modal`

atomiks:
## Summary

This change keeps trigger-side focus guards out of fully trapped modal popovers while preserving the existing tab handoff path when `modal={true}` is used without a close part.

## Changes

- Shared the effective focus-trapping state between `PopoverPopup` and `PopoverTrigger`.
- Removed trigger-adjacent focus guards only when the popup is actually running modal focus management.
- Added regression tests for both the trapped modal path and the no-close fallback path.


atomiks:
## Codex Review

### Overview
This patch makes Popover use the same focus-management decision in both the popup and the trigger, so `modal={true}` only drops the trigger-side guards when focus is actually trapped. It also adds regression coverage for both the trapped modal path and the no-close fallback path.

### Findings (None)

### Confidence: 4/5
High confidence based on a full pass over the branch diff, a second sweep for React/rendering correctness, accessibility and keyboard behavior, API/state consistency, maintainability, performance risk, and test coverage, plus targeted verification with `pnpm test:jsdom PopoverRoot --no-watch`, `pnpm exec eslint ...`, and `pnpm exec tsc -b tsconfig.json --pretty false`.

### Notes
- Latest changes since the earlier draft PR version:
- The branch was rebased onto the latest `master`.
- The follow-up simplify pass replaced the custom popup layout-effect sync with `store.useSyncedValue('focusManagerModal', focusManagerModal)` while preserving the detached-trigger behavior.

PR #4351, opened 2026-03-16
[combobox] Fix iOS viewport settling

atomiks:
When the input is outside the popup, we need to let mobile software keyboards "settle" to avoid 1) flip-flickers, and 2) issues with the selected item not scrolling into view due to the delayed visual viewport resize. 

In iOS 26, this needs to be increased vs. iOS 18 which was faster to adjust the settling visual viewport size. 100ms appears to be robust when manually testing unlike 50ms.


atomiks:
## Codex Review

### Overview
This patch only changes one behavior: it doubles the touch-open delay for the outside-input combobox path from 50ms to 100ms to reduce the visible top-to-bottom repositioning on touch devices when the mobile keyboard settles.

### Findings (None)

### Confidence: 4/5
High confidence because the branch diff is very small, the affected code path is easy to reason about, and the key build and test checks are green. Residual risk is mostly around real-device mobile behavior, which is not directly covered by the automated tests.

### Notes
- Verified the current branch against the full PR diff, not only the latest edit.
- Re-reviewed after aligning the inline comment with the intended cross-touch behavior.
- Checked the broad correctness and maintenance categories, plus focused verification with `pnpm exec tsc -p packages/react/tsconfig.build.json --noEmit`, `pnpm release:build`, and `pnpm test:jsdom ComboboxRoot --no-watch`.

PR #4353, opened 2026-03-16
[drawer] Add virtual keyboard provider

### LINKED ISSUE #4301: [drawer] With keyboard on iOS
# Bug report

## Current behavior

The drawer doesn't appear behind the keyboard input and scroll within the container always dismisses the drawer.

## Expected behavior

The drawer should appear behind the keyboard and the scroll behaviour should be the same as when there is no keyboard.

## Reproducible example

Simply adding an input to the bottom drawer example is enough: https://base-ui.com/react/components/drawer#position

## Base UI version

For example v1.0.2

iOS Safari

## Which OS are you using?

Mac OS/Windows/Other etc.

Voiceover/JAWS etc.

## Additional context

atomiks:
- Preview: https://deploy-preview-4353--base-ui.netlify.app/react/components/drawer#virtual-keyboard-aware (open on a mobile device with a software keyboard)
- Experiments for different slack padding lengths: https://deploy-preview-4353--base-ui.netlify.app/experiments/drawer/virtual-keyboard-aware 

Fixes #4301

Adds `<Drawer.VirtualKeyboardProvider>` as an opt-in wrapper for bottom sheet drawers with form fields. It keeps the drawer popup element stable while the software keyboard is open and scrolls the focused field into view inside the drawer body by temporarily adjusting the nearest scrollable ancestor.

- `padding-bottom` creates real extra scrollable space above the keyboard.
- `scroll-padding-bottom` tells scroll-into-view algorithms to leave breathing room.
- `overflow-anchor: none` prevents browser scroll anchoring from undoing the adjustment.

## Changes

- Exports `<Drawer.VirtualKeyboardProvider>` for opt-in keyboard-aware bottom sheets.
- Uses visual viewport-aware focus handling to add temporary scroll slack and center the focused field in the nearest scrollable drawer body.
- Keeps header/footer layout under user control; the provider does not add a `Drawer.ScrollArea` part or lift footer actions above the keyboard.


flaviendelangle (commented):
Image

The demo is broken in dark mode

flaviendelangle (commented):
# PR Review Summary — #4353 [drawer] Add keyboard-aware drawer

**What it does:** Adds `Drawer.VirtualKeyboardProvider`, an opt-in component that adjusts a bottom-sheet drawer when a software keyboard opens — scroll-slack padding, centering the focused field in the visible visual viewport, and re-focusing tapped inputs without page scroll. +1647 / −110 across 16 files.

Branch `codex/drawer-ios-keyboard` was checked out and reviewed with 5 specialized agents (code, errors, tests, types, comments). `pnpm typescript` and `pnpm eslint` pass on the changed files.

## 🔴 Critical Issues (must resolve before merge)

**1. PR description contradicts the implementation** — *(code-reviewer, comment-analyzer)*
The GitHub description says it "Adds the `virtualKeyboardAware` prop" and "Removes `Drawer.VirtualKeyboardProvider`." The branch does the **opposite**: it adds the `Drawer.VirtualKeyboardProvider` component and there is no `virtualKeyboardAware` prop anywhere. Exports (`index.parts.ts:14`, `index.ts:17`), docs, and demos all consistently describe the provider — so the *code/docs/comments are internally consistent*, but the description (and any changelog tooling reading it) is stale or the wrong commits were pushed. Reconcile this — it's a question of which API the team actually wants to ship.

## 🟠 Important Issues (should address)

**2. Behavioral change to the existing (non-keyboard) swipe path** — *(code-reviewer)* `DrawerViewport.tsx:1064-1072`
`onTouchStart` now resolves `target` as `elementAtPoint ?? getTarget(nativeEvent)` instead of just `getTarget(nativeEvent)`. This `target` feeds the `contains()` guard and `findScrollableTouchTarget()` for **all** drawers, not just keyboard-aware ones, so swipe-vs-scroll arbitration can shift for existing users when `elementFromPoint` returns a different element than the event target. Scope it to the keyboard path, or justify it as a general improvement with a test for the non-keyboard case.

**3. Critical test gaps** — *(pr-test-analyzer)* Only 5 new tests (all Chromium-only) for a 534-line behavioral surface. Highest-value missing tests, in order:
- **Touchend re-focus of an *unfocused* input** (the headline iOS fix) — assert `focus()` called, `preventDefault()` called, and that `opacity`/`transform`/`transition` are **restored** by the `finally`. A regression here leaves an input at `translateY(-2000px)`/`opacity:0` — invisible. Only the *negative* (already-focused) case is tested today. `DrawerVirtualKeyboardProvider.tsx:472-487`
- **Restore on close/unmount** — only the blur restore path is covered; stale 288px padding on reopen would go uncaught. `overflowAnchor: 'none'` restoration is never asserted.
- **Nested-drawer suppression** + restore.
- **focusout → another input** element-switch path (`:91-94`).
- touchmove >10px tap-cancel boundary; hit-slop near-miss tap; label→control resolution; `prefers-reduced-motion` instant-scroll branch.

**4. Brittle test assertions** — *(pr-test-analyzer)* Exact-pixel literals like `paddingBottom).toBe('288px')` and `scrollTop).toBeCloseTo(270)` bake in three internal constants at once (`KEYBOARD_SCROLL_SLACK`, `KEYBOARD_VISIBILITY_MARGIN`, rect math). Tuning a constant breaks tests with an opaque "expected 288px." Prefer asserting the *contract* (input center lands within the visible viewport; padding ≥ base + slack).

## 🟡 Suggestions (nice to have)

- **Missing comments on non-obvious code** — *(comment-analyzer, code-reviewer)* The `translateY(-2000px)` focus hack (`:472`), the keyboard-detection heuristic in `getKeyboardVisualViewport` (`:517`), the hit-slop probe grid, and the `KEYBOARD_*`/`INPUT_TAP_*` magic numbers all warrant a one-line "why."
- **Reword docs prose** — *(comment-analyzer)* `page.mdx:49` "Existing drawers keep the legacy behavior" implies deprecation; the feature is purely opt-in. Say "drawers without this provider are unaffected."
- **Document the `onTouchEnd: boolean` contract** — *(type-design-analyzer)* The `true`-means-"handled, preventDefault called, skip swipe" contract lives only in the implementation. Add JSDoc on the context interface (`DrawerVirtualKeyboardContext.tsx:4`). (The void/boolean asymmetry itself is good design — keep it.)
- **`ScrollAdjustment` polish** — *(type-design-analyzer)* Mark fields `readonly`; the `paddingBottom` (inline string to restore) vs `computedPaddingBottom` (numeric baseline) naming is easy to confuse. Optionally name the `{ top, bottom }` return of `getKeyboardVisualViewport`.
- **Note the looser scroll-ancestor reimplementation** — *(code-reviewer)* `findPotentialScrollAncestor` (`:495`) intentionally omits the `scrollHeight > clientHeight` check that the shared `scrollable.ts` util enforces; a comment would satisfy the AGENTS.md "avoid duplicating logic" guideline.
- **`restoreKeyboardScrollAdjustment` theoretical leak** — *(silent-failure-hunter)* If a `style.*` write threw, `keyboardScrollAdjustmentRef` is never nulled (`:76-79`). Can't happen for real HTMLElements; a one-liner closes it. Borderline, non-blocking.

## ✅ Strengths

- **Clean error handling** — the `try/finally` style-restore is verified safe (restores even if `focus()` throws), `visualViewport` feature-detection degrades to a graceful no-op, and all listeners/rAFs are symmetrically cleaned up. No silent failures, no leaks.
- **Strong guideline compliance** — uses `useAnimationFrame`/`useStableCallback`, owner-safe `ownerDocument`/`ownerWindow`/`contains`/`getTarget`/`activeElement`, realm-safe `instanceof`, `-webkit-user-select` paired in demos, shared touch handlers via context (not duplicated). `react-hooks/exhaustive-deps` passes.
- **Types follow repo conventions** — the empty `State {}` + `Props` + namespace triad is a faithful copy of `DrawerProvider`; correctly does *not* extend `BaseUIComponentProps` since it renders no element. Backward compatible (provider is optional via `?.`).
- **Docs/comments are accurate** vs the implementation, and the demo properly models header/footer-outside-scrollable-body.

## Recommended Action

1. **Resolve the description/implementation mismatch (#1)** — decide whether the shipped API is the provider or a prop.
2. **Clarify/scope the `onTouchStart` target change (#2)** and add a non-keyboard regression test.
3. **Fill the critical test gaps (#3)** — especially the unfocused-input re-focus + style-restore, and close/unmount restore.
4. **De-brittle assertions (#4)** and add the explanatory comments/JSDoc.
5. Re-run the review (and `code-simplifier`) after fixes.

stivaliserna:
Hey! Thanks for addressing this nasty iOS issue.

I was trying this out but I can't get the footer to render above the keyboard. Is this a bug or is it the intended layout? 

Demo:

https://github.com/user-attachments/assets/cc8cc5f6-6cb3-4eb0-a0c9-fa4a564aeaed

atomiks:
@stivaliserna This is intended atm. The provider keeps the drawer popup stable and scrolls the focused field into view inside the scrollable body but it doesn’t move footer/actions above the keyboard. Making the footer keyboard-persistent would force users into an explicit layout, such as a footer/scroll-area part, or keyboard inset CSS variable, which this PR is avoiding to ensure it works "out of the box" without manual config. 

Personally also, I don't think seeing the footer while the keyboard is open is that important and it would actually minimize the viewing height which is already somewhat constrained.

michaldudak:
## Review summary

Ran a multi-aspect review over the source, tests, demos, and docs (general code quality, test coverage, error handling, type design, comment accuracy).

**Verdict: no blocking issues.** This is carefully written, convention-compliant code. The findings below are all hardening/polish — nothing must be fixed before merge.

### Critical issues
None found.

- **Error handling**: the two `try/finally` blocks (`restoreKeyboardScrollAdjustment`, `focusKeyboardInputWithoutPageScroll`) and the two empty `catch` blocks around `CSS.registerProperty` are all correct. The many silent `return` guards are intentional progressive-enhancement no-ops (the feature does nothing when no keyboard is detectable). No NaN-leak path — `parseFloat(...) || 0` plus finite arithmetic. Cleanup is symmetric: inline padding/overflow-anchor styles and the CSS var are always restored on close/unmount.
- **Conventions**: uses `useAnimationFrame`, `useStableCallback`, `ownerDocument`/`ownerWindow`, realm-safe `instanceof` (`win.HTMLInputElement`), shares handlers via context, demo CSS uses raw `oklch()` and pairs `-webkit-user-select`. ESLint clean (incl. `exhaustive-deps`).

### Important — test gaps worth addressing
Three behaviors currently have no direct coverage:

1. **`handleFocusOut` `relatedTarget` re-target branch is entirely untested** — [`DrawerVirtualKeyboardProvider.tsx:248`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx#L248). When focus moves directly from one keyboard input to another, the code re-aligns without clearing inset/slack. A regression (clearing on every focusout) would briefly drop the inset between fields and go unnoticed. The existing "different scroll container" test exercises the `focusin` path, not this branch.
2. **`prefers-reduced-motion` → `behavior: 'auto'` is never asserted** — [`DrawerVirtualKeyboardProvider.tsx:126`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx#L126). Only verifiable via a `scrollTo` spy under mocked `matchMedia`; final `scrollTop` is identical either way, so current tests can't catch a regression.
3. **The touchstart-ordering integration contract isn't locked in** — [`DrawerViewport.tsx:1085`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/packages/react/src/drawer/viewport/DrawerViewport.tsx#L1085). This PR reorders `virtualKeyboard?.onTouchStart` to run *before* the swipe-ignore early return — the load-bearing seam for the iOS fix. A future refactor moving it back would silently break tap-to-focus, but no test asserts the ordering for the common (non-swipe-ignored) input.

The suite is otherwise genuinely strong (touch state machine, scroll-slack lifecycle, resize threshold, pinch-zoom guard, nested-drawer guard, label/contentEditable resolution, hit-slop probe are all well covered, with appropriate `it.skipIf(isJSDOM)` gating).

### Suggestions
- **Comments**: document the `KEYBOARD_SCROLL_SLACK = 48` magic number and why it's added only on top of existing overlap ([`:21`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx#L21) / [`:207`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx#L207)). Add a one-liner to the already-focused early-return guard at [`:356`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx#L356). (All existing platform-quirk comments were verified accurate.)
- **Types**: mark `KeyboardVisualViewport`'s `top`/`bottom` `readonly` to match the immutable-snapshot intent of `ScrollAdjustment` ([`:43`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx#L43)); a clarifying comment on `ScrollAdjustment` (why both string + parsed-number fields exist) would help. The empty `State {}`, the interface/value name sharing, and the optional non-throwing context hook all match existing Base UI conventions — leave as-is.
- **Docs nit**: [`page.mdx:147`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/docs/src/app/(docs)/react/components/drawer/page.mdx#L147) says the provider "scrolls the focused field into the visible viewport," but it actually *centers* the field and only scrolls when the container is scrollable. Minor wording.

### Strengths
- The `data-base-ui-swipe-ignore` focus path, picker-input (`type="date"`) preservation, and the off-screen-transform focus trick are exactly the iOS edge cases that silently regress — and they're tested.
- Comments document *why* (iOS quirks, slack-makes-container-scrollable) rather than restating code; [`:506`](https://github.com/mui/base-ui/blob/aca84a12bfd93a782332c7888f85df70f949fdb1/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx#L506) is exemplary.
- Public API surface (`types.md`, `page.mdx`, CSS var docs) is regenerated and consistent; demos accurately model the documented pattern.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

atomiks:
## Review

No blocking issues — the keyboard provider path is correct. What's left is a couple of non-blocking tradeoffs and some optional test hardening.

### Tradeoffs (non-blocking)
- **Off-screen focus transform** — `focusKeyboardInputWithoutPageScroll` moves the input to `translateY(-2000px)` for the synchronous focus, which is observable to consumer `onFocus` handlers that read layout (they'd see `top ≈ -2000`). Worth keeping since it's doing real iOS work, but a one-line caveat comment would help future maintainers.
- **`--drawer-keyboard-inset` resets to `0px`** in some inactive states rather than being removed. Fine for provider-owned styling, though `removeProperty` would preserve `var(--drawer-keyboard-inset, <fallback>)` semantics more strictly.

### Optional test hardening (later)
Coverage is already strong (shadow-root focus, label targeting, click coordinates, hit-slop, contentEditable host, touchcancel, drag suppression, pinch-zoom, pickers, nested drawers, close cleanup, scroll-container switching, viewport math). Gaps worth closing eventually:
- focused field removed from the DOM **without** `focusout` (the self-correcting restore path)
- provider unmount while the inset is applied (asserts the CSS var is removed)
- a nested drawer opening **during** an active alignment (the `nestedDrawerOpen` branch in `alignFocusedKeyboardTarget`)
- the exact `KEYBOARD_RESIZE_THRESHOLD` (60px) boundary — only the below-threshold case exists
- a tap with jitter just under the 10px movement threshold — only the over-threshold case exists

### Leanness
No dead code — every helper has a call site, and `scrollable.ts` is a one-parameter addition (`allowOverflowIntent`) to an existing util shared with `useSwipeDismiss`, not a new file. The provider's line count maps to genuine iOS complexity (pinch-zoom suppression, tap-vs-drag, off-screen focus, slack save/restore, contentEditable resolution), not over-engineering.

michaldudak (changes requested):
## Codex Review (GPT-5.5)
Reviewed the full PR branch against freshly fetched `upstream/master`. Dominant PR type: feature, with public API, docs, and tests as secondary scope.

### 1. Bugs / Issues

#### 1. 🔴 Disabled text controls can receive synthetic clicks (blocking)

`Drawer.VirtualKeyboardProvider` treats disabled text inputs/textareas as keyboard targets, then prevents the native `touchend` and dispatches its own click. Real taps on a disabled input emit touch events but do not produce a native click; this path can make disabled controls run `onClick` handlers anyway.

`packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx`

```ts
return element instanceof win.HTMLInputElement && KEYBOARD_INPUT_TYPES.has(element.type);
```

```ts
event.preventDefault();
focusKeyboardInputWithoutPageScroll(keyboardFocusTarget);
dispatchKeyboardClick(keyboardClickTarget, touch);
```

**Fix:** Exclude disabled keyboard controls from `isKeyboardInputElement`, preferably using the platform disabled state (`:disabled` or the relevant form-control `disabled` checks), and add coverage for disabled input/textarea and disabled labelled-control taps.

#### 2. 🔴 Shadow-root inputs lose scroll slack/alignment (blocking)

The provider can detect a focused input through a shadow root, but scroll-target lookup starts from `target.parentElement` and `findScrollableTouchTarget` also walks via `parentElement`. For an input inside a shadow root, that traversal stops at the shadow boundary, so the provider can set `--drawer-keyboard-inset` but cannot find the drawer body scroller to add slack or align the field above the keyboard.

`packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx`

```ts
return (
  findScrollableTouchTarget(target.parentElement, root, 'vertical') ??
  findScrollableTouchTarget(target.parentElement, root, 'vertical', true)
);
```

`packages/react/src/utils/scrollable.ts`

```ts
while (node && node !== root) {
  if (isScrollable(node, axis, allowOverflowIntent)) {
    return node;
  }
  node = node.parentElement;
}
```

**Fix:** Use composed-tree ancestor traversal for scroll target discovery so it can cross slots/shadow hosts, and add a browser test where a shadow-root input inside a scrollable drawer content receives slack and scroll alignment.

### Test Coverage Assessment

I ran `pnpm test:chromium DrawerVirtualKeyboardProvider --no-watch` and `pnpm test:chromium DrawerViewport --no-watch`; both passed. I did not verify the new regression tests fail on `upstream/master`.

### Recommendation

Request changes 🔴

The new provider is a public interaction API, and both issues affect core tap/focus/scroll behavior. I’d fix these before shipping it.

packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.tsx
```diff
@@ -0,0 +1,628 @@
...
+      resetTouchTrackingState();
+      return;
+    }
+
+    const touch = event.changedTouches[0] ?? event.touches[0];
+    const doc = ownerDocument(event.currentTarget);
+    const nativeEventTarget = getTarget(event.nativeEvent);
+    const keyboardTarget =
+      touch &&
+      (resolveKeyboardTouchTargetFromPoint(doc, touch.clientX, touch.clientY) ??
```
michaldudak: 🟡 This fallback can still steal the tap that the point-based resolver intentionally rejected. On mobile `touchend.target` stays as the element where the touch started, so if a touch starts on a text input, drifts less than `INPUT_TAP_MOVE_THRESHOLD`, and ends over a nearby button, `resolveKeyboardTouchTargetFromPoint()` returns `null` for the interactive button, but this fallback reselects the original input. The code then calls `preventDefault()`, suppresses the native button click, focuses the input, and dispatches the replacement click to the input instead. Consider making the fallback conditional on the lift point not being an interactive/label target, or returning an explicit blocked result from the point resolver.

docs/src/app/(docs)/react/components/drawer/page.mdx
```diff
@@ -138,6 +140,24 @@ import { DemoDrawerSnapPoints } from './demos/snap-points';
 
 By default, the drawer can skip snap points when swiping quickly. Specify the `snapToSequentialPoints` prop to disable velocity-based skipping so the snap target is determined by drag distance (you can still drag past multiple points).
 
+### Virtual keyboard aware
+
+Wrap a bottom sheet in `<Drawer.VirtualKeyboardProvider>` to make it react to software keyboards when it contains form controls.
+
+Place header and footer content outside a plain scrollable body so the popup frame stays stable. When the keyboard opens, the provider scrolls the body to keep the focused field visible.
+
+If a footer contains its own input, reserve a footer slot below the scroll area and lift the pinned footer with `--drawer-keyboard-inset`. In the demo, the focused footer switches to `position: fixed` — positioned against the popup, since its `transform` contains fixed descendants — and adds the inset to its bottom padding.
```
michaldudak: 🟡 The docs should show the fallback form for this CSS variable. The provider removes `--drawer-keyboard-inset` on cleanup and only sets it during alignment/reset, so CSS like `calc(1rem + var(--drawer-keyboard-inset))` becomes invalid before the first keyboard alignment or after cleanup. The demos already use `var(--drawer-keyboard-inset, 0px)`; this prose/API wording should teach that form too.

michaldudak (commented):
Posted the two review findings from the local high-effort pass.

michaldudak (approved):

atomiks:
Todo: figure out how to prevent native scroll for the input arrow key navigation on the virtual keyboard. Experimented with some things but they didn't work well. Vaul's seems to prevent navigating via input arrow keys on the virtual keyboard entirely (downwards) and doesn't reposition when going upwards.

PR #4354, opened 2026-03-16
[slider] Fix edge thumb alignment after reveal

### LINKED ISSUE #4303: [slider] thumbAlignment="edge" produces NaN positions when mounted inside display:none
# Bug report

## Current behavior

When `Slider` with `thumbAlignment="edge"` mounts inside a `display: none` container (e.g., an inactive tab panel toggled via CSS), the thumb and indicator positions resolve to `NaN%`. When the container later becomes visible (`display: block`), the positions remain broken because no recalculation is triggered.

This is because `getInsetPosition` in `SliderThumb` calls `getBoundingClientRect()`, which returns all zeros for elements inside `display: none`:

```
controlSize = 0 - 0 = 0
thumbOffset = 0 / 2 + 0 * percent / 100 = 0
positionPercent = 0 / 0 * 100 = NaN
```

The `useIsoLayoutEffect` dependencies (`thumbValuePercent`, `inset`) don't change when the container becomes visible, so `getInsetPosition` never re-runs.

Note: `thumbAlignment="center"` (default) is not affected because it uses pure `valueToPercent()` without DOM measurement.

Image

Image

## Expected behavior

The slider should display the correct thumb and indicator positions when the container becomes visible.

## Reproducible example

https://stackblitz.com/github/sunya9/base-ui-slider-edge-repro

## Base UI version

1.2.0

Vivaldi/Chrome

## Which OS are you using?

Mac OS

## Additional context

- Workaround: conditionally render the Slider only when the container is visible, instead of toggling `display`

atomiks:
Fixes #4303

## Summary

This fixes the `thumbAlignment="edge"` slider case where measuring while hidden could produce `NaN%` positions that never recovered after the container became visible.

## Changes

- Ignore non-finite inset measurements instead of storing `NaN`.
- Recompute inset thumb and indicator positions with `ResizeObserver` when the thumb or control size changes.
- Add a regression test covering a slider that mounts inside `display: none` and later becomes visible.

### DIFF
--- a/packages/react/src/slider/thumb/SliderThumb.test.tsx
+++ b/packages/react/src/slider/thumb/SliderThumb.test.tsx
@@ -373,6 +373,75 @@ describe('<Slider.Thumb />', () => {
         expect(indicator.style.getPropertyValue('--start-position')).toBe('32%');
       });
     });
+
+    it.skipIf(isJSDOM)(
+      'recomputes range inset positions when the slider becomes visible',
+      async () => {
+        function App() {
+          const [visible, setVisible] = React.useState(false);
+
+          return (
+            <React.Fragment>
+              <button type="button" onClick={() => setVisible(true)}>
+                show
+              </button>
+              <div style={{ display: visible ? 'block' : 'none' }}>
+                <Slider.Root
+                  defaultValue={[30, 70]}
+                  thumbAlignment="edge"
+                  style={{ width: '100px' }}
+                >
+                  <Slider.Control
+                    data-testid="control"
+                    style={{ position: 'relative', width: '100%', height: '10px' }}
+                  >
+                    <Slider.Track style={{ position: 'relative', width: '100%', height: '10px' }}>
+                      <Slider.Indicator data-testid="indicator" />
+                      <Slider.Thumb
+                        data-testid="start-thumb"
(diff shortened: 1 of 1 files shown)


atomiks:
## Codex Review

### Overview
This patch fixes the slider case where `thumbAlignment="edge"` is mounted while hidden and can end up with invalid thumb and indicator positions after reveal. It now treats unmeasurable inset positions as unavailable, recomputes when layout becomes measurable, and includes browser regression coverage for both single-thumb and range hidden-to-visible cases.

### Findings (None)

I also did a second pass over render and effect timing, hydration safety, accessibility semantics, API and state design, maintainability, performance, and regression coverage, and I did not find any actionable issues in those areas.

### Confidence: 4/5
High confidence based on a full-branch diff review, focused JSDOM and Chromium verification, an explicit category sweep with subagents, and the added range regression test covering the previously missing trailing-indicator branch.

### Notes
- Latest update adds a hidden-to-visible range-slider regression case covering the trailing thumb and indicator-size recomputation path.
- Verified with `pnpm test:jsdom SliderThumb --no-watch`, `pnpm test:chromium SliderThumb --no-watch`, and `pnpm prettier`.

PR #4355, opened 2026-03-16
[navigation menu] Remove invalid aria-orientation

### LINKED ISSUE #4349: [navigation menu] Invalid use of `aria-orientation`
# Bug report

## Current behavior

The `NavigationMenu` component adds an `aria-orientation` attribute to the `ul` from `NavigationMenu.List`. This is not appropriate in it's current state because the `ul` element in this situation is assuming the default `list` role which does not support `aria-orientation`.

## Expected behavior

The `role` of `menubar` should be added to the `ul` element, explicitly defining it as a menu bar that contains menu items. This is appropriate because the `NavigationMenu` component implements keyboard interaction pattern that are expected for a `menubar` role. Then, the `aria-orientation` attribute is allowed and does not throw errors in accessibility checks.

Additionally, the `menuitem` role would need to be added to each `NavigationMenu.Item` to satisfy all accessibility checks.

## Reproducible example

The documentation example includes it: https://base-ui.com/react/components/navigation-menu

## Base UI version

1.3.0

All

## Which OS are you using?

All

Playwright AXE

atomiks:
Fixes #4349

## Summary

NavigationMenu.List currently renders `aria-orientation` on a plain `ul`, which triggers accessibility checks because that implicit role does not support the attribute. This change stops Navigation Menu from outputting `aria-orientation` on either top-level or nested lists.

## Changes

- Override the list props so `NavigationMenu.List` does not render `aria-orientation`.
- Update the Navigation Menu root tests to assert the attribute is absent on both top-level and nested lists.

## Notes

- After rebasing onto `upstream/master`, `pnpm test:jsdom NavigationMenuRoot --no-watch` fails before running tests because Vite cannot resolve `date-fns/addDays` from `packages/react/src/temporal-adapter-date-fns/TemporalAdapterDateFns.ts`.
- `pnpm exec eslint packages/react/src/navigation-menu/list/NavigationMenuList.tsx packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx` passes.

### DIFF
--- a/packages/react/src/navigation-menu/list/NavigationMenuList.tsx
+++ b/packages/react/src/navigation-menu/list/NavigationMenuList.tsx
@@ -66,9 +66,7 @@ export const NavigationMenuList = React.forwardRef(function NavigationMenuList(
   // but we want to block it in this case.
   // When nested, skip this handler so arrow keys can reach the parent CompositeRoot.
   const defaultProps: HTMLProps = nested
-    ? {
-        'aria-orientation': orientation,
-      }
+    ? EMPTY_OBJECT
     : {
         onKeyDown(event: React.KeyboardEvent<HTMLDivElement>) {
           const shouldStop =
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review

### Overview
This patch removes an unsupported `aria-orientation` attribute from `NavigationMenu.List` and updates the regression tests to lock in the new behavior. The change is small and targeted, and it keeps the keyboard-handling behavior intact while fixing the accessibility issue.

### Findings (None)

### Confidence: 4/5
High confidence based on a full review of the branch diff across the changed logic and regression tests. The patch is narrow, the override is necessary because the composite root still injects `aria-orientation` on the top-level path, and I did not find correctness, accessibility, API, maintainability, performance, or test-coverage problems in the updated code.

PR #4357, opened 2026-03-17
[scroll area] Remove onVisible fallback

atomiks:
## Summary

Remove the ScrollArea `onVisible` fallback and rely on the existing mount-time compute plus `ResizeObserver` updates when the viewport becomes measurable.

## Changes

- Remove the unused `onVisible` helper from ScrollArea.
- Simplify `ScrollAreaViewport` by dropping the visibility fallback and leaving CSS property registration in its own effect.
- Keep the existing ScrollArea sizing behavior covered by the focused root tests.

Key experiments where a regression would occur:

- https://deploy-preview-4357--base-ui.netlify.app/experiments/scroll-area/tabs-scroll-area
- https://deploy-preview-4357--base-ui.netlify.app/experiments/scroll-area/inside-select

### DIFF
--- a/packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
+++ b/packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx
@@ -288,11 +288,10 @@ export const ScrollAreaViewport = React.forwardRef(function ScrollAreaViewport(
 
   useIsoLayoutEffect(() => {
     if (!viewportRef.current) {
-      return undefined;
+      return;
     }
 
     removeCSSVariableInheritance();
-    return undefined;
   }, [viewportRef]);
 
   useIsoLayoutEffect(() => {


atomiks:
## Codex Review

### Overview
This patch removes ScrollArea's dedicated `onVisible` helper and leaves the existing mount-time compute plus `ResizeObserver` path in charge of sizing updates. The result is a smaller viewport setup with one less observer path to maintain.

### Findings (None)

### Confidence: 4/5
High confidence based on the narrow two-file diff, a full serial category sweep across rendering correctness, accessibility, API/state design, maintainability, performance, and test coverage, plus fresh focused test coverage on the rebased branch. I reran `pnpm test:jsdom ScrollAreaRoot --no-watch` and `pnpm test:chromium ScrollAreaRoot --no-watch`, and both passed.

### Notes
- The only plausible behavior change I could infer is for browsers without `ResizeObserver`, since `packages/react/src/scroll-area/viewport/ScrollAreaViewport.tsx:311` now returns early in that case and the `onVisible` fallback is gone.
- I did not count that as a finding because the documented support policy targets Baseline modern browsers (`docs/src/app/(docs)/react/overview/about/page.mdx:40`), and the current browser matrix starts at Chrome/Edge 111, Firefox 113, and Safari/iOS 16.4.
- The existing ScrollArea sizing coverage still exercises the supported hidden-to-visible path through `ResizeObserver` in `packages/react/src/scroll-area/root/ScrollAreaRoot.test.tsx:102` and `packages/react/src/scroll-area/root/ScrollAreaRoot.test.tsx:143`.

flaviendelangle:
Claude  Code

## Review of #4357 — Remove `onVisible` fallback in ScrollArea

**Branch:** `codex/scroll-area-remove-on-visible` → `master`
**Files changed:** 2 (3 additions, 34 deletions)

---

### TL;DR

Clean, low-risk simplification. The `IntersectionObserver`-based `onVisible` fallback is removed because `ResizeObserver` (universally supported since ~2020) already covers the "hidden → visible" transition. **Approve with minor suggestions.**

---

### What changed

1. **Deleted** `packages/react/src/scroll-area/utils/onVisible.ts` — a 21-line `IntersectionObserver` utility.
2. **Simplified** the `useIsoLayoutEffect` in `ScrollAreaViewport` — it now only does CSS variable registration; the visibility-recompute branch is gone.
3. Replaced the module-level `HAS_RESIZE_OBSERVER` constant with an inline `typeof ResizeObserver === 'undefined'` check (functionally equivalent, marginally clearer at the call site).
4. Dropped `computeThumbPosition` from the effect's dependency array (correct — it's no longer referenced in the effect body).

---

### Correctness

- The `ResizeObserver` fires when an element transitions from zero to non-zero size (e.g. tab panel becoming visible), which is exactly the scenario `onVisible` was guarding against. The removal is sound.
- The dep array change (`[computeThumbPosition, viewportRef]` → `[viewportRef]`) is correct since `computeThumbPosition` is no longer called inside the effect.

### Edge case: environments without `ResizeObserver`

If `ResizeObserver` is missing, the only mount-time call to `computeThumbPosition` is via `queueMicrotask`. If the viewport is hidden at mount, that measurement would no-op and there's no longer a fallback. This is intentional — `ResizeObserver` is baseline in all supported browsers. Acceptable trade-off.

---

### Suggestions

1. **Remove the unnecessary `return undefined`** in the first `useIsoLayoutEffect`. It was only there because the old code had a conditional cleanup return (`observer.disconnect()`). Now there's no cleanup, so:
   ```tsx
   useIsoLayoutEffect(() => {
     if (!viewportRef.current) {
       return;
     }
     removeCSSVariableInheritance();
   }, [viewportRef]);
Test coverage is the weakest point. No tests were added/modified, and the deleted utility had no isolated test either. The PR relies entirely on existing ScrollAreaRoot/ScrollAreaViewport tests + manual verification of experiment pages. I'd recommend adding a test for the "scroll area inside a hidden container that becomes visible" scenario — this is the exact path that now relies solely on ResizeObserver.

Minor PR description nit: The description says "keep CSS property registration separate from visibility remeasurement," but what actually happened is the visibility branch was removed entirely, not separated. Small clarity point for future archaeology.

Verdict: Approve with the minor nits above. The change is correct, reduces complexity, and the dropped code path was effectively dead for all modern browsers.

PR #4358, opened 2026-03-17
[docs] Update Calendar demos to use Base UI Select

LukasTy:
Currently, one demo uses a Base UI `<Select>` component, while others rely on the native `<select>`.

Aligned all demos to use the Base UI Select.

Had to play around with sizing and spacing on the month and year select demo. 😁

Used icons from `lucide-react`, if we prefer those inlined, I can replace, but IMHO it's slightly cleaner with them imported.


atomiks:
1. The `check` icon looks horizontally squished/too small

Screenshot 2026-03-17 at 7 39 51 pm

> if we prefer those inlined, I can replace, but IMHO it's slightly cleaner with them imported.

I think all demos should continue to use inline icon components for consistency and easy copy-pastability

2. The font-size of the select item text doesn't match the trigger text, but should

3. When `September` is selected, it expands the trigger size, causing subtle layout shift

LukasTy:
> 1. The `check` icon looks horizontally squished/too small
> 
> > if we prefer those inlined, I can replace, but IMHO it's slightly cleaner with them imported.
> 
> I think all demos should continue to use inline icon components for consistency and easy copy-pastability
 
Thanks for noting this. 👍
I replaced the Select relevant icons with inline ones used in Select demos.
Will address the left and right chevrons in a separate PR to reduce noise.

> 2. The font-size of the select item text doesn't match the trigger text, but should

Increased/aligned the font size.
I don't like the limited spacing this demo creates, but not sure how to best solve it.
Maybe we could do one of these sometime:
- increase the overall calendar size (day buttons, etc.)
- remove the arrow navigation

> 3. When `September` is selected, it expands the trigger size, causing subtle layout shift

Must have forgotten to adjust the `min-width` after some last change.
Adjusted and re-adjusted it so that it would not increase on Firefox.

PR #4360, opened 2026-03-17
[combobox] Add `keepFilterText` prop for repeated selections

### LINKED ISSUE #3977: [combobox] Native way to prevent clearing of input after selection

### LINKED ISSUE #4348: [Combobox] More specific event reason when input clearing is caused by item selection
# Feature request

## Summary

We have a combobox that supports multiple selection with the input inside the popover. We would like to prevent the input value from getting cleared when an item is selected.

It doesn't seem like this can be easily done using `details.reason` because the reason value is `input-clear`, which is the same reason given when the clear button is pressed, or other events that we don't want to ignore. 

I'm not sure if other libraries support this or not. I didn't look.

## Motivation

When allowing multiple selection, I would like my users to be able to filter once, and select multiple items based on that filter.

This is similar to https://github.com/mui/base-ui/issues/3977 . The key difference is that the input is inside the popover.

The relevant code is [here](https://github.com/mui/base-ui/blob/master/packages/react/src/combobox/root/AriaCombobox.tsx#L675)

atomiks:
Closes #3977
Closes #4348

This adds `keepFilterText` to `Combobox.Root` for multiple-selection workflows that keep using the same filter after each pick. When the user selects an item while filtering, the current query stays in the input and the list stays open so more matching items can be selected.

A use case is GitHub's Labels combobox:

GitHub Labels combobox

## Changes

- Add `keepFilterText` for repeated filtered selection in multiple mode.
- Cover inputs outside the popup, inside the popup, and inline lists.
- Regenerate the Combobox API docs.

## Bundle size

A fresh same-worktree comparison against `bf2e52725` measured `@base-ui/react` at **0 B parsed / +16 B gzip**. The implementation folds the new behavior into the existing filter check without adding parsed runtime size.


michaldudak (commented):
# PR review

No blocking findings. I found one close-animation regression in the multiple + input-inside-popup path, and the focused `pnpm test:jsdom ComboboxRoot --no-watch` suite passes.

## Bugs (1)

### 1. 🟡 Preserve the filtered popup while closing

**Location:** `packages/react/src/combobox/root/AriaCombobox.tsx:585`

```tsx
if (pendingInputBehaviorRef.current.closeAction !== 'skip') {
  setInputValue('', createChangeEventDetails(REASONS.inputClear, eventDetails.event));
}
```

For multiple comboboxes with the input inside the popup, this now clears the query immediately on close. The previous code explicitly deferred that clear until unmount for this placement, so the filtered list would not flash back to the unfiltered list during an exit animation.

**Failure scenario:** In a multiple combobox with `Combobox.Input` inside `Combobox.Popup`, type `ap`, press Escape or click outside, and the closing popup can visibly reset from filtered results to all items before the exit transition finishes.

**Fix:** Restore the inside-popup guard here, for example only clear immediately when `!inputInsidePopup || inline`, and let `handleUnmount` clear the popup input for the animated inside-popup case.

## Tests (1)

### 1. 🟡 Restore default inside-popup multiple selection coverage

**Location:** `packages/react/src/combobox/root/ComboboxRoot.test.tsx:2544`

```tsx
describe('prop: keepFilterText', () => {
```

The old default behavior test for multiple mode with the input inside the popup was replaced by `keepFilterText` coverage. That leaves the default path unpinned: selecting while filtering inside the popup should keep the popup open but clear the input.

**Failure scenario:** A future change could make the default inside-popup multiple workflow preserve text, close, or fail to clear, and this suite would still pass.

**Fix:** Add back a test without `keepFilterText` for input-inside-popup multiple mode, asserting the popup remains open and the input clears after selecting a filtered item.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve after nits** — the issues are non-blocking but worth fixing before merge.

atomiks:
Closed by https://github.com/mui/base-ui/pull/5362

PR #4362, opened 2026-03-17
[navigation menu] Fix top-level safePolygon pointer events

atomiks:
## Summary

This fixes a Navigation Menu hover bug where top-level safePolygon pointer blocking stops applying after the pointer enters the popup and returns to a trigger, or after switching directly between top-level triggers before traversing back into the popup.

## Changes

- Reapply top-level safePolygon pointer-event blocking when a hover-open trigger becomes active again.
- Preserve same-scope pointer-event mutations while trigger ownership changes, so the previous trigger does not clear the new block.
- Add regressions covering both returning from the popup and switching directly to a different trigger with a top-level link present.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,6 +17,7 @@ This repository contains the source code and documentation for Base UI: a headl
 - Always use the `useTimeout` utility from `@base-ui/utils/useTimeout` instead of `window.setTimeout`, and `useAnimationFrame` from `@base-ui/utils/useAnimationFrame` instead of `requestAnimationFrame`. Search for other example usage in the codebase if unsure how to use them.
 - Use the `useStableCallback` utility from `@base-ui/utils/useStableCallback` instead of `React.useCallback` if the function is called within an effect or event handler. The utility cannot be used to memoize functions that are called directly in the body of a component (during render), so continue with `React.useCallback` in those scenarios.
 - Always use the `useIsoLayoutEffect` utility from `@base-ui/utils/useIsoLayoutEffect` instead of `React.useLayoutEffect`.
+- Always use the shadow DOM-safe utilities for DOM traversal and event targeting: `contains`, `getTarget`, and `activeElement`. Always use the owner utilities `ownerDocument` and `ownerWindow` instead of global `document`/`window` lookups when the code is tied to a DOM node, including realm-sensitive checks such as `instanceof`.
 - Avoid duplicating logic where necessary. If two components can share logic (such as event handlers), define the logic/handlers in the parent and share it through a context to the child; use the existing context if it exists.
 
 ## Linting, typechecking, and formatting
--- a/docs/reference/generated/autocomplete-root.json
+++ b/docs/reference/generated/autocomplete-root.json
@@ -71,6 +71,11 @@
       "description": "Filtered items to display in the list.\nWhen provided, the list will use these items instead of filtering the `items` prop internally.\nUse when you want to control filtering logic externally with the `useFilter()` hook.",
       "detailedType": "any[] | Group[] | undefined"
     },
+    "form": {
+      "type": "string",
+      "description": "Identifies the form that owns the internal input.\nUseful when the combobox is rendered outside the form.",
+      "detailedType": "string | undefined"
+    },
     "grid": {
       "type": "boolean",
       "default": "false",
--- a/docs/reference/generated/checkbox-root.json
+++ b/docs/reference/generated/checkbox-root.json
@@ -36,6 +36,11 @@
       "description": "The value of the selected checkbox.",
       "detailedType": "string | undefined"
     },
+    "form": {
+      "type": "string",
+      "description": "Identifies the form that owns the hidden input.\nUseful when the checkbox is rendered outside the form.",
+      "detailedType": "string | undefined"
+    },
     "nativeButton": {
       "type": "boolean",
       "default": "false",
--- a/docs/reference/generated/combobox-root.json
+++ b/docs/reference/generated/combobox-root.json
@@ -85,6 +85,11 @@
       "description": "Filtered items to display in the list.\nWhen provided, the list will use these items instead of filtering the `items` prop internally.\nUse when you want to control filtering logic externally with the `useFilter()` hook.",
       "detailedType": "any[] | Group[] | undefined"
     },
+    "form": {
+      "type": "string",
+      "description": "Identifies the form that owns the internal input.\nUseful when the combobox is rendered outside the form.",
+      "detailedType": "string | undefined"
+    },
     "grid": {
       "type": "boolean",
       "default": "false",
--- a/docs/reference/generated/menubar.json
+++ b/docs/reference/generated/menubar.json
@@ -41,6 +41,17 @@
       "detailedType": "| ReactElement\n| ((\n    props: HTMLProps,\n    state: Menubar.State,\n  ) => ReactElement)"
     }
   },
-  "dataAttributes": {},
+  "dataAttributes": {
+    "data-orientation": {
+      "description": "Determines the orientation of the menubar.",
+      "type": "'horizontal' | 'vertical'"
+    },
+    "data-has-submenu-open": {
+      "description": "Present when any submenu within the menubar is open."
+    },
+    "data-modal": {
+      "description": "Present when the corresponding menubar is modal."
+    }
+  },
   "cssVariables": {}
 }
--- a/docs/reference/generated/number-field-root.json
+++ b/docs/reference/generated/number-field-root.json
@@ -33,6 +33,11 @@
       "description": "When true, direct text entry may be outside the `min`/`max` range without clamping,\nso native range underflow/overflow validation can occur.\nStep-based interactions (keyboard arrows, buttons, wheel, scrub) still clamp.",
       "detailedType": "boolean | undefined"
     },
+    "form": {
+      "type": "string",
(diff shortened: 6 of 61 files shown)


atomiks:
## Codex Review (GPT-5.4)
This is a focused bug-fix PR that closes a real Navigation Menu hover gap around `safePolygon` pointer-event blocking for top-level triggers. The final branch now looks consistent with the intended list-scoped behavior, and the added regressions cover the tricky trigger-switch and trigger-to-popup paths that originally escaped.

### 1. Bugs / Issues (None)
I did not find any remaining branch-relevant correctness issues in the final diff.

| Area | Assessment | Notes |
|------|:----------:|-------|
| Shared safePolygon options | **Handled** | Preserving `getScope()` on `fn.__options` fixes the actual scope propagation bug instead of layering more trigger-local fallbacks. |
| Pointer-events ownership handoff | **Handled** | The `WeakMap` ownership guard in `useHoverInteractionSharedState` prevents stale same-scope cleanup from tearing down the active mutation during trigger switches. |
| Trigger re-entry / switch behavior | **Handled** | The inactive-trigger reapply path and top-level trigger-side reapply now cover both “return from popup” and direct trigger-to-trigger transitions. |
| Pointerdown cleanup | **Handled** | Clearing the temporary mutation on trigger `pointerdown` matches the intended interaction model and avoids leaving the list blocked after press initiation. |

| Area | Status | Notes |
|------|:------:|-------|
| Top-level scoping | **Consistent** | The branch now consistently scopes top-level blocking to the list container instead of `body`, including the real-hover browser assertion. |
| Nested positioned menus | **Consistent** | Nested positioned menus still fall back to the broader existing hover logic because `getScope()` intentionally returns `null` there. |
| Inline nested menus | **Consistent** | The inline nested expectation was updated to reflect the same list-scoped behavior rather than a `body` mutation. |

**Strengths:** The latest revision removed the extra `closest('ul')` fallbacks from multiple call sites and centralized scope propagation in `safePolygon()`, which is the smallest version of the fix I’d be comfortable merging.

**Gaps:** No additional simplification stood out without reopening the original hover bug.

| Return from popup to original trigger | **Covered** | Explicit regression test added. |
| Direct switch between top-level triggers | **Covered** | Explicit regression test added. |
| Real browser hover path | **Covered** | Browser-only assertion now checks the list scope instead of `body`. |
| Trigger press while blocking is active | **Covered** | Explicit pointerdown cleanup test added. |

| `pnpm test:jsdom NavigationMenuRoot --no-watch` | **Passed** | Covers the added top-level and inline regression cases. |
| `pnpm test:chromium NavigationMenuRoot --no-watch` | **Passed** | Reproduced and fixed the browser-only expectation drift after the scoping change. |
| `pnpm prettier --check ...` | **Passed** | Touched files are formatted. |
| `pnpm eslint ...` | **Passed** | Targeted lint pass completed on the touched file set. |

### Merge Confidence Score
| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Root cause coverage | 5/5 | The final patch addresses the shared scope propagation bug directly. |
| Regression protection | 4/5 | Strong focused coverage for the reported paths and the adjacent browser-only assertion. |
| Residual interaction risk | 3/5 | Hover sequencing remains subtle by nature, so broad CI still matters. |

Overall merge confidence is **4/5**. I would be comfortable merging this once CI re-runs cleanly on the rebased branch.

atomiks:
## Codex Review (GPT-5.4)
This follow-up pass reflects the latest branch tip after the WebKit opt-out was removed. The change broadens the same list-scoped pointer-event blocking behavior to Safari/WebKit, and the updated branch still looks internally consistent.

I did not find any new branch-relevant correctness issues in the latest follow-up diff.

| Area | Assessment | Notes |
|------|:----------:|-------|
| WebKit behavior | **Intentional broadening** | Top-level hover-open menus now use the same non-touch pointer-event blocking path on WebKit as other engines. |
| Scope consistency | **Handled** | The top-level scope still stays on the list container rather than widening back to `body`. |
| Test contract | **Handled** | The Safari-specific regression now asserts the broadened top-level blocking behavior directly. |

| Area | Status | Notes |
|------|:------:|-------|
| Cross-browser behavior | **More consistent** | Top-level blocking is no longer special-cased away on WebKit. |
| Nested behavior | **Unchanged** | Nested WebKit coverage still confirms submenu blocking remains list-scoped. |

**Strengths:** Removing the `isWebKit` branch makes the blocking rule easier to reason about: non-touch pointers block, touch does not.

**Gaps:** No additional simplification stood out in this follow-up delta.

| Safari top-level hover open | **Covered** | The dedicated WebKit test now checks for list-scoped blocking. |
| Safari nested submenu traversal | **Covered** | Existing WebKit nested coverage still passes unchanged. |

| `pnpm test:jsdom NavigationMenuRoot --no-watch` | **Passed** | Includes the WebKit-mocked unit coverage. |
| `pnpm test:chromium NavigationMenuRoot --no-watch` | **Passed** | Browser lane remains green after the WebKit guard removal. |
| `pnpm prettier --check ...` | **Passed** | Touched files are formatted. |
| `pnpm eslint ...` | **Passed** | Targeted lint pass completed on the touched files. |

### Merge Confidence Score
| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Cross-browser consistency | 4/5 | Behavior is simpler and better aligned across engines. |
| Regression protection | 4/5 | Safari-specific expectations are now explicit. |
| Residual interaction risk | 3/5 | This is still subtle hover sequencing, so broader CI remains useful. |

Overall merge confidence is **4/5**. The follow-up delta looks sound if the goal is to keep pointer-event blocking enabled on WebKit for this experiment.

flaviendelangle:
Claude Code

## PR #4362: [navigation menu] Fix top-level safePolygon pointer events

**Author:** atomiks | **Files changed:** 6 | **+201 / -26**

This fixes a bug where `safePolygon` pointer-event blocking stops re-applying after the pointer enters the popup and returns to a trigger, or after switching directly between top-level triggers.

---

### Changes overview

**1. Ownership tracking via WeakMap** — `useHoverInteractionSharedState.ts`

A module-level `WeakMap<HTMLElement, PointerEventsMutationState>` tracks which `HoverInteraction` instance "owns" the pointer-events mutation for a given scope element. This prevents a stale trigger's cleanup from undoing the new trigger's mutation — `clearSafePolygonPointerEventsMutation` now no-ops when the instance is not the current owner. Clean approach; WeakMap keyed on DOM elements avoids memory leaks.

**2. Reapply mutation on `onMouseMove`** — `useHoverReferenceInteraction.ts`

When hovering over an *inactive* trigger while another is already open (`currentOpen && isOverInactive`), pointer-event blocking is reapplied. This uses `instance.handleCloseOptions?.getScope?.()` to locate the scope element, which brings us to...

**3. Propagate all options to `__options`** — `safePolygon.ts`

`fn.__options` now spreads `...options` before explicitly setting `blockPointerEvents`. This ensures `getScope` (and any future options) are available on `handleCloseOptions`. Without this, `getScope` was lost and the `onMouseMove` code above couldn't resolve the scope. Clean fix.

**4. Trigger-level changes** — `NavigationMenuTrigger.tsx`

- **Removed `isWebKit` guard** — `blockPointerEvents` simplified to `pointerType !== 'touch'`. This intentionally enables pointer-event blocking on Safari for top-level triggers (the webkit test is updated to match).
- **`getScope()`** now returns `null` for nested triggers with a positioner (instead of always returning closest `<ul>`), since those use different positioning logic.
- **`handleActivation`** now applies pointer-event mutation for both top-level and nested triggers. When switching items (`value !== itemValue`), it defers via `queueMicrotask` to ensure the old trigger's cleanup runs first.
- **`handleTriggerPointerDown`** clears the mutation on pointer down (click = no more polygon traversal needed).

**5. Tests** — thorough regression coverage for all three scenarios (return from popup, direct trigger switch, pointer down clearing).

---

### Questions / suggestions

1. **Is `queueMicrotask` strictly necessary given the WeakMap ownership?** The WeakMap already prevents the old trigger's cleanup from undoing the new trigger's mutation (different instance = not the owner → skip style removal). I traced through both paths and they arrive at the same outcome. The microtask adds a brief window where pointer events aren't blocked. It's belt-and-suspenders which is fine, but worth confirming whether the WeakMap alone would suffice — it would simplify the logic.

2. **`...options` spread is slightly broader than needed.** It carries `requireIntent` and `buffer` onto `__options` even though only `blockPointerEvents` and `getScope` are read from it. Not harmful today, but a more targeted spread (`{ getScope: options.getScope, blockPointerEvents }`) would be more explicit about intent. Minor nit.

3. **Safari behavioral change** — the removal of the `isWebKit` guard means Safari now gets pointer-event blocking for top-level triggers. The webkit test update confirms this is intentional, but it might be worth a brief note in the PR description for anyone bisecting Safari-specific regressions later.

4. **`hoverFloatingElement` in the microtask closure** — in `applyPointerEventsMutation`, `hoverFloatingElement` is read from the component scope (not captured as a local). By the time the microtask fires (after `flushSync` + effects), the binding should still be valid since it's determined by props/context rather than the active item. Just flagging that it's a live reference, not a snapshot.

---

### Verdict

The fix is well-structured and correct. The ownership-tracking WeakMap is the right primitive for coordinating multiple trigger instances competing over the same scope element's styles. Tests cover the key regression scenarios comprehensively. The only thing I'd probe is whether the `queueMicrotask` path can be removed in favor of relying solely on the WeakMap ownership, since that would simplify the code and eliminate the brief unblocked window.

PR #4363, opened 2026-03-17
[all components] Fix render prop warning false positive

### LINKED ISSUE #4361: Render prop warning false positive on screaming snake case variables
# Bug report

## Current behavior

Since #4077 (and including changes from #4324), passing a `render` prop as a variable named in "screaming snake case" (e.g. "DEFAULT_RENDER") will log a warning about being passed a component.

## Expected behavior

I would not expect logging to occur in a situation where a valid function is passed, as the logging intends to be directed at React components particularly.

Screaming snake case is a common convention for assigning variables to communicate them being constants. It's a convention used in this project itself ([example](https://github.com/mui/base-ui/blob/63be369cf06d054ff9773255eda21a60c5243205/packages/react/src/utils/useSwipeDismiss.ts#L22-L28)).

## Reproducible example

https://stackblitz.com/edit/vitejs-vite-xz4rwjgf?file=src%2FFancyButton.tsx

```
FancyButton.tsx:6 Base UI: The `render` prop received a function named `DEFAULT_RENDER` that starts with an uppercase letter. This usually means a React component was passed directly as `render={Component}`. Base UI calls `render` as a plain function, which can break the Rules of Hooks during reconciliation. If this is an intentional render callback, rename it to start with a lowercase letter. Use `render={<Component />}` or `render={(props) => <Component {...props} />}` instead. https://base-ui.com/r/invalid-render-prop
```

Real-world example: https://github.com/WordPress/gutenberg/blob/26f64ca808f83edf8a58456c2ee6b82e8ecae5b3/packages/ui/src/form/primitives/field/root.tsx#L8-L24

## Base UI version

v1.3.0

Chrome 146

## Which OS are you using?

Mac OS 26.3.1

N/A

## Additional context

With the changes in #4324, maybe the pattern could be updated to expect a lowercase letter after the initial uppercase letter? e.g.

```
const COMPONENT_IDENTIFIER_PATTERN = /^[A-Z][a-z][A-Za-z0-9_$]*$/;
```

atomiks:
Fixes #4361

## Summary

This narrows the development-only render prop warning so constant-style callbacks such as `DEFAULT_RENDER` no longer trigger a false positive. The warning still catches component-like names, including PascalCase names with acronym prefixes.

## Changes

- Require warned render callback names to look component-like instead of matching uppercase constants.
- Add regression coverage for screaming snake case callbacks.
- Keep warning coverage for standard PascalCase and acronym-prefixed component names.


atomiks:
## Codex Review

### Overview
This patch narrows the render prop warning so uppercase constant callbacks like `DEFAULT_RENDER` stop warning while component-like names still do. The tests cover the false-positive case and preserve the earlier warning behavior for PascalCase and acronym-prefixed component names.

### Findings (None)

### Confidence: 4/5
High confidence based on a full pass over the branch diff, targeted inspection of the warning heuristic and regression tests, and successful runs of `pnpm test:jsdom useRenderElement --no-watch` plus targeted ESLint on the touched files.

### Notes
- Reviewed the full `upstream/master...HEAD` diff.
- No additional accessibility, API, maintainability, or performance concerns surfaced for this small change set.

Issue #25, opened 2023-07-19
[accordion] Implement Accordion


michaldudak:
This is the place to gather ideas, measure interest and discuss the API and implementation details of the accordion component and hook.

See https://github.com/mui/base-ui/issues/10 for the umbrella issue to track progress on Base UI and https://github.com/mui/material-ui/issues/6218 for the problem this project solve.

oliviertassinari:
Material UI id API feels so outdated (e.g. `id={summary.props['aria-controls']}`) now that we can use the context and `useId` https://mui.com/material-ui/react-accordion/#basic-accordion

<img width="424" alt="Screenshot 2023-08-15 at 00 52 48" src="https://github.com/mui/material-ui/assets/3165635/98d62f74-3121-459b-9882-40777bb9e60f">

oliviertassinari:
To consider what we should do with the arrow key navigation https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#keyboardinteraction. I don't know if we want this, I guess, why not?

logonoff:
an idea for the base-ui accordion component is to use the `details` and `summary` HTML elements for the component so that the accordion would still work for users who have javascript turned off. i feel like semantically it might also make more sense since its functionality is similar to the accordion component

michaldudak:
Related issue to take into consideration: #1147

Issue #14, opened 2022-09-21
[Select][base-ui] Set the onChange's event.target.value


michaldudak:
As noted in https://github.com/mui/material-ui/pull/34408#discussion_r976288729, it's worth ensuring that the event object in Unstyled Select's `onChange` method has a target property with a valid `value` property.

harsh5902:
shall i work on this issue?

michaldudak:
Sure, go ahead!

jesrodri:
@harsh5902 are you still working on this? if not, can I take it?

harsh5902:
yes i am still working on it.

michaldudak:
I have investigated the problem in more depth. In the case of Select, setting the `event.target.value` may not work the best because:
- [`HTMLSelectElement.value`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement#instance_properties) is defined as a string. In our case, this could be any object.
- In the case of `select multiple`, the `value` contains only the first selected element. The whole collection is available under the `selectedOptions`, which is a set of `HTMLOptionElement` instances. This is not available in the case of our custom Select.

I'd like to wait and see if this becomes a problem in real-world applications. When we see concrete examples of how this should work, we may be able to think of the right solution. Until then, I'm putting it on hold.

A side note: neither Headless UI nor Radix UI provides the `event` object in their selects' change handlers.

Sen-442b:
@michaldudak slightly unrelated, but is there a similar rationale behind not supporting e.target.name as well?

michaldudak:
We are considering changing this API in the new version of the Select but nothing is set in stone yet.

Issue #13, opened 2022-03-31
[Switch] Integrate with FormControlUnstyled


michaldudak:
Make SwitchUnstyled aware of the FormControlUnstyled context, similarly to how Material UI's Switch interacts with FormControl.

EduardoSCosta:
Hi, I'd like to do this issue :smiley:

michaldudak:
Hi. It requires mui/base-ui#14 to be completed first (or adding a second parameter, `newValue`, to all inputs' onChange events), so that FormControlUnstyled is able is figure out the new value after it changes.

michaldudak:
We replaced FormControlUnstyled with Form and Field components. The Switch does support them out of the box.

Issue #12, opened 2022-02-10
[Select] Integrate with FormControlUnstyled


michaldudak:
Make SelectUnstyled aware of the FormControlUnstyled context, similarly to how Material UI's Select interacts with FormControl.

ivan-ngchakming:
Hi, I'd like to work on this.

From my understanding, add the `useFormControlUnstyled` hook in `<SelectUnstyled>`, so `<SelectUnstyled>` can be aware of the following properties: `disabled`, `error`, `focused`, `required`?

I suppose some demos should be added showing how to use `<FormControlUnstyled>` together with `<SelectUnstyled>`, I am thinking I can add them in a separate PR.

michaldudak:
Hi, could you please hold it off for a bit? I have some changes to FormControlUnstyled in progress. I hope to create a PR tomorrow at most.

michaldudak:
@ivan-ngchakming Here's the PR with the changes I mentioned: mui/material-ui#32134

Once it's merged in, you may work on this issue.

michaldudak:
We replaced FormControlUnstyled with Form and Field components. The Select will support them out of the box.

PR #4365, opened 2026-03-18
[otp field] Add OTP Field component

### LINKED ISSUE #75: [base-ui] Add a hook, support OTP/ PinInput `usePinInput`
### Summary

I believe it would be fantastic to introduce a base-ui hook specifically designed for handling OTP inputs. 
Chakra already offers a `PinInput` component along with a headless UI hook in their Zag library (https://zagjs.com/components/react/pin-input).

### Examples

_No response_

### Motivation

_No response_

**Search keywords**:

### Features to support

- Past from plain string, e.g. pasting `990312` on mobile should fill the whole component.
- Caret should be managed as if it was a single `<input>`. 
- Be 6 pins by default, most common use case is 2FA
- Submit form after past when pressing <kbd>Enter</kbd>

### Benchmarks

- https://zagjs.com/components/react/pin-input
- https://chakra-ui.com/docs/components/pin-input
- https://vuetifyjs.com/en/components/otp-input/
- https://mantine.dev/core/pin-input/
- https://www.radix-vue.com/components/pin-input
- https://twitter.com/liampmccabe/status/1754450110992244830

atomiks:
Closes #75

Preview: https://deploy-preview-4365--base-ui.netlify.app/react/components/otp-field

## Summary

This adds a preview OTP Field component with Root and Input parts, along with docs, hero demos, tests, and generated API reference.

**API note:** `length` is required on `OTPField.Root` as the source of truth for the intended OTP size. The component uses it for validation, hidden-input form integration, completion state, and deterministic SSR or initial render behavior, while CompositeList is used only for slot order and indexing.


atomiks:
## Codex Review (GPT-5.4)

This remains a `new-component` PR with secondary tags `public-api`, `docs`, `tests`, and `a11y`. I re-reviewed the full branch against a freshly fetched `upstream/master` and reran `pnpm test:jsdom OTPField --no-watch`, `pnpm test:chromium OTPField --no-watch`, `pnpm typescript`, and `pnpm eslint` on the current head.

I do not currently see any remaining actionable correctness or API issues on the branch head.

### 2. Public API & Composition Assessment

The public shape is now much cleaner than the original branch state. Making `OTPField.Root` the grouping element, keeping `length` as the single source of truth, exposing a single-string value, and leaving forwarded `aria-*` on the root gives the component a more coherent long-term contract than the earlier `Root` + `Group` split.

### 3. Accessibility and Interaction Assessment

The interaction model looks solid at this point: canceled changes no longer move focus, forward delete reselects the shifted value, invalid rejected input preserves selection, readonly slots still participate in focus tracking and keyboard navigation, and the root/label wiring is internally consistent. For a preview component, the keyboard and focus behavior now looks well covered.

### 4. Previous Review Issues Status

| Prior review item | Status | Notes |
|---|:---:|---|
| Canceled changes still moved focus | Resolved | Focus/completion side effects now flush only against committed values |
| Custom validation needed better flexibility | Resolved | `inputMode` and `onValueInvalid` are both present and documented |
| `Group` made the public structure heavier than necessary | Resolved | `Root` now owns the group semantics and the docs/tests match that model |
| Generated OTP reference JSON was unnecessary noise | Resolved | The OTP docs now rely on `types.md` only |
| Readonly slots lost Arrow/Home/End navigation | Resolved | Navigation keys now still work while readonly editing remains blocked |

### Merge Confidence Score

Overall merge confidence is 5/5. I do not see remaining blockers, and the branch now has strong coverage for the keyboard, focus, validation, docs, and form-integration paths that carry most of the risk for this component.

flaviendelangle (commented):
I'll look at the DX more in depth later

# Code Review: [otp field] Add OTP Field component (mui/base-ui#4365)

**Author:** atomiks | **+4,109 / -1** | 44 files

---

## Overview

This PR adds a new `OTPField` component (marked as Preview) with Root, Group, and Input parts. It includes full documentation (hero, alphanumeric, password, custom-sanitize demos), comprehensive tests (~1,240 lines), type-safety spec tests, SSR support, hidden validation input for native form integration, and Field/Form compatibility. The `length` prop on Root is the source of truth, while `CompositeList` handles slot ordering.

---

## Strengths

- **Excellent test coverage** - 912 lines of root tests + 328 lines of input tests covering value handling, controlled/uncontrolled modes, async state reconciliation, cancel semantics, keyboard interactions, pasting, SSR, Field integration, form submission, and data attributes.
- **Well-designed API** - Discriminated union event details (`ChangeEventDetails`, `InvalidEventDetails`, `CompleteEventDetails`) with proper TypeScript narrowing, verified by spec tests.
- **Solid controlled mode handling** - The `pendingFocusRef` / `pendingCompleteValueRef` pattern correctly handles async controlled updates (e.g., `onValueChange` with `setTimeout`), with tests for stale vs. accepted completions.
- **Good accessibility** - `role="group"`, `aria-labelledby` / `aria-describedby` propagation through Field and native labels, per-slot `aria-label`, `enterKeyHint` based on slot position.
- **Follows project conventions** - Uses existing utilities (`useStableCallback`, `useControlled`, `CompositeList`, `useField`, `useRenderElement`), proper data attributes, standard file structure, error code registered.

---

## Issues & Suggestions

### 1. ReadOnly fields skip focus tracking (bug)

In `OTPFieldInput.tsx:221-223`, `onFocus` returns early when `readOnly`:

```tsx
onFocus(event) {
  if (event.defaultPrevented || disabled || readOnly) {
    return;
  }
```

This means `data-focused` is never set on readonly fields, and `handleInputFocus` never runs. Readonly fields should still be focusable and trackable - only editing should be blocked. The `disabled` guard is correct (disabled inputs can't focus natively), but `readOnly` should be removed here. Same for `onMouseDown` - the `event.preventDefault()` and `focusInput` call should still work for readonly (they just select the text).

### 2. Native `contains` instead of shadow DOM-safe utility

In `OTPFieldRoot.tsx:283-284`:

```tsx
if (nextTarget instanceof Node && rootRef.current?.contains(nextTarget)) {
```

Per CLAUDE.md, the shadow DOM-safe `contains` utility should be used instead of the native `.contains()` method.

### 3. Double clipboard read in `onPaste`

In `OTPFieldInput.tsx:259-264`, `event.clipboardData.getData('text/plain')` is called twice. Store it in a variable:

```tsx
const rawValue = event.clipboardData.getData('text/plain');
const nextDigits = normalizeOTPValue(rawValue, length, validationType, sanitizeValue);
const didSanitize = stripOTPWhitespace(rawValue).length > nextDigits.length;
```

### 4. Demo `useInvalidFeedback` uses raw `setTimeout`

`useInvalidFeedback.ts:42-47` uses `setTimeout` / `clearTimeout` directly. CLAUDE.md recommends `useTimeout` from `@base-ui/utils/useTimeout`. This is demo code so it's less critical, but it's shown to users as a reference.

### 5. Missing test coverage for `alpha` validation on typing

The `alpha` validationType is tested via `defaultValue` splitting but there's no test for typing/pasting into an alpha-mode field.

---

## Minor Observations

- Significant CSS duplication across demo CSS modules (hero, alphanumeric, password share ~80% identical styles) - fine for demo independence.
- `{1}` quantifier in slot patterns (e.g. `\\d{1}`) is technically redundant but keeps consistency with the root pattern.
- The `onValueChange` JSDoc in `types.md` is missing line breaks between the bullet points for reason descriptions (renders as a run-on sentence in the generated table).

---

## Rating: 4/5

This is a high-quality, well-architected component addition with excellent test coverage and thoughtful API design. The readonly focus tracking bug (#1) and missing shadow DOM-safe `contains` (#2) are the only issues that should be addressed before merge. Everything else is polish.

docs/reference/generated/otp-field-group.json
```diff
@@ -0,0 +1,53 @@
+{
```
dav-is: You can delete the `docs/reference/` dir, we only need `types.md`

aarongarciah (commented):
I did a quick pass, leaving some feedback/questions before going deeper:
- I'm not quite sure I follow the logic that decides when the text is selected or not. Can you explain the reasoning behind when it should be selected or not? Is it useful to auto-select text on inputs? I see other OTP fields in the wild that don't do this and maybe they feel more natural. But I need to play more with them before having a strong opinion.

  https://github.com/user-attachments/assets/2d609b1d-dec9-4ccd-84d4-6d43f86e6c5c

- In the video above, there's a moment when I press numbers inside an input but nothing happens. Does it happen because the text inside the input is not selected? It felt surprising.
- What's the use case for `OTPField.Group`? Is there a scenario where several of them would be used? If that's the case, a demo would be nice.

docs/src/app/(docs)/react/components/otp-field/demos/hero/css-modules/index.module.css
```diff
@@ -0,0 +1,49 @@
+.Field {
```
aarongarciah, line 1: I find that matching the class name with the part name makes it easier to follow what's going on.


```suggestion
.Root {
```

atomiks:
> I'm not quite sure I follow the logic that decides when the text is selected or not

It should always be selected if you navigate via keyboard or click into an already-filled input. I actually can't get into the state you got in in the recording when pressing <kbd>Backspace</kbd> - it's always selected in that case 🤔

> In the video above, there's a moment when I press numbers inside an input but nothing happens. Does it happen because the text inside the input is not selected? It felt surprising.

I can't get into the state where this happens, do you have the exact steps/keys?

> What's the use case for OTPField.Group? Is there a scenario where several of them would be used? If that's the case, a demo would be nice.

It's mainly to match `NumberField` which has `Root` and `Group` parts. It's possibly not as necessary as NumberField since there's no `ScrubArea` part, so the Root itself could be `role="group"`. This would likely be less confusing regarding how ARIA roles get forwarded as well.

aarongarciah:
> It should always be selected if you navigate via keyboard or click into an already-filled input.

So is it expected that the last input gets selected when filling it?

https://github.com/user-attachments/assets/65d25096-ff75-4f54-9866-39d63f772f88

> I actually can't get into the state you got in in the recording when pressing Backspace - it's always selected in that case
> I can't get into the state where this happens, do you have the exact steps/keys?

You can see the exact keys being pressed in the video, but I can reproduce it consistently with this sequence:

1, 2, 3, 4, 5, 6, ⬅️, Del: at this point the selection is lost and pressing any number doesn't have an effect.

> It's mainly to match NumberField which has Root and Group parts. It's possibly not as necessary as NumberField since there's no ScrubArea part, so the Root itself could be role="group". This would likely be less confusing regarding how ARIA roles get forwarded as well.

Sounds like it's worth considering removing it.

atomiks:
> So is it expected that the last input gets selected when filling it?

I would say it's ambiguous for this particular case because there's no new input to move to and so it selects because it's filled and focused. It probably looks better without it though, so I'll try to stop it in this case.

> 1, 2, 3, 4, 5, 6, ⬅️, Del: at this point the selection is lost and pressing any number doesn't have an effect.

On macOS, is <kbd>Backspace</kbd> the same as  <kbd>Del</kbd> or is it different?  It's likely `'Delete'` vs. `'Backspace'` path, will look into it. As it doesn't happen here:

https://github.com/user-attachments/assets/1f0a252e-b052-4a74-88cc-b58860f95854

aarongarciah:
> On macOS, is Backspace the same as Del or is it different? It's likely 'Delete' vs. 'Backspace' path, will look into it. As it doesn't happen here:

On macOS, <kbd>Backspace</kbd> (backward delete ⌫)  and <kbd>Delete</kbd> (forward delete ⌦) are different, yes.

atomiks:
> On macOS, Backspace (backward delete ⌫) and Delete (forward delete ⌦) are different, yes.

I don't think I've ever used forward delete on macOS 🙈 fixed it though, <kbd>fn + delete</kbd> is the command on a MBP keyboard

atomiks:
@aarongarciah @dav-is @flaviendelangle addressed each

docs/src/app/(docs)/react/components/otp-field/page.mdx
```diff
@@ -0,0 +1,129 @@
...
+<Meta
+  name="description"
+  content="A high-quality, unstyled React OTP field component for one-time password and verification code entry."
+/>
+
+import { DemoOTPFieldHero } from './demos/hero';
+import { DemoOTPFieldAlphanumeric } from './demos/alphanumeric';
+import { DemoOTPFieldPassword } from './demos/password';
+import { DemoOTPFieldCustomSanitize } from './demos/custom-sanitize';
+import { TypesOTPField } from './types';
```
dav-is: I think the imports should be directly above their usage. For types, it should be right under the `API Reference` title

packages/react/src/otp-field/root/OTPFieldRoot.tsx
romgrk, line 1: Coming in late and I haven't followed the discussions, but did you consider implementing the OTP field as a single input instead of multiple ones? It has a lot of performance, bundle-size, maintainability & accessibility benefits.

atomiks: Fair question to ask. The current implementation is basically the status quo across the major libraries (Radix, Ark, Mantine). The odd one out (actually what shadcn happens to use) is `input-otp` but I'm not convinced it's a good idea.

### Maintainability

The most popular implementation of the real input idea appears to have some bugs that this PR's doesn't right now — it seems like syncing the input's state to real visual inputs is actually more challenging than just using real ones just from a basic first user experience glance.

(I know this isn’t proof of better or worse maintainability, but without redesigning the API around a single-input model just to validate that claim, I think it’s reasonable to treat a popular, battle-tested implementation as evidence that this approach also carries meaningful complexity in practice.)

https://github.com/user-attachments/assets/01ee0001-3e78-478f-a74c-8f1052c17c40

From Codex: 

"The strongest single-input-specific issues I found in `input-otp` are:

- Password manager badge overlap.
  The overlaid real input can end up with 1Password-style save/fill badges covering the last visible slot.

- Password manager autofill quirks.
  There are reports of some managers, like Dashlane, not filling 2FA correctly.

- Hidden-input width/layout hacks.
  The invisible absolute input uses width expansion tricks, which has led to overflow and layout-shift issues.

- Caret/selection artifacts on iOS.
  Because there is still one real input underneath, its caret/selection can leak through visually on mobile Safari/iOS.

- Clicking a visual slot maps to the wrong logical position.
  Translating a click on a fake slot back into the correct selection/caret position in one real input turns out to be surprisingly tricky, and there are multiple reports around that.

- Visual focus not matching actual typing focus.
  There are cases where the component still appears focused after programmatic resets, but typing no longer works until the user clicks again.

- Extra CSS/runtime plumbing for autofill styling.
  The architecture needs more special handling around the real hidden/overlay input, which has led to CSS injection, CSP, and WebView-specific issues.

The best piece of evidence is actually the maintainer’s own “limitations/issues caused by core implementation” note, which explicitly calls out architecture-driven downsides like:
- visible iOS selection artifacts
- inability to truly click/touch-select arbitrary slots

So the maintainability point I’d make is:

I’m also not convinced the maintainability advantage is as clear-cut as it sounds. The single-input model clearly works, but it also seems to introduce its own architecture-specific complexity around the overlaid native input: password-manager interactions, width/layout hacks, iOS caret/selection artifacts, and mapping clicks on visual slots back to one real input. I don’t think that proves the model is worse, but it does suggest the tradeoff is more “different complexity” than “obviously simpler to maintain.”
- compressed-letter/caret hacks needed to make the illusion work"

### Performance

This as far as I'm concerned is not relevant, because OTP fields won't be rendered more than once on a single page in 99% of scenarios — I cannot imagine why they would be. 

But based on local bechmarks, it doesn't seem like it brings a benefit anyway:

| Benchmark | Base UI OTPField | input-otp | Relative result |
|---|---:|---:|---|
| Repeated mount/unmount 1,000x | 775 ms | 1164 ms | Base UI ~1.50x faster |
| Repeated mount/unmount per instance | 0.775 ms | 1.164 ms | Base UI ~1.50x faster |
| Bulk mount 1,000 instances | 649 ms | 639 ms | Roughly tied |
| Bulk mount per instance | 0.649 ms | 0.639 ms | Roughly tied |
| DOM elements per instance | 8 | 11 | Base UI renders fewer elements |

(Again, have to defer to an existing solution for this claim - it's not proof on its own)

### Accessibility

This is the strongest concern, but my hunch is that it's not a clear issue. 

Playing around with VoiceOver myself on this comp reads just like it visually appears: "_Character 1 of 6, edit text, Verification code, Enter the 6 character code we sent to your device, group_". This indicates it's a grouped composite widget that has input-like keyboard interactions with individual digit slots quite clearly. The one on the shadcn page is more like a real input field but honestly that seems less like a plainly obvious "this is a 6-digit code with individual slots" widget. Additionally, VoiceOver's virtual cursor appears misplaced with the shadcn one while ours correctly highlights the correct slot.

Edit: After testing NVDA, JAWS, and Narrator on Windows, all have similar semantics to macOS VoiceOver and all make coherent sense regarding what the component does/requires of users. So the multi-input model seems to be AT accessible.

romgrk: I think we could do better with a single input implementation, but this component is not critical enough to spent more time on it if the current PR already works.

oliviertassinari:
I was curious to see if https://github.com/guilhermerodz/input-otp/issues/92 would be an issue here. It behaves like this with Bitwarden:

https://github.com/user-attachments/assets/9f7f9eda-24ba-4e89-9d3e-67a6e3551b11

Even though Shadcn UI uses https://github.com/guilhermerodz/input-otp, the experience is a bit better:

https://github.com/user-attachments/assets/f6cd660d-99d2-4f03-ba02-ae6ddfdb816e

It's the last input that is focused at the end. So overall it's OK, it works (and in the future, we have room to push it one step further).

packages/react/src/otp-field/root/OTPFieldRoot.tsx
oliviertassinari: On bundle size, we do have a large difference:

- input-otp: [3.91 kB gzip](https://bundlejs.com/?q=input-otp%401.4.2&config=%7B%22esbuild%22%3A%7B%22external%22%3A%5B%22react%22%2C%22react-dom%22%5D%7D%7D) or [3.80 kB gzip](https://bundlephobia.com/package/input-otp@1.4.2)
- Base UI OTP: [7.79 kB gzip](https://frontend-public.mui.com/size-comparison/mui/base-ui/diff?prNumber=4365&baseRef=master&baseCommit=1e0f08cce3d8bc7378db747d5f81365030e638d7&headCommit=d72a4bae237a7847444fdc1180b9a3a693116cdf)

But I can't see anything obviously wrong: https://blind-amethyst-p4guydfdkf.edgeone.app/.

- The new useMergedRefs() utils in Base UI seem huge compared to the one in Material UI useForkRef(), but ok, it's reused everywhere.
- else?

atomiks: It's +2.65 KB as an additive component when using Base UI more broadly, which is less than input-otp. So it's better for shadcn-like usage.

But yes standalone, our component API and related machinery increases size a lot.

I cut 7.79 KB down to 7.64 KB in the latest commit.

Edit: I also added some maintainability points above from Codex based on input-otp issues, which shows it has some fundamental limitations

michaldudak:
## Critical Issues (1 found)

  - [error-handling] User-supplied sanitizeValue callback can throw, crashing the component tree. It's called unguarded from 6 locations including during render (otp.ts:58-59). If it throws, the entire tree
  unmounts with no indication the callback is the culprit.

## Important Issues (5 found)

  - [error-handling] clipboardData.getData('text/plain') called without null guard (OTPFieldInput.tsx:250). In edge cases clipboardData can be null, causing a TypeError.
  - [types/validation] length prop typed as number with no dev-mode validation for positive integers. Values like 0, -1, 3.7, NaN, Infinity produce silent degradation. This is the most important invariant of the
  component — everything depends on it. Add a dev warning similar to the existing inputCount !== length warning.
  - [types/DX] sanitizeValue is silently ignored when validationType is not 'none' (otp.ts:56-60). No runtime warning for this misconfiguration — developers will be confused.
  - [tests] No unit tests for otp.ts utility functions (normalizeOTPValue, replaceOTPValue, removeOTPCharacter, stripOTPWhitespace). These are pure foundational functions that are cheap to test and critical to
  correctness. Edge cases like Unicode/emoji, negative length, out-of-bounds index are untested.
  - [tests] onValueInvalid and onValueComplete not tested for paste interactions. The paste codepath has distinct logic and fires different reason values (input-paste) that are part of the public API contract.

## Suggestions (8 found)

  - [tests] No test for data-complete attribute presence/absence on root and inputs (7/10 criticality)
  - [tests] Paste into a middle slot is untested — exercises the splice logic in replaceOTPValue (6/10)
  - [tests] Delete/Backspace not verified as blocked in readOnly mode (6/10)
  - [tests] No negative test for orphaned <OTPField.Input> context error (5/10)
  - [comments] data-focused on Input means "any input in the group is focused", not "this specific input" — description should clarify (OTPFieldInputDataAttributes.ts:9)
  - [comments] onValueInvalid JSDoc says "clamped" which could mislead — truncation to length does not trigger this callback (OTPFieldRoot.tsx:618)
  - [comments] id prop JSDoc should mention derived IDs for subsequent inputs ({id}-2, {id}-3)
  - [docs] types.md auto-generated onValueChange description has list items concatenated into unreadable text (generation tooling issue)

## Strengths

  - Project guideline compliance is excellent. useStableCallback, useIsoLayoutEffect, useTimeout, shadow DOM utilities, ownerDocument — all used correctly. No violations found.
  - Discriminated event detail types are well-crafted. The .spec.tsx files prove narrowing works via @ts-expect-error annotations.
  - Controlled/async pattern is solid. The pendingFocusRef/pendingCompleteValueRef pattern correctly handles race conditions where external state updates arrive asynchronously.
  - Context design is clean — no mutable refs exposed (unlike NumberFieldRootContext), and getOTPFieldInputState is a nice helper for deriving child state.
  - Error message in OTPFieldRootContext.ts follows the required "Base UI:" format perfectly and is registered in error-codes.json.
  - Comprehensive test coverage of value handling, keyboard nav, form integration, accessibility, SSR, disabled/readOnly states, and cancel/veto patterns.

flaviendelangle (commented):
# PR Review: [otp-field] Add OTP Field component (#4365)

**PR:** https://github.com/mui/base-ui/pull/4365
**Author:** atomiks
**Scope:** 4,656 additions across 49 files — new `OTPField` component with Root, Input, and Separator parts, plus docs, demos, tests, and generated API reference.

---

## Critical Issues (3 found)

### 1. User-provided `sanitizeValue` can throw and crash the component (including during render)

**Files:** `packages/react/src/otp-field/utils/otp.ts` (`normalizeOTPValue`), `OTPFieldRoot.tsx`, `OTPFieldInput.tsx`

The user-supplied `sanitizeValue` callback is called without any try-catch protection. It executes during **rendering** (via `normalizeOTPValue` in `OTPFieldRoot.tsx`) and in every event handler (`onChange`, `onPaste`, hidden input `onChange`). If the callback throws, the entire React tree crashes with no indication the issue is in the user's function.

**Recommendation:** Wrap the `sanitizeValue` call in a try-catch in dev mode, log a warning with the error, and fall back to the unsanitized (whitespace-stripped) value.

### 2. `onValueInvalid` JSDoc is misleading

**Files:** `OTPFieldRoot.tsx` (prop JSDoc), `types.md` (generated docs)

The JSDoc says *"Callback fired when entered text is sanitized before the OTP value updates."* This implies it fires every time sanitization occurs. In reality, it only fires when sanitization **removes** characters (the implementation checks `stripOTPWhitespace(rawValue).length > nextDigits.length`).

**Recommendation:** Reword to: *"Callback fired when entered text contains characters that are rejected by sanitization, before the OTP value updates."*

### 3. Generated `types.md` has broken formatting for `onValueChange` reason list

**File:** `docs/src/app/(docs)/react/components/otp-field/types.md`

The bullet-point list of `eventDetails.reason` values renders as a single unbroken sentence:
```
`'input-change'` for typing or autofill`'input-clear'` when a character is removed...
```
The doc generation pipeline strips the list structure from the JSDoc source.

**Recommendation:** Fix the generation pipeline to preserve lists, or reformat the description to use semicolons/commas as inline delimiters.

---

## Important Issues (10 found)

### 4. Clipboard `getData()` can throw in restricted browser contexts

**File:** `OTPFieldInput.tsx`, `onPaste` handler

```ts
const rawValue = event.clipboardData?.getData('text/plain') ?? '';
```

In cross-origin iframes or browsers with restrictive clipboard permissions, `getData()` can throw a `DOMException`. The `??` only handles nullish `clipboardData`, not exceptions from `getData()`.

**Recommendation:** Wrap in try-catch and return early on failure.

### 5. `requestSubmit` silently no-ops when no form is found despite `autoSubmit`

**File:** `OTPFieldRoot.tsx`, `requestSubmit` function

If `autoSubmit={true}` and `form="my-form"` but the ID is misspelled or the element isn't a `<form>`, submission silently fails. The user fills the OTP, `onValueComplete` fires, but the form never submits — very difficult to debug.

**Recommendation:** Add dev-mode warnings when `autoSubmit` is enabled but no form is found, or when the `form` prop points to a non-form element.

### 6. `focusInput` silently no-ops when the target ref is null

**File:** `OTPFieldRoot.tsx`, `focusInput` function

Uses `target?.focus(); target?.select()` which silently skips when the ref is null. If this happens during normal operation, the component appears to "freeze" with no visible error.

**Recommendation:** Add a dev-mode warning when `target == null`.

### 7. Invalid `length` causes silent hidden input omission in production

**File:** `OTPFieldRoot.tsx`, `OTPFieldHiddenInput`

When `length` is not a positive integer, the hidden input returns `null` silently. Form submission, native validation, password manager autofill, and the `name` prop all stop working while visible inputs still render normally.

**Recommendation:** Consider throwing in development rather than just warning, since `length` is a required prop that fundamentally breaks the component when invalid.

### 8. `onValueChange` throwing leaves component in inconsistent state

**File:** `OTPFieldRoot.tsx`, `setValue` function

If the user's `onValueChange` callback throws, the subsequent `setValueUnwrapped` and `pendingCompleteValueRef` updates never execute, leaving internal state inconsistent with what the user sees. This is consistent with React's general callback behavior but worth noting for a new component.

### 9. No tests for `OTPField.Separator` rendering or conformance

**Files:** No Separator test file exists

The Separator is exported as a public part with OTP-specific state attributes (`complete`, `disabled`, `readOnly`, etc. per `types.md`), but no test verifies its state attributes, ref forwarding, or conformance in the OTP context.

### 10. No test for paste blocked in `readOnly` mode

**File:** `OTPFieldInput.tsx` has an early return for `readOnly` in `onPaste`, but no test verifies it. Typing is tested for readonly, but a regression in the paste path could allow modifying a readonly field via paste.

### 11. No test for `onValueComplete` not re-firing on complete-to-complete transitions

If the controlled value changes from one complete value to another (`"123456"` to `"654321"`), the `pendingCompleteValueRef` mechanism should not re-fire `onValueComplete`. This contract is untested.

### 12. `OTPFieldInputState` silently shadows `value` and `filled` from parent type

**File:** `OTPFieldRootContext.ts`

`OTPFieldInputState extends OTPFieldRootState` but overrides `value` (full OTP string -> single character) and `filled` (any character exists -> this slot has a character) with different semantics. TypeScript allows this silently since both are `string`/`boolean`.

**Recommendation:** Use `Omit<OTPFieldRootState, 'value' | 'filled'>` in the extends clause to make the override explicit:
```ts
export interface OTPFieldInputState extends Omit<OTPFieldRootState, 'value' | 'filled'> {
  filled: boolean;   // Whether THIS input contains a character
  index: number;
  value: string;     // The character in THIS slot
}
```

### 13. `setValue` return type `string | null` lacks documentation

**File:** `OTPFieldRootContext.ts`

The context exposes `setValue: (value: string, details: ...) => string | null` where `null` means either "no change occurred" or "change was canceled." This convention is non-obvious from the type alone.

**Recommendation:** Add a JSDoc comment explaining the return semantics.

---

## Suggestions (12 found)

### 14. Expand `otp.test.ts` utility edge cases

Only 6 tests for 91 lines of source. Missing coverage for: `replaceOTPValue` at boundaries (index 0, last index), `removeOTPCharacter` at boundaries, `normalizeOTPValue` with `null`/`undefined` input and `alpha`/`alphanumeric` types, `stripOTPWhitespace` with null/undefined.

### 15. Test cancel during paste + `onValueComplete` interaction

The cancel test for paste verifies values don't change, but doesn't verify `onValueComplete` is NOT fired even if the pasted value would have been complete.

### 16. Test `onBlur` validation commit path

When `validationMode === 'onBlur'`, `validation.commit(valueRef.current)` is called on blur. No test verifies this Field integration, and the OTP-specific blur behavior (only when focus leaves the entire root) could regress.

### 17. Test paste blocked in `disabled` mode

The `onPaste` handler returns early for `disabled`, but disabled tests only check `fireEvent.change`.

### 18. Test backspace on empty non-first slot

The handler has `const deleteIndex = slotValue === '' ? Math.max(0, index - 1) : index` — backspacing on an empty slot 3 when slot 2 has a value should delete slot 2's character. Only the first-empty-slot case is tested.

### 19. Add inline comments on complex event handler logic in `OTPFieldInput.tsx`

Key non-obvious behaviors lack comments:
- `maxLength: index === 0 ? length : 1` — allows autofill of full code into first input
- `slotValue === event.key` check on keyDown — advances focus when re-typing same character
- Backspace shift-left behavior — maintains contiguous fill

### 20. Add block comment on `useValueChanged` async reconciliation pattern

This is the most sophisticated part of the implementation (deferred focus/completion for async controlled mode) and has no comments explaining the pattern.

### 21. Add JSDoc to `replaceOTPValue` and `normalizeOTPValue` utilities

These are key functions with non-obvious behavior (codepoint-aware slicing, conditional sanitization pipeline).

### 22. Document that first input's `aria-label` is silently stripped

The docs say "Let the first input use the field label" but don't mention that `aria-label` on the first input is **ignored** by the implementation. All demos pass `aria-label` on every input including the first (which is silently stripped).

### 23. Add JSDoc to non-obvious `OTPFieldRootContext` members

The interface has 21 members with no JSDoc. At minimum: `queueFocusInput`, `reportValueInvalid`, `inputAriaLabelledBy`, `pattern`.

### 24. Demo uses `setTimeout` instead of `useTimeout` utility

**File:** `docs/src/app/(docs)/react/components/otp-field/demos/custom-sanitize/useInvalidFeedback.ts`

Per AGENTS.md, `useTimeout` should be used instead of `window.setTimeout`. While this is demo code, demos serve as examples for users.

### 25. `sanitizeValue` in context memoization may cause unnecessary re-renders

`sanitizeValue` comes from props and is included in the context `useMemo` dependency array. Any inline arrow function from the consumer will bust the memo on every render. Consider documenting this or wrapping it.

---

## Strengths

### Code Quality
- **Full CLAUDE.md/AGENTS.md compliance**: Uses `useStableCallback`, `useIsoLayoutEffect`, `useTimeout` (in library code), shadow DOM-safe utilities (`contains`, `ownerDocument`), and Vitest-only test APIs. No violations found.
- **Clean architecture**: Root owns all state and shares via context; Input is a pure consumer. Focus management is centralized. The `CompositeList` integration for slot ordering is elegant.
- **Correct `React.useCallback` vs `useStableCallback` usage**: `getInputId` (called during render) uses `useCallback`; all effect/handler callbacks use `useStableCallback`.

### Type Design
- **Required `length` prop** enforced at compile time (verified by spec test with `@ts-expect-error`).
- **Index inferred from render order** rather than accepted as a prop — prevents ordering bugs entirely.
- **Precisely scoped reason unions**: Each callback's reason type only includes semantically valid reasons (e.g., `onValueInvalid` excludes `'keyboard'` because keyboard events don't produce invalid values).
- **Full reuse of centralized event infrastructure** (`BaseUIChangeEventDetails`, `BaseUIGenericEventDetails`, `REASONS`, `ReasonToEventMap`).
- **Consistent with NumberField patterns**: Namespace structure, context creation, data attribute enums, state attribute mappings, and `Omit<..., 'onChange'>` on root props all match.

### Testing
- **~1,500 lines of tests** covering keyboard navigation, paste, validation, completion callbacks, controlled/uncontrolled modes, form integration, accessibility labeling, readonly/disabled states, and data attributes.
- **Controlled mode edge cases are excellent**: Stale controlled value tests with `vi.useFakeTimers()` verify the `pendingFocusRef`/`pendingCompleteValueRef` reconciliation.
- **Cancel semantics tested** for typing, backspace, and paste.
- **Type-level spec tests** verify discriminated union narrowing on event detail reasons.
- **Dev warnings tested**: Length mismatch, invalid length values, and sanitizeValue/validationType misuse.

### Accessibility
- Label propagation via `aria-labelledby` to all inputs.
- Description goes on the group element.
- Hidden input supports form validation and password manager autofill.
- `autoComplete="one-time-code"` default is semantically correct.

### API Design
- Three distinct callbacks (`onValueChange`, `onValueInvalid`, `onValueComplete`) with precise event details.
- `onValueChange` supports cancel mechanism via `details.cancel()`.
- `validationType` with four clean variants plus `sanitizeValue` escape hatch.
- `autoSubmit` for automatic form submission on completion.

---

## Recommended Action

1. **Fix critical issues** (#1-#3) before merge
2. **Address important issues** (#4-#8 error handling, #9-#11 test gaps, #12-#13 type clarity)
3. **Consider suggestions** (#14-#25) for polish
4. Re-run `pnpm docs:api` after JSDoc changes to regenerate `types.md`

atomiks:
@flaviendelangle a few good points there, but a bunch are not relevant. For the "critical error", we shouldn't swallow the sanitize error as users should handle it themselves.

Updated the branch

Codex:
```
A lot of this is stale or over-rotated against the current branch.

What I’d treat as real:
- `#2` `onValueInvalid` wording: fair. “sanitized” is a bit broader than what actually triggers it.
- `#10` readonly paste test: fair small gap if it’s still not covered.
- `#11` complete-to-complete transition test: fair regression test to add.
- `#12` `OTPFieldInputState` using `Omit<...>`: reasonable type-clarity cleanup, but not urgent.
- `#4` `clipboardData.getData()` try/catch: optional hardening, not critical.

What I would not take:
- `#1` catch `sanitizeValue` throws: no. User callback exceptions crashing normally through React is standard; swallowing them would hide the real bug.
- `#5` warn when `autoSubmit` can’t find a form: maybe, but low-value and noisy.
- `#6` warn when `focusInput` target ref is null: no, too noisy/internal.
- `#7` throw for invalid `length`: no. Dev warning is enough.
- `#8` `onValueChange` throwing: not a component bug.
- `#13` document internal `setValue` return semantics: internal context, not worth it.

What looks outdated / incorrect now:
- `#3` broken `types.md` formatting for `onValueChange`: this was already improved.
- `#9` no Separator tests: `Separator` is just a re-export; I wouldn’t add OTP-specific conformance for that.
- `#22` docs should mention first input `aria-label` is stripped: not true anymore on the current branch, since we changed to warn rather than silently strip.
- `#24` demo should use `useTimeout`: stale; we intentionally did not go that route.
- `#25` sanitizeValue in memo deps: not really actionable.
```

michaldudak:
### 1. Silent clipboard paste failure — bare catch block

  [silent-failure-hunter] HIGH — OTPFieldInput.tsx:277-281

  The `onPaste` handler calls `preventDefault()` (killing native paste), then wraps `clipboardData.getData()` in a bare catch that silently returns. If clipboard access is denied (e.g., SecurityError on mobile browsers), the user presses Ctrl+V and nothing happens — no feedback, no console message.

```
  event.preventDefault();  // native paste already killed
  try {
    rawValue = event.clipboardData?.getData('text/plain') ?? '';
  } catch {       // bare catch, no logging
    return;       // silent return
  }
```

  Recommendation: At minimum, log the error in development so developers can diagnose clipboard permission issues.

### 2. autoSubmit silently does nothing when no form is found

  [silent-failure-hunter] MEDIUM — OTPFieldRoot.tsx:179-192

  When `autoSubmit={true}` and the OTP completes, `requestSubmit()` silently no-ops if the field isn't inside a `<form>` and no form prop is provided. A dev-mode warning (matching the existing pattern for length mismatch) would save debugging time.


## Test Coverage Gaps

  - onValueComplete via keyboard reason (criticality 8/10) — The REASONS.keyboard complete path is untested. A Backspace + retype sequence that refills the OTP should verify onValueComplete fires with the correct
  reason.
  - Hidden input onValueInvalid (8/10) — The hidden input's onChange independently calls onValueInvalid during autofill, but this path is untested. The "password manager autofill" test only checks
  onValueChange/onValueComplete.
  - Paste completing from a middle slot (7/10) — Existing paste-to-complete test only pastes from slot 0. Pasting from a non-zero index exercises a different branch in replaceOTPValue.
  - data-disabled / data-readonly state attributes (5/10) — Tests verify the native disabled/readonly attributes but not the data-* state attributes used for CSS styling.
  - useOTPFieldRootContext error when used outside Root (5/10) — The descriptive error thrown by the context is untested.

## Comment Improvements

  - OTPFieldInput.tsx:107 — `maxLength: index === 0 ? length : 1` — Add a comment explaining the first input accepts the full length for browser autofill support.
  - otp.ts:66 — `Array.from(sanitizedValue).slice(...)` — Add a comment noting this handles multi-byte/surrogate-pair characters correctly (vs. `String.slice` which counts UTF-16 code units).

atomiks:
> 1. Silent clipboard paste failure — bare catch block

Fixed

> 2. autoSubmit silently does nothing when no form is found

This one too noisy, if you want `autoSubmit` unconditionally specified in a wrapped component API, even if it's not in a form

> Test Coverage Gaps

Fixed

> Comment Improvements


Fixed

michaldudak (approved):

PR #4366, opened 2026-03-18
[docs][calendar] Replace chevron icons with custom SVG components

LukasTy:
Follow-up on https://github.com/mui/base-ui/pull/4358#issuecomment-4080380695.


LukasTy:
## Replace lucide-react chevron icons with inline SVGs in calendar demos

Removes the `lucide-react` dependency (`ChevronLeft`, `ChevronRight`) from all CSS Modules calendar demos, replacing them with self-contained `ChevronLeftIcon` / `ChevronRightIcon` inline SVG components.

### Changes

- 13 demo files updated across `calendar` and `localization-provider` components
- Each demo now defines its own chevron SVG icons instead of importing from `lucide-react`
- No library source code or API changes

### Motivation

Demo code is meant to be copied by users. Removing the third-party icon dependency makes demos fully self-contained and easier to use without installing additional packages.

PR #4367, opened 2026-03-18
[select] Reset preserved touch modality after close transition

atomiks:
## Summary

Fixes a Select touch regression where reopening after the close transition can reuse the preserved touch interaction type and reopen from the wrong positioning mode.

This regressed in `4fd87cad4` (`[select] Preserve touch exit animations`). That commit landed after `v1.3.0`, so this is a post-`1.3.0` regression. I'm marking this PR `internal` and `type: regression`; it should not need release notes.

## Changes

- Stop preserving the previous touch `openMethod` once the close transition has fully finished and the popup is no longer mounted.
- Add a Chromium regression test that covers touch open, close, and reopen near the bottom of the viewport.

### DIFF
--- a/docs/src/app/(docs)/react/components/calendar/demos/validation/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/calendar/demos/validation/css-modules/index.tsx
@@ -6,8 +6,9 @@ import { ChevronLeft, ChevronRight } from 'lucide-react';
 import indexStyles from './index.module.css';
 import styles from '../../calendar.module.css';
 
+const today = new Date();
+
 export default function MinMaxDateCalendars() {
-  const today = new Date();
   return (
     <div className={indexStyles.Wrapper}>
       <ValidationCalendar minDate={today} />
--- a/docs/src/app/(private)/experiments/calendar/calendar-validation.tsx
+++ b/docs/src/app/(private)/experiments/calendar/calendar-validation.tsx
@@ -6,8 +6,9 @@ import { Calendar } from '@base-ui/react/calendar';
 import { ChevronLeft, ChevronRight } from 'lucide-react';
 import styles from './calendar.module.css';
 
+const today = new Date();
+
 export default function CalendarValidation() {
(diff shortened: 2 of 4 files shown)


atomiks:
## Codex Review

### Overview
This branch fixes the Select touch reopen bug by keeping a hidden keep-mounted popup from recalculating into the wrong side and tiny height while it is fully closed. It also adds a Chromium regression that matches the real long-list repro near the bottom of the viewport.

### Findings (None)

### Confidence: 4/5
High confidence based on a full-branch pass over the `master...HEAD` diff, direct review of the shared positioning changes, and targeted verification in both Select and NavigationMenu paths. The main residual risk is that this touches shared keep-mounted positioning behavior, but the diff is small and the follow-up checks cover the most likely regression path.

### Notes
- The branch diff against `master` is still narrow: `packages/react/src/utils/useAnchorPositioning.ts` passes `open: keepMounted ? mounted : undefined` to `useFloating`, and the `size()` middleware now uses a simple `mountedRef.current` guard without an explicit dep array so closed popups stop overwriting their size variables while hidden.
- `packages/react/src/select/root/SelectRoot.test.tsx` adds a stronger browser regression with the measured `100px` viewport-gap setup, touch dismiss, and reopen assertions against the long-list path that was actually failing.
- I reviewed the explicit second-pass categories for this final diff: React/rendering correctness, accessibility/semantics, API/state design, maintainability/simplification, performance, and test quality. I did not find additional branch-relevant issues beyond the bug this patch fixes.
- Verified with `pnpm test:chromium SelectRoot --no-watch -t "recomputes positioning before the popup becomes visible again after touch dismiss"`, `pnpm test:jsdom SelectRoot --no-watch`, `pnpm exec eslint packages/react/src/utils/useAnchorPositioning.ts packages/react/src/select/root/SelectRoot.test.tsx`, and `pnpm exec prettier --check packages/react/src/utils/useAnchorPositioning.ts packages/react/src/select/root/SelectRoot.test.tsx`.

PR #4369, opened 2026-03-18
[docs] Scaffold new docs theme

aarongarciah:
https://deploy-preview-4369--base-ui.netlify.app/docs-theme

- Font weight doesn't work with our currently loaded Die Grotesk font files.
- Söhne Mono is not loaded.
- Shadows in dark mode are fully transparent.
- Dark mode colors were generated by Codex based on the light ones. Probably not quite right.
- Font sizes and line heights from (website) are included in the scales.

### DIFF
--- a/docs/src/app/(private)/docs-theme/page.module.css
+++ b/docs/src/app/(private)/docs-theme/page.module.css
@@ -12,14 +12,14 @@
 .hero h1 {
   margin: 0;
   font-family: var(--font-sans-b);
-  font-size: var(--font-size-9);
-  font-weight: var(--font-weight-2);
+  font-size: var(--font-size-42);
+  font-weight: var(--font-weight-700);
 }
 
 .kicker {
   margin: 0 0 var(--space-8);
   color: var(--gray-t1);
-  font-size: var(--font-size-1);
+  font-size: var(--font-size-12);
   letter-spacing: 0.06em;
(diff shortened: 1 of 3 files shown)


colmtuite:
- Only need two weights
- Do you wanna do space-24, space-28 naming? 😅

aarongarciah:
@colmtuite 

> Only need two weights

Updated

> Do you wanna do space-24, space-28 naming?

Of course. Only for space? 😏

colmtuite:
For whichever ones you want I guess

aarongarciah:
[Updated](https://github.com/mui/base-ui/pull/4369/changes/fdd0b48c49fe0397ae09bb7061374a7f86d37930) to use "raw" names for space & radius, for now. It'd be nice for font-size as well, but then line-height gets weird if the steps between the two scales are meant to be used together most of the times.

colmtuite:
We don't need line-height vars? Can just bake the line-height into the .Text.size-x classes.

aarongarciah:
IMO it's a bit too much to force all components to use `.Text`, but happy to try that path. 

PR updated: letter spacing tokens are gone and now font size and weight scales use the mad man naming convention.

colmtuite:
Only Text components would use `.Text`? Other components don't need line-height?

PR #4377, opened 2026-03-19
[menubar] Add `data-has-submenu-open` data attribute when submenu is open

chuganzy:
This PR adds missing Menubar data attributes to the doc.

There were two issues:
* The enum name was invalid and therefore the json was missing the field
* `data-has-submenu-open` was missing


atomiks:
## Codex Review (GPT-5.4)

This PR is now a small `bug-fix` to the public Menubar data-attribute contract rather than a docs-only change: it makes `data-has-submenu-open` behave like a presence selector, aligns the generated API docs with that runtime, and adds regression coverage for the updated semantics. The latest state looks coherent, and I do not see a remaining blocker in the patch itself.

I don’t have a remaining branch-level bug to block on after the runtime, docs, and tests were updated together.

The original problem was that `data-has-submenu-open` was being documented as a presence attribute while the runtime actually emitted string values (`"true"` / `"false"`). The latest patch resolves that mismatch by making the runtime presence-only in `packages/react/src/menubar/Menubar.tsx:20-23`, updating the enum/docs source in `packages/react/src/menubar/MenubarDataAttributes.ts:11-14`, and adjusting the menubar tests to assert presence/absence rather than boolean strings in `packages/react/src/menubar/Menubar.test.tsx:81-88` and `packages/react/src/menubar/Menubar.test.tsx:775-806`.

The attribute itself is a DOM reflection of the menubar-wide `hasSubmenuOpen` state, not an item-level state flag. In practice it exists as a styling hook for “some submenu in this menubar is currently open”, mirroring the same internal state that the component already uses for hover-open behavior and related interaction policy (`packages/react/src/menubar/Menubar.tsx:47-58`, `packages/react/src/menubar/Menubar.tsx:89-93`, `packages/react/src/menubar/Menubar.tsx:105-117`).

| Area | Assessment | Notes |
|------|:----------:|-------|
| Presence-style data attributes | Improved | `data-has-submenu-open` now matches the same “present when active” convention used by other Base UI state attrs |
| Generated docs alignment | Improved | `docs/reference/generated/menubar.json:44-54` now describes the actual runtime contract |
| Enum naming / extraction | Improved | Renaming the enum to `MenubarDataAttributes` fixes the docs-generation path and makes the source naming consistent |

**Strengths:**
- Reusing `MenubarDataAttributes.hasSubmenuOpen` directly in the runtime mapping removes the string literal drift that caused the docs issue in the first place.
- The patch stays tightly scoped to the attribute contract and its tests.

**Low risk:**
- I don’t see a worthwhile simplification to push further here; the current diff is already minimal and direct.

| Area | Status | Notes |
|------|:------:|-------|
| Breaking selector change | Worth noting | Consumers who happened to style `[data-has-submenu-open="false"]` / `="true"` will need to switch to presence/absence selectors |
| Previously documented? | Limited blast radius | The attribute was previously exposed in runtime/tests but not in the generated public data-attribute docs, which lowers the risk compared with a documented stable selector contract |
| Release-note value | Helpful | I would not block on this, but a short release note or migration mention would make the semantic change easier to discover for power users |

| Validation | Status | Notes |
| `pnpm test:jsdom Menubar --no-watch` | Passed | 69 passed, 27 skipped |
| `pnpm typescript` | Passed | Full repo typecheck passed |
| Regression coverage for presence semantics | Good | Updated assertions now cover both “submenu open” and “submenu closed” attribute states |

| Previous issue | Latest status | Notes |
|----------------|:-------------:|-------|
| Docs claimed presence semantics while runtime always emitted the attribute | Resolved | Runtime now omits the attribute when no submenu is open |
| Generated Menubar data-attribute table was incomplete due to enum/source issues | Resolved | Generated reference now includes `data-orientation`, `data-has-submenu-open`, and `data-modal` |
| Unclear whether changing the selector shape was acceptable | Acceptable with limited risk | It is technically breaking, but the old string-value behavior was not previously documented in the generated API docs |

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Runtime/docs alignment | 5/5 | The latest patch aligns source, runtime, tests, and generated docs |
| Breaking-change risk | 3/5 | Real but limited to consumers who discovered and relied on the old undocumented boolean-string selector |
| Validation coverage | 4/5 | Targeted menubar tests and typecheck passed |

Overall merge confidence is 4/5. I’m comfortable with the latest patch as long as we accept the small selector-contract break for previously undocumented usage; if we want extra caution, the remaining improvement would be a short release-note callout rather than another code change.

chuganzy:
> but given it wasn't documented and doesn't match other attrs, we could possibly change this

I agree. It's also not yet mentioned in the release doc.
~~To make sure the change is put in the release note, I will create a separate PR soon!
In the meanwhile, changing this PR to the draft state.~~

chuganzy:
> To make sure the change is put in the release note, I will create a separate PR soon!

Since the change itself was almost identical to this PR, I decided to just rename this PR.

PR #4384, opened 2026-03-19
[docs] Die Grotesk font

aarongarciah:
https://deploy-preview-4384--base-ui.netlify.app/react/overview/quick-start

Replaces Unica 77 with [Die Grotesk](https://klim.co.nz/fonts/die-grotesk/).

- Subfamilies/weights used:
  - Docs shell: A/400 & A/700 for copy, B/700 for headings.
  - Demos: A/400, A/700.
- All demos have been adapted to stop using `font-weight: 500`. In some cases, it's been replaced with `400` (buttons, triggers, accordions, menu items, etc.) and with `700` in others (form labels, dialog titles, etc.).


colmtuite:
- I think the weight is 700, not 600?
- SearchTrigger should move to 400
- Gotta remove the letter-spacing from everywhere, or at least everywhere that it's easily achieved.
- Remove bold from .SideNavLink&[data-active]
- Try sidenav headings with 400 weight and low-contrast gray

aarongarciah:
@colmtuite all feedback applied.

- The sidenav heading now uses 400 and low-contrast gray (gray-500, gray-600 doesn't look as good). It doesn't pass AA, though.
- Some letter spacings are left for special cases like badges.
- Letter spacing in demos is untouched. Letter spacing for mono font is also untouched

colmtuite:
> gray-600 doesn't look as good

True but gotta pass AA, so should be 600 for now, same as the headings in the search dialog. Can fix when we have our new theme

colmtuite:
Maybe Avatar fallback can be 400 weight?

What other issues did you see? I think it's ok now?

colmtuite:
Actually, `vertical-align` needs to be removed from `.Kbd`

aarongarciah:
> Maybe Avatar fallback can be 400 weight?

> Actually, vertical-align needs to be removed from .Kbd

Both updated.

> What other issues did you see?

Some of the things I didn't like are fixed now, like the aria-current state of the sidenav links. I changed other stuff like the "latest" badge on the releases page to just use 400, since 700 looked too much.

The rest are just demos that could maybe benefit from a 500 weight, or that could benefit from reducing font-size. I think the problem the is the font-size feels too large when using 400 for the headings in the screenshots below?

|master|400 (this PR)|700|
|---|---|---|
|Screenshot 2026-03-20 at 09 52 42|Screenshot 2026-03-19 at 14 03 15|Screenshot 2026-03-19 at 14 03 48
|Screenshot 2026-03-20 at 09 54 13|Screenshot 2026-03-19 at 14 06 34|Screenshot 2026-03-19 at 14 06 49|

Maybe we could do a mix: use 400 for the NavigationMenu triggers and 700 for headings inside the popup. But I tried and it felt weird.

colmtuite:
> I changed other stuff like the "latest" badge on the releases page to just use 400

It should be 400 yes, but it's still 700?

> I think the problem the is the font-size feels too large when using 400 for the headings

It's just the larger x-height in 400 weight. All will be fixed when we redesign the demos. Fine for now.

aarongarciah:
> It should be 400 yes, but it's still 700?

Sorry, it was uncommitted. Fixed.

colmtuite (approved):

PR #4385, opened 2026-03-19
[docs-infra] SEO Optimizations

brijeshb42:
- Add sitemap generation from the existing docs infra sitemap for search. cc: @dav-is Unless you are already working on some primitive for sitemaps in docs-infra, this would be a good addition there. Unrelated, but moved `robots.txt` to `apps` directory instead of `public` so that it's not treated as a static asset. Could have been `robots.ts` instead with structured data if not for the algolia comment. We can move the Algolia validation to the home page `<meta>` tag.
  - Sitemap generation filters private pages and non-MDX entries (e.g. external links), and derives URL segments from `page.path` rather than `page.slug` to match actual directory names.
- Rename `app/sitemap/index.ts` → `app/computedSiteMap/index.ts` to avoid a naming conflict with Next.js's reserved `sitemap` route convention. Update webpack and Turbopack loader rules in `next.config.mjs` and all import sites (`Header.tsx`, `Search.tsx`, `(docs)/layout.tsx`) accordingly.
- Add `BASE_URL` env var to `next.config.mjs` and use it for `metadataBase` in both layout files instead of a hardcoded string.

Other SEO improvements to the homepage and docs:

- Add `FAQPage` microdata to the homepage accordion using HTML
  `itemscope`/`itemprop` attributes so Google can surface Q&As as rich
  results.

- Add `Organization` JSON-LD schema to the homepage with `logo` and
  `sameAs` links to GitHub, X, npm, Discord, and Bluesky

- Add `twitter:site` (`@base_ui`) to both the website and docs layouts

- Wrap the header logo in an `<a href="/">` with `aria-label="Base UI"`;
  move `aria-hidden` to the SVG so screen readers announce the link
correctly

- Add `rel="noopener noreferrer"` to all external social links in the
  website layout header and footer

- Update homepage `<title>` default from `"Base UI"` to
  `"Base UI · Unstyled UI components for building accessible web apps
and design systems` for better search engine signal. cc: @colmtuite . Some examples -
	- The Foundation for your Design System - shadcn/ui
	- React Router Official Documentation

- Fix `og:title` and `twitter:title` on docs pages always resolving to
  `"Base UI"` instead of the page title: remove the separate
  `openGraph.title` template from the docs layout so `og:title` inherits
  from `metadata.title` (populated per-page by the MDX pipeline)

  https://github.com/mui/mui-public/issues/561 

- Add `/_next/static/` immutable cache rule to `_headers`, replacing
the
  previous `/_next/*.js`-only rule to also cover CSS, fonts, and images

### DIFF
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -148,6 +148,7 @@ const nextConfig = {
     // docs-infra
     LIB_VERSION: rootPackage.version,
     SOURCE_CODE_REPO: 'https://github.com/mui/base-ui',
+    BASE_URL: 'https://base-ui.com',
   },
   ...(process.env.NODE_ENV === 'production' && { distDir: 'export', output: 'export' }),
   devIndicators: false,
--- a/docs/public/_headers
+++ b/docs/public/_headers
@@ -1,4 +1,4 @@
-/_next/*.js
+/_next/static/
   Cache-Control: public, max-age=31536000, immutable
 
 /static/favicon.ico
--- a/docs/public/_redirects
+++ b/docs/public/_redirects
@@ -6,6 +6,9 @@
 # Playground SPA redirects
 /vite-playground/* /vite-playground/index.html 200
 
+# Shortlink to Netlify deploy previews
+/pr/:number https://deploy-preview-:number--base-ui.netlify.app/ 302
+
 # For links that we can't edit later on, for example hosted in the code published on npm
 /r/discord https://discord.com/invite/g6C3hUtuxz 302
 /r/invalid-render-prop /react/handbook/composition 302
--- a/docs/src/app/(website)/page.tsx
+++ b/docs/src/app/(website)/page.tsx
@@ -27,6 +27,28 @@ export default function Homepage() {
           }),
         }}
       />
+      {/* Organization schema for Google Knowledge Panel and entity recognition.
+          https://developers.google.com/search/docs/appearance/structured-data/organization */}
+      <script
+        type="application/ld+json"
+        // eslint-disable-next-line react/no-danger
+        dangerouslySetInnerHTML={{
+          __html: JSON.stringify({
+            '@context': 'https://schema.org',
+            '@type': 'Organization',
+            name: 'Base UI',
+            url: process.env.BASE_URL,
+            logo: `${process.env.BASE_URL}/static/favicon.svg`,
+            sameAs: [
+              process.env.SOURCE_CODE_REPO,
+              'https://x.com/base_ui',
+              'https://www.npmjs.com/package/@base-ui/react',
+              'https://discord.com/invite/g6C3hUtuxz',
+              'https://bsky.app/profile/did:plc:nwr6peuxqzdzlbi72qr5kldc',
+            ],
+          }),
+        }}
+      />
 
       <section className="bui-d-c">
         <h1 className="Text size-3 bp2:size-4 bui-gcs-1 bui-gce-9 bp4:bui-gce-5">
@@ -191,33 +213,59 @@ export default function Homepage() {
           <h2 className="Text size-2">The fine print</h2>
         </div>
         <div className="bui-gcs-1 bui-gce-9 bp2:bui-gcs-3 bp4:bui-gce-7">
-          <Accordion.Root className="AccordionRoot">
-            <Accordion.Item className="AccordionItem">
+          <Accordion.Root
+            className="AccordionRoot"
+            itemScope
+            itemType="https://schema.org/FAQPage"
+            // Setting `keepMounted` so that the content of all panels is available in the DOM for search engines. This is especially important for the homepage, which contains important SEO content.
+            keepMounted
+          >
+            <Accordion.Item
+              className="AccordionItem"
+              itemScope
+              itemProp="mainEntity"
+              itemType="https://schema.org/Question"
+            >
               <Accordion.Header className="AccordionHeader">
-                <Accordion.Trigger className="AccordionTrigger Text size-2">
+                <Accordion.Trigger className="AccordionTrigger Text size-2" itemProp="name">
                   What is Base UI?
                   <PlusIcon className="AccordionIcon AccordionIconPlus" />
                   <MinusIcon className="AccordionIcon AccordionIconMinus" />
                 </Accordion.Trigger>
               </Accordion.Header>
-              <Accordion.Panel className="AccordionPanel">
-                <p className="Text size-2">
+              <Accordion.Panel
+                className="AccordionPanel"
+                itemScope
+                itemProp="acceptedAnswer"
+                itemType="https://schema.org/Answer"
+              >
+                <p className="Text size-2" itemProp="text">
                   Base UI is a library of unstyled UI components for building accessible component
                   libraries, user interfaces, web applications, and websites with React. Base UI
                   components are highly configurable, composable, and customizable.
                 </p>
               </Accordion.Panel>
             </Accordion.Item>
-            <Accordion.Item className="AccordionItem">
+            <Accordion.Item
+              className="AccordionItem"
+              itemScope
+              itemProp="mainEntity"
+              itemType="https://schema.org/Question"
+            >
               <Accordion.Header className="AccordionHeader">
-                <Accordion.Trigger className="AccordionTrigger Text size-2">
+                <Accordion.Trigger className="AccordionTrigger Text size-2" itemProp="name">
                   Does Base UI work with any styling library?
                   <PlusIcon className="AccordionIcon AccordionIconPlus" />
                   <MinusIcon className="AccordionIcon AccordionIconMinus" />
                 </Accordion.Trigger>
               </Accordion.Header>
-              <Accordion.Panel className="AccordionPanel">
-                <p className="Text size-2">
+              <Accordion.Panel
+                className="AccordionPanel"
+                itemScope
+                itemProp="acceptedAnswer"
+                itemType="https://schema.org/Answer"
+              >
+                <p className="Text size-2" itemProp="text">
                   Yes. Base UI works with Tailwind, CSS Modules, CSS-in-JS, plain CSS, and any other
                   styling library you prefer. It also works with JavaScript animation libraries like
                   Motion, or just plain CSS transitions. Base UI is an unstyled component library.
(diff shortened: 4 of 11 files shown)


docs/public/_redirects
```diff
@@ -6,6 +6,9 @@
 # Playground SPA redirects
 /vite-playground/* /vite-playground/index.html 200
 
+# Shortlink to Netlify deploy previews
+/pr/:number https://deploy-preview-:number--base-ui.netlify.app/ 302
```
brijeshb42, line 10: Just a handy shortcut. Already there in other docs.

docs/next.config.mjs
```diff
@@ -71,10 +71,10 @@ const rootPackage = loadPackageJson();
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   trailingSlash: false,
-  pageExtensions: ['mdx', 'tsx'],
+  pageExtensions: ['mdx', 'tsx', 'ts'],
   turbopack: {
     rules: {
-      './src/app/sitemap/index.ts': {
+      './src/app/computedSiteMap/index.ts': {
```
dav-is: What would you think about `./src/app/pages/index.ts` since it compiles a list of the page indexes?

I can follow up with a PR to rename `createSitemap()` to `createPagesList()`, then we could have:

```ts
export const pages = createPagesList([...])
```

Eventually, I want to have a `./src/app/pages/page.tsx` that shows a nice UI that shows all the site's pages.

docs/src/app/sitemap.ts
```diff
@@ -0,0 +1,24 @@
...
+import { sitemap as computedSitemap } from './computedSiteMap/index';
+
+export const dynamic = 'force-static';
+
+export default function sitemap(): MetadataRoute.Sitemap {
+  if (!computedSitemap) {
+    return [];
+  }
+
+  return Object.values(computedSitemap.data).flatMap((section) =>
```
dav-is: Great use for the computed sitemap 👍🏼

docs/src/mdx-components.tsx
```diff
@@ -77,9 +77,13 @@ export const mdxComponents: MDXComponents = {
...
     if (props.name === 'description' && String(props.content).length > 170) {
-      throw new Error('Meta description shouldn’t be longer than 170 chars');
+      throw new Error("Meta description shouldn't be longer than 170 chars");
     }
-    return <meta {...props} />;
+    // At build time, `transformMarkdownMetadata` extracts <Meta> attributes
+    // and injects them as `export const metadata = { ... }` into the compiled
+    // MDX. Next.js picks that export up and emits the <meta> tag itself, so
+    // rendering one here would produce a duplicate.
+    return null;
```
dav-is, line 85: Great, this is the fix I had in mind 👍🏼

colmtuite:
@brijeshb42 

> Wrap the header logo in an <a href="/">

Why do this?

> Update homepage <title>

I think "Unstyled UI components for accessible design systems · Base UI" is better?

brijeshb42:
> Why do this?

To link the logo to the brand so it shows up alongside the search results. I think curently the logo is being taken from favicon which also works but its good to associate the logo on the page as well to the brand -

* https://www.logodesignteam.com/blog/logo-seo-how-to-optimize-logos-for-search/#:~:text=ALT%20attribute,take%20care%20of%20the%20following:
* https://webmasters.stackexchange.com/questions/86722/for-seo-of-the-company-name-should-we-use-a-logo-with-alt-text-or-a-company-nam

docs/src/mdx-components.tsx
```diff
@@ -77,9 +77,13 @@ export const mdxComponents: MDXComponents = {
   QuickNav,
   Meta: (props: React.ComponentProps<'meta'>) => {
```
Janpot, line 77: I'm wondering, we're only using this for

```tsx
<Meta name="description" content="A high-quality, unstyled React input component." />
```

Why aren't we just putting this in the `metadata` export and get rid of this component?

brijeshb42: Yeah. We can do that. I think this is a remnant of older implementation which should have been removed after docs-infra integration.

colmtuite:
@brijeshb42 

> To link the logo to the brand
> I think curently the logo is being taken from favicon

I don't understand either of these statements. Could you elaborate further please? What do you mean by "brand"? The logo has nothing to do with the favicon?

brijeshb42:
Screenshot 2026-03-20 at 1 44 02 PM

The logo shown in google search will use this one instead of favicon. Doesn't matter currently since both are same. But any update in future to the logo in header will also get udpated in google search.

colmtuite:
Gotcha, thanks. I think we should revert that. The favicon almost always makes most sense. Since in the future, we may update the logo to be a "BaseUI" wordmark, which wouldn't work in search results. Whereas the favicon will always work.

What I was wondering about originally was why the logo is now in an `<a>`?

brijeshb42:
> What I was wondering about originally was why the logo is now in an <a>?

General convention across libraries and websites -

1. https://react.dev/
2. https://reactrouter.com/7.12.0/home
3. https://www.radix-ui.com/

brijeshb42:
> Gotcha, thanks. I think we should revert that

Should not be an issue since here, the org metadata is already linked to the favicon.

https://github.com/mui/base-ui/pull/4385/changes#diff-73cfd7da4b348c3f14edbe75988851fa986b91af4a4651205e8f44ab3a742301R41

colmtuite:
> General convention across libraries and websites

Yes but other websites have multiple pages. We just have a single page, so there's never anything to link to. So it's pointless, which is why I didn't do it. Does it have some impact on SEO? If it doesn't impact SEO, it shouldn't be changed.

> the org metadata is already linked to the favicon

Are you saying the logo uses the favicon? I thought you said you changed it so it doesn't use the favicon?

brijeshb42:
Wrapping in link was to associate the logo with the brand, ie, clicking on the logo takes the user to home page which means its part of the brand identity. This will make it appear in image search results as well.
I was wrong in the previous statement about image being shown alongside the result since Google almost always picks the favicon and also trims to square if not already. So this would not replace the logo in previous image I attached. It'll just tell that this is part of the brand unlike all the other images that may be there on the page.

https://webmasters.stackexchange.com/a/86757

> As most optimized variation i would consider the one which contains most SEO-relevant signals. And this is in my opinion this one:

```html
<div itemscope itemtype="http://schema.org/Organization">
  <a itemprop="url" href="http://www.example.com/" title="My Firm">
    <h1 itemprop="name">My Firm</h1>
    <img itemprop="logo" src="http://www.example.com/logo.png" alt="Example" title="My Firm"/>
  </a>
</div>
```

We are already setting the org data through json-ld scripts. I just added the link to logo.

TBH, all these are opinions only which may influence how Google decides what to show. There's no fixed rule.

colmtuite:
> clicking on the logo takes the user to home page which means its part of the brand identity
> This will make it appear in image search results as well.

Are you saying a link is required to make Base UI appear in image search results? I still don't understand. Base UI logo already appears in image search results.

brijeshb42:
the `<a>` links the logo with the brand, since it links to home page, not necessarily to come in the search result. In search result, any image from the site can come based on the alt attribute.

colmtuite:
> the <a> links the logo with the brand

I don't understand what this means. Can you revert the change and put it back how it was before with no `<a>` and no `.Logo` CSS class? The link is pointless, since there is never anything to link to, since the website is just a single page.

docs/src/mdx-components.tsx
dav-is: Agreed, we'll also need a lint rule to prevent descriptions from getting too long (like this mdx-components implementation does)

colmtuite (approved):

docs/src/app/sitemap.ts
```diff
@@ -0,0 +1,24 @@
+import { MetadataRoute } from 'next';
+
+import { sitemap as pagesList } from './pages';
+
+export const dynamic = 'force-static';
+
+export default function sitemap(): MetadataRoute.Sitemap {
```
oliviertassinari: Why are we adding a sitemap?

I'm asking as I'm confused. To what I understand, sitemaps are **only** meant for crawlers to handle large sites (1M+ pages) more effectively. They reduce the number of pages that they need to crawl, a lot of pages for large website like this are just pagination pages, wasted SEO budget. In our case, this is not a problem we have (not enough pages to have to care). So if we don't have this problem, better keep things simple: no sitemaps.

brijeshb42: Any recommendations or guidelines that we should not ? As per [Google](https://developers.google.com/search/docs/crawling-indexing/sitemaps/overview#do-i-need-a-sitemap), in the current stage, we technically don't since the site is still small (less than 500 pages). But its a good to have anyways. I've added a reference to it in `robots.txt` so crawlers get to know about and crawl all the public pages.

Janpot: I'm also in favor of omitting a sitemap unless it brings information a crawler would not be able to crawl efficiently. If crawlers do a correct job on their own, I would avoid adding points where there could be conflicting information. Now or inadvertedly added in the future. A sitemap is not something we should have to maintain if there's no good reason to it.

brijeshb42: Removing it since we only have around 50 pages currently.

docs/src/mdx-components.tsx
brijeshb42: Keeping it for now since the length check is required. Should be handled in a dedicated PR.

docs/src/app/(website)/page.tsx
```diff
@@ -191,50 +213,86 @@ export default function Homepage() {
           <h2 className="Text size-2">The fine print</h2>
         </div>
         <div className="bui-gcs-1 bui-gce-9 bp2:bui-gcs-3 bp4:bui-gce-7">
-          <Accordion.Root className="AccordionRoot">
-            <Accordion.Item className="AccordionItem">
+          <Accordion.Root
+            className="AccordionRoot"
+            itemScope
+            itemType="https://schema.org/FAQPage"
+            // Setting `keepMounted` so that the content of all panels is available in the DOM for search engines. This is especially important for the homepage, which contains important SEO content.
```
dav-is, line 220: Great catch

dav-is (approved):
Nice work

docs/public/_headers
```diff
@@ -1,4 +1,4 @@
-/_next/*.js
+/_next/static/
```
oliviertassinari, line 1: @brijeshb42 We can propagate this to all the repositories, no?

brijeshb42: I'll update

brijeshb42: I saw that netlify is settings the desired headers on its on all the sites own without us specifying these headers. So we can just remove it.

oliviertassinari: > I saw that netlify is settings the desired headers on its on all the sites […] So we can just remove it.

@brijeshb42 This doesn't seem to align with reality. For example, take https://mui.com/_next/static/css/61e6e0c2488b7ca2.css that is loaded from https://mui.com/, it has those response headers:
 
SCR-20260606-otep

So on every consequent page visits, the browser needs to make a request with `if-none-match` to then get a 304 (not changed) response from the CDN. This is wasting network capacity, so not helping to make the page load fast.

Could you check again? Thanks

brijeshb42: This is what I am seeing on first load -

image

And then on reload, directly served from browser cache -

image

brijeshb42: I'll add anyways just to be consistent. I do see the must-revalidate cache control once every few reloads.

oliviertassinari: So .js files behave as expected, but .css files don't, hence the need for propagating the change to all the repositories?

```diff
-/_next/*.js
+/_next/static/
  Cache-Control: public, max-age=31536000, immutable
```

oliviertassinari: > I'll add anyways just to be consistent. 

@brijeshb42 Great thanks, I see we fixed a bunch of other problems at the same time, +1 for going the extra mile.

PR #4401, opened 2026-03-20
[menu] Fix transition status when reopening from ending state

mdm317:
### Issue
When `data-starting-style` and `data-ending-style` are different, reopening a popup before the close animation finishes causes the `data-ending-style` animation to be applied during opening.

| Action | transitionStatus |
|---|---|
| Open | `starting -> undefined` |
| Close | `undefined -> ending` |
| Open again | `ending -> undefined` |

### Fix

Handle the reopen case explicitly: when `open === true`, `mounted === true`, and `transitionStatus === 'ending'`, reset `transitionStatus` to `starting`.

These components are addressed as well:
`MenuDemo, PopoverDemo, SelectDemo, TooltipDemo, PreviewCardDemo, and ComboboxDemo`.

### DIFF
--- a/docs/src/app/(private)/experiments/menu/menu-submenus.tsx
+++ b/docs/src/app/(private)/experiments/menu/menu-submenus.tsx
@@ -16,6 +16,9 @@ interface Settings {
   align: Menu.Positioner.Props['align'];
 }
 
+type MenuPopupPropsWithDataAttributes = React.ComponentProps<typeof Menu.Popup> &
+  Partial<Record<`data-${string}`, string | undefined>>;
+
 export default function MenuSubmenus() {
   const { settings } = useExperimentSettings<Settings>();
 
@@ -35,6 +38,7 @@ export default function MenuSubmenus() {
     submenuTriggerDelay?: Menu.SubmenuTrigger.Props['delay'],
     submenuTriggerClassName: string = classes.SubmenuTrigger,
     popupClassName: string = classes.Popup,
+    popupProps: MenuPopupPropsWithDataAttributes = {},
   ) => (
     <Menu.Root modal={settings.modal} disabled={settings.disabled}>
       <Menu.Trigger
@@ -56,6 +60,7 @@ export default function MenuSubmenus() {
           <Menu.Popup
             className={popupClassName}
             style={{ maxHeight: 'var(--available-height)', overflowY: 'scroll' }}
+            {...popupProps}
           >
             {Array.from({ length: 50 }).map((_, submenuIndex) => (
               <Menu.SubmenuRoot key={submenuIndex}>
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review (GPT-5.4)

This PR fixes the reopen-during-exit path in `useTransitionStatus` by resetting `ending` back to `starting`.
The state change itself looks right, but the branch currently validates it only through a docs experiment.

### 1. Bugs / Issues
#### 1. 🟡 Missing automated regression for the `ending -> starting` reopen path (non-blocking)

In `packages/react/src/utils/useTransitionStatus.ts` (lines 28-30), the new `open && mounted && transitionStatus === 'ending'` branch is the behavior change that fixes the reported bug, but the only verification added in this branch is the manual reproduction in `docs/src/app/(private)/experiments/menu/menu-submenus.tsx` (lines 117-147).

This matters because `useTransitionStatus` is shared by multiple popup consumers, and this exact edge case is timing-sensitive. Without an automated spec, a later refactor of the open/close or animation-complete flow can reintroduce the regression without tripping CI.

**Fix:** Add a focused regression test in a popup/root suite (or a dedicated `useTransitionStatus` test) that closes, reopens before the exit animation finishes, and asserts the reopened popup gets `data-starting-style` instead of retaining `data-ending-style`.

| Shared-hook placement | **Yes** | Fixing the transition state in `useTransitionStatus` keeps the behavior centralized for popup consumers. |
| Render-time state repair | **Yes** | The new branch follows the hook's existing render-time state-repair pattern instead of adding a separate effect. |
| Verification strategy | **No** | Nearby animation timing fixes are usually backed by root-level regression tests, while this branch adds only a manual experiment. |

No separate simplification issue stood out here. The runtime change is already minimal and well-scoped; the missing piece is durable regression coverage rather than a smaller implementation.

| Edge Case | Status | Notes |
|-----------|--------|-------|
| Reopen while exit animation is in flight (`enableIdleState=false`) | **Handled** | The new branch resets `ending` to `starting` before the next open render completes. |
| Reopen while exit animation is in flight (`enableIdleState=true`) | **Likely handled** | The existing idle-state effect should still drive `starting -> idle` after reopen, but this remains inferred rather than directly covered. |
| Deferred ending-state consumers | **Low risk** | This diff does not change the deferred-ending path, but it also does not add targeted coverage around it. |

**Strengths:**
- `pnpm test:jsdom MenuRoot --no-watch` passed.
- `pnpm test:jsdom PopoverRoot --no-watch` passed.
- `pnpm test:jsdom SelectRoot --no-watch` passed.

**Gaps:**
- No new automated regression test covers the reported `ending -> starting` reopen sequence.
- The branch does not prove the fix through at least one real popup with asymmetric enter/exit styles, which is the exact scenario that regressed.

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Correctness | 4/5 | The state-transition change is small, centralized, and matches the bug report. |
| Test coverage | 2/5 | The regression remains protected only by a manual experiment. |
| Cross-category sweep | 4/5 | React/state, accessibility, API shape, maintainability, and performance all look low-risk in this diff. |

Overall merge confidence is 3.5/5. I would add the automated reopen-during-exit regression before merging, even though I did not find a second logic defect.

atomiks:
I'm sort of not convinced this is actually correct because it breaks interruptibility 🤔 At least the example shown in the experiment. But maybe it's better with certain types of animations?

mdm317:
Thanks for the feedback!. I agree that this approach compromises interruptibility in a way that’s not ideal. Given that trade-off, I don’t think this fix is the right direction, so I’ll close this PR.

PR #4404, opened 2026-03-20
[docs] Avoid shipping TypeScript in demo exports

atomiks:
## Summary

This removes the `typescript` package from the docs demo export path so demo-heavy docs pages no longer ship the compiler just to collect Tailwind classes for export-time pre-injection.

## Changes

- Replace the `typescript`-based Tailwind class collector in `demoExportOptions` with a smaller HAST-aware parser that keeps the supported demo patterns.
- Read highlighted HAST sources directly in the Tailwind injection path instead of stringifying them at the call site.
- Add regression tests for highlighted HAST identifier and string-literal cases, while preserving the existing coverage for concatenation, template literals, extra files, and escaping.

Example Tailwind export: https://stackblitz.com/edit/bjhnhpsa?file=src%2FApp.tsx


atomiks:
## Codex Review (GPT-5.4)

This PR removes the `typescript` compiler from the docs demo export path and replaces it with a smaller HAST-aware class collector.
The main review question is whether the lighter implementation still covers the real Tailwind demo patterns that feed export-time class pre-injection.

### 1. Bugs / Issues (None)

| Area | Assessment | Notes |
|-----|:----------:|-------|
| Shipped `typescript` dependency | **Addressed** | `docs/src/utils/demoExportOptions.ts` no longer imports `typescript`, so the original client-bundle issue is removed at the source. |
| Highlighted HAST handling | **Addressed** | The Tailwind path now reads `VariantSource` directly and adds a literal-HAST fast path before falling back to source text parsing. |
| Existing export behavior | **Mostly preserved** | The replacement still handles plain string literals, identifier-backed constants, `+` concatenation, identifier-backed template literals, extra files, and HTML escaping. |

| Pattern checked | Status | Evidence |
|-----|:------:|-------|
| Highlighted HAST string literals | **Covered** | Added regression coverage in `docs/src/utils/demoExportOptions.test.ts`. |
| Highlighted HAST identifier-backed expressions | **Covered** | Added regression coverage in `docs/src/utils/demoExportOptions.test.ts`. |
| Identifier-backed Tailwind class constants in real demos | **Consistent** | The current Tailwind demo set uses the supported shapes in files such as `navigation-menu/demos/nested-inline/tailwind/index.tsx`, `drawer/demos/nested/tailwind/index.tsx`, and `menu/demos/detached-triggers-simple/tailwind/index.tsx`. |
| Richer `className` template expressions | **Not currently present in reviewed Tailwind demos** | I spot-checked the Tailwind demo files for class-related templates that would require ternary or computed interpolation and did not find a branch-relevant case. |

**Strengths:**
- The patch keeps the HAST-specific handling local to `demoExportOptions` instead of reintroducing a heavy parser dependency.
- The replacement is narrower than the earlier AST-based behavior and now documents its supported shapes more clearly through tests.

**Gaps:**
- I did not find a further simplification that rises to a review finding without dropping behavior the current Tailwind demos still rely on.

|-----|:------:|-------|
| String-literal `className="..."` in highlighted code | **Handled** | Covered by the new HAST literal test. |
| Identifier-backed `className={itemClassName}` in highlighted code | **Handled** | Covered by the new HAST identifier test. |
| Inline template composition like ``className={`${itemClass} text-red-600`}`` | **Handled** | Present in the current Tailwind demos and still covered by tests. |
| Bundle-size delta measurement | **Not measured in this review** | I verified the dependency removal in code, but I did not run a before/after bundle benchmark in this pass. |

| Check | Result | Notes |
|-----|:------:|-------|
| `pnpm test:jsdom docs/src/utils/demoExportOptions.test.ts --no-watch` | **Passed** | 8 tests passed. |
| `pnpm typescript` | **Passed** | Typecheck completed successfully. |
| Regression coverage for new HAST path | **Improved** | The suite now covers both highlighted HAST literal and identifier-backed cases. |

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Root-cause coverage | 4/5 | The direct `typescript` import is removed and the replacement path is exercised. |
| Regression risk | 4/5 | The reviewed Tailwind demo patterns line up with the supported parser shapes. |
| Validation depth | 4/5 | Targeted tests and typecheck passed; I did not run a bundle benchmark in this review. |

Overall merge confidence is 4/5. I did not find any concrete branch-relevant bugs in the current patch, and the validation that matters most for this fix passed.

michaldudak:
## Codex Review

### Findings

#### 1. ℹ️ [Informational] The replacement parser narrows supported TypeScript wrappers

**Impact**: Tailwind pre-injection will stop resolving class constants if a demo uses wrappers such as `const popupClassName: string = '…'`, `const popupClassName = '…' as const`, or parenthesized expressions. That would reintroduce the first-render styling bug this utility exists to prevent, even though the previous implementation handled those shapes.

**Evidence**: [`collectStringDeclarations()`](https://github.com/mui/base-ui/blob/f98f2fbe73cd6d2cb3d8c4c1d8076e74c3f855e9/docs/src/utils/demoExportOptions.ts#L413-L460) now records a declaration only when the identifier is followed directly by `=` after trivia, so annotated declarations are skipped. [`resolveStringExpression()`](https://github.com/mui/base-ui/blob/f98f2fbe73cd6d2cb3d8c4c1d8076e74c3f855e9/docs/src/utils/demoExportOptions.ts#L463-L535) now resolves only raw string literals, template literals, top-level `+`, and bare identifiers. The implementation this replaces in [#4219](https://github.com/mui/base-ui/pull/4219) explicitly unwrapped `ts.isAsExpression()` and `ts.isParenthesizedExpression()` before resolving the string.

**Recommendation**: Either extend the string parser and tests to cover type annotations plus `as`/parenthesized wrappers, or document that the exporter now intentionally supports only the narrower subset of demo syntax.

### Confidence: 4/5
High confidence on the current branch behavior based on a full PR diff pass, repo-wide checks of the existing Tailwind demo shapes, and history/previous-PR context. I did not find a current in-repo demo that triggers this, so I’m treating it as compatibility debt rather than a merge blocker.

atomiks:
> Impact: Tailwind pre-injection will stop resolving class constants if a demo uses wrappers such as const popupClassName: string = '…', const popupClassName = '…' as const, or parenthesized expressions. That would reintroduce the first-render styling bug this utility exists to prevent, even though the previous implementation handled those shapes.

To avoid bloating the custom parser code we can easily just avoid this

PR #4405, opened 2026-03-20
[triggers] Fix `ref: HTMLElement` type

mj12albert:
Fixes a type error with trigger's that redeclare their ref as HTMLElement, repro:
```tsx
const triggerRef = React.useRef<HTMLElement>(null);
 // Type 'RefObject<HTMLElement | null>' is not assignable to type '((((instance: HTMLButtonElement | null) => void |…
<Popover.Trigger ref={triggerRef} />;
```

### DIFF
--- a/packages/react/src/popover/trigger/PopoverTrigger.spec.tsx
+++ b/packages/react/src/popover/trigger/PopoverTrigger.spec.tsx
@@ -1,0 +1,15 @@
+import * as React from 'react';
+import { Popover as BasePopover } from '@base-ui/react/popover';
+
+function TriggerWithElementRef() {
+  const triggerRef = React.useRef<HTMLElement>(null);
+
+  return <BasePopover.Trigger ref={triggerRef} />;
+}
+
+const ForwardRefWrapper = React.forwardRef<HTMLElement, BasePopover.Trigger.Props>((props, ref) => (
+  <BasePopover.Trigger {...props} ref={ref} />
+));
+
+<TriggerWithElementRef />;
+<ForwardRefWrapper />;
--- a/packages/react/src/popover/trigger/PopoverTrigger.tsx
+++ b/packages/react/src/popover/trigger/PopoverTrigger.tsx
@@ -41,7 +41,7 @@ import { useTriggerDataForwarding } from '../../utils/popups';
  * Documentation: [Base UI Popover](https://base-ui.com/react/components/popover)
  */
 export const PopoverTrigger = React.forwardRef(function PopoverTrigger(
-  componentProps: PopoverTrigger.Props,
+  componentProps: PopoverTriggerPropsInternal,
   forwardedRef: React.ForwardedRef<HTMLElement>,
 ) {
   const {
@@ -236,7 +236,8 @@ export interface PopoverTriggerState {
 }
 
 export type PopoverTriggerProps<Payload = unknown> = NativeButtonProps &
-  BaseUIComponentProps<'button', PopoverTriggerState> & {
+  Omit<BaseUIComponentProps<'button', PopoverTriggerState>, 'ref'> & {
+    ref?: React.Ref<HTMLElement> | undefined;
     /**
      * Whether the component renders a native `<button>` element when replacing it
      * via the `render` prop.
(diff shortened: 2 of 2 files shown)


atomiks:
## Codex Review (GPT-5.4)

This PR fixes a narrow `Popover.Trigger` typing bug by widening the public ref surface to match the component's forwarded ref type.
The change is type-only and adds a regression spec for the reported `HTMLElement` ref usage and forwarded-wrapper repro.

### 1. Bugs / Issues (None)

I do not see a correctness problem in the narrow `Popover.Trigger` repro this PR is targeting. The change aligns the public ref surface with the implementation's `forwardRef<HTMLElement>` signature, and the new type spec covers both the direct `useRef<HTMLElement>` case and the forwarded-wrapper case described in the PR.

| Root-cause explanation | **Yes** | Before this patch, `HTMLButtonElement` refs already worked because `PopoverTriggerProps` inherited `ref` from `BaseUIComponentProps<'button', ...>`, which flows from `React.ComponentPropsWithRef<'button'>`. That made the public prop type button-specific even though the component itself forwards `HTMLElement`. |
| New public typing | **Yes** | Re-declaring `ref?: React.Ref<HTMLElement>` fixes the actual consumer failure mode here: generic `HTMLElement` refs and `forwardRef<HTMLElement, Popover.Trigger.Props>` wrappers. |
| Internal typing | **Mostly yes** | `PopoverTriggerPropsInternal` keeps the button-shaped internal props for the implementation while widening only the public API. That is a reasonable local repair. |
| Cross-PR alignment | **Partial** | Open draft PR [#3956](https://github.com/mui/base-ui/pull/3956), `[all components] Fix nativeButton type inference`, touches the same `PopoverTrigger` typing surface and may end up subsuming this area structurally. However, its coverage is focused on `nativeButton` inference; it does not explicitly add the failing `useRef<HTMLElement>(null)` or `forwardRef<HTMLElement, Popover.Trigger.Props>` repro that this PR fixes. |

#### 1. ℹ️ Popover now has a one-off ref/style override that overlaps with `#3956` (informational)

In `packages/react/src/popover/trigger/PopoverTrigger.tsx`, this patch fixes the bug by omitting `ref` and `style` from `BaseUIComponentProps<'button', ...>` and re-declaring them locally. That works for the immediate `Popover.Trigger` issue, but draft PR [#3956](https://github.com/mui/base-ui/pull/3956) already rewrites the same file around a shared `NativeButtonComponentProps<...>` helper.

This matters because both changes are operating on the same public typing surface: `nativeButton`, `render`, `style`, and `ref`. Landing both independently is likely to create rebase churn, and `#3956` still would not document this exact `HTMLElement` ref regression unless its spec coverage is expanded.

**Recommendation:** If this is intended as a fast unblocker, I would narrow the PR title/body to `Popover.Trigger` explicitly. If the longer-term goal is to normalize trigger typing more broadly, I would either fold this repro into `#3956` as an extra spec or keep this PR narrowly scoped and let `#3956` rebase over it later.

| Edge Case | Status |
|-----------|--------|
| `useRef<HTMLButtonElement>(null)` passed to `<Popover.Trigger ref={...} />` | **Handled before and after** |
| `useRef<HTMLElement>(null)` passed to `<Popover.Trigger ref={...} />` | **Broken before, fixed here** |
| `React.forwardRef<HTMLElement, Popover.Trigger.Props>(...)` wrapper | **Broken before, fixed here** |
| Equivalent `HTMLElement` refs for sibling triggers such as `Menu.Trigger` and `Dialog.Trigger` | **Not addressed by this PR** |

**Strengths:**
- The new `.spec.tsx` is the right regression layer for this change because it exercises the public type surface rather than runtime behavior.
- It covers both the direct `HTMLElement` ref case and the forwarded-ref wrapper case.

**Gaps:**
- The PR title uses plural “triggers”, but the diff is popover-only.
- If the broader trigger family is expected to support generic `HTMLElement` refs consistently, that broader expectation still belongs in shared typing work rather than in this popover-only patch.

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Correctness | 4/5 | The local popover fix matches the reported repro |
| API typing | 4/5 | Public `ref` typing now matches the component signature |
| Maintainability | 3/5 | Good local fix, but it overlaps with the broader shared typing work in `#3956` |
| Test coverage | 4/5 | Good regression coverage for the narrow bug |

Overall merge confidence is 4/5. I would be comfortable with this as a targeted `Popover.Trigger` fix. The main follow-up is deciding whether the team wants this to land as a narrow unblocker or to fold the same repro into the broader typing work in `#3956`.

PR #4406, opened 2026-03-20
[docs] Fix `style` prop docs generation

mj12albert:
In `BaseUIComponentProps`, style isn't omitted like className even though we redeclared it: https://github.com/mui/base-ui/blob/350b6b19a32e4d922f20d4893ecefc302b2b8cd6/packages/react/src/utils/types.ts#L39

It should also be excluded from `elementProps` like className when passing to useRenderElement

Though this doesn't seem to cause any actual type errors but in the docs the description is missing:

Screenshot 2026-03-21 at 5 00 29 AM


atomiks:
## Codex Review (GPT-5.4)

### 1. Bugs / Issues (None)
No concrete correctness bug stood out in the final branch diff.

| Central prop-model change | **Yes** | Moving the `style` docs fix into the shared `BaseUIComponentProps` contract is the right center of gravity for the change |
| Generated reference output | **Yes** | The generated API files now consistently describe `style` instead of leaving it blank |
| Composite follow-up | **Yes** | The `fix composites` commit restores `style` forwarding in the internal composite path and applies it consistently at the touched call sites |

No meaningful simplification change stands out here. The large file count is mostly generated or mechanical churn driven by a small central typing/docs fix.

| Edge Case | Status |
|-----------|--------|
| Function-valued `style` prop in generated docs | **Handled** |
| Composite-backed wrappers after destructuring `style` | **Handled by follow-up** |
| Slider experiment spreading a function-valued `style` prop | **Handled** |

**Strengths:**
- Ran `pnpm test:jsdom useRenderElement --no-watch`; it passed.
- Ran `pnpm test:jsdom RadioGroup --no-watch`; it passed.
- The generated JSON and snapshot deltas line up with the intended docs output change.

**Gaps:**
- One focused regression test in `useRenderElement.test.tsx` would make this stronger: pass `style` as a function and verify the callback is resolved to a style object rather than leaking a raw function into the DOM, since that is the shared contract this patch is formalizing.

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Docs-generation correctness | 4/5 | The central type change and generated output look internally consistent |
| Regression risk | 3/5 | The broad mechanical churn and follow-up composite fix raise the value of one targeted regression test |
| Verification | 4/5 | Targeted utility and component tests passed |

Overall merge confidence is 4/5. This looks safe to merge, and one small composite-style regression test would make the follow-up fix more durable.

flaviendelangle:
0.37% bundle size increase for that kind of fix is so sad :cry: 
Could we imagine a way to not omit those props in every component but instead do it in `useRenderElement` or would that add to much performance overhead?

mj12albert:
> not omit those props in every component but instead do it in `useRenderElement`

@flaviendelangle Tried this with codex and here's the comparison:
> The prototype was validated in a temp worktree with pnpm typescript, pnpm test:jsdom useRenderElement --no-watch, and pnpm size:snapshot.
> The bundle result was mixed:
> - `@base-ui/react` barrel entry got smaller: `468,552 -> 467,295` parsed `(-0.268%)`, `146,960 -> 146,666` gzip `(-0.200%)`.
> - The full local size snapshot across all measured entrypoints got slightly worse overall: `+1,801` parsed `(+0.075%)` and `+440` gzip `(+0.054%)`.
> - Distribution matters more than the summed total: 35 entrypoints got bigger, 4 smaller, 40 were unchanged. Many leaf entries moved by about `+110` to `+130` parsed bytes and `+24` to `+42` gzip bytes.
>
> Interpretation: the alternative helps the big barrel import a bit, but it pushes extra shared logic into many leaf entrypoints.
> Conclusion: If the goal is minimizing leaf entrypoint size and avoiding extra work in a hot shared helper, the current explicit destructuring is the better trade.

flaviendelangle (approved):

docs/reference/generated/accordion-header.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "AccordionHeader",
+  "description": "A heading that labels the corresponding panel.\nRenders an `<h3>` element.",
```
atomiks, line 3: I think these files were intentionally removed by docs-infra upgrades? @mj12albert

PR #4407, opened 2026-03-21
[popups] Fix positioner not repositioning to a different trigger when reopened with `keepMounted`

### LINKED ISSUE #4222: [preview card] keepMounted breaks detached trigger pattern
# Bug report

## Current behavior

When using PreviewCard with the detached trigger pattern (createHandle) and keepMounted on the Portal, reopening the popup on a different trigger causes it to appear at the previous trigger's position. It does not correct its position.

## Expected behavior

When the popup reopens on a different trigger, it should appear at the correct trigger's position.

## Reproducible example

PreviewCard.createHandle<T>() shared across multiple triggers arranged in a 3x2 grid
Single PreviewCard.Root with handle prop
PreviewCard.Portal with keepMounted
Hover trigger A (row 1) — popup opens correctly
Move cursor away — popup closes
Hover trigger B (row 2) — popup appears at trigger A's position

## Base UI version

Latest (@base-ui/react)

## Which OS are you using?

macOS

## Additional context

mdm317:
fixes #4222 

This issue occurs in all components that have a popup and use keepMounted.
1. Go to https://base-ui.com/experiments/preview-card/triggers
2. Enable the keepMounted option.
3. In Uncontrolled, multiple triggers within Root, do: hover => unhover => hover another trigger.
4. The popup position changes unexpectedly.

The same issue also appears in these components: `tooltip`, `popover`, and `menu`.
new link : https://deploy-preview-4407--base-ui.netlify.app/experiments/preview-card/triggers


atomiks:
## Codex Review (GPT-5.4)

This is a focused bug-fix PR in shared popup positioning. The `positionReference` reset matches the stale-reference failure mode for `keepMounted` reopen flows, and the added Chromium regression reproduces the preview-card case directly.

### 1. Bugs / Issues (None)

| Area | Assessment | Notes |
|-----|:----------:|-------|
| Root-cause fit | Strong | Resetting `positionReference` on the closed `keepMounted` path addresses the store state that was still pointing at the previous trigger when the popup reopened. |
| Patch scope | Narrow | The change is a single store reset in the shared positioning hook, which keeps the blast radius small. |
| Shared-hook impact | Appropriate | This bug lives below the component layer, so fixing it in `useAnchorPositioning` is the right level. |

| Area | Status | Notes |
|-----|:------:|-------|
| Floating root sync | Consistent | The reset now matches the other fields already nulled on close, so the store no longer carries stale reference state across reopen cycles. |
| Reopen behavior | Improved | The hook can now resume from a clean reference state instead of inheriting the last trigger’s positioning input. |

No simplification issue stands out here. The patch is already minimal and directly tied to the bug.

| Area | Status | Notes |
|-----|:------:|-------|
| Preview-card hover path | Covered | The new test exercises the reported close-then-reopen path with `keepMounted`. |
| Nearby shared positioning behavior | Low risk | `pnpm test:chromium PopoverPositioner --no-watch` also passed, which is a useful sanity check on another `keepMounted` consumer. |

**Strengths:** The new Chromium regression is well targeted. It verifies that the positioner aligns to trigger 1, remains mounted while hidden, then realigns to trigger 2 after reopen.

**Gaps:** The PR description says the same bug reproduces in tooltip, popover, and menu as well, but this branch only adds the explicit reopen-and-reposition regression for preview-card. Because the fix is in the shared hook, I do not see evidence of a remaining runtime defect from the code itself, but adding at least one click-driven `keepMounted` multi-trigger regression in popover or menu would better protect that broader claim over time.

**Validation run:** `pnpm test:chromium PreviewCardRoot.detached-triggers --no-watch` and `pnpm test:chromium PopoverPositioner --no-watch` both passed.

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Root cause coverage | 4/5 | The fix lines up with the stale-reference failure mode. |
| Regression coverage | 4/5 | The reported preview-card path is explicitly covered; adjacent popup families are still relying on shared-hook confidence rather than their own dedicated regression. |
| Validation evidence | 4/5 | Targeted Chromium validation passed on the new regression and one nearby shared-hook consumer. |

Overall merge confidence is 4/5. I would be comfortable merging this as-is; adding one more cross-family regression would just improve long-term protection for the broader PR claim.

PR #4410, opened 2026-03-23
[drawer] Fix nested swipe cancel state

atomiks:
Nested drawer swipe-cancel could leave the parent drawer stack visuals out of sync, briefly revealing the previous drawer and then keeping it in the wrong state after the swipe was reversed.

## Changes

- Keep the parent drawer's nested-swiping state aligned with propagated nested swipe progress.
- Clear the parent stack styling when a nested swipe is reversed before release.
- Add regression coverage for the nested swipe cancel path.


atomiks:
## Codex Review (GPT-5.4)

This is a bug-fix PR focused on keeping nested drawer stack visuals in sync when a nested swipe is reversed before release. I reviewed the full diff against a freshly fetched `upstream/master` base and did not find any branch-relevant correctness issues in the current patch.

### 1. Bugs / Issues (None)

| Area | Assessment | Notes |
|------|:----------:|-------|
| Nested swipe cancel state | **Handled** | The parent drawer now stays in its nested-swiping visual state until propagated nested swipe progress returns to `0`, which matches the stack styling contract. |
| Dismiss path consistency | **Handled** | The shared `finishNestedSwipe()` helper keeps release and cleanup behavior aligned across the affected paths. |
| Blast radius | **Low** | The change is localized to nested swipe propagation and does not alter the public API. |

| Area | Assessment | Notes |
|------|:----------:|-------|
| State ownership | **Consistent** | The patch keeps progress propagation and parent swiping state in the same lifecycle instead of letting them drift apart. |
| React/lifecycle correctness | **Low risk** | Cleanup covers both the active release path and viewport teardown. |
| Interaction behavior | **Consistent** | The new regression test matches the reported reverse-before-release scenario directly. |

**Strengths:**
- The new helper removes repeated nested-swipe teardown logic instead of duplicating another inline reset path.

**Gaps:**
- No meaningful simplification concerns stood out in the touched code.

| Nested swipe reversed before release | **Covered** | The new test asserts both `data-nested-drawer-swiping` and `--drawer-swipe-progress` reset correctly. |
| Successful dismiss after nested swipe | **Low risk** | Existing dismiss/release logic remains intact and now routes through the same nested-swipe cleanup helper. |
| Viewport unmount during swipe | **Handled** | Teardown resets backdrop swipe state and nested swiping state together. |

|------------|:------:|-------|
| `pnpm test:jsdom DrawerViewport --no-watch` | **Passed** | Confirms the new nested swipe cancel regression and surrounding viewport behavior. |
| `pnpm typescript` | **Passed** | No type-level regressions in the touched code. |

### Merge Confidence Score

| Dimension | Score | Notes |
|-----------|:-----:|-------|
| Correctness | 4/5 | The fix aligns with the reported bug and the state flow is more coherent after the patch. |
| Regression coverage | 4/5 | The exact reverse-before-release path now has focused coverage. |
| Maintainability | 4/5 | The new helper makes the cleanup path easier to reason about without widening scope. |

Overall merge confidence is **4/5**. I did not find any branch-relevant bugs in the current diff.

flaviendelangle:
Clode Code

## Summary

The PR fixes a bug where cancelling a nested drawer swipe (swiping down then back up before releasing) would leave the parent drawer's visual state out of sync — briefly revealing the previous drawer and keeping stale styles.

## What the PR does

1. **Extracts `finishNestedSwipe` helper** — consolidates the repeated pattern of clearing `nestedSwipeActiveRef` + calling `notifyParentSwipingChange(false)` into one function, used across 4 call sites (onSwipingChange, dismiss, cleanup effect, and the new progress-based reset).

2. **Resets nested swiping when progress returns to 0** — in `applySwipeProgress`, when `nestedSwipeProgress <= 0`, it calls `finishNestedSwipe()`. This is the actual fix: if the user reverses the swipe gesture, progress drops back to 0 and the parent's `data-nested-drawer-swiping` attribute gets cleaned up.

3. **Guards `onSwipingChange` reset** — the `!swiping` branch in `onSwipingChange` now only calls `finishNestedSwipe()` when there's no `notifyParentSwipeProgressChange` callback. This avoids double-resetting: when there is a parent listening to progress, the progress-based path in (2) handles it instead.

## Issues / questions

**The guard condition `!notifyParentSwipeProgressChange` feels indirect.** It uses the presence of a callback as a proxy for "has a parent that tracks progress." If there's ever a case where `notifyParentSwipeProgressChange` is set but the progress path doesn't fire (e.g., swiping stops without a final progress update at 0), nested swiping would never be cleared from `onSwipingChange`. Is this guaranteed? The swipe hook should always emit a final progress of 0 when swiping ends, but it's worth confirming there's no code path where `swiping` becomes `false` without a corresponding `applySwipeProgress({ resolvedProgress: 0, ... })` call.

**Minor: `finishNestedSwipe` in the cleanup effect dependency array** — the effect at line ~927 now depends on `finishNestedSwipe` instead of `notifyParentSwipingChange`. Since `finishNestedSwipe` is wrapped in `useStableCallback`, the identity is stable, so this is fine. Just noting it's correct.

## Test coverage

The new test is well-structured — it simulates a touch-down-then-back-up gesture and asserts both `data-nested-drawer-swiping` removal and `--drawer-swipe-progress` reset to `0`.

## Verdict

The fix is clean and the helper extraction reduces duplication. The one thing I'd want confirmed is the invariant that `applySwipeProgress` is always called with `resolvedProgress <= 0` before `onSwipingChange(false)` when a parent progress listener exists — if that holds, the PR looks good to merge.

atomiks:
@flaviendelangle 

> From Codex:
>
> Thanks, I checked the sequencing in `useSwipeDismiss` more closely.
>
> The exact ordering isn’t "final progress 0 before `onSwipingChange(false)`". In `handleEnd`, `setSwiping(false)` happens first, and the final `updateSwipeProgress(0, ...)` comes afterward on the non-dismiss path.
>
> What this drawer fix relies on is that nested swipe state is cleared by one of the terminal paths:
>
> - reverse / cancel / no-close: `useSwipeDismiss` calls `updateSwipeProgress(0, ...)`, which reaches `applySwipeProgress(...)` and calls `finishNestedSwipe()` when the propagated nested progress is `<= 0`
> - dismiss / close: `startSwipeRelease()` calls `finishNestedSwipe()` directly
> - close / unmount cleanup: the existing effects still reset parent progress to `0` and run nested cleanup
>
> So the `!notifyParentSwipeProgressChange` guard is intentional here: when parent progress propagation exists, the non-dismiss cleanup comes from the progress-reset path rather than `onSwipingChange(false)`.

PR #4413, opened 2026-03-23
[navigation menu] Fix nested inline submenu hover handoff

### LINKED ISSUE #4386: [navigation menu] Nested inline submenus close unexpectedly on hover when entering a third-level submenu
# Bug report

## Current behavior

In a nested inline submenu implementation, when the root menu is opened via hover, opening a secondary submenu and then attempting to move the cursor into a third-level submenu causes the entire menu chain to close.

When the root menu is opened via click, third-level submenus behave as expected and remain open.

## Expected behavior

In a nested inline submenu implementation, submenu behavior should be consistent regardless of how the root menu is opened.

Opening and interacting with deeper submenu levels (e.g., third-level submenus) should not cause ancestor menus to close, whether the root menu was triggered via hover or click.

## Reproducible example

[Hover Product -> Developers -> Composition](https://stackblitz.com/edit/4v8hwphc?file=src%2FApp.tsx)

## Base UI version

v1.3.0

Chrome

## Which OS are you using?

Mac OS

## Additional context

I’ve included a minimal demo based on the current documentation example, making a few assumptions while combining elements from both submenu approaches.

atomiks:
Fixes #4386

This keeps inline nested navigation menus open while the pointer hands off into an already-open positioned descendant popup. The fix stays scoped to the shared hover interaction plus the navigation menu adapter that participates in the floating tree.

## Changes

- Pass the tree node id from `NavigationMenuList` into `useHoverFloatingInteraction` for inline nested menus.
- Treat already-open descendant floating elements as valid hover handoff targets before closing the current menu.
- Add hook-level and `NavigationMenu.Root` regressions covering direct and deeper mixed inline/positioned nesting.

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
+++ b/packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
@@ -200,14 +200,15 @@ export function useHoverFloatingInteraction(
 
       const currentNodeId = dataRef.current.floatingContext?.nodeId ?? nodeIdProp;
       const relatedTarget = event.relatedTarget;
-      if (
+      const isMovingIntoDescendantFloating =
         tree &&
-        currentNodeId != null &&
+        currentNodeId &&
         isElement(relatedTarget) &&
         getNodeChildren(tree.nodesRef.current, currentNodeId, false).some((node) =>
           contains(node.context?.elements.floating, relatedTarget),
-        )
-      ) {
+        );
+
+      if (isMovingIntoDescendantFloating) {
         return;
       }
 


atomiks:
## Codex Review (GPT-5.4)

This is still a `bug-fix` PR, now with `tests`, `docs`, and `react-ui` as the relevant secondary tags. I reviewed the full rebased branch against freshly fetched `upstream/master`, did a second pass over the interaction and maintainability paths touched by the final diff, and reran the targeted regression coverage that matches the current patch shape.

I did not find any blocking or non-blocking defects in the current branch. The final patch now lines up cleanly with the real repro shape: it fixes the tree traversal/root-cause issue in shared hover handling, adds focused regression coverage for that path, and keeps the manual experiment scoped to a single existing navigation menu instead of a throwaway sandbox.

| Area | Assessment | Notes |
| --- | ---: | --- |
| Root cause fit | Good | The behavioral fix is now centered on making open descendants discoverable even when a contextless inline node sits between the parent and positioned popup. |
| Blast radius | Low | Runtime changes stay confined to `getNodeChildren`, `useHoverFloatingInteraction`, and the `NavigationMenuList` adapter. |
| Tree participation | Good | Passing `nodeId` from `NavigationMenuList` remains a small, explicit bridge for inline nested menus that do not carry their own `FloatingContext` node id. |

| Area | Assessment | Notes |
| --- | ---: | --- |
| Shared hook usage | Good | The close guard still lives in `useHoverFloatingInteraction` rather than introducing `NavigationMenu`-specific close logic elsewhere. |
| Utility reuse | Good | The patch leans on `getNodeChildren(..., false)` and the existing shadow-DOM-safe `contains` helper instead of adding another descendant walker. |
| Local cleanup | Good | The small follow-up tidy-up to the hover guard keeps the condition readable without materially expanding the code path. |

| --- | ---: | --- |
| Contextless intermediary node with open descendant popup | Covered | The `nodes` and `safePolygon` regressions now explicitly exercise this shape. |
| Real pointer travel across the submenu gap | Covered | The e2e fixture still drives the full `Product -> Developers -> Composition` mouse path with intermediate movement steps. |
| Manual repro surface | Good | The private experiment was simplified so the mixed nested submenu lives inside the existing navigation menu rather than a separate extra sandbox. |

| Area | Assessment | Notes |
| --- | ---: | --- |
| Utility regression | Good | `nodes.test.ts` now covers open descendants behind a contextless intermediary node. |
| Hover polygon regression | Good | `safePolygon.test.ts` covers the same structural shape at the hover-close utility level. |
| Browser repro | Good | `test/e2e/index.test.ts` plus `InlineSubmenuHoverHandoff.tsx` exercise the actual pointer gap handoff path end-to-end. |
| Validation run | Good | Ran `pnpm test:jsdom safePolygon nodes --no-watch` and `pnpm exec cross-env VITEST_ENV=chromium vitest run --project e2e test/e2e/index.test.ts -t "keeps the inline branch open while moving across the submenu gap into the popup"`. |

### 6. Documentation / Experiment Assessment

| Area | Assessment | Notes |
| --- | ---: | --- |
| Repro clarity | Good | The private experiment now uses the existing top-level menu and removes the extra issue-link sandbox, which makes the manual path easier to exercise. |
| Demo behavior | Good | The nested submenu content and transition styles are scoped to the repro path rather than changing the top-level menu structure wholesale. |

| Earlier review theme | Current status | Notes |
| --- | ---: | --- |
| Deeper mixed inline/positioned nesting coverage | Resolved | The final branch covers the deeper structural case through `nodes`/`safePolygon` regressions plus the Playwright-backed e2e fixture. |
| Manual reproduction support | Improved | The private experiment is now part of the existing navigation menu, which better matches how this was debugged in practice. |

### Merge Confidence Score

Overall merge confidence is 4/5. I’m highly confident in the fix and the regression coverage for the reported pointer-handoff path; the remaining residual risk is mainly in adjacent hover/animation behavior that this branch does not try to normalize beyond the targeted navigation-menu repro.

flaviendelangle:
Claude Code

## PR #4413: [navigation menu] Fix nested inline submenu hover handoff

**Author:** @atomiks | **Fixes:** #4386

### Problem

In mixed nesting scenarios (inline submenus containing positioned popups), moving the pointer from an inline parent's floating element into a child's positioned popup would incorrectly close the parent — because `onFloatingMouseLeave` didn't know about descendant floating elements rendered outside the parent's DOM hierarchy.

### Solution

- Added an optional `nodeId` prop to `useHoverFloatingInteraction` so inline floating elements (which lack a `FloatingContext` with a nodeId) can participate in the floating tree.
- In `onFloatingMouseLeave`, before closing, the handler now checks if the `relatedTarget` is inside any open child floating element — if so, it bails out.
- `NavigationMenuList` reads the tree node id from `useNavigationMenuTreeContext()` and passes it through.

### What looks good

- Minimal, well-scoped fix — only touches the exact code paths needed.
- Reuses existing utilities (`contains`, `getNodeChildren`) instead of new abstractions.
- Correct fallback chain: `dataRef.current.floatingContext?.nodeId ?? nodeIdProp` prefers context's nodeId (positioned menus) and falls back to the explicit one (inline menus).
- Shadow-DOM-aware `contains` is the right choice.
- Both hook-level and integration tests added, including a 4-level deep mixed nesting scenario.
- The boolean simplifications in `NavigationMenuList.tsx` (`!= null` checks replacing `Boolean()` casts and ternaries) are a nice cleanup.

### No blocking issues found

The approach of extending the existing hook with an optional `nodeId` override is the right level of abstraction — solves the problem without restructuring the floating tree internals. Tests are thorough. Clean PR.

PR #4415, opened 2026-03-23
[docs] Serif for italics

aarongarciah:
Preview: https://deploy-preview-4415--base-ui.netlify.app/react/utils/csp-provider#inline-style-attributes (search for "inline style attributes in addition to elements")

Italics had no different styles because we don't load italics font faces for Die Grotesk, and previously we didn't for Unica 77, so this has been broken since forever, but we only use italics in a couple of places. This PR adds a new entry to mdx-components so `<em>` tags render with a class name. `<em>` tags now use the serif font.

### DIFF
--- a/docs/src/css/mdx-components.css
+++ b/docs/src/css/mdx-components.css
@@ -90,4 +90,10 @@
     margin-top: 1.25rem;
     margin-bottom: 1.5rem;
   }
+
+  .MdEm {
+    font-family: var(--font-serif);
+    /* Adjust to visually match the sans-serif font */
+    font-size: 1.06em;
+  }
 }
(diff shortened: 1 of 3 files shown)


docs/src/css/mdx-components.css
```diff
@@ -77,10 +77,6 @@
     list-style-type: decimal;
   }
 
-  .MdStrong {
-    font-weight: 700;
-  }
```
aarongarciah, line 82: We rely on Tailwind's Preflight which takes care of this already for strong tags.

docs/src/css/mdx-components.css
```diff
@@ -94,4 +90,10 @@
     margin-top: 1.25rem;
     margin-bottom: 1.5rem;
   }
+
+  .MdEm {
+    font-family: var(--font-serif);
+    /* Adjust to visually match the sans-serif font */
+    font-size: 1.06em;
+  }
```
aarongarciah, line 98: We could add this as a global style instead if we wanted to treat it as a base style.

PR #3855, opened 2026-01-24
[docs][combobox] Add a `@tanstack/react-query` demo

mj12albert:
Preview: https://deploy-preview-3855--base-ui.netlify.app/react/components/combobox#infinite-scroll

Also added an experiment hooked up to a real API ([openlibrary.org](https://openlibrary.org/dev/docs/api/search)): https://deploy-preview-3855--base-ui.netlify.app/experiments/combobox/tanstack-infinite-query

Closes https://github.com/mui/base-ui/issues/3719

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #4418, opened 2026-03-23
[listbox] Create the Listbox component

michaldudak:
## Summary

- New headless `Listbox` component with single, multiple, and explicit-multiple selection modes (`selectionMode` prop)
- Drag-and-drop reordering via Pragmatic DnD, with `canDrag` and `canDrop` constraints
- Multi-item DnD: in multi-select modes, dragging a selected item moves all selected items together
- Keyboard reorder with Alt+Arrow (also multi-item aware)
- `onItemsReorder` event provides `items`, `referenceItem`, and `edge` for consumer-driven reorder logic
- Shared utilities extracted for reuse with Select: `createItemIndicator`, `createLabel`, `useListItemValueRegistration`, `useSerializedListValue`
- Form integration, lazy loading, scroll area support

## Preview

https://deploy-preview-4418--base-ui.netlify.app/react/components/listbox
https://deploy-preview-4418--base-ui.netlify.app/experiments/listbox/fully-featured


michaldudak:
The [broken Package verification check](https://app.circleci.com/pipelines/github/mui/base-ui/23197/workflows/ebd51e85-27bd-46ee-b7b3-984df554a31b/jobs/259468) is caused by https://github.com/atlassian/pragmatic-drag-and-drop/issues/27.

I tried patching the package and adding it to bundledDependencioes to make the build green but this isn't the best solution in the long run, as it may cause package duplication if our users wants to use Pragmatic DnD outside of Base UI. It also fails with pnpm's `nodeLinker: isolated`.

I'm keeping the build step broken to prevent merging it without a proper resolution.

docs/src/app/(docs)/react/components/listbox/page.mdx
```diff
@@ -47,7 +47,7 @@ The `selectionMode` prop controls how items are selected:
 
 - `"single"` (default) — clicking an item selects it and deselects any other.
 - `"multiple"` — every click toggles an item. Hold Shift and click to select a contiguous range.
-- `"explicitMultiple"` — clicking an item replaces the selection. Hold Ctrl/Cmd and click to toggle, or Shift and click to select a range.
+- `"explicitMultiple"` — clicking an item replaces the selection. Hold Ctrl/Cmd and click to toggle, or Shift and click to select a range. Touch and pen interactions toggle without a modifier, matching `"multiple"` behavior.
```
atomiks: `explicit-multiple` as we use kebab-case for strings unless it matches a native API like for Field

atomiks:
`<Listbox.LoadingTrigger>` is interesting — doesn't that mean Combobox should also have it? For things like https://github.com/mui/base-ui/pull/3855

colmtuite:
@michaldudak 

Awesome progress. I'm particularly excited about this one for some reason, not sure why! I'm probably too early, but anyway, here are some early notes:

**Behaviour**
- `ESC` should deselect all selected items?
- Do we want `[` and `]` to move the focused item(s) to the top/bottom of the list?
- Do we need a way to configure the hotkey bindings?
- `selectionMode`
    - I wonder if we could make `single` always support multiple selection via modifiers? So `explicit-multiple` would be removed, and its behaviour would apply to `single`. I feel like the component working like `single` via pointer input but `multiple` via touch input seems strange/confusing/unexpected to me. I think I would expect and prefer to have to add a toggle button to the UI for touch devices, for toggling `selectionMode`.
    - What is `none` for? If there is no purpose, we could remove it?
    - So ultimately, I wonder if we could just have `single` and `multiple`? Where `single` is single selection without modifiers and multiple selection via modifiers, and `multiple` is multiple selection with or without modifiers. And therefore, we could just have the `multiple` prop, matching other comp APIs.
    - Do we want to support "selection follows focus”? I don’t think so?
- `ItemDragHandle` should not set `cursor: grab`. It’s too opinionated.
- Alt up/down to sort prevents window scroll when focus is inside the list. Are we ok with that?

**Docs**
- Demo content should be something that is naturally a list of items that would be on a screen. Like a list of songs or something. Currently, it looks like the contents of a select popup. Discuss internally if you want ideas. 
- We should have a demo for `draggable=“within-group”`
- We should have a demo for dragging between lists.

**Bugs**
- Alt + up/down works when any other key is pressed as well as Alt. It this desired? Or should it work when only Alt is pressed?
- Shift + up/down currently only selects items, it doesn’t toggle selection?
- I think there are some bugs with Shift + Space contiguous selection? Haven’t spent enough time to accurately reproduce them yet. Apologies for the unhelpful note, just figured I'd mention it to prompt some more testing.
- When you keep the pointer hovered over the list, and reorder using keys, when the item moves to the position above the pointer, it loses its ::before styling. It seems to only happen when sorting upwards.
- When you drag an item and release, the item where you dragged from is highlighted, rather than the item you dragged.
- The text in `Listbox.LoadingTrigger` flashes from some initial content when you quickly scroll to the bottom of the list

michaldudak:
**Behaviour**

> - `ESC` should deselect all selected items?
> - Do we want `[` and `]` to move the focused item(s) to the top/bottom of the list?
> - Do we need a way to configure the hotkey bindings?

I wouldn't handle <kbd>[</kbd> and <kbd>]</kbd> automatically, they seem to be more opinionated. <kbd>Esc</kbd> might be a good idea, though. I'll add a demo showing how devs can set up their own hotkeys.

> - `selectionMode`
    - I wonder if we could make `single` always support multiple selection via modifiers? So `explicit-multiple` would be removed, and its behaviour would apply to `single`. I feel like the component working like `single` via pointer input but `multiple` via touch input seems strange/confusing/unexpected to me. I think I would expect and prefer to have to add a toggle button to the UI for touch devices, for toggling `selectionMode`.
    - What is `none` for? If there is no purpose, we could remove it?
    - So ultimately, I wonder if we could just have `single` and `multiple`? Where `single` is single selection without modifiers and multiple selection via modifiers, and `multiple` is multiple selection with or without modifiers. And therefore, we could just have the `multiple` prop, matching other comp APIs.

I'm quite certain people would ask for a true single-selection mode. It's supported in the native multiline `<select>` and in React Aria.
`selectionMode: 'none'` could be useful when you only want to use the listbox to reorder items  (setting priority among options, for example). RA also has it.
As for different behavior for touch devices - yeah, I'm not sure myself what's the best approach here. Perhaps we can have a Checkbox part for toggling a selection, and tapping on the item outside of the checkbox would deselect all the other items? Windows File Explorer has a similar mode:

https://github.com/user-attachments/assets/86b76c90-e5fa-435b-a5b6-f32262f7afcc

> - Do we want to support "selection follows focus”? I don’t think so?

This should be pretty easy to implement in userland. Also, it wouldn't apply to multi-select mode.
    
> - `ItemDragHandle` should not set `cursor: grab`. It’s too opinionated.

Agree. I haven't noticed that Claude made an inline style in the component itself.
It can still be set in demos' styles, right?

> - Alt up/down to sort prevents window scroll when focus is inside the list. Are we ok with that?

I'm not entirely sure here either. It does feel better than RA's <kbd>Enter</kbd> for toggling drag mode IMO. I personally never used Alt up/down to scroll so it's not confusing to me, but it may be slightly surprising for those who do.

> **Docs**
> - Demo content should be something that is naturally a list of items that would be on a screen. Like a list of songs or something. Currently, it looks like the contents of a select popup. Discuss internally if you want ideas. 

:+1:

> - We should have a demo for `draggable=“within-group”`

I'm also exploring an alternative approach for this that would be more flexible: a function prop (`canDrop` or something along these lines) that would be called when you're dragging over an item. It would expect the user's logic to determine whether the current item is a valid drop target. It would be more explicit, but allow features such as being able to reorder between just certain groups.

> - We should have a demo for dragging between lists.

This is not implemented yet. It seems to me that this feature is used very rarely, so it might be better to create a separate component to handle it (like https://mui.com/material-ui/react-transfer-list/). WDYT?

> **Bugs**

Thanks for the feedback, I'll fix them.

> `<Listbox.LoadingTrigger>` is interesting — doesn't that mean Combobox should also have it? For things like https://github.com/mui/base-ui/pull/3855

@atomiks Yeah, if we don't find any blocking issues with it, we can use it in other components as well

PR #4419, opened 2026-03-23
[combobox] Support keyed values with object items

### LINKED ISSUE #3865: [combobox] value gets cleared if shape of value differs from items

atomiks:
Fixes #3865

Stricter alternative to #4097 without a new prop

---

This makes the implicit keyed combobox path predictable when `items` use the `{ value, label }` shape and the selected value is the primitive `item.value`.

It keeps the supported keyed path consistent across display, selection, comparison, and docs, while tightening TypeScript so unsupported object-item shapes fail earlier instead of drifting at runtime.

## Changes

- Add support for keyed primitive values with `{ value, label }` combobox items without introducing a new prop.
- Keep label resolution, selected-index sync, keyboard selection, and equality handling aligned in the supported keyed path.
- Tighten the type surface so unsupported implicit shapes are rejected, and document the item-shaped vs keyed value modes more clearly.
- Add regression coverage for keyed values, custom equality, empty keyed state behavior, and typed wrapper patterns.


atomiks:
Replacing with https://github.com/mui/base-ui/pull/4985

PR #4429, opened 2026-03-24
[docs] Unify syntax highlighting

aarongarciah:
Preview: https://deploy-preview-4429--base-ui.netlify.app/react/handbook/animation

- Removes `rehype-pretty-code` and custom Rehype plugins in favor of docs-infra Rehype plugins. This means we now only rely on PrettyLights CSS vars for syntax highlighting.
- Removes `shiki` dependency.
- Removes some dead code from the Demo component.
- Removes `data-inline` and `data-table-code` attributes from `<code>` tags.

---

**TODO**
- [x] Scrollbars on fenced code blocks are not visible.
- [x] Fix pre-hydration code block styles and side padding in all scenarios e.g. [Forms handbook page](https://deploy-preview-4429--base-ui.netlify.app/react/handbook/forms)
  - [x] Requires `data-lined` fix in docs-infra (fixed in https://github.com/mui/mui-public/pull/1247).
- [x] Remove nested \<pre\> inside code blocks.
- [x] Do we still need `data-inline`? `.MdInlineCode[data-inline]` uses it for inline margin.
- [x] Do we need `data-table-code`?
- [x] Fix alignment or use a proper code block with border? ([example](https://deploy-preview-4429--base-ui.netlify.app/react/utils/direction-provider#usedirection))
    image
- [x] Double check changes in mdx-components.tsx
- [x] Remove trailing semicolons from anatomy code blocks (fixed in https://github.com/mui/mui-public/pull/1247)
- [x] Fix bugs/missing features on line/word highlighting.
  - [x] `@highlight-text` doesn't work when a line is already highlighted: the whole line gets `data-hl="strong"` but the word is not marked e.g. [preview](https://deploy-preview-4429--base-ui.netlify.app/react/components/preview-card#detached-triggers), [source](https://github.com/mui/base-ui/pull/4429/changes#diff-965624ab886d3cc0db5df4f6f66e30ce300a2d48850daf94623beb41a0399aaaR57-R59)
  - [x] Can’t highlight several words at once with `@highlight-text` e.g. [preview](https://deploy-preview-4429--base-ui.netlify.app/react/components/checkbox#rendering-as-a-native-button), [source](https://github.com/mui/base-ui/pull/4429/changes#diff-027032b8263c0b7d5ae30ca4507cc4061ba82d3a49682102a8bb48eaf69ae2c6R51-R53) (fixed in https://github.com/mui/mui-public/pull/1247)
  - [x] When highlighting characters, some patterns won't work (fixed in https://github.com/mui/mui-public/pull/1247)
    - [x] `@highlight-text "handle={demoAlertDialog}"` e.g. [preview](https://deploy-preview-4429--base-ui.netlify.app/react/components/alert-dialog#detached-triggers), [source](https://github.com/mui/base-ui/pull/4429/changes#diff-849430ac48725ea4c513d7c83a964cf8a98555660cc956bcc5be6295e15f4f72R105-R107)
    - [x] `@highlight-text 'nativeButton="false"'` e.g. [preview](https://deploy-preview-4429--base-ui.netlify.app/react/components/button#rendering-as-another-tag), [source](https://github.com/mui/base-ui/pull/4429/changes#diff-63db4df08c7d4cfc3f97c0f37d9b0a49efb4a04287192ad7c72210f56748b37bR39-R40)
    - [x] `@highlight-text ", 40px"` e.g. [preview](https://deploy-preview-4429--base-ui.netlify.app/react/components/scroll-area#gradient-scroll-fade) ("SSR fallback" code block), [source](https://github.com/mui/base-ui/pull/4429/changes#diff-d7f64d770f49a3135ac26dd0dc163e0835a4de6ab3ee106dcbf0cb2f288941f2R82-R84)


atomiks:
| Before | After |
|---|---|
| After | Before |

RHP has advanced character highlighting that I don't find in other code highlighting libs. Does PrettyLights have parity? Notice `render` isn't highlighted in this PR. You can also control which sets of chars get highlighted. Would be disappointing to lose this feature

aarongarciah:
> Notice render isn't highlighted in this PR

@atomiks I noticed. This is still WIP, not ready for review. I need to chat to docs-infra about the missing/broken stuff, but the idea is that there's full feature parity e.g. being able to highlight different words per line.

docs/src/app/(docs)/react/components/alert-dialog/page.mdx
```diff
@@ -38,7 +38,7 @@ import { AlertDialog } from '@base-ui/react/alert-dialog';
 
 In order to open a dialog using a menu, control the dialog state and open it imperatively using the `onClick` handler on the menu item.
 
-```tsx {12-13,17-18,24-25,28-29} title="Connecting a dialog to a menu"
+```tsx title="Connecting a dialog to a menu"
```
aarongarciah, line 41: Fixing the highlighted lines in this demo, and in Dialog and Menu as well since it's been broken since https://github.com/mui/base-ui/commit/5e474fe1095f6cb2ed7cc5c3f0b3c633735c6662#diff-eb4681ff3f8398dcc2e093c4106508dcbdbd314339bb1ced876eab0b7e054633L59

docs/src/app/(private)/experiments/_components/EditPanel.tsx
```diff
@@ -78,7 +68,6 @@ export async function getDependencyFiles(
         {
           name: basename(path),
           content,
-          prettyContent,
```
aarongarciah, line 81: Seems unused?

docs/src/mdx-components.tsx
```diff
@@ -46,23 +47,18 @@ export const mdxComponents: MDXComponents = {
...
+    if ('data-precompute' in props) {
+      return (
+        <CodeBlock.Root className="MdFigure">
+          <CodeBlockPreComputed {...props} />
+        </CodeBlock.Root>
+      );
     }
 
-    return <figure className="MdFigure" {...props} />;
+    return <CodeBlock.Pre {...props} />;
```
aarongarciah, line 61: I need to double check these changes.

docs/src/app/(private)/experiments/_components/EditPanel.tsx
aarongarciah: cc @atomiks

docs/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -107,15 +128,24 @@ export function PreInline(props: React.ComponentProps<'pre'>) {
...
       <ScrollArea.Viewport
         style={{ overflow: undefined }}
-        render={
-          <pre {...props} id={codeId} className={clsx('CodeBlockPreInline', props.className)} />
-        }
-      />
+        className={viewportClassName}
+        render={viewportRender}
+      >
+        {viewportRender ? null : children}
```
aarongarciah: I'm not super happy with this, but probably out of the scope of the PR to nail it. We need more refactors and this one was about using a single syntax highlighting solution. A future one will be targeted around making more sense of all the kind of code blocks used across the docs and make it easier to reason about them.

aarongarciah:
@dav-is any ideas on how to solve this? You can see it here: https://deploy-preview-4429--base-ui.netlify.app/react/utils/direction-provider#usedirection

Screenshot 2026-04-01 at 17 42 58

Ideally we would align "TextDirection" to the left, but code blocks lines include the padding. I was wondering how could I discriminate these to remove the padding from them, or just use a regular code block with border and all. I'm thinking the latter would make more sense, but I'm not sure if it'd break other use cases where we render code blocks inside popovers, accordions, etc.

I'd love to refactor how we deal with code blocks and their containers (when they have a chrome and when they don't), but that's for another day.

dav-is:
> code blocks lines include the padding

I think the padding should be added only when wrapped by a container like `.container pre :global(.line)`. I think today we add padding to all lines

aarongarciah:
@dav-is do you foresee scenarios where code blocks that are not inside an imaginary `.container` could have highlighted lines?

dav-is:
@aarongarciah wouldn't the `TextDirection` be an example of highlighted lines without a container?

dav-is:
`.container` in my example refers to `.CodeBlockRoot`, `pre` refers to `.CodeBlockPreInline`

Instead of

```css
.CodeBlockRoot .CodeBlockPreInline {
    padding: .5rem 0;
}

.CodeBlockPreInline {
    & code .frame[data-lined] .line {
        display: block;
        white-space: pre;
        padding-left: .75rem;
        padding-right: .75rem;
    }
}
```

It should be:

```css
.CodeBlockRoot {
    & .CodeBlockPreInline {
        padding: .5rem 0;
    }

    & code .frame {
        padding-left: .75rem;
        padding-right: .75rem;
    }

    & code .frame[data-lined] {
        padding-left: 0;
        padding-right: 0;
    }

    & code .frame[data-lined] .line {
        padding-left: .75rem;
        padding-right: .75rem;
    }
}

.CodeBlockPreInline {
    & code .frame[data-lined] .line {
        display: block;
        white-space: pre;
    }
}
```

aarongarciah:
@dav-is thanks for the pointer. It looks good now. Still, I'd love to refactor a bit how we handle code blocks in the future so it's easier to follow.

docs/src/components/Demo/DemoCodeBlock.tsx
```diff
@@ -73,9 +47,7 @@ export function DemoCodeBlock({
     return (
       <Root>
         <ScrollArea.Viewport>
-          <div className="DemoSourceBrowser" data-language={fileNameToLanguage(selectedFileName)}>
```
aarongarciah, line 76: Since we now rely on code-infra setting `.language-*` class name on `<code>` tabs seems ok to remove the `data-language` attribute. Related change: https://github.com/mui/base-ui/pull/4429/changes#diff-367b7b05e2d7475aae46a3c453214e8e78d9d84aca466430f5d57d017df5149dR253-R260

cc @dav-is

colmtuite (approved):

atomiks:
It seems weird to me that line highlighting and syntax colors aren't visible until after hydration? It leads to flickers

aarongarciah:
@atomiks afaik is a way of reducing the HTML size. It's massive in pages with lots of demos/code blocks. We could look into making the first demo/code block on each page to be syntax highlighted on the server? cc @dav-is in case you have thoughts around this.

colmtuite:
> We could look into making the first demo/code block on each page to be syntax highlighted on the server?

Sounds good. It's not noticeable imo when navigating with next/link. Very noticeable with a hard refresh, which is disgusting. It looks awful.

aarongarciah:
BTW, the post-hydration syntax highlighting on demos has been like this for a couple of weeks on master. This PR adopts the same code-infra solution for markdown code blocks, mostly.

aarongarciah:
@atomiks I just pushed a [change](https://github.com/mui/base-ui/pull/4429/changes/5d52c5d244001168fac672e16769f5aed73d1df1) that makes every hero demo syntax highlighted on the server.

I'll try to do the same for markdown code blocks.

aarongarciah:
I added a [new rehype plugin](https://github.com/mui/base-ui/pull/4429/changes/99533b0f0e4c1ac151a9acbeb2d0afefa89b1e6f) so the two first code blocks on each page come syntax highlighted from the server. IMO, this plus the change of syntax highlight the hero demo on each page makes things way better.

---

@dav-is I'd expect that when using `enhanceAfter: 'init'` the highlighted lines would come highlighted from the server, just like with syntax highlighting when using `highlightAfter: 'init'`. Is it working as expected? See the first code block in the [Animation page](https://deploy-preview-4429--base-ui.netlify.app/react/handbook/animation) to see what I mean.

https://github.com/user-attachments/assets/e173d793-c07b-4d4c-bc77-dc888186fe64

---

~PD: I'm looking into improving the syntax highlighting of data attributes in CSS code blocks. They looked better before.~

edit: nah, looks like we'd need rehype or some more complex stuff to achieve it, not worth it IMO.

atomiks:
Highlighting the hero & first two code blocks is a neat solution. Though to me the grayscale colors aren't quite as bad as the highlight only popping in after hydration as it's much more obvious and is still present atm. Of course, you're aware of this already:

>  I'd expect that when using enhanceAfter: 'init' the highlighted lines **would come highlighted from the server**, just like with syntax highlighting when using highlightAfter: 'init'.

(We can release the docs again next week with this change if needed.)

aarongarciah:
@atomiks some changes are needed on the docs-infra side to avoid the flash of highlighted lines.

aarongarciah:
Merging as is. We'll be working on some more improvements.

PR #4430, opened 2026-03-24
[drawer] Align default initial focus with other dialogs

chuganzy:
Align `Drawer.Popup` initial focus with other dialogs; previous difference might not have been intentional, so behavior is now consistent.

### DIFF
--- a/packages/react/src/drawer/popup/DrawerPopup.test.tsx
+++ b/packages/react/src/drawer/popup/DrawerPopup.test.tsx
@@ -18,7 +18,7 @@ describe('<Drawer.Popup />', () => {
     },
   }));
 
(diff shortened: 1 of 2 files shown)


atomiks:
It actually was intentional because drawers start translated beneath the screen and `.focus()` causes scroll containers to glitch on open as it tries to focus out of view. This PR notably breaks the the "Snap drawer", "Indent effect", and "Swipe to open" demos.

Using `preventScroll: true` isn't an option because 1) Android Chrome for some reason doesn't support it yet (based on MDN) and 2) the focused element will be out of view on the screen once the animation settles which is an a11y issue. Focusing the popup instead alleviates all of the issues

chuganzy:
Got it, thank you for sharing! Closing.

PR #4436, opened 2026-03-25
[examples] Align dependency versions with the rest of the repo

LukasTy:
Follow-up on https://github.com/mui/base-ui/pull/4400.

>[!TIP]
> Would it make more sense to not ignore `examples/` folder in renovate so that these don't go out of sync?
> cc @mui/infra 

## Summary

- **Bump dependencies** in `tanstack-start-tailwind-css` and `vite-css` examples to match repo-wide versions (TanStack Router/Start, Tailwind CSS, ESLint, TypeScript, lucide-react, etc.)
- **Remove standalone `pnpm-lock.yaml`** from the `tanstack-start-tailwind-css` example — the root lockfile is sufficient
- **Replace `vite-tsconfig-paths` plugin** with Vite's built-in `resolve.tsconfigPaths` option and remove the now-unused dependency
- **Add `notFoundComponent`** to the TanStack Start root route to suppress the missing route warning
- **Add `suppressHydrationWarning`** on the `<html>` element to avoid noisy console warnings caused by browser extensions ([facebook/react#24430](https://github.com/facebook/react/issues/24430))

### DIFF
--- a/examples/tanstack-start-tailwind-css/package.json
+++ b/examples/tanstack-start-tailwind-css/package.json
@@ -19,12 +19,12 @@
     "@tanstack/react-start": "^1.167.6",
     "@tanstack/router-plugin": "^1.167.4",
     "clsx": "^2.1.1",
-    "lucide-react": "^0.577.0",
+    "lucide-react": "^1.7.0",
     "react": "^19",
     "react-dom": "^19"
   },
   "devDependencies": {
-    "@tailwindcss/vite": "^4.2.1",
+    "@tailwindcss/vite": "^4.2.2",
     "@tanstack/devtools-vite": "^0.6.0",
     "@tanstack/eslint-config": "^0.4.0",
     "@types/node": "22.18.13",
@@ -33,7 +33,7 @@
     "@vitejs/plugin-react": "^6.0.1",
     "eslint": "^10.1.0",
     "prettier": "^3.8.1",
-    "tailwindcss": "^4.2.1",
+    "tailwindcss": "^4.2.2",
     "typescript": "^5.9.3",
     "vite": "^8.0.1"
   }
--- a/examples/vite-css/package.json
+++ b/examples/vite-css/package.json
@@ -25,7 +25,7 @@
     "eslint-plugin-react-refresh": "^0.5.2",
     "globals": "^17.4.0",
     "typescript": "^5.9.3",
-    "typescript-eslint": "^8.57.1",
+    "typescript-eslint": "^8.58.0",
     "vite": "^8.0.1"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -385,8 +385,8 @@ importers:
         specifier: ^2.1.1
         version: 2.1.1
       lucide-react:
-        specifier: ^0.577.0
-        version: 0.577.0(react@19.2.4)
+        specifier: ^1.7.0
+        version: 1.7.0(react@19.2.4)
       react:
         specifier: ^19
         version: 19.2.4
@@ -395,7 +395,7 @@ importers:
         version: 19.2.4(react@19.2.4)
     devDependencies:
       '@tailwindcss/vite':
-        specifier: ^4.2.1
+        specifier: ^4.2.2
         version: 4.2.2(vite@8.0.3(@types/node@22.18.13)(esbuild@0.27.2)(jiti@2.6.1)(terser@5.46.1)(tsx@4.21.0)(yaml@2.8.2))
       '@tanstack/devtools-vite':
         specifier: ^0.6.0
@@ -422,7 +422,7 @@ importers:
         specifier: ^3.8.1
         version: 3.8.1
       tailwindcss:
-        specifier: ^4.2.1
+        specifier: ^4.2.2
         version: 4.2.2
       typescript:
         specifier: ^5.9.3
@@ -474,7 +474,7 @@ importers:
         specifier: ^5.9.3
         version: 5.9.3
       typescript-eslint:
-        specifier: ^8.57.1
+        specifier: ^8.58.0
         version: 8.58.0(eslint@10.1.0(jiti@2.6.1))(typescript@5.9.3)
       vite:
         specifier: ^8.0.1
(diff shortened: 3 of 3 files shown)


examples/tanstack-start-tailwind-css/pnpm-lock.yaml
LukasTy, line 1: Doesn't look relevant, actually; it's a disturbance. 🙈 
cc @mj12albert 

Image

examples/tanstack-start-tailwind-css/vite.config.ts
```diff
@@ -3,19 +3,18 @@ import { devtools } from '@tanstack/devtools-vite';
...
     // this is the plugin that enables path aliases
-    viteTsConfigPaths({
-      projects: ['./tsconfig.json'],
-    }),
     tanstackStart(),
     viteReact(),
     tailwindcss(),
   ],
+  resolve: {
+    tsconfigPaths: true,
```
LukasTy, line 10: After `vite` bump it was warning that `vite-tsconfig-paths` is no longer necessary

examples/tanstack-start-tailwind-css/src/routes/__root.tsx
```diff
@@ -27,11 +27,13 @@ export const Route = createRootRoute({
   }),
 
   shellComponent: RootDocument,
+  notFoundComponent: () => <p>Page not found</p>,
```
LukasTy, line 30: Was complaining that this route doesn't exist

examples/tanstack-start-tailwind-css/src/routes/__root.tsx
```diff
@@ -27,11 +27,13 @@ export const Route = createRootRoute({
...
 });
 
 function RootDocument({ children }: { children: React.ReactNode }) {
   return (
-    <html lang="en">
+    // Use suppressHydrationWarning to avoid https://github.com/facebook/react/issues/24430
+    <html lang="en" suppressHydrationWarning>
```
LukasTy, line 36: Very noisy console after updates when opening started application on a browser with extensions (no warning in incognito window).

LukasTy:
## Claude Opus 4.6 review

### Overall

Clean, well-scoped maintenance PR. Most changes are correct and straightforward. A few items worth addressing before merge.

---

### Issues

#### 1. Stale comment in `vite.config.ts` (bug)

`examples/tanstack-start-tailwind-css/vite.config.ts:10-11`:

```ts
// this is the plugin that enables path aliases
tanstackStart(),
```

This comment originally described the removed `viteTsConfigPaths()` plugin. It now incorrectly attributes path alias support to `tanstackStart()`. Path aliases are now handled by `resolve.tsconfigPaths: true` on line 16. The comment should either be removed or moved/rewritten to describe the new `resolve` option.

#### 2. `devDependencies` sort order broken (nit)

`examples/tanstack-start-tailwind-css/package.json:26-39`: `eslint` (line 30) appears before `@types/node` (line 31). Alphabetically, `@types/*` entries should come before `eslint`. The original file had proper sorting — removing `vite-tsconfig-paths` shifted things. This will likely cause a `prettier --check` / `sort-package-json` failure or at minimum inconsistency with the rest of the repo.

Expected order:
```json
"@tanstack/eslint-config": "^0.4.0",
"@types/node": "22.18.13",
"@types/react": "^19",
"@types/react-dom": "^19",
"@vitejs/plugin-react": "^6.0.1",
"eslint": "^10.1.0",
```

---

### Observations (no action needed)

#### 3. `@types/node` pinned without caret — intentional

`"@types/node": "22.18.13"` (no `^`) matches the root `package.json` which also pins this exact version (and enforces it via `resolutions`). Consistent with monorepo policy.

#### 4. Lockfile removal is correct

`examples/` is included in `pnpm-workspace.yaml`, so the standalone `pnpm-lock.yaml` in the tanstack example was redundant — the root lockfile governs all workspace packages.

#### 5. `resolve.tsconfigPaths` is valid

Confirmed present in Vite 8 type definitions. This is the idiomatic replacement for `vite-tsconfig-paths` in Vite 8+. The tsconfig at `examples/tanstack-start-tailwind-css/tsconfig.json` defines `paths` (`@/*` → `./src/*`), so this option is load-bearing.

#### 6. `suppressHydrationWarning` and `notFoundComponent` — good additions

Both are standard TanStack Start best practices. The linked React issue ([facebook/react#24430](https://github.com/facebook/react/issues/24430)) is the canonical reference for browser-extension-induced hydration mismatches.

#### 7. Examples use `^` ranges for shared deps while root pins exact versions

e.g., `"react": "^19"` in examples vs `"react": "19.2.4"` in root. This is fine — examples are meant to be standalone-reproducible, and `^19` is what a user would write. The workspace hoisting ensures they resolve to the monorepo's pinned version anyway.

---

### Verdict

**Approve with minor changes** — fix the stale comment (#1) and optionally the sort order (#2).

Janpot:
> Would it make more sense to not ignore examples/ folder in renovate so that these don't go out of sync?
> cc @mui/infra

Good question. I believe in core the idea was to define those with very wide version https://github.com/mui/material-ui/blob/master/examples/material-ui-express-ssr/package.json#L9.

michaldudak:
@Janpot, do you see any advantages of this approach? If someone were to use these examples to start a new project, they would likely want more specific dependency versions anyway.

LukasTy:
> Good question. I believe in core the idea was to define those with very wide version https://github.com/mui/material-ui/blob/master/examples/material-ui-express-ssr/package.json#L9.

That's a less maintenance approach, but one that I fear could introduce the most unexpected issues without us noticing. 🙈 
For example, we had `latest` as date library dependencies in MUI X only to see sandboxes broken once date-fns released a major version with BCs.

brijeshb42:
Either way (with `latest` or specific version ranges), there's no guarantee that a particular example will continue to work unless we add a CI job that builds all the examples. This doesn't happen in core neither here.

Janpot:
> @Janpot, do you see any advantages of this approach?

I think the idea is that we don't want to spend a lot of time on maintaining these examples and keeping their versions up-to-date. And so this method was introduced to prevent end-users from accidentally outdated and potentially insecure versions. A potential compromise could also be to "pin" them to their major (e.g. `"6"`). cc @oliviertassinari if you have more context.

LukasTy:
I want to merge this to fix the broken examples.
Let's proceed and discuss potential follow-ups.
Are there any blockers?

PR #4437, opened 2026-03-25
[docs] Paper Mono

vladmoroz:
Back when I was working on the initial docs design, we couldn't find a good monospace font option.

We finally made one at Paper as we've experienced the same problem. This is a pre-release version, but hope it's already an improvement :-)

### DIFF
--- a/packages/react/src/composite/root/CompositeRoot.test.tsx
+++ b/packages/react/src/composite/root/CompositeRoot.test.tsx
@@ -100,6 +100,58 @@ describe('Composite', () => {
       expect(item1).toHaveFocus();
     });
 
+    it('keeps native input behavior when the native target differs from the synthetic target', async () => {
+      render(
+        <CompositeRoot orientation="horizontal">
+          <CompositeItem data-testid="1">1</CompositeItem>
+          <div data-testid="host" />
+          <CompositeItem data-testid="2">2</CompositeItem>
+        </CompositeRoot>,
+      );
+
+      const item1 = screen.getByTestId('1');
+      const item2 = screen.getByTestId('2');
+      const host = screen.getByTestId('host');
+      const input = document.createElement('input');
+
+      input.type = 'text';
+      input.value = 'abcd';
+      input.setSelectionRange(2, 2);
+
+      const focusEvent = new FocusEvent('focusin', { bubbles: true });
+      Object.defineProperty(focusEvent, 'composedPath', {
+        configurable: true,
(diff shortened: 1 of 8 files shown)


docs/src/css/fonts/index.css
```diff
@@ -23,81 +23,18 @@
...
 
 @font-face {
-  font-family: 'DejaVu Sans Mono';
+  font-family: 'Paper Mono';
   font-weight: bold;
-  src: local('DejaVuSansMono-Bold');
+  src: url('/fonts/paper-mono.woff2') format('woff2');
+  font-variation-settings: 'wght' 650;
+  /* Reduce the size to balance with the sans-serif fonts */
   size-adjust: calc(100% * 13.5 / 14);
```
aarongarciah, line 40: Do you think we still need this? With this, the vertical alignment seems to be a bit off.

|with size-adjust|without size-adjust|
|---|---|
|Image|Image|
|Image|Image|
|Image|Image|

vladmoroz: I'd absolutely keep it personally

When zooming in like this, yeah, its obvious that the letters are different height, but when looking at actual paragraphs I think monospace fonts stand out as way larger and overpowering. Visually it looks more balanced to me when the monospace is downsized.

aarongarciah: It's the vertical alignment inside paragraphs what puts me off. It doesn't happen with the current mono font or Söhne Mono.

aarongarciah: A bit hacky, but adding `vertical-align: 0.02em` to inline code tags seems to align them properly. @vladmoroz any better ideas on how to solve the alignment?

vladmoroz: Seeing the same with the current font where the x-height is about 0.5px smaller:

image

***

But overall don't think it's a problem, `vertical-align: 0.02em` makes the baseline misaligned which sticks out much more to me.

It's always tough to balance monospace and sans-serif fonts when used together inline, have never seen it work perfectly. The monospace is either much wider than sans-serif, or has to be condensed as hell, becoming unreadable. That's why I prefer to bring down the size of the monospace via `size-adjust` to make the width difference acceptable

vladmoroz:
Gonna merge, feel free to tweak @aarongarciah @colmtuite

vladmoroz:
Ah I don't have the merge button, someone will have to take over 🙂

PR #4438, opened 2026-03-25
[toast] Fix timers not resuming after window regains focus

### LINKED ISSUE #4439: [toast] Timers never resume after window regains focus
## Summary

When the browser window loses focus while toasts are visible, timers are correctly paused via `handleWindowBlur`. However, when the window regains focus, `handleWindowFocus` in `ToastViewport` exits early without calling `store.resumeTimers()` or `store.setIsWindowFocused(true)`, causing toasts to remain visible indefinitely until manually dismissed.

## Steps to reproduce

1. Go to https://base-ui.com/react/components/toast
2. Trigger a toast
3. Before it auto-dismisses, switch to another window/tab
4. Switch back
5. The toast never auto-dismisses

**Expected:** Toast should dismiss after its remaining timeout once the window regains focus.

## Root cause

In `packages/react/src/toast/viewport/ToastViewport.tsx`, the `handleWindowFocus` handler:

```js
function handleWindowFocus(event: FocusEvent) {
  if (event.relatedTarget || getTarget(event) === win) {
    return;
  }
  // ...resumeTimers() and setIsWindowFocused(true) never reached
}
```

When the user returns to the window (alt-tab, clicking the browser), the focus event fires with `event.target === window`, hitting the `getTarget(event) === win` early return. As a result:

1. `isWindowFocused` stays `false`
2. `resumeTimers()` is never called
3. Toasts remain visible forever

The `event.relatedTarget` check alone is sufficient to filter intra-page focus moves without blocking the window-level refocus case.

## Fix

#4438

kyrregjerstad:
Closes #4439

## Summary

- The `handleWindowFocus` handler in `ToastViewport` had a guard (`getTarget(event) === win`) that caused an early return when the window itself was the focus target — which is exactly the case when alt-tabbing back to the browser.
- This prevented `resumeTimers()` and `setIsWindowFocused(true)` from being called, causing toasts to remain visible indefinitely after switching away and back.
- The `event.relatedTarget` check alone is sufficient to filter intra-page focus moves without blocking window-level refocus.

## Steps to reproduce

1. Open the [Toast demo](https://base-ui.com/react/components/toast)
2. Trigger a toast
3. Before it auto-dismisses, switch to another window (alt-tab)
4. Switch back
5. **Before fix:** toast never auto-dismisses
6. **After fix:** toast dismisses after its remaining timeout

## Test plan

- [x] Existing toast tests pass (`pnpm test:jsdom toast --no-watch` — 206 passed)
- [ ] Manual verification: trigger toast → alt-tab away → alt-tab back → toast dismisses


atomiks:
On https://base-ui.com/react/components/toast, when I switch to a new tab (pointer or cmd+tab) then come back, it resumes correctly for me currently

kyrregjerstad:
@atomiks ah sorry I just realized, this is an Arc bug, not a base-ui toast bug! Seems like window blur events are not sent reliably

atomiks:
@kyrregjerstad so this PR doesn't fix it in Arc still?

atomiks:
## Codex Review (GPT-5.4)

This is a focused bug-fix PR for `ToastViewport` window refocus handling, with a targeted regression test added on top. The runtime fix looks correct to me, and after looking at the tradeoff more closely I think the current test coverage is acceptable for this kind of browser-specific focus bug.

### 1. Bugs / Issues (None)

| Area | Assessment | Notes |
|------|:----------:|-------|
| Alt-tab / tab-return root cause | Handled | Removing the `getTarget(event) === win` early return lets the real window-refocus path reach `resumeTimers()` and `setIsWindowFocused(true)`. |
| `contains()` safety | Handled | The new `target === win` short-circuit prevents passing a `Window` into `contains()`, which matches the Safari follow-up fix intent. |
| In-page focus transitions | Low risk | Keeping the `event.relatedTarget` guard means ordinary element-to-element focus movement inside the page should still bail out before resuming timers. |

The new regression is a logic-level test rather than a full browser-tab automation, but I think that is a reasonable stopping point here. A true alt-tab or tab-backgrounding test is awkward to automate reliably, and the current coverage still protects the important handler path that was broken.

I ran `pnpm test:jsdom ToastViewport --no-watch`, `pnpm test:chromium ToastViewport --no-watch`, `pnpm typescript`, and `pnpm eslint`, and they all passed on the reviewed branch.

### Merge Confidence Score

Overall merge confidence is 4/5. I’m highly confident in the runtime fix and comfortable with the current regression coverage for a small browser-specific bug fix like this. This looks good to merge.

atomiks:
Looks like Safari has the issue as well, but Chrome and Firefox don't seem to for me. This PR fixes Safari.

PR #4440, opened 2026-03-25
[toast] Add upsert support to add method

### LINKED ISSUE #4228: [toast] Built-in support for deduplication?
## Summary

Sometimes, user interaction triggers a toast, and that user interaction can be repeated in a loop. For those cases, we might want to ease the toast deduplication.

- https://honza.codes/articles/taming-toast-notifications
  - https://x.com/HonzaTmn/status/2021298779576533313
- https://x.com/letstri/status/2036764313340223981

## Motivation

Unclear, so far, it feels low priority since toast tends to be meant to use very carefuly: #4229.

atomiks:
Closes #4228

https://deploy-preview-4440--base-ui.netlify.app/react/components/toast#deduplicated-toast

This adds built-in upsert behavior to `toastManager.add()` when callers pass an `id`, so repeated adds update the existing toast instead of creating a duplicate. Matched toasts keep their identity, refresh their auto-dismiss timer, and a toast that is already ending is replaced with a fresh one instead of being revived mid-dismissal.

It also adds `updateKey` to toast objects so custom renderers can replay styles when a toast is updated or upserted. The docs now include a deduplicated toast demo that uses `updateKey` to pulse the existing toast without remounting it.

## Changes

- Treat `add({ id })` as an upsert in the toast store and increment `updateKey` on updates and upserts.
- Add regression coverage for manager and hook flows, including timer refresh, re-adding while a toast is closing, and `updateKey` changes on `add()` and `update()`.
- Document the `id` upsert behavior and add a deduplicated toast demo that replays a pulse via `updateKey`.

### DIFF
--- a/packages/react/src/toast/store.ts
+++ b/packages/react/src/toast/store.ts
@@ -18,7 +18,7 @@ type UpdateToastBehavior = {
   markUpdated?: boolean | undefined;
 };
 type RemoveToastBehavior = {
-  callOnRemove?: boolean | undefined;
+  skipOnRemove?: boolean | undefined;
 };
 
 export type State = {
@@ -130,7 +130,7 @@ export class ToastStore extends ReactStore<State, {}, typeof selectors> {
     }
 
     const toast = this.state.toasts[index];
-    if (behavior.callOnRemove ?? true) {
+    if (!behavior.skipOnRemove) {
       toast?.onRemove?.();
     }
 
@@ -148,7 +148,7 @@ export class ToastStore extends ReactStore<State, {}, typeof selectors> {
 
       if (existingToast) {
         if (existingToast.transitionStatus === 'ending') {
-          this.removeToast(toast.id, { callOnRemove: false });
+          this.removeToast(toast.id, { skipOnRemove: true });
         } else {
           const { id: ignoredId, transitionStatus: ignoredTransitionStatus, ...updates } = toast;
           this.updateToastInternal(toast.id, updates, {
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review (GPT-5.4)

I reviewed the full branch against a freshly fetched `upstream/master`. This is a feature PR with `docs`, `tests`, and `public-api` secondary tags; after the recent edge-case follow-ups, I did not find any remaining correctness issues in the upsert or `updateKey` paths.

### 1. Bugs / Issues (None)

I do not see any remaining branch-relevant correctness defects after the follow-up fixes for stale replacement state, `transitionStatus` filtering, limited-toast coverage, and `onRemove` timing during replacement.

### 2. API Design Assessment

**Strengths:**
- Treating `add({ id })` as an upsert fits the existing toast manager surface naturally and avoids introducing another public method.
- `updateKey` is a small additive primitive that gives custom renderers a reliable replay signal without exposing more internal store mechanics.
- Excluding `updateKey` from both add/update options and filtering `transitionStatus` out of the upsert path keeps internal lifecycle fields owned by the store.

### 3. Documentation / Examples Assessment

**Strengths:**
- The docs now explain both the semantic `id` upsert behavior and the `updateKey` replay mechanism in a way that matches the implementation.
- The deduplicated toast demo shows a realistic pattern for replaying a visual effect while keeping the same toast mounted.
- The added comments in the demo make the alternating animation-name trick and the initial `updateKey: 0` behavior easier to reason about.

**Strengths:**
- The regression suite now covers manager and hook flows for timer refresh, replacing an ending toast, ignoring `transitionStatus` during upsert, incrementing `updateKey` on both `add()` and `update()`, preserving limited-state behavior, and preserving `onRemove` timing during replacement.
- I reran `pnpm test:jsdom useToastManager --no-watch`, `pnpm test:jsdom createToastManager --no-watch`, and `pnpm typescript` on the current branch, and they passed.

### Merge Confidence Score

Overall merge confidence is 4/5. I’m highly confident in the core behavior and regression coverage for the shipped feature, and I would merge this version. The remaining risk is mostly around normal feature-level integration nuance rather than any concrete defect I can point to in the current branch.

PR #4442, opened 2026-03-25
[docs] Enable alternate R and slashed zero for Die Grotesk

aarongarciah:
Preview: https://deploy-preview-4442--base-ui.netlify.app/react/components/radio

|before|after|
|---|---|
|Screenshot 2026-03-25 at 14 08 31|Screenshot 2026-03-25 at 14 08 18|


aarongarciah:
We could disable this for demos if we don't like it there.

edit: we finally disabled the OpenType features on demos.

PR #4446, opened 2026-03-25
[preview card] Fix hover reopen race conditions across close transitions

arturbien:
# Preview Card Hover Timing Fixes

## What was wrong

The `PreviewCard` hover behavior had race conditions around close timing and transitions when using detached triggers:

1. **Detatched triggers Trigger-to-trigger handoff right after close**
   - If trigger A closed and the pointer reached trigger B immediately after (or during the close transition), trigger B failed to open.
   - Root cause: the hover open path still used the normal `openDelay`, and that delayed open could be canceled by close-side cleanup.

https://github.com/user-attachments/assets/97c77459-34fe-41b2-b0fa-8e52e0e13def

2. **Same-trigger re-hover during close transition**
   - If trigger A was hovered again while A's popup was transitioning out, the card could fail to reopen promptly.
   - Root cause: re-entering the same active trigger during hover-driven close transition was still going through the delayed-open path.

https://github.com/user-attachments/assets/74589e3d-39f1-48e1-989b-e54219569aa5

## How we fixed it

The hover reference interaction now uses state-driven transition lifecycle instead of timing heuristics:

- Synced popup `transitionStatus` into `FloatingRootStore` via `useSyncedFloatingRootContext`.
- In `useHoverReferenceInteraction`, immediate-open paths are enabled only when:
  - moving between triggers while already open, or
  - close is hover-driven and `transitionStatus === 'ending'` (including same-trigger re-entry).
- Kept normal `openDelay` behavior after transition lifecycle is finished.
- Added wrapper/delegated trigger resolution in `useHoverReferenceInteraction` so same-trigger transition re-entry works even when hover handlers are attached to a wrapper and the registered trigger is a child element.
This keeps fast handoff behavior only in real transition races and prevents false immediate opens later.

1.

https://github.com/user-attachments/assets/613b2302-1137-4d8a-8fca-4a1694b12d7a

2.

https://github.com/user-attachments/assets/dc42763a-d5a9-4446-a42b-80ab8b378172

## Why this is graceful UX

- **Feels responsive when users expect continuity.** If someone moves from trigger A to trigger B (or back to A) during the close animation, content appears immediately instead of feeling "dropped."
- **Still respects intentional delay in normal browsing.** Once the close lifecycle is fully finished, `openDelay` applies again, so users don't get surprise instant popups while just crossing nearby triggers.
- **Matches visual state with interaction state.** The decision to open immediately is tied to the actual transition lifecycle (`transitionStatus === 'ending'`), so behavior stays predictable and aligned with what users see on screen.

## Test coverage added

Regression tests were added in detached trigger integration tests to cover:

- immediate open when entering trigger B during trigger A close transition,
- normal delayed open when entering trigger B after trigger A close transition completes,
- immediate reopen when re-hovering trigger A during A's close transition,
- normal delayed open for later same-trigger hovers after close lifecycle completes.

@atomiks lmk what you think about this approach 🫡


atomiks:
## Codex Review (GPT-5.4)

This is a focused bug-fix PR for hover timing across close transitions, and the core approach looks directionally right: tying the fast path to `transitionStatus === 'ending'` is a better fit than relying on delay heuristics. I ran targeted validation, and the PreviewCard path itself looks solid, but there is one remaining hook-level gap in the generic wrapper/delegated trigger mode.

### 1. Bugs / Issues
#### 1. 🟡 Same-trigger reopen still misses the wrapper/delegated trigger path (non-blocking)

In `packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts`, the new same-trigger immediate-reopen branch assumes that `event.currentTarget` is the actual trigger element. The hook already supports wrapper/delegated mode, though, where the current target is a wrapper and the real trigger is a child element.

From `packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts`:

```ts
const triggerNode = (event.currentTarget as HTMLElement) ?? null;
const currentDomReference = store.select('domReferenceElement');

const isReenteringSameTriggerDuringCloseTransition =
  !isOverInactive &&
  isElement(triggerNode) &&
  isElement(currentDomReference) &&
  contains(currentDomReference, triggerNode) &&
  isHoverCloseTransition;
```

This matters because the PR changes a generic hook, not just `PreviewCard`, and the hook already has wrapper-mode behavior covered in its existing tests. In that setup, `currentDomReference` is the real trigger while `triggerNode` is the wrapper, so `contains(currentDomReference, triggerNode)` is false and the new fast path falls back to `openDelay` instead of reopening immediately.

**Fix:** Reuse the same wrapper-safe target resolution strategy as the inactive-trigger branch, or base the same-trigger check on `triggerElementRef.current` / `getTarget(event)` rather than `event.currentTarget`, then add a regression test for wrapper-mode same-trigger re-entry during `transitionStatus === 'ending'`.

The root cause analysis looks correct. Syncing `transitionStatus` from the popup store into `FloatingRootStore` gives `useHoverReferenceInteraction` enough lifecycle context to distinguish a real close-transition race from a later fresh hover, which is cleaner than hard-coding timing rules.

Aside from the wrapper-mode hole above, the patch shape is appropriately narrow. The extra shared state is justified by the problem being solved, and I did not find a simpler design that would preserve the same behavior without reintroducing timing coupling.

I ran:

```sh
pnpm test:chromium PreviewCardRoot.detached-triggers --no-watch
pnpm test:jsdom useHoverReferenceInteraction --no-watch
```

Both passed. The new Chromium cases cover the reported PreviewCard regressions well: trigger-to-trigger handoff during close, post-close delayed reopen, same-trigger re-hover during close, and later same-trigger delayed reopen.

The remaining gap is that the generic hook-level wrapper/delegated mode has existing coverage for inactive-trigger detection, but not for this new same-trigger close-transition fast path. That is the one place where the branch still looks under-tested.

### Merge Confidence Score

Overall merge confidence is 4/5. The reported PreviewCard fix appears correct and the targeted regression suite is strong, but I would prefer to close the wrapper/delegated hook gap if this generic behavior is meant to apply uniformly across all `useHoverReferenceInteraction` consumers.

arturbien:
> 🟡 Same-trigger reopen still misses the wrapper/delegated trigger path (non-blocking)

fixed^

arturbien:
@atomiks anything missing here or good to go? the bug is stopping us from using detached triggers in our app 🥺

atomiks:
@arturbien thanks for the fix, looks correct to me. I shortened the tests for CI.

atomiks (approved):

arturbien:
@atomiks you're the goat, thank you so much!

PR #4447, opened 2026-03-25
[select] Fix scroll arrows stopping short of the edges

### LINKED ISSUE #4444: [select] `ScrollUpArrow` and `ScrollDownArrow` not scrolling full content
# Bug report

## Current behavior

Hovering `ScrollUpArrow` or `ScrollDownArrow` not scrolling content up to start/end

https://github.com/user-attachments/assets/d2c849ee-edf1-4b0e-aaf1-7ac4839db981

## Expected behavior

Scrollable content to be scrolled to the end

## Reproducible example

This can be reproduced in docs on this example: https://base-ui.com/react/components/select#grouped

Steps:

1. open url
2. scroll to some select with large list (ex: grouped)
3. make browser window small enough to make scroll to work inside select popower
4. hover arrows

## Base UI version

For example v1.3.0

Chrome latest

## Which OS are you using?

Windows 11

atomiks:
Fixes #4444

This fixes the Select scroll arrows stopping short of the true start or end when constrained layouts leave the popup or item bounds at fractional scroll positions.

The issue is easiest to reproduce when the popup overflows on both sides, including with DevTools open or at non-default browser zoom, but the underlying bug is the fractional edge handling rather than zoom itself.

## Changes

- Normalize edge detection through a shared `scrollEdges` helper used by Select and Scroll Area.
- Normalize aligned popup height measurements before resizing so constrained fractional popup sizes do not stall scrolling.
- Keep hover scrolling advancing past fractional item bounds and add jsdom and Chromium regressions for the grouped overflow case.

### DIFF
--- a/packages/react/src/toast/viewport/ToastViewport.test.tsx
+++ b/packages/react/src/toast/viewport/ToastViewport.test.tsx
@@ -1,4 +1,4 @@
-import { expect } from 'vitest';
+import { expect, vi } from 'vitest';
 import { Toast } from '@base-ui/react/toast';
 import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
 import { act, fireEvent, screen } from '@mui/internal-test-utils';
@@ -398,6 +398,74 @@ describe('<Toast.Viewport />', () => {
       await act(async () => button.focus());
 
       clock.tick(5001);
+
+      expect(screen.queryByTestId('root')).toBe(null);
+    });
+
+    it.skipIf(!isJSDOM)('resumes timers when the window regains focus', async () => {
+      const addEventListenerSpy = vi.spyOn(window, 'addEventListener');
+
+      await renderFakeTimers(
+        <Toast.Provider>
+          <Toast.Viewport>
+            <List />
+          </Toast.Viewport>
+          <Button />
+        </Toast.Provider>,
+      );
+
+      const button = screen.getByRole('button', { name: 'add' });
+
+      fireEvent.click(button);
+
+      expect(screen.queryByTestId('root')).not.toBe(null);
+      const addEventListenerCalls = [...addEventListenerSpy.mock.calls].reverse();
+
+      const blurListener = addEventListenerCalls.find((call) => call[0] === 'blur')?.[1] as
+        | EventListener
+        | undefined;
+      const focusListener = addEventListenerCalls.find((call) => call[0] === 'focus')?.[1] as
+        | EventListener
+        | undefined;
+
+      addEventListenerSpy.mockRestore();
+
+      expect(blurListener).toBeDefined();
+      expect(focusListener).toBeDefined();
+
+      if (!blurListener || !focusListener) {
+        throw new Error('Expected window focus and blur listeners to be registered.');
+      }
+
+      const blurEvent = new FocusEvent('blur');
+      Object.defineProperty(blurEvent, 'composedPath', {
+        value: () => [window],
+      });
+
+      const focusEvent = new FocusEvent('focus');
+      Object.defineProperty(focusEvent, 'composedPath', {
+        value: () => [window],
+      });
+
+      clock.tick(1000);
+
+      await act(async () => {
+        blurListener(blurEvent);
+      });
+
+      clock.tick(5000);
+
+      expect(screen.queryByTestId('root')).not.toBe(null);
+
+      await act(async () => {
+        focusListener(focusEvent);
+      });
+
+      clock.tick(3999);
+
+      expect(screen.queryByTestId('root')).not.toBe(null);
+
+      clock.tick(2);
 
       expect(screen.queryByTestId('root')).toBe(null);
     });
--- a/packages/react/src/toast/viewport/ToastViewport.tsx
+++ b/packages/react/src/toast/viewport/ToastViewport.tsx
@@ -90,13 +90,17 @@ export const ToastViewport = React.forwardRef(function ToastViewport(
     }
 
     function handleWindowFocus(event: FocusEvent) {
-      if (event.relatedTarget || getTarget(event) === win) {
+      if (event.relatedTarget) {
         return;
       }
 
       const target = getTarget(event);
       const activeEl = activeElement(ownerDocument(viewport));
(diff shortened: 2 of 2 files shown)


atomiks:
## Codex Review (GPT-5.4)

This is a `bug-fix` PR with `tests`, `react-ui`, and `public-api` secondary tags. I reviewed the full branch against a freshly fetched `upstream/master`, including separate subagent sweeps for runtime behavior and regression-test quality, and I did not find any remaining branch-relevant issues.

### 1. Bugs / Issues (None)

The fix now addresses the two concrete failure modes that were showing up under constrained fractional layouts:

- `SelectPopup` normalizes measured popup height back into CSS-pixel space before using it in the aligned-scroll resizing path, which closes the mismatch between `getBoundingClientRect()` under zoom/scaling and the CSS-pixel constraints the rest of the algorithm uses.
- `SelectScrollArrow` now normalizes scroll offsets at the edges and applies the same tolerance to item-visibility decisions, which prevents the hover loop from stalling when an item is only fractionally outside the visible range.

The shared `scrollEdges` helper also makes the edge math consistent between popup sizing, arrow visibility, and arrow stepping instead of keeping slightly different `1px` rules in each path.

The branch is more internally consistent than the base version:

- `SelectRoot` and `SelectScrollArrow` now agree on what counts as “at the top” and “at the bottom”.
- `SelectPopup` uses the same edge tolerance constant rather than a local one-off epsilon.
- `ScrollAreaViewport` imports the shared scroll-edge helper directly from the real module instead of through a local re-export shim.

That consistency matters here because the original bug was essentially a disagreement between neighboring pieces of scroll logic under fractional measurements.

The recent simplification pass looks appropriate and bounded. In particular:

- Removing `initialHeightRef` / `maxHeightRef` from `SelectPopup` was a net improvement, since the remaining logic only needed the initial height as a local placement-time value.
- The refactor in `SelectScrollArrow` keeps the meaningful up/down asymmetry while trimming repeated direction checks and centralizing target-scroll computation.

I don’t see another obvious simplification that would materially shrink complexity without also making the edge-case behavior harder to follow.

The branch now explicitly handles the cases I would most worry about for this bug class:

- overlapping start/end tolerance ranges when the total scrollable distance is itself fractional
- bottom-aligned popups that keep growing until they hit the constrained viewport height
- hover scrolling when the next or previous item is only fractionally outside the visible bounds
- arrow visibility near fractional `0` / max edges

I did not find a nearby unhandled path in the same family that looks likely to regress this fix.

The added regression coverage now looks meaningful rather than merely plausible:

- the popup regression asserts the normalized resized height, not just that scrolling still happened
- the scroll-arrow regressions use geometry that actually separates pre-fix and post-fix behavior
- the `SelectRoot` visibility regression exercises the overlapping-tolerance case instead of the old coarse `> 1` behavior

The subagent test sweep also came back with no remaining brittleness or “passes for the wrong reason” issues after the follow-up cleanup.

### Merge Confidence Score

Overall merge confidence is 4/5. I’m highly confident the branch now fixes the reported fractional-scroll bug class and that the regression tests are targeting the real invariants, with only the usual residual risk that comes from DOM-measurement-heavy interaction logic. I’d be comfortable merging this as-is.

flaviendelangle (approved):
# Review: PR #4447 — [select] Fix scroll arrows stopping short of the edges

Fixes #4444. Select scroll arrows stop before reaching the true start/end when fractional scroll positions occur (e.g., at non-default browser zoom).

## What's Good

1. **Shared `scrollEdges` utility** — Moving `scrollEdges.ts` from `scroll-area/utils/` to `src/utils/` is correct. Both Select and ScrollArea need this logic, centralizing avoids duplication.

2. **`getMaxScrollOffset` helper** — Clean extraction of the `Math.max(0, scrollHeight - clientHeight)` pattern repeated across files.

3. **Unified top/bottom scroll logic in `SelectPopup`** — The refactored `handleScroll` collapses duplicated `isTopPositioned`/`isBottomPositioned` branches into shared logic with direction-specific tweaks. Meaningful reduction in duplication.

4. **Extracted `getTargetScrollTop` in `SelectScrollArrow`** — Pulling item-scanning logic out of the inline handler into a standalone function improves readability significantly. The old code had deeply nested direction-specific branches.

5. **Normalized height via `normalizeSize`** — Using `getScale()` to normalize `getBoundingClientRect().height` before comparing against layout values addresses the core fractional mismatch bug.

6. **Good test coverage** — Tests for both scroll directions with fractional positions, plus the grouped overflow Chromium regression test. Clear test descriptions.

## Issues & Questions

1. **`normalizeSize` indirection** — `normalizeSize` (SelectPopup.tsx:579-581) is only used in two places: the `currentHeight` computation and inside `normalizeRect`. The function name doesn't convey much beyond what `size / scale[axis]` already says. Consider whether the helper is worth it vs. inlining.

2. **Early return changes guard behavior** — The PR adds an early return when `!isTopPositioned && !isBottomPositioned` in `handleScroll`. Previously the code would fall through and still set height/scroll. This looks correct (aligned popups shouldn't resize), but it's a subtle behavioral change worth calling out.

3. **`SCROLL_EDGE_TOLERANCE_PX` in item boundary checks** — In `getTargetScrollTop`, the tolerance is applied to `visibleTop` and `visibleBottom` (SelectScrollArrow.tsx:633, 651), treating items within 1px of the visible edge as visible. This is the key fix for fractional boundaries where items that were _almost_ visible would cause the scroll to stall. Looks correct.

4. **Removed `initialHeightRef` / `maxHeightRef`** — Replaced with a local `initialHeight` variable. Works because it's computed and used within the same synchronous `onPlaced` callback. Correct simplification.

5. **Drive-by fix: `SelectScrollDownArrow.test.tsx`** — The existing file incorrectly referenced `ScrollUpArrow` in describe/conformance names. Good catch.

6. **Chromium-only zoom test** — The zoom test modifies `document.documentElement.style.zoom`, properly guarded with `skipIf(isJSDOM)` and wrapped in try/finally. Fine.

## Verdict

Clean, well-tested fix for a real fractional-pixel bug. The refactoring is justified — it reduces duplication and consolidates scroll edge logic. No correctness issues spotted.

PR #4449, opened 2026-03-26
[utils] Move `getFilter`, `serializeValue`, and `stringifyAsLabel` to `@base-ui/utils`

flaviendelangle:
I need to access those utils on the Tree component in the new repo.

## Summary
- Moves `serializeValue`, `stringifyAsLabel`, and the `Filter`/`getFilter` utilities from `@base-ui/react` internal utils to `@base-ui/utils` so they can be shared across packages.
- Updates all imports in `@base-ui/react` to use `@base-ui/utils/filter`, `@base-ui/utils/serializeValue`, and `@base-ui/utils/stringifyAsLabel` directly.
- Deletes the now-unused `packages/react/src/utils/serializeValue.ts` file.

## Test plan
- [x] TypeScript type-checking passes
- [x] ESLint passes on all changed files
- [x] Combobox tests pass (649 tests)


flaviendelangle:
# Claude Code review

## Summary

This PR extracts three utilities from `@base-ui/react` internal code into the shared `@base-ui/utils` package so they can be reused by the Tree component in another package.

**Moved utilities:**
- `getFilter` + `Filter` / `GetFilterParameters` types → `packages/utils/src/filter.ts`
- `serializeValue` → `packages/utils/src/serializeValue.ts` (renamed from `packages/react/src/utils/serializeValue.ts`)
- `stringifyAsLabel` → `packages/utils/src/stringifyAsLabel.ts` (extracted from `resolveValueLabel.tsx`)

## Review Verdict: Looks good ✓

The PR is a clean mechanical extraction with no logic changes. All imports are correctly updated and no stale references remain.

### What's done well

1. **Exact code preservation** — `stringifyAsLabel` and the `getFilter`/`Filter`/cache logic are byte-for-byte identical to the originals. `serializeValue` is a pure rename (100% similarity).

2. **Backward compatibility in `useFilter.ts`** — Re-exports `UseFilterOptions` as an alias for `GetFilterParameters` and keeps `useCoreFilter` as a typed wrapper around `getFilter`, so internal consumers don't break.

3. **No stale imports left** — Verified: no file in `packages/react/` still imports `stringifyAsLabel` from `resolveValueLabel`, `serializeValue` from the old local path, or `Filter` from the old `useFilter` location.

4. **Package exports work** — `packages/utils/package.json` uses a `"./*": "./src/*.ts"` wildcard, so `@base-ui/utils/filter`, `@base-ui/utils/serializeValue`, and `@base-ui/utils/stringifyAsLabel` all resolve correctly without any config changes.

5. **Second commit cleans up** — The follow-up commit removes an unnecessary `stringifyAsLabel` re-export from `resolveValueLabel.tsx` while keeping the import (which is still used by `resolveSelectedLabel` at line 77).

### Minor observations (non-blocking)

1. **No dedicated unit tests for the extracted utils** — `stringifyAsLabel`, `serializeValue`, and `getFilter` are now standalone public utilities in `@base-ui/utils` but don't have their own test files. They're exercised indirectly through the Combobox tests. Consider adding targeted unit tests if these will be consumed by multiple packages going forward.

2. **`useCoreFilter` naming** — This is pre-existing, but `useCoreFilter` follows React hook naming convention (`use` prefix) while it's just `getFilter` (a plain function, not a hook). Not introduced by this PR, so not actionable here.

packages/utils/src/filter.ts
romgrk, line 1: What do you think of depending on `@base-ui/react` in other packages instead of placing some of these in `@base-ui/utils`? Reason is, the utils package is meant to be low-level stuff that could be used in any repository (core, X), whereas `@base-ui/react` is Base UI proper. The utils package is only in the `@base-ui` org because we needed to put it somewhere.

flaviendelangle: I know that we don't want to expose stuff from `@base-ui/react` as stable everytime we want to use it in other package within the company. The package is stable and we care a lot about its DX so anything made public there should be made public because we want people to use it, not because of internal needs.

With that being said, we have two alternatives to this:
- expose stuff from `@base-ui/react` as `unstable_`
- create a new `@base-ui/internals` package

I just don't want us to loose time figuring out how to share some code everytime we need a new util.
Because this will happen more and more as we create new packages.

We have a broader ongoing discussing on [Slack](https://mui-org.slack.com/archives/C02P87NQLJC/p1774510223493279) and [Notion](https://www.notion.so/mui-org/Base-UI-modules-copied-in-other-packages-32fcbfe7b66080738f71c16a7aba2c7e#32fcbfe7b6608098b9e9f6482fc33564)

PR #4452, opened 2026-03-26
[select] Uncaught TypeError: Converting circular structure to JSON

### LINKED ISSUE #4451: [select] Uncaught TypeError: Converting circular structure to JSON
# Bug report

## Current behavior

Image

When I want to use array of object as items for select and pass some item to `defaultValue` or to `value`, I get this error. This happens because each object has `icon`, icon is some react component. And it unable to be serialized in this hook (from stack trace): https://github.com/mui/base-ui/blob/1e0f08cce3d8bc7378db747d5f81365030e638d7/packages/utils/src/useControlled.ts#L56-L66

There is an example:

```ts
// usage
<Select.Root items={shippingMethods} defaultValue={shippingMethods[0]}>
```

```ts
// data
import { AiFillAlipaySquare, AiFillBehanceSquare, AiFillFilter } from "react-icons/ai";

const shippingMethods: ShippingMethod[] = [
  {
    value: "standard",
    label: "Standard",
    duration: "Delivers in 4-6 business days",
    price: "$4.99",
    icon: <AiFillAlipaySquare />,
  },
  {
    value: "express",
    label: "Express",
    duration: "Delivers in 2-3 business days",
    price: "$9.99",
    icon: <AiFillBehanceSquare />,
  },
  {
    value: "overnight",
    label: "Overnight",
    duration: "Delivers next business day",
    price: "$19.99",
    icon: <AiFillFilter />,
  },
];
```

## Expected behavior

No errors

## Reproducible example

[select-circular-bug.zip](https://github.com/user-attachments/files/26283721/select-circular-bug.zip)

## Base UI version

`@base-ui/react@1.3.0`

Chrome latest

## Which OS are you using?

Windows 11

Profesor08:
fixes #4451 [select] Uncaught TypeError: Converting circular structure to JSON


packages/utils/package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@babel/runtime": "^7.29.2",
     "@floating-ui/utils": "^0.2.11",
+    "es-toolkit": "^1.45.1",
```
romgrk: I don't remember, would it still work with this in the `devDependencies` (considering the check only runs in `development`)?

Feels bad to declare a dependency if it's only used in dev mode.

packages/utils/src/useControlled.ts
```diff
@@ -55,15 +56,15 @@ export function useControlled<T = unknown>({
 
     React.useEffect(() => {
       // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is for more details.
-      if (!isControlled && JSON.stringify(defaultValue) !== JSON.stringify(defaultProp)) {
+      if (!isControlled && isEqual(defaultValue, defaultProp) === false) {
```
romgrk: Inconsistent style, `!isControlled` vs `isEqual(...) === false`.

I've never really liked how small the `!` looks, but if we have two negative checks they should use the same format.

packages/utils/package.json
Profesor08: It depends on how bundler works, if bundler has dead code elimination and tree shaking then all be fine. If not, then `es-toolkit` is commonly used ESM library, so bundle will grow a little, but this is not a big deal, because in any way, there should be implemented something to deeply compare objects.

Also bundle size bot reports 0 changes

image

atomiks: We can just use [`deepEqual`](https://github.com/floating-ui/floating-ui/blob/master/packages/react-dom/src/utils/deepEqual.ts) as a util pasted locally:

```jsx
const defaultPropRef = React.useRef(defaultProp);

React.useEffect(() => {
  if (!isControlled && !deepEqual(defaultPropRef.current, defaultProp)) {
    console.error(
      `Base UI: A component is changing the default ${state} state of an uncontrolled ${name} after being initialized. ` +
        `To suppress this warning opt to use a controlled ${name}.`,
    );
  }
}, [defaultProp, isControlled, name, state]);
```

Profesor08: as alternative solution it is ok

atomiks (commented):
## Codex Review (GPT-5.4)

This is a bug-fix PR that resolves the reported circular-JSON crash for React-element defaults, and the targeted `useControlled` JSDOM test still passes on the branch. The main remaining issues are that the warning path now re-logs on every rerender after a mismatch appears, and the new helper is being introduced as a public `@base-ui/utils` entrypoint in a shape that does not look ready to support.

### 1. Bugs / Issues
#### 1. 🟡 The new warning effect re-logs on every rerender after a mismatch appears (non-blocking)

In `packages/utils/src/useControlled.ts`, the `defaultValue` warning effect no longer has a dependency array. Once `defaultValue` differs from the captured initial value, every subsequent rerender schedules the same warning again, even if `defaultValue` itself did not change again.

From `packages/utils/src/useControlled.ts`:

```ts
React.useEffect(() => {
  if (!isControlled && !deepEqual(defaultValue, defaultProp)) {
    console.error(
      [
        `Base UI: A component is changing the default ${state} state of an uncontrolled ${name} after being initialized. ` +
          `To suppress this warning opt to use a controlled ${name}.`,
      ].join('\n'),
    );
  }
});
```

This matters because the old implementation only logged when the tracked value changed. The new version turns that into repeated console noise during unrelated rerenders, which is especially annoying in tests and local debugging.

**Fix:** Add the dependency gate back, or route this through `@base-ui/utils/error` so the warning is deduplicated once the mismatch has been detected.

#### 2. ℹ️ If this helper stays public, the entrypoint should be `deepEqual`, not `deepEquals` (informational)

`packages/utils` exports every top-level source file via the wildcard export map, so adding `packages/utils/src/deepEquals.ts` makes `@base-ui/utils/deepEquals` part of the public package surface immediately. If the helper is going to remain at the package root, that public path should match the actual symbol name (`deepEqual`) instead of introducing a pluralized entrypoint for a singular comparator.

From `packages/utils/package.json`:

```json
"exports": {
  "./*": "./src/*.ts"
}
```

This matters because even a narrow internal helper becomes naming debt once consumers can import it.

**Recommendation:** Prefer keeping the comparator private to `useControlled`, but if it stays as a root-level entrypoint, rename the file to `deepEqual.ts` so the public path matches the export.

#### 3. ℹ️ Drop the top-of-file comment if this helper lands (informational)

The new file starts with an internal implementation note that says the comparator only handles “the comparisons we need”. That kind of comment makes sense for a tightly local helper, but it reads oddly on a file that is being published as a top-level package entrypoint.

From `packages/utils/src/deepEquals.ts`:

```ts
// Fork of `fast-deep-equal` that only does the comparisons we need and compares
// functions
```

This matters because it advertises a partial, intentionally narrow implementation right on what is currently a public utility surface.

**Recommendation:** Remove the comment if the file lands, or keep the helper private so the implementation note stays local.

The patch does address the reported crash on initial render when `defaultValue` contains a React element, and the new regression test proves that narrow scenario. I also checked the old `JSON.stringify` behavior directly against the new comparator behavior: `RegExp` and `Map` were already collapsed by the old code because both serialized to `{}`, so that part is not a new regression. `Date` is the notable semantic difference, because the old check distinguished different dates while the new `deepEqual` implementation treats them as equal. Even so, those value types feel secondary here; the bigger practical regression is still the repeated warning spam from the now-unbounded effect.

I ran `pnpm test:jsdom useControlled --no-watch`, and the touched file passes. The new test in `packages/utils/src/useControlled.test.tsx` is still narrow, though: it only proves the initial render no longer throws, so it does not protect the changed warning behavior itself.

The highest-value addition here would be a regression that changes a circular/default object after mount and asserts the warning only fires once.

### Merge Confidence Score

Overall merge confidence is 3/5. The reported crash looks fixed, but I would still address the repeated-warning behavior before merging, and I would avoid baking the current helper name/comment into the public utils surface unless that exposure is intentional.

packages/utils/package.json
Profesor08: After playng a while with this implementation of `deepEqual`, I can say what I definetely don't like it. Code is weird and unreadable, and it not covering a lot of cases. It is ok for floating-ui and covers their cases, but it is not for public use.

In this example, all values are not equal, but function returns true

```ts
const m1 = new Map().set('a', 1).set('b', 2);
const m2 = new Map().set('a', 1).set('b', 2).set('c', 3);

console.log(deepEqual(m1, m2)); // returns true, should be false

console.log(deepEqual(/test1/, /test2/)); // returns true, should be false

console.log(deepEqual(new Date(0), new Date(1))); // returns true, should be false

class A {
  getPrice = () => 100;
}

console.log(deepEqual(new A(), new A())); // returns true, should be false

class B {
  price = 100;
}

class C {
  price = 100;
}

console.log(deepEqual(new B(), new C())); // returns true, should be false

const b = new B();
const c = new C();

console.log(deepEqual(b, c)); // returns true, should be false
```

`isEqual` from `es-toolkit` covers all this cases correctly, also it is used in other parts of this repo.

atomiks: Those are fairly niche for a dev warning though. es-toolkit's is more accurate but seems overkill. So we should inline it into useControlled and not expose it given it's not intended for public use

oliviertassinari: The Data Grid is most likely going to need a deepEqual logic too, looking at how MUI X Data Grid uses this in a bunch of places. So +1 to have it as an until, to cut down on dependencies.

romgrk: If we need a deep equal function in utils, it probably needs to be a generic one like es-tk's one.

Alternatively, I would suggest [this function as a safe alternative to `JSON.stringify`](https://gist.github.com/romgrk/fccab9852df0aeceb05b321a5296ad7f). It handles circular references. It doesn't handle all edge-cases like es-tk's function does, but neither does `JSON.stringify`:

```bash
$ node -p 'JSON.stringify({ test: new Map([[1, 2]]) })'
{"test":{}}
```

packages/utils/src/useControlled.ts
```diff
@@ -55,15 +56,15 @@ export function useControlled<T = unknown>({
 
     React.useEffect(() => {
       // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is for more details.
```
flaviendelangle: I don't think this comment is relevant anymore

packages/utils/package.json
Profesor08: Just checked how react handles `defaultValue` change on `<input />`, and it not gives any warnings, it allow this prop to be changed on uncontrolled inputs. But it warns then I change behaviour from controlled to uncontrolled and back. So now i'm not sure, if this checks are really usefull. It is rare case, then I use uncontrolled inputs, but when I use, I don't excpect any troubles with them or limitations.

packages/utils/src/useControlled.ts
```diff
@@ -74,3 +75,52 @@ export function useControlled<T = unknown>({
 
   return [value as T, setValueIfUncontrolled];
 }
+if (process.env.NODE_ENV !== 'production') {
+  deepEqual = (a: any, b: any) => {
+    if (a === b) {
+      return true;
+    }
```
atomiks: - This `process.env.NODE_ENV !== 'production'` wrapper around `deepEqual` is redundant, because `deepEqual` is only called from an existing dev-only warning path inside `useControlled`
- The new comparator is realm-unsafe: it relies on `instanceof` and `constructor` checks, which fail for values created in another realm, so `Date`/`RegExp`/`Map`/`Set` handling is not robust
- The implementation is over-scoped for the bug being fixed. The original issue was `JSON.stringify` throwing on React elements inside `defaultValue`, but the patch now adds broad custom handling for `Date`, `RegExp`, `Map`, `Set`, and other cases that aren't critical for a simple dev warning.
- The simpler Floating UI / `fast-deep-equal` style version is a better fit as it avoids the realm bug, does less, and is likely cheaper than the current implementation to improve dev mode perf

romgrk: The FUI version adds more line of code, I would prefer [the `JSON.stringify`-based one](https://gist.github.com/romgrk/fccab9852df0aeceb05b321a5296ad7f).

@Profesor08 Either way, the function should be declared with `function` and placed after the main hook, it focuses the file on what it's really about rather than having to read/scroll past helper utils before getting to the main code.

atomiks (approved):
- Used `error()` for deduping, required a fix with the test setup for React Compiler
- Used @romgrk's func, but needed a tweak for React elements

PR #4464, opened 2026-03-27
[navigation menu] Fix reverse shift-tab submenu re-entry

### LINKED ISSUE #4463: [a11y] Navigation Menu: reverse tab (Shift+Tab) skips submenu items / goes to trigger
# Accessibility Bug
# [a11y] Navigation Menu: reverse tab (Shift+Tab) skips submenu items / goes to trigger
## Current behavior

https://github.com/user-attachments/assets/4494cd1f-c062-445c-9e87-c13a964d60ce

When navigating with the keyboard:

I move to a main nav item with Tab.
I open its submenu with Enter.
I Tab through the items inside the open menu until focus leaves the submenu.

Observed: After leaving the submenu, Shift+Tab returns focus to trigger.

Question : Should reverse navigation follow strict linear tab order (previous focusable = last submenu item), or is jumping to the trigger intentional?

## Expected behavior

In [[ARIA Authoring Practices Guide (APG)]](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/)

Shift tab will move to the latest submenu item 

This pattern is also used in Radix UI nav menu

https://github.com/user-attachments/assets/5c58ed5e-7490-4458-8fc9-e8f9140b6097

## Base UI version
 v1.3.0

Chrome/ Firefox

## Which OS are you using?

Mac OS

Voiceover/keyboard

atomiks:
Fixes #4463

This fixes reverse tab re-entry in Navigation Menu submenus so focus returns to the last submenu item instead of falling back to the trigger or disappearing in inline nested submenus.

## Changes

- Update the `afterOutside` focus guard to re-enter the submenu through the inside guard and synchronously clear inline viewport inert state before redirecting focus.
- Add regression coverage for reverse tab re-entry in both portaled and inline nested submenu flows.

### DIFF
--- a/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
+++ b/packages/react/src/navigation-menu/root/NavigationMenuRoot.test.tsx
@@ -1571,6 +1571,36 @@ describe('<NavigationMenu.Root />', () => {
       expect(screen.queryByTestId('popup-1')).toBe(null);
     });
 
+    it('returns to the last submenu item when shift+tabbing after tabbing out of it', async () => {
+      const { user } = await render(
+        <div>
+          <button data-testid="first" />
+          <TestNavigationMenu />
+          <button data-testid="last" />
+        </div>,
+      );
+      const trigger = screen.getByTestId('trigger-1');
+
+      await act(async () => trigger.focus());
+      fireEvent.click(trigger);
+      await flushMicrotasks();
+
+      expect(screen.getByTestId('popup-1')).not.toBe(null);
+      expect(trigger).toHaveFocus();
+
+      await user.tab();
+      expect(screen.getByText('Link 1')).toHaveFocus();
+
+      await user.tab();
+      expect(screen.getByText('Link 2')).toHaveFocus();
+
+      await user.tab();
+      expect(screen.getByTestId('trigger-2')).toHaveFocus();
+
+      await user.tab({ shift: true });
+      expect(screen.getByText('Link 2')).toHaveFocus();
+    });
+
     it('closes the menu when tabbing back out', async () => {
       const { user } = await render(
         <div>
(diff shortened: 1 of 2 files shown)


flaviendelangle (approved):
### `afterOutside` focus guard fix

The root cause is that when focus leaves the viewport, the viewport's `onBlur` handler sets `viewportInert = true`, making the viewport `inert` on inline (non-portaled) menus. When the user Shift+Tabs back, `isTabbable()` returned false and focus fell back to the trigger.

The `flushSync` approach is the right call here — it ensures React commits the `inert` removal to the DOM before the `.focus()` call, which is necessary since we're inside an event handler that would otherwise batch the state update. The scope of the re-render is small (just the viewport's `inert` attribute), so the perf cost is negligible.

### Minor observations

1. **Asymmetry with `beforeOutside` guard**: The `beforeOutside` guard simply calls `beforeInsideRef.current?.focus()` without clearing inert state. I believe this works because Shift+Tabbing backward hits `beforeOutside` *before* leaving the viewport scope, so the viewport isn't inert yet. Worth confirming this reasoning holds in all cases — if the `beforeOutside` case can also encounter an inert viewport, it would need the same `flushSync` treatment.

2. **Fallback behavior change**: The old code gated on `isTabbable(afterInsideRef.current)` before focusing it. The new code uses `afterInsideRef.current || triggerElement`. Since the `flushSync` ensures the viewport is no longer inert, `afterInsideRef.current` should always be focusable when it exists, so dropping the `isTabbable` gate is safe. But if `afterInsideRef.current` exists yet is somehow not focusable for another reason, focus would silently land on a non-tabbable element instead of falling back to the trigger. Seems unlikely but is a subtle behavioral change.

### Verdict

Overall looks good to merge.

atomiks:
## Codex Review (GPT-5.4)

I reviewed the full PR diff against a freshly fetched `upstream/master` base. This is a narrow `bug-fix` change with secondary tags `tests`, `a11y`, and `react-ui`, focused on the `NavigationMenuTrigger` focus-guard change plus the new reverse-tabbing regression coverage.

### 1. Bugs / Issues (None)

I did not find any concrete correctness, accessibility, or regression issues in the current patch.

| Area | Assessment | Notes |
|------|:----------:|-------|
| Portaled reverse-tab re-entry | Handled | Re-entering through `afterInsideRef` instead of gating on `isTabbable(...)` fixes the fallback-to-trigger behavior on the top-level path. |
| Inline reverse-tab re-entry | Handled | Clearing `viewportInert` synchronously before redirecting focus closes the inline nested regression where focus could disappear instead of landing back in the submenu. |
| Blast radius | Low | The runtime change is tightly scoped to the `afterOutsideRef` focus-guard path in `packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx`. |

The added tests meaningfully improve coverage for the boundary that was previously missing: reverse-tabbing after focus has already left the submenu. I also reran targeted validation on the current PR head with `pnpm test:jsdom NavigationMenuRoot --no-watch`, `pnpm test:chromium NavigationMenuRoot --no-watch`, and `pnpm typescript`, and the latest lint/prettier follow-up for the trigger file is also reflected in the branch.

### Merge Confidence Score

Overall merge confidence is 4/5. The patch is small, the root cause looks correctly addressed for both portaled and inline submenu flows, and the targeted browser/jsdom/type validation is strong for this bug class. I would be comfortable merging this as-is.

PR #4467, opened 2026-03-28
[drawer] Fix interrupted swipe dismiss cleanup

### LINKED ISSUE #4471: [drawer] Outside press dismiss stops working after interrupted swipe-open
# Bug report

  ## Current behavior

  When a drawer is opened with `Drawer.SwipeArea`, an interrupted swipe-open interaction can leave outside press dismissal in a
  broken state.

  After the interrupted gesture, clicking the overlay or outside area may no longer close the drawer, even though the outside click
  target is correct and the click reaches the overlay.

  ## Expected behavior

  After a swipe interaction is interrupted or canceled, outside press dismissal should recover normally, and clicking the overlay or
  outside area should close the drawer.

  ## Reproducible example

  Local reproduction used during investigation:
  - `@base-ui/react` drawer with `Drawer.SwipeArea`
  - open by swipe
  - interrupt the gesture before normal release cleanup completes
  - click the overlay

  I can provide a reduced reproduction if needed.

  ## Base UI version

  v1.3.0

  Primarily Safari and Chrome on macOS during investigation.

  ## Which OS are you using?

  macOS

  None

  ## Additional context

  From investigation, this appears to be related to the swipe cancel path not running release-equivalent cleanup.

  What I observed:
  - the overlay click target is correct
  - the failure is not caused by the overlay click missing the backdrop
  - the issue appears after interrupted swipe interactions
  - the drawer can remain open because outside press dismissal does not recover correctly

  The likely root cause is that the cancel path does not fully reset swipe-related state the same way the normal release path does.

lunaxislu:
Fixes #4471

## Summary

Fix interrupted drawer swipe-open interactions so outside press dismissal is restored even when the swipe is canceled before normal release cleanup runs.

## Changes

- add an `onCancel` hook to `useSwipeDismiss`
- reuse release-equivalent cleanup in `Drawer.SwipeArea` for both release and cancel paths
- add regression coverage for interrupted swipe-open dismissal recovery
- add regression coverage for context-menu-interrupted swipe-open dismissal recovery

## Reproduction

1. Start opening a drawer with `Drawer.SwipeArea`
2. Interrupt the gesture before normal release cleanup finishes
3. Click outside the drawer
4. Before this change, outside dismiss could remain disabled
5. After this change, outside press dismiss works again

## Testing

- `mise exec pnpm@10.32.1 -- pnpm test:jsdom DrawerSwipeArea --no-watch`
- `mise exec pnpm@10.32.1 -- pnpm test:jsdom useSwipeDismiss --no-watch`
- `mise exec pnpm@10.32.1 -- pnpm exec eslint packages/react/src/utils/useSwipeDismiss.ts packages/react/src/drawer/swipe-area/DrawerSwipeArea.tsx packages/react/src/drawer/swipe-area/DrawerSwipeArea.test.tsx`

### DIFF
--- a/packages/react/src/drawer/swipe-area/DrawerSwipeArea.tsx
+++ b/packages/react/src/drawer/swipe-area/DrawerSwipeArea.tsx
@@ -286,16 +286,19 @@ export const DrawerSwipeArea = React.forwardRef(function DrawerSwipeArea(
     );
   }
 
-  const finishSwipeInteraction = useStableCallback(() => {
+  function resetSwipeInteractionState() {
     swipeStartEventRef.current = null;
     openedBySwipeRef.current = false;
-    setSwipeActive(false);
     closedOffsetRef.current = null;
+    setSwipeActive(false);
+  }
 
+  function finishSwipeInteraction() {
+    resetSwipeInteractionState();
     enableDismissAfterRelease();
     resetDragDelta();
     clearSwipeStyles();
-  });
+  }
 
   const swipe = useSwipeDismiss({
     enabled,
@@ -369,9 +372,7 @@ export const DrawerSwipeArea = React.forwardRef(function DrawerSwipeArea(
 
       return false;
     },
-    onCancel() {
-      finishSwipeInteraction();
-    },
+    onCancel: finishSwipeInteraction,
   });
 
   const swipePointerProps = swipe.getPointerProps();
@@ -383,10 +384,7 @@ export const DrawerSwipeArea = React.forwardRef(function DrawerSwipeArea(
       resetSwipe();
       resetDragDelta();
       clearSwipeStyles();
-      setSwipeActive(false);
-      openedBySwipeRef.current = false;
-      swipeStartEventRef.current = null;
-      closedOffsetRef.current = null;
+      resetSwipeInteractionState();
     }
   }, [clearSwipeStyles, enabled, resetDragDelta, resetSwipe]);
 


atomiks:
## Codex Review (GPT-5.4)

This remains a focused bug-fix PR. The follow-up commit simplifies the local cleanup helper in `Drawer.SwipeArea` without changing the underlying fix, and the current tip still looks mergeable based on the full branch pass plus targeted reruns on the touched file.

### 1. Bugs / Issues (None)

The issue report pointed at outside-press dismissal staying disabled when a swipe-open gesture exits through the hook’s cancel path instead of the normal release path. The change in `packages/react/src/utils/useSwipeDismiss.ts` adds an explicit `onCancel` callback at the point where button-loss cancellation is detected, and `packages/react/src/drawer/swipe-area/DrawerSwipeArea.tsx` funnels both `onRelease` and `onCancel` through the same cleanup path. That still matches the reported failure mode and keeps the release/cancel behavior aligned instead of duplicating the cleanup sequence in two places.

The latest follow-up commit stays in `packages/react/src/drawer/swipe-area/DrawerSwipeArea.tsx` and only refactors that cleanup into a plain function plus a small reset helper. I do not see evidence that it changes the runtime behavior of the fix.

**Handled:** the current tip drops the unnecessary `useStableCallback` wrapper around `finishSwipeInteraction`, uses `onCancel: finishSwipeInteraction`, and reuses `resetSwipeInteractionState()` in the disabled-effect path. That makes the cleanup flow easier to read without widening the behavior surface.

The regression coverage in `packages/react/src/drawer/swipe-area/DrawerSwipeArea.test.tsx` still exercises the two interruption modes that matter most here: a non-primary-button takeover during the swipe and a context-menu interruption after the drawer has already been opened by the gesture. That remains the right boundary to test because the bug only appears once outside-dismiss has been disabled and the open-state cleanup has to recover from a non-release exit.

### Merge Confidence Score

Overall merge confidence is 4/5. I still did not find a branch-relevant correctness issue in the diff, and after the follow-up cleanup commit `pnpm test:jsdom DrawerSwipeArea --no-watch`, `pnpm test:chromium DrawerSwipeArea --no-watch`, and `pnpm exec eslint packages/react/src/drawer/swipe-area/DrawerSwipeArea.tsx` all passed. I’d still be comfortable merging this as-is.

Issue #4468, opened 2026-03-28
[avatar] Fallback flashes for cached images on SSR hydration


erdinctekay:
# Bug report

## Current behavior

`Avatar.Fallback` produces a visible flash during SSR hydration even when the image source is already in the browser cache.

`useImageLoadingStatus` has no synchronous cached-image fast-path. It always initializes as `'idle'` and the only path to `'loaded'` is the async `onload` callback. Since `useIsoLayoutEffect` is a no-op on the server, the fallback is baked into the server HTML with no Avatar `<img>` at all. After hydration, the layout effect fires but never checks `image.complete`, which is already `true` synchronously for cached images.

https://github.com/mui/base-ui/blob/master/packages/react/src/avatar/image/useImageLoadingStatus.ts#L17

![Image](https://github.com/user-attachments/assets/280a6ea5-df74-4bed-8cc5-1a09591c4a7b)

Left: base-ui Avatar shows red fallback. Right: plain `<img>` with the same cached URL renders instantly.

Measurements with 4x CPU throttle on a minimal page (image shown above):

| Framework | Red fallback frames | Flash duration |
|-----------|:---:|:---:|
| TanStack Start | 14 | ~304ms |
| Next.js | 3 | ~52ms |
| Vite SPA (client-only) | 1 | ~16ms |

The client-side case is a single-frame flash but the SSR case is the real problem. The flash duration scales with hydration time and gets worse on slower devices or larger pages.

## Expected behavior

For cached images, `Avatar.Image` should render immediately without showing `Avatar.Fallback`. A plain `<img>` with the same cached URL renders instantly on first paint.

## Reproducible example

Any `Avatar` with a cached `src` and a visible fallback during SSR hydration. Tested with both Next.js and TanStack Start.

## Base UI version

v1.3.0

## Which browser are you using?

Chrome, Firefox

## Which OS are you using?

Windows, MacOS, Linux

## Additional context

Radix Avatar handles this with an `image.complete` check in the `useState` initializer to skip the fallback on the first render for cached images: [radix-ui/primitives](https://github.com/radix-ui/primitives/blob/22473d16404bfd446305db5b6c9308aece99fdec/packages/react/avatar/src/avatar.tsx)

Happy to open a PR that addresses this.

atomiks:
[...]

PR #4469, opened 2026-03-28
[avatar] Fix flash when image is cached

mj12albert:
Partially addresses https://github.com/mui/base-ui/issues/4468

Adds an `image.complete` check per the suggestion in the issue matching Radix


atomiks:
## Codex Review (GPT-5.4)

This is a focused bug-fix PR for cached Avatar images. After re-checking the actual Radix demo source and the Base UI patch together, I no longer think this PR should be blocked on changing `Avatar.Root` / `Avatar.Fallback` coordination: the user-visible blank-then-fallback behavior in Radix's demo comes from an explicit fallback delay, while this PR stays narrowly focused on cached-image detection.

### 1. Bugs / Issues (None)

I did not find a concrete, branch-relevant correctness issue in the current patch.

| Area | Assessment | Notes |
|------|:----------:|-------|
| Cached-image path | **Improved** | The new `image.complete` branch avoids waiting for `onload` once the image object has been initialized |
| Scope discipline | **Good** | The patch stays within `useImageLoadingStatus` plus targeted tests instead of spreading new policy across Avatar parts |
| Comparison to Radix demo | **Earlier concern withdrawn** | The blank-first behavior in the demo is driven by delayed fallback rendering, not by extra Root/Fallback coordination in the core component |

The implementation in `packages/react/src/avatar/image/useImageLoadingStatus.ts` is small and directionally sound for the reported cached-image case:

```ts
image.src = src;

if (image.complete) {
  setLoadingStatus(image.naturalWidth > 0 ? 'loaded' : 'error');
}
```

That makes this a reasonable narrow fix for “cached image already available when the loading probe runs,” which is what the PR title and code are actually changing.

The earlier concern I had about needing to suppress fallback during the initial SSR window does not hold up as a requirement from the Radix demo itself. Their docs/demo examples explicitly opt into delayed fallback rendering:

```jsx
<Avatar.Fallback delayMs={600}>CT</Avatar.Fallback>
```

So the demo's blank-until-later behavior is not evidence that the core Avatar component is doing child inspection or special fallback coordination by default. In that light, this PR's narrower approach looks appropriately scoped.

I ran `pnpm test:jsdom AvatarImage --no-watch` and `pnpm test:chromium AvatarImage --no-watch`; both passed.

The added tests cover the important cases for this patch's actual scope:

| Scenario | Covered? | Notes |
|----------|:--------:|-------|
| Cached image resolves immediately | **Yes** | Both mocked cached behavior and browser caching paths are exercised |
| Cached broken image | **Yes** | Confirms the fast path still falls back correctly when `naturalWidth` is `0` |
| Full SSR + delayed-fallback UX | **No** | Not directly covered, but that is broader than the specific behavior this PR changes |

### Merge Confidence Score

Overall merge confidence is 4/5. The patch is small, the targeted tests are green, and the earlier requirement to change fallback/root behavior was based on an incorrect reading of the Radix demo. I would be comfortable merging this version as a focused cached-image fix.

atomiks:
The problem is that the fallback actually shouldn't be rendered until hydration, which makes the image (and caching) take precedence during SSR

See Radix (**uncached**):


https://github.com/user-attachments/assets/b9d095f0-d495-47d1-a19f-0677e2533db0

Afterwards, the image is always present during SSR due to the cache, with no fallback flash.

atomiks (approved):
The fallback-flash UX is usually handled with `delay` on `<Avatar.Fallback>`. Radix uses `delayMs={600}` in its demos, so I’ve updated our demos here to match that.

Radix also does not appear to render the avatar `` in SSR HTML. The underlying issue is really about the pre-hydration SSR, but solving that requires extra coordination between Avatar.Root, Avatar.Image, and Avatar.Fallback such as tree-walking React.Children, which feels too fragile...

---

Unrelated  but important observation, which applies to my comment earlier:

- In Chrome, Radix’s site appears to hide the pre-hydration state on cached reloads much more aggressively than Base UI, which initially made it look like the image was SSR’d. For example its left sidebar doesn't scroll into view (from `.scrollTop === 0`) on cached reloads so it looks like it hydrated immediately, somehow(?)
- In Safari, Radix visibly flashes on reload too, so I don’t think the image is actually present in SSR.
- My takeaway is that the component behavior is now aligned with Radix, and the remaining difference seems to be website/browser hydration behavior rather than this Avatar logic

PR #4475, opened 2026-03-29
[popups] Prevent extra rerenders when opening popups

mdm317:
### summary
When a popup uses a detached trigger,
`domReference: rootContext.useState('domReferenceElement')` already holds a valid DOM element.
However, during the open sequence, the store value is temporarily set to null and then restored to the same DOM element `dom → null → dom`.
This results in unnecessary re-renders.

---

When tested using /experiments/menu/menu-disabled-items,
React Profiler (DevTools) shows:

 - Before: 13 commits
 - After: 12 commits

### fix

Avoid syncing domReferenceElement until localDomReference is set.

### test
 - Implemented benchmark tests.
   - The performance improvement is slight and becomes more noticeable under higher CPU throttling.
 - Confirmed the reduction in commit count using React Profiler (DevTools).


atomiks:
## Codex Review (GPT-5.4)

This is a performance-focused PR that removes an extra detached-trigger store transition during popup open and adds a dedicated benchmark page for that path. I validated the `menu-open` experiment in Chromium with `4x` CPU slowdown using the new `Run 20` benchmark path, and the perf signal is positive, but there is still one hook-level correctness concern worth addressing before merge.

### 1. Bugs / Issues
#### 1. 🟡 Undefined sentinel also leaks into `referenceElement` sync (non-blocking)

In `packages/react/src/floating-ui-react/hooks/useFloating.ts`, the new `undefined` sentinel is intended to preserve `domReferenceElement` until the local ref arrives, but it is also passed through `useSyncedValue('referenceElement', localDomReference)`. This is based on the code path rather than a directly reproduced failure.

From `packages/react/src/floating-ui-react/hooks/useFloating.ts`:

```ts
const [localDomReference, setLocalDomReference] = React.useState<
  NarrowedElement<ReferenceType> | null | undefined
>(undefined);

rootContext.useSyncedValue('referenceElement', localDomReference);
rootContext.useSyncedValue(
  'domReferenceElement',
  localDomReference === undefined ? rootContextElements.domReference : localDomReferenceElement,
);
```

This matters because `ReactStore.useSyncedValue()` writes the value directly into store state, so the store now briefly sees `referenceElement = undefined` even though `FloatingRootState.referenceElement` is `ReferenceType | null`. That can break the `positionReference === referenceElement` invariant that sibling sync code relies on.

**Fix:** Keep the `undefined` sentinel scoped to `domReferenceElement` only, or guard the `referenceElement` sync so it never writes `undefined`.

### 2. Benchmark Summary

I measured the new `menu-open` benchmark in Chromium at `4x` CPU slowdown using `Run 20`, with one uncounted primer run followed by five measured samples.

| Ref | Run 1 | Run 2 | Run 3 | Run 4 | Run 5 | Avg of means |
|-----|------:|------:|------:|------:|------:|-------------:|
| base `b353c1e` | 85.0 ms | 83.6 ms | 83.9 ms | 85.2 ms | 83.3 ms | 84.2 ms |
| PR `d8ee3e1` | 82.0 ms | 76.2 ms | 76.0 ms | 76.9 ms | 74.5 ms | 77.1 ms |

That is a mean improvement of about `7.1 ms` on this benchmark, or roughly `8.4%`. Tail latency improved too: the average `p95` moved from `90.6 ms` on base to `81.3 ms` on the PR.

### 3. Correctness Trade-offs

The performance result is directionally convincing for the path this PR targets: detached-trigger popup open. The main risk is that the current implementation broadens the sentinel from a `domReferenceElement` workaround into a more general `referenceElement` state change inside a shared hook.

The new benchmark page is useful for manual validation, but it still leaves the behavioral fix unguarded in CI. A focused detached-trigger regression test would make this safer to maintain.

### Merge Confidence Score

Overall merge confidence is 3/5. The measured `menu-open` improvement looks real, but I would address issue 1 and add targeted regression coverage before merging so the perf win does not come with a subtle store-state regression.

packages/react/src/floating-ui-react/hooks/useFloating.ts
```diff
@@ -73,7 +73,10 @@ export function useFloating(options: UseFloatingOptions = {}): UseFloatingReturn
     NarrowedElement<ReferenceType> | null | undefined
   >(undefined);
   const [localFloatingElement, setLocalFloatingElement] = React.useState<HTMLElement | null>(null);
-  rootContext.useSyncedValue('referenceElement', localDomReference);
+  rootContext.useSyncedValue(
+    'referenceElement',
+    localDomReference === undefined ? null : localDomReference,
```
atomiks: Wouldn't this work: `localDomReference ?? null`

mdm317: Thanks for pointing that out — I’ve updated it.

PR #4481, opened 2026-03-30
[field] Move field registration into Field.Root

atomiks:
This moves field registration ownership into `Field.Root` so field-aware controls only publish lightweight control descriptors through the field context.

The main tradeoff is that `Field` itself gets a bit bigger because it now owns the optional Base UI form-integration path. That is intentional: consumers who use controls like `Select`, `Slider`, `Checkbox`, `Switch`, `RadioGroup`, `Combobox`, and `NumberField` without Base UI `Field`/`Form` no longer pay for that registration logic by default.

The package-level bundle bot still reports a small increase for `@base-ui/react` because the root barrel exports `Field`, which now owns the shared registration path. That root-package number is a noisy aggregate for this change, and the more useful read is:

- `Field` itself got bigger because it now owns registration work that used to live in each control.
- The field-aware controls got smaller because they stopped each carrying that logic themselves.
- When `Field` is bundled together with those controls, the net effect is still smaller overall because the shared Field-owned path deduplicates better than the old per-control setup.

## Changes

- Move form registration out of each control and into `Field.Root` via `useFieldControlRegistration`.
- Add `useRegisterFieldControl` so the controls share one small registration publisher instead of repeating the same effect boilerplate.
- Remove the intermediate registration state hop in `Field.Root` and keep the active control registration in a ref-backed path.
- Preserve submit-time value normalization for controls that need it, including `Select`, `Combobox`, `Field.Control`, and unselected `RadioGroup`.
- Add a regression test for unselected `RadioGroup` values submitted through `Form.onFormSubmit`.
- Simplify the shared registration path so only the long-lived registry callbacks keep stable identity.

## Bundle impact

Bundle bot for the root barrel currently reports `@base-ui/react` at `+630 B` parsed and `+99 B` gzip.

Local bundle-size snapshot diff against fresh `upstream/master` for the narrower entrypoints affected by this refactor:

The last two rows below are real synthetic-bundle measurements where those entrypoint source files are imported together, rather than sums of the individual deltas.

| Entry | Parsed delta | Gzip delta |
| --- | ---: | ---: |
| `Field` | `+890 B` | `+194 B` |
| `Checkbox` | `-414 B` | `-154 B` |
| `CheckboxGroup` | `-482 B` | `-171 B` |
| `Combobox` | `-504 B` | `-175 B` |
| `NumberField` | `-418 B` | `-111 B` |
| `RadioGroup` | `-424 B` | `-159 B` |
| `Select` | `-503 B` | `-158 B` |
| `Slider` | `-452 B` | `-168 B` |
| `Switch` | `-445 B` | `-191 B` |
| Synthetic bundle: field-aware controls together | `-738 B` | `-270 B` |
| Synthetic bundle: field-aware controls + `Field` | `+541 B` | `+85 B` |

## Benchmark impact

Chromium production benchmark against `upstream/master` after removing the extra `Field.Root` registration state hop:

| Scenario | Master | This branch | Delta |
| --- | ---: | ---: | ---: |
| `Slider x400` without `Field` | `8.2 ms` | `7.3 ms` | `-0.9 ms` |
| `Slider x400` with `Field` | `12.8 ms` | `13.2 ms` | `+0.4 ms` |
| `Select x250` without `Field` | `22.4 ms` | `19.7 ms` | `-2.7 ms` |
| `Select x250` with `Field` | `30.2 ms` | `32.7 ms` | `+2.5 ms` |

So the final shape is:

- Controls used without `Field` are faster and slimmer.
- Wrapped cases are much closer to baseline after removing the extra registration state hop in `Field.Root`.
- The remaining runtime tradeoff is concentrated in the optional `Field` path rather than paid by every field-aware control by default.

### DIFF
--- a/packages/react/src/checkbox-group/CheckboxGroup.tsx
+++ b/packages/react/src/checkbox-group/CheckboxGroup.tsx
@@ -7,10 +7,10 @@ import { useRenderElement } from '../utils/useRenderElement';
 import { CheckboxGroupContext } from './CheckboxGroupContext';
 import type { FieldRootState } from '../field/root/FieldRoot';
 import { useFieldRootContext } from '../field/root/FieldRootContext';
+import { useRegisterFieldControl } from '../field/root/useRegisterFieldControl';
 import { useLabelableContext } from '../labelable-provider/LabelableContext';
 import type { BaseUIComponentProps } from '../utils/types';
 import { fieldValidityMapping } from '../field/utils/constants';
-import { useField } from '../field/useField';
 import { PARENT_CHECKBOX } from '../checkbox/root/CheckboxRoot';
 import { useCheckboxGroupParent } from './useCheckboxGroupParent';
 import type { BaseUIChangeEventDetails } from '../utils/createBaseUIEventDetails';
@@ -92,14 +92,11 @@ export const CheckboxGroup = React.forwardRef(function CheckboxGroup(
     }
   }, []);
 
-  useField({
+  useRegisterFieldControl({
     enabled: !!fieldName,
     id,
-    commit: validation.commit,
     value,
     controlRef,
-    name: fieldName,
-    getValue: () => value,
   });
 
   const resolvedValue = value ?? EMPTY_ARRAY;
--- a/packages/react/src/checkbox-group/CheckboxGroup.tsx
+++ b/packages/react/src/checkbox-group/CheckboxGroup.tsx
@@ -92,11 +92,10 @@ export const CheckboxGroup = React.forwardRef(function CheckboxGroup(
     }
   }, []);
 
-  useRegisterFieldControl({
+  useRegisterFieldControl(controlRef, {
     enabled: !!fieldName,
     id,
     value,
-    controlRef,
   });
 
   const resolvedValue = value ?? EMPTY_ARRAY;
--- a/packages/react/src/checkbox/root/CheckboxRoot.tsx
+++ b/packages/react/src/checkbox/root/CheckboxRoot.tsx
@@ -167,11 +167,10 @@ export const CheckboxRoot = React.forwardRef(function CheckboxRoot(
     };
   }, [registerControlId, controlSourceRef]);
 
-  useRegisterFieldControl({
+  useRegisterFieldControl(controlRef, {
     enabled: !groupContext,
     id,
     value: checked,
-    controlRef,
   });
 
   const inputRef = React.useRef<HTMLInputElement>(null);
--- a/packages/react/src/checkbox/root/CheckboxRoot.tsx
+++ b/packages/react/src/checkbox/root/CheckboxRoot.tsx
@@ -16,8 +16,8 @@ import { mergeProps } from '../../merge-props';
 import { useButton } from '../../use-button/useButton';
 import type { FieldRootState } from '../../field/root/FieldRoot';
 import { useFieldRootContext } from '../../field/root/FieldRootContext';
+import { useRegisterFieldControl } from '../../field/root/useRegisterFieldControl';
 import { useFieldItemContext } from '../../field/item/FieldItemContext';
-import { useField } from '../../field/useField';
 import { useFormContext } from '../../form/FormContext';
 import { useLabelableContext } from '../../labelable-provider/LabelableContext';
 import { useAriaLabelledBy } from '../../labelable-provider/useAriaLabelledBy';
@@ -167,14 +167,11 @@ export const CheckboxRoot = React.forwardRef(function CheckboxRoot(
     };
   }, [registerControlId, controlSourceRef]);
 
-  useField({
+  useRegisterFieldControl({
     enabled: !groupContext,
     id,
-    commit: validation.commit,
     value: checked,
     controlRef,
-    name,
-    getValue: () => checked,
   });
 
   const inputRef = React.useRef<HTMLInputElement>(null);
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -437,11 +437,10 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
   const { openMethod, triggerProps } = useOpenInteractionType(open);
   const getFieldValue = useStableCallback(() => fieldStringValue);
 
-  useRegisterFieldControl({
+  useRegisterFieldControl(inputInsidePopup ? triggerRef : inputRef, {
     id,
     value: fieldRawValue,
     getValue: getFieldValue,
-    controlRef: inputInsidePopup ? triggerRef : inputRef,
   });
 
   const forceMount = useStableCallback(() => {
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -34,7 +34,7 @@ import {
 import { selectors, type State as StoreState } from '../store';
 import { useOpenChangeComplete } from '../../utils/useOpenChangeComplete';
 import { useFieldRootContext } from '../../field/root/FieldRootContext';
-import { useField } from '../../field/useField';
+import { useRegisterFieldControl } from '../../field/root/useRegisterFieldControl';
 import { useFormContext } from '../../form/FormContext';
 import { useLabelableId } from '../../labelable-provider/useLabelableId';
 import { createCollatorItemFilter, createSingleSelectionCollatorFilter } from './utils';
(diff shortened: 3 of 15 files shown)


atomiks:
## Codex Review (GPT-5.4)

This is a `refactor` with secondary tags `perf`, `tests`, and `public-api`. I reviewed the full branch against a freshly fetched `upstream/master`, checked the ownership changes in `Field.Root`, and re-checked the targeted validation and bundle-impact story after the follow-up slimming pass.

### 1. Bugs / Issues (None)

### 2. Behavior Preservation Assessment

The core behavior change looks sound: registration moved from each field-aware control into `Field.Root`, while the controls themselves now publish lightweight descriptors. The branch also preserves the one behavior regression that surfaced during review by restoring `null` submission for an unselected `RadioGroup` and adding coverage for that case.

### 3. Ownership / Architecture Assessment

The new ownership split is easier to reason about. `Field.Root` now owns Base UI form registration, simple controls stop importing the old heavier registration path by default, and the follow-up `useRegisterFieldControl` helper avoids reintroducing the old cost through repeated per-control effect boilerplate.

The latest simplification in `Field.Root` also keeps the implementation smaller without muddying the model: a field now tracks a single active control publisher, which matches the intended “one active field control” mental model much better than keeping registration mechanics scattered across every control.

### 4. Simplification and DRY Assessment

This diff meaningfully reduces duplication rather than just moving it. The old `useField` hook disappears, the shared registration logic is consolidated into `useFieldControlRegistration`, and the controls only keep custom `getValue` serializers where the submit-time value actually differs from the internal state shape.

### 5. Performance / Operational Risk

The bundle tradeoff reads correctly now:

| Area | Result | Notes |
| --- | ---: | --- |
| Root barrel (`@base-ui/react`) | Slight increase | Expected because `Field` now owns the shared registration path |
| `Field` entry | Increase | Also expected for the same reason |
| Field-aware control entries | Decrease | Controls no longer carry Base UI field-registration logic by default |
| Field-aware controls + `Field` | Slight decrease | Shared Field-owned path deduplicates better than the old per-control setup |

That shape matches the design goal here: the optional Base UI form-integration cost is now attached primarily to `Field`, not to every control whether consumers use `Field` or not.

I re-checked the targeted validation that matters for this refactor: `pnpm typescript` and `pnpm test:jsdom FieldRoot RadioGroup SelectRoot SliderRoot CheckboxRoot CheckboxGroup SwitchRoot NumberFieldRoot ComboboxRoot --no-watch`. That covers the moved registration path, the affected field-aware controls, and the `RadioGroup` submit-shape regression that was fixed during the branch.

### Merge Confidence Score

Overall merge confidence is 4/5. I did not find any remaining actionable issues in the current branch, and the architecture, targeted validation, and bundle tradeoff all line up with the stated goal of making Base UI field integration optional from the control side. I would be comfortable merging this as-is.

flaviendelangle (commented):
# Code Review: [field] Move field registration into Field.Root (#4481)

**Confidence: 4/5**

> High confidence in the architectural assessment and the `RadioGroup.getValue` issue. Docked one point because I haven't been able to run the test suite to verify the `NumberField` null coercion change end-to-end, and there may be subtle interaction timing effects from the new state-update path that only surface under specific render scenarios.

## Overview

Centralizes field-control registration from 9 individual controls into `Field.Root` via two new hooks:

- **`useRegisterFieldControl`** — lightweight publisher in each control
- **`useFieldControlRegistration`** — subscriber in `Field.Root` (renamed from `useField`)

Controls used without `Field`/`Form` no longer bundle registration logic, producing meaningful per-control bundle savings (-182B to -362B gzip per control).

## Positives

- Clean publisher/subscriber separation — controls are lighter, `Field.Root` owns integration
- Excellent bundle impact analysis in the PR description
- Good regression test for unselected `RadioGroup` form submission
- Consistent refactoring pattern applied uniformly across all 9 controls

## Issues

### 1. `RadioGroup.getValue` not stabilized (bug-level)

```tsx
// RadioGroup.tsx — creates a new closure every render
getValue: () => checkedValue ?? null,
```

This is in the dependency array of `useIsoLayoutEffect` inside `useRegisterFieldControl`, so the effect tears down and re-registers on **every render**. `Combobox` and `FieldControl` correctly use `useStableCallback` — `RadioGroup` should too:

```tsx
const getFieldValue = useStableCallback(() => checkedValue ?? null);
useRegisterFieldControl({ ..., getValue: getFieldValue });
```

### 2. `NumberField` lost `?? null` coercion (behavioral change)

Old: `getValue: () => value ?? null`
New: passes `value` with no `getValue`, so the fallback in `useFieldControlRegistration` becomes `() => value`. If `value` is `undefined`, form submission gets `undefined` instead of `null`. This could break consumers checking `=== null`.

Same applies to `Slider` (old: `() => valueUnwrapped`, now relies on default) and `Switch`/`Checkbox` (old: `() => checked`), though those are less likely to be `undefined`.

### 3. Removed explanatory comment

The `flushSync` comment was removed:

```diff
-          // Synchronously update the validity state so the submit event can be prevented.
           ReactDOM.flushSync(() => commit(nextValue));
```

`flushSync` is non-obvious enough to warrant keeping a comment.

### 4. Extra state cycle on value changes (minor, informational)

Each value change now triggers effect cleanup → `setRegisteredFieldControl(null)` → setup → `setRegisteredFieldControl(newReg)` in `FieldRoot`. React batches these in layout effects so it's likely one re-render, but it's a new intermediate `null` state that didn't exist before. Worth keeping in mind if validation timing issues arise.

## Test Coverage

- Regression test for unselected `RadioGroup` is solid
- Would benefit from a test verifying that a control's `getValue` is called (not just `value`) during form submission for `Combobox`, `Select`, and `FieldControl` — these are the controls that rely on the `getValue` path for submit-time normalization

## Verdict

Architecturally sound refactor. **Item 1 (RadioGroup)** should be fixed before merge. **Item 2 (null coercion)** needs explicit confirmation that the behavior change is intentional.

atomiks:
## Codex Review (GPT-5.4)

This remains a `refactor` with secondary tags `perf`, `tests`, and `public-api`. I reviewed the full branch against fresh `upstream/master`, did a branch-wide code pass with subagent sweeps for lifecycle/perf and behavior/regression risk, and reran `pnpm typescript` plus the targeted jsdom suite on the current head.

No blocking or follow-up findings on the latest branch state.

### 2. Behavior Preservation Assessment

The risky contracts from the earlier refactor still look preserved on the current head:

- field-aware controls no longer import `field/useField.ts` directly
- `Field.Root` owns registration, but submit-time serialization still stays with the controls that need custom `getValue` behavior
- the `RadioGroup` empty-submit path is explicitly covered by the new regression test, and the current implementation still normalizes the unselected value to `null`
- the `NumberField` null-coercion concern does not reproduce because the registered value is already normalized before it reaches the shared field-registration layer

I also do not see a new stale-closure problem in the simplified `useFieldControlRegistration` shape. The callbacks that remain stabilized are the ones that actually escape into the form registry or sit in long-lived effect dependencies.

### 3. Ownership / Architecture Assessment

The ownership model is clearer than the original per-control registration design:

- controls publish lightweight descriptors through `useRegisterFieldControl`
- `Field.Root` owns the Base UI registration path through `useFieldControlRegistration`
- the extra `registeredFieldControl` state hop is gone, so the wrapped-case performance fix did not require moving ownership back into each control

That is the right boundary for the stated goal of making Base UI field integration optional for consumers who are not using `Field` or `Form`.

### 4. Simplification and DRY Assessment

The latest pass is a real simplification rather than code motion.

**Handled:**
- removed `field/useField.ts` entirely
- collapsed the registration flow into one shared Field-owned path
- removed the intermediate React state used only to relay active registration
- trimmed the latest-value ref boilerplate in `useFieldControlRegistration`
- kept stable identity only where it materially matters, instead of wrapping every local helper

The only duplication left in the shared hook is the repeated `fields.set(...)` shape between immediate registration and metadata refresh, and that reads like an acceptable tradeoff for keeping the fast path explicit.

The targeted validation covers the risky surface of this refactor well. I reran:

- `pnpm typescript`
- `pnpm test:jsdom FieldRoot RadioGroup SelectRoot SliderRoot CheckboxRoot CheckboxGroup SwitchRoot NumberFieldRoot ComboboxRoot --no-watch`

That gives direct coverage for the Field-owned registration path, the controls with custom submit serializers, and the simpler value-publishing controls.

### 6. Previous Review Issues Status

**Resolved:**
- `RadioGroup` no longer re-registers on every render due to an unstable serializer closure
- the `flushSync` rationale comment is restored
- the wrapped-case performance regression source was removed by deleting the extra registration state hop
- the later simplification pass did not reopen the earlier null/submit semantics concerns

### Merge Confidence Score

Overall merge confidence is 4/5. I am highly confident in the current branch state: the ownership model is cleaner, the targeted validation covers the risky behavioral paths, and neither my review nor the subagent sweeps surfaced a remaining correctness regression. I would be comfortable merging this as-is.

PR #4482, opened 2026-03-30
[switch][checkbox][radio][select][combobox] Add suppressHydrationWarning to hidden inputs

devxoul:
Browser extensions (e.g. Dark Reader) can modify the inline styles of visually-hidden `<input>` elements before React hydrates, causing hydration mismatch warnings.

This PR adds `suppressHydrationWarning` to the hidden `<input>` in `SwitchRoot`, `CheckboxRoot`, `RadioRoot`, `SelectRoot`, `AriaCombobox`, `NumberFieldRoot`, and `SliderThumb`.

**Motivation**

When using any of these components in a Next.js (or any SSR) app with a browser extension like Dark Reader installed, the console is flooded with hydration mismatch warnings. The extension injects custom CSS properties (e.g. `--darkreader-inline-*`) and `data-darkreader-*` attributes into the server-rendered HTML before React hydrates, causing React to detect a mismatch between the server-rendered DOM and the client-expected virtual DOM.

Since these inputs are visually hidden and never visible to the user, suppressing hydration warnings on them is safe and consistent with the existing approach used by `TabsIndicator` and `NumberFieldInput`.

**Changes**

| Component | File |
|-----------|------|
| Switch | `packages/react/src/switch/root/SwitchRoot.tsx` |
| Checkbox | `packages/react/src/checkbox/root/CheckboxRoot.tsx` |
| Radio | `packages/react/src/radio/root/RadioRoot.tsx` |
| Select | `packages/react/src/select/root/SelectRoot.tsx` |
| Combobox | `packages/react/src/combobox/root/AriaCombobox.tsx` |
| Number Field | `packages/react/src/number-field/root/NumberFieldRoot.tsx` |
| Slider | `packages/react/src/slider/thumb/SliderThumb.tsx` |


devxoul:
Thanks for the thorough review!

You're right — the original description overstated existing coverage. `SliderThumb`'s nested `<input type="range">` and `NumberFieldRoot`'s hidden form `<input>` do lack the prop.

I've updated the PR description to remove the inaccurate claim and explicitly note the remaining gaps. Happy to cover those two in a follow-up if you'd like.

atomiks:
@devxoul what's the reason not to add the prop for Number Field and Slider?

devxoul:
Good point — no reason not to. Added both, updated the description.

PR #4485, opened 2026-03-30
[popups] Ignore pinch-zoom shifting

atomiks:
Only Navigation Menu popups and the root Context Menu popup use the layout viewport for `shift()`. This keeps them anchored to page content instead of following the smaller visual viewport during pinch zoom. Context Menu submenus and all other anchored popups keep their existing visual viewport behavior.

## Changes

- Use Floating UI's `layoutViewport` boundary for Navigation Menu positioners.
- Use the same boundary for the root Context Menu positioner, replacing its custom layout viewport rectangle.
- Keep Context Menu submenus, Menu, Menubar, and other anchored popup components on the default visual viewport.
- Keep `flip()` and `size()` on the visual viewport so software keyboard collision handling is unchanged.


flaviendelangle (approved):
## Code Review: [popups] Ignore pinch-zoom shifting

**Rating: 4/5** ⭐⭐⭐⭐

### Overview

This PR extends the existing layout-viewport-based `shift()` behavior — previously scoped only to context menus (`shiftCrossAxis`) — to **all popups** when the user is actively pinch-zooming (`visualViewport.scale !== 1`). This prevents disorienting popup movement during pinch zoom on touch devices.

### What's Good

- **Minimal, focused diff** — net +4/−3 in a single file. The change is easy to reason about.
- **Correct use of project utilities** — uses `ownerWindow()` as required by project conventions, instead of accessing the global `window` directly.
- **SSR-safe** — `win.visualViewport?.scale ?? 1` gracefully handles environments where `visualViewport` is `undefined` or `null`.
- **Backward compatible** — the `shiftCrossAxis` condition is preserved in the OR expression, so existing context menu behavior is untouched.
- **Clean variable naming** — `useLayoutViewport` clearly communicates intent and replaces a bare `shiftCrossAxis` that was harder to read in context.

### Concerns

- **Tests were added then removed** — The first commit introduced well-structured unit tests that validated both the default behavior and the pinch-zoom case. The second commit deletes them entirely. This leaves the new `visualViewport.scale` code path without any test coverage. Could these tests be kept, perhaps adjusted to be less reliant on mocking Floating UI internals? Even a simpler assertion (e.g., verifying the `useLayoutViewport` logic in isolation) would improve confidence against regressions.

### Minor Nits

- The dependency array at [line 310](packages/react/src/utils/useAnchorPositioning.ts#L310) doesn't include `visualViewport.scale`, but this is correct — the value is read dynamically inside the callback each time Floating UI recalculates positioning, so it doesn't need to be a memoization dep.

### Summary

Clean, well-scoped fix that correctly generalizes an existing safeguard. The only real concern is the removal of test coverage in the second commit — I'd love to see at least some test retained for the new behavior.

atomiks:
This should be built into Floating UI as a string option due to https://github.com/floating-ui/floating-ui/pull/3387 unless we implement that on our side. It's pretty rare to encounter but it does happen, so I'll try to do a release that handles this

flaviendelangle (approved):
# PR review

This is a clean, behavior-preserving refactor: it extracts the inline trigger-bounds check from the menu, select, and combobox triggers into a shared `isMouseWithinBounds` util, and widens the drift tolerance from `2px` to `5px` so a fast click that drifts slightly during press-release isn't mistaken for a drag-off cancellation. All three call sites are guarded against a null trigger and pass a native `MouseEvent`, matching the util signature. Nothing is merge-blocking. Combobox and util tests pass locally.

## Bugs (0)

No findings.

## Tests (1)

### 1. ℹ️ New 5px tolerance is only integration-tested for combobox, not menu/select

**Location:** `packages/react/src/menu/trigger/MenuTrigger.tsx:149`, `packages/react/src/select/trigger/SelectTrigger.tsx:204`

The behavioral change (release up to 5px outside the trigger keeps the popup open) got dedicated integration tests only in `ComboboxTrigger.test.tsx`. `MenuTrigger` and `SelectTrigger` share the same util, and the exact boundary is directly unit-tested in `getPseudoElementBounds.test.ts`, so the core logic is well covered — but there's no test pinning that menu/select actually route through `isMouseWithinBounds` on release.

**Failure scenario:** A future refactor that removes or bypasses the bounds check in `MenuTrigger`/`SelectTrigger` (e.g. reordering the early-return guards) would not be caught by any test, since only combobox exercises this path end-to-end.

**Fix:** Optional — add a menu and select equivalent of the combobox "keeps open within 5px / closes beyond 5px" test, or leave as-is given the util is directly boundary-tested.

## Simplifications (0)

No findings. The refactor removes three duplicated copies of the inline bounds math and the `BOUNDARY_OFFSET` constant; all consumers were updated consistently and no orphaned `getPseudoElementBounds` direct usages remain.

## Docs (0)

No findings. The new JSDoc and the `BOUNDARY_OFFSET` comment accurately describe the tolerance and its rationale.

## Verdict

**Approve** — clean, well-tested refactor with an intended tolerance widening; the only observation is an optional test-coverage symmetry note across the three triggers.

---

PR #4501, opened 2026-04-01
[tooltip] Fix flickering when nesting tooltip inside another tooltip's trigger

### LINKED ISSUE #4499: [tooltip] Unexpected behaviour when nesting tootip inside another tooltip's trigger
# Bug report

As reported on Discord: https://discord.com/channels/1287292451308048406/1287292451308048409/1488499829104443424

## Current behavior

The inner and outer tooltips do not work correctly. There is flicker.

https://github.com/user-attachments/assets/08c3c2e1-c5e7-4aa8-bda6-fb7304815862

If we add `open={false}` to the outer tooltip:

https://github.com/user-attachments/assets/3d635b23-53de-4aa4-b582-c9592a0128a0

## Expected behavior

The inner and outer tooltips work correctly. No flicker.

## Reproducible example

https://stackblitz.com/edit/zrcwexu5?file=src%2FApp.tsx

## Base UI version

1.3.0

Chrome

## Which OS are you using?

macOS

Voiceover/JAWS etc.

## Additional context

For context, here the real world example of where I need nested tooltips:

https://github.com/user-attachments/assets/a762d5f4-5310-44b5-997d-b30e6fada002

michaldudak:
Prevent the outer tooltip from opening when the cursor is over an enabled nested tooltip trigger. When the outer tooltip is already open and the pointer moves onto a nested trigger, close it when the bubbling `mouseover` event reaches the outer trigger and detects a nested trigger target so that only the innermost tooltip is visible. Disabled nested triggers are excluded from detection so the parent tooltip still opens.

The close path is gated to hover-opened tooltips only — focus-opened tooltips remain visible until blur.

Closes #4499


michaldudak:
## Codex Review

### Overview
This PR fixes the nested-tooltip flicker reported in [#4499](https://github.com/mui/base-ui/issues/4499) by preventing a parent tooltip from opening, or staying open, while the pointer is over an enabled nested trigger. It does that by adding a nested-trigger guard in the tooltip hover path, closing hover-opened ancestors on nested trigger entry, and documenting/testing the new behavior.

### Findings (None)

I reviewed the hover-delay path in `packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts`, the nested-trigger event handling in `packages/react/src/tooltip/trigger/TooltipTrigger.tsx`, and the added delayed-hover regression coverage in `packages/react/src/tooltip/root/TooltipRoot.test.tsx`, and I did not find a remaining correctness issue in the current branch state.

### Confidence: 4/5
High confidence. The branch includes targeted coverage for the key delayed-hover sequence where the pointer enters the outer trigger first and then moves onto the nested trigger before the delay expires. Residual risk is limited to the usual complexity around hover timing and composed event flow.

### Notes
- The PR conversation appears to contain only bot comments, so there is no outstanding human review feedback to reconcile.

atomiks (commented):
## Codex Review (GPT-5.4)

This is a bug-fix PR that does fix the light-DOM nested-tooltip flicker path, and the targeted `TooltipRoot` suite plus fresh size snapshots against `master` back that up. The remaining concern is that the new nested-trigger detection is not shadow-DOM-safe, and the current shape adds roughly `+996 B parsed / +326 B gzip` to `@base-ui/react/tooltip` for a tooltip-only edge case.

### 1. Bugs / Issues
#### 1. 🟡 Nested-trigger detection is not shadow-DOM-safe (non-blocking)

In `packages/react/src/tooltip/trigger/TooltipTrigger.tsx`, the delayed-open guard finds the hovered nested trigger via `document.elementFromPoint(...).closest(...)`, while the already-open close path depends on a retargeted bubbling DOM event. This is code-based reasoning rather than a local repro, but in shadow-root embeddings `closest()` does not walk past the shadow boundary to the trigger host, so the ancestor tooltip can still open even though the pointer is over an enabled nested trigger.

From `packages/react/src/tooltip/trigger/TooltipTrigger.tsx`:

```ts
const elementUnderCursor = doc.elementFromPoint(pos.x, pos.y);
if (!elementUnderCursor) {
  return true;
}

const nearestTrigger = elementUnderCursor.closest(ENABLED_TOOLTIP_TRIGGER_SELECTOR);
if (nearestTrigger && nearestTrigger !== triggerEl && contains(triggerEl, nearestTrigger)) {
  return false;
}
```

This matters because the rest of the floating layer explicitly uses shadow-safe targeting helpers like `contains` and `getTarget`, so this branch narrows Tooltip’s composed-tree guarantees relative to adjacent primitives.

**Fix:** Replace the `closest()`/custom-event combination with a composed-path-aware signal, or track nested-trigger hover from bubbling `pointerover`/`mouseover` state so the guard stays shadow-DOM-safe.

The branch addresses both halves of the reported bug in normal light DOM: it suppresses a delayed outer open when the pointer is already over the inner trigger, and it closes a hover-opened outer tooltip when the pointer moves onto the nested trigger. The new `TooltipRoot` tests cover those light-DOM cases well, and I did not find another concrete interaction defect in that path.

The downside is that the fix is split across two mechanisms in `TooltipTrigger`: cursor-position sampling plus `elementFromPoint()` for future opens, and a bubbling custom event for already-open ancestors. That makes the patch heavier and harder to reason about than the original bug warrants.

Existing floating utilities in this repo consistently lean on store-based trigger tracking and shadow-safe helpers. This patch is the one place that falls back to raw DOM traversal plus a public marker attribute, so it breaks that pattern instead of extending it.

The biggest simplification opportunity is to keep this tooltip-specific behavior out of `useHoverReferenceInteraction`. Adding `shouldOpen` widens a shared hover primitive for a case that only Tooltip uses, which means extra hot-path branching for other consumers as well.

Bundle snapshot results against fresh `master`:

| Entrypoint | `master` | PR | Delta |
|---|---:|---:|---:|
| `@base-ui/react/tooltip` | `95,381 / 32,223` | `96,377 / 32,549` | `+996 parsed / +326 gzip` |
| `@base-ui/react` | `465,728 / 146,125` | `466,689 / 146,437` | `+961 parsed / +312 gzip` |

I would try to collapse this to one local signal in `TooltipTrigger` instead of the current two-part design, and I would avoid documenting `data-tooltip-trigger` unless the intent is to support that marker as public API long-term.

`pnpm test:jsdom TooltipRoot --no-watch` passed, including the new nested-tooltip cases. Those tests cover the light-DOM happy path, delayed-open cancellation, already-open ancestor close, and the explicit `disabled` prop case.

The main missing regression case is a shadow-root embedding, which is exactly where the current detection logic looks fragile. If this approach stays, I would add that case before merging.

### Merge Confidence Score

Overall merge confidence is `3/5`. The reported flicker looks fixed for the newly tested light-DOM scenarios, but I would address issue 1 before merging if Tooltip is expected to preserve the same shadow-DOM behavior as the rest of the floating utilities. Even if that risk is accepted, I would still simplify the implementation so the `~0.3 KB gzip` increase is paying for the smallest possible fix.

atomiks (commented):
Follow-up on the simplification point: I think this can likely be collapsed further than just moving the logic out of `useHoverReferenceInteraction`.

Right now Tooltip is paying for two separate mechanisms for the same nested-trigger behavior:

- suppress a future delayed open when the pointer is already over an inner trigger
- close an already-open outer tooltip when the pointer moves onto that inner trigger

Those could probably be driven from one local `TooltipTrigger` signal such as “pointer is currently over a nested enabled tooltip trigger”, derived from bubbling `pointerover`/`mouseover` state. If that works, you should be able to drop most or all of:

- `lastMousePositionRef`
- `elementFromPoint()`
- the custom bubbling DOM event
- the layout-effect listener for that event
- the need to surface `data-tooltip-trigger` as public docs/API

So splitting the generic hover hook would stop other components from paying for this, but Tooltip would still carry most of the current cost unless the duplicated local machinery is collapsed too. My impression is that the real bundle win is not just “make it tooltip-only”, but “make it one local mechanism instead of two.”

I think the one caveat is that Tooltip still needs some private way to cancel a pending delayed open, so there may still be a small pre-open guard somewhere. But that should be materially smaller than the current `elementFromPoint()` + custom-event combination.

michaldudak:
I believe we have to resort to checking the `closest()` DOM element (with the help of the custom data attribute), as we can't rely on accessing the other Trigger's store from the outside. I improved how it behaves with shadow DOM.

I also simplified the implementation and now use just a single call in `onMouseOver` to detect a nested trigger.

As for `useHoverReferenceInteraction` - the `shouldOpen` is an optional callback with zero overhead when not provided (`shouldOpenRef.current?.() !== false` short-circuits), aside from a couple of extra bytes in the bundle (less than 30 gzipped for components that don't use it). I believe it's the minimal and most readable way to prevent pending delayed opens before they happen.

flaviendelangle (commented):
# PR Review: [tooltip] Fix flickering when nesting tooltip inside another tooltip's trigger

**PR:** https://github.com/mui/base-ui/pull/4501
**Author:** @michaldudak
**Base:** `master`

---

## Summary

This PR prevents the outer tooltip from opening when the cursor is over an enabled nested tooltip trigger. When the outer tooltip is already open and the pointer moves onto a nested trigger, it closes via the bubbling `mouseover` event so that only the innermost tooltip is visible. Disabled nested triggers are excluded from detection so the parent tooltip still opens. The close path is gated to hover-opened tooltips only -- focus-opened tooltips remain visible until blur.

**Files changed:** 10 (+651 / -23)

---

## Critical Issues (1)

### 1. Missing test: focus-opened tooltip should NOT be closed when hovering nested trigger

`handleNestedTriggerHover` explicitly checks `store.select('lastOpenChangeReason') === REASONS.triggerHover` before closing the parent. This means a focus-opened outer tooltip should remain open when the pointer moves onto a nested trigger. This is a critical behavioral distinction -- accidentally closing a keyboard user's tooltip would be an accessibility regression. **There is no test verifying this path.**

A test should: focus the outer trigger to open the outer tooltip, then hover the nested trigger, and assert the outer tooltip remains visible.

---

## Important Issues (4)

### 2. Missing test: rapid oscillation between nested trigger and parent area

The most common flickering scenario in production is rapid back-and-forth pointer movement between the nested trigger and the parent area. The existing tests cover one-directional movements but not the oscillating case. A test should simulate: hover nested trigger, move to outer area, quickly move back to nested trigger, and verify only the inner tooltip is shown.

### 3. PR description inaccuracy: "bubbling custom DOM event"

The PR body states "close it via a bubbling custom DOM event." The actual implementation uses the native `mouseover` event (which naturally bubbles) combined with `handleNestedTriggerHover` on `onMouseOver`. There is no custom DOM event (`new CustomEvent(...)` or `dispatchEvent(...)`) involved. The description should say something like "close it when the bubbling `mouseover` event reaches the outer trigger and detects a nested trigger target."

### 4. Dual state management paths for `isNestedTriggerHoveredRef`

The ref is updated from two different locations:
- `detectNestedTriggerHover` sets it on `onMouseOver`
- `onMouseLeave` unconditionally resets it to `false`

This creates two separate sources of truth. If the logic for what constitutes "nested trigger hovered" ever changes in `detectNestedTriggerHover`, the raw reset in `onMouseLeave` could become inconsistent. Additionally, if a pending open timer fires between the `mouseover` (which set the ref to `true`) and the `mouseleave` (which resets it to `false`), the tooltip open could be incorrectly suppressed, though this window is likely extremely small.

### 5. `shouldOpen` return-value contract is implicit

The `shouldOpenRef.current?.() !== false` pattern means any falsy-but-not-false value (`undefined`, `null`, `0`, `""`) is treated as "should open." If a future contributor modifies `shouldOpen` to return `undefined` as a signal for "don't open," this pattern silently allows opening. TypeScript constrains this at compile time, but at runtime a function that accidentally returns `undefined` (forgot the return statement) would silently allow opening instead of preventing it. A brief inline comment would aid maintainability.

---

## Suggestions (4)

### 6. Extract `checkShouldOpen()` helper

The pattern `shouldOpenRef.current?.() !== false` appears 4 times in `useHoverReferenceInteraction.ts`. Extracting a local helper:

```ts
function checkShouldOpen(): boolean {
  return shouldOpenRef.current?.() !== false;
}
```

would replace the somewhat cryptic double-negation with a readable call at each site.

### 7. Pass `getOpenDelay` directly as `restMs`

Instead of:
```ts
restMs() {
  return getOpenDelay();
},
```

Since `useStableCallback` already returns a stable function reference:
```ts
restMs: getOpenDelay,
```

### 8. Missing test: three levels of nesting

The implementation uses `closest()` to find the nearest enabled `[data-tooltip-trigger]` ancestor, then checks `nearestTrigger !== triggerEl && contains(triggerEl, nearestTrigger)`. With three levels (grandparent > parent > child), hovering the child should suppress both grandparent and parent tooltips. A basic smoke test would confirm the `closest()` traversal works correctly at multiple depths.

### 9. Missing test: `Tooltip.Provider` with nested tooltips

The experiment page (`nested.tsx`) demonstrates nested tooltips wrapped in `Tooltip.Provider`, but no test exercises this combination. The `getOpenDelay` function interacts with `providerContext?.delay` and the group delay logic. A test with `<Tooltip.Provider>` wrapping nested tooltips would verify the delay computation does not interfere with nesting suppression.

---

## Strengths

- **Solid architecture:** The `shouldOpen` callback is a clean, generic integration point in `useHoverReferenceInteraction` that doesn't couple the hook to tooltip-specific nesting knowledge. The three-piece chain (`detectNestedTriggerHover` + `isNestedTriggerHoveredRef` + `shouldOpen`) cleanly separates DOM detection from generic open-gating.

- **Shadow DOM support:** The new `closest()` utility correctly traverses shadow boundaries, and there is a dedicated test for it. This is the kind of environment-specific edge case that often gets missed.

- **Well-written tests:** 8 tests covering core scenarios with realistic event sequences, behavior-oriented assertions, and proper cleanup (shadow DOM test uses `finally` block). The "cancel pending open" test directly targets the flickering bug.

- **Good comments:** The zero-delay comment in `handleNestedTriggerHover` explains "why" rather than "what." The `shouldOpen` JSDoc is well-structured. The `closest()` JSDoc concisely explains the traversal strategy.

- **Correct handling of disabled nested triggers:** The `ENABLED_TOOLTIP_TRIGGER_SELECTOR` uses `:not([data-trigger-disabled])` so disabled nested triggers are transparent to the parent, allowing the parent tooltip to open normally.

- **`handleNestedTriggerHover` structure:** The enter/leave cases are well-paired in a single function with a clear early return. Splitting into separate handlers would duplicate detection logic for no benefit.

---

## Recommended Action

1. Add the focus-opened tooltip test (critical for accessibility)
2. Fix the PR description to accurately describe the mechanism
3. Consider adding the rapid oscillation test
4. Consider the minor code simplifications (items 6 and 7)

flaviendelangle (commented):
# PR Review: [tooltip] Fix flickering when nesting tooltip inside another tooltip's trigger

**PR:** https://github.com/mui/base-ui/pull/4501
**Author:** @michaldudak
**Branch:** `claude/condescending-clarke` → `master`
**Latest commit reviewed:** `00849f4` ("fixes, tests")
**Worktree:** `/tmp/pr4501-review`

This is a re-review that takes into account all prior reviews — specifically the comprehensive review by @flaviendelangle on commit `ad7d99f` and the earlier feedback from @atomiks. The bulk of those action items have been addressed; this review verifies that and surfaces what's still open.

---

## TL;DR

The branch is in good shape. The architecture has been significantly simplified since the earlier reviews (cursor-position sampling + custom DOM event collapsed into a single bubbling-`mouseover` + `shouldOpen` veto). All four prior-flagged test gaps from @flaviendelangle's review (focus-opened tooltip, oscillation, three-level nesting, Provider) are now covered. PR description was corrected.

**One genuine bug remains** (`isMouseLikePointerType(undefined)` returning `true`), plus a handful of low-severity hardening opportunities and one missing test (sibling nested triggers).

---

## Verification of previously-raised concerns

| Concern | Source | Status |
|---|---|---|
| `checkShouldOpen()` helper extracted | flaviendelangle #6 | ✅ [`useHoverReferenceInteraction.ts:101-104`](packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts#L101-L104) |
| `restMs: getOpenDelay` passed directly | flaviendelangle #7 | ✅ [`TooltipTrigger.tsx:152`](packages/react/src/tooltip/trigger/TooltipTrigger.tsx#L152) |
| Shadow-DOM-safe `closest()` added | atomiks | ✅ [`internals/shadowDom.ts:43-54`](packages/react/src/internals/shadowDom.ts#L43-L54), used at [`TooltipTrigger.tsx:124`](packages/react/src/tooltip/trigger/TooltipTrigger.tsx#L124) |
| Single-mechanism architecture | atomiks follow-up | ✅ Cursor sampling + custom event removed; bubbling `onMouseOver` + `shouldOpen` veto + targeted local re-open for the `move:false` case |
| `lastOpenChangeReason === REASONS.triggerHover` gate | flaviendelangle #1 | ✅ [`TooltipTrigger.tsx:183`](packages/react/src/tooltip/trigger/TooltipTrigger.tsx#L183) |
| Focus-opened-tooltip-not-closed test | flaviendelangle #1 | ✅ [`TooltipRoot.test.tsx:1516`](packages/react/src/tooltip/root/TooltipRoot.test.tsx#L1516) (+ controlled-open variant at 1561) |
| Rapid oscillation test | flaviendelangle #2 | ✅ [`TooltipRoot.test.tsx:1271`](packages/react/src/tooltip/root/TooltipRoot.test.tsx#L1271) |
| Three-level nesting test | flaviendelangle #8 | ✅ [`TooltipRoot.test.tsx:1014`](packages/react/src/tooltip/root/TooltipRoot.test.tsx#L1014) |
| Provider + nested tooltips test | flaviendelangle #9 | ✅ [`TooltipRoot.test.tsx:1155`](packages/react/src/tooltip/root/TooltipRoot.test.tsx#L1155), 1208, 1987 |
| Disabled nested trigger test | atomiks | ✅ [`TooltipRoot.test.tsx:1647`](packages/react/src/tooltip/root/TooltipRoot.test.tsx#L1647) |
| PR description accuracy | flaviendelangle #3 | ✅ Updated — describes bubbling `mouseover` mechanism, no false "custom DOM event" claim |
| `data-tooltip-trigger` made private with proper prefix | atomiks | ✅ Renamed to `data-base-ui-tooltip-trigger`, only referenced in `TooltipTrigger.tsx`. Matches established `data-base-ui-` convention |
| `shouldOpen` `!== false` contract documented | flaviendelangle #5 | ⚠️ Inline comment exists at [`useHoverReferenceInteraction.ts:102`](packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts#L102), but the JSDoc on the prop (lines 49–53) says only "Return `false` to veto" without spelling out that `undefined` and other non-`false` values open. See Important #2 below. |

---

## Critical Issues (1)

### C1. `isMouseLikePointerType(undefined)` returns `true`, defeating the touch gate in the local re-open path

[`TooltipTrigger.tsx:196`](packages/react/src/tooltip/trigger/TooltipTrigger.tsx#L196):

```ts
isMouseLikePointerType(pointerTypeRef.current)
```

`isMouseLikePointerType` is defined in [`floating-ui-react/utils/event.ts:46-54`](packages/react/src/floating-ui-react/utils/event.ts#L46-L54):

```ts
export function isMouseLikePointerType(pointerType: string | undefined, strict?: boolean) {
  const values: Array<string | undefined> = ['mouse', 'pen'];
  if (!strict) {
    values.push('', undefined);  // <-- undefined returns true in non-strict mode
  }
  return values.includes(pointerType);
}
```

In the local re-open path, the gate is meant to exclude touch (it sits next to `onMouseLeave` clearing `pointerTypeRef` to `undefined`, and the test at [`TooltipRoot.test.tsx:1838`](packages/react/src/tooltip/root/TooltipRoot.test.tsx#L1838) explicitly verifies that touch-then-mouse handoffs work via this clearing). But because `undefined` is treated as mouse-like in non-strict mode, an event sequence where `pointerTypeRef.current` is `undefined` at the moment of re-open passes the gate.

In the standard browser order (`pointerover` → `pointerenter` → `mouseover`), `pointerenter` will have populated `pointerTypeRef` before `mouseover` fires, so the gate is benign in the happy path. The risk lives at the seams:

- After `mouseLeave` → `clearNestedTriggerHover` resets `pointerTypeRef` to `undefined`. If a `mouseover` fires before the next `pointerEnter` (any timing/ordering quirk in nested or delegated trees), the local re-open is incorrectly allowed.
- Synthetic test events that dispatch `mouseover` without a preceding `pointerEnter`. Tests pass today because they always dispatch a `pointerEnter` first, but the gate is silently disabled for any future test that forgets that step.

**Fix:** pass `strict: true`:

```ts
isMouseLikePointerType(pointerTypeRef.current, true)
```

This matches the gate's intent — "open only if we have positive evidence the input is a mouse or pen." The non-strict default is correct in `useHoverInteractionSharedState` (where the absence of a recorded pointer type means "haven't seen anything yet, assume mouse"), but the wrong default here.

---

## Important Issues (3)

### I1. `getTarget()` non-Element results silently filtered out (shadow-DOM edge cases)

[`TooltipTrigger.tsx:119-135`](packages/react/src/tooltip/trigger/TooltipTrigger.tsx#L119-L135) and [`internals/shadowDom.ts:56-64`](packages/react/src/internals/shadowDom.ts#L56-L64):

`detectNestedTriggerHover` does:

```ts
if (triggerEl && isElement(target)) {
  const nearestTrigger = closest(target, ENABLED_TOOLTIP_TRIGGER_SELECTOR);
  ...
}
isNestedTriggerHoveredRef.current = nestedTriggerHovered;
```

If `target` is a `ShadowRoot` (rare but possible from `composedPath()[0]` when an event is dispatched on the shadow root itself) or `Document` (events on the document directly), `isElement(target)` is `false` and the function silently writes `nestedTriggerHovered = false`. The same chain in `handleNestedTriggerHover` (line 175) silently treats `targetInsideTrigger` as false.

In normal mouse/pointer hover scenarios `composedPath()[0]` is the deepest element under the cursor and `isElement` returns true, so this is a narrow edge case. It is worth flagging because:

1. The ergonomic shape of the bug — silent open of a tooltip that should be suppressed — matches the original flicker bug this PR fixes.
2. Floating UI primitives in this repo specifically aim to be shadow-DOM safe (per AGENTS.md), and this one path narrows that promise relative to adjacent code.

**Fix (defensive):** in `detectNestedTriggerHover`, walk the composed path to find the first Element if the direct target is not one. Or expose a `getTargetElement(event): Element | null` helper that wraps the path-walk and use it everywhere `getTarget` is followed by an `isElement` filter.

---

### I2. `shouldOpen` "veto" contract is fragile and underdocumented on the prop

[`useHoverReferenceInteraction.ts:49-53`](packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts#L49-L53):

```ts
/**
 * Called before hover opens the floating element. Return `false` to veto this
 * particular hover-open attempt.
 */
shouldOpen?: (() => boolean) | undefined;
```

The implementation at line 101–104 uses `!== false`, so any return value other than `false` (`undefined`, `null`, `0`, `""`, `true`) opens. A future caller writing:

```ts
shouldOpen: () => {
  if (someCondition) return; // ← forgot `return false`
}
```

silently opens when they meant to suppress.

The TypeScript signature `() => boolean` constrains compile-time, but a function whose body forgets a `return` still type-checks against `(): boolean | undefined` and is assignment-compatible in many configurations.

**Fix (one or both):**
- Update the JSDoc to spell out the contract: `Called before each hover-driven open attempt (immediate, delayed, and rest-ms). Return false to veto; any other return value, including undefined, permits the open.`
- In dev-mode only, warn if the callback returns a non-boolean: `if (process.env.NODE_ENV !== 'production' && result !== true && result !== false) console.warn(...)`. This is the same pattern used elsewhere in the repo for runtime contract enforcement.

The riskier alternative — flipping the convention to "return `true` to allow" — would change semantics for the existing tooltip caller and is not recommended.

---

### I3. Missing test: sibling nested triggers

[`TooltipRoot.test.tsx`](packages/react/src/tooltip/root/TooltipRoot.test.tsx) — the nested-tooltip block covers parent ↔ child transitions thoroughly but does not test child A ↔ child B transitions inside the same parent.

Sketch:

```ts
// outer trigger contains two nested Tooltip.Roots (innerA, innerB) as siblings
pointerEnter + mouseEnter + mouseOver on innerA;
clock.tick(OPEN_DELAY);
expect(innerA-popup).toBeVisible();
expect(outer-popup).toBeNull();

mouseOut(innerA, { relatedTarget: innerB });
pointerEnter + mouseOver on innerB;
clock.tick(OPEN_DELAY);
expect(innerB-popup).toBeVisible();
expect(innerA-popup).toBeNull();
expect(outer-popup).toBeNull();  // outer must NOT briefly open during the A→B handoff
```

**Why this matters:** the bubbling `mouseover` from `innerB` reaches the outer trigger with target=innerB. `detectNestedTriggerHover` correctly returns `true` (it's still over a nested trigger). But the close branch at line 180 only fires when `store.select('open')` is true — outer was never open, so it's a no-op. The local re-open branch needs `wasNestedTriggerHovered && !nestedTriggerHovered` — both are still nested, so it's also a no-op. Today this is correct, but it's a class of bug that this PR fixes and that could regress easily, e.g. if anyone adds an "outer was suppressed, allow it now" branch.

Severity: Important — covers a class symmetry that the existing oscillation test (parent ↔ same nested) doesn't.

---

## Suggestions (5)

### S1. Tighten the JSDoc on the `closest()` helper

[`internals/shadowDom.ts:43-54`](packages/react/src/internals/shadowDom.ts#L43-L54)

The current docstring describes the algorithm but doesn't document the null-input behavior. Since the helper accepts `Element | null` and returns `null` on null input (loop exits via `while (current)`), a future caller treating "null return" as "no match found" can be misled.

**Fix:** Add `Returns null when the element is null or no match is found in any tree.` Or narrow the input type to `Element` (force callers to handle null themselves).

---

### S2. Improve the `shouldOpen` JSDoc to enumerate which paths invoke it

[`useHoverReferenceInteraction.ts:49-53`](packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts#L49-L53)

The current docstring doesn't tell readers which open code paths invoke the callback. There are four: `shouldOpenImmediately` branch (line 268), delayed-open branch (line 280), immediate-open else branch (line 285), and rest-ms timer in `onMouseMove` (line 451). All currently invoke `checkShouldOpen()`. A reader who later adds a new open path may forget to gate it.

**Suggested copy:** `Called before each hover-driven open attempt (immediate, delayed, and rest-ms paths). Return false to veto; any other return value (including undefined) permits the open.`

---

### S3. Document the rationale for the local re-open path

[`TooltipTrigger.tsx:205-206`](packages/react/src/tooltip/trigger/TooltipTrigger.tsx#L205-L206) currently says:

> Tooltip passes `move: false` to the hover interaction, so moving off a nested trigger inside the same parent trigger needs this local reopen.

This explains the symptom but not the mechanism. With `move: false` the hook attaches only `mouseenter`/`mouseleave` on the parent trigger. Re-entering the parent trigger from a nested child fires no `mouseenter` because the cursor never left the parent — so the hook cannot observe the transition on its own.

**Suggested replacement:** `With move: false, the hook only listens to mouseenter/mouseleave on the parent trigger, so leaving a nested child back into the parent fires no event the hook can react to. Re-open locally here instead.`

---

### S4. Add a one-line note next to `pointerTypeRef`

[`TooltipTrigger.tsx:101`](packages/react/src/tooltip/trigger/TooltipTrigger.tsx#L101)

`pointerTypeRef` is intentionally a duplicate of the hook's internal `instance.pointerType`. The duplicate exists because the hook does not expose its tracked value, and the reset semantics differ — this ref clears on `mouseLeave` so a stale `'touch'` doesn't bleed into a later mouse interaction (covered by [`TooltipRoot.test.tsx:1838`](packages/react/src/tooltip/root/TooltipRoot.test.tsx#L1838)). A one-line comment explaining the duplication would prevent a future contributor from "deduplicating" it.

**Suggested:** `// Tracked locally so we can clear it on mouseLeave; the hook's internal pointer type never resets.`

---

### S5. Nice-to-have: leave-and-return-directly-to-nested test

After the close-on-nested branch fires, leaving the entire outer area and returning directly to the nested trigger (skipping the outer area on re-entry) should not schedule a stale outer reopen via `nestedTriggerOpenTimeout`. The existing test at line 1373 covers cancellation while the timeout is pending; this scenario covers state cleanliness across a full leave/re-enter cycle.

Sketch:

```ts
// Open outer via hover; close-on-nested fires; leave both
pointerEnter + mouseEnter on outer; tick(OPEN_DELAY); // outer open
mouseOver(outer, { target: nested });                // close-on-nested fires; outer closes
mouseLeave(outer, { relatedTarget: body });          // leave nested + outer

// Return directly to nested
pointerEnter + mouseOver on nested;
tick(OPEN_DELAY * 2);
expect(nested-popup).toBeVisible();
expect(outer-popup).toBeNull();
```

Severity: Nice-to-have.

---

## Strengths

- **Architecture simplified per atomiks's follow-up.** The previous dual mechanism (cursor-position sampling + custom bubbling DOM event) is gone. Now: bubbling `onMouseOver` populates `isNestedTriggerHoveredRef`, the hover hook consults it via `shouldOpen`, and a small local re-open path covers the `move: false` gap. Single source of truth.
- **Test coverage is thorough.** 22 nested-tooltip tests covering 3-level nesting, detached parent triggers, disabled nested triggers, focus-opened parents, controlled-open parents, shadow-DOM nested triggers, touch-pointer suppression, stale `pointerType` clearing, popup→nested-trigger backflow, safePolygon suppression, and Provider `delay={0}`. The shadow-root test at line 1686 uses a real `MouseEvent({ composed: true })` — exercises the new `closest()` helper, not just the selector.
- **Tests are behavior-oriented.** Assertions are on popup visibility and `onOpenChange`, not on internal refs. Resilient to refactors.
- **CLAUDE.md / AGENTS.md compliance:** `useTimeout`, `useStableCallback` for all event handlers, shadow-safe `contains`/`getTarget`/`closest` from `internals/shadowDom`, no `as any` casts, no global `document`/`window`. The only `as` casts are necessary ref widening or consistent with established `fastComponentRef` patterns elsewhere.
- **Self-match handled correctly in `closest()`.** Native `Element.closest()` already includes the element itself, and the shadow-host hop only runs after a null result — no double-match possible.
- **Selector ↔ data attribute consistency.** `TooltipTriggerDataAttributes.triggerDisabled = 'data-trigger-disabled'` matches the `:not([data-trigger-disabled])` selector — verified, no typo.
- **PR description now accurate.** The "bubbling `mouseover`" claim, the "Disabled nested triggers are excluded" claim, and the "hover-opened tooltips only" gate all match the implementation.

---

## Recommended Action

1. **Fix C1** (`isMouseLikePointerType(pointerTypeRef.current, true)`) — one-character change, real bug.
2. **Address I2** (`shouldOpen` JSDoc) and **I3** (sibling-trigger test) — both small, both prevent future regressions.
3. **Consider I1** (composed-path walk in `detectNestedTriggerHover`) if shadow-DOM parity is a hard guarantee for the floating layer.
4. Suggestions S1–S5 are polish — apply the ones that match your bar.

After C1 + I2 + I3 are in, this is mergeable.

---

## Notes on the review

- All four review agents (code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter) ran in parallel against worktree `/tmp/pr4501-review` at commit `00849f4`.
- The Critical finding (C1) was independently verified against the source of `isMouseLikePointerType` in `floating-ui-react/utils/event.ts:46-54`.
- Prior reviewers' resolved items were verified by reading the current source rather than trusting commit messages.

atomiks:
## Codex Review (GPT-5.5)
Reviewed the full PR branch against freshly fetched `upstream/master`; dominant type is `bug-fix` with secondary `tests`, `a11y`, and bundle-size impact. The hover-specific fix is well covered, but keyboard focus and delayed-open timing still leave branch-relevant regressions.

### 1. Bugs / Issues

#### 1. 🔴 Inner trigger focus can still open the outer tooltip (blocking)

`packages/react/src/tooltip/trigger/TooltipTrigger.tsx`

```tsx
const focusProps = useFocus(floatingRootContext, { enabled: !disabled }).reference;

// ...

props: [
  hoverProps,
  focusProps,
  rootTriggerProps,
  {
    onMouseOver(event: React.MouseEvent) {
      handleNestedTriggerHover(event.nativeEvent);
    },
```

The nested-trigger guard only covers hover. React `onFocus` bubbles, so focusing an enabled inner `Tooltip.Trigger` inside an outer trigger can run both focus handlers and open both ancestor and inner tooltips for keyboard users.

**Fix:** add the nested-trigger veto to focus opens as well, either through a `useFocus` `shouldOpen` guard or an internal `onFocus` handler that calls `event.preventBaseUIHandler()` when the focus target is an enabled descendant tooltip trigger. Add a regression test for focusing the uncontrolled inner trigger.

#### 2. 🟡 A stale parent hover timer can open too early after returning from the nested trigger (non-blocking)

`packages/react/src/floating-ui-react/hooks/useHoverReferenceInteraction.ts`

```ts
instance.openChangeTimeout.start(openDelay, () => {
  if (shouldOpen && checkShouldOpen()) {
    store.setOpen(true, createChangeEventDetails(REASONS.triggerHover, event, triggerNode));
  }
});
```

If the pointer starts over the outer area, moves onto the inner trigger before the outer delay expires, then moves back to the parent area before the original timer fires, `shouldOpen()` becomes true again and the old timer can open the parent earlier than a fresh parent-area delay.

**Fix:** cancel or invalidate pending hover/rest opens when nested hover starts, or tie delayed callbacks to an epoch. Add a test for outer hover -> half delay -> inner hover -> parent-area return, asserting the parent waits a full new delay.

#### 3. 🟡 Two regression tests do not prove the edge cases their names claim (non-blocking)

`packages/react/src/tooltip/root/TooltipRoot.test.tsx`

```tsx
// Disabled nested trigger case only hovers the outer trigger.
const outerTrigger = screen.getByTestId('outer-trigger');

fireEvent.pointerDown(outerTrigger, { pointerType: 'mouse' });
fireEvent.pointerEnter(outerTrigger, { clientX: 10, clientY: 10 });
```

The disabled nested-trigger test never hovers the disabled inner trigger, so it would pass even if disabled nested triggers were still treated as enabled.

```tsx
const innerShadowTarget = document.createElement('span');
innerShadowRoot.appendChild(innerShadowTarget);

innerShadowRoot.dispatchEvent(new MouseEvent('mouseover', { bubbles: true, composed: true }));
```

The shadow-root test dispatches from the `ShadowRoot`, not `innerShadowTarget`, so it does not really prove traversal from an element inside the shadow tree back to the host trigger.

**Fix:** hover the disabled inner trigger in the disabled case, and dispatch the composed shadow event from `innerShadowTarget`.

### Performance Considerations

I ran `pnpm size:snapshot` on the PR head and on the PR merge-base (`53c3d1e`) to isolate the branch delta.

| Entry | Parsed Δ | Gzip Δ |
|---|---:|---:|
| `@base-ui/react/tooltip` | +1,197 B | +449 B |
| `@base-ui/react` | +1,192 B | +481 B |
| `@base-ui/react/internals/shadowDom` | +220 B | +78 B |
| Shared hover users such as `menu`, `popover`, `preview-card` | +73 B each | +32-40 B each |

Best bundle-size reductions to try:

1. Only mark enabled triggers with `data-base-ui-tooltip-trigger`, then simplify the selector to `[data-base-ui-tooltip-trigger]`.
2. Replace the shared exported `closest()` helper with a tooltip-local composed-path/host scan that checks the marker directly; this should remove the `shadowDom` entrypoint growth and the extra re-export cost.
3. Compute `event.composedPath()` once in `TooltipTrigger` and drop the separate `getTarget` import for this path.

### Test Coverage Assessment

`pnpm test:jsdom TooltipRoot --no-watch` passed on the PR head. Coverage is broad for pointer hover, detached handles, disabled triggers, shadow DOM, and zero-delay providers, but it needs the focus regression and stale-delay regression above before this is safe to merge.

### Recommendation

Request changes 🔴

The pointer-hover fix is directionally solid, but the keyboard focus path can still produce nested ancestor tooltips, which is an accessibility regression. I’d also address the stale timer before merge because it preserves the old flicker/early-open class of bug in a fast pointer path.

packages/react/src/tooltip/trigger/TooltipTrigger.tsx
```diff
@@ -122,7 +123,7 @@ export const TooltipTrigger = fastComponentRef(function TooltipTrigger(
   // Local copy so it can be cleared on mouseLeave without resetting the hover hook's own pointerType.
   const pointerTypeRef = React.useRef<string | undefined>(undefined);
 
-  const getOpenDelay = useStableCallback(() => {
+  const getOpenDelay = () => {
```
atomiks: nit: use function declarations `function {name} {`

michaldudak: OK, but I'll leave `handleNestedTriggerHover` as an arrow function as TS complains `store` may be `undefined` otherwise and I'd have to introduce `!` or additional code to fix it.

atomiks (approved):

PR #4504, opened 2026-04-02
[all components] Reduce listener cleanup bundle size

atomiks:
This refactors repeated DOM listener setup and teardown into tiny shared utilities so the minifier can collapse the repeated patterns across listener-heavy hooks and components. The behavior stays the same while shrinking repeated cleanup code across the floating, menu, toast, number field, and shared utils layers.

## Changes

- Add small `addEventListener` and `mergeCleanups` utilities in `@base-ui/utils` with focused runtime and type-level tests.
- Tighten `addEventListener` typing so known DOM targets infer the correct event type and reject mismatched handlers.
- Replace repeated `addEventListener` and `removeEventListener` cleanup blocks across shared hooks and components.
- Reduce the main `@base-ui/react` bundle by 300B gzipped and 3.05KB parsed in the bundle bot snapshot.

### DIFF
--- a/packages/react/src/context-menu/trigger/ContextMenuTrigger.tsx
+++ b/packages/react/src/context-menu/trigger/ContextMenuTrigger.tsx
@@ -84,7 +84,7 @@ export const ContextMenuTrigger = React.forwardRef(function ContextMenuTrigger(
     addEventListener(
       doc,
       'mouseup',
-      (mouseEvent: MouseEvent) => {
+      (mouseEvent) => {
         allowMouseUpTriggerRef.current = false;
 
         if (!allowMouseUpRef.current) {
--- a/packages/react/src/utils/usePressAndHold.ts
+++ b/packages/react/src/utils/usePressAndHold.ts
@@ -134,7 +134,7 @@ export function usePressAndHold(params: UsePressAndHoldParameters): UsePressAndH
     addEventListener(
       win,
       'pointerup',
-      (event: PointerEvent) => {
+      (event) => {
         isPressedRef.current = false;
         stopAutoChange();
         onStop?.(event);


atomiks:
## Codex Review (GPT-5.4)

This is best classified as a `refactor` PR with secondary `perf` and `typescript` tags: it centralizes repeated listener setup/teardown into tiny shared helpers, tightens the helper typing, and then reapplies the branch cleanly on top of freshly fetched `upstream/master`. I reviewed the full `upstream/master...HEAD` range and reran `pnpm typescript`, `pnpm test:jsdom useDismiss --no-watch`, `pnpm test:jsdom addEventListener --no-watch`, `pnpm test:jsdom mergeCleanups --no-watch`, `pnpm test:jsdom SelectPopup --no-watch`, and `pnpm test:jsdom ScrollAreaScrollbar --no-watch`.

I did not find a concrete correctness regression in the rebased branch. The earlier review nits I would have called out here are now addressed: the helper naming matches behavior (`mergeCleanups`), `addEventListener` infers stricter event types including generic `Element`, the awkward once-listener pattern in `useDismiss` is consolidated into a local helper, and the `useScrollLock` test-teardown guard comment now matches the abstraction being called.

- `packages/utils/src/addEventListener.ts`: The helper now has public-facing JSDoc, stronger overloads, and type-level regression coverage in `addEventListener.spec.ts`, including invalid-handler cases.
- `packages/react/src/floating-ui-react/hooks/useDismiss.ts`: The repeated self-referential unsubscribe pattern is reduced to one local `addTargetEventListenerOnce()` helper, which is easier to audit than four inline copies.
- `packages/utils/src/useScrollLock.ts`: The resize cleanup guard is unchanged behaviorally, but the explanatory comment now accurately describes why the returned cleanup is guarded in tests.

The conversions still read as mechanical 1:1 swaps in the listener-heavy call sites I spot-checked, including `useHover`, `FloatingFocusManager`, `ToastViewport`, `SelectPopup`, and `ScrollAreaScrollbar`. The rebased conflict resolutions were also low risk: both conflicted files only needed import reconciliation, and the resulting listener registration logic matches the intended helper-based cleanup pattern.

### Public Type Surface Assessment

The strongest improvement since the first review is that `addEventListener` is no longer just a size helper; it now encodes useful type checking for known DOM targets while preserving a generic fallback overload for dynamic event names and plain `EventTarget` usage. That makes the public utils surface more defensible than the original version, because the new entrypoint now buys both bundle cleanup and better handler/type alignment.

### Test Coverage Assessment

The rerun coverage is proportionate to the risk in the final branch state: the new helper tests exercise the cleanup mechanics directly, `useDismiss` covers the most behavior-sensitive call site, and the rebased conflict touchpoints (`SelectPopup`, `ScrollAreaScrollbar`) both passed after rebasing onto current `master`. I did not rerun a fresh size snapshot locally in this pass, so I am not restating a new post-rebase benchmark number here.

### Recommendation

Approve ✅

I re-reviewed the rebased full branch rather than just the follow-up commit, and I do not see a remaining blocker. The branch is in a cleaner state than the earlier draft: the naming, typings, readability nits, and conflict resolutions all look settled, and the targeted validation matches the areas of real risk.

packages/utils/src/addEventListener.ts
```diff
@@ -0,0 +1,14 @@
+type EventTargetWithListeners = Pick<EventTarget, 'addEventListener' | 'removeEventListener'>;
+type Listener = EventListenerOrEventListenerObject | ((event: never) => void);
+
+export function addEventListener(
+  target: EventTargetWithListeners,
+  type: string,
+  listener: Listener,
+  options?: boolean | AddEventListenerOptions,
+) {
```
romgrk, line 58: Is it possible to improve the typings? Could `type` be more precise?

packages/utils/src/chain.ts
```diff
@@ -0,0 +1,12 @@
+type Cleanup = false | null | undefined | (() => void);
+
+export function chain(...cleanups: Cleanup[]) {
```
romgrk: The name of the function doesn't match super well what it does.

atomiks: Renamed to `mergeCleanups`

packages/utils/src/addEventListener.ts
atomiks: Updated types so it errors on invalid handler types

michaldudak:
# PR Review: [all components] Reduce listener cleanup bundle size

## Summary

This PR extracts two small utilities (`addEventListener` and `mergeCleanups`) into `@base-ui/utils` and uses them across ~25 files to replace the repetitive `addEventListener`/`removeEventListener` cleanup pattern. The result is a **-300B gzipped / -3KB parsed** bundle reduction with no behavior change.

## Strengths

- **Well-scoped mechanical refactor** — each replacement is a direct 1:1 swap, easy to verify
- **Strong typing** on `addEventListener` — known DOM targets infer the correct event type via overloads, and the fallback overload handles dynamic `event.type` / generic `EventTarget` use cases
- **Type-level tests** (`addEventListener.spec.ts`) verify inference and catch handler mismatches via `@ts-expect-error` — nice touch
- **`mergeCleanups` accepting falsy values** is ergonomic, enabling the clean `condition && addEventListener(...)` pattern

## Issues

### 1. `useScrollLock.ts` — guard around `unsubscribeResize()` is now misleading (Low)

`packages/utils/src/useScrollLock.ts:210-212`:

```ts
if (typeof win.removeEventListener === 'function') {
  unsubscribeResize();
}
```

The original code guarded `win.removeEventListener(...)` directly. Now the guard wraps `unsubscribeResize()`, which internally calls `win.removeEventListener(...)`. It still works correctly, but the comment on line 211 ("in which case `removeEventListener` wouldn't be available, so we check for it") now reads oddly since `removeEventListener` is hidden behind the abstraction. Consider updating the comment or simplifying the guard to just call `unsubscribeResize()` if this defensive check is no longer needed in tests.

### 2. `useDismiss.ts` — `let unsubscribe` + self-referencing closure pattern repeated 4 times (Low)

In `useDismiss.ts` (lines ~493, ~518, ~604, ~632), the same structure appears:

```ts
let unsubscribe: (() => void) | undefined;
function callback() {
  doStuff(event);
  unsubscribe?.();
}
unsubscribe = target ? addEventListener(target, event.type, callback) : undefined;
```

This is correct (closure captures the `let` binding by reference), but it's a bit awkward compared to the original `target?.removeEventListener(event.type, callback)`. Since the whole point is bundle savings, this trade-off is reasonable — but if this pattern keeps growing, a tiny helper like `addOnceEventListener` could encapsulate it more clearly. Not blocking.

### 3. Missing `describe` import in `addEventListener.test.ts` (Nit)

The test file uses `describe` and `it` without importing them. This likely works because Vitest globals are configured, but it's inconsistent with the `import { expect, vi } from 'vitest'` at the top. The `mergeCleanups.test.ts` has the same pattern — both are consistent with each other at least.

### 4. `KnownEventTarget` could include `Element` (Nit)

The union only lists `HTMLElement` and `SVGElement`. If a call site passes a generic `Element`, it falls to the untyped overload. This is fine for current usage but worth noting if the utility is meant to be broadly reusable.

## Verdict

**Looks good.** Clean, low-risk refactoring that achieves its stated goal. The two new utilities are small, well-tested, and well-typed. All call-site changes are mechanical and preserve existing behavior. The issues above are all low-severity / nits.

packages/utils/src/addEventListener.ts
michaldudak, line 1: Add a JSDoc comment to briefly explain what it does and why it's needed.

PR #4505, opened 2026-04-02
[internals] Move the exported internals to a new folder and remove private Calendar component

flaviendelangle:
Applied on external codebases :heavy_check_mark:


docs/src/app/(private)/experiments/dialog/nested.tsx
```diff
@@ -4,7 +4,7 @@ import clsx from 'clsx';
 import { Dialog } from '@base-ui/react/dialog';
 import { animated as springAnimated, useSpring, useSpringRef } from '@react-spring/web';
 import { SettingsMetadata, useExperimentSettings } from '../_components/SettingsPanel';
-import { useTransitionStatus } from '../../../../../../packages/react/src/utils/useTransitionStatus';
+import { useTransitionStatus } from '../../../../../../packages/react/src/internals/useTransitionStatus';
```
flaviendelangle: We could use the new internal import now.
But if it's used in an experiment, maybe we could consider moving it to `@base-ui/utils`

atomiks (commented):
## Codex Review (GPT-5.4)

This is primarily a refactor, but it also expands the published subpath surface under `./internals/*` and `./localization-provider`. The runtime move looks mechanically safe; the remaining risk is in public-contract and guardrail consistency around those new entrypoints.

### 1. Bugs / Issues

#### 1. 🟡 Temporal type augmentation still points at the old module (non-blocking)

In `packages/react/src/internals/temporal/temporal.ts`, the example for registering `TemporalSupportedObjectLookup` still tells consumers to augment `@base-ui/react/types`, but this PR moves the live augmentation target to `@base-ui/react/internals/temporal`.

From `packages/react/src/internals/temporal/temporal.ts` and `packages/react/src/temporal-adapter-date-fns/TemporalAdapterDateFns.ts`:

```ts
// packages/react/src/internals/temporal/temporal.ts
declare module '@base-ui/react/types' {
  interface TemporalSupportedObjectLookup {
    'date-fns': Date;
  }
}

// packages/react/src/temporal-adapter-date-fns/TemporalAdapterDateFns.ts
declare module '@base-ui/react/internals/temporal' {
  interface TemporalSupportedObjectLookup {
    'date-fns': Date;
  }
}
```

This matters because `./internals/temporal` is now a published export. Anyone following the example will augment the wrong module and silently fail to register their temporal object type.

**Fix:** Update the example in `packages/react/src/internals/temporal/temporal.ts` to match `@base-ui/react/internals/temporal`.

#### 2. ℹ️ The deep-import lint exemption is broader than the published export map (informational)

In `eslint.config.mjs`, the new regex exempts every `@base-ui/react/internals/...` import from the deep-import restriction, while `packages/react/package.json` only exports a fixed set of one-level `./internals/*` entrypoints.

From `eslint.config.mjs`:

```js
{
  regex: '@base-ui/react/(?!internals/).+/.+',
  message: OneLevelImportMessage,
}
```

This matters because imports like `@base-ui/react/internals/temporal/getDateManager` would now lint clean in-repo even though the published package would not resolve them. That weakens the guardrail this refactor is supposed to establish.

**Recommendation:** Narrow the exemption to exactly one level under `internals`, or add an explicit restriction for `@base-ui/react/internals/*/*`.

### 2. Behavior Preservation Assessment

I did not see evidence of meaningful runtime churn beyond the export-surface move. Most of the diff is mechanical import rewriting, and the targeted validation I ran passed: `pnpm typescript`, `pnpm test:jsdom index --no-watch`, and `pnpm --filter @base-ui/react build`.

### 3. Ownership / Architecture Assessment

This is no longer just an internal folder move: `packages/react/package.json` now publishes ownership/context primitives such as `./internals/field-root-context`, `./internals/field-register-control`, and `./internals/form-context`. That may be intentional, but it turns future cleanup of those boundaries into semver-governed work, so I would call that out explicitly rather than framing the PR as only a relocation.

The main coverage gap is export-surface validation. `packages/react/src/index.test.ts` now explicitly skips `./localization-provider` and every `./internals/*` subpath:

```ts
![
  '.',
  './utils',
  './temporal-adapter-luxon',
  './temporal-adapter-date-fns',
  './localization-provider',
  './types',
].includes(key) &&
!key.startsWith('./unstable-') &&
!key.startsWith('./internals/')
```

That means the new published entrypoints are not being smoke-tested through `@base-ui/react/...` specifiers for packaged resolution or declaration shape. I would add a small import smoke test for the exported `./internals/*` and `./localization-provider` subpaths before merging.

### Merge Confidence Score

Overall merge confidence is 3/5. The refactor itself looks low-risk and the targeted typecheck, test, and build validation passed, but the new public subpath surface still has one stale consumer-facing type example, one guardrail mismatch, and no direct smoke coverage for the newly published exports. I would fix finding 1 and tighten either the lint rule or the export-surface test coverage before merging.

flaviendelangle (commented):
I'm migrating all the other repos to use this PR in order to make sure everything is correctly migrated 👍

flaviendelangle:
@atomiks all points addressed

docs/src/app/(docs)/react/components/autocomplete/types.md
```diff
@@ -950,11 +950,8 @@ type ReturnValue = T[];
 
 ```typescript
 type AutocompleteFilter = {
-  /** Returns whether the item matches the query anywhere. */
```
michaldudak, line 952: Is this caused by the API extractor ignoring external packages?

michaldudak:
It seems like the whole Temporal* utils aren't used anywhere in this repo anymore. Can we remove them?

michaldudak:
### Self-referential imports inside internals
filter.ts:1 — import { stringifyAsLabel } from '../internals/resolveValueLabel' should be './resolveValueLabel'
resolveValueLabel.tsx:3 — import { serializeValue } from '../internals/serializeValue' should be './serializeValue'
These go up to src/ then back into internals/ — a leftover from the pre-move locations. They resolve correctly today but are inconsistent and fragile.

### Stray useTransitionStatus.d.ts alongside .ts source
useTransitionStatus.d.ts — A hand-written .d.ts file coexists with the .ts source. This is the only file in the package with this pattern. It could produce duplicate/conflicting declarations during the build. Likely a leftover that should be removed.

### Missed import path update in RadioRoot.tsx
RadioRoot.tsx:25 — Still imports from ../../utils/serializeValue while all other consumers have been updated to ../../internals/serializeValue. Works because utils/serializeValue.ts re-exports, but inconsistent with the PR's intent.

### Orphaned Calendar-specific reason constants 
reason-parts.ts:42-44 — dayPress, monthChange, and valuePropChange have zero consumers after Calendar removal. Should be cleaned up.

### ESLint regex patterns could be combined 
eslint.config.mjs:24-31 — The two separate regex entries enforce the same "no deeply nested imports" rule and could be merged into one pattern: `@base-ui/react/(?:(?!internals/).+|internals/.+)/.+`. Minor simplification.

docs/src/app/(docs)/react/components/autocomplete/types.md
flaviendelangle: I think yes

flaviendelangle:
> It seems like the whole Temporal* utils aren't used anywhere in this repo anymore. Can we remove them?

We are using them in the Scheduler.
Since the new repo is private (and not published) I can't use them from there.
I think we can move them to a standalone package in the future, but for now I think the simplest option is to keep them here.

michaldudak:
### 1. Missing React import in constants.ts

  [code-reviewer] packages/react/src/internals/constants.ts:34 — The constant `ownerVisuallyHidden` is typed as `React.CSSProperties` but the file has no `import * as React from 'react'`. This will fail `pnpm typescript`.

  Fix: Add import type * as React from 'react'; at the top of the file.

### 2. luxon missing from peerDependencies

  [code-reviewer] packages/react/src/internals/temporal-adapter-luxon/TemporalAdapterLuxon.ts:6 — The export ./internals/temporal-adapter-luxon runtime-imports from luxon, but luxon is only in devDependencies.
  Consumers would get a runtime error. Compare: date-fns is correctly in both devDependencies and peerDependencies. The file also has `// @ts-nocheck` and a `TODO` suggesting it's not production-ready.

### 3. Roundabout self-referencing imports in labelable-provider/

  [type-design] Three files inside packages/react/src/internals/labelable-provider/ import from ../../internals/useBaseUiId instead of ../useBaseUiId. Both resolve to the same file, but the longer path navigates up
   to src/ and back down, which is misleading:
  - LabelableProvider.tsx:7
  - useAriaLabelledBy.ts:4
  - useLabelableId.ts:8

  Suggestions (2 found)

### 4. Dead re-export shim: utils/serializeValue.ts

  [code-reviewer] packages/react/src/utils/serializeValue.ts — Contains only export { serializeValue } from '../internals/serializeValue'. Nothing in the codebase imports from this path anymore. Can be removed.

flaviendelangle:
> 2. luxon missing from peerDependencies

That one is expected
Fixing the others

atomiks:
## Codex Review (GPT-5.4)
This is a refactor-dominant PR with public internal export, docs, and test-surface implications. I reviewed the full branch against freshly fetched `upstream/master`; the main blockers are mergeability and one newly exported type surface.

### 1. Bugs / Issues

#### 1. 🔴 Branch currently conflicts with latest `master` (blocking)

A merge simulation against the freshly fetched base reports modify/delete conflicts in two docs demo files that this PR deletes while `master` has since modified them:

```text
CONFLICT (modify/delete): docs/src/app/(docs)/react/components/calendar/demos/hero/index.ts deleted in refs/remotes/upstream/pr/4505 and modified in upstream/master.
CONFLICT (modify/delete): docs/src/app/(docs)/react/utils/localization-provider/demos/hero/index.ts deleted in refs/remotes/upstream/pr/4505 and modified in upstream/master.
```

This means the branch cannot be merged cleanly until it is rebased or updated against the current base branch.

**Fix:** Rebase/update the PR on latest `master` and resolve the two modify/delete conflicts by preserving the intended Calendar/LocalizationProvider removal while incorporating any relevant docs infrastructure changes from `master`.

#### 2. 🔴 `temporal-adapter-luxon` emits declarations that import temporal types from the wrong module (blocking)

`packages/react/package.json` now exposes the Luxon adapter as a shipped internal subpath:

```json
"./internals/temporal-adapter-luxon": "./src/internals/temporal-adapter-luxon/index.ts"
```

But the moved adapter still imports temporal types from `../types`:

```ts
import {
  TemporalAdapterFormats,
  DateBuilderReturnType,
  TemporalTimezone,
  TemporalAdapter,
} from '../types';
```

After the move, `../types` resolves to `packages/react/src/internals/types.ts`, which does not export those temporal types. `pnpm typescript` passes because the file is `@ts-nocheck`, but the emitted declaration keeps the bad import, so consumers of `@base-ui/react/internals/temporal-adapter-luxon` get a broken type surface. The date-fns sibling correctly imports these from `../temporal`.

**Fix:** Change the Luxon adapter import to `../temporal`, preferably as a type import where possible.

#### 3. 🟡 Filter API docs lost method descriptions (non-blocking)

The shared filter extraction dropped the property JSDoc that used to document how `contains`, `startsWith`, and `endsWith` differ. The generated public API docs now show the methods without descriptions in `docs/src/app/(docs)/react/components/combobox/types.md` and `docs/src/app/(docs)/react/components/autocomplete/types.md`:

```ts
type ComboboxFilter = {
  contains: <Item>(item: Item, query: string, itemToString?: (item: Item) => string) => boolean;
  startsWith: <Item>(item: Item, query: string, itemToString?: (item: Item) => string) => boolean;
  endsWith: <Item>(item: Item, query: string, itemToString?: (item: Item) => string) => boolean;
};
```

**Fix:** Move the previous method comments onto `packages/react/src/internals/filter.ts`’s `Filter` interface and regenerate the API docs.

#### 4. 🟡 New exported internals only have resolution coverage (non-blocking)

`RequestQueue` and `TimeoutManager` are new exported internal subpaths, but the added coverage only verifies that internals resolve through the package export map. There are no focused tests for queue deduping/concurrency/status behavior or timeout replacement/cleanup.

**Recommendation:** Add small unit tests for `packages/react/src/internals/RequestQueue.ts` and `packages/react/src/internals/TimeoutManager.ts`, or keep them out of the exported internals surface until their behavior is covered.

### Test Coverage Assessment

I ran `pnpm test:jsdom index --no-watch`, `pnpm test:jsdom TemporalAdapterLuxon --no-watch`, and `pnpm typescript` on the PR head in an isolated worktree after installing dependencies. They passed, but they do not catch the emitted-declaration import problem above or the missing behavioral coverage for the two new internal utilities.

### Recommendation

Request changes 🔴

The branch needs a rebase/conflict resolution before it can merge, and the newly exported Luxon adapter declaration currently points at the wrong type module. The docs JSDoc regression and missing tests are smaller follow-ups, but the first two issues should be fixed before approval.

flaviendelangle:
Point 1 is already fixed
I'll fix the other ones :+1: 

1 => ok
2 => ok
3 => ok
4 => ok

docs/src/app/(docs)/react/components/autocomplete/types.md
flaviendelangle: Fixed
Just a mistake on my side

atomiks (approved):

PR #4486, opened 2026-03-30
[docs] Pre-scroll side nav before hydration

aarongarciah:
Follow-up to comment: https://github.com/mui/base-ui/pull/4441#issuecomment-4140510872

Adds a bit of complexity in order to remove the side nav scroll jump on hydration to scroll into view the active link on hard navigations.

## Summary
- Pre-scroll the docs left sidebar to the active page link during initial page load by injecting a lightweight pre-hydration script in the docs layout.
- Keep the existing client-side side-nav auto-scroll for route transitions, but skip the first post-hydration scroll when pre-hydration already handled the current pathname.
- Centralize side-nav scroll constants and pathname normalization in `docs/src/utils/sideNavScroll.ts` to keep server script and client logic aligned.

PR #4508, opened 2026-04-03
[all components] Use native scrollIntoView for composite keyboard navigation

atomiks:
This uses native `scrollIntoView` at event-time, and a custom helper for mount-time.

The tricky part with mount-time scrolling is deciding what container should scroll on mount. Native `.scrollIntoView()` scrolls all containers necessary to make the item appear in the viewport, but this is not desirable for mount-time, only event-time.

Mount-time reveal is still intentionally narrower than keyboard navigation. Previously Tabs + ScrollArea failed to scroll properly because the component assumes TabsList is the scrollable container which is not always true. This change keeps native event-time scrolling for those cases, while mount-time reveal stays bounded to the composite root so arbitrary ancestors or the page do not move on initial render. (Also, Radix doesn't do any mount-time scrolling.)

Furthermore, users might want pre-hydration scrolling which further complicates our internal logic. (See https://github.com/mui/base-ui/pull/4486)

Mount-time reveal for outer containers is now possible using `data-base-ui-composite-root`

Preview: https://deploy-preview-4508--base-ui.netlify.app/experiments/composite-scroll-ancestors

## Changes

- Use native `scrollIntoView({ block: 'nearest', inline: 'nearest' })` for shared composite keyboard navigation and pair the follow-up focus with `preventScroll: true`.
- Keep mount-time selected-item reveal only for local composite-root scrollers, so `Tabs.List` and `RadioGroup` preserve their old behavior without scrolling outer ancestors or the page.
- Delete the old custom `scrollIntoViewIfNeeded` helper instead of keeping two competing scroll behaviors around.
- Add regression tests for keyboard scrolling, local-root mount reveal, and the absence of native mount-time ancestor scrolling.
- Add an experiments page covering plain overflow containers, `ScrollArea`, nested scrollers, and a selected-tab-below-the-fold mount case for manual checking.

### DIFF
--- a/packages/react/src/composite/root/useCompositeRoot.ts
+++ b/packages/react/src/composite/root/useCompositeRoot.ts
@@ -4,7 +4,6 @@ import { isElementDisabled } from '@base-ui/utils/isElementDisabled';
 import { ownerWindow } from '@base-ui/utils/owner';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { useMergedRefs } from '@base-ui/utils/useMergedRefs';
-import type { TextDirection } from '../../direction-provider/DirectionContext';
 import {
   ALL_KEYS,
   ARROW_DOWN,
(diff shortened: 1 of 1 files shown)


atomiks:
Replacing with https://github.com/mui/base-ui/pull/4515

PR #4515, opened 2026-04-03
[scroll area][docs] Document mask fade composition

atomiks:
Updates the Scroll Area docs to show mask-based gradient fades on the viewport itself and adds a short Tabs composition example for reusing `<ScrollArea.Viewport>` through `render`.

I encountered a bug when composing `Tabs` with `ScrollArea` in a local project:

1. The selected tab didn’t scroll into view on mount.
2. Navigating with the arrow keys didn’t scroll items into view correctly using `'nearest'` logic.

The reason is that `Tabs.List` is the scroll container boundary, so it needs to be composed with `ScrollArea.Viewport`.

The `::after` / `::before` approach doesn’t work nicely with this composition method, but a mask is actually better anyway, so I changed the CSS.

https://deploy-preview-4515--base-ui.netlify.app/react/components/scroll-area#gradient-scroll-fade

## Changes

- Replace the pseudo-element fade docs and demo with a CSS mask-based approach.
- Add a `Combining with Tabs` section that shows `<Tabs.List>` rendering `<ScrollArea.Viewport>`.
- Keep the docs examples aligned with the final mask syntax used in the demo.


docs/src/app/(docs)/react/components/scroll-area/demos/scroll-fade/css-modules/index.module.css
```diff
@@ -8,39 +8,23 @@
...
+    to bottom,
+    transparent 0,
+    black min(var(--fade-size), var(--scroll-area-overflow-y-start)),
+    black calc(100% - min(var(--fade-size), var(--scroll-area-overflow-y-end, var(--fade-size)))),
+    transparent 100%
+  );
   height: 100%;
   border-radius: 0.375rem;
   background: var(--color-gray-50);
+  mask-image: var(--viewport-mask-image);
```
aarongarciah: nit: is there a benefit extracting this into a CSS var?

atomiks: No, Codex originally wrote one with `-webkit-mask-image` but the unprefixed version is available broadly since late 2023 (pretty much ~Baseline, which is 2.5 years afterwards) so I'll fix this.

PR #4517, opened 2026-04-03
[code-infra] Remove resolutions

Janpot:
These are harmful as they could hide dependency issues that our users would see.

* If you need them for e.g. docs, then scope them to the docs
* If you need a temporary fix for a specific package, you may want to use `packageExtensions` or patched dependency. This fix must go upstream so that our users benefit from it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,16 +110,5 @@
   "engines": {
     "pnpm": "10.33.0",
     "node": ">=22.18.0"
-  },
-  "resolutions": {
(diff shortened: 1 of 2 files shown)


Janpot:
Propagated to:
- [ ] https://github.com/mui/base-ui-charts/pull/186
- [ ] https://github.com/mui/base-ui-mosaic/pull/143
- [ ] https://github.com/mui/base-ui-plus/pull/10

PR #4518, opened 2026-04-03
[docs-infra] Remove Tailwind vars and unify website and docs stylesheets

aarongarciah:
Preview: https://deploy-preview-4518--base-ui.netlify.app/

- Removes most of Tailwind CSS vars instances from the docs chrome and replaces them with our new CSS vars.
  - Blue is kept for outlines and code for now.
  - Syntax highlighting colors are untouched.
- Loads the main stylesheet in (website) as well.
  - Fixes issues derived from loading the main stylesheet in (website).
  - Removes unused CSS from (website).

### DIFF
--- a/docs/src/app/(website)/css/components/Text.css
+++ b/docs/src/app/(website)/css/components/Text.css
@@ -28,8 +28,8 @@
   }
 
   .Text.sz-3 {
-    font-size: 36px;
-    line-height: 38px;
+    font-size: var(--font-size-34);
+    line-height: 34px;
     text-wrap: balance;
     font-family:
       'die grotesk b',
@@ -67,8 +67,8 @@
     }
 
     .Text.bp2\:sz-3 {
-      font-size: 36px;
-      line-height: 38px;
+      font-size: var(--font-size-34);
+      line-height: 34px;
       text-wrap: balance;
       font-family:
         'die grotesk b',
--- a/docs/src/app/(private)/docs-theme/page.module.css
+++ b/docs/src/app/(private)/docs-theme/page.module.css
@@ -246,6 +246,6 @@
   }
 
   .hero h1 {
-    font-size: var(--font-size-36);
+    font-size: var(--font-size-34);
   }
 }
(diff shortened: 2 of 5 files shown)


docs/src/app/(website)/css/components/Text.css
```diff
@@ -17,17 +17,17 @@
     text-wrap: pretty;
   }
 
-  .Text.size-1 {
+  .Text.Text-size-1 {
```
aarongarciah: `.size-*` is a Tailwind class name so we can't use it.

docs/src/app/(website)/layout.tsx
```diff
@@ -1,9 +1,12 @@
+// Keep CSS imports first to ensure layers are declared before component styles
+import 'docs/src/css/index.css';
+import './css/index.css';
```
aarongarciah, line 3: When CSS imports are placed after component imports like `import { Link } from 'docs/src/components/Link'` below, the layer order gets screwed because component CSS files like `Link.css` contain `@layer components`.

docs/src/app/(website)/css/components/Button.css
aarongarciah, line 1: These and other styles were unused.

docs/src/app/(website)/css/components/Accordion.css
```diff
@@ -1,18 +1,18 @@
 @layer components {
-  .AccordionRoot {
+  .AccordionWebsiteRoot {
```
aarongarciah, line 2: `.Accordion*` class names were clashing with the Accordion component styles from (docs). If we ever unify all accordions into one style, we'll rename this ofc.

docs/src/app/(website)/css/components/Text.css
colmtuite: `sz-*`?

aarongarciah: Updated

colmtuite (approved):

PR #4522, opened 2026-04-05
[popover][menu] De-duplicate focus guard handlers for popover and menu triggers

sai6855:
This PR refactors the focus guard logic used in MenuTrigger and PopoverTrigger by extracting it into a shared hook (useTriggerFocusGuards) to reduce duplication and improve maintainability.


sai6855:
@atomiks added tests recommended by codex

PR #4526, opened 2026-04-06
[all components] Sync hover open event state

atomiks:
This fixes stale hover logic so click-like interactions do not block later hover reopen behavior.

## Changes

- Fix hover reopen during a hover close transition, including same-trigger re-entry.
- Move shared root-side open-change work into `FloatingRootStore`.
- Reuse that shared logic from Menu, Popover, Dialog, Tooltip, and Preview Card.
- Simplify the synced-root mode by replacing the old flag with a smaller `syncOnly` path.
- Add regressions for menu, popover, and dialog around the affected flows.

### DIFF
--- a/packages/react/src/preview-card/root/PreviewCardRoot.test.tsx
+++ b/packages/react/src/preview-card/root/PreviewCardRoot.test.tsx
@@ -330,6 +330,40 @@ describe('<PreviewCard.Root />', () => {
       });
     });
 
+    describe('dismissal', () => {
+      clock.withFakeTimers();
+
+      it('reopens on hover after Escape closes it', async () => {
+        await render(<TestPreviewCard triggerProps={{ delay: 100 }} />);
+
+        const trigger = screen.getByRole('link', { name: 'Link' });
+
+        fireEvent.pointerDown(trigger, { pointerType: 'mouse' });
+        fireEvent.mouseEnter(trigger);
+        fireEvent.mouseMove(trigger);
+
+        clock.tick(100);
+        await flushMicrotasks();
+
+        expect(screen.getByText('Content')).not.toBe(null);
+
+        fireEvent.keyDown(document.body, { key: 'Escape' });
+        await flushMicrotasks();
+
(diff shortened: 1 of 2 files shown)


atomiks:
## Codex Review (GPT-5.4)

Reviewed the full updated branch against a freshly fetched `upstream/master`. Dominant type: `bug-fix` with `tests` and `react-ui` secondary scope; the root-owned `openchange` bookkeeping now looks behaviorally consistent across the touched synced-root consumers, and the remaining coverage gap I previously called out has been closed.

### 1. Bugs / Issues (None)

The core fix still looks sound on the updated branch. Moving `openEvent` syncing and internal `openchange` emission behind `dispatchOpenChange(...)` fixes the old synced-root ordering problem while keeping cancellation store-owned, and the `useHoverReferenceInteraction` change correctly lets the close-transition fast path run before the rest-only early return.

The touched consumers remain aligned after the refactor. Menu, Popover, Dialog, Tooltip, and Preview Card all route root-side side effects through the same `dispatchOpenChange(...)` path, and I did not find a touched component still carrying divergent `openchange` or `openEvent` bookkeeping.

### Test Coverage Assessment

The branch now has proportional regression coverage for the shared bookkeeping change. In addition to the existing Menu, Popover, and Dialog regressions, the latest update adds targeted reopen coverage for Tooltip and Preview Card, which closes the main branch-wide risk I had left open.

I also re-ran the affected jsdom suites on the updated branch:
`pnpm test:jsdom MenuRoot --no-watch`
`pnpm test:jsdom PopoverRoot --no-watch`
`pnpm test:jsdom DialogRoot --no-watch`
`pnpm test:jsdom TooltipRoot --no-watch`
`pnpm test:jsdom PreviewCardRoot --no-watch`

### Recommendation

Approve ✅

I do not see any remaining actionable issues in the implementation or its updated regression coverage. The bookkeeping ownership is cleaner than before, the touched components are consistent with each other, and the added tests cover the previously under-validated consumers.

PR #4527, opened 2026-04-06
[popups] Minify backdrop cutout polygon

atomiks:
This keeps internal backdrop cutouts using a compact `polygon(...)` clip-path string instead of a multiline template literal. The rendered geometry stays the same, but the serialized inline style now minifies cleanly.

## Changes

- Serialize the internal backdrop cutout polygon as a single compact string.
- Add a regression test that asserts the rendered `clip-path` style stays minified.


flaviendelangle:
Obviously we shouldn't apply the `join` suggestion which adds runtime cost :+1:

PR #4528, opened 2026-04-06
[select] Remove custom hover focus handlers

atomiks:
Updated Select to rely on `useListNavigation` for hover-driven highlighting instead of maintaining a separate custom handler path.

## Changes

- Turn `focusItemOnHover` back on in `SelectRoot` so Floating UI handles hover focus again.
- Remove the redundant custom hover and focus handlers from `Select.Item` and `Select.Popup`.
- Drop the now-unused `highlightItemOnHover` context field.

## Benchmarks

Chromium (Incognito) `/select-perf` (1,000 items) on 6x CPU slowdown

| Metric | `master` | PR | Delta |
|---|---:|---:|---:|
| Window duration | 2925.2 ms | 2977.3 ms | +1.8% |
| `mousemove` total | 488.7 ms | 473.9 ms | -3.0% |
| `mousemove` avg | 4.89 ms | 4.74 ms | -3.0% |
| `mousemove` p50 | 0.028 ms | 0.029 ms | basically flat |
| `mousemove` p95 | 21.60 ms | 23.56 ms | +9.1% |
| `mousemove` max | 26.01 ms | 25.77 ms | -0.9% |
| `>16.7 ms` mousemoves | 23 | 16 | -30.4% |
| `>25 ms` mousemoves | 2 | 1 | -50.0% |
| `UpdateLayoutTree` total | 161.4 ms | 161.4 ms | flat |
| `Layout` total | 92.4 ms | 88.3 ms | -4.5% |
| `Paint` total | 75.9 ms | 59.6 ms | -21.4% |


atomiks:
## Codex Review (GPT-5.4)

This PR is best classified as `bug-fix` with secondary tags `perf`, `tests`, and `react-ui`. I reviewed the full branch against freshly fetched `upstream/master`; the current patch both simplifies Select’s hover-highlight ownership and closes the remaining aligned-open highlight gap without reintroducing the old custom handler path.

The core simplification still looks sound: `Select` now relies on `useListNavigation` for hover-driven highlight sync again, and the Safari bounds workaround was preserved by moving it into the shared leave path instead of keeping a Select-only copy.

The follow-up fix in `SelectPopup` also addresses the specific regression reported after that simplification. When `alignItemWithTrigger` is active and there is no selected item, the popup can open under a stationary cursor without any subsequent hover event. The new open-time fallback sets the first item active after placement, which matches the behavior users were expecting on both Chromium and Safari.

`highlightItemOnHover` remains respected for hover behavior rather than being promoted into a blanket “never highlight” switch. The branch still allows highlight to be driven by non-hover paths such as keyboard navigation and typeahead, which is consistent with the existing Select semantics and tests.

### Test Coverage Assessment

The branch now has focused regression coverage for both pieces of risk introduced during the refactor:

- the shared Safari leave-path guard in `useListNavigation`
- the aligned-open, no-selected-value case in `SelectRoot`

I also ran:

- `pnpm test:chromium SelectRoot --no-watch`
- `pnpm typescript`

That is good coverage for the current surface area: one shared interaction hook regression test plus browser-backed validation of the Select path that actually regressed.

### Recommendation

Approve ✅

I did not find any actionable correctness or regression issues in the current branch. The final shape is smaller than the pre-refactor Select-specific implementation, keeps the Safari bounds protection, and covers the aligned-open edge case with focused browser-backed regression tests.

PR #4531, opened 2026-04-06
[select] Fix RTL aligned popup positioning

atomiks:
This fixes `Select.Positioner` with `alignItemWithTrigger` in RTL by aligning the popup against the selected value's logical inline edge instead of always using physical left-edge math. This caused the select to not be aligned perfectly on the x-axis.

https://deploy-preview-4531--base-ui.netlify.app/experiments/select-rtl-align-item-with-trigger

## Changes

- Use the current text direction when computing the aligned popup offset in `Select.Popup`.
- Preserve the existing viewport clamping behavior after the logical-edge alignment step.
- Add real-browser regression coverage for both LTR and RTL selected-item alignment.


atomiks:
## Codex Review (GPT-5.4)

Reviewed the latest PR head `da8ddd4` against a freshly fetched `upstream/master`. Dominant type: `bug-fix` with `tests` and `react-ui` secondary tags. This is a follow-up pass after Claude's review, so I focused on the geometry and regression gaps raised there rather than rehashing the original RTL fix.

### 1. Bugs / Issues (None)

The RTL inline-edge fix still looks correct on the latest head. The measurement path waits for Floating UI's first positioning pass, keeps the selected item text available on first open, and preserves the existing viewport clamp / fallback model.

Claude's remaining notes about the clamp behavior change and the `flushSync` removal still look non-blocking to me on the current branch. The clamp change is consistent with the newer logical-edge positioning, and the fallback state update still runs from layout effect before paint.

### Test Coverage Assessment

Claude's test-gap points were useful, and the latest commit addresses the actionable ones:

- added a real-browser regression case where popup-level padding and border shift the selected text in LTR
- added a no-value RTL case to cover first-open alignment when the first item becomes the measurement target
- removed the stray async flush from the RTL regression and cleaned up the Y-offset naming / comments

I also re-ran:

- `pnpm prettier`
- `pnpm test:jsdom SelectPopup --no-watch`
- `pnpm test:chromium SelectPopup --no-watch`

All passed locally.

### Branch Status

The latest PR head is `da8ddd4`, and the GitHub checks for that commit are still pending at the time of this comment (CircleCI matrix plus Netlify).

### Recommendation

Approve ✅

I don't see a remaining actionable code-review issue on the latest head. Claude's main testing concerns are now covered by the follow-up commit, and the branch looks ready pending CI on `da8ddd4`.

PR #4532, opened 2026-04-06
[tabs] Fix activation direction on first render

atomiks:
This fixes the activation-direction bug on the first render after a tab selection change and stabilizes the Chromium coverage that was catching it. `Tabs.Root` now derives the effective activation direction from committed state plus the current value, then syncs that snapshot after commit, which keeps the first render aligned without relying on mutable ref reads in render.

## Changes

- Replace the render-time ref-driven activation-direction bookkeeping in `Tabs.Root` with state-based previous-value tracking.
- Keep the strict first-render activation-direction assertions for direct tab switches.
- Settle frame-based panel transition work between the two click assertions so delayed renders from the first interaction do not leak into the second assertion.
- Narrow the dynamic add-and-select numeric test to the first render of the newly selected panel.
- Preserve the existing final-render assertion for the out-of-order string-value dynamic case.

### DIFF
--- a/packages/react/src/combobox/positioner/ComboboxPositioner.test.tsx
+++ b/packages/react/src/combobox/positioner/ComboboxPositioner.test.tsx
@@ -1,11 +1,67 @@
-import { expect } from 'vitest';
+import { expect, vi } from 'vitest';
 import * as React from 'react';
+import * as ReactDOMClient from 'react-dom/client';
 import { waitFor } from '@mui/internal-test-utils';
 import { Combobox } from '@base-ui/react/combobox';
 import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
 
 describe('<Combobox.Positioner />', () => {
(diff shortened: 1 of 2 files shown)


atomiks:
@LukasTy yes, randomly on various PRs, but only on React 18 in Chromium

atomiks:
Updated this PR after review feedback.

The final change keeps the strict first-render activation-direction assertions for the direct tab-switch cases, but fixes the underlying `Tabs.Root` logic instead of weakening the tests. `Tabs.Root` now derives the effective activation direction from committed state plus the current value during render, then syncs that snapshot after commit, which avoids reading mutable refs in render and keeps the first render after a selection change in sync. I also tightened the inter-click test cleanup to wait for frame-based panel transition work before clearing mocks, and narrowed the dynamic add-and-select numeric test to the first render of the newly selected panel.

PR #4533, opened 2026-04-06
[toast] Add focus option to move focus to toast on open

### LINKED ISSUE #4253: [toast] accessibility: Undo action can't be reached in time with a keyboard or screen reader, or its timing adjusted
# Bug report

## Current behavior

The "Undo action" Toast example triggers an important Toast with an "undo" button inside it. **Keyboard and screen reader users can't undo the action because focus doesn't move to the toast notification**. 

This should be high priority to fix as it potentially blocks keyboard and screen reader users from being able to complete important actions. 

Potential WCAG failures:
* See [Understanding WCAG 2.2 Success Criterion 2.1.1 Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
* See [Understanding WCAG 2.2 Success Criterion 2.2.1 Timing Adjustable](https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable.html)

## Expected behavior

Keyboard focus either needs to move to these Toast notifications or the toast needs to be the next tab stop, and the toast should persist until dismissed or focus moves back out of them.

## Reproducible example

To replicate

1. Navigate to [Toast: Undo action example](https://base-ui.com/react/components/toast#undo-action)
2. Using only a keyboard, activate the "perform action" button
3. Try to tab to the toast and activate the "Undo" button
4. Note the toast has been loaded at the end of the DOM and disappears before it can be easily reached 
5. Repeat the test using a screen reader with keyboard

## Base UI version

v1.2.0

Chrome

## Which OS are you using?

Mac OS

Keyboard, VoiceOver

## Additional context

In this case, moving focus to the toast and back to the originating control on close (or near to it, if the originating control no longer exists) would be a good solution. The toast should remain until it has been intentionally dismissed.

Jihye2ee:
## Summary

Adds a `focus` option to `toastManager.add()` that moves focus to the toast when it opens. This enables keyboard and screen reader users to reach interactive actions (e.g. Undo) without relying on the F6 shortcut.

Closes #4253

## Changes

- **`useToastManager.ts`** — Add `focus?: boolean` to `ToastObject`
- **`ToastRoot.tsx`** — On mount, if `focus` is `true`: save `prevFocusElement`, focus the toast, pause timers, and set `focused` state (same pattern as F6 in `ToastViewport`)
- **`ToastRoot.test.tsx`** — Add tests for focus behavior
- **Undo demo** — Apply `focus: true` to demonstrate the pattern

## How it works

```
1. toastManager.add({ focus: true, ... })
2. Toast mounts → useEffect saves prevFocusElement, focuses the toast
3. Viewport's handleFocus pauses timers (existing infra)
4. User interacts (Undo/Close) or tabs away → timers resume (existing infra)
5. On close → handleFocusManagement restores focus (existing infra)
```

When `focus` is not set, behavior is identical to before — no breaking changes.

## Screenshot
| Tab → Undo → Click | Tab → Undo → Tab out → auto-dismiss |
|:---:|:---:|
|  |  |

### DIFF
--- a/docs/src/app/(docs)/react/components/toast/demos/undo/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/toast/demos/undo/css-modules/index.tsx
@@ -24,6 +24,7 @@ function Form() {
       title: 'Action performed',
       description: 'You can undo this action.',
       type: 'success',
+      focus: true,
       actionProps: {
         children: 'Undo',
         onClick() {
--- a/docs/src/app/(docs)/react/components/toast/types.md
+++ b/docs/src/app/(docs)/react/components/toast/types.md
@@ -106,6 +106,13 @@ type ToastRootToastObject<Data extends {} = any> = {
   onClose?: () => void;
   /** Callback function to be called when the toast is removed from the list after any animations are complete when closed. */
   onRemove?: () => void;
+  /**
+   * Whether to move focus to the toast when it opens.
+   * When `true`, the previously focused element is saved and restored when the toast closes.
+   * Timers are paused while the toast has focus.
+   * @default false
+   */
+  focus?: boolean;
   /** The props for the action button. */
   actionProps?: Omit<
     React.DetailedHTMLProps<React.ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement>,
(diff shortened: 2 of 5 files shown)


aarongarciah:
Hey @Jihye2ee, thanks for the PR. Unfortunately, we won't be moving forward with it. Toasts are used to notify users without interrupting their workflow, and moving the focus to a toast is not desired.

In scenarios when users' input is required, [AlertDialog](https://base-ui.com/react/components/alert-dialog) is better suited.

A toast is close to the Alert pattern from the [APG guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/alert/):

> Because alerts are intended to provide important and potentially time-sensitive information without interfering with the user's ability to continue working, _it is crucial they do not affect keyboard focus_. The [Alert Dialog Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/) is designed for situations where interrupting work flow is necessary.

PR #4535, opened 2026-04-06
[checkbox group] Fix uncontrolled default initialization

atomiks:
An uncontrolled `CheckboxGroup` without `value` or `defaultValue` should behave like an empty selected list, but it could fall back to per-checkbox local state instead of maintaining a shared group value. This initializes the uncontrolled group value to `[]` so the checkbox group state behaves consistently from the first render.

## Changes

- Initialize uncontrolled `CheckboxGroup` state to an empty array when no `defaultValue` is provided.
- Preserve the existing checkbox-group update flow while making the uncontrolled state concrete.
- Add a regression test covering the omitted-`defaultValue` checkbox-group case.
- Tighten the checkbox-group state typing so type inference matches the runtime fix.


packages/react/src/checkbox-group/CheckboxGroup.tsx
```diff
@@ -57,7 +57,15 @@ export const CheckboxGroup = React.forwardRef(function CheckboxGroup(
 
   const disabled = fieldDisabled || disabledProp;
 
-  const [value, setValueUnwrapped] = useControlled({
+  const defaultValue = React.useMemo<string[] | undefined>(() => {
```
flaviendelangle, line 60: Would that work?


```suggestion
  const defaultValue = externalValue === undefined
    ? (defaultValueProp ?? EMPTY_ARRAY)
    : undefined
```

atomiks: It's used as a dependency, so it has to be wrapped in memo (`EMPTY_ARRAY` is not really safe to use here)

atomiks:
## Codex Review (GPT-5.4)

I reviewed the full branch diff against a freshly fetched `upstream/master`. This is a `bug-fix` PR with secondary `tests` and `public-api` implications, and the current branch looks consistent with the reported checkbox-group failure mode.

### 1. Bugs / Issues (None)

The main fix in `packages/react/src/checkbox-group/CheckboxGroup.tsx` addresses the actual root cause rather than only the symptom: uncontrolled `CheckboxGroup` no longer starts from runtime `undefined` when `defaultValue` is omitted, so the group path now initializes as an empty array and child checkboxes stay attached to shared group state.

The follow-up change in `packages/react/src/checkbox/root/CheckboxRoot.tsx` also resolves the regression risk introduced by routing parent-group updates through the resolved group state. With `!isGroupedWithParent` added to the generic `setGroupValue(...)` branch, parent-enabled groups now have a single update owner instead of two competing mutation paths.

The final shape is consistent with the rest of the controlled/uncontrolled patterns in this codebase: initialize the uncontrolled source once in `CheckboxGroup`, then let children derive their checked state from group context instead of falling back to per-checkbox local state.

I also checked the nearby parent-checkbox flow in `packages/react/src/checkbox-group/useCheckboxGroupParent.ts`. The PR does not widen that hook’s contract; it just ensures the hook now receives the resolved group value in uncontrolled mode, which is the missing piece for the existing logic to work as intended.

### Test Coverage Assessment

Coverage is proportionate to the risk. The new regression in `packages/react/src/checkbox-group/CheckboxGroup.test.tsx` covers the originally broken plain uncontrolled case with omitted `defaultValue`, and the added case in `packages/react/src/checkbox-group/useCheckboxGroupParent.test.tsx` covers the adjacent uncontrolled `allValues` / parent-checkbox path that the first revision had put at risk.

I also ran `pnpm typescript` and `pnpm test:jsdom CheckboxGroup --no-watch`, both of which passed on the current branch.

### Recommendation

Approve ✅

I do not see any actionable issues remaining in the current branch. The patch now fixes the uncontrolled initialization bug at the correct layer, preserves the parent-checkbox ownership model, and includes targeted regression coverage for both affected paths.

PR #4540, opened 2026-04-06
[docs-infra] Support skipping headings in table of contents

aarongarciah:
Adds the possibility of skipping headings in the table of contents by adding a MDX comment. Also adds the relevant comment to all headings that we want to hide.

Example from Dialog:

|before|after|
|---|---|
|Screenshot 2026-04-06 at 18 39 36|Screenshot 2026-04-06 at 18 39 20|

### DIFF
--- a/docs/src/app/(docs)/react/components/alert-dialog/page.mdx
+++ b/docs/src/app/(docs)/react/components/alert-dialog/page.mdx
@@ -221,7 +221,7 @@ import { TypesAlertDialog } from './types';
 
 <TypesAlertDialog.createHandle />
 
-[//]: # 'quick-nav-exclude'
+[//]: # '@exclude-table-of-contents'
 
 ### Handle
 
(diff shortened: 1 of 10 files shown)


docs/src/app/(docs)/react/components/alert-dialog/page.mdx
```diff
@@ -221,6 +221,8 @@ import { TypesAlertDialog } from './types';
 
 <TypesAlertDialog.createHandle />
 
+[//]: # 'quick-nav-exclude'
```
dav-is: What do you think of this? I think we should prefix with a `@` to match the `// @highligth-start` format we have for highlighting

```suggestion
[//]: # '@unlisted-heading'
```

or 

```suggestion
[//]: # '@exclude-table-of-contents'
```

The "quick nav" wording confuses me a bit, I know that's what the rehype plugin calls it

aarongarciah: Updated

colmtuite:
Why would we want to skip a heading?

aarongarciah:
The PR description has an example. The page has a "Handle" heading, which is a type, in this case, what's returned by `createHandle`. IMO hiding "Handle" in the TOC removes noise and reduces potential confusion.

PR #4544, opened 2026-04-07
fix(popover): prevent scroll on initial focus

### LINKED ISSUE #4520: [popover] Scrolls page to top when opened (portal + focus interaction)
## Description

When a `Popover` is opened, the page scrolls to the top. This happens because the popover content is portaled to the end of `<body>` via `PopoverPrimitive.Portal`, and `FloatingFocusManager` calls `focus()` on the popup element without `{ preventScroll: true }`. The browser's default focus behavior scrolls the focused element into view, jumping to the portaled content at the bottom of the DOM.

## Steps to Reproduce

1. Create a page with enough content to scroll
2. Place a `Popover` component lower on the page (e.g., inside a form)
3. Use the default portal behavior (`PopoverContent` wraps with `Portal` + `Positioner` + `Popup`)
4. Click the popover trigger
5. **Observe**: page scrolls to the top

## Expected Behavior

The popover should open positioned near its trigger without causing any page scroll.

## Actual Behavior

The page scrolls to the top when the popover opens. On the second open (if you scroll back and click again), it may not scroll because the portaled element is already mounted and focused in place.

## Root Cause

In `PopoverPopup.js`, the popup is rendered inside a `FloatingFocusManager`:

```jsx
<FloatingFocusManager
  context={floatingContext}
  modal={focusManagerModal}
  initialFocus={resolvedInitialFocus}
  // ...
>
```

When `FloatingFocusManager` moves focus to the popup (or an element inside it), the browser's native `element.focus()` scrolls to make the element visible. Since the portal appends to `<body>`, this scrolls to the top/bottom of the page.

Setting `modal={false}` on the Popover root does not fix the issue.

## Environment

- `@base-ui/react`: 1.3.0
- React: 19.2.4
- Browser: Chrome 146
- OS: macOS

## Workaround

None found so far. Setting `modal={false}` does not prevent the scroll.

## Suggested Fix

The `focus()` calls within `FloatingFocusManager` (or the initial focus logic in `PopoverPopup`) should use `element.focus({ preventScroll: true })` when the popup is portaled, to avoid triggering browser scroll behavior.

addu2612:
## Summary
- prevent initial focus from scrolling the page when a popover opens
- add a regression test covering default initial focus on a tabbable element inside the popup

## What changed
When the popover opens, `FloatingFocusManager` can move focus to the first tabbable element inside the popup. Because the popup is rendered through a portal, that focus move can scroll the page unexpectedly. This change applies `preventScroll: true` to the initial focus enqueue path so opening the popover doesn't jump the page.

## Testing
- `pnpm vitest packages/react/src/popover/popup/PopoverPopup.test.tsx --run`

Fixes #4520

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
@@ -746,7 +746,10 @@ export function FloatingFocusManager(props: FloatingFocusManagerProps): React.JS
       elToFocus = elToFocus || getDefaultFocusElement();
 
       enqueueFocus(elToFocus, {
-        preventScroll: elToFocus === floatingFocusElement,
+        // Opening a portal-backed floating element should not unexpectedly scroll
+        // the page to reveal the focused element. This is especially visible for
+        // popovers rendered at the end of <body>.
+        preventScroll: true,
       });
     });
   }, [
--- a/packages/react/src/popover/popup/PopoverPopup.test.tsx
+++ b/packages/react/src/popover/popup/PopoverPopup.test.tsx
@@ -44,7 +44,7 @@ describe('<Popover.Popup />', () => {
       Object.defineProperty(HTMLElement.prototype, 'focus', {
         configurable: true,
         writable: true,
-        value(options) {
+        value(options?: FocusOptions) {
           lastPreventScroll = options?.preventScroll;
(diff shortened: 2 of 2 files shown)


mj12albert:
https://github.com/mui/base-ui/issues/4520#issuecomment-4190029969

PR #4551, opened 2026-04-07
[switch][checkbox] Prevent input state changes in readOnly mode

tsbehlman:
`Switch` and `Checkbox` are currently able to be toggled by clicking their paired `label` element even in `readOnly` mode, whether the label encloses the control or is arranged as a sibling element:

https://github.com/user-attachments/assets/a764b5da-02d8-4a63-a69c-b266b6fac733

It's possible to work around this issue using a `nativeButton` configuration, which causes the `label` to target the `button` element instead of the hidden `<input type="checkbox" />`. This PR aims to fix the default configuration by marking the hidden input as `disabled` in read-only mode. I chose this over the built-in `readonly` attribute as `readonly` does not prevent a checkbox's `checked` state from changing.


packages/react/src/checkbox/root/CheckboxRoot.tsx
```diff
@@ -211,7 +211,7 @@ export const CheckboxRoot = React.forwardRef(function CheckboxRoot(
   const inputProps = mergeProps<'input'>(
     {
       checked,
-      disabled,
+      disabled: disabled || readOnly,
```
mj12albert: This will cause readOnly fields to be incorrectly excluded from form submission, I think a proper fix could be to preventDefault in the click handler

tsbehlman: Excluding the field the field from the form submission is an issue, thanks for pointing that out.

Unfortunately the control element's click handler isn't called when the label is clicked. It looks like calling preventDefault in the input's change handler works, I can see that it prevents the input's `checked` state from changing.

PR #4564, opened 2026-04-09
[drawer][docs] Keep nested demo close animation alive without shadow

atomiks:
[#4545](https://github.com/mui/base-ui/pull/4545) removed `box-shadow` and surfaced logic that was relying on a popup animation to stay smooth when swipe-dismissing under certain conditions. This keeps the property removed while preserving the transition via `opacity` so the drawer stack still closes smoothly.

I don't consider this solution necessarily ideal but it reverts the logic back to its original state at least.

## Changes

- Add a nearly invisible popup opacity transition to the nested CSS Modules demo.
- Mirror the same transition in the Tailwind nested demo so both versions behave the same.


aarongarciah (dismissed):
Seems to work fine. Ideally we should revisit this and try solving it without the opacity hack.

aarongarciah (approved):

PR #4565, opened 2026-04-10
[collapsible] Refactor panel logic

### LINKED ISSUE #904: [Collapsible] With CSS animations there is a flash on loaded when the panel is initially open
With animations, there is a flash when the page is loaded

https://github.com/user-attachments/assets/7a7cbde8-f34b-4158-8a3a-0ecce9c58648

Can we avoid that? I'd guess it happens because we are disabling transitions and animations when it mounts—could this be something we do in the initial render instead?

_Originally posted by @vladmoroz in https://github.com/mui/base-ui/issues/889#issuecomment-2508002780_

mj12albert:
Experiments:
- [collapsible CSS transitions](https://deploy-preview-4565--base-ui.netlify.app/experiments/collapsible/transitions)
- [collapsible CSS animations](https://deploy-preview-4565--base-ui.netlify.app/experiments/collapsible/animations)
- [accordion CSS transition](https://deploy-preview-4565--base-ui.netlify.app/experiments/accordion/transitions)
- [accordion CSS animations](https://deploy-preview-4565--base-ui.netlify.app/experiments/accordion/animations)

`useCollapsiblePanel` now owns panel mount/unmount, dimension tracking, and motion handling completely. `useCollapsibleRoot` no longer touches any of that and is now very thin. Also:
- Fixes https://github.com/mui/base-ui/issues/904
- Added support for `React.Activity` so already-open panels do not replay their open animation when Activity hides and reveals a collapsible
- Added tests


atomiks (changes requested):
## Codex Review (GPT-5.4)

This is primarily a `refactor` PR with secondary `bug-fix`, `tests`, and `docs` scope. I reviewed the full branch against a freshly fetched `master`; the panel-logic consolidation is moving in the right direction, but there is one remaining public-surface bug in the new animation-suppression path.

### 1. Bugs / Issues

#### 1. 🔴 Inline `style` animations bypass the new mount/activity suppression (blocking)

The refactor now suppresses initial-open / Activity-resume keyframe animations by returning `style: { animationName: 'none' }` from `useCollapsiblePanel`, but that value is merged before the consumer `style` prop. Because `Collapsible.Panel` and `Accordion.Panel` both pass the hook result through `useRenderElement`, a consumer can override the suppression with a supported public API such as `style={{ animationName: 'panel-slide-down', animationDuration: '100ms' }}`.

`packages/react/src/collapsible/panel/useCollapsiblePanel.ts`
```ts
props: {
  hidden,
  id: idParam,
  style: {
    animationName: shouldPreventOpenAnimation ? 'none' : undefined,
  },
},
```

`packages/react/src/collapsible/panel/CollapsiblePanel.tsx`
```ts
props: [
  props,
  {
    style: {
      [CollapsiblePanelCssVars.collapsiblePanelHeight as string]:
        height === undefined ? 'auto' : `${height}px`,
      [CollapsiblePanelCssVars.collapsiblePanelWidth as string]:
        width === undefined ? 'auto' : `${width}px`,
    },
  },
  elementProps,
],
```

I verified this with SSR spot checks on both `Collapsible.Panel` and `Accordion.Panel`: initially-open markup still rendered `animation-name: panel-slide-down` when the animation was supplied via `style`, so the issue-904 fix and the new `React.Activity` suppression are incomplete for a real styling path in an unstyled library. The new tests only cover class-based animations, so this regression is currently unguarded.

**Fix:** Apply the suppression after consumer styles have been resolved/merged, or reintroduce an imperative post-merge override on the element, and add an inline-style regression case for both `Collapsible.Panel` and `Accordion.Panel`.

Moving mount/unmount, measurement, and motion ownership into `useCollapsiblePanel` does make the runtime easier to reason about, and I did a second pass over the touched collapsible/accordion runtime files, docs, and experiments after reading the full diff. Outside the style-merge bug above, I did not find another concrete correctness, React lifecycle, or accessibility issue from code reading.

### Test Coverage Assessment

The added `CollapsiblePanel` tests are pointed at the right behaviors: initial-open keyframes, close measurement, `beforematch`, and `React.Activity` resume paths. I also ran `pnpm typescript`, `pnpm test:jsdom CollapsiblePanel --no-watch`, `pnpm test:chromium CollapsiblePanel --no-watch`, `pnpm test:jsdom AccordionPanel --no-watch`, and `pnpm test:chromium AccordionPanel --no-watch`; they all passed. The missing piece is coverage for animation supplied through the public `style` prop, which is exactly where the remaining bug lives.

### Recommendation

Request changes 🔴

The refactor is close, but the new motion suppression can still be bypassed through a supported public styling path on both Collapsible and Accordion, which leaves the core fix incomplete. Once that path is protected and covered by a regression test, I’d be comfortable approving.

flaviendelangle (commented):
# PR Review: [collapsible] Refactor panel logic (mui/base-ui#4565)

**Author:** mj12albert
**Branch:** `refactor/collapsible` -> `master`
**Stats:** +1647 / -1291 across 25 files

## Overview

This PR is a significant refactor that moves panel mount/unmount, dimension tracking, and motion handling from `useCollapsibleRoot` into `useCollapsiblePanel`. The root hook becomes very thin — it only manages open state, transition status, and the trigger handler. The panel hook now owns:

- Animation type detection (`css-transition` | `css-animation` | `none`)
- Dimension measurement and caching
- Mount/unmount lifecycle
- `React.Activity` animation suppression
- `beforematch` / `hiddenUntilFound` handling

It also fixes #904, adds `React.Activity` support to prevent replaying open animations, and adds comprehensive Chromium-only tests.

## Strengths

- **Much cleaner separation of concerns.** The root no longer needs `panelRef`, `abortControllerRef`, `animationTypeRef`, `transitionDimensionRef`, `visible`, `setVisible`, `setDimensions`, `setHiddenUntilFound`, `setKeepMounted`, `runOnceAnimationsFinish`, `height`, `width`. The context object shrinks dramatically, which reduces re-render scope.

- **Well-structured helper functions.** `setTemporaryStyle`, `resetLayoutStyles`, `scheduleRestore`, `cleanup`, `getDimensions`, `getAnimationType`, `hasNonZeroDuration` are well-extracted, well-documented, and pure. This is a big improvement over the inline imperative code that was previously scattered across multiple `useIsoLayoutEffect` callbacks.

- **Thorough test coverage.** The new tests cover CSS transitions, CSS animations, SSR animation suppression, and four `React.Activity` scenarios (transition + animation, initially open + user opened). These are all Chromium-only (`skipIf(isJSDOM)`) which is appropriate since they depend on real layout.

- **Good experiment cleanup.** Shared accordion/collapsible experiment components are extracted into `_components/` directories, eliminating massive CSS and markup duplication across experiment pages.

## Issues & Suggestions

### 1. Typo in comment

In `useCollapsiblePanel.ts`, the comment on `renderedDimensions` has a typo:

> `// These 2 refs are also safe to read in render, bothhold the last committed`

Should be `both hold`.

### 2. `enabled` parameter on `useOpenChangeComplete`

The hook is now called with an `enabled` field:

```ts
useOpenChangeComplete({
  enabled: open && mounted && panelTransitionStatus === 'idle',
  open: true,
  ref: panelRef,
  onComplete() { ... },
});
```

This parameter doesn't appear in the current `useOpenChangeComplete` interface (at least not in the published `@base-ui/react` version). Is there a companion PR adding this parameter, or is this expected to land together? Without `enabled`, the hook would ignore an unknown property and always run.

### 3. `shouldRender` includes `open`

The new `shouldRender` logic is:

```ts
const shouldRender = keepMounted || hiddenUntilFound || mounted || open;
```

Previously it was `keepMounted || hiddenUntilFound || mounted`. The addition of `|| open` makes sense for the first render cycle before `mounted` is set, but worth confirming this doesn't cause a flash of content for initially-closed panels (the `hidden` prop should handle it, but the DOM element will exist briefly).

### 4. `getAnimationType` called every layout effect

Previously `animationTypeRef` was set once and never changed. Now `getAnimationType(panel, shouldPreventOpenAnimation)` is called on every `useIsoLayoutEffect` run (whenever `mounted`, `open`, `transitionStatus`, or `shouldPreventOpenAnimation` changes). The `getComputedStyle` call inside is not free. Consider whether caching the result when `shouldPreventOpenAnimation` is false would help, or confirm that the frequency is acceptable given that these effects only fire on open/close cycles.

### 5. `lastMeasuredDimensionsRef` never cleared

The ref caches the last measured dimensions indefinitely. While this is intentional for CSS animation close paths (to restore a pixel value when `dimensions` resets to `auto`), it means a stale measurement could be used if panel content changes between open/close cycles. This might be acceptable since it only affects the start of the close animation, but worth documenting.

### 6. Removed `hidden` from `AccordionItem` state

The `AccordionItem` state no longer includes `hidden: !collapsible.mounted`. This is a public API change for `AccordionItem.State` consumers who might be using `state.hidden` in `render` callbacks or `className` functions. Confirm this is documented as a breaking change or that `hidden` was never part of the public type.

### 7. Two `useOpenChangeComplete` calls

The hook is now called twice: once for the open-complete path and once for the close-complete path. Previously there was a single call. This is fine architecturally but doubles the MutationObserver / event listener overhead. Given these only activate when `enabled` is true, the overhead should be negligible.

## Risk Assessment

- **Medium risk** — This is a large refactor of core animation/transition logic that touches both Collapsible and Accordion (via shared hooks). The comprehensive test suite mitigates this well, but real-world edge cases (interrupted animations, rapid toggling, `prefers-reduced-motion`, etc.) are worth manual testing via the experiment links.

- The `React.Activity` support uses `React.useEffect` cleanup to detect Activity teardown, which is the correct approach per React 19 semantics.

## Summary

This is a well-executed refactor that significantly simplifies the root hook and consolidates panel logic where it belongs. The code is cleaner, more testable, and better documented than what it replaces. The main items to address are the typo, confirming the `enabled` parameter on `useOpenChangeComplete`, and verifying the `hidden` state removal from `AccordionItem` isn't a breaking change.

flaviendelangle:
> and verifying the hidden state removal from AccordionItem isn't a breaking change.

I think this is a breaking change

packages/react/src/collapsible/panel/useCollapsiblePanel.ts
```diff
@@ -389,36 +300,200 @@ export function useCollapsiblePanel(
...
+/**
+ * Combines multiple optional cleanup callbacks into a single cleanup function.
+ * @param cleanups - The cleanup callbacks to invoke in order.
+ * @returns A cleanup function that runs each defined callback once.
+ */
+function cleanup(...cleanups: Array<(() => void) | undefined>): () => void {
+  return () => {
+    cleanups.forEach((cleanupFn) => cleanupFn?.());
+  };
+}
```
atomiks: There's `mergeCleanups` now in utils

mj12albert:
@flaviendelangle 
- `enabled` parameter on `useOpenChangeComplete`: [it exists](https://github.com/mui/base-ui/blob/e31aa94d267a32eb67a9cc764537a2d0609021bf/packages/react/src/utils/useOpenChangeComplete.tsx#L35) 😅 
- `shouldRender` includes `open`: it's expected and there's no flash, it's needed because `open === true && mounted === false` is possible
- `lastMeasuredDimensionsRef` never cleared: added code comments explaining the niche use-case of this
- `hidden` removed from `AccordionItem` state: added back, but the condition now includes `!isOpen &&`
- `getAnimationType` runs on every layout-effect pass: previously it was set once and forgotten assuming nobody will realistically change between animations/transitions within the lifetime of the component. However this apparently also breaks live `prefers-reduced-motion` changes (e.g. a setting that toggles "Reduce motion"), so I think it's worth supporting this at the cost of calling getComputedStyle effectively on every open change. Minimizing these calls without breaking the reduced motion case gets verbose and might cost the bundle size reduction gained in this refactor 😪

mj12albert:
> Inline `style` animations bypass the new mount/activity suppression

Fixed in https://github.com/mui/base-ui/commit/3ef5c3d994189f1378446c8d8f9d0e69a5f41c04

packages/react/src/collapsible/panel/useCollapsiblePanel.ts
```diff
@@ -2,384 +2,296 @@
...
-      const abortController = new AbortController();
-      abortControllerRef.current = abortController;
-      const signal = abortController.signal;
-
-      let attributeObserver: MutationObserver | null = null;
-
-      const endingStyleAttribute = CollapsiblePanelDataAttributes.endingStyle;
-
-      // Wait for `[data-ending-style]` to be applied.
-      attributeObserver = new MutationObserver((mutationList) => {
```
atomiks, line 224: Not sure if removing this will cause a regression. It watched for DOM mutations to avoid flake issues, specifically: https://github.com/mui/base-ui/pull/3101. Does the new logic avoid flakes like this?

atomiks: [`useAnimationsFinished`](https://github.com/mui/base-ui/blob/68153460a51e0194157669fb1e8a3e9595460b23/packages/react/src/utils/useAnimationsFinished.ts#L100) actually also waits, to avoid flakes. I found that `rAF`, double `rAF`, `setTimeout` etc are all async hacks  that aren't fully reliable in all scenarios

mj12albert: I've added back this part, now it needs `useAnimationsFinished` + 1 RAF because the timing changed, codex explanation:

> the old fix installed its `MutationObserver` earlier in the close lifecycle, before `data-ending-style` landed. The current refactor’s close-completion effect starts later, when `panelTransitionStatus === 'ending'`, which often means the ending-style render has already committed.

atomiks: I think it actually isn't necessary with your previous (new) logic.

> The observer is not needed for the Base UI-controlled data-ending-style timing. This effect only runs when panelTransitionStatus === 'ending', and that same value is passed into panelState.transitionStatus, which renders data-ending-style via transitionStatusMapping. Since this is a passive useEffect, React has already committed that render before this effect runs, so the attribute is already present when the observer would be created. The old #3101 observer was installed earlier in the close lifecycle; this one is installed after the ending-style commit. Keep the RAF before useAnimationsFinished for Chrome’s transition registration timing, but remove the observer.

I had it prove it with tests as well: 

Old logic:
- close handling started earlier
- so there was a real window where the close-completion machinery could start before `data-ending-style` had landed
- that’s what the old `MutationObserver` was protecting

New logic:
- close handling starts later, after the panel is already in the `'ending'` phase
- with the observer removed, the Chromium timing tests I added still pass
- and the new instrumentation shows the actual close animation check sees `data-ending-style` already present

So the regression that justified the old MO does not reproduce under the new lifecycle. The extra RAF still matters; the MO does not look justified anymore.

atomiks:
## Codex Review (GPT-5.4)

Based on the full branch diff against a freshly fetched `master` base, this is a `refactor` PR with `tests` and `react-ui` secondary tags. The panel/root split is cleaner, and the new render-time ref reads look defensible, but I still see one interruption bug that can leak into the next animation cycle plus one public-surface gap in the new suppression strategy.

### 1. Bugs / Issues

#### 1. 🔴 Deferred zero-duration cleanup can leak into the next cycle (blocking)

`packages/react/src/collapsible/panel/useCollapsiblePanel.ts`

```ts
const restoreTransitionDuration = setTemporaryStyle(panel, 'transition-duration', '0s');
setForcePanelIdle(true);
return mergeCleanups(restoreLayoutStyles, scheduleRestore(restoreTransitionDuration));

function scheduleRestore(restore: () => void): () => void {
  let frame = -1;
  let nextFrame = -1;

  frame = AnimationFrame.request(() => {
    nextFrame = AnimationFrame.request(() => {
      restore();
    });
  });

  return () => {
    AnimationFrame.cancel(frame);
    AnimationFrame.cancel(nextFrame);
  };
}
```

The `beforematch`/skip-motion paths temporarily force `transition-duration: 0s` or `animation-duration: 0s`, but the cleanup only cancels the queued RAFs. If that open is interrupted before the two-frame restore runs, the original inline duration never comes back. On a kept-mounted panel, an Activity-hidden subtree, or any case where the same DOM node survives, that leaked `0s` carries into the next close/open cycle and can silently skip the exit animation you were trying to preserve.

**Fix:** make the cleanup restore the original inline value when the scheduled restore has not run yet, instead of only canceling the RAFs.

#### 2. 🟡 `render={<div style={...} />}` can override the new animation suppression (non-blocking)

`packages/react/src/collapsible/panel/CollapsiblePanel.tsx`

```tsx
props: [
  props,
  ...,
  resolvedStyle ? { style: resolvedStyle } : undefined,
  shouldPreventOpenAnimation ? { style: { animationName: 'none' } } : undefined,
],
```

`packages/react/src/utils/useRenderElement.tsx`

```ts
const mergedProps = mergeProps(props, render.props);
```

The new suppression for initially-open / Activity-resumed keyframe panels now depends on an inline `animationName: 'none'` injected before `useRenderElement` clones a `render` element. Because `render.props` are merged after that, a supported customization like `render={<div style={{ animationName: 'panel-slide-down' }} />}` can reintroduce the animation. So the “don’t replay open animation” behavior is currently reliable for `style`/`elementProps`, but not for the `render` element path.

**Fix:** apply the suppression after the render element has been merged, or keep this specific override in a DOM-mutation path that cannot be undone by `render.props.style`.

I specifically checked the new render-time reads of `shouldPreventMountAnimationRef`, `shouldPreventActivityResumeAnimationRef`, `animationTypeRef`, and `lastMeasuredDimensionsRef`. Those reads look acceptable here: the refs are only written from committed layout/effect paths, and the render is using them as a last-committed snapshot rather than mutating them during render.

### Test Coverage Assessment

I ran `pnpm test:chromium CollapsiblePanel --no-watch`, `pnpm test:chromium AccordionPanel --no-watch`, and `pnpm test:chromium AccordionItem --no-watch`; all three suites passed. The current coverage still does not exercise the interrupted skip-motion cleanup path or a `render={<div style={...} />}` panel, which is why both issues above can slip through.

### Recommendation

Request changes 🔴

The refactor direction looks good, but the leaked `0s` duration cleanup is exactly the kind of timing bug that can make exit animations flake or disappear on the next cycle. I’d fix that interruption path before approving, and ideally harden the new suppression so it also survives `render` element styling.

atomiks:
> 2. 🟡 render={<div style={...} />} can override the new animation suppression (non-blocking)

Doubt this is worth handling, left it off to reduce complexity. Fixed the blocking point

atomiks (approved):

atomiks:
Remaining issue with `hiddenUntilFound`: opening, closing then opening the panel again doesn't transition the `opacity` (`[data-starting-style]`); this issue looked pre-existing before the latest commits I did

mj12albert:
Testing the hiddenUntilFound changes – when [the panel](https://deploy-preview-4565--base-ui.netlify.app/experiments/collapsible/transitions) is opened via beforeMatch, clicking the trigger to close doesn't run the close transition

mj12albert:
There's one more thing that seems mostly cosmetic – CSS animation + `hiddenUntilFound` + initially closed (e.g. [Trigger 24](https://deploy-preview-4565--base-ui.netlify.app/experiments/collapsible/animations)) causes an extra `data-starting-style` attribute on the panel that's harmless (since animations don't use it). It doesn't come back after it's opened once. For Accordion, it's only observable in the server rendered HTML (hydration corrects it). According to Codex it can't be fixed nicely without adding an extra state. Shall we just ignore this? @atomiks

atomiks:
This looks good to me, still ✅

PR #4572, opened 2026-04-10
[combobox] Respect rendered chips for keyboard navigation

### LINKED ISSUE #4244: [combobox] Add `limitChips` prop to limit visible selected chips
## Summary

When using `<Combobox.Chips>` in a multi-select combobox, all selected items are rendered as chips with no way to cap the visible count. This can break layouts when many items are selected.

MUI Material's `Autocomplete` solved this with the [`limitTags` prop](https://github.com/mui/material-ui/issues/19137), which:
- Limits visible chips to N when unfocused
- Shows a "+X" indicator for overflow
- Expands to show all chips on focus

## Proposal

Add a `limitChips` (or `limitTags`) prop to `<Combobox.Chips>` that controls the maximum number of visible chips. When the selected count exceeds the limit, a render prop or slot would allow rendering an overflow indicator (e.g., "+3 more").

```jsx
<Combobox.Chips limitChips={3} renderOverflow={(count) => <span>+{count} more</span>}>
  {(chip) => <Combobox.Chip>{chip.label}</Combobox.Chip>}
</Combobox.Chips>
```

## Motivation

- Prevents layout shifts in constrained spaces (filters, toolbars, forms)
- Common pattern already proven in MUI Material's Autocomplete
- Currently requires a custom wrapper to achieve, which is fragile since `Combobox.Chips` controls iteration internally

atomiks:
Closes #4244

This makes the current Combobox API work for limited-chip rendering when users render a prefix slice of selected values, without adding a new prop/API. 

Keyboard navigation and empty-input Backspace now use the rendered chip count instead of the total selected value count, so hidden selections are skipped and the last visible chip is removed.

https://deploy-preview-4572--base-ui.netlify.app/experiments/combobox/limited-chips

## Changes

- Use the rendered chip collection when moving between chips from the input and chip elements.
- Remove the last rendered chip on empty-input Backspace in multi-select comboboxes.
- Add regression coverage for sliced chips and a private docs experiment for manual testing.


atomiks:
## Codex Review (GPT-5.4)

Reviewed the full branch against freshly fetched `upstream/master`, with focused subagent sweeps for accessibility/interaction, tests, and React/state boundaries. Dominant type: `bug-fix`, with secondary `tests`, `a11y`, and private docs experiment coverage.

### 1. Bugs / Issues (None)

No concrete blocking or non-blocking issues found for the PR’s stated scope: making prefix-limited rendered chips viable without adding a new public `limitChips` API.

### Support Matrix

| Case | Supported? | Realism | Notes |
|---|---:|---:|---|
| Render first N selected chips with `value.slice(0, n)` and an overflow label | Yes | 10/10 | This matches the `limitChips`-style workaround and the private experiment. |
| ArrowLeft from empty input to the last visible chip | Yes | 10/10 | Now uses the rendered chip count instead of total selected values. |
| ArrowRight from the last visible chip back to the input | Yes | 10/10 | `Combobox.Chip` now checks `chipsRef.current.length`. |
| Empty-input Backspace with hidden trailing values | Yes | 9/10 | Now removes the last rendered chip, not the last hidden selected value. |
| `Combobox.ChipRemove` / Delete for prefix-rendered chips | Yes | 8/10 | Rendered chip index still matches selected-value index in the prefix case. |
| Userland “expand all chips while focused” | Yes | 7/10 | Works as a userland rendering pattern; the PR does not add owned focus-expansion semantics. |
| Overflow label announcement in the private experiment | Yes | 8/10 | The experiment uses `aria-describedby` on the input and hidden text like “4 more selected”. |
| Suffix window, e.g. `value.slice(-n)` | No | 3/10 | Rendered index no longer maps to selected-value index, so Backspace/removal can target the wrong selected value. |
| Arbitrary filtering, e.g. hide selected middle items | No | 4/10 | Out of scope for this workaround; would need chip identity/original-index mapping. |
| Reordered or sorted rendered chips | No | 5/10 | Out of scope for this patch for the same rendered-index vs selected-index reason. |
| Rendering zero chips while values remain selected | Not targeted | 2/10 | With no rendered chip elements, there is no visible chip target for chip keyboard navigation/removal. |

The patch addresses the actual blocker for the prefix-limited workaround: chip keyboard navigation and empty-input Backspace were based on `selectedValue.length`, so hidden selected values could still affect focus and deletion. Using the rendered chip collection fixes that for the “first N chips plus overflow” model without adding public API surface.

The key boundary is that removal still maps rendered chip position back to `selectedValue` by index. That is acceptable for prefix-preserving rendering, but it should not be described as general support for arbitrary chip projections.

### Test Coverage Assessment

The new tests cover the core regression paths: navigation through only rendered chips and Backspace removing the last rendered chip in a sliced-prefix setup. I also ran `pnpm test:jsdom ComboboxInput --no-watch`, `pnpm test:jsdom ComboboxChip --no-watch`, and `pnpm --filter docs typescript`; all passed.

### Recommendation

Approve ✅

The branch resolves the original prefix-limited chip workaround blocker and clearly stays inside a small internal behavior change. The unsupported cases are real boundaries, but they are outside the `limitChips`-style prefix rendering model this PR is enabling.

mnajdova (approved):
Logic looks good. Should we add a public example for it?

PR #4574, opened 2026-04-10
[code-infra] Add performance benchmark

Janpot:
Propagated 

 ## Adjustments                                               
  - Playwright image v1.58.2-noble → v1.59.1-noble to match existing jobs                                                                                     
  - vitest and @vitest/browser-playwright 4.1.0 → 4.1.2 to match root workspace versions


test/performance/tests/example.bench.tsx
```diff
@@ -0,0 +1,20 @@
...
+        </li>
+      ))}
+    </ul>
+  );
+}
+
+benchmark('HeavyList mount', () => <HeavyList count={1000} />, {
+  runs: 10,
+  warmupRuns: 5,
+});
```
atomiks: Questions:

- Is there a need for these example tests? Should we start with a real one instead? I presume we'll need to manually add our own over time?
- How much does this affect CI times? e.g. if we write tons of benchmarks and they're quite heavy, what's the best way to write them?

.circleci/config.yml
```diff
@@ -280,6 +298,9 @@ workflows:
       - docs_validate:
           <<: *default-context
           name: 'Validating Docs'
+      - test_benchmark:
+          <<: *default-context
+          name: 'Benchmark tests'
```
atomiks, line 303: Can we get it to post on PRs with a table of results?

test/performance/tests/example.bench.tsx
Janpot: > * Is there a need for these example tests? Should we start with a real one instead? I presume we'll need to manually add our own over time?

I'm just setting up the framework here, if you have pre-existing benchmarks, I can look into porting them, or we can do that in a follow-up. The examples are here just to verify everything works.

> * How much does this affect CI times? e.g. if we write tons of benchmarks and they're quite heavy, what's the best way to write them?

The tests run in CI in a job of their own, CI time depends on how heavy you make the tests. You'll have to find the sweet spot between a representative test and stable results. There is no silver bullet, you can play with iteration count and the size of your component under test. Few points:
- These tests are not aimed at humans. It may make sense to create a test with 1M components if you want to demonstrate perf wins to humans. For measuring, that may be 99% overhead.
- At the moment, this framework limits itself to:
  - Measuring react rendering time
  - Across interactions (e.g. you can scroll)
  - Measuring when content first painted
  - You can also measure when specific elements paint, this is probably meaningless across interactions though
  - in the long run we may be able to leverage the reporting structure to add more metrics. Document any gaps or wishes, I'll look into them.

.circleci/config.yml
Janpot: https://github.com/mui/mui-public/pull/1277 just got merged, so now it's there https://github.com/mui/base-ui/pull/4574#issuecomment-4222359231

atomiks: Does it measure the delta with master, like bundle size does? For it to be useful it needs to show deltas

Edit: ah I see it says:

> 🚫 No benchmark report found for merge base https://github.com/mui/base-ui/commit/70eb4c45be32d62f49cc40be994795ed9c764c2b or any of its 3 parent commits.

So it'll start showing them once this is merged for new PRs?

Janpot: Yeah, it needs a base commit to compare with, if it can't find one it will just show absolute values.

Also want to point out that we're still figuring out how stable the values will be across different runs.

atomiks:
@Janpot added a realistic benchmark suite that are the highest signal for us and it runs in 2 minutes or so, which is faster than the slowest jobs in the CI still.

Though the comment doesn't update with results: https://github.com/mui/base-ui/pull/4574#issuecomment-4222359231

Janpot:
can you try updating `@mui/internal-benchmark` to 0.0.3-canary.0? previous versions don't include the trigger yet for updating the comment.

atomiks:
## Codex Review (GPT-5.4)

I reran the local benchmark suite 10 times to get a feel for same-machine noise before trusting PR deltas too much. The overall suite time is fairly stable, but several individual benchmarks are noisy enough that small percentage changes should not be treated as meaningful on their own.

### 1. Bugs / Issues (None)

### Benchmark Summary

| Benchmark | Mean (ms) | StdDev (ms) | CV |
|:--|--:|--:|--:|
| Select mount (200 instances) | 30.12 | 6.20 | 20.58% |
| Checkbox mount (500 instances) | 18.56 | 2.39 | 12.90% |
| Slider mount (300 instances) | 37.74 | 4.78 | 12.66% |
| Tabs mount (200 instances) | 49.36 | 6.24 | 12.64% |
| Menu mount (300 instances) | 30.77 | 3.37 | 10.96% |
| Select open (500 options) | 13.56 | 1.33 | 9.83% |
| Menu open (500 items) | 19.36 | 1.54 | 7.96% |
| Popover mount (300 instances) | 27.83 | 2.19 | 7.87% |
| Mixed surface mount (app-like density) | 27.22 | 1.83 | 6.72% |
| Tooltip mount (300 contained roots) | 18.30 | 1.20 | 6.54% |
| Dialog mount (300 instances) | 22.05 | 1.40 | 6.34% |
| Scroll Area mount (300 instances) | 23.24 | 0.69 | 2.95% |

| Aggregate | Value |
|:--|--:|
| Average per-benchmark CV | 9.83% |
| Median per-benchmark CV | 8.89% |
| Whole-suite total CV | 3.96% |

### Correctness Trade-offs

For reviewer-facing PR comments, I would not highlight every delta. Even on one machine, a number of cases are already moving by around 10% or more without any code changes to those benchmarks, and CI is clearly noisier than local. This looks suitable for catching large regressions and for long-term trend tracking, but not for strong conclusions from single-run small deltas.

A reasonable first-pass policy would be to only call out a benchmark in the PR comment when both of these are true:

- absolute delta is at least `20%`
- absolute delta is at least `10ms`

That would suppress a lot of noise while still surfacing the changes that are most likely to matter.

### Recommendation

The suite is useful, but the PR comment should probably avoid treating every benchmark delta as significant. A simple thresholded reporting rule would make the comment much more trustworthy for reviewers.

test/performance/tests/menu.bench.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </Menu.Positioner>
+      </Menu.Portal>
+    </Menu.Root>
+  );
+}
+
+benchmark('Menu mount (300 instances)', () => <MenuMountList />);
+
+benchmark(
+  'Menu open (500 items)',
```
Janpot, line 54: Just to note that technically this also measures mount time. Might make `Menu mount (300 instances)` redundant?

Janpot:
Yes local stability has been tested. Real world use-cases such as the ones you've added, running over a longer period in our CI will show how much value we can get out of these tests. I assume at least following the evolution over a longer period on the master branch will be valuable. Hopefully we can do something useful with per-PR diffs as well. For now the PR comment is mostly informational, the thresholds for success/failure are now stddev-based. This will likely need to be changed to a fixed threshold, but I propose we get some mileage first. Eventually we can also think about blocking merge if certain thresholds are met if you want.

atomiks:
> I propose we get some mileage first

Yeah sure, let's see how it behaves on a few new PRs and we can tweak as needed

test/performance/tests/menu.bench.tsx
atomiks: But the table summary shows it takes less time 🤔 

Edit: oh it's only measuring the mount of 1 menu, then open. I'd say that's fine to just measure open. It doesn't make the 300-instance one redundant

atomiks (approved):

PR #4578, opened 2026-04-10
[combobox] Prevent item taps from blurring input

atomiks:
Tapping a combobox item on iOS (seems new in 26.4+) could still blur the input because the item only cancelled the `pointerdown` path. Safari can still deliver a synthetic `mousedown` for touch taps, which left a fallback path that could steal focus.

This keeps the existing `pointerdown` guard and also cancels `mousedown` on the shared item implementation, so the input stays focused while selection is committed. The shared change also covers `Autocomplete.Item`.

## Changes

- Prevent default on `mousedown` in `Combobox.Item` in addition to the existing `pointerdown` handling.
- Add regression coverage for touch selection focus retention and the `mousedown` fallback path.

### DIFF
--- a/packages/react/src/combobox/item/ComboboxItem.test.tsx
+++ b/packages/react/src/combobox/item/ComboboxItem.test.tsx
@@ -184,6 +184,67 @@ describe('<Combobox.Item />', () => {
     expect(screen.queryByRole('listbox')).toBe(null);
   });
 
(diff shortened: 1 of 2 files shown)


michaldudak (approved):

mnajdova (approved):
👌 works as expected

PR #4581, opened 2026-04-10
[menu] Preserve dialog focus on pointer leave

atomiks:
Pointerleave was clearing the shared queued focus request while a dialog opened from a menu item was still waiting for its initial focus to land. This keeps list navigation's stale item-focus cancellation local to the request it owns, so it no longer cancels a later dialog focus request while preserving the existing global focus ordering for menu item navigation.

## Changes

- Return a canceler from `enqueueFocus()` for the specific scheduled focus frame.
- Use that canceler in list navigation pointerleave handling instead of clearing the global focus queue.
- Add regression coverage for stale item focus and the nested menu dialog trigger case.
- Add a private menu experiment with several dialog trigger composition patterns.


atomiks:
## Codex Review (GPT-5.4)

Reviewed the full branch against a freshly fetched `upstream/master`. Dominant type: bug-fix. The change is narrowly scoped to the shared focus queue, and the new regressions match the reported menu-to-dialog focus failure closely.

### 1. Bugs / Issues (None)

I did not find any branch-relevant correctness, regression, typing, or repo-convention issues in the patch.

Returning a per-request canceler from `enqueueFocus()` addresses the actual interference point: `useListNavigation` can now cancel the stale item-focus request it owns without clearing a later focus request scheduled by `FloatingFocusManager`. The `rafId === currentRafId` guard keeps the existing global “latest queued focus wins” behavior intact for the other `enqueueFocus()` call sites.

### Test Coverage Assessment

The added regressions cover both halves of the bug: canceling stale pending item focus in `useListNavigation` and preserving pending dialog focus when a menu item triggers a nested dialog. I also ran `pnpm test:chromium useListNavigation --no-watch` and `pnpm test:chromium MenuRoot --no-watch`; both passed.

### Recommendation

Approve ✅

The fix matches the reported failure mode, stays local to the focus-queue bug, and the targeted Chromium coverage meaningfully reduces regression risk for the affected interaction path.

flaviendelangle (approved):
# PR #4581 Review — `[menu] Preserve dialog focus on pointer leave`

## Context

The user asked me to run the PR review skill on https://github.com/mui/base-ui/pull/4581, dispatching multiple agents to hunt for regressions, then print the consolidated report to an MD file.

PR #4581 is a focused 5-file fix (281 additions / 4 deletions) by @atomiks. It addresses a bug where `pointerleave` on a menu item was clearing the shared module-level rAF used by `enqueueFocus`, which canceled the dialog's pending initial focus when a Dialog opened from a `Menu.Item` (via `Dialog.Trigger` composition). The fix returns a per-rAF canceler from `enqueueFocus` and uses it locally in `useListNavigation` instead of the global cancel.

I dispatched **three parallel regression-hunting agents**:

1. **enqueueFocus refactor auditor** — checked all `enqueueFocus` call sites (`useListNavigation.ts`, `FloatingFocusManager.tsx` ×3) for behavioral changes.
2. **useListNavigation pointer-logic auditor** — checked the new pointerleave/cancel flow, ordering, race conditions, sync path, unmount safety.
3. **Test coverage and consumer impact auditor** — checked which other components compose list-nav with overlay triggers, and assessed test quality.

## Plan

Write a single consolidated review report at **`/tmp/pr-4581-review.md`** based on the agents' findings, organized as:

- TL;DR verdict
- What the PR changes
- Critical regressions (none)
- Coverage gaps and missing tests (the actionable stuff)
- Verified-safe areas
- Recommendations to the author

## Key findings (preview)

### Verdict: APPROVE with minor coverage suggestions

**No critical regressions identified.** The fix is correct, narrowly scoped, and addresses the root cause (shared module-level `rafId` cross-talk between `useListNavigation` and `FloatingFocusManager`).

### Critical regressions
None.

### Highest-value coverage gaps (actionable)

1. **Combobox / Autocomplete are affected but untested.** [`packages/react/src/combobox/root/AriaCombobox.tsx:1041`](packages/react/src/combobox/root/AriaCombobox.tsx#L1041) passes `focusItemOnHover: highlightItemOnHover` into `useListNavigation`, so the same cross-talk bug can hit a `Combobox.Item` rendered as a `Dialog.Trigger`/`Popover.Trigger`. Zero Dialog/Popover references in [`ComboboxRoot.test.tsx`](packages/react/src/combobox/root/ComboboxRoot.test.tsx).
2. **Menubar is affected.** [`packages/react/src/menubar/Menubar.tsx:94`](packages/react/src/menubar/Menubar.tsx#L94) forwards `highlightItemOnHover={hasSubmenuOpen}`. Untested.
3. **ContextMenu** (built on `MenuRoot`) is affected and not tested for this path.
4. **The new useListNavigation test is weak.** [`useListNavigation.test.tsx:689-744`](packages/react/src/floating-ui-react/hooks/useListNavigation.test.tsx#L689-L744) only asserts pointerleave still clears item focus — it does **not** demonstrate the cross-talk fix because there is no competing focus request. Reverting the fix would still pass it. The test exercises the code path, not the behavior the PR claims.
5. **The MenuRoot test is genuinely behavioral.** [`MenuRoot.test.tsx:859-937`](packages/react/src/menu/root/MenuRoot.test.tsx#L859-L937) installs rAF mocks between `click(dialogTrigger)` and `pointerLeave` so both the list-nav item focus and `FloatingFocusManager`'s dialog focus ([`FloatingFocusManager.tsx:748`](packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx#L748)) land in the same mocked frame map. Reverting the fix would make this test fail. ✅

### Verified safe

- Three `enqueueFocus` callers in `FloatingFocusManager.tsx` (lines 748, 954, 974) discard the new return value cleanly — no type/lint impact.
- The canceler's `if (rafId === currentRafId)` guard correctly makes a stale canceler a no-op when a later `enqueueFocus` superseded it.
- Sync path (`forceSyncFocusRef.current === true`) returns `NOOP` and is handled correctly: focus already happened, nothing to cancel.
- Race condition `pointerleave(A) → pointermove(B)`: `cancelQueuedFocusRef.current` is overwritten before any subsequent pointerleave can fire a stale canceler.
- Combobox uses `virtual: true`, so its `runFocus` never touches `cancelQueuedFocusRef` — pointerleave's `?.()` is a no-op (identical to old behavior).
- `Select` is **not** affected: [`SelectRoot.tsx:363`](packages/react/src/select/root/SelectRoot.tsx#L363) sets `focusItemOnHover: false`.
- `NavigationMenu` does not use `useListNavigation` — exempt.
- Order of operations in `onPointerLeave` (cancel → `indexRef = -1` → `onNavigate(event)` → conditional floating refocus) is preserved.
- No reviewer concerns in the PR thread (only an automated Codex review approving).

### Pre-existing (not a regression, worth noting)

- `focusItem`'s outer `requestAnimationFrame` for scroll-into-view ([`useListNavigation.ts:335-357`](packages/react/src/floating-ui-react/hooks/useListNavigation.ts#L335-L357)) is **not** captured by any canceler. Pre-existing; the PR doesn't change it.
- No unmount cleanup for `cancelQueuedFocusRef`. A pending rAF will still call `el?.focus()` on a possibly detached node (no-op). Pre-existing.

## Recommendations to author

1. (P1) Strengthen `useListNavigation.test.tsx` test: add a competing focus request (e.g., a `FloatingFocusManager` initial focus) that would be canceled by the old global `rafId` path, so the test would actually fail if the fix were reverted.
2. (P2) Add a regression test for `Combobox.Item` rendered as `Dialog.Trigger` (and/or `Popover.Trigger`) — same pattern as the new `MenuRoot` test.
3. (P3) Consider documenting why `focusItem`'s outer scroll-into-view rAF remains uncanceled (or capture it too).
4. (P3) Optionally add a docs example showing the supported `Menu.Item` + `Dialog.Trigger` composition pattern (none currently exists in [`docs/src/app/(docs)/react/components/menu/`](docs/src/app/(docs)/react/components/menu/)).

PR #4587, opened 2026-04-12
[navigation-menu] Fix transition width when reopening (#3073)

### LINKED ISSUE #3073: Navigation Menu resets transition when mouse hovers an item without content
# Bug report

## Current behavior
I'm having a navigation menu with 3 items, the first has popup content, the second has none and the third has it again.
When I'm hovering over those items it seems like the hover over the second item (the without the popup) is messing up the transition of the ones which have a popup.

It seems like the `popup-width` change is instant and the arrow component doesn't stay centered.

Here is a video:

https://github.com/user-attachments/assets/5bde69e8-ca53-4d58-8dfc-324498a917e2

## Expected behavior

The transition stays smooth even when there is a item without a popup in between.

## Reproducible example

https://codesandbox.io/p/devbox/sweet-jennings-x3j4x7

## Base UI version

`v1.0.0.beta-4`

Firefox

## Which OS are you using?

Windows

mdm317:
fixes #3073

This happened in this sequence:
`open: true` → `open: false` while the closing animation is still running → `open: true`

During that path:
1. `--popup-width` is removed, so `width: var(--popup-width)` resolves to `auto`.
2. Layout reflow is triggered. (`getComputedStyle `)
3. `--popup-width` is set to the next width.

In this flow, the width transition does not start.

### test
https://deploy-preview-4587--base-ui.netlify.app/experiments/navigation-menu

### DIFF
--- a/docs/src/app/(private)/experiments/navigation-menu.tsx
+++ b/docs/src/app/(private)/experiments/navigation-menu.tsx
@@ -208,6 +208,105 @@ export default function ExampleNavigationMenu() {
         </NavigationMenu.Portal>
       </NavigationMenu.Root>
 
+      <NavigationMenu.Root className={styles.Root} style={{ marginTop: '2rem' }}>
+        <NavigationMenu.List className={styles.List}>
+          <NavigationMenu.Item>
+            <NavigationMenu.Trigger className={styles.Trigger}>
+              Overview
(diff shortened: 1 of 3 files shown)


atomiks (approved):
Thanks for the fix @mdm317

PR #4603, opened 2026-04-14
[docs-infra] Add authority and package rename notes to llms.txt preamble

Janpot:
Extends the llms.txt preamble with two instructions for LLM consumers:

- The documentation is authoritative when it conflicts with prior training data.
- The package was renamed from `@base-ui-components/react` to `@base-ui/react`; imports should use the new name.

See https://deploy-preview-4603--base-ui.netlify.app/react/components/dialog.md


mnajdova (commented):
This makes sense to me, I'd say we should try it.

colmtuite (approved):

Issue #869, opened 2024-11-25
[Select] Initial highlight sometimes flashes in Safari


vladmoroz:
Sometimes opening the Select in Safari briefly flashes the initial highlight and removes it until the next mouse move. Can't figure out any specific reproduction steps.

It doesn't happen often though, but I have been seeing this a lot while using the docs normally. 

https://github.com/user-attachments/assets/61d69704-7f91-4059-a59b-6ab5019a17ea

**Search keywords**:

vladmoroz:
OK I think this happens reliably if I manage to move the mouse between mouse down and mouse up

https://github.com/user-attachments/assets/5d324421-9dd8-4e15-86bb-ed50c91891e9

atomiks:
I just noticed this recently as well. Looks like Radix Select also suffers from this issue.

PR #4604, opened 2026-04-14
[all components] Clear highlight on pointer leave when item is clipped by scroll container

atomiks:
Removing the shared `isMouseWithinBounds` guard lets `useListNavigation` clear hover highlight again when an item is partially clipped by a scroll container.

That guard was originally added as a Safari workaround for an older Select behavior (`alignItemWithTrigger` issue), but it does not seem to reproduce on the current path anymore. Keeping it in the shared hook now blocks legitimate `pointerleave` handling, so this change removes the extra guard instead of adding more conditions.

## Changes

- Remove the shared `isMouseWithinBounds` early return from `useListNavigation` item `onPointerLeave`.
- Add a regression test for a hovered item whose bounds extend past a clipped scrolling container.

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useListNavigation.test.tsx
+++ b/packages/react/src/floating-ui-react/hooks/useListNavigation.test.tsx
@@ -775,15 +775,19 @@ describe('useListNavigation', () => {
       await flushMicrotasks();
     });
 
-    it('does not clear the active item on pointer leave when the pointer is still within bounds', async () => {
+    it('clears the active item when the pointer leaves a clipped container while still within the item bounds', async () => {
       const spy = vi.fn();
       render(<App onNavigate={spy} />);
 
       fireEvent.click(screen.getByRole('button'));
 
+      const menu = screen.getByRole('menu');
       const item = screen.getByTestId('item-1');
 
-      vi.spyOn(item, 'getBoundingClientRect').mockReturnValue({
+      menu.style.overflow = 'auto';
+      menu.style.maxHeight = '40px';
+
+      vi.spyOn(menu, 'getBoundingClientRect').mockReturnValue({
         x: 0,
         y: 0,
         top: 0,
@@ -797,24 +801,39 @@ describe('useListNavigation', () => {
         },
       });
 
-      fireEvent.keyDown(screen.getByRole('menu'), { key: 'ArrowDown' });
+      vi.spyOn(item, 'getBoundingClientRect').mockReturnValue({
+        x: 0,
+        y: 0,
+        top: 0,
+        right: 100,
+        bottom: 80,
+        left: 0,
+        width: 100,
+        height: 80,
+        toJSON() {
+          return {};
+        },
+      });
+
+      fireEvent.mouseMove(item);
 
       await waitFor(() => {
         expect(item).toHaveFocus();
       });
 
-      const callsBeforeLeave = spy.mock.calls.length;
-
       await act(async () => {
         fireEvent.pointerLeave(item, {
           clientX: 50,
-          clientY: 20,
+          clientY: 60,
           pointerType: 'mouse',
+          relatedTarget: document.body,
         });
       });
 
-      await flushMicrotasks();
-      expect(spy).toHaveBeenCalledTimes(callsBeforeLeave);
+      await waitFor(() => {
+        expect(item).toHaveAttribute('aria-selected', 'false');
+      });
+      expect(spy.mock.calls.at(-1)?.[0]).toBe(null);
     });
   });
 
--- a/packages/react/src/floating-ui-react/hooks/useListNavigation.ts
+++ b/packages/react/src/floating-ui-react/hooks/useListNavigation.ts
@@ -1,7 +1,6 @@
 'use client';
 import * as React from 'react';
 import { isHTMLElement } from '@floating-ui/utils/dom';
-import { isMouseWithinBounds } from '@base-ui/utils/isMouseWithinBounds';
 import { useValueAsRef } from '@base-ui/utils/useValueAsRef';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
@@ -534,10 +533,6 @@ export function useListNavigation(
         forceSyncFocusRef.current = true;
 
         const relatedTarget = event.relatedTarget as HTMLElement | null;
-
-        if (isMouseWithinBounds(event)) {
-          return;
-        }
 
         if (!focusItemOnHover || listRef.current.includes(relatedTarget)) {
           return;
--- a/packages/utils/src/isMouseWithinBounds.ts
+++ b/packages/utils/src/isMouseWithinBounds.ts
@@ -1,3 +1,6 @@
+/**
+ * @deprecated This utility is no longer used internally and will be removed in the next version.
+ */
 export function isMouseWithinBounds(
   event: React.MouseEvent<HTMLElement> | React.PointerEvent<HTMLElement>,
 ) {


flaviendelangle (commented):
# PR Review: mui/base-ui#4604

**Title:** [all components] Clear highlight on pointer leave when item is clipped by scroll container
**Author:** atomiks
**Base → Head:** `master` ← `codex/clear-highlight-pointerleave-clipped-scroll-container`
**Files:** 2 changed · +27 / -13

## Context

The `onPointerLeave` handler for list items in [useListNavigation.ts](packages/react/src/floating-ui-react/hooks/useListNavigation.ts) had an early return `if (isMouseWithinBounds(event)) return;`. That util checks whether the pointer's `clientX/Y` is inside `event.currentTarget.getBoundingClientRect()`.

When a menu uses an overflow-scrolling container (`overflow: auto` with a bounded height), an item's `getBoundingClientRect()` can extend past the visible region. A cursor leaving the visible area of the scroll container — while still inside the item's full (clipped) bounding rect — would fire `pointerleave`, but the guard would early-return, so the active-index highlight would never clear.

The guard was originally added as a Safari workaround for [#869](https://github.com/mui/base-ui/issues/869) ("Initial highlight sometimes flashes in Safari" on Select, tied to `alignItemWithTrigger`). The PR description asserts the original bug "does not seem to reproduce on the current path anymore."

## Diff summary

[useListNavigation.ts](packages/react/src/floating-ui-react/hooks/useListNavigation.ts):
- Remove `import { isMouseWithinBounds } from '@base-ui/utils/isMouseWithinBounds';`
- Remove the 4-line `if (isMouseWithinBounds(event)) return;` early return inside `onPointerLeave` (around line 538).

[useListNavigation.test.tsx](packages/react/src/floating-ui-react/hooks/useListNavigation.test.tsx):
- Rename test from *"does not clear the active item on pointer leave when the pointer is still within bounds"* to *"clears the active item when the pointer leaves a clipped container while still within the item bounds"*.
- Mock the menu with `overflow: auto`, `maxHeight: 40px`, and `getBoundingClientRect` → height 40.
- Mock the item's `getBoundingClientRect` → height 80 (clipped by the 40px menu).
- Replace `fireEvent.keyDown(..., { key: 'ArrowDown' })` with `fireEvent.mouseMove(item)` to activate hover.
- Fire `pointerLeave` at `clientY: 60` (below the menu, still inside the item rect), with `relatedTarget: document.body`.
- Assert `aria-selected="false"` and `spy.mock.calls.at(-1)?.[0] === null`.

## Critical Issues

None — the fix itself is sound and the test meaningfully exercises the new behavior.

## Important Issues

### 1. `isMouseWithinBounds` becomes dead code after this PR

A repo-wide grep (excluding `node_modules` and build output) shows the util has exactly one call site, which this PR removes:

```
packages/utils/src/isMouseWithinBounds.ts:1        (definition)
packages/react/src/floating-ui-react/hooks/useListNavigation.ts:4    (removed by PR)
packages/react/src/floating-ui-react/hooks/useListNavigation.ts:538  (removed by PR)
```

After merge, [packages/utils/src/isMouseWithinBounds.ts](packages/utils/src/isMouseWithinBounds.ts) has no internal consumers. Since `@base-ui/utils` is a published package, there's a question of whether to:
- **(Preferred)** Delete the file outright. It carries a Safari-workaround comment and a specific bounds heuristic that's only meaningful in the context of the fix this PR is removing — keeping it around invites misuse.
- Leave it as an intentional util with a deprecation note.

Recommend deleting, since the comment inside the util (tying it to issue #869) becomes misleading the moment the hook stops calling it.

### 2. No regression guard for the original Safari bug (#869)

The PR description softens the claim: *"it does not seem to reproduce on the current path anymore"*. That's a reasonable engineering judgment, but the PR:
- Doesn't cite evidence (e.g., a linked WebKit fix, a before/after reproduction on current Safari).
- Doesn't add a test that would fail if #869's symptom returned.

Issue #869's video showed the bug flashing when moving the mouse between mousedown and mouseup on a `Select` item — not something trivially covered by jsdom. A short manual verification checklist in the PR description, or a comment in `useListNavigation.ts` pointing at the history so a future investigator can find the context, would harden this.

Minimum ask: manually re-verify the #869 scenario (aligned Select, Safari, mousedown → move → mouseup on an item) and record the result in the PR description.

## Suggestions

### 3. Update the test's hover activation comment, or inline a short note

The new test switches from `keyDown ArrowDown` to `fireEvent.mouseMove(item)` for hover activation. That's correct (the hook's `onMouseMove` calls `syncCurrentTarget` when `focusItemOnHover` is true), but the intent is non-obvious. A single-line comment like `// Activate hover via mouse, not keyboard, so onPointerLeave's isPointerModalityRef is true` would spare future readers the grep.

(Optional — the existing codebase leans toward no comments, so this is a judgement call.)

### 4. Verify behavior across all `useListNavigation` consumers

`useListNavigation` is shared by Menu, Select, Autocomplete, Combobox, and Toolbar item nav. The early-return removal affects all of them. For consumers without clipped scroll containers, behavior is unchanged (pointerleave only fires when the cursor genuinely exits the element). Still, a quick smoke test of:
- Menu with long item list + inner scroll
- Select with `alignItemWithTrigger` (the historical trouble spot)
- Combobox popup with scroll

would be cheap insurance.

## Strengths

- **Surgical**: 5 lines removed from the hook, targeted rewrite of a single test — easy to review and revert if needed.
- **Clear narrative**: Commit message and PR description explain *why* (removing a stale workaround) rather than just *what*.
- **Test strengthened, not weakened**: The new assertions (`aria-selected="false"` and `spy.mock.calls.at(-1)?.[0] === null`) are positive assertions about the desired behavior, stronger than the old "spy call count didn't change" negative check.
- **Correct guard preservation**: The other early returns in `onPointerLeave` (`latestOpenRef`, `isPointerModalityRef`, `event.pointerType === 'touch'`, `focusItemOnHover`, `listRef.includes(relatedTarget)`, `resetOnPointerLeaveRef`) are all untouched, so keyboard-driven and touch flows remain unaffected.
- **CLAUDE.md compliance**: No `window.setTimeout`/`requestAnimationFrame`, no `as any`, no new public API, Vitest APIs only, test file alongside source, scope label `[all components]` fits AGENTS.md guidance.

## Recommended Action

1. **Delete** [packages/utils/src/isMouseWithinBounds.ts](packages/utils/src/isMouseWithinBounds.ts) as part of this PR, or open a follow-up explicitly scoped to the cleanup. Leaving it in-tree with a comment that points at an already-removed workaround is a rot hazard.
2. **Manually re-verify issue #869** on current Safari with `alignItemWithTrigger` and note the result in the PR description. If the bug still reproduces, this PR reopens it and should be rethought (possibly scoping the removal to the non-Safari path, or keying on clipped-container detection instead of removing the guard entirely).
3. Optionally, smoke-test Menu / Select / Autocomplete / Combobox hover-highlight behavior end-to-end before merging.
4. Suggestions #3 and #4 are non-blocking polish.

---

## Verification

To reproduce the fix locally:

```bash
pnpm test:jsdom useListNavigation --no-watch
```

Manual check for the clipped-scroll scenario (any Menu/Select with a constrained-height popup):
1. Open the popup so hover highlight tracks the pointer.
2. Hover an item near the bottom edge of the scroll container so the item rect extends past the visible area.
3. Move the cursor just past the container's clip line.
4. Expected (post-fix): highlight clears. Pre-fix: highlight sticks.

Safari regression check for #869:
1. Open a `Select` with `alignItemWithTrigger` in Safari.
2. Press and hold on an item, drag the cursor slightly, release.
3. Expected: the highlight should not flicker or disappear.

flaviendelangle:
I think we can drop `isMouseWithinBounds`, it's super specific and the utils package is unstable anyway :eyes:

atomiks:
> I think we can drop isMouseWithinBounds, it's super specific and the utils package is unstable anyway 👀

It would require bumping the package to `0.3.0`, which might be okay, but I think we can leave it in for now until we really need to break an existing util and do it in the same release, wdyt? To reduce (breaking) version churn on consumers

flaviendelangle:
As you want, it's clearly not an important decision either way.
Maybe at least add a tag on the function to identify the one we no longer use.

flaviendelangle (approved):

romgrk:
I think `isMouseWithinBounds` should have been removed. There isn't more work if we release a minor or patch version — in `0.x` versions, any version change must be considered like a breaking version.

PR #4607, opened 2026-04-14
[combobox][docs] Fix TalkBack option wrappers in docs demos

### LINKED ISSUE #4435: [docs][combobox] Wrapping `<div>` around `<Combobox.Item>` text breaks TalkBack
# Docs feedback

TalkBack breaks when a `<div>` wraps the text label in `<Combobox.Item>`:

```html
<div role="option">
  <div>Label</div>
</div>
```

You can't swipe over options to select them, it only ever focuses the listbox parent element.

A `<span>` works (or no wrapper):

```html
<div role="option">
  <span>Label</span>
</div>
```

atomiks:
Fixes #4435

TalkBack breaks in the docs demos when an option renders block-level `div` wrappers inside `Combobox.Item` or `Autocomplete.Item`. This change replaces those wrappers with `span`s across the affected demos and handbook form examples while keeping the existing layouts intact.

## Changes

- Replace option-internal `div` wrappers with `span` wrappers in the affected combobox demos.
- Apply the same wrapper change to the affected autocomplete demos and handbook form demos.
- Preserve the existing grid and flex layout styling so the docs presentation stays the same.

### DIFF
--- a/docs/src/app/(docs)/react/components/autocomplete/demos/async/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/autocomplete/demos/async/css-modules/index.tsx
@@ -89,10 +89,10 @@ export default function ExampleAsyncAutocomplete() {
             <Autocomplete.List>
               {(movie: Movie) => (
                 <Autocomplete.Item key={movie.id} className={styles.Item} value={movie}>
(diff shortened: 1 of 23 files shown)


aarongarciah (approved):
Nice catch

PR #4609, opened 2026-04-14
[otp] Refactor OTPFieldRoot and OTPFieldHiddenInput 

sai6855:
Tiny improvements to OTP component to reduce bundle size while keeping same code readability

### DIFF
--- a/packages/react/src/otp-field/input/OTPFieldInput.tsx
+++ b/packages/react/src/otp-field/input/OTPFieldInput.tsx
@@ -128,12 +128,6 @@ export const OTPFieldInput = React.forwardRef(function OTPFieldInput(
         return;
       }
 
-      if (index > value.length) {
-        const nextIndex = Math.min(value.length, length - 1);
-        focusInput(nextIndex);
-        return;
-      }
-
       handleInputFocus(index, event);
     },
     onBlur(event) {
--- a/packages/react/src/otp-field/root/OTPFieldRoot.tsx
+++ b/packages/react/src/otp-field/root/OTPFieldRoot.tsx
@@ -428,6 +428,7 @@ function OTPFieldHiddenInput(props: OTPFieldHiddenInputProps) {
     length,
     validationType,
   } = useOTPFieldRootContext();
+
   if (!Number.isInteger(length) || length <= 0) {
     return null;
   }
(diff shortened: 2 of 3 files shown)


packages/react/src/otp-field/root/OTPFieldRoot.tsx
```diff
@@ -391,72 +391,43 @@ export const OTPFieldRoot = React.forwardRef(function OTPFieldRoot(
...
           validation={validation}
-          value={value}
         />
       </OTPFieldRootContext.Provider>
     </CompositeList>
   );
 });
 
 interface OTPFieldHiddenInputProps {
-  autoComplete: string | undefined;
```
sai6855, line 419: OTPFieldHiddenInput is rendered inside the OTPFieldRootContext.Provider but was receiving 18 individual
  props instead of reading from context. Replaced 14 of those with a useOTPFieldRootContext() call, keeping only the  props that aren't in
  context

packages/react/src/otp-field/input/OTPFieldInput.tsx
```diff
@@ -128,12 +128,6 @@ export const OTPFieldInput = React.forwardRef(function OTPFieldInput(
         return;
       }
 
-      if (index > value.length) {
```
sai6855, line 131: OTPFieldInput's onFocus had an index > value.length redirect that was already handled by handleInputFocus
  in OTPFieldRoot (which uses the valueRef.current). Removed the dead copy.

packages/react/src/otp-field/utils/otp.ts
```diff
@@ -36,12 +36,8 @@ export function getOTPValidationConfig(validationType: OTPValidationType) {
   return OTP_VALIDATION_CONFIG[validationType];
 }
 
-function removeWhitespace(value: string) {
-  return value.replace(/\s/g, '');
-}
-
 export function stripOTPWhitespace(value: string | null | undefined) {
-  return removeWhitespace(value ?? '');
```
sai6855, line 44: The private removeWhitespace one-liner in otp.ts was only called by stripOTPWhitespace. Inlined the .replace()
  call directly.

atomiks (approved):

PR #4618, opened 2026-04-15
[test] Inline merge-base baseline in CircleCI benchmark job

Janpot:
Rework the CircleCI \`test_benchmark\` job to inline a baseline report into the head upload:

- Compute the merge-base commit with \`origin/master\`.
- Run benchmarks at that commit in a \`git worktree\` at \`/tmp/base-worktree\`.
- Export \`BENCHMARK_BASELINE_PATH\` via \`\$BASH_ENV\` so the head run picks it up automatically as the reference to use for comparison.


michaldudak:
_Codex-authored comment._

I think the brittle case is this pair of commands:

```sh
git fetch --no-tags origin master
BASE_SHA=$(git merge-base HEAD origin/master)
```

`git fetch origin master` fetches the remote branch, but without an explicit destination ref it only reliably records that result in `FETCH_HEAD`. In a CI checkout where `refs/remotes/origin/master` was not created, was narrowed by the checkout refspec, or is simply stale, the next line can either fail with `origin/master` missing or compute the baseline from an older remote-tracking ref than the one that was just fetched.

The safer forms are either:

```sh
git fetch --no-tags origin master
BASE_SHA=$(git merge-base HEAD FETCH_HEAD)
```

or, if we want to keep using `origin/master`:

```sh
git fetch --no-tags origin master:refs/remotes/origin/master
BASE_SHA=$(git merge-base HEAD origin/master)
```

That makes the benchmark baseline unambiguously use the base branch commit fetched in this step.

michaldudak (approved):

PR #4626, opened 2026-04-16
[docs-infra][experiments] Separate stylesheet for experiments and playground

aarongarciah:
Preview: https://deploy-preview-4626--base-ui.netlify.app/experiments

This PR isolates styles in the private area so changes to the public area don't affect it.

Changes included:
- `(private)` root layout: simplifies it and stops loading the docs stylesheet and removes fonts preloading.
- `/experiments`
  - Removes Tailwind support and all of the related CSS classes and vars.
  - Adds a copy of Tailwind's preflight as the reset layer as all experiments were created using it. We can look into removing it and simplifying the reset layer in the future.
  - Copies `--color-gray-*`, `--color-blue` and `--color-red` CSS vars (which change with the OS color scheme, just like the ones we have in the public demos today) into its stylesheet.
  - Stops using styles specific to the public website in experiments.
- `/playground`
  - Adds a new isolated stylesheet with Tailwind, using Tailwind's default theme.
- docs main stylesheet: files from `/experiments` and `/playground` are no longer taken into account for Tailwind generation.

---

Notes: 

- Ideally all folders inside `(private)` should have their own root layout to have full-style isolation, but the different folders will never be linked between them i.e. keeping a single root layout for all `(private)` folders makes the folder structure simpler.
- Ideally we should merge `(docs)` and `(website)` under the same root layout, but it's outside the scope of this PR. It'd require moving a lot of files.
- There are a bunch of experiments importing CSS files from demos. I think this is a terrible idea and we should avoid it, but it's outside the scope of this PR.


.gitignore
```diff
@@ -10,8 +10,6 @@
 /docs/.env.local
 /docs/.next
 /docs/export
-/docs/src/app/(private)/playground/*
-!/docs/src/app/(private)/playground/\[slug\]
```
aarongarciah, line 14: Tailwind respects `.gitignore` when it comes to detecting sources. This change is a bit annoying since we'll need to remember to not commit the playground files we create locally.

If we kept the `.gitignore` untouched, we'd need to list specific files and folders in the playground's Tailwind config, which we'd also need to remember not to commit. So apparently there's no perfect solution.

colmtuite (approved):

colmtuite:
> I think this is a terrible idea and we should avoid it

It is terrible yes

dav-is (approved):

Issue #4487, opened 2026-03-30
ToggleGroup orientation prop does not affect keyboard navigation


peoray:
# Bug report

`ToggleGroup` accepts an `orientation` prop, but that prop is not forwarded to `CompositeRoot`. As a result, it affects the rendered state/data attributes, but not the composite keyboard navigation behavior.

Today, arrow key navigation still works on both axes because `CompositeRoot` falls back to its default orientation behavior.

## Current behavior

`ToggleGroup` accepts an `orientation` prop (default `'horizontal'`), but it is not forwarded to `CompositeRoot`. This means the prop only affects the rendered `data-orientation` attribute — it has no effect on keyboard navigation.

Since `CompositeRoot` defaults to `orientation: 'both'`, arrow keys in all four directions always navigate between toggles, regardless of what `orientation` is set to.

```tsx
// orientation="vertical" is set, but Left/Right arrows still navigate
<ToggleGroup orientation="vertical" defaultValue={['bold']}>
  <Toggle value="bold">B</Toggle>
  <Toggle value="italic">I</Toggle>
  <Toggle value="underline">U</Toggle>
</ToggleGroup>
```

## Expected behavior

If orientation="vertical" is provided, I would expect navigation to be constrained to the vertical axis:

- `ArrowUp` / `ArrowDown` move focus
- `ArrowLeft` / `ArrowRight` should not be captured by the composite navigation

If orientation="horizontal" is provided, I would expect the inverse:

- `ArrowLeft` / `ArrowRight` move focus
- `ArrowUp` / `ArrowDown` are not handled by the group

Provide a clear and concise description of the expected behavior.

## Reproducible example

Link to a CodeSandbox, StackBlitz, or other IDE.

## Base UI version

For example v1.0.2

## Which browser are you using?

Chrome/Safari/Firefox/Safari iOS etc.

## Which OS are you using?

Mac OS/Windows/Other etc.

## Which assistive tech are you using (if applicable)?

Voiceover/JAWS etc.

## Additional context

Provide any additional context that might help us identify the problem and find a solution.

CiscoFran10:
Not sure if this was intentional, but I think `orientation` should affect styling/layout only, not keyboard navigation.

To me, multi-axis navigation here seems reasonable, so the main issue is user expectations, if current behavior is intentional, it should be documented more clearly i guess. Radio component seems to work multi-axis as well although there isnt a orientation prop there.

CiscoFran10:
[...]

PR #4628, opened 2026-04-16
[toggle-group] Remove invalid aria-orientation from role="group" element

### LINKED ISSUE #4623: [ToggleGroup] aria-orientation not allowed on role=group element
# Bug report

## Current behavior

With the 1.4.0 update the ToggleGroup component violates the  WAI-ARIA 1.2 rule `aria-allowed-attr`.

`aria-orientation` is not a supported aria attribute for `role="group"` elements. (https://www.w3.org/TR/wai-aria-1.2/#group)

## Expected behavior

The simplest approach seems to use a different role. the `toolbar` role supports `aria-orientation`, but I'm not sure it matches really the ToggleGroup usage

## Reproducible example

base-ui doc: https://base-ui.com/react/components/toggle-group

## Base UI version

1.4.0

Chrome

## Which OS are you using?

Mac OS

## Additional context

We can override the role on the ToggleGroup so there is an easy workaround, but I guess it should be valid from the start :)

sernstberger:
## Summary

`ToggleGroup` renders `role="group"` but was also rendering `aria-orientation`, which is not a valid ARIA attribute for the `group` role ([WAI-ARIA spec](https://www.w3.org/TR/wai-aria-1.2/#group)). This was flagged by axe's `aria-allowed-attr` rule.

The `aria-orientation` attribute was being added by `useCompositeRoot` (which always includes it in its returned props). This was introduced in #4487 when keyboard navigation was added via `CompositeRoot`.

## Fix

Strip `aria-orientation` from the DOM output by adding `{ 'aria-orientation': undefined }` to the `CompositeRoot` props array. This follows the same pattern used by `NavigationMenuList` and `SelectPopup` in the codebase.

The `orientation` prop still reaches `CompositeRoot` for keyboard navigation — it just won't appear as an ARIA attribute on the rendered element.

## Test

Added a test asserting that `aria-orientation` is not present on the `role="group"` element.

Closes #4623
Related: #4487


packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -132,7 +132,7 @@ export const ToggleGroup = React.forwardRef(function ToggleGroup<Value extends s
           style={style}
           state={state}
           refs={[forwardedRef]}
-          props={[defaultProps, elementProps]}
+          props={[defaultProps, { 'aria-orientation': undefined }, elementProps]}
```
mj12albert: `SelectPopup` and `NavigationMenuList` are already overwriting `aria-orientation` like this, perhaps it should be moved out of Composite and its consumers should be required to set `aria-orientation` themselves accordingly

atomiks (approved):

PR #4642, opened 2026-04-17
[all components] Fix display: contents tabbability

### LINKED ISSUE #4622: `checkVisibility` isn't sufficient for tabbability checks
# Bug report

## Current behavior

After updating to Base UI 1.4, we noticed a regression that breaks tabbability inside our overlays in the WordPress/Gutenberg repository.

https://github.com/mui/base-ui/pull/4336 replaced the external `tabbable` dependency with Base UI's internal helper that relies on the `Element.checkVisibility()` method.

We find this check for generic visibility detection too strict / not sufficient for focusability / tabbability decisions. We have at least two downstream cases that point in this direction:

- In [WordPress/gutenberg#77381](https://github.com/WordPress/gutenberg/pull/77381), a floating popup wrapped by an ancestor with `display: contents` (our `ThemeProvider` intentionally renders a layout-inert element to apply styles to DOM subtrees) ended up with no tabbable descendants, breaking focus-trap cycling.
- In [WordPress/gutenberg#77191](https://github.com/WordPress/gutenberg/pull/77191), we found that generic visibility checks were also not sufficient for `VisuallyHidden`-style elements, because those components are intentionally designed to evade normal visibility heuristics while still mattering for accessibility.

Those are different cases, but they seem to share the same broader issue: **some elements fail a generic visibility check even though the resulting focusability behavior should not treat them the same way as actually hidden content**.

## Expected behavior

We can think of two potentisl options:

1. Harden the visibility detection logic.

   In particular, distinguish between cases that are truly hidden vs cases that only look "non-visible" to a generic layout-based check.

   For example, a wrapper with `display: contents` shouldn't be considered non-visible from a tabbability POC. More generally, there may be accessibility-related or layout-related patterns where a generic visibility heuristic is too coarse for focus-management decisions.

2. Expose an escape hatch for specific elements or wrappers.

   For example, allow a data attribute or option that tells Base UI to bypass or relax the visibility check for a known wrapper / element 

   This would be less ideal than fixing the default behavior, but it could be a useful fallback if there are edge cases where the generic check cannot be made reliable.

## Reproducible example

https://stackblitz.com/edit/4egwyytd?file=src%2FApp.tsx

## Base UI version

1.4

N/A

## Which OS are you using?

N/A

N/A

## Additional context

N/A

### LINKED ISSUE #4629: [dialog] display: contents ancestor blocks initial focus
# Bug report

Since v1.4.0 Base UI dialogs stop auto-focusing their content whenever the popup has an ancestor with `display: contents` (for example, a `<form className="contents">` wrapper). The dialog opens, but no field gets focus.

## Reproducible example

https://stackblitz.com/edit/stm1da5g?file=src%2FApp.tsx

## Base UI version

For example v1.4.0

Chrome

## Which OS are you using?

Mac OS

https://github.com/user-attachments/assets/69470db2-ab77-42a5-84a7-c159ccba7f21

atomiks:
Fixes #4622
Fixes #4629

Tabbability regressed when the internal visibility helper walked ancestors and treated `display: contents` wrappers as hidden. This narrows the `display: contents` exception to the ancestor walk, keeps the stricter element-level visibility checks for restore focus and typeahead, and adds regressions around the adjacent hidden-state cases.

## Changes

- Treat `display: contents` ancestors as transparent only in the tabbable tree walk.
- Keep element-level visibility checks strict for restore focus and list navigation paths.
- Add regressions for hidden `display: contents` ancestors, effective-visibility typeahead filtering, and restore-focus when a focused element becomes hidden.

### DIFF
--- a/packages/react/src/dialog/popup/DialogPopup.test.tsx
+++ b/packages/react/src/dialog/popup/DialogPopup.test.tsx
@@ -309,6 +309,77 @@ describe('<Dialog.Popup />', () => {
     });
   });
 
+  describe.skipIf(isJSDOM)('display: contents ancestors', () => {
+    it('keeps initial focus working when the popup is wrapped by a display: contents ancestor', async () => {
+      const { user } = await render(
+        <div>
+          <button data-testid="outside-before">Outside before</button>
+          <Dialog.Root modal={false}>
+            <Dialog.Trigger>Open</Dialog.Trigger>
+            <Dialog.Portal>
+              <form style={{ display: 'contents' }}>
+                <Dialog.Popup data-testid="dialog-popup">
+                  <input data-testid="dialog-input" />
+                  <button type="button">Close</button>
+                </Dialog.Popup>
+              </form>
+            </Dialog.Portal>
+          </Dialog.Root>
+          <button data-testid="outside-after">Outside after</button>
+        </div>,
+      );
+
+      await user.click(screen.getByText('Open'));
+
+      await waitFor(() => {
+        expect(screen.getByTestId('dialog-input')).toHaveFocus();
+      });
+    });
+
+    it('keeps trap-focus tab cycling inside the popup when wrapped by a display: contents ancestor', async () => {
+      const { user } = await render(
+        <div>
+          <button data-testid="outside-before">Outside before</button>
+          <Dialog.Root defaultOpen modal="trap-focus">
+            <Dialog.Portal>
+              <form style={{ display: 'contents' }}>
+                <Dialog.Popup data-testid="dialog-popup">
+                  <input data-testid="first-input" />
+                  <button type="button" data-testid="second-button">
+                    Second
+                  </button>
+                </Dialog.Popup>
+              </form>
+            </Dialog.Portal>
+          </Dialog.Root>
+          <button data-testid="outside-after">Outside after</button>
+        </div>,
+      );
+
+      const popup = screen.getByTestId('dialog-popup');
+
+      await waitFor(() => {
+        expect(screen.getByTestId('first-input')).toHaveFocus();
+      });
+
+      await user.keyboard('[Tab]');
+      expect(screen.getByTestId('second-button')).toHaveFocus();
+
+      await user.keyboard('[Tab]');
+      await waitFor(() => {
+        expect(screen.getByTestId('first-input')).toHaveFocus();
+      });
+      expect(screen.getByTestId('outside-before')).not.toHaveFocus();
+      expect(screen.getByTestId('outside-after')).not.toHaveFocus();
+
+      await user.keyboard('[ShiftLeft>][Tab][/ShiftLeft]');
+      await waitFor(() => {
+        expect(screen.getByTestId('second-button')).toHaveFocus();
+      });
+      expect(popup.contains(document.activeElement)).toBe(true);
+    });
+  });
+
   describe('prop: finalFocus', () => {
     it('should focus the trigger by default when closed', async () => {
       const { user } = await render(
--- a/packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
@@ -182,279 +182,335 @@ function Dialog({ render, open: passedOpen = false, children }: DialogProps) {
   );
 }
 
-describe.skipIf(!isJSDOM)('FloatingFocusManager', () => {
-  describe('prop: initialFocus', () => {
-    test('default behavior focuses first tabbable element', async () => {
-      render(<App />);
+describe('FloatingFocusManager', () => {
+  describe.skipIf(!isJSDOM)('JSDOM-only coverage', () => {
+    describe('prop: initialFocus', () => {
+      test('default behavior focuses first tabbable element', async () => {
+        render(<App />);
 
-      fireEvent.click(screen.getByTestId('reference'));
-      await flushMicrotasks();
+        fireEvent.click(screen.getByTestId('reference'));
+        await flushMicrotasks();
 
-      expect(screen.getByTestId('one')).toHaveFocus();
-    });
+        expect(screen.getByTestId('one')).toHaveFocus();
+      });
 
-    test('default behavior focuses the checked radio in a named group', async () => {
-      render(<RadioApp />);
+      test('default behavior focuses the checked radio in a named group', async () => {
+        render(<RadioApp />);
 
-      fireEvent.click(screen.getByTestId('reference'));
-      await flushMicrotasks();
+        fireEvent.click(screen.getByTestId('reference'));
+        await flushMicrotasks();
 
-      expect(screen.getByTestId('radio-two')).toHaveFocus();
-    });
+        expect(screen.getByTestId('radio-two')).toHaveFocus();
+      });
 
-    test('ref', async () => {
-      render(<App initialFocus="two" />);
-      fireEvent.click(screen.getByTestId('reference'));
-      await flushMicrotasks();
+      test('ref', async () => {
+        render(<App initialFocus="two" />);
+        fireEvent.click(screen.getByTestId('reference'));
+        await flushMicrotasks();
 
-      expect(screen.getByTestId('two')).toHaveFocus();
-    });
+        expect(screen.getByTestId('two')).toHaveFocus();
+      });
 
-    test('respects autoFocus', async () => {
-      render(
-        <App>
-          <input autoFocus data-testid="input" />
-        </App>,
-      );
-      fireEvent.click(screen.getByTestId('reference'));
-      await flushMicrotasks();
-      expect(screen.getByTestId('input')).toHaveFocus();
-    });
-  });
-
-  describe('prop: returnFocus', () => {
-    test('when true', async () => {
-      const { rerender } = render(<App />);
-
-      screen.getByTestId('reference').focus();
-      fireEvent.click(screen.getByTestId('reference'));
-      await flushMicrotasks();
-
-      expect(screen.getByTestId('one')).toHaveFocus();
-
-      act(() => screen.getByTestId('two').focus());
-
-      rerender(<App returnFocus={false} />);
-
-      expect(screen.getByTestId('two')).toHaveFocus();
-
-      fireEvent.click(screen.getByTestId('three'));
-      expect(screen.getByTestId('reference')).not.toHaveFocus();
-    });
-
-    test('when false', async () => {
-      render(<App returnFocus={false} />);
-
-      screen.getByTestId('reference').focus();
-      fireEvent.click(screen.getByTestId('reference'));
-      await flushMicrotasks();
-
-      expect(screen.getByTestId('one')).toHaveFocus();
-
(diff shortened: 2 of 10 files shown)


LukasTy:
# Claude Opus 4.7 review

## Overview

Fixes regressions from #4336 (internal replacement of the external `tabbable` package) where the ancestor visibility walk in tabbable.ts treated `display: contents` wrappers as hidden. Two reported breaks: Gutenberg's `ThemeProvider` (#4622) and dialog auto-focus under `<form className="contents">` (#4629).

**Fix strategy**

- Narrow the `display: contents` exception to the ancestor walk only.
- Harden `isElementVisible` to explicitly cover `visibility: hidden|collapse` and `content-visibility: hidden` (since `element.checkVisibility()` with default options only catches `display: none`).
- Drop `FloatingFocusManager`'s local `isFocusable` in favor of the hardened `isElementVisible`.
- Share a new `isHiddenByStyles` predicate between the tabbable walk and the element-level check.

## Correctness

Walked through the case matrix:

- `display:contents` wrapper (visible) -> descendant tabbable: first clause of `isVisibleInTabbableTree` returns `true`, bypassing `checkVisibility()` on the boxless wrapper. Correct.
- `display:contents` wrapper with `visibility:hidden` -> rejected: first clause fails on `isHiddenByStyles`, fallback `isElementVisible` short-circuits. Correct.
- Candidate itself `display:contents` with `checkVisibility()===false` -> rejected (ancestor flag is false, so the transparency exception doesn't apply). Correct and tested.
- `display:none` ancestor -> rejected via `checkVisibility()` or the `display !== 'none'` fallback. Correct.

Refactor of the `details` clause (`current !== element` -> `isAncestor`) is a pure rename — equivalent.

## Behavior changes worth flagging (not called out in PR description)

The PR says it "keeps the stricter element-level visibility checks for restore focus and list navigation paths," but the element-level check itself got stricter. These are user-visible shifts:

1. **Restore focus on `visibility: hidden`** — `FloatingFocusManager` line 492. Old `isFocusable` delegated to `checkVisibility()` whose defaults ignore `visibility`, so a focused element turned `visibility:hidden` would not trigger restoration. New `isElementVisible` rejects it via `isHiddenByStyles` first, so restore now fires. The added test asserts the new behavior.
2. **Typeahead skips `visibility:hidden` items** — useTypeahead.ts:119. Same root cause. JSDoc was softened to "effectively visible", which is a bit vague — concretely list what counts (display:none / visibility:hidden|collapse / content-visibility:hidden).
3. **`isListIndexDisabled`** — composite.ts:489 inherits the same new strictness. Keyboard list navigation now skips `visibility:hidden` / `content-visibility:hidden` items. Not tested directly.

None of these are unreasonable (arguably all more correct), but they're semantic changes beyond just the display:contents fix. Worth a changelog note.

## Performance

- `getComputedStyle` is now called unconditionally for every ancestor during the tabbable walk. The previous code could short-circuit on `checkVisibility()` and skip layout lookups. Benchmark in the PR shows Menu open (500 items) +8% / +5 ms — small but real. Deeper trees would compound.
- `isVisibleInTabbableTree` threads `styles` into `isElementVisible` to avoid double computation. Good.
- Minor optimization: when `!isAncestor`, `isVisibleInTabbableTree` doesn't need the `display === 'contents'` branch at all. You could skip the explicit `getComputedStyle` and let `isElementVisible`'s default param do it:

      function isVisibleInTabbableTree(element, isAncestor) {
        if (!isAncestor) return isElementVisible(element);
        const styles = getComputedStyle(element);
        if (styles.display === 'contents' && !isHiddenByStyles(styles)) return true;
        return isElementVisible(element, styles);
      }

  Functionally equivalent, avoids one getComputedStyle on the candidate path.

## API / Style nits

- `isElementVisible(element: Element | null, ...)` now accepts `null`. No caller passes `null`, and the `null` branch is purely defensive — tightening to `Element` plus letting callers guard would keep the signature honest.
- The default-parameter expression `styles: CSSStyleDeclaration | null = element ? getComputedStyle(element) : null` is correct (only evaluated when `undefined`), but mildly unusual. A guard inside the body reads more plainly.
- `isHiddenByStyles` is a good shared predicate — consider one direct unit test so regressions in the rule set (e.g., adding a new value) don't rely purely on integration tests.

## Test coverage

Solid on the reported cases. Gaps:

- No explicit test for `content-visibility: hidden` on the **candidate itself**, only on an ancestor.
- No test for `visibility: collapse` (uncommon, mainly `<tr>` / `<col>`, but it's in `isHiddenByStyles`).
- The `keeps descendants of display: contents ancestors` test fakes `checkVisibility` via `Object.defineProperty`. That's fine for JSDOM but sidesteps the question of real browser behavior, which varies by engine/version. A browser-mode test exercising the WordPress pattern (wrapper `display:contents` + real DOM) would be more convincing.
- `isListIndexDisabled` behavior under the new `isElementVisible` isn't covered directly.

## Risks

- **Restore-focus / typeahead / list-nav behavior change** on `visibility:hidden` items (see above). Low risk of regression but worth mentioning in release notes.
- **Candidate with `display:contents`**: behavior now hinges on the browser's `checkVisibility()` semantics for boxless elements. Consumers who rely on `<button style="display:contents">` being tabbable in browsers where `checkVisibility()` returns false will break. The test locks in this choice.
- **VisuallyHidden pattern** (also mentioned in #4622) isn't addressed. Most `visually-hidden` implementations use `clip-path`/`position:absolute` which `checkVisibility()` still reports visible, so they should be unaffected — but if the team wants to close #4622 fully, that second case should either be validated or deferred to a follow-up with a note.

## Verdict

Targeted fix, correct reasoning, good tests for the core regression. The unflagged behavior changes for `visibility: hidden` across restore focus / typeahead / list navigation are the main thing I'd want the author to call out explicitly before merge. Perf regression is small but real and proportional to the correctness win.

LukasTy (approved):
LGTM overall, both issues are resolved with this fix.

I've also asked Claude to compare our own rolled solution with the previous `tabbable` in case you didn't check it.

## How `tabbable` (the library) handles visibility

Its default `displayCheck: 'full'` mode, for each candidate, does this (`isHidden` at src/index.js:403-529):

1. `getComputedStyle(node).visibility === 'hidden'` -> hidden.
2. `matches('details:not([open]) *')` -> hidden.
3. If the node is attached: `return !node.getClientRects().length`.

That third line is the silver bullet: "if this element has any client rect, it's rendered". `getClientRects` naturally captures `display: none` on *any* ancestor, `display: contents` on self (no rect), closed `<details>`, etc. — without walking the tree.

It also has an opt-in `displayCheck: 'full-native'` mode that delegates to `Element.checkVisibility({ contentVisibilityAuto: true, visibilityProperty: true, checkVisibilityCSS: true })` — note the explicit options.

## How base-ui (after this PR) handles visibility

No `getClientRects` check. Instead, per candidate, it walks every ancestor and runs a hand-rolled predicate combining:

- `checkVisibility()` (no options),
- an explicit `isHiddenByStyles` (visibility hidden/collapse + content-visibility hidden),
- a special-case to treat `display: contents` ancestors as transparent unless they have `isHiddenByStyles`.

## Side-by-side

| Concern | `tabbable` (default) | base-ui (this PR) |
|---|---|---|
| `display: contents` ancestor | Non-issue — candidate's own rects are present | Special-cased in the ancestor walk |
| `display: contents` on the candidate itself | Hidden (no rects) | Hidden if `checkVisibility()===false`; otherwise tabbable |
| `display: none` on ancestor | Caught by `!getClientRects().length` | Caught by `checkVisibility()` or the `display !== 'none'` fallback per ancestor |
| `visibility: hidden` | Explicit style check (self only) | Explicit `isHiddenByStyles` (self + every ancestor) |
| `content-visibility: hidden` | Only caught in `full-native` via `contentVisibilityAuto: true` | Explicit predicate |
| `visibility: collapse` | Not checked | Checked |
| Ancestor walking | Avoided — one rect check does it all | Full manual walk, `getComputedStyle` per ancestor |
| `inert` / closed `details` | Selector-based (`closest('[inert]')`, `details:not([open]) *`) | Manual walk |
| Disabled fieldset | Explicit `isDisabledFromFieldset` traversal | Implicit via `matches(':disabled')` (browser handles fieldset rules natively) |
| Shadow DOM | Configurable via `getShadowRoot` / `shadowRootFilter`, iterative walker | Via `getParentElement` (assignedSlot) + `getComposedChildren` |
| Sort order | Sorts by `tabIndex` first, then document order | Document order only |
| Perf cost | Forces layout (`getClientRects`) per candidate | Forces style recalc (`getComputedStyle`) per candidate x ancestor depth |

## Observations

1. **`tabbable` solves `display: contents` structurally, base-ui solves it with a special case.** `getClientRects().length` is one line that handles display:none ancestors, display:contents on self, off-DOM nodes, and closed details all at once. The base-ui approach needs an ancestor walk plus an explicit `display === 'contents'` branch — functionally equivalent for the reported bug, but more moving parts and more ways to regress.

2. **base-ui could simplify by passing options to `checkVisibility()`.** `checkVisibility({ visibilityProperty: true, contentVisibilityAuto: true, checkVisibilityCSS: true })` is exactly what `tabbable`'s `full-native` mode does, and it would make `isHiddenByStyles` unnecessary for browsers that support it (Chrome 105+, Firefox 106+, Safari 17.4+). The `isHiddenByStyles` fallback would only be needed for JSDOM and old browsers. That would cut the per-ancestor cost too, since the native implementation is faster than a JS-side style walk.

3. **base-ui gets `visibility: collapse` that tabbable doesn't.** Minor, but genuine.

4. **base-ui doesn't sort by `tabIndex`.** This is a real behavior difference from the `tabbable` library and from native browser tab order. Probably fine for focus-trap-style use (where elements rarely carry positive tabIndex), but worth knowing if someone hits it.

5. **Perf profile is different, not strictly better or worse.** `getClientRects` forces layout; `getComputedStyle` forces style recalc. For `N` candidates in a tree of depth `D`: tabbable is `O(N)` layout hits; base-ui after this PR is `O(N*D)` style hits. For shallow trees with many candidates they're comparable; for deep trees, base-ui is proportionally more expensive.

6. **Shadow DOM support is more complete in `tabbable`.** The `getShadowRoot` callback lets consumers declare "there's a shadow root here even though you can't see it". base-ui handles open shadow roots via composed children traversal but doesn't have the undisclosed-shadow escape hatch.

## Practical takeaway for the PR

The fix is correct and closes the reported regressions. But if base-ui wants to match `tabbable`'s robustness without re-adopting the dependency, two low-cost upgrades would get it closer:

- Either pass `{ visibilityProperty: true, contentVisibilityAuto: true, checkVisibilityCSS: true }` to `checkVisibility()` and drop most of `isHiddenByStyles` for the native path;
- Or consider `!element.getClientRects().length` as the primary rendered-ness check — it would replace the `display: contents` special case, the ancestor `display: none` walk, and the content-visibility handling with one line (at the cost of forcing layout).

The current PR lands something that works; those are future-direction notes rather than blockers.

ciampo:
I'd be also in favour of making base ui check as close as possible to `tabbable` — as shows in the detailed reviews above, there are many edge cases where we may still want to "white list" a tabbable scenario that `checkVisibility()` would not allow.

PR #4646, opened 2026-04-18
[navigation menu] Fix stale popup size on rapid trigger hover

### LINKED ISSUE #4619: [navigation menu] Popup dimensions become stale when rapidly hovering between menu triggers
# Bug report

## Current behavior

When rapidly hovering across multiple navigation menu triggers that open popups of varying sizes, the popup dimensions can become incorrect or stale. Specifically, a newly opened smaller menu may render at the width or height of a larger submenu that was opened just before it.

This mainly occurs during quick hover interactions between triggers, though I’ve occasionally observed it at more normal cursor speeds as well (albeit less frequently).

## Expected behavior

Each navigation menu popup should always render with the correct dimensions based on its own content, regardless of how quickly the user moves between triggers.

Rapid hover interactions should not cause popups to retain dimensions from previously opened menus.

## Reproducible example
[StackBlitz](https://stackblitz.com/edit/4v8hwphc?file=src%2FApp.tsx)

I've included a screen recording because it isn't always easy to reproduce.

https://github.com/user-attachments/assets/a4412da8-1250-49dd-8220-29ba07cc3476

## Base UI version

v1.4.0

Chrome (likely reproducible in other browsers)

## Which OS are you using?

MacOS

N/A

## Additional context

- This occurs when moving the cursor quickly across multiple top-level navigation triggers.
- The issue is more noticeable when popups have significantly different sizes.
- It appears that layout measurements or animations may not be recalculated in time before the next popup renders.
- I’ve included a screen recording where a smaller submenu takes on the dimensions of a previously opened larger submenu.

atomiks:
Fixes #4619

Rapid trigger hovers could leave a queued popup size update behind from the previously active trigger. That extra size pass could reapply the old panel dimensions after the next trigger had already measured a smaller popup.

## Changes

- Cancel pending popup size work when a trigger stops being the active item.
- Ignore deferred popup size writes once the trigger that scheduled them is no longer active.
- Add a Chromium regression test that rapidly switches between differently sized popups and asserts the old width is not written again.

### DIFF
--- a/packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.test.tsx
+++ b/packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.test.tsx
@@ -45,9 +45,7 @@ function TestNavigationMenuRapidHoverSizing() {
   );
 }
 
-function getPopupWidthCalls(
-  calls: Array<[property: string, value: string, priority?: string]>,
-) {
+function getPopupWidthCalls(calls: Array<[property: string, value: string, priority?: string]>) {
   return calls.filter((call) => call[0] === '--popup-width').map((call) => call[1]);
 }
 
--- a/packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.test.tsx
+++ b/packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.test.tsx
@@ -1,3 +1,4 @@
+import * as React from 'react';
 import { expect, vi } from 'vitest';
 import { NavigationMenu } from '@base-ui/react/navigation-menu';
 import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
@@ -15,6 +16,8 @@ const rapidHoverAnimationStyles = `
 `;
 
 function TestNavigationMenuRapidHoverSizing() {
+  const [isProductExpanded, setIsProductExpanded] = React.useState(false);
+
   return (
     <NavigationMenu.Root>
       {/* eslint-disable-next-line react/no-danger */}
@@ -23,7 +26,16 @@ function TestNavigationMenuRapidHoverSizing() {
         <NavigationMenu.Item value="product">
           <NavigationMenu.Trigger>Product</NavigationMenu.Trigger>
           <NavigationMenu.Content>
-            <div style={{ width: 700, height: 420 }}>Product panel</div>
+            <div>
+              <button type="button" onClick={() => setIsProductExpanded(true)}>
+                Expand Product
+              </button>
+              {isProductExpanded ? (
+                <div style={{ width: 760, height: 460 }}>Expanded product panel</div>
+              ) : (
+                <div style={{ width: 700, height: 420 }}>Product panel</div>
+              )}
+            </div>
           </NavigationMenu.Content>
         </NavigationMenu.Item>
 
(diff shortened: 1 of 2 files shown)


flaviendelangle (approved):
# PR #4646 Review: [navigation menu] Fix stale popup size on rapid trigger hover

**Author:** atomiks
**Branch:** `codex/navigation-menu-fix-stale-hover-size` -> `master`
**Fixes:** #4619

---

## Summary

When rapidly hovering between NavigationMenu triggers with differently-sized popups, a queued `sizeFrame` callback from the previous trigger could fire after the new trigger became active, reapplying the old (stale) popup dimensions. This PR cancels pending size work when a trigger is deactivated and guards the deferred write against stale execution.

---

## Changes Analysis

### Source change (`NavigationMenuTrigger.tsx`, +13 -1)

1. **`React.useEffect` -> `useIsoLayoutEffect` with early return + cancellation (line ~148)**
   - When `isActiveItem` becomes `false`, `sizeFrame.cancel()` and `cancelAutoSizeReset()` are called immediately.
   - When `isActiveItem` is `true`, the effect is a no-op (early return), which is correct — the active trigger should not cancel its own pending work.
   - Using `useIsoLayoutEffect` ensures cancellation happens synchronously before paint, preventing a single frame where the stale callback could fire. Good choice.

2. **Guard inside `sizeFrame.request` callback (line ~253)**
   - `if (!isActiveItemRef.current) return;` prevents a queued animation frame from writing dimensions if the trigger has since become inactive.
   - Uses a ref (`isActiveItemRef`) which is correct for reading current state inside an async callback. Double-checked: `isActiveItemRef` is defined via `useValueAsRef(isActiveItem)` at line 126.

3. **Whitespace-only change (line ~502)** — blank line addition, harmless.

### Test change (`NavigationMenuTrigger.test.tsx`, +92 -1)

- Chromium-only regression test (`it.skipIf(isJSDOM)`) — appropriate since this requires real layout measurements.
- Sets up two triggers with different panel sizes (700px vs 500px), hovers Product first, then switches to Solutions.
- Spies on `popupRoot.style.setProperty` and asserts that after the Solutions panel becomes visible, no `700px` width is written after the `500px` write. This directly tests the race condition.
- Final assertion checks the popup width resets to `auto` after animations finish, validating the full lifecycle.
- Test fixture and helper (`TestNavigationMenuRapidHoverSizing`, `getPopupWidthCalls`) are clean and well-scoped.

---

## Potential Concerns

| Area | Assessment |
|------|-----------|
| **Race condition coverage** | The two-pronged fix (cancel on deactivation + guard on callback) provides belt-and-suspenders protection. Solid. |
| **Effect type change** | Switching from `useEffect` to `useIsoLayoutEffect` is intentional and correct — cancellation must happen before paint. No SSR concern since the effect body only runs when `isActiveItem` is false (cleanup path). |
| **Bundle size** | +31B parsed, +7B gzipped. Negligible. |
| **Performance regressions** | CI benchmark shows some component mount regressions (Slider +16%, Select +10%) but these are unrelated to navigation menu and likely noise — the PR doesn't touch those components. |
| **Missing cleanup** | The `useIsoLayoutEffect` doesn't return a cleanup function, but that's fine — `sizeFrame` cleanup is already handled in the unmount effect at line ~358. |
| **Edge cases** | The guard uses `isActiveItemRef.current` which reads the latest value at callback time. If React batches multiple state updates, the ref will still reflect the final state correctly. |

---

## Verdict

This is a clean, minimal, and well-targeted bug fix. The root cause is clearly identified (queued animation frame outliving trigger activation), the fix addresses it at two levels (proactive cancellation + defensive guard), and the regression test is thorough and Chromium-only where appropriate. No API changes, no unnecessary refactoring.

## Confidence Rating: 4.5 / 5

Deducting half a point only because the race condition is timing-sensitive and the test relies on `setProperty` spy ordering, which could theoretically be flaky under extreme CI load — but the fix itself is sound and low-risk. Safe to merge.

PR #4649, opened 2026-04-20
[popups] Improve mount performance

atomiks:
This PR now focuses on the popup performance work that is still in this branch after rebasing on the interaction-split changes from `master`. It keeps popup coordination in the popup store, removes the remaining generic interaction wrapper paths, and limits popup-root interaction prop syncing to the times a popup is actually open or mounted.

## Changes

- Make popup stores own their synced `FloatingRootStore` setup, including floating ids, nesting, trigger maps, and shared popup selectors.
- Replace the remaining `useInteractions`/`useRole` wrapper paths with direct prop merging in popup roots, Select, Combobox, and Floating UI tests.
- Centralize popup trigger ownership and ARIA state for active triggers, inactive triggers, focusable popup props, and detached/multiple-trigger cases.
- Defer popup-root interaction prop syncing behind `open || mounted` where safe, while keeping Menu on its more guarded path.
- Keep trigger element lookups fast with a mirrored trigger element set in `PopupTriggerMap`.
- Add and update regression coverage for detached triggers, active trigger ownership, first-open/controlled-open behavior, open-method reset, and the Floating UI hook changes.

## Notes

- This no longer includes the older benchmark/experiment work or the broader cleanup that has since landed separately.
- Select and Combobox are touched only to remove `useInteractions` from their prop plumbing; the main perf work here is still the popup store/root path.


michaldudak:
## Critical (must fix before merge)

### C1 — Tooltip popup no longer sets `role="tooltip"`
**File**: [packages/react/src/tooltip/popup/TooltipPopup.tsx](packages/react/src/tooltip/popup/TooltipPopup.tsx)
With `useRole` deleted, PopoverPopup (`role: 'dialog'`) and MenuPopup (`role: 'menu'`) got hardcoded replacements; TooltipPopup did not. The new test in [TooltipRoot.test.tsx:38-55](packages/react/src/tooltip/root/TooltipRoot.test.tsx:38) asserts `aria-describedby` is gone from the trigger — so tooltips are now invisible to assistive tech. A11y regression with no loud failure.

### C2 — `TooltipTrigger` onClick handler is either dead or double-fires
**File**: [packages/react/src/tooltip/trigger/TooltipTrigger.tsx:143-147](packages/react/src/tooltip/trigger/TooltipTrigger.tsx:143)
```ts
onClick(event) {
  if (closeOnClick && !store.select('open')) {
    store.setOpen(false, createChangeEventDetails(REASONS.triggerPress, event.nativeEvent));
  }
},
```
Guard fires only when the tooltip is *not* open, then calls `setOpen(false)` — either a no-op, or (since `TooltipStore.setOpen` has no same-state early-return) a duplicate `onOpenChange` fire. The new "calls onOpenChange once" test at [TooltipRoot.test.tsx:891-914](packages/react/src/tooltip/root/TooltipRoot.test.tsx:891) likely passes by accident. Condition looks inverted.

### C3 — Render-time mutation of `store.state.floatingRootContext`
**File**: [packages/react/src/utils/popups/popupStoreUtils.ts:223](packages/react/src/utils/popups/popupStoreUtils.ts:223)
`usePopupRootSync` assigns `store.state.floatingRootContext = …` during render. Only the external-store branch calls `notifyAll()` inside a layout effect; non-external subscribers via `store.useState('floatingRootContext')` are not woken. Works today only because `useSyncedFloatingRootContext` returns a stable reference — any future change to that ref (e.g., Strict Mode double-render) silently desyncs subscribers. Use `store.set(...)` or `useSyncedValue` to make notification explicit.

### C4 — `openMethod` never cleared on unmount; detached-trigger click never writes it
- [popupStoreUtils.ts:236-240](packages/react/src/utils/popups/popupStoreUtils.ts:236) resets `openMethod` only when `!open`. If root unmounts while `open === true`, or a handle-based consumer has no mounted root, `openMethod` leaks to the next opener.
- [DialogTrigger.tsx:88](packages/react/src/dialog/trigger/DialogTrigger.tsx:88) and [PopoverTrigger.tsx:145](packages/react/src/popover/trigger/PopoverTrigger.tsx:145) route handle-equipped triggers to `rootTriggerProps`, which relies on a mounted root running `useOpenMethodTriggerProps`. Handle-only consumers get no `openMethod` written. Silent focus-style / iOS touch regression.

---

## Important

### I1 — MenuRoot rewritten with zero new Menu tests
[MenuRoot.tsx](packages/react/src/menu/root/MenuRoot.tsx) lost `useRole`, gained `usePopupRootSync`, gated dismiss/typeahead on `open`, yet `MenuRoot.test.tsx` and `MenuRoot.detached-triggers.test.tsx` are untouched. Gaps:
- Dismiss/typeahead now `enabled: open && !disabled` — while the close-animation plays, Escape / outside-press / typeahead silently do nothing. Inconsistent with `popupProps` memo at [MenuRoot.tsx:513](packages/react/src/menu/root/MenuRoot.tsx:513) which uses `open || mounted`.
- `aria-haspopup: 'menu'` and popup `aria-labelledby: activeTriggerId` moved from `useRole` to manual wiring — no tests assert these remain.
- [MenuRoot.tsx:103](packages/react/src/menu/root/MenuRoot.tsx:103) changed `open: defaultOpen` → `open: openProp ?? defaultOpen` (SSR-visible first-paint change) with no test.

### I2 — PreviewCard refactor entirely untested
No changes to preview-card test files. With `useRole` deleted, the `aria-expanded`/trigger ARIA path is gone and no replacement is visible in the PR. Also:
- [PreviewCardRoot.tsx:85-86](packages/react/src/preview-card/root/PreviewCardRoot.tsx:85) uses `dismiss.reference ?? EMPTY_OBJECT` — Dialog/Popover roots omit the fallback, creating an inconsistent contract (see I3).
- [PreviewCardStore.ts:122-134](packages/react/src/preview-card/store/PreviewCardStore.ts:122) conditionally calls `useSyncedValue('floatingRootContext', …)` only for external stores; no handle-based preview-card test exercises this.

### I3 — `inactiveTriggerProps = dismiss.trigger` may be undefined in Dialog/Popover
[useDialogRoot.ts:152-153](packages/react/src/dialog/root/useDialogRoot.ts:152) and the equivalent in [PopoverRoot.tsx:129-150](packages/react/src/popover/root/PopoverRoot.tsx:129) don't apply the `?? EMPTY_OBJECT` fallback. `useDismiss` returns `{}` when disabled, so current behavior may be safe, but the divergence from `PreviewCardRoot` signals unclear contract. Either remove the preview-card fallback or add it to Dialog/Popover.

### I4 — Stale `open` read in Menu triggers
[MenuTrigger.tsx:223-225](packages/react/src/menu/trigger/MenuTrigger.tsx:223) and [MenuSubmenuTrigger.tsx:161-163](packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx:161) call `useOpenMethodTriggerProps(store.select('open'), …)`. `store.select` doesn't subscribe, so the click handler sees a stale `open`. `DialogTrigger`/`PopoverTrigger` use `store.useState('open')` here — Menu should match.

### I5 — `usePopupRootSync` helper has no tests
[popupStoreUtils.ts:199-239](packages/react/src/utils/popups/popupStoreUtils.ts:199) is shared by Dialog/Popover/Menu/Tooltip roots but [popupStoreUtils.test.tsx](packages/react/src/utils/popups/popupStoreUtils.test.tsx) wasn't updated. The external-store `notifyAll()` path is entirely uncovered.

### I6 — Duplicated trigger-prop assembly across triggers
`aria-haspopup` / `aria-expanded` / `aria-controls` and the `controlsPopup = open && (isOpenedByThisTrigger || activeTriggerId == null || store.context.triggerElements.size === 1)` expression are copy-pasted across DialogTrigger, PopoverTrigger, MenuTrigger. AGENTS.md explicitly calls out avoiding duplication — extract into a shared hook next to `useTriggerDataForwarding`.

### I7 — Inconsistent Root context shapes after refactor
[DialogRootContext.ts](packages/react/src/dialog/root/DialogRootContext.ts) and [PopoverRootContext.ts](packages/react/src/popover/root/PopoverRootContext.ts) switched from `{ store }` to the store directly, but [MenuRootContext.ts](packages/react/src/menu/root/MenuRootContext.ts) still wraps `{ store, parent }`. Internal types only, but confusing to readers.

---

## Medium / Low

- **M1** — `MenuRoot.tsx:372` reads `store.state.floatingRootContext` directly; every other consumer uses `store.useState(...)`. Diverges from the pattern.
- **M2** — One-frame stale `triggerProps` after popup opens in Dialog/Popover (`*Interactions` renders only when `open || mounted`, so the first render missing dismiss handlers). Minor — one lost frame of outside-press handling.
- **L1** — `useOpenChangeComplete` now gated `mounted && !open && !preventUnmountingOnClose` ([popupStoreUtils.ts:186](packages/react/src/utils/popups/popupStoreUtils.ts:186)) — intentional but changes when the completion callback fires.
- **L2** — `usePopupRootSync` types `eventDetails` as `any` at [popupStoreUtils.ts:214](packages/react/src/utils/popups/popupStoreUtils.ts:214). Erodes type safety around reason codes. AGENTS.md discourages `as any` casts.
- **L3** — Dropped sentence in the iOS Safari comment at [useOpenInteractionType.ts:11-12](packages/react/src/utils/useOpenInteractionType.ts:11) — the "`interactionType` will be `''` in that case" clause explained *why* the `||` fallback trips. Restore it.
- **L4** — Nested `role: 'menuitem'` for nested-menu triggers previously supplied by `useRole` — not restored. Verify MenuSubmenuTrigger still produces the right role.
- **L5** — `useRole` previously used `getFloatingFocusElement(floatingElement)?.id || defaultFloatingId`; the replacements use plain `floatingId`. Minor edge case when a child owns focus.

packages/react/src/utils/popups/popupStoreUtils.ts
```diff
@@ -181,7 +195,7 @@ export function useOpenStateTransitions<State extends PopupStoreState<any>>(
   const preventUnmountingOnClose = store.useState('preventUnmountingOnClose');
 
   useOpenChangeComplete({
-    enabled: !preventUnmountingOnClose,
+    enabled: mounted && !open && !preventUnmountingOnClose,
     open,
```
atomiks: This line fixes unmount/teardown perf:

> 
Contained triggers warm unmount (all) | 119.95 ms | 17.25 ms | -102.7 ms (-85.6%)
-- | -- | -- | --

packages/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -17,7 +17,7 @@ export const IsDrawerContext = React.createContext(false);
  *
  * Documentation: [Base UI Dialog](https://base-ui.com/react/components/dialog)
  */
-export function DialogRoot<Payload>(props: DialogRoot.Props<Payload>) {
+export const DialogRoot = function DialogRoot<Payload>(props: DialogRoot.Props<Payload>) {
```
romgrk: Is this useful?

packages/react/src/dialog/store/DialogStore.ts
```diff
@@ -109,12 +123,29 @@ export class DialogStore<Payload> extends ReactStore<
     externalStore: DialogStore<Payload> | undefined,
     initialState?: Partial<State<Payload>>,
   ) {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
romgrk: You can also wrap the method body with enable/disable eslint comments.

packages/react/src/drawer/root/DrawerRoot.tsx
```diff
@@ -427,9 +427,9 @@ function DrawerProviderReporter() {
   const providerContext = useDrawerProviderContext(true);
   const dialogRootContext = useDialogRootContext(false);
 
-  const open = dialogRootContext.store.useState('open');
-  const nestedOpenDialogCount = dialogRootContext.store.useState('nestedOpenDialogCount');
-  const popupElement = dialogRootContext.store.useState('popupElement');
+  const open = dialogRootContext.useState('open');
+  const nestedOpenDialogCount = dialogRootContext.useState('nestedOpenDialogCount');
+  const popupElement = dialogRootContext.useState('popupElement');
```
romgrk: Rename the variable to `store`? It would match the rest of the codebase.

packages/react/src/dialog/root/DialogRoot.tsx
atomiks: No, was a leftover when I tried to put `fastComponent` around `Dialog`. Apparently it slowed it down but makes the other ones better

romgrk (commented):
It would have been really nice to split this PR into conceptual chunks, reviewing 2k+ lines is not really fun nor safe, and we can't see the individual effect of each change (e.g. adding `fastComponent()`, avoiding work before mount, ...).

atomiks:
@romgrk yeah totally fair 😅 I'll split this into multiple PRs so it's easier to review and we can see the perf differences incrementally. At least I got it into (what I think should be) mostly the final state for perf + bundle size though so it shouldn't be that hard to split it from here

packages/react/src/dialog/root/useDialogRoot.ts
```diff
@@ -38,17 +34,40 @@ export function useDialogRoot(params: UseDialogRootParameters): UseDialogRootRet
...
+  parentContext?: DialogStore<unknown>['context'] | undefined;
+  isDrawer: boolean;
+}
+
+export interface UseDialogRootReturnValue {
+  parentContext: DialogStore<unknown>['context'] | undefined;
+  isDrawer: boolean;
+}
+
+export function DialogInteractions({
```
flaviendelangle, line 57: Could be worth using `fastComponent` on this one since it's used a lot and has at least 5 selectors

atomiks: It only affects opening, so not really. One single dialog opening is not a perf cliff. 

(Also I'm drafting this PR, as we should start with a review on #4661, we'll get back to this once that lands)

atomiks:
Creating separate PRs for:

- Deleting `useRole` + `useInteractions`
- Improving Menu perf

PR #4651, opened 2026-04-20
[docs] Highlight Releases sidenav item on version subpages

LukasTy:
## Summary

The docs sidebar used an exact pathname match to decide the active item, so visiting a version subpage such as `/react/overview/releases/v1-4-0` left the "Releases" entry inactive — readers lost their place in the nav.

This updates `SideNav.Item` to also treat a nav item as active when the current pathname is a descendant of its `href`:

Screenshot 2026-04-20 at 12 36 09

```ts
const active = pathname === href || pathname.startsWith(`${href}/`);
```

The trailing `/` guard prevents spurious matches (e.g. `releases-archive` would not count as a child of `releases`).

## Why this is safe

All other sidebar sections (Overview, Handbook, Components, Utils) list flat leaf pages — no sibling href is a prefix of another, and `releases/` is the only section with rendered subpages that are not themselves sidebar items. External items use full URLs, so the new `startsWith` check cannot match app pathnames.

## Test plan

- [x] Visit `/react/overview/releases` — "Releases" highlighted.
- [x] Visit `/react/overview/releases/v1-4-0` — "Releases" highlighted (`data-active` / `aria-current` set).
- [x] Visit `/react/overview/releases/v1-0-0-beta-7` — "Releases" highlighted.
- [x] Visit `/react/overview/quick-start` — only "Quick start" highlighted.
- [x] Visit `/react/components/accordion` — only "Accordion" highlighted.
- [x] No console errors.


michaldudak:
## Codex Review (GPT-5.5)
I reviewed PR #4651 as a docs-only/react-ui change against freshly fetched `upstream/master`. The sidebar highlight fix is directionally right, but it broadens `active` in a way that also changes link/navigation behavior, not just styling.

### 1. Bugs / Issues

#### 1. 🟡 Descendant pages inherit exact-page link behavior (non-blocking)

`active` now covers both the exact link target and descendant routes, but the same value also controls `scroll={false}` and the “already on this page” click handler.

`docs/src/components/SideNav.tsx`
```tsx
const active = pathname === href || pathname.startsWith(`${href}/`);

scroll={external ? undefined : !active}

...(active
  ? {
      'aria-current': true,
      'data-active': true,
      onClick: () => {
        window.scrollTo({ top: 0, behavior: 'smooth' });
      },
    }
  : {})
```

On `/react/overview/releases/v1-4-0`, the Releases item is visually active, but clicking it is still a real navigation to `/react/overview/releases`. This branch now treats that click like an exact-current-page click: it disables Next’s normal scroll handling and runs the smooth scroll handler intended for the already-current page.

**Fix:** Split the concepts, e.g. `exactActive = pathname === href` and `sectionActive = exactActive || pathname.startsWith(`${href}/`)`; use `sectionActive` for `data-active`, but keep `scroll` and the current-page click behavior tied to `exactActive`. Consider whether `aria-current` should also remain exact-only.

### Accuracy and Completeness Assessment

The slash-delimited prefix check avoids the common `/foo` versus `/foobar` false positive and matches the release-version URL shape produced by `ReleaseTimeline`.

### Validation Coverage

I checked the full PR diff and ran `git diff --check upstream/master...refs/remotes/upstream/pr/4651` plus `pnpm eslint docs/src/components/SideNav.tsx`; both passed. I did not run a browser repro.

### Recommendation

The visible highlighting behavior is a reasonable fix, but the active state currently carries extra behavior. I’d approve once the visual section-active state is decoupled from exact-page navigation behavior.

michaldudak:
I updated the PR and I noticed one issue: we're removing the `cursor: pointer` and underline on hover for active navbar links, but they make sense for nested pages like releases. @aarongarciah, how would you solve it? Can we make this link look more clickable?

aarongarciah:
@michaldudak it's fine if we bring back `cursor: pointer` and underline on hover for active links IMO.

michaldudak (approved):

PR #4661, opened 2026-04-22
[popups] Improve mount performance with interaction splitting

atomiks:
This is a simpler version of the earlier #4649.

It keeps the interaction split that reduces closed-mount work (and second render) in Tooltip, Popover, Dialog, Drawer, and Preview Card, along with the supporting fixes needed to ensure tests pass. It does not include the further follow-up improvements from that PR, especially the extra Menu work.

Because this is the smaller branch, **it is not the final bundle-size**. Some shared popup helper code is added here without yet removing all of the older popup interaction logic. The broader cleanup branch can be rebased on top of this once it lands on master, which should reduce how much needs to be reviewed again.

## Changes

- Split popup interaction setup out of the root path so closed popups skip that work until they are open or mounted.
- Includes teardown/unmount perf fix.
- Adds `fastComponent`/`fastComponentRef` around most of the rest of the components. Benchmarks below show this is a small improvement in a real browser env; we could probably remove it everywhere except the Menu and not lose much
- Update the contained-triggers perf experiment for the smaller branch, including Preview Card and the unmount measurement cleanup.

## Benchmarks

The table below uses the docs contained-triggers experiment and compares `upstream/master` against this branch.

| Benchmark | Baseline | Branch | Delta |
| --- | ---: | ---: | ---: |
| Contained triggers mount (all: tooltip + popover + dialog + menu + preview card) | 165.50 ms | 115.84 ms | -49.66 ms (-30.0%) |
| Contained triggers unmount (all: tooltip + popover + dialog + menu + preview card) | 168.76 ms | 21.58 ms | -147.18 ms (-87.2%) |
| Contained triggers mount (tooltip only) | 20.36 ms | 10.40 ms | -9.96 ms (-48.9%) |
| Contained triggers mount (popover only) | 23.78 ms | 11.90 ms | -11.88 ms (-50.0%) |
| Contained triggers mount (dialog only) | 22.20 ms | 11.12 ms | -11.08 ms (-49.9%) |
| Contained triggers mount (menu only) | 32.80 ms | 30.66 ms | -2.14 ms (-6.5%) |
| Contained triggers mount (preview card only) | 18.52 ms | 9.06 ms | -9.46 ms (-51.1%) |

I also compared the current branch against the same code without the `fastComponent` / `fastComponentRef` wrappers used here, using the same docs experiment for the single-popup rows below.

| Benchmark | Without `fastComponent` | With `fastComponent` | Delta |
| --- | ---: | ---: | ---: |
| Contained triggers mount (tooltip only) | 12.42 ms | 10.74 ms | -1.68 ms (-13.5%) |
| Contained triggers mount (popover only) | 13.36 ms | 12.58 ms | -0.78 ms (-5.8%) |
| Contained triggers mount (menu only) | 35.34 ms | 32.20 ms | -3.14 ms (-8.9%) |
| Contained triggers mount (dialog only) | 11.90 ms | 11.64 ms | -0.26 ms (-2.2%) |


michaldudak:
# PR Review Summary — `[popups] Improve mount performance with interaction splitting`

Synthesized from five specialized review passes (code, silent-failure, tests, comments, types).

## Critical issues — confirm before merge

**1. `Escape` on inactive Dialog / AlertDialog triggers no longer dismisses** (~82% confidence)
- `packages/react/src/dialog/root/useDialogRoot.ts:152`
- On `master`, `useInteractions([role, dismiss])` merged `dismiss.trigger` into props for *inactive* triggers, so any focused detached trigger could close the popup with `Escape`. The new `DialogInteractions` sets `inactiveTriggerProps = EMPTY_OBJECT` while the parallel `PopoverInteractions` correctly preserves `dismiss.trigger ?? EMPTY_OBJECT`. Likely a real regression for non-modal dialogs and detached-trigger setups.
- **Fix:** mirror the Popover refactor — `inactiveTriggerProps = dismiss.trigger ?? EMPTY_OBJECT` in `DialogInteractions`.

**2. `triggerCount` is read non-reactively from a mutable `Map`**
- `packages/react/src/popover/trigger/PopoverTrigger.tsx:60`, `packages/react/src/dialog/trigger/DialogTrigger.tsx:48`
- `store.context.triggerElements.size` is captured during render, but `PopupTriggerMap.add/delete` does not notify the store. The `triggerOwnsOpenPopupOrIsOnlyTrigger` fallback (which gates `aria-controls` for the lone-trigger / no-`activeTriggerId` case) relies on `triggerCount === 1` and so is unreachable in the most common first-render path. Today the bug is masked because `useImplicitActiveTrigger` writes `activeTriggerId` in a follow-up layout effect — but if a second trigger mounts later without other state changes, the first trigger keeps a stale `triggerCount = 1` and ARIA ownership goes wrong silently.
- **Fix:** track trigger count inside the reactive store (write to state in `useTriggerRegistration`), or remove the fallback and rely solely on `useImplicitActiveTrigger`.

## Important issues

**3. `openMethod` reset switched from `useIsoLayoutEffect` to `useEffect`**
- `packages/react/src/utils/popups/popupStoreUtils.ts:253-272` — `usePopupRootSync`
- The previous `useOpenInteractionType` reset `openMethod` synchronously after render via `useValueChanged` + `useIsoLayoutEffect`. The replacement uses `React.useEffect`, which runs after paint. On a close→reopen sequence in the same task (rapid taps on iOS), `openMethod` carries the stale value into the next open's `defaultInitialFocus` decision in `DialogPopup.tsx:73-78`. Symptom: wrong initial focus for touch reopens.
- **Fix:** switch to `useIsoLayoutEffect`.

**4. `useFloatingRootContextSync` mutates `store.state` during render**
- `packages/react/src/utils/popups/popupStoreUtils.ts:285-300`
- `(store.state as State).floatingRootContext = floatingRootContext` bypasses `Store.set`, so subscribers are only woken when `notifyOnChange === true` (Dialog with `externalStore` only). Internal-store callers see a silent mutation: future code that recreates the floating context will desync subscribers without warning. Also breaks the `Store.state` "do not modify directly" contract.
- **Fix (preferred):** route through `store.set('floatingRootContext', …)`. **Or:** narrow the parameter type so the unsafe-without-`notifyOnChange` case can't be expressed, and add a comment.

**5. `PopupStoreWithOpen` and `useSyncedFloatingRootContext` use `any` in their generics**
- `packages/react/src/utils/popups/popupStoreUtils.ts:26-28`, `packages/react/src/floating-ui-react/hooks/useSyncedFloatingRootContext.ts:13`
- `setOpen(open, eventDetails: any)` and `PopupStoreContext<any>` accept any `setOpen` whose second arg is anything, so the handoff to `BaseUIChangeEventDetails<string>` is unchecked.
- **Fix:** parameterize `PopupStoreWithOpen<State, EventDetails>` and add `<EventDetails extends BaseUIChangeEventDetails<string>>` to `useSyncedFloatingRootContext`. ~10 lines, removes two `any`s.

## Test coverage gaps worth filling

**6. Tooltip `trackCursorAxis !== 'none'` keeps interactions live while closed**
- `TooltipRoot.tsx:140` added the `trackCursorAxis !== 'none'` clause to the gate; nothing in the tooltip test suite uses `trackCursorAxis`. A future refactor can drop the clause silently.
- **Suggested:** Chromium-only test that hovers a `Tooltip` with `trackCursorAxis=\"x\"` and asserts the positioner tracks the cursor on the *first* open.

**7. Dialog handle recreation while open**
- The new `useFloatingRootContextSync(notifyOnChange)` path is only used by Dialog's external-store flow. The existing \"Fast Refresh-like handle recreation\" tests cycle through closed states. Add a case that swaps the handle *while the dialog is open* and asserts `trigger.aria-controls === popup.id` after the swap.

**8. Single detached trigger, controlled-open, no `triggerId` prop** (relates to issue 2)
- The trigger-claim path inside `useTriggerDataForwarding` is exercised only when `triggerId`/`defaultTriggerId` is set. Render `<Popover.Root open><Popover.Trigger>…` with no `triggerId` and assert `aria-controls` lands.

**9. Hover-to-open via detached trigger**
- Detached-trigger suites only cover click. With interaction splitting, first-hover wiring is the most likely class of regression.

## Minor / suggestions

- **`useSyncedFloatingRootContext` always allocates an unused `internalStore`** when `floatingRootContextProp` is supplied (`useSyncedFloatingRootContext.ts:53-66`) — small perf win available, skip the `useRefWithInit` when the prop is present.
- **`useFloatingRootContextSync` has exactly one caller** (DialogStore) — inline it or apply to the other three stores.
- **Repeated 4× boilerplate** in tooltip/popover/dialog/preview-card store constructors creating `triggerElements` + `floatingRootContext`. A `createPopupStoreFoundation()` helper would lock the pairing invariant.
- **`setOpenTriggerState` mutates a parameter** (`popupStoreUtils.ts:98-111`) — return a partial instead, eliminates a `as Partial<State>` cast at both call sites.
- **`usePopupRootSync` constraint** uses `openMethod: unknown` — should be `openMethod: InteractionType | null`.
- **`FOCUSABLE_POPUP_PROPS`** uses `as` instead of `satisfies`.
- **JSDoc on `triggerPopupId` selector** at `store.ts:216-218` doesn't mention the `triggerCount === 1` fallback branch — and as noted in issue 2, callers must pass `triggerCount` for that branch to fire.
- **Stale parenthetical** in `popupStoreUtils.ts:149-152`: \"(or will immediately open via defaultOpen)\" — there's no `defaultOpen` branch in this function.
- **`setOpenTriggerState` comment** says \"may be null\" but the parameter is `Element | undefined`.

## Strengths

- The `popupStoreUtils.test.tsx` unit tests pin the bug fixes from the popup id syncing and trigger ownership commits directly — including the explicit `popupElement.id` overrides `floatingId` precedence used by all popup components.
- The four parallel \"synchronizes ARIA attributes in controlled mode\" tests across AlertDialog/Dialog/Popover/Drawer pin the cross-component contract — exactly the regression mode the refactor risks.
- `PopupFloatingRootContextOptions` + `createPopupFloatingRootContext` centralize a previously-duplicated 11-line `syncOnly: true` setup. New selectors (`popupId`, `triggerPopupId`, `triggerOwnsOpenPopup`) are well-named and replace ad-hoc inline logic in `PopoverTrigger`/`DialogTrigger`.
- AGENTS.md compliance is clean: no new `as any`, no native `setTimeout`/`requestAnimationFrame`, `useStableCallback`/`useIsoLayoutEffect` used correctly, Vitest-only matchers in tests, no public error changes.
- `?? EMPTY_OBJECT` fallbacks for `dismiss.*` are correct — `useDismiss` returns `{}` when disabled, and `EMPTY_OBJECT` keeps stable identity for `useSyncedValues`.

## Recommended action

1. **Fix 1 (Dialog Escape regression)** — small change, prevents a behavior regression that current tests don't catch.
2. **Decide on 2 (`triggerCount` reactivity)** — either fix the reactivity or drop the dead fallback branch. Don't ship the silently-unreachable code.
3. **Fix 3 (`useEffect` → `useIsoLayoutEffect`)** — one-line restore of prior timing semantics.
4. **Address 4 properly or document why it's safe** — at minimum a comment near the render-time mutation; ideally route through `store.set`.
5. **Tighten generics in 5** — small, high-leverage cleanup before more code piles on top.
6. Add tests 6, 7, 8 to lock in the changed behaviors.
7. Defer the comment nits, single-caller helper inlining, and store constructor dedup to a follow-up — they're code-health items, not blockers.

michaldudak:
## PR Review

No critical issues found — the interaction split preserves previously-tested invariants and the new tests cover the regression set the PR was scoped to fix.

### Important Issues

**1. Test gap: no test asserts the conditional-mount behavior itself**

The whole PR hinges on `shouldRenderInteractions = open || mounted`. No test verifies that listeners (outside press, Escape) are actually inert against a closed-and-unmounted popup. A regression that inverted the gate would silently still pass every test.

A single high-value test (e.g. on Popover): render closed `Popover.Root`, dispatch a document-level `mousedown`, assert `onOpenChange` is not invoked.

Files: `packages/react/src/popover/root/PopoverRoot.tsx:103`, `packages/react/src/dialog/root/DialogRoot.tsx:81`, `packages/react/src/tooltip/root/TooltipRoot.tsx:140`, `packages/react/src/preview-card/root/PreviewCardRoot.tsx:88`

**2. Test gap: floating-root-context handoff on first-frame controlled `open`**

The dialog/popover/preview-card stores eagerly construct a standalone `FloatingRootStore`, then `useFloatingRootContextSync` swaps it for the synced one when interactions mount. A `<Popover.Root open>` + immediate Escape test would guard the most subtle regression vector this structural change introduces. AlertDialog has this; Popover/PreviewCard/Drawer do not.

`packages/react/src/dialog/store/DialogStore.ts:138`, `packages/react/src/utils/popups/popupStoreUtils.ts:298`

**3. Test gap: Tooltip `trackCursorAxis` / `disabled` branches of `shouldRenderInteractions`**

The new test covers `trackCursorAxis !== 'none'` first-hover positioning, but never asserts the inverse fallbacks (`trackCursorAxis="none"` closed-and-unmounted hover-open at trigger anchor, `disabled=true` short-circuit). A `||` ↔ `&&` swap would not be caught.

`packages/react/src/tooltip/root/TooltipRoot.tsx:140`, `packages/react/src/tooltip/root/TooltipRoot.test.tsx:916`

**4. Duplicate test in `PopoverRoot.detached-triggers.test.tsx`**

`should open the popover with any trigger` is defined twice (lines 17-55 and 57-95) with identical bodies — likely a copy-paste. Vitest runs both but the second shadows reporting.

**5. `user: any` in test helpers violates the "no unnecessary casts" rule**

`packages/react/src/popover/root/PopoverRoot.detached-triggers.test.tsx:361`, `packages/react/src/dialog/root/DialogRoot.detached-triggers.test.tsx:333`, `packages/react/src/alert-dialog/root/AlertDialogRoot.test.tsx:420`. Use `UserEvent` from `@testing-library/user-event`.

### Suggestions

**6. `useFloatingRootContextSync` is dead code in `DialogStore.useStore`**

It writes `floatingRootContextProp ?? internalStore` back into the store, but the prop is the same reference, so `Store.set` no-ops on `Object.is` equality. Tooltip and PreviewCard correctly skip it. Either remove the call at `packages/react/src/dialog/store/DialogStore.ts:141` or document why it exists.

**7. `usePopupStore` allocates an unused `internalStore` when an external store is passed**

`packages/react/src/utils/popups/popupStoreUtils.ts:47` and `packages/react/src/floating-ui-react/hooks/useSyncedFloatingRootContext.ts:65` always construct one, including a `FloatingRootStore` and `PopupTriggerMap`. Wasted memory per detached-trigger render — gate `useRefWithInit` on `externalStore === undefined`.

**8. `syncTriggerCount` fires twice in same frame when an id is replaced**

`packages/react/src/utils/popups/popupStoreUtils.ts:99-111` — the unregister-then-register flow notifies subscribers twice synchronously (delete drops by 1, add restores it). Functionally correct, just an extra render. Compute new count once at end of the callback.

**9. Pre-existing: `instantType` widens to `'focus'` at runtime but not in type**

`packages/react/src/popover/store/PopoverStore.ts:171` assigns `'focus'` though the union at line 25 is `'dismiss' | 'click' | undefined`. Pre-dates this PR but worth surfacing while the area is open.

**10. Test gap: trigger replacement during open state**

Reparenting tests cover wrapper add/remove; none covers swapping one trigger element for another with controlled `triggerId` while the popup is open, which exercises the new `triggerCount`-reactive `aria-controls` claim path at `packages/react/src/utils/popups/popupStoreUtils.ts:160-175`.

michaldudak:
## Codex Review (GPT-5.5)

Reviewed the full PR diff against freshly fetched `upstream/master` with PR head `14e39f6`. This is a performance-led popup interaction refactor with secondary test and perf-experiment changes.

### 1. Bugs / Issues

#### 1. 🔴 Tooltip can keep stale cursor-tracking handlers after tracking is disabled (blocking)

`packages/react/src/tooltip/root/TooltipRoot.tsx`

```tsx
const shouldRenderInteractions = open || mounted || (!disabled && trackCursorAxis !== 'none');

{shouldRenderInteractions && (
  <TooltipInteractions store={store} disabled={disabled} trackCursorAxis={trackCursorAxis} />
)}
```

```tsx
store.useSyncedValues({
  activeTriggerProps,
  inactiveTriggerProps,
  popupProps,
});
```

When a closed tooltip changes `trackCursorAxis` from a tracking value back to `'none'`, `TooltipInteractions` unmounts, but the store still contains the previous `clientPoint` trigger props. Those stale mouse handlers can continue setting `positionReference` to a cursor virtual element, so the next open may still position from the cursor even though tracking is disabled.

**Fix:** Clear the synced trigger/popup props on `TooltipInteractions` cleanup, or keep the interaction component mounted long enough with `enabled: false` to overwrite the store with non-client-point props.

#### 2. 🔴 Unmount benchmark undercounts cleanup work (blocking)

`docs/src/app/(private)/experiments/perf/utils/benchmark.tsx`

```tsx
const finish = () => {
  resolved = true;
  observer?.disconnect();
  settleTimeout.clear();
  resolve(Math.max(0, lastMutationAt - start));
};
```

For `measureUnmount`, the benchmark records time until the last observed DOM mutation, not until teardown cleanup has actually finished. Passive effect cleanup, listener removal, timer disposal, store unsubscriptions, and other non-mutating work can run after or without a mutation, so the unmount numbers can make cleanup look much cheaper than it is.

**Fix:** Measure through `finish` time for unmount mode, or relabel this as DOM-mutation settle time and avoid using it as cleanup/unmount evidence.

#### 3. 🟡 Lazily-created internal Popover stores miss timeout disposal (non-blocking)

`packages/react/src/utils/popups/popupStoreUtils.ts`

```ts
if (externalStore === undefined && internalStoreRef.current === null) {
  internalStoreRef.current = createStore(floatingId, nested);
}

return { store, internalStore: internalStoreRef.current };
```

`packages/react/src/popover/store/PopoverStore.ts`

```ts
const { store, internalStore } = usePopupStore(...);

useOnMount(() => internalStore?.disposeEffect());
```

If `Popover.Root` first receives an external `handle` store and later drops it, `usePopupStore` creates the internal store after the one-time `useOnMount` callback already ran with `internalStore === null`. That new internal store’s `stickIfOpenTimeout` is then never disposed on unmount.

**Fix:** Use an effect keyed on `internalStore`, or keep Popover’s internal store eagerly created.

#### 4. ℹ️ New ARIA assertions can pass when both ids are missing (informational)

Several new tests use this pattern:

```tsx
expect(trigger.getAttribute('aria-controls')).toBe(popup.getAttribute('id'));
```

That passes when both sides are `null`, so it would not catch a regression in the new popup id / trigger ownership sync. The closed-trigger “does not notify” tests also spy only on `store.set`, which would miss `update`, `setState`, or `notifyAll`.

**Fix:** Assert a real popup id first and use `toHaveAttribute('aria-controls', popup.id)`; for no-notification tests, assert via `store.subscribe(vi.fn())` or spy on `setState`.

### Benchmark / Performance Assessment

The interaction split itself is directionally reasonable, but the current benchmark evidence needs tightening before the unmount claim is reliable. I also noticed the `PopoverRoot` `fastComponent` wrapper mostly covers the outer shell, while the subscribed `PopoverRootComponent` still renders as a normal child component; that makes the `fastComponent` contribution easy to overstate unless it is benchmarked separately.

### Test Coverage Assessment

Validated locally: `pnpm typescript`, `pnpm test:jsdom popupStoreUtils --no-watch`, `pnpm test:jsdom ReactStore --no-watch`, and Chromium suites for `TooltipRoot`, `PopoverRoot.detached-triggers`, `DialogRoot.detached-triggers`, `AlertDialogRoot`, and `DrawerRoot`.

### Recommendation

Request changes 🔴

The implementation is close, and the targeted suites pass, but the stale tooltip cursor handlers are a real user-facing regression and the unmount benchmark does not yet substantiate one of the PR’s central performance claims.

michaldudak:
# PR Review Summary

Reviewed `pr-4661` (36 files, +1309 / −384) against base `4597333a`. Four review passes ran in parallel: general code review, silent-failure hunt, test coverage, and comments. Findings consolidated and de-duplicated below.

## Critical Issues

None. The interaction split looks correct: closed popups skip dismiss/role/click setup, the parent re-renders interactions when `open || mounted`, and the bug-fix commits (cursor staleness, trigger ID sync, popup ownership) are appropriate.

## Important Issues (worth addressing before merge)

1. **Stale `triggerCount` after closed-time register/unregister** — `packages/react/src/utils/popups/popupStoreUtils.ts:228-250` — `useTriggerRegistration` skips `syncTriggerCountIfOpen` while closed and `useImplicitActiveTrigger` early-returns when `!open`. A controlled-open popup whose first effect run is delayed can read stale `state.triggerCount`, causing `triggerOwnsOpenPopupOrIsOnlyTrigger` to misreport `aria-controls`. The new test passes only because of `waitFor`. Fix: reset `triggerCount` to 0 on close, or sync it unconditionally.
2. **Stale `data-instant` JSDoc / docs** — three sites disagree on the union after this PR widened it to include `'focus' | 'trigger-change'`:
   - `packages/react/src/popover/popup/PopoverPopupDataAttributes.ts:32`
   - `packages/react/src/popover/viewport/PopoverViewportDataAttributes.ts:23`
   - `docs/src/app/(docs)/react/components/popover/types.md:313` and `:480` — the `State` blocks were updated, the data-attribute rows weren't. Run `pnpm docs:api` after fixing.
3. **`cancelPendingOpen` only routes through floating events, not the popup store** — `packages/react/src/tooltip/store/TooltipStore.ts:131-136` — calls `floatingRootContext.dispatchOpenChange(false)` which doesn't invoke the popup store's `onOpenChange`. If hover is disabled, `useHoverReferenceInteraction` short-circuits its subscription, so a click on a focus-only-opened tooltip would not cancel a pending open. Verify the disabled-hover code-path or replace with `store.setOpen(false, ...)`.
4. **Test gap — full close→reopen cycle for Popover/Dialog** — there is no test that opens, asserts the popup unmounts, then reopens and verifies dismiss/Escape/outside-click still work. The new `*Interactions` subtree must remount and re-wire on the second open; this is the most likely regression class introduced by the refactor.
5. **Test gap — `openMethod` reset across mixed-method open cycles** for Popover/Dialog (Menu/Select have a dedicated `interaction type tracking` describe; these don't). Mouse-then-keyboard reopen could carry stale `openMethod` and produce wrong initial focus.
6. **Test gap — Tooltip reopen with stale cursor position**. The two new cursor-tracking tests cover first-hover and disable-while-closed but not the very scenario commit `6f4dbd2e1` ("Avoid stale cursor tracking") fixes: open→close→move→reopen with `trackCursorAxis="x"`.
7. **Removed multi-trigger Popover test** — `packages/react/src/popover/root/PopoverRoot.detached-triggers.test.tsx` — the contained-triggers `'should open the popover with any trigger'` was deleted; only the detached-handle equivalent remains. Sequential opens through different contained triggers go through a meaningfully different path post-refactor; re-add an equivalent.

## Suggestions

- **`useSyncedFloatingRootContext` allocates an unused `useId`** when callers pass `floatingId` — wasted work in a hook this PR is trying to streamline. `packages/react/src/floating-ui-react/hooks/useSyncedFloatingRootContext.ts:48`
- **`internalStoreRef` ↔ `externalStore` switch is fragile** — the lazy init only fires when the ref is null, so a defined→undefined→defined toggle reuses a stale internal store. Pre-existing across all popup stores; document the controlled/uncontrolled constraint or add a runtime warning. `packages/react/src/utils/popups/popupStoreUtils.ts:46-51`
- **`useTriggerDataForwarding` skips `stateUpdates` when popup is closed** — a re-mounted active trigger with changed `payload`/`closeOnClick`/`closeDelay` keeps stale data until next open. `packages/react/src/utils/popups/popupStoreUtils.ts:175-181`
- **`usePopupRootSync` unmount cleanup writes to a possibly-shared external store** — could wipe in-flight `openMethod` for handle-shared roots. `packages/react/src/utils/popups/popupStoreUtils.ts:311-318`
- **Nested-dialog count notification semantics changed** — `onNestedDialogOpen/Close` now only run while local dialog is `open || mounted`. Worth calling out in the PR description. `packages/react/src/dialog/root/useDialogRoot.ts:122-150`
- **JSDoc / comment drift**:
  - `packages/react/src/utils/popups/popupStoreUtils.ts:218-220` — `@param open` documented but signature only takes `store`.
  - `packages/react/src/floating-ui-react/hooks/useSyncedFloatingRootContext.ts:28-31` — JSDoc only describes the lazy-create branch; the new caller-provides branch isn't mentioned.
  - `packages/react/src/floating-ui-react/hooks/useClientPoint.ts:223` — the standalone unmount-only `useEffect(() => resetReference, ...)` deserves a one-line WHY (the reason it can't fold into the main effect cleanup is non-obvious).
  - `packages/react/src/tooltip/store/TooltipStore.ts:131-136` — one-line comment on `cancelPendingOpen` would help (only caller is `TooltipTrigger.onClick` with `closeOnClick`).
  - Re-evaluate the `// TODO` at `packages/react/src/floating-ui-react/hooks/useSyncedFloatingRootContext.ts:103` — this PR moves significantly closer to obsoleting it.
  - Typos: `interactling` at `packages/react/src/utils/popups/store.ts:40`; `Mangages` at `packages/react/src/utils/popups/popupStoreUtils.ts:254`.
- **Test gaps (lower priority)**:
  - `triggerCount` reactivity inverse: open popup → second trigger registers → implicit-active should un-set.
  - Popover detached-trigger ARIA round-trip on `floatingId` regeneration (Dialog has it; Popover doesn't).
  - Drawer `aria-controls` after close→reopen and with `triggerId` / detached trigger.
  - More direct assertion on `cancelPendingOpen` invocation (current test passes via timer-clearing path).

## Strengths

- Clean, consistent split: `shouldRenderInteractions = open || mounted` is readable and applied uniformly. Tooltip's `trackCursorAxis !== 'none'` extension makes sense.
- New `triggerPopupId` selector + `triggerOwnsOpenPopupOrIsOnlyTrigger` helper drive `aria-controls` from the store without depending on popup mount; well unit-tested.
- `useTriggerRegistration` + `triggerCount` skip-while-closed is a good optimization, and the "registers and unregisters closed triggers without notifying the store" test pins it down precisely.
- Detached-trigger tests for `defaultOpen` + handle recreation while open + escape from inactive detached trigger materially improve confidence.
- `useClientPoint` reset paths fix two real staleness bugs.
- `ReactStore.useControlledProp`'s `super.setState` → `store.setState` change is subtle but correct (closure vs. prototype binding) and the new test pins it.
- `useOpenInteractionType` split into `useOpenMethodTriggerProps` is the right shape.
- `fastComponent`/`fastComponentRef` wrappers are a clean perf win without component-specific behavior.

## Recommended Action

1. Fix the docs/JSDoc drift on `data-instant` (`#2`) and run `pnpm docs:api`. Quick.
2. Address the stale-`triggerCount` issue (`#1`) and verify `cancelPendingOpen`'s click-cancel path on focus-only tooltips (`#3`).
3. Add 2-3 tests covering the lazy-mount lifecycle (`#4`–`#7`) — full close-then-reopen for Popover/Dialog dismiss, Tooltip cursor-staleness across reopens, and a multi-trigger Popover sequence.
4. Address suggestions opportunistically; most are minor.

Overall this is a high-quality, high-impact change with strong test coverage on the bug-fix commits. The remaining gaps cluster around the new lazy lifecycle itself rather than any single defect.

packages/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -63,22 +63,31 @@ export function DialogRoot<Payload>(props: DialogRoot.Props<Payload>) {
...
 
   return (
     <IsDrawerContext.Provider value={false}>
       <DialogRootContext.Provider value={contextValue as DialogRootContext}>
+        {shouldRenderInteractions && (
+          <DialogInteractions
+            store={store}
+            parentContext={dialogRootResult.parentContext}
+            isDrawer={dialogRootResult.isDrawer}
+          />
```
romgrk: Might be nice to pass the `dialogRootResult` as a single prop directly rather than in multiple props. Right now it exposes variables/fields that could stay internal to `useDialogRoot.ts` instead. It would also make maintainance/updates easier if that shape changes. The name of that shape's variable could use an update.

michaldudak (approved):

PR #4662, opened 2026-04-22
[react] Use floating document for virtual arrow

lyzno1:
## Summary

- create the virtual arrow element from the floating element's owner document

## Why

When no arrow element is mounted, the fallback element should use the same document as the positioned floating element instead of falling back through a null arrow ref.

## Checks

- pnpm typescript
- pnpm eslint packages/react/src/utils/useAnchorPositioning.ts
- pnpm prettier --check packages/react/src/utils/useAnchorPositioning.ts


atomiks (approved):

atomiks:
Thanks for the contribution @lyzno1

PR #4665, opened 2026-04-22
[dialog] Fix scroll lock handoff with external overlays

### LINKED ISSUE #4640: [Dialog] scroll locking mechanism doesn't handoff well with non-base ui modal overlays
# Bug report

## Current behavior

When opening a base ui `Dialog` from another non-base ui modal overlay, it is possible that the page's scroll ends up in an unrecoverable locked state:

For example, [Ariakit `Menu`](https://ariakit.com/reference/menu) adds a delay to its "scroll lock cleanup" to wait for animations/transitions, which causes the following bug:

- `Menu` opens and sets `overflow: hidden` on the page
- `Menu` item is clicked — `Menu` starts closing, but delays its scroll-lock cleanup to respect the animation
- `Dialog` opens before the `Menu`'s scroll lock cleanup has happened. It snapshots the situation (eg page is scroll-locked), then applies its own scroll lock logic, which **wrongly restores page scrolling while dialog is opened**
- When the `Dialog` closes, it restores the previous snapshot and **the page stays scoll-locked**

So far, I managed to avoid the issue with a couple of workarounds:

- trick ariakit by moving the animation on child of the popup element, so that ariakit executes scroll lock cleanup without any delays;
- manually [prevent scroll lock](https://ariakit.com/reference/menu#preventbodyscroll) on the Ariakit `Menu` (not ideal, but seems to work);

Although I wonder if base UI could make improvemens on their side to improve scroll-lock handoff

## Expected behavior

I wish base ui could coordinate cross-library scroll-lock handovers.

A few ideas:

- Come up with a robust heuristic for better coordination/restoration. For example:
  - it should not be possible for `Dialog`'s scroll-lock logic to enable page scroll when opening;
  - it should not be possible for `Dialog`'s scroll-lock cleanup logic to scroll-lock the page (ie add `overflow: hidden`) when closing
- expose a pluggable scroll-lock manager and/or explicit interop hooks

## Reproducible example

https://stackblitz.com/edit/vitejs-vite-altbnksf?file=src%2FApp.tsx,src%2Findex.css

https://github.com/user-attachments/assets/945bf96c-1a72-429c-83de-3a5c4ebbc43e

## Base UI version

1.4

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

N/A

atomiks:
Fixes #4640

Base UI was treating an externally scroll-locked page as settled state, which breaks delayed handoffs from non-Base-UI overlays. If a third-party overlay clears its `html` or `body` lock after `Dialog` opens, the page can become scrollable under the open popup or stay locked after close.

## Changes

- Wait for external `html` or `body` scroll locks to clear before Base UI takes over its own lock.
- Add a regression test that covers both `body` and `html` handoffs into `Dialog`.


packages/utils/src/useScrollLock.ts
```diff
@@ -242,18 +248,54 @@ class ScrollLocker {
     }
   };
 
+  private observePageScrollUnlock(referenceElement: Element | null) {
```
romgrk: Lint: method ordering is not great for readability. The `lock`, `unlock`, and `release` methods being grouped makes it easier to understand how the overall fields logic works. Inserting this method between them makes it harder to read (at least for a human).

packages/utils/src/useScrollLock.ts
```diff
@@ -242,18 +248,54 @@ class ScrollLocker {
...
     const html = doc.documentElement;
-    const htmlOverflowY = ownerWindow(html).getComputedStyle(html).overflowY;
+    const body = doc.body;
+    const win = ownerWindow(html);
 
-    // If the site author already hid overflow on <html>, respect it and bail out.
-    if (htmlOverflowY === 'hidden' || htmlOverflowY === 'clip') {
-      this.restore = NOOP;
+    if (isPageScrollLocked(win, html, body)) {
+      this.restore = this.observePageScrollUnlock(referenceElement);
```
romgrk: So IIUC, we're handling the case where scroll-lock is applied by someone else before we run. Do we also need to handle the case where someone tries to apply a scroll-lock after we have already applied a scroll-lock?

atomiks: Yep, and the inverse is harder. Before we lock, we can safely detect an existing external lock and wait, but after Base UI has locked, later locks write to the same html/body style/clase attr, so there isn’t a reliable signal to tell which `overflow: hidden` belongs to which lib

packages/utils/src/useScrollLock.ts
```diff
@@ -249,11 +255,11 @@ class ScrollLocker {
...
     const doc = ownerDocument(referenceElement);
     const html = doc.documentElement;
-    const htmlOverflowY = ownerWindow(html).getComputedStyle(html).overflowY;
+    const body = doc.body;
+    const win = ownerWindow(html);
 
-    // If the site author already hid overflow on <html>, respect it and bail out.
-    if (htmlOverflowY === 'hidden' || htmlOverflowY === 'clip') {
-      this.restore = NOOP;
+    if (isPageScrollLocked(win, html, body)) {
```
romgrk, line 269: I would keep a comment to explain the necessity of this guard/logic.

packages/utils/src/useScrollLock.ts
romgrk: I've thought about this a bit more, and I think we should avoid including too much complexity in our codebase to deal with this issue. We should aim/optimize for the case where we're the sole low-level library and not try to fix every edge-case with users mix and matching libraries. We're adding complexity & bundle-size costs for everyone by doing so.

I think dealing with this edge-case is fine, but if there are more we should expose the `ScrollLocker` and let users integrate libraries cleanly instead of us trying to guess what happens by observing the DOM.

romgrk (approved):
LGTM beyond the missing comment. Maybe the names could be reworked to `hasExternalScrollLock` and `observeExternalScrollUnlock` — "external" makes it more clear what those are for, and "page" doesn't add much as the scroll lock always apply to the full page.

ciampo:
Hey @atomiks , do you plan on resuming work on this anytime soon? If not, can I help to push this PR to the finish line?

PR #4669, opened 2026-04-23
[autocomplete]  Prevent data-highlighted on disabled items (#4666)

### LINKED ISSUE #4666: [Autocomplete] `disabled` item can be highlighted
# Bug report

## Current behavior

A `disabled` Autocomplete Item can be focused/highlighted using keyboard and pointer events and ends up with the `data-highlighted` attribute.
```jsx
<Autocomplete.Item value="test" disabled>
   Test
</Autocomplete.Item>
```

## Expected behavior

The disabled items should not be interactive and shouldn't have `data-highlighted` attribute.

## Reproducible example

https://stackblitz.com/edit/2fydq8ct?file=src%2FApp.tsx

## Base UI version

v1.4.1

Chrome

## Which OS are you using?

Mac OS

Not applicable

## Additional context

None

albertdugba:
This fixes #4666 

Changes
- Remove the disabledIndices override in AriaCombobox so default disabled/aria-disabled detection is respected.
- Ignore pointer/keyboard navigation updates that target disabled indices in combobox onNavigate.

### DIFF
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -18,7 +18,7 @@ import {
   useListNavigation,
   useClick,
 } from '../../floating-ui-react';
-import { contains, getTarget } from '../../floating-ui-react/utils';
+import { contains, getTarget, isListIndexDisabled } from '../../floating-ui-react/utils';
 import {
   createChangeEventDetails,
   createGenericEventDetails,
@@ -1048,11 +1048,17 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
     // Floating UI tests don't require `role="row"` wrappers, so retains the number API.
     cols: grid ? 2 : 1,
     orientation: grid ? 'horizontal' : undefined,
-    disabledIndices: EMPTY_ARRAY as number[],
     onNavigate(nextActiveIndex, event) {
       // Retain the highlight only while actually transitioning out or closed.
       if ((!event && !open) || transitionStatus === 'ending') {
         return;
+      }
+
+      if (nextActiveIndex != null) {
+        const nextItem = listRef.current[nextActiveIndex];
+        if (nextItem?.hasAttribute('data-disabled') || isListIndexDisabled(listRef, nextActiveIndex)) {
+          return;
+        }
       }
 
       if (!event) {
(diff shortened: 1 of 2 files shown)


packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1048,13 +1048,22 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
...
     cols: grid ? 2 : 1,
     orientation: grid ? 'horizontal' : undefined,
-    disabledIndices: EMPTY_ARRAY as number[],
     onNavigate(nextActiveIndex, event) {
       // Retain the highlight only while actually transitioning out or closed.
       if ((!event && !open) || transitionStatus === 'ending') {
         return;
       }
 
+      if (nextActiveIndex != null) {
```
mj12albert, line 1057: Codex review:

> This only blocks disabled items inside `onNavigate,` but this component still assigns `activeIndex = 0` directly in the `autoHighlight` paths. Those writes bypass `onNavigate,` and `useListNavigation` accepts any in-bounds `activeIndex` without re-checking `disabled` state. In practice, `autoHighlight` and `autoHighlight="always"` can still put `data-highlighted` on a disabled first option.

atomiks:
Disabled items are intentionally focusable/highlightable in our implementation per WAI-ARIA recommendation for listboxes

`AriaCombobox` explicitly passes `disabledIndices: EMPTY_ARRAY` (same as Select + Menu) so disabled items remain navigable, and `Combobox.Item` uses `focusableWhenDisabled: true`. Disabled items should remain reachable/highlightable for navigation/discovery, while activation/selection stays blocked.

(This change also doesn’t fully fix the reported state, since `autoHighlight` writes `activeIndex = 0` directly and bypasses `onNavigate`)

albertdugba:
> Disabled items are intentionally focusable/highlightable in our implementation per WAI-ARIA recommendation for listboxes
> 
> `AriaCombobox` explicitly passes `disabledIndices: EMPTY_ARRAY` (same as Select + Menu) so disabled items remain navigable, and `Combobox.Item` uses `focusableWhenDisabled: true`. Disabled items should remain reachable/highlightable for navigation/discovery, while activation/selection stays blocked.
> 
> (This change also doesn’t fully fix the reported state, since `autoHighlight` writes `activeIndex = 0` directly and bypasses `onNavigate`)

Got it, thanks for the clarification. I didn't realize disabled items are intentionally kept highlightable per WAI-ARIA. sounds like the current behaviour is correct by design. I will close this PR.

PR #4671, opened 2026-04-23
[select] Add `data-popup-side` to trigger

### LINKED ISSUE #4667: [Select] Expose resolved popup side on Trigger
# Feature request

## Summary

Please expose the resolved popup side on `Select.Trigger`, ideally in the same way as other select-like components:

- add a `data-popup-side` attribute to `Select.Trigger`
- add a `popupSide` value to `Select.Trigger.State`

`Select.Positioner`, `Select.Popup`, `Select.Arrow`, and scroll arrow parts already expose the resolved placement via `data-side`, but the trigger currently only exposes `data-popup-open`. For trigger styling, it would be helpful if `Select.Trigger` also knew whether its popup is currently positioned above, below, or on another side of the trigger.

This would make `Select.Trigger` consistent with `Autocomplete.Trigger`, whose docs list both `data-popup-side` and a `popupSide` state value.

Base UI itself already provides this on related select-like components:

- `Autocomplete.Trigger` exposes `data-popup-side` and `popupSide`.
- `Autocomplete.Input` and `Autocomplete.InputGroup` also expose `data-popup-side`.

That API is useful because consumers can style the anchor element based on the actual popup placement without wiring popup positioning state manually.

## Motivation

We maintain shared styling for `Select`, `Combobox`, and `Autocomplete` fields. The fields use different border radius classes depending on whether the popup opens above or below the anchor, so the anchor and popup visually connect as one control.

For `Combobox` and `Autocomplete`, this can be done directly with the Base UI data attributes on the field/trigger parts. For `Select`, we currently need a workaround:

1. Render `Select.Positioner` with a custom `render` function.
2. Read `state.side` from the positioner.
3. Store that value in React state.
4. Manually write `data-popup-side` back onto `Select.Trigger`.

In simplified form:

```tsx
function SelectPositionerBridge(props) {
  useLayoutEffect(() => {
    onPopupSideChange(props.popupSide);
  }, [props.popupSide]);

  return <div {...props} />;
}

<Select.Positioner
  render={(positionerProps, state) => (
    <SelectPositionerBridge
      {...positionerProps}
      popupSide={state.side}
      onPopupSideChange={setPopupSide}
    />
  )}
>
  <Select.Popup>{/* ... */}</Select.Popup>
</Select.Positioner>

<Select.Trigger data-popup-side={open ? popupSide : undefined}>
  {/* ... */}
</Select.Trigger>
```

This works, but it adds component-local state and a layout effect just to mirror internal positioning state that Base UI already computes.

The requested API would let consumers style `Select.Trigger` declaratively:

```css
.Trigger[data-popup-side="bottom"] {
  border-bottom-left-radius: 0;
  border-bottom-right-radius: 0;
}

.Trigger[data-popup-side="top"] {
  border-top-left-radius: 0;
  border-top-right-radius: 0;
}
```

This is primarily a consistency and ergonomics improvement rather than a bug. The current behavior can be worked around, but exposing the resolved popup side on `Select.Trigger` would make custom select styling simpler and align `Select` with the existing `Autocomplete` API.

mj12albert:
Add `data-popup-side` to trigger matching Combobox

Closes https://github.com/mui/base-ui/issues/4667


LukasTy (approved):
LGTM. 👍

flaviendelangle (commented):
# PR Review Summary — mui/base-ui#4671

**[select] Add `data-popup-side` to trigger** by @mj12albert · 8 files, +70 / −16 · Closes mui/base-ui#4667

Three agents reviewed the diff against the upstream `mui/base-ui` repo (code-reviewer, pr-test-analyzer, silent-failure-hunter). Bottom line: **ship it**, with one design question worth confirming and a small follow-up test.

## Critical Issues
None.

## Important Issues

### 1. `positioning.side` vs `renderedSide` divergence — design question, not a bug
`packages/react/src/select/positioner/SelectPositioner.tsx:139-142`

Select's positioner derives `renderedSide = alignItemWithTriggerActive ? 'none' : positioning.side` and uses it for `SelectPositionerState.side`. The new effect pushes the **raw** `positioning.side` to the store. So under the default `alignItemWithTrigger={true}`, the trigger will expose `data-popup-side="bottom"` (or whichever side floating-ui resolved) while the positioner itself reports `data-side="none"`.

- The code-reviewer flagged this as an inconsistency.
- The silent-failure-hunter argues this is **intentionally correct**: `data-popup-side` should reflect actual placement, not the rendered-as-`none` override.

Both views are defensible. Worth a one-line note in the PR confirming the intent — and an explicit test asserting the contract under default `alignItemWithTrigger`.

### 2. Test gaps
`packages/react/src/select/trigger/SelectTrigger.test.tsx:178-203`

Only `side="right"` with `alignItemWithTrigger={false}` is covered. Worth adding (and ideally for Combobox too, for parity):

- **Default `alignItemWithTrigger` mode** — the common case, currently untested. Settles the question above.
- **Collision-flip** — requested `side="bottom"` but viewport forces flip to `top`. Most failure-prone path; protects against a refactor that ever read the requested prop instead of the resolved side.

`defaultOpen` initial-render gating is a nice-to-have but lower value.

Suggested collision-flip test:

```tsx
it.skipIf(isJSDOM)('reflects the resolved side after collision flip', async () => {
  const { user } = await render(
    <div style={{ position: 'fixed', top: 0, left: 0, height: '100vh', width: '100vw' }}>
      <div style={{ position: 'absolute', bottom: 4, left: 100 }}>
        <Select.Root>
          <Select.Trigger data-testid="trigger">Trigger</Select.Trigger>
          <Select.Portal>
            <Select.Positioner side="bottom" alignItemWithTrigger={false}>
              <Select.Popup style={{ height: 200 }}>
                <Select.Item value="a">a</Select.Item>
              </Select.Popup>
            </Select.Positioner>
          </Select.Portal>
        </Select.Root>
      </div>
    </div>,
  );
  const trigger = screen.getByTestId('trigger');
  await user.click(trigger);
  await waitFor(() => expect(screen.queryByRole('listbox')).not.toBe(null));
  expect(trigger).toHaveAttribute('data-popup-side', 'top');
});
```

## Suggestions

- **Cleanup the store on positioner unmount** (`packages/react/src/select/positioner/SelectPositioner.tsx:142-145`). The `useIsoLayoutEffect` has no cleanup, so `store.popupSide` lingers across open/close cycles. It's currently safe because the trigger gates on `mounted && positionerElement` — but `positionerElement` and `listElement` are nulled on unmount, so leaving `popupSide` stale is a foot-gun for future readers. A one-liner `return () => store.set('popupSide', null);` makes the lifecycle pattern self-consistent.
- **Shared mapping for parity with Combobox**. Combobox extracts `triggerStateAttributesMapping` into `combobox/utils/stateAttributesMapping.ts`; Select inlines it in `SelectTrigger.tsx:32-37`. Extracting a shared helper would reduce future drift. Non-blocking.

## Strengths

- **Pattern faithfully mirrors Combobox**: identical store field + selector, identical `useIsoLayoutEffect` push, identical `mounted && positionerElement` gating, identical state-attribute mapping, identical `Side | null` typing from the canonical `useAnchorPositioning` module.
- **No stale-attribute window on close**. Render-time computation `popupSide = mounted && positionerElement ? popupSideValue : null` (`SelectTrigger.tsx:94`) is synchronous with the positioner's ref-callback `setPositionerElement(null)`, so the attribute disappears in the same render.
- **Conventions respected**: `useIsoLayoutEffect`, no `as any`, no shadow-DOM-unsafe globals, no `flushMicrotasks` in the test, `it.skipIf(isJSDOM)` correctly applied.
- **Docs kept in sync**: `page.mdx`, `select/types.md`, JSDoc on `SelectTriggerState.popupSide`, and `SelectTriggerDataAttributes.ts` all match — looks like `pnpm docs:api` was run.
- **No silent failures**. The `null` fallback is a correct expression of "popup is closed", not error-hiding. `getStateAttributesProps` properly omits the attribute when the mapping returns `null`.
- **Initial state is `null`** in `SelectRoot.tsx:154` — attribute correctly absent on first render.

## Recommended Action

1. Confirm with @mj12albert whether exposing `data-popup-side="bottom"` while `data-side="none"` (under default `alignItemWithTrigger`) is intentional.
2. If yes, add a default-mode test plus a collision-flip test (consider mirroring on Combobox for parity).
3. Optional: add the unmount cleanup one-liner.

No re-review needed after these — they're all small.

PR #4676, opened 2026-04-24
[docs] Revamp demos

aarongarciah:
Preview: https://deploy-preview-4676--base-ui.netlify.app/react/handbook/forms

- Updates all demos to look something like this:
  docs-examples
- Adopts tokens from the default Tailwind config instead of rolling our own tokens.
  - Removes color CSS vars from CSS Modules demos to use raw values.
  - Tailwind demos now use `dark:` prefixed class names, and CSS Modules demos now use `@media (prefers-color-scheme: dark)` instead of relying on vars changing their value between light/dark.
  - Removes the `theme.css` file from all CSS Modules demos (also removed from Stackblitz export).


colmtuite:
@aarongarciah 

- I think focus rings should be black (or whatever very high-contrast gray we use)
- Accordion will prob look better if the border encloses the whole Root, so border on left + right sides too. Then we can add padding-inline and a hover state to the buttons.
- What are the icon styles for? Icons shouldn't have any styles besides `display: block`?
- Why not use `line-height: 1` on buttons?
- Add `gap` to buttons that contain text?
- The `padding-block` in Autocomplete + Combobox popup needs to be half its current size.
- Popup/widget group labels should have the exact same styles as items, but just low-contrast text color. Same weight, height, letter-spacing, text-transform etc.
- Sticky group labels seem unnecessarily opinionated to me?
- Dialog padding should be 16px computed.
- Dialog title should be 16px computed.
- Dialog title needs 4px space beneath it.
- The `margin-top` on DialogActions is weird. `gap` is better.
- Why does Drawer have a cropped black border that's only visible on the top edge? In the nested drawers example, when you open the nested drawer, one drawer just has a black border along the top edge?
- Drawer shadows extend upwards, when the drawer opens from the bottom? That's not how shadows work, they come from a light source, which is always in the same place. x + y values should be the same everywhere.
- Numberfield input padding should be 8px computed.
- Numberfield input `outline-offset` should be `-1px`.
- Popover title needs 4px space beneath it.
- Popover trigger should be a text label button, not an icon button.
- Text inside preview card should not be low-contrast.
- Scrollbar is enormous, needs to be way thinner.
- What is the `::before` for in ScrollArea styles?
- Tab panels should contain text, not icons.
- Toast titles need 4px space beneath them

colmtuite:
@aarongarciah Remove the :active style from Accordion buttons. It's too noisy.

atomiks:
1. The outline being inset 1px when focused feels odd to me and makes it look quite cramped
Screenshot 2026-05-11 at 9 26 56 pm


2. The icons in the Combobox input look oddly aligned?
3. The height of the Combobox popup doesn't align with the scroll-margin anymore so it shifts once it reaches the scroll boundary
4. The checkbox/radio buttons feel slightly too small?

atomiks:
1. The number not being centered here feels off:

Screenshot 2026-05-11 at 9 36 54 pm

---

2. Focus outline looks cramped here:

Screenshot 2026-05-11 at 9 36 51 pm

---

3.  Popup should cover right edge of trigger:

Screenshot 2026-05-11 at 9 36 46 pm

---

4. Focus state glitch:

Screenshot 2026-05-11 at 9 36 33 pm

---

5. This Navigation Menu demo doesn't seem to transition the content on the x-axis or its opacity?

Screenshot 2026-05-11 at 9 39 44 pm

---

6. Selection highlight should be gray or even transparent?

Screenshot 2026-05-11 at 9 41 33 pm

atomiks:
Scale animation is broken for this -- no transition. `opacity` works though.

Screenshot 2026-05-15 at 6 29 05 pm

Is it intentional the text  isn't centered? 

Screenshot 2026-05-15 at 6 29 48 pm

Should the switch's background-color transition? It looks slightly odd that the thumb transitions but not the background-color.

Screenshot 2026-05-15 at 6 30 10 pm

This should have a black focus ring:

Screenshot 2026-05-15 at 6 30 37 pm

The bottom padding of combobox lists seems to be smaller than the top padding:

Screenshot 2026-05-15 at 6 31 21 pm

The year text feels slightly hard to read:

Screenshot 2026-05-15 at 6 31 46 pm

The Grid layout example looks off center:

Screenshot 2026-05-15 at 6 32 23 pm

There's too much space beneath the emoji list?:

Screenshot 2026-05-15 at 6 32 47 pm

This example looks too small?:

Screenshot 2026-05-15 at 6 33 08 pm

Mismatched padding-y here:

Screenshot 2026-05-15 at 6 34 00 pm

The Dialog demos' transition duration seems too slow compared to the hero demo.

The Outside Scroll Dialog's translateY animation seems much too fast?

atomiks:
Not sure how I feel about pure white borders in dark mode. Kind of feels too harsh. Maybe a light gray would work better?

Screenshot 2026-05-15 at 6 41 36 pm

atomiks:
## Code Review (GPT-5.5)
Request changes 🔴 because CSS Modules demo exports lost the global theme variables that the revamped demo CSS now depends on.

### 1. Bugs / Issues

#### 1. 🔴 CSS Modules exports no longer include the demo color theme (blocking)

`docs/src/utils/createDemo.ts` now creates demos without the CSS Modules global theme:

```ts
export const createDemo = createDemoFactory({
  DemoContent,
});

export const createDemoWithVariants = createDemoWithVariantsFactory({
  DemoContent,
});
```

`docs/src/utils/demoExportOptions.ts` also only emits the minimal `demo.css` for CSS Modules variants:

```ts
const htmlHeadTemplate: ExportConfig['headTemplate'] = (props) => {
  const head = [htmlHeadWithDefaultStyles(props)];

  if (props.variantName === 'Tailwind') {
    head.push(htmlHeadWithTailwind(props));
  }

  return head.filter(Boolean).join('\n');
};

const transformVariant: ExportConfig['transformVariant'] = (variant, variantName, globals) => {
  globals = { ...globals };
  globals['demo.css'] = { source: demoCss };

  return { globals };
};
```

But the CSS Modules demos now rely heavily on theme tokens such as:

```css
.Accordion {
  border: 1px solid var(--color-neutral-950);
  color: var(--color-neutral-950);
}

.Trigger:hover:not([data-disabled]) {
  background-color: var(--color-neutral-100);
}
```

Those variables are defined for the docs site in `docs/src/css/index.css` via `@theme`, but they are not included in isolated CSS Modules exports. As a result, copied/exported CSS Modules demos in CodeSandbox/StackBlitz lose many colors, borders, backgrounds, and states because the referenced custom properties are undefined.

**Fix:** Keep exporting a CSS Modules theme file, update it to define the new `--color-neutral-*`, `--color-red-*`, `--color-blue-*`, and `--color-green-*` tokens, and continue linking/exposing it for `CssModules` variants; alternatively inline those variables into the generated `demo.css` for CSS Modules exports.

### 2. Accuracy and Completeness Assessment

I reviewed the full branch diff against freshly fetched `upstream/master`. The in-docs rendering path has the new tokens registered globally, so this issue is specifically with the exported/copied CSS Modules demo path rather than the main docs page styling.

### 3. Validation Coverage

I ran `git diff --check` and found no whitespace/conflict-marker issues. I did not run the full docs build or browser visual regression pass.

aarongarciah:
@atomiks 

> 1. The outline being inset 1px when focused feels odd to me and makes it look quite cramped

All focus rings have been updated.

> 2. The icons in the Combobox input look oddly aligned?

We have new icons now.

> 3. The height of the Combobox popup doesn't align with the scroll-margin anymore so it shifts once it reaches the scroll boundary

I think this is fixed now.

> 4. The checkbox/radio buttons feel slightly too small?

You mean the checkmark icon? We have a new one now.

---

> 1. The number not being centered here feels off:

We decided to leave it left-aligned.

> 2. Focus outline looks cramped here:

Fixed

> 3.  Popup should cover right edge of trigger:

Fixed

> 4. Focus state glitch:

Fixed (the outline could be improved then is pressed but that's another issue)

> 5. This Navigation Menu demo doesn't seem to transition the content on the x-axis or its opacity?

I _think_ I fixed the regression

> 6. Selection highlight should be gray or even transparent?

That change was part of another PR. I rebased and now it's gray.

---

> 1. 🔴 CSS Modules exports no longer include the demo color theme (blocking)

The issue reported by Codex is not really an issue. Before merging, we'll replace all CSS vars in CSS modules demos to use raw values.

colmtuite (approved):

aarongarciah:
@atomiks 

> Scale animation is broken for this -- no transition. opacity works though.

Fixed

> Is it intentional the text isn't centered?  Yes  > Should the switch's background-color transition? It looks slightly odd that the thumb transitions but not the background-color.

Updated. I wasn’t sure about it.

> This should have a black focus ring:

Updated all focus rings to be "black" on the docs shell.

> The bottom padding of combobox lists seems to be smaller than the top padding:

I added `overflow: clip` to the `.Popup` part to avoid the scrollbar overflowing it, but then it made the bottom padding smaller. Turns out I didn’t realize it was a miscalculation on the max-height 😅. Should be fixed now on all popups.
￼
> The year text feels slightly hard to read:

Tweaked. There might be some more cases like this in dark mode. There will be plenty more PRs to tweak stuff.

> The Grid layout example looks off center:

Fixed

> There's too much space beneath the emoji list?:

Fixed

> This example looks too small?:

Why is that? It’s a 32x32 avatar, same size as buttons, inputs, etc.

> The Dialog demos' transition duration seems too slow compared to the hero demo.

Fixed

> The Outside Scroll Dialog's translateY animation seems much too fast?

Fixed

> Not sure how I feel about pure white borders in dark mode. Kind of feels too harsh. Maybe a light gray would work better?

We might change it in another PR to gray-50. Leaving as is in this PR.

atomiks:
> Why is that? It’s a 32x32 avatar, same size as buttons, inputs, etc.

For a demo, it just looks too small, even if the size is realistic in actual apps. Just my opinion though.

>  Updated. I wasn’t sure about it.

Thumb still needs it too:

https://github.com/user-attachments/assets/36259c80-c3a5-4815-bf50-29995edba175

atomiks:
8px width scrollbar thumbs aren't accessible without an extended hit area to be 16-18px (Windows/macOS defaults). The original demo had a hit area iirc.

Screenshot 2026-05-17 at 5 47 43 pm

This includes the Dialog demos.

aarongarciah:
> Thumb still needs it too:

There was a bug on TW demos with separate `transition-transform` and `transition-colors` instead of a single `transition-[translate,background-color]`.

> 8px width scrollbar thumbs aren't accessible without an extended hit area to be 16-18px (Windows/macOS defaults). The original demo had a hit area iirc.

Good point. Re-added on all demos using ScrollArea. What do you think about expanding the hit area outside of the scrolling container? In some scenarios is not useful (e.g. when the ScrollAre touches the viewport edge) and in others could be harmful (e.g. if the scrolling container's edge is touching some interactive element). So I was thinking on expanding the hit area only inwards.

Screenshot 2026-05-17 at 14 20 12

This wasn't a problem before this PR because the visible ScrollArea was inset.

Screenshot 2026-05-17 at 14 04 11

colmtuite:
> What do you think about expanding the hit area outside of the scrolling container?

No. Delete. Put it back to a simple 15px/16px scrollbar and delete the pseudo element like I said before.

atomiks:
> Put it back to a simple 15px/16px scrollbar and delete the pseudo element like I said before.

The thick scrollbar looked pretty terrible to me. The extended hit area is a more realistic use case with a thin track/thumb which is the case in ~every app.

colmtuite:
It looks awful yeah. But it's more important that initial demos are simple. The plan is to move fancy hit target tricks, and animation demos, to future examples. Just trying to keep this PR simple so we can merge it asap.

aarongarciah:
Updated the scrollbars to be thiiiick again (I love them)

atomiks:
The Tailwind version of the Accordion demo (and maybe Collapsible) has weird jumpy easing. I don't notice it with CSS Modules.

https://github.com/user-attachments/assets/3200634d-2ab6-47e3-a4fe-d8a3ca522e98

atomiks:
Nested Dialogs + Close confirmation TW examples are mispositioned

Screenshot 2026-05-18 at 4 42 28 pm

atomiks:
Screenshot 2026-05-18 at 4 44 35 pm

`padding-y` of both CSS Modules and is still off in this example. The bottom has slightly more padding

atomiks:
Found a blue ring:

Screenshot 2026-05-18 at 4 49 04 pm

atomiks:
Shouldn't the link be black now?

Screenshot 2026-05-18 at 4 52 41 pm

atomiks:
I think these classes should go beneath the component definition. Comps like NavigationMenu do this.

Screenshot 2026-05-18 at 4 55 30 pm

atomiks:
The highlight should probably be darker here, as it's not close to being accessible and on LQ screens probably is not differentiable at all. There's an open issue about this, but I think the 3:1 contrast requirement stated is excessive. 

I think it should match gray text selection color or a bit lighter

Screenshot 2026-05-18 at 4 58 12 pm

I also don't think the scrollbars should fade out when hovering out. It's quite distracting

atomiks:
The focus outline here seems weird?

Screenshot 2026-05-18 at 5 01 25 pm

aarongarciah:
> `padding-y` of both CSS Modules and is still off in this example. The bottom has slightly more padding

That's caused by the focus ring being offset 1px.

flaviendelangle (commented):
Small nitpick on the positioning of the context menu.
With the new theme, we can clearly see that the menu is placed slightly higher than where we click. On the other context menu I tested (VSCode, Chrome and the File Explorer, all on Linux Mint)  the Context Menu is placed exactly so that your mouse is at the top left of the menu (or other edge depending on the available space).

I think we already had the same behavior on the demos before your revamp, but the new sharp edges make it more visible.

aarongarciah:
> The Tailwind version of the Accordion demo (and maybe Collapsible) has weird jumpy easing. I don't notice it with CSS Modules.

Updated the easing on the Tailwind demo to match the CSS modules one.

> Found a blue ring

Updated.

> Shouldn't the link be black now?

Good point. Updated.

> Nested Dialogs + Close confirmation TW examples are mispositioned

Fixed.

> I think these classes should go beneath the component definition. Comps like NavigationMenu do this.

Not really related to this PR, but updated all demos that I saw following this pattern.

> The highlight should probably be darker here,

Updated to use neutral-200 / neutral-700 (light mode / dark mode) for the highlighted bg.

> I also don't think the scrollbars should fade out when hovering out. It's quite distracting.

You mean it should always be visible or only when scrolling?

> The focus outline here seems weird?

Do you prefer this?

image

atomiks:
> Do you prefer this?

I think what's on the live site is fine? (no focus ring)

> Small nitpick on the positioning of the context menu.

Yeah, it matches macOS

> You mean it should always be visible or only when scrolling?

Always visible, same as the docs shell one

docs/src/app/(docs)/react/components/avatar/demos/hero/css-modules/index.module.css
```diff
@@ -6,13 +6,18 @@
   border-radius: 100%;
   user-select: none;
   font-weight: 400;
-  color: var(--color-gray-900);
```
flaviendelangle, line 9: This component looks quite different from the other now, I think it's the only one with round borders

docs/src/app/(docs)/react/components/combobox/demos/input-inside-popup/css-modules/index.module.css
```diff
@@ -11,131 +11,191 @@
   align-items: center;
```
flaviendelangle, line 11: Image


The border of the input is wider than the border of the popup

atomiks:
Empty state looks off here - should be centered?

Screenshot 2026-05-18 at 5 58 05 pm

Padding is off in this one:

Screenshot 2026-05-18 at 5 58 39 pm

docs/src/app/(docs)/react/components/navigation-menu/demos/nested/css-modules/index.module.css
```diff
@@ -1,14 +1,17 @@
 .Root {
```
flaviendelangle, line 1: Nitpick: the horizontal alignment is not the same across nested Menu and nested Navigation Menu. Might be expected though
Image

Image

And on the Navigation Menu the bottom aligment is 1px off 😆 

Image

atomiks:
Spacing seems off:

Screenshot 2026-05-18 at 5 59 09 pm

docs/src/app/(docs)/react/components/input/demos/hero/css-modules/index.module.css
```diff
@@ -6,25 +6,53 @@
...
   box-sizing: border-box;
-  padding-left: 0.875rem;
+  padding: 0 0.5rem;
   margin: 0;
-  border: 1px solid var(--color-gray-200);
-  width: 14rem;
-  height: 2.5rem;
-  border-radius: 0.375rem;
+  border-radius: 0;
+  border: 1px solid oklch(14.5% 0 0deg);
```
flaviendelangle, line 21: Is this intended to keep a radius here unlike almost all other components (e.g. Number Field)

flaviendelangle (commented):
The revamp is insane 🤩

docs/src/app/(docs)/react/components/combobox/demos/input-inside-popup/css-modules/index.module.css
aarongarciah: that's the focus ring

docs/src/app/(docs)/react/components/avatar/demos/hero/css-modules/index.module.css
aarongarciah: yeah, I think it's fine

docs/src/app/(docs)/react/components/combobox/demos/input-inside-popup/css-modules/index.module.css
flaviendelangle: It feels ugly imo, the input always have the focus in this demo (since blurring closes), so you have this weird border mis-alignment whenever you open the popup

atomiks:
- Accordion triggers are too wide to fit on iPhone (base)
- Navigation Menu on iPhone causes text reflow as it transitions the width (`Styling` menu for example)
- Object values Select demo is missing highlight for touch despite being a single selection mode
- This isn't new, but `overscroll-behavior: contain` should be added to the Select scrollable lists for mobile
- Long pressing Context Menu on iPhone causes a selection bubble to appear (not on live site)
- Opening Select inside the Toolbar demo on touch reveals a broken trigger state 
- The search box when it has no focus state [default for mobile] for autocomplete and combobox when the trigger is inside the popup looks off as there's no separation

aarongarciah:
@atomiks @flaviendelangle @michaldudak I'll apply some more improvements over the few next hours, but we want to merge this before the release. Many demos need improvements, but I think it'd be better to merge and iterate if we can't include every possible improvement in this PR.

One important thing I'd love to have feedback on: what do you think about the removal of `theme.css` from CSS Modules demos? The idea is to have as few abstractions as possible. As you probably noticed, we no longer have a custom set of CSS vars: we use default Tailwind CSS vars. That means, we now rely on `dark:` class names for Tailwind demos, and we rely on `@media (prefers-color-scheme: dark)` for CSS Modules demos. In CSS Modules demos, we use raw values (from Tailwind, so both demo variants look the same).

docs/src/app/(docs)/react/components/input/demos/hero/css-modules/index.module.css
aarongarciah: Not sure I follow. I don't see any focus ring on Input.

flaviendelangle: :facepalm: I was on the wrong URL, forget that

flaviendelangle:
I never worked on the doc infra and theme so I don't feel confortable approving this kind of change.
The doc looks good enough to be shipped as is imo, but I'll let the other get the final approval :pray:

aarongarciah:
> I think what's on the live site is fine? (no focus ring)

Updated since we'll be working on new stuff in the near future.

> Always visible, same as the docs shell one

Updated.

> Empty state looks off here - should be centered?

I think it's fine. The items are left aligned and they're filtered as you write. If we center the empty state we change where the user needs to look. Not a strong opinion but it looks fine to me.

> Padding is off in this one:

Fixed

> Spacing seems off:

Fixed

docs/src/app/(docs)/react/components/combobox/demos/input-inside-popup/css-modules/index.module.css
aarongarciah: changed to be inset, also in the Autocomplete grid demo which is a similar scenario

aarongarciah:
> Accordion triggers are too wide to fit on iPhone (base)

Fixed

> Navigation Menu on iPhone causes text reflow as it transitions the width (Styling menu for example)

Should be fixed now

> Object values Select demo is missing highlight for touch despite being a single selection mode

Fixed (it was broken on master's the Tailwind variant)

> This isn't new, but overscroll-behavior: contain should be added to the Select scrollable lists for mobile

I'll do a follow up PR for this

> Long pressing Context Menu on iPhone causes a selection bubble to appear (not on live site)

Fixed. Looks like `-webkit-user-select: none` was needed after all.

> Opening Select inside the Toolbar demo on touch reveals a broken trigger state

Fixed

> The search box when it has no focus state [default for mobile] for autocomplete and combobox when the trigger is inside the popup looks off as there's no separation

Fixed. Also made the focus outline inset in this scenarios as it looks better.

PR #4683, opened 2026-04-26
[select] Clear stale item text ref

### LINKED ISSUE #1922: Select with dynamic items fails to show popup upon update

atomiks:
Fixes https://github.com/mui/base-ui/issues/1922#issuecomment-4314884068

When a controlled Select resets its value to `null` while replacing its options, React can reuse an old item text node if the rendered items are keyed by index. That leaves `selectedItemTextRef` connected but stale, so `alignItemWithTrigger` measures the wrong item and reopens the popup scrolled to the old selected position.

## Changes

- Clear the selected item text ref whenever the closed Select resolves to no selected index.
- Add a Chromium regression test for replacing options while resetting the controlled value to `null`.


flaviendelangle (approved):
# PR #4683 Review — `[select] Clear stale item text ref`

**URL:** https://github.com/mui/base-ui/pull/4683
**Author:** atomiks
**Base:** `master` ← `codex/select-stale-item-text-ref-pr`
**Files changed:** 2 (+160 / −11)

- `packages/react/src/select/root/SelectRoot.tsx` (+30 / −11)
- `packages/react/src/select/root/SelectRoot.test.tsx` (+130 / −0)

---

## Summary

Targeted bug fix in `SelectRoot.tsx` that nulls `selectedItemTextRef` when the resolved selection becomes empty, plus two Chromium regression tests in `SelectRoot.test.tsx`. The change addresses [issue #1922](https://github.com/mui/base-ui/issues/1922#issuecomment-4314884068): when items are keyed by index and the controlled value resets to `null` while options are replaced, the previously-set ref keeps pointing to a recycled DOM node. `alignItemWithTrigger` then misaligns the popup on next open.

---

## Why the existing clearance paths miss it

Tracing the three places that touch `selectedItemTextRef`:

- **`SelectItemText.tsx:29-33`** — only reassigns when `selectedItemTextRef.current === null || !isConnected`. With key-by-index reuse the node remains connected, so this guard does nothing.
- **`SelectItem.tsx:102-112`** — only writes the ref when an item value matches the selected value; with `value=null` no item matches, so nothing overwrites or clears.
- **`SelectPositioner.tsx:186-209`** — clears via `onMapChange`, but that path is gated on `map.size` *changing*. Index-keyed reuse keeps the size identical, so `onMapChange` skips the clearance.

The PR plugs the gap inside `syncSelectedIndex`, which is already triggered on the relevant transitions.

---

## Critical issues

None.

---

## Important issues

None. The fix is minimal, scoped to the documented scenario, and behavior-preserves all existing branches. I walked each old/new code path:

| Case                                | Old behavior                  | New behavior                                              |
| ----------------------------------- | ----------------------------- | --------------------------------------------------------- |
| multiple + value `[]`               | set `selectedIndex=null`      | clear ref, set `selectedIndex=null`                       |
| multiple + last value found         | set `selectedIndex=lastIndex` | set `selectedIndex=lastIndex`                             |
| multiple + last value not in list   | set `selectedIndex=null`      | clear ref, set `selectedIndex=null`                       |
| single + value found                | set `selectedIndex=index`     | set `selectedIndex=index`                                 |
| single + value not found            | set `selectedIndex=null`      | clear ref, set `selectedIndex=null`                       |
| open transition + `value=null`      | (no-op, early return)         | clear ref (new), early return                             |

Every existing case reaches an equivalent `selectedIndex`; the only added behavior is the ref-clearance, which is the intended fix.

---

## Suggestions

### 1. Open-branch only clears on `!hasSelectedValue`

The closed branch clears whenever `nextIndex === null` (covers both `value=null` and `value=stale-not-in-list`). The open branch at the new `if (open) { if (!hasSelectedValue) { ... } return; }` only clears when `!hasSelectedValue`.

If a consumer batches `setItems(newItems-without-stale-x)` + `setOpen(true)` while leaving `value="x"` set, `hasSelectedValue` stays `true` and the stale ref isn't cleared. Probably out of scope here (the issue is specifically about resetting to `null`), but worth either:

- mentioning the limitation in the PR description, or
- mirroring the closed-branch logic in the open branch (resolve the index and clear when `-1`).

### 2. Optional one-line comment

Inside `syncSelectedIndex`, a short note on *why* the ref is cleared (e.g. `// Drop a possibly-recycled DOM node from index-keyed item replacement before alignment reads it.`) would help future readers. The two existing clearances in `SelectPositioner.tsx` are also uncommented, so this is consistency-neutral — pick one.

### 3. Dependency-array (no action needed)

Adding `selectedItemTextRef` to the deps is redundant (refs are stable) but matches the project's existing pattern of including refs (see `valuesRef` in the same array). Fine as-is.

---

## Strengths

- Surgical fix; no API or public-types churn.
- Refactor extracting `hasSelectedValue` removes a duplicate expression between the `setFilled` effect and the new check.
- The `nextIndex` rewrite unifies the previously branching `multiple` / single early-returns into one assignment + one `store.set`, making the ref-clearing site obvious.
- Behavior parity verified for all five branches (table above).

---

## Tests

Both tests meet AGENTS.md guidelines:

- ✅ `it.skipIf(isJSDOM)` — correct, both rely on real layout (`getBoundingClientRect`, `scrollTop`).
- ✅ No `await flushMicrotasks()` after `render`.
- ✅ Vitest-only APIs (`expect`, `waitFor`).
- ✅ `key={index}` in fixtures is the precise reproducer; selecting `a-35` (deep in a 40-item list) makes the misalignment unambiguous.
- ✅ Three assertions cover open state, scroll reset, and the first option being below the positioner top — together they pin down the bug.

**Test coverage:**
- **Test 1** — Unbatched path (closed-branch clear via separate render → click trigger).
- **Test 2** — Batched path (`setGroup` + `setValue(null)` + `setOpen(true)` in one handler, exercising the open-branch clear).

Both are necessary; Test 2 specifically justifies the open-branch addition.

**Possible enhancement (non-blocking):** if you later want to harden the batched scenario, a third variant where `value` is set to a *new-but-stale-against-old-options* value (instead of `null`) would exercise the gap mentioned in suggestion (1).

---

## Compliance check (AGENTS.md)

- ✅ `useIsoLayoutEffect`, `useStableCallback` patterns intact
- ✅ Shadow-DOM-safe utilities — not relevant here, no new DOM traversal
- ✅ No new comments where they aren't needed
- ✅ No duplicated logic (deduped via `hasSelectedValue`)
- ✅ Commit messages follow `[scope]` format (`[select] Clear stale item text ref`, etc.)
- ✅ No public API change → `pnpm docs:api` not required
- ✅ No new error messages → `pnpm extract-error-codes` not required

---

## Recommendation

**Approve / merge after** the author confirms whether suggestion (1) — the open-branch stale-non-null-value case — is worth covering now or deferring. Everything else is approval-level.

PR #4684, opened 2026-04-26
[all components] Clean up component internals

atomiks:
This is mostly an internal cleanup pass across component implementations. It makes field control registration terser, removes unnecessary stable callback wrappers, and normalizes spacing between unrelated setup blocks. It also preserves live menu change-event detail getters when adding `preventUnmountOnClose`, which fixes canceled checkbox/radio menu item changes.

## Changes

- Simplify `useRegisterFieldControl` call sites to positional arguments and remove the extra options type.
- Keep form value overrides lazy only where the submitted form value differs from the field validation value.
- Replace unnecessary `useStableCallback` wrappers with plain local handlers where the handler is not used as a dependency.
- Preserve live menu item change-event details while adding `preventUnmountOnClose`, and cover canceled checkbox/radio item changes with regression tests.
- Add whitespace between semantically separate setup groups in touched component files.


michaldudak:
## Review summary

Cleanup is sound — no correctness regressions. One point worth surfacing before merge.

### Important

**PR description claim "behavior unchanged" is inaccurate for the Menu items change.**

`MenuCheckboxItem.tsx:77-89` and `MenuRadioItem.tsx:76-82` changed from:

```ts
const details = { ...createChangeEventDetails(...), preventUnmountOnClose: () => {} };
```

to:

```ts
const details = createChangeEventDetails(..., undefined, { preventUnmountOnClose() {} });
```

The old spread **flattened the live `isCanceled` / `isPropagationAllowed` getters into stale `false` snapshots**, so a consumer calling `details.cancel()` from `onCheckedChange` had no effect on the subsequent `details.isCanceled` read. The new form preserves the live getters — this is a **latent bug fix**, not a no-op.

Suggested follow-up:
- Update the PR description to reflect this is a bug fix.
- Add a regression test asserting `details.cancel()` from `onCheckedChange` prevents the toggle in `MenuCheckboxItem`.
- Note that `MenuRadioItem.handleClick` doesn't read `details.isCanceled` at all (pre-existing asymmetry — likely out of scope here, but worth filing).

### Suggestions

- The new 4th positional arg in `useRegisterFieldControl(ref, id, value, getValueOverride?)` loses some self-documentation. The call-site renames (`getStringifiedValueForForm`, `getValueFromInput`) help, but `getFormValueOverride` is slightly misleading: when `value === undefined`, the override is the *primary* source rather than an override. `getFormValue` may read better.

### Strengths

- "Disabled" inference from `id === undefined` traces correctly across all call sites. `CheckboxGroup` (`fieldName ? id : undefined`) and `CheckboxRoot` (`groupContext ? undefined : id`) preserve gating semantics, and the no-id case was already a soft no-op deeper in `useFieldControlRegistration`.
- **OTPFieldRoot dropping `getValue: () => valueRef.current` is safe.** `value` is `normalizeOTPValue(...)` — always a defined string — and `useValueAsRef` updates `valueRef.current` in a layout effect, the same phase as `registration.value`. No freshness gap.
- **RadioGroup change is equivalent.** Old: `value: checkedValue, getValue: () => checkedValue ?? null`. New: `value: checkedValue ?? null`. The fallback path in `useFieldControlRegistration` only fires when `value === undefined`, which produced `null`; the new form just inlines that.
- Keeping `getValue` overrides only on `FieldControl`, `AriaCombobox`, and `SelectRoot` is correct — those are exactly the cases where the form-submitted (stringified) value differs from the validation (raw) value.
- Renames (`getValueForForm`, `getValueFromInput`, `getStringifiedValueForForm`) make each override's role explicit.

michaldudak (approved):

PR #4685, opened 2026-04-26
[form] Avoid flushSync during validation

atomiks:
Calling form validation during OTP auto-submit could trigger `flushSync` from inside a lifecycle method. Field validation now updates the form registry synchronously as part of the normal validation pass, so submit blocking still sees current validity without forcing a React flush.

## Changes

- Remove the `flushSync` wrapper from field control validation.
- Update form validation callers to use the simplified validate contract.
- Keep matching registered field entries in sync when field validity changes.
- Add an OTP auto-submit regression with `Form`.


flaviendelangle (commented):
# PR #4685 Review — `[form] Avoid flushSync during validation`

**PR:** https://github.com/mui/base-ui/pull/4685
**Author:** atomiks · **Branch:** `codex/form-validation-no-flushsync` · **Diff:** 7 files, +173 −34

## Context

Calling form validation during `OTPField.autoSubmit` triggered React's "flushSync was called from inside a lifecycle method" warning, because `useValueChanged` invokes `requestSubmit()` from a layout effect and the submit handler then called `flushSync(commit)` to keep `formRef.current.fields` up to date for submit-blocking. The PR removes the `flushSync` wrapper and instead has `useFieldValidation.commit` write the registry entry synchronously as part of the normal validation pass, threading a new `registeredFieldIdRef` through `FieldRoot` so registry writes can address controls (Switch / OTPField) whose registry id is not the labelable `controlId`.

Five specialized reviewers ran in parallel: **code-reviewer**, **pr-test-analyzer**, **silent-failure-hunter**, **comment-analyzer**, **type-design-analyzer**. Findings, deduplicated and reconciled, follow.

---

## Critical

1. **Test for "same-name field validity scoped on submit" does not exercise the new code path.**
   `packages/react/src/form/Form.test.tsx:60-86`. Both fields use `<Field.Control>`, where the registry key has historically equaled `controlId`. The bug class the PR fixes is precisely controls where the registry id is *not* `controlId` (Switch, OTPField). To actually exercise `registeredFieldIdRef`, replace one or both controls with `<Switch.Root>` (or two OTPFields). As-is, the test would pass on master.
   *(pr-test-analyzer)*

2. **The flushSync regression assertion is fragile.**
   `packages/react/src/otp-field/root/OTPFieldRoot.test.tsx:883-892`. The `console.error` spy filters on the substring "flushSync was called from inside a lifecycle method", which is React internal-warning copy and not a stable contract. Replace with a direct `vi.spyOn(ReactDOM, 'flushSync')` and `expect(...).not.toHaveBeenCalled()` — wording-independent and tests the actual contract named in the PR title.
   *(pr-test-analyzer)*

## Important

3. **`registeredFieldIdRef` weakly encodes its writer/reader contract.**
   `FieldRoot.tsx:58`, written by `useFieldControlRegistration.ts:130,139`, read by `useFieldValidation.ts:71`. Both hooks receive an identically typed `RefObject<string | undefined>`, hiding the role asymmetry (registration writes, validation reads). Consider splitting into `getRegisteredFieldId: () => string | undefined` for the validation hook and `setRegisteredFieldId: (id: string | undefined) => void` for the registration hook — same memory model, structurally honest about ownership, and a stray write would no longer type-check.
   *(type-design-analyzer)*

4. **`commit` parameter type unnecessarily widened in `UseFieldControlRegistrationParameters`.**
   `useFieldControlRegistration.ts:151` declares `commit: (value: unknown, revalidate?: boolean) => void`, but the registration's own `validate` only ever calls `commit(nextValue)` (no second arg). The widening allows future code in registration to reach into `useFieldValidation`'s internal `revalidate`-only path, which it should not. Narrow back to `(value: unknown) => void` (parameter bivariance still permits assigning the wider concrete function), or document via JSDoc that registration must not set the flag.
   *(type-design-analyzer, code-reviewer)*

5. **Comment in `Form.tsx` second sentence is rot-prone.**
   `packages/react/src/form/Form.tsx:103-104`:
   > Async validation isn't supported to stop the submit event. Field validation updates the form registry synchronously so the submit event can be prevented.

   First sentence captures real WHY (load-bearing async limitation) — keep. Second sentence narrates WHAT the next ten lines of code do and pins the comment to the current implementation; refactoring this flow again will leave the comment stale. Drop the second sentence; if the synchronous-write contract is worth documenting, attach it to `validate`/`commit`, not to this call site.
   *(comment-analyzer)*

6. **No coverage for the `revalidate` early-return path on unnamed controls.**
   `useFieldValidation.ts:106-123` is the "user typing while errored" path — it now reads `registeredFieldIdRef.current ?? controlId` and was migrated by this PR, but no test exercises it for `<Switch.Root>` / `<OTPField>`. Suggested: submit invalid Switch → toggle valid → assert `aria-invalid` clears and a subsequent submit succeeds.
   *(pr-test-analyzer)*

7. **No coverage for registration id changes after first registration.**
   `useFieldControlRegistration.ts:117-141` handles `previousId !== registration.id` and updates `registeredFieldIdRef`. No test covers a control swap inside one `<Field.Root>`. Stale-id bugs hide easily in this kind of ref-passing pattern.
   *(pr-test-analyzer)*

## Suggestions

- **Add a JSDoc post-condition on `FormContext.formRef.current.fields[id].validate`**: "After this call returns, the field's registry entry reflects the latest synchronous validity verdict; async validators do not block submit." Captures the load-bearing invariant the PR introduces. *(type-design-analyzer)*
- **Clarify the `false` literal at `useFieldValidation.ts:108`**. The previously removed `// invalid = false` was a WHAT comment, but in dropping it the WHY is now silent — the user just resolved the `valueMissing` condition, so we deliberately ignore the externally-derived `invalid` (which may still reflect form-level errors). One line of WHY here would help. *(comment-analyzer)*
- **`expect(node).not.toBe(null)` is tautological** when the node was just retrieved with `getByTestId` (which throws on miss): `Form.test.tsx:56,84`, `OTPFieldRoot.test.tsx`. Use `toBeInTheDocument()` or assert via `toHaveTextContent`. Mirrors prior style in this file, so not a regression — but new tests can do better. *(code-reviewer, pr-test-analyzer)*
- **`try/finally { mockRestore() }`** at `OTPFieldRoot.test.tsx:944-991` is unusual for the suite; prefer `beforeEach`/`afterEach` at the `describe('prop: autoSubmit')` level. *(pr-test-analyzer)*
- **Throwing inside `console.error` mock impl** at `OTPFieldRoot.test.tsx:885-892` propagates into React's error handling. Prefer non-throwing filtering (e.g. record-and-assert-after) or rely on `createRenderer`'s default unexpected-error-detection. *(pr-test-analyzer)*

## Note on async-validator concerns

The **silent-failure-hunter** flagged a long list of paths around `commit` being `async` and `Form.tsx`'s submit handler not awaiting validators. These are accurate observations but **pre-existing**: the prior `flushSync(commit)` only awaited the synchronous portion of `commit` either; async validators could not block submit on master either. The new comment in `Form.tsx:103` ("Async validation isn't supported to stop the submit event") makes this limitation more visible than before, not less. The PR's `JSDoc` on `FieldRoot.validate` already documents this. Treat any improvements here (warning when an async validator's result is discarded during submit; `await Promise.all` of `field.validate()`) as separate hardening work, not as blockers for this PR.

## Strengths

- Correct root cause: the layout-effect → `requestSubmit` → `field.validate()` → `flushSync` chain in OTP autoSubmit is removed cleanly.
- New `updateRegisteredFieldValidity` helper deduplicates two registry-write sites in `useFieldValidation.ts` (per AGENTS.md "avoid duplicating logic").
- AGENTS.md compliance: `useStableCallback`, `useIsoLayoutEffect`, no `flushSync`, no `react-dom` import, Vitest-only test APIs.
- Three orthogonal regression cases covered: unnamed registered control invalid (Switch), same-name fields (though see Critical #1), OTP autoSubmit success and blocked.
- Type signature simplification — `validate: (flushSync?: boolean) => void` → `validate: () => void` — removes a parameter that leaked React-DOM scheduling into a domain contract.

## Recommended actions

**Must fix before merge:**
- (1) Rewrite the same-name-scoped test to use `<Switch.Root>` so it exercises `registeredFieldIdRef`.
- (2) Replace the `console.error` substring spy with a direct `flushSync` spy.

**Should address in this PR:**
- (4) Narrow `commit` parameter in `UseFieldControlRegistrationParameters`.
- (5) Drop second sentence of the `Form.tsx` submit comment.
- (6, 7) Add tests for the `revalidate` path and id-change scenarios.

**Nice to have:**
- (3) Split the shared ref into typed accessor functions.
- JSDoc post-condition on `validate`.

## Files referenced

- `packages/react/src/field/root/FieldRoot.tsx`
- `packages/react/src/field/root/useFieldValidation.ts`
- `packages/react/src/form/Form.tsx`
- `packages/react/src/form/Form.test.tsx`
- `packages/react/src/internals/field-register-control/useFieldControlRegistration.ts`
- `packages/react/src/internals/form-context/FormContext.ts`
- `packages/react/src/otp-field/root/OTPFieldRoot.test.tsx`

atomiks:
> (2) Replace the console.error substring spy with a direct flushSync spy.

I disagree with this one since the original issue is about the error appearing during a specific React phase, and this tests that more directly. Checking if `flushSync` was called doesn't test the timing of the call. The error messages haven't changed in a long time so that's not a valid concern imo

Addressed others

atomiks:
Approve ✅ The patch removes the lifecycle `flushSync` dependency while preserving synchronous submit blocking through the form registry, and the targeted regression coverage/typecheck look solid.

### 1. Bugs / Issues (None)

No concrete blocking, non-blocking, or informational issues found.

The fix addresses the real submit-time requirement: validation no longer needs React state to flush synchronously because `validate()` now updates the form registry’s `validityData` before submit checks read it. The registered-control id handoff also keeps non-native field controls, same-name fields, and control replacement scoped to the correct registry entry.

The changed code follows the local hook conventions: `useStableCallback` remains on effect/event-facing callbacks, `useIsoLayoutEffect` usage is unchanged where needed, and the tests stick to Vitest/testing-library matcher style.

### Test Coverage Assessment

Validated locally:

- `pnpm test:jsdom Form --no-watch`
- `pnpm typescript`

The new OTP auto-submit regression tests were also applied to the PR merge-base implementation and failed there, confirming they catch the removed `flushSync` lifecycle warning path.

PR #4687, opened 2026-04-26
[combobox] Fix popup input form submit

### LINKED ISSUE #4681: [combobox] form submit doesn't work when using combobox (input inside list pattern)
# Bug report

## Current behavior

When following combobox input inside list pattern, Form submit doesn't work properly if you have render prop on combobox input.

## Expected behavior

It should submit properly every time

## Reproducible example

https://stackblitz.com/edit/vitejs-vite-nvkvpy5e?file=src%2FApp.jsx

You will notice that after each refresh you have select twice to submit the form. submit always doesn't at first time

## Base UI version

v1.4.1

Chrome

## Which OS are you using?

Mac OS

NA

## Additional context

atomiks:
Fixes #4681

When a combobox uses the input-inside-popup pattern and renders `Combobox.Input` through field-aware `Input`, the popup search input inherited the surrounding `Field.Root` name and registered as the field control. That left the submitted field value empty until the interaction was repeated.

## Changes

- Render positioned popup inputs under the default no-op Field context so field-aware render targets stay search-only.
- Keep the combobox's hidden form control responsible for field registration and submission.
- Add regression coverage for a required combobox using `render={<Input />}` inside the popup.


flaviendelangle (approved):
# PR #4687 Review Summary — `[combobox] Fix popup input form submit`

**Verdict: Ship it.** The fix is correct, minimal, and well-scoped. No blocking issues.

## Critical Issues

*None found.* Both agents independently confirmed the fix correctly addresses the root cause — `useRegisterFieldControl` overwriting the hidden combobox input's registration when a popup `Combobox.Input` renders the field-aware `Input` component.

## Important Issues

### 1. Missing inverse-regression test for the non-popup branch *(severity 9)*

The fix is gated on `hasPositionerParent` at `packages/react/src/combobox/input/ComboboxInput.tsx:491-497`, but no test pins the non-popup case. If someone refactors and accidentally always wraps in `DEFAULT_FIELD_ROOT_CONTEXT`, `<Combobox.Input render={<Input />} />` *outside* a popup would silently stop submitting. Suggest adding a test that renders the field-aware `Input` outside a popup and asserts `name="country"` is applied + form submits correctly.

### 2. Validation behavior with field-aware popup input is uncovered *(severity 8)*

The fix replaces the *entire* `FieldRootContext`, including `validate`, `validation.commit`, and `validationMode`. The existing test at line 5269 only covers bare `<Combobox.Input />`, not the `render={<Input/>}` variant. Suggest a test asserting `Field.Error match="valueMissing"` still surfaces on the trigger when popup input is field-aware.

### 3. Multi-select / chips path not exercised *(severity 7)*

Both agents flagged this independently. The most popular real-world combobox usage (`multiple` mode with chips) isn't covered with the popup field-aware input + form submit combination.

## Suggestions

- **Tighten the new test's assertions** — also assert the *hidden* serializer input has `name="country"` and the correct value (mirroring patterns at lines 4882, 4922). Currently only the negative assertion (`not.toHaveAttribute('name')`) is made.
- **Freeze `DEFAULT_FIELD_ROOT_CONTEXT`** (and nested `validityData` / `validation`) so accidental writes — e.g. `Field.Control` writing `validation.inputRef.current` through the noop context — fail loudly in dev. Mirrors existing `EMPTY_OBJECT` / `EMPTY_ARRAY` guards.
- **Re-mount cycle test** — the original bug "resolved on a second interaction," suggesting registration symbols can leak across mounts. A test that opens → closes → re-opens → selects → submits would lock this in.
- **No-Field.Root regression** — quick render+submit with `Combobox.Root name="country"` and no `Field.Root` ancestor.

## Strengths

- **Root cause correctly identified.** The placement of the provider is sound: `useFieldRootContext()` is called at line 60 *before* the provider takes effect, so `ComboboxInput`'s own focus/touch/validation logic still uses the outer field root — only the cloned child element sees the noop context.
- **Right scope** — `hasPositionerParent` correctly excludes the inline-combobox case (where the visible input *is* the field control).
- **Refactor of the inline default to `DEFAULT_FIELD_ROOT_CONTEXT`** follows project conventions (`DEFAULT_FIELD_ROOT_STATE` siblings) and is intentionally not exported from the package's `index.ts` — internal-only.
- **New test directly targets the user-visible failure** (form data) rather than implementation details.
- **Adheres to project guidelines** — no banned APIs introduced, no Chai/Sinon idioms, no unnecessary `flushMicrotasks`.

## Recommended Action

1. Optional: ask the PR author to add the **non-popup branch regression test** (severity 9) — it's cheap and catches the most likely future regression.
2. Optional: add a **multi-select + popup + field-aware Input** form-submit test.
3. Approve and merge — the remaining items are polish, not blockers.

PR #4690, opened 2026-04-27
[otp field] Fire onValueComplete for full paste

### LINKED ISSUE #4679: [otp] onValueComplete is not triggered when a pasted number fully replaces the previous one
# Bug report

## Current behavior

When I paste a number to the OTP field, it correctly triggers the `onValueComplete`, but if I paste a new number that fully replaces the previous number, it doesn't trigger the `onValueComplete anymore.

## Expected behavior

Whenever a complete value is pasted and completely fills the OTP field the `onValueComplete` should trigger, either that's a new value or an old value.

## Base UI version

1.4.1

Chrome 147.0.7727.103

## Which OS are you using?

Mac OS 26.3.1

atomiks:
Fixes #4679

Pasting a full OTP over an already complete value kept the field complete but skipped `onValueComplete` because completion was only queued when the previous value was incomplete. This keeps the existing transition behavior and adds a paste-specific path for complete-to-complete pastes, including when the pasted value matches the current value.

## Changes

- Fire `onValueComplete` for complete paste interactions while the field is already complete.
- Add regression coverage for replacing a complete OTP and re-pasting the same complete value.
- Update generated API docs for the callback semantics.


mj12albert (approved):
Looks good ~

PR #4693, opened 2026-04-27
[all components] Drop unnecessary memoization

flaviendelangle:
useRenderElement reads its props/state/ref only during render — it has no useMemo/useEffect/useCallback internally — so referential stability of values passed to it is never required. mergeProps always returns fresh objects too. Likewise, a useStableCallback whose only consumer is inside another useStableCallback's body, or inside a JSX-inline event handler, is dead weight: the outer wrapper (or fresh JSX render) already provides freshness.

Decision rule applied per wrapper: KEEP if consumed by a context Provider's value, another hook's deps array, an effect, an event listener stored imperatively, an external store, or useImperativeHandle. DROP otherwise.

Switch
- SwitchRoot: drop inputProps useMemo (only spread on a hidden <input>) and useStableCallback(onCheckedChangeProp) (only fed the dead useMemo).

Checkbox
- CheckboxRoot: drop useStableCallback(onCheckedChangeProp); only call site is inline in the input's onChange handler.
- CheckboxIndicator: drop stateAttributesMapping useMemo; only consumer is useRenderElement.

Radio
- RadioRoot: drop serializedValue useMemo; only consumer is the input's value prop. Inlined as serializeValue(value).

Toggle
- Toggle: drop useStableCallback(onPressedChangeProp); body inlined into the JSX-inline onClick handler (re-bound each render).

ToggleGroup
- ToggleGroup: drop defaultValue useMemo; useControlled compares defaults by serialization, not identity. Inlined with EMPTY_ARRAY for stability.

Progress
- ProgressIndicator: drop getStyles useCallback; the function was invoked inline (style: getStyles()) so its identity was never observed.

Slider
- SliderThumb: replace getThumbStyle useCallback with a let thumbStyle block; same reason as ProgressIndicator.
- SliderValue: drop outputFor and defaultDisplayValue useMemos; only consumed in the JSX prop/children passed to useRenderElement.

Popover
- PopoverTrigger: drop stateAttributesMapping useMemo; only consumer is useRenderElement.

Tabs
- TabsList: drop notifyIndicatorUpdateListeners useStableCallback; only consumer is the body of a single useEffect, which can capture the ref set directly without identity stability.
- TabsIndicator: drop activeTabPosition, activeTabSize, and style useMemos; all three are consumed only by useRenderElement.

Accordion
- AccordionRoot: drop useStableCallback(onValueChangeProp); only consumer is the body of handleValueChange (also a useStableCallback), whose trampoline already re-binds the latest closure each render.
- AccordionPanel: drop panelState useMemo; only consumed by resolveStyle and useRenderElement.
- AccordionTrigger: drop props useMemo; only passed to useRenderElement.

Collapsible
- CollapsibleTrigger: drop props useMemo; only passed to useRenderElement.
- CollapsiblePanel: drop panelState useMemo; same shape as AccordionPanel.

Menu
- MenuCheckboxItem: drop handleClick useStableCallback; only used as the JSX-inline onClick handler (re-bound each render).
- MenuLinkItem: drop state useMemo; only consumed by useRenderElement.
- MenuRadioItem: same as MenuCheckboxItem.

Select
- SelectTrigger: drop ariaControlsId useMemo; only consumed in the aria-controls JSX prop.

Drawer
- DrawerRoot: drop useStableCallback(onSnapPointChangeProp); only consumer is the body of setActiveSnapPoint (also a useStableCallback).

### DIFF
--- a/packages/react/src/toggle-group/ToggleGroup.tsx
+++ b/packages/react/src/toggle-group/ToggleGroup.tsx
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { useControlled } from '@base-ui/utils/useControlled';
+import { EMPTY_ARRAY } from '@base-ui/utils/empty';
 import { useRenderElement } from '../internals/useRenderElement';
 import type { BaseUIComponentProps, HTMLProps, Orientation } from '../internals/types';
 import { CompositeRoot } from '../internals/composite/root/CompositeRoot';
@@ -45,14 +46,6 @@ export const ToggleGroup = React.forwardRef(function ToggleGroup<Value extends s
 
   const toolbarContext = useToolbarRootContext(true);
 
-  const defaultValue = React.useMemo(() => {
-    if (valueProp === undefined) {
-      return defaultValueProp ?? [];
-    }
-
-    return undefined;
-  }, [valueProp, defaultValueProp]);
-
   const isValueInitialized = React.useMemo(
     () => valueProp !== undefined || defaultValueProp !== undefined,
     [valueProp, defaultValueProp],
@@ -62,7 +55,7 @@ export const ToggleGroup = React.forwardRef(function ToggleGroup<Value extends s
 
   const [groupValue, setValueState] = useControlled({
     controlled: valueProp,
-    default: defaultValue,
+    default: valueProp === undefined ? (defaultValueProp ?? EMPTY_ARRAY) : undefined,
     name: 'ToggleGroup',
     state: 'value',
   });
--- a/packages/react/src/accordion/panel/AccordionPanel.tsx
+++ b/packages/react/src/accordion/panel/AccordionPanel.tsx
@@ -96,13 +96,10 @@ export const AccordionPanel = React.forwardRef(function AccordionPanel(
 
   const { state, triggerId } = useAccordionItemContext();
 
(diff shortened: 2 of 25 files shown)


packages/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -62,7 +55,7 @@ export const ToggleGroup = React.forwardRef(function ToggleGroup<Value extends s
 
   const [groupValue, setValueState] = useControlled({
     controlled: valueProp,
-    default: defaultValue,
+    default: valueProp === undefined ? (defaultValueProp ?? EMPTY_ARRAY) : undefined,
```
flaviendelangle, line 58: We have a similar pattern with `useMemo` on the accordion with a comment I fail to understand 😆 
So not sure if we want to drop the memoization here.

cc @atomiks

atomiks: Accordion one returns a non-stable array (but maybe not intentionally) so it has to stay memoized unless replaced with `EMPTY_ARRAY`

atomiks (approved):

packages/react/src/toggle-group/ToggleGroup.tsx
flaviendelangle: Yes, we would need to use `EMPTY_ARRAY` before dropping memoization.
But I assumes something smarter was happening that was worth writting a comment :laughing:

Issue #4698, opened 2026-04-28
[select] Mouse selection is a no-op in dev StrictMode


rachel-church:
# Bug report

## Current behavior

When a Base UI `<Select>` is rendered in a React StrictMode dev environment (e.g. a Next.js dev server, which has `reactStrictMode: true` by default), clicking a `Select.Item` with the mouse is a no-op:

- The dropdown opens normally and the CSS `:hover` style appears on items.
- Clicking an item does **not** fire `onValueChange`. The trigger keeps its placeholder.
- Depending on which item is hovered, the dropdown either stays open or closes without committing.
- Keyboard navigation (`ArrowDown` + `Enter`) works correctly.
- Production builds (no StrictMode → single mount) work correctly.  

The root cause appears to be in `useCompositeListItem` / `CompositeList`: under StrictMode, the cleanup effect that sets `elementsRef.current = []` runs (twice, as expected), but in the Next.js dev runtime the items' ref callbacks do not re-fire to repopulate `elementsRef`. As a result `listRef.current.indexOf(liveOptionNode) === -1` inside `useListNavigation.syncCurrentTarget`, so `activeIndex` never updates on hover and `SelectItem.onClick` bails on its `pointerTypeRef.current !== 'touch' && !highlighted` guard.

## Expected behavior

Hovering a `Select.Item` updates `activeIndex` so a subsequent click commits the value — regardless of React StrictMode. Equivalently: `useCompositeListItem` registration should be StrictMode-resilient so the live DOM nodes always end up in `listRef.current`.

## Reproducible example

- **StackBlitz:** https://stackblitz.com/github/rachel-church/base-ui-select-strictmode-repro?file=src%2Fapp%2Fpage.tsx
- **GitHub:** https://github.com/rachel-church/base-ui-select-strictmode-repro

Steps:

1. Open the StackBlitz (it boots `next dev` automatically).
3. Click the **Pick a flavor** trigger.
4. Click any flavor — the trigger keeps showing the placeholder.

The component tree is:

```tsx
"use client";
import * as React from "react";
import { Select } from "@base-ui/react/select";
import { useState } from "react";

export default function Page() {
  const [value, setValue] = useState<string | null>(null);
  return (
    <React.StrictMode>
      <Select.Root value={value} onValueChange={setValue}>
        <Select.Trigger>
          <Select.Value placeholder="Pick a flavor" />
        </Select.Trigger>
        <Select.Portal>
          <Select.Positioner>
            <Select.Popup>
[...]

PR #4699, opened 2026-04-28
[select] Allow mouse selection without highlight

### LINKED ISSUE #4663: [Select][Accordion] `Select.Item` in `Accordion.Panel` is not clickable after `v1.4.0`
# Bug report

## Current behavior (`>=1.4.0`)

Image

## Expected behavior (`<1.4.0`)

Image

## Reproducible example

Example of not working (^1.4.0): https://stackblitz.com/edit/vitejs-vite-mygxv7so?file=src%2FApp.tsx
Example of working (~1.3.0): https://stackblitz.com/edit/vitejs-vite-ske857uh?file=src%2FApp.tsx&terminal=dev

## Base UI version

v1.4.0

Chrome

## Which OS are you using?

Mac OS

atomiks:
Partially addresses #4698
Closes #4663

Mouse selection could become a no-op when a Select item was not highlighted, even if the user clicked or drag-selected the item intentionally. This allows real mouse selection to commit when the press starts on the item, while keeping the guard for `alignItemWithTrigger` cases where an item appears under the cursor as the popup opens.

This does not claim to fix the underlying `[data-highlighted]` styling/registration issue; it only prevents mouse selection from bailing when highlight state is unavailable. For #4663, it only addresses pointer selection becoming a no-op; it does not make nested Accordion content a supported Select composition.

## Changes

- Allow real mouse clicks that start on a `Select.Item` to select even when the item is not highlighted.
- Allow drag-to-select to commit when `highlightItemOnHover` is disabled.
- Preserve the aligned-popup guard that ignores opening clicks that did not start on an item.
- Add regression tests for unhighlighted mouse clicks, drag-to-select, and the aligned opening-click case.

### DIFF
--- a/packages/react/src/select/item/SelectItem.test.tsx
+++ b/packages/react/src/select/item/SelectItem.test.tsx
@@ -203,6 +203,64 @@ describe('<Select.Item />', () => {
     expect(handleClick).toHaveBeenCalledOnce();
   });
 
+  it('should select an unhighlighted item with the mouse', async () => {
+    await render(
+      <Select.Root defaultOpen highlightItemOnHover={false}>
+        <Select.Trigger data-testid="trigger">
+          <Select.Value data-testid="value" />
+        </Select.Trigger>
+        <Select.Portal>
+          <Select.Positioner>
+            <Select.Popup>
+              <Select.Item value="one">one</Select.Item>
+              <Select.Item value="two">two</Select.Item>
+            </Select.Popup>
+          </Select.Positioner>
+        </Select.Portal>
+      </Select.Root>,
+    );
+
+    const option = screen.getByRole('option', { name: 'two' });
+
+    expect(option).not.toHaveAttribute('data-highlighted');
+
+    fireEvent.pointerDown(option, { pointerType: 'mouse' });
+    fireEvent.mouseDown(option);
+    fireEvent.mouseUp(option);
+    fireEvent.click(option, { detail: 1 });
+
+    await waitFor(() => {
+      expect(screen.getByTestId('value').textContent).toBe('two');
+    });
+  });
+
+  it('should ignore an unhighlighted item with a generic virtual click', async () => {
+    await render(
+      <Select.Root defaultOpen highlightItemOnHover={false}>
+        <Select.Trigger data-testid="trigger">
+          <Select.Value data-testid="value" />
+        </Select.Trigger>
+        <Select.Portal>
+          <Select.Positioner>
+            <Select.Popup>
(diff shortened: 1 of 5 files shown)


rachel-church:
Confirmed that this is a partial fix for #4698. Options can be selected with mouse events, but the hover state does not appear.

atomiks:
## Code Review (GPT-5.5)

Approve ✅ — the branch fixes the unhighlighted mouse-selection regression while preserving the alignItemWithTrigger opening-release guard and the drag-to-select fast path, with focused Select coverage passing in both JSDOM and Chromium.

### 1. Bugs / Issues (None)

No actionable issues found in the current branch diff.

The previous selection guard conflated “this item is highlighted” with “the user intended to select this item.” That broke real mouse selection whenever hover highlighting was disabled or unavailable. The patch now tracks mouse intent separately:

- real mouse clicks must start on the item via `allowMouseSelectionRef`
- drag-to-select still commits through `onMouseUp`
- generic virtual clicks remain tied to highlight state
- virtual clicks carrying pointer metadata can activate unhighlighted items for assistive technology paths
- opening-release selection is delayed for both selected and unselected items unless real drag movement opts into the unselected mouseup path sooner

Moving the drag accumulator into `selectionRef.current.dragY` keeps cross-item drag accumulation scoped to one Select root, avoiding module-level shared state across Select instances.

### Edge Cases Worth Noting

The small-movement behavior is now covered both ways: a small pointer movement does not unlock the fast drag path before the opening delay, but it also does not permanently block selection after the delay elapses.

The 400ms opening guard is simpler than the previous selected/unselected split and matches the safer behavior for cases where the popup opens under a neighboring item rather than the selected item.

### Test Coverage Assessment

The added tests cover the main regression and adjacent guardrails:

- unhighlighted mouse click selection
- generic virtual click rejection when unhighlighted
- opening-release guard for unselected items
- fast drag-to-select before the delay
- accumulated drag movement across items
- drag-to-select with hover highlighting disabled
- small pointer movement not counting as drag
- small pointer movement still selecting after the delay
- opening click that did not start on the item being ignored

Validation run on the final squashed branch:

```bash
pnpm prettier
git diff --check
pnpm test:jsdom SelectItem --no-watch
pnpm test:chromium SelectItem --no-watch
pnpm test:jsdom SelectRoot --no-watch
pnpm test:chromium SelectRoot --no-watch
pnpm typescript
pnpm size:snapshot
```

The regression tests were not run against `upstream/master`; their upstream failure is by inspection from the old `!highlighted` guard.

PR #4702, opened 2026-04-28
[select] Fix no-value popup alignment after reopen

atomiks:
This fixes a post-release regression from #4683 where a Select with no value could reopen aligned to the popup top instead of the first item. The stale selected-item text ref still clears when the value disappears, but no-value alignment now keeps a separate first-item text ref for measurement.

## Changes

- Track the first item text separately from the selected item text used for value alignment.
- Fall back to the first item text only when the select has no value.
- Add a Chromium regression test for open, close, and reopen alignment.

### DIFF
--- a/packages/react/src/select/item-text/SelectItemText.tsx
+++ b/packages/react/src/select/item-text/SelectItemText.tsx
@@ -16,23 +16,25 @@ export const SelectItemText = React.memo(
     componentProps: SelectItemText.Props,
     forwardedRef: React.ForwardedRef<HTMLDivElement>,
   ) {
-    const { indexRef, textRef, selectedByFocus, hasRegistered } = useSelectItemContext();
-    const { selectedItemTextRef } = useSelectRootContext();
+    const { index, textRef, selectedByFocus, hasRegistered } = useSelectItemContext();
+    const { firstItemTextRef, selectedItemTextRef } = useSelectRootContext();
 
     const { render, className, style, ...elementProps } = componentProps;
 
     const localRef = React.useCallback(
       (node: HTMLElement | null) => {
-        if (!node || !hasRegistered) {
+        if (!node) {
           return;
         }
-        const hasNoSelectedItemText =
-          selectedItemTextRef.current === null || !selectedItemTextRef.current.isConnected;
-        if (selectedByFocus || (hasNoSelectedItemText && indexRef.current === 0)) {
+
+        if (!firstItemTextRef.current?.isConnected || (hasRegistered && index === 0)) {
+          firstItemTextRef.current = node;
+        }
+        if (hasRegistered && selectedByFocus) {
           selectedItemTextRef.current = node;
         }
       },
-      [selectedItemTextRef, indexRef, selectedByFocus, hasRegistered],
+      [firstItemTextRef, selectedItemTextRef, index, selectedByFocus, hasRegistered],
     );
 
     const element = useRenderElement('div', componentProps, {
--- a/packages/react/src/select/item-text/SelectItemText.tsx
+++ b/packages/react/src/select/item-text/SelectItemText.tsx
@@ -27,7 +27,7 @@ export const SelectItemText = React.memo(
           return;
         }
 
-        if (!firstItemTextRef.current?.isConnected || (hasRegistered && index === 0)) {
+        if (hasRegistered && index === 0) {
           firstItemTextRef.current = node;
         }
         if (hasRegistered && selectedByFocus) {
--- a/packages/react/src/select/item/SelectItem.tsx
+++ b/packages/react/src/select/item/SelectItem.tsx
@@ -1,7 +1,6 @@
 'use client';
 import * as React from 'react';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
-import { useValueAsRef } from '@base-ui/utils/useValueAsRef';
 import { useStore } from '@base-ui/utils/store';
 import { useSelectRootContext } from '../root/SelectRootContext';
 import {
@@ -72,7 +71,6 @@ export const SelectItem = React.memo(
     const hasRegistered = index !== -1;
 
     const itemRef = React.useRef<HTMLDivElement | null>(null);
-    const indexRef = useValueAsRef(index);
 
     useIsoLayoutEffect(() => {
       if (!hasRegistered) {
@@ -225,12 +223,12 @@ export const SelectItem = React.memo(
     const contextValue: SelectItemContext = React.useMemo(
       () => ({
         selected,
-        indexRef,
+        index,
         textRef,
         selectedByFocus,
         hasRegistered,
       }),
-      [selected, indexRef, textRef, selectedByFocus, hasRegistered],
+      [selected, index, textRef, selectedByFocus, hasRegistered],
     );
 
     return <SelectItemContext.Provider value={contextValue}>{element}</SelectItemContext.Provider>;
--- a/packages/react/src/select/item/SelectItemContext.ts
+++ b/packages/react/src/select/item/SelectItemContext.ts
@@ -3,7 +3,7 @@ import * as React from 'react';
 
 export interface SelectItemContext {
   selected: boolean;
-  indexRef: React.RefObject<number>;
+  index: number;
   textRef: React.RefObject<HTMLElement | null>;
   selectedByFocus: boolean;
   hasRegistered: boolean;
--- a/packages/react/src/select/popup/SelectPopup.test.tsx
+++ b/packages/react/src/select/popup/SelectPopup.test.tsx
@@ -409,6 +409,96 @@ describe('<Select.Popup />', () => {
     },
   );
 
+  it.skipIf(isJSDOM)('keeps item text aligned with the trigger after reopening', async () => {
+    function getCenterY(rect: DOMRect) {
+      return rect.top + rect.height / 2;
+    }
+
+    const itemStyle: React.CSSProperties = {
+      padding: '8px 16px 8px 10px',
+      fontSize: 14,
+      lineHeight: '16px',
+    };
+
+    const { user } = await render(
+      <div style={{ marginLeft: 100, minHeight: 600, paddingTop: 96 }}>
+        <Select.Root>
+          <Select.Trigger
+            data-testid="trigger"
+            style={{
+              boxSizing: 'border-box',
+              width: 176,
+              height: 40,
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'flex-start',
+              paddingInlineStart: 14,
+            }}
+          >
+            <Select.Value data-testid="value" placeholder="Select produce" />
+          </Select.Trigger>
+          <Select.Portal>
+            <Select.Positioner data-testid="positioner">
+              <Select.Popup style={{ width: 220, maxHeight: 'none' }}>
+                <Select.List style={{ paddingBlock: 4, overflowY: 'auto' }}>
+                  <Select.Group>
+                    <Select.GroupLabel
+                      style={{
+                        padding: '8px 16px 4px 30px',
+                        fontSize: 11,
+                        lineHeight: '16px',
+                      }}
+                    >
+                      Fruits
+                    </Select.GroupLabel>
+                    <Select.Item value="apple" style={itemStyle}>
+                      <Select.ItemText data-testid="first-item-text">Apple</Select.ItemText>
+                    </Select.Item>
+                    <Select.Item value="banana" style={itemStyle}>
+                      <Select.ItemText data-testid="selected-item-text">Banana</Select.ItemText>
+                    </Select.Item>
+                  </Select.Group>
+                </Select.List>
+              </Select.Popup>
+            </Select.Positioner>
+          </Select.Portal>
+        </Select.Root>
+      </div>,
+    );
+
+    const trigger = screen.getByTestId('trigger');
+
+    async function expectItemAligned(testId: string) {
+      const value = screen.getByTestId('value');
+      const itemText = screen.getByTestId(testId);
+
+      await waitFor(() => {
+        expect(screen.getByTestId('positioner')).toHaveAttribute('data-side', 'none');
+        expect(
+          Math.abs(
+            getCenterY(value.getBoundingClientRect()) -
+              getCenterY(itemText.getBoundingClientRect()),
+          ),
+        ).toBeLessThan(1);
+      });
+    }
+
+    await user.click(trigger);
+    await expectItemAligned('first-item-text');
+
+    await user.keyboard('{Escape}');
+    await waitFor(() => expect(trigger).toHaveAttribute('aria-expanded', 'false'));
+
+    await user.click(trigger);
+    await expectItemAligned('first-item-text');
+
+    await user.click(screen.getByTestId('selected-item-text'));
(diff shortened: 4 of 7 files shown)


flaviendelangle (commented):
# PR #4702 Review Summary — `[select] Fix no-value popup alignment after reopen`

Four review agents (code-reviewer, test-analyzer, silent-failure-hunter, comment-analyzer) examined the diff. Findings converge on a few concerns.

## Critical Issues
None. The fix is correctly scoped — the new test confirms both halves are load-bearing (revert either side and it fails).

## Important Issues

1. **`firstItemTextRef` is never cleared** — flagged by code-reviewer + silent-failure-hunter
   - [SelectItemText.tsx:30](packages/react/src/select/item-text/SelectItemText.tsx#L30), [SelectRoot.tsx:120](packages/react/src/select/root/SelectRoot.tsx#L120)
   - Only stale-detection is `?.isConnected`, which catches DOM unmount but **not reorder**: if the index-0 item moves to index 5 with its DOM node intact (re-keyed list, conditional rendering, virtualization), the ref keeps pointing at it. Result: silent few-pixel misalignment.
   - Suggested fix: mirror the `selectedItemTextRef.current = null` reset in [SelectRoot.tsx:222-223](packages/react/src/select/root/SelectRoot.tsx#L222-L223), or null on unmount via the ref-callback cleanup.

2. **Missing test: value-present alignment after reopen** — flagged by test-analyzer
   - The PR also rewrote the `selectedItemTextRef` assignment branch in [SelectItemText.tsx:33](packages/react/src/select/item-text/SelectItemText.tsx#L33) (now gated on `hasRegistered && selectedByFocus` only). All existing alignment tests in [SelectPopup.test.tsx](packages/react/src/select/popup/SelectPopup.test.tsx) use `<Select.Root open>` one-shot — no test exercises open / close / reopen with a `defaultValue` set. A regression here would be silent for users with selected values.

3. **Race: registered-but-no-text-ref with `hasSelectedValue=true`** — flagged by silent-failure-hunter
   - [SelectPopup.tsx:289-297](packages/react/src/select/popup/SelectPopup.tsx#L289-L297) gates the `firstItemTextRef` fallback on `!hasSelectedValue`. If a controlled value is set but `selectedItemTextRef` hasn't been populated yet (no `selectedByFocus` pass before the layout effect runs), `textElement` is null and the alignment math is silently skipped — popup falls back to plain trigger-edge positioning.

## Suggestions

- **Multi-select + empty array**: `selectors.hasSelectedValue` returns `false` for `[]`, so the new fallback engages. No alignment test covers `<Select.Root multiple defaultValue={[]}>`. Worth a small assertion. ([SelectRoot.test.tsx:2307](packages/react/src/select/root/SelectRoot.test.tsx#L2307) tests `data-filled` only.)
- **Invalid controlled value**: value set, no matching item → `selectedIndex=null`, `selectedItemTextRef=null`, `hasSelectedValue=true` → fallback refused → alignment skipped. Acceptable as graceful degrade, but a dev-mode warning in `syncSelectedIndex` ([SelectRoot.tsx:222](packages/react/src/select/root/SelectRoot.tsx#L222)) when `value != null && nextIndex === null` would surface the root cause.
- **Inline test boilerplate**: `gridTemplateColumns: '12px 1fr'` and the leading `<span />` in [SelectPopup.test.tsx:417-460](packages/react/src/select/popup/SelectPopup.test.tsx#L417-L460) don't contribute to vertical-center geometry — could be trimmed.
- **Optional WHY comment** on [SelectItemText.tsx:30](packages/react/src/select/item-text/SelectItemText.tsx#L30): the two-clause condition (provisional capture + upgrade-on-registration) is non-obvious; a future "simplifier" could collapse it to `hasRegistered && indexRef.current === 0` and silently break first-open alignment.

## Strengths

- Test directly reproduces the regression and is correctly Chromium-scoped via `it.skipIf(isJSDOM)`.
- Clean separation: `firstItemTextRef` (eager fallback) vs `selectedItemTextRef` (cleared when no selection) preserves #4683's intent without overloading a single ref.
- `isConnected` checks at read sites are defensive against detached nodes.
- Dependency arrays correctly extended with `firstItemTextRef` and `hasSelectedValue` ([SelectPopup.tsx:438,450](packages/react/src/select/popup/SelectPopup.tsx#L438-L450)).
- `HTMLSpanElement → HTMLElement` widening in [SelectRootContext.ts:36-37](packages/react/src/select/root/SelectRootContext.ts#L36-L37) corrects a misleading prior type (SelectItemText renders a `<div>`).
- No comments added; no pre-existing comments rendered stale by the change. Fully complies with project comment policy.
- All AGENTS.md hooks compliance checks pass: `useIsoLayoutEffect`, `useStableCallback`, `useStore`, `ownerDocument`/`ownerWindow` used throughout; no `as any`, no raw `setTimeout`/`requestAnimationFrame`.

## Recommended Action
1. Add a clear of `firstItemTextRef` on close (parallel to the existing `selectedItemTextRef` clear) — addresses the highest-confidence latent bug.
2. Add a Chromium test for open/close/reopen alignment **with a `defaultValue` set** — the PR rewrote that path too but leaves it untested.
3. Consider a dev-mode warn for the invalid-controlled-value path.
4. Other items above are nice-to-haves and not blockers for merge.

flaviendelangle (approved):
# PR #4702 Review Summary

**PR:** [mui/base-ui#4702 — `[select] Fix no-value popup alignment after reopen`](https://github.com/mui/base-ui/pull/4702)
**Author:** atomiks
**Commit reviewed:** `7d90d803d`
**Files changed:** 7 files, +120 / -15

**Goal:** Fix a post-release regression from #4683 where a Select with no value could reopen aligned to the popup top instead of the first item. Adds a separate `firstItemTextRef` (independent of `selectedItemTextRef`) so that no-value alignment has its own measurement anchor.

---

## Review aspects run

- code-reviewer (general project guidelines + correctness)
- pr-test-analyzer (test coverage of the new code paths)
- silent-failure-hunter (silent fallbacks and stale state)
- comment-analyzer (comment accuracy and missing rationale)

> Note: type-design-analyzer was not run because the PR introduces no new types — only a widening of an existing ref type (`HTMLSpanElement` → `HTMLElement`).

---

## Critical Issues

None of the agents flagged a hard blocker. The closest to one is the silent-failure-hunter's findings about ref cleanup.

### [silent-failure-hunter] No unmount cleanup; refs persist as stale, disconnected nodes
**Where:** [SelectItemText.tsx:24-38](packages/react/src/select/item-text/SelectItemText.tsx#L24-L38)
The new `localRef` returns early on `node === null`, so neither `firstItemTextRef.current` nor `selectedItemTextRef.current` are cleared when an item unmounts. The fix relies entirely on `isConnected` checks at *read time* in `SelectPopup` (and on the next mount overwriting). This is the same class of bug the PR is patching — catching the disconnect at write time would close the loop properly.

**Suggested fix:** return a cleanup from the callback ref that nulls each ref it owns when the same node is referenced.

```ts
const localRef = React.useCallback((node) => {
  if (!node) return;
  if (!firstItemTextRef.current?.isConnected || (hasRegistered && index === 0)) {
    firstItemTextRef.current = node;
  }
  if (hasRegistered && selectedByFocus) {
    selectedItemTextRef.current = node;
  }
  return () => {
    if (firstItemTextRef.current === node) firstItemTextRef.current = null;
    if (selectedItemTextRef.current === node) selectedItemTextRef.current = null;
  };
}, [...]);
```

> **Note (assessor):** code-reviewer rated the same observation as low-confidence (~45) because the read-time `isConnected` guard is sufficient in practice. Treat as a quality-of-implementation improvement, not a blocker.

---

## Important Issues

### [silent-failure-hunter] Selected-value-but-stale-text silently degrades to no-alignment
**Where:** [SelectPopup.tsx:289-297](packages/react/src/select/popup/SelectPopup.tsx#L289-L297)

```ts
let textElement = selectedItemTextRef.current;
if (!textElement?.isConnected) {
  textElement = !hasSelectedValue && firstItemTextRef.current?.isConnected
    ? firstItemTextRef.current
    : null;
}
```

When `hasSelectedValue === true` but `selectedItemTextRef.current` is stale/disconnected (virtualized lists, controlled value pointing to an unregistered item), `textElement` becomes `null`, the `if (textElement && valueElement) { ... }` block at ~line 333 is skipped, the `--transform-origin` block at ~line 400 is also skipped, and the popup opens visibly mis-aligned with no diagnostic.

**Suggested fix:** in development, emit a `console.warn` or `logError` when this case is hit, so virtualization/controlled-value bugs become visible. Document the intentional null-out with a one-liner.

### [comment-analyzer] Non-obvious branch mixing two concerns lacks a WHY comment
**Where:** [SelectItemText.tsx:30](packages/react/src/select/item-text/SelectItemText.tsx#L30)

```ts
if (!firstItemTextRef.current?.isConnected || (hasRegistered && index === 0)) {
```

The OR mixes two distinct concerns:
1. "Nothing valid is currently stored" (initial mount or stale node from a previous open cycle).
2. "Once the registered first item is known, it should win even if a placeholder was stored earlier."

A future maintainer is likely to "simplify" this to `hasRegistered && index === 0` and reintroduce the very bug being fixed. A two-line comment explaining the seed-then-upgrade intent would protect this code from regression.

### [silent-failure-hunter] Order-dependent claim of `firstItemTextRef`
**Where:** [SelectItemText.tsx:30](packages/react/src/select/item-text/SelectItemText.tsx#L30)

The current condition lets *any* item with a connected node take ownership of `firstItemTextRef` whenever the existing ref is null/disconnected. React mount order normally matches `index` order, but suspense/portals/async list registration can break that assumption. During the window before the index-0 item's ref callback fires, alignment may use the wrong item.

**Suggested fix:** make the assignment deterministic — `if (hasRegistered && index === 0) firstItemTextRef.current = node;` — and pair it with the unmount cleanup from the critical issue above. The combination removes the `!isConnected` half of the OR.

> **Note (assessor):** code-reviewer also looked at this and concluded the `(hasRegistered && index === 0)` branch self-corrects on subsequent renders. The two views agree on the timing of correction; they disagree on whether the transient window is acceptable.

---

## Suggestions

### [pr-test-analyzer] Coverage gaps for the new fallback paths
None are critical; all are nice-to-haves.

1. **Unmount of the first item between close and reopen** — exercise the `null` fallback path in `SelectPopup` and confirm the popup still opens (uncentered) without crashing.
2. **`defaultValue` set + LTR alignment** — pin the contract that the *selected* item, not the first item, becomes the anchor on initial open even though `firstItemTextRef` is also populated. Existing rtl test at [SelectPopup.test.tsx:309](packages/react/src/select/popup/SelectPopup.test.tsx#L309) only covers RTL.
3. **Disabled first item** — does it still become the alignment anchor? Currently un-tested and likely un-decided.
4. **Multi-select with no values** — `multiple` Select with empty array should also fall back to first item.
5. **Items inserted dynamically while popup is open** — exercise the "first item registered after a placeholder" path.
6. **Minor:** the new test renders a `Select.Group` + `Select.GroupLabel`; not strictly required to reproduce the regression and could be removed for clarity.

### [comment-analyzer] Add brief WHY comments

- [SelectPopup.tsx:292-297](packages/react/src/select/popup/SelectPopup.tsx#L292-L297) — explain that the `!hasSelectedValue` guard is load-bearing: when a value *is* selected but its text is gone, intentionally bail to `null` rather than misalign on the first item.
- [SelectRoot.tsx:120-121](packages/react/src/select/root/SelectRoot.tsx#L120-L121) — one-liner distinguishing `firstItemTextRef` (anchor when nothing is selected) from `selectedItemTextRef` (anchor for the selected value), since the names alone don't convey their alignment role.

### [code-reviewer] Lower-confidence observations
- `firstItemTextRef` assignment when `!isConnected` does not gate on `hasRegistered`. Confidence ~55. The `(hasRegistered && index === 0)` clause self-corrects.
- `localRef` re-attaches when `index` changes (now in deps). Verified to self-correct on reorder.

---

## Strengths

- **The test at [SelectPopup.test.tsx:412](packages/react/src/select/popup/SelectPopup.test.tsx#L412) directly reproduces the regression** (open → close → reopen with no value, then select → close → reopen). Without the SelectPopup fallback to `firstItemTextRef`, the second open would mis-align — the assertion catches it. (pr-test-analyzer, code-reviewer)
- **Behavioral, not implementation-coupled.** Asserts geometry (center-Y alignment) via the public API only — no white-box hooks or internal-ref reads. Survives refactor of the ref tracking. (pr-test-analyzer)
- **`it.skipIf(isJSDOM)` is correct** since the assertion depends on real layout, per AGENTS.md. (pr-test-analyzer)
- **Removal of `useValueAsRef(index)` in favor of `index` directly** is a simplification. Depending on a primitive `index` is more idiomatic than ref-as-value. (code-reviewer)
- **Type widening from `HTMLSpanElement` to `HTMLElement`** at [SelectRoot.tsx:120](packages/react/src/select/root/SelectRoot.tsx#L120) and [SelectRootContext.ts:36](packages/react/src/select/root/SelectRootContext.ts#L36) is a corrective change — `SelectItemText` actually renders a `div` via `useRenderElement('div', ...)`. (code-reviewer)
- **Public API surface unchanged** — both refs are context-internal, no `pnpm docs:api` regeneration needed. (code-reviewer)
- **Pre-existing comment style at SelectPopup.tsx:284-285** ("Ensure we remove any transforms…") remains accurate and explains WHY rather than WHAT — a good local model for the new code to follow. (comment-analyzer)
- **AGENTS.md compliance:** uses `useIsoLayoutEffect` and `useStableCallback` correctly throughout; no new violations. (code-reviewer)

---

## Recommended Action

1. **Optional but worthwhile before merge:**
   - Add unmount cleanup to `SelectItemText` `localRef` (closes the stale-ref loop properly; also addresses the silent-failure-hunter critical and the order-dependent claim issue).
   - Add a 2-line comment at `SelectItemText.tsx:30` explaining the seed-then-upgrade intent of the OR.
   - Add a 1-line comment at `SelectPopup.tsx:292-297` explaining why `!hasSelectedValue` is load-bearing.

2. **Nice-to-have for follow-up (not blocking):**
   - Dev-only warning in `SelectPopup.tsx` when `hasSelectedValue && !selectedItemTextRef.current?.isConnected` — surfaces virtualization/controlled-value bugs that currently degrade silently.
   - One additional regression test for unmount-between-close-and-reopen.
   - Consider whether disabled first item should be the alignment anchor (decision + test).

3. **Approve as-is is also defensible.** The fix is logically sound, the regression is locked down by a behavioral test, and the silent-failure concerns are about hardening the code path rather than fixing an active bug. The code-reviewer agent's bottom line was "Recommend approval."

---

## Files reviewed

- [packages/react/src/select/item-text/SelectItemText.tsx](packages/react/src/select/item-text/SelectItemText.tsx)
- [packages/react/src/select/item/SelectItem.tsx](packages/react/src/select/item/SelectItem.tsx)
- [packages/react/src/select/item/SelectItemContext.ts](packages/react/src/select/item/SelectItemContext.ts)
- [packages/react/src/select/popup/SelectPopup.tsx](packages/react/src/select/popup/SelectPopup.tsx)
- [packages/react/src/select/popup/SelectPopup.test.tsx](packages/react/src/select/popup/SelectPopup.test.tsx)
- [packages/react/src/select/root/SelectRoot.tsx](packages/react/src/select/root/SelectRoot.tsx)
- [packages/react/src/select/root/SelectRootContext.ts](packages/react/src/select/root/SelectRootContext.ts)

PR #4704, opened 2026-04-29
[tabs] Fire onValueChange for automatic tab selection (alternative)

### LINKED ISSUE #2097: [tabs] `onValueChange` does not fire when the initially selected tab is picked by the `Tabs` component

michaldudak:
## Summary

This adds `onValueChange` notifications for automatic Tabs value changes while keeping the implementation intentionally small.

Tabs can already change their selected value without direct user interaction, for example when the initially selected tab is disabled, when the selected tab becomes disabled, or when the selected tab is removed. This PR makes those automatic changes observable through `onValueChange`, using shared Base UI reason constants:

- `initial`: automatic initial selection when tabs first register
- `disabled`: selected tab became disabled
- `missing`: selected tab was removed

Controlled Tabs keep ownership of `value`, so automatic fallback notifications are only emitted for uncontrolled roots.

## Behavior

- Notifies `onValueChange` when an uncontrolled root selects the first enabled tab automatically.
- Notifies `onValueChange(null, { reason: 'initial' })` when all tabs are initially disabled.
- Notifies `onValueChange` when the selected uncontrolled tab becomes disabled or missing.
- Keeps automatic event details non-cancelable, since these transitions are internal consistency updates rather than user events.
- Keeps activation direction stable for automatic fallbacks.
- Preserves controlled panel visibility semantics.

---

This is a smaller alternative to #3758.
Closes #2097.


atomiks:
## Codex Review (GPT-5.5)

Reviewed the full PR against freshly fetched `upstream/master`. This is primarily a `bug-fix` PR with public API/docs and test coverage impact: it makes automatic Tabs value changes observable through `onValueChange`.

### 1. Bugs / Issues

#### 1. 🔴 Automatic fallback changes ignore `eventDetails.cancel()` (blocking)

`packages/react/src/tabs/root/TabsRoot.tsx`

```tsx
const notifyValueChange = (nextValue: TabsTab.Value, reason: TabsRoot.ChangeEventReason) => {
  onValueChangeProp?.(
    nextValue,
    createChangeEventDetails(reason, undefined, undefined, {
      activationDirection: 'none',
    }),
  );
};

notifyValueChange(fallbackValue, selectionIsDisabled ? REASONS.disabled : REASONS.missing);

setValue(fallbackValue);
```

The new auto-selection path creates normal cancelable change details, but then discards them and always commits `setValue(fallbackValue)`. That differs from the existing user-initiated Tabs path, where `eventDetails.cancel()` prevents the internal update.

**Fix:** Return the created details from `notifyValueChange`, and skip `setValue` plus the activation-direction reset when `details.isCanceled` is true. Add regression coverage for canceling an automatic disabled/missing fallback.

#### 2. 🟡 Initial all-disabled fallback changes value without firing `onValueChange` (non-blocking)

`packages/react/src/tabs/root/TabsRoot.tsx`

```tsx
if (shouldNotifyInitialValueChange) {
  shouldNotifyInitialValueChangeRef.current = false;

  if (firstEnabledTabValue !== undefined) {
    notifyValueChange(firstEnabledTabValue, REASONS.initial);
  }
}

if (!shouldNotifyInitialValueChange) {
  notifyValueChange(fallbackValue, selectionIsDisabled ? REASONS.disabled : REASONS.missing);
}

setValue(fallbackValue);
```

When all tabs are disabled, `firstEnabledTabValue` is `undefined`, `fallbackValue` becomes `null`, and `setValue(null)` still runs. Because this is also the initial notification pass, no `onValueChange` event is emitted for that automatic transition.

**Fix:** Ensure every automatic `setValue`, including the initial all-disabled `null` fallback, emits exactly one change event. Add an all-disabled test with no `defaultValue`.

#### 3. 🟡 New reason constants are missing from the shared event map (non-blocking)

`packages/react/src/internals/reason-parts.ts`

```ts
export const disabled = 'disabled' as const;
export const missing = 'missing' as const;
export const initial = 'initial' as const;
```

`packages/react/src/internals/createBaseUIEventDetails.ts`

```ts
[REASONS.disabled]: Event;
[REASONS.imperativeAction]: Event;
```

The PR exports `missing` and `initial` as shared Base UI reason constants, but does not add them to `ReasonToEventMap`. Tabs still gets `Event` through the fallback conditional type, but the central reason-to-event registry becomes incomplete and those reasons are not available to helpers constrained to `keyof ReasonToEventMap`.

**Fix:** Add `[REASONS.missing]: Event` and `[REASONS.initial]: Event` to `ReasonToEventMap`.

### Test Coverage Assessment

I ran `pnpm test:jsdom TabsRoot --no-watch`; the TabsRoot suite reported 90 passed and 25 skipped. The added tests cover the main enabled, disabled-first, explicit `defaultValue`, `defaultValue={undefined}`, disabled-after-mount, and removed-tab paths, but they miss the cancellation and all-disabled `null` fallback cases above.

### Recommendation

Request changes 🔴

The main behavior works for the covered paths, but the automatic change path needs to preserve Tabs’ existing cancelable `onValueChange` contract before this lands.

atomiks (approved):
Approved based on the latest re-review. The earlier automatic fallback issues are addressed, and I validated the updated head with `pnpm test:jsdom TabsRoot --no-watch`, `pnpm test:chromium TabsRoot --no-watch`, and `pnpm typescript`. The remaining SSR/pre-hydration behavior is the existing documented caveat rather than a new blocker from this PR.

PR #4710, opened 2026-04-29
Fix outdated accesibility information in color contrast section

markmiro:
Removed reference to APCA becoming the new standard in WCAG 3. APCA was removed from the WCAG 3 draft in July 2023. It may come back, as the APCA algorithm has had updates since then. The contrast algorithm used in WCAG 3 has not been determined.

Source:
https://adrianroselli.com/2026/04/wcag3-contrast-as-of-april-2026.html

### DIFF
--- a/docs/src/app/(docs)/react/overview/accessibility/page.mdx
+++ b/docs/src/app/(docs)/react/overview/accessibility/page.mdx
@@ -31,7 +31,7 @@ WCAG provides [guidelines on focus appearance](https://www.w3.org/WAI/WCAG22/Und
 ## Color contrast
 
 When styling elements, it's important to meet the minimum requirements for color contrast between each foreground element and its corresponding background element.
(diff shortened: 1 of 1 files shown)


markmiro:
I make this PR because I almost mentioned in my company Slack that we should use APCA over WCAG 2 for accessibility since it'll be part of WCAG 3. I've personally not liked WCAG 2's contrast guidelines because I don't find them ideal for accessible contrast, so APCA is promising.

However, some key points:
- WCAG 3 is a long ways away
- APCA has been removed from the WCAG 3 draft (though it may come back because the version removed was obsolete)

markmiro:
CI failing because there's no tags. Probably these two would fit but I don't have a way to add tags to the PR:
- docs
- accessibility

colmtuite (approved):

PR #4712, opened 2026-04-30
[popups] Consider the controlled `open` prop for open state detection

### LINKED ISSUE #4709: [menu] Keep `[data-popup-open]` on trigger when `Menu.Popup` is open
# Bug report

### Context

Let's consider a `Menu` which has a bunch of actions a user can take. 
One of them is destructive and requires their confirmation. While the user sees the confirmation dialog, the `Menu` should remain open.

For that, I used the `Menu.Item` with the `closeOnClick={false}`. Additionally, to prevent closure on any interactions within the dialog, I overrode the `onOpenChange` with:
```tsx
const [menuOpen, setMenuOpen] = useState(false)
const handleMenuOpenChange = useCallback(
  (open: boolean, { reason, event }: Menu.Root.ChangeEventDetails) => {
    if (open) {
      setMenuOpen(true)
      return
    }

    if (
      reason === 'outside-press' &&
      event.target instanceof HTMLElement &&
      event.target.closest(DIALOG_POPUP_SELECTOR)
    ) {
      return
    }

    setMenuOpen(open)
  },
  []
)
```

### Issue
When a user clicks somewhere within the confirmation dialog, the `[data-popup-open]` is removed from the `Menu.Trigger` element ⚠️ 

| Action | Header |
|--------|--------|
| User clicks on the trigger to open the `Menu` | Image |
| User clicks within the confirmation dialog | <video src="https://github.com/user-attachments/assets/cc646d44-a524-4d56-ae52-bbaf7c888641" /> | 

With the `[data-popup-open]` removed, the trigger loses the "active" styling too. 
Even when the confirmation dialog is closed, the `[data-popup-open]` isn't getting restored.

## Expected behavior

Instead, I'd expect the `[data-popup-open]` to remain on the `Menu.Trigger` element when the `Menu.Popup` is present. External interactions shouldn't impact that attribute.

## Reproducible example

https://stackblitz.com/edit/tc57mcm9?file=src%2FApp.tsx

## Base UI version

`@base-ui/react@^1.4.1`

Chrome/Safari/Firefox.

## Which OS are you using?

Mac OS.

## Relations
This issue is related to the https://github.com/mui/base-ui/issues/1761. There, a similar unexpected `[data-popup-open]` cleanup happens, but for the `Select` component.

## Workaround
Fortunately, there's a workaround recommended by @atomiks: https://github.com/mui/base-ui/issues/1761#issuecomment-2821243308

michaldudak:
Fixes the controlled popup trigger state after a close request is vetoed.

Updated shared popup trigger state to derive `isOpenedByTrigger` from the effective open state, `openProp ?? open`, instead of raw internal `open`.

Fixes #4709


michaldudak:
## Codex Review (GPT-5.5)
This follow-up review covers the full PR branch against freshly fetched `upstream/master`. The implementation now addresses the prior public-handle mismatch and I did not find additional blocking or non-blocking issues.

No remaining actionable issues found.

The prior issue with public handles reporting stale raw `store.state.open` after a controlled close veto is resolved. `DialogHandle`, `MenuHandle`, `PopoverHandle`, `PreviewCardHandle`, and `TooltipHandle` now read the effective popup state via `store.select('open')`, matching the selector-backed state used by rendered popup parts and triggers.

The core fix is consistent: `isOpenedByTrigger` now uses the same effective open calculation as the public `open` selector, so controlled `openProp` takes precedence over attempted internal state changes. The extracted `getOpen` helper keeps the selector logic shared and avoids a second raw `openProp ?? open` implementation.

The handle updates are also aligned with that model. Since the affected stores are constructed with popup selectors, `this.store.select('open')` is valid for both internally owned and handle-owned stores.

### Test Coverage Assessment

The added tests cover the low-level selector cases and the controlled close-veto flow through Menu and AlertDialog, including `handle.isOpen`. I also validated the broader handle/detached-trigger surface.

Commands run:

```txt
pnpm typescript
pnpm test:jsdom MenuTrigger AlertDialogRoot popups/store --no-watch
pnpm test:jsdom detached-triggers --no-watch
pnpm test:chromium detached-triggers --no-watch
git diff --check upstream/master...origin/menu-popup-open
```

All passed.

### Recommendation

Approve ✅

The implementation is small, consistent with the existing popup store architecture, and the previously identified public API mismatch is fixed with targeted coverage. I don’t see further issues that should block this PR.

mj12albert (approved):
Looks good 👍

PR #4713, opened 2026-04-30
[checkbox] Submit the associated form on Enter key press

### LINKED ISSUE #4677: [checkbox] Pressing Enter does not submit the form
## Current behavior

Pressing Enter on a focused base-ui `<Checkbox>` in the form does not submit the form.

## Expected behavior

Pressing Enter on a focused base-ui `<Checkbox>` should submit the form, consistent with native checkbox behavior. 

## Reproducible example

https://stackblitz.com/edit/jnekgnni?file=src%2FApp.tsx

## Base UI version

1.4.1

Chrome

## Which OS are you using?

Mac OS

## Additional context

I found that a similar issue was previously reported/observed in React Spectrum as well. It seems like the same problem exists in this library.
https://github.com/adobe/react-spectrum/issues/6235

michaldudak:
Matches the native `<input type="checkbox">` behavior regarding form submission. <kbd>Enter</kbd> no longer toggles the checkbox but submits the associated form. It supports both parent forms and the ones linked with the `form` attribute.

Added a `getDefaultFormSubmitter` utility to `@base-ui/utils`, so it can be used in other repos (cc @flaviendelangle)

Fixes #4677.


packages/utils/src/getDefaultFormSubmitter.test.ts
```diff
@@ -0,0 +1,71 @@
+import { expect } from 'vitest';
+import { getDefaultFormSubmitter } from './getDefaultFormSubmitter';
+
+describe('getDefaultFormSubmitter', () => {
+  it('returns the first submit button associated with the form', () => {
+    document.body.innerHTML = [
```
mj12albert: Maybe something like this is nicer to read than Array+join
```ts
function setup(html: string) {
  document.body.innerHTML = html;
  return document.querySelector<HTMLFormElement>('#test-form')!;
}

afterEach(() => {
  document.body.innerHTML = '';
});

it('returns the first submit button associated with the form', () => {
  const form = setup(`
    <button id="external-before" form="test-form" type="submit">External before</button>
    <form id="test-form">
      <input type="checkbox" />
      <button id="internal" type="submit">Internal</button>
    </form>
  `);
});
```

packages/utils/src/getDefaultFormSubmitter.ts
```diff
@@ -0,0 +1,40 @@
...
+
+      if (button.type === 'submit') {
+        return button;
+      }
+    }
+
+    if (tagName === 'INPUT') {
+      const input = candidate as HTMLInputElement;
+
+      if (input.type === 'submit') {
```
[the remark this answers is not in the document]
michaldudak: I chose to ignore `input[type=image]` as Chromium omits it from `form.elements` and it would have to be handled separately. Considering that it's quite exotic, I don't want to inflate the bundle because of it.

packages/utils/src/getDefaultFormSubmitter.ts
```diff
@@ -0,0 +1,40 @@
...
+        return input;
+      }
+    }
```
romgrk: Could these branches be merged? It would save a bit of bundle size.

packages/utils/src/getDefaultFormSubmitter.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * This is useful for custom form controls that need to mirror native Enter key behavior.
+ * Clicking the returned submitter preserves browser semantics such as the submitter's click
+ * event, `SubmitEvent.submitter`, and submitter-specific attributes.
+ *
+ * The function follows the controls exposed by `form.elements`, which includes controls associated
+ * through the `form` attribute. Disabled submitters can be returned because the default button is
+ * determined before disabled state is considered; clicking a disabled submitter is a no-op.
+ */
+export function getDefaultFormSubmitter(form: HTMLFormElement | null): DefaultFormSubmitter | null {
```
flaviendelangle, line 14: Do you see a different with this one liner?

```ts
const defaultButton = form.querySelector<DefaultFormSubmitter>(
  'input[type="submit"], button[type="submit"], button:not([type])',
);
```

packages/utils/src/getDefaultFormSubmitter.ts
```diff
@@ -0,0 +1,32 @@
...
+    return null;
+  }
+
+  for (const candidate of form.elements) {
+    const tagName = candidate.tagName;
+
+    if (tagName === 'BUTTON' || tagName === 'INPUT') {
+      const button = candidate as HTMLButtonElement | HTMLInputElement;
+
+      if (button.type === 'submit') {
```
flaviendelangle, line 27: If there is no type then it also submits (unless `button.type` defaults to `submit` when the DOM element doesn't have its type explicitely defined, I did not check)

packages/utils/src/getDefaultFormSubmitter.ts
michaldudak: Yes. This code only finds submit buttons within a form, not the ones associated with it via the `form` attribute.

packages/utils/src/getDefaultFormSubmitter.ts
michaldudak: Yup, the `type` property is set to `"submit"` by default.

packages/utils/src/getDefaultFormSubmitter.ts
flaviendelangle: Wow I did not know about this feature :laughing:

flaviendelangle (commented):
# PR Review: [checkbox] Submit the associated form on Enter key press

**PR:** [mui/base-ui#4713](https://github.com/mui/base-ui/pull/4713)
**Scope:** 4 files, +366/-14 — checkbox Enter behavior + new utility

---

## Critical Issues

### 1. Missing `input type="image"` as a submit button

`getDefaultFormSubmitter` only checks for `type === 'submit'`, but per the HTML Living Standard, **both `submit` and `image` are considered submit buttons**:

> The following input types are considered submit buttons: `submit`, `image`

Real-world example:
```html
<form>
  <input type="image" src="go.png" alt="Go">
</form>
```

Native browsers will use `<input type="image">` as the default submitter when there's no other submit button. The PR's utility would return `null` here, diverging from native behavior.

**Suggested fix** in `getDefaultFormSubmitter.ts`:
```ts
if (tagName === 'BUTTON') {
  const button = candidate as HTMLButtonElement;
  if (button.type === 'submit') return button;
} else if (tagName === 'INPUT') {
  const input = candidate as HTMLInputElement;
  if (input.type === 'submit' || input.type === 'image') return input;
}
```

A test for `<input type="image">` should also be added.

---

## Important Issues

### 2. `event.preventDefault()` is called even when there is no form

In `CheckboxRoot.tsx`:
```ts
event.preventDefault();
getDefaultFormSubmitter(inputRef.current?.form ?? null)?.click();
```

When the checkbox is rendered with `nativeButton` outside any form, `preventDefault()` will swallow the Enter keypress without performing any action. That mirrors the goal ("Enter must not toggle the checkbox") so the resulting behavior is fine, but the *intent* of the unconditional `preventDefault()` could be clearer. Consider hoisting the rationale into a comment, or only calling `preventDefault()` when there's actually a form to submit. (Behavior-equivalent today, but easier to reason about.)

### 3. Inconsistent assertions across new tests

Most new Enter-submit tests assert `expect(checkbox).toHaveAttribute('aria-checked', 'false')` afterwards to confirm the checkbox didn't toggle. But the new `submits to an external form with Enter when 'form' is provided` test (`packages/react/src/checkbox/root/CheckboxRoot.test.tsx`, around L670–698) is missing that assertion. Add it for consistency and to defend against a regression where Enter both submits and toggles.

---

## Suggestions

### 4. Iteration could be slightly clearer

The current pattern works, but the local `tagName` + cast pattern is a touch awkward. A narrower check via DOM type guards would mix concerns. Either keep the current shape, or compress to:

```ts
for (const candidate of form.elements) {
  if (candidate instanceof HTMLButtonElement && candidate.type === 'submit') {
    return candidate;
  }
  if (candidate instanceof HTMLInputElement &&
      (candidate.type === 'submit' || candidate.type === 'image')) {
    return candidate;
  }
}
```

Note: if you adopt `instanceof`, `form.ownerDocument.defaultView`'s constructors should be used for cross-realm safety (matching base-ui-plus convention; check whether base-ui has the same rule).

### 5. JSDoc nit

The doc says:
> Disabled submitters can be returned because the default button is determined before disabled state is considered; clicking a disabled submitter is a no-op.

This is accurate, and the test `does not submit the form with Enter when the default button is disabled` is the right behavioral assertion. Consider adding a one-line note pointing readers to that test (or vice versa) so future maintainers understand the "no-op click" is intentional, not a bug.

---

## Strengths

- **Comprehensive test matrix.** Covers: focus + keyboard, consumer `preventDefault`, `readOnly`, `nativeButton` (with the important "submits once" check), no submit button, disabled default submitter, and external `form=` association. This is the right shape for a behavior change like this.
- **`preventBaseUIHandler` + `defaultPrevented` order is correct.** Calling `preventBaseUIHandler` first defensively stops Base UI's button-style Enter handling regardless of what the consumer does, then the `defaultPrevented` check honors the consumer's opt-out before attempting submission. Good escape hatch design.
- **Using `.click()` on the default submitter (rather than `form.requestSubmit()`)** correctly preserves browser-native semantics: the submitter's own click handlers run, `SubmitEvent.submitter` is set, and submitter-specific attributes (`formaction`, `formmethod`, etc.) are honored. This is the subtle correct choice.
- **Refactor of the previous `['Enter', 'Space'].forEach` parametric test** into two separate tests is clean and clearly states the new contract (Space activates, Enter does not).
- **The bottom-of-file integration test was updated correctly** to flip the Enter/Space/click expectations rather than being deleted.
- **Comment on the `onKeyDown` handler explains *why*** (consumer opt-out + defensive Base UI handler suppression) rather than what — matches the comment-quality bar.

---

## Recommended Action

1. **Add `input type="image"` support** in `getDefaultFormSubmitter` and a corresponding test (Critical).
2. **Add the `aria-checked='false'` assertion** to the external-form Enter test (Important).
3. Optional: minor doc/code cleanup per suggestions 2, 4, 5.

The core design is sound — the only behavioral correctness gap is the missing `type="image"` case.

michaldudak:
> Missing input type="image" as a submit button

This was a deliberate decision: https://github.com/mui/base-ui/pull/4713#discussion_r3172532677

> Iteration could be slightly clearer

This would make TS slighly happier, but I think there's no real value in it (and we'd introduce more code to make sure the right `instanceof` is called). Checking the `tagName` should be enough.

flaviendelangle:
> This was a deliberate decision: https://github.com/mui/base-ui/pull/4713#discussion_r3172532677

I'd just add a comment on the method then to avoid having Claude Code or Codex propose to change that in the future :+1:

flaviendelangle (approved):

PR #4719, opened 2026-05-01
[number field] Allow Persian digits in keyboard input

### LINKED ISSUE #4718: [NumberField] Persian digit keyboard input is silently blocked
## Bug report

  ## Current behavior

  Typing Persian digits into `NumberField.Input` does not update the input value.

  This happens when using a Persian keyboard layout. The key presses are ignored before the value can be parsed.

  ## Expected behavior

  Persian digits should be accepted and parsed like other supported numeral systems, such as Arabic and fullwidth digits.

  ## Reproducible example

  ```tsx
  <NumberField.Root onValueChange={console.log}>
    <NumberField.Input />
  </NumberField.Root>
```
  Switch your keyboard layout to Persian and type ۱۲۳.

  Expected: the input accepts the digits and onValueChange receives 123.

  Actual: the input stays empty.

  ## Base UI version

  1.4.1

  ## Browser

  All browsers

  ## OS

  Any OS with a Persian keyboard layout

### PR DESCRIPTION
**Pull request by lunaxislu**
Fixes #4718

## Problem

Persian digits could not be typed into Number Field inputs because they were blocked during keyboard handling, even though the parser already supports them.

## Solution

Allow Persian numerals through the same keyboard input path as Arabic, Han, and fullwidth numerals.

## Test

Added a regression test that types Persian digits with `user.type()`.

Ran:

```bash
pnpm test:jsdom NumberField --no-watch
```

### DIFF
--- a/packages/react/src/number-field/input/NumberFieldInput.tsx
+++ b/packages/react/src/number-field/input/NumberFieldInput.tsx
@@ -335,6 +335,7 @@ export const NumberFieldInput = React.forwardRef(function NumberFieldInput(
       const isAsciiDigit = event.key >= '0' && event.key <= '9';
       const isArabicNumeral = ARABIC_DETECT_RE.test(event.key);
       const isHanNumeral = HAN_DETECT_RE.test(event.key);
+      const isPersianNumeral = PERSIAN_DETECT_RE.test(event.key);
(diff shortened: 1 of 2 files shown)


michaldudak (approved):
Looks good! Thanks for your contribution!

PR #4720, opened 2026-05-01
[docs] Add Ink UI and Fig UI to community libraries

anxndsgn:


colmtuite:
Fig looks great, and it's quite a novel idea. Would love to add that. But perhaps update the description to mention "for use in Figma plugins"

Unfortunately, I think Ink is not yet a good fit. It needs to be more complete, have more of a community around it, more adoption, or become more novel/unique.

anxndsgn:
> Fig looks great, and it's quite a novel idea. Would love to add that. But perhaps update the description to mention "for use in Figma plugins"
> 
> Unfortunately, I think Ink is not yet a good fit. It needs to be more complete, have more of a community around it, more adoption, or become more novel/unique.

hi @colmtuite, I just removed the Ink UI and update the Fig UI description. Thank you!

colmtuite (approved):

PR #4721, opened 2026-05-01
[focus] Fix enqueueFocus cleanup not cancelling RAF when a newer call has overwritten rafId

stefee:
The cleanup function returned by `enqueueFocus` only cancels the queued RAF if `rafId` still equals `currentRafId`:

```ts
return () => {
  if (rafId === currentRafId) {  // misses if a newer call has overwritten rafId
    cancelAnimationFrame(currentRafId);
    rafId = 0;
  }
};
```

When a second call to `enqueueFocus` happens after the first (e.g. a nested or concurrent popup interaction), `rafId` gets overwritten. The first call's cleanup no longer matches, so its RAF is never cancelled on unmount and fires as a ghost after the component is gone.

Moving the `cancelAnimationFrame` outside the guard fixes this — it's always safe to cancel a specific RAF id, even if it has already fired or been cancelled by another call.

Fixes a scenario where a `Combobox` inside a `Dialog` leaves a pending RAF after unmount that fires during a subsequent render and steals focus, preventing the popup from opening.

### DIFF
--- a/packages/react/src/floating-ui-react/utils/enqueueFocus.ts
+++ b/packages/react/src/floating-ui-react/utils/enqueueFocus.ts
@@ -22,8 +22,8 @@ export function enqueueFocus(el: FocusableElement | null, options: Options = {})
   const currentRafId = requestAnimationFrame(exec);
   rafId = currentRafId;
   return () => {
+    cancelAnimationFrame(currentRafId);
     if (rafId === currentRafId) {
-      cancelAnimationFrame(currentRafId);
       rafId = 0;
     }
   };


atomiks:
Do you have a regression test for the Combobox-in-Dialog repro? I can see a bug with `cancelPrevious: false`, but all callsites use the default `cancelPrevious: true`, so I’m not seeing how the stale rAF survives a later `enqueueFocus` call in any component 🤔

stefee:
@atomiks Hmm, yeah fair - I think maybe the scenario I hit is actually the component unmounting while a RAF from `enqueueFocus` is still pending (before a subsequent call with `cancelPrevious: true`).

So maybe the correct thing would be to cancel the pending RAF in `FloatingFocusManager`'s unmount handler?

stefee:
This is a weird thing and a bit tricky to repro outside of a real project. I actually encountered it when upgrading from a M1 Pro macbook to M5 Pro and running jsdom tests (with React Testing Library). There was a race condition in the tests that meant the tests passed 100% of the time on M1 Pro and failed 100% of the time on M5 Pro.

This was the fix that led to tests passing on both machines:



Screenshot 2026-05-05 at 10 17 30 am

stefee:
OK I was actually able to put together a minimal repro here: https://github.com/stefee/base-ui-4721-repro

Hope this helps!

atomiks:
Yeah the `enqueueFocus` part isn't the right fix for this because `enqueueFocus` isn't called again in the test's flow. The issue is that `enqueueFocus`'s rAF doesn't get canceled if focus redirects before it fires, but there's no simple way (e.g. effect cleanup) to cancel it in this case. Rather it needs to check if focus moved somewhere acceptable before the rAF completed so it can decide if it should be ignored. (Or a `focusin` handler needs to listen for this and call `cancelAnimationFrame` from there.)

In a real app, there's no practical bug because user interaction doesn't happen before that rAF fires (1 frame ~5-20ms), so it only shows up in a test or automated interactions

The root cause: 

- Dialog opens, `FloatingFocusManager` queues initial focus via queueMicrotask + rAF
- The test waits for Fruit to become enabled, then clicks the Combobox input
- The Combobox opens on click, which focuses the input
- The Dialog’s pending initial focus rAF fires later than the above interactions and focuses the Dialog’s default tabbable element, which in the repro is Cancel, not the Combobox input
- The Combobox input loses focus because Dialog's initialFocus "steals" it away, so the Combobox closes - `findByText("apple")` fails because the popup was opened and then immediately closed

stefee:
I think this makes sense @atomiks - what do you think of something like this: https://github.com/stefee/base-ui/commit/133ac7242467c6ebea267da423e5d1ba4a26d454 ?

atomiks:
Yes that looks like it'd work, if all tests pass

PR #4723, opened 2026-05-02
[menu] Fix submenu safePolygon scope for keepMounted portals

### LINKED ISSUE #4722: [menu] Submenu safePolygon fails with keepMounted portal
# Bug report

## Current behavior

When a Menu submenu uses `<Menu.Portal keepMounted>`, diagonal pointer movement from a submenu trigger into the submenu can close the submenu and highlight the next sibling item instead.

This does not happen when `keepMounted={false}`.

## Expected behavior

`safePolygon` should allow the same diagonal movement whether the submenu portal is unmounted while closed or mounted but hidden.

## Reproducible example

https://github.com/user-attachments/assets/5b5142a7-1ee6-4dc2-925b-a52c051baebd

https://stackblitz.com/edit/vitejs-vite-kt2dyjc1

## Base UI version

v1.4.x / current main

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

The issue appears to come from safePolygon pointer-events scoping. In the kept-mounted case, the current floating element can be selected as its own pointer-events scope, so sibling menu items remain targetable and emit `itemhover`, closing the submenu.

twillhorn:

Closes #4722

This fixes submenu hover traversal when a submenu uses `<Menu.Portal keepMounted>`. In that case, the kept-mounted submenu can already be present in the floating tree before it opens, so safePolygon pointer-events scoping can resolve to the submenu itself. That scope does not shield sibling items in the parent menu, allowing sibling hover handling to close the submenu during diagonal pointer movement.

The change keeps the existing scope fallback order, but ignores cached or parent scope candidates when they are the current floating element. This lets the parent menu remain the pointer-events scope while the submenu positioner stays interactive.

Validation:
- Confirmed the new regression test fails without the implementation change.
- `PLAYWRIGHT_BROWSERS_PATH=.playwright VITEST_ENV=chromium pnpm vitest run --project @base-ui/react packages/react/src/menu/root/MenuRoot.test.tsx -t "scopes submenu safePolygon"`

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
+++ b/packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
@@ -116,10 +116,18 @@ export function useHoverFloatingInteraction(
         parentFloating.style.pointerEvents = '';
       }
 
+      // A keep-mounted submenu can appear in the tree before it opens, so a
+      // cached scope or parent lookup may resolve to the submenu itself. That
+      // would not shield sibling items in the parent menu.
+      const cachedScopeElement =
+        instance.pointerEventsScopeElement !== floatingEl
+          ? instance.pointerEventsScopeElement
+          : null;
+      const parentScopeElement = parentFloating !== floatingEl ? parentFloating : null;
       const scopeElement =
         instance.handleCloseOptions?.getScope?.() ??
-        instance.pointerEventsScopeElement ??
-        parentFloating ??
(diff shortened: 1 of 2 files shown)


twillhorn:
Tagging @atomiks because this touches the submenu hover pointer-events/safePolygon scope logic from #4231/#4199.

atomiks (approved):
Thanks for the fix @twillhorn!

PR #4725, opened 2026-05-02
[autocomplete] Fix popup input FormData submission

### LINKED ISSUE #4724: [autocomplete] Popup input value is missing from native FormData
# Bug report

## Current behavior

When `Autocomplete.Input` is rendered inside `Autocomplete.Popup` and a surrounding `Field.Root` provides the field name, selecting an item updates the visible trigger value and Base UI `Form.onFormSubmit` receives the selected value. However, native `FormData` does not include the value.

This affects native submit paths such as:

- `new FormData(event.currentTarget)` in a React `onSubmit` handler
- `fetch(new FormData(form))`
- server action style submissions that read from `FormData`
- form libraries that serialize from DOM form controls

Observed console output from the reproduction below after selecting `alpha` and submitting:

```text
onFormSubmit value: alpha
native FormData value: null
```

This is an Autocomplete sibling case to #4687. #4687 fixed the popup-input **Field registration** path for Combobox; this issue is the related **native `FormData` routing** problem in Autocomplete (`selectionMode="none"`), where the submitted value is the input value rather than a selected value.

## Expected behavior

Native `FormData` should include the Autocomplete input value when the input is rendered inside the popup.

```text
onFormSubmit value: alpha
native FormData value: alpha
```

## Reproducible example

```tsx
import { Autocomplete } from '@base-ui/react/autocomplete';
import { Field } from '@base-ui/react/field';
import { Form } from '@base-ui/react/form';
import { Input } from '@base-ui/react/input';

  return (
    <Form
      onSubmit={(event) => {
        event.preventDefault();
        const data = new FormData(event.currentTarget);
        console.log('native FormData value:', data.get('search'));
      }}
      onFormSubmit={(values) => {
        console.log('onFormSubmit value:', values.search);
      }}
    >
      <Field.Root name="search">
        <Autocomplete.Root items={['alpha', 'alpine']}>
          <Autocomplete.Trigger>
            <Autocomplete.Value />
          </Autocomplete.Trigger>

          <Autocomplete.Portal>
            <Autocomplete.Positioner>
              <Autocomplete.Popup>
                <Autocomplete.Input render={<Input />} />
                <Autocomplete.List>
                  {(item: string) => (
                    <Autocomplete.Item key={item} value={item}>
                      {item}
                    </Autocomplete.Item>
                  )}
                </Autocomplete.List>
              </Autocomplete.Popup>
            </Autocomplete.Positioner>
          </Autocomplete.Portal>
        </Autocomplete.Root>
      </Field.Root>

      <button type="submit">Submit</button>
    </Form>
  );
}
```

Steps:

1. Open the popup from the trigger.
2. Select `alpha`.
3. Submit the form.
4. Compare `Form.onFormSubmit` with `new FormData(form).get('search')`.

## Workaround

Base UI `Form.onFormSubmit` receives the value, so consumers that can use that API can avoid the issue. Consumers relying on native `FormData` (server actions, `fetch(new FormData(form))`, third-party form libraries) currently cannot read the value from the DOM form data path.

## Base UI version

v 1.4.1

Reproduced in both jsdom and Chromium environments via the repository test runner.

## Which OS are you using?

macOS.

Not applicable.

## Additional context

Root cause:

- `Autocomplete.Input` reuses `ComboboxInput`; `Autocomplete.Root` delegates to `AriaCombobox` with `selectionMode="none"`, so the submitted value is `inputValue`.
- The popup input is transient and can unmount with the portal, so its `name` attribute cannot reliably back native `FormData`.
- The persistent hidden input next to the trigger exists, but is unnamed in the `selectionMode="none"` path, so native `FormData` has no named DOM control to read.

A fix should keep inline/non-popup Autocomplete inputs submitting through the visible input, while popup-positioned inputs route the `name` to the persistent hidden input.

lunaxislu:
Fixes #4724.

## Summary

- keep popup-positioned Autocomplete inputs from owning the native form `name`
- let the persistent hidden input submit `selectionMode="none"` values when the input is inside a `Positioner`
- preserve outside and inline Autocomplete native submission through the visible input
- add regression coverage for popup, inline, and default-value-before-open native `FormData` paths

## Root cause

`Autocomplete.Root` delegates to `AriaCombobox` with `selectionMode="none"`, so the submitted field value is the input value rather than a selected value. Base UI `Form.onFormSubmit` could read that value through the Field registry, but native `FormData` only reads named DOM controls.

When the input is rendered inside the popup, the visible input is transient and can unmount after selection. The persistent hidden input existed, but it was unnamed for `selectionMode="none"`, so native `FormData` had no stable named control to read from.

This complements #4687, which fixed the adjacent Combobox popup-input Field registration path.

## Validation

- `pnpm test:jsdom AutocompleteRoot --no-watch -t "native FormData"`
- `pnpm test:chromium AutocompleteRoot --no-watch -t "native FormData"`
- `pnpm test:jsdom ComboboxRoot --no-watch -t "submits when input inside popup renders a field-aware input"`
- `pnpm test:chromium ComboboxRoot --no-watch -t "submits when input inside popup renders a field-aware input"`
- `pnpm test:jsdom AutocompleteRoot --no-watch`
- `pnpm typescript`
- `pnpm prettier`
- `pnpm eslint`
- `git diff --check`


atomiks:
The names here get a bit confusing.

`inputInsidePositioner` fixes the immediate ambiguity, but it adds a second boolean next to `inputInsidePopup`, which sounds similar. (To be fair, its existing meaning is already broader than the name suggests (`hasPositionerParent || inline`) which should possibly be refactored later). The two similar-sounding names makes this hard to reason about because one flag is about focus/layout behavior and the new one is really about native form ownership.

Could we keep `inputInsidePopup` unchanged to reduce how much gets changed in this PR, but rename the new state to the concept it controls?

```ts
inputOwnsFormValue: selectionMode === 'none' && !hasPositionerParent,
```

Then the hidden input branch reads:

```ts
const hiddenInputOwnsNoneValue =
  !inputOwnsFormValue ||
  // Before the popup input mounts, trigger-control comboboxes still need
  // the persistent hidden input to own native form submission.
  (!inline && triggerElement != null && inputElement == null);
```

lunaxislu:
> The names here get a bit confusing.
> 
> `inputInsidePositioner` fixes the immediate ambiguity, but it adds a second boolean next to `inputInsidePopup`, which sounds similar. (To be fair, its existing meaning is already broader than the name suggests (`hasPositionerParent || inline`) which should possibly be refactored later). The two similar-sounding names makes this hard to reason about because one flag is about focus/layout behavior and the new one is really about native form ownership.
> 
> Could we keep `inputInsidePopup` unchanged to reduce how much gets changed in this PR, but rename the new state to the concept it controls?
> 
> ```ts
> inputOwnsFormValue: selectionMode === 'none' && !hasPositionerParent,
> ```
> 
> Then the hidden input branch reads:
> 
> ```ts
> const hiddenInputOwnsNoneValue =
>   !inputOwnsFormValue ||
>   // Before the popup input mounts, trigger-control comboboxes still need
>   // the persistent hidden input to own native form submission.
>   (!inline && triggerElement != null && inputElement == null);
> ```

That makes perfect sense !!. 
Separating the layout/focus flag from the native form ownership makes the intent much clearer. 

I'll revert `inputInsidePopup` to keep the PR scope small and rename the new state to `inputOwnsFormValue` as you suggested. I also reused `inputOwnsFormValue` in the input's name branch to keep the same concept in one place. 

Updating the PR shortly! Thanks for the review.

atomiks (approved):
Thanks for the fix @lunaxislu

PR #4732, opened 2026-05-04
[all components] Do not use `Math.random` in `useStableCallback`

### LINKED ISSUE #4729: @base-ui/utils/useStableCallback calls Math.random() at module scope
### Problem

`@base-ui/utils/useStableCallback` evaluates `Math.random()` at module scope:

```js
const useInsertionEffect = React[`useInsertionEffect${Math.random().toFixed(1)}`.slice(0, -3)];
```

In Cloudflare Workers, generating random values during module evaluation is disallowed. This crashes requests when Base UI is included in a server-evaluated bundle, for example React Router RSC client-reference or SSR chunks.

Cloudflare error:

```text
Error: Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope.
```

### Why this matters

RSC frameworks can evaluate `"use client"` dependencies in server/client-reference chunks to build client references. The component still runs on the client, but top-level module code can run in the Worker runtime.

### Reproduction shape

- React 19.2
- React Router 7 RSC / Vite RSC
- Cloudflare Workers
- Import any Base UI component path that depends on `@base-ui/utils/useStableCallback`
- Request a route whose server/client-reference chunk imports that component

### Suggested fix

Avoid top-level randomness. For React 18/19, direct access works:

```js
const useInsertionEffect = React.useInsertionEffect;
```

If the computed-property workaround is still needed for older environments, it should avoid randomness during module evaluation.

michaldudak:
Use `SafeReact` instead of nondeterministic code using `Math.random` to reference `useInsertionEffect` that is missing in older versions of React. This matches the usage for other features introduced after React 17.

Fixes #4729


LukasTy (approved):
LGTM. 👍 

## Claude Opus 4.7 review

Surgical fix that replaces a `Math.random()`-at-module-scope minification workaround in `useStableCallback.ts` with a read from `SafeReact`, the existing namespace clone already used in this codebase for the same purpose (`useId.ts`, plus a handful of components for `captureOwnerStack`). Resolves a real production crash on Cloudflare Workers, where generating random values during module evaluation is disallowed and breaks any RSC/SSR chunk that imports a Base UI component depending on `useStableCallback`. Behavior is preserved across React 17, React 18+, and Preact. CI is fully green.

## Context

- Original workaround used `(React as any)[\`useInsertionEffect${Math.random().toFixed(1)}\`.slice(0, -3)]` to defeat esbuild 0.19+ rewriting computed-string property access into direct named imports (mui/material-ui#41190). That static-string defeat works, but it pulls `Math.random()` into the module's top-level evaluation.
- Cloudflare Workers disallow random value generation in global scope, so the call throws `Error: Disallowed operation called within global scope` whenever Base UI is evaluated in a Worker (RSC client-reference chunks, SSR). See issue #4729 for the repro shape (React 19.2 + React Router 7 RSC + Vite RSC + Cloudflare).
- `SafeReact = { ...React } as typeof React` defeats the same bundler optimization differently: reads go through a runtime-cloned object, which no minifier will inline back into a named React import. This is the same approach already in `packages/utils/src/safeReact.ts` and consumed by `useId.ts:24` and several components.

## Critical issues

None.

## Suggestions

| # | File | Line | Suggestion | Category |
|---|------|------|------------|----------|
| 1 | packages/utils/src/useStableCallback.ts | 5 | Optional: explicitly annotate the type to make the React 17 (undefined) case visible, mirroring `useId.ts:24` (`const maybeReactUseId: undefined \| (() => string) = SafeReact.useId;`). The trailing truthy check covers the runtime side either way, so this is purely a readability nit. | Style |

## What looks good

- Reuses existing infrastructure rather than introducing a new escape hatch. `SafeReact` is the established way Base UI reads APIs that may be missing on older React versions: `useId.ts:24` (React 18 hook), and ~7 components reading `captureOwnerStack` (React 19 hook). Adding `useInsertionEffect` to the list is the same shape of change.
- The expanded JSDoc on `SafeReact` is a real readability win. The previous file-level link to a year-old `material-ui` comment forced anyone curious to chase context off-repo. The new comment names the failure mode (bundlers rewriting direct property reads into named imports, which breaks React 17) right at the source.
- Behavior preservation across all three target runtimes:
  - React 17: `SafeReact.useInsertionEffect` is `undefined`, truthy check fails, falls back to the no-op trampoline `(fn) => fn()`. Same as before.
  - Preact: spreading copies both `useInsertionEffect` and `useLayoutEffect` to the same function, so the `useInsertionEffect !== SafeReact.useLayoutEffect` identity check is false; falls back to the no-op trampoline. Same as before.
  - React 18+: `SafeReact.useInsertionEffect` is the real hook, distinct from `useLayoutEffect`; uses the real hook. Same as before.
- Drops the now-unused `import * as React from 'react'` from `useStableCallback.ts` ‚Äî clean removal.
- Slight bundle reduction (parsed -52B / gzip -39B per the code-infra-dashboard comment). The fix is essentially free.
- All CI green: `test-dev` (macos/windows/ubuntu), JSDOM tests + React 18 variant, browser tests + React 18 + React Compiler variants, typechecking + React 18 variant, E2E, regression, linting, package verification, validating docs, argos visual tests.

## Things worth verifying / non-blocking notes

- The fix's robustness depends on bundlers being unable to track that `SafeReact` is structurally identical to `React` (which would let them re-inline `SafeReact.useInsertionEffect` into a named import). That tracking is well outside what esbuild and similar minifiers attempt, and the same pattern has lived in `useId.ts` without regression. No real risk, just naming the assumption.
- There is no `useStableCallback.test.ts` (the neighboring `useId.ts` has `useId.test.tsx`). The change is structural and the React 17 / Preact branches are exercised indirectly by the wider component test suites (which CI runs against React 18 too), so the absence is not a blocker for this PR. Worth flagging if maintainers want a regression test for the fallback paths.
- The Preact identity comparison (`useInsertionEffect !== SafeReact.useLayoutEffect`) reads `useLayoutEffect` from the same spread snapshot, so the comparison stays semantically equivalent to the previous `React.useLayoutEffect` read. Only matters if something mutates `React.useLayoutEffect` after `safeReact.ts` evaluates, which is not a realistic scenario.

## Verdict

Approve. Tight, well-scoped fix that resolves a reproducible production bug, aligns with an existing pattern in the codebase, preserves behavior across all supported React/Preact runtimes, and improves readability of the workaround it leans on.

PR #4734, opened 2026-05-04
[docs] Fix HTML validation issues from link-check

Janpot:
Enables html-validate's `mui:recommended` ruleset in the docs link-check and fixes all issues

Component/demo fixes:
- `Table`: `TableCellInner` span → div (block content no longer under `<span>`).
- `ReferenceTable`: render fallback type signatures inline so the type cell no longer nests a block `CodeBlock` div in the trigger `<button>`; h4 → h3 for contiguous heading levels; unique caption id per table instance.
- `rehypeSlug`: dedupe heading ids (`no-dup-id`). Fixes `mergeProps` link in TOC on https://base-ui.com/react/utils/merge-props
- `CodeBlock`: default `codeId`/`titleId` to `undefined` (no `id=""` outside a `CodeBlock.Root`).
- `ReleaseTimeline`: card h3 → h2; spine as a `::before` pseudo-element (no `<div>` under `<ul>`).
- `DemoPlayground`: add `role="figure"` to the labelled demo container.
- Avatar page: render `<TypesAvatarAdditional>` so the `ImageLoadingStatus` TypeRef resolves (its `#imageloadingstatus` anchor was never emitted → broken link).
- Autocomplete grid demo: `aria-label` on the composer and search inputs.


docs/src/app/(docs)/react/components/avatar/page.mdx
```diff
@@ -39,6 +39,10 @@ import { TypesAvatar } from './types';
 
 <TypesAvatar.Fallback />
 
+## Additional Types
+
+<TypesAvatarAdditional showAdditionalTypes={['imageloadingstatus']} />
+
```
aarongarciah: I'm a bit lost about this. Why do we need it in Avatar but not in e.g. Drawer (see screenshot below). The `TransitionStatus` type doesn't link anywhere in `Drawer.Popup` but `ImageLoadingStatus` does in `Avatar.Root`.

We intentionally don't display these types in component pages, so if we could find another solution, that'd be great.

Image

aarongarciah (commented):
Code blocks are broken inside prop tables.

Image

docs/src/app/(docs)/react/components/avatar/page.mdx
aarongarciah: I was wrong, Toast also has the `Additional types` section. It's just that it hides it via `[//]: # '@exclude-table-of-contents'`. We should do the same here if we end up displaying this section in Avatar.

Janpot: > Why do we need it in Avatar but not in e.g. Drawer 

Looks like `ImageLoadingStatus` is publicly [exported](https://github.com/mui/base-ui/blob/5b1bfea02ea9ae8ec8aa5007da4df43ac3ee1b3d/packages/react/src/avatar/root/AvatarRoot.tsx#L44), while `TransitionStatus` is coming from [`./internals/`](https://github.com/mui/base-ui/blob/5b1bfea02ea9ae8ec8aa5007da4df43ac3ee1b3d/packages/react/src/internals/useTransitionStatus.ts#L6)

patches/@mui__internal-docs-infra@0.11.1-canary.22.patch
Janpot, line 1: temporary patch until adopted in https://github.com/mui/mui-public/pull/1639

docs/src/app/(docs)/react/components/avatar/page.mdx
aarongarciah: Looks like this section is not needed anymore since now ImageLoadingStatus displays its own popup?

Image

docs/src/components/ReleaseTimeline/ReleaseTimeline.css
```diff
@@ -20,16 +20,15 @@
...
   .TimelineItem {
     position: relative;
   }
 
   @media (min-width: 48rem) {
-    .TimelineSpine {
+    /* Decorative vertical spine. A pseudo-element rather than a child <div> so
+       the <ul> only contains <li> elements. */
+    .ReleaseTimeline::before {
+      content: '';
```
aarongarciah, line 34: The first release card is now aligned to the right

|master|pr|
|---|---|
|Image|Image|

aarongarciah (commented):
Sharing only the interesting findings from Fable 5:

## Simplifications

### 🟡 Methods/Properties accordions keep the deterministic caption-id pattern the PR removes from ReferenceAccordion

**Location:** `docs/src/components/ReferenceTable/MethodsReferenceAccordion.tsx:21`

```tsx
const captionId = `${partName}-methods-caption`;
```

The PR switches `ReferenceAccordion` to `React.useId()` because `${partName}-caption` duplicated when the same part name rendered twice on a page. `MethodsReferenceAccordion` and `PropertiesReferenceAccordion` (`${partName}-properties-caption`, [[MethodsReferenceAccordion.tsx:21](https://claude.ai/epitaxy/docs/src/components/ReferenceTable/MethodsReferenceAccordion.tsx:21)](docs/src/components/ReferenceTable/MethodsReferenceAccordion.tsx:21), [[PropertiesReferenceAccordion.tsx:21](https://claude.ai/epitaxy/docs/src/components/ReferenceTable/PropertiesReferenceAccordion.tsx:21)](docs/src/components/ReferenceTable/PropertiesReferenceAccordion.tsx:21)) have the identical latent bug — and `ReferenceTable` renders `MethodsReferenceAccordion` twice with the same `name` (static methods + instance methods, [[ReferenceTable.tsx](https://claude.ai/epitaxy/docs/src/components/ReferenceTable/ReferenceTable.tsx)](docs/src/components/ReferenceTable/ReferenceTable.tsx)). No currently documented class has both, so `no-dup-id` passes today.

**Failure scenario:** The first documented class with both static and instance methods emits two `X-methods-caption` ids and fails the newly enabled `no-dup-id` check in CI, in a PR that has nothing to do with the docs infra.

**Fix:** Switch both components to `React.useId()` in this PR, matching `ReferenceAccordion`.

## Docs

### ℹ️ TypeInline drops all props except children, unlike its TypePre counterpart

**Location:** `docs/src/components/CodeBlock/CodeBlock.tsx:149`

```tsx
export function TypeInline({ children }: { children?: React.ReactNode }) {
  return <span className="Code TableCode">{children}</span>;
}
```

`TypePre` (`PreInline`) forwards full `React.ComponentProps<'pre'>`, but `TypeInline` accepts only `children`. If the docs-infra factory ever passes `className` or data attributes to `ShortTypePre`, they're silently discarded; the hardcoded `"Code TableCode"` string also duplicates the class names owned by the `Code`/`TableCode` components. Couldn't verify the factory contract locally (new canary not installed), so this is a heads-up only.

## Verdict

**Approve after nits** — no correctness issues; the nits are a latent duplicate-id trap in the sibling accordions, a missing test for the new slug-dedupe logic, and a heading-case inconsistency.

docs/src/app/(docs)/react/components/avatar/page.mdx
Janpot: Removing it breaks the popover. Not 100% sure why.

docs/src/components/ReleaseTimeline/ReleaseTimeline.css
Janpot: ah yes, saw it but at the time my local was one version behind, I assumed it alternated per version. Should be fixed now.

docs/src/app/(docs)/react/components/avatar/page.mdx
Janpot: Ok, so apparently these popups are populated from a page-wide registry, which in turn gets populated by the reference tables. If we remove AdditionalTypes there is nothing being registered for that type and the popup falls back to a (broken) link. Conceptually it also seems like the section is meant to serve as the target for the anchor link that the popup falls back to for SSR (a bit useless if hidden, but ok)

aarongarciah: 👍

aarongarciah (commented):
Fable 5 only found one remaining issue (apart from the CI failing). I tested manually and couldn't any issues.

---

### 2. 🟡 `rehypeSlug` dedup ignores pre-existing heading ids, so collisions with them remain possible

**Location:** `docs/src/components/QuickNav/rehypeSlug.mjs:46`

```js
visit(tree, 'element', (node) => {
  if (headingRank(node) && !node.properties.id) {
    ...
    occurrences.set(id, 0);
    node.properties.id = id;
  }
```

Headings that already carry an `id` (set by an author or an earlier plugin) are skipped entirely and never seeded into the `occurrences` map. A later heading whose text slugs to that same value still gets the bare slug, reintroducing exactly the duplicate-id class of bug this PR fixes — and it would surface as a confusing `no-dup-id` CI failure now that the rule is enforced. The new test suite covers the "already has an id" passthrough but not this collision case.

**Failure scenario:** A page contains `<h2 id="props">` (explicit) followed by `## Props`; both end up with `id="props"`, the TOC link targets the wrong heading, and the link-check job fails on `no-dup-id`.

**Fix:** Seed `occurrences` with existing heading ids as they are visited (e.g. `occurrences.set(node.properties.id, 0)` in an `else` branch), and add a test for the case.

aarongarciah (approved):

PR #4740, opened 2026-05-04
fix(FloatingPortal): guard addEventListener against null portalNode under concurrent React

ossaidqadri:
## Bug

Under React 18+ concurrent rendering (Next.js App Router with `useTransition`/client-side navigation), `FloatingPortal` throws:

```
Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at Provider (<anonymous>)
```

## Environment

- `@base-ui/react`: 1.4.1
- React: 19
- Next.js: 16 (App Router)
- Reproduction: client-side navigation to a page that contains `Dialog` (used as Sheet) and `Select` components inside a `Suspense` boundary

## Root cause

In `packages/react/src/floating-ui-react/components/FloatingPortal.tsx`, the `useEffect` that attaches `focusin`/`focusout` listeners has an early-return guard at the top:

```ts
React.useEffect(() => {
  if (!portalNode || modal) {
    return undefined;
  }
  // ...
  return mergeCleanups(
    addEventListener(portalNode, 'focusin', onFocus, true),
    addEventListener(portalNode, 'focusout', onFocus, true),
  );
}, [portalNode, modal]);
```

Under concurrent React, the effect can fire with a `portalNode` that passes the `!portalNode` guard at the start but has since become `null` — or `portalNode` was briefly truthy during the transition phase and the effect was scheduled before it reset to `null`.

The `@base-ui/utils/addEventListener` utility has **no null guard**:

```ts
function addEventListener(target, type, listener, options) {
  target.addEventListener(type, listener, options);  // throws if target is null
  return () => { target.removeEventListener(type, listener, options); };
}
```

Compare this with the guarded pattern used elsewhere in `FloatingFocusManager.tsx`:

```ts
// safe — short-circuit prevents null call
floating && addEventListener(floating, 'focusin', handleFocusIn)
```

But the `FloatingPortal` call site does NOT use this pattern:

```ts
// unguarded — portalNode can be null under concurrent rendering
return mergeCleanups(
  addEventListener(portalNode, 'focusin', onFocus, true),
  addEventListener(portalNode, 'focusout', onFocus, true),
);
```

## Suggested fix

Align the call site with the guarded pattern already used in `FloatingFocusManager`:

```ts
return mergeCleanups(
  portalNode && addEventListener(portalNode, 'focusin', onFocus, true),
  portalNode && addEventListener(portalNode, 'focusout', onFocus, true),
);
```

Or add a null guard to `@base-ui/utils/addEventListener`:

```ts
function addEventListener(target, type, listener, options) {
  if (!target) return () => {};
  target.addEventListener(type, listener, options);
  return () => { target.removeEventListener(type, listener, options); };
}
```

## Impact

Non-fatal console error on every client-side navigation to a page with `Dialog`/`Select` components. Does not break functionality but pollutes the console and will surface as an error in monitoring tools (Sentry, etc.).

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingPortal.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingPortal.tsx
@@ -216,9 +216,11 @@ export const FloatingPortal = React.forwardRef(function FloatingPortal(
 
     // Listen to the event on the capture phase so they run before the focus
     // trap elements onFocus prop is called.
+    // Guard with `portalNode &&` to defend against concurrent React renders
+    // where the effect fires before the ref is attached (issue #4739).
     return mergeCleanups(
-      addEventListener(portalNode, 'focusin', onFocus, true),
-      addEventListener(portalNode, 'focusout', onFocus, true),
+      portalNode && addEventListener(portalNode, 'focusin', onFocus, true),
+      portalNode && addEventListener(portalNode, 'focusout', onFocus, true),
     );
   }, [portalNode, modal]);
 


atomiks:
The root cause and fix here seems impossible to me? 

The closure means the variable can't randomly be `null` since it already early-returns before ever reaching `mergeCleanups`, which is why TS is fine with it

Codex analysis agrees:

> Short answer: with the code in `v1.4.1`, that exact explanation is basically not possible.
>
> The effect in `packages/react/src/floating-ui-react/components/FloatingPortal.tsx:196` closes over one render’s `portalNode` value. If it passes:
>
> ```ts
> if (!portalNode || modal) {
>   return undefined;
> }
> ```
>
> then the later calls at `packages/react/src/floating-ui-react/components/FloatingPortal.tsx:220` are using that same captured value. Concurrent React can delay effects, flush old passive effects, or schedule a later render where `portalNode` is `null`, but it does not mutate the old closure’s `portalNode` binding from an element into `null`. Even if the DOM node was removed, the old `HTMLElement` object still has `addEventListener`.
>
> So PR [#4740](https://github.com/mui/base-ui/pull/4740)’s change:
>
> ```ts
> portalNode && addEventListener(portalNode, ...)
> ```
>
> is runtime-equivalent after the existing early return. It would not fix the claimed path.
>
> Most likely explanations: the stack/source map is pointing at the wrong callsite, the app is not actually running the code/version described, or the real nullable target is coming from some other `addEventListener` call. I’d ask for a minimal repro or a full unminified stack with source maps before accepting this fix.

PR #4742, opened 2026-05-05
[toast] Fix stuck drag state on release

aeterno-caspian:
This fixes a toast drag edge case where release cleanup can be missed after dragging a toast far in the direction other than that of the expected dismissal directions.

In practice, the toast can feel like it is still being controlled by the mouse after the press has already been released. In a stack, that same stale drag state can leave the card suspended out of place with its content hidden behind the frontmost toast.

The change keeps the active drag state in refs, listens for release/cancel events on the owner document for the lifetime of the gesture, and safely handles pointer capture APIs so release cleanup still runs.

Before:

https://github.com/user-attachments/assets/52b19912-c666-4bc5-93bf-5582a3368d32

After:

https://github.com/user-attachments/assets/48f46c30-6a48-47fd-bd6d-2c58d5e573e3

I tested this with:

- `pnpm test:jsdom toast --no-watch`
- `pnpm typescript --pretty false`


aeterno-caspian:
I don’t have permission to add labels on this repo, and the required-label check is failing because the PR currently has none.

atomiks:
I can't seem to repro the issue myself, do you have steps beyond the recording? It seems like the problem is that the viewport collapses because it thinks it lost focus or such? I can't tell if that's just from you releasing the pointer though.

---

**The main reason this PR's solution gives me pause is it adds a lot of code for what I'd think should be a simpler fix. Here's Codex's suggestion:**

The most minimal robust fix is: keep the existing React drag logic, but add a **backup document-level end listener** for the active pointer. Don’t refactor the whole gesture path.

I’d aim for only these additions:

1. `activePointerIdRef`
   This replaces the need for `isSwipingRef`; `null` means no active drag.

2. `dragOffsetRef`
   Needed because a native `document` listener installed during `pointerdown` would otherwise read stale React state.

3. `cleanupDocumentPointerListenersRef`
   Needed to remove the temporary document listeners.

Then:

```ts
onPointerDown:
  activePointerIdRef.current = event.pointerId
  add document pointerup/pointercancel listeners
  try setPointerCapture(pointerId)

onPointerMove:
  keep existing logic
  also write latest drag offset to dragOffsetRef

onPointerUp / document pointerup:
  if pointerId !== activePointerIdRef.current return
  activePointerIdRef.current = null
  remove document listeners
  clear swiping using dragOffsetRef + initialTransformRef

onPointerCancel / document pointercancel:
  same pointerId guard
  reset drag state
```

What I’d cut from the current PR:

- `isSwipingRef`: redundant if `activePointerIdRef` is the source of truth for native cleanup.
- `setSwiping()` wrapper: then unnecessary.
- thin `handlePointerUp` / `handlePointerCancel` wrappers.
- `cleanupDocumentPointerListeners` as a `useStableCallback`; just call `cleanupDocumentPointerListenersRef.current?.()` inline.
- possibly `releasePointerCapture()` entirely, since capture is implicitly released after `pointerup`/`pointercancel`; or guard it with `hasPointerCapture()` instead of a whole helper.

The only helper I might keep is a tiny guarded `setPointerCapture` call, if they’ve actually seen it throw. But the core fix should be much smaller than the PR: backup release listener + pointer id ref + latest drag offset ref.

aeterno-caspian:
Repro steps in the docs hero (toast page):

1. Click "Create toast" several times so a stack exists (so a non-frontmost toast is present).
2. Hover the stack to expand it.
3. Press and hold on a non-frontmost toast.
4. Drag far to the left, past the toast viewport.
5. Release.

Without the fix, the dragged toast does not snap back into the stack and its content stays at opacity 0, because the toast remains in `data-swiping` state with a non-zero `--toast-swipe-movement-x`.

The new regression test in `ToastRoot.test.tsx` reproduces this in jsdom by releasing `pointerup` on `document` instead of the toast root. It fails on `master` and passes after this PR; the failing assertion is `--toast-swipe-movement-x` not resetting to `0px` after release.

aeterno-caspian:
I also verified the regression test against the pre-PR source.

Setup: temporary worktree at `origin/master`, with only the updated `ToastRoot.test.tsx` regression test applied.

Command:

```bash
pnpm test:jsdom ToastRoot --no-watch
```

Result: the new test fails as expected on `master`:

```text
FAIL src/toast/root/ToastRoot.test.tsx > <Toast.Root /> > drag behavior regression > resets drag state after releasing a far swipe even when the release lands on the document

expect(element).not.toHaveAttribute("data-swiping")

Expected the element not to have attribute:
  data-swiping
Received:
  data-swiping=""
```

So the reduced test captures the stale drag state directly, independently of the visual stack styling in the recording.

atomiks:
With those exact steps, including a resized browser window like in the recording (so I drag out of the window), it doesn't reproduce for me - the `data-swiping` state is removed upon release every time.

That said, I pushed another simplified version in a new commit, can you let me know if it still works for you?

aeterno-caspian:
> With those exact steps, including a resized browser window like in the recording (so I drag out of the window), it doesn't reproduce for me - the `data-swiping` state is removed upon release every time.
> 
> That said, I pushed another simplified version in a new commit, can you let me know if it still works for you?

To calrify, I did not drag outside the window (I was also in full screen) in the video, the desktop background is purely an video editor effect.

And yes, with the latest commit I cannot reproduce the bug and the test passes.

atomiks:
@aeterno-caspian Ah, got it — it still doesn't reproduce though regardless, for me

I further trimmed the diff to get bundle size down from 🔺+147B(+0.10%) to 🔺+119B(+0.08%), but it should work the same. Can you let me know if it still passes? If so, I'll approve and merge

aeterno-caspian:
> @aeterno-caspian Ah, got it — it still doesn't reproduce though regardless, for me
> 
> I further trimmed the diff to get bundle size down from 🔺+147B(+0.10%) to 🔺+119B(+0.08%), but it should work the same. Can you let me know if it still passes? If so, I'll approve and merge

Yes, the latest commit@324b07d still fixes it for me.

I can no longer reproduce the stale drag state with the original interaction. The tests also pass locally:

`pnpm test:jsdom ToastRoot --no-watch`
`pnpm test:jsdom toast --no-watch`

Thanks for tightening this up.

atomiks (approved):

PR #4744, opened 2026-05-05
[accordion][slider] Refactor key handling to use COMPOSITE_KEYS for consistency

sai6855:
This PR refactors how supported keyboard keys are managed and referenced across the accordion and slider components.


michaldudak:
It seems that `ALL_KEYS` is used just in useCompositeRoot, so I'd refactor it to use `COMPOSITE_KEYS` instead and just remove `ALL_KEYS` completely.

michaldudak (approved):

PR #4751, opened 2026-05-06
[toast] Remove memoized selector

atomiks:
Removes Toast's memoized selector use while keeping the ID-based toast metadata lookup fast.

## Changes

- Store toast metadata directly when the toast list changes.
- Keep toast metadata synchronized through close transitions.
- Leave `createSelectorMemoized` and `reselect` in utils for other consumers.

## Performance

Local Node microbenchmark comparing upstream/master to this PR with 50 toasts:

| Operation | upstream/master | PR |
| --- | ---: | ---: |
| Metadata map rebuild after toast list change | 1.864 us/op | 1.100 us/op |
| Cached selector read by id | 0.017 us/op | 0.007 us/op |

The rebuild timing moved: upstream rebuilds lazily on the first selector read after a toast list change, while this PR rebuilds when the toast list is written. Toast lists are normally small in practice, so this is not a meaningful cost.


flaviendelangle (commented):
We are using `createSelectorMemoized` heavily in advanced components (MUI X and Base), what is the proposed alternative here?

atomiks:
> We are using createSelectorMemoized heavily in advanced components

If it's still needed externally, we can leave it in the package. But it wasn't necessary for Toast or any other component and adds needless bundle size, so it feels like a waste to have the `reselect` dep in our package.json.

flaviendelangle:
I think we should leave it in the package, because it's the only centralized place we have to easily import it from everything without duplicating, even if it means keeping `reselect` as a dep.

For Toast specifically, if we can have the same performances without the bundle size cost, great :+1:

romgrk:
@atomiks Please keep in mind that `@base-ui/utils`, despite the name, is not utils just for BUI — it's utils that are used organization-wide in all our repositories. It's placed here because it needs to be in a repository, not because it's owned by BUI. For any planned changes to the Store (or to the utils), you can check in with me if you want to know if/how something is used elsewhere, I'm the code owner for that package because I have a good overview of how it's used by the rest of the org.

atomiks:
@romgrk I've kept the deps and memoized module after @flaviendelangle's review. I wasn't sure if it was being used elsewhere previously, but now I know

> I'm the code owner for that package because I have a good overview of how it's used by the rest of the org.

Looks like you were marked as a code reviewer here automatically. Since there are no changes to the utils package anymore, there shouldn't be any issue there

romgrk:
:+1: 

And if there are BUI-specific utils, they should actually go in `@base-ui/react`, not in `@base-ui/utils` which should be reserved for general-purpose utils.

atomiks:

Approve ✅ because the full Toast store diff keeps metadata synchronized across every affected mutation path, preserves behavior, and the test/typecheck evidence plus local benchmarking support the eager metadata tradeoff.

### 1. Bugs / Issues (None)

Checked constructor initialization, `addToast`, duplicate-id upsert, replacing an ending toast, `updateToastInternal`, single-toast close, close-all, `removeToast`, promise updates, limit recalculation, focus-management lookups, and the `Toast.Root` / `Toast.Positioner` selector consumers.

The new `toastMetadata` map carries the same `value`, `domIndex`, `visibleIndex`, and `offsetY` data as the previous memoized selector. The change is timing: metadata is rebuilt when the toast list is written instead of lazily on first selector read.

### Performance Assessment

I reran a local Node microbenchmark against freshly fetched `upstream/master` (`7a6cf63b1`) and the PR head (`2bdaf4d3e`). Values below are median microseconds per operation from 9 rounds, with the store-construction baseline subtracted for non-construction scenarios. Because this refactor intentionally moves work from selector reads to toast-list writes, I measured both write-only and write-plus-first-read paths.

| Scenario | 5 toasts | 100 toasts | Takeaway |
|---|---:|---:|---|
| `createStore` with preloaded toasts | master 1.49 µs / PR 1.61 µs | master 1.52 µs / PR 4.61 µs | Slower when constructed with existing toasts; mostly synthetic for public `Toast.Provider`, which initializes with `toasts: []`. |
| `addToast` without an id, write only | master 0.33 µs / PR 0.64 µs | master 0.96 µs / PR 4.38 µs | Slower on the PR because metadata is rebuilt eagerly before render reads. |
| `addToast` without an id + first selector reads | master 1.93 µs / PR 0.64 µs | master 6.59 µs / PR 4.42 µs | Faster once the render/positioning read path is included. |
| `addToast` with an id + first selector reads | master 2.18 µs / PR 0.43 µs | master 11.08 µs / PR 4.17 µs | Faster for the ID lookup/upsert path. |
| `updateToastInternal` on a cold selector state | master 1.45 µs / PR 0.32 µs | master 5.71 µs / PR 3.76 µs | Faster in the cold-selector path. |
| `closeToast` on a cold selector state | master 1.48 µs / PR 0.31 µs | master 6.67 µs / PR 4.70 µs | Faster in the cold-selector path. |
| first 4 selector reads | master 1.33 µs / PR 0.04 µs | master 3.62 µs / PR 0.04 µs | Much faster because selectors now read through direct `Map.get()` calls. |

The code confirms the underlying tradeoff: toast-list writes now rebuild metadata eagerly, while render/positioning selectors read from a direct `Map.get()` path. The one real write-only regression above is expected from moving the work earlier; the common write-plus-read paths still look favorable for Toast usage.

### Simplification and DRY Assessment

This is a clean simplification. `createToastMetadata` is a small pure function, and toast selectors now use direct `Map.get()` lookups without the `createSelectorMemoized` / reselect layer. `createSelectorMemoized` and `reselect` remain available in utils for other consumers, so this does not broaden the change beyond Toast.

The new `store.test.ts` follows the local Vitest style and directly exercises the metadata invariant. I did not find a repo-convention issue in the touched files.

### Test Coverage Assessment

Validated with `pnpm typescript`, `pnpm test:jsdom toast --no-watch`, `pnpm test:chromium toast --no-watch`, targeted eslint, and `git diff --check`. The added store test covers metadata sync through update, close/ending transition, remove, and add; existing Toast tests cover duplicate-id, limit, close-all, promise, viewport, root, and positioner paths.

PR #4756, opened 2026-05-06
[menu] Reduce closed root work

atomiks:
Menu roots were still doing some trigger interaction setup and popup prop syncing work while closed. This keeps that work on the trigger or delays it until the menu is open or mounted.

## Changes

- Move open-method tracking from `Menu.Root` to menu triggers and submenu triggers.
- Reset the stored open method through the shared popup root sync helper.
- Skip menu popup and item prop syncing while the popup is closed and unmounted.


atomiks:
https://github.com/mui/base-ui/pull/4757 shows a strong enough result not to work further here

PR #4757, opened 2026-05-06
[floating-ui] Remove interaction wrappers

atomiks:
This removes the internal interaction wrapper hooks and keeps interaction prop merging at the call sites that still need it. Runtime code now uses direct hook props, while tests use a local helper for the old getter-style ergonomics.

## Changes

- Remove `useInteractions` from the Floating UI runtime exports and move the helper behavior into `useTestInteractions` for tests.
- Remove `useRole` and spell out the small role/ARIA props directly in the Menu and test fixture call sites that used it.
- Convert Select, Combobox, Menu, Navigation Menu, and Floating UI tests to merge direct interaction props.
- Keep the wrapper-style prop getter only in test utilities.


atomiks:
Approve ✅: the two previously noted nits are addressed, and I do not see remaining actionable issues in the refreshed full-branch diff.

No remaining branch-relevant bugs or review nits found.

The stale `ElementProps.item` callback contract was removed from the production Floating UI type surface, with the legacy callback shape kept inside the test-only helper where it is still needed.

The moved interaction-merging helper now has direct regression coverage for handler composition, undefined handlers, non-event `on*` props, return values, floating focusability props, item callback behavior, and memoized getters.

The direct `mergeProps` call sites still preserve the old interaction handler order, and the inline ARIA replacements line up with the removed `useRole` behavior for Select, Combobox, Menu, Navigation Menu, and the Floating UI test fixtures.

### Test Coverage Assessment

I verified the follow-up locally with `pnpm test:jsdom useTestInteractions --no-watch`, `pnpm typescript`, and `pnpm eslint`. GitHub CI has started on the pushed commit and is still pending at the time of this comment.

PR #4758, opened 2026-05-07
[code-infra] Enable mui/no-presentation-role lint rule

Janpot:
Propagated from https://github.com/mui/material-ui/pull/48486

Enables the `mui/no-presentation-role` ESLint rule, which auto-fixes `role="presentation"` to its preferred alias `role="none"`.

Scope goes beyond the propagation. The rule only visits `JSXAttribute`, but Base UI sets the role through a props object almost everywhere, so the auto-fix alone would have left the library rendering `role="none"` on the internal backdrop next to a `Dialog.Backdrop` still rendering `role="presentation"`. The remaining 23 object-form sites are migrated here too, so the rendered DOM settles on one spelling, and a `no-restricted-syntax` selector guards the form the rule cannot see.

`role="none"` and `role="presentation"` are exact ARIA synonyms, so this is behavior-neutral for assistive tech. It is observable to anyone selecting on `[role="presentation"]`.

The `jsx-a11y/role-supports-aria-props` override from the upstream PR is dropped: material-ui needed it for a case under `docs/data`, but no `role="none"` element here carries an `aria-*` prop and lint is green without it.


Janpot:
@claude review

Janpot:
@claude review low

PR #4765, opened 2026-05-07
[utils] Import `Store` changes from `x-internals`

romgrk:
This PR imports the `Store` changes from `x-internals` in preparation for deleting the duplicated code from the MUI X repository.


flaviendelangle (approved):

packages/utils/src/store/useStoreEffect.ts
```diff
@@ -0,0 +1,58 @@
...
+    onMount: () => {
+      instance.subscribe();
+      return () => {
+        instance.dispose?.();
+        instance.dispose = null;
+      };
+    },
+  };
+
+  instance.subscribe();
```
michaldudak: This creates a store subscription before the component commits. In concurrent rendering, an abandoned render will not run the useOnMount cleanup, leaving a listener for an instance that never mounted.

packages/utils/src/store/useStoreEffect.ts
```diff
@@ -0,0 +1,58 @@
...
+ * compute and store derived state, use `createSelectorMemoized` instead.
+ */
+export function useStoreEffect<State, Value>(
+  store: ReadonlyStore<State>,
+  selector: (state: State) => Value,
+  effect: (previous: Value, next: Value) => void,
+): void {
+  const instance = useRefWithInit(initialize, { store, selector }).current;
+  instance.effect = effect;
+  useOnMount(instance.onMount);
```
michaldudak: The hook updates only `effect`; `initialize` closes over the initial `store` and `selector`, and `useOnMount` never reruns. If a component receives a replacement `store`, or passes a selector that closes over props, updates continue to be computed from stale inputs while invoking the latest effect callback.

packages/utils/src/store/createSelector.ts
michaldudak, line 1: JSDoc is stale after the arity change

packages/utils/src/store/ReactStore.test.tsx
michaldudak, line 1: New public store behavior is effectively untested.
The only changed test line widens an existing nested-store type. The PR adds/exports `useStoreEffect`, `Store.create`, `createSelectorMemoizedWithOptions`, single-combiner handling in `createSelectorMemoized`, and expanded `createSelector` arity, but none of those behaviors have focused regression tests.

packages/utils/src/store/useStoreEffect.ts
romgrk: Good catch. The `useStoreEffect` early subscription has always felt too hacky for my taste but it fixed initialization issues in some parts of the MUI X codebase. I'm going to remove the hook altogether from this PR and see if we can rework things there to get rid of it.

flaviendelangle:
I think we could move forward with this one

PR #4806, opened 2026-05-11
[otp field] Ignore hidden-input changes while readonly or disabled

### LINKED ISSUE #4805: [otp field] Hidden validation input still accepts changes while readOnly or disabled
# Bug report

## Current behavior

`OTPField.Input` ignores changes while the field is `readOnly` or `disabled`, but the hidden validation input rendered by `OTPField.Root` still accepts `change` events.

As a result, updates coming through the hidden input path can still mutate the value and fire `onValueChange` / `onValueComplete` while the field is locked.

I noticed this while writing regression tests for locked OTP states.

## Expected behavior

If `OTPField.Root` is `readOnly` or `disabled`, the hidden validation input should ignore changes the same way the visible slot inputs do.

## Reproducible example

```tsx
<OTPField.Root
  length={6}
  name="otp"
  readOnly
  onValueChange={(value) => {
    console.log('change', value);
  }}
  onValueComplete={(value) => {
    console.log('complete', value);
  }}
>
  <OTPField.Input />
  <OTPField.Input />
  <OTPField.Input />
  <OTPField.Input />
  <OTPField.Input />
  <OTPField.Input />
</OTPField.Root>
```

Then dispatch a change event against the hidden validation input:

```ts
const hiddenInput = document.querySelector('input[name="otp"]');
fireEvent.change(hiddenInput!, { target: { value: '123456' } });
```

Current result:

- visible slots update to `123456`
- `onValueChange` fires
- `onValueComplete` fires

Expected result:

- value remains unchanged
- callbacks do not fire

The same issue also happens with `disabled`.

## Base UI version

`@base-ui/react` v1.4.1

Chrome

## Which OS are you using?

macOS

N/A

## Additional context

There is already existing coverage for hidden-input autofill (`handles password manager autofill through the hidden input`), so this seems to be the same supported path not honoring the locked state consistently.

I already have a small patch and regression tests prepared and can open a PR if this is the expected fix direction.

lunaxislu:
Fixes #4805

## Summary

`OTPField.Input` already ignores changes while the field is `readOnly` or `disabled`, but the hidden validation input rendered by `OTPField.Root` still accepted `change` events.

This patch applies the same guard to the hidden input's `onChange` handler so locked OTP fields cannot be updated through that path.

## Why

There is already existing coverage for hidden-input autofill, so this aligns the locked-state behavior across both input paths instead of changing the intended hidden-input flow.

## Changes

- add `disabled || readOnly` guard to the hidden validation input `onChange`
- add a regression test for readonly hidden-input autofill
- add a regression test for disabled hidden-input autofill

## Testing

- `pnpm test:chromium OTPField --no-watch`


atomiks (approved):
Thanks for the fix @lunaxislu

PR #4807, opened 2026-05-11
[tooltip] Cancel pending open on pointerdown

atomiks:
A post-1.4.1 regression let a tooltip open if its trigger received pointerdown while the hover delay was pending and pointerup/click had not fired yet. This is marked internal and should not be mentioned in release notes.

## Root cause

After interaction splitting, closed tooltips no longer had dismiss interactions mounted, so pointerdown did not emit the internal close event that clears pending hover timers.

## Changes

- Cancel pending tooltip opens on pointerdown when the trigger is configured to close on click.
- Cover pointerdown without pointerup while the hover delay is pending.


mj12albert (approved):
Looks good~

PR #4810, opened 2026-05-11
[select][combobox][number field] Ignore hidden-input changes while readonly or disabled

### LINKED ISSUE #4809: [select][combobox] Hidden input ignores readonly/disabled during browser autofill
## Summary

Follow-up to #4805 — the same hidden-input autofill pattern (fixed in #4806 for OTPField) is also present in `SelectRoot` and `AriaCombobox`

`SelectRoot` and `AriaCombobox` use the same hidden-input autofill pattern that `OTPFieldRoot` used before #4806, but their hidden-input `onChange` handlers only guard `event.nativeEvent.defaultPrevented`, not `disabled` or `readOnly`.

As a result, a hidden-input change can still update component state even when the component is in a locked state.

## Where

- `packages/react/src/select/root/SelectRoot.tsx`
- `packages/react/src/combobox/root/AriaCombobox.tsx`

`AriaCombobox` is the shared root used by both `Combobox.Root` and `Autocomplete.Root`.

## Why it matters

This is a narrow follow-up consistency gap from #4806:

- hidden-input autofill is already treated as a supported path
- `disabled` and `readOnly` are already wired to the hidden input
- but the autofill handler still ignores those locked states

The user-facing case is narrower than OTP, but the contract drift is the same shape.

lunaxislu:
Fixes #4809.

Follow-up to #4806.

`SelectRoot` and `AriaCombobox` use the same hidden-input autofill pattern that
`OTPFieldRoot` used before #4806, but their hidden-input `onChange` handlers
still only guard `defaultPrevented`, not `disabled || readOnly`.

This PR applies the same locked-state guard to those remaining handlers.

`AriaCombobox` is the shared root used by:
- `Combobox.Root` (`selectionMode='single' | 'multiple'`)
- `Autocomplete.Root` (`selectionMode='none'`)

## What changed

- `SelectRoot.tsx`: hidden-input `onChange` guard now matches OTP's
- `AriaCombobox.tsx`: same one-line guard update
- Regression tests added for:
  - `Select`
  - `Combobox` (`selectionMode='single'`)
  - `Autocomplete` (`selectionMode='none'`)

Each regression test fires `change` on the hidden input and verifies that no public state change happens while `readOnly` or `disabled`.

No public API change.


packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1260,7 +1260,7 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
           // Handle browser autofill.
           onChange(event: React.ChangeEvent<HTMLInputElement>) {
             // Workaround for https://github.com/facebook/react/issues/9023
-            if (event.nativeEvent.defaultPrevented) {
+            if (event.nativeEvent.defaultPrevented || disabled || readOnly) {
```
mj12albert, line 1263: Codex finding, applies to Select too:
> The new guard returns from the component’s hidden-input `onChange,` but it does not stop the `validation.getInputValidationProps` handler that is merged onto the same input. `mergeProps` runs the external handler first, then the internal handler unless `event.preventBaseUIHandler()` was called. That internal Field handler still calls `clearErrors(name)` and `commit(event.currentTarget.value, ...)`, so a readonly/disabled hidden-input change can still clear Form errors or update Field validity with the autofilled value.

lunaxislu: Fixed.
 
The hidden-input autofill guard now calls `event.preventBaseUIHandler()` before returning in both Select and Combobox, preventing the merged Field validation `onChange` from clearing errors or committing validation while `readOnly` or `disabled`.

Added regression tests for `readOnly`/`disabled` hidden-input autofill with existing form errors.

packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1258,9 +1258,10 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
             (inputRef.current || triggerElement)?.focus();
           },
           // Handle browser autofill.
-          onChange(event: React.ChangeEvent<HTMLInputElement>) {
+          onChange(event: BaseUIEvent<React.ChangeEvent<HTMLInputElement>>) {
             // Workaround for https://github.com/facebook/react/issues/9023
-            if (event.nativeEvent.defaultPrevented) {
+            if (event.nativeEvent.defaultPrevented || disabled || readOnly) {
+              event.preventBaseUIHandler();
```
mj12albert:
```suggestion
              event.preventBaseUIHandler?.();
```
Could you see if this is needed? preventBaseUIHandler  here may throw if the Select/Combobox isn't wrapped in Field, all the tests wrap with Field but `<Combobox.Root name="x" disabled>` is supported too

lunaxislu: Good catch. Outside `Field.Root`, `getInputValidationProps` returns theprops directly, so the change event is not wrapped by `mergeProps` and `preventBaseUIHandler` may be absent. I kept the optional call and added outside-Field regression coverage for both `readOnly` and `disabled` in
Select and Combobox

atomiks:
## Code Review (GPT-5.5)
Request changes 🔴 because the same readOnly hidden-input autofill path still exists in NumberField, and the new regression coverage can be much smaller.

### 1. Bugs / Issues

#### 1. 🔴 NumberField still accepts hidden-input autofill while readOnly (blocking)

The PR fixes Select and AriaCombobox, but NumberField has the same supported hidden-input autofill path and still only guards `defaultPrevented`. By code reading, a `readOnly` NumberField can still update through this hidden input because the input does not receive `readOnly`, and the handler does not check the root locked state.

`packages/react/src/number-field/root/NumberFieldRoot.tsx`

```tsx
onChange(event) {
  // Workaround for https://github.com/facebook/react/issues/9023
  if (event.nativeEvent.defaultPrevented) {
    return;
  }

  // Handle browser autofill.
  const nextValue = event.currentTarget.valueAsNumber;
  const parsedValue = Number.isNaN(nextValue) ? null : nextValue;

  setDirty(parsedValue !== validityData.initialValue);
  setValue(parsedValue, details);
}
```

This leaves the same class of bug open in another public component that already has a hidden-input browser autofill test.

**Fix:** Apply the same locked-state guard to `NumberFieldRoot` and add a regression for at least `readOnly` hidden-input autofill; include `event.preventBaseUIHandler?.()` there too when the handler is wrapped by `Field.Root`.

#### 2. 🟡 The locked-state regression tests can be table-driven (non-blocking)

The new tests repeat the same matrix several times: `readOnly` vs `disabled`, inside `Field.Root` vs outside `Field.Root`, then the same hidden-input change and “no public state changed” assertions. That makes this bug fix look larger than it is.

`packages/react/src/select/root/SelectRoot.test.tsx`

```tsx
it('ignores hidden-input autofill when readOnly', async () => { ... });
it('ignores hidden-input autofill when disabled', async () => { ... });
it('ignores hidden-input autofill when readOnly outside Field', async () => { ... });
it('ignores hidden-input autofill when disabled outside Field', async () => { ... });
```

**Recommendation:** Extract small render/assert helpers per component and drive the lock state with `it.each(['readOnly', 'disabled'] as const)`. For Select and Combobox, a second table dimension for `withField` would keep the Field error assertion where it matters while removing most of the duplicated JSX.

### 2. Simplification Opportunities

The production change is already close to minimal for Select and AriaCombobox. If NumberField is included too, a tiny shared predicate may be worth it, for example “ignore this hidden input change when the native event was already prevented or the root is locked, and prevent the Field validation handler when possible.” I would only add that helper if it removes duplication across Select, AriaCombobox, and NumberField; otherwise the source-level inline guard is fine.

lunaxislu:
Addressed the NumberField gap in 5b5865dc5.

- Added the same locked-state guard to NumberField's hidden input autofill path: `defaultPrevented || disabled || readOnly` now returns early and calls `event.preventBaseUIHandler?.()` so Field validation handlers do not clear existing errors while locked.
- Added NumberField regression coverage for `readOnly` and `disabled`, both inside and outside `Field.Root`.
- Reduced the existing Select and Combobox locked-state regression tests to table-driven coverage while preserving the inside/outside Field assertions.

I considered adding `readOnly={readOnly}` to NumberField's hidden input for consistency with Select/Combobox/OTPField, but kept this as a handler guard only. Unlike those text hidden inputs, NumberField uses a `type="number"` hidden input with native `min`/`max`/`step`/`required` validation; setting `readOnly` on the input would bar it from constraint validation and could silently allow invalid read-only values through form submission.

Verified with:

- `pnpm test:jsdom NumberField --no-watch`
- `pnpm test:jsdom Select --no-watch`
- `pnpm test:jsdom Combobox --no-watch`

atomiks (approved):
Thanks for the fix @lunaxislu

PR #4814, opened 2026-05-12
[internal] `useLayoutEffect` with `ResizeObserver`

romgrk:
Make sure we use `useLayoutEffect` with `ResizeObserver`.


flaviendelangle (commented):
Worth double checking, but I think the `registerTabResizeObserverElement` in `TabsTab` should also be a layout effect

flaviendelangle (approved):

Issue #4745, opened 2026-05-05
[autocomplete] Manually reset highlighted item


timakro:
I'm building a more complex autocomplete component based on Base UI Autocomplete (similar to GitHub issue search). When one item is autocompleted I need to keep the autocomplete open but show a new set of options -- the highlighted item should reset to null (no item highlighted). How can I control the highlighted item from the outside or call into the component to reset the highlighted item to null?

atomiks:
@timakro  Can you post a Stackblitz demo of what you're trying to build?

timakro:
https://stackblitz.com/edit/rfuxdqib?file=src%2FApp.tsx

First select category (Animals or Cars), then the specific animal or car. What I am actually building is very close to GitHub issue search where you first select e.g. "author" and then the specific author. This StackBlitz is missing a lot of course but it should demonstrate the basic issue. When swapping the set of items the Base UI Autocomplete component does not reset the highlighted item index.

One workaround is to temporarily clear the list of items, but this requires some unsightly coordination (first setting item list to empty list, remembering what it should be set to shortly after, etc.) also it causes a visual flash of the empty list.

I'm not well versed in Base UI API design, but I've noticed the APIs are very well thought through. So if this is too niche/not worth adding to Base UI I do understand.

oliviertassinari:
I recall a similar conversation in  https://github.com/mui/material-ui/issues/20852, around the need for more control over what's highlighted.

PR #4815, opened 2026-05-12
[menu] Add highlightItem('first' | 'last' | 'none') imperative action

chsmc-ant:
When a menu is opened programmatically (via controlled `open`, or `MenuHandle.open()`, or `onOpenChange(true)` from a custom interaction that isn't one of `Menu.Trigger`'s built-in open keys), `useListNavigation` leaves `activeIndex` as `null`. This is correct for pointer-driven opens, but for custom keyboard shortcuts it means every item renders with `tabindex="-1"` and `FloatingFocusManager` moves focus to the popup container rather than the first item — the user then has to press ↓ once more before arrow navigation works.

This PR adds an imperative `highlightItem(target: 'first' | 'last' | 'none')` action so callers can ask the menu to highlight an item once it opens:

- **`Menu.Root` `actionsRef`** — alongside `unmount` / `close`:

  ```tsx
  const actionsRef = React.useRef<Menu.Root.Actions>(null);

  function onCustomKeyDown(event: React.KeyboardEvent) {
    if (event.key === 'ArrowRight') {
      setOpen(true);
      actionsRef.current?.highlightItem('first');
    }
  }

  <Menu.Root open={open} onOpenChange={setOpen} actionsRef={actionsRef}>
  ```

- **`MenuHandle.open(triggerId, highlightItem?)`** — same values as a second parameter:

  ```tsx
  handle.open('my-trigger', 'first');
  ```

The requested target is stored as `pendingHighlightItem` on `MenuStore` and resolved inside `useListNavigation`'s initial-sync layout effect once the list is populated, using `getMinListIndex` / `getMaxListIndex` so hidden items are skipped — the same logic used for the built-in `focusItemOnOpen` path. `useListNavigation` then moves DOM focus to the resolved item (for Menu; virtual-focus consumers like Combobox would only update `activeIndex`).

A `Menu.Root.HighlightItem` type alias is exported for the `'first' | 'last' | 'none'` union.

Preview: https://deploy-preview-4815--base-ui.netlify.app/react/components/menu#controlled-mode-with-multiple-triggers

---

> Earlier revisions exposed a raw `setActiveIndex(index)` setter and then a `focusItem()` action; reworked per review to use `highlightItem()` with directional targets and to cover `MenuHandle.open` as well.


oliviertassinari:
To help the team a bit (and your team on Claude), here is a reproduction that seems to reproduce the problem: https://stackblitz.com/edit/va1fcbzq?file=src%2FApp.tsx (repro based on [this demo](https://master--base-ui.netlify.app/react/components/menu#controlled-mode-with-multiple-triggers)):

https://github.com/user-attachments/assets/41d3d1ea-e22d-4acb-9931-cc37b101cb91

---

It's not clear that this should be solved with a new API; this feels like a bug: I would expect the same behavior regardless of how the menu is open: with a trigger, with the `open` prop.

For example, in https://mui.com/material-ui/react-menu/, the focus is placed on the first menu item when the menu opens (with the keyboard) but the behavior can be configured with [a prop](https://mui.com/material-ui/api/menu/#menu-prop-disableAutoFocusItem).

Let's wait for team inputs.

chsmc-ant:
Thanks @oliviertassinari and good call! I opened #4816 as well so we can compare this approach to that one. Will defer to the team on which they'd prefer.

atomiks:
Thanks for the PR @chsmc-ant

Makes sense as controlled-opens are assumed to always be pointer-driven, but a keyboard-driven controlled open is also valid in which case the highlight should be set on the first item.

An imperative action also makes more sense than a prop like `autoHighlight` (which Autocomplete has) since this is a one-shot/event-driven action that probably shouldn't need state.

The only issue is naming: `setActiveIndex` is quite broad and "active" is more so an internal name as we use `highlighted` or `focus` publicly. 

I'm also not sure if exposing the index is a good idea, because 1) disabled items should be skipped on initial open, which should pass through the internal logic to skip them, and 2) it's likely hard to keep track of which index is the last item if the controlled open is an `ArrowUp` for example. 

Possible alternative: `.focusItem('first' | 'last' | 'none')` or `.focusItem('start' | 'end' | 'none')`

cc: @colmtuite 

---

Side note: in case you need this behavior before release, this technically already works, if hacky:

```jsx
firstItem.focus();
// or firstItem.dispatchEvent(new MouseEvent('mousemove', { bubbles: true }))
```

michaldudak:
We have another imperative API that has the same issue: `MenuHandle.open`. Whatever solution we land on for the `actionsRef`, we should also apply it there. I like @atomiks idea with `.focusItem('first' | 'last' | 'none')`. We could add an optional parameter to the handle's `open` method: `open(triggerId: string, focusItem?: 'first' | 'last' | 'none')`.

chsmc-ant:
Thanks for the feedback @atomiks @michaldudak — pushed 974094c5f which:

- Replaces `setActiveIndex(index)` with `focusItem('first' | 'last' | 'none')` on `actionsRef`, resolved via `getMinListIndex`/`getMaxListIndex` so hidden items are skipped.
- Adds the same option as a second parameter to `MenuHandle.open(triggerId, focusItem?)`.
- Exports `Menu.Root.FocusItem` for the union type.

PR description updated to match. Happy to bikeshed `'first' | 'last'` vs `'start' | 'end'` if there's a preference.

packages/react/src/menu/root/MenuRoot.tsx
```diff
@@ -423,6 +435,10 @@ export const MenuRoot = fastComponent(function MenuRoot<Payload>(props: MenuRoot
     openOnArrowKeyDown: parent.type !== 'context-menu',
     externalTree: nested ? floatingTreeRoot : undefined,
     focusItemOnHover: highlightItemOnHover,
+    pendingFocusItem,
+    onPendingFocusItemChange() {
+      store.set('pendingFocusItem', null);
+    },
```
atomiks, line 440: Folded this into `useListNavigation` as it removes the duplicate effect, and AriaCombobox will likely want this behavior as well: https://github.com/mui/base-ui/issues/4745

atomiks:
I think `.highlightItem()` is likely a better term for this, because the public API is about the `highlighted` state, even though focus is in sync with it. And AriaCombobox doesn't use DOM focus, so `highlight` makes more sense if this is to be added there — @michaldudak

chsmc-ant:
thanks @atomiks happy to go ahead and implement that rename if you think it's the right approach

chsmc-ant:
Renamed to `highlightItem` in f4aa2a88f — covers `actionsRef.highlightItem()`, `MenuHandle.open(triggerId, highlightItem?)`, the exported `Menu.Root.HighlightItem` type, and the internal `pendingHighlightItem` plumbing through `MenuStore` / `useListNavigation`. Left the pre-existing internal `focusItem`/`resolveFocusItem` helpers in `useListNavigation` alone since those are about DOM focus and shared with the `focusItemOnOpen` path.

oliviertassinari:
I wonder how we will fix this demo https://base-ui.com/react/components/menu#controlled-mode-with-multiple-triggers later on. I guess we will add something like:

```jsx
<button
  type="button"
  className={styles.Button}
  onClick={(event) => {
    // Detect if the interaction was explicitly driven by a keyboard. Unless we need isVirtualClick().
    const isKeyboard = event.detail === 0;
    setOpen(true);
    actionsRef.current!.highlightItem(isKeyboard ? 'first' : null);
```

Not fully self managed by the popover (testing for :focus-visible), but this stays kind of simple.

atomiks (approved):
This looks right now, thanks for the contribution @chsmc-ant 

I ran Codex & Claude reviews on it and they both found some issues which I hardened in the follow up commits.

Something intentional (that they reported): I removed the `waitForListPopulatedFrame` as that was unnecessary on top of the `queueMicrotask` based on all tests:

> CompositeList does the index reconciliation, then useCompositeListItem writes the actual DOM node into listRef.current[index] once the item gets its resolved index. That can require one extra layout-effect/update turn for composed items, so the PR waits with queueMicrotask before deciding the list is empty.

It was from the `floating-ui` fork which added an extra rAF just as a precaution, but only the microtask is needed

---

_Unrelated to this PR, but needs to be fixed as well_: there's a bug where if you mouse into the menu that was controlled-open by a different trigger, it closes upon `mouseleave` unexpectedly, because the hover hook requires a click on its trigger to block the close on `mouseleave`. I'll fix it in a new PR

chsmc-ant:
Thanks @atomiks I think the failing test is flake/unrelated, is there anything else I need to do before we can merge this?

docs/src/app/(docs)/react/components/context-menu/types.md
```diff
@@ -11,22 +11,22 @@ Doesn't render its own HTML element.
...
+| actionsRef           | `React.RefObject<MenuRoot.Actions \| null>`                                    | -            | A ref to imperative actions. `unmount`: When specified, the menu will not be unmounted when closed.&#xA;Instead, the `unmount` function must be called to unmount the menu manually.&#xA;Useful when the menu's animation is controlled by an external library.`close`: When specified, the menu can be closed imperatively.`highlightItem`: Highlight the `'first'` or `'last'` item, or `'none'` to clear the highlight.&#xA;Useful when opening the menu programmatically from a custom interaction&#xA;so that the first item is highlighted instead of the popup container receiving focus. |
```
aarongarciah, line 20: _Unrelated to this PR_

@dav-is something to improve in the types.md file generation? `close` and `highlightItem` are bullets in the original code (see https://github.com/mui/base-ui/pull/4815/changes#diff-0c9aa23d488587208914fe2b3474310cf0586456d23b0cd370e448b2a5ddc7dfR639-R640). I we can't use bullets inside the markdown table (I've seen some MD renderers allowing to use them when using `<br>` for new lines), we could add a space at least.

aarongarciah:
> Thanks @atomiks I think the failing test is flake/unrelated, is there anything else I need to do before we can merge this?

@chsmc-ant yes, we have a flacky test. I've re-run it and now it's all green.

colmtuite:
Hey @chsmc-ant 👋 Apologies for jumping in at the last minute here, but I'm curious about this API.

It seems that this might be solvable already with the [detached triggers API](https://base-ui.com/react/components/menu#detached-triggers)?

Also, I'm struggling to imagine a use case where you might want to highlight the last item? Pressing up/down on an isolated trigger seems like a strange case to me, I'd expect that only Space/Enter would work. And when opening programatically, I'm also struggling to imagine any use case where you'd want to highlight anything other than the first item.

So I'm wondering if this should be handled automatically, somehow. Or whether it's already achievable with our detached triggers API.

Are you able to share more about exactly what your use case is?

oliviertassinari:
Side note, I wondered if it would make sense to have a global logic to record the modality in the background, e.g. somethigg like this:

```jsx
// Keep track of the current interaction modality
let currentModality = 'pointer';

window.addEventListener('keydown', (event) => {
  if (!event.metaKey && !event.altKey && !event.ctrlKey) {
    currentModality = 'keyboard';
  }
}, { capture: true });

window.addEventListener('mousedown', () => {
  currentModality = 'pointer';
}, { capture: true });

window.addEventListener('touchstart', () => {
  currentModality = 'pointer';
}, { capture: true });
```

or (assuming .matches(':focus-visible') doesn't kill performance)

```js
const AppState = {
  modality: 'pointer' // Default fallback ('pointer' or 'keyboard')
};

window.addEventListener('focusin', (event) => {
  // Use a microtask to let the browser update the pseudo-class state
  queueMicrotask(() => {
    if (event.target.matches(':focus-visible')) {
      AppState.modality = 'keyboard';
    } else {
      AppState.modality = 'pointer';
    }
  });
});
```

This way, when there is a programmatic change of state from a prop (where there is no access to the `event`), the focus can default to the modality. Maybe that would solve OG's problem, out of the box.

chsmc-ant:
@colmtuite thanks for taking a look — happy to share the use case.

**Use case:** A vertical app sidebar where each row has an associated flyout menu. Desired keyboard UX: `→` on the sidebar row opens the flyout with the first menuitem highlighted (so `↑`/`↓` work immediately), `←` inside the flyout closes it and returns focus to the row. It's the same cross-orientation pattern Menubar/submenus already implement, but the parent list isn't a Base UI Menubar — it's an existing sidebar with its own routing/selection state, and converting it to a Menubar/Composite parent is a larger refactor than we can take on.

**Why detached triggers don't cover it:** The row *is* a `Menu.Trigger` (in-tree or detached — same result). The trigger's keydown handler only sets `useListNavigation`'s internal `keyRef` for main-orientation keys (`↑`/`↓` on a vertical menu) plus Enter/Space; `→` isn't a navigation key for a non-nested vertical menu, so `keyRef` stays `null` and `focusItemOnOpen: 'auto'` falls through to focusing the popup. `MenuHandle.open(triggerId)` is purely programmatic and never touches the keydown path, so it lands in the same place. Either way we end up writing a custom `→` handler that opens the menu and then needs a way to say "and highlight the first item" — which is what this PR adds.

**On `'last'`:** Agreed it's not needed for our case — we only use `'first'`. It's there for symmetry with the existing `↑`-on-trigger → last-item behaviour, per @atomiks's original suggestion. Happy to drop it and ship `'first' | 'none'` if you'd rather keep the surface minimal.

**On handling it automatically:** Would love that. @oliviertassinari's modality-tracking idea would make this zero-config for the common case. This PR was scoped as the minimal escape hatch that doesn't commit to a heuristic; if the team would rather pursue auto-detection instead, happy to close this in favour of that (or land this now and treat auto-detection as a follow-up that makes the explicit call unnecessary in most cases).

atomiks:
@chsmc-ant thanks for the explanation. Would users know `→` works in that pattern if it's not a menubar? Or is there an ARIA pattern you're using that lets AT know it's menubar-like with keyboard actions?

On @oliviertassinari's point: this seems interesting. I wonder if `:focus-visible` can work instead on the `.Popup` element so we can automatically know if it was pointer- or keyboard-opened as the browser should track the modality via that state for us. That would more cleanly solve this issue than adding an imperative escape hatch.

---

On the (original) implementation:

I'm okay with ditching `'last'` until an actual use case comes up instead of adding it just for symmetry.

I also think, for future-proofing/clarity, we should make `handle.open(id, ...)` take an object with `highlightItem` as the only parameter (for now). 

So the API looks like:

- `actionsRef.current?.highlightItem('first' | 'none')`
- `handle.open('id', {highlightItem: 'first' | 'none'})`

But again, maybe `:focus-visible` checking is a better solution for this

chsmc-ant:
@atomiks

> Would users know `→` works in that pattern if it's not a menubar? Or is there an ARIA pattern you're using that lets AT know it's menubar-like with keyboard actions?

They don't need to — `→` isn't the accessible path, it's an additive shortcut. The sidebar row is a `Menu.Trigger`, so it already gets `aria-haspopup="menu"` + `aria-expanded` from Base UI; screen readers announce "menu button, collapsed" and AT users open it with Enter/Space, which Base UI already handles correctly (first item focused via the built-in `keyRef` path). That flow is unchanged.

The `→` binding is a supplementary shortcut for sighted-keyboard / power users, mirroring the convention from native vertical-nav UIs (macOS Finder sidebar, VS Code Explorer, Slack channel list) where `→` enters and `←` backs out. WAI-ARIA APG's [Disclosure Navigation Menu](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/) pattern explicitly calls out optional arrow-key support that supplements (not replaces) the standard Tab/Enter interaction, without requiring `menubar` semantics — that's the model here. So no special role is advertising `→`; `aria-haspopup` covers AT, and `→` is just there for users who try it by muscle memory.

> maybe `:focus-visible` checking is a better solution for this

I'd be happy if it were — it'd cover the `→` case nicely. The bit I'm less sure it covers is fully-detached opens like `handle.open('id')` from a command palette / global hotkey, where there's no focused trigger element for the browser to carry modality from; the popup mounts fresh and `:focus-visible` on it would depend on whatever the last interaction in the document happened to be. The imperative escape hatch is unambiguous there. Not opposed to exploring the heuristic though — they're not mutually exclusive.

> ditching `'last'` / `handle.open(id, { highlightItem })` object param

Both sound good to me — happy to push those changes if we're going ahead with the imperative API.

oliviertassinari:
> The imperative escape hatch is unambiguous there. Not opposed to exploring the heuristic though — they're not mutually exclusive.

**Some historical context**: We had this in Material UI for a while (at the time where `:focus-visible` had no enough browser support). The logic was here: https://github.com/mui/material-ui/blob/v5.11.0/packages/mui-utils/src/useIsFocusVisible.ts. We adopted it from: https://github.com/WICG/focus-visible. On the bundle size cost, it might be too high for this path to be viable. 

*Oops, closed by mistake, wrong button on my phone.*

atomiks:
We already use `matchesFocusVisible` internally which checks the native CSS `:focus-visible` selector, which you can use once the popup receives focus

Suggestion: try adding this to the floating props in `useListNavigation`  to see if it acts how you'd expect from a UX perspective (instead of the imperative API when testing)

```tsx
onFocus(event) {
  const target = getTarget(event.nativeEvent) as Element | null;

  if (
    focusItemOnOpenRef.current === 'auto' &&
    open &&
    store.select('activeIndex') == null &&
    keyRef.current == null &&
    target === event.currentTarget &&
    matchesFocusVisible(target)
  ) {
    indexRef.current = getMinListIndex(listRef, disabledIndicesRef.current);
    onNavigate(event);
  }
}

```

PR #4817, opened 2026-05-13
[navigation menu] Fix keepMounted content sizing

atomiks:
`<NavigationMenu.Content keepMounted>` could transition the size of the popup on open incorrectly + fail to transition the popup dimensions when hovering over nested inline submenus.

## Root cause

Kept-mounted content can stay in the DOM while inactive, so it must not keep or regain ownership of the shared content ref used for popup sizing.

Before: https://stackblitz.com/edit/dwjmyhms
After: https://stackblitz.com/edit/dwjmyhms-hwgbd7w8

`<NavigationMenu.Portal keepMounted>` was one workaround before. Though that has a bug with the arrow transitioning (separate issue to fix)

## Changes

- Keep the shared current content ref owned only by active content, and clear it only when that same content closes.
- Observe `hidden` changes and defer the first nested-content opening measurement so the popup uses the final nested content size.
- Add regression coverage for kept nested content opening, hidden changes, close/reopen, and mixed kept/unmounted siblings.


atomiks:
## Code Review (GPT-5.5)
Approve ✅: the keepMounted sizing bug is fixed at the shared-content-ref ownership point, the hidden-attribute sizing path is covered, and the earlier ref-stability nit is resolved.

### 1. Bugs / Issues (None)

Reviewed as a `bug-fix` with secondary `tests` / `react-ui` scope against freshly fetched `upstream/master` (merge-base `646f5803`). The patch addresses the root cause rather than only the visible animation symptom: inactive kept-mounted `NavigationMenu.Content` nodes can now mount in the viewport without taking ownership of `currentContentRef`, while active content still re-asserts ownership through the existing layout effect when the same DOM node re-enters.

The follow-up change to use `useStableCallback` keeps `handleCurrentContentRef` from changing identity on every `open` flip, so it preserves latest-state reads without churning the merged ref passed through `CompositeRoot` / `useMergedRefsN`.

The Trigger-side MutationObserver expansion is also targeted: kept nested submenu switches toggle `hidden` instead of inserting or removing nodes, and observing `attributeFilter: ['hidden']` makes those visibility changes feed the existing sizing sync path without watching unrelated attributes.

### Test Coverage Assessment
The regression coverage is proportional to the bug. The new tests cover kept nested default content during first portal entry, switching kept inline nested content, direct `hidden` attribute mutation, and the existing top-level keepMounted trigger-switch path remains covered.

Validation performed:

- `pnpm test:jsdom NavigationMenu --no-watch`
- `pnpm test:chromium NavigationMenu --no-watch`
- `pnpm exec eslint packages/react/src/navigation-menu/content/NavigationMenuContent.tsx --max-warnings 0`
- `pnpm exec prettier --check packages/react/src/navigation-menu/content/NavigationMenuContent.tsx`

I also verified the new kept-content regression expectations against the refreshed base implementation before the follow-up commit; the targeted cases failed there as expected, which confirms they gate the bug being fixed.

PR #4819, opened 2026-05-13
[useButton] Stop setting type on non-native buttons

chuganzy:
This is a minor PR to avoid passing `type={undefined}` from `useButton` when `nativeButton` is `false`.

Previously, custom non-native button renderers could receive an unnecessary `type` prop:

```tsx
import { Button } from "@base-ui/react/button";

function App() {
  return <Button render={<CustomButton />} nativeButton={false} />
}


function CustomButton(props) {
  // Previously props included type={undefined}
  return <span {...props} />
}
```

### DIFF
--- a/packages/react/src/internals/use-button/useButton.test.tsx
+++ b/packages/react/src/internals/use-button/useButton.test.tsx
@@ -41,6 +41,19 @@ describe('useButton', () => {
         });
       });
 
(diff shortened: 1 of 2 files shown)


atomiks (approved):
Thanks for the fix @chuganzy

PR #4820, opened 2026-05-13
[select][popover] Fix nested Select requiring extra outside clicks to dismiss parent

mattrothenberg:
## Summary

Fixes nested `Select`/`Popover` dismissal paths where a parent Popover could require extra outside presses after Select interactions.

The final fix is intentionally small:

- `Select.Root` now uses `useDismiss(floatingContext)` defaults instead of hardcoding `bubbles: false`. This keeps Escape isolated while allowing outside-press dismissal to cascade like other popup components.
- `useDismiss` clears leftover drag-out suppression when a new press begins outside the floating element and its trigger, so stale suppression from a previous Select interaction cannot consume the next real outside click.

## Details

`Popover` uses intentional mouse outside-press handling. In nested popup flows, an earlier interaction can leave `suppressNextOutsideClickRef` set even though the next press is a fresh outside dismissal attempt. The guard clears that stale suppression on the new outside `pointerdown`/`mousedown`; it does not clear suppression for the eventual `click`, so legitimate inside-to-outside drag suppression is preserved.

Removing Select's explicit `bubbles: false` relies on `normalizeProp(undefined)` defaults: `escapeKey` remains non-bubbling and `outsidePress` can bubble.

## Tests

Added Chromium coverage for Select inside Popover dismissal:

- reopened aligned Select dismisses the Popover with one outside press
- selected/reopened Select dismisses the Popover with one outside press
- modal Popover stays open when choosing a non-modal aligned Select item
- Escape from a non-modal Select does not bubble to the Popover
- drag from a modal Select trigger outside all popups does not consume the next outside press


atomiks:
Coudn't get an automated failing test for:

- Open Popover
- Open Select
- Click outside both (force-click, not soft-tap) requiring 3 clicks instead of 2 (or 1 depending on if the popups both close simultaneously which depends on the pointer flow.)

But I manually verified that it works now in browser, with regular dismiss and drag dismiss flows

Before: https://stackblitz.com/edit/emel6xue
After: https://stackblitz.com/edit/emel6xue-thchawh7

atomiks (approved):

PR #4821, opened 2026-05-14
[code-infra] Deduplicate dependencies using Linux

michaldudak:
The stale lockfile errors are back (see https://app.circleci.com/pipelines/gh/mui/base-ui/25342/workflows/83dc889b-1767-4f31-8889-7cba4570af94/jobs/292838). This is an attempt to resolve them. It seems that the source of these issues is a different OS `dedupe` runs on. I created a script that runs `dedupe` in a Docker container with the same image that's used on CI.

The script runs pnpm install and pnpm dedupe inside the same `cimg/node:22.21.1` Docker image used by CircleCI’s shared mui-node executor, while isolating node_modules and pnpm’s store from the host checkout. It also supports passing args through to pnpm dedupe, e.g., `pnpm dedupe-harder -- --check`, and allows overriding the image with DEDUPE_HARDER_IMAGE.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "deduplicate": "pnpm dedupe",
+    "dedupe-harder": "tsx ./scripts/dedupeHarder.mts",
     "build": "lerna run --no-private build",
     "release:version": "lerna version --no-changelog --no-push --no-git-tag-version --no-private",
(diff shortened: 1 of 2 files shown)


Janpot:
We recently solved a similar issue on @ZeeshanTamboli 's machine using `pnpm cache delete`. Did you try that already?

michaldudak:
It does help indeed. Thanks!

PR #4824, opened 2026-05-14
[number field] Sync pasted input during step interactions

### LINKED ISSUE #4823: [number field] Pasted input does not visually sync during wheel or scrub interactions
## Problem

After pasting a valid value into an uncontrolled NumberField, using a step-based interaction can update the numeric value while leaving the visible input text stale.

Repro:

1. Render a NumberField with defaultValue={10}.
2. Paste 20 into NumberField.Input.
3. Use mouse wheel scrubbing or NumberField.ScrubArea to increment the value.

Expected:

- onValueChange reports the stepped value, for example 21 or 22.
- The visible input text updates to the same stepped value.

Observed:

- onValueChange reports the stepped value.
- The visible input can remain at the pasted text, for example 20.

## Cause

Pasting intentionally sets allowInputSyncRef.current = false so raw typed/pasted text is preserved until blur. Keyboard and increment/decrement button step interactions restore input syncing before applying a stepped value, but the wheel and scrub paths did not.

## Notes

This is not about commit timing. Wheel still calls onValueChange immediately and commits on blur; scrub commits on pointerup. The issue is only that the visible input text can stay stale after the stepped value changes.

lunaxislu:
Fixes #4823.

## Problem

After pasting into an uncontrolled NumberField, wheel and scrub interactions could update the numeric value while leaving the visible input text unchanged.

## Cause

Pasting intentionally disables input syncing so raw user text is preserved until blur. Keyboard and increment/decrement button step interactions already restore input syncing before applying a stepped value, but wheel and scrub did not.

## Fix

Restore input syncing at the wheel and scrub interaction boundaries before applying the stepped value. This keeps direct text editing behavior intact while aligning wheel/scrub with the other step-based interactions.

## Verification

- pnpm test:jsdom NumberField --no-watch
- pnpm test:chromium NumberField --no-watch
- pnpm prettier
- pnpm typescript
- pnpm stylelint

### DIFF
--- a/packages/react/src/number-field/root/NumberFieldRoot.test.tsx
+++ b/packages/react/src/number-field/root/NumberFieldRoot.test.tsx
@@ -10,6 +10,26 @@ import { REASONS } from '../../internals/reasons';
 describe('<NumberField />', () => {
   const { render } = createRenderer();
 
+  function pasteText(target: HTMLElement, value: string) {
+    if (isJSDOM) {
+      fireEvent.paste(target, {
+        clipboardData: {
+          getData: (type: string) => (type === 'text/plain' ? value : ''),
+        },
+      });
+      return;
+    }
+
+    const pasteEvent = new Event('paste', { bubbles: true, cancelable: true });
+    Object.defineProperty(pasteEvent, 'clipboardData', {
+      value: {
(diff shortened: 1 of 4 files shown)


atomiks (approved):
Thanks for the fix @lunaxislu

PR #4827, opened 2026-05-15
[otp field] Prevent locked hidden autofill validation

### LINKED ISSUE #4828: [otp field] Field validation runs after locked hidden-input autofill
## Problem

#4806 added a locked-state guard to OTP Field's hidden autofill input, but when the component is wrapped in `Field.Root`, the merged Field validation `onChange` can still run after the OTP Field handler returns.

This means a locked hidden-input autofill change can still clear existing Form errors or commit validation while the OTP Field is `readOnly` or `disabled`, even though the OTP Field value itself does not change.

This is the same Field validation side effect handled for Select, Combobox, and NumberField in #4810.

## Expected behavior

- Hidden-input autofill changes should be ignored while OTP Field is `readOnly` or `disabled`.
- When wrapped in `Field.Root`, the internal Field validation handler should also be skipped.
- Existing Form errors should remain intact.

lunaxislu:

Fixes #4828.

Follow-up to #4806 and #4810.

#4806 added a locked-state guard to OTP Field's hidden autofill input, but when the component is wrapped in `Field.Root`, the merged Field validation `onChange` can still run after the OTP Field handler returns. This can clear existing Form errors or commit validation while the field is `readOnly` or `disabled`.

This PR calls `event.preventBaseUIHandler?.()` before returning from the locked hidden-input autofill path, matching the Select, Combobox, and NumberField handling from #4810.

The tests now cover the locked-state matrix for hidden-input autofill:

- `readOnly` / `disabled`
- inside `Field.Root` / outside `Field.Root`

The outside-Field cases keep asserting that OTP Field public callbacks are not called. The inside-Field cases additionally assert that existing `Field.Error` content is preserved.

Verified with:

- `pnpm typescript`

### DIFF
--- a/packages/react/src/internals/types.ts
+++ b/packages/react/src/internals/types.ts
@@ -3,6 +3,9 @@ import type { BaseUIEvent, ComponentRenderFn, HTMLProps } from '../types';
 
 export type { HTMLProps, BaseUIEvent, ComponentRenderFn };
 
+export type MaybeBaseUIEvent<E extends React.SyntheticEvent<Element, Event>> = E &
+  Partial<Pick<BaseUIEvent<E>, 'preventBaseUIHandler' | 'baseUIHandlerPrevented'>>;
+
 export interface FloatingUIOpenChangeDetails {
   open: boolean;
   reason: string;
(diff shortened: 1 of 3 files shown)


atomiks (approved):
Thanks for the fix @lunaxislu

PR #4830, opened 2026-05-17
[tabs] Add keepMounted prop to TabsRoot

### LINKED ISSUE #4822: [tabs] Add keepMounted prop to Tabs.Root for consistency with Accordion.Root
## Summary

Add a `keepMounted` prop to `Tabs.Root` for consistency with `Accordion.Root`.

`Tabs.Panel` already supports `keepMounted`, but there is no way to set it once for all panels from `Tabs.Root`.

## Motivation

This mainly improves API consistency between Accordion and Tabs.

Today, users can write `<Accordion.Root keepMounted>`, but need to repeat `keepMounted` on every `Tabs.Panel`. A root-level prop would make Tabs match Accordion and provide a convenient default for preserving panel state, avoiding remounts, and supporting measured or animated panel content.

sarthakmalik0810:
Closes https://github.com/mui/base-ui/issues/4822

Add `keepMounted` prop to `<Tabs.Root />`. Can be overridden per panel with `keepMounted` on `<Tabs.Panel />`


aarongarciah:
Thanks for the PR, @sarthakmalik0810. However, we’ve decided not to proceed with it. We prefer to keep `keepMounted` at the panel level. It’s easy to apply to all panels and it’s semantically accurate.

PR #4836, opened 2026-05-18
[preview card] Keep inline preview anchored while open

atomiks:
The React Compiler browser CI exposed a multiline Preview Card test that expected the popup to move to a newly entered line while it was already open. Keeping the opened anchor stable is less disorienting; a new line is only used when opening or reopening.

https://deploy-preview-4836--base-ui.netlify.app/experiments/inline-positioning

With `delay` enabled, it won't choose a new rect while open. Matches Wikipedia

## Root cause

Inline hover coordinates were refreshed on same-trigger mouseenter while open, so incidental positioning updates could make the popup jump.

## Changes

- Keep stored inline coordinates unchanged while the preview card is open.
- Update the multiline trigger test to assert stable positioning while open.


michaldudak:
## Codex Review (GPT-5.5)

I reviewed PR 4836 as a bug-fix against freshly fetched `upstream/master`. The implementation change is narrowly scoped and matches the stated root cause, but one updated browser regression test does not actually guard the old user-facing failure.

### 1. Bugs / Issues

#### 1. 🟡 Browser regression test still false-passes against the old behavior (non-blocking)

`packages/react/src/preview-card/positioner/PreviewCardPositioner.test.tsx`

```tsx
await user.pointer([
  { target: document.body },
  {
    target: trigger,
    coords: {
      clientX: secondLineRect.left + secondLineRect.width / 2,
      clientY: secondLineRect.top + secondLineRect.height / 2,
    },
  },
]);

expectWithin(positioner.getBoundingClientRect().y, firstLineRect.bottom + 5);
```

The old implementation updates `inlineRectCoordsRef` on same-trigger `mouseenter`, but the visible jump only appears after a later positioning update. I verified the PR’s updated browser test still passes when paired with the old `inlineRect.ts` behavior, while the helper-level `inlineRect` test fails as expected. That means the browser test can pass before the regression has a chance to manifest.

**Fix:** After entering the second line, force or wait for the same positioning update path that exposed the jump, then assert the position still tracks `firstLineRect`.

The code change itself is sound: `onMouseEnter` now uses the same “only while closed” guard as `onMouseMove`, so the opened trigger’s captured inline rect remains stable. Passing `isOpenedByThisTrigger` rather than global open state also preserves trigger switching and close-transition re-entry behavior.

### Test Coverage Assessment

I ran:

```bash
pnpm test:jsdom inlineRect --no-watch
pnpm test:chromium PreviewCardPositioner --no-watch
```

Both passed on the PR branch. I also checked regression sensitivity against the old implementation: the updated helper test fails as expected, but the updated browser test still passes, which is the coverage gap above.

### Recommendation

The runtime fix looks correct, and the helper regression coverage is meaningful. I’d tighten the browser test before merging so the user-facing Preview Card behavior is protected too.

atomiks:
## Code Review (GPT-5.5)
Approve ✅ because the inline-rect change matches the intended stable-open behavior, keeps reopen/new-trigger paths intact, and the targeted regression coverage passes.

### 1. Bugs / Issues (None)

The fix is appropriately narrow: `getInlineRectTriggerProps` now gates both React `onMouseEnter` and `onMouseMove` coordinate refreshes while that trigger is already open, so incidental same-trigger re-entry no longer moves the popup to a different wrapped line.

The existing native `updateInlineRectCoords` path remains available for hover-open handling, and other triggers can still capture coordinates because `isOpenedByThisTrigger` is scoped per trigger.

### Test Coverage Assessment
Validated against a freshly fetched `upstream/master` diff.

Ran:
```bash
pnpm test:jsdom inlineRect --no-watch
pnpm test:chromium PreviewCardPositioner --no-watch
```

Both passed. I did not separately execute the updated regression test against base-only source, but by inspection the previous `onMouseEnter: updateCoords` path is the behavior this test now guards against.

PR #4838, opened 2026-05-18
[button] Fix custom tag keyboard click handlers

atomiks:
Custom-tag Buttons handled keyboard activation by directly calling the Button prop handler. This makes Enter and Space dispatch a real click instead, so render-element and delegated click handlers run through the normal event path.

## Root cause

For non-native buttons, `useButton` called the captured `onClick` prop directly:

```ts
externalOnClick?.(event);
```

That skipped other click handlers on the rendered element, including cases like:

```tsx
<Button nativeButton={false} render={<span onClick={handleClick} />} />
```

## Changes

- Dispatch a real click for Enter and Space activation on non-native, non-composite buttons.
- Add missing public Button/Toolbar coverage around the affected keyboard paths and nearby documented behavior: disabled state callbacks and focusable disabled transitions.
- Remove the old duplicate-click test for two merged `useButton` instances. With keyboard activation now using a real `.click()`, that contrived case follows normal click semantics instead of a keyboard-only dedupe path.
- Keep the fix narrow and avoid broader render-prop disabled guard machinery for rare handler placement on disabled render elements.


atomiks:
## Code Review (GPT-5.5 + Opus)
Approve ✅: The real-click dispatch fix addresses the custom-button keyboard regression, the added tests cover the meaningful Button/Toolbar/useButton paths, and the mixed-composite radio concern appears outside Base UI’s supported composition contract.

### 1. Bugs / Issues (None)

The patch fixes the important behavioral mismatch: non-native custom buttons no longer call only the merged Base UI `onClick` handler during keyboard activation. Dispatching a real click means `onClickCapture`, render-element `onClick`, component `onClick`, delegated/ancestor handlers, and normal event propagation now align with pointer-click behavior.

The same-mode nested `useButton` case is covered by the updated composite test: when both layers activate on the same keyboard event, `preventBaseUIHandler` still suppresses the duplicate handler path.

### Mixed Composite / Radio Assessment

I also stress-tested the concern that a `composite: true` `useButton` merged onto the same DOM node as a `composite: false` `useButton` can double-activate Space, because one layer acts on keydown and the other on keyup. That mechanism is real at the hook level, but Opus and I both concluded the reachable public case is not a realistic supported composition.

The concrete reproduction requires rendering `Radio.Root` through another button-like component, for example `Toolbar.Button nativeButton={false} render={<Radio.Root />}`. That shape is already semantically invalid for Base UI: it corrupts or replaces the radio accessibility contract, and for a toolbar-exclusive choice the appropriate component model is `ToggleGroup`/toggle buttons rather than a radio rendered inside another button. Normal supported radio usage (`RadioGroup` → `Radio.Root`) has a single `useButton`, preserves `role="radio"`, and is unaffected.

So I would not block this PR or add machinery to support the invalid radio-inside-button composition family.

### Test Coverage Assessment

I validated the branch with:

- `pnpm test:jsdom Button --no-watch`
- `pnpm test:chromium Button --no-watch`
- `pnpm test:chromium MenuTrigger PopoverTrigger DialogTrigger SelectTrigger Toggle SwitchRoot CheckboxRoot RadioRoot Button --no-watch`
- `pnpm eslint packages/react/src/internals/use-button/useButton.ts packages/react/src/internals/use-button/useButton.test.tsx packages/react/src/button/Button.test.tsx packages/react/src/toolbar/button/ToolbarButton.test.tsx`
- `pnpm typescript`
- `git diff --check upstream/master...HEAD`

I also verified the new regression tests fail against the current upstream implementation in the expected real-click-dispatch places.

flaviendelangle (approved):
# PR review

This PR makes non-native (`nativeButton={false}`) keyboard activation dispatch a real bubbling `click` event instead of calling the captured `onClick` prop directly, so render-element and ancestor click handlers run through the normal event path. The implementation is clean and well-tested, and all CI (jsdom, chromium, E2E, React 18, React Compiler) is green. Nothing is merge-blocking. The main thing worth a maintainer's awareness is the breadth of the behavioral change: it touches every component that can render as a non-native button, and keyboard activation now produces a DOM click that bubbles where it previously did not.

## Bugs (0)

No findings.

I traced each activation path and could not construct a correctness failure:

- **Composite Space** (`useButton.ts:138-150`) and **non-composite Enter/Space** (`:152-166`, `:200`) all route through the new `activateElement`, which fires exactly one click; `preventBaseUIHandler()` correctly suppresses the duplicate in nested/merged `useButton` cases (verified against the rewritten `useButton.test.tsx` nested test).
- The `event.defaultPrevented` early-return for Enter (`:158`) matches native semantics (Enter activates on keydown), and `RadioRoot`'s new Enter `preventDefault` correctly leans on it. Space still activates on keyup, also matching native — the keydown-`preventDefault` asymmetry I initially suspected is *not* a defect, since native Space activation is keyup-driven (consistent with the existing keyup comment in the file).
- The dispatched `PointerEvent('click')` has `detail===0`, `pointerType===''`, `isTrusted===false`, so `isVirtualClick` still returns `true` via its third branch — `SelectItem`'s rewritten `onClick` (with `lastKeyRef` removed) commits correctly on keyboard activation.
- Untrusted dispatched clicks do run native activation behavior (form submission, link navigation), so replacing `currentTarget.click()` with the dispatch in the composite native/link branch is behavior-preserving while additionally carrying modifier state — exactly as the code comment claims.

## Tests (0)

No findings.

Coverage is appropriately expanded: render-element + capture + ancestor click counts, modifier propagation, `preventBaseUIHandler`/`preventDefault` suppression, focusable-when-disabled transitions, and the RadioGroup Enter case. The nested-button test rename matches the new real-click semantics.

## Simplifications (0)

No findings. The `SelectItem` `lastKeyRef` removal and the unified `activateElement` branch are genuine simplifications enabled by the new approach, and the `mergeProps` import was correctly dropped from the test.

## Docs (0)

No findings. The new inline comments in `RadioRoot.tsx`, `SelectItem.tsx`, and `activateElement` accurately describe the new behavior and the `useButton` contract they depend on.

## Verdict

**Approve** — no correctness or test gaps survived verification; the one residual is scope-of-impact, not a defect.

ℹ️ **Heads-up (not blocking):** this changes activation semantics for *every* component that can render non-native (Menu items, Combobox items, Tabs, Accordion/Collapsible triggers, Switch, Checkbox, Select items, dialog/popover triggers, etc. — ~30 components). For those, keyboard activation now bubbles a real DOM `click` to ancestors and reports `isTrusted: false` / `detail: 0`, where the old code called the handler directly without bubbling. CI (including E2E) passing covers the tested integrations, but it's worth a spot-check of any place a non-native Base UI button sits under an ancestor `onClick` (or relies on `event.isTrusted`) that should not react to keyboard activation.

---

PR #4841, opened 2026-05-18
[drawer] Forward viewport style prop

atomiks:
Drawer.Viewport now forwards the `style` prop to the rendered viewport, matching the rest of the component API and keeping conformance coverage in place.

## Root cause

`Drawer.Viewport` destructured `style` but never passed it to `DialogViewport`, so styles set on the component were dropped.

## Changes

- Forward `style` from `Drawer.Viewport` to `DialogViewport`.
- Add the standard conformance suite for `Drawer.Viewport`.

### DIFF
--- a/packages/react/src/drawer/viewport/DrawerViewport.test.tsx
+++ b/packages/react/src/drawer/viewport/DrawerViewport.test.tsx
@@ -4,7 +4,7 @@ import { Combobox } from '@base-ui/react/combobox';
 import { Drawer } from '@base-ui/react/drawer';
 import { Slider } from '@base-ui/react/slider';
 import { fireEvent, flushMicrotasks, screen, waitFor } from '@mui/internal-test-utils';
-import { createRenderer, isJSDOM } from '#test-utils';
+import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
 
 describe('<Drawer.Viewport />', () => {
   beforeAll(function beforeHook() {
@@ -14,6 +14,17 @@ describe('<Drawer.Viewport />', () => {
   });
 
   const { render } = createRenderer();
+
+  describeConformance(<Drawer.Viewport />, () => ({
+    refInstanceof: window.HTMLDivElement,
+    render(node) {
+      return render(
+        <Drawer.Root open>
+          <Drawer.Portal>{node}</Drawer.Portal>
+        </Drawer.Root>,
+      );
+    },
+  }));
 
   function createTouch(target: EventTarget, point: { clientX: number; clientY: number }) {
     if (typeof Touch === 'function') {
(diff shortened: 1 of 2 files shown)


atomiks:
## Code Review: [#4841 — [drawer] Forward viewport style prop](https://github.com/mui/base-ui/pull/4841)

### Overview
Fixes a bug where `Drawer.Viewport` destructured `style` out of its props but never forwarded it to the underlying `DialogViewport` — so any `style` set on `Drawer.Viewport` was silently dropped. Also adds the standard `describeConformance` suite for the component (13 additions, 1 deletion).

### Verification performed
- **Root cause confirmed.** [DrawerViewport.tsx:70](packages/react/src/drawer/viewport/DrawerViewport.tsx#L70) destructures `style` into a local variable, which excludes it from `...elementProps`. Before the fix, `style` reached neither `DialogViewport` directly nor the spread `elementProps`, so it was lost. `render`, `className`, and `children` were already forwarded — `style` was the lone omission, so "matching the rest of the component API" is accurate.
- **Fix is correct.** `DialogViewport` accepts `style` via `BaseUIComponentProps<'divّ>` and resolves it in `useRenderElementProps` — `resolveStyle(styleProp, state)` then `mergeObjects(outProps.style, style)` ([useRenderElement.tsx:106](packages/react/src/internals/useRenderElement.tsx#L106)). The user style merges with `DialogViewport`'s internal `pointerEvents` style, with user style winning on conflict — consistent with `DialogViewport`'s own behavior.
- **Tests pass.** Ran `pnpm test:jsdom DrawerViewport` — all 50 pass (6 browser-only skipped).
- **Regression test is genuine.** Reverted just the one-line source fix and re-ran: the conformance test `forwards the custom 'style' attribute defined on the component` **fails**. So the new conformance suite catches this exact bug.

### Code quality & conventions
- The fix mirrors the `className={className}` line directly above it — minimal and idiomatic.
- The conformance setup correctly follows the **drawer** convention (`<Drawer.Root open><Drawer.Portal>` — same as [DrawerPopup.test.tsx](packages/react/src/drawer/popup/DrawerPopup.test.tsx) and [DrawerContent.test.tsx](packages/react/src/drawer/content/DrawerContent.test.tsx)), rather than copying the `Dialog.Viewport` pattern (`modal={false}` + nested `<Dialog.Popup />`). Drawers default to `modal`, so this is the right call.
- Commit message follows `[scope] Imperative summary`.

### Minor notes (non-blocking)
- Function-form `style` (`(state) => CSSProperties`) is now forwarded too, and `DialogViewport` resolves it against `DialogViewportState`. `DrawerViewportState` and `DialogViewportState` are structurally identical, and this is the same path the already-forwarded function-form `className` takes — so no issue, just worth being aware the function receives the dialog-typed state.
- The conformance test renders an empty `Drawer.Viewport` (no `Drawer.Popup` child). Conformance only exercises the viewport element itself, so this is fine and all tests pass — adding a popup child would be marginally more realistic but isn't needed.

### Risks
Effectively none. The change is purely additive — it restores a previously-dropped prop. No security, performance, or API-surface concerns.

**Verdict: Looks good to merge.** Small, correct, well-tested bug fix with a verified regression test.

PR #4844, opened 2026-05-18
[otp field] Fix vertical arrow slot navigation

atomiks:
Pressing <kbd>ArrowUp</kbd> or <kbd>ArrowDown</kbd> in an OTP slot could collapse the slot selection, so a later typed character could be blocked by the one-character input length.

## Root cause

Vertical arrows fell through to native text-input caret handling instead of being handled as OTP field navigation.

## Changes

- Map <kbd>ArrowUp</kbd> to the first slot, matching <kbd>Home</kbd>.
- Map <kbd>ArrowDown</kbd> to the last filled slot when the current slot has a character, matching <kbd>End</kbd>.
- Keep focus in place when pressing <kbd>ArrowDown</kbd> on an empty slot.
- Only advance same-character typing when another slot exists.
- Add regression coverage for vertical arrow navigation and final-slot same-character handling.

### DIFF
--- a/packages/react/src/otp-field/input/OTPFieldInput.test.tsx
+++ b/packages/react/src/otp-field/input/OTPFieldInput.test.tsx
@@ -196,6 +196,57 @@ describe('<OTPField.Input />', () => {
     expect(document.activeElement).toBe(inputs[2]);
   });
 
+  it.each([
+    ['ArrowUp', 0],
+    ['ArrowDown', 3],
+  ] as const)('moves focus to the field boundary with %s', async (key, targetIndex) => {
+    await render(<OTPFieldTest defaultValue="1234" />);
+
+    const inputs = screen.getAllByRole<HTMLInputElement>('textbox');
+
+    await act(async () => {
+      inputs[1].focus();
+    });
+
+    expect(fireEvent.keyDown(inputs[1], { key })).toBe(false);
+    expect(inputs[targetIndex]).toHaveFocus();
+  });
+
+  it('keeps focus on an empty slot with ArrowDown', async () => {
+    await render(<OTPFieldTest defaultValue="12" />);
+
+    const inputs = screen.getAllByRole<HTMLInputElement>('textbox');
+
+    await act(async () => {
+      inputs[2].focus();
+    });
+
+    expect(fireEvent.keyDown(inputs[2], { key: 'ArrowDown' })).toBe(false);
+    expect(inputs[2]).toHaveFocus();
+  });
+
+  it('does not reselect the final slot when typing the same character', async () => {
+    const { user } = await render(<OTPFieldTest defaultValue="123456" />);
+
+    const inputs = screen.getAllByRole<HTMLInputElement>('textbox');
+    const lastInput = inputs[5];
+
+    await act(async () => {
+      lastInput.focus();
+    });
+
+    const select = vi.spyOn(lastInput, 'select');
+
+    try {
+      await user.keyboard('6');
+
+      expect(select).not.toHaveBeenCalled();
+      expect(lastInput).toHaveFocus();
+    } finally {
+      select.mockRestore();
+    }
+  });
+
   it('moves focus to the first slot with Home', async () => {
     await render(<OTPFieldTest defaultValue="1234" />);
 
@@ -255,10 +306,10 @@ describe('<OTPField.Input />', () => {
     fireEvent.keyDown(inputs[1], { key: 'ArrowRight' });
     expect(document.activeElement).toBe(inputs[2]);
 
-    fireEvent.keyDown(inputs[2], { key: 'Home' });
+    fireEvent.keyDown(inputs[2], { key: 'ArrowUp' });
     expect(document.activeElement).toBe(inputs[0]);
 
-    fireEvent.keyDown(inputs[0], { key: 'End' });
+    fireEvent.keyDown(inputs[0], { key: 'ArrowDown' });
     expect(document.activeElement).toBe(inputs[3]);
   });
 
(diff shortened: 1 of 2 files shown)


flaviendelangle (approved):
# PR Review Summary — #4844 [otp field] Fix vertical arrow slot navigation

**Author:** atomiks  •  **Base:** `master` → **Head:** `codex/fix-otp-field-selection`  •  **Diff:** +63 / −6 across 2 files

The PR maps `ArrowUp`/`ArrowDown` in `OTPField.Input` to `Home`/`End`-like slot navigation (with `ArrowDown` keeping focus in place on an empty slot), and fixes a latent re-`.select()` bug when the same character is retyped into the final slot.

## Critical Issues

**None blocking.** Both review agents converged — no correctness, security, or merge-blocker concerns.

## Important Issues

- **[tests] Lost Home/End readonly coverage** — `packages/react/src/otp-field/input/OTPFieldInput.test.tsx:306-312`. The existing readonly navigation test was modified to swap `Home`/`End` assertions for `ArrowUp`/`ArrowDown`, removing readonly coverage for the original keys. Either restore the Home/End assertions alongside the new ones, or add a dedicated arrow-key readonly test. A future regression breaking readonly Home/End would now go unnoticed.

- **[tests] No ArrowDown boundary case (already on last filled slot)** — `packages/react/src/otp-field/input/OTPFieldInput.test.tsx:199-211`. The parametrized test only exercises ArrowDown from a non-boundary index (`inputs[1]` of `defaultValue="1234"`, lastFilledIndex=3). The case where current index already equals `lastFilledIndex` is exactly where the new `slotValue !== ''` guard interacts — focus should stay, `keyDown` should still return `false`. Without it, a refactor that flipped the guard polarity could pass silently.

## Suggestions

- **[code-review] `lastFilledIndex` is a misleading name** — `packages/react/src/otp-field/input/OTPFieldInput.tsx:199`. `Math.max(value.length - 1, 0)` returns `0` when `value === ''`, which isn't actually a "filled" index. Current callers happen to be safe (focusing slot 0 when empty is correct; the new `slotValue !== ''` guard on ArrowDown short-circuits before it matters), but the name implies a postcondition the value doesn't guarantee. Consider `endTargetIndex`.

- **[code-review] ArrowDown on an empty slot still calls `stopEvent`** — `packages/react/src/otp-field/input/OTPFieldInput.tsx:218-223`. Functionally fine for `type=text|password` (browsers have no default ArrowDown behavior to suppress), but if the input `type` ever changes, suppressing it silently could mask issues. Either move `stopEvent` inside the conditional, or leave a one-line WHY.

- **[code-review] Modifier-semantics asymmetry vs. ArrowLeft/Right** — `packages/react/src/otp-field/input/OTPFieldInput.tsx:200-212` vs `218-223`. ArrowLeft/Right only jump to boundary with `Ctrl`/`Meta`; ArrowUp/Down jump unconditionally. Not wrong (WAI-ARIA APG doesn't standardize an OTP keyboard model), but worth confirming with the design system that the divergence is intentional — some macOS users expect `Cmd+Arrow` semantics across all four directions.

- **[tests] No coverage for ArrowUp/ArrowDown in `disabled` mode** — the early-return at `OTPFieldInput.tsx:194` now gates the new keys too. A `defaultValue="12" disabled` + ArrowUp test would harden the gate; no existing test covers disabled keyboard behavior.

- **[tests] `stopPropagation` not directly asserted** — `OTPFieldInput.test.tsx:215-220` asserts `preventDefault` via `fireEvent.keyDown(...) === false`. `stopEvent` also calls `stopPropagation`, which the return value doesn't capture. Minor — both calls live in the same helper.

## Strengths

- **The re-`.select()` fix is the standout.** Replacing `focusInput(Math.min(length - 1, index + 1))` with `if (index < length - 1) focusInput(index + 1)` at `packages/react/src/otp-field/input/OTPFieldInput.tsx:258-262` correctly avoids the spurious `.select()` call (root invokes `target.select()` unconditionally) when the user retypes the same character on the last slot. The test at `OTPFieldInput.test.tsx:222-241` uses `vi.spyOn(lastInput, 'select')` — a precise behavioral assertion that would catch any regression.

- **ArrowDown-on-empty-slot semantics match the user mental model.** When the focus is already past the filled region, the natural "end" is the current position. Captured cleanly by the test at `OTPFieldInput.test.tsx:213-221`.

- **Tests follow project conventions.** Uses `fireEvent.keyDown` return-value to assert `preventDefault`, Vitest APIs only (`vi.spyOn`, `vi.fn` — no Chai/Sinon chains), `screen.getAllByRole` with explicit `<HTMLInputElement>` typing, and `act`-wrapped focus changes. Aligns with AGENTS.md testing guidelines.

- **No CLAUDE.md / AGENTS.md violations** — no `as any`, no new hooks (so `useTimeout`/`useStableCallback`/`useIsoLayoutEffect` N/A), no `document`/`window` lookups added.

- **Small, focused diff** that fixes two real bugs (missing ArrowUp/Down navigation + re-`.select()` on the last slot) without scope creep.

## Recommended Action

1. **Address before merge** (low-effort):
   - Restore Home/End readonly coverage at `OTPFieldInput.test.tsx:306-312` (or add a sibling test for ArrowUp/Down in readonly).
   - Add the ArrowDown-on-last-filled-slot boundary test case.

2. **Consider** (cheap polish):
   - Rename `lastFilledIndex` → `endTargetIndex` at `OTPFieldInput.tsx:199`.
   - Confirm with design system that ArrowUp/Down unconditional boundary-jump (vs ArrowLeft/Right requiring Ctrl/Meta) is intentional.

3. **Follow-up** (optional):
   - Disabled-mode coverage for the new keys.

PR #4840, opened 2026-05-18
[direction provider] Fix RTL component behavior

atomiks:
Several direction-sensitive paths still treated left/right or wheel movement as LTR, so RTL keyboard and scroll interactions could stall or move the wrong way.

This also uncovered a bug with Scroll Area where the `wheel` listener was not registered depending on the render timing.

## Root cause

Combobox, Scroll Area, Navigation Menu, and logical positioning coverage did not consistently pass or assert `DirectionProvider` state.

## Changes

- Pass RTL direction into Combobox grid navigation.
- Mirror Combobox chip left/right keyboard behavior in RTL.
- Clamp horizontal Scroll Area wheel movement to the RTL negative `scrollLeft` range.
- Mirror vertical Navigation Menu trigger opening in RTL.
- Add RTL coverage for Popover logical-side positioning and Direction Provider runtime/types.

PR #4847, opened 2026-05-19
[otp field] Fix RTL arrow navigation

atomiks:
Follow-up to #4840. OTP Field still handled slot arrow navigation as LTR, so RTL users moved to the visually wrong slot.

## Root cause

`OTPField.Input` handled left/right keys itself and did not read `DirectionProvider`.

## Changes

- Read Base UI direction in OTP Field inputs.
- Map previous/next arrow keys according to LTR or RTL.
- Add RTL coverage for normal and Ctrl/Cmd boundary arrow navigation.

### DIFF
--- a/packages/react/src/otp-field/input/OTPFieldInput.test.tsx
+++ b/packages/react/src/otp-field/input/OTPFieldInput.test.tsx
@@ -4,6 +4,7 @@ import userEvent from '@testing-library/user-event';
 import { act, fireEvent, screen } from '@mui/internal-test-utils';
 import { OTPFieldPreview as OTPField } from '@base-ui/react/otp-field';
 import { Field } from '@base-ui/react/field';
+import { DirectionProvider } from '@base-ui/react/direction-provider';
 import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
 
 describe('<OTPField.Input />', () => {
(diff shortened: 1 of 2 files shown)


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The patch fixes the OTP RTL arrow-key bug at the right abstraction level and matches both Base UI’s RTL composite convention and native RTL input caret behavior.

### 1. Bugs / Issues (None)

The bug was that `OTPField.Input` owned its own arrow-key handling and always treated `ArrowLeft` as previous and `ArrowRight` as next. Reading `useDirection()` in `packages/react/src/otp-field/input/OTPFieldInput.tsx` and mapping physical keys to logical previous/next fixes the root cause without changing the existing focus math.

This is the right shape: Home/End and Backspace/Delete semantics remain unchanged, while only horizontal navigation is direction-aware.

This matches the repo’s existing RTL pattern for horizontal composites: RTL reverses the previous/next arrow keys. It is also consistent with native RTL text input behavior. I verified in headless Chromium that a native `dir="rtl"` input moves the caret logically forward on `ArrowLeft` and logically backward on `ArrowRight`, which corresponds to this OTP change: `ArrowLeft` moves to the next slot index in RTL, `ArrowRight` moves to the previous slot index.

One caveat, expected for Base UI: this follows `DirectionProvider`, not DOM `dir` auto-detection. That matches the rest of the library’s direction-aware component behavior.

### Test Coverage Assessment
The added RTL tests cover normal arrow movement and Ctrl/Cmd boundary movement, which are the affected branches. Validation run:

- `pnpm test:jsdom OTPFieldInput --no-watch`
- `pnpm test:chromium OTPFieldInput --no-watch`
- `pnpm prettier`
- `pnpm eslint`
- `pnpm typescript`
- `pnpm stylelint`
- `git diff --check`

PR #4849, opened 2026-05-19
[popups] Fix programmatic focus return

### LINKED ISSUE #4169: Accessibility: Focus moves to the wrong trigger when using "Open programmatically" buttons
# Bug report

## Current behavior

In the Alert Dialog, Dialog and Popover components, there is a variant that includes the option to use multiple triggers to open the dialogs. When using the "Open Programmatically" button to open the dialog/popup, then closing it, focus moves to a different control than the one that triggered the dialog/popup.

See:
* [Alert Dialog: Controlled mode with multiple triggers](https://base-ui.com/react/components/alert-dialog#controlled-mode-with-multiple-triggers
* [Dialog: Controlled mode with multiple triggers](https://base-ui.com/react/components/dialog#controlled-mode-with-multiple-triggers)
* [Popover: Controlled mode with multiple triggers](https://base-ui.com/react/components/popover#controlled-mode-with-multiple-triggers)

This could cause failures against [WCAG 2.2 Success Criterion 2.4.3 Focus Order](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html) when implemented.

## Expected behavior

Keyboard focus should move back to the originating control (if available) when closing a dialog as long as a page refresh or redirection is not triggered. 

If the control is no longer available, focus should move next to that originating control's position in the DOM.

## Reproducible example

1. Go to [Alert Dialog: Controlled mode with multiple triggers](https://base-ui.com/react/components/alert-dialog#controlled-mode-with-multiple-triggers
2. Using keyboard, move focus to the "open programmatically" button
3. Press Enter to activate the control and open the dialog
4. Press Esc or activate the close button to close the dialog
5. Note focus has moved back to the "Delete" button and not the "open programmatically" button as expected.

## Base UI version

v1.2.0

Chrome 145.0.7632.76

## Which OS are you using?

Mac OS Tahoe 26.2

Keyboard

## Additional context

Recommend keeping track of focus and returning it to the originating control.

atomiks:
Fixes #4169

Controlled dialogs and popovers can be associated with a triggerId even when a separate control opened them programmatically. This returns focus to that originating control on close instead of always preferring the associated trigger.

## Root cause

The default return-focus path preferred the associated reference trigger, even for programmatic opens where the focused opener was a different control.

## Changes

- Prefer the previously focused opener for programmatic popup opens.
- Keep reference-first focus restoration for trigger-opened popups and disconnected-reference fallback behavior.
- Add coverage for a controlled Dialog triggerId opened from a separate control.


flaviendelangle (approved):
# PR #4849 Review — `[popups] Fix programmatic focus return`

**Verdict: Sound, well-targeted fix. No blocking issues.** The PR diff is exactly 3 files (1 source, 2 tests). The core change correctly makes programmatically-opened controlled dialogs/popovers return focus to the originating control instead of the associated `triggerId` trigger, while preserving reference-first behavior for trigger-opened popups.

## How the fix works (verified)
`preferPreviousFocus = openInteractionTypeRef.current == null` (`FloatingFocusManager.tsx:734`). Only Dialog, Popover, Combobox, and Drawer pass `openInteractionType={openMethod}` — which is `null` on a programmatic open and a non-null `InteractionType` on a trigger interaction. So trigger opens keep reference-first; programmatic opens prefer the previously-focused control. `getNodeName` is already imported and the disconnected/`body` staleness guard mirrors the existing `addPreviouslyFocusedElement` logic. Logic is correct.

## Critical Issues
**None.** (The test-analyzer's reported Critical — Dialog test vacuous in jsdom — is **incorrect**: the test is Chromium-gated; jsdom skips it. Disregard.)

## Important / Worth a decision
1. **Menu and Select are silently excluded from the fix** — `MenuPopup.tsx` / `SelectPopup.tsx` use `FloatingFocusManager` but don't pass `openInteractionType`, so it defaults to `''` (`FloatingFocusManager.tsx:264`) and `'' == null` is `false` → `preferPreviousFocus` is always `false` for them. This is *not a regression* (they keep prior reference-first behavior), and the `== null` check (rather than `!value`) is in fact what scopes the feature to opt-in components. But it's subtle. **Recommendation:** confirm excluding Menu/Select is intentional, and add a one-line comment at line 734 explaining that the `== null` check (vs. falsy) deliberately limits the behavior to components that pass `openMethod`. Otherwise a future reader will "fix" it to `!openInteractionTypeRef.current` and silently change Menu/Select behavior.

2. **No test for the preserved path** — both new tests open via a *separate* programmatic control, exercising only the `preferPreviousFocus === true` branch. A regression that *always* preferred previous focus would pass every new assertion. The broader component suites likely cover trigger-open→return-to-trigger, but a dedicated assertion here (open via the trigger, close, assert focus returns to that trigger) would lock in the preserved behavior.

## Suggestions
- **Disconnected-reference fallback is untested** — the new `referenceReturnElement = null` (disconnected trigger) and `getPreviouslyFocusedElement()` fallback (`FloatingFocusManager.tsx:797-801`) is exactly the kind of edge this code touches, but no test removes the originating control/trigger while open.
- **`openInteractionTypeRef` dep-array entry is a no-op** (`:858`) — `useValueAsRef` returns a stable ref, so it never re-triggers the effect. Behavior is still correct (the trigger sets `openMethod` synchronously before mount). Harmless but slightly misleading; fine to leave.
- **`'html'`/`documentElement` not rejected** by the staleness guard at `:799` (only `'body'`). This matches the existing `addPreviouslyFocusedElement` convention, so it's consistent — purely theoretical.
- **Dead "Close externally" button** in `PopoverRoot.detached-triggers.test.tsx:216` — the test now closes via the in-popup `Popover.Close`, so the renamed external button is never clicked. Either remove it or add an assertion exercising the externally-driven close.

## Strengths
- Minimal, surgical change to a notoriously subtle area (focus restoration).
- Preserves all existing fallback semantics (disconnected reference, outside-press, explicit `returnFocus`).
- New tests are correctly Chromium-gated (focus assertions need real layout/focus) and capture the button reference before close rather than re-querying.
- Fully CLAUDE.md/AGENTS.md compliant (`useIsoLayoutEffect`, `activeElement`, `ownerDocument`, `contains`).

## Recommended actions
1. Decide on Menu/Select scope and add the clarifying comment at line 734 (Important).
2. Optionally add the preserved-path and disconnected-reference tests.
3. Confirm CI Chromium is green (the two new tests are the actual regression coverage; jsdom skips them).

## Reviewer notes
- jsdom run confirmed: 17 of 20 Dialog detached-trigger tests are skipped — the programmatic test at line 179 sits inside `describe.skipIf(isJSDOM)` (line 15), so it is Chromium-only.
- A couple of the sub-agents were misled by `git diff master...HEAD` (local `master` is behind, pulling in unrelated NumberField/menu commits) — those are **not** part of this PR; the canonical `gh pr diff` is clean at 3 files.

PR #4851, opened 2026-05-19
[docs][experiments] Prefix user-select CSS property

aarongarciah:
`-webkit-*` prefix is still needed for Safari, and we don't auto-prefix. The corresponding Tailwind class comes prefixed.


flaviendelangle (commented):
I think it's worth adding a line in the `AGENDS.md` to make sure the future examples follow this rule.
Unless we can have a Stylelint rule for it, but I couldn't find one 😬

aarongarciah:
@flaviendelangle added a note in `AGENTS.md` (https://github.com/mui/base-ui/pull/4851/changes/3097ddc97cf1f970b1b5da33dae7aeda7ef46105)

flaviendelangle (approved):

PR #4856, opened 2026-05-19
[ContextMenu] Use pointer events to support stylus input

### LINKED ISSUE #4854: [context menu] `ContextMenuTrigger` long-press doesn't fire for stylus input
Hi! `ContextMenuTrigger`'s long-press path uses TouchEvents (`onTouchStart`/`Move`/`End`/`Cancel`), so it doesn't fire for stylus input. Apple Pencil on iOS Safari dispatches PointerEvents only, and most Surface pen and Wacom-style tablet setups are pointer-primary too. Would you consider switching to PointerEvents filtered to `pointerType === 'touch' || pointerType === 'pen'`? The 500ms timer, 10px movement tolerance, and cancellation logic would all carry over unchanged for finger touch, and stylus users would get the same long-press UX. 

Happy to send a PR if you'd accept it.

flaviendelangle:
Fixes #4854

@michaldudak if you can test on your setup, and confirm the bug on master :pray: 
It's working fine on my Android touch device without stylus.

### DIFF
--- a/packages/react/src/context-menu/trigger/ContextMenuTrigger.test.tsx
+++ b/packages/react/src/context-menu/trigger/ContextMenuTrigger.test.tsx
@@ -202,15 +202,12 @@ describe('<ContextMenu.Trigger />', () => {
 
       const trigger = screen.getByTestId('trigger');
 
-      const touchObj = new Touch({
-        identifier: 0,
-        target: trigger,
+      fireEvent.pointerDown(trigger, {
+        pointerType: 'touch',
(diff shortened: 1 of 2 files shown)


michaldudak:
This works OK on master on both an iPad and an Android tablet with pens for me.
The PR version also works well.

flaviendelangle:
I'll wait for a proper reproduction case then, since it's a tiny breaking change for people opening context menu with touch on tests :+1:

PR #4866, opened 2026-05-20
[internal] Update the hook value when store/arguments change

chuganzy:
This PR fixes an issue where `useStoreFast` returns a stale value when selector arguments changes during render.
This was not an issue in all test cases and dev mode because of the `StrictMode`, but was causing issues in production build.

ex: MenuTrigger in the Menu hero demo does not update `aria-expanded`.
https://base-ui.com/react/components/menu

### DIFF
--- a/packages/utils/src/store/useStore.ts
+++ b/packages/utils/src/store/useStore.ts
@@ -69,7 +69,6 @@ export type StoreInstance = Instance & {
     a2: unknown;
     a3: unknown;
     value: unknown;
-    didChange: boolean;
   }[];
   didChangeStore: boolean;
   subscribe: (onStoreChange: any) => () => void;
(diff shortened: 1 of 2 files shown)


chuganzy:
@atomiks 
Thanks for labeling! Unfortunately, this might affect all places that utilize `fastComponent`, not only menu. Although I have not confirmed any active issues in them, we might want to also add the "tooltip" and "preview-card" labels just in case.

packages/utils/src/store/useStore.ts
```diff
@@ -173,7 +170,7 @@ function useStoreFast(
       hook.a1 = a1;
       hook.a2 = a2;
       hook.a3 = a3;
-      hook.didChange = true;
+      hook.value = selector(store.getSnapshot(), a1, a2, a3);
```
romgrk, line 173: This was an optimization because we have complex memoized selectors (in the MUI X repository), so avoiding running the selectors is desirable. It's indeed a bug to avoid it here, and the behavior was relying on react updating in response to our `getSnapshot()` (`instance.syncTick`) changing.

I will investigate avoiding the double selector call (here and in `getSnapshot()`) by keeping a copy of the snapshot on the hook, but in a different PR. The gains may not be worth the bookkeeping cost for some/all components.

romgrk (approved):

romgrk:
Thanks for the PR!

PR #4868, opened 2026-05-20
[otp field] Avoid password manager bubbles after first input

atomiks:
Only the first OTP field input needs the full-code max length. Later visible inputs now omit `maxlength`, so password managers do not treat every slot as a separate one-character target while typing still handles multi-character 
changes.

This notably affected iCloud password manager

## Root cause

Each visible slot had a `maxlength`, which could cause password manager UI to appear after later inputs.

## Changes

- Keep the full-code `maxlength` only on the first visible OTP input.
- Assert later visible inputs omit `maxlength` and cover multi-character replacement from a later slot.

### DIFF
--- a/packages/react/src/otp-field/input/OTPFieldInput.tsx
+++ b/packages/react/src/otp-field/input/OTPFieldInput.tsx
@@ -101,8 +101,8 @@ export const OTPFieldInput = React.forwardRef(function OTPFieldInput(
     autoCorrect: 'off',
     spellCheck: 'false',
     enterKeyHint: index === length - 1 ? 'done' : 'next',
-    // Allow the first slot to accept a full code so browser paste/autofill can target it directly.
-    maxLength: index === 0 ? length : 1,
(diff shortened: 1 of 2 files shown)


aarongarciah (commented):
Very nice. The iCloud Password extension popup doesn't appear anymore after the first input. In terms of assistive tech, it looks like we don't lose anything by removing `maxlength="1"`?

before

https://github.com/user-attachments/assets/9ab680a7-e528-4c54-be15-689673c7a153

after

https://github.com/user-attachments/assets/6ca7a8e7-6e4a-4454-97c3-50f0aebb0296

atomiks:
> In terms of assistive tech, it looks like we don't lose anything by removing maxlength="1"?

When I was testing VoiceOver, NVDA, Narrator, etc., they didn't announce anything relating to the length of the input, it's the `aria-label` (and group labeling/`<input>` tag) that are announced. So yeah that concept appears hidden to them currently

aarongarciah (approved):

PR #4869, opened 2026-05-20
[docs-infra] Add suffix to component page titles

aarongarciah:
Adds "React Component" suffix to all component page titles to improve SEO e.g. "Accordion" -> "Accordion React Component"

### DIFF
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -33,7 +33,6 @@ const withMdx = nextMdx({
       [
         transformMarkdownMetadata,
         {
-          titleSuffix: ' · Base UI',
           extractToIndex: {
             include: ['src/app/react'],
             exclude: [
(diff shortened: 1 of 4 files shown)


docs/next.config.mjs
```diff
@@ -33,7 +33,6 @@ const withMdx = nextMdx({
       [
         transformMarkdownMetadata,
         {
-          titleSuffix: ' · Base UI',
```
aarongarciah, line 36: Removing this to rely on Next.js' `metadata.title.template` instead. @dav-is any downsides I might be missing?

aarongarciah: I think this is nicer than:
- Creating a new remark plugin that appends "React Component" to component page titles. This plugin would ran after the `transformMarkdownMetadata` from docs-infra.
- Manually adding the suffix to all component pages (via Next.js' `metadata`), which opts out of the automatic `titleSuffix` from docs-infra.

dav-is: I think this makes sense as an alternative to `titleSuffix` 👍🏼

dav-is (approved):
I think this looks good, is there any context to why this change is helpful?

aarongarciah:
@dav-is we're looking into improving some SEO metrics.

PR #4870, opened 2026-05-20
[drawer][docs] Fix snap points iOS bottom bar paint

atomiks:
The snap-points Drawer demo now paints the iOS 26+ bottom browser bar area immediately on open. On iOS 26.1, a 1px swipe up previously caused Safari to repaint and fill the bottom bar; this makes that fill happen without requiring movement. iOS 26.0 doesn't paint anything and is always bugged.

## Issue

The snap-points demo only painted the bottom bleed through the transformed popup, so iOS Safari could leave the browser bar area unfilled until a tiny drag forced a repaint.

## Changes

- Add a fixed bottom paint strip on the snap-points viewport in both CSS Modules and Tailwind demos.
- Keep the popup above the strip and preserve the real bottom bleed sizing so the existing snap behavior stays the same.


aarongarciah:
_Not related to this PR_

Interesting how the band's color doesn't change when switching between light/dark mode.

https://github.com/user-attachments/assets/b1c541e8-6ceb-49f3-a827-528861ac2fbc

aarongarciah (approved):
Working nicely

PR #4873, opened 2026-05-21
[field] Fix form validation bugs

atomiks:
This is part of the Codex component behavior and test coverage sweep. It fixes Field/Form behavior bugs found during review and adds focused test and adjacent sweep coverage for the affected paths.

## Changes

- Fix Field control name registration and Form values when the name lives on `Field.Control`.
- Fix stale validation values and stale async validation results.
- Apply `validationDebounceTime` consistently to field-aware controls.
- Preserve user `aria-describedby` values when Field adds description or error ids.
- Keep Field-provided `aria-invalid` authoritative when validation marks a control invalid.
- Fix controlled `dirty` required validation and `Form.errors` arrays.
- Reflect `Field.Item disabled` state on the item itself.
- Add test coverage for the failing and adjacent field-aware control paths.

- [P2] `Field.Control name` is ignored by the Form registry. `Field.Root` documents its `name` as taking precedence over `Field.Control`'s `name`, but registration stores only the root name, so `<Field.Root><Field.Control name="email" /></Field.Root>` is omitted from `onFormSubmit` values and `validate(_, formValues)`.
- [P2] `actionsRef.validate()` can validate a stale value for custom validators. The imperative action calls `validation.commit(validityData.value)`, but normal typing only updates dirty/filled until a validation commit happens, so custom `validate(value)` can receive `null` or the previous committed value. Existing coverage only checks native `required`.
- [P2] Async validation has no stale-result guard. Each `commit` awaits `validate` and then unconditionally writes validity state, so a slower validation for an older value can overwrite the newer result. The async test only covers one validation pass.
- [P2] `validationDebounceTime` only works for `Field.Control`. Field-aware controls such as Select, RadioGroup, Checkbox, NumberField, Combobox, Slider, and Switch call `validation.commit()` directly when validating on change, bypassing the debounce path. Existing debounce coverage uses only `Field.Control`.
- [P2] User `aria-describedby` props can drop Field-provided description/error IDs. `Field.Control` merges validation props before `elementProps`, and `mergeProps` overwrites plain props, so consumer `aria-describedby` replaces Field's generated IDs instead of concatenating them. Tests only cover the automatic case.
- [P2] Controlled `dirty` does not affect required-field validation. `dirty={true}` updates public state, but `setDirty` returns before updating `markedDirtyRef`, so `valueMissing` can still be suppressed as if the field were pristine. The dirty prop test only checks style hooks.
- [P2] `Form.errors` arrays are typed but render as raw adjacent strings. Client validation arrays become a `<ul>`, while `Form errors={{ field: ['a', 'b'] }}` flows directly into `children`; tests only cover string server errors.
- [P3] `Field.Item disabled` is not reflected in `Field.Item` state or data attributes. It computes item-level disabled and passes it to wrapped controls, but `useRenderElement` receives the unmodified root state, so render props and `data-disabled` are wrong for the item itself.


flaviendelangle (approved):
# PR #4873 Review — `[field] Fix form validation bugs`

**Scope:** 34 files, +410/−219. A field/form validation bug-fix sweep with one regression test per fix. Reviewed via the full diff (`/tmp/pr4873.diff`) by four specialized agents (general code, test coverage, silent-failure, type design).

**Overall:** Solid PR. The eight documented fixes are all correct, each backed by a real regression test that would fail without its fix. No critical or merge-blocking issues. Findings are mostly type-design footguns and a couple of behavior changes worth a conscious sign-off.

---

## Critical Issues
**None.** All four agents independently traced the core logic (async stale-guard, `change()` debounce, registration refactor, `markedDirtyRef`, `FieldError` arrays) and found it sound.

---

## Important Issues

**1. `aria-invalid` precedence silently flipped for field-aware controls** — `useFieldValidation.ts:756-765` + prop-array reorderings in CheckboxRoot.tsx, RadioGroup.tsx, SliderRoot.tsx, SwitchRoot.tsx, RadioRoot.tsx, NumberFieldInput.tsx.
`validation.getValidationProps` now sits *after* `elementProps`, and `{ 'aria-invalid': true }` now applies after `externalProps`. Previously a consumer's explicit `aria-invalid` won; now the field's invalid state wins. Likely *more* correct, but it's an unannounced behavior change with no test asserting the new precedence. **Recommend:** confirm intentional and add a note to the PR description.

**2. Debounce regression coverage is Checkbox-only** — `FieldRoot.test.tsx:531-558`.
The `commit`→`change` migration touched **8** field-aware controls (Checkbox, CheckboxGroup, Combobox, NumberField, OTP, RadioGroup, Select, Slider, Switch), but only Checkbox is verified to actually debounce. A regression in any other control's `change()` wiring would go uncaught. **Recommend:** one more debounce test on a non-input control with different value semantics (e.g. Select or RadioGroup).

---

## Suggestions

**3. `useRegisterFieldControl` is now 6 positional params** — `useRegisterFieldControl.ts:959-963`. Call sites repeatedly pass `undefined, true, name` (and `value: unknown` masks any misalignment). Highest-density footgun in the PR. **Recommend:** convert to an options object `{ controlRef, id, value, getValue?, enabled?, name? }`.

**4. `useFieldControlRegistration` returns a positional tuple `[validate, register]`** — `useFieldControlRegistration.ts:951`. Both elements are `() => void`-shaped, so swapping the destructure compiles cleanly and breaks both silently. **Recommend:** return a named object `{ validate, register }` (matches `useFieldValidation`'s own convention).

**5. `commit` type/impl divergence** — `useFieldValidation.ts` interface (`commit: (value) => Promise<void>`) vs impl (`(value, revalidate = false)`). The hidden param contradicts the narrowed public type; a future caller reading the impl could re-introduce the duplicated debounce/revalidate policy this PR centralized. **Recommend:** split the revalidate branch into a private named function so "only `change()` may revalidate" is structural, not convention.

**6. `LabelableProvider` splits `aria-describedby` on a single space** — `LabelableProvider.tsx:1014-1022`. `split(' ')` + `Set` leaves empty tokens on leading/trailing/consecutive spaces, emitting a stray space. Cosmetic. **Recommend:** `split(/\s+/).filter(Boolean)`.

**7. `change()` empty-value bypass is string-only** — `useFieldValidation.ts:747`. A text input clearing to `''` skips the debounce for instant feedback, but a checkbox/select returning to `false`/`null`/`[]` waits the full debounce. Minor UX asymmetry, not a bug. **Recommend:** a one-line comment clarifying the `''` bypass is intentionally string-input-only.

**8. Untested edge cases (low):** Field.Control vs Field.Root `name` *conflict* precedence (`name ?? registration.name` — root wins) isn't asserted; the async-guard test only covers the late-old-result direction. Optional adds.

---

## Strengths

- **One real regression test per fix** — each would fail without its change; verified by tracing.
- **Broad aria-describedby coverage** — preserved-user-value tests across 9 component test files.
- **`FieldError` empty-array handling fixes a genuine pre-existing silent bug** — `[]` was truthy and used to suppress real client errors with no message; now correctly falls through.
- **Imperative `validate()` relocation** into the registration hook correctly fixes the stale-value bug by reading live registration/DOM value.
- **Async stale-guard is sound** — no validity state left inconsistent on early return (no `setCustomValidity` runs before the await on the discarded path).
- **AGENTS.md compliant** — `useIsoLayoutEffect`, `useStableCallback`, `useTimeout` all used correctly; no unsafe DOM/global access; fake-timers and `flushMicrotasks` usage follows conventions.

---

## Recommended Action

1. Confirm the `aria-invalid` precedence change (#1) is intentional; note it in the PR.
2. Add a debounce test for a non-Checkbox field-aware control (#2).
3. Consider the two type-design refactors (#3, #4) — both are contained, in-repo, and remove real footguns.
4. The rest (#5–#8) are nice-to-haves.

Nothing here blocks merge; #1 and #2 are the two worth addressing before merging.

PR #4876, opened 2026-05-21
[number-field] Handle unreadable clipboard paste

lyzno1:
Handle clipboard read failures during NumberField paste events without throwing, matching the existing OTPField paste behavior.


atomiks (approved):
Thanks for the fix @lyzno1

PR #4877, opened 2026-05-21
[combobox] Fix chip context error

lyzno1:
Align the Combobox chip context error with Base UI error message conventions.


atomiks (approved):
Thanks for the fix @lyzno1!

PR #4878, opened 2026-05-21
[preview-card] Fix inline trigger repositioning

lyzno1:
Try to fix https://app.circleci.com/jobs/github/mui/base-ui/299088

- Requests a Preview Card position update when inline trigger coordinates change while open.
- Keeps inline rect coordinates in a ref while letting the positioner own the Floating UI update.


lyzno1:
I will close this PR at any time if there are any problems with the proposal.

atomiks:
@lyzno1 yeah I noticed this flaky test too, I actually had a PR already open for it: https://github.com/mui/base-ui/pull/4836

PR #4884, opened 2026-05-22
[tooltip] Preserve focus-opened tooltip on hover leave

michaldudak:
## Summary

- Keep a tooltip opened by focus open when the pointer leaves the focused trigger.
- Continue closing it on blur or Escape.
- Add regression coverage for focus -> mouseLeave -> blur/Escape.

## Bug

A tooltip can be opened by focus and then closed by hover leave while the trigger still has focus. That leaves keyboard users with a tooltip that disappears even though the focus-open condition is still true.

## Reproduction

- Before: https://codesandbox.io/s/yt6z9g?file=/src/App.tsx
- After: https://codesandbox.io/s/lr75l2?file=/src/App.tsx

## Test plan

- `pnpm test:jsdom TooltipRoot --no-watch`
- `pnpm test:chromium TooltipRoot --no-watch`

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingRootStore.ts
+++ b/packages/react/src/floating-ui-react/components/FloatingRootStore.ts
@@ -5,7 +5,7 @@ import { type BaseUIChangeEventDetails } from '../../internals/createBaseUIEvent
 import { createEventEmitter } from '../utils/createEventEmitter';
 import { type FloatingUIOpenChangeDetails } from '../../internals/types';
 import { type PopupTriggerMap } from '../../utils/popups';
-import { isClickLikeEvent } from '../utils';
+import { isClickLikeEvent, isFocusLikeEvent } from '../utils';
 import type { TransitionStatus } from '../../internals/useTransitionStatus';
 
 export interface FloatingRootState {
@@ -97,8 +97,8 @@ export class FloatingRootStore extends ReactStore<
       !newOpen ||
       !this.state.open ||
       // Prevent a pending hover-open from overwriting a click-open event, while allowing
-      // click events to upgrade a hover-open.
-      (event != null && isClickLikeEvent(event))
+      // click or focus events to upgrade a hover-open.
+      (event != null && (isClickLikeEvent(event) || isFocusLikeEvent(event)))
     ) {
       this.context.dataRef.current.openEvent = newOpen ? event : undefined;
     }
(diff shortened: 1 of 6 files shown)


michaldudak:
## Codex Review (GPT-5.5)

Reviewed the full PR diff against freshly fetched `upstream/master`. Dominant type: bug-fix, with regression-test coverage for Tooltip focus/hover ownership.

### 1. Bugs / Issues (None)

I did not find any branch-relevant correctness issues in the changed hover ownership path.

The patch addresses the actual ownership problem: a focus-opened tooltip should not be treated as hover-owned while focus remains on the trigger or popup. Updating `syncOpenEvent` to let focus upgrade a hover-open event, then gating hover-close through `isNonHoverOpenEvent(..., focusOwned)`, keeps the tooltip open for focus ownership without making stale focus events permanently sticky after blur.

### Test Coverage Assessment

The new Tooltip tests cover the important regressions: focus-open -> mouse leave -> blur/Escape, hover-open -> focus -> mouse leave, and pending hover-open not overwriting focus ownership. These run through the existing contained/detached/multiple-detached trigger matrix.

Validated:
`pnpm test:jsdom TooltipRoot --no-watch`
`pnpm test:chromium TooltipRoot --no-watch`
`pnpm test:jsdom useHoverReferenceInteraction --no-watch`
`pnpm test:chromium useHoverReferenceInteraction --no-watch`
`pnpm prettier --check ...`
`pnpm eslint ...`
`pnpm typescript`

I did not separately replay the new regression tests on `upstream/master`.

### Recommendation

Approve ✅

The implementation is narrowly scoped, matches the shared hover-interaction architecture, follows the repo’s DOM utility conventions, and the relevant Tooltip/hook/type/lint validation passes.

atomiks:
I've noticed this in the past, but didn't consider it to be important to fix or possibly even desirable, especially without `:focus-visible` checks.  I think we can leave this off?

michaldudak:
It might be worth discussing. [WCAG](https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html) doesn't address this case directly, but says:

> ### Persistent
> The intent of this condition is to ensure users have adequate time to perceive the additional content after it becomes visible. Users with disabilities may require more time for many reasons, such as to change magnification, move the pointer, or simply to bring the new content into their visual field. Once it appears, the content should remain visible until:
> - The user removes hover or focus from the trigger and the additional content, consistent with the typical user experience;

It would also be consistent with other cases where we consider open modality to determine whether to close something (like open-on-hover Popover persists if opened by click, but closes when open on hover)

michaldudak:
We decided not to implement this.

PR #4886, opened 2026-05-22
[tooltip][preview card] Close when active trigger unmounts

michaldudak:
## Summary

- Clear stale active-trigger state when the active trigger unregisters while the tooltip/preview card is open.
- Close the popup instead of implicitly re-anchoring it to another trigger.
- Cover contained and detached trigger removal.

## Bug

When the active trigger unmounts, unregistering removes the DOM node from the trigger map but leaves the active trigger id/element in store state. Conditional or virtualized triggers can then leave the tooltip open against stale anchor state instead of closing.

### Tooltip

- Before: https://codesandbox.io/s/v8rgxd?file=/src/App.tsx
- After: https://codesandbox.io/s/n87sc5?file=/src/App.tsx

### Preview Card

- Before: https://codesandbox.io/s/zdpz96?file=/src/App.tsx
- After: https://codesandbox.io/s/c65d6g?file=/src/App.tsx

## Test plan

- `pnpm test:jsdom popupStoreUtils --no-watch`
- `pnpm test:chromium TooltipRoot.detached-triggers --no-watch`


michaldudak:
## Codex Review (GPT-5.5)

Dominant type: `bug-fix`. I reviewed the full PR diff against freshly fetched `upstream/master`; the patch addresses stale tooltip active-trigger state when the owning trigger unregisters, with focused coverage for contained and detached trigger setups.

### 1. Bugs / Issues (None)

I did not find any concrete blocking or non-blocking issues in the implementation.

The fix targets the shared active-trigger reconciliation point in `useImplicitActiveTrigger`, which is the right layer for detecting that the active trigger id no longer maps to a registered element. Tooltip opts into closing on that condition, while other popup consumers retain the previous “clear stale active trigger” behavior without being forced closed.

The microtask guard is doing useful work here: it avoids closing if the same trigger remounts before the deferred check runs, and it respects canceled `onOpenChange` events before clearing active-trigger state.

### Test Coverage Assessment

Coverage looks proportionate for this bug. The PR adds utility-level coverage for unregistering the active trigger and Chromium tooltip coverage for both contained and detached triggers, including a canceled close path.

I ran:

```bash
pnpm test:jsdom popupStoreUtils --no-watch
pnpm test:chromium TooltipRoot.detached-triggers --no-watch
pnpm typescript
```

All passed. I did not replay the new regression assertions against `upstream/master`.

### Recommendation

Approve ✅

The change is narrowly scoped, covered at the shared utility and Tooltip integration levels, and the relevant targeted tests plus typecheck pass.

flaviendelangle (commented):
I tried to fix the same think (in a less clean way) for `PreviewCard` the other day 😆 
#4953

If you could also apply your fix there and add some tests 🙏

michaldudak:
> If you could also apply your fix there and add some tests 🙏

Done in [2db0116](https://github.com/mui/base-ui/pull/4886/commits/2db01165153d5e32d9e684e23e247b313bdb25ba)

flaviendelangle (commented):
# PR Review Summary — #4886 "[tooltip][preview card] Close when active trigger unmounts"

Four specialized agents reviewed the PR. Aggregated findings below. The PR adds `closeOnActiveTriggerUnmount` to `useImplicitActiveTrigger` and opts in `TooltipRoot` / `PreviewCardRoot`. Three agents converge on the same handful of soft spots; one (silent-failure-hunter) is meaningfully harsher and worth weighing.

## Critical Issues

**None blocking.** Type-check, eslint, and the affected Vitest suites pass on the PR branch (verified by the code-reviewer agent).

The silent-failure-hunter raised two "Critical" items, but they are debatable rather than blocking — see Important.

## Important Issues

1. **Stale JSDoc on `useImplicitActiveTrigger`** — `popupStoreUtils.ts:~208-216`. The doc still describes only the implicit-promotion behavior. The hook now also (a) reconciles `activeTriggerElement` when the registered element for an id changes, (b) clears stale active-trigger state on unmount, (c) optionally closes via `closeOnActiveTriggerUnmount`. New `options` param is undocumented. Multiple agents independently flagged this.

2. **Microtask deferral is load-bearing but unexplained** — `popupStoreUtils.ts:~273`. The codebase consistently annotates `queueMicrotask` elsewhere (`FloatingFocusManager.tsx:663`, `useCompositeRoot.ts:356`). A future reader will likely "simplify" this to synchronous; one-line comment recommended.

3. **Cancel branch intentionally leaves stale `activeTriggerId`/`activeTriggerElement`** — `popupStoreUtils.ts:~281-286`. Genuinely surprising design — popup stays anchored to a detached DOM node. Two implications:
   - Reads as a bug to a future maintainer; an inline `// intentional` comment is cheap insurance.
   - Silent-failure-hunter argues this should emit a dev `console.warn` per AGENTS.md "actionable feedback" guidance. Worth considering but probably out of scope for this PR — file a follow-up if not addressed here.

4. **Behavior change for Popover/Menu/Dialog (not opted in)** — `popupStoreUtils.ts:~289-294`. The default `else` branch now nulls `activeTriggerId`/`activeTriggerElement` whenever the active trigger element disappears, even when `closeOnActiveTriggerUnmount: false`. All existing tests pass, but this is an undocumented side effect for a PR titled tooltip/preview-card-only. Either:
   - Gate the clear behind the option too, or
   - Add a regression test for "controlled Popover with disappearing active trigger stays open and doesn't crash" and call the change out in the PR description.

5. **Missing test for `triggerCount === 1` after active unmount** — would lock in the "don't briefly swap to the other trigger before closing" ordering. The component tests have two triggers but don't assert `activeTriggerId` doesn't transiently equal `trigger-2`.

## Suggestions

- **Add `setOpen` `toHaveBeenCalledTimes(1)`** to `popupStoreUtils.test.tsx:286-289` to catch double-fire regressions.
- **Microtask cleanup**: no explicit cancellation if `Root` unmounts between schedule and fire. In practice the in-microtask guards (`open`, `activeTriggerId` match, element absent) make it a no-op, so low priority unless a concrete repro appears.
- **No detached + canceled combo test** for Tooltip/PreviewCard — the contained variant covers the same hook path, so low value.
- **`as Partial<State>` casts** at three sites (lines 268, 285, 293) — would catch typos if `stateUpdates` were typed as `Partial<State>` from the start.
- **`REASONS.none` as the close reason** — silent-failure-hunter suggests a dedicated `REASONS.activeTriggerUnmount` for debuggability. Reasonable but not blocking; this is a project-wide convention question.

## Strengths

- `queueMicrotask` correctly avoids "update during render" and re-checks `open` / `activeTriggerId` / element presence inside the microtask — robust against fast remount.
- `BaseUIChangeEventDetails<typeof REASONS.none>` typing is contravariant-compatible with all existing consumers; `pnpm typescript` clean.
- Cancel-then-keep-anchored test (`PreviewCardRoot.detached-triggers.test.tsx:326-389`) explicitly locks the intentional behavior.
- New unit test at `popupStoreUtils.test.tsx:252-292` isolates the lost-trigger flow at the right level.
- `!lostActiveTriggerId && !activeTriggerId && triggerCount === 1` correctly suppresses implicit promotion when the active trigger has just been lost — prevents a flicker where the popup briefly re-anchors before closing.
- Conventions followed: `useIsoLayoutEffect`, no `setTimeout`, `queueMicrotask` matches existing repo usage. Tests use `describe.skipIf(isJSDOM)`, `await render`, `await act`, no `flushMicrotasks` directly after `render`.

## Recommended Action

1. **Update the JSDoc on `useImplicitActiveTrigger`** to document the new `options` arg and the hook's expanded responsibilities (Important #1).
2. **Add a one-line comment above the `queueMicrotask`** explaining why deferral is needed (Important #2).
3. **Add a one-line comment on the `if (!eventDetails.isCanceled)` branch** noting that stale active-trigger state on cancel is intentional (Important #3).
4. **Decide on Important #4**: either narrow the default-branch clear behind the new option, or document the behavior change for Popover/Menu/Dialog and add the suggested regression test.
5. The other findings are nice-to-have follow-ups.

PR #4888, opened 2026-05-22
[docs-infra] Update docs to support turbopack mode

brijeshb42:
### TL;DR

- **Dev startup**: ~3× faster (1.0s → 0.35s)
- **Cold-route compile**: ~1.4× faster on average (5.87s → 4.26s across 4 routes); per-route range from 0.8× to 2.0×. Still dominated by `loadPrecomputedTypes` (1–2s per types.ts) in both modes. We need to improve this on the docs-infra package side.
- **HMR rebuilds**: ~2.8× faster (1.77s → 0.63s mean across 3 edit targets)
- **Production build**: 2.3× faster (61.5s → 27.0s), 2.7× less CPU

1. Dev server startup (cold, after `rm -rf .next`)

| | webpack | Turbopack |
|---|---|---|
| `Ready in` | 1.32–1.51s | **0.34–0.53s** |

~3× faster ready time.

2. First-request compile per route (no cache)

| route | webpack | Turbopack |
|---|---|---|
| `/` | 1.9s | **1.4s** |
| `/react/components/popover` | 6.1s* | **5.1s** |
| `/react/components/dialog` | 4.2s | **2.0s** |
| `/react/components/select` | 2.5s | 3.7s |
| `/react/components/combobox` | 3.5s | **2.7s** |

3. HMR / incremental recompile (3 edit cycles × 2 add+revert)

| file edited | webpack | Turbopack | speedup |
|---|---|---|---|
| `popover/page.mdx` | 1.86s | **0.48s** | 3.9× |
| `PopoverPopup.tsx` (workspace pkg) | 2.50s | **0.95s** | 2.6× |
| docs `CodeBlock.tsx` | 0.95s | **0.45s** | 2.1× |
| **mean** | **1.77s** | **0.63s** | **2.8×** |

4. Production build (cold, `rm -rf .next export`)

| | wall | user CPU |
|---|---|---|
| webpack | 61.5s | 128.3s |
| Turbopack | **27.0s** | **47.5s** |
| ratio | **2.3×** faster | 2.7× less CPU |

@dav-is When building the docs, I got a bunch of -

```
Unable to handle a type with flag "IncludesInstantiable". Using any instead.
```

Also, I need you to check if everything is working as far as docs-infra related transforms are concerned since you have larger context. I have already verified that a bunch of pages are working as expected.

### DIFF
--- a/docs/src/components/CodeBlock/rehypeEagerCodeBlocks.mjs
+++ b/docs/src/components/CodeBlock/rehypeEagerCodeBlocks.mjs
@@ -2,7 +2,7 @@ import { visit, EXIT, CONTINUE } from 'unist-util-visit';
 
 const EAGER_COUNT = 2;
 
-export function rehypeEagerCodeBlocks() {
+export default function rehypeEagerCodeBlocks() {
   return (tree) => {
     let count = 0;
     visit(tree, 'element', (node) => {
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -111,7 +111,6 @@
     "#test-utils": "./test/index.ts",
     "#formatErrorMessage": "@base-ui/utils/formatErrorMessage"
   },
-  "type": "commonjs",
   "scripts": {
     "prebuild": "rimraf --glob build build-tests \"*.tsbuildinfo\"",
     "build": "code-infra build --ignore \"**/*.template.js\" --copy .npmignore",
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -154,6 +154,7 @@ const nextConfig = {
   devIndicators: false,
   experimental: {
     globalNotFound: true,
+    turbopackFileSystemCacheForBuild: true,
   },
 };
 
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -3,35 +3,25 @@ import * as path from 'path';
 import * as url from 'url';
 import * as fs from 'fs';
 import { withDeploymentConfig } from '@mui/internal-docs-infra/withDocsInfra';
-import transformMarkdownMetadata from '@mui/internal-docs-infra/pipeline/transformMarkdownMetadata';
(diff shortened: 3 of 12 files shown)


docs/src/components/CodeBlock/rehypeEagerCodeBlocks.mjs
```diff
@@ -2,7 +2,7 @@ import { visit, EXIT, CONTINUE } from 'unist-util-visit';
 
 const EAGER_COUNT = 2;
 
-export function rehypeEagerCodeBlocks() {
```
brijeshb42, line 5: Default exports are read through string based plugin paths.

packages/react/package.json
```diff
@@ -111,7 +111,6 @@
     "#test-utils": "./test/index.ts",
     "#formatErrorMessage": "@base-ui/utils/formatErrorMessage"
   },
-  "type": "commonjs",
```
brijeshb42, line 114: Explicit `type` to `commonjs` leads turbopack to fail with error -

```txt
./packages/utils/src/warn.ts
Specified module format (CommonJs) is not matching the module format of the source code (EcmaScript Modules)
The CommonJs module format was specified in the package.json that is affecting this source file or by using an special extension, but Ecmascript import/export syntax is used in the source code.
The module was automatically converted to an EcmaScript module, but that is in conflict with the specified module format. Either change the "type" field in the package.json or replace EcmaScript import/export syntax with CommonJs syntas in the source file.
```

So we can either omit the field (as in PR) or add a nested package.json inside `src` stating `type: module`.

dav-is:
We should enable `turbopackFileSystemCacheForBuild`:

https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache

aarongarciah (approved):
Seems to work fine. The page compilation time in dev is noticeably faster.

PR #4890, opened 2026-05-22
[fieldset] Fix disabled fieldset form bugs

atomiks:
This is part of the Codex component behavior and test coverage sweep. It fixes Fieldset/Form disabled-state bugs found during review and adds focused test coverage for the affected paths.

Disabled field-aware controls now follow native HTML disabled semantics more broadly: whether disabled directly, through `Field.Root`, or through `Fieldset.Root`, they are skipped by Base UI `Form` validation and omitted from `onFormSubmit` values.

## Changes

- Preserve native `disabled` semantics on `Fieldset.Root`, including nested fieldsets and render-prop compositions.
- Stop disabled fieldset and individually disabled field-aware controls from registering with Base UI `Form` validation and submission.
- Make the `Fieldset.Legend` missing-root guard reachable.
- Add coverage for disabled fieldsets, field-aware controls, Form submit values, and Legend root usage.

- [P1] `<Fieldset.Root disabled>` drops the native disabled attribute.
- [P1] Fieldset-disabled Field controls still participate in Base UI Form validation and submission.
- [P2] `disabled` does not flow through documented Fieldset render compositions.
- [P2] Nested Base UI fieldsets can shadow an outer disabled fieldset.
- [P3] The missing-root guard for `Fieldset.Legend` is dead code.


atomiks:
Approve ✅: the current branch fixes the Fieldset/Form disabled-state bugs without introducing actionable issues, and the broader disabled-control behavior is now documented in the PR body.

### 1. Bugs / Issues (None)

No concrete blocking or non-blocking issues found in the full diff against refreshed `upstream/master`.

The fix is targeted at the right ownership points. `Fieldset.Root` now preserves native `disabled` semantics, nested fieldsets inherit disabled state through context, and `Field.Root` consumes the Fieldset context optionally so fields outside a fieldset keep working.

Changing `FieldsetRootContext` to default to `undefined` also makes the `Fieldset.Legend` missing-root guard reachable instead of silently using a no-op context. The remaining non-optional consumer is the intended throwing path.

The Form registration changes are consistent across the affected field-aware controls: disabled controls stop registering with the Base UI Form validation map, which excludes them from validation and `onFormSubmit` values in line with native disabled-control behavior.

All touched `useRegisterFieldControl` call sites now gate registration on the resolved disabled state, while preserving the existing group-specific conditions such as `!groupContext` and `!!fieldName`. The resolved disabled state flows through the existing Field/Fieldset ownership chain, so the implementation stays centralized instead of adding Form-specific filtering later in submission.

### Opus Notes

Opus called out that the behavior change is broader than fieldsets: individually disabled field-aware controls are also excluded from Form validation and submitted values. That is native-aligned and is now called out in the PR body, so I do not see a remaining documentation issue.

Opus also noted the synchronous `Fieldset.Legend` throw test differs from an async `.rejects.toThrow()` pattern used elsewhere. The test passes in the targeted suite and is asserting the render-phase error directly, so I would not block on changing it.

### Test Coverage Assessment

Validation run on the current PR head:

- `pnpm test:jsdom Fieldset --no-watch`
- `pnpm test:jsdom Form --no-watch`
- `pnpm typescript`

The focused coverage exercises native Fieldset disabled semantics, nested fieldsets, render-prop disabled propagation, Form submit values, and the Legend missing-root guard. I do not see a remaining coverage gap that needs to block this PR.

PR #4892, opened 2026-05-22
[menu] Fix submenu trigger interactions

atomiks:
This is part of the Codex component behavior and test coverage sweep. It fixes Menu behavior bugs found during review and adds focused test coverage for the affected paths.

## Changes

- Register `Menu.SubmenuTrigger` labels for typeahead.
- Respect `highlightItemOnHover={false}` for submenu triggers.
- Stop controlled-open disabled menus from accepting typeahead, item clicks, and submenu trigger interaction.
- Run `Menu.SubmenuTrigger` conformance against the actual submenu trigger element.
- Add coverage for submenu trigger typeahead, hover highlighting, disabled controlled-open menus, and conformance.

- [P2] `Menu.SubmenuTrigger` ignores its `label` prop for typeahead.
- [P2] `highlightItemOnHover={false}` is bypassed for submenu triggers.
- [P2] Controlled-open disabled menus still accept item/typeahead interaction.
- [P2] `Menu.SubmenuTrigger` conformance is effectively untested.


atomiks:
Approve after nits are addressed 🟡 The refreshed full diff against `upstream/master` fixes the menu interaction regressions at the right layers and the affected jsdom/Chromium suites pass; only a small test-name cleanup remains.

Opus 4.7 supplied the static-review notes synthesized here; GPT-5.5 refreshed the diff and ran the validation suites before posting.

### 1. Bugs / Issues

#### 1. ℹ️ Clarify the hover regression test name (informational)

`packages/react/src/menu/root/MenuRoot.test.tsx`

```tsx
it('does not highlight submenu triggers from mouse enter when disabled', async () => {
  await render(
    <TestMenuContents
      rootProps={{ open: true, highlightItemOnHover: false }}
```

This test is in the `highlightItemOnHover` block and exercises `highlightItemOnHover={false}`, not the `disabled` prop. The assertion is correct, but the name points future readers at the wrong behavior.

**Recommendation:** Rename the test to mention `highlightItemOnHover={false}` rather than `disabled`.

The fixes line up with the actual failure points:

- `Menu.SubmenuTrigger` now passes `label` into `useCompositeListItem({ label })`, which lets typeahead use the explicit label while preserving the existing text-content fallback when `label` is undefined.
- `highlightItemOnHover={false}` is now synced into the menu store and checked by the submenu trigger's extra `onMouseEnter -> setActive()` path, closing the gap that regular items already avoided through `useListNavigation({ focusItemOnHover })`.
- Controlled-open disabled menus are made inert across the relevant interaction paths: typeahead is gated with `enabled: !disabled`, regular items inherit `store.disabled` through `useMenuItem`, and submenu triggers also include the parent menu's disabled state.
- The conformance test now renders `Menu.SubmenuTrigger` itself and expects the correct `HTMLDivElement` shape instead of accidentally testing `Menu.Trigger`.

The root-disabled behavior follows the existing menu model: `MenuStore` already centralizes root and menubar-disabled state, and the PR extends that state to item activation instead of only trigger/open behavior. Reading `highlightItemOnHover` with `select()` inside the event handler is also the right shape because it needs the latest store value at event time without adding another reactive subscription.

### Test Coverage Assessment

The added coverage maps directly to the fixed regressions: submenu trigger typeahead labels, hover highlighting opt-out, disabled controlled-open item/typeahead behavior, disabled submenu triggers, and real submenu trigger conformance.

Validated with:

```bash
pnpm test:jsdom MenuSubmenuTrigger --no-watch
pnpm test:jsdom MenuRoot --no-watch
pnpm test:chromium MenuSubmenuTrigger --no-watch
pnpm test:chromium MenuRoot --no-watch
```

PR #4893, opened 2026-05-22
[popups] Fix controlled hover leave close

atomiks:
Fixes https://github.com/mui/base-ui/pull/4815#pullrequestreview-4336715913

Hover-close now only runs when the current open state came from trigger hover. Eventless opens, including external controlled opens and default-open popups, no longer close just because the pointer leaves the floating element.

## Root cause

Floating `mouseleave` closed any non-click-like open, so an eventless open was treated like a trigger-hover open.

## Changes

- Gate floating `mouseleave` safe-polygon and close handling on the existing hover open event.
- Keep trigger-hover opens closing normally, including controlled popups opened by their own trigger.
- Cover Menu, Popover, and PreviewCard eventless opens plus trigger-hover controlled opens.

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
+++ b/packages/react/src/floating-ui-react/hooks/useHoverFloatingInteraction.ts
@@ -224,12 +224,14 @@ export function useHoverFloatingInteraction(
 
       // If the safePolygon handler is active, let it handle the close logic.
       if (instance.handler) {
-        instance.handler(event);
+        if (isHoverOpen()) {
+          instance.handler(event);
+        }
         return;
       }
 
       clearPointerEvents();
-      if (!isClickLikeOpenEvent()) {
+      if (isHoverOpen() && !isClickLikeOpenEvent()) {
         closeWithDelay(event);
       }
     }
@@ -262,6 +264,7 @@ export function useHoverFloatingInteraction(
     dataRef,
     closeDelayProp,
     nodeIdProp,
+    isHoverOpen,
     isClickLikeOpenEvent,
     clearPointerEvents,
     instance,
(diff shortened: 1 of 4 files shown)


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The updated patch preserves safePolygon cleanup while fixing non-hover-owned popup leave behavior, and the relevant regression/type checks now pass on the rebased branch.

The current implementation keeps the important distinction: plain floating `mouseleave` only hover-closes when the current open is hover-owned, while an already-installed safePolygon handler still runs unconditionally so it can close or clean up its document `mousemove` listener. That avoids the stale-handler regression from the earlier guarded version.

### Test Coverage Assessment
Coverage now exercises the main bug across Menu, Popover, and PreviewCard, plus the Popover safePolygon cleanup path after a hover-opened popup becomes click-sticky. The externally-opened Popover regression was verified to fail on `upstream/master`; the safePolygon cleanup test is a defensive follow-up for the intermediate guarded-handler regression.

Validation run:

```text
pnpm typescript
pnpm code-infra validate-built-types
pnpm test:jsdom PopoverRoot --no-watch -t "cleans up the safe polygon handler after a hover-opened popup becomes click-sticky|does not close after hovering out of a popup opened externally|does not close after hovering out of a popup opened without trigger hover|closes after hovering out of a popup opened by its trigger"
pnpm test:jsdom PreviewCardRoot --no-watch -t "does not close after hovering out of a popup opened externally|does not close after hovering out of a popup opened without trigger hover|closes after hovering out of a popup opened by its trigger"
pnpm test:jsdom MenuRoot --no-watch -t "does not close after hovering out of a popup opened externally|does not close after hovering out of a popup opened without trigger hover|closes after hovering out of a popup opened by its trigger"
```

PR #4897, opened 2026-05-23
[select] Fix mouse selection in StrictMode dev

### LINKED ISSUE #4698: [select] Mouse selection is a no-op in dev StrictMode

uixmat:
## Summary
- Re-sync `CompositeList` `elementsRef` from `sortedMap` on each layout effect run so list navigation can resolve live DOM nodes after StrictMode cleanup wipes refs without re-running item ref callbacks
- Fixes mouse selection in `Select` (and other composite list consumers) in React StrictMode dev environments such as Next.js dev

Fixes https://github.com/mui/base-ui/issues/4698

## Test plan
- [x] `pnpm test:jsdom CompositeList --no-watch`
- [x] `pnpm test:jsdom SelectRoot --no-watch`
- [x] Manual verification in docs dev: open select, click item, value commits

### DIFF
--- a/packages/react/src/internals/composite/list/CompositeList.test.tsx
+++ b/packages/react/src/internals/composite/list/CompositeList.test.tsx
@@ -50,7 +50,7 @@ describe('<CompositeList />', () => {
         const [showExtraItem, setShowExtraItem] = React.useState(false);
 
         return (
-          <>
+          <React.Fragment>
             <button type="button" onClick={() => setShowExtraItem(true)}>
               Add item
             </button>
@@ -60,7 +60,7 @@ describe('<CompositeList />', () => {
               <Item id="c" />
               {showExtraItem ? <Item id="d" /> : null}
             </CompositeList>
-          </>
+          </React.Fragment>
         );
       }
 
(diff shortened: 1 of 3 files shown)


atomiks:
@uixmat thanks for the fix —  mouse selection without highlight was already fixed in https://github.com/mui/base-ui/pull/4699, so the tests should prove that `[data-highlighted]` is correctly set in Strict Mode. I have a feeling it may not be reproducible in our test env though.

Note this also adds an extra O(n) scan just to fix a rare dev mode problem, ideally it would be cheaper.

uixmat:
> @uixmat thanks for the fix — mouse selection without highlight was already fixed in #4699, so the tests should prove that `[data-highlighted]` is correctly set in Strict Mode. I have a feeling it may not be reproducible in our test env though.
> 
> Note this also adds an extra O(n) scan just to fix a rare dev mode problem, ideally it would be cheaper.

No problem, go-ahead and close the PR if you like, i'll hunt down some other issues!

PR #4900, opened 2026-05-24
[utils] Fix UA-CH browser detection

lyzno1:
This fixes `detectBrowser` so `navigator.userAgentData.brands` is no longer converted into a synthetic user-agent string before browser checks.

UA Client Hints expose structured brand data, so Edge detection now checks the `Microsoft Edge` brand directly and keeps the legacy `navigator.userAgent` fallback for browsers without UA-CH support.

References:
- https://developer.chrome.com/docs/privacy-security/user-agent-client-hints
- https://learn.microsoft.com/en-us/microsoft-edge/web-platform/user-agent-guidance
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent

Verification:
- Added focused tests for UA-CH brands and legacy Edge UA fallback
- TypeScript and lint pass for the touched utils files

### DIFF
--- a/packages/utils/src/detectBrowser.test.ts
+++ b/packages/utils/src/detectBrowser.test.ts
@@ -1,0 +1,96 @@
+import { expect, vi } from 'vitest';
+
+const originalNavigatorDescriptors = new Map<string, PropertyDescriptor | undefined>();
+
+function setNavigatorProperty(name: string, value: unknown) {
+  if (!originalNavigatorDescriptors.has(name)) {
+    originalNavigatorDescriptors.set(name, Object.getOwnPropertyDescriptor(navigator, name));
+  }
+
+  Object.defineProperty(navigator, name, {
+    configurable: true,
+    value,
+  });
+}
+
+async function importDetectBrowser() {
+  vi.resetModules();
+  return import('./detectBrowser');
+}
+
+afterEach(() => {
+  vi.resetModules();
+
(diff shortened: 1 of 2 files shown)


romgrk:
What's the motivation for this change? As much as I enjoy getting rid of legacy APIs, if it's working, I wouldn't touch it. The UA string is not going away.

lyzno1:
> What's the motivation for this change? As much as I enjoy getting rid of legacy APIs, if it's working, I wouldn't touch it. The UA string is not going away. 

That’s fair.

The existing behavior works for current mainstream UA values. The motivation here wasn’t to remove the legacy UA fallback — this still keeps `navigator.userAgent` for browsers without UA-CH support.

The narrow issue I was trying to address is that `userAgentData.brands` is structured data, but the current helper turns it into a synthetic UA string and then runs the existing browser regexes against it. That can make UA-CH GREASE/arbitrary brands participate in checks that were originally written for the legacy UA string.

That said, I agree this is more of a correctness hardening than a user-visible bug. If you’d rather avoid touching this while the current behavior works, I’m happy to close the PR.

romgrk:
Thanks but I wouldn't merge it. I see that Edge (and other chromium browsers) have some older versions without support for UA-CH, just on the off chance someone uses such browser, I'd keep the UA check as it is.

PR #4903, opened 2026-05-25
[tabs] Fix suspended panel activation

### LINKED ISSUE #4880: [tabs] Tabs.Panel content not rendered if a child component suspends without boundary within Tabs.Root
# Bug report

## Current behavior

If a tab panel has a child that suspends and the suspense unmounts the tab root (i.e., the next boundary is outside the `Tabs.Root` component), the panel content is not rendered when opening a tab for the first time.

Only when clicking the trigger a second time, the panel is rendered.

## Expected behavior

The panel content is rendered also when first opening a tab like it did in v1.4.1.

## Reproducible example
Click on tab B in this
[Stackblitz](https://stackblitz.com/edit/vitejs-vite-ddcpzhwd?file=src%2Fmain.tsx)

Alternatively, run this code snippet locally:
```
import { Tabs } from '@base-ui/react/tabs';
import { Suspense, use } from 'react';

// Promise used to suspend.
const dataPromise = new Promise<string>((resolve) =>
  setTimeout(() => resolve('Panel B is rendered'), 500)
);

function SuspendingChild() {
  return <div>{use(dataPromise)}</div>;
}

  return (
    // The Suspense boundary is outside the <Tabs.Root> and the panels/panel content don't use a separate suspense boundary.
    <Suspense fallback={<div>Loading…</div>}>
      <Tabs.Root defaultValue="a">
        <Tabs.List>
          <Tabs.Tab value="a">Tab A</Tabs.Tab>
          <Tabs.Tab value="b">Tab B</Tabs.Tab>
        </Tabs.List>
        <Tabs.Panel value="a">Panel A (no suspense)</Tabs.Panel>
        <Tabs.Panel value="b">
          <SuspendingChild />
        </Tabs.Panel>
      </Tabs.Root>
    </Suspense>
  );
}

```

## Base UI version

The only affected version is v1.5.0

v1.4.1 is not affected

Reproducible in Chrome/Safari/Firefox

## Additional context

Potentially a side effect of https://github.com/mui/base-ui/pull/4704

atomiks:
Fixes #4880

Opening a tab whose panel suspends no longer makes Tabs treat the temporarily empty tab registry as real tab removal.

## Root cause

Suspense can clean up tab registration effects while keeping the previous tab nodes connected, so the automatic missing-tab fallback reset the value before the suspended panel could resume. The connectivity check has to run from the root effect after the DOM settles; during a real unmount, unregister can run before React detaches the DOM node.

## Changes

- Track a previously registered tab element and only commit the missing-tab fallback when that element is disconnected.
- Add a React 19 Suspense regression test for opening a suspending panel with the boundary outside `Tabs.Root`.


atomiks:
## Code Review (GPT-5.5 + Opus 4.8)

Approve ✅: Both reviews found the fix correctly scoped and validated by a regression test that fails on master and passes with the PR.

### 1. Bugs / Issues (None)

No blocking or non-blocking defects found.

### Combined Assessment

This fixes #4880, where a `Tabs.Panel` child suspending with the `Suspense` boundary outside `Tabs.Root` left the panel blank until a second click.

The root cause is that Suspense can clean up tab registration effects while keeping the DOM nodes connected. `TabsRoot` then saw `tabMap.size === 0` and incorrectly treated it as real tab removal, committing the `missing` fallback and resetting the value.

The PR’s fix is minimal and well-placed: `TabsRoot` tracks a previously registered tab node and only commits the missing-tab fallback when that node is actually disconnected.

### Verification

Validated across both reviews:

- The new Suspense regression test fails on current `upstream/master`.
- The same test passes on the PR.
- The failure matches issue #4880: `Panel B` never renders after the first activation.
- `pnpm test:chromium TabsPanel --no-watch` passes on the PR.
- `pnpm test:jsdom TabsPanel --no-watch` passes on the PR.
- `pnpm test:chromium TabsRoot --no-watch` passes on the PR.

Opus 4.8 additionally reported:

- The test passes with the fix and fails when only the source fix is removed.
- ESLint and Prettier passed for the changed files.
- Typecheck showed only unrelated pre-existing TS6305 build-cache noise, with no tabs-related type errors.

### Opus 4.8 Notes

Opus also called out a few optional considerations:

- Tracking the first tab node is fine because any surviving tab works as the connectivity probe, but a short comment could make that clearer.
- A theoretical reparenting case could keep the probe connected, though this is not a realistic Tabs concern.
- Other composite components still share the cleanup-as-removal assumption, but Tabs is the one with missing-value auto-fallback behavior, so this scope is appropriate.

flaviendelangle (approved):
Branch restored. Here is the synthesized review.

# PR #4903 Review Summary — [tabs] Fix suspended panel activation

Four specialized agents reviewed the PR (code quality, test coverage, comments, silent failures). Across all four, **no critical or blocking issues were found**. The fix is correct, minimal, and well-scoped. Below is the consolidated action plan.

## Critical Issues
**None.** All four agents independently concluded the PR is ready to merge.

## Important Issues
**None.** The previously-existing GitHub review comment (GPT-5.5 + Opus 4.8) had already cleared the fix; this deeper pass confirms it.

## Suggestions (nice-to-have)

### Source code (`TabsRoot.tsx`)
- **Add a comment on the probe-selection line** at `TabsRoot.tsx:299` — `lastKnownTabElementRef.current = tabMap.keys().next().value;`. This is the load-bearing line; the "any surviving tab works as the probe" rationale isn't self-evident. A future maintainer could "improve" it to `getTabElementBySelectedValue(value)` and break the fix, because during a Suspense reveal the selected tab may not yet be in `tabMap`.
- **Tighten the comment at `TabsRoot.tsx:286-287`** to mention the `isConnected` mechanism explicitly (e.g. *"Use a previously-registered tab as a liveness probe: if it is still connected, the empty tabMap is a Suspense effect-teardown artifact, not a real removal"*). Current wording "previous tabs connected" is ambiguous (could be read as *prior selection*).
- Consider broadening "Suspense boundary outside the root" to "an ancestor Suspense boundary" — the same code path triggers wherever a Suspense ancestor is above the registered tabs.
- Optional: clear `lastKnownTabElementRef.current = undefined` after `commitAutomaticValueChange(null, REASONS.missing)` to drop the strong reference to a now-detached `Node`. Not a real leak (single node), but tidies up.

### Tests (`TabsPanel.test.tsx`)
- **Drop `await Promise.resolve();` at line 71** — empirically verified the test still passes both jsdom and chromium without it. `act`'s async-callback resolution already drains microtasks, and `findByText` retries. The test analyzer ran this 5× without flake.
- Consider adding **`aria-selected` assertions** after `findByText('Panel B')` (lines 74-77) to catch a future regression where panel content renders by accident while value state is wrong.
- Consider a follow-up test for **initial-mount suspension** (`defaultValue="b"` with a Tab B panel that suspends on first render). This locks in the `didRegisterTabsRef.current` guard, which is a different code path than the click-to-suspend case the new test covers.
- The test arguably belongs in `TabsRoot.test.tsx` (the fix lives in TabsRoot's effect) per the project's file-per-source-file convention, but this is purely cosmetic.

## Strengths
- **Empirically validated**: the new test fails on master, passes on the PR, and mutation tests confirm each guard in the rewritten branch is covered by *some* existing test.
- **Pre-existing real-removal coverage** at `TabsRoot.test.tsx:880` already locks in the `!isConnected` path — no coverage gap there.
- **Suspense vs unmount discrimination is sound**: when a sibling suspends, React tears down effects but leaves DOM connected; when tabs actually unmount, the host node is detached. `isConnected` is the correct minimal probe and is already shadow-DOM-safe by spec (consistent with `CompositeList.tsx:58`'s same usage).
- **No CLAUDE.md/AGENTS.md violations**: uses `useIsoLayoutEffect`, no `as any`, no global `document`/`window`, no shadow-DOM-unsafe traversal.
- **The "lonely tab removal" edge case can't fire** — the new condition is inside `if (tabMap.size === 0)`, so by definition no tab survives; if even one did, the `else` branch reassigns the probe ref.
- **`shouldNotifyInitialValueChangeRef` concern is unfounded** — by the time the new Suspense-skip branch can fire (requires `didRegisterTabsRef.current === true`), the initial-notify ref has already been cleared in a prior commit.

## Recommended Action
1. **Merge as-is is defensible** — none of the items above are blocking.
2. If the author has a spare minute: add the comment on `TabsRoot.tsx:299` (load-bearing line, biggest maintainability win) and drop the redundant `await Promise.resolve();` in the test.
3. The suggested follow-up tests (initial-mount suspension, re-suspension) could be a separate small PR if desired but aren't required.

PR #4905, opened 2026-05-25
[number field] Fix committed values and keyboard stepping

atomiks:
Fixes a few Number Field commit paths that could report a different value than the one shown or stored.

## Changes

- Use the clamped blur value for commit callbacks, display text, and on-blur field validation.
- Let Alt+ArrowUp/ArrowDown use smallStep instead of bypassing keyboard stepping.
- Snap before clamping so non-step-aligned min/max boundaries remain reachable.
- Update smallStep API docs to say the alt key.


flaviendelangle (approved):
# PR #4905 Review Summary — [number field] Fix committed values and keyboard stepping

**Verdict:** Solid, well-scoped fix. No blocking issues. The three behavioral changes are correct, the comments are accurate, and tests pass (validate, NumberFieldInput, NumberFieldRoot suites all green). The main opportunities are an architectural fragility worth addressing and several test-coverage gaps in the *mirror* directions of each fix.

---

## Critical Issues
None.

## Important Issues

**1. Implicit cross-file contract behind `lastChangedValueRef.current ?? committed`** — `NumberFieldInput.tsx:195-201`
The blur handler reads the clamped value back through a shared ref that `setValue` writes only when `shouldFireChange` is true. Today it's provably safe — on every blur path that calls `setValue`, `isInputReason && unvalidatedValue !== value` forces `shouldFireChange`, so the ref is always fresh and `?? committed` never fires. But the correctness depends on the exact composition of `shouldFireChange` in a *different* file (`NumberFieldRoot.tsx:248-253`). If a future edit lets an `inputBlur` early-out before the ref write (e.g. an added guard, or the `isCanceled` short-circuit moving earlier), the handler would silently commit/validate the *pre-clamp* value with no error.
*Recommendation:* have `setValue` return the validated value (it already returns `shouldFireChange`) and consume that directly, instead of reaching into the ref — or at minimum document the dependency at the ref write site.

**2. Min/underflow direction untested for both clamping fixes** — `NumberFieldRoot.test.tsx`, `validate.test.ts:255`
The blur-clamp test and the snap-before-clamp test both only cover the **max/overflow** side. The mirror cases are the highest-value missing tests:
- Blur with `min={0}`, input `-50` → expect committed/displayed `0`.
- `toValidatedNumber` with raw `-13`, `step: -3`, `min: -10` → expect `-10` (the directional-negative counterpart of the exact bug being fixed).

**3. Alt+Arrow → `smallStep` → nearest-mode snapping not covered end-to-end** — `NumberFieldInput.test.tsx:125`
Alt uniquely routes through `small: true` (`NumberFieldRoot.tsx:235`), which switches `toValidatedNumber` to `'nearest'` mode and changes base selection. The new Alt test runs with `snapOnStep` defaulting to false, so the genuinely-new nearest-snap path is never exercised via the keyboard. A test with `snapOnStep` enabled + a non-aligned start (e.g. `0.15`, Alt+ArrowUp) would prove the full chain.

## Suggestions
- **No-update blur path** (`NumberFieldInput.tsx:189`): when `value === committed`, `committedValue` stays `committed` and is never read back. Add a test asserting committed-value correctness when already at the clamped boundary.
- **Formatted canonical text after clamp**: tests only check default integer formatting; add a `format={{ minimumFractionDigits: 2 }}`, `max={10}`, input `1000` → displays `10.00` case (`NumberFieldInput.tsx:204-208`).
- **Regression guard for Alt-bypass**: no test asserts Alt+ArrowLeft/ArrowRight (caret nav) or Alt+letter still bypass. Since `isStepKey` is narrowly Arrow Up/Down, a negative test would lock the contract against a future widening.
- **`snapOnStep=false` + clamp**: the `if (shouldClamp)` now runs unconditionally outside the snap block — worth a pin test that out-of-range still clamps with `snapOnStep: false`.
- **Optional comment expansion** at `NumberFieldInput.tsx:191`: note the `?? committed` fallback handles the no-change case.

## Strengths
- **Genuine doc-rot fix**: `smallStep` JSDoc and the docs table said "meta key", but `getStepAmount` keys off `event.altKey` (`NumberFieldRoot.tsx:201-209`). "alt key" is now correct and consistent with the neighboring `largeStep`/"shift key" entry. JSDoc and auto-generated table stay in sync.
- **Snap-before-clamp is a real correctness improvement**: old code could return out-of-range snapped values and made non-step-aligned boundaries unreachable (max=10, step=3 snapped down to 9). All three comments are accurate; the `snapToStep` param rename `clampedValue`→`value` correctly reflects that the arg is no longer pre-clamped.
- **No regressions in the reorder**: `step === 0`, division-by-zero, NaN/Infinity, and the `small` base selection (De Morgan-equivalent) all verified safe.
- **Alt+Arrow change has no side effects**: Alt+Left/Right nav and Alt+digit still bypass; `stopEvent` + `incrementValue` still fire correctly.
- **CLAUDE.md compliance clean**: no banned globals, no `as any`, stable callbacks preserved, Vitest-only matchers.
- New tests are correctly placed, JSDOM-runnable (so they actually execute in the fast suite), and the `validationMode=onBlur` test verifies the value flows all the way to `Field.Root`'s `validate`.

## Recommended Action
1. Consider Important #1 (return validated value from `setValue`) — low effort, removes a silent-failure class.
2. Add the min-direction tests (Important #2) and the Alt+snapOnStep test (Important #3).
3. Suggestions are nice-to-haves; safe to merge without them if the team prefers.

PR #4907, opened 2026-05-25
[dropzone] Add Dropzone component

mbrookes:
# Base UI Dropzone

## Background

This component was originally developed as a personal project, but seemed like a useful addition to Base UI, so submitting here for consideration. Colm had input on the API, and I've reviewed the code, but it was developed with agent assistance, therefore all errors and omissions are mine.

## Overview

Dropzone is an unstyled drop target primitive that provides both drag-and-drop and click-to-select file picking in a single composable component. It is intentionally scoped to the drop/selection surface only — it does not manage file state, upload logic, or progress — making it composable with whatever file handling the consuming application needs.

## API
```jsx
 import { Dropzone } from '@base-ui/react/dropzone';

 <Dropzone.Root onFilesDrop={(files) => console.log(files)}>
   <Dropzone.HiddenInput />
   Drop files here, or click to browse
 </Dropzone.Root>
```

## Components

 - `Dropzone.Root` — the interactive drop target `<div role="button">`. Handles drag events, click-to-open, and keyboard activation.
 - `Dropzone.HiddenInput` — a visually hidden `<input type="file">` that powers the native file picker and enables form participation when a name prop is provided.

## Key features

 - Drag-and-drop with data-dragging state attribute for styling
 - Click and keyboard (Enter/Space) activation of the native file picker
 - Uncontrolled dragging state (`onDraggingChange`)
 - Disabled support with `data-disabled` attribute
 - Render function children receive `isDragging` for inline conditional rendering
 - Built-in aria-live region announces drag enter, drag leave, and drop events to screen readers
 - Nested interactive elements (buttons, links) inside the dropzone do not accidentally trigger the file picker
 - `suppressHydrationWarning` on the hidden input to avoid SSR mismatches

## Documentation

 - Component page with usage guidelines, anatomy, and labeling examples: `react/components/dropzone/page.mdx`
 - Hero demo (CSS Modules and Tailwind CSS versions): `/react/components/dropzone/demos/hero/`

## Comparison with other implementations

Most existing dropzone libraries (react-dropzone, react-files, etc.) are monolithic hooks or components that bundle file validation, preview, upload queuing, and state management together. Dropzone follows the Base UI philosophy of doing one thing: providing the interactive drop target surface. Consumers own file handling.

### Notable differences

 - No bundled file management — just gives you the `File[]` array
 - Composable — works standalone or wrapped with other components
 - Fully styleable — completely unstyled, with presence-based data attributes (`data-dragging`, `data-disabled`) rather than `className` injection
 - render prop — like all Base UI components, the root element can be replaced via the render prop

## Architecture

Follows standard Base UI compound component conventions:

 - `useRenderElement` for the render prop / className / style / state-attribute contract
 - `useStableCallback` for all event handlers to avoid stale closures and unnecessary re-renders
 - Context (`DropzoneRootContext`) to connect `HiddenInput` to `Root` without prop drilling — `HiddenInput` registers its ref into Root so Root can imperatively call `.click()` on it
 - Uses `contains` from the `floating-ui-react` utils for correct drag-leave detection across child elements (avoids false drag-end when cursor moves over children)
 - State attributes use `stateAttributesMapping` + `DropzoneRootDataAttributes` enum, consistent with the rest of the library
 - `visuallyHidden`/`visuallyHiddenInput` utilities for the hidden input
 
 ## Deploy preview
 
 https://deploy-preview-4907--base-ui.netlify.app/react/components/dropzone


mbrookes:
Converted to draft until the Copilot review fixes are complete

michaldudak:
We're currently investigating how to handle drag and drop library-wide. We're considering an external library or an in-house solution. We can revisit this PR when we have a solution.

mbrookes:
Yes, the thinking was that a fully fledged DnD library would be overkill for this use-case, and will bring with it a lot of non-tree-shakeable baggage. 

Let's see when that lands, then we can evaluate with vs without.

PR #4909, opened 2026-05-25
[website] Add search to home page

aarongarciah:
Preview: https://deploy-preview-4909--base-ui.netlify.app/

Adds search to the home page. Changes included:

- Adds a search trigger on the home page, and remove the "Components" link.
- Refactors the Search component to support different triggers.
- Lazy loads the search dialog.
- Removes search dialog's exit animation. It's snappier, and better.

Note: the PR is not meant as a full refactor of the Search component. It's quite weird currently with the `containedScroll` and `enableKeyboardShortcut` props (which are used to have another instance for mobile), but it works and the mobile UX might change in the future, so better to not spend more time on it.

---

before
Capture-2026-05-26-220740

after
Capture-2026-05-26-220806


colmtuite (approved):

aarongarciah:
Removed the exit animation.

PR #4910, opened 2026-05-26
Update reselect to 5.2.0 in base-ui/utils

IzumiSy:

Currently `@base-ui/utils` depends reselect 5.1.1 which does not have provenonce due to manual release workflow, but the latest (v5.2.0) has the proper provenonce. Without it, users would manually have to set `reselect@5.1.1` in `trustPolisyExclude` in using pnpm v11. 

I would like to remove the entry by updating `@base-ui/utils` side to have the latest reselect package that seems a bit more secured.


IzumiSy:
@zannager Thanks for labels. How do I get review on this PR?

flaviendelangle (approved):
The bump make total sense 👍

flaviendelangle:
@mui/infra we just noticed that you mentionned Reselect 5.1.1 in https://github.com/mui/base-ui/pull/4929
Is everything good bumping it now?

brijeshb42:
Yes. `5.2.0` has provenance. So good to bump.

PR #4920, opened 2026-05-26
[utils] Improve platform detection

romgrk:
### Summary

The previous `detectBrowser.ts` exported a flat list of booleans (`isWebKit`, `isFirefox`, `isMac`, `isIOS`, `isAndroid`, `isJSDOM`) that conflated **OS**, **rendering engine**, and **environment** into one undifferentiated namespace. Several call sites had picked the wrong flag because the names suggest a single coherent category that doesn't actually exist:

- **"Browser" is imprecise.** All iOS browsers (Chrome, Firefox, Edge on iOS) use WKWebView, so they're WebKit at the engine level regardless of the user-facing brand. A check that's really about "Safari macOS focus quirk" should not also fire for GNOME Web on Linux; a check that's really about "VoiceOver virtual cursor" should not also fire for WebKitGTK + Orca.
- **`isFirefox` was a brand, not an engine.** Firefox-on-iOS is WebKit (its UA marker is `FxiOS/`), so checks like "Firefox text-direction bug" need to target the Gecko engine specifically, not the Firefox brand.
- **OS vs. form factor were tangled.** `isIOS` was being used both as a true OS check (iOS-only software keyboard behavior) and as a "touch is primary input" proxy — those are different things and call for different signals.
- **`isWebKit` was overloaded.** Some sites meant "WebKit rendering bug" (CSS, layout), others meant "VoiceOver accessibility quirk" (Apple platform + WebKit). Conflating them made one or the other site wrong on platforms like WebKitGTK.

The rule going forward: **a quirk patch should name the thing it's actually patching** — an OS behavior, an engine bug, or an accessibility-tree quirk. Never "this browser".

Scope is intentionally limited to traits that don't change at runtime. Dynamic capabilities like input modality (pointer/touch/hover) belong in a separate feature-detection layer because they can change mid-session (e.g. plugging in a mouse) and will be handled in a follow-up.

### Changes

Old (`@base-ui/utils/detectBrowser`) — flat booleans:

```ts
import { isWebKit, isMac, isIOS, isAndroid, isFirefox, isJSDOM } from '@base-ui/utils/detectBrowser';
```

New (`@base-ui/utils/platform`) — single grouped object:

```ts
import { platform } from '@base-ui/utils/platform';

platform.os.mac
platform.os.ios
platform.os.android
platform.os.windows
platform.os.linux
platform.os.apple          // mac || ios

platform.engine.webkit     // Safari, all iOS browsers, GNOME Web (NOT Blink)
platform.engine.blink      // Chrome, Edge, Opera, Brave, …
platform.engine.gecko      // Firefox (desktop / Android — Firefox iOS is WebKit)

platform.screenReader.voiceOver  // Apple OS + WebKit (the AX path with the VoiceOver virtual-cursor quirk)

platform.env.jsdom         // jsdom or HappyDOM (tests)
```

### Migration

| Old | New |
| --- | --- |
| `isWebKit` (engine bug) | `platform.engine.webkit` |
| `isWebKit` (VoiceOver workaround) | `platform.screenReader.voiceOver` |
| `isFirefox` | `platform.engine.gecko` |
| `isMac` | `platform.os.mac` |
| `isIOS` | `platform.os.ios` |
| `isAndroid` | `platform.os.android` |
| `isJSDOM` (source code) | `platform.env.jsdom` |
| `isJSDOM` (tests) | `import { isJSDOM } from '#test-utils'` |

### Notable specifics

- **`FocusGuard`** previously gated its `role="button"` workaround on `isWebKit`, which was over-broad — on WebKitGTK + Orca the workaround was setting a meaningless role on a hidden span. Now gated on `platform.screenReader.voiceOver`, which is `(os.mac || os.ios) && engine.webkit` — the only place where VoiceOver's NSAccessibility virtual-cursor path is active.
- **`useDismiss.ts`** was importing a *function-form* `isWebKit()` from `@floating-ui/utils/dom` while the rest of the codebase used the boolean from `detectBrowser`. Unified on `platform.engine.webkit`.
- **Engines are mutually exclusive by construction.** `gecko` and `blink` are both anchored to `!webkit`, so Firefox-on-iOS (`FxiOS/`, WebKit) and Chrome-on-iOS (`CriOS/`, WebKit) classify correctly as WebKit-not-Gecko/Blink regardless of future UA-string drift.
- **SSR defaults.** Every flag resolves to `false` when `navigator` is undefined.
- **`isJSDOM` now also matches HappyDOM**, matching the equivalent helper in `@mui/x-internals/platform`.

### Breaking change

`@base-ui/utils/detectBrowser` is removed. Consumers must import from `@base-ui/utils/platform` and update field references per the migration table above. `vi.mock('@base-ui/utils/platform', …)` users must mock the nested `platform` object shape rather than individual booleans.


docs/src/components/Demo/Demo.tsx
```diff
@@ -161,7 +161,7 @@ export function Demo({
 
   const [fallbackToCodeSandbox, setFallbackToCodeSandbox] = React.useState(false);
   React.useEffect(() => {
-    if (isSafari || isEdge) {
+    if (platform.engine.webkit) {
```
romgrk, line 164: Check if Edge works with stackblitz.

packages/react/src/floating-ui-react/hooks/useFocus.ts
```diff
@@ -20,7 +20,7 @@ import { createChangeEventDetails } from '../../internals/createBaseUIEventDetai
 import { REASONS } from '../../internals/reasons';
 import { FloatingUIOpenChangeDetails } from '../../internals/types';
 
-const isMacSafari = isMac && isSafari;
+const isMacSafari = platform.os.mac && platform.engine.webkit;
```
romgrk, line 23: Flag name could potentially be inaccurate.

oliviertassinari:
Most use of `platform` are usually considered a code smell since they don't age well; so we should rely on feature detection whenever possible. So, could we systematically add a comment on why we can't use feature detection in each case?

romgrk:
I think we need both concerns:
 - Workarounds for bugs/quirks (OS, engine, screen reader)
 - Feature detection

But yes feature detection should be the default, and we should be *very* precise about what we're detecting, a flag like `isMobile` is going to be wrong in some cases (either logically or semantically) because it conflates many things (pointer type, screen size, touch support, keyboard).

I've added more `pointer` categories, because many combinations of `hover: boolean, touch: boolean, coarse: boolean` are possible, the most confusing ones:
 - a mobile device can use a pen pointer: `hover: false, touch: true, coarse: false`
 - a laptop can have a touch screen: `hover: true, touch: true, coarse: false`
 - a device with a remote pointer (wii) is: `hover: true, touch: false, coarse: true`
And some devices are pretty fluid (e.g. iPad can be used as a tablet or as a desktop, in which case it's going to pretend to be a mac).

atomiks:
My main concern is the bundle size increase here, especially for a change not critical to the lib. I've been trying to reduce the overall size of the library but size increases keep returning over time and I'd rather not implement a noticeable increase for just a tidy-up.

Moving from individual booleans to one exported `platform` object makes every consumer pull in all detections, even when a call site only needs one flag like `platform.engine.webkit`. That also includes the new `pointer.*` checks, which are not used anywhere in this PR yet.

A better shape would be a namespace export that remains treeshakable. The grouped object reads nicely, but it turns the utility into an all-or-nothing import, which feels too expensive for this kind of helper.

romgrk:
It's not just a tidy-up, I've seen some issues across our repositories where a flag/check would produce incorrect results for specific environments, it's also a correctness change.

> That also includes the new pointer.* checks, which are not used anywhere in this PR yet.

Note that this module (like all `@base-ui/utils`) is meant for use in all our products. That part covers checks that aren't in this repository. Other codebases use partial copies or equivalents of `detectBrowser.ts`, I'm aiming to centralize & standardize platform detection across all of them.

> My main concern is the bundle size increase

So you'd suggest breaking the object into its direct fields, which would be exported directly? I also considered that, although I felt like readability was a bit less optimal. Also, `os` collides with the `node:os` module which is usually imported as `os`. I guess it would be fairly rare for those two modules to be used in the same file though. Please let me know what you prefer.

romgrk:
I've added a few commits to improve bundle-size, one of them being avoiding the `brands` in prod. IIUC, that was meant to avoid a devtools warning so it's not relevant to keep it in the bundle in prod.

From a quick search, it seems that bundlers can tree-shake `import * as x` imports, so we could keep optimal readability & bundle size by using this pattern, although `import *` is not something we usually do in our codebases. Thoughts?

```js
import * as platform from '@base-ui/utils/platform'
// ...
if (platform.os.ios) { /* ... */ }
```

aarongarciah:
+1 in favor of using `import * as platform` instead of an object. AI (and humans) will use whatever is used in the code base (we can also add something to `AGENTS.md`), so I don't think it's a problem.

atomiks:
@romgrk you can check `reason-parts.ts` and `reasons.ts`. They group them into a namespace export so you can just do `import { platform } from '...'` and it stays treeshakeable, like our components do. But this doesn't treeshake the object parts; ideally you'd namespace the exports like `import { engine }` so you don't have two levels of nesting.

romgrk:
I have applied the suggestions as well as went a bit overboard with bundle-size micro-optimization by avoiding duplicate inline strings and abusing regexes. The combined result is that PR now decreases bundle-size.

packages/utils/src/platform/os.ts
```diff
@@ -0,0 +1,23 @@
+import { lowerUserAgent, lowerPlatform, maxTouchPoints } from './shared';
+
+// iPadOS 13+ reports `MacIntel` for `navigator.platform`; disambiguated via
+// `maxTouchPoints` so iPad is classified as iOS, not macOS.
+/** iPhone, iPad (including iPadOS 13+ reporting as macOS), iPod. */
+export const ios =
+  /^i(os$|p)/.test(lowerPlatform) || (lowerPlatform === 'macintel' && maxTouchPoints > 1);
```
romgrk, line 8: The second check (`(lowerPlatform === 'macintel' && maxTouchPoints > 1)` will become incorrect as Apple plans to ship a mac latop with a touchscreen. This needs either strenghtening or abandoning.

If `iOS` wants to pretend to be a `macOS`, we should probably respect that. iPads can be used as a laptop with keyboard & mouse plugged.

Thoughts on dropping the second check?

docs/src/app/(private)/experiments/scroll-lock.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 import * as React from 'react';
 import { useScrollLock } from '@base-ui/utils/useScrollLock';
-import { isWebKit } from '@base-ui/utils/detectBrowser';
+import * as platform from '@base-ui/utils/platform';
```
atomiks: PR body says `import { platform }`?

Seems like this should be added:

```js
// platform/parts.ts
export * as os from './os';
export * as engine from './engine';
export * as pointer from './pointer';
export * as screenReader from './screen-reader';
export * as env from './env';

// platform/index.ts
export * as platform from './parts';
```

Then consumers can do:

```js
import { platform } from '@base-ui/utils/platform';

platform.os.mac;
platform.engine.webkit;
```

packages/utils/src/platform/screen-reader.ts
```diff
@@ -0,0 +1,17 @@
...
+// VoiceOver moves browser focus as its virtual cursor moves and only triggers
+// `onFocus` on focusable/role-button elements via WebKit's NSAccessibility
+// path. That path only exists on Apple platforms running WebKit — Chrome on
+// macOS uses Blink and has a different AX integration.
+/**
+ * The user *may* be using VoiceOver — actual activation is not detectable.
+ * True on Apple platforms running WebKit, where VoiceOver's virtual-cursor /
+ * NSAccessibility focus quirks apply.
+ */
+export const voiceOver = apple && webkit;
```
atomiks: Chrome with VoiceOver also moves focus on virtual cursor moves

romgrk: Right, so the flag should actually be `voiceOver = apple`, right? And then the check should be for `voiceOver && (webkit || blink)`? And to be sure, this also means that voiceover+gecko doesn't have that issue?

packages/react/src/utils/FocusGuard.tsx
```diff
@@ -14,10 +14,11 @@ export const FocusGuard = React.forwardRef(function FocusGuard(
...
   useIsoLayoutEffect(() => {
-    if (isSafari) {
-      // Unlike other screen readers such as NVDA and JAWS, the virtual cursor
-      // on VoiceOver does trigger the onFocus event, so we can use the focus
-      // trap element. On Safari, only buttons trigger the onFocus event.
+    // Unlike NVDA and JAWS, VoiceOver's virtual cursor triggers `onFocus` as
+    // it moves — but only on focusable/role-button elements, and only on
+    // engines that route AX through macOS NSAccessibility (WebKit and Blink).
+    // Firefox on macOS uses Gecko's own AX integration and is excluded.
+    if (platform.screenReader.voiceOver && (platform.engine.webkit || platform.engine.blink)) {
```
atomiks: I think this should just be:

```suggestion
if (platform.engine.webkit) {
```

It's about Safari (macOS or iOS) only triggering onFocus when `role="button"` is present. Chrome/FF on Mac always move it without that attribute.

romgrk: The problem with that check is that it would also trigger for browsers that use webkit without using voiceOver, for example Gnome Web on linux (the stock browser for Gnome). Which we don't officially support, but there's no reason not to target our quirk fixes to the specific platforms where they're needed.

From your comment, I'm also not sure what you mean by "it's about safari": as I wrote in the top-comment, although only safari uses webkit on macOS, all browsers use webkit on iOS (including chrome, firefox, edge), so using browser names to target quirks is imprecise.

I'm unsure how to proceed, please add more details and/or answer the questions in [the comment above](https://github.com/mui/base-ui/pull/4920#discussion_r3331332668).

atomiks: > all browsers use webkit on iOS (including chrome, firefox, edge), so using browser names to target quirks is imprecise

I personally view "webkit" as meaning Mac Safari or iOS (any browser), regardless of how precise that is 

> The problem with that check is that it would also trigger for browsers that use webkit without using voiceOver, for example Gnome Web on linux (the stock browser for Gnome)

`platform.os.apple && platform.engine.webkit` should work then no? False negatives are a worse tradeoff than false positives here regardless, so it shouldn't matter that it fails on such a niche platform/browser combo

romgrk: > I personally view "webkit" as meaning Mac Safari or iOS (any browser), regardless of how precise that is

Right but you used "safari", not "webkit". Also equating `webkit = apple` is false, as that engine can be used outside of apple. I'm not advocating that we start testing on niche platforms, but there's no reason to have inaccurate flags.

So IIUC, this would be correct?
```
platform.screenReader.voiceOver && platform.engine.webkit
```
I'm using the screen reader flag for semantical correctness, not just for logic correctness.

atomiks: Given `voiceOver = apple`, yes that looks correct; same as before with additional edge case handled

docs/src/components/Demo/Demo.tsx
```diff
@@ -161,7 +161,7 @@ export function Demo({
...
       setFallbackToCodeSandbox(true);
     }
```
atomiks, line 166: Not sure why Stackblitz would break edge if it works on Chrome? Not sure if it's a Windows-only thing or macOS either, but I don't have Edge installed

romgrk: See https://mui-org.slack.com/archives/C02P87NQLJC/p1779833012842919

atomiks (approved):
## PR #4920 Claude Review — `[utils] Improve platform detection`

## Semantic equivalence (old → new)

| Call site | Old | New | Equivalent? |
|---|---|---|---|
| [Demo.tsx:163](docs/src/components/Demo/Demo.tsx:163) | `isSafari \|\| isEdge` | `engine.webkit` | **No — Edge dropped** |
| [useFocus.ts:23](packages/react/src/floating-ui-react/hooks/useFocus.ts:23) | `isMac && isSafari` | `os.mac && engine.webkit` | Yes |
| [FocusGuard.tsx:21](packages/react/src/utils/FocusGuard.tsx:21) | `isSafari` | `voiceOver && engine.webkit` (`= (mac\|\|ios) && webkit`) | Yes (more precise) |
| [ComboboxInput.tsx:94](packages/react/src/combobox/input/ComboboxInput.tsx:94) | `isFirefox && isRTL` | `engine.gecko && isRTL` | Yes |
| [NumberFieldScrubArea.tsx:186](packages/react/src/number-field/scrub-area/NumberFieldScrubArea.tsx:186) | `isFirefox` | `engine.gecko` | Yes |
| `isWebKit` sites (scrub area, scroll area, select, focus mgr, scroll lock) | `isWebKit` | `engine.webkit` | Yes (CSS check byte-identical) |
| `isAndroid` / `isIOS` / `isMac` sites | flat flag | `os.android/ios/mac` | Yes |
| `isJSDOM` | `ua.includes('jsdom/')` | `env.jsdom` `/jsdom\|happydom/` | Yes (intentionally broader) |

## Important

**1. ~~Edge loses the CodeSandbox fallback~~ — [Demo.tsx:163] (seems like a non-issue now) (docs/src/components/Demo/Demo.tsx:163)**
Old `isSafari || isEdge` fired for Chromium Edge (`/Edg/` UA, a **Blink** browser). New `platform.engine.webkit` is `false` on Edge, so Edge users now get the **StackBlitz** button instead of the CodeSandbox fallback. The `3e168a0fa refactor: delete isEdge` commit suggests this is deliberate, and it's docs-only/low-impact — but the original Edge branch implies someone once hit an Edge+StackBlitz problem. **Confirm modern Edge renders demos on StackBlitz before merging**, otherwise restore an Edge check.

**2. PR description misdescribes `screenReader.voiceOver` (two reviewers flagged independently)**
The description's API doc and migration notes claim `screenReader.voiceOver` is `"Apple OS + WebKit"` / `(os.mac || os.ios) && engine.webkit`. The implementation ([screen-reader.ts:12](packages/utils/src/platform/screen-reader.ts:12)) is just `voiceOver = apple` (OS-only) — the `&& engine.webkit` lives only at the [FocusGuard call site](packages/react/src/utils/FocusGuard.tsx:21). The in-code JSDoc is correct; **the PR description is the inaccurate artifact.** A future consumer trusting the description and writing `if (platform.screenReader.voiceOver)` alone gets `true` on macOS Chrome/Firefox — exactly the over-broad behavior this PR sets out to eliminate. Also, the migration table row labels FocusGuard's source as `isWebKit` when it was actually `isSafari`, omits the `&& engine.webkit` term, and has no row for the removed `isSafari` flag. Fix the description (or make `voiceOver` actually include the engine term it advertises).

## Minor

**3. Speculative unused API surface — contradicts lean-implementation preference**
Zero consumers anywhere in `packages/` or `docs/` for: the entire [pointer.ts](packages/utils/src/platform/pointer.ts) module (`hover`/`touch`/`coarse`), `os.windows`, `os.linux`, and `engine.blink` ([os.ts:17-20](packages/utils/src/platform/os.ts:17), [engine.ts:22](packages/utils/src/platform/engine.ts:22)). Tree-shaking means no bundle cost, but each ships untested correctness risk (e.g. `os.linux`'s `chrome os` branch, `pointer.touch`'s `compat/compat` lint-disable) with no caller. Recommend dropping them and re-adding when a real call site appears.

**4. `pointer.touch` doc overstates the check** — [pointer.ts:25](packages/utils/src/platform/pointer.ts:25) only probes `window.Touch`, but the PR description says `Touch`/`TouchEvent`. (Also a stray duplicated `//` comment block sits above the JSDoc at [pointer.ts:14-23](packages/utils/src/platform/pointer.ts:14).)

**5. `os.mac` theoretical divergence** — old `isMac` required `!maxTouchPoints`; new `os.mac = !ios && startsWith('mac')` would return `true` on a Mac reporting `maxTouchPoints === 1` (old: `false`). Real Macs report `0`, so this is academic — worth a line in the rationale.

**6. New `platform` detection logic has no unit tests** — all component tests *mock* `platform`, so the actual derivations (the `/^i(os$|p)/` iOS regex, iPad `macintel && maxTouchPoints > 1` disambiguation, `!webkit`/`!ios`/`!android` mutual-exclusion anchoring) never execute under test. The old `detectBrowser` was also untested, so it's not a regression — but a small table-driven test feeding synthetic `userAgent`/`platform`/`maxTouchPoints` would protect this low-level utility cheaply.

**7. Test-mock cross-module staleness (latent, harmless)** — mocks set `os.apple: true/false` ([ComboboxStatus.iOS.test.tsx:15](packages/react/src/combobox/status/ComboboxStatus.iOS.test.tsx:15), [ContextMenuRoot.test.tsx](packages/react/src/context-menu/root/ContextMenuRoot.test.tsx)), but `screenReader.voiceOver` captured `apple` at module load and won't track the override. No current test depends on it; flag only so a future FocusGuard/VoiceOver test mocks `screenReader` directly. Relatedly, [NavigationMenuRoot.webkit.test.tsx](packages/react/src/navigation-menu/root/NavigationMenuRoot.webkit.test.tsx) — the `engine.webkit: true` mock is inert (the exercised path never reads it); pre-existing, preserved faithfully.

## Verified non-issues (the detection-regression suspects all cleared)

- **Firefox `engine.gecko` is NOT broken by the new `!webkit` anchor.** Firefox ships only unprefixed `backdrop-filter`, so `CSS.supports('-webkit-backdrop-filter:none')` is `false` → `webkit=false` → `gecko=true`. ComboboxInput RTL Home/End and NumberFieldScrubArea pointer-lock release are safe.
- **FocusGuard `(mac||ios) && webkit` exactly reproduces old `isSafari`** (`vendor=Apple` covers macOS Safari + all iOS browsers, all WebKit; excludes desktop Chrome/Firefox).
- **SSR-safe:** every flag `false` except `pointer.hover=true`, matching the documented contract.
- **`os.android` (exact-match vs substring), `os.ios` regex, and dev-vs-prod `userAgentData` paths** don't flip any used flag for real browsers/devices.
- **`engine.webkit` CSS check is byte-for-byte identical to the deleted code** — no new risk there.

**Recommended actions:** (1) confirm Edge/StackBlitz then keep or restore the Edge branch; (2) fix the PR description's `voiceOver` claim + migration table; (3) trim unused `pointer.*`/`os.windows`/`os.linux`/`engine.blink`. Nothing here blocks merge on correctness grounds — the detection migration itself is faithful.

packages/utils/src/platform/pointer.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Device can dispatch touch events (the `Touch` constructor exists). True on
+ * hybrid laptops with a touchscreen even when the primary input is a mouse.
+ * Use this for *capability* decisions (e.g. binding touch handlers); use
+ * `hover` to choose between hover vs. tap UX.
+ */
+/* eslint-disable-next-line compat/compat */
+export const touch = hasWindow && typeof window.Touch !== 'undefined';
+
+/** A coarse pointer is available (e.g. touchscreen on a hybrid laptop). */
+export const coarse = matchMedia('(pointer: coarse)');
```
michaldudak: `(pointer: coarse)` reflects the primary pointer, not whether any coarse pointer is available. On a hybrid laptop with a mouse as the primary pointer and a touchscreen also present, this returns false, which contradicts both the flag name/comment and the nearby hybrid-device rationale.

Use `matchMedia('(any-pointer: coarse)')` instead.

romgrk: I will drop the pointer module from here. The rest is all static flags for production, whereas pointer could change in theory (e.g. plugged mouse) and may require dynamic logic. I'll do a separate pass for feature-detection and keep this focused on platform-detection.

PR #4921, opened 2026-05-27
[checkbox][radio][switch] Fix the type of forwarded refs

chuganzy:
Updates the forwarded ref types for Checkbox, Radio, and Switch roots from `HTMLElement` to `HTMLSpanElement`.

These components render a `<span>` by default, and their public props already extend `BaseUIComponentProps<'span', ...>`, so this aligns the implementation ref type with the rendered element.

Not sure about the breaking change policy in Base UI, but opened for now since the current type seems inaccurate.


atomiks:
We've kept components that are commonly replaced to another tag loose intentionally (`HTMLElement`). Only if they're not intended to be rendered as another tag (e.g. `div`s like `Positioner`) do we make the type narrow

chuganzy:
I got it, so this is intentional. Closing then! Thank you for having a look!

PR #4922, opened 2026-05-27
[menubar] Fix vertical menu focus behavior

atomiks:
This fixes Menubar behavior gaps around keyboard access, disabled state, and vertical orientation placement.

## Changes

- Move the initial composite highlight to the first enabled item when the default item is disabled.
- Enable Home and End navigation for Menubar triggers.
- Open vertical Menubar popups on the inline-end side by default while keeping horizontal Menubars opening downward.
- Keep open menu items disabled when the Menubar is disabled.
- Add CompositeRoot, Menubar, and positioner regression coverage.

### DIFF
--- a/packages/react/src/internals/composite/root/CompositeRoot.test.tsx
+++ b/packages/react/src/internals/composite/root/CompositeRoot.test.tsx
@@ -612,6 +612,32 @@ describe('Composite', () => {
   });
 
   describe('prop: disabledIndices', () => {
+    it('moves the initial tab stop to the first enabled item when the default item is disabled', async () => {
+      await render(
+        <CompositeRoot disabledIndices={[0]}>
+          <CompositeItem data-testid="1" />
+          <CompositeItem data-testid="2" />
+          <CompositeItem data-testid="3" />
+        </CompositeRoot>,
+      );
+
+      expect(screen.getByTestId('1')).toHaveAttribute('tabindex', '-1');
+      expect(screen.getByTestId('2')).toHaveAttribute('tabindex', '0');
+      expect(screen.getByTestId('3')).toHaveAttribute('tabindex', '-1');
+    });
+
+    it('keeps the initial tab stop when all items are disabled', async () => {
+      await render(
+        <CompositeRoot disabledIndices={[0, 1]}>
+          <CompositeItem data-testid="1" />
+          <CompositeItem data-testid="2" />
+        </CompositeRoot>,
+      );
+
+      expect(screen.getByTestId('1')).toHaveAttribute('tabindex', '0');
+      expect(screen.getByTestId('2')).toHaveAttribute('tabindex', '-1');
+    });
+
     it('disables navigating item when their index is included', async () => {
       function App() {
         const [highlightedIndex, setHighlightedIndex] = React.useState(0);
--- a/packages/react/src/internals/composite/root/useCompositeRoot.ts
+++ b/packages/react/src/internals/composite/root/useCompositeRoot.ts
@@ -33,8 +33,8 @@ import {
   type ModifierKey,
 } from '../composite';
 import { ACTIVE_COMPOSITE_ITEM } from '../constants';
-import { CompositeMetadata } from '../list/CompositeList';
-import { HTMLProps } from '../../types';
+import type { CompositeMetadata } from '../list/CompositeList';
+import type { HTMLProps } from '../../types';
 import { getTarget } from '../../../floating-ui-react/utils';
 
 export interface UseCompositeRootParameters {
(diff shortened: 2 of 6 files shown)


flaviendelangle (approved):
# PR Review: [menubar] Fix vertical menu focus behavior

**[mui/base-ui#4922](https://github.com/mui/base-ui/pull/4922)** · +138/−7 across 5 files · CI fully green (browser + JSDOM tests, lint, typecheck all pass)

**Verdict: Mergeable as-is.** The core fix is correct, well-targeted, and actually *removes* a pre-existing keyboard-accessibility silent failure. No blocking issues. The findings below are about test coverage, comment precision, and documenting an incidental cross-component impact.

Agents run: code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter. Skipped: type-design-analyzer (no new types — only `import type` reordering + a cast) and a standalone code-simplifier pass (the changed logic is already minimal; no simplification surfaced).

## Critical Issues (0)
None. The `else if` logic is correct, calls `onHighlightedIndexChange` with a single arg so `shouldScrollIntoView` defaults to `false` (no surprise scroll/focus), and the all-disabled case is safely guarded. `parent.context.orientation` is type-guaranteed non-null for menubar parents. `inline-end` is the correct RTL-aware logical side.

## Important — should address

**1. Shared-code change has an incidental, undocumented impact on Toolbar** — `useCompositeRoot.ts:137-144`
This fix lives in `useCompositeRoot`, shared by Menu, Select, Tabs, Toolbar, etc. — not just Menubar. Tracing each consumer's `disabledIndices`:
- Menubar → `undefined` → reads DOM `disabled`/`aria-disabled` (the intended target). ✓
- Select/Combobox/Menu/Tabs → `[]` (truthy) → suppresses the DOM-attribute fallback (`composite.ts:496`), so the branch effectively never fires. Safe.
- **Toolbar → a computed non-empty array** (`ToolbarRoot.tsx:36-44`), so a Toolbar with a disabled first item *will* also have its roving tab stop moved. This is the same bug being fixed and is almost certainly desirable, but the PR is titled/scoped "[menubar]" and doesn't mention it. → Note it in the PR description and ideally add a Toolbar regression test.

**2. Comment over-generalizes Tab behavior for `aria-disabled`** — `useCompositeRoot.ts:138-140`
> *"a disabled element is skipped by Tab while its enabled siblings sit at `tabIndex=-1`"*

True only for **natively `disabled`** elements (the Menu.Trigger in the test). But `isListIndexDisabled` also treats `aria-disabled="true"` as disabled, and an `aria-disabled` element is **not** removed from the tab order. Toolbar buttons default to `focusableWhenDisabled` → render `aria-disabled`, not native `disabled` (`useFocusableWhenDisabled.ts:38-47`) — yet the same code path runs. Suggested rewrite that stays accurate across both cases:
```ts
// The default highlighted item is disabled, so it should not hold the single
// roving tab stop (tabIndex=0): a natively disabled element is removed from the
// tab order entirely, and even a still-tabbable aria-disabled one shouldn't be
// the entry point. Move the tab stop to the first enabled item instead.
```

**3. Test coverage is narrowly scoped and asymmetric**
- **The core shared fix has no test at the composite level.** `CompositeRoot.test.tsx` (lines 614-713) already has the perfect harness (`prop: disabledIndices`, controlled `highlightedIndex` at 0) — none of its cases start with index 0 disabled. The fix is currently validated only indirectly through one menubar test.
- **No coverage of the non-menubar impact** (e.g., Toolbar with a disabled first button).
- **Horizontal-menubar `bottom` default is untested** — only the vertical `inline-end` arm of the `MenuPositioner.tsx:105-106` ternary has a test. A typo flipping the branches would pass CI. A symmetric "uses bottom when horizontal" test is cheap.
- **The "all items disabled" guard branch (no-op) is untested** — this is the defensive branch the source comment relies on.

## Suggestions — nice to have

- **S1.** Document the intentional all-disabled no-op (`useCompositeRoot.ts:142`) so a future maintainer doesn't "fix" it into a regression. *(silent-failure-hunter: behavior is correct, just opaque.)*
- **S2.** The new branch can write back over a **controlled** `highlightedIndex`. Harmless for every current consumer (only `TabsList` controls it, via internal state + empty `disabledIndices`), but consider gating on `externalHighlightedIndex == null` — or a comment — to make the contract "we never override a caller-set index" explicit.
- **S3.** Strengthen Home/End and disabled-trigger tests: cover a disabled first/last trigger, vertical orientation (no menubar test currently uses `orientation="vertical"` despite the PR title), and assert the reachable trigger's menu actually opens / arrows skip the disabled one — not just tab order.
- **S4.** The `disables menu items while the menubar is disabled` test (`Menubar.test.tsx:1035`) covers **pre-existing** behavior (`MenuTrigger.tsx:113` already folds in `parent.context.disabled`), independent of all three source changes. Keep it, but its fresh `describe` block implies it exercises a new path — a note would prevent confusion.
- **S5.** Optional one-line comment on the `inline-end` vertical-menubar default (`MenuPositioner.tsx:105`), mirroring intent. Low priority — it parallels the adjacent submenu branch.

## Strengths
- Correct root-cause fix: `useCompositeItem.ts:28` (`tabIndex: isHighlighted ? 0 : -1`) confirms a disabled item at index 0 genuinely traps the only tab stop. The change **reduces** a real, previously-silent a11y failure rather than introducing risk.
- RTL-aware logical `inline-end` (`useAnchorPositioning.ts:177-178`), consistent with the existing submenu default.
- `enableHomeAndEndKeys` matches the established opt-in pattern (TabsList, ToggleGroup) and the WAI-ARIA menubar spec.
- Deterministic tests (`act` + `waitFor`), layout assertions correctly gated behind `describe.skipIf(isJSDOM)`, observing resolved `side` via the `sideOffset` callback rather than brittle pixel math.
- `import type` conversions and the `as Array<HTMLElement | null>` cast are correct (confirmed by green typecheck).

## Recommended action
Nothing blocks merge. In priority order: **(2)** tighten the comment, **(3)** add the composite-level + horizontal-default tests, **(1)** mention the Toolbar impact in the PR description. S1–S2 are quick defensive polish; S3–S5 are optional.

PR #4911, opened 2026-05-26
[checkbox][switch][slider] Fix extra `validate` fn calls

mj12albert:
Fixes [4902](https://github.com/mui/base-ui/issues/4902)

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #4923, opened 2026-05-27
[field] Remove input validation prop helper

atomiks:
Cleanup pass following #4911, which fixed double validation for Checkbox, Switch, and Slider. This removes the shared input-specific validation prop helper so each field-aware control validates the semantic value it owns.

## Changes

- Removed `getInputValidationProps` from the field validation context.
- Moved plain string input validation into `Field.Control`.
- Updated Select, Combobox, NumberField, and OTP hidden inputs to use validation props without a default `onChange`.
- Added Field.Control regression tests for one on-change validation call and for clearing a prefilled required value.

## Behavior note

`Field.Control` now marks a prefilled required control invalid immediately when the user clears it in `validationMode="onChange"`, because clearing the value is treated as a dirty user edit.

### DIFF
--- a/packages/react/src/combobox/root/AriaCombobox.tsx
+++ b/packages/react/src/combobox/root/AriaCombobox.tsx
@@ -41,7 +41,7 @@ import { createCollatorItemFilter, createSingleSelectionCollatorFilter } from '.
 import { useCoreFilter } from './utils/useFilter';
 import { useTransitionStatus } from '../../internals/useTransitionStatus';
 import { useOpenInteractionType } from '../../utils/useOpenInteractionType';
-import type { MaybeBaseUIEvent, HTMLProps } from '../../internals/types';
+import type { HTMLProps } from '../../internals/types';
 import { useValueChanged } from '../../internals/useValueChanged';
 import { NOOP } from '../../internals/noop';
 import { FOCUSABLE_POPUP_PROPS } from '../../utils/popups';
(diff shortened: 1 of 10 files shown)


atomiks (commented):
## Code Review (GPT-5.5)
Approve ✅ No actionable issues remain after the follow-up edits; the validation helper removal preserves the semantic validation paths, and the targeted field-backed component tests pass on the current PR head.

### 1. Bugs / Issues (None)

Dominant type: `refactor`. Reviewed the full branch against freshly fetched `upstream/master`. Removing `getInputValidationProps` no longer drops validation because `Field.Control` owns its change validation, Select/Combobox/NumberField call `validation.change` from their hidden-input change paths, and OTP validates through `useValueChanged(value)`. The ordering-sensitive `Field.Control` path now documents that `setDirty` must precede `validation.change`, which covers the prefilled-required clear behavior.

The useful Opus review points are accounted for: the weak test assertion was tightened to `toBeInTheDocument()`, the ordering comment was added, `pnpm-lock.yaml` is no longer in the PR diff, and the stale `MaybeBaseUIEvent`/`preventBaseUIHandler` indirection is gone. I also rechecked the name-resolution and props-getter notes; neither needs a code change.

### Test Coverage Assessment
Validated with `pnpm test:jsdom FieldControl Select Combobox NumberField OTPField --no-watch` on the current PR head: 53 test files passed, with 1,587 tests passed and 85 skipped. I also verified the new `Field.Control` required-clear regression fails against `upstream/master` at the `aria-invalid` assertion, so it covers the behavior change.

PR #4925, opened 2026-05-27
[popups] Fix positioning and viewport edge cases

atomiks:
Updates shared popup behavior so keep-mounted/custom-anchor positioning, preventUnmountOnClose reopen cycles, and viewport API docs are handled consistently across popup components.

## Changes

- Observes the actual positioning reference for keep-mounted popups so custom anchors trigger updates.
- Clears preventUnmountOnClose when popups reopen, including controlled external reopens.
- Shares active-trigger open-state updates across popup stores and menu root.
- Prevents disabled hover popover triggers from opening.
- Exports consistent viewport props/state types and regenerates popup viewport API docs.

### DIFF
--- a/docs/src/app/(docs)/react/components/menu/types.md
+++ b/docs/src/app/(docs)/react/components/menu/types.md
@@ -481,6 +481,7 @@ Re-export of [Viewport](#viewport) props.
 
 ```typescript
 type MenuViewportState = {
+  /** The activation direction of the transitioned content. */
   activationDirection: string | undefined;
   /** Whether the viewport is currently transitioning between contents. */
   transitioning: boolean;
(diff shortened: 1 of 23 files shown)


flaviendelangle (approved):
# PR #4925 Review — `[popups] Fix positioning and viewport edge cases`

**Scope:** 22 files, +276/−91. Three behavioral fixes (custom-anchor tracking, `preventUnmountOnClose` reset-on-reopen, disabled-hover) + a viewport type/namespace refactor + a docs/JSDoc correction. Reviewed by four specialized agents (code, tests, comments, types); their key claims were independently re-verified (typescript, `docs:api`, eslint, prettier, and regression runs that revert each fix).

## Critical Issues (0)
None. All three fixes were empirically confirmed to fail on `master` and pass on the branch — they are genuine regression guards, not tautological.

## Important Issues (1)

**1. Menu's reset-on-reopen behavior is untested — `MenuRoot.tsx:314`**
This PR replaced Menu's hand-written active-trigger block with the shared `setPopupOpenState`, which newly resets `preventUnmountingOnClose = false` on open. Menu fully supports `preventUnmountOnClose` (`MenuRoot.tsx:248-250`), making it the **highest-regression-risk** component — it's the only one whose call site is bespoke rather than a uniform store method (Dialog/Tooltip/PreviewCard all call `setPopupOpenState` identically inside their store classes). The only existing Menu test (`MenuRoot.test.tsx:1562`) covers the `actionsRef.unmount()` path, not reopen-then-normal-close.
→ **Recommend** adding a Menu analog of `PopoverRoot.test.tsx:644`: open → close once with a gated `preventUnmountOnClose()` (menu stays mounted) → reopen via trigger → normal-close → assert `queryByRole('menu')` is null.

## Suggestions (3, all optional)

1. **PreviewCardViewport namespace order** — `PreviewCardViewport.tsx:85-88` re-exports `State` before `Props`; the other three viewports use `Props` then `State`. Cosmetic (alias semantics are order-independent) but it propagates into the generated `preview-card/types.md` ordering. One-line tidy for intra-PR symmetry.
2. **Direct coverage for Dialog/Tooltip/PreviewCard reset-on-reopen** — currently rely on Popover as a proxy. Defensible since they share the store-level call site; a single mirrored test would close the loop cheaply.
3. **Undocumented "why"** — the functional core of the bug fix (`popupStoreUtils.ts:124-126` and the `open ? false : …` sync at `:270`) has no comment. A one-liner like *"Opening clears any pending prevent-unmount request from a previous close cycle"* would help future maintainers.

## ⚠️ One agent finding to disregard
The type-design agent suggested collapsing the multi-line `extends BaseUIComponentProps<…>` in `PreviewCardViewportProps` onto one line. I ran `pnpm prettier --check` — it passes as-is; the wrap is **forced by line length** (PreviewCard's identifiers are longer than the siblings'). Collapsing it would fail formatting. Leave it.

## Strengths
- **Docs fix corrects a real, pre-existing error.** `data-activation-direction` was documented as `top`/`bottom`, but the producer (`usePopupViewport.tsx:285`) emits `right`/`left` + `down`/`up` with an `''` fallback. The PR fixes the tokens, fixes a missing-`$` typo in the popover variant, adds the `''` option, and unifies all four files with the already-correct Menu definition.
- **`domReference` → `reference`** (`useAnchorPositioning.ts:507`) correctly tracks custom anchors for `keepMounted` popups (a custom `anchor` sets `elements.reference` via `setPositionReference` while `domReference` stays the trigger), and aligns the keepMounted path with the non-keepMounted `whileElementsMounted` path. Safe for virtual elements.
- **`!disabled` hover guard** (`PopoverTrigger.tsx:83`) brings Popover in line with Tooltip/Menu/SubmenuTrigger, which already guard this way; PreviewCard's `<a>` trigger has no `disabled` concept. No shared abstraction warranted for a one-token guard.
- **Type refactor** finishes a partial migration to the codebase's dominant convention (~180 components use top-level interface + namespace alias); `instant` unions are correctly differentiated per component (verified against each store), and the `string` widening for `activationDirection` matches its runtime provider.
- **Verification clean:** `pnpm typescript`, `pnpm docs:api` ("No files needed updating"), `pnpm eslint`, `pnpm prettier`, plus PopoverRoot/PopoverTrigger/PopoverPositioner/Menu/Tooltip/PreviewCard test suites all pass.

## Recommended Action
1. Add the Menu reset-on-reopen test (the one Important item). 
2. Optionally apply suggestions 1–3.
3. Everything else is merge-ready.

PR #4926, opened 2026-05-27
[radio group] Fix disabled selected form submission

atomiks:
Radio groups now match native form submission when the selected radio is disabled, and the indicator state type reflects the root state it already receives.

## Changes

- Return `null` from form values when the tracked selected radio is disabled.
- Add Chromium coverage for radios disabled after mount and initially disabled.
- Expose the full radio root state through `Radio.Indicator.State` docs and types.

- [P1] Disabled selected radios were excluded from native `FormData` but still reported by `onFormSubmit`.
- [P2] `RadioIndicatorState` only declared `checked` even though it receives root field state at runtime.

### DIFF
--- a/docs/src/app/(docs)/react/components/radio/types.md
+++ b/docs/src/app/(docs)/react/components/radio/types.md
@@ -108,10 +108,26 @@ Re-export of [Indicator](#indicator) props.
 
 ```typescript
 type RadioIndicatorState = {
-  /** Whether the radio button is currently selected. */
-  checked: boolean;
   /** The transition status of the component. */
   transitionStatus: TransitionStatus;
+  /** Whether the radio button is currently selected. */
+  checked: boolean;
+  /** Whether the component should ignore user interaction. */
+  disabled: boolean;
+  /** Whether the user should be unable to select the radio button. */
+  readOnly: boolean;
+  /** Whether the user must choose a value before submitting a form. */
+  required: boolean;
+  /** Whether the field has been touched. */
+  touched: boolean;
+  /** Whether the field value has changed from its initial value. */
+  dirty: boolean;
+  /** Whether the field is valid. */
+  valid: boolean | null;
+  /** Whether the field has a value. */
+  filled: boolean;
+  /** Whether the field is focused. */
+  focused: boolean;
 };
 ```
 
--- a/docs/src/app/(docs)/react/components/radio/types.md
+++ b/docs/src/app/(docs)/react/components/radio/types.md
@@ -55,15 +55,15 @@ type RadioRootState = {
   readOnly: boolean;
   /** Whether the user must choose a value before submitting a form. */
   required: boolean;
-  /** Whether the field has been touched. */
+  /** Whether the radio button has been touched (when wrapped in Field.Root). */
   touched: boolean;
-  /** Whether the field value has changed from its initial value. */
+  /** Whether the radio button's value has changed from its initial value (when wrapped in Field.Root). */
   dirty: boolean;
-  /** Whether the field is valid. */
+  /** Whether the radio button is in a valid state (when wrapped in Field.Root). */
   valid: boolean | null;
-  /** Whether the field has a value. */
+  /** Whether the radio button has a value (when wrapped in Field.Root). */
   filled: boolean;
-  /** Whether the field is focused. */
+  /** Whether the radio button is focused (when wrapped in Field.Root). */
   focused: boolean;
 };
 ```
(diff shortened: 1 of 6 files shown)


flaviendelangle (approved):
# PR #4926 Review Summary

**PR:** [radio group] Fix disabled selected form submission — by @atomiks
**Scope:** 5 files, narrow & focused. Fixes form-data parity for disabled radios + widens `Radio.Indicator.State` / `Radio.Root` context types to match runtime.

---

## Critical Issues
**None.** All five reviewers concur: no blocking concerns.

---

## Important Issues

### 1. Silent value divergence when controlled `value` doesn't match any rendered radio
**File:** `packages/react/src/radio-group/RadioGroup.tsx:148-153`
**Source:** silent-failure-hunter

Pre-PR: `<RadioGroup value="missing">` submits `"missing"` (a signal).
Post-PR: same code submits `null` silently — because `input?.checked` is false. Same `null` is reported for "no input registered", "disabled input", and "controlled/DOM desync". A dev warning when `checkedValue != null && !input.checked && !input.disabled` would catch this.

> ⚠️ Counter-evidence from code-reviewer: the validation path still uses `checkedValue` (via `getRegistrationValue` preferring `registration.value` over the override at `useFieldControlRegistration.ts:49`), so `required` validation isn't affected. Only the form-submit payload changes. Worth confirming this is the intended trade-off, not a regression.

### 2. Drop `it.skipIf(isJSDOM)` on the two new tests
**File:** `packages/react/src/radio-group/RadioGroup.test.tsx:1035, 1072`
**Source:** pr-test-analyzer (verified locally — both tests pass in jsdom ~10ms)

The neighboring "submits null to onFormSubmit when no radio is selected" test at line 1038 uses the same `fireEvent.click(submit)` + `handleSubmit.mock.calls[0][0]` pattern and is not skipped. jsdom 27.4 correctly returns `null` from `FormData(form).get('group')` for disabled radios. The skip needlessly halves coverage at default test runs.

### 3. Remove the two redundant test comments
**File:** `packages/react/src/radio-group/RadioGroup.test.tsx:1062, 1093`
**Source:** comment-analyzer

```ts
// Native form submission excludes disabled inputs.
expect(new FormData(form).get('group')).toBe(null);
```
The test name already says "to match native form data" and the `FormData(...)` call itself self-documents. Per `AGENTS.md` ("default to writing no comments"), remove these.

---

## Suggestions

### 4. Clarify the ternary in `getFormValue` with explicit guards
**File:** `packages/react/src/radio-group/RadioGroup.tsx:152`
```ts
return input?.checked && !input.disabled ? (checkedValue ?? null) : null;
```
Four distinct outcomes are collapsed into one ternary. An early-return form (`if (!input || input.disabled || !input.checked) return null;`) survives future edits (e.g., adding the dev warning from #1) without rewriting.

### 5. Add a "disabled → re-enabled" transition test
Test 1 toggles `disabled=true` once; nothing covers the reverse transition. A future refactor of `registerInputRef` swap logic could silently keep returning `null` after re-enable.

### 6. Add a test for group-level `<RadioGroup disabled>` form submission
`RadioGroup.tsx:155` passes `!disabled` as the `enabled` flag, so a group-level disabled never registers the field — yielding `{}` instead of `{ test: null }`. This distinct path is uncovered.

### 7. Note `Radio.Indicator.State` widening in the changelog
**Source:** type-design-analyzer

Public type extends from 2 fields to 10. Render-prop consumers and `Pick`/`Omit` users are safe; consumers doing exact structural construction (`const s: RadioIndicatorState = {checked:true, transitionStatus:'idle'}`) get `TS2741`. Trivial migration, but worth a one-line release note.

### 8. Pre-existing staleness in `groupInputRef.current` (follow-up)
`RadioRoot.tsx:108-123` never nulls `groupInputRef.current` on unmount. The new `getFormValue` actively reads from this potentially-detached node where the prior code didn't. Out of scope for this PR — flag for follow-up.

---

## Strengths

- **Correct fix.** Cross-verified by code-reviewer: `groupInputRef.current` is fresh through `useStableCallback` + ref read; on disable-toggle, React commits `disabled` to the DOM before the effect re-registers, so `input.disabled` is `true` when `getFormValue` runs. ✅
- **Brings Radio into alignment with Checkbox/Switch** — both already use `type RootContext = RootState` and `IndicatorState extends RootState`. Radio was the outlier; PR fixes the inconsistency.
- **Type widening matches what the provider already passes** — `RadioRoot.tsx:243` already set `contextValue = state` (full memoized state), so this is type-as-truth cleanup, not a runtime broadening.
- **`useStableCallback` + ref-read pattern is idiomatic** and AGENTS.md-compliant. No `useTimeout`/`useIsoLayoutEffect`/shadow-DOM regressions.
- **The new source comment in `RadioGroup.tsx:149-150` earns its place** — explains a non-obvious native-parity decision (kept per comment-analyzer).
- **Existing `data-*` attribute test at `RadioGroup.test.tsx:172-198`** indirectly verifies the new `RadioIndicatorState` fields propagate correctly through `stateAttributesMapping`. No additional indicator tests needed.

---

## Recommended Action

1. **Decide on #1:** add a dev warning for the controlled/DOM desync case, or document the new "missing value silently becomes null" behavior. This is the only finding where reviewers materially disagreed.
2. **Drop `it.skipIf(isJSDOM)`** on the two new tests (#2). Quick win.
3. **Remove the two redundant test comments** (#3). Quick win.
4. *(Optional)* Apply #4 (clarify ternary) and add the missing test cases (#5, #6).
5. *(Optional)* Add a changelog note for the type widening (#7).
6. *(Follow-up)* File an issue for the `groupInputRef` unmount staleness (#8).

The PR is ship-ready as a bug fix; items 1-3 would strengthen it before merge.

PR #4930, opened 2026-05-28
[radio] Fix Radio to trigger on space key press

chuganzy:
Radio should trigger on Space as per the WAI-ARIA radio pattern:
https://www.w3.org/WAI/ARIA/apg/patterns/radio/

Previously, `useButton` ran `onClick` from `onKeyDown`, but Radio's click handler ignored it because the event was `defaultPrevented`.

Unlike other "composite items", it is more natural for Radio to activate on `keyUp`, so this fixes Space key activation by setting `composite: false` on `useButton`.


atomiks (approved):
Looks like a regression from #4053

PR #4931, opened 2026-05-28
[menu][select] Remove kept-mounted tabIndex workaround

atomiks:
A kept-mounted Menu or Select item could be re-enabled as tabbable after focus left the popup. This removes the root-level microtask repair and keeps item tabbability tied to the popup state instead.

## Changes

- Removed the Menu and Select `queueMicrotask` tabIndex workaround.
- Kept highlighted Menu and Select items tabbable only while their popup is open.
- Restored temporarily disabled portal focus before focus management runs on reopen.
- Added keep-mounted regression coverage for Menu, Select, and Popover.


atomiks:
## Code Review (GPT-5.5 + Opus 4.8)
Approve ✅: Reviewed as a bug-fix/refactor against freshly fetched `upstream/master`; the shared focus timing change is sound, the workaround removal preserves Menu/Select tabbability behavior, and the targeted suites pass on the PR.

### 1. Bugs / Issues (None)

No blocking or non-blocking defects found.

This PR removes the Menu/Select close-time `queueMicrotask` tabIndex repair and makes item tabbability declarative via `open && highlighted`. That is the right ownership boundary: when the popup closes, React re-renders the highlighted item to `tabindex="-1"` instead of relying on an imperative race with Floating UI’s focus restoration.

The `FloatingPortal` change also makes sense. Moving `enableFocusInside` from close-time passive effect to reopen-time `useIsoLayoutEffect` means temporarily disabled focus is restored before `FloatingFocusManager`’s queued focus-on-open microtask runs. I checked the surrounding `FloatingFocusManager` and `tabbable` logic; Opus’s timing analysis matches the code.

The change removes two component-specific hacks from `MenuRoot` and `SelectRoot` while keeping the shared behavior in the floating portal layer. The deleted workaround sites are fully addressed, and the remaining `activeIndex`/`listRef` values are still used elsewhere, so there is no obvious cleanup or lint fallout from the removal.

The Select test is valid even without an explicit `keepMounted` prop: Select’s portal/positioner path keeps the popup mounted after close until explicit unmount, so the test exercises the intended closed-but-mounted state.

### Opus 4.8 Notes

Opus’s main points are consistent with my review:

- The close-time passive effect previously restored saved `tabindex` values after Menu/Select tried to repair them.
- The new layout-effect timing is intentional, because it runs before the focus manager’s queued initial-focus step.
- The blast radius is narrow: modal portals do not use the non-modal focusout path, and non-kept-mounted portals unmount instead.
- The added per-item `open` subscription is a small cost, but appropriate for making tabbability declarative.
- Dialog `keepMounted` is not covered directly, but Menu, Select, and Popover cover the shared mechanism sufficiently.

### Test Coverage Assessment

Validated on the PR:

```sh
pnpm test:chromium MenuRoot --no-watch
pnpm test:chromium SelectRoot --no-watch
pnpm test:chromium PopoverRoot --no-watch
```

`MenuRoot` and `SelectRoot` passed. `PopoverRoot` had one transient focus-order failure on the first full run, but the isolated case passed and the full suite passed on rerun.

I also overlaid the new tests onto current `upstream/master`. The new Popover reopen-focus test fails on master and passes on the PR, confirming the shared portal timing fix. The new Menu/Select tests pass on master as well because the old workaround already covered those behaviors; they are still useful behavior-preservation coverage for safely removing that workaround.

PR #4932, opened 2026-05-28
[switch][test] Add regression coverage and fix `id` doc

atomiks:
Adds regression coverage for Switch cancellation and unchecked form values, plus aligns the id API docs with the current hidden-input/native-button behavior.

## Changes

- Added tests that cancel onCheckedChange from root and hidden input clicks.
- Added Chromium coverage for custom value and uncheckedValue across an off/on/off cycle.
- Updated the Switch id JSDoc and generated API docs.


flaviendelangle (approved):
Not applicable — the parallel review agents have already returned and I'm just aggregating their findings into the final summary now.

# PR #4932 Review Summary

**PR**: [switch][test] Add regression coverage and fix `id` doc — 83 additions / 2 deletions across 3 files.

## Critical Issues
None. The PR is small, low-risk, and the new tests would catch the targeted regressions.

## Important Issues
1. **Missing keyboard-driven cancel coverage** (`SwitchRoot.test.tsx`) — the two new cancel tests cover root click and hidden-input click, but Enter/Space activation goes through a different path (`useButton`/native bubble) and is left untested. A regression that bypasses the `isCanceled` guard on keyboard activation would not be caught. Existing pattern at `SwitchRoot.test.tsx:74-87` shows how to add it cheaply.
2. **Missing `nativeButton=true` cancel coverage** — the `rootProps.onClick` synthetic-dispatch path (`SwitchRoot.tsx:151-172`) is skipped for native buttons, so cancel behavior in that mode is structurally different and untested.

## Suggestions
1. **JSDoc wording is slightly ambiguous** (`SwitchRoot.tsx:275-278`) — "is applied to the root element" can read as "also applied to," but the runtime moves the id rather than duplicating it. Suggest appending "instead": *"When `nativeButton` is `true`, the id is applied to the root element instead."* Then re-run `pnpm docs:api`.
2. **Minor style nit in JSDoc** — for consistency with `inputRef` on line 298, consider "the hidden `<input>` element" (backticks) rather than plain "hidden input element."
3. **Cross-component follow-up** — `CheckboxRoot.tsx:438-440` has the same incomplete "The id of the input element" JSDoc and identical `nativeButton` branching. Worth applying the same fix there to keep documentation in sync (out of scope for this PR, but flag for a follow-up).
4. **Overspecified assertions in cancel tests** — `data-dirty` and `data-filled` checks restate the same condition as the `aria-checked`/`input.checked` checks (all derive from `setCheckedState` firing). Trimming would sharpen intent, but it's a nit.
5. **Cycle test redundancy is borderline** — the new off/on/off cycle overlaps with the existing test at `SwitchRoot.test.tsx:553-591`. The reviewers split on whether it's redundant: it does uniquely exercise `value="yes"` + `uncheckedValue="no"` *together*, which no other test covers. Keep, but consider narrowing assertions to the unique aspect.

## Strengths
- New cancel tests use `user.click` (not `fireEvent.click`), exercising the full pointer-dispatch path that includes the `isCanceled` guard.
- Tests hit both entry points (root and hidden input), confirming the guard is uniform.
- Browser-only test correctly gated via `it.skipIf(isJSDOM)`; placement in the existing `Form` describe block is appropriate.
- JSDoc fix is factually correct against `SwitchRoot.tsx:89, 127, 180` and is reinforced by the existing test at `SwitchRoot.test.tsx:371-391`.
- All AGENTS.md conventions satisfied (Vitest-only APIs, no stray `flushMicrotasks`, file naming, commit-scope format).

## Recommended Action
1. (Optional) Add Enter/Space cancel test and a `nativeButton=true` cancel test for completeness.
2. (Optional) Tighten the JSDoc wording with "instead" and re-run `pnpm docs:api`.
3. Ship — nothing here blocks merge.

PR #4933, opened 2026-05-28
[toast] Fix timer and limit edge cases

atomiks:
Toast now handles several limit, timer, and focus edge cases that could leave hidden or inactive toasts in the wrong state.

## Changes

- Recomputed limited toast state when the provider `limit` changes.
- Shared limited-state recomputation across add, close, and provider sync paths.
- Reset paused timer state when all active timers are cleared so future toasts can pause and resume normally.
- Only prevented touch scrolling while a toast swipe is active.
- Kept timers paused while the window is blurred, even if mouseleave or blur handlers run.
- Skipped ending and limited toasts when focus enters the viewport.
- Updated Toast docs and API output for the current limit and manager behavior.
- Added regression coverage for Toast root, viewport, content, manager, and store behavior.

### DIFF
--- a/docs/src/app/(docs)/react/components/toast/page.mdx
+++ b/docs/src/app/(docs)/react/components/toast/page.mdx
@@ -60,7 +60,7 @@ import { Toast } from '@base-ui/react/toast';
 A global toast manager can be created by passing the `toastManager` prop to the `<Toast.Provider>`.
 This enables you to queue a toast from anywhere in the app (such as in functions outside the React tree) while still using the same toast renderer.
 
-The created `toastManager` object has the same properties and methods as the `Toast.useToastManager()` hook.
+The created `toastManager` exposes the same `add`, `close`, `update`, and `promise` methods as the `Toast.useToastManager()` hook. Unlike the hook, it does not return the reactive `toasts` array, since it lives outside the React tree.
 
 ```tsx title="Creating a manager instance"
 const toastManager = Toast.createToastManager();
@@ -148,8 +148,8 @@ The `data-swipe-direction` attribute can be used to determine the swipe directio
 }
 ```
 
-The `data-limited` attribute indicates that the toast was removed from the list due to exceeding the `limit` option.
-This is useful for animating the toast differently when it is removed from the list.
+The `data-limited` attribute indicates that the toast exceeded the `limit` option.
+Limited toasts remain mounted but inert, so this is useful for hiding them or animating them differently.
 
 The `updateKey` property increments when a toast is updated or upserted.
 This can be used to replay attention-grabbing styles by switching animation names or, when remounting is acceptable, by including it in a React `key`.
--- a/docs/src/app/(docs)/react/components/toast/page.mdx
+++ b/docs/src/app/(docs)/react/components/toast/page.mdx
@@ -149,7 +149,7 @@ The `data-swipe-direction` attribute can be used to determine the swipe directio
 ```
 
 The `data-limited` attribute indicates that the toast exceeded the `limit` option.
-Limited toasts remain mounted but inert, so this is useful for hiding them or animating them differently.
+Limited toasts remain mounted with the HTML `inert` attribute, so this is useful for hiding them or animating them differently.
 
 The `updateKey` property increments when a toast is updated or upserted.
 This can be used to replay attention-grabbing styles by switching animation names or, when remounting is acceptable, by including it in a React `key`.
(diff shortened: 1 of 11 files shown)


flaviendelangle (approved):
## Critical Issues
None — the PR is mergeable as is.

## Important Issues

1. **`clearTimer` reset bypassed when timer fires for a removed toast** — `packages/react/src/toast/store.ts:444-447`, `:411-419`, `:312-318`
   `scheduleTimer` and `resumeTimers` inline `this.timers.delete(id); callback()` instead of routing through `clearTimer`. `closeToast` then early-returns at `:314` if the toast was concurrently removed via `removeToast` — so the `areTimersPaused = false` reset never happens, re-introducing the bug this PR fixes. Narrow path (requires external `removeToast` of a toast with a running timer), but easy to close: extract `private onTimerFired(id) { this.timers.delete(id); if (this.timers.size === 0) this.areTimersPaused = false; }` and call it from both inline callbacks.

2. **`store.handleDocumentPointerDown` and `ToastViewport.handleKeyDown` (Shift+Tab) still call `resumeTimers()` unconditionally** — `store.ts:425-437`, `ToastViewport.tsx:152-162`
   The PR added `resumeTimersIfWindowFocused` to `handleMouseLeave`/`handleBlur` but didn't apply the same gate elsewhere. Likely intentional (these paths imply explicit user activity), but worth confirming; if intentional, a one-line comment explains the asymmetry.

3. **Touchmove test doesn't verify scroll re-enables after pointer end** — `packages/react/src/toast/root/ToastRoot.test.tsx:584-596`
   The new test fires `pointerCancel` but doesn't re-dispatch a `touchmove` afterward to assert `defaultPrevented === false`. Add one more dispatch after `pointerCancel` to lock in the contract.

4. **Brittle `addEventListenerSpy` reversed-find pattern** — `ToastViewport.test.tsx` ~`:492-494`, `:539-541`
   Selecting the listener by last-registered ordering is fragile. Prefer dispatching a real `window.dispatchEvent(blurEvent)` or filter by the capture flag (the viewport's `blur` listener is the only `true`-capture one in these tests).

## Suggestions

5. **`applyLimited` rot risk** — `store.ts:69-83`
   Comment asserts "Toasts are ordered newest-first" but the function trusts iteration order. Tighten to "Caller guarantees newest-first ordering (see `addToast`)."

6. **Docs use "inert" colloquially** — `docs/.../page.mdx:151`, `types.md`, `ToastProvider.tsx:75`
   Implementation does set the HTML `inert` attribute (`ToastRoot.tsx:521`), so consider naming it: "remain mounted with the [`inert`](https://developer.mozilla.org/docs/Web/HTML/Global_attributes/inert) attribute".

7. **Missing test coverage**:
   - Limit change while toasts are mid-`ending` (current "limit" test uses only `starting` toasts).
   - Timer-elapsed callback path for `clearTimer` (covers Finding 1).
   - Focus guard with `limit > 0` and *all visible toasts ending* (only `limit=0` covered).
   - `syncProviderProps` early-return reference stability.

8. **Limited flag not cleared on ending toasts** — `store.ts:69-83`
   `applyLimited` skips ending toasts entirely, whereas the old `addToast` reset `limited: false` on every toast. Almost certainly benign (toast is on its way out), but no test pins this transition down.

## Strengths

- `applyLimited` cleanly deduplicates three previously-divergent code paths and is reference-stable to avoid extra re-renders.
- `syncProviderProps` correctly uses `useIsoLayoutEffect` per AGENTS.md with the right dep array and an early-return for unchanged values.
- The `hasActiveToasts` check (replacing `toasts.length === 1`) genuinely fixes a corner case where ending-only toasts left `hovering`/`focused` stuck `true`.
- The 5 `timer pausing` unit tests in `store.test.ts` enumerate failure modes in DAMP style — each test clearly documents the bug it pins down.
- Public docs accurately reflect the new behavior (`data-limited` ↔ `inert` attribute is correctly wired up).
- F6 focus-guard skip correctly falls back to `restoreFocusToPrevElement` when no toast can receive focus.
- Tests follow project conventions (vitest + jest-dom only, `it.skipIf` used correctly, no chai/sinon).

## Recommended Action

1. (Optional but cheap) Centralize the `areTimersPaused` reset into a single helper called from both `scheduleTimer`/`resumeTimers` inline callbacks (Finding 1).
2. Add the missing post-cancel `touchmove` assertion (Finding 3).
3. Decide intent on `handleDocumentPointerDown`/`handleKeyDown` resume paths (Finding 2) and either gate them or add a comment.
4. Consider the docs/comment polish (Findings 5, 6).
5. Test gaps (Finding 7) are nice-to-haves — current coverage is already strong for the fixes shipped.

PR #4934, opened 2026-05-28
[select] Fix autofill and selected state edge cases

atomiks:
Fixes several Select edge cases where displayed selection state, autofill matching, and locked interactions could drift from the current value.

## Changes

- Match Select autofill against rendered item text after value and label checks so primitive values like `CA` can accept browser-filled text like `Canada`.
- Keep item `data-selected` tied to the current value when controlled value changes while the popup is open.
- Keep Field filled state consistent with placeholder semantics for empty string values.
- Ignore item commits when a controlled-open Select is disabled or read-only.
- Keep scroll arrows mounted through transition status so ending styles can apply.
- Add regression coverage for the fixed Select behaviors.


packages/react/src/select/store.ts
```diff
@@ -97,12 +97,9 @@ export const selectors = {
       );
     }
 
-    // `selectedIndex` is only updated after the items mount for the first time,
-    // the value check avoids a re-render for the initially selected item.
-    if (state.selectedIndex === index && state.selectedIndex !== null) {
-      return true;
-    }
```
atomiks, line 104: Select open renders are unchanged (14) according to the benchmark and running the benchmark showed ~no difference

What the old code was trying to do: use `selectedIndex` as a fast path after items mount, while falling back to the value comparison before `selectedIndex` is initialized so the initially selected item doesn’t need a second render. The problem is that `selectedIndex` is not kept in sync while open, so it can become stale and override the actual controlled value.

flaviendelangle (commented):
# PR #4934 Review Summary — `[select] Fix autofill and selected state edge cases`

Four review agents (code, comments, tests, silent failures) ran in parallel against branch `pr-4934`. Tests, lint, and types were verified clean by the code reviewer.

The fixes themselves are solid — every behavior change is correctly implemented and aligned with peer components. **The main risk is test coverage**: several behavioral changes ship under-tested, and one existing test may not actually exercise the new guard it appears to verify.

---

## Critical (should fix before merge)

1. **`commitSelection` disabled/readOnly test may not isolate the new guard** — `SelectRoot.test.tsx:872-904` uses `user.click(...)`, but `useButton({disabled})` already swallows the click upstream. Verify the test fails when you remove `SelectItem.tsx:134-138`; if not, add a path that reaches `commitSelection` (keyboard Enter on a highlighted item, or `defaultOpen` variant).
2. **No coverage for `SelectScrollArrow` `visible → mounted` change** — the entire fix (and the new `transitionStatusMapping`) is untested. At minimum add a jsdom smoke test that the arrow remains in the DOM through the exit tick; ideally a chromium test asserting `data-ending-style` / `data-transition-status="ending"` during exit.
3. **Initial-mount regression scenario for the removed `isSelected` short-circuit is untested** — add `<Select.Root defaultOpen value="b">` (and an object-value + `isItemEqualToValue` variant) asserting `option b` has `data-selected` on first paint, without any user interaction. The deleted short-circuit existed precisely for that scenario.
4. **Multi-select stale-index not covered** — the line 168 assertion is single-select. Mirror it for `value={['a','b']} → ['b']` while open.

## Important (should address)

5. **Autofill matcher precedence is untested** — the rewrite changed the closure but the test only covers "rendered label matches". Add tests that pin (a) value wins over rendered text of a later item, (b) case-insensitive matching of the new rendered-text path, (c) tolerance of `null` rendered labels.
6. **Autofill *success* path with `useValueChanged`** — only the cancel path is tested. Add a Field-wrapped test that asserts a successful autofill produces `data-dirty=""` and triggers a `validate` spy. Pins that `setDirty` + `validation.change` still run after the refactor.
7. **Rewrite comment at `SelectRoot.tsx:584-586`** — currently restates the three matching paths (the WHAT). Suggested condensed form:
   ```
   // Browsers autofill the rendered text (e.g. "United States"), which is neither the
   // serialized value nor the serialized label, so match it against `labelsRef` too.
   ```

## Suggestions

8. Delete redundant test comments at `SelectRoot.test.tsx:167` and `:806` — both restate what the `it()` title + assertion already convey. (If keeping :806, expand it to describe the `cancel → useValueChanged-doesn't-fire` chain.)
9. `hasSelectedValue`: add a coverage case using a non-string value + custom `itemToStringValue` returning `''` (the present test happens to also pass on the old code).
10. UX follow-up (out of scope): items in a forced-open `disabled`/`readOnly` Select have no visible disabled state or `aria-disabled`, so the new no-op click is invisible to users. Worth a separate PR cascading the lock state to items.

## Strengths

- Each fix correctly delivers what its description claims (verified by code reviewer and silent-failure hunter independently).
- `isSelected` change makes `value` the single source of truth and correctly fixes the simultaneous-`data-selected` bug when controlled value changes while the popup is open (`syncSelectedIndex` defers while `open`).
- `hasSelectedValue` local definition correctly mirrors `selectors.hasSelectedValue` — Field's filled state now agrees with the trigger placeholder semantics.
- Removing the explicit `setDirty` / `validation.change` and deferring to `useValueChanged` is the right refactor — it fixes the canceled-autofill-still-dirty bug on master.
- `SelectScrollArrow` changes align it with peer components (`SelectItemIndicator`, `SelectPopup`, `SelectArrow`, `SelectBackdrop`) which all already merge `transitionStatusMapping`.
- Comments at `SelectItem.tsx:134`, `SelectRoot.tsx:197,600`, and `store.ts:100` are good examples of non-obvious WHY documentation.
- All AGENTS.md rules observed: no `as any`, no `React.useLayoutEffect`, no `window.setTimeout`, Vitest-only matchers, etc.

## Recommended action

1. Investigate finding #1 first (test isolation) — if confirmed, expand to also cover keyboard activation.
2. Add the missing tests for findings #2, #3, #4 — all are short additions.
3. Apply finding #7 (comment rewrite) and #8 (comment deletions).
4. Re-run `pnpm test:jsdom SelectRoot --no-watch` and `pnpm test:chromium SelectRoot --no-watch`.

flaviendelangle (approved):
# PR #4934 Review Summary — `[select] Fix autofill and selected state edge cases`

Reviewed via 4 specialist agents (code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter) against a worktree at the PR HEAD (`/tmp/pr-4934`, commit `6f30a2d56`).

**Overall:** Solid, focused bug-fix PR. No critical or important defects. The logic changes are correct, well-tested, and CLAUDE.md-compliant. Findings below are improvements and risks worth a quick second pass before merge.

---

## Critical Issues
**None.**

## Important Issues

1. **Missing unmount guard on `queueMicrotask(handleChange)`** — `packages/react/src/select/root/SelectRoot.tsx:605-606`
   If the root unmounts between the autofill change event and the microtask, `setValue` runs on an unmounted root and React will warn in dev. Gate with a mountedRef or equivalent.

2. **Multiple-mode `data-filled` doesn't honor `''` serialization** — `packages/react/src/select/root/SelectRoot.tsx:199-201`
   The PR unified single-value semantics so a value serializing to `''` counts as empty, but the `multiple` branch still only checks `Array.isArray(value) && value.length > 0`. A non-empty array of items that all serialize to `''` will still mark the field as filled. Either apply the same serialize-check in the multiple branch (with a test) or add a comment explaining the asymmetry.

3. **Comment under-explains the autofill matcher** — `packages/react/src/select/root/SelectRoot.tsx:583-585`
   Three OR-arms exist (`stringifyAsValue` → `stringifyAsLabel` → `labelsRef[index]`) but the new comment only justifies the third. Either restore a brief mention of the value/label arms, e.g.: *"Browsers can autofill either the serialized value, the serialized label, or the rendered text (e.g. 'United States'); the last one needs `labelsRef`, so we also check it."*

## Suggestions

4. **Add typeahead-on-disabled-forced-open test** — currently the disabled/readOnly guard test at `SelectRoot.test.tsx:1003` only exercises the mouse-click path. Typeahead is already gated upstream via Floating UI's `enabled: !readOnly && !disabled` (`SelectRoot.tsx:384`), but a test would prevent regression if someone refactors that flag.

5. **Assert eventual scroll-arrow unmount** — `SelectRoot.test.tsx:1722` reaches `data-ending-style` but never asserts that the arrow is later removed from the DOM. A regression where `setMounted(false)` never fires (e.g., `useOpenChangeComplete` callback breaks) would not be caught. Add `await waitForElementToBeRemoved(...)` after the 100ms keyframe.

6. **Replace fragile `screen.getByText('▼')` query** — `SelectRoot.test.tsx:1714`. Querying by the literal arrow glyph silently breaks if the default children change; prefer a data attribute or test id.

7. **Defensive try/catch around `stringifyAsValue` in `hasSelectedValue`** — `SelectRoot.tsx:199`. If a user-supplied `itemToStringValue` throws on a particular value shape, it crashes render with a generic React error. A dev-mode console.error pointing to the user's `itemToStringValue` would help. Low priority.

## Behavior Changes Worth a Changelog Note

- **Autofill no longer marks dirty when re-matching the current value** — the old code marked dirty whenever `matchingValue !== validityData.initialValue`; the new code only marks dirty when the value actually changes. More correct, but a narrow behavior change.
- **`isSelected` no longer falls back on `selectedIndex === index`** — object-value users without a custom `isItemEqualToValue` who mutated the value in place will see items deselect. This is the right fix (the PR explicitly targets the stale-index bug), but worth calling out.
- **`SelectScrollArrow` with `keepMounted=false`** — the arrow now participates in the exit transition (`mounted || keepMounted`) rather than unmounting the instant `visible` flips false. Intentional and the entire point of the fix.

## Strengths

- `isSelected` selector change is genuinely correct: value is now the single source of truth; the old early return masked the stale-index bug.
- `labelsRef`/`valuesRef` index parity verified — both refs use the same `useCompositeListItem` index, so the new positional `labelsRef[index]` lookup is safe.
- Lock-state guard is placed correctly: the click path on `SelectItem` is the only commit path that bypasses Floating UI's `enabled:` gates. Click also covers keyboard Enter (synthetic click).
- `useValueChanged` chain confirmed to fire **both** `setDirty` AND `validation.change` on real value transitions, and correctly skips them when `setValue` is canceled.
- Negative assertions added in tests (`not.toHaveAttribute('data-selected')`) lock in the exact regression that was fixed.
- The animation test injects its own keyframe rather than depending on app CSS — good isolation.
- All AGENTS.md rules (`useIsoLayoutEffect`, `useStableCallback`, owner utilities, no raw `setTimeout`) are followed in the changed files.

## Recommended Action

1. Address #1 (microtask unmount guard) — small, safe addition.
2. Decide on #2 (multiple-mode `data-filled` semantics) — fix or document.
3. Adopt #3 (comment expansion) — quick wording tweak.
4. Optional but low-cost: add tests #4 and #5.
5. Add changelog notes for the three behavior changes above.

PR #4935, opened 2026-05-28
[tabs] Fix state edge cases

atomiks:
Fixes a few Tabs state edge cases where fallback state could drift from the selected tab or omit state data from customized parts.

## Changes

- Preserve the initial uncontrolled `defaultValue` when deciding whether an explicitly selected disabled tab should stay selected.
- Expose `data-activation-direction` and `tabActivationDirection` on `Tabs.Tab`.
- Report `null` indicator position and size when the selected value has no matching tab.
- Add regression coverage for these cases plus nearby keyboard and rendered-element behavior.

### DIFF
--- a/docs/src/app/(docs)/react/components/tabs/types.md
+++ b/docs/src/app/(docs)/react/components/tabs/types.md
@@ -258,6 +258,8 @@ type TabsTabState = {
   active: boolean;
   /** The component orientation. */
   orientation: Tabs.Root.Orientation;
+  /** The direction used for tab activation. */
+  tabActivationDirection: Tabs.Tab.ActivationDirection;
 };
 ```
 
--- a/packages/react/src/tabs/indicator/TabsIndicator.test.tsx
+++ b/packages/react/src/tabs/indicator/TabsIndicator.test.tsx
@@ -23,6 +23,37 @@ describe('<Tabs.Indicator />', () => {
     testRenderPropWith: 'div',
   }));
 
+  it('exposes null active tab state when the selected value has no matching tab', async () => {
+    const indicatorStates: Tabs.Indicator.State[] = [];
+
+    function renderIndicator(
+      props: React.HTMLAttributes<HTMLSpanElement>,
+      state: Tabs.Indicator.State,
+    ) {
+      indicatorStates.push(state);
+      return <span data-testid="bubble" {...props} />;
+    }
+
+    await render(
+      <Tabs.Root value="missing">
+        <Tabs.List>
+          <Tabs.Tab value="one">One</Tabs.Tab>
+          <Tabs.Indicator render={renderIndicator} />
+        </Tabs.List>
+      </Tabs.Root>,
+    );
+
+    // Wait for Tabs.List to register its element; before that no tab can be measured.
+    await waitFor(() => {
+      expect(indicatorStates.length).toBeGreaterThan(1);
+    });
+
+    const state = indicatorStates.at(-1)!;
+    expect(state.activeTabPosition).toBe(null);
+    expect(state.activeTabSize).toBe(null);
+    expect(screen.getByTestId('bubble')).toHaveAttribute('hidden');
+  });
+
   describe.skipIf(isJSDOM)('rendering', () => {
     it('should not render when no tab is active', async () => {
       await render(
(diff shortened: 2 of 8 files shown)


flaviendelangle (approved):
# PR Review Summary: #4935 [tabs] Fix state edge cases

## Critical Issues (0 found)
None. All three fixes are correct and ready to merge from a correctness perspective.

## Important Issues (2 found)

### 1. JSDoc Wording Inconsistency
**comment-analyzer + code-reviewer**: `packages/react/src/tabs/tab/TabsTab.tsx:256` and `docs/src/app/(docs)/react/components/tabs/types.md:261` use `"The direction of the activation."` while every other `tabActivationDirection` doc in the codebase (e.g., `TabsRoot.tsx:449`) uses `"The direction used for tab activation."` — divergent JSDoc for the same underlying value across `Root.State` vs `Tab.State` is a maintainability hazard. Mirror the existing wording.

### 2. Misleading Snapshot-Ref Comment
**comment-analyzer**: `packages/react/src/tabs/root/TabsRoot.tsx:45-47` — "`defaultValue` is read-once at mount" is factually inaccurate (the prop is destructured every render at line 32; only `useControlled`'s state-init is once). Also doesn't mention that `useControlled` already warns on the scenario this guards. Suggested rewrite:
> `useControlled` warns when `defaultValue` changes after mount, but the disabled-default-honor policy in the fallback effect below still needs a stable reference — snapshot the initial value so the policy can't shift post-mount even if a consumer ignores the warning.

## Suggestions (nice to have)

### Tests
- **pr-test-analyzer**: Vertical-orientation `data-activation-direction` test at `TabsRoot.test.tsx:1944-1978` was NOT extended like the horizontal one — only horizontal asserts per-tab attribute. A regression breaking vertical-only exposure would slip through.
- **pr-test-analyzer**: No test confirms `tabActivationDirection` is reachable via `<Tabs.Tab render={(_, state) => ...} />`. Only the data-attribute is checked.
- **pr-test-analyzer**: `initialDefaultValueRef` is not tested for the symmetric case — defaultValue=0 (enabled, selected), then changed to defaultValue=1 (where tab 1 is now disabled). The snapshot's value depends on both directions of the policy.
- **pr-test-analyzer**: The new cancellation test (`TabsRoot.test.tsx:650-677`) passes the spec but is loose — `'none'` reason fires for both click and focus activation, so the `.some()` assertion is too permissive.

### Style
- **code-reviewer**: `TabsTab.tsx:196` places `stateAttributesMapping` before `props`; sibling parts (`TabsRoot.tsx:362`, `TabsPanel.tsx:89`, `TabsIndicator.tsx:136`) put it last. Cosmetic.
- **code-reviewer**: `TabsIndicator.test.tsx:46-48`'s `waitFor(() => length > 1)` only succeeds due to Strict Mode double-rendering. Would be more robust to assert on `indicatorStates[0]` directly.
- **comment-analyzer**: Snapshot-ref comment at `TabsRoot.tsx:45-47` lives 200+ lines from its use site at `:303,:310`. Consider co-locating with `shouldHonorDisabledDefaultValueRef` at line 255 where the policy is actually defined.

## Strengths
- The `initialDefaultValueRef` snapshot correctly mirrors `useControlled`'s read-once-at-mount semantics; safe under Strict Mode.
- Removing `defaultValueProp` from effect deps is sound — all internal reads now go through the constant ref.
- TabsIndicator fix is minimally invasive (one hoisted line) and downstream null-aware logic does the right thing.
- TabsTab wiring is consistent with sibling components already using `tabsStateAttributesMapping`.
- The "continues honoring an initially disabled explicit defaultValue" test smartly asserts both the `useControlled` dev warning *and* selection stability via `toErrorDev`.
- The TabsIndicator test asserts the user-visible `hidden` attribute alongside the state values — exactly the regression vector.
- The test comment at `TabsRoot.test.tsx:674` is exemplary — explains *why*, not what.

## Recommended Action
1. Fix JSDoc wording on `tabActivationDirection` to match codebase convention (Important).
2. Rewrite the snapshot-ref comment to remove the "read-once at mount" inaccuracy and add `useControlled`-warning context (Important).
3. Extend the vertical-orientation `data-activation-direction` test for Tabs.Tab attribute symmetry (Suggestion).
4. Consider adding the symmetric `initialDefaultValueRef` test case (Suggestion).

Verification baseline: tests pass, ESLint clean, TypeScript clean, `pnpm docs:api` no drift.

PR #4936, opened 2026-05-28
[scroll area] Fix overflow and scrolling state

atomiks:
This fixes a batch of Scroll Area state bugs where delayed content, threshold updates, and custom scrollbar interactions could leave overflow or scrolling attributes out of sync.

## Changes

- Fixed overflow state when `ScrollArea.Content` mounts after the viewport's first measurement.
- Recomputed overflow edge attributes when `overflowEdgeThreshold` changes.
- Allowed wheel events on scrollbar tracks to chain at scroll edges.
- Marked custom scrollbar wheel and track-click interactions as scrolling.
- Cleaned up thumb drag state on `pointercancel` without releasing stale pointer capture.
- Documented Content data attributes.


flaviendelangle (commented):
# PR #4936 Review Summary

**[scroll area] Fix overflow and scrolling state** — by @atomiks, 6 bug fixes across 11 files (+342/-23).

## Critical Issues (0)

None. All five reviewers found this to be a well-scoped, well-tested fix.

## Important Issues (5)

### 1. Asymmetric `pointercancel` handling on scrollbar track *(silent-failure-hunter)*
**`packages/react/src/scroll-area/scrollbar/ScrollAreaScrollbar.tsx:205`** — The thumb now listens to `onPointerCancel` (via `endDrag`), but the scrollbar element only listens to `onPointerUp`. When the user clicks the track in a track-only configuration (no `ScrollAreaThumb` child), `handlePointerDown` sets `thumbDraggingRef.current = true` and starts scrolling state — but if `pointercancel` fires (browser-cancelled gesture, context menu, focus stolen), neither cleanup runs. The next `pointermove` would mutate viewport scroll erroneously. **This is the same bug the PR fixes for the thumb, applied asymmetrically.** Recommend wiring `onPointerCancel` on the scrollbar to a handler that clears scrolling state + `thumbDraggingRef`.

### 2. Horizontal-orientation thumb pointercancel branch is not tested *(pr-test-analyzer)*
**`packages/react/src/scroll-area/thumb/ScrollAreaThumb.test.tsx:21-89`** only covers the vertical thumb. The fix in `ScrollAreaThumb.tsx:36-44` switches on orientation, but a regression dropping the horizontal branch would still pass the suite. Add a mirrored horizontal-orientation test.

### 3. Horizontal wheel doesn't assert `data-scrolling` *(pr-test-analyzer)*
**`ScrollAreaScrollbar.test.tsx:387-393`** exercises vertical only. Add a horizontal variant — ideally with RTL — to validate the negative-`scrollLeft` math path that's the most likely regression site.

### 4. Wheel-at-edge missing negative `data-scrolling` assertion *(pr-test-analyzer)*
**`ScrollAreaScrollbar.test.tsx:374-385`** verifies `preventDefault` semantics but doesn't assert `expect(scrollbar).not.toHaveAttribute('data-scrolling')` after the chained wheel. A regression that moved `handleScroll(...)` above the edge guard would silently mark the area as scrolling while chaining.

### 5. Data-attribute drift across 4 sources of truth *(type-design-analyzer)*
The new `ScrollAreaContentDataAttributes.ts` is byte-for-byte identical to `ScrollAreaRootDataAttributes` and `ScrollAreaViewportDataAttributes`. The runtime mapping in `root/stateAttributes.ts` only references `Root`; the others are docs-only twins kept in sync by convention. This pre-existed but the PR widens the surface. Optional: alias `Content`/`Viewport` enums to `Root` (verify docs-infra resolves re-exports).

## Suggestions (7)

- **Lost ResizeObserver rationale** — `ScrollAreaContent.tsx:39-43` comment now describes the late-mount case but lost the original "ResizeObserver fires once on observe" context. Consider folding both: *"ResizeObserver fires once on observe; ignore that fire unless content mounted after viewport's initial measurement."* *(code-reviewer + comment-analyzer agree)*
- **"No overflow" wheel-chaining case** — extend the edge-chaining test to cover `maxScroll === 0` (common with `keepMounted`). *(pr-test-analyzer)*
- **Numeric form of `overflowEdgeThreshold`** — recompute test only covers the `{ yStart }` partial-object form, not the single-`number` form. *(pr-test-analyzer)*
- **Symmetric "at start edge" wheel-chaining** — one-liner adding `scrollTop=0, deltaY=-50` would cover the `scrollValue <= minScroll && delta < 0` branch. *(pr-test-analyzer)*
- **`currentOrientationRef` orientation gate** — if anything changes the ref mid-drag (none does today), `releasePointerCapture` is silently skipped. Pre-existing fragility made visible by the new `hasPointerCapture` check. *(silent-failure-hunter)*
- **`queueMicrotask(computeThumbPosition)` non-cancellable** — already safe via `if (!viewportEl) return;` guard, but the new effect runs more often now with threshold deps; consider an effect-scoped flag. *(silent-failure-hunter)*
- **`handlePointerUp` is no longer drag-guarded** — `thumbDraggingRef.current = false` always runs, fine today, but a `if (!thumbDraggingRef.current) return;` early-return would make intent explicit. *(silent-failure-hunter)*

## Strengths

- **Each of the 6 fixes has a targeted regression test** with descriptive DAMP-style names.
- **`hasPointerCapture` guard is exemplary error-aware code** — comment explains *why* (spec behavior, would throw), test mock at `ScrollAreaThumb.test.tsx:73-78` throws if `releasePointerCapture` is called.
- **`fireEvent.wheel(...).toBe(false/true)` cleverly asserts `preventDefault` semantics** via the `dispatchEvent` return value, with an explanatory comment.
- **`computeOnInitialResizeRef` captures `hasMeasuredScrollbar` at mount-time** — correctly handles both early-mount (skip first ResizeObserver fire) and late-mount (compute on first fire) cases.
- **`overflowEdgeThreshold` destructuring fix** is a latent perf win — context value identity is now stable when threshold primitives are stable.
- **Conformance** — `useStableCallback`, `useIsoLayoutEffect`, `useTimeout`, shadow-DOM-safe `contains`/`getTarget`, Vitest-only matchers, `describe.skipIf(isJSDOM)` for layout-dependent assertions — all guidelines respected.
- **Comments respect no-task-reference rule** — no "for the X flow" / "issue #123" anywhere.

## Recommended Action

1. **Important #1 is the most actionable** — wire `onPointerCancel` on the scrollbar element to fully close the loop opened by the thumb fix. The asymmetry is a real (though narrow) bug.
2. **Importants #2-4 are inexpensive test additions** — a horizontal-orientation pointercancel test, a horizontal wheel `data-scrolling` assertion, and a negative `data-scrolling` assertion at edge-chain. Together ~30 lines.
3. **Suggestion** — restore the ResizeObserver rationale in the `ScrollAreaContent.tsx` comment (both code-reviewer and comment-analyzer independently flagged this).
4. **Important #5 (4-way data-attribute duplication) is pre-existing**; defer unless tackling the broader pattern.

flaviendelangle (approved):
# PR #4936 Re-Review Summary

**[scroll area] Fix overflow and scrolling state** — re-review of commit `8de937c2f` (the follow-up commit that addressed my prior review feedback).

## What the follow-up commit addressed

All 5 items from my prior review were addressed:

1. ✅ Wired `onPointerCancel` on the scrollbar element to mirror `onPointerUp` — `ScrollAreaScrollbar.tsx:208`
2. ✅ Added horizontal thumb pointercancel test — `ScrollAreaThumb.test.tsx:91-159`
3. ✅ Added horizontal wheel `data-scrolling` test — `ScrollAreaScrollbar.test.tsx:387-401`
4. ✅ Added negative `data-scrolling` assertion at edge-chain — `ScrollAreaScrollbar.test.tsx:403-412`
5. ✅ Restored "ResizeObserver fires once on observe" rationale — `ScrollAreaContent.tsx:39-42`

## Critical Issues (0)

None. (silent-failure-hunter flagged a multi-pointer scenario, but two concurrent gestures on two scrollbars is sufficiently niche to defer; see Suggestion S6.)

## Important Issues (4)

### I1. The newly-wired scrollbar `onPointerCancel` has no direct test
**`ScrollAreaScrollbar.tsx:208`** — The whole motivation for half of `8de937c2f` is "wire onPointerCancel on the scrollbar track to mirror onPointerUp", but the new tests fire `pointerCancel` only on `<ScrollArea.Thumb>` (which has its own `endDrag`). Removing the new line would still pass the suite. Add a track-only test: `pointerDown` on the scrollbar (not the thumb), `pointerCancel`, then assert a subsequent `pointerMove` does not scroll the viewport — proving `thumbDraggingRef` was cleared. *(pr-test-analyzer + silent-failure-hunter agreed independently.)*

### I2. `delta === 0` wheel events silently call `preventDefault`
**`ScrollAreaScrollbar.tsx:105-117`** — Both edge predicates are false when `delta === 0`, so the handler falls through, calls `preventDefault()`, and writes `scrollValue + 0` back to the viewport. macOS trackpad gestures occasionally emit zero-delta wheels; the page-level wheel chain is silently swallowed for those. One-line fix at top of handler: `if (delta === 0) return;` and add a test. *(silent-failure-hunter)*

### I3. Start-edge wheel chaining is untested
**`ScrollAreaScrollbar.test.tsx:374-385`** — The `preventDefaults only when it consumes the scroll` test covers only `scrollTop=800, deltaY=+50` (end edge). The start-edge predicate `(scrollValue <= minScroll && delta < 0)` is the common "scroll up at the top" UX path. Add one mirror case: `scrollTop=0, deltaY=-50`, `expect(...).toBe(true)`. *(pr-test-analyzer)*

### I4. `overflowEdgeThreshold: number` form is untested
**`ScrollAreaRoot.tsx:388-396`** — `normalizeOverflowEdgeThreshold` has two branches; only the partial-object form is exercised. Documented usage example (`@default 0` + "accepts a single number") could silently break. Add a render-with-`overflowEdgeThreshold={20}` test asserting all four edges respect the threshold. *(pr-test-analyzer)*

## Suggestions (6)

- **S1. Tighten over-claiming comment** — `ScrollAreaScrollbar.tsx:206` says `onPointerCancel: handlePointerUp` "clears the drag state", but `handlePointerUp` only clears `thumbDraggingRef` (and conditionally releases capture). Reword to "resets `thumbDraggingRef`" to avoid implying it touches `data-scrolling`. *(comment-analyzer)*
- **S2. Document destructure-and-re-bundle intent** — `ScrollAreaRoot.tsx:46` destructures `{ xStart, xEnd, yStart, yEnd }` from `normalizeOverflowEdgeThreshold(...)` so the memo deps (line 313-316) track primitive values rather than the freshly-constructed object identity. A future "cleanup" reverting to a single variable would silently invalidate the memo each render. One-line comment recommended. *(comment-analyzer)*
- **S3. RTL wheel-chaining uncovered** — The new chaining/scrolling-state tests are all LTR. Add a `direction: 'rtl'` mirror at start/end edges to exercise the `maxScrollValue = 0` and `minScroll = -maxScroll` math. *(pr-test-analyzer)*
- **S4. `maxScroll === 0` (no overflow) wheel path uncovered** — The new comment calls out "(or with no overflow)" but no test exercises a non-overflowing scroll area's wheel handling. Cheap to add. *(pr-test-analyzer + silent-failure-hunter)*
- **S5. Real-browser `hasPointerCapture` test** — Both new pointercancel tests mock `hasPointerCapture: () => false`, which only simulates JSDOM. A `describe.skipIf(isJSDOM)` Chromium test would validate the actual browser path where pointercancel implicitly releases capture. *(silent-failure-hunter)*
- **S6. Multi-pointer race (defer)** — `currentOrientationRef`/`thumbDraggingRef` are single scalars; concurrent drags on horizontal + vertical scrollbars can mismatch and leak pointer capture. Real but extremely niche; document a single-pointer constraint or convert to `Map<pointerId, ...>`. Out of scope for this PR. *(silent-failure-hunter)*

## Strengths

- **All 5 prior review items addressed cleanly**, with the new tests using the established `Object.defineProperty` mocking pattern.
- **`hasPointerCapture` guard is exemplary defensive code** — comment explains spec behavior, tests install throwing `releasePointerCapture` stubs to prove the guard is load-bearing.
- **`computeOnInitialResizeRef = useRef(hasMeasuredScrollbar)`** correctly captures mount-time state in both early- and late-mount flows (walked both flows; `setHasMeasuredScrollbar(true)` at viewport line 141 runs before the `scrollableContentHeight === 0` early-return, so late-mount Content sees `true`).
- **`overflowEdgeThreshold` destructure** breaks context churn from fresh object identity each render — latent perf win.
- **Wheel `fireEvent(...).toBe(false/true)` idiom** cleverly asserts `preventDefault` semantics via the `dispatchEvent` return value, with an inline comment explaining the API.
- **Restored ResizeObserver rationale comment** captures both branches of the predicate accurately.
- **No `useStableCallback`/`useIsoLayoutEffect`/`useTimeout`/shadow-DOM-helper violations** in the diff; commit messages follow `[scope] Imperative summary`; Vitest-only matchers throughout.

## Recommended Action

1. **I1 is the most actionable** — one test (~30 lines) directly verifies the line of production code the follow-up commit was written to add. Without it, removing line 208 would still pass.
2. **I2 is a 1-line code fix + 1-line test** — `delta === 0` short-circuit.
3. **I3 and I4 are cheap test additions** (~10 lines each) closing documented branches.
4. **Defer S6** (multi-pointer race) unless the team wants to tackle the broader pattern.

**Verdict: very close to ship; the I1 + I2 + I3 + I4 additions would close the remaining gaps in ~60 lines total.**

PR #4940, opened 2026-05-29
[toggle] Fix grouped cancelation and JSDoc

atomiks:
This is part of the Codex review sweep

## Changes

- Stop grouped Toggle handling immediately when `onPressedChange` cancels the event.
- Add regression coverage for canceled Toggle and ToggleGroup changes, including the parent callback staying quiet when child cancellation vetoes the change.
- Document `data-disabled` for Toggle and add accessible labels to the ToggleGroup hero demos.


flaviendelangle (approved):
# PR #4940 Review Summary — `[toggle] Fix grouped cancelation and JSDoc`

The fix itself is correct and well-tested for the headline scenarios. Main gap: the new tests don't cover several adjacent code paths that could regress silently. No blocking issues.

## Critical Issues
None.

## Important Issues

**Test coverage gaps** (`packages/react/src/toggle/Toggle.test.tsx`, `packages/react/src/toggle-group/ToggleGroup.test.tsx`):
- No test for cancellation in `multiple` mode — `setGroupValue` (ToggleGroup.tsx:70-76) has a distinct array-mutation branch that's exercised only in multi mode. A regression there would slip through.
- No test for keyboard activation (Enter/Space) — the existing tests use `button.click()`. The repo's convention (see `ToggleGroup.test.tsx:402-426`) tests both paths separately.
- No test asserting `onPressedChange` is called **exactly once** in the cancel case. A future bug that skips or double-fires it would still pass assertions on `aria-pressed`.

## Suggestions

- **Toggle.tsx:90-91 comment wording**: "before the group commits" is slightly imprecise — `setGroupValue` itself fires `onValueChange` before committing. Consider tightening to: *"before `setGroupValue` runs, so canceling here also vetoes `onValueChange` and the group value commit (they share this `details` object)."*
- **Toggle.tsx:101-103 (second `isCanceled` check)**: still meaningful — guards the case where `onValueChange` cancels — but could use a mirroring one-line comment so future readers understand why both checks exist.
- **Pre-existing silent-failure hazard** (not introduced by this PR): if a Toggle inside a ToggleGroup has a falsy `value`, the group commit is silently skipped in production (the dev-only `error()` at Toggle.tsx:52-63 doesn't run in prod). Worth tracking separately.

## Strengths

- The reorder genuinely fixes the bug: pre-PR, canceling in `onPressedChange` couldn't veto the group commit because `setGroupValue` had already run.
- The added comment correctly explains the **WHY** (shared `details` object, cancel ordering), not the WHAT — aligned with AGENTS.md guidance.
- `data-disabled` is already emitted via `state.disabled` (verified by `Toggle.test.tsx:93`); this PR just documents the existing attribute appropriately in `ToggleDataAttributes.ts`, `types.md`, and `page.mdx`.
- `aria-label="Text alignment"` on the ToggleGroup hero demos is a real a11y improvement.
- Tests follow conventions: Vitest matchers, no Chai chains, no unnecessary `flushMicrotasks`, mirror existing describe structure.
- No new silent failures, no broad catches, exception propagation is unaffected.
- AGENTS.md compliance: no `window.setTimeout`/`useCallback`/`useLayoutEffect` violations.

## Recommended Action

1. (Optional) Add a `multiple` mode cancellation test and a keyboard-activation cancellation test — closest to "important" of any finding here.
2. (Optional) Tighten the Toggle.tsx:90-91 comment wording and mirror it above the second `isCanceled` check.
3. Approve and merge — the fix is correct and reasonably covered.

PR #4941, opened 2026-05-29
[checkbox] Fix parent group cancellation and indeterminate state

atomiks:
This fixes Checkbox and CheckboxGroup edge cases where internal handling could continue after a canceled change, or manually indeterminate parent checkboxes could expose the wrong state.

## Changes

- Let `onCheckedChange` cancellation run before CheckboxGroup parent/child handling.
- Keep parent-enabled group state unchanged when a parent or child change is canceled.
- Apply manual parent `indeterminate` state to ARIA and the hidden input state.
- Tighten parent-group tests so child checkbox assertions do not silently skip the default `span` root.

- [P1] Parent checkboxes could be visually mixed while exposing `aria-checked="false"`.
- [P2] `eventDetails.cancel()` was too late for parent-enabled CheckboxGroup changes.
- [P2] Parent-group tests had no-op child assertions.


mj12albert (approved):

flaviendelangle (commented):
# PR #4941 Review Summary

**[checkbox] Fix parent group cancellation and indeterminate state** — https://github.com/mui/base-ui/pull/4941

The PR is small, well-motivated, and the headline fix is correct. All four review agents flagged a small, overlapping set of concerns. The biggest is that **two of the four new tests don't actually exercise the source-code lines they appear to target**, so the regression risk for those lines is unprotected.

## Critical issues

### 1. The new `useCheckboxGroupParent` cancellation tests are misnamed and don't cover the new guards
`useCheckboxGroupParent.test.tsx:253-299` cancels inside the child/parent's own `Checkbox.Root.onCheckedChange`. That triggers the early return at `CheckboxRoot.tsx:242` — *before* `groupOnChange` (the `useCheckboxGroupParent` callback) is ever called. The new `if (!eventDetails.isCanceled)` guards at `useCheckboxGroupParent.ts:75` (parent path) and `:96` (child path) are never hit. Reverting either of those guards would NOT cause any current test to fail.

To actually exercise them, cancellation must come from the **CheckboxGroup**'s `onValueChange`, not from an individual checkbox's `onCheckedChange`. The regression is the parent "ghost-advancing" through its `mixed → on → off → mixed` cycle on a canceled change — the entire point of the PR.

**Suggested test shape (add two tests):**
```jsx
const cancelSpy = vi.fn((_, d) => d.cancel());
render(
  <CheckboxGroup allValues={['a','b','c']} onValueChange={cancelSpy}>
    <Checkbox.Root parent data-testid="parent" />
    <Checkbox.Root value="a" />
    <Checkbox.Root value="b" />
    <Checkbox.Root value="c" />
  </CheckboxGroup>,
);
// Click parent once → group cancels → status must stay 'mixed'.
fireEvent.click(screen.getByTestId('parent'));
// Click parent again — should still transition from 'mixed' to 'on', not skip ahead.
fireEvent.click(screen.getByTestId('parent'));
// Assert children flipped to checked (the actual 'mixed → on' transition).
```

A parallel child-path test (click child, group cancels, then click parent and assert correct `all` set is passed) covers `useCheckboxGroupParent.ts:96-99`.

## Important issues

### 2. Brittle DOM selection in the new "manually indeterminate" test
`CheckboxRoot.test.tsx:303-313` uses `document.querySelector<HTMLInputElement>('input[type="checkbox"]')`, which returns the *first* checkbox input in document order. It only happens to be the parent's because of JSX ordering. The sibling test at `:272-279` already uses the conventional pattern in this file:
```ts
const [, input] = screen.getAllByRole<HTMLInputElement>('checkbox', { hidden: true });
```
Use the same here for consistency and robustness against future JSX reorderings.

### 3. Test description mismatch
`useCheckboxGroupParent.test.tsx:253` and `:278` say "lets a parent/child checkbox cancel a parent-enabled group change" but, as noted in #1, they exercise the reorder in `CheckboxRoot.tsx`, not anything specific to `useCheckboxGroupParent`. Consider moving them to `CheckboxRoot.test.tsx` (or renaming) and keep separate tests for the actual `useCheckboxGroupParent` guards.

### 4. Inconsistency check: `Checkbox.Root.onCheckedChange` ordering vs `Toggle`
`CheckboxRoot.tsx:240-250` now calls the consumer `onCheckedChange` **before** the group handler. `Toggle.tsx` calls the group handler first. This is an intentional contract divergence that's not documented anywhere. Either:
- Add a brief JSDoc note on `CheckboxRootProps.onCheckedChange` clarifying that cancellation also prevents the surrounding `CheckboxGroup.onValueChange` from firing, or
- Add a short comment above line 240 explaining the order (consumer first → group second is the new chain of cancellation points).

## Suggestions

### 5. `computedIndeterminate = groupIndeterminate || indeterminate` looks redundant but isn't
At `CheckboxRoot.tsx:119`, `groupIndeterminate` is destructured with a default of `indeterminate`. A future maintainer will reasonably try to simplify the `|| indeterminate` away. It isn't redundant: `getParentProps()` always returns a boolean for `indeterminate`, so the destructuring default never fires for grouped parents. The `|| indeterminate` exists specifically so a parent in a non-mixed group still honors a manually-set `indeterminate` prop. Consider a one-line comment.

### 6. `uncontrolledStateRef.current = newValue` moved inside the cancel guard
`useCheckboxGroupParent.ts:94-99` — moving the ref update behind `!eventDetails.isCanceled` is the right call but is subtle. A short comment like "scope this so a canceled change doesn't pollute the snapshot used by the next parent click" would protect the invariant from being undone.

### 7. `allOnOrOff` branch in `useCheckboxGroupParent.ts:53-60` doesn't follow the new cancel-check pattern
Today it's safe because that branch has no local state to roll back — `status` is derived from `value`. But the surrounding code (lines 75, 96) consistently gates state mutations behind `!isCanceled`. The next developer adding a side effect here will likely miss the pattern. Either add a `// status/uncontrolledStateRef are derived from value here; nothing to roll back` comment, or flatten the structure to match.

### 8. Pre-existing `splice(-1, 1)` bug surfaces more strongly now
`useCheckboxGroupParent.ts:91`: `newValue.splice(newValue.indexOf(childValue), 1)` silently removes the last element if `indexOf` returns `-1`. Now that `uncontrolledStateRef.current = newValue` is the only writer of the ref, a wrong `newValue` poisons the ref persistently across renders. Pre-existing, but worth guarding while you're in the file. Trivial fix: `if (idx !== -1) newValue.splice(idx, 1)`.

### 9. `uncontrolledStateRef` initialization-once vs externally-changed `value`
`useCheckboxGroupParent.ts:15` initializes the ref to `value` once. If a consumer mutates the controlled `value` from outside, the ref is silently stale. Cancellation now extends the staleness window (any canceled child click leaves it behind). Consider syncing the ref to `value` in a layout effect. Pre-existing issue, but cancellation deepens it.

### 10. Cosmetic: `expect(input).toHaveProperty('indeterminate', true)` at `CheckboxRoot.test.tsx:312`
Inconsistent with the sibling test at `:278` which uses `expect(input.indeterminate).toBe(true)`. Prefer the latter form.

## Strengths

- **Headline fix is correct.** Reordering `onCheckedChange` before `groupOnChange` with paired `isCanceled` checks (`CheckboxRoot.tsx:240-250`) is the right contract: a checkbox-level cancel must abort the group-level write, and previously it didn't.
- **`computedIndeterminate` usage is now consistent.** Both `aria-checked` (`:309`) and `inputRef.current.indeterminate` (`:191`) read from `computedIndeterminate`, fixing a genuine accessibility silent failure: a parent in a fully-checked group could announce `aria-checked="true"` while visually mixed.
- **Effect dependency tightening** (`:196`) correctly subscribes to `computedIndeterminate` instead of `groupIndeterminate`.
- **Pre-existing test bug fixed.** `granny-smith` → `granny-smith-apple` in `CheckboxGroup.test.tsx:679` aligns `allValues` with the child's `value` prop.
- **AGENTS.md compliance is clean** — no new `setTimeout`/`requestAnimationFrame`/`useLayoutEffect`/`document`/`window`. No new "what" comments. No try/catch swallowing errors.
- **The `CheckboxGroup.test.tsx:132-153` cancellation test is well-scoped and meaningful.**

## Recommended action

1. **Must fix:** Add tests that cancel via `CheckboxGroup.onValueChange` (gaps #1) to actually cover `useCheckboxGroupParent.ts:75` and `:96`.
2. **Should fix:** Replace `document.querySelector` in `CheckboxRoot.test.tsx:303-313` with the `getAllByRole` pattern used elsewhere in the file (#2). Rename or relocate the two misnamed tests (#3).
3. **Nice to have:** Decide on documenting the consumer-vs-group ordering contract (#4) and add 1-2 short WHY comments at the non-obvious spots (#5, #6, #7).
4. **Pre-existing, optional:** Guard the `splice(-1)` case (#8) and consider syncing the uncontrolled ref to `value` (#9).

PR #4942, opened 2026-05-29
[navigation menu] Fix interaction, value, and styling-hook bugs

atomiks:
This is part of the Codex sweep for Navigation Menu. It restores the local fixes from the recovered branch while preserving newer upstream changes.

## Changes

- Prevented disabled triggers from opening through hover, click, or keyboard paths.
- Avoided duplicate `onValueChange` calls when switching items via keyboard.
- Treated non-nullish falsy item values as valid open values.
- Exposed the documented `actionsRef.unmount()` action.
- Pinned left-side popup transitions to the correct edge.
- Documented the missing Navigation Menu styling hooks.

- Disabled triggers could still activate their content.
- Keyboard activation could emit duplicate value changes.
- Falsy item values were handled like closed values in root cleanup.
- The documented imperative unmount action was not wired up.
- Popup and data-attribute styling hooks were incomplete.

### DIFF
--- a/docs/src/app/(docs)/react/components/navigation-menu/demos/hero/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/navigation-menu/demos/hero/css-modules/index.tsx
@@ -27,7 +27,7 @@ export default function ExampleNavigationMenu() {
           </NavigationMenu.Content>
         </NavigationMenu.Item>
 
-        <NavigationMenu.Item value="b">
+        <NavigationMenu.Item>
           <NavigationMenu.Trigger className={styles.Trigger}>
             Handbook
             <NavigationMenu.Icon className={styles.Icon}>
--- a/docs/src/app/(docs)/react/components/navigation-menu/types.md
+++ b/docs/src/app/(docs)/react/components/navigation-menu/types.md
@@ -139,6 +139,12 @@ An icon that indicates that the trigger button opens a menu.
 | style     | `React.CSSProperties \| ((state: NavigationMenu.Icon.State) => React.CSSProperties \| undefined)` | -       | Style applied to the element, or a function that&#xA;returns a style object based on the component's state.                                                                                   |
 | render    | `ReactElement \| ((props: HTMLProps, state: NavigationMenu.Icon.State) => ReactElement)`          | -       | Allows you to replace the component's HTML element&#xA;with a different tag, or compose it with another component. Accepts a `ReactElement` or a function that returns the element to render. |
 
+**Icon Data Attributes:**
+
+| Attribute       | Type | Description                                                      |
+| :-------------- | :--- | :--------------------------------------------------------------- |
+| data-popup-open | -    | Present when the navigation menu is open and the item is active. |
+
 ### Icon.Props
 
 Re-export of [Icon](#icon) props.
@@ -361,6 +367,7 @@ Renders a `<nav>` element.
 | :------------------ | :------------------------------------------------------------------------- | :-------------------------------------------------------------------- |
 | data-open           | -                                                                          | Present when the popup is open.                                       |
 | data-closed         | -                                                                          | Present when the popup is closed.                                     |
+| data-anchor-hidden  | -                                                                          | Present when the anchor is hidden.                                    |
 | data-align          | `'start' \| 'center' \| 'end'`                                             | Indicates how the popup is aligned relative to the specified side.    |
 | data-side           | `'top' \| 'bottom' \| 'left' \| 'right' \| 'inline-end' \| 'inline-start'` | Indicates which side the popup is positioned relative to the trigger. |
 | data-starting-style | -                                                                          | Present when the popup is animating in.                               |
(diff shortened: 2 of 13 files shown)


flaviendelangle (approved):
# PR Review Summary — #4942 [navigation menu] Fix interaction, value, and styling-hook bugs

**Branch:** `codex/navigation-menu-sweep` (5 nav-menu commits, 12 files, +324/−28)
**Reviewers:** code-reviewer · pr-test-analyzer · silent-failure-hunter · comment-analyzer
**Verdict:** Fixes are sound; the gaps are almost entirely in **test coverage**.

## Critical Issues (0 found)
None. The four claimed fixes (disabled-trigger guards, `onValueChange` dedup, falsy item values, `actionsRef.unmount()`) all hold up under inspection, and no silent failures, swallowed errors, or improper fallbacks were introduced.

## Important Issues (5 found — all in test coverage)

1. **Disabled-trigger touch path uncovered** — `NavigationMenuRoot.test.tsx:1820-1866`
   The source guard at `NavigationMenuTrigger.tsx:700-702` covers all pointer types, but the suite only exercises mouse/keyboard. Add a `pointerDown`/`pointerUp` with `pointerType: 'touch'` (pattern at `:1188-1262`).

2. **Falsy value coverage incomplete** — `NavigationMenuRoot.test.tsx:1587-1618`, `2637-2656`
   Only `value={0}` is tested, but the predicate change at `NavigationMenuRoot.tsx:157,169,178` applies equally to `''` and `false`. Parametrize with all three (`it.each`).

3. **Duplicate `onValueChange` test doesn't faithfully traverse** — `NavigationMenuRoot.test.tsx:1561-1585`
   The test programmatically focuses `trigger2` before pressing `ArrowDown`, so the listNavigation traversal path the guard at `NavigationMenuTrigger.tsx:663` actually targets isn't exercised. Use real `ArrowRight`/`ArrowLeft` across triggers with the popup open.

4. **`actionsRef.unmount()` not exercised mid-transition or while open** — `NavigationMenuRoot.test.tsx:1868-1894`
   The test only calls `unmount()` after a clean close. The very scenario that justifies the API — unmount while open or mid-animation — isn't covered.

5. **`side="left"` test is layout-dependent in jsdom** — `NavigationMenuRoot.test.tsx:1896-1925`
   Without real bounding rects, `useAnchorPositioning` may resolve to its default side, silently passing the negative case. Either wrap with `describe.skipIf(isJSDOM)` or assert via `data-side="left"`. RTL `inline-start`/`inline-end` branch at `NavigationMenuPopup.tsx:49-53` is also entirely untested.

## Suggestions (6 found)

6. **Hover test uses `fireEvent.mouseEnter` instead of `pointerEnter`** — `NavigationMenuRoot.test.tsx:1821-1834`
   `useHoverFloatingInteraction` requires `pointerEnter` with `pointerType: 'mouse'` (pattern at `:1192,1222`). Current test likely passes only because `disabled` short-circuits `interactionsEnabled`, not because the hover code path runs.

7. **"Suppress initial open transitions" undertested** — `NavigationMenuRoot.test.tsx:1507-1540`
   Asserts `transition: 'none'` immediately after click but never asserts it's cleared on a subsequent tick. A regression that pins `instant: true` forever would pass.

8. **Nested falsy `setValue(null)` propagation untested** — `NavigationMenuRoot.tsx:177-184`
   New nested branch using `nextValue == null` has no regression test.

9. **Positioner resize-listener re-subscription is fragile** — `NavigationMenuPositioner.tsx:139-162`
   Adding `instant` to deps causes the resize listener to unsubscribe/re-subscribe on every flip. Works only because `handleResize` does `flushSync(setInstant(true))` *before* `instantTimeout.start(100, …)`. Either add a comment explaining the ordering invariant, or split the effects.

10. **Three comment-clarity nits**
    - `NavigationMenuRoot.tsx:214` — `useImperativeHandle` for `actionsRef` is silent about the contract that consumer opt-in disables auto-unmount via `enabled: !actionsRef` on both `useOpenChangeComplete` calls. One line of "why" warranted.
    - `NavigationMenuTrigger.tsx:661-663` — comment cites `onKeyDown` but the guard sits inside `handleActivation`; clarify.
    - `NavigationMenuList.tsx:99-101` — comment about "skipping the onKeyDown handler" describes work done earlier at line 72, not here.

11. **Public docs likely need regeneration** — `docs/src/app/(docs)/react/components/page.mdx:1102`
    The Trigger prop list does not include `disabled`. This PR is the first that makes `disabled` materially observable on the Trigger. Run `pnpm docs:api` to regenerate.

## Strengths

- All four claimed fixes verified correct; `nextValue == null` replaces `!nextValue` consistently in every site (Root lines 157, 169, 178), and the trigger-side guards (`NavigationMenuList.tsx:44-46`, `NavigationMenuTrigger.tsx:130-132`) mirror that.
- Disabled-trigger guard is layered correctly: `&& !disabled` on the interaction hooks, an early-return in `handleOpenEvent`, and `useButton`'s own keyboard short-circuit one layer down.
- `actionsRef` wiring is null-safe (no error swallowing in `handleUnmount`).
- New `data-attribute` files (`anchorHidden` on Popup, `popupOpen` on Icon, typed `activationDirection`) are accurately wired and match `types.md` + `page.mdx`.
- Tests use `vi.fn()` with call-count assertions, `flushMicrotasks()`, and `await act(...)` deterministically.
- No `window.setTimeout`, `requestAnimationFrame`, raw `document`/`window`, or `React.useLayoutEffect` were introduced — AGENTS.md guidelines followed.

## Recommended Action

1. **Tests are the only blocker.** Add coverage for items 1–5 before merging — they're the meaningful regressions a future change could re-introduce undetected.
2. Apply comment clarifications (item 10) and run `pnpm docs:api` (item 11) — both are mechanical.
3. Address suggestions 6–9 opportunistically.

PR #4943, opened 2026-05-29
[composite] Make grid navigation tree-shakeable

atomiks:
## Summary

`useCompositeRoot`/`CompositeRoot` bundled the full grid-navigation algorithm into every composite consumer, even though no consumer passes grid props. Grid navigation is only used by floating-ui's `useListNavigation` (Combobox), which imports the grid helpers directly from a separate, untouched module — so the grid code was dead weight in `tabs`, `toolbar`, `radio-group`, `menubar`, `toggle-group`, and `navigation-menu`.

Rather than dropping grid support, this makes it **tree-shakeable** via dependency injection:

- Extract the grid algorithm out of `useCompositeRoot` into `internals/composite/root/gridNavigation.ts`, exported as `gridNavigation` together with the `CompositeGridConfig` / `CompositeGridItemSize` types.
- Replace the flat `cols` / `itemSizes` / `dense` props on `CompositeRoot` and `useCompositeRoot` with a single optional `grid?: CompositeGridNavigator` — an opaque navigator function produced by the `gridNavigation(config)` factory. The static grid config (`cols`, `dense`, `itemSizes`) is captured in the closure; runtime navigation state, including `onLoop`, is still supplied by `useCompositeRoot`.
- Keep `onLoop` as a top-level `CompositeRoot` / `useCompositeRoot` prop so flat lists and grid consumers share the same loop customization hook.
- `useCompositeRoot` calls the caller-supplied `grid` navigator instead of importing the helpers directly. Consumers that don't pass `grid` never reference the algorithm, so bundlers tree-shake it out.
- Remove the now-unused grid re-exports and the `Dimensions` interface from `internals/composite/composite.ts`.

No current consumer passes `grid`, so every composite consumer drops the grid mass. Opt-in grid navigation stays available by passing `grid={gridNavigation({ cols, ... })}`.

## Behavior fix

While extracting the algorithm, the spanning-item corner selection was made direction-aware: the forward corner is now chosen for `ArrowLeft` in RTL (previously only the physical `ArrowRight` was treated as forward). Navigating from an item that spans multiple columns in RTL no longer resolves back to itself. Covered by a new RTL test.

## Breaking change

This changes the `CompositeRoot` / `useCompositeRoot` API, so any Composite grid consumer must update. The flat grid layout props are replaced by a single `grid` navigator built with the `gridNavigation` factory, while `onLoop` remains a top-level prop:

```diff
-<CompositeRoot cols={3} dense itemSizes={sizes} onLoop={onLoop}>
+<CompositeRoot grid={gridNavigation({ cols: 3, dense: true, itemSizes: sizes })} onLoop={onLoop}>
```

`gridNavigation` is exported from `@base-ui/react/internals/composite`. Importing and calling it is what keeps the algorithm out of consumers that don't use grid navigation.

## Bundle size

Measured with `pnpm size:snapshot` against `master`.

| Entrypoint | Parsed | Gzip |
|---|---:|---:|
| `@base-ui/react/radio-group` | `-3,363 B` | `-1,298 B` |
| `@base-ui/react/navigation-menu` | `-3,300 B` | `-1,294 B` |
| `@base-ui/react/menubar` | `-3,355 B` | `-1,291 B` |
| `@base-ui/react/toolbar` | `-3,354 B` | `-1,283 B` |
| `@base-ui/react/toggle-group` | `-3,354 B` | `-1,263 B` |
| `@base-ui/react/tabs` | `-3,355 B` | `-1,258 B` |
| `@base-ui/react` | `-525 B` | `-249 B` |
| `@base-ui/react/internals/composite` | `+201 B` | `+36 B` |

The `internals/composite` entrypoint grows slightly because it now exposes `gridNavigation` as a dedicated export.


flaviendelangle (dismissed):
It is used by the Calendar component
But we can probably refactor the Composite to avoid bundling the logic for components that don't need it.

atomiks:
> But we can probably refactor the Composite to avoid bundling the logic for components that don't need it.

Can make it a treeshakeable prop instead. I'll refactor

flaviendelangle (commented):
WIP

PR #4946, opened 2026-05-29
[combobox] Clear value on Escape from Trigger in input-inside-popup

### LINKED ISSUE #4938: [combobox] clearing value using the input-inside-popup pattern
# Bug report

## Current behavior

When you build a Combobox that uses the [input-inside-popup](https://base-ui.com/react/components/combobox#input-inside-popup) pattern, the ESC key does not clear the selection. 

Clearing the value of the Combobox with the ESC key is currently managed by the `Combobox.Input`. If you are using `Combobox.Input` with the popup pattern, the behavior closes the popup, which is expected. If youre using this outside the popup, it also clears the value. However, there doesnt seem to be a clean way to clear the value when using a `Combobox.Trigger`,  as this component does not have keyboard event handlers for `Escape`.

As a workaround you can render a `Combobox.Clear` element and manually invoke the click action on the `Combobox.Clear` element. 

```tsx
  const clearRef = useRef<HTMLButtonElement | null>(null); 
  const handleKeyDown: KeyboardEventHandler<HTMLButtonElement> = (event) => {
    if (event.key === 'Escape' && event.target === event.currentTarget && clearRef.current) {
      clearRef.current?.click();
      event.stopPropagation();
    }
  };

return (
   <>
      <Combobox.Trigger
        ref={ctx.ref}
        onKeyDown={handleKeyDown}
      />
      <Combobox.Clear ref={clearRef} /> // Style it with display: none
   </>
);
```

This could be by design perhaps. But it does seem odd that the "input-inside-popup" behaves slightly different than the other variants of the Combobox.

## Expected behavior

Pressing the ESC key on a `Combobox.Trigger` clears the input.

## Reproducible example

https://base-ui.com/react/components/combobox#input-inside-popup

## Base UI version

1.5.0

Edge

## Which OS are you using?

Mac OS

Voiceover

## Additional context

Edit: My workaround is too aggressive by using `event.stopPropagation`. This prevents dialogs from closing

spokodev:
Fixes #4938.

In the input-inside-popup pattern (`<Combobox.Trigger>` outside, `<Combobox.Input>` inside `<Combobox.Popup>`), `Combobox.Trigger` was the keyboard entry point when the popup was closed, but it had no handler for `Escape`. So users could not clear a selected value without first opening the popup, focusing the Input, and pressing `Escape` there. The issue author posted a manual workaround that simulated a click on `Combobox.Clear`.

This adds an `Escape` branch to `Combobox.Trigger`'s `onKeyDown` that mirrors the existing clear path in `ComboboxInput.tsx` (around line 399): it only fires when `inputInsidePopup` is true and the popup is not mounted, calls `setSelectedValue` with `null` (single) or `[]` (multiple), and stops propagation only when there was something to clear, not inline, and propagation was not explicitly allowed. `disabled` and `readOnly` are respected via the early return at the top of the handler.

`Combobox.Input` in the outside-popup pattern keeps its existing behavior. The Trigger handler only activates when the popup is closed, so when the popup is open the Escape goes through the Input as before.

11 new tests in `ComboboxTrigger.test.tsx` cover:

- Two positive cases (single value cleared to `null`, multi to `[]`)
- Three negative cases (Tab on Trigger no-op; Trigger in input-outside-popup no-op; Trigger Escape while popup is open does not duplicate Input's clear)
- Six edge cases (no value selected, Trigger `disabled`, `Combobox.Root disabled`, `Combobox.Root readOnly`, controlled value flow, `details.reason === REASONS.escapeKey`)

All four fail on master and pass on this branch.

## Verification

```
pnpm test:jsdom ComboboxTrigger --no-watch
# 54 passed | 1 skipped (55)

pnpm test:jsdom Combobox --no-watch
# 685 passed | 17 skipped (702) across 25 test files

pnpm typescript
# clean

pnpm eslint packages/react/src/combobox/trigger/ComboboxTrigger.tsx packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
# clean

pnpm prettier --check packages/react/src/combobox/trigger/
# clean
```

Open as draft until CI confirms.

### DIFF
--- a/packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
+++ b/packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
@@ -1,3 +1,4 @@
+import * as React from 'react';
 import { expect, vi } from 'vitest';
 import { Combobox } from '@base-ui/react/combobox';
 import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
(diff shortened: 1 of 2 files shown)


michaldudak:
@spokodev, are you planning to continue working on this PR?

spokodev:
Yes — sorry, this was left as a draft by mistake. It's complete and ready for review now.

It mirrors `ComboboxInput`'s Escape-to-clear onto the `Trigger` for the input-inside-popup pattern, where the Trigger is the keyboard entry point while the popup is closed. Single-value clears to `null`, multiple to `[]`, with the same `escapeKey` change reason. Covered by tests for both modes plus negatives (Tab doesn't clear, and the input-outside-popup pattern is unchanged).

Let me know if you'd like it rebased or anything adjusted.

atomiks:
# PR review

This PR adds an `Escape`-to-clear branch to `Combobox.Trigger` for the input-inside-popup pattern (#4938), closely mirroring the existing `ComboboxInput` clear path. The core change is sound and correctly guarded (`disabled`/`readOnly` early return, `!mounted` to avoid double-handling while the popup is open, `inputInsidePopup` scoping). Nothing here is merge-blocking. The findings are a spurious change event on the empty-Escape path, two weak/mislabeled tests, and the growing duplication of the clear logic.

## Bugs (1)

### 1. 🟡 Escape on the Trigger fires `onValueChange` even when there is nothing to clear

**Location:** `packages/react/src/combobox/trigger/ComboboxTrigger.tsx:284`

```tsx
const isClear =
  selectionMode === 'multiple' && Array.isArray(selectedValue)
    ? selectedValue.length === 0
    : selectedValue === null;

const details = createChangeEventDetails(REASONS.escapeKey, event.nativeEvent);
const value = selectionMode === 'multiple' ? [] : null;
store.state.setSelectedValue(value, details); // called unconditionally
```

`setSelectedValue` invokes `onSelectedValueChange` (→ the public `onValueChange`) unconditionally — it does not diff against the current value (see `AriaCombobox.tsx`). So pressing `Escape` on the Trigger when nothing is selected still fires `onValueChange(null, { reason: escapeKey })`, and in multiple mode fires `onValueChange([])` with a fresh array identity. `isClear` is computed but only gates `stopPropagation`, not the call itself.

This mirrors `ComboboxInput`'s existing behavior, so it's arguably consistent-by-design, but it's a real wart: a `Combobox.Trigger` inside a `Dialog` will emit a spurious `null`/`[]` change on the same `Escape` the user presses to dismiss the dialog, which can mark a pristine form dirty, trigger analytics/side effects in the handler, or force a re-render on a controlled multiple-value combobox.

**Failure scenario:** Trigger with no selection, focused, user presses `Escape` → `onValueChange(null, {reason: 'escape-key'})` fires despite no change.

**Fix:** Only clear (and stop propagation) when there is something to clear:

```tsx
if (!isClear) {
  store.state.setSelectedValue(value, details);
  if (!store.state.inline && !details.isPropagationAllowed) {
    event.stopPropagation();
  }
}
```

This also makes the "is a no-op when no value is selected" test (below) actually true.

## Tests (2)

### 1. 🟠 "is a no-op when no value is selected" asserts nothing and is mislabeled

**Location:** `packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx:991`

```tsx
await user.keyboard('{Escape}');

// Setting null when already null still calls onValueChange but does not throw.
const calls = onValueChange.mock.calls;
if (calls.length > 0) {
  expect(calls[0][0]).toBe(null);
}
```

The assertion is behind an `if`, so it may never run — the test passes whether `onValueChange` is called or not, and even if the clear path were removed entirely for the empty case. It pins no behavior. It's also misnamed: as shown in Bug 1, the code is *not* a no-op — it fires `onValueChange(null)`. The comment concedes this. Either fix the code to be a genuine no-op (Bug 1) and assert `expect(onValueChange).not.toHaveBeenCalled()`, or rename the test and assert the actual spurious-call behavior unconditionally.

**Failure scenario:** A regression that stops the empty-Escape path from behaving as intended would not be caught — the conditional assertion short-circuits.

### 2. 🟡 "does not duplicate clear when popup is open" uses a loose assertion and a wrong comment

**Location:** `packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx:973`

```tsx
await user.keyboard('{Escape}');
// The Trigger-side clear should not have fired a second time after Input handles it.
// We accept at most one onValueChange call (from Input's clear path), not two.
expect(onValueChange.mock.calls.length).toBeLessThanOrEqual(1);
```

When the popup is open, `mounted` is `true`, so `ComboboxInput`'s Escape-clear branch (`!mounted && event.key === 'Escape'`) does **not** run — Escape just dismisses the popup and the value is retained. So the actual call count here is `0`, not "one from Input's clear path" as the comment claims. The `<= 1` bound passes vacuously and the comment misdescribes the mechanism. Assert the real expectation (`expect(onValueChange).not.toHaveBeenCalled()` after opening then pressing Escape) and correct the comment.

**Failure scenario:** If the Trigger *did* start double-firing (e.g. a future change to the `!mounted` guard), `toBeLessThanOrEqual(1)` would still pass for the 1-call case, so the "no duplication" intent isn't fully pinned.

## Simplifications (1)

### 1. 🟡 Clear logic is now duplicated across three components with subtle divergence

**Location:** `packages/react/src/combobox/trigger/ComboboxTrigger.tsx:277`

The `isClear` computation + `setSelectedValue([] / null)` + `stopPropagation` block is now copy-pasted from `ComboboxInput.tsx`, and a third variant lives in `ComboboxClear.tsx`. The three have already drifted:

- `ComboboxClear`: `setInputValue('')` + `setSelectedValue` + `setIndices(...)`
- `ComboboxInput`: `setInputValue('')` + `setSelectedValue`
- `ComboboxTrigger` (new): `setSelectedValue` only

The Trigger's reduced set happens to be correct (the input is unmounted and already cleared to `''` by `handleUnmount` in this pattern), but AGENTS.md calls for sharing such logic through the root/context rather than re-implementing it. A single `store.state.clearSelectedValue(details)` helper on the root that all three call would remove the divergence risk — the next change to clear semantics would otherwise need to be made in three places and is easy to miss in one.

**Failure scenario:** A future fix to clear behavior updates `ComboboxInput`/`ComboboxClear` but not the Trigger (or vice-versa), silently reintroducing the exact inconsistency this pattern already shows.

## Docs (0)

No findings.

## Verdict

**Approve after nits** — no correctness or accessibility regression; the deciding factors are the vacuous/mislabeled "no-op" test and the spurious `onValueChange`, both worth tightening before merge.

---

spokodev:
Thanks for the review — fixed both deciding nits:

- The empty-Escape path now guards on `!isClear`, so Escape with nothing selected is a true no-op (no `onValueChange`).
- Tightened the two tests: the no-op case now asserts `onValueChange` isn't called, and the popup-open case asserts the popup closes and the value is preserved (dropped the vacuous `<= 1` bound and the mislabeled comment).

On the shared `clearSelectedValue` helper: the three paths have diverged beyond the duplication — `ComboboxClear` also resets the active/selected indices, refocuses the input, and uses `clearPress` rather than `escapeKey`, and the `isClear` guard + `stopPropagation` act on the synthetic event a store helper wouldn't have. Unifying them would change `ComboboxInput`/`ComboboxClear` behaviour, so I've kept this PR to the Trigger and would rather do that as a deliberate follow-up. Happy to pick it up if you'd like.

atomiks:
# PR review

I found one merge-blocking controlled-state regression. The new Escape tests pass, but a focused controlled `inputValue` probe fails; this also covers the `inputInsidePopup` edge for Positioner and inline/dialog-style compositions.

## Bugs (1)

### 1. 🔴 controlled inputValue is not cleared from the trigger Escape path

**Location:** `packages/react/src/combobox/trigger/ComboboxTrigger.tsx:287`

```tsx
const value = selectionMode === 'multiple' ? [] : null;
store.state.setSelectedValue(value, details);
```

This mirrors only half of `ComboboxInput`’s Escape behavior. The existing input path calls `setInputValue('', details)` before `setSelectedValue(...)`, but the new trigger path only clears the selected value. In a controlled `inputValue` combobox, the parent never receives `onInputValueChange('')`, so reopening the popup shows the stale query.

**Failure scenario:** A controlled input-inside-popup combobox has `value="apple"` and `inputValue="apple"`. Pressing Escape on the closed trigger clears the visible selected value, but reopening the popup still shows `apple` in the input and keeps the stale filter.

**Fix:** Use the same event details to call `store.state.setInputValue('', details)` in the trigger Escape branch, before `setSelectedValue(...)`, and add a controlled `inputValue` test.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Request changes** - the new trigger clear path breaks the controlled `inputValue` contract.

---

🤖 Review generated with Codex

spokodev:
Fixed. The Trigger Escape path now calls `setInputValue('', details)` before `setSelectedValue`, matching `ComboboxInput`, so a controlled `inputValue` gets `onInputValueChange('')` and a reopened popup no longer shows the stale query. I kept `setSelectedValue` behind the `!isClear` guard so an already-empty value doesn't fire a spurious `onValueChange` (per the earlier review). Added a controlled-`inputValue` regression test.

spokodev:
Rebased onto master. Two things the refactor since June needed:

- `store.state.setInputValue` / `setSelectedValue` moved to `store.context`, matching what `ComboboxInput` does now.
- `const mounted = store.useState('mounted')` and the `disabled`/`readOnly` guard had been dropped from the Trigger; both are restored, so the `!mounted` gate and the guards you reviewed still hold.

75/75 in `ComboboxTrigger.test.tsx`, 978 across the combobox suite, `tsc -b` and Prettier clean.

PR #4948, opened 2026-05-30
[combobox] Keep ArrowLeft/ArrowRight on the input caret in grid mode

### LINKED ISSUE #4947: [autocomplete] AutocompleteInput cursor navigation when using grid layout
# Bug report

## Current behavior

The AutocompleteInput cursor navigation using your keyboard is being highjacked by the AutocompleteList when using a grid layout.

In this demo, I click the autocomplete trigger, type "smi", and then press my left arrow key to navigate to the left on what I typed. However, it actually navigates back on the autocomplete items. You can see that clearly because it starts at the last item of the list.

https://github.com/user-attachments/assets/12052734-7185-4309-b3f9-571d3adfa9a8

## Expected behavior

If none of the items are highlighted, the left and right arrows should navigate the input. If the user already navigated down to the items, or autoHighlight is used, the left and right arrows should navigate the items.

## Reproducible example

[StackBlitz](https://stackblitz.com/edit/mpzppufs?file=src%2FApp.tsx)

## Base UI version

1.5.0

Chrome

## Which OS are you using?

Mac OS

spokodev:
Fixes #4947.

Preview: https://deploy-preview-4948--base-ui.netlify.app/react/components/autocomplete#grid-layout

In grid mode `useListNavigation` is configured with `cols: 2` and `orientation: 'horizontal'`, so it reacts to ArrowLeft/ArrowRight on the reference element. The reference element is the input. When the input has focus and no item is highlighted yet, the user is still editing the query, and the arrow keys silently jump the highlight into the list (typically looping into the last cell) instead of moving the caret.

The fix adds a guard props object to the input's `mergeProps` chain, placed right after `listNavigation.reference`: while `activeIndex == null` in grid mode, ArrowLeft/ArrowRight call `event.preventBaseUIHandler()` so the navigation handler is skipped and the input keeps its native caret behavior. Once an item is highlighted (via ArrowDown, `autoHighlight`, or earlier grid navigation) the event reaches the navigation hook as before and grid traversal continues unchanged. The guard reads `store.state.activeIndex` non-reactively at event time, so no extra reactive dependencies are introduced.

This matches the ARIA APG "Editable Combobox with Grid Popup" pattern: Left/Right operate on the caret while no suggestion is focused, and move between cells once one is.

### Tests

Added three cases under `prop: grid` in `ComboboxRoot.test.tsx`, two RED on master, one regression-preserve:

- `moves the input caret on ArrowLeft when no item is highlighted in grid mode` (RED on master)
- `moves the input caret on ArrowRight when no item is highlighted in grid mode` (RED on master)
- `keeps grid navigation when autoHighlight surfaces an item before typing arrow keys` (pins the only path where an item becomes highlighted without arrow keys)

Horizontal navigation with a highlighted item is already covered by the existing `arrow keys navigate across rows and columns in grid mode` test. Existing grid tests (`arrow keys navigate across rows and columns in grid mode`, `mirrors horizontal grid navigation in RTL mode`, `supports uneven rows navigation`) still pass, along with the rest of the Combobox suite and the Autocomplete suite.


packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -1086,8 +1086,28 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
...
+        listNavigation.reference,
+        {
+          onKeyDown(event: BaseUIEvent<React.KeyboardEvent>) {
+            // In grid mode the navigation hook treats ArrowLeft/ArrowRight as horizontal
+            // grid movement. When the input has focus and no item is highlighted the user
+            // is still editing the query, so let the input keep its native caret behavior.
+            if (
+              grid &&
+              store.state.activeIndex == null &&
+              (event.key === 'ArrowLeft' || event.key === 'ArrowRight')
```
atomiks, line 1100: This matches macOS' emoji picker, where <kbd>ArrowLeft</kbd> and <kbd>ArrowRight</kbd> do not work even if the text caret is at the end/boundary of the textbox (e.g. when it's empty). You need to press <kbd>ArrowDown</kbd>/<kbd>ArrowUp</kbd>. I guess that is fine.

atomiks (approved):

PR #4949, opened 2026-05-30
[autocomplete] Improve command palette accessibility

### LINKED ISSUE #4247: [autocomplete] Command palette input not announced by screen readers
# Bug report

## Current behavior

The "command palette" variant has an input at the top of the dialog. This is not announced by VoiceOver. Virtual focus lands on the first option in the list and remains there even when keyboard focus moves to the input. There is no announcement of the autocomplete behaviour or input placeholder text, only of the list of options and that you can use arrow keys to move between them.

In NVDA using Edge, I was able to move to the input but only after moving focus _backwards_ to the close button and then using arrow keys. This is not intuitive and means other screen reader users may not be informed of the available functionality.

## Expected behavior

The input is announced by screen readers. 
Keyboard and screen reader focus can move to the input. 

Ideally, focus should move to the **input** when the dialog opens. This is a combobox with autocomplete behaviour similar to the [Aria Authoring Practices Guide (APG) Editable Combobox With List Autocomplete Example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/) except that it remains in an open/expanded state. 

It should be possible to move focus between options and to/from the input itself.

## Reproducible example

Steps to replicate:
1. Turn on VoiceOver on a Mac
2. Navigate to the [Autocomplete: Command Palette example](https://base-ui.com/react/components/autocomplete#command-palette)
3. Activate the "Open command palette" button
4. Note the input (which has an accessible name) is never announced, nor is the autocomplete behaviour
5. Use the keyboard to try and move focus to/from the input (e.g. using arrow keys, tab and shift+tab)

## Base UI version

v1.2.0

Chrome

## Which OS are you using?

Mac OS

Voiceover 

## Additional context

Removing the `aria-activedescendent` by default may fix the behaviour. 

Ensure the focus order is in a sensible order. E.g. Input > List items > Keyboard commands group > close button

### LINKED ISSUE #4248: [autocomplete] accessibility: Command palette example keyboard shortcuts not announced by screen readers
# Bug report

## Current behavior

On opening the command palette dialog, there are some keyboard shortcuts shown at the bottom. These are not announced by screen readers and wouldn't be unless they switched from forms to reading mode (which is unlikely when the dialog only contains an autocomplete combobox). 

People using a screen reader will almost certainly miss these keyboard shortcuts. 

This could justifiably count as a fail against [Understanding WCAG 2.2 Success Criterion 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html) or possibly [Understanding WCAG 2.2 Success Criterion 3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html) as this content is potentially important/helpful information that is not referenced by the autocomplete.

## Expected behavior

Keyboard commands should be announced to screen reader users. This could be done by referencing them via `aria-describedby` on the input or by placing them inside a named focusable region 

## Reproducible example

Steps to reproduce: 
1. Turn on VoiceOver on a Mac (or other screen reader
2. Navigate to the [Autocomplete: Command Palette example](https://base-ui.com/react/components/autocomplete#command-palette)
3. Activate the "Open command palette" button
4. Move around the dialog content. (e.g. using Tab, Shift+Tab and Arrow keys)
5. Note the keyboard commands are not announced 

## Base UI version

v1.2.0

Chrome

## Which OS are you using?

Mac OS

Voiceover

## Additional context

Note: The text also instructs users to press `cmd+k` when Windows users would need to use `Ctrl`. The content may need to reflect this.

SSDWGG:
## Summary
- focus the command palette autocomplete input when the dialog opens
- give the input an explicit accessible name and describe the available keyboard commands
- make the visible Actions shortcut platform-neutral by showing Ctrl/Cmd + K

## Why
The command palette demo currently opens with the inline autocomplete list already active, and the footer shortcuts are only visible text. Screen reader users can miss the search input and the available command shortcuts. This updates both CSS Modules and Tailwind variants so the demo announces the input and exposes the instructions through aria-describedby.

Closes #4247
Closes #4248

## Tests
- pnpm prettier
- pnpm typescript
- pnpm exec eslint docs/src/app/(docs)/react/components/autocomplete/demos/command-palette/css-modules/index.tsx docs/src/app/(docs)/react/components/autocomplete/demos/command-palette/tailwind/index.tsx --max-warnings 0
- pnpm stylelint docs/src/app/(docs)/react/components/autocomplete/demos/command-palette/css-modules/index.module.css


SSDWGG:
The code/test checks are green; the remaining failing check is `test-label-applied`, which requires at least one PR label. I do not have permission to add labels from this fork; could a maintainer please add `accessibility` when reviewing?

atomiks:
Thanks for the PR @SSDWGG  

Few changes:

- Remove `initialFocus` related changes, the input already receives focus
- Not sure if "Use Control or Command K for command actions." is needed as it's (at least in the example) purely decorative
- `.VisuallyHiddenClose` seems redundant with this change as `.VisuallyHidden` covers it

SSDWGG:
Thanks for the feedback. I've updated the PR to remove the initialFocus/ref changes, remove the Command/Ctrl K footer text, and reuse the existing visually-hidden close styling instead of adding a separate class.

atomiks (approved):
On:

> #4247 — virtual focus lands on the first option, and “removing the aria-activedescendent by default may fix the behaviour.”


`autoHighlight="always"` is intentional and matches command-palette UX. The remaining issue is that VoiceOver/NVDA don't announce the input name/description while the first item remains active upon open, but this is somewhat of out of scope and would require some hacks to get working (e.g. https://github.com/mui/base-ui/pull/4286).

PR #4953, opened 2026-06-01
[preview-card] Close popup when its active detached trigger unmounts

flaviendelangle:
When a PreviewCard used with createHandle() + multiple detached Triggers had its active trigger removed from the DOM (e.g. virtualized out by scroll), the store kept a stale reference to the detached node. floating-ui's autoUpdate kept calling getBoundingClientRect() on it (returning {0,0,0,0}) and the popup jumped to the top-left of the viewport.

useTriggerDataForwarding now clears activeTriggerId / activeTriggerElement when the active trigger detaches from the document, and additionally closes the popup when the new closeOnActiveUnmount opt-in is passed. PreviewCard opts in (handle !== undefined); the other popup components keep the default so legitimate "stay open after trigger unmount" patterns (Dialog opened from a Menu item, Popover with a disappearing trigger) keep working.

FloatingFocusManager.getReturnElement already falls back to elementFocusedBeforeOpen when the reference is detached, so focus return keeps working for components that just clear references.

[Reproduction case on master](https://stackblitz.com/edit/izmxex8x?file=src%2FApp.tsx)


packages/react/src/utils/popups/popupStoreUtils.ts
```diff
@@ -184,6 +219,25 @@ export function useTriggerDataForwarding<State extends PopupStoreState<unknown>>
...
+      store.select('activeTriggerId') !== triggerId
+    ) {
+      return;
+    }
+    store.update({
+      activeTriggerId: null,
+      activeTriggerElement: null,
+    } as Partial<State>);
+    if (closeOnActiveUnmount && store.select('open')) {
+      store.setOpen(false, createChangeEventDetails(REASONS.none));
```
flaviendelangle, line 237: We can add a new `REASONS.triggerUnmounted` if you want, or your `REASONS.imperativeAction`

flaviendelangle:
Closed in favor of #4886

PR #4955, opened 2026-06-01
[dialog][popover] Reset detached handles on Root unmount

### LINKED ISSUE #4951: [dialog] dialog with createHandle survives navigation and preserve pre-navigation store state
# Bug report
1. Open the dialog with attached handle
2. Unmount the dialog while open (eg in my case - by navigating away)
3. Return to the previous screen and mount the dialog back
## Current behavior

If you try to follow steps above , you will see that dialog stays with state from before navigation.

And another example is straight [base-ui.com - Controlled mode with multiple triggers](https://base-ui.com/react/components/dialog#controlled-mode-with-multiple-triggers). 
Click any detached triggers from an example , navigate away and then navigate back. You won't see opened dialog but you won't be able to do anything on the page either. Definitely something going on with store or handle state clean up.

https://github.com/user-attachments/assets/ccf67561-29db-4e80-8b34-459b80f39978

## Expected behavior
Feels like it should properly clean-up it's current state , so it behaves how any other dialogs without handle attached.

## Reproducible example

https://stackblitz.com/edit/a9jtaf9i-rxzlgxqh?file=src%2Froute-with-dialog.tsx)

## Base UI version

1.5.0 - but i am pretty sure i encountered the same behaviour on other versions.

Chrome

## Which OS are you using?

Mac OS

-
## Additional context
https://discord.com/channels/1287292451308048406/1507703359660884079

michaldudak:
Fixes #4951.

## What changed

- Reset detached popup-handle state when the last owning Root unmounts. This now applies through the shared popup cleanup path used by Dialog, AlertDialog, Drawer, Popover, Tooltip, and PreviewCard.
- Reuse each component's initial state factory in the reset path while preserving Root-level invariants such as modal, nested, role, and other root-owned configuration.
- Clear Floating UI root-owned refs and synchronized elements so disconnected trigger and popup nodes do not survive navigation or remounts.
- Clear shared hover interaction state, timers, and safe-polygon pointer-event mutations so pending hover opens cannot reopen a handle after its Root unmounts.
- Add regression coverage for shared popup cleanup plus detached-handle unmount/remount flows across Dialog, AlertDialog, Drawer, Popover, Tooltip, and PreviewCard, including Floating UI hover/reset edge cases.

## Why

External handles intentionally outlive the Root so detached triggers can keep controlling the same popup. When an open Root unmounted during navigation, the handle store stayed open with its previous active trigger, Floating UI refs, and in hover-driven cases pending interaction state. Returning to the route could therefore remount into stale open state, reuse disconnected elements, or allow delayed hover logic to reopen a rootless handle.

The shared cleanup waits one microtask before resetting. That allows Strict Mode and keyed same-commit remounts to claim the same store without losing valid controlled state, while still resetting when no Root remains mounted.

## Reproduction

Before: https://master--base-ui.netlify.app/react/components/dialog
After: https://deploy-preview-4955--base-ui.netlify.app/react/components/dialog

1. Open any page in the docs other than the Dialog.
2. Open the Dialog page.
3. Open any Dialog in the Controlled open with multiple triggers demo.
4. Use the browser back button to navigate to the first docs page.
5. Use the browser forward button to navigate to the Dialog again.
6. Verify the page remains interactive.


flaviendelangle (commented):
# PR #4955 Review — `[dialog][popover] Reset detached handles on Root unmount`

**Author:** Michał Dudak · **Fixes:** #4951 · 21 files, +1084/−17

**Overall:** Solid, well-targeted fix. The reference-count + microtask design is correct and robust to React Strict Mode, keyed remounts, and concurrent owners. Regression coverage is strong at three layers (store primitive, Floating UI sub-systems, per-component E2E). No project-guideline violations. **No true blocking issues** — the items below are hardening, comment-accuracy, test-gap, and type-tightening points.

Two findings that one agent rated *Critical* I **downgraded after verifying against the code** — neither affects correctness (details in Suggestions §C).

---

## Important

**1. Deferred `reset()` fires the consumer's `onOpenChange` after the Root unmounts — intended, but undocumented and unguarded.**
`popupStoreUtils.ts:383-385` — the microtask `store.reset()` calls `setOpen(false, …)`, which runs the consumer's `onOpenChange(false, {reason:'none'})` *after* the Root tree unmounted. Two agents converged here.
- This is **intended and load-bearing**: the controlled detached-handle tests rely on it (the parent `TestRoute` stays mounted and its `open` state must flip to `false`). The popup store's `onOpenChange` trampoline is deliberately not cleared on unmount. ✅ Not a bug.
- But (a) it deserves a one-line comment at the `setOpen` call so a future maintainer doesn't mistake the post-unmount callback for a leak; and (b) if a consumer's `onOpenChange` throws, it becomes a **detached unhandled microtask rejection** with no React stack / error boundary. Optional hardening: wrap `store.reset()` in `try { … } catch (e) { setTimeout(() => { throw e; }) }` so it reaches `window.onerror` with context. Minimum action: the comment.

**2. Test gap — real `<React.StrictMode>` is never exercised.** `popupStoreUtils.ts:430` and the PR description both name Strict Mode as *the* reason for the microtask + ref-counting, yet no test wraps anything in `StrictMode`. The two ref-count unit tests simulate the mechanism (two concurrent keyed owners) but not the Strict-Mode mount→cleanup→mount double-invoke that the `?? 1` underflow guard exists for. Add one Strict-Mode test: open a detached Root, unmount, assert the handle still reopens cleanly.

**3. Test gap — closed-root unmount is untested.** `usePopupRootUnmountCleanup` queues `reset()` unconditionally; `resetPopupRootStore` still runs `resetHoverInteraction()` / `update(nextState)` / `floatingRootContext.reset()` even when the popup was closed at unmount. All six component tests open before unmounting, so the common "navigate away with popup closed, navigate back" path has zero coverage.

**4. Test gap — per-store preserved fields aren't asserted across reset+remount.** Dialog preserves `modal`/`disablePointerDismissal`/`nested`/`role`; Popover preserves `modal`/`nested`; Tooltip/PreviewCard preserve only `floatingRootContext`. Only `role` is indirectly covered (AlertDialog reopens via `findByRole('alertdialog')`). `modal`/`disablePointerDismissal` survival is asserted nowhere — a refactor dropping one from the preserved set would pass CI. The Popover/Tooltip `nested` asymmetry (Popover preserves, Tooltip resets) is intentional-looking but unverified.

---

## Suggestions

**A. Test quality**
- `useFloating.test.tsx:52` — `await flushMicrotasks()` immediately after `await render(...)` with no intervening state change violates AGENTS.md (`floatingContext` is set in a layout effect already run by the time `render` resolves). Remove it. *(The other `flushMicrotasks` calls in this PR are legitimate — they each follow a `setProps`/unmount that queues the deferred reset.)*
- Hover-cleanup tests reuse the same single trigger after remount; consider one case where trigger **B** reclaims the shared `HoverInteraction` that trigger **A** left pending (the instance is shared across a handle's detached triggers).
- The six near-duplicate "resets … when an open root unmounts" tests are **justified** (each drives a different store class + real DOM/ARIA wiring) — not a finding, just confirming they shouldn't be collapsed.

**B. Type design**
- `popupStoreUtils.ts:34` — `PopupStoreWithOpen` was promoted from private `type` to `export type` but has **no external consumer**. Revert to non-exported unless one is intended.
- `popupStoreUtils.ts:349-351` — `interface ResettablePopupStore { reset(): void }` is too loose: any `{ reset(): void }` satisfies it (the name promises "popup store"). Either tighten to `extends PopupStoreWithOpen<any, any>`, or rename to plain `Resettable` to stop overselling the guarantee.
- `FloatingRootStore.ts:63-65` — new `HoverInteractionContextData` duplicates the existing, stronger `HoverContextData` (`useHoverInteractionSharedState.ts:147-149`), which uses the real `HoverInteraction` type and is what *writes* the field. Reuse `HoverContextData` (or `Pick<HoverInteraction,'reset'>`) to avoid two types drifting over the same `dataRef` slot.
- `useHoverInteractionSharedState.ts:14,24` — `handler`/`handlerDocument` must be set/cleared as a pair, but both are public mutable fields; the invariant is enforced only by convention (all writers now route through the new methods — good). Consider modeling them as one nullable object `{ handler, document } | null`, or privatizing both (the only external read is `useHoverFloatingInteraction.ts:226-227`), so "both or neither" becomes unrepresentable-otherwise. *(This change correctly fixes a real prior bug — see Strengths.)*

**C. Comment accuracy** *(the two downgraded-from-Critical items — verified)*
- `popupStoreUtils.ts:374-376` — the `NoInfer` comment cites a specific failure ("`nextState.openProp` … would otherwise conflict with `boolean | undefined`"). One agent ran `tsc -b` with the wrappers removed and got **zero errors**, so the cited scenario doesn't reproduce. `NoInfer` is defensible as defensive typing (pinning inference to `store`), but the *specific claim* is unverifiable. Either confirm a real conflict case or soften the comment to "defensive: keep these args from participating in inference of `State`/`SetOpenEventDetails`."
- `popupStoreUtils.ts:28-31` + `:407` — the parenthetical example "a keyed remount mounts the next root before the previous one's cleanup runs" has the ordering **inverted**: for a keyed swap in one commit React runs the old fiber's cleanup *before* the new fiber's setup. The general statement ("a store can briefly be claimed by more than one root") is *correct* (true for route transitions / concurrent mounts) and is the real justification for ref-counting; the mechanism works regardless of order. Reword the example to the Strict-Mode / concurrent-transition case, or drop the "keyed remount" framing. **Not a correctness issue.**
- `popupStoreUtils.ts:389` — "No-op for stores without hover interactions (e.g. Dialog)" reads as if Dialog lacks the method. Verified: Dialog's `floatingRootContext` *is* a full `FloatingRootStore` with a real `resetHoverInteraction`; the no-op comes from `dataRef.current.hoverInteractionState` never being populated (no Dialog trigger uses a hover hook). Clarify: "No-op unless a hover trigger registered shared state on `dataRef` (Dialog never does)."

---

## Strengths

- **Reference-count + microtask deferral is correct** under Strict Mode double-invoke, keyed remounts, and multiple concurrent owners; the counter can't leak or go negative, and `queueMicrotask` matches established patterns (`MenuRoot.tsx`, `SelectRoot.tsx`). Internal root-owned stores are never reset (hook is only called with `handle?.store`).
- **`HoverInteraction` document-capture genuinely fixes a real bug**: the old cleanup derived the document from `domReferenceElement` at teardown, which can be detached after unmount → `ownerDocument(null)` falls back to global `document`, leaking the listener or removing from the wrong realm. Capturing `handlerDocument` at attach time is the right fix.
- **Regression coverage fails on `master`, passes after** — directly guarded at all three layers, with thoughtful setups (hover tests unmount *before* `clock.tick` to reproduce the freeze; `useFloating` test pins that `reset()` clears `floatingContext` but preserves the shared `hoverInteractionState`).
- **No project-guideline violations** — `useIsoLayoutEffect`, owner/shadow-DOM-safe utils, timer utilities, Vitest-only APIs, `name.test.tsx` colocation, correct `skipIf(isJSDOM)` gating throughout.

---

## Recommended action
1. **Important #1** (comment at the `setOpen` call; optionally guard the microtask) and the three **test gaps (#2–#4)** are the highest-value follow-ups.
2. Comment-accuracy fixes (§C) are quick and prevent future maintainers being misled.
3. Type-tightening (§B) and the `flushMicrotasks` nit (§A) are polish — safe to defer.

atomiks:
## Code Review (GPT-5.5)
Informational: I think there are a few bundle-size simplifications worth trying before accepting the +451 B gzip increase.

### 1. Bugs / Issues (None)
I don't see a correctness issue in the reset strategy here; this is a size/maintainability pass.

### Simplification Opportunities

The best-looking reductions are in the runtime cleanup plumbing rather than the tests:

1. Move `usePopupRootUnmountCleanup(handle?.store)` into `usePopupStore(externalStore, ...)`.

   All four changed roots already call their store's `useStore(handle?.store, ...)`, so the cleanup can be automatic for external handle stores. That would delete the repeated root imports and hook calls in Dialog, Popover, Tooltip, and PreviewCard while keeping internal stores untouched.

2. Replace `HoverInteraction.additionalOpenTimeouts: Set<Timeout>` with a single shared nested-trigger timeout on `HoverInteraction`.

   The Set/register/unregister API looks heavier than needed for the tooltip nested-trigger reopen case. Since hover state is already shared per floating root/handle and the reopen timer is just another hover-open timer, a single `nestedTriggerOpenTimeout = new Timeout()` on `HoverInteraction` should let `reset()` clear it directly. Then `TooltipTrigger` can drop the `useTimeout()` allocation, registration effect, and `registerAdditionalOpenTimeout` API.

3. Fold `FloatingRootStore.resetHoverInteraction()` into `FloatingRootStore.reset()`.

   The popup reset path always calls both `resetHoverInteraction()` and `reset()`. Having `FloatingRootStore.reset()` clear `dataRef.current.hoverInteractionState?.reset()` first would remove one public-ish method and one call site without changing ordering in a meaningful way.

These should preserve the current behavior while trimming some of the new helper/API surface that is contributing to the bundle bump.

michaldudak:
Used a different approach in #4992

Issue #4956, opened 2026-06-01
[menu] Menu.Trigger aria-expanded stays "false" after open in v1.5.0


mjcampagna:
# Bug report

## Current behavior

After upgrading from `1.4.1` to `1.5.0`, `Menu.Trigger` no longer flips `aria-expanded` to `"true"` when the menu opens. The trigger correctly receives `data-popup-open=""` and `data-pressed=""`, but `aria-expanded` stays `"false"`, so screen readers announce the menu as collapsed while it's visibly open.

After clicking the trigger:

```html
<button data-popup-open="" data-pressed="" aria-expanded="false" …>Open</button>
```

Reproducible in both the contained pattern and the detached `handle` pattern — not specific to any wrapper code.

## Expected behavior

`aria-expanded` should be `"true"` whenever the menu's popup is open, matching v1.4.1 behavior and the [WAI-ARIA Menu Button pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/).

## Reproducible example

```tsx
import { Menu } from "@base-ui/react/menu";

export function Repro() {
  return (
    <Menu.Root>
      <Menu.Trigger>Open</Menu.Trigger>
      <Menu.Portal>
        <Menu.Positioner>
          <Menu.Popup>
            <Menu.Item>Item</Menu.Item>
          </Menu.Popup>
        </Menu.Positioner>
      </Menu.Portal>
    </Menu.Root>
  );
}
```

Click the trigger and inspect — `aria-expanded` stays `"false"` while `data-popup-open=""` is correctly set.

## Base UI version

`@base-ui/react@1.5.0` (regression from `1.4.1`)

## Which browser are you using?

Reproducible in jsdom (test environment) and confirmed in Chrome — browser-agnostic, the attribute simply isn't written.

## Which OS are you using?

macOS

## Which assistive tech are you using (if applicable)?

Not directly tested, but `aria-expanded` is a primary signal for VoiceOver / NVDA / JAWS to announce "expanded" vs "collapsed" on menu buttons.

## Additional context

`MenuTrigger.tsx` selects `triggerProps` from the store gated on `isMountedByThisTrigger`:

```ts
// packages/react/src/menu/trigger/MenuTrigger.tsx#L220
const rootTriggerProps = store.useState('triggerProps', isMountedByThisTrigger);
```

```ts
// packages/react/src/utils/popups/store.ts#L205-L211
isMountedByTrigger: createSelector(
  (state, triggerId) =>
[...]

PR #4957, opened 2026-06-01
[test] Test menu trigger `aria-expanded`

mj12albert:
https://github.com/mui/base-ui/issues/4956


atomiks:
I think it was caused by adding `fastComponentRef`, at least based on the issue description (and on `master` deploy, `aria-expanded` is visible): https://github.com/mui/base-ui/pull/4866

atomiks:
I think the runtime change is redundant after #4866. `triggerProps` already carries `aria-expanded`; the issue was that `useStoreFast` returned stale `triggerProps` when the selector arg changed in `MenuTrigger`. That’s fixed on `master`.

The updated test is still valuable though, since the previous controlled `open` test didn’t exercise the transition that regressed in 1.5.0.

atomiks (approved):

PR #4958, opened 2026-06-01
[checkbox group] Fix validation with multiple required checkboxes

### LINKED ISSUE #4952: [CheckboxGroup] Multiple required child checkboxes are not all validated
# Bug report

## Current behavior

When multiple `Checkbox.Root` components inside a `CheckboxGroup` are marked as `required`, validation can incorrectly pass after only one of the required checkboxes is checked.

As a result, the `Field.Root` can be considered valid even while another required checkbox in the same group is still unchecked.

## Expected behavior

Each required child checkbox inside a `CheckboxGroup` should continue to enforce its own required constraint.

For example, if two child checkboxes are marked as `required`, checking only one of them should not make the `Field.Root` valid while the other required checkbox remains unchecked.

## Reproducible example

A failing test case reproduces the issue:

`packages/react/src/checkbox-group/CheckboxGroup.test.tsx`

```tsx
it('does not clear the required error when another required child checkbox is unchecked', async () => {
  const { user } = render(
    <Form>
      <Field.Root name="apple">
        <Fieldset.Root render={<CheckboxGroup defaultValue={[]} />}>
          <Fieldset.Legend>Allowed network protocols</Fieldset.Legend>
          <Field.Item>
            <Field.Label>
              <Checkbox.Root value="one" data-testid="checkbox" required />
              HTTP
            </Field.Label>
          </Field.Item>
          <Field.Item>
            <Field.Label>
              <Checkbox.Root value="two" data-testid="checkbox" required />
              HTTPS
            </Field.Label>
          </Field.Item>
        </Fieldset.Root>
        <Field.Error match="valueMissing" data-testid="error">
          required
        </Field.Error>
      </Field.Root>
      <button type="submit">submit</button>
    </Form>,
  );

  const checkboxes = screen.getAllByTestId('checkbox');

  await user.click(screen.getByText('submit'));
  expect(screen.getByTestId('error')).toHaveTextContent('required');

  await user.click(checkboxes[1]);
  await user.click(screen.getByText('submit'));

  expect(screen.getByTestId('error')).toHaveTextContent('required');
});
```

The test fails with `TestingLibraryElementError: Unable to find an element by: [data-testid="error"]` on the final assertion, because `Field.Error` is removed after only the second required checkbox is checked.

## Base UI version

v1.5.0

Reproduced in the test environment.

## Which OS are you using?

macOS

N/A

## Additional context

The issue seems related to how `Field.Root` tracks a single validation input ref. Multiple checkbox inputs inside the same field can register with that ref, but only one input ultimately becomes the active validation target.

atomiks:
Fixes #4952

In a `Field` wrapping a `CheckboxGroup`, all checkboxes share one validation instance but each merges its hidden input into a single shared ref, so only the last-mounted input was ever checked. With multiple `required` checkboxes, ticking any one of them cleared the field's error even while the others were still empty.

## Root cause

Field validation read native validity from a single representative input, so the group looked valid as soon as one of several `required` checkboxes was ticked.

## Changes

- Track every input registered against a field, and when more than one shares it, validate against a currently-invalid one so each `required` checkbox is enforced independently (matching native per-checkbox behavior).


packages/react/src/field/root/useFieldValidation.ts
```diff
@@ -57,10 +58,43 @@ export function useFieldValidation(
 
   const timeout = useTimeout();
   const inputRef = React.useRef<HTMLInputElement | null>(null);
+  const registeredInputs = useRefWithInit(() => new Set<HTMLInputElement>()).current;
```
romgrk, line 94: We should probably expose factory functions for built-in objects, this hook was meant for classes with a static function like `useRefWithInit(Timeout.create)` to avoid creating GC work. Here, the inline function still needs to be GC'ed later. I think it's still better than `useRef(new Set())` though, `Set` probably gets a non-zero allocated capacity even if it's created empty.

flaviendelangle (commented):
# PR Review Summary — #4958 [checkbox group] Fix validation with multiple required checkboxes

Reviewed in an isolated worktree by 5 specialized agents (general code, tests, comments, silent-failures, simplification). The agents independently verified claims by running `pnpm typescript`/`eslint`/the test suites and by reverting the fix to confirm what the tests actually guard. **Verdict: the fix is correct and well-scoped for the core bug — but the test suite under-guards it, and there are two edge cases worth tightening.**

---

## The fix is sound
- Root cause is correctly diagnosed: all group checkboxes share one `validation.inputRef`, so the last-mounted input won. The new `registeredInputs` Set + "prefer a currently-invalid input" selection in `commit` correctly restores native per-checkbox `required` semantics. Confirmed by reverting the fix — test 1 fails without it.
- Backward-compatible: only `CheckboxRoot` merges `validation.registerInput`; every other field control (Radio, Switch, Select, NumberField, Combobox, OTPField…) leaves the Set empty and keeps the `?? inputRef.current` fallback unchanged.
- `useMergedRefs` honors ref-callback cleanups, so the Set is correctly pruned on unmount. Types/lint/tests are green.
- `React.useCallback` for `registerInput` (not `useStableCallback`) is the **correct** choice for a ref callback — its only dep `registeredInputs` is stable, so identity never changes. Don't "fix" this.

## Important (should address before merge)

**1. Two of the three new tests don't actually guard the regression** *(tests — empirically verified)*
Reverting only the source hunks while keeping the tests: **test 1 fails (good), but tests 2 and 3 pass on the buggy code.** Their names imply they protect the disabled-skip and registry/unmount branches, but neither does, because in both the correct input is also the last-mounted one (so old `inputRef.current` already worked).
- Test 2 (`CheckboxGroup.test.tsx:296`): mount the disabled+required checkbox **last** (unchecked) so it's the would-be-selected input — only then does the `input.disabled` skip decide the outcome.
- Test 3 (`CheckboxGroup.test.tsx:332`): unmount a **checked/valid** sibling while an **unchecked required** sibling remains — that's the exact "shared `inputRef` nulled by a sibling unmount" hazard the registry exists for.
- **Missing entirely:** `validationMode="onChange"` with two `required` checkboxes (the mode where users most notice the bug — error clearing the instant one box is ticked). This exercises the `revalidate` branch (`useFieldValidation.ts:130-174`).

**2. Stale custom validity on a rotating element** *(silent-failures — flagged, partially disputed)*
`element.setCustomValidity(...)` is only ever set/cleared on the single element chosen each commit. Master always chose the same last-mounted input; this PR rotates the chosen element, so for a Field with a **custom `validate` + multiple required checkboxes**, a custom error set on input A may not be cleared when a later commit selects B. The code-reviewer's probe of the *typical* flow passed (the `break`-on-invalid tends to re-select and self-heal the stale element), so this is a narrow, possibly self-healing case rather than a confirmed blocker — but it's PR-introduced and worth a targeted test (custom `validate` + group, check/uncheck cycle) to confirm no phantom error lingers.

## Suggestions (nice to have)

**3. All-disabled group: comment imprecision + edge behavior** *(converged by 3 agents)*
The comment "fall back to it only when empty" (`useFieldValidation.ts:83`) is inaccurate — the `registeredElement ?? inputRef.current` fallback also fires when the registry is **non-empty but all inputs are disabled**, in which case it falls back to a possibly-disabled `inputRef.current`, contradicting the loop's own disabled-skip rule. The behavior largely matches master (not a clear regression), but a one-line comment fix is warranted, e.g. *"…fall back only when no enabled input was found (empty registry, or every registered input is disabled)."* Optionally skip a disabled `inputRef.current` so the all-disabled case can't surface a stale `valueMissing`.

**4. Extract the selection loop into a named helper** *(simplify — optional)*
Lifting the loop to a module-level `findRepresentativeInput(inputs)` (returning `input` directly on the invalid case drops the `= input; break` pair) makes `commit` read as intent and shortens the inline comment. Pure clarity, behavior-preserving.

## Comments
Generally high-quality "why" comments worth keeping. The only inaccurate one is the "only when empty" phrasing above. All test inline comments accurately match their assertions.

---

### Recommended action
1. Strengthen tests 2 & 3 to actually fail without the fix, and add a `validationMode="onChange"` multi-required test (item 1) — highest value.
2. Fix the "only when empty" comment (item 3).
3. Add a custom-`validate` + group test to settle item 2; harden if it reproduces.
4. Consider the helper extraction (item 4).

flaviendelangle (approved):
I've completed a comprehensive multi-agent review of PR #4958, plus empirical test verification. Here's the consolidated report.

---

# PR Review Summary — #4958 `[checkbox group] Fix validation with multiple required checkboxes`

**What it does:** In a `Field` wrapping a `CheckboxGroup`, every checkbox merged its hidden input into a single shared `inputRef`, so only the last-mounted input was ever validated. With multiple `required` checkboxes, ticking any one cleared the field error while others were still empty. The fix tracks all inputs in a `Set` (`registerInput`) and validates against a "representative" — the first enabled currently-invalid input (`findRepresentativeInput`), falling back to the shared `inputRef`.

**Verdict: Approve.** The fix is correct, well-targeted, well-commented, and well-tested. No critical or blocking issues found across six specialized review passes. Only minor polish suggestions remain.

## ✅ Empirical verification (I ran the tests)

I checked the PR out in a worktree and ran the suites both ways:

- **New tests against pre-fix source:** the **4 core regression tests fail**, test E passes — proving the tests genuinely capture the bug.
- **New tests against the fixed source:** **all pass** (49 passed / 5 skipped in CheckboxGroup).
- **No regressions:** Field + Checkbox + Radio + RadioGroup suites → **357 passed / 36 skipped**, all green.

This matches the test-analyst's per-test prediction exactly (A–D are fix-specific, E is a general guard that passes on both).

## Critical Issues
None.

## Important Issues
None. The concerns probed by the agents were all investigated and cleared:
- `commit()` early-return preserves prior semantics (custom `validate` still runs even when all checkboxes are disabled — verified).
- Ref-callback cleanup composes correctly with `useMergedRefs` across React 18/19 (the registry is pruned on unmount; test C proves it).
- No stale `customValidity` in normal toggling — the "prefer first-invalid" selection re-selects the same input that holds the custom error and clears it (test E + manual trace confirm).
- `Set` "first invalid wins" is order-independent for field-level validity.

## Suggestions (nice-to-have polish)

1. **Use `NOOP` for the default `registerInput`** *(highest-confidence; simplifier + type-designer agreed)* — `FieldRootContext.ts:76`. The same `validation` block already does `change: NOOP`; `registerInput: () => undefined` breaks the pattern. `NOOP` is type-compatible with `React.RefCallback`.
   ```ts
   registerInput: NOOP,   // instead of () => undefined
   ```

2. **Tighten test E's comment** *(test-analyzer + comment-analyzer agreed)* — `CheckboxGroup.test.tsx:415`. It says "no stale custom validity must linger on **either input**," but the test only asserts `aria-invalid`, which is **field-level** state (`getValidationProps` → `state.valid`), not per-input native validity. Reword to reflect that it guards the field-level result.

3. **Clarify the `findRepresentativeInput` JSDoc** *(comment-analyzer + type-designer agreed)* — `useFieldValidation.ts:40-45`. Lead with the enabled-only constraint ("the first *enabled* currently-invalid input") since disabled inputs are skipped *before* the validity check, and note that "first" relies on `Set` insertion (mount) order.

4. **(Optional) Harden the phantom-customValidity edge** *(silent-failure-hunter + type-designer)* — low severity, latent. If the input holding a custom error becomes **disabled or detached** between the set and clear commits, the clear targets a different input and the original keeps a stale `setCustomValidity`. Inert today because the form uses `noValidate` (JS-driven validation) and disabled inputs are excluded from submission. Defensive options: skip `!input.isConnected` in `findRepresentativeInput`, or clear custom validity across the whole registry when validation resolves valid.

5. **(Optional) Add an `onBlur`-mode test** *(test-analyzer)* — the only validationMode × multiple-`required` combination not exercised by the new tests (the existing `onBlur` test uses a custom validator, not native `required`). Submit and onChange paths are covered.

6. **(Consistency note) Radio-group precedent** *(type-designer)* — `RadioGroup` solves the analogous problem with `registerInputRef: (element) => void` via `useStableCallback`. This PR uses `registerInput: React.RefCallback` via `React.useCallback`. Both are correct; worth a glance for naming/typing consistency, but the differing approach (lazy Set vs. eager single ref) is justified — checkbox groups need per-checkbox `required`, radios have a single value.

## Strengths

- **Root cause correctly identified and fixed** at the right layer (shared `useFieldValidation`), matching native per-checkbox constraint semantics.
- **Excellent, accurate comments** explaining the non-obvious "last-mounted wins the shared ref" rationale — verified accurate by the comment analyst.
- **Strong, behavior-driven tests** that fail on pre-fix code (empirically confirmed), including a clever unmount case (test C) that proves *why* a registry is needed.
- **Clean, minimal diff** — no over-engineering; the single-loop `findRepresentativeInput` is the simplest correct form.
- **No silent failures, no swallowed errors**; the one fallback (`?? inputRef.current`) is explicit and documented.

## Recommended action

All findings are non-blocking. If you want a quick polish pass before merge: apply **#1** (trivial, `NOOP`) and **#2/#3** (comment tweaks). **#4** and **#5** are worth a line in the issue/PR as optional follow-ups but don't gate this change.

PR #4959, opened 2026-06-02
[code-infra] Add flat build setup

brijeshb42:
Adds the flat build setup (`code-infra build --flat`): build output is flat with extension-based module type (`.js` CJS, `.mjs` ESM, `.d.ts`/`.d.mts` types) instead of separate `esm`/`cjs` subdirectories.

Verified `publint` and `attw` pass.

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -12,15 +12,17 @@
   "exports": {
     "./store": "./src/store/index.ts",
     "./platform": "./src/platform/index.ts",
-    "./*": "./src/*.ts"
+    "./*": "./src/*.ts",
+    "./*.test": null,
+    "./*.spec": null
   },
   "imports": {
     "#formatErrorMessage": "./src/formatErrorMessage.ts"
   },
   "scripts": {
     "prebuild": "rimraf --glob build build-tests \"*.tsbuildinfo\"",
-    "build": "code-infra build --copy .npmignore --tsgo",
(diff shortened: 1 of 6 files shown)


packages/utils/package.json
```diff
@@ -11,23 +11,25 @@
   },
   "exports": {
     "./store": "./src/store/index.ts",
-    "./*": "./src/*.ts"
+    "./*": "./src/*.ts",
+    "./*.test": null,
```
brijeshb42, line 16: required because `./src/*.ts` expands to also include the test/spec files.

test/bundle-size/bundle-size-checker.config.mjs
Janpot, line 65: Ah, we can probably use [`expand`](https://github.com/mui/material-ui/blob/a389a2d9bc2a89c6c86c4d98aca51c3fa809a753/test/bundle-size/bundle-size-checker.config.mjs#L15) now to simplify this. But doesn't have to be part of this effort.

packages/utils/package.json
Janpot: Should we do this in the build tool itself?

Janpot (commented):
Looks mostly good, but I'm not keen on the package.json update. The build output shouldn't contain spec/test files. Maybe we should only expand those paths that have build output, even though there are more source files that match?

packages/utils/package.json
brijeshb42: Yeah. I'll take that up separately and create a new PR here.

Janpot (approved):

PR #4965, opened 2026-06-03
[accordion] Proposal: align Accordion keyboard navigation with APG

chuganzy:
The Accordion APG pattern no longer recommends roving focus, as documented in https://github.com/w3c/aria-practices/pull/3434.

This PR proposes removing roving focus from Accordion to align with the latest APG guidance.


atomiks:
Probably more accurate: "`@deprecated Roving focus was removed from Accordion following the APG guidance update`"

Not sure if props becoming no-ops is confusing but given the guidelines update and deletion of a bunch of code, I'm okay with this as it's not critical behavior.

According to Codex: 
- Arrow keys conflicted with normal page scrolling. The original W3C issue says Up/Down moved focus between accordion headers, so the user could no longer scroll the page to read opened accordion content. ([github.com](https://github.com/w3c/aria-practices/issues/3406))
- APG concluded accordion is not really a composite widget like menu/listbox/tabs/grid. It’s “a collection of independent headings,” closer to disclosure sections, so composite roving-focus navigation was considered contrived. ([w3.org](https://www.w3.org/2026/04/01-aria-apg-minutes.html)) (**Arguably, this is also the case for Navigation Menu**)
- The shortcuts had poor discoverability: there’s no native/browser precedent or role="accordion" signal that would tell users arrow/Home/End are available. ([w3.org](https://www.w3.org/2026/04/01-aria-apg-minutes.html)) Current APG keyboard guidance now only lists Enter/Space and Tab/Shift+Tab for accordion headers. ([w3.org](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/))

colmtuite:
I'm fine with this 👍

docs/scripts/generateLlmTxt/__snapshots__/mdxToMarkdown.test.mjs.snap
```diff
@@ -676,7 +684,15 @@ type AccordionTriggerState = {
   value: any[];
   /** Whether the component should ignore user interaction. */
   disabled: boolean;
-  /** The component orientation. */
+  /**
+   * The component orientation.
+   *
+   * Deprecated following the [APG guidance update](https://github.com/w3c/aria-practices/pull/3434)
+   * to remove roving focus.
```
atomiks, line 691: `@deprecated` doesn't show up in the docs so added it to the main text. Given it's a no-op now it makes sense to remove the original description as it's otherwise misleading

atomiks (approved):

docs/scripts/generateLlmTxt/__snapshots__/mdxToMarkdown.test.mjs.snap
chuganzy: I was originally thinking to improve the docgen separately to check `@deprecated`, but this change makes more sense. Thank you for the update.

PR #4967, opened 2026-06-03
[toolbar] Do not forward `disabled` to default toolbar button

### LINKED ISSUE #4966: Tooltip doesn't trigger for pseudo-disabled toolbar button
# Bug report

## Current behavior

The tooltip does not shown on mouse over. It does however show on keyboard focus.

```tsx
    <Toolbar.Root>
      <Tooltip.Root>
        <Tooltip.Trigger render={<Toolbar.Button disabled />}>
          foo
        </Tooltip.Trigger>
        <Tooltip.Portal>
          <Tooltip.Positioner>
            <Tooltip.Popup>foo</Tooltip.Popup>
          </Tooltip.Positioner>
        </Tooltip.Portal>
      </Tooltip.Root>
    </Toolbar.Root>
```

## Expected behavior

The tooltip is shown on mouse over.

## Reproducible example

https://stackblitz.com/edit/y3hfssor?file=src%2FApp.tsx

## Base UI version

1.5.0

Chrome

## Which OS are you using?

Mac OS

N/A

## Additional context

This is related to but not the same as https://github.com/mui/base-ui/issues/3075. I believe that issue deals with buttons that are fully disabled, whereas this is pseudo-disabled.

Note it does seem to work correctly when using `<Button disabled focusableWhenDisabled />`.

For context, I am building a Git client and we have a toolbar for branches which includes a push button. The push button is always visible but might be disabled e.g. because there's nothing to push or it's loading, in which case we want to show a tooltip still to explain why it's disabled.

Related: https://base-ui.com/react/components/toolbar#using-with-tooltip

mj12albert:
Stackblitz: https://stackblitz.com/edit/jv4sr229?file=package.json

Fixes https://github.com/mui/base-ui/issues/4966

### DIFF
--- a/packages/react/src/button/Button.test.tsx
+++ b/packages/react/src/button/Button.test.tsx
@@ -148,7 +148,7 @@ describe('<Button />', () => {
 
         await user.hover(button);
 
-        expect(handleMouseMove).toHaveBeenCalledTimes(1);
+        expect(handleMouseMove).toHaveBeenCalled();
 
         await user.click(button);
 
--- a/packages/react/src/button/Button.test.tsx
+++ b/packages/react/src/button/Button.test.tsx
@@ -1,7 +1,7 @@
 import { expect, vi } from 'vitest';
 import { Button } from '@base-ui/react/button';
 import { screen } from '@mui/internal-test-utils';
-import { describeConformance, createRenderer } from '#test-utils';
+import { describeConformance, createRenderer, isJSDOM } from '#test-utils';
 
 describe('<Button />', () => {
   const { render } = createRenderer();
@@ -124,6 +124,37 @@ describe('<Button />', () => {
       expect(handlePointerDown.mock.calls.length).toBe(0);
       expect(handleKeyDown.mock.calls.length).toBe(0);
     });
+
+    it.skipIf(isJSDOM)(
+      'native button: allows hover handlers while blocking activation',
+      async () => {
+        const handleClick = vi.fn();
+        const handleMouseMove = vi.fn();
+
+        const { user } = await render(
+          <Button
+            disabled
+            focusableWhenDisabled
+            onClick={handleClick}
+            onMouseMove={handleMouseMove}
+          />,
+        );
+
+        const button = screen.getByRole('button');
+
+        expect(button).not.toHaveAttribute('disabled');
+        expect(button).toHaveAttribute('data-disabled');
+        expect(button).toHaveAttribute('aria-disabled', 'true');
+
+        await user.hover(button);
+
+        expect(handleMouseMove).toHaveBeenCalledTimes(1);
+
+        await user.click(button);
+
+        expect(handleClick).toHaveBeenCalledTimes(0);
+      },
+    );
 
     it('custom element: prevents interactions but remains focusable', async () => {
       const handleClick = vi.fn();
--- a/packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.test.tsx
+++ b/packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.test.tsx
@@ -170,6 +170,40 @@ describe('<Menu.SubmenuTrigger />', () => {
       expect(submenuTrigger).toHaveAttribute('aria-disabled', 'true');
     });
 
+    it('does not open on hover when disabled', async () => {
+      const { user } = await render(
+        <Menu.Root open>
+          <Menu.Trigger>Open menu</Menu.Trigger>
+          <Menu.Portal>
+            <Menu.Positioner>
+              <Menu.Popup>
+                <Menu.Item>1</Menu.Item>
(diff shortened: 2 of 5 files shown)


flaviendelangle (approved):
Here's the consolidated review.

# PR #4967 Review Summary — `[toolbar] Do not forward disabled to default toolbar button`

**Author:** mj12albert · **Fixes:** #4966 · **Size:** +130/−5 across 5 files (1 source, 4 tests)

The single source change in `ToolbarButton.tsx:67`:
```diff
-        { disabled },
+        render ? { disabled } : EMPTY_OBJECT,
```
Stops forcing the native `disabled` attribute onto the default toolbar button. A focusable-disabled native button now keeps only `aria-disabled` and stays hoverable, so tooltips/hover handlers work on it (issue #4966). The `disabled` prop is still forwarded when composing via `render`. All three agents agree this is a small, correct, idiomatic, well-targeted fix.

## Critical Issues (0)
None.

## Important Issues (0)
None.

## Suggestions (4)

1. **[tests] Add an explicit guard for `focusableWhenDisabled={false}` on a default toolbar button** — this is the one risky regression path (it must *still* render the native `disabled` attribute, now relying entirely on `getButtonProps`→`useFocusableWhenDisabled`). It is correct in code, and *is* covered in Chromium by `ToolbarRoot.test.tsx:184` (`toHaveAttribute('disabled')`) — but that runs only under `describe.skipIf(isJSDOM)`, so there's no JSDOM-level guard. A small explicit test would lock it in. *(code-reviewer + pr-test-analyzer)*

2. **[comments] The new comment over-narrows the condition** — `ToolbarButton.tsx:62-63` says "When composing with **another Base UI component**, `disabled` must be passed through," but the actual branch is `render` being *any* truthy value (a plain `<a>`, `<div>`, or render function also triggers forwarding). Suggested rewording anchors it to the `render` prop while keeping the Base-UI-composition intent:
   > `// When a render prop is provided (typically another Base UI component such as Menu.Trigger), forward disabled so the rendered component can derive its own disabled state. For the default toolbar button, avoid forwarding a React disabled prop so focusable disabled buttons remain hoverable for interactions like tooltips.`
   *(comment-analyzer)*

3. **[tests] Brittle assertion** — the hover tests assert `handleMouseMove` fired *exactly once* (`toHaveBeenCalledTimes(1)`); `user.hover` can dispatch more than one `mousemove` depending on pointer path. `toHaveBeenCalled()` is more robust. `Button.test.tsx`, `ToolbarButton.test.tsx`. *(pr-test-analyzer)*

4. **[tests] Two of the four new tests don't actually guard this PR's change** (not a defect, just scoping clarity):
   - `Button.test.tsx` "native button: allows hover handlers…" — `Button.tsx` was untouched and never forwarded a separate `{ disabled }`; this passes identically on master. Useful characterization test, not a regression guard.
   - `MenuSubmenuTrigger.test.tsx` "does not open on hover when disabled" — `MenuSubmenuTrigger` is a non-native div whose hover gate is unrelated to this change; also passes on master. A fine standalone test, but incidental to #4966.
   The two **strong** regression guards (would fail on old code in a real browser) are `ToolbarButton.test.tsx:94` and `TooltipTrigger.test.tsx:60`. *(pr-test-analyzer)*

## Strengths
- **Conditional verified correct across all four cases**: default focusable (aria-disabled, hoverable, no native `disabled`), `focusableWhenDisabled={false}` (native `disabled` still applied via `getButtonProps`), `render` with a Base UI button (unchanged), and `render` function/non-button (unchanged).
- **`EMPTY_OBJECT` is the established idiom** for this exact `condition ? {…} : EMPTY_OBJECT` pattern in a `CompositeItem` props array (same usage in `CheckboxRoot`, `NavigationMenuContent`, `SwitchRoot`, `RadioRoot`, `ComboboxChips`) and preserves referential stability. No type cast needed.
- **The comment rewrite fixes a previously *inaccurate* claim** — the old comment said `getButtonProps` "includes it already," which was false for the default focusable button (it only sets `aria-disabled`). The TODO (issue #1976) is correctly retained since the render-branch forwarding still exists.
- **`data-disabled` styling hook is unaffected** (it derives from component state, not the React prop) — and tests assert this.
- **Project-guideline compliant**: `@base-ui/utils/empty` import, correct `it.skipIf(isJSDOM)` gating for hover-dependent tests, Vitest/jest-dom matchers only, no redundant `flushMicrotasks` after `render`.
- The `updateDisabled` layout-effect hack in `useButton.ts:67-88` is **not** dead code — still required for the `render` composition path; its existing comment remains accurate.

## Recommended Action
This PR is in good shape to merge. None of the findings are blocking. If the author wants to polish: address Suggestion 2 (1-line comment accuracy fix) and optionally Suggestion 1 (lock in the `focusableWhenDisabled={false}` path with a JSDOM-level test). Suggestions 3–4 are minor/optional.

PR #4968, opened 2026-06-03
[toolbar] Fix disabled state and roving focus bugs

atomiks:
This is part of the Codex sweep for Toolbar. It fixes disabled-state and keyboard-navigation bugs found during review and adds focused regression coverage.

## Changes

- Removed the redundant `stopEvent` keydown handler from `Toolbar.Input` that trapped Tab and blocked vertical roving focus when disabled; `useFocusableWhenDisabled` already prevents text entry without stopping propagation.
- Computed `Toolbar.Root`'s `disabledIndices` from each item's actual disabled state, fixed the truthy index-0 check, and carried the disabled state in item metadata for `Toolbar.Button`, `Toolbar.Input`, `Toolbar.Link`, and direct `ToggleGroup > Toggle` children, so runtime `disabled` changes stay in sync.
- Moved the composite's default tab stop off a disabled, non-focusable first item when `disabledIndices` resolves a render after the initial map population, so the toolbar stays keyboard-reachable.
- Propagated `Toolbar.Group` disabled into a direct `ToggleGroup` child so its toggles disable along with the group.
- Corrected the `Toolbar.Separator` data-attribute description and the misleading inherited `orientation` default; the separator is perpendicular to the toolbar, not always horizontal.

- [P1] Disabled `Toolbar.Input` trapped focus and blocked vertical roving.
- [P1] `Toolbar.Root` miscomputed `disabledIndices` (index-0 missed, enabled non-focusable items skipped, raw `Toggle` children treated as disabled).
- [P2] Composite's initial tab stop left a disabled, non-focusable first item as the toolbar's only `tabIndex=0` entry.
- [P2] `Toolbar.Group` disabled did not disable a direct `ToggleGroup` child.
- [P3] `Toolbar.Separator` API docs showed the misleading inherited default `horizontal` and described `data-orientation` as the toolbar's orientation.


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The branch fixes the Toolbar disabled/focus regressions at the metadata source, and the added tests fail on freshly fetched `upstream/master` but pass on the PR head.

### 1. Bugs / Issues (None)

No branch-relevant bugs, repo-convention issues, or documentation mismatches found in the full `upstream/master...upstream/pr/4968` diff.

The fix addresses the underlying metadata problem rather than only the observed keyboard symptoms: Toolbar items now report both `disabled` and `focusableWhenDisabled`, direct `ToggleGroup > Toggle` items register disabled state with the outer Toolbar composite, and `Toolbar.Group disabled` now flows into direct toggle groups.

The shared `useCompositeRoot` fallback is narrowly scoped to explicit `disabledIndices`, so it covers Toolbar’s late metadata update without changing composites that rely on DOM disabled fallback behavior.

### Test Coverage Assessment

Validation run on the PR head:

- `pnpm test:jsdom Toolbar --no-watch` passed.
- `pnpm test:chromium Toolbar --no-watch` passed.
- `pnpm test:jsdom CompositeRoot --no-watch` passed.
- `pnpm typescript` passed.
- `pnpm eslint` passed.
- `git diff --check upstream/master...upstream/pr/4968` passed.

I also applied the new Toolbar regression tests onto freshly fetched `upstream/master`; they failed there as expected, including the direct `ToggleGroup > Toggle`, disabled input, initial tab stop, and runtime-disabled toggle cases. That gives good confidence the tests are guarding the intended regressions rather than only asserting the new implementation.

PR #4969, opened 2026-06-03
[select] Fix flaky tab-out test from stale focus-on-open

atomiks:
A Select test flaked on React 18 (`does not leave a tabbable option while closed and kept mounted after tabbing out`). The initial focus queued by `FloatingFocusManager` could run after the popup had already closed and pull focus back onto the item.

## Root cause

`FloatingFocusManager` queues the initial focus on the next animation frame. When the popup closes before that frame runs (tabbing out of a kept-mounted popup), the queued focus still ran and stole focus back onto the item.

## Changes

- Skip the queued initial focus when the floating element is no longer open.

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx
@@ -283,6 +283,7 @@ export function FloatingFocusManager(props: FloatingFocusManagerProps): React.JS
   const initialFocusRef = useValueAsRef(initialFocus);
   const returnFocusRef = useValueAsRef(returnFocus);
   const openInteractionTypeRef = useValueAsRef(openInteractionType);
+  const openRef = useValueAsRef(open);
 
   const tree = useFloatingTree(externalTree);
   const portalContext = usePortalContext();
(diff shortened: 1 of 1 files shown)


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The stale deferred focus guard is correctly scoped to the closed-popup case, and I don’t see an edge case this breaks.

### 1. Bugs / Issues (None)

The change in `packages/react/src/floating-ui-react/components/FloatingFocusManager.tsx` fixes the right race: `enqueueFocus()` runs on the next animation frame, so a kept-mounted popup can close before that frame fires. Checking `openRef.current` inside `shouldFocus()` prevents that stale frame from pulling focus back into a closed floating element.

The guard is also placed before the existing `hadFocusInside` branch, which is important. Once `open` is false, even “focus was inside when queued” should not win over the user’s later tab movement.

### Edge Cases Worth Noting

Rapid close/reopen looks fine: a fresh open schedules a new `enqueueFocus()`, and `enqueueFocus()` already cancels the previous queued frame globally.

`initialFocus={false}`, `undefined`, and custom `initialFocus` resolution are unaffected because the new check only runs after the existing initial-focus resolution decides to enqueue focus.

This specifically fixes the open-state stale focus path. It does not broaden behavior for disabled/unmounted-only stale frames, but I don’t see that as a regression from this PR.

### Test Coverage Assessment

Validated the PR head with:

```sh
pnpm test:chromium Select --no-watch
pnpm use-react-18
pnpm test:chromium Select --no-watch
```

Both Chromium Select runs passed, including under React 18.3.1.

PR #4971, opened 2026-06-03
[select] Fix dirty state not clearing in multiple mode

atomiks:
In a multiple Select inside a Field, `[data-dirty]` stayed set after the selection returned to its initial value.

## Root cause

The dirty check used `value !== validityData.initialValue`. In multiple mode `value` is a fresh array on every change, so the reference comparison was always true once the user interacted — even after deselecting and reselecting back to the initial items.

## Changes

- Compare array values element-by-element via `isItemEqualToValue` when computing dirty state in multiple mode, falling back to reference comparison for non-array values. This mirrors the equivalent combobox fix.
- Add regression tests for returning to the initial value, object values compared via `isItemEqualToValue`, and the order-sensitive case.

## Notes

- The comparison is order-sensitive, matching `CheckboxGroup` and `Slider`. Returning to the same items in a different order is still treated as dirty.


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The dirty-state fix is narrowly scoped, uses the existing order-sensitive equality utilities correctly, and the added regression tests cover the important multiple-select cases.

### 1. Bugs / Issues (None)

No branch-relevant bugs or blocking issues found.

The patch addresses the actual failure mode: multiple Select values are arrays, so reference comparison kept `[data-dirty]` set after returning to an equivalent initial value. Comparing array entries through `compareItemEquality(..., isItemEqualToValue)` preserves object-value support while leaving single-value behavior unchanged.

The order-sensitive comparison is consistent with the existing `areArraysEqual` usage in CheckboxGroup and Slider, and the object-value path matches Select’s existing `isSelected` semantics.

Coverage looks appropriate for this bug fix: it includes returning to the initial primitive value, object values compared through `isItemEqualToValue`, and the intentional different-order case that should stay dirty.

Validation run:
- `pnpm typescript`
- `git diff --check upstream/master...upstream/pr/4971`

PR #4972, opened 2026-06-03
[combobox] Fix autofill and selected state edge cases

atomiks:
Part of the Codex sweep. Ports the Select autofill and selected-state fixes (#4934) to Combobox, with the review findings from that port folded in.

## Changes

- Browser autofill matches an item's rendered label, not just its serialized value, so a primitive `value="US">United States` autofills correctly. Works for inline children and the `items` prop.
- Field state (dirty, validation, error clearing) runs through the shared `useValueChanged` effects and respects `onValueChange`/`onInputValueChange` cancellation. A canceled autofill, selection, or input change no longer touches field state, and a canceled single selection keeps the popup open.
- Multiple mode clears `data-dirty` when the value returns to its initial set, comparing element-by-element with the item comparator instead of by array reference.
- When the input is inside the popup, the trigger's `aria-controls` points at the dialog popup (a stable `${rootId}-popup` id shared by the popup and trigger) instead of the listbox.
- `Combobox.Label` resolves to `Combobox.Label.State` in the generated docs.

- Canceling a single-select item press still closed the popup; it now stays open, with a test.
- The `${rootId}-popup` id was built in both the popup and the trigger; it's now one `getComboboxPopupId` helper, with the custom-id path covered.
- The autofill `forceMounted` mount was sticky and ran for `multiple` and `none` modes that never match; it's now scoped to single-selection autofill.


atomiks:
# PR review

No merge-blocking issues found. After the `waitFor` cleanup, the combined Codex + Claude review has no actionable findings: the autofill matching, cancellation guards, dirty-state comparison, and popup `aria-controls` changes are covered thoroughly. Residual note only: the `items` rendered-label autofill path still relies on the `forceMounted` external-store update and label registration completing before the queued microtask, which matches the existing pattern and is covered by tests.

No findings.

## Verdict

**Approve** — the only test-style nit was fixed and the remaining review found no regressions.

---

🤖 Review generated with Codex and Claude Code combined

michaldudak (approved):

PR #4975, opened 2026-06-05
[toast][docs] Increase timeout on undo demo

aarongarciah:
From 5s to 10s

Preview https://deploy-preview-4975--base-ui.netlify.app/react/components/toast#undo-action


flaviendelangle (approved):
Make sense to me
Not sure if you want another approval

PR #4978, opened 2026-06-05
[docs] Add background to inline code tags

### LINKED ISSUE #4009: [docs] Talking about accessibility, why do you use the blue color for non-links?

aarongarciah:
Closes https://github.com/mui/base-ui/issues/4009
Previous attempt: https://github.com/mui/base-ui/pull/4043

|before|after|
|---|---|
|Screenshot 2026-06-05 at 12 12 07|Screenshot 2026-06-05 at 12 11 55|

Preview: https://deploy-preview-4978--base-ui.netlify.app/react/overview/quick-start#portals


aarongarciah:
@dav-is the new local rehype plugin seems the only reliable way of doing this without introducing any runtime. The idea is that instances like the one in the screenshot below (HTML tags) are kept without a background.

image

If you have any ideas on how to make this simpler, please share. CSS-only solutions required checking for certain `.pl-*` class names being present inside the `<code>` tag.

dav-is:
> If you have any ideas on how to make this simpler

Are you opposed to using `:has()` selectors? The performance should be fine, given it only needs to check one level deep on inline code blocks. It's set to become Baseline Widely Avaiable in July (supported in all browsers for 30 months).

```css
code[data-inline]:not(:has(> .di-ht)) {
  /* */
}
```

If not, your PR uses the enhancers API as intended 👍🏼

aarongarciah:
> Are you opposed to using :has() selectors?

@dav-is that was my first attempt, but targeting `.pl-*` classes instead of the docs infra ones (`.di-*`). It felt wrong, but it's growing on me. Seems way simpler, although more brittle.

aarongarciah:
@dav-is you convinced me. Updated the PR to stick to the CSS-only approach.

dav-is (approved):

oliviertassinari:
A quick feedback coming from #4009:

## 1. Vertical alignment

If we are adding vertical padding, shouldn't we also set `display: inline-block;` so we continue to have the vertical-align baseline style that currently exist?

PR: https://deploy-preview-4978--base-ui.netlify.app/react/overview/quick-start#ios-26-safari

SCR-20260610-tzlk

Proposal:
SCR-20260610-tzct

## 2. Information priority

Are we sure about keeping the font-size at 16px?. 15.5px feels more balanced (15px is maybe even better), the code block drawing less attention away from the sentence I'm trying to read:

PR: https://deploy-preview-4978--base-ui.netlify.app/react/overview/quick-start#ios-26-safari
SCR-20260610-uadp

Proposal:
SCR-20260610-ubjh

aarongarciah:
@oliviertassinari I missed `inline-block`. re reducing the font-size, we're already reducing the optical size of the Paper Mono font via `size-adjust`, which is currently at `100% * 13.5 / 14 = 96,42%`, so it's not really 16px. I'm adding `inline-block` and reducing the optical size even further in https://github.com/mui/base-ui/pull/5009

oliviertassinari:
> we're already reducing the optical size of the Paper Mono font via size-adjust, which is currently at 100% * 13.5 / 14 = 96,42%, so it's not really 16px

@aarongarciah On https://deploy-preview-5009--base-ui.netlify.app/react/overview/quick-start#portals the code becomes really small vs. https://master--base-ui.netlify.app/react/overview/quick-start#portals, no?

SCR-20260612-cyfo

---

Something else I saw: https://master--base-ui.netlify.app/react/utils/csp-provider#disable-inline-style-elements this afternoon (while looking at CSP with Jan). If there is a blue background, should there be a green background too?

SCR-20260612-cxps

---

> I'm adding inline-block and reducing the optical size even further in https://github.com/mui/base-ui/pull/5009

👍

aarongarciah:
Maybe the font-size is a bit small on code blocks. I've changed the approach and left `size-adjust` alone for Paper Mono and changed the font-size of inline code tags instead.

~re green ones, I felt like it wasn't necessary but after playing with it, it looks cool. Added to https://github.com/mui/base-ui/pull/5009 as well.~

edit: we are not adding the green bg.

oliviertassinari:
> #5009

@aarongarciah 👍 

> Maybe the font-size is a bit small on code blocks

https://master--base-ui.netlify.app/react/handbook/animation#animating-components-unmounted-from-dom-when-closed-with-motion looks like this today:

SCR-20260614-sfug

On my end, I struggle to read it, maybe I have some issue with my contact lens. I have to go into +110% zoom mode in my browser to enjoy it. I imagine 13px could work since it works on https://web.archive.org/web/20260205072416/https://base-ui.com/react/components/accordion#root, but then the issue is with the font-weight, it's really thin at that 13px size.

PR #4979, opened 2026-06-08
[drawer] Fix Safari backdrop positioning demo

lyzno1:
Align the snap points Drawer CSS Modules backdrop demo with the existing iOS Safari positioning pattern.


aarongarciah:
With this change, the nested and position demos have regressed:

|before|after|
|---|---|
|nested-before|nested-after|
|position-before|position-after|

PR #4980, opened 2026-06-08
[drawer] Drive swipe gestures natively to stop per-frame re-rasterization

### LINKED ISSUE #4863: [drawer] Swipe on low end devices has perf considerations

atomiks:
Fixes #4863

### Problem

While a Drawer was open, every swipe `touchmove` re-rasterized the sheet's content instead of staying compositor-only. The `touchmove` bubbled into React's delegated touch system, and processing it there made Chrome invalidate and repaint the document and full-screen popup once per move (~2 paints per move). The translate was already a cheap composited transform — the repaints came from the event-dispatch path.

### Fix

Drive the swipe from a native, capture-phase `touchmove` listener. Once the Drawer claims the gesture it calls `stopPropagation()` so the move never reaches React's delegated handlers, forwarding it via a typed `moveFromNativeTouch` hook method. Propagation is preserved for cases the Drawer must not hijack (native scroll, range inputs, pinch-zoom, text selection). `getDragTransform` also moves to `translate3d`. No consumer CSS or workarounds required.

### Results

[Heavy drawer stress benchmark](https://mui-2026-drawer-performance.netlify.app/) (800-node payload, includes Vaul), headless Chromium, 20× CPU slowdown, paired same-session runs, medians:

| Lib | Total busy | Paint | Paint ev | Avg move task |
| Base UI (before) | 981 ms | 341 ms | 78 | 17.1 ms |
| Base UI (after) | 631 ms | 70 ms | 6 | 8.9 ms |
| Material UI | 723 ms | 328 ms | 79 | 14.4 ms |
| Silk | 586 ms | 80 ms | 0 | 4.2 ms |
| Vaul | 1121 ms | 86 ms | 6 | 10.2 ms |

Base UI before → after: paint 341 → 70 ms (−79%), paint events 78 → 6, avg move task 17.1 → 8.9 ms (−48%), total busy −36%. The per-move content re-rasterization is gone; paint no longer scales with the heavy content.

Notes: numbers reflect main-thread work during the gesture. `Total busy` = all main-thread time in the swipe window; `Avg move task` = per-move input-handler latency — not comparable across libraries (Silk travels via native scroll, so it offloads the drag to the compositor and delivers fewer JS moves). Material UI's `SwipeableDrawer` keeps an 800-node paper mounted while closed, which inflates absolute numbers for every row equally, so treat cross-library absolutes loosely; the Base UI before/after delta is unaffected.

## Before

Screenshot 2026-06-10 at 9 51 44 am

## After

Screenshot 2026-06-10 at 9 52 25 am

### Changes

- Handle swipe moves in a native capture-phase `touchmove` listener and call `stopPropagation()` so they don't reach React's delegated handler. Native scroll, range inputs, pinch-zoom, and text selection still propagate.
- Add `moveFromNativeTouch` to `useSwipeDismiss` so the native listener feeds the gesture without a synthetic React event.
- Switch the swipe transform to `translate3d`.


atomiks:
## Review (Claude Fable 5): base-ui#4980 — [drawer] Drive swipe gestures natively

This routes drawer swipe moves through the existing document-level capture-phase `touchmove` listener: once the drawer claims the gesture it calls `stopPropagation()` and forwards the event into `useSwipeDismiss` via the new `moveNative` method, so the move never reaches React's delegated touch system — which was what triggered ~2 paints per move. The drag transform also moves to `translate3d`, and the old React `onTouchMove` plumbing in `DrawerViewport` is removed.

**Correctness** — traced the forwarding path end to end:

- The synthetic object passed to `handleMove` supplies exactly the five fields the move pipeline reads (`touches`, `currentTarget`, `nativeEvent`, `defaultPrevented`, `timeStamp`); `getPrimaryPointerPosition` and `isTouchLikeEvent` branch on `'touches' in event`, so the native shape flows through the touch path identically to a React event. The `defaultPrevented` snapshot (vs. React's live getter) is only read on the non-touch branch, so freezing it is harmless.
- `currentTarget` is the same element the React touch props were spread onto (`viewportElement ?? popupElement`), so the scrollable-boundary check in `handleMoveCore` is unchanged.
- Propagation is preserved everywhere the drawer must not hijack the gesture: range inputs, two-finger pinch, text selection, cross-axis native scroll, and in-progress scrolling of inner containers. The old React-side guard (`allowSwipe === false || (scrollTarget != null && !allowSwipe)`) maps cleanly onto the new branches — forward only when `scrollTarget` is null/doc/body, or when `allowSwipe === true`.
- One theoretical gap: the doc/body branch forwards without re-checking `allowSwipe === false`, where the old guard would have skipped. In practice it's unreachable — native touch events keep targeting the touchstart element, so a gesture that began on a range input never reaches that branch (and a touchstart on a range input nulls out `touchState` entirely).

**Notes:**

1. **Observable behavior change.** User-attached React `onTouchMove` handlers inside the drawer no longer fire while the drawer owns the gesture (the new test asserts this). That's the point of the fix, but it's API-observable — worth a line in the changelog for anyone reading swipe moves themselves. `onTouchStart`/`onTouchEnd` still fire, so consumers see start → end with no moves in between.
2. The "non-overflowing scroll container" branch stops propagation without forwarding to the swipe — consistent, since `preventDefault` already made the gesture inert there; just noting it's the one spot where propagation stops without the swipe consuming the move.
3. The hook method is named `moveNative` and documented in terms of the mechanism (capture-phase consumers claiming the gesture) rather than the Drawer specifically, keeping the shared util component-agnostic — Toast also consumes it. Since the property key survives minification, the short name also saves a few bundle bytes.

**Tests** cover both the claimed path (consumer handler not called) and the preserved native-scroll path (handler called once), and the `translate3d` assertion tolerates browser serialization differences. Benchmark methodology in the description is appropriately honest about cross-library caveats.

LGTM — the notes above are observations, not blockers.

PR #4982, opened 2026-06-08
[scroll area] Add scrolling state to Thumb

aarongarciah:
Adds the missing `scrolling` state to `ScrollArea.Thumb`, which ends like `data-scrolling` in the DOM.


packages/react/src/scroll-area/thumb/ScrollAreaThumb.tsx
```diff
@@ -70,10 +75,14 @@ export const ScrollAreaThumb = React.forwardRef(function ScrollAreaThumb(
...
 export interface ScrollAreaThumbState {
+  /**
+   * Whether the scroll area is being scrolled.
+   */
+  scrolling: boolean;
   /**
    * The component orientation.
    */
-  orientation?: 'horizontal' | 'vertical' | undefined;
+  orientation: 'horizontal' | 'vertical';
```
aarongarciah, line 85: This change seems consistent with runtime behavior, but it could affect users relying on this type. Do you think it's better to leave outside of the PR? cc @michaldudak @atomiks

atomiks: Seems like a correctness fix? `ScrollArea.Thumb` can only render within `ScrollArea.Scrollbar`, and the context always provides an orientation, so `orientation | undefined` was inaccurate

In which case, I'm ok with the change

atomiks (approved):

PR #4983, opened 2026-06-08
[docs] Demo chrome tweaks

aarongarciah:
Preview:
- Demo: https://deploy-preview-4983--base-ui.netlify.app/react/components/accordion
- Installation block: https://deploy-preview-4983--base-ui.netlify.app/react/overview/quick-start

Changes:
- Demos
  - Removes compact demos.
  - Reduces the height of the code block when collapsed, so only 4 lines are displayed.
  - The whole code block is clickable to expand the demo code block when collapsed.
  - New tabs style.
  - New "Show code" button styles.
  - Removes the "Hide code" button once it's expanded.
  - Moves the copy code button from the demo toolbar to the demo code block.
  - Hides file tabs when there's only one file (that's the case for most Tailwind demos).
- Updated installation block (present in [quick start page](https://deploy-preview-4983--base-ui.netlify.app/react/overview/quick-start)) styles so tabs have the same vibe as demo tabs.
- Removes the "Copy" visual label from regular code blocks.
- Updates regular code blocks, prop tables, etc., to match the demos' border radius.

|before|after|
|---|---|
|Screenshot 2026-06-08 at 18 34 54|Screenshot 2026-06-12 at 15 02 21|
|Screenshot 2026-06-08 at 18 35 52|image|

### DIFF
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -103,7 +103,12 @@ const nextConfig = {
       },
       './src/app/**/demos/*/index.ts': {
         as: '*.ts',
-        loaders: ['@mui/internal-docs-infra/pipeline/loadPrecomputedCodeHighlighter'],
+        loaders: [
+          {
+            loader: '@mui/internal-docs-infra/pipeline/loadPrecomputedCodeHighlighter',
+            options: { emphasisOptions: { focusFramesMaxSize: 4 } },
+          },
+        ],
       },
       './src/demo-data/*/index.ts': {
         as: '*.ts',
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -136,7 +136,10 @@ const nextConfig = {
       test: /[/\\\\]demos[/\\\\][^/\\\\]+[/\\\\]index\.ts$/,
       use: [
         defaultLoaders.babel,
-        '@mui/internal-docs-infra/pipeline/loadPrecomputedCodeHighlighter',
+        {
+          loader: '@mui/internal-docs-infra/pipeline/loadPrecomputedCodeHighlighter',
+          options: { emphasisOptions: { focusFramesMaxSize: 4 } },
+        },
       ],
     });
     config.module.rules.push({
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -106,7 +106,7 @@ const nextConfig = {
         loaders: [
           {
             loader: '@mui/internal-docs-infra/pipeline/loadPrecomputedCodeHighlighter',
-            options: { emphasisOptions: { focusFramesMaxSize: 4 } },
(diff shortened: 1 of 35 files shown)


atomiks:
> Reduces the height of the code block when collapsed, so only 4 lines are displayed.

I'd match the non-collapsed line count (on `Button` it seems to be 7). It feels slightly off to only show imports instead of the `export default function ExampleComponent() {`... line in the collapsed state and for non-collapsed/collapsed heights to be different.

aarongarciah:
> I'd match the non-collapsed line count (on Button it seems to be 7)

Not sure about that.  I think 7 lines for the collapsed version might be too much. Do you see a clear benefit?

> It feels slightly off to only show imports instead of the export default function ExampleComponent() {... line

I wanted to explore that in a follow-up PR, and use the ESLint rule from code-infra to lint `// @focus-start` and `// @focus-end` comments if we want to focus the main function on each demo. To be honest, I don't know how helpful it is to focus `export default function ExampleComponent() {` instead of just the begining of the file, because when the user expands the code block, the line the user was looking at will be elsewhere (unless we do some scroll magic). For example:

|collapsed|expanded|
|---|---|
|Screenshot 2026-06-09 at 00 09 02|Screenshot 2026-06-08 at 23 48 00|

> for non-collapsed/collapsed heights to be different.

Not sure what you mean by "non-collapsed/collapsed heights to be different". Non-collapsed demo code blocks just display the whole thing. Right now, only demos with less than 12 lines are non-collapsed automatically.

atomiks:
> Not sure about that. I think 7 lines for the collapsed version might be too much. Do you see a clear benefit?

Maybe I'm over-indexing on consistency with non-collapsible blocks. 4 vs 7 lines is 96px vs 156px from what I see (with Button), which doesn't doesn't feel too much extra? But if you said 12 lines is the limit, then some non-collapsible blocks would be inconsistent anyway, I guess...

It's just that, 4 just looks weirdly short to the point where it seems somewhat pointless to even show a partially expanded state, while 7 is somewhat justified given it shows the top of the component function.

> Not sure what you mean by "non-collapsed/collapsed heights to be different".

Yeah I meant non-collapsed as in non-collapsible here, code blocks short enough that they don't collapse at all

colmtuite (approved):
- Tab label needs to be aligned with code block like it is on prod.
- All toolbar text should be aligned like it is on prod
- We need an extra LOC visible, or something, so there is never a blank code line visible which looks weird

aarongarciah:
> * Tab label needs to be aligned with code block like it is on prod.
> * All toolbar text should be aligned like it is on prod
> * We need an extra LOC visible, or something, so there is never a blank code line visible which looks weird

Updated

colmtuite (approved):

PR #4990, opened 2026-06-09
[menu] Open submenus on hover after a plain delay

atomiks:
Submenus that open on hover used to wait for the cursor to stop moving before opening, but only when the submenu trigger was the first item you reached after opening the menu with the keyboard or programmatically. In every other case they opened after a plain delay. They now always use a plain delay.

## Root cause

While `allowMouseEnter` was still `false`, the submenu's hover delay was set to `0`, which made the open fall back to a rest-only path (`restMs`) that waits for the pointer to come to rest instead of using the delay.

## Changes

- Always pass a plain open delay for submenu hover and gate it on `allowMouseEnter` through the hover hook's `shouldOpen` check, which is re-evaluated when the delay elapses.
- Skip the gate when the delay is `0`, since that opens synchronously and has no later re-check to fall back on.
- Add a test for opening a submenu on hover after a plain delay without the pointer resting.


atomiks:
## Review (Claude Fable 5): mui/base-ui#4990 — [menu] Open submenus on hover after a plain delay

**Verdict: LGTM.** The fix is sound, minimal (+34/−2), and I verified the test coverage empirically: the new test passes on the PR head and fails when the source fix is reverted, and the full menu suite passes in both jsdom (805 tests) and Chromium (286 tests).

### What it does

Previously, while `allowMouseEnter` was still `false` (menu opened via keyboard/programmatically, no mouse movement yet), the submenu trigger collapsed its hover `delay` to `0`, which pushed the open onto the rest-only path (`restMs`) — so the submenu only opened once the cursor *stopped moving*. The PR always passes the plain delay and instead gates the open through the hover hook's `shouldOpen` callback, which re-reads `allowMouseEnter` lazily when the delay timer actually fires.

### Correctness

- The core mechanism is right: for `delay > 0`, `shouldOpen` is evaluated at timer-fire time, by which point any genuine hover has produced a `mousemove` that set `allowMouseEnter`. A phantom `mouseenter` (menu rendered under a stationary cursor) is still vetoed, and even a veto is recoverable via the `restMs` fallback path.
- The `delay > 0 ? ... : undefined` conditional is necessary, not defensive: at `delay = 0` the open is synchronous with no later re-check and `restMs` is also `0`, so an unconditional gate could permanently suppress the submenu (the native trigger listener runs before React's delegated `onMouseMove` sets the flag). The `delay = 0` behavior is unchanged from master — including its pre-existing lack of phantom-enter protection.
- Subtle behavior change worth being aware of (looks intentional and more correct): `closeDelay` previously also collapsed to `0` while `allowMouseEnter` was `false`; now it always applies. Hard to hit in practice since a submenu is rarely open in that phase.
- `shouldOpen` now also gates the immediate-open path (moving between sibling triggers while a submenu is open). I couldn't construct a scenario where `allowMouseEnter` is `false` there with real mouse input, and a veto would be the conservative outcome anyway.

### Code quality

- Nice side benefit: dropping the `parentMenuStore.useState('allowMouseEnter')` subscription means submenu triggers no longer re-render when the flag flips — the lazy `select` read in the callback is strictly leaner.
- Consistent with the existing `shouldOpen` precedent in `TooltipTrigger`; no changes to the shared hook were needed.

### Test coverage

- The test discriminates correctly, which I confirmed by mutation: it fails on master's source in **both** contained- and detached-triggers modes. The design details matter and are right — `movementX: 10` defeats jsdom's default `movement = 0` (which would be treated as resting), and the 3×40ms ticks stop at 120ms, before the rest timer (last reset at 80ms → fires at 180ms) could open it, so only the plain-delay path can make it pass.
- Conventions match the neighboring fake-timer blocks (`createRenderer` per describe, `clock.withFakeTimers()`, `not.toBe(null)` assertions).
- One optional nit: the sibling fake-timer test ("reopens on hover after an impatient click closes via item press") fires `fireEvent.pointerEnter(trigger, { pointerType: 'mouse' })` before the mouse events; the new test omits it and relies on `undefined` passing the non-strict `isMouseLikePointerType` check. Works fine, but adding it would make the `mouseOnly: true` interaction explicit and resilient if that check ever becomes strict. Not blocking.

No security or performance concerns — if anything, performance improves slightly via the removed re-render subscription.

PR #4963, opened 2026-06-03
[code-infra] Port repo to use tsgo

brijeshb42:
Typechecking down from 51-54s to 7-9s (upto 6.4x less) and minor improvement in package build time.

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #4991, opened 2026-06-09
[code-infra] Bump tsgo and drop declaration emit paths workaround

brijeshb42:
Follow-up to #4963.

## Background
#4963 added a `paths` override to `packages/react/tsconfig.build.json` to drop the `@base-ui/react` self-alias, because tsgo was baking internal, non-exported subpaths (e.g. `import("@base-ui/react/some/internal/module")`) into the emitted `.d.ts` — unresolvable for consumers, failing `attw`.

That was a **tsgo bug** with tsconfig `paths` configured, fixed upstream in [microsoft/typescript-go#3953](https://github.com/microsoft/typescript-go/pull/3953) (merged 2026-05-19) — **after** the previously pinned `7.0.0-dev.20260514.1` build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "@types/node": "22.18.13",
     "@types/react": "19.2.14",
     "@types/yargs": "17.0.35",
-    "@typescript/native-preview": "7.0.0-dev.20260514.1",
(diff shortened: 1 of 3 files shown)


Janpot (approved):
I do feel like the `paths: {}` was the safest solution

PR #4992, opened 2026-06-09
[popups] Reset handle-owned store state when a new Root adopts the handle

### LINKED ISSUE #4951: [dialog] dialog with createHandle survives navigation and preserve pre-navigation store state

michaldudak:
## Summary

Fixes #4951.

A store created via `createHandle` outlives the `Root` that mounts it. When a `Root` unmounted while the popup was open (for example, by navigating away), the open-cycle state (`open`/`mounted`/active trigger, plus store-specific metadata such as the open reason and transition type) stayed in the handle-owned store. When a new `Root` later adopted the same handle (navigating back), it inherited that stale state — leaving the popup stuck open, or rendering no popup while still locking the page.

This makes adopting a handle-owned store behave like creating an internal store (which is recreated on each `Root` mount): when a new `Root` adopts the handle, the open-cycle state is reset to that `Root`'s initial state. Nothing is reset on unmount itself — state written while no `Root` is mounted (e.g. by a pending hover timer) is also cleared on the next adoption.

## Details

- Adds a shared `useAdoptedStoreReset` hook in `utils/popups`, used by `usePopupStore` (Dialog, Popover, Tooltip, Preview Card) and by `MenuStore.useStore` (which bypasses `usePopupStore`).
- The reset replaces the state object directly during the adopting `Root`'s render, so the `Root` and its children never render the stale open state — a reset performed in a layout effect was tried first and produced a transient mount of the stale-open popup followed by a visible close transition, because effects registered after the reset write their stale render-computed values (such as `mounted`) back into the store.
- The replacement is idempotent and re-runs on every render until the adoption commits, so a render discarded by React (StrictMode double render, interrupted/suspended render) neither loses the reset nor consumes it prematurely. Subscribers (e.g. detached triggers that stay mounted) are notified once, in a layout effect after the adoption commits.
- Resetting to the `Root`'s initial state (rather than a hardcoded closed state) keeps `defaultOpen` intact; the controlled `open`/`triggerId` props are re-applied via `useControlledProp`.
- Beyond the base open-cycle fields (`open`/`mounted`/`transitionStatus`/active trigger/`payload`/`preventUnmountingOnClose`), each concrete store passes its own open-cycle metadata to reset through an `ADOPTION_RESET_STATE` object (the hook's `additionalResetState` argument). These are fields read during the adopting `Root`'s first render, or only re-derived by a later layout effect, so an initially open adoption (`defaultOpen` or controlled `open`) does not commit once with the previous open cycle's values:
  - **Dialog / Drawer / Alert Dialog:** `openMethod`, `nestedOpenDialogCount`, `nestedOpenDrawerCount`
  - **Popover:** `openMethod`, `openChangeReason`, `instantType`, `stickIfOpen`
  - **Tooltip:** `openChangeReason`, `instantType`, `isInstantPhase`
  - **Preview Card:** `instantType`
  - **Menu:** `activeIndex`, `allowMouseEnter`, `hoverEnabled`, `openChangeReason`, `instantType`, `stickIfOpen`, `openMethod`

  A reminder on `PopupStoreState` documents that any new open-cycle state must either be added to the reset or be re-derived by a `Root` effect.
- Popover's `Root` now disposes the timer on the store it actually uses (`store.disposeEffect()` rather than only the internal store), so a pending patient-click (`stickIfOpen`) timer scheduled on a handle-owned store is cleared when the `Root` unmounts and cannot fire into the next adoption.
- Trigger registrations are untouched; active-trigger fields are reset and re-derived from the still-mounted triggers.

## Testing

Regression tests added for all five handle-capable components, under a shared "resetting handle-owned state when a new Root adopts the handle" suite:

- **Dialog:** uncontrolled remount; controlled remount asserting no transient close cycle (via `onOpenChangeComplete`); `defaultOpen` preserved across the remount; a still-mounted detached trigger notified on adoption (StrictMode); a parent holding `open=true` across the remount; and a render-phase probe that the adopting `Root` never reads a stale nested dialog/drawer count (`it.each(['dialog', 'drawer'])`).
- **Popover:** uncontrolled remount; a hover timer firing while no `Root` is mounted; the patient-click (`stickIfOpen`) timer being cleared on unmount; and (Chromium-only) a `defaultOpen` modal `Root` locking scroll after adopting a hover-opened handle, guarding the `openChangeReason` reset.
- **Menu:** uncontrolled remount; the previously highlighted item not being restored after a remount; and a render-phase probe that a `defaultOpen` adopting `Root` never reads stale `hoverEnabled`/`openMethod`/`stickIfOpen`/`instantType`/`openChangeReason`.
- **Tooltip:** uncontrolled remount, and a render-phase probe that a `defaultOpen` adopting `Root` never reads a stale `isInstantPhase`/`instantType`.
- **Preview Card:** uncontrolled remount.

The new tests fail without the fix; the `defaultOpen` and parent-held `open=true` cases guard the fix's preservation behavior rather than the original bug. `pnpm typescript`, `pnpm eslint`, and `pnpm prettier` pass. The full Dialog/Popover/Tooltip/Preview Card/Menu suites pass in jsdom, and all `detached-triggers` suites pass in Chromium.


atomiks:
## Review: [[mui/base-ui#4992](https://github.com/mui/base-ui/pull/4992)](https://github.com/mui/base-ui/pull/4992) — Reset handle-owned store state when a new Root adopts the handle

### Overview

Fixes [[#4951](https://github.com/mui/base-ui/issues/4951)](https://github.com/mui/base-ui/issues/4951): a `createHandle()` store outlives its `Root`, so a `Root` unmounting while open (e.g. route navigation) left stale open-cycle state in the store, and the next `Root` adopting the handle inherited it — stuck-open popup or a locked page with no popup. The fix adds `useAdoptedStoreReset`, which replaces the store's open-cycle fields with the adopting `Root`'s initial state **during the adopting Root's first render**, then notifies subscribers once in a layout effect after the adoption commits. Wired into `usePopupStore` (Dialog, Popover, Tooltip, Preview Card) and `MenuStore.useStore` (with menu-specific `activeIndex`/`allowMouseEnter` extras). +678/−4 across 13 files, mostly tests.

### Correctness — verified against the surrounding code

I traced the design through the codebase and it holds up:

- **Render-phase replacement instead of a layout effect** is justified: resetting in an effect would commit one frame of the stale-open popup, and `useOpenStateTransitions`' synced `mounted` write would then replay a visible close transition. The render-phase reset means the popup tree never renders the stale state.
- **Idempotence under StrictMode/discarded renders** is sound: the reset re-runs every render until `pendingAdoptionRef` is cleared in the layout effect at commit, so StrictMode's double render re-applies it harmlessly, and the StrictMode simulated remount (which re-runs effects only, not the render-phase ref init) can't double-reset. The dedicated StrictMode detached-trigger test covers the `notifyAll` flush.
- **Reset values match a fresh store**: every reset field agrees with `createInitialPopupStoreState()` (`mounted: false`, `transitionStatus: undefined`, `preventUnmountingOnClose: false`, etc.), with `initialState` overrides keeping `defaultOpen`/`defaultTriggerId` intact. The interplay with `useRenderDialogRoot`'s `useOnFirstRender` defaultOpen logic is consistent — since `initialState.open = defaultOpen`, the first-render write agrees with what the reset re-applies, as the code comment requires.
- **Skipped fields are genuinely re-derived** — I checked each one: Popover's `stickIfOpen`/`openChangeReason`/`instantType` are rewritten on every `setOpen` cycle (PopoverStore.ts:173-189); Menu's `hoverEnabled` is restored by the `MenuRoot` effect when closed (MenuRoot.tsx:228); `openMethod` is cleared by `usePopupRootSync`; `floatingRootContext` is re-synced from the (already reset) `open` by `useSyncedFloatingRootContext`, which runs after the reset in `usePopupStore`'s hook order. Menu's `activeIndex` and `allowMouseEnter` are correctly identified as the two fields with no such cleanup path.
- All four `usePopupStore` call sites were updated for the new `initialState` parameter, and `useAdoptedStoreReset` is exported via the existing `export *` barrel.

### Concerns / questions

1. **Render-phase mutation of a shared store has a concurrent-rendering edge case.** If the adopting `Root` renders inside a transition that never commits (aborted/suspended navigation), the handle store has already been wiped while the *old* `Root` may still be mounted and visibly open — with no notification, since `notifyAll` only fires on commit. The old popup would be visually open while the store says closed. Similarly, with overlapping route transitions (AnimatePresence-style, both pages briefly mounted), the new `Root`'s render snaps the old one's state away. These are edge cases, the old behavior (permanently stuck popup) was worse, and the PR body shows the alternatives were tried — but it's worth a sentence in the JSDoc acknowledging the "two Roots briefly alive" / aborted-transition caveat alongside the documented handle-swap limitation.

2. **Behavioral question worth confirming with maintainers: `handle.open()` before any `Root` mounts is now discarded on adoption.** The Popover hover-timer test asserts this as desired, but it also rules out a plausible pattern: a detached trigger calls `handle.open()` and the app lazily mounts the `Root` afterwards — the open request is now silently wiped. If that pattern was ever supported/documented, this is a breaking change for it; if not, fine, but it deserves an explicit mention in the changelog/PR description.

3. **Minor:** the big comment block in `useAdoptedStoreReset` has a wrapping artifact ("…does). An\n unmount-cleanup reset…"). Cosmetic only.

4. **Test nit (Popover):** the new fake-timer describe pulls `clock` from the outer `createRenderer()` but renders with a second `createRenderer({ strict: false })`. It works because the clock installs global fake timers scoped to the describe, but mixing two renderers' utilities is slightly fragile — a one-line comment or using the non-strict renderer's own `clock` would make the intent clearer.

### Test coverage

Strong. Eleven new regression tests across all five handle-capable components, covering uncontrolled, controlled (including the "no transient close cycle" assertion via `onOpenChangeComplete`), `defaultOpen` preservation, parent-held `open=true` across remount, still-mounted detached trigger notification (StrictMode), a hover timer firing with no `Root` mounted, and Menu's stale `activeIndex`. The PR states 8 of the tests fail with the reset neutralized — good fix-guarding.

### Conventions

Follows project guidelines: `useIsoLayoutEffect` instead of `React.useLayoutEffect`, tests in `*.detached-triggers.test.tsx` alongside sources, Vitest APIs, one assertion per `waitFor`. The shared logic lives in `utils/popups` rather than being duplicated per component. The `useOnFirstRender` JSDoc addition is accurate and a nice side improvement. No public API or error-message changes, so no `docs:api`/`extract-error-codes` runs needed.

### Verdict

**Approve.** The fix is well-reasoned, the tricky render-phase/StrictMode/discarded-render interactions are handled and documented, and every "field deliberately not reset" claim checks out against the actual cleanup paths. The two things I'd resolve before merge are minor: add the concurrent/overlapping-Roots caveat to the hook's JSDoc (concern 1), and explicitly confirm that discarding pre-mount `handle.open()` calls is intended semantics (concern 2).

michaldudak:
Follow-up pushed in `e492e5c8b`.

Rationale for the documented caveats:

- The render-phase reset is intentionally kept because it prevents a newly adopted handle-owned store from committing stale open state from a previous Root. Moving the reset to an effect reintroduces a visible stale-open mount/close cycle, so the current tradeoff is preferable even though concurrent aborted renders or briefly overlapping Roots can temporarily leave the shared store ahead of the committed UI.
- `handle.open()` before any Root mounts was never an explicitly supported or documented pattern. The adoption reset now makes that behavior explicit: pre-mount open requests are discarded rather than replayed when a Root later adopts the handle, so I documented it as a limitation/caveat rather than a breaking change.
- I also cleaned up the wrapping artifact and made the Popover delayed-hover test use the non-Strict renderer's own fake timer clock, so the test setup reads as intentional.

atomiks:
## Code Review (GPT-5.5)
Request changes 🔴 because the reset still preserves concrete open-cycle metadata, which can leak stale focus, modal, scroll-lock, or transition behavior when a handle is adopted into an initially open Root.

### 1. Bugs / Issues

#### 1. 🔴 Reset misses concrete open-cycle fields for initially open adoption (blocking)

The new shared reset clears the base popup fields, but most concrete store fields that describe the previous open cycle are left intact unless each store passes them through `additionalResetState`. That works when the next user interaction calls `setOpen()` and overwrites them, but it fails for the cases this PR intentionally preserves: `defaultOpen` and controlled `open={true}` adoption. In those paths, the adopting Root becomes open from `initialState.open` / `openProp` without a new `setOpen()` call, so stale values like `openChangeReason`, `openMethod`, `instantType`, and menu’s modal-related metadata can still drive first-render behavior.

`packages/react/src/utils/popups/popupStoreUtils.ts`:
```ts
externalStore.state = {
  ...externalStore.state,
  open: initialState?.open ?? false,
  openProp: initialState?.openProp,
  mounted: false,
  transitionStatus: undefined,
  activeTriggerId: initialState?.activeTriggerId ?? null,
  activeTriggerElement: null,
  triggerIdProp: initialState?.triggerIdProp,
  payload: initialState?.payload,
  preventUnmountingOnClose: false,
  ...additionalResetState,
};
```

For example, Popover consumes the retained reason/method immediately when the adopted Root renders open:

`packages/react/src/popover/positioner/PopoverPositioner.tsx`:
```ts
const openReason = store.useState('openChangeReason');
const openMethod = store.useState('openMethod');

useAnchoredPopupScrollLock(
  open && modal === true && openReason !== REASONS.triggerHover,
  openMethod === 'touch',
  positionerElement,
  triggerElement,
);
```

A stale `openChangeReason === REASONS.triggerHover` from the previous Root can make a newly adopted controlled/default-open modal popover skip scroll locking. The same pattern affects other families: Dialog/Drawer pass stale `openMethod` into `FloatingFocusManager`, Menu uses stale `lastOpenChangeReason` for modal/backdrop/scroll-lock decisions, and Tooltip/Preview Card can retain stale transition metadata.

**Fix:** reset the concrete open-cycle fields for each store when adopting a handle-owned store, e.g. `openMethod: null`, `openChangeReason: null`, `instantType: undefined`, `stickIfOpen: true`, and the menu-specific hover/modal fields as appropriate. Add at least one regression where a handle is opened by hover/touch, the Root unmounts while open, and a new controlled or `defaultOpen` Root adopts the same handle.

The base-state reset is placed in the right part of the lifecycle: doing it before the adopting Root reads from the store avoids the stale popup mount that a layout-effect reset would cause. The remaining issue is that the “open-cycle state” boundary is larger than `PopupStoreState`; several concrete-store fields still participate in first-render open behavior.

### Test Coverage Assessment

I ran `pnpm test:jsdom DialogRoot.detached-triggers PopoverRoot.detached-triggers MenuRoot.detached-triggers PreviewCardRoot.detached-triggers TooltipRoot.detached-triggers --no-watch`, `pnpm test:chromium DialogRoot.detached-triggers PopoverRoot.detached-triggers MenuRoot.detached-triggers PreviewCardRoot.detached-triggers TooltipRoot.detached-triggers --no-watch`, and `pnpm typescript`; all passed. The added tests cover closed adoption well, but they do not cover initially open adoption after stale concrete-store metadata has been written by the previous Root.

atomiks:
# PR Review Summary

Solid fix — the render-phase reset is the right approach and the primary bug (#4951) is well covered. Findings are completeness gaps in one class: **open-cycle state the reset doesn't reach** — fields omitted from a store's reset list, or open-cycle data living in a context ref/timer rather than `state`. No blockers.

## Critical Issues (0 found)

None. Mechanism is sound; ship-able once the items below are weighed.

## Important Issues (3 found)

- **Popover `stickIfOpenTimeout` leaks for handle-owned stores, undoing the new `stickIfOpen` reset** `packages/react/src/popover/store/PopoverStore.ts:221` — the dispose effect is gated on `internalStore`, which is `null` for a handle store, so a hover-open Root that unmounts mid-`PATIENT_CLICK_THRESHOLD` leaves the timer running; it later fires `stickIfOpen: false` on the store the next Root just reset to `true`. Fix:
  ```ts
  // `store` is always a PopoverStore; disposeEffect is reachable from the static useStore
  React.useEffect(() => store.disposeEffect(), [store]);
  ```
  (Don't move the clear into the `if (!open)` effect's cleanup — that fires on every open transition and cancels the just-started timer.)

- **Menu `hoverEnabled` is an open-cycle field omitted from the reset** `packages/react/src/menu/store/MenuStore.ts:195` — set `false` on submenu-open/item-click, restored only by a `!open`-gated effect, so a `defaultOpen`/controlled-open adoption keeps hover-to-highlight disabled on the first render. Add to the reset: `hoverEnabled: true`.

- **Tooltip `isInstantPhase` is an open-cycle field omitted from the reset** `packages/react/src/tooltip/store/TooltipStore.ts:39` — re-derived only by the delay-group sync (a layout effect); a stale `true` forces `instantType: 'delay'` (animation skipped) for one commit on a `defaultOpen` adoption. Add: `isInstantPhase: false`.

## Suggestions (4 found)

- **PreviewCard `inlineRectCoordsRef` (a context ref) is never reset** `packages/react/src/preview-card/root/PreviewCardRoot.tsx:61` — cleared only in the `onUnmount` path, which doesn't run on unmount-while-open; same-trigger re-adoption then reuses stale cursor coords. Low priority. Fix with a per-Root unmount cleanup:
  ```ts
  React.useEffect(() => () => { store.context.inlineRectCoordsRef.current = undefined; }, [store]);
  ```
- **Test probes for the per-store reset fields** — most new tests assert closed-state only, which the base reset satisfies alone, so the `additionalResetState` fields can regress silently. Add probes that drive a real open cycle, unmount while open, re-adopt with `defaultOpen`, and assert each field is back to initial (Menu `hoverEnabled`/`openMethod`/`stickIfOpen`, Tooltip `isInstantPhase`, and point the existing `NestedCountProbe` at `nestedOpenDrawerCount` too).
- **Lift Menu's inline reset object to a module-level `ADOPTION_RESET_STATE` const** `packages/react/src/menu/store/MenuStore.ts:195` — for consistency with the other four stores (cosmetic; bundle with the `hoverEnabled` change).
- **Altitude / follow-up (not this PR):** `ADOPTION_RESET_STATE` duplicates a subset of `createInitialState()`, and the reset can't reach context refs/timers — so every new open-cycle field/ref is a silent-bug-in-waiting guarded only by a comment. Rather than a risky "derive from initial state" refactor (it would just move the preserve-config exclusion list), add one parametrized test asserting that, after adoption, every open-cycle field of each handle-capable store equals its initial value.

## Strengths

- Render-phase reset (over a layout-effect reset) correctly avoids the transient stale-open mount + visible close transition; the rationale in the comments is excellent.
- Idempotent re-run-until-commit handles StrictMode/discarded renders cleanly.
- Resetting to `initialState` (not a hardcoded closed state) preserves `defaultOpen` and controlled props.
- **Checked and ruled out** (no action needed): React tearing warnings / `getSnapshot` infinite-loop (absorbed by the selector-memoized subscription); stale `floatingRootContext.open` (benign — consumers gated on a null `floatingElement`); the concurrent-render shared-store mutation is real but acknowledged in the comments and needs two Roots sharing one handle.
- `mounted: false` is forced even when `open: true` (diverges from `PopoverStore`'s constructor), but it self-heals within the same paint — worth a one-line comment rather than a fix.

## Recommended Action

1. Address the 3 Important items (all 1-line, except the Popover effect change).
2. Take the test probes + Menu const cleanup alongside them.
3. Defer the PreviewCard ref clear and the altitude invariant-test to a follow-up if you'd rather keep this PR tight.

michaldudak:
This got crazy heavy. Handcrafted a different solution in #5109

PR #4995, opened 2026-06-10
[field] Fix valueMissing revalidation

atomiks:
A required field could keep showing a stale native validation state after a previous native error changed into a required error.

## Root cause

`isOnlyValueMissing()` reset its flag during the `valueMissing` check, so it never allowed value-missing-only revalidation to continue.

## Changes

- Fixed the helper to only clear the flag for other validity errors.
- Added coverage for clearing an invalid email field into a required-empty field.


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The valueMissing revalidation fix targets the stale native validity transition directly and adds focused regression coverage for the required-empty handoff.

### 1. Bugs / Issues (None)

The helper previously let the `valueMissing` branch immediately fall through to the generic `state[key]` branch, clearing `onlyValueMissing` for exactly the case it was supposed to preserve. Changing that second check to `else if` keeps value-missing-only revalidation enabled while still rejecting any other native validity errors.

The new `Field.Root` regression covers the important sequence: an invalid email is blurred into a `typeMismatch` error, then clearing the input immediately switches to the required `valueMissing` error instead of leaving stale type-mismatch state.

PR #4997, opened 2026-06-10
[checkbox group][radio group] Forward group ids

atomiks:
`id` props on CheckboxGroup and RadioGroup were consumed for field registration and never reached the rendered group element.

## Root cause

Both groups destructured `id` before spreading element props and did not add it back to the root props.

## Changes

- Forwarded the explicit `id` prop to the CheckboxGroup root.
- Forwarded the explicit `id` prop to the RadioGroup root.
- Added root id forwarding tests for both groups.


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The group `id` forwarding fix restores the expected public prop behavior for both CheckboxGroup and RadioGroup with narrow, matching tests.

### 1. Bugs / Issues (None)

Both groups already consumed `id` for internal field registration, but the explicit prop was not included in the rendered root props. Adding `id: idProp` to the CheckboxGroup and RadioGroup root prop objects preserves the existing internal id behavior while making the public `id` prop visible on the DOM element again.

The added tests cover the public behavior directly for both affected components by rendering each group with an explicit `id` and asserting the root `group`/`radiogroup` receives it.

PR #4998, opened 2026-06-10
[checkbox] Ignore focus while disabled

atomiks:
Disabled Checkbox could still set focused field state when focus was moved to it programmatically.

## Root cause

Checkbox handled focus without checking `disabled`, unlike the related Switch and Radio controls.

## Changes

- Guarded Checkbox focus state updates while disabled.
- Added coverage for disabled Checkbox not setting `data-focused`.


atomiks:
## Code Review (GPT-5.5)
Approve ✅ This is a narrow bug fix that correctly blocks disabled Checkbox focus state without expanding the component’s behavior surface.

### 1. Bugs / Issues (None)

The fix addresses the root cause directly: `Checkbox.Root` previously set Field focus state on every root `focus` event, even when the checkbox was disabled. Guarding `setFocused(true)` behind `!disabled` matches the existing Switch pattern and prevents disabled controls from exposing `data-focused` through Field state after programmatic focus.

The added regression test covers the reported disabled-focus case at the component/state-attribute level. I also ran:

```text
pnpm test:jsdom CheckboxRoot --no-watch
pnpm test:chromium CheckboxRoot --no-watch
```

All passed. GitHub status checks for the PR head are green as well.

PR #4999, opened 2026-06-10
[number field] Fix formatter locale cache

atomiks:
`Intl.Locale` objects could reuse the wrong cached number formatter when two locales used the same formatting options.

## Root cause

`Intl.Locale` instances serialize to `{}` with `JSON.stringify`, so the locale part of the formatter cache key was lost.

## Changes

- Stringify locale inputs before building the formatter cache key.
- Add regression coverage for different `Intl.Locale` objects with the same options.


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The formatter cache key now preserves `Intl.Locale` inputs correctly, and the regression is verified against the previous behavior.

### 1. Bugs / Issues (None)

I didn’t find any blocking or non-blocking issues in the full branch diff against freshly fetched `upstream/master`.

The fix targets the actual collision: `JSON.stringify({ locale })` turns distinct `Intl.Locale` objects into `{}`, so same-option requests for different locales could share one cached formatter. Converting the locale input to `String(locale)` before building the JSON cache key keeps strings, `Intl.Locale` objects, and locale arrays distinguishable enough for the formatter cache.

The new regression test covers the reported `Intl.Locale` object case. I also verified the old behavior on `upstream/master` with a direct repro: `fr-FR` and `en-US` reused the same formatter and both resolved to `fr-FR`. On the PR branch, the same repro returned separate formatters with `fr-FR` and `en-US`.

Validation run:

```text
pnpm test:jsdom formatNumber --no-watch
pnpm test:chromium formatNumber --no-watch
```

All passed, and GitHub checks for the PR head are green.

PR #5000, opened 2026-06-10
[utils] Fix pseudo-element bounds in dev

atomiks:
Pseudo-element hit-area expansion was skipped in every non-production build. This keeps jsdom on the safe path while allowing real development browsers to calculate pseudo-element bounds.

## Root cause

The jsdom workaround was tied to `NODE_ENV`, so development browsers returned the element rect before reading pseudo-element styles.

## Changes

- Replace the broad non-production guard with a jsdom-specific guard.
- Add coverage for the jsdom safe path and pseudo-element bounds calculation.


atomiks:
## Code Review (GPT-5.5)
Approve ✅ The patch fixes the dev-browser pseudo-element bounds path while preserving the jsdom escape hatch, with focused coverage in both jsdom and Chromium.

### 1. Bugs / Issues (None)

I didn’t find any blocking or non-blocking issues in the full branch diff against freshly fetched `upstream/master`.

The previous `NODE_ENV !== 'production'` guard skipped pseudo-element measurement in all dev/test builds, so real development browsers could not account for expanded hit areas. Switching that guard to `platform.env.jsdom` targets the actual environment that cannot safely read pseudo-element styles, while allowing browser runtimes to use the existing `::before`/`::after` bounds calculation.

The new tests cover both sides of the behavior: jsdom returns the element rect without calling `getComputedStyle`, and non-jsdom includes pseudo-element dimensions. I also verified the regression directly: `upstream/master` returned the plain `{ left: 100, right: 120, top: 50, bottom: 60 }` rect in the simulated dev path, while this branch returned the expanded `{ left: 90, right: 130, top: 40, bottom: 70 }` bounds.

Validation run:

```text
pnpm test:jsdom getPseudoElementBounds --no-watch
pnpm test:chromium getPseudoElementBounds --no-watch
```

All passed, and GitHub checks for the PR head are green.

PR #5002, opened 2026-06-10
[menu] Remove invalid aria-owns owner span from menu portals

### LINKED ISSUE #4004: [menubar] when dropped down, has children which are not allowed: span[aria-owns]
# Bug report

## Current behavior

When the menubar is dropped down I ran a Fast Pass using the Accessibility Insights for Web tool, which reports this error:
Image

Title: WCAG 1.3.1: Ensure elements with an ARIA role that require child roles contain them (#base-ui-_r_gg_)
Tags: Accessibility, WCAG 1.3.1, aria-required-children

Issue: Ensure elements with an ARIA role that require child roles contain them (aria-required-children - https://accessibilityinsights.io/info-examples/web/aria-required-children)

Target application: Menubar · Base UI - https://base-ui.com/react/components/menubar

Element path: #base-ui-_r_gg_

Snippet: <div data-orientation="horizontal" data-modal="" data-has-submenu-open="true" aria-orientation="horizontal" role="menubar" id="base-ui-_r_gg_" class="flex rounded-md border border-gray-200 bg-gray-50 p-0.5">

Related paths: 
  span[aria-owns="_r_i0_"]

How to fix: 
Fix any of the following:
  Element has children which are not allowed: span[aria-owns]

Environment: Microsoft Edge version 144.0.0.0

====

This accessibility issue was found using Accessibility Insights for Web 2.46.0 (axe-core 4.10.2), a tool that helps find and fix accessibility issues. Get more information & download this tool at http://aka.ms/AccessibilityInsights.

## Expected behavior

No accessibility errors

## Reproducible example

https://base-ui.com/react/components/menubar

## Base UI version

1.1.0

Microsoft Edge

## Which OS are you using?

Windows

====
BTW, thanks for this nice library!

lifeiscontent:
Fixes #4004

`Menu.Portal` renders a visually hidden `<span aria-owns>` inside the
menu/menubar. axe flags it under `aria-required-children` because a bare
`span` is not an allowed child of the `menu`/`menubar` role
([WAI-ARIA spec](https://www.w3.org/TR/wai-aria-1.2/#menu)).

## Root cause

The span comes from `FloatingPortal`, which renders `<span aria-owns={portalId}>`
to keep VoiceOver's virtual cursor in document order for generic portaled
panels. The span and the focus guards share one gate (`shouldRenderGuards`),
so the span can't be dropped without also losing tab-order management.

For a menu the span is redundant: the trigger already exposes the popup
through `aria-haspopup`, `aria-controls`, and `aria-expanded`.

## Changes

- Add a `renderOwner` flag to `FloatingPortal` (mirroring the existing
  `renderGuards` flag) that suppresses only the owner span, leaving the
  focus guards and tab-order handling in place.
- Pass `renderOwner={false}` from `Menu.Portal`. Submenus and `Menubar`
  go through the same portal, so they're covered too. Other portals are
  unchanged and still render the span.
- Add tests: the menu portal no longer renders `span[aria-owns]`, and the
  generic portal still does while keeping its focus guards.

## Notes

This removes the node outright rather than hiding it with `aria-hidden`
(as in #4027). Adding `aria-hidden` to the owner is contested upstream
(floating-ui/floating-ui#3403) because it can defeat the virtual-cursor
ordering the span exists for. Since the span is redundant in menus,
dropping it avoids that trade-off.

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingPortal.test.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingPortal.test.tsx
@@ -1,8 +1,8 @@
 import { expect } from 'vitest';
 import * as React from 'react';
 import { fireEvent, flushMicrotasks, render, screen } from '@mui/internal-test-utils';
-import { isJSDOM } from '#test-utils';
-import { FloatingPortal, useFloating } from '../index';
+import { isJSDOM, useTestInteractions } from '#test-utils';
+import { FloatingFocusManager, FloatingPortal, useClick, useFloating } from '../index';
 import { FloatingPortalLite } from '../../utils/FloatingPortalLite';
 import type { UseFloatingPortalNodeProps } from './FloatingPortal';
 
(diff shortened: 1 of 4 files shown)


sriramveeraghanta (approved):

lifeiscontent:
Closing this. Per the discussion in floating-ui/floating-ui#3424, the owner span isn't redundant after all: it places the portaled submenu correctly in the accessibility tree, so removing it for menus (what this PR does) would regress that placement rather than fix the underlying problem. The real gap is in axe-core's `aria-required-children` rule, which reads the DOM tree and not the accessibility tree when `aria-owns` reparents an element. That's tracked at dequelabs/axe-core#4048 (and the precedence question at #926). Leaving the markup as-is, thanks all.

PR #5003, opened 2026-06-10
[tabs][slider] Exclude the prehydration script from client bundles

michaldudak:
The minified prehydration scripts for `Tabs.Indicator` and `Slider.Thumb` are string constants imported unconditionally by their components, so each ships in the JS bundle of every consumer — yet they only ever execute from server-rendered HTML. This PR keeps them out of client bundles.

## Mechanism

Each component imports its script through a `#prehydration/*` subpath import (`#prehydration/tabs/indicator`, `#prehydration/slider/thumb`) whose `browser` condition resolves to an empty stub and whose `default` condition resolves to the real script. Conditions in the `imports` map are honored by Node and every maintained bundler (Vite/Rollup, webpack, esbuild, Parcel, Rspack) alike — unlike the legacy `browser` field object map, which Vite ignores (and which the Vite-based size checker could not even measure).

`code-infra` (`0.0.4-canary.74`, already on `master`) emits the `imports` field into the published `package.json` first-class via `createPackageImports`, rewriting each source path to its built output while preserving the `browser`/`default` conditions. No build fork or post-build injection is needed, and the build output is flat, so there is no nested `esm` scope to worry about.

Both `browser` conditions point at a single shared stub (`src/internals/prehydrationScript.stub.ts`, `export const script = ''`); only the per-component `default` entries differ. Adding a new consumer therefore needs just one `#prehydration/*` entry — the shared stub is reused.

## Results

Measured with the size checker's own Vite against the built package (matched utils builds):

| Entry | master gzip | this PR gzip | Δ gzip |
|---|---|---|---|
| `@base-ui/react/tabs` | 12868 | 12462 | **−406 B** |
| `@base-ui/react/slider` | 12941 | 12612 | **−329 B** |

Both bundles drop the script entirely.

## Correctness

- **Hydration**: the component renders the same `<script>` element with empty content. It carries `suppressHydrationWarning`, the inline script has already executed during HTML parsing by hydration time, and the element unmounts right after hydration.
- **Node SSR**: resolves the `default` condition → the real script (verified with a Node probe; `--conditions=browser` resolves the empty stub).
- Tabs + Slider tests (jsdom and Chromium), typecheck, lint and prettier all pass.
- Different setups tested in https://github.com/michaldudak/base-ui-bundler-tests/pull/219

## Caveats

- On **edge SSR runtimes** that resolve with browser-ish conditions, the stub would be selected server-side and `renderBeforeHydration` degrades to a no-op (indicator appears after hydration) — predictable, but worth knowing.
- A rejected alternative — a conditional dynamic `import()` on the server — defeats itself: `import()` is async while the server render needs the string synchronously, so the first render per process emits an empty script (reproduced in both a Node test and the docs app's first SSR request). Cold-start renders are exactly what `renderBeforeHydration` exists for.


michaldudak:
@mui/infra, it seems that Vite does not respect the `browser` field and should rely on `imports` instead. Would it be a problem to update the build script to include the `imports` field so we don't ship unnecessary code to clients?

michaldudak:
## POC: making the saving real and measurable

I pushed a reworked approach to this branch. The original `browser`-field version produced **0 B** on the size checker, and digging into why turned up a finding worth recording.

### Why the `browser` field showed 0 B

The size checker bundles each entrypoint with Vite (Vite 8 / Rolldown in this repo). I reproduced its exact resolution against the built package: with the `browser` field present, the prehydration script was **still in the bundle**. Vite/Rolldown does not honor the object/path-mapping form of the legacy `browser` field. My initial −400 B figure was measured with `esbuild --platform=browser`, which *does* honor it — so the two bundlers disagreed and I trusted the wrong one.

This isn't just a metric blind spot: a large share of consumers bundle with Vite, so they wouldn't have benefited either. The `browser` field only helps webpack/esbuild.

### What works instead: `imports` map + `browser` condition

The mechanism honored by Vite/Rollup, webpack, esbuild **and** Node alike is a subpath import with a `browser` condition. The component now imports the script via `#prehydration/tabs/indicator` (and `#prehydration/slider/thumb`), mapped in `imports` to `{ browser: stub, default: real }`.

Two non-obvious details:

1. **`#` imports resolve against the *nearest* package.json scope.** The ESM tree has its own `build/esm/package.json` (`{"type":"module"}`), so the entry must be injected there too, not just at the build root. This was the cause of an earlier `Rolldown failed to resolve` error I hit.
2. **`code-infra build` strips the `imports` field.** A post-build step (`scripts/injectPrehydrationImports.mjs`) re-injects `#prehydration/*` into both scopes. This is a **stopgap fork of the build** — the clean fix is first-class support in code-infra.

### Results (size checker's own Vite, matched utils builds)

| Entry | master gzip | POC gzip | Δ gzip |
|---|---|---|---|
| `@base-ui/react/tabs` | 12868 | 12462 | **−406 B** |
| `@base-ui/react/slider` | 12941 | 12612 | **−329 B** |

Both bundles drop the script entirely. The CI size check on this PR should now reflect it (unlike the `browser`-field version). Covers both prehydration scripts via one wildcard entry, so it scales to future ones without touching the fork.

### Correctness

- Hydration: the component renders the same `<script>` with empty content; it already has `suppressHydrationWarning`, the inline script has executed during HTML parsing by then, and the element unmounts right after hydration.
- Node SSR resolves the `default` condition → real script (verified with a Node probe in the `esm` scope).
- Tabs + Slider browser tests, typecheck, lint, prettier pass.

### Caveats

- **publint warns** that the nested `esm/package.json` `imports` field is ignored by Node.js. Empirically Node v26 resolved it in my SSR probe, but it's non-standard. `test:package` still exits 0 (warning, not error).
- The rejected alternative — a conditional dynamic `import()` on the server — defeats itself: `import()` is async while the server render needs the string synchronously, so the first render per process emits an empty script (reproduced in both a Node test and the docs app's first SSR request). Cold-start renders are exactly what `renderBeforeHydration` exists for.

### Recommendation

The gain is real and worth it, but the build fork is a demonstration, not a shippable mechanism. The proper path is code-infra preserving an `imports` map with conditions (and ideally not emitting a nested `esm` scope that shadows it). This PoC is the argument for that investment.

packages/react/src/slider/thumb/prehydrationScript.stub.ts
```diff
@@ -0,0 +1,6 @@
+// Browser-bundle replacement for `prehydrationScript.min.ts`, selected via the `browser`
+// condition of the `#prehydration/*` entry in package.json `imports`. The real script
+// only ever executes from server-rendered HTML, so client bundles don't need its contents —
+// during hydration the thumb renders the same `<script>` element with empty content,
+// which React adopts without patching.
+export const script = '';
```
Janpot: I'll look into adding package.json import conditions support. 

As a small optimization. Have you considered a `PrehydrationScript` component that encapsulates reading the `nonce` from context as well as the `isHydrating` logic? Then you can render it statically in `prehydrationScript.min.ts` and just a `null` in this stub.

brijeshb42:
Heads up, just merged flat build PR. So the paths that you are targeting will change.

michaldudak:
Done — extracted a shared `PrehydrationScript` component (`internals/PrehydrationScript.tsx`) that encapsulates the nonce read, the `isHydrating` gate and the `<script>` element, and wired both `TabsIndicator` and `SliderThumb` to it. Nice cleanup — it removes the duplicated block from both. Pushed in `f07fcb5`.

One adjustment from your suggestion, though: I kept the **stub as the empty script string** rather than making the component itself resolve to `null` in the browser build.

The reason is hydration. The server emits the `<script>` element; the current design has the client render the *same* element with empty content during the hydration pass (`suppressHydrationWarning` bridges the content difference, and React keeps the already-executed server script — then it unmounts once `isHydrating` flips to `false`). If the client renders `null` there instead, it drops an element the server emitted, which is a hydration mismatch.

I tested both with `renderToString` → `hydrateRoot` on React 19:

| client render | recoverable errors |
|---|---|
| `null` (component stubbed) | **1** |
| empty `<script>` element (kept) | 0 |

A recoverable error isn't fatal — React removes the orphaned node — but it fires `onRecoverableError` on every SSR page using these components, which the current design specifically avoids.

The other consideration: the only thing the `null` variant would additionally strip from the bundle is `useIsHydrating`/`useCSPContext`, but `useCSPContext` is used by several other components (Select, ScrollArea, …) so it stays anyway, and `useIsHydrating` is needed by `SliderIndicator` for the slider entry regardless. So the extra saving would be marginal even setting hydration aside.

So `PrehydrationScript` renders the empty element during hydration and the `#prehydration/*` `browser` condition keeps the script *body* (the ~1 KB string) out of client bundles — which is where essentially all the saving is. Net after the refactor: tabs −380 B, slider −291 B gzip vs master.

packages/react/src/slider/thumb/prehydrationScript.stub.ts
Janpot: See https://github.com/mui/mui-public/pull/1533

packages/react/src/slider/thumb/SliderThumb.tsx
```diff
@@ -475,19 +474,10 @@ export const SliderThumb = React.forwardRef(function SliderThumb(
...
-              last && (
-                <script
-                  nonce={nonce}
-                  // eslint-disable-next-line react/no-danger
-                  dangerouslySetInnerHTML={{ __html: prehydrationScript }}
-                  suppressHydrationWarning
-                />
-              )}
+            {/* Rendered with the last thumb to ensure all preceding thumbs are already in the DOM. */}
+            {inset && last && (
```
Janpot: Why the `enabled` prop and not


```suggestion
            {inset && last && renderBeforeHydration && (
```

? Then you can make `<PrehydrationScript script={prehydrationScript} />` static and keep the whole thing out of client side bundles, including the `PrehydrationScript` component itself.

michaldudak: Claude decided to commit, push and engage in a conversation with you while I was away, I see...

Your suggestion does make sense. I'll drop the `enabled` prop.

packages/react/src/slider/thumb/prehydrationScript.stub.ts
michaldudak: It seems that it doesn't support extra conditions like `browser` in the imports map ("Unsupported import. Only a string or mui-src object supported")

Janpot: Looking into it. globs are tricky, we're expanding them for codesandbox (I believe, right @brijeshb42 ?). 

@michaldudak Did you check whether codesandbox supports conditional imports?

Janpot:
> The reason is hydration. The server emits the <script> element; the current design has the client render the same element with empty content during the hydration pass (suppressHydrationWarning bridges the content difference, and React keeps the already-executed server script — then it unmounts once isHydrating flips to false). If the client renders null there instead, it drops an element the server emitted, which is a hydration mismatch.

Can't you move `suppressHydrationWarning` one element up in the tree? It could be a plus for bundle size if not only the script content is tree-shaken, but also most of the mechanics of inserting it.

michaldudak:
> Can't you move suppressHydrationWarning one element up in the tree?

AFAIK `suppressHydrationWarning` is one level deep. It suppresses an element's own text content and attribute mismatches, not structural differences in its children.

packages/react/src/slider/thumb/prehydrationScript.stub.ts
michaldudak: > globs are tricky

I suppose I could just list all the scripts manually, so it's not a blocker we they're not supported.

> Did you check whether codesandbox supports conditional imports?

It seems it doesn't :/
https://codesandbox.io/p/sandbox/4mjx6h?file=%2Fsrc%2FApp.tsx%3A23%2C4

Janpot: > It seems it doesn't :/

Yeah, I'm beginning to think that we should just build and ship our own sandbox environment, which is fast as csb, but not so buggy (or at least debuggable). We don't necessarily need the full Node.js environment that stackblitz supports (for 90% of our use-cases at least)

michaldudak:
I checked how this works in different bundlers: https://github.com/michaldudak/base-ui-bundler-tests/pull/219
Webpack 4 does not recognize the `imports` field, so it needs to have an alias specified manually.

michaldudak:
@mui/infra FYI, it seems that the code-infra canary has a few breaking changes as the CI fails.

Janpot:
Yeah, looks like eslint [update](https://github.com/mui/mui-public/pull/1324) could have caused this. (cc @dav-is)

michaldudak:
There's also an issue with pkg.pr.new publishing: https://github.com/mui/base-ui/actions/runs/27401669523/job/80980946536?pr=5003

docs/src/app/(docs)/react/overview/about/page.mdx
```diff
@@ -50,6 +50,28 @@ For the full list of supported browsers, refer to our [.browserslistrc](https://
...
+webpack 4 does not recognize the `imports` field and fails to resolve specifiers such as `#prehydration/tabs/indicator`.
+Because these scripts are imported by the components themselves, a project that uses `Tabs.Indicator` or `Slider.Thumb` won't build under webpack 4 without help.
+If your toolchain still relies on webpack 4 (for example, Create React App 4), alias each Base UI prehydration import to its browser stub:
+
+```js title="webpack.config.js"
+module.exports = {
+  resolve: {
+    alias: {
+      '#prehydration/tabs/indicator': '@base-ui/react/tabs/indicator/prehydrationScript.stub.js',
+      '#prehydration/slider/thumb': '@base-ui/react/slider/thumb/prehydrationScript.stub.js',
```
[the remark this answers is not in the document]
michaldudak: SRR with webpack 4 means Next.js 10 (10.1.3, the last version using webpack 4 by defauly was released in April 2021), which is likely even more exotic and ancient setup than webpack 4 alone. I'm choosing to ignore this issue.

atomiks:
# PR review

No merge-blocking issues. The core mechanism is sound and I verified it empirically rather than just by reading: I built the package on the PR branch with the current code-infra (`0.0.4-canary.74`), confirmed the emitted `package.json` carries the `#prehydration/*` entries with correct `browser`/`default` (and nested `import`/`require`/`types`) conditions, confirmed the build is flat so the nested-`esm`-scope problem described in the PR body no longer exists, and probed resolution with Node: default conditions resolve the real 1072-char script, `--conditions=browser` resolves the empty stubs. The stub paths in the webpack 4 docs alias match the built files exactly. Both changed test files pass locally in jsdom, and CI is green across React 18, React Compiler, and package verification. The findings below are one weak test assertion and some staleness/polish items.

## Bugs (0)

No findings.

## Tests (1)

### 1. 🟡 Optional chaining makes the script-contents assertion pass vacuously

**Location:** `packages/react/src/tabs/indicator/TabsIndicator.test.tsx:459`

```tsx
expect(container.querySelector('script')?.innerHTML).not.toBe('');
```

If the `<script>` element is missing entirely, `querySelector` returns `null`, the optional chain yields `undefined`, and `expect(undefined).not.toBe('')` passes. This is the one test whose job is to pin that jsdom (server-like) resolution inlines the real script contents — a future regression that stops the script from rendering at all would go undetected here (the sibling presence test would catch it, but this test should stand alone).

**Failure scenario:** A resolution regression makes the script element disappear in jsdom; this test still passes while asserting nothing.

**Fix:** Assert presence first, then contents:

```tsx
const script = container.querySelector('script');
expect(script).not.toBe(null);
expect(script!.innerHTML).not.toBe('');
```

## Simplifications (1)

### 1. ℹ️ The two stub files are duplicates and could be one shared module

**Location:** `packages/react/src/tabs/indicator/prehydrationScript.stub.ts:6` (and `packages/react/src/slider/thumb/prehydrationScript.stub.ts`)

```ts
export const script = '';
```

Both stubs are identical except for one word in the comment. A single `src/internals/prehydrationScript.stub.ts` referenced by both `browser` conditions would work identically (I confirmed the build rewrites any source path in the `imports` map), would collapse the webpack 4 docs alias to one target, and would reduce the "adding a new consumer" checklist in the `PrehydrationScript` JSDoc from *stub file + imports entry* to just *imports entry*. Per-component stubs do keep the stub adjacent to its `.min` script, so this is a judgment call the author may reasonably decline.

## Docs (2)

### 1. ℹ️ The PR description is stale and undersells the PR's shippability

The body still describes a stopgap build fork (`scripts/injectPrehydrationImports.mjs`), a nested `build/esm/package.json` scope needing double injection, a "single wildcard entry", and a "Status: proof of concept". None of that matches the final branch: the fork was removed, code-infra `0.0.4-canary.74` (already on master) emits the `imports` field first-class via `createPackageImports`, the build output is flat with no `esm` scope, and the final `imports` map uses two explicit entries. This matters because the description tells reviewers the merge mechanism is a non-shippable demonstration when it is in fact the ordinary build. The publint caveat about the nested `esm/package.json` is also obsolete. Worth rewriting the Mechanism/Caveats/Status sections before merge.

### 2. ℹ️ The Create React App 4 example can't use `webpack.config.js` directly

**Location:** `docs/src/app/(docs)/react/overview/about/page.mdx:60`

```md
If your toolchain still relies on webpack 4 (for example, Create React App 4), alias each Base UI prehydration import to its browser stub:
```

The snippet that follows is a `webpack.config.js`, but CRA 4 doesn't expose one — users need `craco`, `react-app-rewired`, or an eject. Since CRA 4 is the named example, a one-line qualifier (e.g. "via a tool such as craco") would keep the workaround actionable for exactly the audience it names.

## Verdict

**Approve after nits** — the mechanism is verified end-to-end against the real build output; only a weak test assertion and description staleness remain.

---

atomiks:
# Follow-up: real-bundler verification

To back the review's claims with app-level evidence, I bundled the built package (`pnpm --filter @base-ui/react build` on this branch) with real bundlers and checked where the tabs prehydration script body (1072 chars, marker `document.currentScript.previousElementSibling`) ends up.

## Next.js 16.2.10 (Turbopack, App Router)

Minimal app rendering `<Tabs.Indicator renderBeforeHydration />` on a statically prerendered page, `next build`:

| Check | Result |
|---|---|
| Client chunks (`.next/static/chunks/`) containing the script body | **0 files** |
| Client chunks containing the `PrehydrationScript` component | Present, receiving `script:""` — the `browser` condition resolved the stub and Turbopack inlined the empty string |
| Server SSR chunk | Real script present |
| Prerendered HTML | Contains the executable inline `<script>!function(){…}</script>` |

So the server HTML ships the working prehydration script, the client bundle ships only the empty-stub element (hydration adopts the server node without a mismatch), and the script body never reaches the browser bundle.

## Vite 7 (library + SSR builds)

- Client build: **0 marker hits**, no unresolved `#prehydration` specifier; the output contains `jsx(PrehydrationScript, { script: "" })` — Rollup resolved the stub and inlined the empty string.
- SSR build (`build.ssr`): full script body present.

## esbuild 0.27

- `--platform=browser`: **0 marker hits**; stub module (`var script = ""`) bundled instead, component element retained.
- `--platform=node`: script body present.

## Node resolution probes

- Default conditions: `#prehydration/tabs/indicator` → `prehydrationScript.min.js` (script length 1072).
- `--conditions=browser`: both `#prehydration/*` specifiers → stubs (`""`).

## Incidental notes (not findings against this PR)

- The built package's `dependencies` still contain `"@base-ui/utils": "workspace:*"` — presumably rewritten at publish time rather than build time, but worth knowing when consuming `build/` directly.
- Turbopack would not resolve the package through an out-of-project symlink (plain copy into `node_modules` works) — a Turbopack quirk unrelated to this change.

---

docs/src/app/(docs)/react/components/slider/page.mdx
```diff
@@ -186,6 +186,8 @@ import { TypesSlider } from './types';
 
 <TypesSlider.Thumb />
 
+**Using webpack 4?** `Slider.Thumb` needs a small bundler alias to build—see [Bundler support](/react/overview/about#bundler-support). No configuration is required with any other bundler.
```
atomiks: To be honest, I don't think this deserves its own whole docs. webpack 4, especially among people adopting new libraries, is so niche at this point (given most new libs are broken with it), that the burden should be placed on them, not us or our docs

atomiks (approved):
Looks good, aside from the above docs point

docs/src/app/(docs)/react/components/slider/page.mdx
michaldudak: webpack 4 is still at around 9% usage (compared to all versions), so unfortunately, it's not dead yet. But you're right, it should be uncommon to use Base UI with it. I'll add a note about supported bundlers instead.

PR #5006, opened 2026-06-10
[docs] Add view source link

aarongarciah:
Adds a "View source" link to each component/util/hook page.

|before|after|
|---|---|
|Screenshot 2026-06-10 at 22 01 03|image|


colmtuite (approved):
Subtitle gap needs to be 16px.
Subtitle links gap needs to be 24px.

aarongarciah:
> Subtitle gap needs to be 16px. Subtitle links gap needs to be 24px.

Updated

PR #5007, opened 2026-06-10
[code-infra] Filter packages to release

michaldudak:
Pass the packages to release to the `pnpm code-infra publish --ci` call.


Janpot:
Isn't it enough to just not update the version of the packages you don't want to release?

PR #5009, opened 2026-06-10
[docs] Inline code tags adjustments

aarongarciah:
Preview:
- Regular inline code tags (blue): https://deploy-preview-5009--base-ui.netlify.app/react/handbook/animation#javascript-animations
- Code highlighted lines and terms: https://deploy-preview-5009--base-ui.netlify.app/react/handbook/forms#react-hook-form-integrate-components

Some adjustments for inline code tags:
- Slightly reduces the font-size on inline code tags.
- Adjusts the blue highlight color, used in inline code tags and code blocks highlighted lines.

### DIFF
--- a/docs/src/components/Code.css
+++ b/docs/src/components/Code.css
@@ -6,7 +6,7 @@
     p &[data-inline],
     li &[data-inline] {
       font-size: 0.96875em; /* 16px becomes 15.5px */
-      letter-spacing: -0.02em;
+      letter-spacing: -0.01em;
       word-spacing: -0.2em;
       line-height: normal;
     }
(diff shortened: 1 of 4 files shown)


docs/src/components/Code.css
```diff
@@ -30,5 +31,9 @@
         }
       }
     }
+
+    .Link & {
+      text-decoration: inherit;
```
aarongarciah, line 37: Block elements don't inherit text-decoration.

colmtuite (approved):
The blue needs to be brighter and more saturated. Something in the 200–208 range.
The `-.05px` trick looks better to me, but it feels horrible. Not sure what to do here.

aarongarciah:
Updated some stuff:
- Blue bg color.
- Reduced font-size.

PR #5010, opened 2026-06-11
[popups] Restore viewport morphing after reopen

atomiks:
A kept-mounted popup viewport could skip its morphing snapshot after closing, reopening through one trigger, then switching back to the trigger it had handled before close. Reset the handled-trigger guard when the popup is closed so each open cycle can animate trigger switches normally.

## Root cause

The last handled trigger was kept in a ref across close/reopen cycles, so a later switch to that same trigger was ignored.

## Changes

- Reset the handled trigger ref when the popup is closed or unmounted.
- Add a kept-mounted viewport regression test for close, reopen, and trigger switch.


atomiks:
## Code Review (GPT-5.5)
Approve ✅ No actionable issues found; the patch resets only the stale trigger guard across closed/unmounted cycles and the regression test reproduces the bug on base.

### 1. Bugs / Issues (None)

The fix targets the stale state directly: `lastHandledTriggerRef` can outlive a close/reopen cycle when the portal is kept mounted, so a later switch back to that trigger is incorrectly ignored. Resetting that ref when `open` or `mounted` goes false preserves the existing duplicate-handling guard while the popup is open.

### Test Coverage Assessment
I verified the added regression test fails on freshly fetched `upstream/master` with only the test applied: after close/reopen/switch, `[data-previous]` stays `null`. On the PR branch, these targeted suites pass:

- `pnpm test:chromium PopoverViewport --no-watch`
- `pnpm test:chromium MenuViewport --no-watch`
- `pnpm test:jsdom PopoverViewport --no-watch`
- `pnpm test:jsdom MenuViewport --no-watch`

### Performance / Operational Risk
| Evidence | Result | Review take |
| --- | --- | --- |
| code-infra bot | +42B parsed, +16B gzip; total duration +41.25ms; renders +0 | Not a blocker for this bug fix; the change adds one ref-reset layout effect and no render-count increase, but worth watching if a rerun repeats the duration/paint signal. |

PR #5016, opened 2026-06-11
[popups] Keep position transition running when the rendered side changes

atomiks:

https://deploy-preview-5016--base-ui.netlify.app/experiments/popover/calendar

Fixes animation bugs when the rendered side changes while a popup moves between detached triggers (one trigger's popup fits below it, the other's flips above): the position teleported instead of transitioning, and the size morph snapped. Same-anchor collision flips remain instant.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/93f55a27-1b1d-4b38-ba68-e6ddc4117f21"></video> | <video src="https://github.com/user-attachments/assets/556adabf-e580-44e4-afb6-9f2c4e386f8b"></video> |

With a `Viewport`, `adaptiveOrigin` positions from the inset property nearest the anchor (`bottom` for side `top`, `right` for side `left`). A rendered side change swaps that property, and CSS can’t transition from `auto`, so the move jumped. The middleware now tracks each floating element’s last rendered side, dimensions, and anchor in a `WeakMap` and animates anchor moves only: an anchor change opens a time-bound window derived from the computed transition duration, and a side change inside the window is re-expressed in the new inset properties — compensating on a swapped axis for the positioner size snap that precedes repositioning — so the transition continues from where the popup is. Outside the window the middleware suppresses the declared transitions with an inline `transition-property: none`, so scroll and resize tracking is rigid and same-anchor collision flips apply instantly. This also fixes positioners with inset transitions lagging behind their anchor while scrolling: previously every reposition animated, since nothing disabled the transitions for mouse-opened popups in the steady state.

The size morph snapped for a separate reason: `usePopupAutoResize` applied the side-dependent anchoring styles inside its measurement effect, so a side change restarted that effect and cancelled the in-flight size transition. They're applied in their own effect now.


aarongarciah (commented):
# PR review

Two merge-blocking issues remain. The deployed A→B animation works smoothly.

## Bugs (1)

### 1. 🔴 Preserve inline transition longhands while disabling a flip

**Location:** `packages/react/src/utils/adaptiveOriginMiddleware.ts:132`

```ts
const inlineTransition = floatingStyle.transition;
floatingStyle.transition = 'none';
// ...
floatingStyle.transition = inlineTransition;
```

When consumers configure transitions using individual inline properties, `floatingStyle.transition` is empty. Assigning the `transition` shorthand clears those longhands, and restoring the empty value leaves them deleted.

I reproduced this through `Popover.Positioner`’s public `style` prop: after a same-anchor flip, `transitionProperty`, `transitionDuration`, and `transitionTimingFunction` were all empty. React also reported conflicting shorthand/longhand mutations.

**Failure scenario:** After a collision flip, subsequent detached-trigger movements lose their transition and teleport.

**Fix:** Temporarily override and restore `transition-duration`, including its original value and priority, instead of writing the `transition` shorthand. Add an inline-longhand regression test.

## Tests (1)

### 1. 🔴 Wrap the frame loop in act

**Location:** `packages/react/src/popover/positioner/PopoverPositioner.test.tsx:635`

```tsx
for (let i = 0; i < 20; i += 1) {
  await new Promise(requestAnimationFrame);
  // ...
}
```

Animation completion updates `PopoverViewport` state during these frames outside `act`. The [[React 18 browser job](https://circleci.com/gh/mui/base-ui/338332)](https://circleci.com/gh/mui/base-ui/338332) deterministically fails because `vitest-fail-on-console` catches the resulting warning. The targeted test also failed after merging the PR into current `master`.

**Failure scenario:** The required React 18 browser check stays red, blocking the PR.

**Fix:** Use the already-imported `waitSingleFrame` and wrap each awaited frame in `act`.

## Verdict

**Request changes** - the middleware deletes valid consumer transition styles, and the new test leaves a required browser check failing.

---

🤖 Review generated with Codex

michaldudak:
# PR review

Two merge-blocking transition regressions remain. The existing 81 focused Chromium tests pass, but two targeted regressions added temporarily against the exact PR head both failed; the temporary worktree was removed afterward.

## Bugs (4)

### 1. 🔴 Preserve consumer transition properties

**Location:** `packages/react/src/utils/adaptiveOriginMiddleware.ts:75`

```ts
const suppressedProperty = moving ? '' : 'none';
floating.style.transitionProperty = suppressedProperty;
```

This destructively replaces a consumer’s inline `transitionProperty`. Steady tracking changes it to `none`; an anchor move changes it to an empty string rather than restoring the authored value. It also suppresses unrelated transitions such as opacity.

A targeted Chromium regression expected `top, bottom, opacity` after scrolling but received `none`.

**Failure scenario:** A positioner using inline inset and opacity transitions loses those settings after tracking, so subsequent movement teleports and appearance transitions stop.

**Fix:** Preserve the authored inline property and suppress only positional insets, restoring the exact value and priority during moves and cleanup.

### 2. 🔴 Keep same-anchor collision flips instant during a move

**Location:** `packages/react/src/utils/adaptiveOriginMiddleware.ts:54`

```ts
if (anchorChanged || (sideChanged && now < movingUntil)) {
  movingUntil = now + getMaxTransitionTime(styles);
}
```

A collision flip during the movement window is treated as another animated move. Scrolling while a horizontal trigger transition is running therefore animates the newly swapped vertical inset across the anchor.

The targeted regression found the popup still 190px from its correct flipped position when `data-side="bottom"` was already applied.

**Failure scenario:** A user switches triggers and immediately scrolls; the popup glides through or remains temporarily detached from its anchor.

**Fix:** Snap the swapped axis when the anchor itself did not change, while allowing any unchanged-axis anchor transition to continue.

### 3. 🟠 Derive the movement window from positional transitions

**Location:** `packages/react/src/utils/adaptiveOriginMiddleware.ts:18`

```ts
return (parseList(styles.transitionDuration) + parseList(styles.transitionDelay)) * 1000;
```

This combines the independently longest duration and delay, even when they belong to different transition entries, and includes unrelated properties. The resulting window can considerably outlive every positional transition.

**Failure scenario:** A long opacity transition keeps the positioner in moving mode after its inset transition finished, causing later scroll and resize updates to animate and lag.

**Fix:** Pair duration and delay lists according to CSS repetition rules, then take the maximum only for transitioned inset properties.

### 4. 🟡 Keep calendar events in the same grid row

**Location:** `docs/src/app/(private)/experiments/popover/calendar.tsx:53`

```tsx
dayOfWeek: 1,
```

The experiment now contains multiple events per column, but events specify only `gridColumn` while the calendar defines one explicit row. Repeated-column events are auto-placed into implicit rows, so their percentage-based time offsets no longer share the calendar’s full height.

**Failure scenario:** The manual experiment’s newly added edge-position events appear at incorrect vertical positions, undermining the scenarios it was expanded to test.

**Fix:** Give every event `gridRow: 1`, or position events absolutely within each day column.

## Tests (1)

### 1. 🟠 Assert that side-swap transitions actually move

**Location:** `packages/react/src/popover/positioner/PopoverPositioner.test.tsx:658`

```tsx
expect(Math.abs(top - previousTop)).toBeLessThan(20);
```

The continuity tests only impose a maximum per-frame delta. A popup frozen at its starting coordinate produces a zero delta and passes as long as `data-side` changes.

**Failure scenario:** A regression completely preventing movement remains green.

**Fix:** Also require measurable progress and eventually assert the final anchor-relative position.

## Simplifications (1)

### 1. 🟡 Remove redundant resize state and memoization

**Location:** `packages/react/src/utils/usePopupAutoResize.ts:31`

```ts
const committedDimensionsRef = React.useRef<Dimensions | null>(null);
const isInitialRenderRef = React.useRef(true);
```

`committedDimensionsRef.current === null` already represents the initial state, making `isInitialRenderRef` redundant. After splitting the anchoring effect, its `useMemo` can likewise be replaced by computing the styles inside that effect from `side` and `direction`.

**Failure scenario:** The shared popup path retains avoidable hooks and bundle code without distinguishing additional state.

**Fix:** Use the nullable dimensions ref as the sole sentinel and calculate anchoring styles inside the layout effect.

## Verdict

**Request changes** - consumer transition styles are overwritten, and collision flips can animate across the anchor during an active move.

---

🤖 Review generated with Codex

PR #5021, opened 2026-06-11
[docs-infra] Ban :has selector

aarongarciah:
Bans `:has()` in the docs (except demos) using the [`selector-pseudo-class-disallowed-list`](https://stylelint.io/user-guide/rules/selector-pseudo-class-disallowed-list/) Stylelint rule. `:has()` can still be used, but requires a `stylelint-disable` comment.


stylelint.config.mjs
```diff
@@ -29,12 +29,21 @@ export default {
         'declaration-property-value-no-unknown': null,
         // Remove when ready
         'keyframes-name-pattern': null,
+        // `:has()` can be heavy on perf, be cautious
+        'selector-pseudo-class-disallowed-list': ['has'],
```
atomiks, line 33: It can't be limited to `body`? I think it's usually fine elsewhere, though I get that on big roots, it could still be an issue.

aarongarciah: Yeah, we could use [selector-disallowed-list](https://stylelint.io/user-guide/rules/selector-disallowed-list/) to limit it to `body` (using a regex).

Note that this change doesn't affect demos.

colmtuite: Why limit it to body?

aarongarciah: I think disallowing it everywhere is fine. If the use case is not risky, you just add a `stylelint-disable` comment and move on.

colmtuite (approved):

PR #5025, opened 2026-06-12
[select] Skip disabled items in typeahead and fix multiple-mode serialization

atomiks:
## Summary

Fixes a few Select correctness issues plus adjacent cleanup, with regression tests.

### Bugs
- **Typeahead now skips disabled items.** Disabled items register their value/label unconditionally and `useTypeahead` only skipped invisible elements, so typeahead could stop on a disabled item (and, on a closed trigger, commit it). It now skips disabled items *while matching*, so a single keypress advances to the next selectable item — matching native `<select>` and arrow-key navigation — in both open and closed states. This is done via a new opt-in `disabledIndices` option on `useTypeahead` (the same array-or-predicate shape as `useListNavigation`), decoupled from the `elementsRef`/visibility filter so Select's mounted-but-hidden closed items still match. Menu and Combobox don't pass it, so their behavior is unchanged.
- **Typeahead matching is now locale-neutral.** `useTypeahead` no longer relies on locale-sensitive lowercasing, so Turkish/Azeri casing rules don't prevent typing `i` from matching labels like `Istanbul`.
- **Multiple mode no longer passes the whole value array to `itemToStringValue`.** The shared hidden input is nameless in multiple mode (per-value `<input>`s carry the data), so `serializedValue` is now `''`; previously a user `itemToStringValue` written for a single item would throw or yield `''`.
- `positionerElement.style.maxHeight = 'auto'` (invalid, silently ignored) → `'none'`, so the explicit height governs in align mode and isn't clamped by user CSS.
- Trigger callback refs no longer fire twice (they were attached via both a merged-ref prop and the `useRenderElement` ref array).
- A custom `isItemEqualToValue` is no longer handed the raw empty array as the "selected value" in multiple mode (`undefined` when nothing is selected).

### Cleanup
- Remove the write-only `keyboardActiveRef` and the unconsumed `events` context field.
- Use `ownerWindow(el).getComputedStyle(el)` in `SelectPopup` (repo convention for realm-sensitive lookups).
- Drop a stray `useStore` argument in `SelectArrow` and unused scroll-arrow refs.
- Comment fixes (backdrop description + generated docs, `parseFloat` fallback, outside-press); remove a shipped `XXX` note.

### Tests
- Typeahead skips a disabled item and commits the next match.
- Commits nothing when the only typeahead match is disabled.
- Locale-neutral typeahead matching still matches ASCII `I...` labels when Turkish/Azeri casing would differ.
- Multiple-mode `itemToStringValue` isn't invoked with the array.

Typecheck, eslint, prettier, and the Select / Menu / Combobox / useTypeahead suites pass. Based on current `master`.


michaldudak:
## Codex Review (GPT-5.5)

Reviewed the full PR 5025 branch against freshly fetched `upstream/master`. This is a focused `bug-fix` PR with secondary test/docs cleanup, and most of the patch looks sound, but one disabled-typeahead edge case remains.

### 1. Bugs / Issues

#### 1. 🔴 Disabled double-letter items still block rapid same-letter cycling (blocking)

`packages/react/src/floating-ui-react/hooks/useTypeahead.ts`

```ts
const allowRapidSuccessionOfFirstLetter = listContent.every((text) =>
  text ? text[0]?.toLocaleLowerCase() !== text[1]?.toLocaleLowerCase() : true,
);
```

The new matching path skips disabled items, but this rapid-succession guard still evaluates every label in `listContent`, including disabled items. That means a disabled option like `Aaron` can prevent cycling through enabled `A...` options: disabled `Aaron`, enabled `Apple`, enabled `Avocado`; pressing `a` selects/highlights `Apple`, but pressing `a` again builds `"aa"` instead of cycling to `Avocado`, then clears because the only `"aa"` match is disabled.

**Fix:** Make the rapid-succession guard use the same eligibility rules as matching, at least ignoring `isIndexDisabled?.(index)` and likely hidden items from `elementsRef`, then add a regression test for disabled double-letter labels not blocking cycling among enabled matches.

The multiple-mode serialization fix is well targeted: the nameless shared input no longer serializes the full array, while per-value hidden inputs still carry form data.

The disabled-item typeahead fix addresses the single-key matching path, but the rapid repeated-letter path still has a separate eligibility decision. That leaves the branch short of fully aligning disabled item handling across typeahead behavior.

### Test Coverage Assessment

I ran:

```text
pnpm test:jsdom SelectRoot --no-watch
pnpm test:jsdom useTypeahead --no-watch
```

Both passed. The added Select tests cover single-key disabled skipping and the multiple-mode serialization regression, but they do not cover the disabled double-letter cycling case above. I did not verify the new regression tests against `upstream/master`.

### Recommendation

Request changes 🔴

The PR is close, but disabled items can still alter enabled typeahead behavior through the repeated-letter guard. I’d fix that guard and add the focused regression before merging.

flaviendelangle (commented):
# PR Review Summary

Five specialized agents reviewed the PR in an isolated worktree against the merge base. Headline: **the PR is correct and merge-ready.** No critical bugs, no new silent failures, strong project-guideline compliance. Findings are about polish — one naming/convention question, one misleading JSDoc phrase, and one untested fix.

## Critical Issues (0)
None. All seven flagged risk areas (keyboardActiveRef removal, `events` removal, ref-dedup, `serializedValue=''`, outside-press simplification, `useStore` arg drop, `maxHeight`) were independently verified safe.

## Important Issues (2)

**1. `isIndexDisabled` naming collides with the established convention** — `useTypeahead.ts:41` *(type-design-analyzer, Invariant Expression 5/10)*
The codebase already expresses "disabled index" as `disabledIndices?: DisabledIndices` (`useListNavigation.ts:149`, type at `composite.ts:8`) with a shared resolver `isListIndexDisabled` (`composite.ts:472`). The new field `isIndexDisabled` sits one word from `isListIndexDisabled` but has a *different shape* (bare predicate vs. array-or-predicate union). Two sibling hooks in the same directory now name the same concept differently — a future maintainer could wire the wrong one. Recommend renaming to `disabledIndices` (ideally adopting the shared union, which makes the array form and attribute fallback free). The JSDoc's three-line "this isn't `isListIndexDisabled`" disclaimer is itself a signal the name fights the convention. *(Judgment call — reasonable to defer, but cheap to fix now.)*

**2. The `isItemEqualToValue` → `undefined` fix has zero test coverage** — `SelectItem.tsx:100-110` *(pr-test-analyzer, Gap A, 8/10)*
This is one of the two fixes named in the PR title, yet no test exercises it. The pre-fix bug: `multiple` mode + empty selection (`[]`) + a custom `isItemEqualToValue` written for single items → the raw `[]` was passed to the comparer → throws on `[].value`. Existing `isItemEqualToValue` tests all use *non-empty* values and miss this branch. Add: render `Select.Root multiple` with `defaultValue={[]}` (or unset) and `isItemEqualToValue={(a, b) => a.value === b.value}` over object items; assert it renders without throwing. Fails pre-fix, passes post-fix.

## Suggestions (5)

- **Misleading JSDoc wording** — `useTypeahead.ts:34-40` *(flagged independently by comment-analyzer AND type-design-analyzer)*. Says the predicate is "independent of `elementsRef`/visibility," but the code AND-combines them: `isItemAvailable = isVisible(index) && !isIndexDisabled?.(index)`. A disabled-but-also-hidden item is still rejected by visibility. Reword to "the *disabled check* doesn't read `elementsRef`, so consumers that don't pass `elementsRef` still get disabled-skipping."
- **Untested open-state skip path** — `SelectRoot.tsx:402-407` *(Gap B, 6/10)*. All three new typeahead tests use the closed-trigger (`setValue`) branch; the open branch (`store.set('activeIndex')`) is unexercised. A `describe.skipIf(isJSDOM)` test opening the popup and asserting the disabled item gets no `data-highlighted` would close it.
- **Pre-existing latent quirk, now documented not fixed** — `SelectPopup.tsx:316-318` *(silent-failure-hunter)*. `parseFloat(x) || 10` treats an explicit `margin:0`/`min-height:0` as `10`/`100`. The PR only *adds a comment* about it (no regression), but it's a candidate for a follow-up using `Number.isNaN(parsed) ? N : parsed`.
- **Comment over-generalizes** — `SelectPopup.tsx:314-315`. "the align algorithm always reserves some spacing" is true for the two margins but the same `|| N` also guards `minHeight` (not spacing). Narrow the clause.
- **Minor perf note** — `useTypeahead.ts:164`. `allowRapidSuccessionOfFirstLetter` now calls `isItemAvailable` (which runs `isElementVisible` + the predicate) for every item on every keypress. Not a correctness issue; worth a glance if it shows in the combobox bench.

## Strengths
- **No new bugs and no new silent failures** across all risk areas; the `serializedValue=''` change correctly avoids breaking form submission (shared input is nameless in multiple mode; per-value inputs carry data) and `isElementDisabled(null)` returns `true` (skips gracefully, never throws).
- **Cleanups verified complete**: `keyboardActiveRef` was genuinely write-only in Select (remaining reads belong to Combobox's own store); `events` had no consumer; the ref refactor still applies all four refs via `useRenderElement`; import + memo body + dep array all pruned together.
- **4 of 6 new tests are true regression guards** (verified by tracing pre-fix behavior), including a negative case and the subtle disabled double-letter cycling interaction.
- **Several exemplary "why" comments** (multiple-mode serialization, the `isIndexDisabled` rationale in `SelectRoot`, `maxHeight: 'none'` documenting the invalid-`auto` trap) and **good guideline compliance** (migrates two `getComputedStyle` globals to `ownerWindow`, shares logic via context).
- `SelectRootContext` slimming is a clean encapsulation win (9/10) — drops a leaked `FloatingEvents` type and cross-component mutable ref.

## Recommended Action
1. **Add the Gap A test** (empty multiple + custom `isItemEqualToValue`) — the one named fix with no coverage. *(highest value)*
2. **Reword the `isIndexDisabled` JSDoc** at `useTypeahead.ts:34-40` — quick, two reviewers flagged it.
3. **Decide on the `isIndexDisabled` → `disabledIndices` rename** — align with `useListNavigation`, or consciously accept the divergence.
4. Optional polish: open-state test (Gap B), narrow the `SelectPopup:314` comment, file a follow-up for the `parseFloat || N` quirk.

flaviendelangle (approved):
# PR Review Summary — Round 2 (PR #5025)

The author pushed one follow-up commit, `23a59f264` "Address review", which addresses **all four** findings from the previous round. I re-reviewed only the delta. Net: the source is solid and the rename is a clean win — but **one of the two new tests is a vacuous pass** and must be fixed before it provides any protection.

## How the prior findings were addressed
| Prior finding | Status |
|---|---|
| ① `isIndexDisabled` naming collision | ✅ Renamed to `disabledIndices: DisabledIndices`, routed through the shared `isListIndexDisabled` resolver — matches `useListNavigation` exactly |
| ② Misleading "independent of visibility" JSDoc | ✅ Reworded to "the disabled check doesn't read `elementsRef`" |
| ③ Gap A — `isItemEqualToValue` empty-array fix untested | ⚠️ Test added but **vacuous** (see below) |
| ④ Gap B — open-state skip path untested | ✅ Valid, non-flaky guard added |
| (suggestion) `\|\| N` comment over-generalized | ✅ Narrowed |

## Important Issues (1 found)

- **[pr-test-analyzer] The new Gap A test is a vacuous pass — it cannot fail even if the bug is reintroduced.** `SelectRoot.test.tsx:1073` ("does not invoke isItemEqualToValue with the value array in multiple mode when empty"). The test renders a **closed** `Select.Root multiple defaultValue={[]}` and never opens/focuses it. The comparison it targets lives in a `useIsoLayoutEffect` gated on `hasRegistered` (`SelectItem.tsx:92-93`), so it only runs once items **mount** — which a closed, unfocused Select never does. The comparer is never called, so both the `Array.isArray` throw and `not.toHaveBeenCalledWith(...)` pass vacuously. The analyst empirically confirmed: reverting the `SelectItem.tsx` fix leaves the test green; adding `defaultOpen` makes it correctly fail pre-fix and pass post-fix. I independently confirmed the mounting gate by reading the source.
  - **Fix:** add `defaultOpen` (or `forceMount` the items) to the test's `Select.Root` so the `SelectItem` effect actually runs. The comparison path needs no layout, so it should remain a jsdom test (no `skipIf(isJSDOM)` needed). The underlying *source* fix (`SelectItem.tsx:100-105`) is correct — only the test is inadequate.

## Suggestions (2 — optional, below merge-blocking threshold)

- **[code-reviewer] JSDoc wording polish** — `useTypeahead.ts:36-42`. "...without opting into the visibility filter" is slightly imprecise since `isItemAvailable` always AND-combines visibility regardless of `disabledIndices`. Meaning is clear from context; pure wording.
- **[type-design-analyzer] `isListIndexDisabled(EMPTY_ARRAY, …)` is a mild API smell** — `useTypeahead.ts`. Passing a sentinel empty list to defeat the resolver's element-based fallbacks is correct and well-commented, but the signature still advertises a list it doesn't use. A future cleanup could extract `isIndexExplicitlyDisabled(index, disabledIndices)` from `composite.ts:477-480` and call that directly. Not worth blocking.

## Strengths
- **The `isIndexDisabled` → `disabledIndices` rename is behavior-preserving and verified.** For Select's predicate, `!isListIndexDisabled(EMPTY_ARRAY, i, fn)` reduces exactly to `!fn(i)`; for `disabledIndices == null` it short-circuits to visibility-only — Menu/Combobox confirmed unchanged. It also correctly broadens to the array-or-predicate union (the old `?.()` call would have thrown on an array).
- **Type design measurably improved:** Invariant Expression **5/10 → 8/10**, Enforcement **7 → 8** (deduplicated onto the shared resolver). The "rename before merge" recommendation is fully satisfied.
- **Gap B test (open-popup skip) is a genuine, non-flaky guard** — correctly gated with `it.skipIf(isJSDOM)`, one assertion per `waitFor()`, and verified to fail pre-fix. The wrong-reason/start-index flakiness concern was traced and ruled out.
- No new bugs, no silent failures, no guideline violations in the delta.

## Recommended Action
1. **Add `defaultOpen` (or `forceMount`) to the Gap A test at `SelectRoot.test.tsx:1092`** so it actually exercises the `SelectItem` comparison — currently it guards nothing.
2. Optionally apply the two wording/idiom polish suggestions.
3. Everything else is merge-ready.

One actionable item this round. The rest of the follow-up is clean.

PR #5026, opened 2026-06-12
[menu] Fix detached-trigger store sync and context-menu navigation

atomiks:
## Summary

Correctness fixes around the menu store lifecycle, detached triggers, and context menus, plus cleanup.

### Bugs
- **`MenuStore` parent observer no longer detaches itself / leaks its subscription.** `ReactStore.observe` fires its listener synchronously before returning, so the single unsubscriber field was clobbered at construction (leaking the per-parent store subscription) and a later `parent` change tore down the observer itself. The observer's own unsubscriber and the per-parent subscription now live in separate fields, and the root-level branch restores this menu's own `allowMouseUpTriggerRef` rather than keeping a borrowed parent ref.
- **Hover-open `mouseup` listener is now cleaned up.** A hover-open that closed again without a `mouseup` left a `{ once: true }` document listener armed to fire on an unrelated later interaction.
- **Touch-to-close guard runs before `onOpenChange` / `dispatchOpenChange`.** Previously the guard bailed *after* already notifying controlled consumers (and floating-ui's own state), so a controlled menu closed anyway.
- **ArrowLeft no longer closes a root context menu.** A root context menu was treated as `nested`, so the cross-orientation close key closed it; native context menus don't (there's no parent list to return to). Its submenus remain regular `parent.type === 'menu'` menus.

### Cleanup
- Remove the never-constructed `nested-context-menu` `MenuParent` variant.
- Drop the no-op render-time `delete rootTriggerProps.id`.
- Use the deduping `warn` util for the modal-on-child-menu warning (was `console.warn` on every render) with corrected wording; use `REASONS.triggerHover` instead of the string literal; fix the `stickIfOpen` comment.

The duplicated trigger drag-release bounds check is kept inline for now (a shared helper is a planned follow-up).

Typecheck, eslint, prettier, and the Menu / Menubar / ContextMenu suites pass. Based on current `master`.

### DIFF
--- a/docs/src/app/(docs)/react/components/menu/types.md
+++ b/docs/src/app/(docs)/react/components/menu/types.md
@@ -1078,7 +1078,6 @@ type MenuParent =
   | { type: 'menu'; store: MenuStore<unknown> }
   | { type: 'menubar'; context: MenubarContext }
   | { type: 'context-menu'; context: ContextMenuRootContext }
-  | { type: 'nested-context-menu'; context: ContextMenuRootContext; menuContext: MenuRootContext }
   | { type: undefined };
 ```
 
--- a/packages/react/src/context-menu/root/ContextMenuRoot.test.tsx
+++ b/packages/react/src/context-menu/root/ContextMenuRoot.test.tsx
@@ -1,5 +1,6 @@
 import { vi, expect } from 'vitest';
 import {
+  act,
   fireEvent,
   flushMicrotasks,
   ignoreActWarnings,
@@ -230,6 +231,78 @@ describe('<ContextMenu.Root />', () => {
 
       expect(screen.queryByTestId('context-popup')).toBe(null);
       expect(onOpenChange.mock.calls.length).toBe(0);
+    });
+  });
+
+  describe.skipIf(isJSDOM)('keyboard navigation', () => {
+    it('does not close the root context menu on the cross-orientation close key, but closes its submenus', async () => {
+      const rootOnOpenChange = vi.fn();
+
+      const { user } = await render(
+        <ContextMenu.Root onOpenChange={rootOnOpenChange}>
+          <ContextMenu.Trigger data-testid="context-trigger">Surface</ContextMenu.Trigger>
+          <ContextMenu.Portal>
+            <ContextMenu.Positioner>
+              <ContextMenu.Popup data-testid="context-root-popup">
+                <ContextMenu.SubmenuRoot>
+                  <ContextMenu.SubmenuTrigger data-testid="context-submenu-trigger">
+                    More options
+                  </ContextMenu.SubmenuTrigger>
+                  <ContextMenu.Portal>
+                    <ContextMenu.Positioner>
+                      <ContextMenu.Popup data-testid="context-submenu-popup">
+                        <ContextMenu.Item data-testid="context-submenu-item">
+                          Deep action
+                        </ContextMenu.Item>
+                      </ContextMenu.Popup>
+                    </ContextMenu.Positioner>
+                  </ContextMenu.Portal>
+                </ContextMenu.SubmenuRoot>
+              </ContextMenu.Popup>
+            </ContextMenu.Positioner>
+          </ContextMenu.Portal>
+        </ContextMenu.Root>,
+      );
+
+      const trigger = screen.getByTestId('context-trigger');
+      fireEvent.contextMenu(trigger, { clientX: 10, clientY: 10, button: 2 });
+      await flushMicrotasks();
+
+      await screen.findByTestId('context-root-popup');
+
+      const submenuTrigger = screen.getByTestId('context-submenu-trigger');
+      await act(async () => {
+        submenuTrigger.focus();
+      });
+      await waitFor(() => {
+        expect(submenuTrigger).toHaveFocus();
+      });
+
+      // ArrowRight opens the (nested) submenu and moves focus to its first item.
+      await user.keyboard('[ArrowRight]');
+      await screen.findByTestId('context-submenu-popup');
+      const submenuItem = screen.getByTestId('context-submenu-item');
+      await waitFor(() => {
+        expect(submenuItem).toHaveFocus();
+      });
+
+      // ArrowLeft closes the submenu because it is nested, returning focus to its trigger.
+      await user.keyboard('[ArrowLeft]');
+      await waitFor(() => {
+        expect(screen.queryByTestId('context-submenu-popup')).toBe(null);
+      });
+      expect(screen.queryByTestId('context-root-popup')).not.toBe(null);
+      await waitFor(() => {
+        expect(submenuTrigger).toHaveFocus();
+      });
+
+      // ArrowLeft on the root context menu must NOT close it: unlike a submenu, a root context
+      // menu has no parent list to return to, and native context menus never close on this key.
+      await user.keyboard('[ArrowLeft]');
+      await flushMicrotasks();
+
+      expect(screen.queryByTestId('context-root-popup')).not.toBe(null);
+      expect(rootOnOpenChange.mock.calls.some((call) => call[0] === false)).toBe(false);
     });
   });
 
--- a/packages/react/src/menu/positioner/MenuPositioner.tsx
+++ b/packages/react/src/menu/positioner/MenuPositioner.tsx
@@ -298,11 +298,7 @@ export const MenuPositioner = React.forwardRef(function MenuPositioner(
     <MenuPositionerContext.Provider value={positioner}>
       {shouldRenderBackdrop && (
         <InternalBackdrop
-          ref={
-            parent.type === 'context-menu' || parent.type === 'nested-context-menu'
-              ? parent.context.internalBackdropRef
-              : null
-          }
+          ref={parent.type === 'context-menu' ? parent.context.internalBackdropRef : null}
           inert={inertValue(!open)}
           cutout={backdropCutout}
         />
--- a/packages/react/src/menu/root/MenuRoot.tsx
+++ b/packages/react/src/menu/root/MenuRoot.tsx
@@ -6,6 +6,7 @@ import { useId } from '@base-ui/utils/useId';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
 import { EMPTY_ARRAY, EMPTY_OBJECT } from '@base-ui/utils/empty';
 import { fastComponent } from '@base-ui/utils/fastHooks';
+import { warn } from '@base-ui/utils/warn';
 import {
   FloatingTree,
   useDismiss,
@@ -147,8 +148,10 @@ export const MenuRoot = fastComponent(function MenuRoot<Payload>(props: MenuRoot
 
   if (process.env.NODE_ENV !== 'production') {
     if (parent.type !== undefined && modalProp !== undefined) {
-      console.warn(
-        'Base UI: The `modal` prop is not supported on nested menus. It will be ignored.',
+      // `warn` dedupes, so this won't spam on every render. `parent.type !== undefined` also
+      // covers menubar and context menus, not just submenus.
+      warn(
+        'The `modal` prop is not supported on submenus, menubar menus, or context menus. It will be ignored.',
       );
     }
   }
@@ -238,6 +241,22 @@ export const MenuRoot = fastComponent(function MenuRoot<Payload>(props: MenuRoot
         return;
       }
 
+      const nativeEvent = eventDetails.event as Event;
+
+      // Prevent the menu from closing on mobile devices that have a delayed click event.
+      // In some cases the menu, when tapped, will fire the focus event first and then the click
+      // event. Without this guard, the menu will close immediately after opening.
+      // This must bail before notifying `onOpenChange` and dispatching the open change, otherwise
+      // controlled consumers (and floating-ui's own state) would close the menu regardless.
+      if (
+        nextOpen === false &&
+        nativeEvent?.type === 'click' &&
+        (nativeEvent as PointerEvent).pointerType === 'touch' &&
+        !allowTouchToCloseRef.current
+      ) {
+        return;
+      }
+
       const shouldPreventUnmountOnClose = attachPreventUnmountOnClose(
         eventDetails as MenuRoot.ChangeEventDetails,
       );
@@ -256,19 +275,8 @@ export const MenuRoot = fastComponent(function MenuRoot<Payload>(props: MenuRoot
 
       store.state.floatingRootContext.dispatchOpenChange(nextOpen, eventDetails);
 
-      const nativeEvent = eventDetails.event as Event;
-      if (
-        nextOpen === false &&
-        nativeEvent?.type === 'click' &&
-        (nativeEvent as PointerEvent).pointerType === 'touch' &&
-        !allowTouchToCloseRef.current
-      ) {
-        return;
-      }
-
-      // Prevent the menu from closing on mobile devices that have a delayed click event.
-      // In some cases the menu, when tapped, will fire the focus event first and then the click event.
-      // Without this guard, the menu will close immediately after opening.
+      // Reset the touch-to-close guard so a delayed click after a focus-driven open is ignored
+      // (see the early return above).
       if (nextOpen && reason === REASONS.triggerFocus) {
         allowTouchToCloseRef.current = false;
         allowTouchToCloseTimeout.start(300, () => {
@@ -394,7 +402,10 @@ export const MenuRoot = fastComponent(function MenuRoot<Payload>(props: MenuRoot
     enabled: !disabled,
     listRef: store.context.itemDomElements,
     activeIndex,
-    nested: parent.type !== undefined,
+    // A root context menu has no parent menu to return to, so it must not be treated as nested:
+    // otherwise the cross-orientation close key (e.g. ArrowLeft) would close it, which native
+    // context menus never do. Its submenus are regular menus with `parent.type === 'menu'`.
+    nested: parent.type !== undefined && parent.type !== 'context-menu',
     loopFocus,
     orientation,
     parentOrientation: parent.type === 'menubar' ? parent.context.orientation : undefined,
@@ -665,11 +676,6 @@ export type MenuParent =
   | {
       type: 'context-menu';
       context: ContextMenuRootContext;
-    }
-  | {
-      type: 'nested-context-menu';
-      context: ContextMenuRootContext;
-      menuContext: MenuRootContext;
     }
   | {
       type: undefined;
--- a/packages/react/src/menu/root/MenuRoot.tsx
+++ b/packages/react/src/menu/root/MenuRoot.tsx
@@ -275,8 +275,9 @@ export const MenuRoot = fastComponent(function MenuRoot<Payload>(props: MenuRoot
 
       store.state.floatingRootContext.dispatchOpenChange(nextOpen, eventDetails);
 
-      // Reset the touch-to-close guard so a delayed click after a focus-driven open is ignored
-      // (see the early return above).
+      // Arm the touch-to-close guard for 300ms after a focus-driven open, so the delayed click
+      // that follows a tap is ignored (see the early return above). Any other open/close clears
+      // the guard so a normal click can close the menu right away.
       if (nextOpen && reason === REASONS.triggerFocus) {
         allowTouchToCloseRef.current = false;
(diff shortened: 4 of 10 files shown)


flaviendelangle (commented):
# PR #5026 Review Summary — `[menu] Fix detached-trigger store sync and context-menu navigation`

**Verdict: Approve with one recommendation (add regression tests).** Six specialized agents reviewed the diff against the actual source. The four bug fixes are each verified correct, the cleanups are safe, and the affected suites pass (code-reviewer ran `MenuRoot` 201✓, `detached-triggers` 21✓, `ContextMenu` 33✓, `MenuTrigger` 106✓, `MenuSubmenuTrigger` 22✓; lint clean). The one TS error seen (`Demo.tsx`) is pre-existing and unrelated.

## Critical Issues (0)
None.

## Important Issues (1)

- **[tests] The PR fixes four distinct correctness bugs with zero tests added.** Three of the four are essentially uncovered today, so a future refactor can silently reintroduce them. Prioritized gaps:
  - **ArrowLeft must NOT close a root context menu (sev 9)** — no `ArrowLeft`/`ArrowRight` tests exist in any context-menu test file. The fix is a one-line predicate (`MenuRoot.tsx:408`). Add to `context-menu/root/ContextMenuRoot.test.tsx` (likely `it.skipIf(isJSDOM)`): root context menu stays open on ArrowLeft, but a submenu inside it *does* close on ArrowLeft (proves root is non-nested while children remain nested).
  - **Controlled menu must not close on a delayed touch-click (sev 8)** — `MenuRoot.test.tsx`; controlled `open`+`onOpenChange`, focus-then-touch-click, assert `onOpenChange(false)` is not called. Hard in jsdom (needs fake timers for the 300ms gate); a Chromium variant is higher-confidence.
  - **Hover-open → hover-out leaves no armed mouseup listener (sev 7)** — `MenuTrigger.test.tsx`; jsdom-friendly. Spy on `document` add/remove `'mouseup'` for balance, or fire an unrelated later `mouseup` and assert no side effect.
  - **Detached-store subscription teardown (sev 6)** — a store-level unit test (jsdom): set `parent` to a `{type:'menu'}`, change it, assert the prior parent-store subscription was torn down; mount/unmount a submenu and assert the parent store's listener count returns to baseline.

## Suggestions (4, all low / optional)

- **[silent-failure / types] The `observe('parent', …)` return value is discarded (`MenuStore.ts:141`), so the observer subscription is never explicitly torn down.** This is **safe today** (it's a self-subscription, GC'd with the store — confirmed by 3 agents). Suggestion only: capture and dispose it in `disposeEffect` alongside `unsubscribeParentStore` to prevent a *future* silent leak if the subscription target ever changes. Optionally add an `assertNever(parent)` default in one selector switch so a future 5th `MenuParent` variant fails loudly.
- **[comments] Two MenuStore.ts comments are accurate but slightly dense/overstated.** The `observe` comment (`:137-140`) describes a now-removed historical bug; consider stating the invariant positively. The `disposeEffect` comment's "needs no explicit cleanup" (`:209-213`) sits awkwardly next to an explicit-cleanup method — clarify it refers to the observer's *self*-subscription, not the parent subscription.
- **[comments] MenuRoot.tsx:278-279** — the "Reset the touch-to-close guard" comment sits above an `if/else` but only describes the `if`. Minor wording nit; rewording risks less accuracy, so optional.
- **[warn dedupe] The modal warning is now deduped globally-by-message** (`MenuRoot.tsx:153`) — a second offending menu with an identical message won't re-warn. This is the standard React tradeoff and a strict improvement over per-render `console.warn`; flagged for awareness only.

## Resolved disagreement
The type agent flagged `disposeEffect` as a confusing name; the simplifier **rebutted it with evidence** — it's an established repo convention (`useTimeout.ts`, `PopoverStore.ts` use the identical "method returning a cleanup closure" pattern, and `MenuStore` now mirrors `PopoverStore` exactly). **Do not rename.**

## Strengths
- **Genuine bug fixes, each verified against source**, not just the PR description: the synchronous-`observe` field-clobber (validated against `ReactStore.ts:249`), the touch-guard state-desync (guard moved before `onOpenChange`/`dispatchOpenChange`), the leaked `{once:true}` hover mouseup listener, and the nested/ArrowLeft predicate (validated against the vendored `useListNavigation`).
- **Safe cleanups**: removing `delete rootTriggerProps.id` eliminates a shared-state mutation (the prop never had an `id`); dropping `mouseUpTarget === triggerRef.current` is verified-redundant with `contains()`'s native self-match; `'trigger-hover'` → `REASONS.triggerHover` removes a magic string; the `nested-context-menu` variant was provably unconstructable dead code (carrying an unread `menuContext`), so removal tightens the union to "declared states = reachable states."
- Excellent, accurate explanatory comments throughout (comment agent found **zero inaccuracies**).

## Recommended action
1. **Add the four regression tests** (start with sev 9 ArrowLeft and sev 7 hover-cleanup, which are the cheapest/most jsdom-friendly). This is the only thing I'd ask for before merge.
2. Optionally apply the low-effort comment-wording tweaks.
3. The `observe`-unsubscriber capture and `assertNever` are nice-to-haves; fine as a follow-up.

flaviendelangle (commented):
# PR #5026 Review Summary — `[menu] Fix detached-trigger store sync and context-menu navigation`

Reviewed at PR head `8e312ed56` against merge-base `3c5e55f97`, using five specialized agents (code, tests, silent-failures, comments, types) plus my own empirical verification of the headline finding.

The four underlying fixes are each **correct in their primary intent**, the regression tests genuinely fail pre-fix, and the cleanup is sound. However, the new unmount-disposal logic — the one part meant to *fix* a leak — introduces a confirmed reactivity regression under React StrictMode.

## 🔴 Critical (must address before merge)

**1. `MenuStore` tears down its parent subscription under StrictMode and never re-establishes it** — `packages/react/src/menu/store/MenuStore.ts:141`, `:199-202`, `:216-223`

I **empirically confirmed** this: a submenu store mounted under `<React.StrictMode>` (React's dev default, and the repo's own `createRenderer`) ends up with its subscription into the parent store **gone** — parent listener count was `1` (baseline) under StrictMode vs `2` without it.

Root cause: the `observe('parent')` wiring (and the per-parent `parent.store.subscribe(...)`) is set up **only in the constructor**, which runs once via `useRefWithInit`. The new `useEffect` cleanup (`disposeEffect`) tears both down on every effect-cleanup. StrictMode's mount → simulated-unmount → remount runs the cleanup but not the constructor, so the subscriptions are permanently severed while the component is still mounted. The same hazard exists in production if the `handle` prop toggles defined→undefined (the `[externalStore, internalStore]` dep makes the cleanup run mid-life).

Impact is **latent but real**: `.select()` reads the parent live so there's no crash and existing suites pass, but the *notify/re-render* path is dead — a submenu won't re-render when the parent's `rootId` / `floatingTreeRoot` / `keyboardEventRelay` change, and won't re-wire if its `parent` reference changes. This is the exact opposite of the leak fix's intent.

Note: a `[internalStore]`-only dep array (the Popover pattern the silent-failure agent cited) would **not** fix this — StrictMode re-runs effects on remount regardless of deps. Popover is safe only because it disposes a *re-creatable* timeout, not a constructor-only subscription. The real fix is to make the parent-subscription lifecycle **symmetric and re-establishable** — e.g. set up `observe('parent')` inside the effect (returning its teardown as the cleanup) rather than in the constructor — so StrictMode's remount re-arms it. Then add a regression test using `createRenderer` (StrictMode) and a `handle` defined→undefined→defined toggle.

## 🟠 Important

**2. The new `MenuStore.test.tsx` bypasses StrictMode, which is exactly why the bug above slips through** — `packages/react/src/menu/store/MenuStore.test.tsx:3`. It imports `render` from `@testing-library/react` instead of the repo's `createRenderer` (StrictMode on). Switching follows convention *and* would surface finding #1. The `handle`-toggle path is also untested.

**3. Touch-to-close guard may swallow a legitimate item-press close within the 300ms window** — `packages/react/src/menu/root/MenuRoot.tsx:251-258` (narrow / not empirically confirmed). The guard fires for *any* `nextOpen===false` carrying a touch `click` while armed. A real menu **item tap** (reason `itemPress`, native touch PointerEvent) landing <300ms after a focus-driven menubar open would early-return and silently drop the close (`onOpenChange(false)` never fires). The guard's *scope* predates this PR, but the reorder changes the resulting behavior from "inconsistent partial close" to "fully stays open." Consider excluding `itemPress` (or popup-targeted) closes from the guard, and add a test for "tap an item <300ms after a focus-driven menubar open closes the menu."

## 🟡 Suggestions

- **Untested branch**: the `allowMouseUpTriggerRef` "stop borrowing parent ref at root" branch (`MenuStore.ts:178-179`) has no coverage — a cheap unit assertion (attach parent → detach → assert own ref restored) would guard it.
- **Untested behavior**: the modal-warning dedup + broadened wording (`MenuRoot.tsx:151-156`) is unverified.
- **Misleading comment**: the `disposeEffect` comment (`MenuStore.ts:211-215`) frames the observer teardown as "collected anyway… merely symmetric and future-proof." The GC claim is technically accurate, but the framing presents the teardown as a harmless no-op when it is in fact the **cause** of finding #1. Update it once the lifecycle is fixed.
- **Comment nit**: `MenuRoot.tsx:278-280` "Any other open/close *clears* the guard" — the else-branch *disarms* it (sets the ref to `true`) and clears the timer; "disarm" is more precise than "clear."
- **Naming**: two different predicates are both called `nested` in `MenuRoot.tsx` (`:147` floating-tree-based vs `:409` the list-nav flag). Give the line-409 predicate an intent-named local (e.g. `treatAsNestedForListNav`). Do **not** add a `nested` field to `MenuParent` — deriving from the discriminant keeps a single source of truth.
- **Build hygiene**: a stale build artifact (`packages/react/build/esm/menu/root/MenuRoot.d.ts:124`) still references the removed `nested-context-menu` variant; regenerate so it doesn't ship in published types.

## ✅ Strengths (verified)

- **All four regression tests genuinely fail on pre-fix source and pass post-fix** — confirmed by reverting the five source files to merge-base and re-running each. No decorative tests.
- The four fixes are each correct in intent: the split-unsubscriber clobbering fix (real bug per `ReactStore.observe` firing synchronously), the hover-open `{once:true}` mouseup cleanup, the touch-guard reorder (keeps floating-ui and store consistent), and `nested` excluding root context menus (ArrowLeft no longer closes them; submenus still do — the test verifies both).
- `delete rootTriggerProps.id` removal, the redundant `mouseUpTarget === triggerRef.current` check removal (`contains` is inclusive), and the dead `nested-context-menu` MenuParent variant removal are all **verified safe** — none is constructed/relied upon anywhere.
- Comments are high-quality "why" explanations and accurate (aside from the two nits above). `MenuParent` is a clean discriminated union; the removal improves invariant expression.
- Guideline compliance is good: `warn` util, `addEventListener`/`ownerDocument`/`contains` utils, `REASONS.triggerHover`, `'use client'`, one-assertion-per-`waitFor`, no post-`render` `flushMicrotasks`, correct env gating.

## Recommended action

1. **Fix finding #1** — make the parent-subscription setup/teardown symmetric (effect-owned) so it survives StrictMode remount and `handle` toggles; add a `createRenderer`-based + handle-toggle regression test.
2. Decide on finding #3 (scope the touch guard so a real item-press isn't dropped).
3. Apply the comment/naming/coverage nits.
4. Re-run the Menu / Menubar / ContextMenu suites under StrictMode after the fix.

Everything except finding #1 is polish; #1 is the one item I'd block merge on. Want me to draft the fix for the `MenuStore` lifecycle (finding #1) and the guard scoping (finding #3)?

atomiks:
Thanks for the thorough review — addressed in 0c00c1a4b.

## 🔴 Finding 1 — StrictMode severs the parent subscription (fixed)

Made the parent-subscription lifecycle symmetric and re-establishable, as suggested. The `observe('parent')` wiring now lives in `connectParentObserver()`, which is **idempotent** (it tears down any prior observer/subscription first) and is re-run from the mount effect via `connectEffect()`:

- Mount → `connectEffect()` re-arms the observer; unmount → tears it down. A StrictMode mount → simulated-unmount → remount now re-arms on the remount instead of leaving it severed, and the same holds for a `handle` prop toggling defined → undefined → defined.
- The constructor still calls `connectParentObserver()` so **external/`handle` stores** (which never run the effect) stay wired up; for internal stores the effect's first run just reconnects idempotently.

## 🟠 Finding 2 — test bypassed StrictMode (fixed)

`MenuStore.test.tsx` now uses the repo's `createRenderer` (StrictMode on) instead of `@testing-library/react`. Two of the render-based tests **genuinely fail on the pre-fix source** (`expected 2, received 1` — subscription severed after the StrictMode remount) and pass post-fix:
- `stays subscribed to its parent under StrictMode and removes it on unmount`
- `re-arms the internal store subscription when the handle prop toggles`

Also added (per the suggestions) a unit test for the `allowMouseUpTriggerRef` root branch (attach parent → borrows it; detach → restores its own ref).

## 🟠 Finding 3 — touch guard swallowing an item press

I couldn't reproduce this as a live bug, and I think the premise (item close = "native touch PointerEvent") doesn't hold in practice. An item-originated close goes through `useMenuItemCommonProps`, which emits `close` with `domEvent: event` — the **synthetic** React `MouseEvent`, which has no `pointerType`. By the time it reaches the guard, `(nativeEvent as PointerEvent).pointerType === 'touch'` is already `false`, so the guard never matched item closes (only the trigger's own native-PointerEvent click via `useClick`/`triggerPress` ever did). I verified empirically: an item tap inside the guard window closes the menu with or without a scope change.

I still applied the scoping you suggested as **defensive hardening + intent clarification** — the guard now requires `reason === REASONS.triggerPress`, making explicit that it only ever suppresses the trigger's own delayed click. Added `still closes the menu when an item is tapped within the guard window` to lock that invariant in (it passes pre-fix too, since the bug doesn't currently reproduce; it guards against a future change that routed a native pointer event into an item close).

## 🟡 Suggestions

- **`disposeEffect` comment / naming** — the method is gone; replaced by `connectParentObserver` / `disconnectParentObserver` / `connectEffect` with comments that describe the lifecycle accurately (no more "collected anyway… merely symmetric" framing).
- **`MenuRoot.tsx` "clears" → "disarms"** — done.
- **`nested` naming** — extracted `treatAsNestedForListNav` (with a note distinguishing it from the floating-tree `nested = floatingParentNodeId != null`). I kept it derived from the discriminant rather than adding a `nested` field to `MenuParent`, per your note.
- **Modal-warning dedup test** — skipped intentionally: `warn` dedupes via a process-wide module-level `Set` with no reset hook, so a test would be order-dependent (an earlier test emitting the same message would hide the warning) and would pollute later tests. Happy to add one if there's an existing reset pattern I missed.
- **Stale `build/esm/.../MenuRoot.d.ts`** — `packages/react/build/` is git-ignored (`git check-ignore` confirms), so the stale artifact isn't committed and is regenerated at publish; nothing to ship from this branch.

Suites: `MenuStore` 5✓, `Menubar` (jsdom) 79✓, `MenuRoot` 201✓, `ContextMenu` 29✓; typecheck/eslint/prettier clean. (The Chromium `#2222` hover test fails identically on pre-fix HEAD here — pre-existing/environment, unrelated to this change.)

flaviendelangle (commented):
I'm playing with your skill 😆 




I've completed my analysis. Let me verify the production path is truly unaffected by confirming the effect dependency behavior, then finalize. I have enough to write the review.

Key findings verified:
- **MenuStore lifecycle**: confirmed via failing StrictMode test that the constructor-created parent observer/subscription is torn down by the new dispose effect and never recreated.
- **Touch-guard reorder**: correct — `attachPreventUnmountOnClose` only mutates `eventDetails`, harmless to skip when bailing.
- **`nested` fix**: correct and well-scoped — only the cross-orientation close-key path is affected for root context menus.
- **`addEventListener` cleanup, removed self-`contains` check, `delete rootTriggerProps.id` removal, `REASONS.triggerHover`, `nested-context-menu` removal**: all verified safe.

# PR review

This PR fixes four real menu lifecycle/navigation bugs, and most of them are correct. However, the store-lifecycle fix is **not StrictMode-safe**: the parent observer and parent-store subscription are created once in the constructor but torn down by the new dispose effect and never recreated, so after the strict mount → unmount → remount cycle the menu store stops observing its parent. This is merge-blocking (🔴). Production (no StrictMode) is unaffected, but the repo's own test renderer and the React dev default both run StrictMode, and the new regression tests deliberately use a non-strict renderer that hides it. The other three fixes (touch-to-close reorder, ArrowLeft/`nested`, hover `mouseup` cleanup) are correct.

## Bugs (1)

### 1. 🔴 Parent observer/subscription is destroyed under StrictMode and never recreated

**Location:** `packages/react/src/menu/store/MenuStore.ts:199` (with constructor `:141` and `disposeEffect` `:216`)

```ts
// constructor — runs ONCE via useRefWithInit; sets up the observer + parent-store subscription
this.unsubscribeParentObserver = this.observe('parent', (parent) => { ... });

// useStore — the new dispose effect
React.useEffect(
  () => (externalStore ? undefined : internalStore.disposeEffect()),
  [externalStore, internalStore],
);

// disposeEffect — tears both down, but nothing re-creates them on the next setup
private disposeEffect = () => () => {
  this.unsubscribeParentStore?.();
  this.unsubscribeParentStore = null;
  this.unsubscribeParentObserver?.();
  this.unsubscribeParentObserver = null;
};
```

`observe('parent', …)` is called in the constructor, which runs exactly once (`useRefWithInit`). The dispose effect's cleanup tears down both the observer's self-subscription and the per-parent store subscription. Under StrictMode, React runs the effect: setup → cleanup → setup. The cleanup removes both subscriptions, and the second setup does **not** recreate them — `disposeEffect()` only returns a teardown. So after the first strict cycle the store no longer observes `parent`, and its subscription into the parent store is gone for the component's lifetime.

This is a regression introduced by this PR: before it, there was no teardown at all, so under StrictMode the subscription leaked but stayed *functional*; now it is destroyed and not restored.

I confirmed it with a scratch test rendered in `<React.StrictMode>` (now removed):

```
AssertionError: expected 1 to be 2   // parent subscriber count fell back to baseline
  expect(subscriberCount(parent)).toBe(baseline + 1);
```

Production is unaffected (no double-invoke; the observer lives until real unmount, and a toggling `handle` store is handled correctly). The dev impact is bounded today because the cross-parent selectors (`rootId`, `keyboardEventRelay`, `floatingTreeRoot`) are mostly read non-reactively at event time — but this silently disables the reactive parent→child propagation that this PR's own "detached-trigger store sync" purpose depends on, in the default dev environment.

**Failure scenario:** In React StrictMode (the project's `createRenderer` default and the Next.js/Vite dev default), open a nested submenu or detached-trigger menu. After the initial strict mount/remount, the child store no longer re-renders subscribers when the parent's `rootId`/`keyboardEventRelay` changes — the leak "fix" has detached the subscription entirely.

**Fix:** Make the parent-sync lifecycle owned by the effect so a remount re-establishes it. Extract the `observe('parent', …)` setup into a method and call it from the effect setup (recreating the observer when it was previously torn down), keeping `disposeEffect` as the cleanup — e.g.:

```ts
React.useEffect(() => {
  if (externalStore) {
    return undefined;
  }
  internalStore.setupParentSync(); // (re)creates observer + parent subscription if disposed
  return internalStore.disposeEffect();
}, [externalStore, internalStore]);
```

## Tests (1)

### 1. 🟡 New regression tests bypass the project's default StrictMode, so they don't guard the regression above

**Location:** `packages/react/src/menu/store/MenuStore.test.tsx:3`

```ts
import { render } from '@testing-library/react';
```

`MenuStore.test.tsx` imports `render` from `@testing-library/react` (no StrictMode) instead of `#test-utils`'s `createRenderer`, which enables strict effects by default. The whole point of these tests is to pin subscription lifecycle, yet they run in the one mode where the lifecycle is correct. Under strict effects, the "subscription survives while mounted" assertion would fail — exactly the bug in finding #1 — so CI as written cannot catch a StrictMode regression in this code.

**Failure scenario:** The store's parent subscription can be detached under StrictMode (finding #1) with every test still green.

**Fix:** Add a strict-mode case (render inside `<React.StrictMode>`, or via `createRenderer`) asserting the parent subscription survives a mount/remount cycle and that parent updates still propagate after it. It will fail until finding #1 is fixed, which is the intent.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Request changes** — the store lifecycle fix is not StrictMode-resilient (confirmed by a failing strict-mode reproduction in the project's default test environment), and the accompanying regression tests run without StrictMode, so they won't catch it.

---

PR #5035, opened 2026-06-12
[dialog][alert-dialog] Correct inaccurate prop JSDoc

atomiks:
> ⚠️ **Stacked on #5034.** Diff is cumulative until it merges — review this branch's top commit.

Corrects inaccurate public prop JSDoc (regenerated API reference included):

- **`disablePointerDismissal`** — the description was inverted (and omitted that for non-modal dialogs it also disables closing on focus-out).
- **`actionsRef`** — the `unmount` action requires calling `eventDetails.preventUnmountOnClose()` from `onOpenChange`; specifying the ref alone no longer prevents unmount.
- **`Dialog.Trigger` state** — `disabled` reflects the trigger (not the dialog); `open` is true only when the dialog is open **and** was opened by that trigger.
- **`Dialog.Popup` `initialFocus`/`finalFocus`** — document the `null`-fallback and the touch default (the popup itself is focused on touch).
- Document `createAlertDialogHandle`.


atomiks:
Superseded by #5036, which fixes the same JSDoc rot across all components instead of just Dialog/AlertDialog.

PR #5039, opened 2026-06-12
[tooltip][preview-card] Small cleanups and comment fixes

atomiks:
Low-risk, behavior-preserving cleanups in Tooltip and Preview Card.

### Changes

- Memoize the per-trigger `safePolygon()` instance (it captures a `Timeout`, so it must be per-instance, not shared) instead of allocating one per render.
- Collapse `TooltipProvider`'s two `useMemo`s into one; tighten `previousInstantTypeRef`'s type; flatten nested `if (open)` blocks in both roots.
- Move/correct a few comments (the `TooltipRoot` instant-animation rationale, the Preview Card reposition-on-open effect) and fix the `PreviewCardPopup` test's copy-pasted `describe` title.

### Verification

`pnpm typescript`, `pnpm eslint`, jsdom Tooltip / PreviewCard all green.

### DIFF
--- a/packages/react/src/tooltip/provider/TooltipProvider.tsx
+++ b/packages/react/src/tooltip/provider/TooltipProvider.tsx
@@ -12,17 +12,14 @@ import { TooltipProviderContext } from './TooltipProviderContext';
 export const TooltipProvider: React.FC<TooltipProvider.Props> = function TooltipProvider(props) {
   const { delay, closeDelay, timeout = 400 } = props;
 
-  const { contextValue, delayValue } = React.useMemo(
-    () => ({
-      contextValue: { delay, closeDelay } satisfies TooltipProviderContext,
-      delayValue: { open: delay, close: closeDelay },
-    }),
+  const contextValue: TooltipProviderContext = React.useMemo(
+    () => ({ open: delay, close: closeDelay }),
     [delay, closeDelay],
   );
 
   return (
     <TooltipProviderContext.Provider value={contextValue}>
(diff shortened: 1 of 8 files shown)


packages/react/src/tooltip/provider/TooltipProvider.tsx
```diff
@@ -12,16 +12,14 @@ import { TooltipProviderContext } from './TooltipProviderContext';
 export const TooltipProvider: React.FC<TooltipProvider.Props> = function TooltipProvider(props) {
   const { delay, closeDelay, timeout = 400 } = props;
 
-  const contextValue: TooltipProviderContext = React.useMemo(
+  const { contextValue, delayValue } = React.useMemo(
```
michaldudak: We can optimize it even further and use just one of these objects.

atomiks: Done

packages/react/src/tooltip/provider/TooltipProviderContext.ts
```diff
@@ -2,8 +2,8 @@
 import * as React from 'react';
 
 export interface TooltipProviderContext {
-  delay: number | undefined;
-  closeDelay: number | undefined;
+  open: number | undefined;
```
michaldudak: `open` doesn't give a clue this is about the delay. I'd change it to `delay`/`closeDelay` (or, even better. `openDelay`, `closeDelay`) and update FloatingDelayGroup to understand the new name

michaldudak (approved):

PR #5045, opened 2026-06-15
[number field] Deduplicate steppers and simplify internals

atomiks:
Headless refactor of the Number Field internals.

## Changes

- **Deduplicate steppers**: the near-identical `NumberFieldIncrement`/`NumberFieldDecrement` now share a `useNumberFieldStepperButton` hook.
- **Internals tidy**:
  - `getStepAmount` typed `() => number`; dead `DEFAULT_STEP` fallbacks and the constant removed.
  - Remove dead `UseNumberFieldButton*` exports and unused `NumberFieldScrubAreaContext` fields.
  - Extract `isNumeralChar` and dedupe the plus/minus keydown handling; type `getAllowedNonNumericKeys` as `Set<string>`; derive the numeral-detection regexes from their arrays.
  - Rename `getViewportRect`'s misleading `width`/`height` to `right`/`bottom` (they're absolute edge coords).
  - Hoist static button/scrub/cursor style objects; drop a per-stepper `useMemo`.
  - Use the shared `stateAttributesMapping` directly; `Decrement` type-only import.
- **Validation coverage**: Added coverage for pre-round clamping with fractional bounds.
- **Comment/doc corrections**: `onValueChange` keyboard reason, `step="any"` modifier stepping, `teleportDistance`, `aria-roledescription`, `smallStep`/`largeStep` snapping.


flaviendelangle (approved):
## Verdict: ✅ Clean. Approve.

No critical or important issues. This is a careful, behavior-preserving refactor with a few intentional, well-tested fixes. TypeScript checks clean.

## Critical Issues
None.

## Important Issues
None.

## Suggestions (all optional)

1. **`step="any"` interactive stepping is untested — and this PR now leans on it.** *(tests, criticality 6)* Removing `?? DEFAULT_STEP` makes the `step = stepProp === 'any' ? 1 : stepProp` resolution (`NumberFieldRoot.tsx:95`) load-bearing, but the only `step="any"` test covers form validation, not stepping. A regression in that resolution would fail no test. Suggest: `<NumberField defaultValue={5} step="any" />` → click Increase expects `'6'` (and Alt→`5.1`, Shift→`15`, which the new JSDoc now explicitly promises).

2. **Scrub-area pointer-lock `flushSync` removal is untested.** *(tests, criticality 4)* The removed nested `flushSync` relies on `onScrubbingChange` already flushing — verified correct by reading, but no test exercises the post-pointer-lock cursor path (Chromium-only, low value).

3. **Cosmetic: unused `catch (error)` binding** at `NumberFieldScrubArea.tsx:311` (pre-existing, not from this PR). Could be `catch {` to match line 163's style.

4. **Minor asymmetry now visible side-by-side in the shared hook:** `onStop` commits unconditionally while `onClick` commits only when `committed !== prev` (`useNumberFieldButton.ts:84` vs `:118`). Pre-existing and preserved — noting only because dedup put them adjacent.

## What was verified sound (the load-bearing changes)

- **Dropped `useMemo` for `buttonState`** in the new shared `useNumberFieldStepperButton` — safe; `useRenderElement` never uses `state` identity, and a fresh object per render is the established pattern.
- **`commitValue` no-direction + `!details.isCanceled` guard** — the intended fix; avoids a spurious directionally-snapped intermediate and correctly steps from the applied (not rejected) base. New test at `NumberFieldIncrement.test.tsx:397` **was confirmed to fail when the fix is reverted**.
- **`getStepAmount: () => number`** and all `?? DEFAULT_STEP` removals — every path returns a finite number (including `step="any"`→1); the guards were already dead code.
- **`validate.ts` two-pass clamp** — both clamps genuinely needed for non-integer bounds; new test at `validate.test.ts:547` **uniquely covers** the before-rounding pass (fails only when that clamp is removed).
- **`parse.ts`** — `{ regex: groupRegex }` is equivalent (groupRegex is null when group is falsy), `input.includes('∞')` ≡ the old regex, DETECT_REs derived from arrays have no TDZ issue.
- **`getViewportRect` width/height→right/bottom** rename — all branches and the sole consumer updated consistently.
- **ScrubArea context field removals** and **`[minWithDefault, formatStyle]`→`[minWithDefault]`** dep change — no remaining consumers / no stale closure.
- **All 10 modified comments/JSDoc audited accurate** — no comment rot introduced.

## Strengths
- Genuine duplication elimination (two near-identical steppers → one shared hook), directly advancing the "avoid duplicating logic" guideline.
- The risky behavior fixes each ship with a dedicated regression test that the analyzer empirically confirmed fails without the fix.
- High-quality "why over what" comments paired with locking tests; docs (`types.md`) and JSDoc updated together.

## Recommended action
Mergeable as-is. If you want to harden it, add the `step="any"` stepping test (suggestion #1) — that's the one real coverage hole and it's now load-bearing.

PR #5046, opened 2026-06-15
[number field] Fix input and scrub edge cases

atomiks:
Fixes several Number Field interaction, validation, and commit edge cases.

## Changes

- Fixed read-only focus state, paste handling, hidden input validation, and wheel scrubbing guards.
- Fixed seeded increment behavior for empty negative ranges with `snapOnStep`.
- Fixed redundant or missing `onValueCommitted` calls for wheel, blur, keyboard, and canceled changes.
- Fixed caret-aware paste insertion and allowed native handling for unhandled multi-character keys.
- Fixed scrub area cleanup when it unmounts mid-scrub.
- Restored pre-round clamping for fractional bounds.
- Removed invalid `aria-readonly` from stepper buttons; read-only steppers are now exposed as unavailable through button disabled semantics (supersedes #5047).
- Added targeted regression tests.


flaviendelangle (approved):
# PR Review Summary — #5046 `[number field] Fix input and scrub edge cases`

**Verdict: Approve.** This is a high-quality, well-tested bug-fix PR. The code-reviewer agent ran the full suite in the PR worktree: **354 tests pass** (16 skipped), `pnpm typescript` clean, `pnpm eslint` clean on all four changed source files. No correctness bugs, silent failures, or guideline violations were found across all four review dimensions. The findings below are all minor/advisory.

## Critical Issues (0)
None.

## Important Issues (0)
None blocking. The closest items are coverage and maintainability notes below.

## Suggestions

**Test coverage gaps** (pr-test-analyzer):
- `NumberFieldInput.tsx:226-230` — The `blockRevalidationRef` reset (`if (committedValue === value)`) fixes a real stale-validation regression but has no end-to-end test for the two-step sequence: a blur that normalizes text back to the *current* stored value, followed by an external value change that must re-validate. Existing tests cover each half but not the combination. Worth a targeted `validationMode="onBlur"` test.
- `usePressAndHold.ts:166-172` — The unmount cleanup of the global `pointerup` listener is untested. The mouseleave/mouseenter double-commit fix *is* well covered; only unmounting the stepper mid-hold is not. Low priority (scrub-area got an analogous unmount test).

**Comment precision** (comment-analyzer):
- `NumberFieldInput.tsx:180-182` — The empty-field-blur commit comment names only two of the three OR clauses; it omits the `value !== null` case ("an actual clear"). A reader could mistakenly simplify the guard. Add a clause covering it.
- `NumberFieldInput.tsx:95-103` — The `pendingCaretRef` `useIsoLayoutEffect` runs on every render (no deps) intentionally. A one-line note would stop a future reader from "fixing" it with a dependency array and breaking caret timing.

**Maintainability** (silent-failure-hunter):
- `NumberFieldRoot.tsx:246` — On cancel, `setValue` returns `false` *before* `lastChangedValueRef` is updated (correct). A brief comment that the ref is deliberately left untouched on cancel — so the per-interaction reset guards remain authoritative — would prevent a future edit from reintroducing staleness.

## Strengths
- **Correctness of the core mechanism verified end-to-end**: the `setValue`→`changed` return-value plumbing is consistent across all four `incrementValue` call sites; boundary no-ops and canceled changes correctly avoid committing, and `lastChangedValueRef.current ?? valueRef.current` always resolves to the freshly-set value because the ref is updated synchronously inside `setValue`.
- **Excellent test coverage** of the behavioral fixes: onValueCommitted dedup across wheel/blur/keyboard/canceled paths, negative-range seeding (`-5` not `-6`, no directional snap), wheel guards (disabled/readOnly/ctrlKey/unfocused/largeStep), paste-at-caret variants, Home/End without min/max, non-printing key passthrough, and the scrub-area unmount cleanup (correctly Chromium-gated).
- **The `exitPointerLock` try/catch suppression** (`NumberFieldScrubArea.tsx:266-274`) is justified — it mirrors the pre-existing pattern at lines 161-164, the call genuinely can throw on a detached node, local state is reset before the call, and it's internal teardown (not a public error).
- **Comments are unusually high-value** — they explain *why*, not *what*, and every one was verified accurate against the implementation.
- **Full guideline compliance**: `useIsoLayoutEffect`, `useTimeout`/`useInterval`, `useStableCallback`, `ownerDocument`, and the `NOOP`/`clamp` utils all used per AGENTS.md.

## Recommended Action
Mergeable as-is. Optionally, before merge: add the `blockRevalidationRef` regression test (the one untested behavioral fix), and tighten the two comment notes at `NumberFieldInput.tsx:180-182` and `:95-103`. Everything else is nice-to-have.

PR #5047, opened 2026-06-15
[number field]: Fix readonly stepper semantics

lyzno1:
## Summary

Remove invalid `aria-readonly` from `NumberField` increment and decrement buttons.

`aria-readonly` is only valid on value-oriented roles such as `spinbutton` and `textbox`; it is not supported on the `button` role ([WAI-ARIA spec](https://www.w3.org/TR/wai-aria-1.2/#aria-readonly), [button role](https://www.w3.org/TR/wai-aria-1.2/#button)). The readonly state should be exposed by the value control, not by the stepper command buttons.

## Fix

Keep the readonly state on `NumberField.Input`, remove `aria-readonly` from the stepper button props, and route readonly steppers through the existing button disabled semantics so they are exposed as unavailable while preserving `data-readonly` for styling.


atomiks:
Folding this into #5046, which now removes `aria-readonly` from the stepper buttons and routes read-only steppers through the existing button disabled semantics (same change as here, plus a regression test in `NumberFieldRoot.test.tsx`). Closing in favor of that PR.

PR #5048, opened 2026-06-15
[docs] Shell UI updates

aarongarciah:
Preview: https://deploy-preview-5048--base-ui.netlify.app/react/overview/quick-start

Updating a bunch of stuff on the docs shell. Successor of https://github.com/mui/base-ui/pull/4875

- Header
  - Fixed position.
  - Transparent (on desktop).
  - Removes npm and GitHub links from the header.
  - Updates the search trigger styles to match the home page search trigger.
  - Updates the search trigger on mobile so it only displays an icon.
- Sidebars
  - Fixed position.
  - Reduces font size.
  - Updates hover style for left sidebar links.
- Demo shell
  - Brings back the hide code button.
  - Tweaks the show/hide code button styles.
  - Fixes code overflowing the bottom rounded corners in the code block.
  - Fixes the toolbar displaying the scrim when there wasn't content overflowing.

before

master--base-ui netlify app_react_components_accordion

after

localhost_3005_react_components_accordion

PR #5055, opened 2026-06-16
[docs] Demo tabs UI refresh

aarongarciah:
Preview:
- Non-scrollable tabs: https://deploy-preview-5055--base-ui.netlify.app/react/components/accordion
- Scrollable tabs: https://deploy-preview-5055--base-ui.netlify.app/react/handbook/forms

We recently changed the design of the demo tabs in https://github.com/mui/base-ui/pull/4983, but I think this new design is nicer, probably less noisy, and does a better job at separating the demo playground from the toolbar+codeblock.

This new design uses ScrollArea to benefit from its data attributes and CSS vars to display subtle left/right scroll indicators.

It only adds `overscroll-behavior-x: contain` when `data-has-overflow-x` is present i.e. when the tabs' content overflows, so we don't create new layers that could [harm performance](https://github.com/mui/base-ui/pull/5022).

I'd like to improve the "jumping" borders when switching the active tab, so it behaves like e.g. VS Code, where borders between tabs are stable. Maybe another day. This thing got complex enough with focus rings and all.

|before|after|
|---|---|
|Screenshot 2026-06-16 at 14 41 55|Screenshot 2026-06-16 at 14 42 22|
|Screenshot 2026-06-16 at 14 44 11|Screenshot 2026-06-16 at 14 44 28|
|Screenshot 2026-06-16 at 14 42 07|Screenshot 2026-06-16 at 14 42 35|

### DIFF
--- a/docs/src/components/Demo/Demo.css
+++ b/docs/src/components/Demo/Demo.css
@@ -59,71 +59,106 @@
     background-clip: padding-box;
     display: flex;
     align-items: end;
-    gap: 1.5rem;
     height: 2.25rem;
-    padding: 0 0 0 calc(0.25rem - 1px);
+    border-top: 1px solid var(--color-border);
     -webkit-user-select: none;
     user-select: none;
+    overflow: hidden;
+  }
 
-    /* Scroll */
-    overflow: auto hidden;
-    overscroll-behavior-x: contain;
-    scrollbar-width: none;
+  /* The single horizontal scroller for the toolbar. On mobile it holds the tabs
+     and the (in-scroll) actions; at --sm+ it holds just the tabs and the sticky
+     actions sit outside it. */
+  .DemoToolbarScrollAreaRoot {
+    display: flex;
+    flex: 1 1 auto;
+    min-width: 0;
+    align-self: stretch;
+  }
 
-    &::-webkit-scrollbar {
-      display: none;
-    }
+  .DemoToolbarViewport {
+    display: flex;
+    flex: 1 1 auto;
+    min-width: 0;
+    width: calc(100% + 1px);
+    margin-left: -1px;
 
-    /* Scroll containers may be focusable */
+    /* ScrollArea hardcodes `overflow: scroll` on both axes; force horizontal-only
+       so a trackpad gesture can't rubber-band the (non-scrolling) vertical axis. */
+    overflow-y: hidden !important;
+
+    /* Edge-fade scroll indicators at every viewport, driven by the
+       distance-from-edge ScrollArea exposes (`--scroll-area-overflow-x-*`), so
+       each fade is absent at rest and grows in only once scrolled past that
+       edge. The second mask layer keeps the bottom 1px opaque so the active
+       tab's background can bridge the border into the code block below. */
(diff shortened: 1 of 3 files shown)


colmtuite (approved):

oliviertassinari:
Is the alignment on mobile expected? Raising in case, It's left aligned, not right aligned like on desktop.

Screenshot_2026-06-16-20-44-13-89_40deb401b9ffe8e1df2f1cc5ba480b12

aarongarciah:
> Is the alignment on mobile expected? Raising in case, It's left aligned, not right aligned like on desktop.

Fix coming in a follow-up PR: #5056

aarongarciah:
Fix https://github.com/mui/base-ui/pull/5056

oliviertassinari:
@aarongarciah 👍

---

On the tab order:
SCR-20260617-bbjy

I wonder if we don't want:

1. (quick win?) to have the copy icon before the show code so the tab goes to the nearest element; right now, it zigzags (12 -> 13 -> 14 -> 15).
2. (overkill?) to use a Toolbar so as not to have to tab on each element of the demo toolbar (9 -> 10 -> 11 -> 12).

aarongarciah:
@oliviertassinari

> (quick win?) to have the copy icon before the show code so the tab goes to the nearest element; right now, it zigzags (12 -> 13 -> 14 -> 15).

The tab order is correct on [master](https://master--base-ui.netlify.app/react/components/accordion) at the time of writing. It was changed via https://github.com/mui/base-ui/pull/5048

> (overkill?) to use a Toolbar so as not to have to tab on each element of the demo toolbar (9 -> 10 -> 11 -> 12).

I thought about it, but since we're planning to improve the mobile experience across the docs and this might change, I decided to not include this for now.

oliviertassinari:
@aarongarciah

1. Coool.
2. Yeah, fair enough, it's outside of the scope. OK, I will be using #2766 instead to record these kinds of things.

---

One **final** thing, there seems to be a regression, #4983 removes the `window.scrollBy({` logic, but isn't this breaking the docs behavior? Shouldn't we add this back?

Before: https://base-ui.com/react/components/collapsible ✅

https://github.com/user-attachments/assets/1740e551-084d-4261-8549-21e3f90740a3

After: https://master--base-ui.netlify.app/react/components/collapsible ❌

https://github.com/user-attachments/assets/36e39638-d7dc-40de-949f-71be828cb5c9

aarongarciah:
> One final thing, there seems to be a regression, https://github.com/mui/base-ui/pull/4983 removes the window.scrollBy({ logic, but isn't this breaking the docs behavior? Shouldn't we add this back?

@oliviertassinari thanks for spotting! I'll push a fix tomorrow. It was probably removed when I tested without a "hide code" button.

aarongarciah:
> One final thing, there seems to be a regression, #4983 removes the window.scrollBy({ logic, but isn't this breaking the docs behavior? Shouldn't we add this back?

Fixed via https://github.com/mui/base-ui/pull/5071

oliviertassinari:
@aarongarciah Perfect 👍

PR #5067, opened 2026-06-17
[menu] Fix submenu item VoiceOver announcement when opened by keyboard

mj12albert:
Fixes a bug where VoiceOver (reproed with Chrome & Safari) doesn't announce the initially focused submenu item when the submenu opens via keyboard (arrow keys, enter space), recording:

https://github.com/user-attachments/assets/6f7b9f6e-fcd2-41c3-a370-74ee0bf42b0e

With the fix:

https://github.com/user-attachments/assets/952493a5-a480-4ce9-9682-277d5750ee49

This doesn't occur with NVDA

### DIFF
--- a/packages/react/src/floating-ui-react/hooks/useListNavigation.ts
+++ b/packages/react/src/floating-ui-react/hooks/useListNavigation.ts
@@ -447,9 +447,10 @@ export function useListNavigation(
             // omitted here so attribute-disabled items (`disabled`/`aria-disabled`) are skipped
             // on open even when the consumer passes an empty `disabledIndices` array. Passing it
             // would regress that behavior (see mui/base-ui#2604).
-            // macOS Safari VoiceOver can miss this first focus event as a freshly mounted
-            // submenu enters the accessibility tree. Delay only DOM focus; activeIndex still
-            // updates immediately so visual highlight and navigation state are not delayed.
+            // With macOS VoiceOver, WebKit can miss this first focus event as a freshly
+            // mounted submenu enters the accessibility tree. Delay only DOM focus for this
+            // engine-specific path; activeIndex still updates immediately so visual highlight
+            // and navigation state are not delayed.
             delayInitialFocusRef.current =
               platform.os.mac &&
               platform.screenReader.voiceOver &&
--- a/packages/react/src/floating-ui-react/hooks/useListNavigation.ts
+++ b/packages/react/src/floating-ui-react/hooks/useListNavigation.ts
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { useAnimationFrame } from '@base-ui/utils/useAnimationFrame';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
 import { ownerDocument } from '@base-ui/utils/owner';
+import { platform } from '@base-ui/utils/platform';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { useValueAsRef } from '@base-ui/utils/useValueAsRef';
 import { isHTMLElement } from '@floating-ui/utils/dom';
@@ -294,6 +295,7 @@ export function useListNavigation(
   const previousOpenRef = React.useRef(open);
   const forceSyncFocusRef = React.useRef(false);
   const forceScrollIntoViewRef = React.useRef(false);
+  const delayInitialFocusRef = React.useRef(false);
   const cancelQueuedFocusRef = React.useRef<(() => void) | null>(null);
 
   const disabledIndicesRef = useValueAsRef(disabledIndices);
@@ -302,15 +304,16 @@ export function useListNavigation(
   const resetOnPointerLeaveRef = useValueAsRef(resetOnPointerLeave);
 
   const focusFrame = useAnimationFrame();
+  const delayFocusFrame = useAnimationFrame();
   const waitForListPopulatedFrame = useAnimationFrame();
 
   const focusItem = useStableCallback(() => {
-    function runFocus(item: HTMLElement) {
+    function runFocus(item: HTMLElement, sync = forceSyncFocusRef.current) {
       if (virtual) {
         tree?.events.emit('virtualfocus', item);
       } else {
         cancelQueuedFocusRef.current = enqueueFocus(item, {
-          sync: forceSyncFocusRef.current,
+          sync,
           preventScroll: true,
         });
       }
@@ -318,8 +321,10 @@ export function useListNavigation(
 
     const initialItem = listRef.current[indexRef.current];
     const forceScrollIntoView = forceScrollIntoViewRef.current;
+    const shouldDelayInitialFocus =
+      delayInitialFocusRef.current && !forceSyncFocusRef.current && !virtual;
 
-    if (initialItem) {
+    if (initialItem && !shouldDelayInitialFocus) {
       runFocus(initialItem);
     }
 
@@ -331,10 +336,21 @@ export function useListNavigation(
       const waitedItem = listRef.current[indexRef.current] || initialItem;
 
       if (!waitedItem) {
+        delayInitialFocusRef.current = false;
         return;
       }
 
-      if (!initialItem) {
+      if (shouldDelayInitialFocus) {
+        delayInitialFocusRef.current = false;
+        delayFocusFrame.request(() => {
+          if (!latestOpenRef.current) {
+            return;
+          }
+
+          const latestItem = listRef.current[indexRef.current] || waitedItem;
+          runFocus(latestItem, true);
+        });
+      } else if (!initialItem) {
         runFocus(waitedItem);
       }
 
@@ -386,6 +402,8 @@ export function useListNavigation(
     }
     if (!open) {
       forceSyncFocusRef.current = false;
+      delayInitialFocusRef.current = false;
+      delayFocusFrame.cancel();
       return;
     }
     if (!floatingElement) {
@@ -429,6 +447,16 @@ export function useListNavigation(
             // omitted here so attribute-disabled items (`disabled`/`aria-disabled`) are skipped
             // on open even when the consumer passes an empty `disabledIndices` array. Passing it
             // would regress that behavior (see mui/base-ui#2604).
+            // macOS Safari VoiceOver can miss this first focus event as a freshly mounted
+            // submenu enters the accessibility tree. Delay only DOM focus; activeIndex still
+            // updates immediately so visual highlight and navigation state are not delayed.
+            delayInitialFocusRef.current =
+              platform.os.mac &&
+              platform.screenReader.voiceOver &&
+              platform.engine.webkit &&
+              nested &&
+              keyRef.current != null &&
+              !previousMountedRef.current;
             indexRef.current =
               keyRef.current == null ||
               isMainOrientationToEndKey(keyRef.current, orientation, rtl) ||
@@ -459,6 +487,7 @@ export function useListNavigation(
     rtl,
     onNavigate,
     focusItem,
+    delayFocusFrame,
     waitForListPopulatedFrame,
   ]);
 
--- a/packages/react/src/floating-ui-react/hooks/useListNavigation.webkit.test.tsx
+++ b/packages/react/src/floating-ui-react/hooks/useListNavigation.webkit.test.tsx
@@ -1,0 +1,180 @@
+import { vi, it, describe, expect } from 'vitest';
+import * as React from 'react';
+import { act, render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { isJSDOM, useTestInteractions } from '#test-utils';
+import { useFloating, useListNavigation } from '../index';
+
+vi.mock('@base-ui/utils/platform', async () => {
+  const actual =
+    await vi.importActual<typeof import('@base-ui/utils/platform')>('@base-ui/utils/platform');
+
+  return {
+    ...actual,
+    platform: {
+      ...actual.platform,
+      os: { ...actual.platform.os, mac: true, apple: true },
+      engine: { ...actual.platform.engine, webkit: true },
+      screenReader: { ...actual.platform.screenReader, voiceOver: true },
+    },
+  };
+});
+
+function NestedKeyboardMenu({ keepMountedAfterOpen = false }: { keepMountedAfterOpen?: boolean }) {
+  const [open, setOpen] = React.useState(false);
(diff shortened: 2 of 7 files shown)


packages/react/src/floating-ui-react/hooks/useListNavigation.ts
```diff
@@ -429,6 +447,16 @@ export function useListNavigation(
             // omitted here so attribute-disabled items (`disabled`/`aria-disabled`) are skipped
             // on open even when the consumer passes an empty `disabledIndices` array. Passing it
             // would regress that behavior (see mui/base-ui#2604).
+            // macOS Safari VoiceOver can miss this first focus event as a freshly mounted
```
mj12albert: Delaying the ARIA attrs in the menu fixes the issue fully for Chrome, but without this part Safari still skips the submenu item item on the very first open (a subsequent close and reopen of the submenu works)

atomiks (commented):
# PR review

This is a focused VoiceOver fix: it delays the submenu's initial DOM focus by an extra frame on macOS Safari (platform-gated) and defers the trigger↔popup ARIA relationship while a keyboard-opened submenu is active. The focus-delay path is correctly confined to macOS/WebKit/VoiceOver, so it can't regress other platforms. Nothing is merge-blocking. The one thing worth a second look is that the **ARIA deferral is not platform-scoped** — unlike the focus delay, it changes the accessibility tree for every browser and screen reader. All menu tests pass locally (808 passed).

## Bugs (1)

### 1. 🟡 ARIA deferral applies to all screen readers, not just the targeted VoiceOver case

**Location:** `packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx:175` and `packages/react/src/menu/root/MenuRoot.tsx:472`

```tsx
// MenuSubmenuTrigger.tsx
'aria-controls':
  open && lastOpenChangeReason === REASONS.listNavigation ? undefined : popupId,
'aria-expanded':
  open && lastOpenChangeReason === REASONS.listNavigation
    ? false
    : rootTriggerProps['aria-expanded'],
```

```tsx
// MenuRoot.tsx
'aria-labelledby':
  parent.type === 'menu' && open && lastOpenChangeReason === REASONS.listNavigation
    ? undefined
    : activeTriggerElement?.id,
```

The *focus delay* in `useListNavigation.ts` is carefully gated to `platform.os.mac && …voiceOver && …webkit`, but this ARIA deferral has no platform gate. Whenever a submenu is opened by keyboard, on **any** platform/browser/AT (NVDA, JAWS, Orca, Chrome, Firefox), and for the **entire** time it stays open (the condition keys off `lastOpenChangeReason`, which doesn't change while open — the second new test asserts it stays deferred even after focus enters the submenu), the result is:

- the open submenu's trigger reports `aria-expanded="false"` and drops `aria-controls`, and
- the submenu popup (`role="menu"`) drops its `aria-labelledby`, so it has **no accessible name**.

The most concrete consequence: a keyboard-opened submenu has no accessible name, while a click-opened one keeps `aria-labelledby` (the third new test encodes exactly this divergence). That's an inconsistency every screen reader sees, even though the PR description notes the original bug "doesn't occur with NVDA" — i.e. NVDA worked before this change.

Gating the ARIA deferral to the same `platform.os.mac && …voiceOver && …webkit` condition is feasible and carries no SSR/hydration risk: these flags are module-level constants (false on the server), and the deferral only ever activates *after* a client interaction (`open && lastOpenChangeReason === listNavigation`), never during the initial render of a closed submenu — so server and first client render agree regardless.

**Failure scenario:** An NVDA/JAWS user opens a submenu with ArrowRight. Before this PR the submenu announced its name (from the trigger) and the trigger reported expanded; now the submenu is unnamed and the trigger reports collapsed while open.

**Fix:** Either scope the three ARIA overrides behind the same platform predicate used for the focus delay (extract it to a shared constant so the two stay in sync), or, if the cross-AT behavior is intentional and was validated against NVDA/JAWS, leave a comment recording that decision.

## Tests (1)

### 1. ℹ️ The core focus-delay fix has no automated coverage

**Location:** `packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.test.tsx`

The three new tests exercise the (non-platform-gated) ARIA deferral, which is good. But the actual subject of the PR — the two-frame `delayInitialFocusRef` → `delayFocusFrame` focus delay in `useListNavigation.ts` — is unreachable in jsdom because `platform.os.mac`/`webkit`/`voiceOver` are all false there, so `shouldDelayInitialFocus` is never true under test.

**Failure scenario:** A future refactor of the `focusItem` delay machinery (e.g. dropping the `latestOpenRef` guard in the `delayFocusFrame` callback, or the close-path `delayFocusFrame.cancel()`) would pass CI while silently breaking the VoiceOver fix.

**Fix:** This is inherently hard to assert without the real platform, so a unit test may not be practical — but consider noting it as a manual/VoiceOver-only verification step, or factoring the platform predicate so the delay branch can be unit-tested with the flags stubbed.

## Simplifications (1)

### 1. ℹ️ `platform.screenReader.voiceOver` is redundant alongside `platform.os.mac`

**Location:** `packages/react/src/floating-ui-react/hooks/useListNavigation.ts`

```tsx
delayInitialFocusRef.current =
  platform.os.mac &&
  platform.screenReader.voiceOver &&   // voiceOver === apple === (mac || ios)
  platform.engine.webkit &&
  ...
```

`voiceOver` is defined as `apple` (`mac || ios`) in `screen-reader.ts`, so once `platform.os.mac` is required the `voiceOver` term is always true and adds nothing at runtime. `platform.engine.webkit && platform.os.mac` already *is* "Safari on macOS" (the only WebKit on mac), where VoiceOver is the only screen reader — so the term can simply be dropped. It does read as intent-documenting, and the comment above already states the VoiceOver intent, so this is optional.

## Docs (0)

No findings.

## Verdict

**Approve after nits** — no correctness regressions, but the ARIA deferral's unscoped reach across all screen readers (Bug 1) is worth gating or explicitly confirming before merge.

atomiks:
# PR review

Nothing merge-blocking, and the fix's structure is sound — but the diff ships two independent workarounds (the WebKit focus delay and the menu-side ARIA deferral) plus two new util files, and most of the +200 gzip can be recovered: the menu side re-implements keyboard-click detection that already exists in the store as `openMethod`, and the hook side re-implements a delayed-focus path that already exists in `focusItem`. There's also a real chance one of the two mechanisms can be dropped entirely — see Simplifications No. 1.

## Bugs (2)

### 1. 🟡 ARIA relationship stays wrong for the entire lifetime of a keyboard-opened submenu, for all macOS users

**Location:** `packages/react/src/menu/submenu-trigger/MenuSubmenuTrigger.tsx:180`

```tsx
macVoiceOverKeyboardOpen
  ? {
      'aria-controls': undefined,
      'aria-expanded': false,
    }
```

`platform.screenReader.voiceOver` is just the `apple` OS flag (`screen-reader.ts:12` — actual activation is undetectable), so every mac user in every browser gets a trigger that reports `aria-expanded="false"` with no `aria-controls`, and an unnamed submenu popup, for as long as the submenu stays open — the test "keeps the ARIA relationship deferred after keyboard focus moves into the submenu" pins this as permanent, not a one-frame deferral. The deferral is presumably needed only around the moment focus moves into the popup; nothing restores the attributes afterwards.

**Failure scenario:** A VoiceOver user opens a submenu with ArrowRight, browses it, then navigates back to the trigger (VO+Left) while it's still open — VO announces the trigger as collapsed/unexpanded with no controls relationship. Same degraded semantics for any other mac AT, with VoiceOver not even running.

**Fix:** Restore the attributes once initial focus has settled in the submenu (e.g. after the same frame window the hook fix uses), if that doesn't re-trigger the announcement bug. If it does, keep the current behavior but document the tradeoff (see Docs No. 1).

### 2. ℹ️ Inconsistent `keepMounted` gating between the two mechanisms

**Location:** `packages/react/src/floating-ui-react/hooks/useListNavigation.ts:457`

```ts
delayInitialFocusRef.current =
  ... && nested && keyRef.current != null && !previousMountedRef.current;
```

The focus delay only engages when the submenu is freshly mounted, but the ARIA deferral applies to `keepMounted` submenus too. If VoiceOver also misses the focus event when a kept-mounted subtree is *unhidden* (it enters the accessibility tree at that point, same as a fresh mount), the WebKit half of the fix silently won't engage there.

**Failure scenario:** `<Menu.Portal keepMounted>` submenu reopened with ArrowRight in Safari + VoiceOver — the item may again go unannounced.

**Fix:** Verify with VO; either drop `!previousMountedRef.current` or confirm the bug doesn't reproduce with `keepMounted` and leave a comment saying so.

## Tests (1)

### 1. ℹ️ No coverage for keyboard-opening a `keepMounted` submenu

**Location:** `packages/react/src/floating-ui-react/hooks/useListNavigation.webkit.test.tsx:169`

The `keepMountedAfterOpen` variant is only used for the close-before-focus test. There's no test pinning what happens on keyboard *reopen* of an already-mounted list — which is exactly the boundary of the `!previousMountedRef.current` gate above.

**Failure scenario:** The gate's behavior on remount-less reopen is unpinned; a future refactor could flip it either way without a failing test.

**Fix:** Add one case: reopen the kept-mounted menu via keyboard and assert whether focus is delayed or immediate (whichever is intended).

## Simplifications (4)

### 1. 🟡 Establish whether both mechanisms are needed — deleting one is the real byte win

The diff contains two independent fixes, and their gates don't line up with the PR description. `platform.engine.webkit` explicitly excludes Blink (`engine.ts:7` — detected via `-webkit-backdrop-filter`), so on **Chrome/mac the focus delay never runs** and the ARIA deferral alone is apparently fixing the announcement there. That raises two cheap experiments before landing:

- If the ARIA deferral alone fixes Chrome, test whether it alone fixes Safari → delete the entire `useListNavigation` change.
- Conversely, test the focus delay gated on `os.mac` only (all engines), with the ARIA deferral removed → delete both util files and the MenuRoot/MenuSubmenuTrigger changes, which is the bulk of the bytes **and** removes the spec-violating `aria-expanded="false"` from Bugs No. 1.

One of these likely halves-or-zeroes the remaining diff. If both are genuinely required per engine, a comment should say which mechanism serves which engine.

### 2. 🟡 The keyboard-open detection already exists in the store — both new util files can go

**Location:** `packages/react/src/menu/utils/isMacVoiceOverKeyboardOpen.ts:15`

```ts
return openEvent.type === 'click' && (openEvent as MouseEvent).detail === 0;
```

This re-implements the `detail === 0` → keyboard heuristic that already ships in `useEnhancedClickHandler` (`useEnhancedClickHandler.ts:33`), whose result MenuRoot already stores per-menu as `openMethod` and already merges into the submenu trigger's props (`MenuRoot.tsx:440, 457`). The whole check reduces to state already in the `MenuStore`:

```ts
const macVoiceOverKeyboardOpen =
  open &&
  platform.screenReader.voiceOver &&
  (lastOpenChangeReason === REASONS.listNavigation || openMethod === 'keyboard');
```

Add it as one derived selector in `MenuStore.ts` and both call sites (MenuRoot popup props and MenuSubmenuTrigger) become a single `store.useState(...)` read. This deletes `isMacVoiceOver.ts` and `isMacVoiceOverKeyboardOpen.ts`, removes the duplicated computation at two render sites, removes the redundant `'aria-expanded': rootTriggerProps['aria-expanded']` re-spread in the else branch (it's already in `rootTriggerProps`), and as a bonus stops reading the mutable `floatingRootContext.context.dataRef.current.openEvent` during render, which is a rules-of-React smell the current version introduces in both components. The new Enter/Space tests (re-pointed to mock `platform` instead of `isMacVoiceOver`) would confirm `openMethod` covers the raw-`keydown` open path the removed util handled.

### 3. 🟡 The hook already has a delayed-focus path — `delayFocusFrame` and the `sync` param are unnecessary

**Location:** `packages/react/src/floating-ui-react/hooks/useListNavigation.ts:344`

The PR's delayed path is focusFrame rAF → `delayFocusFrame` rAF → sync focus: two frames. The existing `!initialItem` waited path is focusFrame rAF → async `enqueueFocus` rAF: also exactly two frames. So routing the delayed case through the code that's already there gives identical timing:

```ts
let initialItem = listRef.current[indexRef.current];
if (delayInitialFocusRef.current && !forceSyncFocusRef.current && !virtual) {
  // Defer to the waited path below (one extra frame) so the freshly mounted
  // subtree is in the accessibility tree before focus lands.
  delayInitialFocusRef.current = false;
  initialItem = undefined;
}
```

The existing `else if (!initialItem) runFocus(waitedItem)` branch then does the delayed focus. This deletes the `delayFocusFrame` instance, the `sync` parameter on `runFocus`, the nested request block, the `delayFocusFrame.cancel()` on close, and the extra effect dep. For the closed-before-focus guard (the PR's `latestOpenRef` check), `enqueueFocus` already accepts a `shouldFocus` callback built for exactly this — passing `shouldFocus: () => latestOpenRef.current` unconditionally in `runFocus` is one line and is a correct invariant for every path, not just this one. The new webkit tests should pass unchanged since frame counts are identical.

### 4. ℹ️ Redundant platform conjunction

**Location:** `packages/react/src/floating-ui-react/hooks/useListNavigation.ts:452`

`platform.screenReader.voiceOver` is defined as `apple` (= `mac || ios`), so `platform.os.mac && platform.screenReader.voiceOver` is just `platform.os.mac` with an extra property read the minifier can't remove, and `isMacVoiceOver()` is likewise ≡ `platform.os.mac`. Pick one flag per site — `screenReader.voiceOver` alone reads best per the platform module's own guidance (and additionally covers iPadOS hardware keyboards, if that's desirable); add `os.mac &&` only if iOS must be excluded on purpose.

## Docs (1)

### 1. 🟡 The ARIA-deferral mechanism has no explanation of why it works

**Location:** `packages/react/src/menu/utils/isMacVoiceOverKeyboardOpen.ts:4`

The hook change carries a good comment, but the menu-side workaround — the more surprising one, since it deliberately publishes incorrect ARIA — has none. Nothing records why removing `aria-controls`/`aria-labelledby`/`aria-expanded` makes VoiceOver announce the focused item, which engines need it, or why it must persist for the whole open duration rather than just the opening frames.

**Failure scenario:** A future contributor sees `aria-expanded: false` on an open trigger, reasonably assumes it's a bug, "fixes" it, and silently regresses VoiceOver.

**Fix:** A short comment where the boolean is computed: the VO behavior it works around, the engine coverage split with the `useListNavigation` delay, and why it isn't restored while open.

## Verdict

**Approve after nits** — the fix works and nothing is blocking, but the permanent wrong-ARIA window (Bugs No. 1) deserves a look and Simplifications No. 1–3 can reclaim most of the +200 gzip using machinery the codebase already has.

---

PR #5072, opened 2026-06-18
[test] Restore slider listener spies

atomiks:
Restores document listener spies in the slider touchend cleanup regression test so global event methods are not left wrapped.

### DIFF
--- a/packages/react/src/slider/root/SliderRoot.test.tsx
+++ b/packages/react/src/slider/root/SliderRoot.test.tsx
@@ -1391,35 +1391,46 @@ describe('<Slider.Root />', () => {
         const removedListeners: EventListenerOrEventListenerObject[] = [];
         const originalAdd = document.addEventListener.bind(document);
         const originalRemove = document.removeEventListener.bind(document);
-        vi.spyOn(document, 'addEventListener').mockImplementation((type, listener, options) => {
-          if (type === 'touchend' && listener != null) {
-            addedListeners.push(listener);
-          }
-          return originalAdd(type, listener, options);
-        });
-        vi.spyOn(document, 'removeEventListener').mockImplementation((type, listener, options) => {
-          if (type === 'touchend' && listener != null) {
-            removedListeners.push(listener);
-          }
-          return originalRemove(type, listener, options);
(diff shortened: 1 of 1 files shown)


michaldudak (approved):

flaviendelangle (approved):
# PR review

This is a test-hygiene fix that wraps the slider touchend-cleanup regression test in `try/finally` and calls `mockRestore()` on the two `document` event-listener spies. The change is correct and addresses a real cross-test leak, not just cosmetics — I verified against the Vitest 4.1.8 spy source and the repo's global teardown. Nothing is merge-blocking.

The key thing I checked: the global teardown in `test/setupVitest.ts:24` runs `vi.resetAllMocks()`, **not** `vi.restoreAllMocks()`. In `@vitest/spy@4.1.8`, `resetAllMocks()` → `mockReset()` only clears state and nulls the implementation (so the spy falls back to calling through to the original); it never invokes the spy's `restore()`. Only `mockRestore()`/`restoreAllMocks()` actually un-wrap the object property. So without this PR, `document.addEventListener`/`removeEventListener` stay permanently wrapped as Vitest mocks for the rest of the worker's lifetime — breaking function identity and, more concretely, causing any later `vi.spyOn(document, 'addEventListener')` to short-circuit and return the *stale* spy (`spyOn` returns early when the target is already a mock function). The `finally` block also guarantees restoration even when the assertions throw. The fix is well-targeted.

## Bugs (0)

No findings.

## Tests (0)

No findings. The assertions are unchanged (only reflowed by Prettier) — `addedListeners.length > 0` and the every-added-was-removed check still pin the same behavior. This is the only test in the file that spies on the global `document` listener methods; the other `document.addEventListener` usages (head lines ~1509/1548) are inside component effects that remove their own listeners on cleanup, so there is no sibling leak left unaddressed.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve** — a correct, well-scoped fix for a real spy leak (`resetAllMocks` doesn't un-wrap spies), with no behavior change to the assertions and no remaining unrestored global spies.

---

PR #5075, opened 2026-06-18
[code-infra] Remove duplicated path from docs tsconfig

brijeshb42:

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -20,7 +20,7 @@
     "@base-ui/utils": "workspace:*",
     "@mdx-js/loader": "^3.1.1",
     "@mdx-js/react": "^3.1.1",
-    "@mui/internal-docs-infra": "0.11.1-canary.22",
+    "@mui/internal-docs-infra": "0.12.1-canary.3",
     "@next/mdx": "^16.1.6",
     "@react-spring/web": "^10.0.3",
     "@stefanprobst/rehype-extract-toc": "^3.0.0",
--- a/docs/tsconfig.json
+++ b/docs/tsconfig.json
@@ -30,11 +30,6 @@
         "name": "next"
       }
     ],
-    "paths": {
-      "@base-ui/react": ["../packages/react/src"],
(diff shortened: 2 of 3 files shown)


flaviendelangle (approved):

brijeshb42:
Turns out its needed by docs-infra. I'll revisit it.

dav-is:
@brijeshb42 for when you take another look, this is where we parse the paths:

https://github.com/mui/mui-public/blob/b652e3d89b09d3c556f2b1df043b0dc9cbaf7a88/packages/docs-infra/src/pipeline/loadServerTypesMeta/loadServerTypesMeta.ts#L201

https://github.com/mui/mui-public/blob/b652e3d89b09d3c556f2b1df043b0dc9cbaf7a88/packages/docs-infra/src/pipeline/loadServerTypesMeta/loadTypescriptConfig.ts#L131-L149

https://github.com/mui/mui-public/blob/b652e3d89b09d3c556f2b1df043b0dc9cbaf7a88/packages/docs-infra/src/pipeline/loadServerTypesMeta/loadServerTypesMeta.ts#L221

https://github.com/mui/mui-public/blob/b652e3d89b09d3c556f2b1df043b0dc9cbaf7a88/packages/docs-infra/src/pipeline/loadServerTypesMeta/resolveLibrarySourceFiles.ts#L98

If Ctrl+Click still works from the `docs/` when the `docs/tsconfig.json` is missing paths, it should still support parsing the parent paths

brijeshb42:
It seems the main issue somewhere is related to ts's resolution. If I remove the paths, then it fallbacks to the root tsconfig paths which is correct. But if the packages have been built locally, ie, the `build` dir is present, then that is picked over the actual ts paths. The issue is that the `build` might be stale. So devs may get different local and CI behaviour.
So its better to keep the paths in `docs/tsconfig` to always point to the source and have a deterministic behaviour always. The docs-infra assumption was wrong.

PR #5076, opened 2026-06-18
[docs] Fix side nav focus ring

lyzno1:
Fixes the docs side nav focus frame so active items do not visually collide with the scroll area outline.

The side nav uses a negative link margin to align text while allowing the active pill to extend left. This adds matching focus padding around the scroll viewport and draws the viewport focus frame after the content.

| Before | After |
| --- | --- |
| CleanShot 2026-06-18 at 21 55 59 | CleanShot 2026-06-18 at 21 55 23 |

### DIFF
--- a/docs/src/components/SideNav.css
+++ b/docs/src/components/SideNav.css
@@ -56,15 +56,13 @@
     outline: 0;
 
     /* stylelint-disable-next-line selector-pseudo-class-disallowed-list */
-    .SideNavRoot:has(&:focus-visible)::before {
+    .SideNavRoot:has(&:focus-visible)::after {
       content: '';
-      inset: 0;
+      inset: 0 -2px;
       pointer-events: none;
       position: absolute;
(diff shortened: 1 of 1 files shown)


aarongarciah (changes requested):
Thanks for the PR! This is on my to-do list. Sadly, this solution moves the side nav a few pixels down.

https://github.com/user-attachments/assets/ac73e5e7-2662-4982-9c32-5da913a58a79

AFAIK, a viable and simpler solution is to do `inset: 0 -2px` on `.SideNavRoot:has(&:focus-visible)::before`

lyzno1:
Updated, thanks. I switched this to the suggested `inset: 0 -2px` approach so the side nav layout no longer shifts.

I kept the focus frame as `::after` so it is painted above the scrolling content, which prevents the active item from visually covering the top edge of the scroll area focus ring when scrolled to the boundary.

aarongarciah (approved):
Thanks!

PR #5079, opened 2026-06-18
[docs][preview card] Remove text-wrap pretty

aarongarciah:
Safari sucks at `text-wrap: pretty`, so removing it. It doesn't provide any value in this case anyway.

||before|after|
|---|---|---|
|Safari 26.5|safari-before|safari-after|
|Chrome 149|chrome-before|chrome-after|

### DIFF
--- a/docs/src/app/(docs)/react/components/preview-card/demos/index.module.css
+++ b/docs/src/app/(docs)/react/components/preview-card/demos/index.module.css
@@ -96,7 +96,6 @@
   margin: 0;
   font-size: 0.875rem;
   line-height: 1.25rem;
-  text-wrap: pretty;
 }
 
 .Container {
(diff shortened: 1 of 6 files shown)


docs/src/app/(docs)/react/components/preview-card/demos/index.module.css
```diff
@@ -96,7 +96,6 @@
   margin: 0;
   font-size: 0.875rem;
   line-height: 1.25rem;
-  text-wrap: pretty;
 }
 
 .Container {
```
atomiks, line 101: Tailwind?

aarongarciah: Updated. Also removed some unnecessary `m-0` classes.

atomiks (approved):

PR #5081, opened 2026-06-18
[combobox] Expose focusItemOnOpen prop

Tboodle:
Having worked with the combobox component, I noticed that focus was being set to the first selected item when `selectionMode` is set to `multiple`, and that there was no way to instead set focus upon re-opening to the first item in the list. By exposing the `focusItemOnOpen` prop, users can better control this behavior, which I believe makes sense to support.


aarongarciah:
@Tboodle thanks for your interest. Can you open an issue where you describe your exact use case? Any meaningful change to behavior and/or API needs proper discussion and thinking before jumping into implementation.

I'll close this PR; let's move the discussion to the issue once you create it. Thanks!

PR #5082, opened 2026-06-18
[docs-infra] Use versioned GitHub URLs

aarongarciah:
Point 18. in #2766.

Use the lib version instead of HEAD for the "View source on GitHub" and the "Copy link to source" demo links (visible under the three dot menu).

Old: https://github.com/mui/base-ui/tree/HEAD/docs/src/app/(docs)/react/components/accordion/demos/hero/css-modules
New: https://github.com/mui/base-ui/tree/v1.6.0/docs/src/app/(docs)/react/components/accordion/demos/hero/css-modules

Preview: https://deploy-preview-5082--base-ui.netlify.app/react/components/context-menu

### DIFF
--- a/docs/src/utils/getGitHubDemoUrl.test.ts
+++ b/docs/src/utils/getGitHubDemoUrl.test.ts
@@ -1,11 +1,21 @@
-import { describe, it, expect } from 'vitest';
-import { getGitHubDemoUrl, GITHUB_BASE } from './getGitHubDemoUrl';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { getGitHubDemoUrl } from './getGitHubDemoUrl';
 
 describe('getGitHubDemoUrl', () => {
+  const GITHUB_BASE = 'https://github.com/mui/base-ui/tree/v1.6.0';
   const unixUrl =
     'file:///home/user/base-ui/docs/src/app/(docs)/react/components/accordion/demos/hero/index.ts';
   const windowsUrl =
     'file:///C:/Users/Dev/base-ui/docs/src/app/(docs)/react/components/accordion/demos/hero/index.ts';
+
+  beforeEach(() => {
+    vi.stubEnv('SOURCE_CODE_REPO', 'https://github.com/mui/base-ui');
+    vi.stubEnv('LIB_VERSION', '1.6.0');
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
 
   it('converts a Unix file URL to a GitHub directory URL', () => {
     expect(getGitHubDemoUrl(unixUrl)).toBe(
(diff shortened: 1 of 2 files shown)


colmtuite (approved):

michaldudak:
## Review summary

Reviewed the two changed files (`docs/src/utils/getGitHubDemoUrl.ts` + test) across three angles: general code quality, test coverage, and silent-failure/error handling.

**Verdict:** Correct, well-tested, and ready to merge. All 11 tests pass. No Critical and no blocking Important issues. The removed `GITHUB_BASE` export was only consumed by its own test, and the consumer (`Demo.tsx`) already handles the new `null` return gracefully (early-return in the copy handler + conditional render of the menu items), so missing env simply hides the links rather than producing broken hrefs.

### Recommended improvements (optional — these are docs-site infra, not user-facing library code)

1. **Silent degradation on missing env metadata** — `getGitHubDemoUrl.ts:7-9`
   Since `next.config.mjs` always injects `SOURCE_CODE_REPO` and `LIB_VERSION` at build time, a `null` from `getGitHubBaseUrl()` in practice signals a broken build config — but the only symptom is "View source" links silently vanishing site-wide. Consider a dev-only `console.warn` before returning `null`, preserving graceful degradation while making misconfiguration loud locally.

2. **The `v`-prefix transform is asserted only via a coupled literal** — `getGitHubDemoUrl.test.ts:5,13`
   The headline behavior (`LIB_VERSION=1.6.0` → `tree/v1.6.0`) is verified through the hardcoded `GITHUB_BASE` constant, which couples "version is 1.6.0" with "prefix is v". A focused test with a distinctive version (`stubEnv('LIB_VERSION', '9.9.9')` → `toContain('/tree/v9.9.9/')`) would make the contract explicit and regression-proof.

### Suggestions (all optional, none reachable today)

- **Duplicated logic with `ViewSourceLink.tsx:5-26`** — the `${repo}/tree/v${version}` derivation now exists twice. A shared helper would consolidate it, though the two sites read env at different times (module-load vs. call-time), so it's a judgment call.
- **Malformed-but-present env** (trailing slash in `SOURCE_CODE_REPO`, non-semver `LIB_VERSION`) would yield a broken non-`null` URL — the guard checks absence, not validity. Not reachable today since the values are controlled literals.
- **Pre-existing bare `catch { return null }`** (`getGitHubDemoUrl.ts:55-56`) now also wraps the new code; can't realistically throw, but a dev-only `console.warn(error)` would cost nothing.
- **Both env vars missing simultaneously is untested** — the single-operand tests already cover the `||` branch, so marginal value.

### Strengths

- Moving env reads into a function body (vs. `ViewSourceLink`'s module-level reads) is a deliberate tradeoff that enables `vi.stubEnv` testing.
- Removing the `GITHUB_BASE` export and re-declaring it as a test-local literal stops the test from trivially agreeing with the production constant — existing tests now genuinely assert the full versioned URL.
- The `string | null` contract is honored end-to-end.

<sub>🤖 Automated multi-agent review via Claude Code.</sub>

PR #5083, opened 2026-06-19
[number field] Add regression test for interleaved percent sign (1%2)

### LINKED ISSUE #2963: [number field] dont allow to input `1%2` or parse it accordingly
# Bug report

## Current behavior

Suppose the current value of a percentage input is `0.01`.

* It renders as `1%`.
* When I click the input, the cursor is after `1%`.
* I type `2`, the input displays `1%2`.
* On losing focus, the value resets to `1%`.

## Expected behavior

* Typing `2` after `1%` should result in `12%`.

## Reproducible example

https://codesandbox.io/p/sandbox/c6z7f5

## Proposal 1

change [parseNumber](https://github.com/mui/base-ui/blob/c204fef0ac467db20492aff1145a97a379e63a4f/packages/react/src/number-field/utils/parse.ts#L93):

```ts
it('handles percentages with style: "percent"', () => {
  expect(parseNumber('12%', 'en-US', { style: 'percent' })).to.equal(0.12);
  expect(parseNumber('1%2', 'en-US', { style: 'percent' })).to.equal(0.12); // <- new
});

it('handles percentages with style: "unit" and unit: "percent"', () => {
  expect(parseNumber('12%', 'en-US', { style: 'unit', unit: 'percent' })).to.equal(12);
  expect(parseNumber('1%2', 'en-US', { style: 'unit', unit: 'percent' })).to.equal(12); // <- new
});
// ... and other styles: 
```

## Proposal 2:

if user click on input field on the right THEN it should not add focus bar here `1%|` but to here `1|%`

## Proposal 3:

the @react-input/number-format handles this ok

https://codesandbox.io/p/sandbox/react-input-number-format-r234d9

- when input is `11.13%` then trying to input `11.13%9` doesnt change anything
- when input is `1%` then trying to input `1%9` will change to `19%`

so we can do same way https://github.com/GoncharukOrg/react-input/blob/main/packages/number-format/src/utils/filter.ts and just clean out any non decimals

atomiks:
Adds the missing regression coverage for [#2963](https://github.com/mui/base-ui/issues/2963) ("[number field] dont allow to input `1%2` or parse it accordingly").

The behavior was already fixed by #4804 ("[number field] Respect Intl rounding options on blur"), which introduced global `%` stripping via `PERCENT_GLOBAL_RE` in `packages/react/src/number-field/utils/parse.ts`. However, no test covered the specific interleaved `1%2` case — the existing percent tests only exercise `12` / `12%`. This PR adds that coverage so the issue can be closed.

## What's tested

The exact repro from the issue: a percent-formatted field with value `0.01` renders as `1%`. Typing `2` at the end (cursor after `1%`) produces `1%2`, which must reformat to `12%` on blur and commit `0.12`. Before #4804 it incorrectly reset to `1%` (parsing only the leading `1`).

- **Component test** in `NumberFieldRoot.test.tsx`: drives the focus → `1%2` → blur flow and asserts the input reformats to `12%` and `onValueCommitted` fires with `0.12`.
- **Parser unit test** in `parse.test.ts`: `parseNumber('1%2', ...)` returns `0.12` for `style: 'percent'` and `12` for `style: 'unit', unit: 'percent'`.

Closes #2963


flaviendelangle (approved):
# PR review

Test-only PR adding regression coverage for the `1%2` interleaved-percent case fixed by #4804. I ran both touched files — all assertions pass (`parse.test.ts`: 48 passing; `NumberFieldRoot.test.tsx`: 139 passing) — and traced `parseNumber` to confirm the tests exercise the claimed path (`%` stripped via `PERCENT_GLOBAL_RE` → `12` → `shiftDecimal(-2)` → `0.12`; unit-percent skips the shift → `12`). Nothing is merge-blocking. One optional robustness note below.

## Bugs (0)

No findings.

## Tests (1)

### 1. ℹ️ Component test asserts locale-formatted strings while relying on the default locale

**Location:** `packages/react/src/number-field/root/NumberFieldRoot.test.tsx:520` (and `:525`)

```tsx
expect(input).toHaveValue('1%');
// ...
expect(input).toHaveValue('12%');
```

Unlike the sibling parser test in this same PR, which pins the locale explicitly (`parseNumber('1%2', 'en-US', …)`), the component test passes no `locale` and asserts the *rendered* strings `'1%'` / `'12%'`. Percent formatting is locale-sensitive: the default resolves to `en-US` → `'1%'` here and in CI, but e.g. `fr-FR` produces `'1 %'` (narrow no-break space `U+202F`), which would fail the exact-string match. The numeric assertion (`onValueCommitted` → `0.12`) is locale-stable; only the two `toHaveValue` string checks carry this dependency.

This is consistent with the neighboring `parses percent while typing` test (line 484), which also relies on the default locale — so it follows established convention and is not a defect. Adding `locale="en-US"` to the `<NumberField>` would make the rendered-string assertions self-contained and match the parser tests, if you want to harden it.

**Failure scenario:** A CI runner or contributor machine whose ICU default locale is not `en-US` would see `toHaveValue('1%')` fail against a space-separated format, even though the behavior under test is correct.

**Fix:** Optionally pass `locale="en-US"` to the `<NumberField>` in this test so the literal `'1%'`/`'12%'` assertions don't depend on the environment's default locale.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve** — the two tests are correct, pass against the current implementation, pin a real pre-#4804 regression (`1%2` → `12%`/`0.12`), follow existing conventions, and are complementary rather than redundant; the only note is an optional locale-robustness nit.

---

PR #5086, opened 2026-06-19
[combobox] Stop filtering grouped items after limit

lyzno1:
## Summary
- stop filtering grouped combobox items once the configured limit has been reached
- add a regression test that verifies items after the limit are not passed through the custom filter

## Root cause
Grouped combobox filtering built a full filtered array for each group before slicing it to the remaining limit. This performed unnecessary filter work when the limit had already been satisfied inside the current group.

## Validation
- pnpm test:jsdom Combobox --no-watch
- pnpm eslint -- packages/react/src/combobox/root/AriaCombobox.tsx packages/react/src/combobox/root/ComboboxRoot.test.tsx
- pnpm typescript
- pnpm exec prettier --check packages/react/src/combobox/root/AriaCombobox.tsx packages/react/src/combobox/root/ComboboxRoot.test.tsx


flaviendelangle (commented):
# PR review

This is a self-contained performance optimization to grouped combobox filtering: instead of filtering an entire group and then slicing to the remaining limit, it filters item-by-item and stops once the per-group remaining limit is reached. I verified the output is behaviorally identical to the old code across all branches (`filterQuery === ''`, `limit === -1`/`Infinity`, partial-limit groups, empty groups), so the only observable change is fewer `filter` calls — which is the intent and matches the already-existing non-grouped path. Nothing is merge-blocking; only two minor optional notes.

## Bugs (0)

No findings.

The `remainingLimit` is always ≥ 1 when the loop body runs (the top-of-loop guard `currentCount >= limit` breaks first), so the `slice(0, remainingLimit)` and `length >= remainingLimit` checks never degenerate to a zero/empty window. The removed `if (candidateItems.length === 0) continue;` is correctly subsumed by the existing `if (itemsToTake.length > 0)` push guard, so empty groups still don't advance `currentCount` or get pushed. Output and ordering match the previous implementation exactly.

## Tests (1)

### 1. ℹ️ Redundant `.not.toBe(null)` on `getByRole`

**Location:** `packages/react/src/combobox/root/ComboboxRoot.test.tsx:5189`

```tsx
expect(screen.getByRole('option', { name: 'apple' })).not.toBe(null);
expect(screen.getByRole('option', { name: 'apricot' })).not.toBe(null);
```

`getByRole` throws when the element is absent, so the `.not.toBe(null)` assertion can never observe a `null` — the presence check is already enforced by the call itself. The adjacent absence assertions correctly use `queryByRole(...).toBe(null)`. This is cosmetic; the test still pins the intended behavior (including the strong assertion on `filter.mock.calls` that `avocado` and `artichoke` are never filtered, which is what makes this a real regression test for the optimization).

**Failure scenario:** No functional gap — purely a slightly misleading assertion. Dropping `.not.toBe(null)` (or switching to `findByRole`) would read more idiomatically.

**Fix:** Use `screen.getByRole('option', { name: 'apple' });` as the presence assertion, or `expect(screen.queryByRole(...)).not.toBe(null)`.

## Simplifications (1)

### 1. ℹ️ `filterQuery` emptiness is checked twice

**Location:** `packages/react/src/combobox/root/AriaCombobox.tsx:281`

```tsx
const itemsToTake = filterQuery === '' ? group.items.slice(0, remainingLimit) : [];

if (filterQuery !== '') {
  for (const item of group.items) {
    ...
  }
}
```

The ternary and the following `if` test the same `filterQuery` condition in opposite directions. An `if/else` would express the two mutually exclusive paths once and avoid the empty-array placeholder. This is a readability nit, not a correctness issue.

**Failure scenario:** Slightly harder to read — a reader must connect the `: []` placeholder to the populating loop below it.

**Fix (optional):**
```tsx
const remainingLimit = limit > -1 ? limit - currentCount : Infinity;
let itemsToTake: Value[];
if (filterQuery === '') {
  itemsToTake = group.items.slice(0, remainingLimit);
} else {
  itemsToTake = [];
  for (const item of group.items) {
    if (itemsToTake.length >= remainingLimit) {
      break;
    }
    if (filter(item, filterQuery, itemToStringLabel)) {
      itemsToTake.push(item);
    }
  }
}
```

## Docs (0)

No findings.

## Verdict

**Approve** — the optimization is behavior-preserving across every branch, the regression test correctly asserts on filter-call count, and the only remaining items are two optional style notes.

---

atomiks (approved):

Issue #3938, opened 2026-02-02
[discussion] Base UI Integration with Angular?


claridgicus:
## Ask a question

Is there an appetite for an angular port of base-ui to add to the supported frameworks list,
Totally understand if the answer is no, it's overhead I get it.

I would happily submit a PR for this if there is appetite.

mj12albert:
@claridgicus See https://github.com/mui/base-ui/issues/2200#issuecomment-3172874197

claridgicus:
Understandable position @mj12albert 

Guess I'll keep betting on the underdog 😉

oliviertassinari:
<img width="238" height="50" alt="Image" src="https://github.com/user-attachments/assets/5f817c08-cf4a-4b2e-bc19-64c5fc691b62" />

The current status of this issue is wrong; it's not "completed"

Considering the other ones: https://github.com/mui/base-ui/issues?q=is%3Aissue%20%22%5Bdiscussion%5D%20Base%20UI%20Integration%20with%22, we can keep this one open, especially because Angular is the biggest community outside of React.

Overall, I would be curious to understand what's missing from Angular CDK https://material.angular.dev/cdk/accordion/overview

oliviertassinari:
One project that works on this direction: https://github.com/radix-ng/primitives

> A low-level, accessible component library that adapts Base UI 's architecture and behavior to Angular.

PR #5090, opened 2026-06-20
[docs] Add new library on community page

pimenovoleg:
Added new library to community page: [radix-ng.com](https://radix-ng.com/)

Related to #3938

### DIFF
--- a/docs/src/app/(docs)/react/overview/community/page.mdx
+++ b/docs/src/app/(docs)/react/overview/community/page.mdx
@@ -29,6 +29,7 @@ Here's a non-exhaustive list of open-source styled libraries built with Base UI
 
 - [base-ui-solid](https://github.com/msviderok/base-ui-solid): unofficial port of Base UI to SolidJS.
 - [base-ui-vue](https://github.com/vuepont/base-ui-vue): unofficial port of Base UI to Vue.
+- [radix-ng](https://github.com/radix-ng/primitives): unofficial Angular adaptation of Base UI architecture and behavior (originally a Radix UI port).
 
 ## Get help and contribute
 


mj12albert:
Closing as this seems to be an Angular lib that doesn't actually use Base UI

pimenovoleg:
Hi @mj12albert , thanks for taking a look. 
Just to clarify — while the project started as a Radix port, the API and behavior now primarily align with Base UI, and we're actively converging on Base UI's architecture. I actually checked with @colmtuite before opening this and he gave the go-ahead.
Happy to adjust the wording on the community entry to make the relationship clearer if that helps.

mj12albert:
Reopened, I'll leave it to Colm to review it then

> the API and behavior now primarily align with Base UI, and we're actively converging on Base UI's architecture

@pimenovoleg It's still going to be an Angular lib that doesn't use Base UI as a dependency though right?

pimenovoleg:
> Reopened, I'll leave it to Colm to review it then
> 
> > the API and behavior now primarily align with Base UI, and we're actively converging on Base UI's architecture
> 
> @pimenovoleg It's still going to be an Angular lib that doesn't use Base UI as a dependency though right?

Correct — it doesn't use Base UI as an npm dependency, since Base UI is React-specific and can't be consumed directly from Angular. 
It's a from-scratch Angular implementation that follows Base UI's component API, behavior, and architecture (anatomy, prop/part naming, interaction patterns), the same way the other unofficial ports listed on the page reimplement Radix/Base UI for non-React frameworks. 
So it's a "port" in the same sense as the existing community entries, not a wrapper around the React package.

aarongarciah (approved):
Thanks! The library could use a rename, though 😅

PR #5093, opened 2026-06-20
[all components] Restore visible focus after keyboard close

### LINKED ISSUE #5092: [focus] Visible focus is not restored after keyboard dismissal in Safari
## Problem

In Safari/Firefox, popup components can restore DOM focus to their trigger after being dismissed with <kbd>Escape</kbd>, but the trigger does not match `:focus-visible`.

This makes it look like focus was lost even though `document.activeElement` is the trigger.

## Reproduction

1. Open a popup with a pointer/mouse click.
2. Press <kbd>Escape</kbd> to close it.
3. Inspect the trigger.

Confirmed with Menu and Dialog demos:

```js
document.activeElement === trigger // true
trigger.matches(':focus-visible')  // false in WebKit, true in Chromium
```

## Expected behavior

When a popup is dismissed by keyboard, the returned focus should be visible so keyboard users can see where focus landed.

## Notes

This appears to affect the shared `FloatingFocusManager` return-focus path rather than individual Menu/Dialog triggers. The return target is correct; the visible focus state is not restored in WebKit after pointer-open + keyboard-dismiss.

lyzno1:
Fixes #5092

This is a tentative fix for a Safari/Firefox focus-visible issue observed across popup components that use the shared `FloatingFocusManager` return-focus path.

When a popup is opened with a pointer and dismissed with <kbd>Escape</kbd>, WebKit restores DOM focus to the trigger but does not mark it as `:focus-visible`. Since the focus manager already tracks keyboard dismissal, this change requests visible focus only for keyboard-driven return focus.

| Before | After |
| --- | --- |
| Before: focus is restored without visible focus indication | After: focus is restored with visible focus indication |

This PR is intentionally small and can be closed if this behavior should be handled differently.

Validation:

- `pnpm test:webkit MenuPopup --no-watch`
- manual Playwright WebKit check against the Menu e2e fixture: pointer open + Escape close returns focus with `:focus-visible`

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
@@ -551,6 +551,58 @@ describe('FloatingFocusManager', () => {
         HTMLElement.prototype.focus = originalFocus;
       });
 
+      test('passes focusVisible when returning focus after keyboard close', async () => {
+        function App() {
+          const [isOpen, setIsOpen] = React.useState(false);
+
+          const { refs, context } = useFloating({
+            open: isOpen,
+            onOpenChange: setIsOpen,
+          });
+
+          const click = useClick(context);
+          const dismiss = useDismiss(context);
+
+          const { getReferenceProps, getFloatingProps } = useTestInteractions([click, dismiss]);
+
+          return (
+            <>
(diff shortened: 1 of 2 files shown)


atomiks (approved):
Thanks for the fix @lyzno1

PR #4904, opened 2026-05-25
[meter] Sync value text with indicator

atomiks:
## Changes

- Keep Meter's default value text, aria-valuetext, and indicator width aligned for custom ranges.
- Clamp Meter indicator width and aria-valuenow when values fall outside the range.
- Add Meter coverage for localized text, custom ranges, bounded ARIA values, and value updates.

PR #5095, opened 2026-06-22
[progress] Fix custom min/max semantics to match the indicator

atomiks:
## Summary

`Progress.Indicator` already normalized the value against `min`/`max` (`valueToPercent`), but the rest of `Progress.Root` used the raw value, so custom-range progress bars were inconsistent. With `min={20} max={40} value={30}` the indicator filled to 50% while `Progress.Value` and `aria-valuetext` reported `30%`.

This mirrors the fix Meter already shipped (#4904): compute a single clamped value and normalized percentage once in the Root and use it everywhere.

## What changed

- **`ProgressRoot.tsx`** — for finite values, compute one `clampedValue` and one clamped `percentageValue`, then use them consistently:
  - completion status is now `clampedValue === max` (an overshoot like `value=45, max=40` correctly reports `complete` instead of `progressing`);
  - `aria-valuenow` is `clampedValue`, so it can no longer fall outside `aria-valuemin`/`aria-valuemax`;
  - the formatted text / default `aria-valuetext` use `formatNumber(percentageValue / 100, …, { style: 'percent' })` when no `format` is given, so they match the indicator for any range;
  - `format` is read directly instead of through `useValueAsRef`, removing the one-commit lag when `format` changes.
- **`ProgressRootContext.tsx`** — exposes `percentageValue` (clamped, `null` while indeterminate).
- **`ProgressIndicator.tsx`** — reads `percentageValue` from context instead of recomputing it, making the Root the single source of truth (this also clamps the indicator width).
- **`formatNumber.ts`** — removed the now-unused `formatNumberValue` helper that hard-coded `value / 100` and ignored `min`/`max`.

`value={null}` (indeterminate) behavior is unchanged.

## Tests

Added regression tests in `ProgressRoot.test.tsx` (verified failing before the fix, passing after):
- custom range `min=20 max=40 value=30` normalizes the value text, `aria-valuetext`, and indicator width to 50%;
- `aria-valuenow` is clamped to `max` when the value overshoots;
- `complete` is reported when the value reaches/exceeds `max`;
- a `format` change is reflected immediately (no stale-ref lag).

Full `Progress` and `Meter` jsdom suites pass; `pnpm typescript`, `pnpm eslint`, and `pnpm prettier` are clean.

### DIFF
--- a/packages/react/src/progress/indicator/ProgressIndicator.tsx
+++ b/packages/react/src/progress/indicator/ProgressIndicator.tsx
@@ -1,7 +1,6 @@
 'use client';
 import * as React from 'react';
 import { useRenderElement } from '../../internals/useRenderElement';
-import { valueToPercent } from '../../utils/valueToPercent';
 import type { ProgressRootState } from '../root/ProgressRoot';
 import { useProgressRootContext } from '../root/ProgressRootContext';
 import { progressStateAttributesMapping } from '../root/stateAttributesMapping';
@@ -19,10 +18,7 @@ export const ProgressIndicator = React.forwardRef(function ProgressIndicator(
 ) {
   const { render, className, style, ...elementProps } = componentProps;
 
-  const { max, min, value, state } = useProgressRootContext();
-
-  const percentageValue =
-    Number.isFinite(value) && value !== null ? valueToPercent(value, min, max) : null;
+  const { percentageValue, state } = useProgressRootContext();
 
   const indicatorStyle: React.CSSProperties =
     percentageValue == null
--- a/packages/react/src/progress/root/ProgressRoot.test.tsx
+++ b/packages/react/src/progress/root/ProgressRoot.test.tsx
@@ -55,6 +55,46 @@ describe('<Progress.Root />', () => {
     });
   });
 
+  describe('range', () => {
+    it('normalizes the formatted value, aria-valuetext, and indicator within a custom range', async () => {
+      const expected = (0.5).toLocaleString(undefined, { style: 'percent' });
+
+      await render(
+        <Progress.Root min={20} max={40} value={30}>
+          <Progress.Value data-testid="value" />
+          <Progress.Track>
+            <Progress.Indicator data-testid="indicator" />
+          </Progress.Track>
+        </Progress.Root>,
+      );
+
+      const progressbar = screen.getByRole('progressbar');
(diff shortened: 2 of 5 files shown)


flaviendelangle (approved):
# PR review

This is a clean, well-scoped fix that makes `Progress.Root` normalize value/percentage once and use it everywhere, mirroring the Meter fix (#4904). For the common default range (`min=0, max=100`) the output is byte-for-byte identical to before (`clamp(valueToPercent(v,0,100),0,100) === v` for in-range `v`), so there's no regression on the hot path; the changes only affect custom ranges and overshoot, which is the intended fix. The `useValueAsRef` removal genuinely fixes the one-commit `format` lag (confirmed: `useValueAsRef` updates `.current` in a layout effect, so the render reading `.current` sees the stale value). Tests cover the new cases, and no orphaned references to the removed `formatNumberValue` remain in source. Nothing is merge-blocking.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (1)

### 1. ℹ️ `${value}%` fallback in `getDefaultAriaValueText` is now unreachable and would be wrong for custom ranges

**Location:** `packages/react/src/progress/root/ProgressRoot.tsx:16`

```tsx
function getDefaultAriaValueText(formattedValue: string | null, value: number | null) {
  if (value == null) {
    return 'indeterminate progress';
  }
  return formattedValue || `${value}%`;
}
```

After this PR `formattedValue` is always a non-empty string for any finite `value` (it's a percent or formatted number), and `value == null` returns early — so the `` `${value}%` `` branch is dead. It's pre-existing and harmless, but it's now also semantically stale: it hard-codes the old "raw value as percent" assumption the PR deliberately moves away from, so if it ever did fire on a custom range it'd contradict the indicator. Not introduced by this PR and safe to leave; worth dropping to `return formattedValue` while this file is being touched.

**Failure scenario:** Future refactor that allows an empty `formattedValue` would silently resurrect the inconsistent `30%`-style text this PR fixed.

**Fix:** Simplify to `return formattedValue;` (the `value == null` guard already handles the only empty case).

## Verdict

**Approve** — correctness-improving fix with matching tests; no 🔴 or 🟡 findings, only one optional dead-code cleanup. I reasoned through the behavior rather than executing the suite (the working tree is on `master`, so the PR's new tests can't run against the unmodified source locally); the logic and the Meter parity check out.

---

PR #5096, opened 2026-06-22
[dialog] Fix touch outside-press dismissal without a backdrop

atomiks:
## Summary

Tapping outside a `Dialog` with touch did not dismiss it within a **5px-10px touchmove range** when the dialog had no `Dialog.Backdrop` — i.e. `modal={false}` or `modal="trap-focus"`. Mouse outside-press worked; only the touchend dismissal path was broken.

## Cause

`useDismiss` invokes the dialog's `outsidePress` callback with the **`touchend`** event after a small outside touch movement. On `touchend`, the lifted point is reported in `event.changedTouches` and `event.touches` is empty for a completed single-finger gesture. The dialog-only guard rejected anything where `event.touches.length !== 1`:

```js
if ('touches' in event && event.touches.length !== 1) {
  return false;
}
```

So in the touchend-dismiss window, `0 !== 1` → the press was wrongly rejected and the dialog stayed open. This is narrow to reproduce manually: movement under 5px does not arm touchend dismissal, while movement over 10px dismisses earlier during `touchmove` where `touches.length === 1`. The broken case is the middle window (>5px and <=10px) that waits for `touchend`. This guard is dialog-only legacy from #3011; `Popover` has no such guard and relies on floating-ui defaults.

## Fix

Handle `touchend` separately: accept it only when exactly one touch ended and no other touches remain active. `touchstart`/`touchmove` still require exactly one active touch, preserving the original multi-touch rejection.

```js
if ('touches' in event) {
  if (event.type === 'touchend') {
    if (event.changedTouches.length !== 1 || event.touches.length !== 0) {
      return false;
    }
  } else if (event.touches.length !== 1) {
    return false;
  }
}
```

## Testing

Added Chromium-only regression coverage for the narrow touchend-dismiss window by dispatching a real single-finger sequence with a ~6px move (`touchstart` → `touchmove` → `touchend` with the lifted point in `changedTouches`) across both `modal={false}` and `modal="trap-focus"` with no backdrop. The 6px move intentionally arms touchend dismissal without crossing the >10px immediate touchmove-dismiss threshold. Also added a multi-touch `touchend` case that keeps the dialog open when another touch remains active.


flaviendelangle (approved):
**Verification summary:**

1. **The bug is real.** `useDismiss` `handleTouchEnd` (line 640–655) calls `closeOnPressOutside(event)` with the actual `touchend` event when `dismissOnTouchEnd` is set (arming happens at `distance > 5` in `handleTouchMove`). `closeOnPressOutside` invokes the dialog's `outsidePress(event)` at line 465. On a completed single-finger `touchend`, `event.touches.length === 0`, so the old guard `event.touches.length !== 1` rejected it → dialog stayed open. Confirmed.

2. **The fix is correct.** The new `touchend` branch accepts only `changedTouches.length === 1 && touches.length === 0` (true single-finger tap), and `touchstart`/`touchmove` still require `touches.length === 1`, preserving multi-touch rejection. The `>10px` immediate-dismiss path goes through a `touchmove` event (`touches.length === 1`) → still accepted via the `else if`. No regression.

3. **No path-confusion regressions.** For `modal === true`, `InternalBackdrop` renders only when `modal === true` (DialogPortal.tsx:35), so `outsidePressEvent` returns `'intentional'` → touch dismissal goes through `click` (no `touches` property) → guard block skipped entirely. For `modal={false}` and `modal="trap-focus"` there's no backdrop → `'sloppy'` → `touchend`/`touchmove` path. Test premises are accurate.

4. **No sibling copies.** The guard is unique to `useDialogRoot.ts` (shared by Dialog + Drawer); Popover has no such guard. Other `touches.length` references are unrelated handlers.

5. **Tests pin the behavior** and would fail without the fix (`waitFor` would time out). Assertion shape `mock.calls[0][1].reason` matches file conventions.

# PR review

A correct, well-scoped fix for a narrow touch-dismissal bug. I verified the central claim against the vendored `useDismiss` source: the dialog's `outsidePress` guard does receive the real `touchend` event (where `touches.length === 0`), the old guard wrongly rejected it, and the new branch accepts exactly the single-finger-tap case while preserving multi-touch rejection. Nothing is merge-blocking. One optional note on an untested branch of the new guard.

## Bugs (0)

No findings.

## Tests (1)

### 1. ℹ️ The `changedTouches.length !== 1` rejection branch is untested

**Location:** `packages/react/src/dialog/root/useDialogRoot.ts:90`

```js
if (event.type === 'touchend') {
  if (event.changedTouches.length !== 1 || event.touches.length !== 0) {
    return false;
  }
}
```

The new `touchend` guard rejects on two conditions. The added tests cover the accept case (1 ended / 0 remaining) and the `touches.length !== 0` rejection (a second finger still down). The other half — two fingers lifting simultaneously, so `changedTouches.length === 2` and `touches.length === 0` — is not exercised. It's a rare real-world gesture, so a maintainer may reasonably skip it.

**Failure scenario:** A future refactor that drops the `changedTouches.length !== 1` check (e.g. simplifying to only test `touches.length`) would let a two-finger simultaneous lift dismiss the dialog, and no test would catch the regression.

**Fix:** Optionally extend the multi-touch test with a `touchEnd` whose `changedTouches` has two entries and `touches: []`, asserting the dialog stays open — or leave as-is given how marginal the gesture is.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve** — the fix is correct, regression-tested across both backdrop-less modal variants plus a multi-touch negative case, and introduces no new public surface; the only note is an optional test-coverage nicety.

---

PR #5097, opened 2026-06-22
[slider] Fix assorted issues

atomiks:
Fixes five runtime bugs in `Slider`, each covered by a regression test (added failing first, then confirmed passing).

### Range values aren't clamped
`SliderRoot`'s scalar path clamped to `[min, max]`, but the range path only sorted. So `defaultValue={[19, 41]}` with `min={20} max={40}` rendered thumbs outside the track and produced out-of-range `aria-valuenow`. Each range item is now clamped before sorting.

### Thumb `onFocus`/`onBlur` fire on the wrapper div
The internal focus/blur/keydown handlers live on the nested `<input>`, and `onKeyDown` is forwarded there, but `onFocus`/`onBlur` were attached to the wrapper `<div>` — so `event.currentTarget` was the div, inconsistent with `onKeyDown` (both are typed `FocusEventHandler<HTMLInputElement>`). They're now forwarded to the input. The synthetic blur/focus dispatched while restoring `:focus-visible` is still suppressed so user handlers don't observe those internal events.

### `Slider.Value` keeps stale formatting
`formattedValues` was memoized on a stable ref whose `.current` only updates in a layout effect, so changing `format` (e.g. plain → currency) without changing the value/locale never recomputed. `format` is now threaded through context and the memo keys on it. This also makes `Thumb`'s `aria-valuetext` react to `format` changes, which had the same latent staleness.

### Range validation commits while focus moves between thumbs
The input `onBlur` unconditionally ran field blur logic (`validation.commit`, `setFocused(false)`) under `validationMode="onBlur"`, so tabbing from one range thumb to its sibling committed validation mid-interaction. Guarded with a shadow-safe `contains(controlRef.current, event.relatedTarget)`, mirroring `RadioGroup`.

### Range pointer handling can emit a malformed array
`getFingerState` guarded the upper bound only for non-range sliders; the range branch checked `thumbIndex < 0` only. If the pressed thumb index pointed past the end of a value array that shrank mid-drag, the collision/push helpers wrote past the array end and grew it. Added `|| thumbIndex >= values.length` to the range guard.

`pnpm test:jsdom Slider` and `pnpm test:chromium Slider` pass; `typescript`, `eslint`, and `prettier` are clean.


flaviendelangle (approved):
# PR review

This PR fixes five distinct Slider runtime bugs, each with a regression test. The changes are correct and well-targeted: range clamping, the `getFingerState` upper-bound guard, the field-blur `contains` guard, and the `format`-through-context fix all hold up under scrutiny, and the focus/blur forwarding interacts correctly with `mergeProps` (the user handlers run *before* the internal `onFocus` resets `restoringFocusVisibleRef`, so the synthetic restore-focus suppression still works). Nothing is merge-blocking. One informational note on a re-render tradeoff.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Notes

### ℹ️ Threading `format` through context defeats the context memo for inline `format` objects

**Location:** `packages/react/src/slider/root/SliderRoot.tsx:318,361`

```tsx
format,            // now a memo dependency of contextValue (and `state` consumers downstream)
```

Previously `formatOptionsRef` was a stable ref, so the slider `contextValue` memo never invalidated on `format` identity changes. Now `format` is a direct dependency. This is *necessary* for the staleness fix — but it means the idiomatic usage `<Slider.Root format={{ style: 'currency', currency: 'USD' }}>` (a fresh object each parent render) now recreates `contextValue` on every render, re-rendering `Thumb`/`Value`/`Control`. `SliderValue`'s `formattedValues` memo (`[format, locale, values]`) likewise recomputes each render.

**Failure scenario:** A slider whose parent re-renders frequently (e.g. on each drag tick elsewhere) and passes an inline `format` object now does extra reconciliation work per render that the stable-ref version avoided. Note this also diverges from `NumberField`/`Progress`, which still use the `formatOptionsRef` pattern.

**Fix:** Acceptable as-is given the correctness win; no action required. If the extra renders matter, the doc/JSDoc could advise memoizing `format`, or `format` could be excluded from the memo and read via a ref that's *also* tracked by a content-based signal — but that reintroduces complexity the PR deliberately removed.

## Verdict

**Approve** — five real bugs fixed with matching regression tests, no correctness or contract issues found; the only note is a minor, justified re-render tradeoff.

---

Issue #2942, opened 2025-10-08
[combobox][autocomplete] Input doesn't filter using Android Samsung keyboard until closed


joncoronel:
# Bug report

## Current behavior

The combobox/autocomplete components don't update the item list after entering text into the input when using the samsung keyboard on mobile android. The entered text shows up in the input box like normal so the keyboard is working but the the list does not filter until after I close the keyboard. It's as if the internal state for input does not update when entering text using the samsung keyboard. But it works fine with google's gboard. I then tested the samsung keyboard on combobox/autocomplete components from other ui libraries and it seemed to work fine. Not sure what's causing the issue. 

## Expected behavior

As I type text into the input box, the list of items should filter based on the text entered. 

## Reproducible example

This happens will all examples in the base ui combobox/autocomplete docs. 

![Screenshot_20251008_141843_Chrome.jpg](https://github.com/user-attachments/assets/57618fe3-9c85-4d65-8a21-44f97ca170ab)

## Base UI version

1.0.0-beta.4

## Which browser are you using?

Android chrome/firefox

## Which OS are you using?

Android

atomiks:
Most likely this is due to `Combobox.Input`'s composition/IME handling where it won't start filtering until a composed character has been committed. This prevents the empty state from appearing while typing before the full character has been committed. But it's preventing it from filtering at all on this keyboard incorrectly

mj12albert:
I can reproduce this in Browserstack, turning off "predictive text" (on by default) in the keyboard settings seems to fix it

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/c8781567-18b6-4b61-bae8-a86fa39cc644" />

https://github.com/user-attachments/assets/a2566dab-daab-4a5a-98f1-cea1328f001a

joncoronel:
Yep just tried it and that seems to be a way to fix it. Probably still a good idea to have a fix for it still since this is the default keyboard configuration for all samsung phones. It seems like theres already a pr for the fix. Very quick.
> I can reproduce this in Browserstack, turning off "predictive text" (on by default) in the keyboard settings seems to fix it

mj12albert:
> This prevents the empty state from appearing while typing before the full character has been committed

Yup, this occurs with the RA combobox for example:

https://github.com/user-attachments/assets/4630a533-2dbb-44d1-8705-01a17867245a

PR #5098, opened 2026-06-22
[otp field] Fix Safari IME composition over-filling slots

### LINKED ISSUE #5091: [otp field] Safari IME input duplicates alphanumeric characters
# Bug report

## Current behavior

When `OTPField.Root` uses `validationType="alphanumeric"`, Safari can duplicate characters while an IME/composition input is active.

With a 6-character OTP field, typing `d` three times in Safari while the input method is active fills all six slots with `dddddd`:

- first `d` writes one `d`
- second `d` writes two more `d` characters
- third `d` writes three more `d` characters

Chrome does not reproduce this in the same scenario.

## Expected behavior

Each committed character should be written once. Typing `d` three times should produce `ddd`, not `dddddd`.

## Reproducible example

Use an OTP Field with `length={6}` and `validationType="alphanumeric"`, for example the alphanumeric OTP Field demo on the docs page:
https://base-ui.com/react/components/otp-field#alphanumeric-verification-codes

## Base UI version

v1.6.0 / current `main` source (`70524bdf5`)

Safari and Safari iOS.

## Which OS are you using?

macOS and iOS.

N/A

## Additional context

This appears to happen because Safari exposes the composing input value as an accumulating string. `OTPField.Input` currently handles `onChange` by reading the whole `event.currentTarget.value` and treating it as a multi-character OTP input, then advances focus by that length. That is correct for paste/bulk input, but it misinterprets Safari IME composition updates as committed OTP text.

atomiks:
## Summary

`OTPField.Input` committed every IME `onChange` value during composition. In Safari, the in-progress text can arrive as accumulating values (`d`, `dd`, `ddd`), so the field wrote each intermediate value into the slots and could end up with `dddddd` instead of the final `ddd`. Chrome does not reproduce the same over-fill in the reported flow.

## Changes

This makes OTP input composition-aware, following the Combobox pattern:

- Track active composition and keep the temporary text in local input state.
- During composition, let `onChange` update only that temporary text.
- On `compositionend`, commit the final value once through the normal normalization/replacement path.
- Reuse that commit path for normal input and composition end, preserving paste, autofill, invalid-character reporting, cancellation, and focus movement.
- Keep the Android guard used by Combobox for inconsistent keyboard composition events.

Fixes https://github.com/mui/base-ui/issues/5091

### DIFF
--- a/packages/react/src/otp-field/input/OTPFieldInput.android.test.tsx
+++ b/packages/react/src/otp-field/input/OTPFieldInput.android.test.tsx
@@ -1,0 +1,52 @@
+import { expect, vi } from 'vitest';
+import * as React from 'react';
+import { act, fireEvent, screen } from '@mui/internal-test-utils';
+import { OTPField } from '@base-ui/react/otp-field';
+import { createRenderer } from '#test-utils';
+
+vi.mock('@base-ui/utils/platform', async () => {
+  const actual =
+    await vi.importActual<typeof import('@base-ui/utils/platform')>('@base-ui/utils/platform');
+
+  return {
+    ...actual,
+    platform: {
+      ...actual.platform,
+      os: { ...actual.platform.os, android: true },
+    },
+  };
+});
+
+describe('<OTPField.Input /> Android', () => {
+  const { render } = createRenderer();
+
+  it('commits each change during an IME composition', async () => {
+    const onValueChange = vi.fn();
+
+    await render(
+      <OTPField.Root length={3} validationType="alphanumeric" onValueChange={onValueChange}>
+        <OTPField.Input />
+        <OTPField.Input />
+        <OTPField.Input />
+      </OTPField.Root>,
+    );
+
+    const inputs = screen.getAllByRole<HTMLInputElement>('textbox');
+
+    await act(async () => {
+      inputs[0].focus();
+    });
+
+    fireEvent.compositionStart(inputs[0]);
+    fireEvent.change(inputs[0], { target: { value: 'a' } });
+
+    expect(onValueChange).toHaveBeenCalledTimes(1);
+    expect(onValueChange).toHaveBeenLastCalledWith('a', expect.anything());
+    expect(document.activeElement).toBe(inputs[1]);
+
+    fireEvent.compositionEnd(inputs[0]);
+
+    expect(onValueChange).toHaveBeenCalledTimes(1);
+    expect(inputs.map((input) => input.value)).toEqual(['a', '', '']);
+  });
+});
--- a/packages/react/src/otp-field/input/OTPFieldInput.test.tsx
+++ b/packages/react/src/otp-field/input/OTPFieldInput.test.tsx
@@ -190,6 +190,38 @@ describe('<OTPField.Input />', () => {
     expect(firstInput.selectionEnd).toBe(1);
   });
 
+  it('commits an IME composition once on compositionend instead of per intermediate change', async () => {
+    const onValueChange = vi.fn();
+
+    await render(<OTPFieldTest validationType="alphanumeric" onValueChange={onValueChange} />);
+
+    const inputs = screen.getAllByRole<HTMLInputElement>('textbox');
+    const firstInput = inputs[0];
+
+    await act(async () => {
+      firstInput.focus();
+    });
+
+    fireEvent.compositionStart(firstInput);
+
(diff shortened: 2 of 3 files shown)


mj12albert (approved):
The fix works for Safari and matches the behavior in Chrome now:  

https://github.com/user-attachments/assets/f4ac5565-5430-4627-b12f-62527a2c86c9

Though the "correct" behavior still feels very strange to use, if I had to type the code it would be just as fast (if not faster) to switch the input method, type the code, and switch it back

flaviendelangle (commented):
# PR review

This is a clean, well-targeted fix for the Safari IME over-fill (#5091). It mirrors the established Combobox composition pattern, the extracted `commitInputValue` is a faithful move of the original `onChange` body, and the 62 existing + 1 new test pass. Nothing is merge-blocking. The notes below are minor test/comment observations only.

## Bugs (0)

No findings. I specifically checked the highest-risk paths and they hold:

- **Double-commit on `compositionend` then trailing `change`** (old WebKit/Chrome ordering): the trailing `onChange` re-enters `commitInputValue` with `composingValue == null`, but it carries the same final value, and `setValue` dedupes it — no extra `onValueChange`, focus stays correct.
- **State-vs-ref for the composing branch**: unlike Combobox, this uses only `composingValue` state (no `isComposingRef`) to branch in `onChange`. This is safe because React classifies composition events as discrete and flushes the `onCompositionStart` update synchronously before the next `input`/`change`, so the first intermediate value is buffered, not committed.
- **Android path**: `onCompositionStart` returns early, leaving `composingValue` null, so `onChange` commits as before and `onCompositionEnd` no-ops — no regression versus current behavior.
- **`composingValue === ''`**: a composition that starts and ends with no `change` still commits, because `'' != null` keeps the composition treated as active (`?? ` only falls through on null/undefined).

## Tests (2)

### 1. ℹ️ Android composition branch is untested

**Location:** `packages/react/src/otp-field/input/OTPFieldInput.tsx:194` (PR version)

```tsx
onCompositionStart(event) {
  if (platform.os.android) {
    return;
  }
  setComposingValue(event.currentTarget.value);
},
```

The new `platform.os.android` early-return is a distinct branch with no coverage. The added test exercises only the non-Android buffering path.

**Failure scenario:** A future refactor could break the Android passthrough (where each `onChange` must still commit immediately) without any test failing. This is the same constraint Combobox depends on (#2942).

**Fix:** If practical to mock the `@base-ui/utils/platform` module, add a case asserting that with `platform.os.android === true`, intermediate `change` events commit immediately (matching pre-PR behavior). Acceptable to skip if the platform module can't be mocked cleanly, as Combobox does.

### 2. ℹ️ No test for invalid characters surfaced during composition commit

**Location:** `packages/react/src/otp-field/input/OTPFieldInput.tsx` (PR version, `commitInputValue` → `reportValueInvalid`)

The composition commit routes through the same `didRejectCharacters` / `reportValueInvalid` path as normal input, but the new test only commits valid alphanumerics.

**Failure scenario:** A regression where invalid-character reporting is dropped specifically on the composition-commit path would go unnoticed.

**Fix:** Add a composition test where `compositionend` carries a value containing a rejected character and assert `onValueChange`/invalid reporting behaves as it does for direct input.

## Simplifications (0)

No findings. Extracting the shared body into `commitInputValue` is itself the right de-duplication, and it's consistent with the file's existing inline-handler style (no `useStableCallback` needed since the handlers are recreated per render).

## Docs (1)

### 1. ℹ️ Comment claims Gecko/Firefox reproduces, but the issue is Safari-only

**Location:** `packages/react/src/otp-field/input/OTPFieldInput.tsx:75` (PR version)

```tsx
// While an IME composition is active, Safari (and other WebKit/Gecko engines) can expose the
// in-progress text through `onChange` as an accumulating string (`d`, then `dd`, then `ddd`).
```

The linked issue reports the over-fill in Safari/iOS only and explicitly notes Chrome does not reproduce it; there's no evidence cited for Firefox (Gecko). The fix is correctly defensive regardless, but the comment slightly overclaims the reproduction surface.

**Fix:** Narrow to "Safari/WebKit" or add "(Chrome does not reproduce the reported flow)" to match the issue, so the comment doesn't mislead a future reader about which engines were actually observed.

## Verdict

**Approve** — correct, well-scoped fix that matches the Combobox composition pattern and passes its tests; only optional test-coverage and comment-accuracy notes remain.

---

PR #5101, opened 2026-06-22
[code-infra] Enable mui/no-floating-cleanup ESLint rule

romgrk:
## Summary

Enables the type-aware **`mui/no-floating-cleanup`** rule (from [mui/mui-public#1538](https://github.com/mui/mui-public/pull/1538), already shipped in the installed `@mui/internal-code-infra` canary) for `packages/*/src`, and turns on typed linting for that scope so the rule can resolve type information.

The rule is like `@typescript-eslint/no-floating-promises`, but for functions: it flags call statements whose return value is a discarded callable (typically an `unsubscribe` / cleanup callback). Opt out per call site with `void`.

## What the rule caught

7 sites in `packages/*/src`. Reviewing each, most are intentional discards opted out with `void` + a one-line rationale — but **two were genuine leaks**, now fixed:

| Site | Discarded return | Disposition |
|---|---|---|
| `floating-ui-react/.../FloatingFocusManager.tsx` (×3) | `enqueueFocus(...)` | `void` — returns a rAF-cancel function we don't need |
| `internals/useRenderElement.tsx` | `useMergedRefs(null, null)` | `void` — hook called only to keep hook order stable when disabled |
| `utils/useMergedRefs.ts` | `ref(null)` | `void` — legacy detach for a ref with no attach-time cleanup; React 19 cleanups run in the sibling branch |
| `context-menu/.../ContextMenuTrigger.tsx` | `addEventListener(..., { once: true })` remover | **real fix** (see below) |
| `radio-group/RadioGroup.tsx` | `setInputRef(fallbackInput)` | **real fix** (see below) |

## ContextMenuTrigger fix

The long-press / right-press flow registers a one-shot `mouseup` listener on the document from inside the `onContextMenu` handler and discarded the remover. `{ once: true }` only removes the listener *after* it fires, so if the trigger unmounts before the release — or the mouseup never arrives (pointer released outside the window, keyboard-driven contextmenu) — the listener lingers on the document and later runs its handler against a torn-down trigger. Now the remover is retained in a ref, cleared before re-registering, and torn down on unmount (matching the sibling document `contextmenu` listener). Regression test added.

## RadioGroup fix

The group's `inputRef` is re-pointed from three places — a radio input mounting (`registerInputRef`, a real ref callback), the selection changing (`RadioRoot` effect), and the value-cleared fallback in `useValueChanged`. A function `inputRef` can return a cleanup (React 19 ref-callback semantics), but only the mount path forwarded it to React; the other two discarded the cleanup and never ran the previous attachment's cleanup before re-pointing, leaking it. (The rule only flagged one of the three because the context types `registerInputRef` as returning `void`.) Fix: give `setInputRef` sole ownership of the cleanup lifecycle — run the prior cleanup before each re-point, store the new one, and detach on unmount. Regression test added.

## Test plan

`pnpm typescript`, `pnpm eslint` (incl. the new rule over `packages/*/src`), and `pnpm prettier` are clean. `pnpm test:jsdom Radio` (158 tests) and `pnpm test:jsdom ContextMenu` (incl. the two new regression tests) pass.

### DIFF
--- a/packages/react/src/context-menu/trigger/ContextMenuTrigger.tsx
+++ b/packages/react/src/context-menu/trigger/ContextMenuTrigger.tsx
@@ -47,7 +47,7 @@ export const ContextMenuTrigger = React.forwardRef(function ContextMenuTrigger(
   const longPressTimeout = useTimeout();
   const allowMouseUpTimeout = useTimeout();
   const allowMouseUpRef = React.useRef(false);
-  const removeMouseUpListenerRef = React.useRef<(() => void) | null>(null);
+  const mouseUpAbortControllerRef = React.useRef<AbortController | null>(null);
 
   function handleLongPress(x: number, y: number, event: MouseEvent | TouchEvent) {
     const isTouchEvent = event.type.startsWith('touch');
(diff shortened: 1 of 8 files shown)


packages/react/src/context-menu/trigger/ContextMenuTrigger.tsx
```diff
@@ -81,10 +82,16 @@ export const ContextMenuTrigger = React.forwardRef(function ContextMenuTrigger(
...
+    // Tear down a listener from a previous trigger that never saw its mouseup, then
+    // retain this one so it can be removed on unmount if the mouseup never arrives.
+    removeMouseUpListenerRef.current?.();
+    removeMouseUpListenerRef.current = addEventListener(
       doc,
       'mouseup',
       (mouseEvent) => {
+        // `{ once: true }` has already removed this listener.
+        removeMouseUpListenerRef.current = null;
+
```
romgrk: Using event listeners outside of react properly requires a lot of boilerplate, we should encapsulate this logic in something more readable/usable.

michaldudak (approved):

PR #5104, opened 2026-06-23
[all components] Type render callback props to the rendered element

### LINKED ISSUE #5102: Avatar.Image's render prop has the wrong prop types
# Bug report

## Current behavior

When trying to use the `render` prop on `<Avatar.Image />` with a function the `props` type is missing standard `` tag prop types like `src`.

For instance if you wanted to use next/image through the render function like so you will get the TypeScript errors "Property 'src' does not exist on type" and "Property 'alt' does not exist on type".
```
function NextAvatarImage({ ...props }: Avatar.Image.Props) {
  return (
    <Avatar.Image
      {...props}
      render={({ ref, placeholder, ...rest }) => {
        return <Image ref={ref} {...rest} src={rest.src} alt={rest.alt} />;
      }}
    />
  );
}
```

## Expected behavior

I would expect the `render` function's `props` to inherit from `React.ComponentProps<'img'>` which contains the `src` and `alt` props.

I don't know too much about Base UI's types but looking at how `BaseUIComponentProps` is defined it looks like this might just be an issue of `AvatarImageProps` not providing the 3rd generic argument to it.

## Reproducible example

https://stackblitz.com/edit/stackblitz-starters-dqfq2juz?file=app%2Fpage.tsx

## Base UI version

v1.6.0

atomiks:

Fixes #5102

The third generic of `BaseUIComponentProps` types the `render` callback's props. It defaults to `HTMLProps` so the callback can spread onto any element. Specializing it to `ComponentPropsWithRef<'tag'>` exposes element-specific props (`src`, `href`, `disabled`, …) but blocks rendering as a different element. So we specialize only parts locked to one element, and keep `HTMLProps` for parts that can legitimately render as something else.

### Specialized

| Part | Element |
|---|---|
| `Avatar.Image` | `` |
| `Form` | `<form>` |
| `NumberField.Input` | `<input>` |
| `OTPField.Input` | `<input>` |
| `PreviewCard.Trigger` | `<a>` |
| `Menu.LinkItem` | `<a>` |
| `NavigationMenu.Link` | `<a>` |
| `Toolbar.Link` | `<a>` |

### Kept loose (`HTMLProps`)

| Part | Reason |
|---|---|
| `Combobox.Input` / `Autocomplete.Input` | can render as `<textarea>` |
| `Field.Control`, `Toolbar.Input` | generic input, can be `<textarea>` |
| `Field.Label` | `nativeLabel={false}` supports `<div>` |
| `Fieldset.Root` | can render group controls such as `RadioGroup`, `CheckboxGroup`, or `Slider.Root` |
| `Slider.Value` | often a `<span>`/`<div>`; shouldn't force `htmlFor` |
| `NavigationMenu.List` | `<ul>` has no element-specific props |
| `*.Button` parts | polymorphic via `nativeButton` |

### Also

- `Combobox.Input` / `Autocomplete.Input`: dropped the injected `type="text"` (native default already; was invalid when rendered as a `<textarea>`).
- Added type-level `*.spec.tsx` checks for the specialized callbacks.


flaviendelangle (approved):
# PR review

This PR is type-only except for one safe runtime change (dropping the injected `type="text"` on `Combobox.Input`/`Autocomplete.Input`). The native-input-vs-other-element ARIA branch keys off `tagName === 'INPUT'` (`AriaCombobox.tsx:992`), not the `type` attribute, so the removal doesn't disturb it, and the new tests pin both the editability and the open-only ARIA behavior. The render-callback specialization only propagates each component's already-declared element type (`'img'`, `'form'`, `'input'`, `'a'`) to the callback param, so it can't tighten anything beyond what the component's own props already required. Nothing is merge-blocking.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (2)

### 1. ℹ️ Auto-generated `render` type is now noisy in public API tables

**Location:** `docs/src/app/(docs)/react/components/form/types.md:11` (and the avatar/number-field/otp-field/preview-card/menu/navigation-menu/toolbar tables)

```md
| render | `ReactElement \| ((props: React.DetailedHTMLProps<React.FormHTMLAttributes<HTMLFormElement>, HTMLFormElement>, state: Form.State) => ReactElement)` | ... |
```

The specialized third generic causes `pnpm docs:api` to emit the fully-expanded `React.DetailedHTMLProps<React.FormHTMLAttributes<HTMLFormElement>, HTMLFormElement>` in place of the prior clean `HTMLProps`. It's accurate but considerably harder to read in the published prop table.

**Failure scenario:** A user scanning the Form/Avatar/Input prop tables sees a long leaky type for `render` instead of a concise one; purely a readability regression in the generated docs.

**Fix:** Optional — if the docs generator supports a type alias/display override, map these back to a short name; otherwise accept it as the cost of the more precise type.

### 2. ℹ️ `Combobox.Input` no longer emits a `type="text"` attribute

**Location:** `packages/react/src/combobox/input/ComboboxInput.tsx:205`

```tsx
// removed:
type: 'text',
```

Behaviorally inert for a native `<input>` (text is the default), and the PR documents it as intentional. The only observable effect is that the rendered DOM element no longer carries `type="text"`, so an app styling the input via an `input[type="text"]` selector would stop matching. The in-repo demos that rely on `type="text"` pass it explicitly (`autocomplete/demos/grid/...`), so they're unaffected.

**Failure scenario:** A consumer with CSS like `.MyCombobox input[type="text"] { ... }` loses those styles after upgrading.

**Fix:** None needed; consider a one-line changelog/migration note since it's a visible DOM-attribute change.

## Verdict

**Approve** — type-only specialization plus a safe, well-tested `type="text"` removal; no correctness or test gaps, only minor cosmetic notes on generated docs.

---

PR #5110, opened 2026-06-24
[popups] Fix rendered trigger id ownership

### LINKED ISSUE #5108: 1.6.0 regression: tooltip doesn't appear if button has ID
# Bug report

## Current behavior

```tsx
<Tooltip.Trigger
  // ❌
  render={<button id="foo" />}
  // ✅
  // render={<button />}
>
```

## Expected behavior

It works regardless of `id`.

## Reproducible example

1.6.0 ❌: https://stackblitz.com/edit/wvguwzda?file=src%2FApp.tsx
1.5.0 ✅: https://stackblitz.com/edit/wvguwzda-sia2cyk3?file=src%2FApp.tsx

## Base UI version

1.6.0

Chrome

## Which OS are you using?

N/A

Voiceover/JAWS etc.

## Additional context

atomiks:
Fixes #5108

A trigger rendered with an element that already has its own `id` (e.g. `render={<button id="custom-button" />}`) detaches the trigger's DOM `id` from the internal id Base UI registers it under. The active trigger is tracked by DOM `id` (`setPopupOpenState` reads `trigger.id`), while the registry and ownership selectors key off the internal id, so the trigger is never recognized as active. This affects every popup family that shares `useImplicitActiveTrigger`, at two severity levels:

### 🔴 Closes immediately after opening — Tooltip, Preview Card

Both enable `closeOnActiveTriggerUnmount`, so the failed registry lookup is treated as the active trigger unmounting and the popup closes right after it opens. This is the #5108 regression introduced in #4886 (`[tooltip][preview card]`) — 1.6.0 fails, 1.5.0 works.

### 🟡 Opens, but trigger ownership / ARIA is wrong — Popover, Menu, Dialog

No close-on-unmount, so the popup stays open, but the trigger never registers as active: `data-popup-open` is missing and `aria-expanded` / `aria-controls` are stale. This one is pre-existing (independent of #4886), surfaced by the same id mismatch.

> Passing the id directly to the trigger (`<Popover.Trigger id="custom-button">`) instead of to the rendered element sidesteps all of this and already works on 1.6.0.

## Root cause

`setPopupOpenState` records `activeTriggerId` from the trigger's DOM `id`. When a `render` id overrides Base UI's internal id, that value is not in the trigger registry, so `getById` misses and the ownership selectors (`isOpenedByTrigger`, etc.) never match the trigger.

## Fix

In `useImplicitActiveTrigger`, when the active id misses a direct registry lookup, scan for the same element registered under a different id and reassociate to the registered id instead of treating it as lost. The reconciliation is also made reactive to active-trigger-id changes so it covers trigger-to-trigger handoffs while the popup stays open.

## Perf

The scan only runs after the fast `getById` lookup misses, and it is self-healing (subsequent lookups hit the fast path). Measured across the multi-trigger suite: **0 scans in normal flows** (108 effect runs → 5 scans total, all in the rare custom-id / lost-trigger paths), ~1.4 iterations per scan, bounded by the number of registered triggers.

## Changes

- Reassociate stale active trigger ids with the registered element before closing; rerun the reconciliation on active-trigger handoffs.
- Add Tooltip store + component coverage for rendered trigger ids and handoffs.

### DIFF
--- a/packages/react/src/utils/popups/popupStoreUtils.ts
+++ b/packages/react/src/utils/popups/popupStoreUtils.ts
@@ -351,7 +351,7 @@ export function useImplicitActiveTrigger<State extends PopupStoreState<unknown>>
   const reactiveTriggerCount = store.useState('triggerCount');
   // Subscribe to the active trigger id so the reconciliation below reruns when ownership moves to
   // another trigger while the popup stays open (e.g. a focus/hover handoff between triggers).
-  const reactiveActiveTriggerId = store.useState('activeTriggerId');
+  const activeTriggerId = store.useState('activeTriggerId');
 
   useIsoLayoutEffect(() => {
     if (!open) {
@@ -368,11 +368,11 @@ export function useImplicitActiveTrigger<State extends PopupStoreState<unknown>>
       stateUpdates.triggerCount = triggerCount;
     }
 
-    const activeTriggerId = store.select('activeTriggerId');
+    const currentActiveTriggerId = store.select('activeTriggerId');
     let lostActiveTriggerId: string | null = null;
 
-    if (activeTriggerId) {
-      const activeTriggerElement = store.context.triggerElements.getById(activeTriggerId);
+    if (currentActiveTriggerId) {
+      const activeTriggerElement = store.context.triggerElements.getById(currentActiveTriggerId);
       if (!activeTriggerElement) {
         for (const [triggerId, triggerElement] of store.context.triggerElements.entries()) {
           if (triggerElement === store.state.activeTriggerElement) {
@@ -383,14 +383,14 @@ export function useImplicitActiveTrigger<State extends PopupStoreState<unknown>>
         }
 
         if (stateUpdates.activeTriggerId === undefined) {
-          lostActiveTriggerId = activeTriggerId;
+          lostActiveTriggerId = currentActiveTriggerId;
         }
       } else if (activeTriggerElement !== store.state.activeTriggerElement) {
         stateUpdates.activeTriggerElement = activeTriggerElement;
       }
     }
 
-    if (!lostActiveTriggerId && !activeTriggerId && triggerCount === 1) {
+    if (!lostActiveTriggerId && !currentActiveTriggerId && triggerCount === 1) {
       const iteratorResult = store.context.triggerElements.entries().next();
       if (!iteratorResult.done) {
         const [implicitTriggerId, implicitTriggerElement] = iteratorResult.value;
@@ -430,7 +430,7 @@ export function useImplicitActiveTrigger<State extends PopupStoreState<unknown>>
         });
       }
     }
-  }, [open, store, reactiveTriggerCount, reactiveActiveTriggerId, closeOnActiveTriggerUnmount]);
+  }, [open, store, reactiveTriggerCount, activeTriggerId, closeOnActiveTriggerUnmount]);
 }
 
 /**
--- a/packages/react/src/utils/popups/popupStoreUtils.ts
+++ b/packages/react/src/utils/popups/popupStoreUtils.ts
@@ -325,9 +325,12 @@ export type PayloadChildRenderFunction<Payload> = (arg: {
  *
  * When a popup opens without an explicit trigger id and exactly one trigger is registered, that
  * trigger is claimed as the active trigger. When the active trigger id is still registered but its
- * element changed, the active element is refreshed. When the active trigger unregisters, the
- * default path preserves existing ownership so non-closing popup families do not silently claim a
- * different trigger while staying open.
+ * element changed, the active element is refreshed. When the active trigger id is missing from the
+ * registry but the same element is still registered under a different id (e.g. the rendered trigger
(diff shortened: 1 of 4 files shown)


flaviendelangle (commented):
# PR review

This is a tight, well-targeted fix for #5108 (tooltip triggers rendered with a custom `id` closing immediately after opening). The reassociation logic in `useImplicitActiveTrigger` is correct, self-limiting (slow-path only, runs once per open), and properly regression-tested at both the store and component level. Nothing here is merge-blocking. Two informational notes below: a one-time viewport content remount that this fix newly introduces for a narrow combination, and a stale function-level comment.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (1)

### 1. ℹ️ note — function comment doesn't mention the new reassociation path

**Location:** `packages/react/src/utils/popups/popupStoreUtils.ts:323-339`

```ts
 * When a popup opens without an explicit trigger id and exactly one trigger is registered, that
 * trigger is claimed as the active trigger. When the active trigger id is still registered but its
 * element changed, the active element is refreshed. When the active trigger unregisters, the
 * default path preserves existing ownership so non-closing popup families do not silently claim a
 * different trigger while staying open.
```

The JSDoc enumerates the three existing reconciliation paths but not the new one this PR adds: when the active trigger id is *not* found by direct lookup, the registry is scanned for the same element under a different id and the active id is reassociated (the whole point of the fix). The non-obvious branch — a DOM `id` differing from Base UI's internal trigger id — is exactly the kind of thing the next maintainer will want explained here.

**Failure scenario:** A future reader sees the `getById` miss fall through to a loop with no comment explaining why an element can be registered under an id other than `activeTriggerId`, and either misreads it as dead code or breaks it during refactor.

**Fix:** Add a sentence, e.g. "When the active trigger id is missing from the registry but the same element is registered under a different id (e.g. the rendered element carries its own DOM `id`), the active id is reassociated to the registered id instead of being treated as lost."

## Verdict

**Approve** — correct, minimal fix with solid regression coverage; only an optional comment update and one narrow heads-up below.

---

A heads-up worth recording even though it doesn't rise to a finding: the reassociation flips `activeTriggerId` from the DOM id to the internal registered id once, just after open. For consumers of `<X.Viewport>` (`usePopupContentKey` keys its remount on `activeTriggerId`), a trigger rendered with a custom `id` will now produce a single content remount right after opening. For Tooltip/PreviewCard (`closeOnActiveTriggerUnmount: true`) this is strictly better than the prior behavior (full close). For Popover/Menu it's a new one-time remount in exchange for correctly marking the trigger active (`isOpenedByTrigger` / `data-popup-open` were previously wrong in this case). It's an acceptable trade, but if the remount ever matters, keying `usePopupContentKey` on `activeTriggerElement` identity instead of `activeTriggerId` would be stable across the reassociation.

---

packages/react/src/utils/popups/popupStoreUtils.ts
```diff
@@ -346,6 +349,9 @@ export function useImplicitActiveTrigger<State extends PopupStoreState<unknown>>
   const { closeOnActiveTriggerUnmount = false } = options;
   const open = store.useState('open');
   const reactiveTriggerCount = store.useState('triggerCount');
+  // Subscribe to the active trigger id so the reconciliation below reruns when ownership moves to
+  // another trigger while the popup stays open (e.g. a focus/hover handoff between triggers).
+  const reactiveActiveTriggerId = store.useState('activeTriggerId');
```
michaldudak: nit: I'm not fan of this name. Could we call it just `activeTriggerId` (and rename what's currently `activeTriggerId` in effects to `currentActiveTriggerId` - there are some places that already use this convention)

michaldudak (approved):

ciampo:
@atomiks and @michaldudak , do you intend to release a point release that includes this fix? Or will it be part of `1.7.0`?

atomiks:
@ciampo whether a patch release goes out before 1.7.0 depends on how often this is actually hit — I'm not sure how common putting an id on the element passed to the `render` prop of a tooltip trigger is (my guess for the most commonly affected case would be form inputs with a `<label>` `htmlFor` link).

The workaround for the time being is simple at least: move the `id` prop to the `<Tooltip.Trigger>` component itself.

ciampo:
@atomiks sounds good, thank you for the reply

Kosai106:
@atomiks this issue is hurting users, why would a patch release not be the best solution here?

atomiks:
@Kosai106  for context we limit stable releases to once per month and reserve patch releases for widespread regressions without a practical workaround. So far, we’ve received relatively few reports of this issue, and there is a straightforward workaround.

If the workaround isn’t viable (placing the `id` on the component directly, instead of the `render`ed element), we publish [a `pkg.pr.new` canary build for every commit](https://base-ui.com/react/overview/releases#canary-releases), so the fix can be used immediately when you need it, without needing to wait for a release.

The next stable release is scheduled for next week as well, so the fix will be live shortly for everyone. (This release came a little later than usual because we’ve shifted our monthly cadence to the first week of each month)

PR #5112, opened 2026-06-24
[drawer] Fix popup flashing fully open for a frame on swipe re-grab

atomiks:
## Summary

Fixes a one-frame glitch where the Drawer popup flashes to its fully-open position when a swipe gesture is interrupted and quickly re-grabbed. Two independent code paths caused it.

### SwipeArea re-grab during close

`Drawer.SwipeArea` writes the popup's `--drawer-swipe-movement-*` imperatively, but the viewport also resets them to `0px` (on render via `getDragStyles`, and post-paint via the open-reset effect's `resetSwipe`) whenever it isn't swiping — clobbering the swipe area's values for a frame. A shared `swipeAreaActiveRef` lets the swipe area re-assert its movement in a layout effect and the viewport skip `resetSwipe` while the swipe area drives the open.

### Popup re-grab

This one is mostly impossible to encounter in real usage

Regression from #4867, which moved the per-move drag transform onto imperative inline writes (`syncDragStyles`). The render-path writer (`getDragStyles`) still gated the transform on the lagging `isSwiping` **state**, so a render committing before `setSwiping(true)` flushed would strip the transform `syncDragStyles` had just written, dropping the popup to its resting position for a frame. It now reads `isSwipingRef.current` to match the imperative writer.

### DIFF
--- a/packages/react/src/drawer/root/DrawerRoot.tsx
+++ b/packages/react/src/drawer/root/DrawerRoot.tsx
@@ -76,6 +76,7 @@ export function DrawerRoot<Payload = unknown>(props: DrawerRoot.Props<Payload>) 
   });
 
   const isNestedDrawerOpenRef = React.useRef(false);
+  const swipeAreaActiveRef = React.useRef(false);
 
   const setActiveSnapPoint = useStableCallback(
     (
@@ -172,6 +173,7 @@ export function DrawerRoot<Payload = unknown>(props: DrawerRoot.Props<Payload>) 
   const contextValue: DrawerRootContext = React.useMemo(
     () => ({
       swipeDirection,
+      swipeAreaActiveRef,
       snapToSequentialPoints,
       snapPoints,
       activeSnapPoint: resolvedActiveSnapPoint,
@@ -210,6 +212,7 @@ export function DrawerRoot<Payload = unknown>(props: DrawerRoot.Props<Payload>) 
       setActiveSnapPoint,
       snapPoints,
       snapToSequentialPoints,
+      swipeAreaActiveRef,
       swipeDirection,
     ],
   );
--- a/packages/react/src/drawer/root/DrawerRootContext.ts
+++ b/packages/react/src/drawer/root/DrawerRootContext.ts
@@ -14,6 +14,11 @@ export interface DrawerNestedSwipeProgressStore {
 export interface DrawerRootContext {
   swipeDirection: DrawerSwipeDirection;
   /**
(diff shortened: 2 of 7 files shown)


flaviendelangle (approved):
# PR review

A focused, well-scoped fix for a one-frame flash where the Drawer popup snaps to its fully-open position when a swipe gesture is re-grabbed. The two mechanisms (SwipeArea re-grab via the shared `swipeAreaActiveRef`, and the `getDragStyles` ref-vs-state divergence) are both real and correctly addressed. Nothing is merge-blocking. I verified both new tests pass in Chromium, confirmed the `useSwipeDismiss` test genuinely fails when the `isSwipingRef` read is reverted, and confirmed ESLint passes on all changed files.

## Bugs (0)

No findings.

I traced the lifetime of `swipeAreaActiveRef` for the regression I was most worried about — the viewport skipping `resetSwipe()` leaving stale viewport swipe state. The ref is true only *during* an active swipe-area gesture (set in `applySwipeMovement`, cleared in `clearSwipeStyles`, which runs on both `finishSwipeInteraction` and the `!enabled` effect), and the viewport's open-reset effect reads it at the instant `open` flips true — exactly when the ref is legitimately true. After release the ref is already false, so a subsequent non-swipe open still runs `resetSwipe()` normally. No stuck-true path leaks.

One residual edge (benign, not actionable): in `DrawerPopup`, `getDragStyles()` now reads `isSwipingRef` while the adjacent snap-point branch (`if (swiping && …)`) still reads the lagging `swiping` *state*. On a snap-points-down drawer during a lagging frame, the snap-point offset is computed from the raw freeze movement rather than `getSnapPointSwipeMovement`. The transform is stripped by design in both branches and the movement var is preserved, so there's no visible flash — but a maintainer extending the snap-point path may want the two reads to agree.

## Tests (0)

No findings.

Both added tests are strong regression guards. The `useSwipeDismiss` test captures `getDragStyles` from a render where `swiping` state is false and asserts the closure still emits the freeze transform — I confirmed it fails (`transform` undefined) when the fix is reverted. The SwipeArea test exercises a real exit-animation re-grab and pins the exact `--drawer-swipe-movement-y: 120px` value (200px popup − 80px displacement).

## Simplifications (0)

No findings.

## Docs (0)

No findings. The new `swipeAreaActiveRef` JSDoc and the inline comments on the viewport effect, the `useIsoLayoutEffect`, and the `getDragStyles` ref read accurately describe the mechanism.

## Verdict

**Approve** — correct, minimal, and well-covered fix; no blocking or non-blocking issues survived verification (tests pass, fix proven necessary by reverting it).

---

PR #5114, opened 2026-06-24
[password field] Create new PasswordField component

atomiks:
Adds a new `PasswordField` component (in preview) — a headless password input with a button to reveal or hide the value.

Preview: https://deploy-preview-5114--base-ui.netlify.app/react/components/password-field


flaviendelangle (commented):
# PR review

A clean, well-tested new `PasswordToggleField` (preview). Nothing is merge-blocking — no 🔴. The design closely tracks Radix's `unstable_PasswordToggleField` (same name, parts, props, form-reset semantics, and autofill defaults), which is a good sign, but the port **drops Radix's explicit caret save/restore** in favor of relying on the browser, and that path is only verified in Chromium. The other actionable items are test gaps (controlled-mode click, `nativeButton={false}`) and a labeling inconsistency between the hero demo and the docs.

## How it compares to competitors

| | Base UI (this PR) | Radix `unstable_PasswordToggleField` | Ark UI `PasswordInput` | React Aria / Headless UI / AriaKit |
|---|---|---|---|---|
| Dedicated component | ✅ Root / Input / Toggle | ✅ Root / Input / Toggle / Slot / Icon | ✅ Root / Label / Control / Input / VisibilityTrigger / Indicator | ❌ none — hand-rolled |
| Visibility props | `visible` / `defaultVisible` / `onVisibleChange` | `visible` / `defaultVisible` / `onVisibilityChange` | `visible` / `defaultVisible` / `onVisibilityChange` | — |
| Toggle a11y | `aria-pressed` + `aria-controls`, no default label | `aria-pressed` + `aria-controls` | `data-state`, `aria-controls` | (manual) |
| Caret on toggle | **relies on focus-retention + browser** | **explicit save-on-blur + `setSelectionRange` in rAF** | not documented | — |
| Hide on submit/reset | ✅ (reset respects `defaultPrevented`, submit unconditional) | ✅ (identical asymmetry) | ✅ | — |
| autofill defaults | `current-password` / `autocapitalize=off` / `spellcheck=false` | same | `autoComplete` prop | — |
| Password-manager escape hatch | ❌ | ❌ | ✅ `ignorePasswordManagers` | — |
| Imperative API | ❌ | ❌ | ✅ `usePasswordInput()` (`focus`/`setVisible`/`toggleVisible`) | — |

The naming and part split are well-judged: `visible`/`onVisibleChange` diverges from Radix/Ark's `onVisibilityChange` but stays consistent with Base UI's own `open`/`onOpenChange`, `checked`/`onCheckedChange` conventions, and the toggle correctly mirrors Base UI's `Toggle` surface (`state.pressed`, `data-pressed`, `aria-pressed`). The minimal three-part anatomy (no `Indicator`/`Icon` part — icons swap via `render` + `state.pressed`) is idiomatic Base UI and a reasonable simplification of Ark/Radix.

## Bugs (2)

### 1. 🟡 Caret restoration is delegated to the browser and verified only in Chromium

**Location:** `packages/react/src/password-toggle-field/toggle/PasswordToggleFieldToggle.tsx:55-57`

```tsx
onMouseDown(event) {
  event.preventDefault();
},
```

There is no `setSelectionRange`/`selectionStart` logic anywhere in the component, yet a Chromium-only test asserts the caret selection `(2,4)` survives a `password`→`text` switch (`PasswordToggleFieldToggle.test.tsx:185-205`). Base UI's strategy — `preventDefault` on `mousedown` so focus never leaves the input — is actually cleaner than Radix's (Radix lets focus move to the button, then calls `input.focus()` and restores selection in a `requestAnimationFrame`), and because focus is retained the browser will *usually* preserve the selection on its own. But Radix went to the trouble of explicit restoration for a reason: WebKit/Firefox have historically reset the caret when an input's `type` mutates. This port keeps zero non-Chromium coverage and zero safety net.

**Failure scenario:** A Safari/Firefox user places the caret mid-password, clicks the reveal toggle, and the caret jumps to the end — losing their place mid-edit. No test would catch it.

**Fix:** Either add a Firefox/WebKit (playwright) assertion confirming the browser preserves selection under the focus-retention approach, or port Radix's explicit `setSelectionRange` restore as a safety net. At minimum, rename the test from "restores the caret position" (no restoration code exists) to "preserves" and treat it as a focus-retention assertion.

### 2. 🟡 Form `reset`/`submit` listener binds once and never re-binds for a late or changing form

**Location:** `packages/react/src/password-toggle-field/input/PasswordToggleFieldInput.tsx:42-60`

```tsx
useIsoLayoutEffect(() => {
  const form = inputRef.current?.form;
  if (!form) {
    return undefined;
  }
  return mergeCleanups(/* reset + submit listeners */);
}, [inputRef, setVisible]);
```

Both deps are stable (`inputRef` is a ref object, `setVisible` is a `useStableCallback`), so this effect runs exactly once and reads `inputRef.current?.form` a single time. If the input mounts before its ancestor `<form>` exists (form conditionally added later, or the input reparented into a form), `form` is `undefined`, the effect returns early, and the hide-on-submit/reset listeners are **never attached** — so a revealed password can persist past a submit, defeating the stated security intent. (Radix uses the same one-shot pattern, so this is not a regression against the reference, but it is a real latent edge.)

**Failure scenario:** A field rendered, then wrapped in a `<form>` on a later render, stays revealed across submit.

**Fix:** Key the effect on the resolved form node (e.g. track `inputRef.current?.form` via state/`useIsoLayoutEffect` that re-runs when the element/form changes) so listeners re-bind. Low urgency given the normal ancestor-form case works.

## Tests (3)

### 1. 🟡 Controlled `visible` + toggle click is never exercised

**Location:** `packages/react/src/password-toggle-field/root/PasswordToggleFieldRoot.test.tsx:41-54`

The only controlled test drives change via `setProps`, never a click. In controlled mode a click must fire `onVisibleChange` while leaving the type pinned (since `setVisibleState` is a no-op). The core controlled-component contract — click does *not* locally flip the type — has no coverage.

**Failure scenario:** A regression where a click locally toggles type in controlled mode ships undetected.

**Fix:** Add a test: `<Root visible={false} onVisibleChange={spy}>`, click the toggle, assert `spy` fired with `true` and the input stays `type="password"`.

### 2. 🟡 The `nativeButton={false}` path is untested

**Location:** `packages/react/src/password-toggle-field/toggle/PasswordToggleFieldToggle.tsx:39`

`nativeButton` is forwarded to `useButton`, changing role/`type`/keyboard handling and how `disabled` is expressed (`aria-disabled` vs the native attribute). No test renders a non-native toggle, including the disabled-click-suppression path. (I verified the suppression *is* correct — `getButtonProps` receives the toggle's `onClick` as an external handler and gates it on `!disabled` — but it's unguarded by any test.)

**Fix:** Add a `<Toggle nativeButton={false} render={<span />} disabled />` case asserting it does not toggle and exposes `aria-disabled`.

### 3. ℹ️ Loose / decorative assertions

- `PasswordToggleFieldRoot.test.tsx:85` — `expect(...event).toBeInstanceOf(Event)` passes for nearly any DOM event; asserting `MouseEvent` for the click would actually pin that `event.nativeEvent` is forwarded.
- `PasswordToggleFieldRoot.test.tsx:188-208` — the capture-phase `submit` `preventDefault` is decorative: the source hides unconditionally on submit, so the assertion passes whether or not the listener exists. Asserting hide both with and without `preventDefault` would pin the intended reset-vs-submit asymmetry.

## Simplifications (1)

### 1. 🟡 Root's `inputId`/`setInputId` channel duplicates Field's `LabelableContext.controlId` when inside a Field

**Location:** `packages/react/src/password-toggle-field/root/PasswordToggleFieldRoot.tsx:48,72-73`; `input/PasswordToggleFieldInput.tsx:38`

The Input renders `Field.Control`, which already registers its resolved id into `LabelableContext.controlId` (via `useLabelableId`). The component maintains a *second*, parallel id channel (`inputId` state + `useRegisteredLabelId` + context) solely to feed the toggle's `aria-controls`. Inside a `Field.Root` there are then two sources of truth for the same id.

**Cost:** Redundant state/registration. Note this can't be *fully* removed — `LabelableContext` is only populated when a `Field.Root` ancestor exists and this component supports standalone use — so the cleanup is a `controlId ?? inputId` consolidation or at least a comment explaining why both exist.

## Docs (2)

### 1. 🟡 Hero demo and the docs disagree on how to label the input

**Location:** `docs/.../password-toggle-field/page.mdx:52-66` vs `demos/hero/css-modules/index.tsx:10-18` (and `tailwind/index.tsx:9-16`)

The "Form integration" prose recommends `Field.Root` + `Field.Label` for labeling/validation, but the shipped hero demo uses a plain `<label htmlFor>` + manual `React.useId()`. Users copy the hero and miss the Field path the docs sell.

**Fix:** Make the hero use the recommended Field labeling (or align the prose to the hero's approach).

### 2. ℹ️ The `autoComplete="current-password"` default is undocumented

**Location:** `packages/react/src/password-toggle-field/input/PasswordToggleFieldInput.tsx:65`

The Input hardcodes `autoComplete="current-password"` (matching Radix and Ark's defaults), which is wrong for signup/change-password forms (`new-password`). It's overridable, but the docs never mention the default exists. Worth a one-line note in the Input section. (Ark surfaces this as an explicit `autoComplete` Root prop and even adds an `ignorePasswordManagers` escape hatch — a possible future addition, not needed now.)

## Verdict

**Approve after nits** — no correctness defect blocks merge; the component faithfully mirrors Radix's proven design, but it drops Radix's explicit caret restoration in favor of a browser-reliant path that's only Chromium-tested, and it's missing controlled-mode and `nativeButton={false}` coverage. Address the caret test coverage and the demo/docs labeling mismatch before merge.

---

aarongarciah:
I pushed a couple of commits adjusting the icons and the toggle size.

sss-Mihail-sss:
When toggling password visibility, caret moves to the front of the input field.

https://github.com/user-attachments/assets/3508dde5-b0a1-4290-8d8c-0b11b8703d2b

PR #5115, opened 2026-06-24
[listbox] Create the Listbox component

michaldudak:


atomiks:
# PR review

PR #5115 adds a new `Listbox` component (~8k lines: selection modes, keyboard nav, Pragmatic drag-and-drop, lazy loading, form integration). **There is one merge-blocker:** `itemToStringLabel` is a documented public prop that is wired into the store but has no consumer, so it silently does nothing. Beyond that, the notable items are a confirmed wrong-highlight after non-contiguous multi-select reorder, a first-ArrowDown that skips the first option, a duplicate-key form bug, an RTL gap, a stale group-label id, a double tab-stop, and two large-list perf cliffs. The scariest candidates I chased — composite-index vs. DOM-order desync and a Shift+Arrow double-move — are **non-issues** (composite sorts by `compareDocumentPosition` and bails on Shift). Verification was static (no test run / profiler).

## Bugs (12)

### 1. 🔴 `itemToStringLabel` ships as a no-op public API

**Location:** `packages/react/src/listbox/root/ListboxRoot.tsx:414`

```tsx
itemToStringLabel?: ((itemValue: Value) => string) | undefined;
```

The prop is stored on the Listbox store (`store.ts:14,66`) and passed through Root, but **nothing reads `selectors.itemToStringLabel`**. `labelsRef` (used by typeahead) is populated only from each item's `label` prop or rendered text content (via `useCompositeListItem`). Unlike Combobox — where `itemToStringLabel` is consumed in `ComboboxValue.tsx` — Listbox has no consumer.

**Failure scenario:** A user passes `itemToStringLabel={(item) => item.name}` expecting object-value typeahead to work; matching still fails unless every `Listbox.Item` also has an explicit `label`. The documented prop does nothing.

**Fix:** Wire it into item label registration with `stringifyAsLabel` (so typeahead can match object values), or remove the prop and its docs before release.

### 2. 🟡 Multi-select keyboard reorder highlights/focuses the wrong item for non-contiguous selections

**Location:** `packages/react/src/listbox/item/ListboxItem.tsx:446`

```tsx
const offsetInSelection = selectedIndices.indexOf(resolvedIndex);
const newFirstIdx = moveUp ? firstIdx - 1 : firstIdx + 1;
store.set('activeIndex', newFirstIdx + offsetInSelection);
```

`newFirstIdx = firstIdx ± 1` only equals the moved block's landing index when the selection is **contiguous**. For a non-contiguous selection the highlight (and `restoreFocusAndScroll`, which focuses `[role="option"][tabindex="0"]`) lands on the wrong row. Confirmed by tracing.

**Failure scenario:** `multiple` mode, items `[0,1,2,3,4]`. Toggle-select **1 and 3**, focus item 3, Alt+ArrowDown → items move correctly to `[0,2,4,1,3]` but `activeIndex` is set to 3 (value `1`) while the initiating item is now at index 4. Focus jumps to the wrong item, and a follow-up Alt+Arrow then reorders the wrong item. The contiguous case works.

**Fix:** After the move, recompute the initiating item's index from the registry: `store.set('activeIndex', findItemIndex(valuesRef.current, itemValue, eqFn))`.

### 3. 🟡 First ArrowDown skips the first option

**Location:** `packages/react/src/listbox/list/ListboxList.tsx:50`

```tsx
highlightedIndex: activeIndex ?? 0,
```

When the list container is focused with nothing highlighted, `activeIndex` is `null` so composite is seeded with `highlightedIndex: 0` — but no item is actually highlighted/focused. The first ArrowDown advances *from* the phantom index 0 to index 1 (`useCompositeRoot` treats `nextIndex === highlightedIndex` as "already there" and steps forward).

**Failure scenario:** A keyboard user tabs into a fresh listbox and presses ArrowDown; option 1 is highlighted and option 0 is skipped.

**Fix:** Report the first item as highlighted on list focus, or pass a sentinel (e.g. `null`/`-1` semantics) so the first forward arrow lands on index 0.

### 4. 🟡 Duplicate React keys on hidden inputs drop a form value

**Location:** `packages/react/src/listbox/root/ListboxRoot.tsx:280`

```tsx
return <input key={currentSerializedValue} type="hidden" name={name} value={currentSerializedValue} />;
```

The key is the serialized value, so two selected values that serialize to the same string collide → React drops one input and the form submits fewer values than selected (plus a console warning).

**Failure scenario:** A controlled `value={['a', 'a']}` (controlled mode bypasses the reducer dedup), or two distinct objects with a by-id `isItemEqualToValue` and a non-injective `itemToStringValue` → one `<input>` dropped → `FormData` has one entry instead of two.

**Fix:** Include the index in the key: `` key={`${i}-${currentSerializedValue}`} ``. (The comma-`join` for `serializedValue` is *not* a submission bug — per-value hidden inputs carry submission; the joined string only feeds validity tracking.)

### 5. 🟡 Horizontal Alt+Arrow reorder ignores RTL

**Location:** `packages/react/src/listbox/item/ListboxItem.tsx:317`

```tsx
const moveUp = (isVertical && event.key === 'ArrowUp') || (!isVertical && event.key === 'ArrowLeft');
const moveDown = (isVertical && event.key === 'ArrowDown') || (!isVertical && event.key === 'ArrowRight');
```

`ListboxList.handleSelectionKeyDown` swaps Left/Right for `direction === 'rtl'` (List.tsx:155-161), but this keyboard-reorder handler does not.

**Failure scenario:** Horizontal listbox with `dir="rtl"` and DnD enabled — ArrowLeft reorders the opposite direction from how Shift+Arrow navigates the same widget.

**Fix:** Mirror the RTL swap: in horizontal RTL, ArrowRight → `before`/moveUp, ArrowLeft → `after`/moveDown.

### 6. 🟡 Group labels leave a stale `aria-labelledby`

**Location:** `packages/react/src/listbox/group-label/ListboxGroupLabel.tsx:25`

```tsx
useIsoLayoutEffect(() => {
  setLabelId(id);
}, [id, setLabelId]);
```

Unlike `MenuGroupLabel` and `ComboboxGroupLabel` (both `return () => setLabelId(undefined)`), this effect never unregisters the id.

**Failure scenario:** A conditional `Listbox.GroupLabel` unmounts and the parent group keeps `aria-labelledby` pointing at a removed element.

**Fix:** `return () => setLabelId(undefined);`

### 7. 🟡 Double tab stop: list container and roving item both `tabIndex={0}`

**Location:** `packages/react/src/listbox/list/ListboxList.tsx:233` + `packages/react/src/listbox/item/ListboxItem.tsx:478`

The list is permanently `tabIndex: disabled ? undefined : 0` and items use roving tabindex (`highlighted ? 0 : -1`) with real DOM focus (no `aria-activedescendant`). When an item is highlighted, both the container and that item are in the tab order. Sibling `Combobox.List` uses `tabIndex: -1` (ComboboxList.tsx:76).

**Failure scenario:** With a highlighted item, Tab into the listbox stops on the container, then on the option — an extra, semantically empty stop.

**Fix:** Set the container `tabIndex={-1}` once items can take the roving stop (keep `0` only for the empty-list case); confirm the intended focus model first.

### 8. 🟡 Typeahead can't cycle on a repeated character and skips the current item

**Location:** `packages/react/src/listbox/list/ListboxList.tsx:85,95`

```tsx
typeaheadStringRef.current += event.key.toLowerCase();
// ...
const startIndex = activeIndex != null ? activeIndex + 1 : 0;
```

The comment claims "typing the same character repeatedly cycles through matches," but the buffer always appends, so pressing `a` then `a` searches for `aa` and cannot cycle through `Apple`, `Apricot`, … Separately, starting at `activeIndex + 1` means a growing multi-char buffer can jump away from a still-matching current item.

**Failure scenario:** Repeatedly pressing the same starting letter stops matching instead of cycling through same-initial options.

**Fix:** Special-case a single repeated-character buffer to cycle from `activeIndex + 1`, and start at `activeIndex` (inclusive) for genuine multi-char searches — or reuse the existing `useTypeahead` behavior.

### 9. 🟡 `onHighlightChange` runs a full `querySelectorAll` on every store mutation

**Location:** `packages/react/src/listbox/utils/useHighlightChangeNotifier.ts:61` (called from `store.subscribe` at :116)

```tsx
const element = listEl?.querySelectorAll<HTMLElement>('[role="option"]')[activeIndex] ?? null;
```

The subscription fires on every `store.set` (selection, hover, drag, loading), and `resolveHighlightedItem` walks all N option nodes *before* the `areHighlightsEqual` early-out — even when `activeIndex` didn't change. O(N) DOM query per mutation, only when `onHighlightChange` is set.

**Failure scenario:** A 5k-item listbox with an `onHighlightChange` handler; holding ArrowDown or rapid hover does a full DOM walk per event.

**Fix:** Gate on an actual `activeIndex` change (`store.observe('activeIndex', …)`) and resolve the element from `composite.elementsRef` rather than re-querying.

### 10. 🟡 Every item re-registers with Pragmatic DnD on any reorder

**Location:** `packages/react/src/listbox/utils/useDragAndDrop.ts:60-87` (effect deps include `index`, `itemValue`)

The setup `useIsoLayoutEffect` tears down and re-creates `draggable()` + `dropTargetForElements()` whenever `index` changes; a reorder shifts the index of every item between source and target. The handlers already read live data from refs/store.

**Failure scenario:** Dragging the first item of a 2k-item list to the bottom re-registers ~2k items synchronously post-drop → visible hitch.

**Fix:** Read `index`/`itemValue` from refs (the component has `indexRef`) inside the Pragmatic callbacks and drop them from the effect deps.

### 11. ℹ️ Smaller correctness/realm notes

- **Stale closure `index`** in `onClick`/`onFocus`/`onMouseMove` (`ListboxItem.tsx:480,484,525`) — only keydown re-resolves via `findItemIndex`; a click/hover in the window between an optimistic registry reorder and React re-render can act on a stale index.
- **`IntersectionObserver` uses the global constructor** (`ListboxLoadingTrigger.tsx:76`) instead of `ownerWindow(root).IntersectionObserver` — wrong realm inside a cross-document iframe/portal (AGENTS.md owner-utilities rule); SSR-safe otherwise.
- **Non-memoized selectors** (`store.ts`) — `isSelected` does `value.some(...)` per item, so a toggle in `multiple` mode is O(N·M) in `getSnapshot`; `createSelectorMemoized` would short-circuit on unchanged `value`.

### 12. ℹ️ API-design footguns on the new public surface (hard to reverse post-release)

- **`value` is always `Value[]`, even in `'single'` mode** (`ListboxRoot.tsx:424`) — `Select`/`Combobox` give a scalar in single mode; the single-select hero demo is forced into `defaultValue={['billie-jean']}` and consumers read `value[0]` everywhere.
- **`selectionMode: 'single' | 'multiple' | 'explicit-multiple'`** vs. the library-wide public `multiple` boolean (Combobox maps it internally); `'explicit-multiple'` is undiscoverable by name.
- **`onHighlightChange(value, element)`** exposes a raw `HTMLElement` positionally, whereas the sibling ships `onItemHighlighted(value, eventDetails)` with a `reason` (`ComboboxRoot.tsx:127`); `actionsRef.highlightValue(value, element?)` similarly leaks keyed-reconciliation internals.

## Tests (2)

### 1. 🟡 Core public behaviors ship with no test

Several headline behaviors are uncovered, so a regression in any would stay green: **multi-select Alt+Arrow reorder** (including the non-contiguous case from bug #2), **single-item pointer-drag reorder** (only multi-drag/`canDrop=false`/drop-on-self are tested), **`actionsRef.highlightValue`**, **Escape-to-clear** and **Ctrl+A toggle-all**, **autofill `onChange`** resolution, **`required`/form-submission round-trip** (incl. a duplicate value, which exercises bug #4), and **typeahead reset/wrap/cycle**. Gate layout-dependent cases in Chromium via `it.skipIf(isJSDOM)`.

### 2. 🟡 Split the multi-assertion `waitFor`

**Location:** `packages/react/src/listbox/drag-and-drop-provider/ListboxDragAndDropProvider.test.tsx:131`

```tsx
await waitFor(() => {
  expect(screen.getAllByRole('option').map((el) => el.textContent)).toEqual([...]);
  expect(screen.getByRole('option', { name: 'b' })).toBe(document.activeElement);
});
```

AGENTS.md asks for one assertion per `waitFor` so retries stay scoped. **Fix:** Use separate `waitFor` calls for the order and the focus assertions.

## Simplifications (4)

### 1. 🟡 Re-use `selectedValueIncludes` instead of hand-rolling `.some(eqFn)`

`packages/react/src/internals/itemEquality.ts` exports `selectedValueIncludes(values, item, comparer)` (with null guards), re-implemented at ~5 sites: `store.ts:82`, `ListboxList.tsx:211`, `ListboxItem.tsx:423`, `ListboxDragAndDropProvider.tsx:158,168`. **Fix:** Call the helper; several sites can drop their `compareItemEquality` import.

### 2. 🟡 Dead `policySignature` field on the DnD context

**Location:** `ListboxDragAndDropProviderContext.ts:41` + `ListboxDragAndDropProvider.tsx:326`. Declared and constructed but never read (and it pins `canDrag, canDrop` in the memo deps). **Fix:** Delete it.

### 3. 🟡 Raw `setTimeout` in `restoreFocusAndScroll` bypasses the `useTimeout` lifecycle

**Location:** `packages/react/src/listbox/item/ListboxItem.tsx:347`. The code documents *why* it avoids `useTimeout` (the item can unmount during a cross-group reorder and a hook timer would be canceled). The trade-off is a real, uncancelable timer that can run post-unmount against stale state (it mostly no-ops against the external store). **Fix:** Move the scheduler to a root/provider-scoped callback (which doesn't unmount mid-reorder) using `useTimeout`/`afterDomSettle`, keeping the cross-group robustness without a raw timer.

### 4. ℹ️ Dead/duplicated state

- **`typingRef`** (`store.ts:49`, written in `ListboxList` typeahead) is never read — remove it, or use it to suppress hover-steal during typeahead (the likely intent).
- The "iterate values, skip undefined/disabled" predicate (`selectAll`, Ctrl+A pre-check, DnD multi-drag), the "collect selected indices" loop (Item reorder vs. DnD provider), and the 4× `reorderRegistry` pass over 4 parallel arrays can each collapse to one shared helper.

## Docs (2)

### 1. 🟡 DnD-context error message is terser than its siblings

**Location:** `packages/react/src/listbox/drag-and-drop-provider/ListboxDragAndDropProviderContext.ts:59` — `'… is missing. Use <Listbox.DragAndDropProvider>.'` drops the convention the other three Listbox context errors use. **Fix:** `'Base UI: ListboxDragAndDropProviderContext is missing. Listbox drag-and-drop parts must be placed within <Listbox.DragAndDropProvider>.'`, then re-run `pnpm extract-error-codes` (same arg count → update code 99 in place).

### 2. ℹ️ Smaller doc gaps

- **DnD payload type shapes are not rendered.** `types.md:321-323` references `ListboxDragAndDropItem<Value>` and `ListboxDragAndDropProviderOnItemsReorderEvent<Value>` in the prop table, but only `ListboxDragAndDropTargetEdge` is rendered (`:411`). Users can't see that items expose `{ value, index, groupId, disabled }` or that reorder events carry `reason`. Add those shapes.
- The substantial multi-select keyboard contract (Shift+Arrow, Ctrl/Cmd+A, Ctrl/Cmd+Shift+Home/End, Escape) is implemented but **undocumented** on `listbox/page.mdx` (only Alt+Arrow is mentioned).
- The "two render cycles" comments (`ListboxRoot.tsx:191`, `useHighlightChangeNotifier.ts:20`) understate the actual `afterDomSettle` mechanism (a macrotask + two animation frames).

## Verdict

**Request changes** — the no-op `itemToStringLabel` public prop (#1) should be wired up or removed before this new API ships; the multi-select-reorder highlight (#2), first-ArrowDown skip (#3), and duplicate-key form bug (#4) are the next most worth fixing.

---

PR #5116, opened 2026-06-24
[field] Keep invalid state on disabled fields

### LINKED ISSUE #5100: [Field] Control "valid" when `disabled`
# Bug report

## Current behavior

Behavior change between 1.5.0 and 1.6.0 related to `disabled` and `invalid` state. In 1.5.0, the `invalid` state would be added to the Control **even** when the control was `disabled`. In 1.6.0, this changed, and when the control is `disabled`, an `invalid` state is ignored.

### Related Change

https://github.com/atomiks/base-ui/blob/e6dc73dfa253e0c11513e93e0ce36939f828351f/packages/react/src/field/root/FieldRoot.tsx#L111

Relates to #4890

## Expected behavior

This _might_ be the intended behavior, I just wanted to double-check with the maintainers.

### Base UI 1.5.0

displays `data-invalid` on a control when both `disabled` and `invalid` are set on the `Field.Root`.

Image

https://stackblitz.com/edit/yogmuxar-9lbmyjxq?file=src%2FApp.tsx

### Base UI 1.6.0

If a field is `disabled`, `data-invalid` is no longer present.

Image

https://stackblitz.com/edit/yogmuxar-rpzzwdhj?file=src%2FApp.tsx

atomiks:
Closes #5100

## Problem

When a field was marked invalid by the app — through the `invalid` prop or `<Form>` errors — `data-invalid` was removed as soon as the field became disabled. This used to work in 1.5.0 and changed in 1.6.0 (#4890), which started excluding disabled controls from validation.

## Fix

An app-set invalid state now stays on the field even while disabled. Only computed validity (native constraints and `validate`) is still suppressed when disabled, matching how `:disabled` controls don't take part in constraint validation.

`aria-invalid` is still not added to disabled controls, so this only restores the `data-invalid` styling hook.

This lines up with React Hook Form, which skips validation on disabled fields but keeps errors set by the app.

## Tests

Added tests for the `invalid` prop and `<Form>` errors on a disabled field.

### DIFF
--- a/packages/react/src/field/root/FieldRoot.test.tsx
+++ b/packages/react/src/field/root/FieldRoot.test.tsx
@@ -342,6 +342,45 @@ describe('<Field.Root />', () => {
       expect(label).toHaveAttribute('data-disabled', '');
       expect(message).toHaveAttribute('data-disabled', '');
     });
+
+    it('keeps an explicitly invalid field marked invalid while disabled', async () => {
+      await render(
+        <Field.Root data-testid="field" disabled invalid>
+          <Field.Control data-testid="control" />
+          <Field.Label data-testid="label" />
+          <Field.Description data-testid="description" />
+        </Field.Root>,
+      );
+
+      const field = screen.getByTestId('field');
+      const control = screen.getByTestId('control');
+      const label = screen.getByTestId('label');
+      const description = screen.getByTestId('description');
+
+      expect(field).toHaveAttribute('data-invalid', '');
+      expect(control).toHaveAttribute('data-invalid', '');
+      expect(label).toHaveAttribute('data-invalid', '');
+      expect(description).toHaveAttribute('data-invalid', '');
+
+      // It does not participate in native constraint validation.
+      expect(control).not.toHaveAttribute('aria-invalid');
+    });
+
+    it('keeps a disabled field with form errors marked invalid', async () => {
+      await render(
+        <Form errors={{ name: 'Server error' }}>
+          <Field.Root name="name" disabled>
+            <Field.Control data-testid="control" />
+          </Field.Root>
+        </Form>,
+      );
+
+      const control = screen.getByTestId('control');
+
+      expect(control).toHaveAttribute('data-invalid', '');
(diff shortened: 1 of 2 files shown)


flaviendelangle (approved):
# PR review

This is a focused, correct fix: the single-line change to `valid` restores `data-invalid` on app-marked-invalid fields while they're disabled, without re-enabling computed (native/`validate`) validity or `aria-invalid` on disabled controls. I traced all branches of the rewritten boolean and every consumer of the computed `valid`; behavior changes only for the disabled+invalid case, exactly as intended. Nothing is merge-blocking.

## Bugs (0)

No findings.

The rewrite from `disabled ? null : !invalid && validityData.state.valid` to `!invalid && (disabled ? null : validityData.state.valid)` is equivalent on every branch except disabled+invalid:
- `invalid` true → `valid = false` → `data-invalid` (now also when disabled). ✓
- `invalid` false + disabled → `null` → no attribute (unchanged). ✓
- `invalid` false + enabled → `validityData.state.valid` (unchanged). ✓

`aria-invalid` stays suppressed because `getValidationProps` (`useFieldValidation.ts:323`) gates on `!disabled`, so the new `state.valid === false` never reaches it while disabled. The `revalidate` early-return at `useFieldValidation.ts:149` (`state.valid !== false`) is reachable only via input change events, which disabled controls don't fire, so the flipped value there has no practical effect. `FieldError` rendering and `getCombinedFieldValidityData` read raw `validityData.state.valid`, not the computed `valid`, so they're untouched.

## Tests (0)

No findings.

Both added tests pin the right behavior: `data-invalid` present on root/control/label/description, and `aria-invalid` explicitly asserted absent — covering both the `invalid` prop and `<Form errors>` paths. They don't conflict with the existing computed-validity tests.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

The inline comment accurately describes the new semantics.

## Verdict

**Approve** — minimal, behavior-equivalent except for the intended disabled+invalid case, with matching tests and no regressions found.

---

PR #5119, opened 2026-06-25
[select] Don't force-mount the popup on programmatic value changes

atomiks:

Spotted on the docs site: pages like the combobox one had a stack of empty `data-base-ui-portal` nodes mounted on load — one per demo toolbar's "Styling method" select.

## Cause

Each toolbar select is controlled, and its value resolves from a stored preference one tick after hydration (SSR renders the default, then the persisted variant is applied in an effect). That `value` change tripped this effect in `SelectRoot`:

```tsx
if (value !== initialValueRef.current) {
  store.set('forceMount', true);
}
```

`forceMount` is sticky and `SelectPortal` renders whenever `mounted || forceMount`, so the popup stayed mounted permanently — for every such select on the page.

## Fix

Remove the effect. It turns out to be redundant in the current implementation:

- `Select.Value` resolves the label from `value`/`items` via `resolveSelectedLabel`, never from the mounted list.
- `selectedIndex` and the selected-item text ref are restored by `SelectItem` registration when the popup mounts on open.
- Closed-trigger typeahead force-mounts via the trigger's `onFocus`, and autofill via the hidden input's `onChange`.

(The first iteration only gated the effect behind `items`; review surfaced that the whole effect is unnecessary, so it's removed outright — fixing every controlled select whose value settles after mount, not just `items`-based ones.)

## Testing

- Regression test: a programmatic value change (with and without `items`) doesn't force-mount the popup, and the label still updates.
- Test pinning closed-trigger typeahead with `items` (popup never opened) to guard the trigger-focus force-mount path the removal relies on.
- Full select suite passes in jsdom and Chromium; `pnpm typescript` / `eslint` / `prettier` clean. No public API change.

### DIFF
--- a/packages/react/src/select/root/SelectRoot.test.tsx
+++ b/packages/react/src/select/root/SelectRoot.test.tsx
@@ -2858,17 +2858,28 @@ describe('<Select.Root />', () => {
     });
   });
 
-  it('does not force-mount the popup on a programmatic value change when items are provided', async () => {
+  it('does not force-mount the popup on a programmatic value change', async () => {
     function App() {
-      const [value, setValue] = React.useState<string | null>(null);
+      const [withItems, setWithItems] = React.useState<string | null>(null);
+      const [withoutItems, setWithoutItems] = React.useState<string | null>(null);
       return (
         <div>
-          <button type="button" onClick={() => setValue('b')}>
+          <button
+            type="button"
+            onClick={() => {
+              setWithItems('b');
+              setWithoutItems('b');
+            }}
+          >
             set
           </button>
-          <Select.Root items={{ a: 'Apple', b: 'Banana' }} value={value} onValueChange={setValue}>
-            <Select.Trigger data-testid="trigger">
-              <Select.Value data-testid="value" />
+          <Select.Root
+            items={{ a: 'Apple', b: 'Banana' }}
+            value={withItems}
+            onValueChange={setWithItems}
+          >
+            <Select.Trigger>
+              <Select.Value data-testid="items-value" />
             </Select.Trigger>
             <Select.Portal>
               <Select.Positioner>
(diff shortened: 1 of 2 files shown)


flaviendelangle (approved):
# PR review

A clean, well-scoped fix: the force-mount on programmatic value change is correctly gated behind `!hasItemsProp`, which stops the docs-site DOM bloat (permanently-mounted empty `data-base-ui-portal` nodes) without regressing label resolution or typeahead. Nothing is merge-blocking. I verified the three behavioral claims the fix relies on — label resolution from `items` (`SelectValue.tsx:57-59`), closed-trigger typeahead force-mounting via the trigger's `onFocus` (`SelectTrigger.tsx:165-167`), and autofill via the hidden input (`SelectRoot.tsx:617`) — and the new gate matches Combobox's existing pattern (`AriaCombobox.tsx:460-468`).

## Bugs (0)

No findings.

## Tests (1)

### 1. ℹ️ Closed-trigger typeahead with `items` is the load-bearing claim but isn't pinned by a test

**Location:** `packages/react/src/select/root/SelectRoot.test.tsx:2861`

The fix is only safe because, with `items` provided, closed-trigger typeahead force-mounts "through its own path" — the trigger `onFocus` → `timeoutFocus.start(0, …)` → `store.set('forceMount', true)` chain (`SelectTrigger.tsx:165`). The new test pins label resolution (`Select.Value` shows `Banana`) but not that path. If a future refactor drops the focus force-mount, typeahead on a closed trigger with `items` would silently break and this suite would stay green.

**Failure scenario:** A regression that removes the trigger-focus force-mount leaves closed-trigger typeahead unable to populate `labelsRef`/`valuesRef`, so typing a letter on the closed trigger no longer selects the matching item — with no failing test to catch it.

**Fix:** Add a Chromium test (`it.skipIf(isJSDOM)`) that, with `items` provided and the popup never opened, focuses the trigger, types a letter, and asserts the selected value updates. Optional but it converts the PR's prose argument into a guard.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve** — correct, minimal fix; tests pass and the only note is an optional test-coverage addition.

---

PR #5120, opened 2026-06-25
[popups] Prevent unwanted flip with capped scrollable content

### LINKED ISSUE #5118: [combobox] Popup positioning gets tricky with scrollable content
# Bug report

## Current behavior

When the popup has a lot of items, its placement becomes a bit unpredictable:

- If the trigger is in the upper half of the screen, the popup jumps above it
- If it's in the lower half, it goes below

This happens even when there’s plenty of visual space in the viewport to display the popup normally.

I noticed that the [Input inside popup](https://base-ui.com/react/components/combobox#input-inside-popup) example in the docs handles it by constraining the max-height of both the popup and the inner List. While that works for simple cases, it feels a bit fragile for real-world apps because:

1. It forces us to hardcode or pre-calculate the height of whatever is inside the popup (like headers or inputs)
2. In my specific case, I’m wrapping the List inside a custom ScrollArea component, and forcing a max-height on the list completely breaks my vertical paddings.

## Expected behavior

It feels like the positioning logic might be looking at the full, uncollapsed height of the children to decide where to place the popup rather than checking the actual height of the popup itself.

Since max-height is usually styled on the popup container anyway, wouldn't it be safer to base the positioning logic strictly on the popup’s bounding box?

## Reproducible example

[Demo]: https://stackblitz.com/edit/2wrfuw59 - I just replicated the very first example from the docs. The first element renders the popup below (since it's in the upper half of the viewport) while the other one renders it above, even though there's plenty of space available at the bottom to show it

https://github.com/user-attachments/assets/79fb886a-62b6-45b6-9ac4-a8be10c88fc1

## Base UI version

v1.6.0

Chrome

## Which OS are you using?

Mac OS

atomiks:
## Summary

Fixes #5118.

When a popup's scrollable content is taller than the viewport and capped with the documented `max-height: min(<x>, var(--available-height))` pattern, the popup could open on the wrong side: it flipped away from its preferred side even when the capped popup fit there comfortably (e.g. a combobox in the lower half of the viewport opening upward despite plenty of room below).

**Root cause:** `flip()` runs before `size()` sets `--available-height`. On the first positioning pass that variable is undefined, so `min(<x>, var(--available-height))` resolves to an invalid value and the whole `max-height` declaration is dropped. The list is therefore measured at its full, uncapped content height while `flip()` decides the side, so it picks the side with more raw space rather than the side where the capped popup actually fits.

## Fix

Seed `--available-width: 100vw` and `--available-height: 100vh` on the positioner before `size()` writes the real values. This keeps the consumer's `min(<x>, var(--available-height))` valid on the first `flip()` pass, so the popup is measured at its capped height and the side is chosen correctly. `size()` overwrites the vars with the real px values immediately, and React's style reconciliation does not clobber those imperative values (the seeded value string is constant across renders, so React never rewrites the property after mount).

No changes to the docs demos are required — the existing patterns now behave as intended.

## Testing

- Added a Chromium regression test in `ComboboxPositioner.test.tsx` (fails without the fix: the popup flips to `top`; passes with it).
- All positioner suites pass: Combobox, Select, Menu, Navigation Menu, Popover, Tooltip, Preview Card.

### DIFF
--- a/packages/react/src/combobox/positioner/ComboboxPositioner.test.tsx
+++ b/packages/react/src/combobox/positioner/ComboboxPositioner.test.tsx
@@ -1,7 +1,7 @@
 import { expect, vi } from 'vitest';
 import * as React from 'react';
 import * as ReactDOMClient from 'react-dom/client';
-import { waitFor } from '@mui/internal-test-utils';
+import { screen, waitFor } from '@mui/internal-test-utils';
 import { Combobox } from '@base-ui/react/combobox';
 import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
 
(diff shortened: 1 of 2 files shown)


flaviendelangle (approved):
# PR review

This is a tight, well-reasoned fix for an unwanted-flip bug: it seeds `--available-width: 100vw` / `--available-height: 100vh` in the positioner's React style object so consumer `min(<x>, var(--available-height))` rules resolve to a valid length on the first `flip()` pass, before `size()` writes the real px values. The mechanism is correct — React's per-property style diff never rewrites a key whose value is constant across renders, so `size()`'s imperative px values survive — and the change only ever makes the flip pass measure a *smaller* (capped) popup, which strictly reduces unwanted flips rather than introducing new ones. Nothing is merge-blocking.

## Bugs (0)

No findings.

## Tests (1)

### 1. ℹ️ Regression test covers only the `min(...)` consumer pattern, not the bare `var(--available-height)` pattern

**Location:** `packages/react/src/combobox/positioner/ComboboxPositioner.test.tsx:90`

```tsx
style={{ maxHeight: 'min(80px, var(--available-height))', overflowY: 'auto' }}
```

The new test exercises the documented `min(<x>, var(--available-height))` shape. But the default Select/Menu dropdowns (and several demos) use the bare form `max-height: var(--available-height)` — see `docs/src/components/Select.css:11` and `Menu.css:11`. The first-pass behavior change differs between the two: for the bare form, the declaration previously dropped entirely (uncapped) and now resolves to `100vh`. Since the fix lives in shared `useAnchorPositioning`, the Combobox test does cover the shared code path, so this is informational rather than a real gap — but a Select/Menu case asserting the bare pattern no longer over-flips would pin the behavior these components actually ship with.

**Failure scenario:** A future change to how the bare `var(--available-height)` first-pass value resolves could regress Select/Menu flip behavior without any test failing.

**Fix:** Optionally add an analogous `it.skipIf(isJSDOM)` flip test to `SelectPositioner` (or Menu) using `max-height: var(--available-height)`.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve** — correct, minimal fix; the explanatory comment accurately documents the React-reconciliation invariant the fix depends on, and the only note is an optional test-breadth suggestion.

---

PR #5121, opened 2026-06-25
[docs][context menu] Add usage guidelines and with menu demo

### LINKED ISSUE #4171: [Context menu] component is inaccessible and has no warning in the docs #4170
# Bug report

## Current behavior

There is no way to activate the Context Menu popup using a keyboard, screen reader, touch or speech input. There is no warning against using it or its accessibility limitations in the docs. 

This fails [WCAG 2.2 Success Criterion 2.1.1 Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)

## Expected behavior

Preferably, remove this component. It isn't necessary, and creates an inequitable experience between users of different modalities. The menu component already exists, which would be an accessible implementation of this functionality.

If it cannot be removed:
- either document that it is inaccessible to keyboard, screen reader, touch and speech input users and should not be used, including updating BaseUi's accessibility statement
- or update the component so that it has a named control to open the context menu 
- or update the component so that it is focusable and the usual keyboard shortcuts work (for example, <kbd>Control + Enter</kbd> on Mac or <kbd>Shift + F10</kbd> or dedicated menu keys on Windows)

## Reproducible example

Steps to reproduce: 
1. Navigate to the [Context menu](https://base-ui.com/react/components/context-menu) component demo
3. Attempt to open the menu using only a keyboard

## Base UI version

v1.2.0

Chrome 

## Which OS are you using?

Mac OS Tahoe 26.2

Keyboard, VoiceOver, Apple Voice Control

## Additional context

The simplest solution would be to remove/retire this component from Base UI. 

Alternatively:
- Implement a focusable named region with appropriate keyboard shortcuts to open the context menu  
- Implement a separate visual indicator or additional control to open the context menu

aarongarciah:
Closes https://github.com/mui/base-ui/issues/4171

Preview: https://deploy-preview-5121--base-ui.netlify.app/react/components/context-menu

- Adds usage guidelines to recommend always adding visual controls to perform actions present in a context menu.
- Adds a demo using ContextMenu alongside Menu.
- Improves two existing demos by reducing font-size and avoiding them from becoming wider than the container on small viewports.

### DIFF
--- a/docs/src/app/(docs)/react/components/context-menu/demos/with-menu/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/context-menu/demos/with-menu/css-modules/index.tsx
@@ -27,7 +27,7 @@ export default function ContextMenuWithMenuDemo() {
           <Menu.Portal>
             <Menu.Positioner align="end" sideOffset={8} className={styles.Positioner}>
               <Menu.Popup className={styles.Popup}>
-                <SharedMenuItems Item={Menu.Item} Separator={Menu.Separator} />
+                <SharedMenuItems />
               </Menu.Popup>
             </Menu.Positioner>
           </Menu.Portal>
@@ -37,7 +37,7 @@ export default function ContextMenuWithMenuDemo() {
       <ContextMenu.Portal>
         <ContextMenu.Positioner className={styles.Positioner}>
           <ContextMenu.Popup className={styles.Popup}>
-            <SharedMenuItems Item={ContextMenu.Item} Separator={ContextMenu.Separator} />
+            <SharedMenuItems type="context-menu" />
           </ContextMenu.Popup>
         </ContextMenu.Positioner>
       </ContextMenu.Portal>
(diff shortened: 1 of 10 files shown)


docs/src/app/(docs)/react/components/context-menu/demos/with-menu/css-modules/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      {actions.map((action) => (
+        <Item key={action} className={styles.Item}>
+          {action}
+        </Item>
+      ))}
+      <Separator className={styles.Separator} />
+      <Item className={`${styles.Item} ${styles.ItemDestructive}`}>Delete</Item>
+    </React.Fragment>
+  );
+}
```
aarongarciah, line 64: Not a big fan of this abstraction, but the current API requires a way of rendering different components depending on where items are being rendered. We could remove this but I think it's better to showcase how to share logic between context and regular menus. We could also include the Popup part in this abstraction for this demo.

AFAIK, one could use `Menu.Item` inside `ContextMenu`, but that's not officially supported, so not using that approach here.

colmtuite (approved):

PR #5123, opened 2026-06-25
[docs-infra] Enable Live Editing Demos in Development Mode

dav-is:
> [!NOTE]
> This PR is a work in progress
>
> Currently blocked by a bug fixed in Next.js `16.2`, but we are stuck on `16.1` due to a webpack bug. We can either wait for a fix to the webpack bug, or completely switch to turbopack.

Allows editing in development mode, extra deps are loaded only in development. `client.ts` files are automatically created when running `pnpm docs:validate` and are required to separate server and client bundles in RSC.

[Screencast From 2026-06-25 11-44-06.webm](https://github.com/user-attachments/assets/4f182237-b8d1-43d7-9314-15c561d81b9d)


colmtuite:
Does anyone want this? I don't think we do, right? @michaldudak @atomiks @jjenzz @flaviendelangle?

dav-is:
@colmtuite, @mnajdova asked for it here: https://mui-org.slack.com/archives/C042YB5RB3N/p1781529410713419?thread_ts=1781520672.787749&cid=C042YB5RB3N

Just in dev mode

mnajdova:
The ask was for the data grid team, we can close this PR and open one there. Also, ideally the config should be much simpler (one config boolean that will apply everywhere).

PR #5137, opened 2026-06-29
[docs] New mobile nav/search

aarongarciah:
### Before

The "mobile" navigation and search are two separate dialogs that are highly repetitive: before typing in the search input, a set of default search results are displayed, and these results are basically the same as the nav links.

https://github.com/user-attachments/assets/7f156602-71e9-41cc-994c-639d31b71c30

### After

A new drawer with a unified nav and search for "mobile": the drawer displays a static link list when opened, but when the user types in the input, the static link list gets hidden and the autocomplete results are displayed in its place, using the same styles.

The mobile nav/search trigger now uses a magnifying glass icon instead of the "menu" icon to help users discover the search capability.

I was a bit worried about hiding/showing the static link list, but after playing a while with it, feels pretty nice. It could get a bit confusing for a keyboard user in a small viewport because the static link list and the autocomplete results have different tab interactions (multiple tab stops vs single tab stop).

An alternative approach would be to remove the static link list altogether and use default autocomplete results instead (same approach as the desktop search).

https://github.com/user-attachments/assets/e6691883-3ad1-4900-a7ca-a72d87207bfc

The drawer radius and handle are inspired by iOS. This is a screenshot from the iOS Maps app.

Screenshot 2026-06-30 at 23 51 09


docs/src/components/MobileNavDrawer.tsx
aarongarciah, line 1: @dav-is I'm adding analytics to the mobile nav. May I ask you to check if it looks good?

atomiks:
# PR review

One merge-blocking finding: a unitless `0` fallback in the new drawer CSS invalidates the `calc()` expressions that drive the scroll viewport's bottom inset, so the bottom padding, fade scrims, and scroll padding silently collapse in the drawer's default state. The rest is a low-impact keyboard-shortcut regression and duplication between the two search implementations.

## Bugs (3)

### 1. 🔴 Unitless `0` fallback makes the scroll-inset `calc()` invalid until the search input is focused

**Location:** `docs/src/components/MobileNav.css` (`.MobileNavScrollAreaRoot`)

```css
--mobile-nav-scroll-bottom-inset: calc(
  var(--mobile-nav-bleed) + var(--drawer-keyboard-inset, 0)
);
```

`--drawer-keyboard-inset` is not a registered CSS property; `DrawerVirtualKeyboardProvider` only sets it (as `0px` or a real inset) once a keyboard-input target inside the drawer gains focus. On a fresh open where the user just browses nav links (initial focus goes to the popup, and iOS doesn't focus links on tap), the var is unset, so the fallback applies — and `calc(3rem + 0)` is `<length> + <number>`, which is invalid CSS. Substituted through `var()`, every property consuming it becomes invalid at computed-value time and falls back to `unset`: the viewport's `padding-bottom` and `scroll-padding-bottom`, the entire `mask-image` (top and bottom fade scrims disappear), and the scrollbar's `padding-bottom`.

**Failure scenario:** Open the drawer and scroll to the bottom without ever tapping the search field: the last nav item sits behind the 3rem off-screen bleed (only partially rescued by `env(safe-area-inset-bottom)` on notched iPhones — fully clipped elsewhere), and no fade scrims render. Everything self-heals the moment the search input is focused once, which makes it easy to miss in testing.

**Fix:** `var(--drawer-keyboard-inset, 0px)`.

### 2. 🟡 Cmd+K is dead on the website layout at small viewports

**Location:** `docs/src/components/SearchControls.tsx`

```tsx
const visibleTarget =
  getVisibleTarget(desktopTriggerRef.current) ?? getVisibleTarget(mobileTriggerRef.current);
if (!visibleTarget) {
  return;
}
```

On the `(website)` layout, `SearchControls` renders without `mobileNavContent`, and its container is `display: none` below `bp3`. Previously `enableKeyboardShortcut` called `handle.open()` unconditionally, so Cmd+K opened the search dialog even with the trigger hidden. Now both triggers fail `getClientRects().length` and the shortcut does nothing.

**Failure scenario:** iPad in portrait with a hardware keyboard (or a narrow desktop window) on the homepage — Cmd+K silently does nothing where it used to open search.

**Fix:** When no trigger is visible, fall back to opening the desktop dialog (or the drawer when `hasMobileNav`), rather than returning.

### 3. ℹ️ `titleMap` removal changes the desktop side nav label

**Location:** `docs/src/app/(docs)/layout.tsx`

Removing `titleMap` means the desktop `SideNav` now renders "About Base UI" instead of the shortened "About". That's a desktop-visible change in a mobile-nav PR — confirm it's intended (and that the longer label doesn't wrap awkwardly in the sidebar).

## Tests (0)

No findings.

## Simplifications (2)

### 1. 🟡 ~120 lines of search plumbing duplicated between the drawer and the dialog

**Location:** `docs/src/components/MobileNavDrawer.tsx`

`MobileNavDrawer` re-implements, nearly verbatim from `SearchDialog`: the GA search-session tracking (`searchQueryRef`/`resultCountRef`/`attemptRef`/`selectedResultRef`/`lastTrackedQueryRef`, the 1500ms query debounce, and the select/dismiss close event), `SEMVER_PATTERN` + `slugifyWithParentContext`, `normalizeGroup`, and the 400ms empty-results delay. It also declares its own `sitemapPromise` instead of reusing `loadSearchSitemap` — which this same PR added specifically to share the promise and retry after a failed load, so the drawer's copy misses the catch-and-reset behavior and the two entry points don't share the cached import.

**Failure scenario:** The two search surfaces drift — an analytics or slug fix lands in one and not the other, and the duplicated logic is the bulk of the new 475-line file.

**Fix:** Extract a shared `useSearchTracking(ga)` hook plus shared slug/group helpers (e.g. in `Search/`), and use `loadSearchSitemap` in `MobileNavDrawer`.

### 2. ℹ️ `MobileNav.Item` link branches can collapse into a spread

**Location:** `docs/src/components/MobileNav.tsx`

`linkProps` uses camelCase keys (`ariaCurrent`, `dataActive`) that are then manually re-mapped onto two near-identical JSX branches. Building the object with real attribute keys (`'aria-current'`, `'data-active'`, …) lets both branches spread it: `<a {...linkProps} />` vs `<NextLink {...linkProps} scroll={false} onNavigate={handleNavigate} />`, removing ~20 lines.

## Docs (0)

No findings.

## Verdict

**Request changes** — the unitless `calc()` fallback silently disables the drawer's bottom inset, scrims, and scroll padding in its default state (one-character fix).

---

aarongarciah:
@atomiks tackled the feedback points:

- Fixed the drawer inset fallback to `0px`, so the scroll padding/scrims stay valid before the keyboard inset variable is set.
- Restored the shortened `About` label and moved the logic into a shared `getDisplayTitle` helper so the side nav/mobile nav/search stay consistent.
- Fixed the small-viewport `Cmd+K` fallback so it still opens the desktop dialog on website pages, or the drawer when mobile nav is available.
- Reduced the duplicated search plumbing by extracting shared search tracking, result helpers, deferred empty-state handling, and the shared docs search setup.
- Simplified `MobileNav.Item` by using real attribute keys in `linkProps` and spreading them into both link branches.

I also kept the dialog/drawer lazy-loaded, with idle preloading plus immediate request-time loading so the first interaction does not silently no-op while the chunk is loading.

docs/src/components/CodeBlock/CodeBlock.css
```diff
@@ -1,3 +1,5 @@
+@layer all, base, theme, components, utilities;
```
aarongarciah, line 2: Adding this to every component to avoid future issues with async loaded components declaring CSS layers before the first `@layer all, base, theme, components, utilities;` instance and hence causing styles to break.

colmtuite (approved):

atomiks:
The drawer now pauses on release which stops its momentum:

https://github.com/user-attachments/assets/ac08eb60-170e-4273-bf40-ad78a641d028

aarongarciah:
> The drawer now pauses on release which stops its momentum:

Should be fixed now.

PR #5141, opened 2026-06-30
[docs] Adjust cursor for demo playground link and file tabs

atomiks:
## What changed

Two cursor tweaks in the docs demo chrome:

- **"Open in StackBlitz" / "Open in CodeSandbox" button** now uses `cursor: pointer` to indicate it opens in a new tab, despite not being a link.
- **Demo file tabs** (`.DemoTab`) now use `cursor: default`. They're rendered as `<a>` anchors (which default to `pointer`) but behave like in-page tabs, because left-click goes nowhere (only right-click has a special `<a>` affordance).

## Why

The cursor should reflect behavior, not the underlying element: the playground link opens an external playground in a new tab (link-like), while the file tabs switch the displayed source in place (tab-like). The goal is to make the `cursor` style feel less random rather than attempting to strictly adhere to tag/`<a>` conventions

### DIFF
--- a/docs/src/components/Demo/Demo.css
+++ b/docs/src/components/Demo/Demo.css
@@ -14,6 +14,12 @@
     right: 1.125rem;
   }
 
+  /* The external playground link opens a new tab, so use a link-style cursor
+     even though it's rendered as a plain button. */
+  .DemoExternalPlaygroundButton {
+    cursor: pointer;
+  }
+
   .DemoPlayground {
     position: relative;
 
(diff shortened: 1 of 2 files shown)


aarongarciah:
> "Open in StackBlitz" / "Open in CodeSandbox" button now uses cursor: pointer to indicate it opens in a new tab, despite not being a link.
> The cursor should reflect behavior, not the underlying element

Are we sure about this? When I see `cursor:pointer` I expect it to be a link i.e. I expect to be able to Cmd+Click it, right-click it, etc. It's tricky because the "external" icon can confuse some users about it being a link. That's why I prefer to keep `cursor:default` if it's a button.

> Demo file tabs (.DemoTab) now use cursor: default. They're rendered as <a> anchors (which default to pointer) but behave like in-page tabs, because left-click goes nowhere (only right-click has a special <a> affordance).

`cursor:pointer` is the only affordance that lets users know this is a link that can be opened in a new tab. To be honest, if we don't find this feature appealing (I personally don't, since linking to a certain demo is enough IMO), I'd prefer to use buttons instead of links for the tabs.

PR #5143, opened 2026-06-30
[popups] Fix collision padding off-by-one on the biased side

### LINKED ISSUE #5131: [popover] Collision padding top is off by `1px`
# Bug report

Collision padding top is off by `1px`

Image

## Reproducible example

https://stackblitz.com/edit/afvglxcy-ucb3pecu?file=src%2FApp.tsx

## Base UI version

1.6.0

atomiks:
## Summary

Fixes #5131.

`collisionPadding` was being applied 1px farther from the edge than requested on the popup's preferred-side axis (e.g. the top gap measured `6px` for a default `collisionPadding={5}`).

## Cause

A 1px "preferred side" bias was added so that an iOS soft-keyboard-centered input doesn't undesirably flip its popup to the top. That bias was baked directly into the shared `collisionPadding` object, so instead of only nudging `flip()`'s decision, it leaked into **every** middleware that reads `collisionPadding`:

- `shift()` rested the popup 1px farther from the biased edge than requested.
- `size()` computed `--available-height` 1px too small.
- the arrow `limitShift` offset was also slightly off.

## Fix

Normalize `collisionPadding` to the user's raw values, and apply the directional bias **only inside `flip()`**. The `flip()` padding is algebraically unchanged (`collisionPadding + bias + directionalBias`), so the iOS anti-flip behavior is fully preserved — the bias just no longer shifts the resting position.

This lives in the shared `useAnchorPositioning`, so it applies to every anchored component (Popover, Tooltip, Menu, Select, Combobox, Preview Card, etc.), but the user-visible symptom was reported on Popover.

## Verification

I tested the Combobox page in iOS (iPhone) and couldn't see the popup flip to the top by default. Tested every input on the demos page.

- Added a regression test in `PopoverPositioner.test.tsx` asserting the popup rests exactly at `collisionPadding` from the colliding edge (fails at `13` pre-fix, passes at `12` post-fix).
- All 205 positioner tests pass across components.
- Reproduced the off-by-one in the browser and confirmed the fix (`6px → 5px`, `--available-height` `968px → 969px`).

---


flaviendelangle (approved):
# PR review

This is a tightly-scoped positioning fix in the shared `useAnchorPositioning`: the iOS anti-flip 1px bias was baked into the normalized `collisionPadding` object and leaked into `shift()`/`size()`, shifting the resting position and `--available-height` 1px off. The fix normalizes to raw user values and re-adds the bias only inside `flip()`. Nothing is merge-blocking — the `flip()` padding is algebraically unchanged so the anti-flip behavior is preserved, and the fix is well-targeted and tested.

## Bugs (0)

No findings.

I verified the algebra across every consumer of `collisionPadding`:
- `flip()`: old `(user + biasTop) + bias` vs new `user + bias + biasTop` — identical, so the iOS anti-flip is fully preserved and still strictly exceeds the `size()` padding (precedence maintained).
- `shift()` / `size()`: now use raw `user` padding — the intended fix.
- `limitShift`: uses raw padding; correct.

## Tests (0)

No findings.

The added `PopoverPositioner.test.tsx` regression test is well-constructed: it pins the anchor near the bottom to force a flip to the top edge, then asserts the resting `top` equals `collisionPadding` exactly (fails at 13 pre-fix, passes at 12 post-fix). Because the flipped popup collides with the top edge, this also indirectly exercises the corrected `size()`/`shift()` padding. `triggerStyle`, `Trigger`, `isJSDOM`, and `waitFor` are all in scope, and the test is correctly Chromium-only via `it.skipIf(isJSDOM)`.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

ℹ️ Note (PR description only, not the code): the description lists a third symptom — "the arrow `limitShift` offset was also slightly off." That claim doesn't hold. `biasTop`/`biasBottom` are nonzero only when `sideParam` is `top`/`bottom`, but `limitShift`'s `offsetAmount` for a top/bottom (y-axis) placement sums the **cross-axis** `left + right` paddings, where `biasLeft`/`biasRight` are always `0` — and symmetrically for left/right popups. So `limitShift` never carried the directional bias, and its change here is a harmless no-op. Worth correcting in the PR body to avoid implying a behavior change that isn't there; no code action needed.

## Verdict

**Approve** — a minimal, algebraically-verified fix with a targeted regression test; the only note is a minor inaccuracy in the PR description, not the code.

---

PR #5148, opened 2026-06-30
[composite] Support controlled item indexes

### LINKED ISSUE #4657: [Combobox][Autocomplete] Combobox.Item/Autocomplete.Item `index` breaks hover state
# Bug report

## Current behavior

(Written for Autocomplete, applies to Combobox as well though)

When using `Autocomplete.Item` with the documented `index` prop in a non-virtualized `Autocomplete.List`, pointer hover does not update the highlighted item. The `data-highlighted` attribute remains on the previously highlighted item, while keyboard navigation with Arrow Up/Down still works.

This appears to happen because providing `index` makes the item skip normal `CompositeList` DOM registration, but the non-virtualized list still uses `CompositeList` to manage `listRef`. Hover navigation then calls `listRef.current.indexOf(event.currentTarget)`, but the registry is empty/stale, so the hovered item cannot be resolved.

## Expected behavior

Providing `index` to `Autocomplete.Item` should work as documented in non-virtualized lists, or the docs should clarify that `index` is only supported/recommended when `Autocomplete.Root virtualized` is enabled.

Hovering different options should update the active item and move `data-highlighted` accordingly.

## Reproducible example

Minimal shape:

```tsx
<Autocomplete.Root items={items} autoHighlight="always">
  <Autocomplete.Input />
  <Autocomplete.Portal>
    <Autocomplete.Positioner>
      <Autocomplete.Popup>
        <Autocomplete.List>
          {items.map((item, index) => (
            <Autocomplete.Item key={item} value={item} index={index}>
              {item}
            </Autocomplete.Item>
          ))}
        </Autocomplete.List>
      </Autocomplete.Popup>
    </Autocomplete.Positioner>
  </Autocomplete.Portal>
</Autocomplete.Root>
```

Remove `index={index}` and hover highlighting works again.

## Base UI version
v1.4.0

Observed in tests / local reproduction and Chrome.

## Which OS are you using?
macOS

Not applicable.

## Additional context
`Autocomplete.Root virtualized={true}` changes the behavior because `Autocomplete.List` skips `CompositeList`, and items are expected to register themselves by explicit index. In the non-virtualized path, `CompositeList` still owns registration, so passing `index` appears to put the item/list registration mechanisms out of sync.

### LINKED ISSUE #4698: [select] Mouse selection is a no-op in dev StrictMode

jjenzz:
Fixes #4657
Fixes #4698

This updates `useCompositeListItem` so passing an explicit `index` makes the item index controlled without opting the item out of `CompositeList` registration.

Previously, items with an explicit `index` skipped registration entirely. In non-virtualized Combobox/Autocomplete lists, `useListNavigation` still relies on `listRef` to resolve pointer hover targets, so `listRef.current.indexOf(event.currentTarget)` could not find hovered items and `data-highlighted` stayed on the previous item.

The new behavior:
- explicit `index` non-virtualized items still register with `CompositeList` (source-of-truth)
- controlled-index lists skip DOM reorder observation
- mixed controlled/uncontrolled indexes in one `CompositeList` report a development error

Added test coverage for:
- mixed controlled/uncontrolled index warnings
- Combobox hover highlighting with manually indexed items

@atomiks it looks like you've touched this file the most, so would really appreciate your eyes on this one 🙏 lmk if this could have been achieved more simply elsewhere.

| BEFORE | AFTER |
| - | - |
| <video src="https://private-user-images.githubusercontent.com/175330/615272151-200687f9-ccf9-4b7d-a4a0-720a566709d6.mp4" controls="controls" muted="muted" class="d-block rounded-bottom-2 border-top width-fit" style="max-height:640px; min-height: 200px"></video> | <video src="https://github.com/user-attachments/assets/8aa88e41-2b12-4bb3-a6e8-a7b8b44dd081" controls="controls" muted="muted" class="d-block rounded-bottom-2 border-top width-fit" style="max-height:640px; min-height: 200px"></video> |


packages/react/src/internals/composite/list/CompositeListContext.ts
```diff
@@ -1,8 +1,12 @@
 'use client';
 import * as React from 'react';
 
+export type CompositeMetadata<CustomMetadata> = {
+  index?: number | null | undefined;
+} & CustomMetadata;
```
jjenzz, line 7: moved this here to avoid circular imports

packages/react/src/internals/composite/list/CompositeList.tsx
```diff
@@ -98,13 +125,26 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
...
+          elementsRef.current[metadata.index] = node as HTMLElement;
+
+          if (labelsRef) {
+            labelsRef.current[metadata.index] =
+              metadata.label ?? metadata.textRef?.current?.textContent ?? node.textContent ?? null;
+          }
+        }
+      });
+
+      nextIndexRef.current = nextLength;
```
jjenzz: The important part here is that `CompositeList` becomes the source-of-truth re: maintaining `elementsRef`/`labelsRef`.

Previously, each item populated those refs [from its callback ref](https://github.com/mui/base-ui/pull/5148/changes#diff-c56a12aa02e59ddea82b623ac22cf1a3a6a3f796466725002c662b9e68fdf070L55-L71), while `CompositeList` adjusted their length and cleared them on cleanup. That split ownership was fragile because callback refs and layout-effect cleanups can run in different phases, especially in React 18 dev/StrictMode. 

With this change, items only register/unregister their node and metadata. `CompositeList` then builds `elementsRef` and `labelsRef` from its `sortedMap` so that ordering, sparse controlled indexes, labels, cleanup, and lengths all come from the same source of truth.

atomiks: This PR's design direction makes sense to me. Curious if this would fix https://github.com/mui/base-ui/issues/4698 as well.

Perf-wise: `Array.from({ length: maxIndex + 1 })` allocates a dense array up to the largest index on every tick. Keeping the arrays sparse is strictly cheaper (and allows deletion of `getMaxIndexLength`) with identical behavior:

```diff
   useIsoLayoutEffect(() => {
     const shouldUpdateLengths = lastTickRef.current === mapTick;
     if (shouldUpdateLengths) {
-      const isControlled = controlledRef.current === true;
-      const nextLength = isControlled ? getMaxIndexLength(sortedMap) : sortedMap.size;
-      elementsRef.current = Array.from({ length: nextLength });
-
-      if (labelsRef) {
-        labelsRef.current = Array.from({ length: nextLength });
-      }
+      // Rebuild from the sorted map, but keep the arrays sparse so a large
+      // controlled index doesn't allocate a dense array up to it.
+      elementsRef.current = [];
+      if (labelsRef) {
+        labelsRef.current = [];
+      }

       sortedMap.forEach((metadata, node) => {
         if (metadata.index != null) {
           elementsRef.current[metadata.index] = node as HTMLElement;

           if (labelsRef) {
             labelsRef.current[metadata.index] =
               metadata.label ?? metadata.textRef?.current?.textContent ?? node.textContent ?? null;
           }
         }
       });

-      nextIndexRef.current = nextLength;
+      nextIndexRef.current = elementsRef.current.length;
     }

     onMapChange(sortedMap);
   }, [onMapChange, sortedMap, elementsRef, labelsRef, mapTick]);
```

```diff
-// Controlled indexes can be sparse if, for example, an item in the middle suspends.
-// So, we might receive [0, 2] and length should be 3.
-function getMaxIndexLength<Metadata>(map: Map<Element, CompositeMetadata<Metadata>>) {
-  let maxIndex = -1;
-
-  map.forEach((metadata) => {
-    const index = metadata.index;
-    if (index != null && index > maxIndex) {
-      maxIndex = index;
-    }
-  });
-
-  return maxIndex + 1;
-}
```

atomiks: As for the R18 failures, it seems like `ref` callback still needs to fill the refs synchronously as a fallback?

jjenzz: oooooh, i really didn't like the `getMaxIndexLength` thing (thx codex lol) and dno why i didn't think to do that  🙈 much simpler, thanks!

packages/react/src/internals/composite/list/CompositeList.tsx
```diff
@@ -93,48 +121,34 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
...
       }
     };
-  }, [labelsRef]);
+  }, [elementsRef, labelsRef]);
 
   const subscribeMapChange = useStableCallback((fn) => {
     listeners.add(fn);
+    // Delayed items need the latest snapshot when they subscribe, otherwise
+    // they can stay on a guessed index until another list change happens.
+    fn(sortedMapRef.current);
```
jjenzz, line 192: The `OTPField` highlighted an edge case where an item could keep its guessed index because items subscribe to map changes _after_ registering themselves, and this PR now produces the sorted map during registering (before the subscription). 

AFAIUI, this wasn't a problem before because map listeners were notified after the `mapTick` state update, from within a layout effect tied to that state change, so index correction happened later in the render/effect cycle. The old callback ref writes were covering up that correction delay.

Now that the sorted map and refs are produced during registration,`elementsRef`/`labelsRef` are ready without waiting for that state/effect cycle. That means listener notification also happens earlier, so if a child subscribes _after_ registering it wouldn't get the correction snapshot. 

Long story short, we probably want to make sure we subscribe before we register, but this fix seems more correct/robust.

atomiks:
# PR review

The controlled-index design is sound and fixes #4657 the right way (registering controlled items instead of bypassing the list), and I measured no perf cost — local 500-item typing benchmarks are within noise of master, with the narrowing scenario actually dropping from 17 to 15 renders. However, the rewrite moved the flush trigger from `register`/`unregister` (which used to schedule state) into `CompositeList`'s own layout effect, and that effect only runs when `CompositeList` itself re-renders — I confirmed a merge-blocking regression with a failing repro: items mounted by a state update in a deep subtree are never indexed.

## Bugs (2)

### 1. 🔴 Items mounted from a deep-subtree state update are never registered into the list

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:41`

```ts
const register = useStableCallback((node: Element, metadata: CompositeMetadata<Metadata>) => {
  // ...
  map.set(node, metadata);
  isDirtyRef.current = true;
});
```

On master, `register`/`unregister` call `setMapTick`, so any item mounting anywhere re-renders `CompositeList` and re-syncs. In this PR they only set `isDirtyRef`, and the flush lives in a layout effect that runs **only when `CompositeList` re-renders**. The MutationObserver can't cover it either: additions produce `diff.size !== 0`, which deliberately skips `flush()` on the assumption that registration handles adds/removes. So a commit that mounts or unmounts an item **without re-rendering `CompositeList`** — local state in a subcomponent inside the popup, a memoized subtree — leaves `elementsRef` stale and the new item at index `-1` (its immediate `subscribeMapChange` snapshot predates the registration) until an unrelated re-render happens to flush.

**Failure scenario:** confirmed with a repro test that passes on master and fails on this branch:

```tsx
function DeepSection() {
  const [extra, setExtra] = React.useState(false);
  addItem = () => setExtra(true);
  return extra ? <Item label="extra" /> : null;
}

await render(
  <CompositeList elementsRef={elementsRef}>
    <Item label="a" />
    <DeepSection />
    <Item label="b" />
  </CompositeList>,
);

act(() => addItem!());
expect(elementsRef.current).toHaveLength(3); // PR: still 2; item "extra" keeps index -1
```

In product terms: a menu section that expands via its own state (`{open && <Menu.Item/>}`) mounts items that keyboard navigation, typeahead, and hover resolution all skip.

**Fix:** keep the new flush pipeline but restore a wake-up from registration — e.g. `setMapTick((t) => t + 1)` in `register`/`unregister` (React batches it per commit, and the dirty-check effect still coalesces the work), or a `queueMicrotask` fallback that flushes when `isDirtyRef` is still set after the commit. Please also add the repro above as a regression test.

### 2. ℹ️ `label: null` now falls through to `textContent`, losing the typeahead-exclusion semantic

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx` (syncRefs)

Old code wrote labels with `label !== undefined ? label : (textRef?.current?.textContent ?? node.textContent)`, so an explicit `label={null}` stored `null` — which `useTypeahead` treats as "skip this item". The new `metadata.label ?? metadata.textRef?.current?.textContent ?? node.textContent ?? null` coalesces `null` away and substitutes the text content. No first-party component passes `null` today (public `label` props are `string | undefined`), but `useCompositeListItem`'s param type is `string | null | undefined` and the hook is exposed via `/internals`, so the exclusion semantic is silently lost for anyone using it. Preserving the `!== undefined` check keeps both behaviors.

## Tests (1)

### 1. 🟡 The second mixed-index warning assertion is vacuous — `error()` dedupes by message

**Location:** `packages/react/src/internals/composite/list/CompositeList.test.tsx`

```tsx
await render(/* controlled first */);
expect(console.error).toHaveBeenCalledWith(expectedMessage);

await render(/* uncontrolled first */);
expect(console.error).toHaveBeenCalledWith(expectedMessage);
```

`error()` from `@base-ui/utils/error` deduplicates by message string, and the test setup only calls its `reset()` **between tests** (`test/setupVitest.ts`). Within this single test, the second render's warning is swallowed by the dedup set, and the second `toHaveBeenCalledWith` passes via the *first* call — the uncontrolled-then-controlled ordering is never actually verified. Splitting into two tests (so `reset()` runs in between) or asserting `toHaveBeenCalledTimes` makes the coverage real.

## Simplifications (1)

### 1. 🟡 `ComboboxItem`'s parallel `listRef` bookkeeping is now redundant for non-virtualized controlled items

**Location:** `packages/react/src/combobox/item/ComboboxItem.tsx:84`

```ts
const shouldRun = hasRegistered && (virtualized || indexProp != null);
// ...
list[index] = itemRef.current;
```

This effect was the registration mechanism for controlled items back when they bypassed `CompositeList` — the wipe-out interaction between it and `CompositeList`'s length-truncation is the actual root cause of #4657. With this PR, non-virtualized controlled items register normally and `syncRefs` rebuilds `elementsRef` wholesale (replacing the array identity, which also makes this effect's cleanup operate on a dead array). The `indexProp != null` arm can be dropped, leaving `virtualized` only (where `CompositeList` isn't rendered and the effect is still load-bearing). Two writers for the same slot is exactly the kind of coupling that made the original bug hard to see.

## Docs (1)

### 1. 🟡 The mixed-index error links to React's controlled-inputs page and misstates when the mode is fixed

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx` (register)

```ts
'More info: https://fb.me/react-controlled-components',
```

That shortlink targets React's controlled *form inputs* docs — unrelated to composite item indexes, and borrowed from the classic MUI form-state warning. The repo's error guidelines call for a Base UI docs link when applicable; here I'd drop the link entirely (or point at the Combobox `index` prop docs). Also, "The nature of the state is determined during the first render" isn't quite what the code does: it's determined by the **first registered item** and resets whenever the list empties. Worth rewording so the message matches the mechanism.

## Verdict

**Request changes** — the deep-subtree registration regression (Bugs No. 1) is confirmed by a failing repro against this branch that passes on master.

---

packages/react/src/internals/composite/list/useCompositeListItem.ts
```diff
@@ -32,58 +32,44 @@ export function useCompositeListItem<Metadata>(
...
+        textRef,
+      };
+      register(node, registeredMetadata);
       return () => {
         unregister(node);
       };
     }
     return undefined;
-  }, [externalIndex, register, unregister, metadata]);
+  }, [externalIndex, register, unregister, metadata, label, textRef]);
```
atomiks: 🟡 **A label change now costs a full list flush instead of one array write**

`label` is now a dependency of the register effect, so any item whose `label` changes unregisters and re-registers, which dirties the list and triggers `flush()`: an O(n log n) `compareDocumentPosition` sort, an O(n) `syncRefs` rebuild including a `textContent` DOM walk per item, N listener callbacks, a parent re-render, and a MutationObserver rebuild. On master, a label change was handled inside the item ref callback with a single `labelsRef.current[index] = label` write — O(1), no sort, no render.

**Failure scenario:** a 1,000-item Menu/Select with live-updating labels (localized or data-driven text) performs a full DOM-position sort plus 1,000 `textContent` reads on every label update — visible jank on low-end devices.

**Suggestion:** keep `label`/`textRef` out of the register-effect deps and publish label changes through a cheap path (e.g. store the label in a ref read by `syncRefs`, or write `labelsRef` directly when only the label changed).

packages/react/src/internals/composite/list/CompositeList.tsx
```diff
@@ -146,13 +162,38 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
...
   return new Set<Function>();
 }
 
+function sortMap<Metadata>(map: Map<Element, CompositeMetadata<Metadata>>, isControlled: boolean) {
+  const newMap = new Map<Element, CompositeMetadata<Metadata>>();
+
+  if (isControlled) {
+    const sortedNodes = Array.from(map.entries())
+      .filter(([node]) => node.isConnected)
+      .sort((a, b) => (a[1]?.index ?? 0) - (b[1]?.index ?? 0));
```
atomiks: ℹ️ **Duplicate or negative controlled indexes are silently accepted**

The controlled branch never validates indexes: two items passing the same `index` collide in `elementsRef.current[metadata.index]` (last write wins, the other item becomes invisible to keyboard navigation and hover resolution), and a negative index is dropped from the array while still sorting first. The PR adds a dev error for mixed modes but nothing for this adjacent, equally likely misuse of the new API.

**Failure scenario:** a consumer maps filtered items but passes the unfiltered array index with an off-by-one — one item silently unreachable via keyboard, no warning.

**Suggestion:** in dev, warn when a registration would overwrite an existing controlled index (or when `index < 0`).

packages/react/src/internals/composite/list/CompositeList.tsx
```diff
@@ -93,48 +121,36 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
     return () => {
       mutationObserver.disconnect();
     };
-  }, [sortedMap]);
+  }, [flush, mapTick]);
 
+  // React runs parent layout effects after child layout effects, so all item
+  // registrations from this commit are already in `map`. We flush here to
+  // make sure elementsRef/labelsRef updates are not coupled to an extra render.
   useIsoLayoutEffect(() => {
```
atomiks, line 174: ℹ️ **`elementsRef`/`labelsRef` are now populated later in the commit than before**

On master, each item wrote itself into `elementsRef.current[index]` inside its ref callback, i.e. at ref-attach time before any layout effect ran. Now the arrays are only rebuilt in this layout effect, which runs after all item-level layout effects (and, for items mounting from a nested state update, only in the follow-up render that `setMapTick` schedules — still pre-paint). I found no in-repo consumer that reads `listRef` from an item-level layout effect on mount (floating-ui `useListNavigation` reads it above `CompositeList`, after flush), so this is a heads-up for external/floating-ui-internal consumers rather than a named breakage — but it is a real narrowing of when the arrays are trustworthy within a commit.

**Suggestion:** a short comment documenting that the refs are valid only from the `CompositeList` layout effect onward would prevent surprises.

packages/react/src/internals/composite/list/CompositeList.tsx
```diff
@@ -20,54 +17,86 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
...
+        elementsRef.current[metadata.index] = node as HTMLElement;
 
-    sortedNodes.forEach((node, index) => {
-      const metadata = map.get(node) ?? ({} as CompositeMetadata<Metadata>);
-      newMap.set(node, { ...metadata, index });
+        if (labelsRef) {
+          labelsRef.current[metadata.index] =
+            metadata.label !== undefined
+              ? metadata.label
+              : (metadata.textRef?.current?.textContent ?? node.textContent ?? null);
```
atomiks: 🟡 **`syncRefs` re-reads `textContent` for every item on every flush**

Every flush walks the DOM text of all N items, where master read `textContent` only in the ref callback of items whose index/label actually changed. In a JSX-children Combobox (which passes `labelsRef`) or an open 1,000-item Select, each filter keystroke or single item add/remove now performs ~N subtree text walks (~0.1–1ms at N=1000) on top of the pre-existing sort. Mitigations already exist (`label` prop, `textRef` narrowing, lists without `labelsRef` skip this entirely), so non-blocking — but this is the line most likely to show up in a large-list profile.

**Suggestion:** cache the resolved label per node at registration and only re-read `textContent` for newly registered nodes.

packages/react/src/internals/composite/list/CompositeList.tsx
```diff
@@ -20,54 +17,86 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
...
 
-    return newMap;
-  }, [map, mapTick]);
+    nextIndexRef.current = elementsRef.current.length;
+  });
+
+  const flush = useStableCallback(() => {
+    const nextSortedMap = sortMap(map, controlledRef.current === true);
+    syncRefs(nextSortedMap);
+    setMapTick((tick) => tick + 1);
```
atomiks: ℹ️ **Item-local re-registrations cause two parent renders and two observer rebuilds**

`flush()` bumps the tick unconditionally. In a commit that does not include `CompositeList` (an item re-registering from local state), the `register` bump causes render A, whose dirty-check effect flushes and bumps again, causing render B — and the MutationObserver effect (deps `[flush, mapTick]`) is torn down and rebuilt on both, the first time from a stale `sortedMapRef`. Master did one render and one rebuild. Both extra renders bail out cheaply (stable `children` element, memoized context), so this is only a note.

**Suggestion:** skip the tick bump in `flush()` when the registration bump already scheduled this render.

packages/react/src/internals/composite/list/CompositeList.tsx
jjenzz: >  Now the arrays are only rebuilt in this layout effect, which runs after all item-level layout effects (and, for items mounting from a nested state update, only in the follow-up render that `setMapTick` schedules — still pre-paint)

ah yeah, adding `setMapTick` to `register` to fix [this bug](https://github.com/mui/base-ui/pull/5148#issuecomment-4861021313) will have killed some of the perf gains i added haha. okay, back to the ref callback approach for that then 🙈

packages/react/src/internals/composite/list/CompositeList.tsx
jjenzz: good point, they could produce broken `elementsRef` entries but this PR doesn't make invalid user provided indexes more likely than before (it's not a new API). sounds like a separate discussion around whether we want to provide warnings/errors for invalid `index` props?

packages/react/src/internals/composite/list/CompositeList.tsx
jjenzz: oops good catch, forgot to remove that one! not needed now

packages/react/src/internals/composite/list/CompositeList.tsx
jjenzz: okay no, the ref callback brings two sources-of-truth again which i'm trying to avoid, because i'm getting diff bug reports now when running `/base-ui-review`. gonna implement another solution there.

packages/react/src/internals/composite/list/CompositeList.tsx
atomiks: Yeah imo, this is not worth handling.

atomiks:
# Performance: measured, one real inefficiency found

Benchmarked `CompositeList` directly in headless Chromium (1000 items, `flushSync` mount/update, median of 12 rounds), on this PR's head vs `master`:

| Scenario (1000 items) | master | PR #5148 | PR + guard |
|---|---|---|---|
| Mount, uncontrolled | 26.0 ms | 26.7 ms | **6.4 ms** |
| Mount, controlled (`index` prop) | 4.8 ms † | 23.6 ms | **4.1 ms** |
| Replace half the items, uncontrolled | 47.0 ms | 47.6 ms | **26.6 ms** |
| Replace half the items, controlled | 4.6 ms † | 47.6 ms | **5.6 ms** |
| Filter 1000→500, controlled (all indexes reassigned) | 2.4 ms † | 36.6 ms | **3.3 ms** |
| Filter 1000→500, uncontrolled | 23.7 ms | 24.5 ms | **12.3 ms** |

† master's controlled numbers are the skip-registration path — fast because it did nothing, which is exactly the bug being fixed.

Reading the middle column: the PR is **perf-neutral for uncontrolled lists** (the common case), but large controlled lists pay real new cost — a filter keystroke on a 1000-item indexed Combobox costs ~34 ms more than master, per keystroke.

**The cost is not the sort — it's the state-update enqueue.** Both `register` and `unregister` call `setMapTick` once per item, so a 1000-item commit enqueues 1000 updates that React coalesces into one render anyway. Guarding the enqueue on the dirty-flag transition eliminates it:

```tsx
map.set(node, metadata);
if (!isDirtyRef.current) {
  isDirtyRef.current = true;
  setMapTick((tick) => tick + 1);
}
```

(and the same in `unregister`). This is safe by construction: every false→true dirty transition enqueues exactly one tick, and the only thing that clears the flag is `flush()`, which is guaranteed to run afterwards — either in the same commit's parent layout effect or in the tick-triggered render. With this guard, **all 2679 tests across the eleven affected suites pass**, and every scenario becomes faster than master — including the previously "free" controlled paths (4.1 vs 4.8 ms mount) and a 4× improvement on uncontrolled mounts, since master paid the same per-register enqueue tax.

# Minimality assessment

The architecture is close to minimal for what it has to preserve — I looked for structural cuts and found none that survive scrutiny: the `mapTick` state is the only mechanism that gets `CompositeList` to re-run its flush effect after child registrations (a dep-less effect wouldn't fire without a re-render); `sortedMapRef` is needed so late subscribers and the MutationObserver path get a current snapshot without a render; `isDirtyRef` is what lets one flush serve n registrations. The head commit already removed the redundant tick from `flush`. Remaining trims are cosmetic:

- `sortMap`'s controlled comparator `(a[1]?.index ?? 0)` — metadata can no longer be `null`/`undefined` in the map (the type changed in this PR), so the optional chaining and fallback are dead defensiveness.
- The `ComboboxItem` virtualized manual `listRef` effect now overlaps with registration-driven `syncRefs` for items that pass `index` (both write the same node into the same slot; the registration flush runs after the manual write in the same commit) — consolidating to a single writer is the one real simplification available.

The single actionable outcome of this pass is the `setMapTick` guard: a two-site, behavior-preserving change that turns the PR's only measurable regression into an across-the-board win over master.

---

packages/react/src/internals/composite/list/CompositeList.tsx
```diff
@@ -20,54 +17,89 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
...
-  // information for list navigation.
-
+  // `mapTick` is our re-render trigger mechanism.
   const map = useRefWithInit(createMap<Metadata>).current;
   // `mapTick` uses a counter rather than objects for low precision-loss risk and better memory efficiency
   const [mapTick, setMapTick] = React.useState(0);
-  const lastTickRef = React.useRef(mapTick);
+  // The sorted map is kept in a ref so registration flushes can publish the
+  // current ordering immediately.
+  const sortedMapRef = React.useRef(createMap<Metadata>());
```
atomiks: Seems like this should use `useRefWithInit` like `map` above it

packages/react/src/select/root/SelectRoot.test.tsx
```diff
@@ -4799,6 +4799,7 @@ describe('<Select.Root />', () => {
     it('does not jump matches after a closed-state value reset', async () => {
       function App() {
         const [value, setValue] = React.useState<string | null>('dog');
+
```
atomiks: Dead diff

atomiks:
# Second pass: virtualized regressions

Follow-up hunt on the current head — the earlier findings (deep-subtree registration, per-register tick cost, `label: null`, vacuous warning tests) are all resolved now. This pass focused on the newly-changed virtualized path, since removing the `if (virtualized) return element` early-return means virtualized lists mount a `CompositeList` for the first time. Found one merge-blocking regression there plus a few smaller things.

## 1. 🔴 Virtualized: `listRef.current.length` is truncated to the rendered window, breaking keyboard bounds

On master, virtualized mode renders no `CompositeList`; items write `store.state.listRef.current[index] = el` directly (and `delete` on cleanup never shrinks `length`), while `AriaCombobox` maintains the invariant `listRef.current.length = flatFilteredItems.length` in a layout effect keyed on `[items, flatFilteredItems]`.

With this PR, every flush runs `syncRefs`, which does `elementsRef.current = []` and sparse-fills only the currently rendered window — the array is **replaced** with one whose `length` is `maxRenderedIndex + 1`, not the filtered total. The `AriaCombobox` fixup effect doesn't re-run (its deps haven't changed), and even when it does run in the same commit, the next window shift (scroll or key-repeat arrowing → items re-register with a new `index` → flush in the tick-triggered render) clobbers it again with no counter-fixup.

`useListNavigation` computes bounds from that length on every keydown (`getMaxListIndex` → `length - 1`; the combobox passes `disabledIndices: EMPTY_ARRAY`, so holes aren't skipped):

- `loopFocus` (default `true`) + `allowEscape` (true whenever `autoHighlight` is off): at `currentIndex >= maxIndex` the highlight **escapes to `-1` in the middle of the list** — reachable with key-repeat ArrowDown, which outruns the virtualizer's scroll-triggered re-window.
- With `autoHighlight` (no escape): **wraps to the top mid-list** instead of advancing.
- Trigger-based (non-typeable) virtualized combobox: **`End` jumps to the last *rendered* item**, not the last filtered item.

The new test `registers items at their filtered indexes when the rendered window starts after index 0` actually asserts the broken invariant: snapshot `'||three|four'` is length 4 while `flatFilteredItems.length` is 5 — pre-PR that snapshot would be `'||three|four|'`. JSDOM tests can't see the fallout because they don't exercise virtualizer window shifts.

**Fix direction:** either `syncRefs` shouldn't shrink/replace the array for controlled lists (mutate in place, never reduce `length`, leaving total-length ownership with `AriaCombobox`), or `ComboboxList` should re-assert `listRef.current.length = flatFilteredItems.length` after each flush (it already has `filteredItems` in scope and could pass an `onMapChange`).

## 2. 🟡 Virtualized items missing from the filter register with controlled index `-1`

`ComboboxItem.tsx`:

```ts
const index = indexProp ?? (virtualized ? (indexFromFilter ?? -1) : undefined);
```

`findItemIndex` returns `-1` (never `undefined`) when a rendered item isn't in `flatFilteredItems` — a normal transient state while a virtualizer's window lags a keystroke, so the `?? -1` arm is dead and `-1` flows through as a *controlled* index. On master these items were excluded (`hasRegistered` gated the manual write). Now they land in the sorted map (sorted first), and `syncRefs` executes `elementsRef.current[-1] = node` / `labelsRef.current[-1] = …` — junk string-keyed properties holding node references. If a whole stale window misses the filter, that flush produces an essentially empty `listRef` (length 0) for the commit — which also feeds regression 1. A `metadata.index >= 0` guard in `syncRefs` (or skipping registration of `-1` at the call site) closes it.

## 3. 🟡 MutationObserver is never re-created after an observer-triggered flush (uncontrolled lists)

On master, the MO callback called `setMapTick` → re-render → new `sortedMap` → the observer effect re-ran and re-observed the current parents. Now the MO callback calls `flush()` directly, which updates `sortedMapRef` but never bumps `mapTick`, and the observer effect only re-runs on `mapTick`. After a reorder that moves a node to a *different* parent (DnD libraries, manual DOM re-parenting — exactly the no-React-commit scenario this observer exists for), the new parent is never observed, so subsequent reorders under it go undetected until an unrelated register/unregister. Edge-case, but a strict capability loss vs master's re-observe-every-flush.

## 4. ℹ️ Stale type on `subscribeMapChange`

`CompositeListContextValue.subscribeMapChange` is still typed `(map: Map<Element, Metadata | null>) => void`, but listeners now always receive `Map<Element, CompositeMetadata<Metadata>>` — values are never `null` after this PR, and `index`/`labelRef` are what the subscriber in `useCompositeListItem` actually reads. Cosmetic, but it hides type errors in listener code.

## Checked and clear

- **Flush timing** (refs filled in `CompositeList`'s layout effect instead of item ref callbacks): every descendant reader checks out — `SelectPopup`'s `listRef.current[0]` read waits on `isPositioned` (a post-flush commit), `useCompositeRoot.onMapChange` runs inside the flush, and event handlers run long after.
- **`onMapChange` consumers** (Tabs, Toolbar, Slider, OTPField, SelectPositioner): all size-guarded or idempotent against the new call pattern (no more initial empty-map call; unregister-to-empty still flushes).
- **Mode transitions:** `controlledRef` resets correctly when the map empties (cleanups run before creates within a commit), and StrictMode double-invoke is safe.
- **Label semantics:** `label={null}` typeahead-exclusion survives, and label-change-without-reregister works via the `labelRef` indirection.
- **Registration churn:** the register effect's deps (`metadata`, `externalIndex`) have the same identity requirements as master — no new infinite-render risk.

No. 1 is the one I'd block on — it likely decides what `syncRefs`'s ownership contract over `elementsRef` should be.

jjenzz (commented):
# PR review

No findings.

I reviewed the branch changes against `master`, including the composite index lifecycle, controlled sparse indexes, downstream tabs/toolbar/slider/select/combobox usage, the submenu hover timeout guard, tests, docs, and config changes. Focused verification passed with Node 22.18.0 via Corepack pnpm 11.5.2: `CompositeList` and `MenuRoot` jsdom test runs were green.

## Verdict

**Approve** — no actionable regressions or correctness issues found.

---

🤖 Review generated with Codex

jjenzz:
@atomiks GPT-5.5 High on medium effort was happy but, pretty sure Fable won't be 😅 I've run out of API usage there though, would you mind giving it a spin for me plz?

atomiks:
# PR review

The core redesign is sound: registration-time flush ordering is correct (child layout effects registering before the parent's `flush`), the coalesced `isDirtyRef`/`mapTick` mechanism cannot miss a flush, and the wholesale `syncRefs` rebuild fixes real staleness the old optimistic ref-callback writes had. Nothing is merge-blocking; the findings below are robustness gaps at the edges of the new controlled mode.

## Bugs (3)

### 1. 🟡 Indexing mode is locked forever, so a legitimate full remount into the other mode silently breaks in production

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:21`

```tsx
// Whether item indexes are controlled (`index` props) or derived from DOM order. The mode is
// intentionally fixed to the first registered item and never reset so that during development
// any runtime mode switch surfaces an uncontrolled/controlled mixing error.
const controlledRef = React.useRef<boolean | null>(null);
```

The lock is intentional for catching *mixing*, but it also covers the non-mixing case where one item population fully unregisters and a new population registers in the opposite mode (e.g. a consumer that passes `index` only under some condition, without remounting the list). Once locked `controlled`, later uncontrolled registrations have `metadata.index === undefined`, so `sortMap`'s controlled branch never assigns indexes and `syncRefs` skips every item — `elementsRef`/`labelsRef` stay empty, keyboard navigation, hover highlighting, and typeahead all go dead. In the old code this swap worked (index-less items simply registered in DOM-order mode). In development you get one warning (and `error()` dedupes by its static message, so only the first offending list in a session logs it); in production the list is silently inert.

**Failure scenario:** An app renders `<Combobox.Item index={i}>` for a large dataset and plain `<Combobox.Item>` when a smaller dataset loads into the same open list — after the swap, no item can be highlighted or selected in a production build.

**Fix:** In `unregister`, reset `controlledRef.current = null` when `map.size` drops to 0. Genuine mixing within one population still errors; a wholesale swap re-derives the mode.

### 2. 🟡 `syncRefs` replaces the `elementsRef` array, dropping the externally maintained `listRef` length for partially rendered lists

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:65`

```tsx
elementsRef.current = [];
...
nextIndexRef.current = elementsRef.current.length;
```

`AriaCombobox` maintains an invariant on the same array: `listRef.current.length = flatFilteredItems.length` (`AriaCombobox.tsx:841`), but its effect only re-runs when `flatFilteredItems` changes. Previously, explicit-`index` items never registered, so `CompositeList` never touched the array after mount and the length stuck. Now every register/unregister flush replaces the array with one whose length is `maxRegisteredIndex + 1`. For a non-virtualized combobox with `items` that renders only a window of indexed items (the configuration `index` exists for, minus the `virtualized` flag — with `virtualized`, `ComboboxList` renders no `CompositeList` and this path is bypassed), moving the window re-registers items and shrinks `listRef.current.length` from the total item count to the window bound, clamping `useListNavigation`'s range.

**Failure scenario:** A combobox with 1,000 filtered items renders indexed items 0–20 without `virtualized`; after any window change, ArrowDown navigation is bounded at the last rendered index instead of index 999.

**Fix:** Mutate in place — clear and refill the existing array (or restore its prior `length` when the controlled max index is lower) instead of assigning a fresh array, matching the old identity-preserving behavior.

### 3. ℹ️ Duplicate controlled indexes silently collide

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:70`

```tsx
nextMap.forEach((metadata, node) => {
  if (metadata.index != null && metadata.index > -1) {
    elementsRef.current[metadata.index] = node;
```

Two items registering the same `index` (an easy off-by-one in user mapping code) last-write-wins in `elementsRef`, and both items report `highlighted` for the same `activeIndex`, plus a duplicate DOM `id` from `${rootId}-${index}`. The PR adds a dev error for mixing but nothing for duplicates.

**Failure scenario:** A user maps two item groups both starting at index 0; the first group's items can never be highlighted, with no diagnostic.

**Fix:** In the flush (dev only), warn when a controlled index is seen twice.

## Tests (1)

### 1. 🟡 The "controlled lists skip DOM reorder observation" behavior is untested

**Location:** `packages/react/src/internals/composite/list/CompositeList.test.tsx:287`

The PR description lists three behaviors; two are pinned (registration with explicit indexes, mixing warnings), but nothing asserts that a controlled list ignores DOM reorders — i.e. that after `list.insertBefore(...)` moves controlled items, their indexes still follow the `index` props rather than document order (mirroring the existing "updates indexes when keyed groups reorder" test, inverted). A future refactor could silently re-enable the observer for controlled lists — or break the early return — without any test failing.

**Failure scenario:** The `isControlled` early-return in the observer effect is lost in a refactor; controlled indexes get clobbered by document order and no test catches it.

**Fix:** Add a test that reorders controlled items in the DOM and asserts `elementsRef` still reflects the `index` props.

## Simplifications (0)

No findings.

## Docs (1)

### 1. ℹ️ Stale and missing JSDoc for the new controlled-index contract

**Location:** `packages/react/src/internals/composite/list/useCompositeListItem.ts:27`

```ts
/**
 * Used to register a list item and its index (DOM position) in the `CompositeList`.
 */
```

"(DOM position)" now describes only the uncontrolled mode. The `index` parameter in `UseCompositeListItemParameters` is also undocumented, and the public `ComboboxItem.index` JSDoc ("Improves performance when specified by avoiding the need to calculate the index automatically from the DOM", `ComboboxItem.tsx:310`) doesn't mention the new all-or-nothing constraint the dev error enforces.

**Failure scenario:** A user passes `index` to only the items they consider expensive, gets a console error (or silent prod misbehavior) and finds no documented rule explaining it.

**Fix:** Update the hook JSDoc to cover controlled indexes, and add a sentence to the `index` prop docs: either all items in a list receive `index` or none do (then run `pnpm docs:api`).

## Verdict

**Approve after nits** — the redesign is correct on the main paths and well tested, but the permanent mode lock and the `elementsRef` identity change each regress an edge case the old code handled.

---

atomiks:
# PR review

This is a well-executed rewrite of `CompositeList`'s registration/flush model so that an explicit `index` makes an item's index *controlled* while still registering it (fixing the broken pointer-hover highlighting on manually-indexed lists in issue 4657). **Nothing is merge-blocking** — no 🔴. All affected suites pass (Combobox 716, Tabs 211, Toolbar 129, Slider 368, Select 492, Menu 956, OTP 156, Accordion 113), typecheck and eslint are clean, and a dedicated core-mechanism sweep found no lost registrations, stale-ref, or ordering regressions. The residual risk is concentrated in the new `itemCount` contract and the mode latch: both are latent edge cases that fail *silently* rather than loudly, plus a handful of test/doc gaps.

## Bugs (4)

### 1. 🟡 `itemCount` is gated on `hasItems`, so an externally-filtered virtualized list can't navigate past the rendered window

**Location:** `packages/react/src/combobox/list/ComboboxList.tsx:127`, `packages/react/src/combobox/root/AriaCombobox.tsx:177`

```tsx
itemCount={hasItems ? flatFilteredItems.length : undefined}   // hasItems = items !== undefined
```

The deleted `AriaCombobox` line `listRef.current.length = flatFilteredItems.length` ran regardless of the `items` prop. Its replacement, `itemCount`, is only fed when `items` is passed. When `itemCount == null`, `syncRefs` lets `elementsRef.current.length` collapse to `highestRenderedIndex + 1` instead of the logical total.

**Failure scenario:** a virtualized combobox that drives its window from `filteredItems` via external `useFilter()` **without** also passing `items` renders logical indices 900–910 with controlled `index` props; `store.state.listRef.current.length` becomes ~911 instead of 1000, and `useListNavigation`'s `getMaxListIndex`/`listRef.current.length` reads make Home/End and arrow-past-window unreachable. The `filteredItems` prop is documented to be used *alongside* `items` (in which case this is fine), so this only bites the `filteredItems`-without-`items` config — but the implicit gate deserves a deliberate decision or a comment rather than silently depending on that combination.

**Fix:** derive `itemCount` from the total filtered length whenever the list is virtualized/windowed (not only when `items` is set), or document + assert that controlled-index windowing requires `items`.

### 2. 🟡 The controlled/uncontrolled mode is latched for the list's lifetime and the mixing guard is a prod no-op

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:36`

```js
if (controlledRef.current === null) {
  controlledRef.current = isControlled;
} else if (controlledRef.current !== isControlled) {
  error(`A CompositeList is mixing controlled and uncontrolled indexes.` /* … */);
}
```

`controlledRef` is fixed to the first-registered item and never reset; `error()` is a no-op under `NODE_ENV === 'production'`. Because `ComboboxList` now wraps children in one persistent `CompositeList` (the old `if (virtualized) return element` early-return is gone), the instance survives a runtime mode switch.

**Failure scenario:** start a combobox non-virtualized (`controlledRef = false`), then toggle `virtualized` at runtime. Items remount and re-register as controlled, but the latch stays `false`, the dev error doesn't fire in prod, and `sortMap(map, false)` re-derives indexes by document position — discarding the intended filtered indexes so navigation/selection targets the wrong element. Same corruption on an empty→refill with the opposite mode. `virtualized` is documented as stable-per-instance, so this is edge/unsupported — but it degrades silently. Consider resetting the latch when the list empties, or keeping the mixing check outside the prod-stripped path for this specific corruption. Relatedly, the message names the internal `CompositeList`, which an app author using `Combobox.Item` can't act on — naming the consumer component would make it diagnosable.

### 3. 🟡 `itemCount` going number→`undefined` never shrinks `elementsRef.length` back

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:111-121`

```js
if (itemCount == null) {
  return;                          // ← bails before un-applying a previous length
}
if (elementsRef.current.length !== itemCount) {
  elementsRef.current.length = itemCount;
}
```

The standalone `itemCount` effect early-returns on `null`, and the flush effect (`[mapTick, flush]`, gated on `isDirtyRef`) doesn't run without a registration change. So if `itemCount` transitions `5 → undefined` in a commit with no mount/unmount, `elementsRef.current.length` stays `5` with phantom empty slots. Number→number transitions are handled; only →`undefined` leaks. Self-corrects on the next registration-triggered flush, and unlikely in current Combobox usage, but it's a real asymmetry in the new feature.

### 4. ℹ️ Duplicate/out-of-range controlled indexes silently drop an element

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:70-83`

`syncRefs` does `elementsRef.current[metadata.index] = node` with no collision detection (last-write-wins), then `elementsRef.current.length = itemCount` truncates any placement at or beyond `itemCount`. Two items with the same `index` prop, or an `index >= itemCount`, drop an element from `elementsRef` with no warning — navigation to that slot reaches only one (or none). Benign for Combobox (`indexFromFilter < itemCount`, indices unique), but a footgun for future controlled consumers; a dev warning on collision would mirror the existing mixing check.

## Tests (5)

### 1. 🟡 The `itemCount` dynamic-change effect is untested

**Location:** `packages/react/src/internals/composite/list/CompositeList.test.tsx` — no test passes `itemCount`

No `CompositeList` unit test passes `itemCount`; the only coverage is one combobox integration test with a *static* window. The effect that trims/grows `elementsRef`/`labelsRef` length when `itemCount` changes without a registration change (the exact case its comment cites, and findings 1/3 above) is entirely uncovered — a regression there would leave navigation reaching stale indices with nothing failing.

### 2. 🟡 The hover test doesn't isolate the controlled-index path

**Location:** `packages/react/src/combobox/root/ComboboxRoot.test.tsx:1109`

Items render with `index={index}` where `index` is the natural map order, so the manual indices coincide with DOM order. The test passes even if the `index` prop were ignored — it proves the code path doesn't crash, not that the supplied index drives highlighting/`aria-activedescendant`. Reversed or non-contiguous indices would make it a real pin (and directly regression-guard the issue this PR fixes).

### 3. 🟡 Controlled-index collision is untested and undetected

Two items with the same `index` (bug 4 above) silently drop one from `elementsRef` with no assertion or warning anywhere.

### 4. 🟡 Controlled-item label change is untested

**Location:** `packages/react/src/internals/composite/list/CompositeList.test.tsx:500`

"updates labels without re-registering" uses an *uncontrolled* item. For a controlled item the register deps exclude `label`, so the update rides entirely on the `ref` callback's optimistic `labelsRef.current[index]` write — a distinct path with no coverage; a stale `labelsRef` entry (broken typeahead) wouldn't be caught.

### 5. ℹ️ Smaller gaps

No test asserts `listRef`/`labelsRef` length for an `items`-based combobox while the popup is closed (the invariant the deleted `AriaCombobox` line held); grouped + controlled-index reordering (MutationObserver off) is uncovered; and the mixing-error tests use `.mockImplementation(() => {})` (swallows *all* `console.error`, hiding unexpected warnings) with `toHaveBeenCalledWith` rather than `toHaveBeenCalledExactlyOnceWith`. The dedup-across-tests worry is a non-issue — `afterEach` calls `resetBuiltError()`.

## Simplifications (2)

### 1. 🟡 Gate the mixing-error branch behind `NODE_ENV !== 'production'`

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:38-45`

The `else if (controlledRef.current !== isControlled) { error(...) }` runs on every `register` in production, and its ~250 bytes of message strings ship to every consumer of every composite (Tabs/Toolbar/Slider/Select/Menu/Combobox…). Wrapping just the error branch (keeping the `controlledRef.current = isControlled` assignment live, since prod `sortMap` reads it) lets terser drop it — relevant to the bundle-size initiative and matching `Toggle.tsx`/`useButton.ts`. Note the repo is inconsistent here (ComboboxLabel/FieldLabel/DrawerPopup call `error()` ungated), so this is an optional win, not a convention fix. No `error-codes.json` update is needed — that minifier only registers `new Error(...)`, not the `error()` warn util.

### 2. ℹ️ `CompositeMetadata` import paths are inconsistent

Post-move, the type is imported three ways — barrel `'../../internals/composite'` (Tabs/Toolbar/Slider) vs direct `'../list/CompositeListContext'` (CompositeRoot/useCompositeRoot); `SliderRoot.tsx` even splits one import into two statements. All `type`-only, so zero runtime cost — worth a one-line normalization only while touching these files.

## Docs (3)

### 1. 🟡 `useCompositeListItem` header JSDoc "(DOM position)" is now misleading

**Location:** `packages/react/src/internals/composite/list/useCompositeListItem.ts:28`

```
 * Used to register a list item and its index (DOM position) in the `CompositeList`.
```

For a controlled item the registered index is the caller-provided `index`, not the DOM position — exactly what this PR adds. Should read e.g. "…its index in the `CompositeList` (derived from DOM position, or the explicit `index` when controlled)."

### 2. 🟡 `Combobox.Item#index` JSDoc doesn't reflect the new all-or-nothing contract

**Location:** `packages/react/src/combobox/item/ComboboxItem.tsx:298`

```
 * The index of the item in the list. Improves performance when specified by avoiding the need to calculate the index automatically from the DOM.
```

`index` is no longer just a perf hint — setting it on some items but not others in a list now trips the mixing path. The JSDoc should state that if you set `index` on any item you must set it on all items of that list (and run `pnpm docs:api` if the JSDoc changes so `types.md` stays in sync).

### 3. ℹ️ Two smaller doc nits

The `itemCount` JSDoc (`packages/react/src/internals/composite/list/CompositeList.tsx:343`) says "when only a subset is rendered … with controlled indexes," but `ComboboxList` also feeds it for fully-rendered uncontrolled `items` lists — the description is narrower than the usage. And the mixing `error()` message states what/how-to-fix but omits the *why* (the consequence — indexes become inconsistent so navigation/highlighting break), which the repo's error format calls for.

## Verdict

**Approve after nits** — no blocking issues; the substantive items are latent, silent-failure edges in the new `itemCount`/mode-latch contract plus test and doc gaps, none merge-blocking, and every existing suite, typecheck, and lint passes.

---

atomiks:
# PR review

The direction is good, and the rewrite plausibly fixes both the manually indexed hover bug and the StrictMode ref wipe. However, there are three consumer-facing correctness issues plus one lower-priority internal bug, so the current head is not ready to merge.

## Bugs (4)

### 1. 🔴 Recompute the indexing mode when every item switches modes

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:20`

```tsx
const controlledRef = React.useRef<boolean | null>(null);
```

`controlledRef` is permanently fixed by the first item ever registered to that `CompositeList`. Mode mismatches only produce an error in development; production continues using the original mode.

That breaks valid runtime updates where all items change together:

- Controlled → uncontrolled: `sortMap` still takes the controlled branch, but the new metadata has no index. `syncRefs` skips every entry, leaving `elementsRef` empty or containing only `itemCount` holes.
- Uncontrolled → controlled: `sortMap` still takes the uncontrolled branch and overwrites every supplied index with DOM order.

**Failure scenario:** The public `virtualized` prop can trigger exactly this transition. `Combobox.Item` remounts when the root changes virtualization mode, but the surrounding `CompositeList` remains mounted, so its sticky mode survives.

**Fix:** Derive the mode from the currently registered map during each flush, while still warning if the current entries are mixed. Alternatively, reset it once the map becomes empty. Add controlled→uncontrolled and uncontrolled→controlled tests, including a `virtualized` toggle.

### 2. 🔴 Keep the DOM and value registries aligned with controlled indexes

**Location:** `packages/react/src/combobox/root/AriaCombobox.tsx:501`

```tsx
const activeValue = valuesRef.current[options.activeIndex];
```

The new test deliberately supplies reverse indexes:

```tsx
apple  -> 2
banana -> 1
cherry -> 0
```

It then verifies that keyboard navigation follows that controlled order and initially highlights `cherry`. That establishes that explicit indexes are intended to control semantic ordering, not merely duplicate the existing array index.

But when `items` is provided, `valuesRef` remains in `flatFilteredItems` order:

```tsx
['apple', 'banana', 'cherry']
```

while `listRef` becomes:

```tsx
[cherryElement, bananaElement, appleElement]
```

**Failure scenario:** `setIndices({ activeIndex: 0 })` visually highlights `cherry`, but immediately reports `valuesRef.current[0]`, which is `apple`, through `onItemHighlighted`. Initial selected-index resolution has the same mismatch: selected `apple` resolves to index `0`, which now points at the `cherry` element.

**Fix:** Either make `valuesRef`, selected-index resolution, and related registries controlled-index aware; or define that `index` must exactly equal the item's filtered data index, document that restriction, and stop testing reversed indexes. The current halfway state gives incorrect callbacks and incorrect initial highlighting.

### 3. 🔴 Populate the complete value registry for filteredItems-only virtualization

**Location:** `packages/react/src/combobox/root/AriaCombobox.tsx:838`

```tsx
useIsoLayoutEffect(() => {
  if (items) {
    valuesRef.current = flatFilteredItems;
  }
}, [items, flatFilteredItems]);
```

The latest test adds support for:

```tsx
<Combobox.Root virtualized filteredItems={allItems}>
```

but only verifies that `listRef.length === 5`.

`CompositeList` now exposes the complete logical length, including unmounted slots. But `valuesRef` is populated from the complete filtered array only when `items` exists. Without `items`, each mounted `Combobox.Item` fills only its own windowed slot, while `allValuesRef` accumulates only values that happened to mount.

**Failure scenario:** Null virtualized slots are deliberately treated as navigable rather than disabled. Keyboard navigation can therefore move to logical index `0` while only indexes `2` and `3` are mounted. The immediate `onItemHighlighted` call reads `valuesRef[0]` and reports `undefined`; selected-value synchronization can also fail because it searches the incomplete `allValuesRef`.

**Fix:** Treat `hasFilteredItemsProp` as another authoritative data-backed mode:

```tsx
if (hasItems || hasFilteredItemsProp) {
  valuesRef.current = flatFilteredItems;
}
```

The selected-index registry should make the same distinction. Add a test that navigates from a window beginning after index zero and asserts `onItemHighlighted`, selected-index synchronization, and selection—not just array length.

### 4. 🟠 Resynchronize replacement ref objects

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:112`

```tsx
useIsoLayoutEffect(() => {
  syncRefs(sortedMapRef.current);
}, [itemCount, sortedMapRef, syncRefs]);
```

`syncRefs` uses the latest `elementsRef`, but the effect that invokes it only reacts to `itemCount`. A new `elementsRef` object with unchanged registrations and unchanged `itemCount` therefore stays empty.

Previously, item callback refs depended on `elementsRef` and repopulated the replacement. The item hook no longer even reads `elementsRef`, and neither its ref callback nor registration effect changes when that prop changes.

**Failure scenario:** An internal consumer replaces `elementsRef` while keeping the same items and count. The replacement ref remains empty, so navigation reads no registered elements.

**Fix:** Include `elementsRef` and `labelsRef` in the resynchronization effect and add a ref-swap test.

## Tests (0)

No findings.

## Simplifications (1)

### 1. ℹ️ Recheck the targeted performance regressions after rebasing

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:63`

```tsx
elementsRef.current = [];
if (labelsRef) {
  labelsRef.current = [];
}
```

The reported overall benchmark result is positive, but there are targeted alarms:

- Menu open, 500 items: **+27.1%**, including **+34.14 ms paint**
- Combobox typing: **+15.1%**, including **+13.39 ms paint**

There is a plausible mechanism: every flush now allocates fresh arrays and performs a complete map-to-array rebuild, whereas previously items populated their own slots and `CompositeList` generally adjusted only array lengths.

**Failure scenario:** Large menus and comboboxes may incur noticeably more work when opening or typing, even though the aggregate benchmark result remains positive.

**Fix:** Rerun the benchmark several times after rebasing. The branch is currently diverged and about 20 commits behind `master`, so it needs a fresh benchmark and CI pass. The numbers are not conclusive from one run, but they are credible enough not to dismiss as noise.

## Docs (0)

No findings.

## Verdict

**Request changes** - three consumer-facing correctness issues can desynchronize navigation, highlighting, callbacks, and selection.

---

🤖 Review generated with Codex

PR #5149, opened 2026-06-30
[menu][popover][preview card][tooltip] Make the Root part own the store

michaldudak:
Adapts the store-ownership model introduced for Dialog in #5109 to Menu, Popover, Preview Card, and Tooltip, and extracts the shared machinery so all five popup families use one implementation.

### Background

Previously each handle (`Menu.createHandle()`, etc.) owned a single long-lived store for its entire lifetime. A Root adopted that store, which meant remounting a root (e.g. during an animated route transition) reused stale state and required patches like useInitialOpenSync to re-apply defaultOpen. #5109 inverted this for Dialog: the Root owns the store and the handle merely attaches to it. This PR brings the remaining anchored popups onto that model.

### What changed

- The Root now owns the store. Each root creates its own store once per mount (`useRefWithInit`) and attaches it to the handle; the handle exposes an inert fallback store to detached triggers while no root is mounted, and migrates triggers onto the live store as roots attach/detach. Because the store is constructed fresh every mount, `useInitialOpenSync` and the per-component null stores are no longer needed and have been removed.
- Shared implementation in packages/react/src/utils/popups:
  - `BasePopupHandle` — fallback store, root-attachment stack (with the deferred overlapping-roots dev warning), subscriber notifications, and the shared open/close logic. Menu, Popover, Preview Card, Tooltip, and Dialog handles now subclass it and only narrow their public method signatures.
  - `usePopupRootStore` / `useAttachHandle` — create the root-owned store and attach it to the handle.
  - `usePopupHandleStore` — the hook detached triggers use to follow the handle's current store pointer (replaces the deleted per-component `useDialogHandleStore`).
- `open()` on a missing trigger now throws for anchored popups. Calling `handle.open(id)` with an `id` that isn't registered throws for Menu, Popover, Preview Card, and Tooltip — opening an anchored popup with no valid trigger has no sensible anchor. Dialog (and Alert Dialog / Drawer, which shares its handle) is not anchored, so it keeps the existing behavior of opening unassociated with a dev-only warning. This is controlled by a `throwOnMissingTrigger` flag on `BasePopupHandle` (default `true`; Dialog opts out).
- `createChangeEventDetails` trigger param widened from `HTMLElement` to `Element` to match `PopupTriggerMap.getById`.


flaviendelangle (commented):
Oups didn't see the "draft", I removed the review :eyes:

atomiks:
# PR review

The port of the Dialog store-ownership model to Menu, Popover, Preview Card, and Tooltip is faithful and well-tested: the shared `BasePopupHandle` preserves the attach-stack semantics, the fallback-map lookup during the attach commit, and the deferred overlap warning, and the removed `useInitialOpenSync`/null-store machinery is genuinely obsolete under root-owned stores. Nothing is merge-blocking. The one real defect is in the dev-only `StoreInspector` change, which doesn't follow the handle's store pointer it was redesigned to consume.

## Bugs (1)

### 1. 🟡 StoreInspector's new `handle` mode captures the fallback store and never follows the store pointer

**Location:** `packages/utils/src/store/StoreInspector.tsx:177`

```tsx
// A handle exposes a narrowed store view for its public API; at runtime it is a full `Store`.
const store = (props.store ?? props.handle?.store) as Store<any>;
```

`props.handle.store` is read once per render, and `StoreInspectorPanel` subscribes only to that instance's state changes (`store.subscribe`), never to the handle's store-pointer changes (`subscribeStore`). With root-owned stores, the pointer now swaps on every root attach/detach — the very thing the prop was changed from `store` to `handle` to accommodate.

**Failure scenario:** In the updated experiments (e.g. `docs/src/app/(private)/experiments/menu/triggers.tsx`), `<StoreInspector handle={menu1} />` renders before the root's attach layout effect, so it captures the inert fallback `NullStore`, whose state is frozen and never notifies — the panel shows the initial closed state and doesn't update until an unrelated re-render happens to re-read `handle.store`. Similarly, if a root remounts while the panel is open (the route-transition case this model exists for), the panel stays subscribed to the dead previous store and freezes.

**Fix:** Give `StoreOwner` an optional `subscribeStore(listener): () => void` (already on `BasePopupHandle`) and resolve the store via `useSyncExternalStore(handle.subscribeStore, () => handle.store)` when a handle is passed.

## Tests (0)

No findings. The four new suites cover attach/detach ignoring, fresh state on remount, late trigger registration, the missing-id throw, and the overlap warning; the negative cases (clean-handoff no-warn, same-commit fallback-map association) are covered by the Dialog suite and now exercise the shared `BasePopupHandle` code path.

## Simplifications (0)

No findings.

## Docs (1)

### 1. ℹ️ `overlapWarningFrame` is now a declared class field, dropping the deliberate dev-only-instance convention

**Location:** `packages/react/src/utils/popups/popupHandle.ts:110`

```tsx
private overlapWarningFrame: AnimationFrame | undefined;
```

The Dialog original attached this lazily via `(dev as any).overlapWarningFrame ??=` with a comment explaining it was kept off instances "so it never appears on instances in production (like `controlledValues` in `ReactStore`)". With the repo's ES2022 target (`useDefineForClassFields` defaults to true), the declared field is now defined (`undefined`) on every handle instance in production, and the explanatory comment was dropped in the port.

**Failure scenario:** Purely a lost micro-convention — one extra property slot per handle instance and a silent departure from the pattern the deleted comment documented. Fine to decline.

**Fix:** Either restore the lazy dev-only assignment with its comment, or keep the field and accept the convention change knowingly.

## Verdict

**Approve after nits** — the component-library changes are correct and well-covered; only the dev-only `StoreInspector` handle mode needs a subscription to the store pointer.

---

atomiks:
## Code Review (GPT-5.5)
Request changes 🔴, because the new public thrown error is minified but does not meet Base UI's required error-message contract.

### 1. Bugs / Issues

#### 1. 🔴 Expand the missing-trigger error message (blocking)

`packages/react/src/utils/popups/popupHandle.ts`

```ts
throw new Error(
  `Base UI: ${this.componentName}Handle.open: No trigger found with id "${triggerId}".`,
);
```

This new public package error says what happened, but it does not say why that is a problem or how to fix it. The repo's error guidelines require all public errors to explain the consequence and give actionable guidance. Since `docs/src/error-codes.json` now exposes this as error code `99`, the terse text would ship to users as the canonical minified-error message too.

**Fix:** Expand the message to explain that anchored popups cannot open without a registered trigger and tell users to pass the id of a mounted matching trigger using the same handle; then rerun `pnpm extract-error-codes`.

The root-owned store refactor and shared `BasePopupHandle` shape look consistent across Menu, Popover, Preview Card, Tooltip, and Dialog. I specifically checked fallback-store registration, store-pointer subscription, overlapping root handoff, and the anchored-popup missing-trigger path.

### Test Coverage Assessment

Focused validation passed:

```sh
pnpm test:chromium detached-triggers --no-watch
pnpm typescript
```

The focused Chromium run covered 5 detached-trigger files and 154 tests. `git diff --check upstream/master...HEAD` was also clean.

atomiks (approved):
Error message needs to be updated, but looks good aside from that.

PR #5153, opened 2026-07-01
[menu] Cancel stale submenu hover-open when Chrome drops mouseleave

### LINKED ISSUE #5152: [menu] Multiple submenus are broken when moving mouse fast
# Bug report

Noticed a new issue in Paper:

When sliding over multiple submenu triggers, a delayed popup open seems to take over the currently hovered item

https://github.com/user-attachments/assets/b70ebd24-4f05-4f7c-a741-eff19f0b825f

## Reproducible example

Isolated repro
https://stackblitz.com/edit/pexbtjdc-4kzfacv8?file=src%2FApp.tsx

## Base UI version

1.6.0 – this used to work correctly on some older version

atomiks:
Fixes a Chrome regression where sweeping the pointer quickly across adjacent submenu triggers leaves a submenu stuck open and the parent menu unresponsive. Firefox and Safari are unaffected, and it reproduces on old Base UI versions (e.g. 1.3.0), confirming a browser-side change rather than our code.

## Root cause

Each `Menu.SubmenuTrigger` arms a delayed hover-open on `mouseenter` and cancels it on the trigger's `mouseleave`. During a fast sweep, recent Chrome drops the trigger's **non-bubbling `mouseleave`** (the synthesized boundary event), so the delayed open is never cancelled and fires for a trigger the pointer has already left. That stale open takes over the currently-hovered item and leaves the parent scope's safePolygon `pointer-events: none` shielding in place, so the menu appears stuck.

The specific trigger is a common highlight pattern — the hovered item becomes a `position: relative` stacking context with an absolutely-positioned `::before` — which restructures the item's paint box on every hover and trips Chrome's `mouseleave` synthesis. The **bubbling `mouseout` survives** the same crossing.

## Fix

Add an opt-in `guardStaleOpen` to `useHoverReferenceInteraction` that also cancels a pending hover-open from the trigger's `mouseout` (when the pointer genuinely leaves, i.e. `relatedTarget` is outside the trigger), not just from its `mouseleave`. Because `mouseout` bubbles and isn't dropped, the stale open is cancelled at the source and never fires. `Menu.SubmenuTrigger` opts in — tooltips, navigation menu, popover, preview card, etc. are unaffected because the flag defaults to `false`.

This targets the root cause (the un-cancelled open) rather than the visible symptom, is independent of whatever CSS provokes Chrome's drop, and — unlike a `:hover`-based guard — is exercisable with synthetic events, so it ships with a regression test (`fireEvent.mouseOut` on a submenu trigger; verified to fail without the fix).

Closes #5152


flaviendelangle (approved):
# PR review

Fixes a real Chrome regression (#5152) where a fast pointer sweep across adjacent submenu triggers strands a stale hover-open. The fix is narrow and correct: an opt-in `guardStaleOpen` that cancels the pending open from the bubbling `mouseout` (which Chrome doesn't drop), defaulting to `false` so only `Menu.SubmenuTrigger` is affected. Nothing is merge-blocking. I verified the new jsdom tests pass (159 passing in `MenuRoot.test.tsx`, incl. the 2 new ones) and traced the interaction with the existing open/close paths.

## Bugs (0)

No findings.

I specifically checked the highest-risk interaction — that clearing `openChangeTimeout` from `onMouseOut` could cancel a legitimately-scheduled submenu **close**. It does not: `Menu.SubmenuTrigger` always passes `handleClose: safePolygon(...)`, so on leave `onMouseLeave` installs the document `mousemove` handler and schedules the close later via that handler's `onClose → closeWithDelay`, not synchronously during the leave crossing. `onMouseOut` neither removes that handler nor races the not-yet-scheduled close, so the browser ordering of `mouseout` vs `mouseleave` is immaterial here. The `contains(trigger, relatedTarget)` guard correctly returns `false` for `relatedTarget === null` (leave to nothing) via the `contains` null-check, and `true` for within-subtree crossings, matching the two tests.

## Tests (0)

No findings.

The two added tests pin the guard's two branches (cancel on genuine leave; keep on within-subtree crossing) with synthetic events, which is the correct level given the real Chrome drop can't be reproduced synthetically. The `relatedTarget: submenuTrigger` case in the second test exercises the `contains(trigger, trigger) === true` self-containment path. Normal browser open/close flows remain covered by the existing `describe.skipIf(isJSDOM)` suites.

## Simplifications (0)

No findings.

## Docs (0)

No findings. The JSDoc on `guardStaleOpen` and the two inline comments accurately describe the mechanism and its single-trigger-root intent (each `Menu.SubmenuRoot` has exactly one trigger, so the "cancels even for sibling triggers" caveat is safe here).

## Verdict

**Approve** — a correctly-scoped, tested fix for a confirmed browser regression with no behavioral risk to other consumers (flag defaults to `false`). One benign asymmetry worth noting for future readers: unlike `onMouseLeave`, `onMouseOut` has no `isClickLikeOpenEvent()` early-return and clears the timeouts unconditionally — harmless because a click-opened submenu has no pending hover-open or rest timeout to clear, but a one-liner if you want the two handlers to read symmetrically.

---

atomiks:
Still an issue in v150, so this bug will probably stick around — merging

PR #5156, opened 2026-07-01
[composite] Fix nested list reorder detection

### LINKED ISSUE #5125: [docs] Command palette arrow keys out of order
# Bug report

## Current behavior

When searching the docs and using arrow keys to navigate, it seems that the `API Reference` section is out of order, so arrowing up or down doesn't move to the section you expect it to.

Image

https://github.com/user-attachments/assets/bc338767-b5e0-4f45-9e8e-50c0e3f102d7

## Expected behavior

Moving up or down via arrow keys moves the selection to the expected visible items.

## Reproducible example

See screenshot and video example

Helium (Chrome)

## Which OS are you using?

macOS

N/A

jjenzz:
Fixes #5125 

## Summary

- Walks up the DOM tree to find the common ancestor of each adjacent pair of registered items, and observes those boundaries for child-list mutations, covering nested/grouped list changes.
- Only re-sorts when a mutation batch contains a moved node and the registered items are no longer in their last known document order.

## Why

`CompositeList` derives item indexes from DOM order so keyboard navigation and typeahead can move through the same order the user sees. Before this change, reorder detection watched each registered item's immediate parent:

```ts
mutationObserver.observe(node.parentElement, { childList: true });
```

That works for a flat list where items are direct siblings. If React moves an item from one position to another under the same parent, that parent receives child-list mutations and `CompositeList` can re-sort the map.

Grouped lists are different. Items are often nested under group wrappers:

```txt
List
  Group: API Reference
    Item: Tabs > Tab
    Item: Tabs > Root
  Group: Components
    Item: Tabs
```

If the item keeps the same key (`key={item.id}`), React can preserve the component and DOM node which means it doesn't unmount/remount. That is usually desirable, but in this case, if a later query moves the highlighted item's _group_ elsewhere in the list (`key={group.id}`), the `register`/`unregister` callbacks do not fire for the item, nor the mutation observer, so the item keeps the index it had before the reorder.

If `CompositeList` misses that move, the item may still think it has composite index `0`, so `activeIndex = 0` continues to highlight it even though another item is now visually first, and this created the strange keyboard navigation ordering.

Observing the common ancestor of each adjacent pair of items fixes this by watching the boundaries that actually own the composite ordering: a reorder that changes item indexes must invert at least one adjacent pair, and the mutation that causes it happens at that pair's common ancestor. For a flat list this collapses to the single item parent; for grouped lists it includes the group wrapper boundaries.

## Performance Notes

The observer still only watches child-list mutations (no `subtree`), and observes one node per adjacent-pair boundary instead of every item's parent, so a flat list collapses to a single observation. The callback also filters out batches that contain no moved node, and verifies the items' relative document order actually changed before re-sorting.

### DIFF
--- a/packages/react/src/internals/composite/list/CompositeList.tsx
+++ b/packages/react/src/internals/composite/list/CompositeList.tsx
@@ -4,9 +4,7 @@ import * as React from 'react';
 import { useRefWithInit } from '@base-ui/utils/useRefWithInit';
 import { useStableCallback } from '@base-ui/utils/useStableCallback';
 import { useIsoLayoutEffect } from '@base-ui/utils/useIsoLayoutEffect';
-import { contains } from '../../shadowDom';
 import { CompositeListContext } from './CompositeListContext';
-import { isElement } from '@floating-ui/utils/dom';
 
 export type CompositeMetadata<CustomMetadata> = {
   index?: number | null | undefined;
@@ -49,29 +47,34 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
     setMapTick(lastTickRef.current);
   });
 
-  const { sortedMap, sortedNodes } = React.useMemo(() => {
+  const sortedMap = React.useMemo(() => {
     // `mapTick` is the `useMemo` trigger as `map` is stable.
     disableEslintWarning(mapTick);
 
-    const nextSortedMap = new Map<Element, CompositeMetadata<Metadata>>();
+    const newMap = new Map<Element, CompositeMetadata<Metadata>>();
     // Filter out disconnected elements before sorting to avoid inconsistent
     // compareDocumentPosition results when elements are detached from the DOM.
-    const nextSortedNodes = Array.from(map.keys())
+    const sortedNodes = Array.from(map.keys())
       .filter((node) => node.isConnected)
       .sort(sortByDocumentPosition);
 
-    nextSortedNodes.forEach((node, index) => {
+    sortedNodes.forEach((node, index) => {
       const metadata = map.get(node) ?? ({} as CompositeMetadata<Metadata>);
-      nextSortedMap.set(node, { ...metadata, index });
+      newMap.set(node, { ...metadata, index });
     });
 
-    return { sortedMap: nextSortedMap, sortedNodes: nextSortedNodes };
+    return newMap;
   }, [map, mapTick]);
 
   useIsoLayoutEffect(() => {
-    if (typeof MutationObserver !== 'function' || sortedNodes.length < 2) {
+    // A single item can't reorder.
+    if (typeof MutationObserver !== 'function' || sortedMap.size < 2) {
       return undefined;
     }
+
+    // `sortedMap` is populated in sorted order, so its keys are the last known
+    // document order of the items.
+    const sortedNodes = Array.from(sortedMap.keys());
 
     // Rather than observing the whole composite container, observe the smallest
     // direct-child lists whose order can affect registered item order. For flat
@@ -83,15 +86,18 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
     }
 
     const mutationObserver = new MutationObserver((entries) => {
-      if (!entries.some((entry) => hasSortedNode(entry, sortedNodes))) {
+      // Only verify the order after a move: a node that was removed and later
+      // re-added within the same batch. Additions and removals alone can't
+      // change the relative order of the remaining items, and items that mount
+      // or unmount re-sort through `register`/`unregister`.
+      if (!hasMovedNode(entries)) {
         return;
       }
 
       let previousConnectedNode: Element | null = null;
 
-      // `sortedNodes` is the last known document order. If any connected node now
-      // appears before the previous connected node, wrappers/items moved and the
-      // index map needs to be rebuilt.
+      // If any connected node now appears before the previous connected node,
+      // wrappers/items moved and the index map needs to be rebuilt.
       for (const node of sortedNodes) {
         if (!node.isConnected) {
           continue;
@@ -115,7 +121,7 @@ export function CompositeList<Metadata>(props: CompositeList.Props<Metadata>) {
     return () => {
       mutationObserver.disconnect();
     };
-  }, [sortedNodes]);
+  }, [sortedMap]);
 
   useIsoLayoutEffect(() => {
     const shouldUpdateLengths = lastTickRef.current === mapTick;
@@ -195,35 +201,41 @@ function getAdjacentNodeRoots(nodes: Element[]) {
 function getCommonAncestor(firstNode: Element, lastNode: Element) {
   let ancestor = firstNode.parentElement;
 
-  while (ancestor && !contains(ancestor, lastNode)) {
+  // The `parentElement` walk cannot cross shadow boundaries, so the native
+  // `contains` is sufficient here.
+  while (ancestor && !ancestor.contains(lastNode)) {
     ancestor = ancestor.parentElement;
   }
 
   return ancestor;
 }
 
-function hasSortedNode(entry: MutationRecord, sortedNodes: Element[]) {
-  for (let i = 0; i < entry.addedNodes.length; i += 1) {
-    if (containsSortedNode(entry.addedNodes[i], sortedNodes)) {
-      return true;
+function hasMovedNode(entries: MutationRecord[]) {
+  const removed = new Set<Node>();
+
+  // The records are chronological: an addition following a removal is a move,
+  // while a removal following an addition is a net removal.
+  for (const entry of entries) {
+    for (let i = 0; i < entry.addedNodes.length; i += 1) {
+      if (removed.has(entry.addedNodes[i])) {
+        return true;
+      }
+    }
+
+    for (let i = 0; i < entry.removedNodes.length; i += 1) {
+      removed.add(entry.removedNodes[i]);
     }
   }
 
-  for (let i = 0; i < entry.removedNodes.length; i += 1) {
-    if (containsSortedNode(entry.removedNodes[i], sortedNodes)) {
+  // A removed node that is still connected was reparented into a container
+  // this observer doesn't watch.
+  for (const node of removed) {
+    if (node.isConnected) {
       return true;
     }
   }
 
   return false;
-}
-
-function containsSortedNode(node: Node, sortedNodes: Element[]) {
-  if (!isElement(node)) {
-    return false;
-  }
-
-  return sortedNodes.some((sortedNode) => contains(node, sortedNode));
 }
 
 function sortByDocumentPosition(a: Element, b: Element) {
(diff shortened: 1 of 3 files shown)


atomiks:
# PR review

The fix's mechanism is correct — observing the list container is the right boundary for the "composite index = DOM order" invariant — and I measured no typing-path perf regression (local `combobox` benchmarks on master vs this PR, flat and grouped 500-item lists, all deltas within noise; the PR also replaces ~500 per-item `observe()` calls per keystroke with one). Nothing is merge-blocking, but the move-detection heuristic has a widened masking window, the bug that motivated the PR has no regression test, and roughly half the diff is avoidable churn.

## Bugs (2)

### 1. 🟡 Reorder detection can be masked or spuriously triggered by unrelated mutations in the same batch

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:81`

```ts
if (diff.size === 0) {
  lastTickRef.current += 1;
  setMapTick(lastTickRef.current);
}
```

The "was this a move?" heuristic (removed+added node cancels out → re-sort) is inherited from master, but `subtree: true` feeds it a much wider mutation stream, and it keys on *any* node, not registered items. Two directions (plausible — the mechanism is real, the trigger needs a specific commit shape):

- **False negative:** a group reorder batched in the same React commit as any unregistered childList change (a conditional separator, badge, or decorative node mounting inside an item) yields `diff.size !== 0`, the tick is skipped, no `register` fires (the extra node isn't an item), and the stale-index bug this PR fixes reappears for that commit.
- **False positive:** any node move *inside* an item (a component reordering its own children) nets to `diff.size === 0` and triggers a full `compareDocumentPosition` re-sort plus a new map identity flowing to `onMapChange` and all item listeners, for no ordering change.

**Failure scenario:** in a grouped command palette, a query that reorders groups *and* mounts a decorative node in the same commit leaves arrow-key order stale — the exact symptom of #5125, now on a rarer path.

**Fix:** make the trigger item-aware instead of count-based: tick when any added/removed node in the batch is, or contains, a registered element (the `diff`-cancellation can stay as a fast path). That closes both directions.

### 2. ℹ️ Items outside the container ref are silently unobserved

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:88`

When `listRef` is provided, per-parent observation is skipped entirely. An item rendered outside that container (portaled, or an external `/internals` consumer passing a too-narrow ref) loses reorder detection it previously had, with no warning. All current internal consumers keep items inside the passed container, so this is a contract note for the newly public `/internals` surface, not a live bug.

**Failure scenario:** an external consumer passes their scroll-viewport ref while rendering one composite item outside it; that item's reorders are never detected.

**Fix:** none required now; consider a dev-mode warning when a registered node is not contained by `listRef.current`.

## Tests (1)

### 1. 🟡 No regression test for the bug being fixed

**Location:** `packages/react/src/internals/composite/list/CompositeList.test.tsx:22`

```tsx
<CompositeList listRef={{ current: null }} elementsRef={elementsRef} labelsRef={labelsRef}>
```

The only test change adapts the new required prop — and passes `{ current: null }`, which exercises the old fallback path, not the new container-observation path. The fixture updates (`ListboxFocus`, `Menu`, `MenuOrientation`) pass real refs but are flat lists, where old and new behavior are identical. Nothing reproduces #5125: keyed group wrappers reordering around items that never remount. This exact blind spot survived #2675 unnoticed, which is an argument for pinning it.

**Failure scenario:** a future refactor (e.g. someone "optimizing" `subtree: true` away) reintroduces the stale-index bug and every test stays green.

**Fix:** add a `CompositeList` test that renders items under two keyed group wrappers, swaps the group order via state (stable keys so nodes move rather than remount), and asserts the registered indexes match the new DOM order.

## Simplifications (3)

### 1. 🟡 The `listRef` API may be unnecessary — the container is derivable from registered nodes

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:187`

`sortedMap` is sorted by document position, and a subtree is a contiguous preorder interval — so the lowest common ancestor of the first and last registered node contains every registered node:

```ts
let root = nodes[0].parentElement;
while (root && !root.contains(nodes[nodes.length - 1])) {
  root = root.parentElement;
}
```

Observing that root (falling back to per-parent if it climbs to `body`) fixes every consumer — including external `/internals` users, who get no new required prop to learn — in one file. It eliminates the 10-file ref threading, the fixture churn, and most of the +319 B bundle cost, and for every current consumer resolves to the same element the PR threads manually or a tighter one.

**Failure scenario:** every future `CompositeList` consumer must remember to thread a container ref (and merge it into their element's ref array) or silently land on the weaker fallback — a per-consumer tax the library can absorb once instead.

**Fix:** derive the observation root in the effect as above; keep the explicit prop only as an optional override if wanted.

### 2. 🟡 Existing refs not reused: Select creates a duplicate, OTP passes `null` past its own root ref

**Location:** `packages/react/src/select/positioner/SelectPositioner.tsx:72`

`SelectPositioner` builds a new ref and splices it into the ref array, while this same PR reuses `store.context.popupRef` in `MenuPositioner`. Select has the identical `popupRef` in context (`packages/react/src/select/root/SelectRoot.tsx:496`); the popup contains every item and is a tighter observation scope than the positioner (excludes the arrow). Likewise `OTPFieldRoot` passes `listRef={null}` even though its `rootRef` (`packages/react/src/otp-field/root/OTPFieldRoot.tsx:102`) is already attached to the container — passing it would leave zero internal consumers on the fallback branch.

**Failure scenario:** redundant ref plus merged-ref work on every Select render, an inconsistent pattern between two components changed in the same PR, and a fallback branch kept alive by a single consumer that didn't need it.

**Fix:** `listRef={store.context.popupRef}` in `SelectPositioner` (dropping the new ref); `listRef={rootRef}` in `OTPFieldRoot`.

### 3. 🟡 Required prop and `listRef` naming generate churn and a collision

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:187`

Making the prop required-with-`| null` forces `listRef={null}`/`{ current: null }` boilerplate through the test and would-be-untouched consumers, and the three floating-ui fixture files gain `useMergedRefs` plumbing solely to satisfy it on flat lists the fallback already handles. Separately, `listRef` already means "the elements array" in floating-ui land, producing `<CompositeList listRef={ref} elementsRef={store.state.listRef}>` — two `listRef`s with different meanings in one expression (`packages/react/src/combobox/store.ts:52`).

**Failure scenario:** readers must disambiguate `listRef` by context on every touched call site; fixture/test diffs obscure the real change.

**Fix:** rename to `containerRef` and make it optional (moot if Simplifications No. 1 is taken).

## Docs (1)

### 1. ℹ️ New JSDoc contradicts the prop's type

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:185`

```ts
/**
 * A ref to the composite list container to listen for reorder mutations.
 * If this is not passed, it will observe each item's parent.
 */
listRef: React.RefObject<HTMLElement | null> | null;
```

"If this is not passed" describes an optional prop, but the type requires an explicit value (consumers pass `null`). Say "If `null`, each item's parent is observed instead" — or make the prop optional, which makes the sentence true.

**Failure scenario:** an `/internals` consumer reads the JSDoc, omits the prop, and gets a type error the doc told them was fine.

**Fix:** align the sentence with whichever prop shape ships.

## Verdict

**Approve after nits** — the fix is correct with no measured perf cost, but the batched-mutation masking window, the missing regression test, and the avoidable API/threading churn are worth addressing before merge.

---

jjenzz:
@atomiks i love these reviews you do, what prompt/skill are you using to generate them? and i wonder if it is worth setting up some infra to do it automatically on `member` PRs.

jjenzz:
@atomiks i've addressed the feedback but i've kept the explicit `containerRef` prop rather than deriving the container from registered items. Deriving can work, but the component already knows the intended ordering boundary, so passing it explicitly avoids repeated ancestor scans whenever the registered map changes, and seems more predictable. lmk if you have any thoughts there though

packages/react/src/internals/composite/list/CompositeList.tsx
```diff
@@ -153,6 +174,31 @@ function createListeners() {
...
+  for (const registeredNode of map.keys()) {
+    if (contains(node, registeredNode)) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+function isElement(node: Node): node is Element {
```
jjenzz: i did see that there is already a `@floating-ui` util for this that we appear to use elsewhere, but it didn't seem apt. for `CompositeList` to have a floating-ui dependency. lmk if we feel that's fine here tho, happy to change.

atomiks: It's treeshakable and used ~everywhere, so it's fine to import from `@floating-ui` 👍

atomiks:
@jjenzz [Base UI Review Skill](https://docs.google.com/document/d/1V51nL8KDanLVenKr6lgKK3yMLmBq1Da9eWqUB0cnQWw/edit?tab=t.0) — Fable is by far the best reviewer though (which I used here), no other model is as good. `/base-ui-review high effort` makes the other models get ~close, but `/base-ui-review` (default medium review effort, cheaper) is still better with Fable I noticed.

> i've kept the explicit containerRef prop rather than deriving the container from registered items.

IMO the case for simplicity is worth it here, because the perf seems like it's not an issue. I got Fable to benchmark it:

> I benchmarked the derive-vs-explicit question in a real Chrome DOM, mirroring this PR's sortedMap memo and observer setup exactly (realistic fixture: 10-deep app wrappers, popup with arrow siblings, keyed group wrappers). The LCA walk (first.parentElement up until it contains(last)) would run in the same effect as the observer setup, i.e. exactly as often as the compareDocumentPosition sort — so per-map-change cost is the fair comparison:

Scenario | sort+rebuild (already per map change) | LCA derivation | overhead
-- | -- | -- | --
20 items, 4 groups | 2 µs | 0.07 µs | 3.5%
500 flat | 306 µs | 0.38 µs | 0.12%
500 grouped (50×10) | 70 µs | 0.43 µs | 0.61%
500 grouped, items 10 wrappers deep | 86 µs | 0.71 µs | 0.82%
5000 grouped (500×10) | 988 µs | 3.7 µs | 0.37%

> The "repeated ancestor scans" are sub-microsecond — under 1% of work the memo already does on the identical trigger, and ~100–200× cheaper than the per-parent observe() loop the fallback runs per map change today. On predictability: the derived LCA resolved to exactly the list element in every scenario, which is tighter than the popupRef we pass explicitly in Menu/Select (that subtree includes the arrow nodes). Deriving would also let us drop the prop, the 10-file ref threading, the per-render merged refs in CompositeRoot/Tabs/Accordion/ComboboxList, and most of the +351B — the PR becomes CompositeList + tests. One real edge: items split across portals can push the LCA toward body; guarding with "if the walk reaches body, use the per-parent fallback" covers it.

Might be worth checking also with Codex to see if it can verify+reproduce that, or do it manually with DevTools

atomiks (approved):

jjenzz:
thanks @atomiks  🙌

Issue #3185, opened 2025-11-11
[scroll area] Prop to disable custom scrollbars on mobile (iOS/Android)


atomiks:
# Feature request

## Summary

Scroll Area currently disables native scrollbars on mobile, so the styles are consistent everywhere.

On iOS, the Scroll Area's scrollbars lack the rubber band effect (though this could potentially be tweaked given the scroll area component disables it intentionally to fix some issues), and the `transform` updates are pretty laggy on iOS due to heavily async scroll events. 

Since mobile scrollbars are already optimized and don't look bad, we can add an option to disable custom scrollbars on iOS & Android specifically.

## Motivation

The scrollbars on the [Dialog examples](https://master--base-ui.netlify.app/react/components/dialog#outside-scroll-dialog) feel worse than the native scrollbars on iOS due to the aforementioned issues (no rubber band, laggy transform updates)

PR #5157, opened 2026-07-02
[scroll area] Fix scrollbar visibility during touch scrolling on iOS

atomiks:

Spotted while testing the new nav drawer in #5137: on iPhone, the scrollbar sometimes never appears while scrolling, depending on prior momentum/overscroll timing.

The viewport treats a scroll as user-driven only after a `wheel`/`touchmove`/`pointermove`/`pointerenter`/`keydown` event, and flips back to programmatic after 100ms without scroll events. On iOS, a touch that catches an in-flight momentum scroll or rubber-band bounce is consumed natively by WebKit and doesn't dispatch the expected touch events (confirmed by @aarongarciah — `touchmove` is swallowed after re-grabbing), so the gesture gets classified as programmatic and `data-scrolling` never appears.

Since no DOM event reliably marks these gestures, event-based attribution can't work on touch. Fix: while in touch modality (last pointer to interact with the root was a touch), every scroll is treated as user-driven and the programmatic check is skipped. This matches native iOS behavior, where the system scroll indicator also shows for programmatic scrolls. Programmatic suppression still applies on desktop (wheel/mouse/keyboard modality) and before the first touch (for example scroll restoration on load).

This needs manual verification on an actual iPhone, since the bug depends on native gesture handling that tests can't reproduce.

Related: #3185


aarongarciah:
I'm testing locally and the issue persists. When I "re-grab" the scrollarea while it's scrolling due to momentum, the thumb disappears.

https://github.com/user-attachments/assets/0a3d63fe-3cb6-440c-b150-c13319e6a9e5

Also happening in the PR preview:

https://github.com/user-attachments/assets/ec974505-de7a-4dc6-ad53-0e6df0854d93

aarongarciah:
It works now. I haven't tried it on Android, though.

https://github.com/user-attachments/assets/2aadc194-a1dd-46a5-831f-c5adaa283b48

PR #5159, opened 2026-07-02
[menu][select][combobox] Widen trigger slip-out release tolerance

### LINKED ISSUE #5158: [menu] Slip-click silently cancels opening: mouseup a few px outside the trigger closes the just-opened menu (2px tolerance, no time threshold)
# Bug report

## Current behavior

`Menu.Trigger` opens the menu on `mousedown`. A one-time document `mouseup` listener is then attached (see `handleDocumentMouseUp` in [`packages/react/src/menu/trigger/MenuTrigger.tsx`](https://github.com/mui/base-ui/blob/master/packages/react/src/menu/trigger/MenuTrigger.tsx)): if the `mouseup` lands outside the trigger's bounds (+2px `BOUNDARY_OFFSET`) and outside the popup, the just-opened menu is immediately closed (`cancelOpen`).

As a result, an ordinary quick click whose pointer drifts a few pixels past the trigger's edge between press and release opens the menu for a few tens of milliseconds and then silently closes it. There is **no time threshold** — even a <100ms click is treated as a "drag off and release" cancel.

On small triggers (icon buttons / chevrons in the 12–24px range) this happens surprisingly often with fast mouse movement or trackpad taps. To the user, the menu appears to "flash" or simply not react.

Real-world impact: we ship Base UI menus in production, and our product analytics recorded **15 rage-click events from 8 distinct users in a single day**, all landing on menu triggers — users clicking repeatedly because the menu kept cancelling itself.

(Aside: on 1.5.0, `aria-expanded` also stayed `"false"` on the trigger while the popup was open, which made this harder to diagnose from telemetry. That appears to be fixed in 1.6.0.)

## Expected behavior

A quick click that starts on the trigger should open the menu and leave it open, even if the pointer drifts slightly past the trigger's edge before release. Possible directions, in order of preference:

1. Treat a short press (e.g. released within ~200ms of the press, similar in spirit to the existing `PATIENT_CLICK_THRESHOLD`) as a click rather than a drag-cancel — native macOS menus behave this way.
2. Or widen the tolerance well beyond 2px (e.g. 16–24px) — 2px is far below normal click jitter.
3. Or make cancel-on-release-outside opt-out via a prop.

Note the same `BOUNDARY_OFFSET = 2` pattern also exists in `SelectTrigger.tsx` and `ComboboxTrigger.tsx`, so those components likely share the behavior.

## Reproducible example

Reproducible on the official docs demo at https://base-ui.com/react/components/menu (the "Song" hero demo):

1. Press the mouse button down on the trigger — the menu opens.
2. While holding, move the pointer just past the trigger's edge (3–10px is enough).
3. Release.
4. The menu silently closes.

I verified this with synthesized real input events (CDP `Input.dispatchMouseEvent`): `mousedown` at the trigger center → menu open (`aria-expanded="true"`, popup mounted) → `mousemove` to 9px past the right edge → `mouseup` → menu closed. Also reproduced in a minimal `<Menu.Root><Menu.Trigger>…` app on 1.6.0 with a 4px overshoot.

No special code is needed — any default Menu reproduces it.

## Base UI version

1.6.0 (also present in 1.5.0)

Chrome (reproduced on Chrome 149 and Chromium via CDP)

## Which OS are you using?

macOS

## Additional context

The press → drag off → release = cancel behavior itself is presumably intentional (press-drag-release menu interaction). The problem is that it applies with no time threshold and an extremely tight 2px boundary, so ordinary fast clicks on small triggers get misclassified as drag-cancels. Rage-click telemetry from production suggests this hits real users noticeably often.

atomiks:
Increases the slip-out release tolerance (`BOUNDARY_OFFSET`) from 2px to 5px in `Menu.Trigger`, `Select.Trigger`, and `Combobox.Trigger`, and extracts the duplicated bounds check into a shared `isMouseWithinBounds` util.

When a popup opens on `mousedown`, releasing the mouse outside the trigger's bounds cancels the open (`cancel-open` reason). The previous 2px tolerance was tight enough that fast clicks with slight pointer drift were treated as slip-out cancellations, causing popups to flash closed. 5px matches typical OS/browser drag-initiation thresholds, so ordinary click jitter no longer cancels while a deliberate drag-off-to-dismiss still does.

Note on history: the 2px offset was introduced in #1250 as slack around the approximate `getPseudoElementBounds` measurement rather than as a click-drift tolerance. This PR repurposes it as an explicit drift tolerance (which also still covers the measurement imprecision) and documents it as such.

Related: #4994 takes the same position-driven (rather than time-driven) approach for the context menu open gesture.

Closes #5158


flaviendelangle (approved):
# PR review

Clean, well-scoped refactor: extracts the duplicated slip-out bounds check into a shared `isMouseWithinBounds` util and bumps the tolerance from 2px to 5px. The change is behavior-preserving at each call site except the intentional, documented tolerance increase. Nothing is merge-blocking; the only note is that the new coverage is at the util level rather than component-level.

## Bugs (0)

No findings.

## Tests (1)

### 1. ℹ️ Tolerance change is pinned only at the util level, not per-component

**Location:** `packages/react/src/utils/getPseudoElementBounds.test.ts:60`

```ts
it('allows up to 5px of mouse drift around element bounds', () => {
```

The new test verifies the arithmetic of `isMouseWithinBounds` (95px inside, 94px outside), which is good and directly pins the 5px boundary. The actual regression being fixed (a fast click with slight drift no longer emits `cancel-open` / closes the popup) has no component-level test in Menu/Select/Combobox. The util test is a reasonable proxy since all three delegate to it, so this is optional rather than required.

**Failure scenario:** A future change to the mouseup handler in one trigger (e.g. reordering the bounds check, or reverting to a direct `getPseudoElementBounds` call with a different offset) would not be caught by the util test — only by a per-component test asserting the popup stays open on a slightly-drifted release.

**Fix:** Optionally add one Chromium-env test per trigger (`describe.skipIf(isJSDOM)`) that opens on mousedown, releases ~3–4px outside the trigger, and asserts the popup remains open.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve** — pure consolidation plus an intentional, documented tolerance bump with matching util test coverage; no correctness or API concerns.

---

flaviendelangle (approved):
# PR review

This PR widens the slip-out release tolerance from 2px to 5px across `Menu`/`Select`/`Combobox` triggers and extracts the duplicated bounds-check into a shared `isMouseWithinBounds` util. It's a faithful refactor: all three call sites previously used identical `BOUNDARY_OFFSET = 2` logic, and each retains its own null-guard before calling the new helper, so the extraction is type-safe and behavior-preserving apart from the intended constant change. Tests and lint pass locally. Nothing is merge-blocking.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Notes

### ℹ️ Boundary behavior for `Menu`/`Select` triggers is only covered indirectly

**Location:** `packages/react/src/utils/getPseudoElementBounds.test.ts:60`, `packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx:484`

The new integration tests exercise the 5px tolerance through `Combobox.Trigger` only; `Menu.Trigger` and `Select.Trigger` changed behavior too but get no trigger-level test. The shared `isMouseWithinBounds` unit test pins the boundary math centrally, so the arithmetic is covered — this is only a heads-up that a future regression in the menu/select mouseup wiring (e.g. an accidentally-dropped `isMouseWithinBounds` guard) wouldn't be caught by these files. Reasonable to decline given the central coverage.

### ℹ️ `isMouseWithinBounds` lives in `getPseudoElementBounds.ts`

**Location:** `packages/react/src/utils/getPseudoElementBounds.ts:20`

The helper's name reads as generic "is the mouse over this element" rather than pseudo-element-specific. Co-locating it with `getPseudoElementBounds` (which it wraps) is defensible, so this is purely a naming/altitude observation, not something to change.

## Verdict

**Approve** — a behavior-preserving refactor plus an intentional, documented tolerance bump; no correctness issues found, and the only gaps are optional test-coverage notes.

---

PR #5165, opened 2026-07-03
[typescript] Strip internal types from published .d.ts

michaldudak:
## Summary

- Enable TypeScript `stripInternal` for the `@base-ui/react` declaration build so `@internal` declarations are omitted from the published `.d.ts` output. This should make it harder for consumers to accidentally use APIs that are not intended for public use.
- Keep internal Floating UI helper components undocumented and stripped, while updating public portal prop types to inline the small container shape they expose instead of naming internal helper props. This can prevent accidental type breaking changes (when we change an `@internal` type that's also used in a public interface)
- Keep workspace project references on an unstripped declaration config so docs/tests can still typecheck against internal helpers.
- Add an API Extractor check for the `@base-ui/react` declaration entry point and run it in CircleCI's package-verification job next to `Check public types`, reusing the generated package build output.

## Motivation

Base UI marks implementation-only declarations with `@internal`, but those declarations were still being emitted in the published type files. Enabling `stripInternal` removes them from the generated declarations.

The extra API Extractor check catches the important failure mode after stripping: a public declaration accidentally referencing an internal declaration that TypeScript removed from the emitted `.d.ts` graph. The existing public-types check still validates the broader published declaration graph.

Where public APIs previously reached through Floating UI helper prop aliases, this PR now duplicates the exported shape at the component boundary so the helpers can remain internal and absent from generated docs.

cc @mui/infra, not sure if this flag was already used in other projects. IMO it makes sense as it makes the public contract clearer.


package.json
```diff
@@ -57,6 +57,7 @@
     "test:browsers": "cross-env VITEST_ENV=all-browsers pnpm test:_unit",
     "test:argos": "code-infra argos-push --folder test/regressions/screenshots/chrome",
     "typescript": "tsgo -b tsconfig.json",
+    "types:check": "corepack pnpm --filter @base-ui/react build && corepack pnpm --filter @base-ui/react types:check",
```
brijeshb42: Why corepack?

Janpot:
> cc @mui/infra, not sure if this flag was already used in other projects. 

ATM It's being used [selectively](https://github.com/mui/material-ui/blob/2bcf35ff620b46d5a2e84ee691c9f3d7a51f03f7/packages/mui-system/tsconfig.build.json#L12) in `mui/material` for projects that are converted to typescript, for similar reasons.

packages/react/package.json
```diff
@@ -125,6 +125,7 @@
     "test:package": "publint --pack pnpm && attw --pack ./build --exclude-entrypoints package.json",
     "release": "pnpm build && pnpm publish",
     "test": "cross-env VITEST_ENV=jsdom vitest",
+    "types:check": "api-extractor run --config api-extractor.json --typescript-compiler-folder ../../node_modules/typescript",
```
brijeshb42, line 128: Is this script just checking for CI or also stripping ?

package.json
michaldudak: whoops, a leftover from an earlier issue.

packages/react/package.json
michaldudak: This is just for verifying if none of the internals are used as part of public types. Stripping is done entirely with tsc by the `stripInternal` flag.

Janpot (approved):

PR #5167, opened 2026-07-03
[utils] Move clamp, areArraysEqual, shadowDom, and formatNumber to utils

romgrk:
## Summary

First phase of moving domain-agnostic primitives out of `@base-ui/react`'s `internals/` and `utils/` into the shared `@base-ui/utils` stdlib (usable across MUI + Base UI). These are pure, dependency-clean utilities with no Base UI-specific semantics.

Moved (history preserved via `git mv`):

- **`clamp`** — pure math
- **`areArraysEqual`** — reimplemented inspired by [`fastArrayCompare`](https://github.com/mui/mui-x/blob/master/packages/x-internals/src/fastArrayCompare/fastArrayCompare.ts) from `@mui/x-internals`: referential-equality short-circuit, early length bail-out, and a reverse loop that inlines the element check when no custom comparer is passed. Uses `Object.is` for the default comparison (consistent with the sibling `itemEquality` util) so `NaN` values are treated as equal. The optional `itemComparer` parameter is retained for the Select/Combobox call sites.
- **`shadowDom`** — `activeElement`, `contains`, `getTarget` (shadow-DOM-safe traversal)
- **`formatNumber`** + its `stringifyLocale` dependency

## Changes

- All import sites repointed to `@base-ui/utils/*` (17 `clamp`, 4 `areArraysEqual`, 8 `formatNumber`, 1 `shadowDom`, 1 `stringifyLocale`).
- Dropped the now-obsolete `./internals/{areArraysEqual,clamp,shadowDom}` export subpaths from `packages/react/package.json`. The moved files are auto-exported via the utils package's `./*` glob.
- Added a test for `areArraysEqual` (previously untested); `clamp`, `formatNumber`, and `stringifyLocale` tests moved alongside their sources and now run under the `@base-ui/utils` project.

## Follow-up

A subsequent phase will import the broader set of generic utilities currently living in `@mui/x-internals` (e.g. `hash`, `isDeepEqual`, `throttle`, `disposable`, …) into `@base-ui/utils`, in coordination with a parallel `mui-x` PR that re-exports them from the stdlib. That work is intentionally kept out of this PR to keep the move small and reviewable.

## Verification

- `pnpm typescript` ✓
- `pnpm eslint` ✓
- `pnpm prettier` ✓
- `pnpm test:jsdom clamp areArraysEqual formatNumber stringifyLocale` ✓ (13 tests)


packages/utils/src/formatNumber.ts
michaldudak, line 1: The utils package’s wildcard export exposes every named export from @base-ui/utils/formatNumber. This makes a generic implementation cache public and allows consumers to mutate shared formatter behavior globally.
Fix: Change this to a module-private const cache.

mnajdova (approved):

michaldudak:
It's worth noting that if anyone relies on the `internals`, they'll have to update their imports (CC @kenanyusuf, @DiegoAndai)

michaldudak (approved):

michaldudak:
# PR review

The move itself is clean — no dangling specifiers, no `@base-ui/utils` → `@base-ui/react` back-edge, and the moved tests still get collected by the utils project. What follows is scoped to tests, cleanup, and docs; nothing here is merge-blocking. The recurring theme is the `areArraysEqual` rewrite: it changes default equality semantics and adds a per-element branch and a dead short-circuit, and the new test file pins only half of the behavior change. I could not run the suite locally, so all conclusions come from reading the code plus standalone arithmetic checks of the slider math.

## Tests (3)

### 1. 🟠 Only the loosening half of the `Object.is` switch is pinned

**Location:** `packages/utils/src/areArraysEqual.test.ts:26`

```ts
it('treats NaN as equal to NaN (Object.is semantics)', () => {
  expect(areArraysEqual([NaN], [NaN])).toBe(true);
});
```

The default comparison changed from `(a, b) => a === b` to `Object.is`, which flips two things: `NaN` becomes equal to itself, and `-0` stops being equal to `0`. The `NaN` direction (false → true) is covered here. The `+0`/`-0` direction (true → false) is not, and neither is the custom-comparer escape hatch. Those semantics reach `SliderRoot.tsx:45` (`areValuesEqual`, the gate that decides whether `setValue` fires) and `SliderRoot.tsx:186` (the field dirty check) — the only two call sites that omit a comparer over numeric arrays.

**Failure scenario:** a later revert to `===`, or a swap to a generic shallow-equality helper, flips Slider's dirty-state and change-suppression behavior back with no red test.

**Fix:** add `expect(areArraysEqual([0], [-0])).toBe(false)` and `expect(areArraysEqual([0], [-0], (a, b) => a === b)).toBe(true)`.

### 2. 🟠 The short-circuit's right to skip a caller-supplied comparer is untested

**Location:** `packages/utils/src/areArraysEqual.test.ts:30`

```ts
it('uses the provided comparer', () => {
  const a = [{ id: 1 }, { id: 2 }];
  const b = [{ id: 1 }, { id: 2 }];
  expect(areArraysEqual(a, b)).toBe(false);
  expect(areArraysEqual(a, b, (x, y) => x.id === y.id)).toBe(true);
});
```

Both arrays are distinct references, so this can never observe `areArraysEqual.ts:18` returning `true` before `itemComparer` is consulted. That matters because `SelectRoot.tsx:248` and `AriaCombobox.tsx:1033` route the public `isItemEqualToValue` prop through this path — there is now an undocumented carve-out where user code is bypassed on referential identity. The single `it` also asserts two unrelated behaviors under a name describing only the second.

**Failure scenario:** the bypass is never exercised, so a future change that makes the comparer authoritative — or that removes the short-circuit — produces no signal either way.

**Fix:** split the test, and add `const spy = vi.fn(() => false); expect(areArraysEqual(arr, arr, spy)).toBe(true); expect(spy).not.toHaveBeenCalled();`.

### 3. 🟡 The rewritten default-comparer path has no component-level coverage

**Location:** `packages/react/src/slider/root/SliderRoot.test.tsx:3086`

```tsx
it('[data-dirty]', async () => {
  await render(
    <Field.Root>
      <Slider.Root data-testid="root">
        <Slider.Control><Slider.Thumb /></Slider.Control>
```

With no `defaultValue` array, `Array.isArray(fieldValue)` is false and `SliderRoot.tsx:186` takes the scalar branch — so the array branch this PR rewrote is never reached, despite the test name reading as general slider dirty coverage. CheckboxGroup is worse: `grep -rn dirty packages/react/src/checkbox-group/` returns nothing, so `CheckboxGroup.tsx:129` — the only site running the default comparer over a `string[]` — has zero coverage, while every sibling Field control has a `[data-dirty]` test. (Select's array path *is* covered, at `SelectRoot.test.tsx:1096`.)

**Failure scenario:** a regression in the reverse loop or the short-circuit affecting Slider range values or CheckboxGroup dirty tracking ships green.

**Fix:** parameterize the Slider test with `defaultValue={[20, 40]}` (move a thumb, expect dirty; move it back, expect clean), and add the equivalent for CheckboxGroup with `defaultValue={['apple']}`.

## Simplifications (4)

### 1. 🟠 The rewrite costs a per-element branch and buys nothing measurable

**Location:** `packages/utils/src/areArraysEqual.ts:29`

```ts
const equal = itemComparer
  ? itemComparer(array1[i], array2[i])
  : Object.is(array1[i], array2[i]);
```

The `itemComparer ?` test is loop-invariant but sits inside the loop, so it re-evaluates on every element — the deleted default-parameter form resolved it once at call entry. Against that, the rewrite genuinely saves the `every` callback closure per call.

Whether that trade wins depends on the workload, and the workload is tiny. `areValuesEqual` only reaches `areArraysEqual` when both operands are arrays, i.e. a range slider: n = 2, occasionally 3–4. The one pointermove-adjacent site (`SliderRoot.tsx:195`) already pays a forced layout via `control.getBoundingClientRect()` plus a native event clone and a React re-render on the same tick — a two-element comparison is nanoseconds against that. Every other site runs once per committed value change inside `useValueChanged`. And at the two sites where the comparer actually does work, the "avoid the closure" saving is defeated outright: `SelectRoot.tsx:248` and `AriaCombobox.tsx:1033` allocate a fresh arrow per call regardless. The repo's only slider benchmark (`test/performance/tests/slider.bench.tsx:26`) measures mount, so nothing in CI would move either way.

**Failure scenario:** 26 lines plus a lint suppression plus a semantics change, shipped into a bundle-size-sensitive package, for a win no call site can observe.

**Fix:** restore the defaulted parameter and drop the ternary and the `equal` temp:

```ts
export function areArraysEqual<Item>(
  array1: ReadonlyArray<Item>,
  array2: ReadonlyArray<Item>,
  itemComparer: ItemComparer<Item> = Object.is,
): boolean {
  let i = array1.length;
  if (i !== array2.length) {
    return false;
  }
  while (i--) {
    if (!itemComparer(array1[i], array2[i])) {
      return false;
    }
  }
  return true;
}
```

### 2. 🟡 The referential short-circuit cannot fire at any call site

**Location:** `packages/utils/src/areArraysEqual.ts:18`

```ts
if (array1 === array2) {
  return true;
}
```

`SliderRoot.tsx:43` already tests `newValue === oldValue ||` before delegating, making the inner check structurally unreachable. The other four sites run inside `useValueChanged`, which only fires when the reference changed (`useValueChanged.ts:11`: `if (valueRef.current !== value)`), and compare against a separately-held `validityData.initialValue`. The one case that can hit — CheckboxGroup with `EMPTY_ARRAY` on both sides — would run zero iterations anyway.

**Failure scenario:** a never-taken comparison plus ~25 minified bytes on every call, and a JSDoc selling a benefit that does not exist here.

**Fix:** delete the block (already folded into the replacement above).

### 3. 🟡 The `./*` glob promotes two package-private modules to published API

**Location:** `packages/utils/package.json:15`

```json
"./*": "./src/*.ts",
```

`formatNumber.ts` and `stringifyLocale.ts` previously lived in `packages/react/src/utils/`, which is not in react's exports map — they were package-private. Landing them in `packages/utils/src/` makes `@base-ui/utils/formatNumber` (exposing `getFormatter` as well) and `@base-ui/utils/stringifyLocale` semver-bound subpaths, with no decision to publish them and no `packages/utils/CHANGELOG.md` entry.

**Failure scenario:** maintainers now owe compatibility on `getFormatter(locale, options)`, an internal `Intl` caching helper never designed as API. The follow-up phase described in the PR body makes this systemic — every utility pulled in from `@mui/x-internals` becomes public automatically, with no review gate.

**Fix:** the opt-out mechanism already exists in the same map (`"./*.test": null`). Either add explicit `null` keys for genuinely internal modules, or put them under a `src/internal/` subfolder the glob doesn't reach.

### 4. ℹ️ `isSelectedValueDirty` is duplicated verbatim in two touched files

**Location:** `packages/react/src/select/root/SelectRoot.tsx:244`

The same 11-line function appears byte-identically in `SelectRoot.tsx` and `AriaCombobox.tsx`. This predates the PR, but both files are touched here and it is the exact shape `AGENTS.md` warns against. Hoisting it into `internals/itemEquality.ts` — which already owns `compareItemEquality` and `defaultItemEquality` — would delete the duplication and let the wrapper closure be allocated once at module scope instead of per call, a larger real win than anything the `areArraysEqual` rewrite achieves.

## Docs (2)

### 1. 🟠 The new JSDoc is inaccurate on two counts and omits the semantics it changed

**Location:** `packages/utils/src/areArraysEqual.ts:6`

```
 * Short-circuits on referential equality and bails out early on a length
 * mismatch, then iterates in reverse to avoid the closure allocation of
 * `Array.prototype.every`. When no `itemComparer` is provided the elements are
 * compared with `Object.is`, inlined to skip a per-call callback.
```

Reverse iteration has nothing to do with closure allocation — a forward `for` loop allocates none either; it's the manual loop that avoids it. And "inlined to skip a per-call callback" is not what the code does: the branch is inside the loop and re-tested per element. Meanwhile the one thing a reader actually needs — that the default now treats `NaN` as equal and `-0` as distinct from `0`, unlike the old `@base-ui/react/internals/areArraysEqual` — goes unmentioned. This comment ships in the published `.d.ts`.

**Failure scenario:** a consumer reads the shipped types, assumes `===` semantics carried over, and writes an equality check that silently disagrees with the old behavior.

**Fix:** state the equality semantics and drop the unverified perf narrative. The sibling in the same directory handles attribution with a single-line source URL (`fastObjectShallowCompare.ts:1`) — match that.

### 2. 🟡 Moved test suite is still named after NumberField

**Location:** `packages/utils/src/formatNumber.test.ts:11`

```ts
describe('NumberField format', () => {
```

The module now lives in `@base-ui/utils` and serves Meter, Progress, Slider and NumberField. Failure output will misattribute to NumberField. Rename to `formatNumber`. While there, the inner `describe('formatNumber')` block only ever calls `getFormatter`, so the `if (value == null) { return ''; }` guard — the one behavior distinguishing `formatNumber` from `getFormatter(...).format` — is untested, and Progress and Meter are exactly the callers that can pass `null`.

## Verdict

**Approve after nits** — no blocking issues in these categories; the `areArraysEqual` rewrite is the one thing worth revisiting, since its perf rationale doesn't hold at any call site and its semantics change is only half pinned by tests.

---

PR #5169, opened 2026-07-04
[docs] Add Safari user-select prefix

lyzno1:
## Summary
- add the Safari `-webkit-user-select` prefix for the website Figure component

## Verification
- pnpm stylelint
- git diff --check


aarongarciah (approved):
Thanks!

PR #5171, opened 2026-07-04
[tabs] Position pre-hydration indicator inside streamed Suspense

brijeshb42:

## Problem

With streaming SSR, `<Tabs.Indicator renderBeforeHydration />` stays `[hidden]` until hydration when the tabs are below a Suspense boundary that resolves after the shell. React parses such content inside a hidden container (`<div hidden id="S:…">`), so everything measures 0×0 and the prehydration script bails permanently on its zero-size guard.

Repro: https://github.com/brijeshb42/base-ui-tabs-indicator-streaming-repro

There's also an experiment page in this PR (`/experiments/tabs/tabs-streaming-ssr`, needs `pnpm docs:dev` since it relies on server rendering) with the same setup against the repo source: a control tab bar, a streamed one, and hydration held back 5s so the pre-hydration window is easy to observe. Can be removed before merge.

https://github.com/user-attachments/assets/38f8231a-03fd-4d51-aefe-0ae79c52d875

## Fix

On zero size, observe the tabs list with a ResizeObserver: inside the hidden segment it has no box so nothing fires, and when React reveals the segment it gains one — the callback runs after layout but before paint, so the indicator still never flashes in late. It disconnects on success or when hydration gets there first. The measurable-at-parse-time path is unchanged, and browsers without ResizeObserver degrade to today's behavior (wait for hydration).

Verified with the repro in headless Chromium (hydration delayed 4s): the streamed indicator goes from `{hidden: true, width: 0}` to `{hidden: false, width: 62}` pre-hydration.

The slider thumb prehydration script has the same failure mode; left out to keep this focused. Can be updated to have same behaviour.


atomiks (commented):
## Review summary

The core rewrite looks correct. I regenerated the minified script from the template with terser (`ecma: 2020`) and it matches the committed `prehydrationScript.min.ts` byte-for-byte, and the common (non-streaming) SSR path is behaviorally identical to before (the removed `if (activeTab != null && tabsList != null)` block was already dead — both were guaranteed non-null by the earlier early-returns). The diff is a net simplification on top of the fix.

One non-blocking finding left inline: a newly-introduced ResizeObserver/DOM retention on an abandoned-before-reveal lifecycle.

**Note — test coverage:** the new ResizeObserver path has no automated coverage; the existing `pre-hydration rendering` tests only assert the script's presence, nonce, and post-hydration removal, not its runtime effect. That's hard to address in the current harness (jsdom has no layout so `offsetWidth` is always 0 and inline scripts aren't executed; the chromium env resolves `#prehydration/tabs/indicator` to the empty stub), so relying on the headless-Chromium repro + the experiment page is a reasonable call — flagging so the gap is a deliberate choice.

**Verdict: Approve after nits** — no correctness regressions; the only actionable item is the narrow-window observer retention, which can be mitigated or knowingly accepted.

---

packages/react/src/tabs/indicator/prehydrationScript.template.js
```diff
@@ -64,24 +58,48 @@
...
+
+      if (activeTab.offsetWidth > 0 && tabsList.offsetWidth > 0) {
+        observer.disconnect();
+        position();
+      }
+    });
 
-  if (width > 0 && height > 0) {
-    indicator.removeAttribute('hidden');
+    observer.observe(tabsList);
```
atomiks: 🟡 **ResizeObserver never disconnects if the streamed segment is unmounted before it's revealed**

The observer only disconnects from *inside its own callback*, and that callback only runs on a size change of `tabsList`. In the target streaming case that's fine — the reveal resizes `tabsList`, the callback fires, and it disconnects. But if the tabs subtree is removed while still `0×0` (user navigates away, or React discards the boundary, before the Suspense content is ever revealed), there's no `0×0` → size change to report, so the callback never fires again and never reaches the `!indicator.isConnected` self-heal. The browser keeps an observer with a live observation alive, and the observer holds `tabsList` (plus its `activeTab` / `indicator` / script children), so that subtree is retained for the document's lifetime.

This is a *new* risk relative to the old code, which did `return` on zero size and set nothing up. Per-occurrence it's tiny, but in a client-routed app that streams these tabs and gets navigated away mid-stream, the retained subtrees accumulate until a full page load.

Genuinely awkward to fix cleanly from an inline pre-hydration script (no unmount hook, every byte counts), so accepting it may be reasonable given the narrow window — but worth a conscious decision. If mitigating, the cheapest option is a bounded lifetime armed alongside `observe`, e.g. `setTimeout(() => observer.disconnect(), N)`, or dropping the observer after a fixed number of no-progress callbacks.

brijeshb42:
@atomiks Updated to add a lifetime timer which cancels the observer after 10sec of inactivity.

atomiks:
# PR review

Two correctness issues remain, one merge-blocking. I reproduced both DOM-level mechanisms in headless Chromium using the PR’s observer logic. I could not run the repository test suite locally.

## Bugs (2)

### 1. 🔴 Hydration can resurrect an indicator for a stale server tab

**Location:** `packages/react/src/tabs/indicator/prehydrationScript.template.js:95`

```js
const activeTab = tabsList.querySelector('[data-active]');

observer = new ResizeObserver(() => {
  if (!indicator.isConnected || !indicator.hasAttribute('hidden')) {
    stop();
    return;
  }

  if (activeTab.offsetWidth > 0 && tabsList.offsetWidth > 0) {
    stop();
    position();
  }
});
```

`activeTab` is captured when the streamed HTML is parsed, but the delayed callback never verifies that it is still active. It also treats the indicator remaining `hidden` as proof that hydration has not taken ownership.

That assumption is incorrect. The hydrated component intentionally leaves the indicator hidden when the controlled value has no matching tab, and this behavior already has explicit test coverage. The observer can nevertheless call `position()` using the old server-side tab and remove `hidden`.

**Failure scenario:** The server renders tab `"one"` as selected inside a streamed boundary. Before the first `ResizeObserver` delivery, hydration or an immediate controlled update changes the value to `"missing"`. React removes `data-active` and correctly keeps the indicator hidden. The callback still holds the old tab element, sees that it has a width, positions the indicator over it, and unhides it. With no subsequent resize or render, the stale indicator can remain visible indefinitely.

**Fix:** Revalidate ownership and selection before mutating:

```js
const currentActiveTab = tabsList.querySelector('[data-active]');

if (currentActiveTab !== activeTab) {
  stop();
  return;
}
```

Capturing `document.currentScript` and stopping when that script is removed by hydration would provide an additional ownership signal. Re-querying the current active tab inside `position()` would be safer still.

### 2. 🟠 The observer disconnects before positioning has actually succeeded

**Location:** `packages/react/src/tabs/indicator/prehydrationScript.template.js:102`

```js
if (activeTab.offsetWidth > 0 && tabsList.offsetWidth > 0) {
  stop();
  position();
}
```

The callback considers only width before permanently disconnecting, while `position()` only reveals the indicator when both width and height are positive. The observer also watches only `tabsList`, not `activeTab`.

**Failure scenario:** A streamed list has a fixed nonzero size, while an active icon-only tab initially has a CSS/flex width but zero height until its image or styles resolve. Revealing the segment makes the width check pass, so the observer disconnects. `position()` records a zero height and leaves the indicator hidden. When the tab later gains height without changing the fixed-size list, nothing is observed and the indicator stays hidden until hydration.

**Fix:** Have `position()` return whether it successfully revealed the indicator and only call `stop()` after success. Observe `activeTab` as well as `tabsList`, or require all relevant width and height measurements to be positive before disconnecting.

## Tests (1)

### 1. 🟠 The new runtime observer path has no automated coverage

**Location:** `packages/react/src/tabs/indicator/TabsIndicator.test.tsx:431`

The new runtime observer path has no automated coverage. The existing pre-hydration tests verify only that the script is emitted, has contents and a nonce, survives hydration, and is subsequently removed. None executes the script or exercises its delayed `ResizeObserver` branch.

**Failure scenario:** Both bugs above, or a future regression where revealing the streamed segment no longer positions the indicator, pass the entire current test suite.

**Fix:** Add a Chromium test or E2E fixture using the real generated script that covers:

1. A hidden streamed segment becoming visible before hydration.
2. Hydration changing the selected value to one with no matching tab before observer delivery.
3. The active tab becoming fully measurable after the list has already become visible.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Request changes** - the delayed observer can override hydrated state and display an indicator for a tab that is no longer active.

---

🤖 Review generated with Codex

atomiks (approved):
# PR review

No correctness regressions or merge-blocking issues found. The implementation is appropriately narrow overall, but one observer target appears unnecessary.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (1)

### 1. 🟡 Observe only the active tab

**Location:** `packages/react/src/tabs/indicator/prehydrationScript.template.js:114`

```js
observer.observe(tabsList);
observer.observe(activeTab);
```

Observing `tabsList` appears redundant. `updatePosition()` returns `false` only while the captured active tab has zero width or height, and it can return `true` only after that tab’s box becomes measurable.

Revealing the streamed Suspense segment changes the active tab from `0×0` to its rendered dimensions, so observing `activeTab` already covers the primary case. It also covers the independently-delayed width or height case exercised by the new test.

**Failure scenario:** There is no behavioral failure, but each pre-hydration indicator installs an additional observation that cannot independently make the positioning retry succeed.

**Fix:** Remove the list observation and retain only:

```js
observer.observe(activeTab);
```

The initial measurement, hydration guards, stale-selection guard, bounded timeout, and success-returning `updatePosition()` are otherwise all justified.

## Docs (0)

No findings.

## Verdict

**Approve after nits** - the implementation is correct, but the redundant `tabsList` observation can be removed to make the fix as small as possible.

---

🤖 Review generated with Codex

PR #5174, opened 2026-07-06
[docs] URL updates for SEO

aarongarciah:
Updates all docs URLs to remove the "section" segment. This could lead to clashing URLs in the future, but it's an unlikely scenario.

IMPORTANT: the `overview`, `handbook`, `components`, and `utils` have been renamed to `(overview)`, `(handbook)`, etc. so they don't participate in the URL ([Next.js route groups](https://nextjs.org/docs/app/api-reference/file-conventions/route-groups)).

Some examples:

|old|new|
|---|---|
|`/react/overview/quick-start`|`/react/quick-start`|
|`/react/handbook/styling`|`/react/styling`|
|`/react/components/accordion`|`/react/accordion`|
|`/react/utils/csp-provider`|`/react/csp-provider`|

A bunch of index pages have been removed, and now they redirect to the first page in the section:

|old|new|
|---|---|
|`/react/overview`|`/react/quick-start`|
|`/react/handbook`|`/react/styling`|
|`/react/components`|`/react/accordion`|
|`/react/utils`|`/react/csp-provider`|

Preview: https://deploy-preview-5174--base-ui.netlify.app/


aarongarciah:
@Janpot may I ask you to take a look at how the `/react/page.mdx` index is being generated? I added a new script to do it. I couldn't find a way to support it using code-infra now that "components", "handbook", etc. are Next.js route groups (`(components)`) and they don't count toward the URL creation.

Janpot:
> @Janpot may I ask you to take a look at how the /react/page.mdx index is being generated? I added a new script to do it. I couldn't find a way to support it using code-infra now that "components", "handbook", etc. are Next.js route groups ((components)) and they don't count toward the URL creation.

Taking a [look](https://github.com/mui/mui-public/pull/1640). To be honest, at the moment this package is as much a black box to me as it is to you. I see you immediately reach for a prebuild script to implement this behavior. Would that be your preferred way of achieving the generated parts of the docs? Over doing it as side effect in turbopack loaders? Personally I find it a rather more straightforward paradigm to reason about.
Second thing that makes things rather complicated is using markdown as a data interchange format that is both human and machine editable.

aarongarciah:
> Over doing it as side effect in turbopack loaders? 

I have no hard preference. I think whatever approach blends better with our current/future docs-infra approach would be the best one.

Janpot:
> I think whatever approach blends better with our current/future docs-infra approach would be the best one.

If the kind of ad-hoc temporary overrides like this PR does become regular practice, I consider that a failure of the docs-infra abstractions. Doesn't mean we have to turn them around tomorrow, but it's a strong signal.

aarongarciah:
Let’s wait for @colmtuite and @oliviertassinari to confirm if we’re on the same page regarding the new URLs.

colmtuite:
My understanding was that we agreed on:

```
/styling
/react-accordion
```

aarongarciah:
> ```
> /styling
> /react-accordion
> ```

@colmtuite a conversation with an LLM pointed out that `/react/accordion` was equivalent and maybe a bit more natural than `/react-accordion`, but if we are removing `/react` from the non-component pages it's fine to use `react-`.

@oliviertassinari are you aware of any potential differences between both approaches? I'd like to avoid  adding new redirects in the near future.

docs/public/_redirects
```diff
@@ -24,3 +24,163 @@
 # At the time of adding this rule, there was requirement of at max only a single args[] param
 /production-error/:code args\[\]=:args /production-error?code=:code&args[]=:args 301
 /production-error/:code /production-error?code=:code 301
+
+# 2026
+
+# React docs URL migration
+/react/overview /react/quick-start 301!
```
oliviertassinari: Normally the! are here so that even if there is a html file under that slug, the rule takes precedence. There should never be a file, so that ! seems not necessary, and could hide problem. So, I'm proposing 

```suggestion
/react/overview /react/quick-start 301
```

docs/public/_redirects
```diff
@@ -24,3 +24,163 @@
...
+/react/overview/about.md /react/about.md 301!
+/react/overview/accessibility /react/accessibility 301!
+/react/overview/accessibility.md /react/accessibility.md 301!
+/react/overview/community /react/community 301!
+/react/overview/community.md /react/community.md 301!
+/react/overview/quick-start /react/quick-start 301!
+/react/overview/quick-start.md /react/quick-start.md 301!
+/react/overview/releases /react/releases 301!
+/react/overview/releases.md /react/releases.md 301!
+/react/overview/releases/v1-0-0 /react/releases/v1-0-0 301!
```
oliviertassinari: We should use pattern matching here, no? For example: https://github.com/mui/material-ui/blob/ba0c0afa128335250ba63339c70f5e499cac0071/docs/public/_redirects#L51

docs/scripts/generateLlmTxt/index.mjs
```diff
@@ -23,6 +23,45 @@ const NETLIFY_DEPLOYMENT_URL =
   process.env.PULL_REQUEST === 'true' ? process.env.DEPLOY_PRIME_URL : process.env.URL;
 // Use the deployment URL if available, just root relative otherwise
 const BASE_URL = NETLIFY_DEPLOYMENT_URL || '/';
+const REACT_DOCS_ROUTE_MAPPINGS = [
```
oliviertassinari: This seems strange, the `REACT_DOCS_ROUTE_MAPPINGS` seems to exist twice. And the logic seems custom. Can't we rely a lot more on the folder structure to encode the logic? In https://github.com/mui/base-ui/issues/1615#issuecomment-4416507869 I was arguing for more exceptions, but so if we normalize this more, it looks like we can simplify this logic even more.

README.md
```diff
@@ -6,15 +6,15 @@ From the creators of Radix, Floating UI, and Material UI, Base UI is an unsty
 
 ## Documentation
 
-To get started, check out the [Base UI documentation](https://base-ui.com/react/overview/quick-start).
+To get started, check out the [Base UI documentation](https://base-ui.com/react/quick-start).
```
oliviertassinari: To move to a second PR, to merge after the next release.

oliviertassinari (commented):
> @oliviertassinari are you aware of any potential differences between both approaches? I'd like to avoid adding new redirects in the near future.

@aarongarciah I also explored this with Gemini https://share.gemini.google/yBkuF1ZKmnEo. I couldn't find much difference between the options in the short term. What seems important is that everything is prefixed with the prefix of the framework for consistency in the site architecture, assuming content about styling is specific to React or it will be duplicated for each framework.

In the long term, the current state of the PR (/react/) look better because it allows to accumulate framework specific authority to the subfolder (which is key for primitives), to organize source more easily, and to have a clearer site hierarchy in the Google result page. 

For the advanced components, we want to treat each a lot more as individual products than the small components https://app.notion.com/p/mui-org/product-2025-GTM-Brand-Strategy-for-our-components-1ddcbfe7b66080179b7dd32b45ba1198#1ddcbfe7b66080c9a1cdc1bfa1d9f35c

If we go with a subdomain, I could see something like this work ok
- base-ui.com/primitives/react/slider
- data-grid.base-ui.com/react-data-grid/lazy-loading

If we go with a subfolder, we might need to change to something like this:
- base-ui.com/primitives/react/slider
- base-ui.com/data-grid/react/lazy-loading

But overall, to answer the question: yes it's important that it doesn't change, Google index needs 4-8 months to adapt. And because this depends on the overall brand architecture, it seems that this is stuck until the overall architecture is fully defined.

CONTRIBUTING.md
```diff
@@ -72,11 +71,11 @@ There are, however, a few differences:
    <TypesCheckbox.Indicator />
    ```
 
-4. Start the dev server with `pnpm docs:dev` and visit the page (for example, `http://localhost:3005/react/components/toggle`). The `types.md` file is generated automatically. If you prefer not to open the browser, run `pnpm docs:build` to generate all files.
+4. Start the dev server with `pnpm docs:dev` and visit the page (for example, `http://localhost:3005/react/toggle`). The `types.md` file is generated automatically. If you prefer not to open the browser, run `pnpm docs:build` to generate all files.
```
aarongarciah:
```suggestion
4. Start the dev server with `pnpm docs:dev` and visit the page (for example, `http://localhost:3005/react/button`). The `types.md` file is generated automatically. If you prefer not to open the browser, run `pnpm docs:build` to generate all files.
```

docs/scripts/generateLlmTxt/index.mjs
```diff
@@ -11,6 +11,9 @@ import remarkParse from 'remark-parse';
 import remarkGfm from 'remark-gfm';
 import remarkStringify from 'remark-stringify';
 import { visit } from 'unist-util-visit';
+// Node ESM requires the extension; this script runs with --experimental-strip-types.
+// eslint-disable-next-line import/extensions
+import { getCanonicalReactDocsUrl } from '../../src/utils/canonicalReactDocsUrl.ts';
```
aarongarciah, line 16: @Janpot does this look good?

docs/public/_redirects
aarongarciah: Updated

docs/public/_redirects
aarongarciah: Updated

docs/scripts/generateLlmTxt/index.mjs
aarongarciah: Updated

README.md
aarongarciah: Updated

docs/scripts/generateLlmTxt/index.mjs
Janpot: I can live with this, alternatively, you can run it with `tsx`

docs/src/utils/canonicalReactDocsUrl.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function isReactDocsSection(section: string | undefined): section is ReactDocsSection {
+  return section !== undefined && Object.hasOwn(REACT_DOCS_SECTION_DEFAULTS, section);
+}
+
+/**
+ * Converts internal route-group paths and legacy section URLs to their public React docs URL.
+ */
+export function getCanonicalReactDocsUrl(url: string) {
```
Janpot, line 29: 🤔  I imagine there could already be some functions in `@mui/internal-docs-infra` that can do this. 

_Edit: I'm adding a helper in docs infra._
_Edit: nvm, this seems fine for now_

aarongarciah:
> If we go with a subdomain, I could see something like this work ok
> 
> * base-ui.com/primitives/react/slider
> * data-grid.base-ui.com/react-data-grid/lazy-loading
> 
> If we go with a subfolder, we might need to change to something like this:
> 
> * base-ui.com/primitives/react/slider
> * base-ui.com/data-grid/react/lazy-loading
> 
> But overall, to answer the question: yes it's important that it doesn't change, Google index needs 4-8 months to adapt. And because this depends on the overall brand architecture, it seems that this is stuck until the overall architecture is fully defined.

The team was chatting today and it seems unlikely that we end up adding `/primitives` to the url. In that case, this PR feels safe to merge if we want to get the (potential?) SEO benefits.

PR #5178, opened 2026-07-07
[menu] Fix duplicate onOpenChange calls when closing a submenu

atomiks:

When a submenu item is clicked, the submenu's `onOpenChange(false)` currently fires three times for a single close (reproducible on published 1.6.0 and master with trusted pointer input):

1. The legit close from the tree-wide `close` event emitted by the item press.
2. A duplicate: when the root closes, the child's `onParentClose` relay in `MenuPositioner` closes the already-closed submenu again. The redundancy guard in `setOpen` doesn't catch it because the relayed event arrives in the same tick as the first close (the render-closure `open` is still `true`), and its `trigger` is compared against `activeTriggerElement` before the trigger normalization runs.
3. A stale `trigger-hover` close from a pending hover timer, which passes the guard because its reason differs from the last one.

## Fix

`setOpen` now ignores a close request when the menu is already closed, reading the synchronous store state (`store.select('open')`) rather than the render closure so same-tick relays are caught too. Since the `open` selector resolves to `openProp ?? state.open`, a controlled menu whose owner rejects a close keeps receiving close events, and the intentional same-open refire used for detached-trigger switching is unaffected (the guard only applies when the next state is closed).

The regression test asserts each of the root's and submenu's `onOpenChange` receives exactly one `false` call in the nested item-click scenario; it fails 2/2 (contained + detached triggers) without the fix. The item click is dispatched with `fireEvent` because the synthetic pointer teleport of `user.click` closes the submenu via hover before the click lands.

### DIFF
--- a/packages/react/src/menu/root/MenuRoot.test.tsx
+++ b/packages/react/src/menu/root/MenuRoot.test.tsx
@@ -817,6 +817,50 @@ describe('<Menu.Root />', () => {
       });
 
       it.skipIf(isJSDOM)(
+        'calls onOpenChange with false exactly once per menu when a submenu item is clicked',
+        async () => {
+          const rootOnOpenChange = vi.fn();
+          const submenuOnOpenChange = vi.fn();
+
+          const { user } = await render(
+            <TestMenu
+              rootProps={{ onOpenChange: rootOnOpenChange }}
+              submenuProps={{ onOpenChange: submenuOnOpenChange }}
+              submenuTriggerProps={{ delay: 0 }}
+            />,
+          );
+
+          await user.click(screen.getByRole('button', { name: 'Toggle' }));
+          await screen.findByTestId('menu');
+
+          const submenuTrigger = await screen.findByTestId('submenu-trigger');
+          await user.hover(submenuTrigger);
+
+          await waitFor(() => {
+            expect(screen.queryByTestId('submenu')).not.toBe(null);
+          });
+
+          // Click the item directly. Moving the pointer to it teleports it out of the
+          // submenu trigger, closing the submenu before the click could land.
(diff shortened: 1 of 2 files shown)


flaviendelangle (approved):
# PR review

This PR extracts the three duplicated trigger slip-out boundary checks (menu, select, combobox) into a single shared `isMouseWithinBounds` util and widens the drift tolerance from 2px to 5px. The refactor is faithful: all three call sites already null-guard the trigger element before calling, the boundary math is unchanged apart from the intended widening, and no inline copies remain. Nothing merge-blocking.

## Bugs (0)

No findings.

## Tests (1)

### 1. ℹ️ Menu and select tolerance change covered only indirectly

**Location:** `packages/react/src/utils/getPseudoElementBounds.test.ts:60`

The behavior change (2px → 5px) applies to `MenuTrigger` and `SelectTrigger` too, but only `ComboboxTrigger` gained boundary integration tests. The widened math is exercised by the new shared-util unit test, so menu/select are covered at the logic level — just not at the component-interaction level.

**Failure scenario:** A future change to how `MenuTrigger`/`SelectTrigger` wire up `handleMouseUp` (e.g. passing the wrong element or dropping the call) would not be caught by a dedicated test; the shared unit test would still pass. Low risk given the parallel combobox tests.

**Fix:** Optional — mirror the combobox `within/outside 5px` tests for menu and select, or leave as-is since the shared util is directly tested.

## Simplifications (0)

No findings.

## Docs (0)

No findings. The new JSDoc and the tolerance comment accurately describe the util.

## Verdict

**Approve** — clean, well-tested consolidation; the only note is an optional test-coverage parity suggestion for menu/select.

---

PR #5179, opened 2026-07-07
[drawer] Fix scroll handling when focus moves while the virtual keyboard is open

atomiks:
Steps (bug A):
1. [Virtual keyboard aware demo](https://deploy-preview-5179--base-ui.netlify.app/react/components/drawer#virtual-keyboard-aware)
2. Tap an input on iOS
3. Press arrow navigation keys to go between inputs (up/down)
4. Should not use native handling

Steps (bug B):
1. [Virtual keyboard aware demo](https://deploy-preview-5179--base-ui.netlify.app/react/components/drawer#virtual-keyboard-aware)
2. Tap the footer input on iOS
3. Scroll the drawer container down to reach textarea
4. Tap textarea
5. Should correctly scroll it into view

With the keyboard already open, moving focus without a fresh tap broke the drawer's scroll containment:

- The iOS keyboard's previous/next arrows move focus natively (no touch events, no `preventScroll`), so WebKit's reveal scroll panned the page and pushed the drawer offscreen.
- Tapping a field in the scrollable body while a pinned footer field held focus didn't scroll it into view — no viewport events follow, and the single alignment pass measured layout still reacting to the focus change.

Fixes in `DrawerVirtualKeyboardProvider`:

- Preempt the native reveal in `focusout`: briefly present the incoming field as already visible above the keyboard so WebKit has nothing to reveal, restoring in `focusin`.
- Commit the centering scroll only once its destination stops moving between checks, so it never chases mid-transition layout.
- Re-run alignment on a short interval after keyboard-open focus changes, re-issuing only scrolls that stalled (WebKit cancels in-flight smooth scrolls when its reveal resolves) — never restarting one in progress.
- Pin the page scroll for modal drawers as a recovery net.

Verified on a physical iPhone (taps, arrows both directions, footer→body transitions, keyboard open/close); each mechanism has Chromium tests with a mocked visual viewport.

### DIFF
--- a/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.test.tsx
+++ b/packages/react/src/drawer/virtual-keyboard-provider/DrawerVirtualKeyboardProvider.test.tsx
@@ -1528,6 +1528,760 @@ describe('<Drawer.VirtualKeyboardProvider />', () => {
   );
 
   it.skipIf(isJSDOM)(
+    'overrides the incoming field geometry when focus moves without a tap while the keyboard is open',
+    async () => {
+      const restoreInnerHeight = mockWindowInnerHeight(800);
+      const visualViewport = mockVisualViewport(800);
+
+      try {
+        await render(
+          <Drawer.Root open modal={false}>
+            <Drawer.VirtualKeyboardProvider>
+              <Drawer.Portal>
+                <Drawer.Viewport>
+                  <Drawer.Popup>
+                    <input data-testid="first" type="text" />
+                    <input data-testid="second" type="text" />
+                  </Drawer.Popup>
+                </Drawer.Viewport>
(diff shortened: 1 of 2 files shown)


aarongarciah:
Very nice. Tested on Safari 26.5 and it seems to work fine.

I wonder if it'd be desirable or if it's possible to avoid the animated scroll when focusing the textarea in the video below?

https://github.com/user-attachments/assets/ce98cd66-98e0-4398-91b2-f81a445382e1

atomiks:
@aarongarciah yeah I thought that looked kinda bad but it at least fixed the previous broken behavior.

If there's a way to detect this particular edge case and disable scroll animations, I will apply it.

atomiks:
Wasn't able to get that working smoothly. It's not severe and the current fix is better than before, so deferring to an issue or community fix.

PR #5180, opened 2026-07-07
[docs] Fix scrollable focus ring

lyzno1:
## Summary
- draw the scrollable API cell focus ring on the outer cell
- hide the overflow fade while the cell is keyboard-focused

## Screenshots
| Before | After |
| --- | --- |
| Before focus ring | After focus ring |

## Test plan
- pnpm stylelint docs/src/components/Accordion.css
- pnpm prettier docs/src/components/Accordion.css --check
- git diff --check -- docs/src/components/Accordion.css

### DIFF
--- a/docs/src/components/Accordion.css
+++ b/docs/src/components/Accordion.css
@@ -119,6 +119,22 @@
   .AccordionScrollable {
     overflow-x: hidden;
 
+    /* stylelint-disable-next-line selector-pseudo-class-disallowed-list */
+    &:has(.AccordionScrollableInner:focus-visible) {
+      position: relative;
+      z-index: 1;
+
+      &::before {
+        content: '';
+        position: absolute;
+        pointer-events: none;
+        inset: 0;
+        z-index: 1;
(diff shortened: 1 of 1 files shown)


docs/src/components/Accordion.css
```diff
@@ -119,6 +119,17 @@
...
 
+    /* stylelint-disable-next-line selector-pseudo-class-disallowed-list */
+    &:has(.AccordionScrollableInner:focus-visible) {
+      z-index: 1;
+      outline: 2px solid var(--gray-t2);
+      outline-offset: -1px;
+
+      &::after {
+        display: none;
+      }
```
aarongarciah: Can we find a solution that doesn't require hiding the scroll indicator?

lyzno1: I’ll keep the scroll indicator and move the focus ring to a separate `::before` overlay, so `::after` can continue rendering the overflow fade.
CleanShot 2026-07-07 at 16 24 16

aarongarciah (approved):
Thanks!

PR #5186, opened 2026-07-08
[docs] Add shared Base UI review skill

atomiks:
Adds a shared Base UI review skill so agents can use and improve the same review workflow from the repo.

The skill reviews diffs across bugs/regressions, tests, simplifications, and docs. It also calls out when API design or runtime performance specialists should join the review.

## Changes

- Add the `base-ui-review` skill under `.agents/skills`.
- Add OpenAI skill metadata for the shared skill.
- Add a Claude Code discovery entrypoint under `.claude/skills/base-ui-review`.
- Document the shared skill location in `AGENTS.md`.

## Usage

Invoke it as `/base-ui-review [effort] [--fix] [--comment [inline]] [target]`. Plain `--comment` posts one top-level PR comment; `--comment inline` posts inline finding comments when they map to PR diff lines and falls back to a top-level comment for non-diff findings.

Effort controls review depth and cost:

- `low`: Fastest bug-only pass.
- `medium` (default): One agent checks bugs, tests, simplifications, docs, and any triggered API design/performance concerns; may add a bug/regression subagent for large or risky diffs.
- `high`: Separate subagents for each review area, plus API design/performance specialists when triggered.
- `xhigh`: `high` plus a fresh sweep for missed findings.
- `max`: `xhigh` plus verifier subagents for surviving candidates.

Use higher effort when missed regressions would be expensive.

### DIFF
--- a/.agents/skills/base-ui-review/SKILL.md
+++ b/.agents/skills/base-ui-review/SKILL.md
@@ -1,22 +1,23 @@
 ---
 name: base-ui-review
-description: 'Review the current diff for regressions, correctness bugs, tests, simplifications, and docs issues, scaling depth to a low/medium/high effort level. Use when the user asks to review changes, review a diff/branch/PR, or runs /base-ui-review. Pass --comment to post findings as inline PR comments, or --fix to apply the findings to the working tree after the review.'
+description: 'Review the current diff for regressions, correctness bugs, tests, simplifications, and docs issues, scaling depth to a low/medium/high/xhigh/max effort level. Use when the user asks to review changes, review a diff/branch/PR, or runs /base-ui-review. Pass --comment to post findings as inline PR comments, or --fix to apply the findings to the working tree after the review.'
 ---
 
 # Base UI Review
 
 Review the current diff and report **regressions and correctness bugs** alongside
 **cleanup** (reuse / simplification / efficiency), scaling depth to the chosen effort
-level. The effort level (`low` / `medium` / `high`, default `medium`) controls how
-many areas are reviewed, whether work fans out to subagents, and the precision/recall
-bias - see [Effort levels](#effort-levels) for the canonical behavior of each.
+level. The effort level (`low` / `medium` / `high` / `xhigh` / `max`, default
+`medium`) controls how many areas are reviewed, whether work fans out to subagents,
+and the precision/recall bias - see [Effort levels](#effort-levels) for the
+canonical behavior of each.
 
-Argument hint: `[low|medium|high] [--fix] [--comment] [<target>]`
+Argument hint: `[low|medium|high|xhigh|max] [--fix] [--comment] [<target>]`
 
 You are reviewing for **precision** at medium effort: every finding you surface
-should be one a maintainer would act on. At **high** the bias shifts to **recall**
-to catch every real bug a careful reviewer would catch; a missed bug ships, so err
-on the side of surfacing (uncertain findings are allowed, labeled).
+should be one a maintainer would act on. At **high**, **xhigh**, and especially
+**max**, the bias shifts toward **recall**; a missed bug ships, so surface uncertain
+findings when the mechanism is realistic and label them clearly.
 
 ## Scope
 
@@ -28,15 +29,16 @@ that target instead.
 
 ## Phase 1 - Find candidates
 
-Review the four main areas below: Bugs, Tests, Simplifications, and Docs. At
-`medium` (default) do this locally in the main agent with no subagents; at `high`
-run those four areas as **independent subagents**. At `low`, only Area 1 (Bugs)
-applies, and only its high-confidence runtime-correctness subset - see
-[Effort levels](#effort-levels). Each area surfaces actionable candidate findings
-with `file`, `line`, a one-line `summary`, a `severity` (🔴 / 🟡 / ℹ️ - see
-[Output](#output) for the rubric), and a concrete `failure_scenario`. Do NOT let
-one area's conclusions suppress another's - if two areas flag the same line for
-different reasons, record both.
+Review the four main areas below: Bugs, Tests, Simplifications, and Docs. At `low`,
+only Area 1 (Bugs) applies, and only its high-confidence runtime-correctness subset.
+At `medium` (default), do this locally in the main agent; if the change is large,
+risky, or touches fragile behavior, also spawn an independent bug/regression hunter
+subagent. At `high`, `xhigh`, and `max`, run the four areas as **independent
+subagents** - see [Effort levels](#effort-levels). Each area surfaces actionable
+candidate findings with `file`, `line`, a one-line `summary`, a `severity` (🟣 / 🔴 /
+🟠 / 🟡 / ℹ️ - see [Output](#output) for the rubric), and a concrete
+`failure_scenario`. Do NOT let one area's conclusions suppress another's - if two
+areas flag the same line for different reasons, record both.
 
 Pass every candidate with a nameable failure scenario through - finders that
 silently drop half-believed candidates bypass the verify step and are the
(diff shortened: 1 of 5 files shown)


.agents/skills/base-ui-review/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+Prefix every finding title with a severity marker:
+
+- 🔴 **high / critical - blocking.** A regression or correctness bug that breaks
+  behavior, loses data, crashes, or ships a broken public contract. Should block
+  merge until fixed.
+- 🟡 **low / medium - non-blocking.** A real issue worth fixing but not
+  merge-blocking: a minor bug on a rare path, a simplification, a missing test, or
+  stale docs.
+- ℹ️ **note.** Informational - an observation, heads-up, or optional suggestion the
+  maintainer may reasonably decline.
```
atomiks: Maybe it would be nice to granularize this a bit more

🟣 - Critical
🔴 - High
🟠 - Medium
🟡 - Low
ℹ️ - Note

.agents/skills/base-ui-review/SKILL.md
```diff
@@ -0,0 +1,325 @@
...
+
+# Base UI Review
+
+Review the current diff and report **regressions and correctness bugs** alongside
+**cleanup** (reuse / simplification / efficiency), scaling depth to the chosen effort
+level. The effort level (`low` / `medium` / `high`, default `medium`) controls how
+many areas are reviewed, whether work fans out to subagents, and the precision/recall
+bias - see [Effort levels](#effort-levels) for the canonical behavior of each.
+
+Argument hint: `[low|medium|high] [--fix] [--comment] [<target>]`
```
atomiks: `high` is super expensive, which might be another thing to granularize more?

- `low` - bug fixes only
- `medium` - default, one agent checks the four areas. Might be worth still running a separate agent for bugs too in this case, in case it causes it to skim too much - based on what @michaldudak said before
- `high` - independent subagents, no verifier or sweep re-check
- `xhigh` - this one? Or `max`?

.agents/skills/base-ui-review/SKILL.md
aarongarciah: Sounds good

atomiks: Applied this in a new commit.

.agents/skills/base-ui-review/SKILL.md
atomiks: Applied this in a new commit.

aarongarciah (commented):
# PR review

This PR ships prompt/config files only — no runtime code — so risk is limited to skill discovery and agent behavior. Nothing is merge-blocking. The two findings worth addressing are behavioral: the `--comment`/`--fix` sections are written as if those flags were always passed, and the `.claude/skills` symlink is the first symlink in the repository, which silently degrades on Windows checkouts. Formatting and lint gates were verified locally (prettier with the repo config, vale, and ESLint/remark all pass on the new files).

## Bugs (2)

### 1. 🟡 `--comment` and `--fix` sections assert the flags were passed unconditionally

**Location:** `.agents/skills/base-ui-review/SKILL.md:325`

```md
## Posting to GitHub (--comment)

The `--comment` flag was passed. After producing the findings list, if the review
target is a GitHub PR, post each finding as an inline PR comment via `gh api`
```

The entire SKILL.md is loaded into the agent's context on every invocation, so on a plain `/base-ui-review` run the agent still reads the flat statements "The `--comment` flag was passed." and "The `--fix` flag was passed." Only the section headings hint that these are conditional. A less careful model can act on the assertion — posting inline comments to a public GitHub PR or editing the working tree without being asked. Posting comments is an outward-facing, hard-to-retract action, which makes this the most consequential wording issue in the file.

**Failure scenario:** A maintainer runs `/base-ui-review <PR link>` with no flags in an auto-accepting permission mode; the agent reaches the `--comment` section, takes "The `--comment` flag was passed" at face value, and posts review comments to the upstream PR uninvited.

**Fix:** Make both sections explicitly conditional, e.g. "If the `--comment` flag was passed in the arguments, then after producing the findings list…" and "If the `--fix` flag was passed…", and state that with no flags the review stops at the report.

### 2. 🟡 `.claude/skills/base-ui-review` symlink silently breaks on Windows checkouts

**Location:** `.claude/skills/base-ui-review:1`

```text
../../.agents/skills/base-ui-review
```

This is the first symlink ever committed to this repository (`git ls-files -s` shows no other mode-120000 entries), so no contributor workflow has exercised symlinks here before. On Windows, git checks out symlinks as plain text files containing the target path unless `core.symlinks=true` and Developer Mode/admin rights are enabled — which is not the default. On such checkouts `.claude/skills/base-ui-review` becomes a regular file, Claude Code finds no `SKILL.md` under it, and `/base-ui-review` is silently unavailable with no error to explain why. The relative target itself is correct (`../../` resolves to the repo root), and prettier does not follow the symlink, so nothing else breaks — the skill just vanishes for that class of contributors.

**Failure scenario:** A contributor on Windows clones the repo, runs `/base-ui-review`, and gets "unknown skill" with no indication that a degraded symlink is the cause.

**Fix:** Either accept and document the limitation in the new AGENTS.md section ("requires symlink support; on Windows enable `core.symlinks`"), or replace the symlink with a real `.claude/skills/base-ui-review/SKILL.md` stub whose frontmatter matches and whose body instructs the agent to read and follow `.agents/skills/base-ui-review/SKILL.md`.

## Verdict

**Approve after nits** — no correctness regressions, but the unconditional `--comment`/`--fix` wording can trigger unrequested outward-facing actions and is a one-line fix.

---

aarongarciah:
## Bugs (1)

### 1. 🟡 Stub skill description drops the trigger guidance, weakening auto-invocation in Claude Code

**Location:** `.claude/skills/base-ui-review/SKILL.md:3`

```md
description: 'Review Base UI diffs across bugs, tests, simplifications, docs, and triggered API design or performance concerns.'
```

Claude Code decides whether to invoke a skill from its frontmatter `name` and `description` alone — the body is only read after invocation. The canonical skill's description deliberately includes trigger phrasing ("Use when the user asks to review changes, review a diff/branch/PR, or runs /base-ui-review. Pass --comment to post a top-level PR comment, --comment inline for inline PR comments, or --fix to apply findings."), but the stub — the only surface Claude Code actually sees — omits it. Explicit `/base-ui-review` invocations still work (name-based), but model-initiated triggering on requests like "review my changes before I commit" depends entirely on this shorter description, and the flag/effort vocabulary is invisible until after the skill fires.

**Failure scenario:** A contributor asks Claude Code "can you review this PR?" without typing the slash command; the model doesn't match the stub's terser description and runs an ad-hoc review instead of the shared workflow the PR exists to standardize.

**Fix:** Copy the canonical description's trigger sentence into the stub's frontmatter description (they can be identical — the stub body already handles the delegation).

## Docs (1)

### 1. ℹ️ Severity rubric gives conflicting cues for stale public docs

**Location:** `.agents/skills/base-ui-review/SKILL.md:227`

```md
- 🟠 **medium - non-blocking.** A real issue worth fixing before merge when
  practical but not normally blocking: a narrow or uncommon bug, meaningful missing
  test, stale public docs, or cleanup that avoids likely future mistakes.
```

The 🟠 definition names "stale public docs" as a canonical medium example, while the tie-break paragraph a few lines below instructs "choose the higher severity if the failure reaches … a public API/docs contract" and the 🔴 definition includes "a public docs example". A reviewer model handed a stale-docs finding gets pushed toward 🟠 by the example and toward 🔴 by the tie-break, so severities for the same class of finding may drift between runs.

**Failure scenario:** Two runs of `/base-ui-review` on the same diff rate an outdated docs page 🟠 in one pass and 🔴 (changing the verdict to Request changes) in the other.

**Fix:** Draw the line explicitly, e.g. 🔴 for a public docs example that no longer compiles or teaches broken API usage, 🟠 for stale prose that doesn't break copy-pasted code.

## Verdict

**Approve after nits** — only remaining issue is the stub's weakened trigger description, a one-line metadata fix.

---

jjenzz (approved):
thanks for doing this, i've not reviewed too closely because something is better than nothing at this point. we can iterate if needed 🙂

aarongarciah (approved):

PR #5188, opened 2026-07-08
[checkbox] Remove redundant validation input ref

nami8824:

My understanding is that `validation.inputRef` is redundant here.

`Checkbox.Root` already registers its hidden input through `validation.registerInput`, and validation uses the registered inputs first before falling back to `validation.inputRef`.


atomiks (approved):
Thanks @nami8824

PR #5189, opened 2026-07-08
[menu] Add initialFocus prop to Menu.Popup

kylemcd:
`Menu.Popup` hardcodes initial focus on its focus manager, so there's no way to stop the popup grabbing focus on open. `Popover.Popup` and `Dialog.Popup` already expose `initialFocus` for exactly this — this adds the same prop to `Menu.Popup`.

motivation: a trigger that renders a typeable input (combobox-style) can't keep focus when the menu opens, since the popup always pulls focus inside. `initialFocus={false}` lets the consumer opt out.

- forwards `initialFocus` to `FloatingFocusManager`, same type/JSDoc as `Popover`/`Dialog`
- default is unchanged: `undefined` still resolves to `parent.type !== 'menu'`, so it's a no-op unless set
- deliberately did not reuse `createDefaultInitialFocus` like popover/dialog — that adds touch/virtual-keyboard handling menus don't have today, which would change behavior
- adds tests + regenerates the `types.md` tables (menu + context-menu)

## test plan

`pnpm test:jsdom` on `MenuPopup.test.tsx` — 23 pass, incl. two new cases: default focuses the first item, and `initialFocus={false}` keeps focus on the trigger


aarongarciah:
@kylemcd can you share some screenshots/wireframes of what you're trying to build to better understand the motivation? If I understood correctly, you are using an input as a Menu trigger, but I want to understand why you didn't reach for [Autocomplete](https://base-ui.com/react/components/autocomplete) in this case.

aarongarciah:
I just noticed https://telegraph-storybook-git-kyle-kno-14086-telegra-21a570-knocklabs.vercel.app/?path=/story/components-menu--typeable-trigger from https://github.com/knocklabs/telegraph/pull/869. Although I not have the necessary context, looking at that Storybook link it looks like [Autocomplete](https://base-ui.com/react/components/autocomplete) might be what you're looking for.

image

kylemcd:
@aarongarciah 

yeah here's some context, but feel free to close this if there isn't utility in it for anyone else. 

we recently completed a radix to base ui migration for our component library [telegraph](https://github.com/knocklabs/telegraph). within telegraph we have the [menu](https://github.com/knocklabs/telegraph/tree/main/packages/menu) component which under the hood used radix's internal menu API so that we could get the keyboard / focus management a11y for free. within the knock dashboard, we have this super complex variable selector component that utilized the radix menu to get some of the keyboard management for free (see screenshot). this component does a lot of heavy lifting across our dashboard and refactoring that carries a ton of risk we're not wanting to take on right now. (eventually though you're right, we should migrate to the base ui autocomplete component)

Screenshot 2026-07-10 at 07-59-55


the base-ui menu fit the bill perfectly besides [not being able to manage the initial focus](https://github.com/mui/base-ui/blob/0cfa86d10814bdd0df2ed89d804868b7ce0371cf/packages/react/src/menu/popup/MenuPopup.tsx#L135). in the popover and dialog components, there's the `initialFocus` prop available which we're able to use to create a shim to manage auto focus in the same way that radix did, see this bit of code: https://github.com/knocklabs/telegraph/blob/13c3aa24c301d33a00701376ed54eca70205da0d/packages/popover/src/Popover/Popover.tsx#L232

because the base ui menu component doesn't allow for setting this prop in the same way dialog and popover does, we have to write this gross code in order to manage the focus in a way that doesn't break the variable selector code for example, see code: https://github.com/knocklabs/telegraph/pull/869/changes#diff-68ffc570fefcc972b7f997962aae62f5d4fc8be21fa7522ebb47128900d48dd8

so the ask is just to have parity between these. (there's already access to the `finalFocus` prop on the menu) maybe there's a reason why we wouldn't want the `initiaFocus` prop available to be set or maybe was just a consistency miss? either way it would help us immensely to have a more predictable focus code path for the menu.

aarongarciah:
From what you describe, it looks like you're after Autocomplete (or Combobox, depending on the exact UX). The problem with the pattern you're using (Menu with typeable trigger) is that it breaks the WAI-ARIA pattern: the trigger is supposed to be a button, and the focus always moves focus to the first or last menu item (see https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/#keyboardinteraction).

Since the focus must always be moved to a menu item, there wasn't a use case for `initialFocus` AFAIK. There's been some discussions around an imperative API like `highlightItem('first' | 'last' | 'none')` (https://github.com/mui/base-ui/pull/4815) to have control when the menu is programmatically opened, which keeps being WAI-ARIA compliant, but that's not related to your pattern.

In short, it's unlikely we merge this since there's no use case for `initialFocus`.

aarongarciah:
Thanks for the PR @kylemcd, but we won't go ahead with it as we don't think there's a good use case for `initialFocus` on `Menu.Popup`.

kylemcd:
Thanks for the guidance here. We've now moved this typeable-trigger use case to a Combobox/Autocomplete-based implementation in Telegraph and deprecated the Menu-based recipe, so this can stay closed. I'll remove the fork branch as well.

PR #5198, opened 2026-07-10
[combobox] Remove leaky value registry

atomiks:
## Changes

- Removed the retained all-values registry used when `items` is omitted.
- Closed-state selected-index synchronization uses the `items` prop when provided, otherwise the live registry of mounted item values; mounted items re-assert their index while closed so `keepMounted`/force-mounted lists stay in sync.


flaviendelangle (approved):
Restored to the original branch with a clean tree. Here's the review.

# PR review

Nothing merge-blocking. The core change is sound: `selectedIndex` is now resolved against the same registry (`valuesRef`) that typeahead and list navigation index into, which is exactly the misalignment the removed `allValuesRef` created, and the unbounded `push`-without-cleanup registry is gone. I ran the full Combobox suite in jsdom on the PR branch (710 passed) and chased the one removed-behavior candidate I could name — the `open` dependency dropped from the item effect, which used to re-assert `selectedIndex` on every open — but I could not construct a case where that safety net actually mattered; `master` behaves identically on the path I probed. The findings below are a missing test for the one genuinely new piece of logic and a note on effect shape.

**Verification limit:** the Chromium project could not run locally (Playwright's `chrome-headless-shell` isn't installed), so the reworked `pointerdown`/`mouseup` gesture path was only exercised in jsdom.

## Bugs (0)

No findings.

## Tests (2)

### 1. 🟡 The new `isPrimary` guard is untested

**Location:** `packages/react/src/combobox/item/ComboboxItem.tsx:161`

```tsx
onPointerDownCapture(event) {
  // The compat `mouseup` only fires for the primary pointer, so a non-primary
  // touch must not overwrite the shared ref — a mismatch would make the primary
  // pointer's release read as a drag-select and commit a second time after `click`.
  if (event.isPrimary) {
    store.state.pointerDownItemRef.current = event.currentTarget;
  }
  event.preventDefault();
}
```

Moving from a per-item boolean (`didPointerDownRef`) to a single shared element ref on the store is the one place in this PR that introduces genuinely new logic rather than relocating existing logic, and the `isPrimary` guard is the only thing standing between it and a double commit. `isPrimary` appears nowhere else in `packages/react/src`, so there is no precedent test to lean on and no coverage of the case the comment describes.

**Failure scenario:** a second (non-primary) touch lands on item B while the primary touch is held on item A. Without the guard the shared ref becomes B, so the primary pointer's `mouseup` on A sees `pointerDownItemRef.current !== event.currentTarget`, treats the release as a drag-select, and commits A — then the subsequent `click` on A commits it a second time. Nothing in the suite would catch a regression that drops or inverts this guard.

**Fix:** add a Chromium-only test that dispatches `pointerdown` on item A with `isPrimary: true`, a second `pointerdown` on item B with `isPrimary: false`, then `mouseup` + `click` on A, and asserts `onValueChange` fired once.

### 2. 🟡 `keepMounted` closed lists now register `selectedIndex`, with no test pinning it

**Location:** `packages/react/src/combobox/item/ComboboxItem.tsx:117`

Dropping the `if (!open) return;` guard changes observable state on a path none of the new tests assert. With `<Combobox.Portal keepMounted>` and no `items` prop, `selectedIndex` on `master` is `null` while the popup has never been opened (the item effect bailed out, and the closed sync read the still-empty `allValuesRef` because `FloatingPortal` mounts its children in a second commit). On this branch the item effect runs on that second commit and `selectedIndex` is correctly the selected item's index.

**Failure scenario:** this is an improvement, but it's unpinned — the `it.each([false, true])` typeahead test in `ComboboxTrigger.test.tsx` passes on `master` for both `keepMounted` values, so it does not cover it. Of the seven new tests, only `cycles from the selected item after reordering items while closed` fails on `master`; the five new `ComboboxRoot.test.tsx` tests all pass there, so they guard against regression rather than pin the fix.

**Fix:** assert `selectedIndex` (via the existing `SelectedIndexProbe`) is the selected item's index on initial render with a `keepMounted` portal and no `items` prop, before any open.

## Simplifications (1)

### 1. ℹ️ The pointer-gesture reset is coupled to the selection-sync effect's dependency list

**Location:** `packages/react/src/combobox/root/AriaCombobox.tsx:817`

```tsx
// State-driven (not tied to the internal event path) so controlled closes
// also clear a pointerdown that never received a matching item mouseup.
pointerDownItemRef.current = null;
```

Clearing the ref is correct, but the only thing it cares about is `open`, while the effect it lives in re-runs on `selectedValue`, `flatItems`, `multiple`, `isItemEqualToValue`, and `setIndices`. Those extra runs are harmless no-ops today, but the coupling means a future change to the selection-sync dependencies silently changes when the pointer state is reset.

**Failure scenario:** the concrete cost is readability — a reader of `syncClosedState` has to work out that a pointer-gesture reset is riding along in a selection-index effect.

**Fix:** a dedicated one-liner states the intent and cannot drift:

```tsx
useIsoLayoutEffect(() => {
  if (!open) {
    pointerDownItemRef.current = null;
  }
}, [open]);
```

## Docs (0)

No findings.

## Verdict

**Approve after nits** — no correctness problems found, but the new `isPrimary` guard and the new closed-state `selectedIndex` registration both ship without a test.

---

flaviendelangle (approved):

PR #5203, opened 2026-07-10
[utils] Correctly infer value type in `useControlled`

### LINKED ISSUE #5202: Incorrect types in `useControlled` util
# Bug report

## Current behavior

Note how I haven't defined a default value for any of the props below:

```tsx
function Popup({ open: openProp, defaultOpen, onOpenChange }) {
  const [open, setOpen] = useControlled({
    controlled: openProp,
    default: defaultOpen,
    // ...
  })
}
```

When we have an example like this where the types are a union of `undefined`, the above example excludes `undefined` from the types so `open` is `boolean` when it should be `undefined`.

Image

## Expected behavior

I expected the type system to correctly reflect that `open` could be `undefined` here.

## Reproducible example

https://stackblitz.com/edit/vitejs-vite-drsu38kk?file=src%2FApp.tsx

## Base UI version

Utils version 0.3.1

N/A

## Which OS are you using?

N/A

N/A

jjenzz:
Fixes #5202

## Summary

- Fixed `useControlled` type inference so returned values preserve `undefined` when both `controlled` and `default` can be undefined.
- Added a type regression test for optional vs. defined defaults.
- Made a few existing component fallback defaults explicit where their state contracts are non-optional.

## Verification

- `pnpm typescript`
- `pnpm prettier`
- Focused jsdom tests for `useControlled`, Accordion, Collapsible, ToggleGroup, and Toggle
- Targeted ESLint on changed TypeScript files


packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -219,7 +219,7 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
 
   const [inputValue, setInputValueUnwrapped] = useControlled({
     controlled: inputValueProp,
-    default: initialDefaultInputValue,
+    default: initialDefaultInputValue ?? '',
```
jjenzz: there is code on line 235 that does `String(inputValue)`, which would have been `"undefined"`

romgrk (approved):

packages/react/src/accordion/root/AccordionRoot.tsx
```diff
@@ -66,7 +66,7 @@ export const AccordionRoot = React.forwardRef(function AccordionRoot<Value = any
 
   const [value, setValue] = useControlled({
     controlled: valueProp,
-    default: defaultValue,
```
flaviendelangle, line 69: Right now it creates a new array on every re-render.
We can use `EMPTY_ARRAY` from `@base-ui/utils` to avoid that 👍

flaviendelangle (commented):
# PR review

The type fix itself is sound — I verified with `tsc` that an inferred call with optional `controlled`/`default` now yields `boolean | undefined`. Nothing here is merge-blocking. The main gap is that the new fallback overload silently reintroduces the exact bug from #5202 whenever the first overload doesn't match, which is the case at four call sites the PR didn't touch (they pass an explicit type argument). The rest are cleanups: the `AccordionRoot` memo and the `Toggle` `defaultPressed` ternary are now dead weight, and one comment is stale.

## Bugs (1)

### 1. 🟡 The fallback overload silently restores the old (wrong) type

**Location:** `packages/utils/src/useControlled.ts:26`

```ts
export function useControlled<T = unknown>(
  props: Omit<UseControlledProps<T>, 'default'> & { default: T },
): [T, React.Dispatch<React.SetStateAction<T>>];
export function useControlled<T = unknown>(
  props: UseControlledProps<T>,   // default: T | undefined  ->  returns [T, ...]
): [T, React.Dispatch<React.SetStateAction<T>>];
```

Overload resolution falls through to the second signature — which is the pre-PR signature, undefined-swallowing and all — whenever the first one fails to match. That happens in two realistic cases: when the caller passes an explicit type argument (inference is skipped, so `default: T | undefined` no longer satisfies `default: T`), and when `controlled` is wider than `default`. Both fail silently, with no error at the call site to hint that the stricter overload was skipped.

Four existing call sites hit the first case and were not updated, so they still have exactly the bug the PR fixes:

- `packages/react/src/otp-field/root/OTPFieldRoot.tsx:95` — `useControlled<string>` with `default: defaultValue` (`string | undefined`)
- `packages/react/src/number-field/root/NumberFieldRoot.tsx:111` — `useControlled<number | null>` with `default: defaultValue` (`number | undefined`)
- `packages/react/src/checkbox-group/CheckboxGroup.tsx:67` — `useControlled<string[]>`
- `packages/react/src/combobox/root/AriaCombobox.tsx:187` — `useControlled<any>`

I confirmed the resolution behavior against the repo's TypeScript with a standalone reproduction of the two overloads: `useControlled<string>({ controlled: strProp, default: defaultStr, name: 'OTPField' })` type-checks and returns `[string, …]`, and `expectType<string, typeof b>` passes — the `undefined` is dropped.

**Failure scenario:** Render an uncontrolled `<OTPField.Root />` with no `defaultValue`. On the first render `valueUnwrapped` is `undefined` at runtime, but TypeScript reports it as `string`. It doesn't crash today only because `normalizeOTPValue` happens to accept `string | null | undefined` (`OTPFieldRoot.tsx:136`); `NumberFieldRoot` is likewise saved by a defensive `valueUnwrapped ?? null`. So the type is still lying at exactly the sites where the fix was supposed to land, and a maintainer reading the merged PR would reasonably believe #5202 is fully closed.

**Fix:** Give those four sites a defined default the same way the other five got one — `default: defaultValue ?? ''` (OTPField), `?? null` (NumberField), `?? EMPTY_ARRAY` (CheckboxGroup) — or drop the explicit type argument so inference applies. Keeping the second overload is fine as a fallback for the wider-`controlled` case, but it shouldn't be the thing that quietly rescues call sites that ought to be fixed.

## Tests (1)

### 1. 🟡 The type test doesn't pin the explicit-type-argument case

**Location:** `packages/utils/src/useControlled.spec.ts:5`

The spec covers the two inferred cases (optional default, defined default) but nothing exercises `useControlled<T>(...)` with an explicit type argument, which is precisely where the new overload does not apply. Nothing in the suite would catch it if someone reintroduced the pattern, and the spec currently reads as if the fix is universal.

**Failure scenario:** A contributor adds `useControlled<string>({ controlled, default: maybeUndefined, name: 'X' })` in a new component. It compiles, `value` is typed `string`, it is `undefined` on first render, and no test or type test objects.

**Fix:** Add a third case asserting the current (or, once fixed, the intended) behavior with an explicit type argument, so the boundary of the fix is documented and locked down.

## Simplifications (2)

### 1. 🟡 The `AccordionRoot` `defaultValue` memo is now dead weight

**Location:** `packages/react/src/accordion/root/AccordionRoot.tsx:55`

```tsx
// memoized to allow omitting both defaultValue and value
// which would otherwise trigger a warning in useControlled
const defaultValue = React.useMemo(() => {
  if (valueProp === undefined) {
    return defaultValueProp ?? [];
  }
  return undefined;
}, [valueProp, defaultValueProp]);

const [value, setValue] = useControlled({
  controlled: valueProp,
  default: defaultValue ?? [],
  ...
```

With `?? []` applied at the call site, the memo's only remaining job — returning `undefined` in the controlled case — is immediately undone. `default: defaultValueProp ?? EMPTY_ARRAY` is equivalent for both the controlled and uncontrolled paths, and it's the shape `ToggleGroup` already uses (which this PR just aligned it toward). The comment above the memo is also now inaccurate: the memo no longer determines what `useControlled` receives.

**Failure scenario:** In controlled mode `defaultValue` is `undefined`, so `?? []` allocates a fresh array on every render; that array is the `[defaultProp]` dependency of the dev-mode default-change effect in `useControlled`, so the effect re-fires on every render in development (its body short-circuits on `isControlled`, so the cost is small — the real cost is a memo and a stale comment that a reader has to reason through).

**Fix:** Delete the memo and its comment, import `EMPTY_ARRAY` from `@base-ui/utils/empty`, and pass `default: defaultValueProp ?? EMPTY_ARRAY`.

### 2. ℹ️ `Toggle`'s `defaultPressed` ternary is now a no-op

**Location:** `packages/react/src/toggle/Toggle.tsx:49`

```tsx
const defaultPressed = groupContext ? undefined : defaultPressedProp;
...
default: defaultPressed ?? false,
```

`defaultPressedProp` already defaults to `false` (line 31), so after `?? false` both branches of the ternary produce `false` for the group case and `defaultPressedProp` otherwise — and inside a group `controlled` is always a boolean, so `default` is ignored entirely. The intermediate variable no longer expresses anything.

**Failure scenario:** No runtime cost; a reader has to trace three lines to conclude the branch is meaningless.

**Fix:** Drop line 49 and pass `default: defaultPressedProp` directly.

## Docs (1)

### 1. ℹ️ The two overloads have no explanation

**Location:** `packages/utils/src/useControlled.ts:26`

The pair of overloads encodes a non-obvious contract: the first is the strict one that preserves `undefined`, the second is a permissive fallback for calls where the first can't match. Without a comment, the next reader is likely to assume the second is redundant and delete it (or, more likely, not realize their explicit-type-argument call is quietly getting it).

**Failure scenario:** A future contributor removes the second overload as "dead", breaking every call site whose `controlled` type is wider than its `default` (e.g. `AriaCombobox`'s `inputValue`, where `controlled` is `string | undefined` and `default` is `string | number | readonly string[]`).

**Fix:** Add a short comment above the overloads naming what each one is for and when the fallback kicks in.

## Verdict

**Approve after nits** — no correctness regressions in the shipped behavior (all five component changes are runtime no-ops), but the fix leaves four explicit-generic call sites still typed incorrectly, and that gap is worth closing before this lands as "fixes #5202".

---

packages/utils/src/empty.ts
```diff
@@ -1,5 +1,5 @@
 export function NOOP() {}
 
-export const EMPTY_ARRAY: readonly any[] = Object.freeze([]);
```
jjenzz, line 3: @flaviendelangle i noticed that wherever we used this we were casting to compensate for the fact that freezing makes this `readonly`. using `never[]` will make TS flag the array if someone tries to mutate it, without forcing everywhere we use it to cast. can you think of any issue with this?

packages/react/src/combobox/root/AriaCombobox.tsx
```diff
@@ -203,21 +205,19 @@ export function AriaCombobox<Value = any, Mode extends SelectionMode = 'none'>(
 
   // If neither inputValue nor defaultInputValue are provided, derive it from the
   // selected value for single mode so the input reflects the selection on mount.
-  const initialDefaultInputValue = useRefWithInit<React.ComponentProps<'input'>['defaultValue']>(
-    () => {
-      if (hasInputValue) {
```
jjenzz, line 214: this was typing the value as potentially `undefined` when it's not. the ref returns a string when undefined.

packages/utils/src/useControlled.ts
```diff
@@ -69,13 +77,16 @@ export function useControlled<T = unknown>({
     }, [defaultProp]);
   }
 
-  const setValueIfUncontrolled = React.useCallback((newValue: React.SetStateAction<T>) => {
-    if (!isControlled) {
-      setValue(newValue as T);
-    }
-  }, []);
+  const setValueIfUncontrolled = React.useCallback(
+    (newValue: React.SetStateAction<T | undefined>) => {
```
jjenzz, line 83: the `T | undefined` here is just to satisfy the overloads, but will be `SetStateAction<T>` when a default is provided.

packages/utils/src/empty.ts
flaviendelangle: The main difference I see is that I'm now able to do this:

```ts
const EMPTY_ARRAY = Object.freeze([]) as never[]
const a: string[] = EMPTY_ARRAY
a.push('test')
```

jjenzz: that's the issue my change is attempting to solve, because previously we were doing this at call-sites:

```tsx
const EMPTY_ARRAY = Object.freeze([]);
// call-sites were casting it
const a = EMPTY_ARRAY as string[];
// types don't complain but runtime would explode here
a.push('test');
```

types that don't match the runtime is obvs less than ideal. instead now, i've made the empty array match runtime expectations:

```tsx
export const EMPTY_ARRAY: never[] = [];
```

so this will be a [TS error](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwgOwM7wKIFkAKAVATQH0BBAJTJIIC45lgA3YKAbQF04BeOdgbgCgkaeAEMucbPmLlKBASIB0YAK6oAFgAoA5DGDotASiA):

```tsx
const a = EMPTY_ARRAY;
a.push('test')
```

but [this won't](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwgOwM7wKIFkAKAVATQH0BBAJTJIIC45lgA3YKAbQF04BeOdgbgCgkaeAEMucbPmLlKBOCNRx0UAJbIA5n34iAdGACuqABYAKAOQxg6MwEogA), which is fine because casting and runtime match:

```tsx
const a = EMPTY_ARRAY as string[];
a.push('test')
```

flaviendelangle: I guess it's fine then :ok_hand:

jjenzz:
@claude review

PR #5206, opened 2026-07-10
[docs] Fix demo outline not being fully masked

aarongarciah:
One pixel of the outline wasn't being masked. A refactor was required to fix it.

|before|after|
|---|---|
|Screenshot 2026-07-10 at 20 33 48|Screenshot 2026-07-10 at 20 35 36|


flaviendelangle (approved):
LGTM

PR #5207, opened 2026-07-11
[composite] Keep item refs populated across StrictMode remounts

### LINKED ISSUE #4698: [select] Mouse selection is a no-op in dev StrictMode

Arman-Luthra:
Fixes #4698

## Problem

When a `CompositeList` subtree mounts in StrictMode dev, the simulated unmount runs the cleanup that empties `elementsRef`/`labelsRef`, and the simulated remount re-registers the items — but items whose index didn't change bail out of `setIndex`, so their ref callbacks never re-run and the arrays stay empty. `useListNavigation` then can't resolve hovered items (`listRef.current.indexOf(event.currentTarget) === -1`), so `activeIndex`/`data-highlighted` never update and mouse interaction breaks.

Whether this is user-visible depends on the exact React build, which is why it has been hard to reproduce (#4897):

- `react@19` stable: StrictMode reattaches refs during the simulated remount, refilling the arrays.
- `react@18.3.1` stable: the StrictMode double render double-invokes the `useState` initializer, which bumps the shared `nextIndexRef` twice per item, so `IndexGuessBehavior.GuessFromOrder` guesses are always off (`1, 3, 5, …`). Every index then shifts after registration, producing new ref callbacks that refill the arrays as a side effect.
- `react@18.3.0-canary-178c267a4e-20241218` (the build Next 14.2 vendors for the App Router, which the issue's repro runs on): guesses are already correct and refs are not reattached, so nothing repopulates the arrays and the select becomes unusable with the mouse, exactly as reported.

## Changes

- `useCompositeListItem` writes the element/label back into the ref arrays from the map-change subscription it already registers, whenever the stored element is stale. Compared to #4897 this adds no extra pass over the list: the per-item listener already runs for `setIndex`, and the `elementsRef.current[i] !== node` guard keeps the steady-state cost at one comparison per item, with no label/`textContent` reads unless the slot is actually stale.
- Test: `CompositeList` keeps the ref arrays populated for items whose guessed index is already correct.
- Test: hovering a `Select.Item` with the mouse sets `data-highlighted` and a subsequent click commits the value.

## Verification

Running the suite against the Next-vendored canary above, the two new tests plus five existing `SelectItem` tests (keyboard focus, Enter select, click select, `onClick` once, `data-selected`) fail on unmodified `master` in 3/3 runs and pass with this fix in 3/3 runs. The select, combobox, menu, and internals suites are green with the fix under `react@19.2.5`, `react@18.3.1`, and that canary.

Honest limitation: on the React versions CI runs (`19.2.5` and `^18` = `18.3.1`), the new tests also pass on `master` because of the accidental repopulation described above, so CI alone cannot demonstrate the regression — the canary runs are the fail-before-fix evidence.

### DIFF
--- a/packages/react/src/internals/composite/list/CompositeList.test.tsx
+++ b/packages/react/src/internals/composite/list/CompositeList.test.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { screen, waitFor } from '@mui/internal-test-utils';
 import { createRenderer } from '#test-utils';
 import { CompositeList } from './CompositeList';
-import { useCompositeListItem } from './useCompositeListItem';
+import { IndexGuessBehavior, useCompositeListItem } from './useCompositeListItem';
 
 describe('<CompositeList />', () => {
   const { render } = createRenderer();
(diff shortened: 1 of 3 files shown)


atomiks:
Thanks for the PR @Arman-Luthra, it seems https://github.com/mui/base-ui/pull/5148 will already fix this structurally. The tests from here can be added onto that PR however.

atomiks:
Replaced with https://github.com/mui/base-ui/pull/5256

PR #5210, opened 2026-07-13
[toast] Render content passed through the render prop in Title, Description, and Action

### LINKED ISSUE #2809: [toast] Toast.Description does not render when a custom component is passed to render

m2na7:

### Description

`Toast.Title`, `Toast.Description`, and `Toast.Action` gate rendering on the presence of resolved children. When the part is provided through the `render` prop with content but no `children` — e.g. `<Toast.Description render={<div>…</div>} />`, the form in the #2809 repro — that content was dropped: `shouldRender` was `false` and the element (along with its `aria-*` wiring) never mounted.

The fix gates on the **evaluated element's resolved children**, which are the source of truth for content — a `render` element's / render function's own children win over the toast-supplied content in the merge. Whether resolved children count as content is decided by a shared `isRenderableNode` predicate matching React renderability: numeric/bigint children (including `0`, `0n`, and `NaN`) are content; nullish values, booleans, empty strings, and arrays containing only those are not. A `React.isValidElement` check guards render functions that return null, so the children lookup can't crash. So:

- `render={<div>…</div>}` (carries content) mounts;
- a childless, styling-only `render={<MyHeading />}` — where content comes from `toast.title` / `add({ title })`, or is intentionally empty — stays conditional and does **not** force-mount an empty element;
- `<Toast.Title render={<MyHeading />}>…</Toast.Title>` (explicit children) mounts;
- a plain `{0}` child now mounts, and an empty-array child no longer mounts an empty element.

On top of #5219's structure: `useToastLabelPart` keeps resolving the content, and a new `useToastLabelElement` gates mounting on the evaluated element and registers the part's id with the root only while it actually renders, keeping `aria-labelledby` / `aria-describedby` pointing at real nodes. `Toast.Action` drops the `enabled` option because render-prop content only exists once the element is evaluated.

### Linked Issues

Closes #2809

### Additional context

- New shared util `packages/react/src/toast/utils/isRenderableNode.ts`, with the value-kind edge cases (`0`, empty/nested arrays, null-returning render functions) pinned in its unit tests; each part's suite covers its own call-site paths, including the childless styling-only case and the `aria-*` wiring/cleanup for a render-only part.
- Supersedes #5211 — with the gate shared across the three parts, the `Toast.Action` fix lives here.


atomiks:
# PR review

No merge-blocking `.props`, Suspense, RSC, or React-version issue. Both current heads—#5210 `a9e92cf` and #5211 `c1edef0`—now short-circuit with `React.isValidElement(element)`.

Genuine React elements always contain a props object in [React 18](https://github.com/facebook/react/blob/v18.3.1/packages/react/src/ReactElement.js) and [current React](https://github.com/facebook/react/blob/main/packages/react/src/jsx/ReactJSXElement.js). A callback can return `null` at runtime, but the guard handles it. Only a manually forged internal `$$typeof` object could pass `isValidElement` without props; that is outside React’s supported contract.

Suspension occurs before the property read, and React retries suspended trees from scratch, so there is no partially initialized element. Client/Server Component JSX can cross a client boundary, while ordinary render callbacks cannot; they must originate in client code.

Local verification passed 44 Title/Description tests, TypeScript, SSR, and lazy/Suspense smoke testing. The React 18 CI job currently stops during dependency installation before running tests.

## Bugs (1)

### 1. 🟡 Treat renderable children separately from truthiness

**Location:** `packages/react/src/toast/title/ToastTitle.tsx:48`, mirrored in `ToastDescription.tsx:49` and PR #5211’s `ToastAction.tsx:53`

```tsx
const shouldRender =
  React.isValidElement(element) &&
  Boolean((element.props as { children?: React.ReactNode }).children);
```

ReactNode renderability does not match JavaScript truthiness. Numeric `0`, `0n`, and `NaN` render visible text but are rejected here. Conversely, `[]` and `[null]` are truthy and mount an empty element.

**Failure scenario:** `<Toast.Title render={<div>{count}</div>} />` disappears when `count` is `0`, leaving the toast without its intended visible and accessible title.

**Fix:** Use a shared renderable-children predicate across Title, Description, and Action that preserves numeric/bigint children while filtering empty arrays, nullish values, booleans, and empty strings. Add `0` and empty-array coverage. This behavior predates these PRs for ordinary children, but the rewritten guard is the right place to correct it.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve after nits** - the only remaining issue is the narrow falsy-numeric children case.

---

🤖 Review generated with Codex

m2na7:
Addressed with a shared `isRenderableNode` predicate (also used by Action in #5211): numeric/bigint children like `0` count as content; nullish values, booleans, empty strings, and empty arrays don't. Added `0` and empty-array coverage (render prop + plain children) and a regression test for a render function returning null.

m2na7:
@atomiks Updated the branch on top of #5219. The gate now lives in the shared layer (`useToastLabelPart` + a new `useToastLabelElement` that mounts on the evaluated element), and `Toast.Action` is folded in here via the same check, supersedes #5211. Also pinned the effect's cleanup (removing the title content clears `aria-labelledby`) and moved the value-kind edge cases into the `isRenderableNode` unit tests.

atomiks (approved):
Gating each part on the evaluated element's resolved children is the right fix for #2809, and matches React renderability semantics. Added two tests pinning the `Toast.Description` render-prop paths (aria-describedby wiring and the childless styling-only fallback to `toast.description`). CI green.

PR #5211, opened 2026-07-13
[toast] Render content passed via the render prop in Action

m2na7:

### Description

`Toast.Action` gates rendering on the presence of resolved children (`toast.actionProps?.children ?? children`). When the action is provided through the `render` prop with content but no children — `<Toast.Action render={<button>Undo</button>} />` — that content was dropped: `shouldRender` was `false` and the element never mounted.

The fix gates on the **evaluated element's resolved children**, so a `render` that carries content mounts, while a childless, styling-only `render={<MyButton />}` — used purely to customize the element in a shared template, with the action content coming from `toast.actionProps` — stays conditional and does not force-mount an empty button for every toast. Whether resolved children count as content is decided by a shared `isRenderableNode` predicate matching React renderability (numeric/bigint children including `0` are content; nullish values, booleans, empty strings, and arrays containing only those are not), behind a `React.isValidElement` check so a render function returning null can't crash the children lookup.

### Linked Issues

Related to #2809 (same root cause, for Title and Description).

### Additional context

Follow-up to #5210, kept separate to scope that PR to the reported issue. The shared util `packages/react/src/toast/utils/isRenderableNode.ts` (+ tests) is included byte-identically on both PRs, so whichever merges second resolves it cleanly and its squash then effectively touches only the `Toast.Action` files. Tests cover the `render` prop as an element and as a function, the childless-render case, a render function returning null, numeric `0` as render-prop and plain child, and empty-array children.

### DIFF
--- a/packages/react/src/toast/action/ToastAction.test.tsx
+++ b/packages/react/src/toast/action/ToastAction.test.tsx
@@ -140,4 +140,47 @@ describe('<Toast.Action />', () => {
 
     expect(screen.getByText('0')).not.toBe(null);
   });
+
+  it('renders a numeric zero child', async () => {
+    await render(
+      <Toast.Provider>
+        <Toast.Viewport>
+          <Toast.Root toast={toast}>
+            <Toast.Action>{0}</Toast.Action>
+          </Toast.Root>
+        </Toast.Viewport>
+      </Toast.Provider>,
+    );
+
+    expect(screen.getByText('0')).not.toBe(null);
+  });
+
+  it('does not render an empty-array child', async () => {
+    await render(
+      <Toast.Provider>
+        <Toast.Viewport>
+          <Toast.Root toast={toast}>
+            <Toast.Action data-testid="action-render">{[]}</Toast.Action>
+          </Toast.Root>
+        </Toast.Viewport>
+      </Toast.Provider>,
+    );
+
+    expect(screen.queryByTestId('action-render')).toBe(null);
+  });
+
+  it('does not render when a render function returns no element', async () => {
+    await render(
+      <Toast.Provider>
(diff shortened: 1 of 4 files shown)


atomiks:
`render !== undefined` isn't quite the right check, as `render` describes how to render, not whether. It's common to pass `render` purely to customize the element in a shared template:

```tsx
<Toast.Root toast={toast}>
  <Toast.Action render={<MyButton />} /> {/* content comes from toast.actionProps */}
</Toast.Root>
```

With this change, an empty action button mounts for every toast, even those without an action.

A better check is to also treat the render element's own children as content, so `render={<button>Undo</button>}` mounts while childless `render={<MyButton />}` stays conditional. Components that encapsulate their own content still must specify children externally (`<Toast.Action render={<MyButton />}>Undo</Toast.Action>`)

```diff
-  const shouldRender = Boolean(computedChildren);
+  // The render prop's evaluated element is the source of truth for content:
+  // `render` element children override `computedChildren` in the merge, and a
+  // render function's returned element carries its own children.
+  const shouldRender = Boolean(
+    (element.props as { children?: React.ReactNode }).children,
+  );
+
```

m2na7:
Thanks, you're right
`render !== undefined` over-renders, mounting an empty element for every toast without an action. Adopted your evaluated-element check (`Boolean(element.props.children)`) so `render={<button>Undo</button>}` mounts while a childless `render={<MyButton />}` stays conditional, and added a regression test for the childless case. Applying the same fix to #5210 for Title/Description.

m2na7:
Consolidated into #5210 — after #5219 deduplicated the Title/Description logic into a shared helper, the same gate now covers all three parts, so the Action fix lives there.

PR #5229, opened 2026-07-13
[utils] Add idle callback utility

bernardobelchior:
Adds `useIdleCallback`, a cancellable idle-time scheduler with a Safari-compatible `Timeout` fallback, plus coverage for scheduling and lifecycle cleanup.

### DIFF
--- a/packages/utils/src/useIdleCallback.ts
+++ b/packages/utils/src/useIdleCallback.ts
@@ -4,6 +4,13 @@ import { useOnMount } from './useOnMount';
 
 const supportsIdleCallback = typeof requestIdleCallback === 'function';
 
+// Macrotask fallback for environments without `requestIdleCallback` (e.g. Safari),
+// which still runs after the current commit and paint.
+const requestCallback = supportsIdleCallback
+  ? requestIdleCallback
+  : (fn: () => void) => setTimeout(fn, 0);
+const cancelCallback = supportsIdleCallback ? cancelIdleCallback : clearTimeout;
+
 export class IdleCallback {
   static create() {
     return new IdleCallback();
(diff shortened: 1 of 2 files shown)


packages/utils/src/useIdleCallback.ts
```diff
@@ -0,0 +1,61 @@
...
+export class IdleCallback {
+  static create() {
+    return new IdleCallback();
+  }
+
+  currentId: number | null = null;
+
+  // Macrotask fallback for environments without `requestIdleCallback` (e.g. Safari),
+  // which still runs after the current commit and paint.
+  timeout = new Timeout();
```
romgrk: We could skip instanciating it unless we need it, keeps memory usage lower.

romgrk: We could also skip the `Timeout` import and use `setTimeout` and `clearTimeout` directly.

packages/utils/src/useIdleCallback.ts
```diff
@@ -0,0 +1,55 @@
...
+import { useOnMount } from './useOnMount';
+
+const supportsIdleCallback = typeof requestIdleCallback === 'function';
+
+// Macrotask fallback for environments without `requestIdleCallback` (e.g. Safari),
+// which still runs after the current commit and paint.
+const requestCallback = supportsIdleCallback
+  ? requestIdleCallback
+  : (fn: () => void) => setTimeout(fn, 0);
+const cancelCallback = supportsIdleCallback ? cancelIdleCallback : clearTimeout;
```
bernardobelchior, line 13: I'm assuming that support for `requestIdleCallback` doesn't change in runtime.

I think it's a valid assumption IMO.

packages/utils/src/useIdleCallback.ts
bernardobelchior: Done 👍

michaldudak:
# PR review

No merge-blocking issues. Targeted jsdom and Chromium suites both pass 8/8; two non-blocking coverage/documentation gaps remain.

## Bugs (0)

No findings.

## Tests (1)

### 1. 🟠 The completed-callback state reset is not regression-tested

**Location:** `packages/utils/src/useIdleCallback.test.tsx:60`

```tsx
idleCallback.start(fn);
await waitFor(() => expect(fn).toHaveBeenCalledTimes(1));

idleCallback.start(fn);
await waitFor(() => expect(fn).toHaveBeenCalledTimes(2));
```

This test still passes if `this.currentId = null` is removed from the callback wrapper: the second `start()` merely cancels the stale handle before scheduling another callback. It therefore does not cover the correctness fix added in `11091b7`.

**Failure scenario:** A future refactor drops the reset while all tests remain green. If the browser reuses the completed handle, a later `clear()` can cancel another pending idle callback.

**Fix:** After the first callback runs, assert that `idleCallback.currentId` is `null`, covering both native and fallback environments.

## Simplifications (0)

No findings.

## Docs (1)

### 1. 🟠 The method documentation still promises after-paint execution for the fallback

**Location:** `packages/utils/src/useIdleCallback.ts:23`

```tsx
/**
 * Schedules `fn` to run during idle time, after the current commit and paint, clearing any
 * previously scheduled call.
 */
```

The latest comments correctly explain that the `setTimeout(0)` fallback is not guaranteed to run after the next paint, but this public method JSDoc still makes that guarantee. The test comment at `useIdleCallback.test.tsx:31` repeats the same stale claim.

**Failure scenario:** A consumer relies on `IdleCallback.start()` to defer work until after paint, but in Safari’s fallback path the callback may run before the next paint.

**Fix:** Describe the common guarantee as asynchronous execution after the current task, then qualify the native idle/after-paint behavior. Update the test comment similarly.

## Verdict

**Approve after nits** - the implementation is sound, but the state-reset regression needs a direct assertion and the remaining timing documentation should be corrected.

---

🤖 Review generated with Codex

romgrk (approved):

PR #5232, opened 2026-07-14
[combobox] Return highlight to selected item on query clear

### LINKED ISSUE #5227: [combobox] Return highlight to the selected item
# Feature request

I'm missing a consistent way to return and rehighlight the selected item when clearing the combobox input. This feels like it could be either a feature request to be able to compose the behavior yourself, or a bug – I am not sure the default behavior makes sense to me at all.

Using the base [Input inside popup](https://base-ui.com/react/components/combobox#input-inside-popup) example:

Initial open scrolls to and highlights the selected value, but clearing the input doesn't restore the initial state:

https://github.com/user-attachments/assets/a083aebb-a649-4f9a-8d5f-0be2d3996ed5

Using it with `autoHighlight` makes it worse, setting a strong emphasis on the first item which feels random for the user

https://github.com/user-attachments/assets/7850544c-f8fa-45b1-afed-bb295d4c4915

In Paper we had to resort to a bun patch to achieve the experience we need (note ours is also inline and virtualized):

https://github.com/user-attachments/assets/8d455809-c23f-45ab-b60f-c1acf5c60c00

As an example in other apps, Figma has a somewhat buggy implementation of the same (they re-highlight the item the first time you clear the input, but only scroll to it the second time):

https://github.com/user-attachments/assets/f6e7e89b-be52-4f2c-b3e3-cd803c31c684

atomiks:
Closes #5227

Opening the popup highlights the selected item and scrolls it into view, but clearing a typed query did not. The highlight stayed wherever filtering left it, and with `autoHighlight` it snapped to the first item, which reads as arbitrary. Clearing now returns to the same anchor opening uses, including when selecting an item auto-clears the query with the input inside the popup.

Typing deliberately drops `selectedIndex`, since a stale index would misdirect arrow-key navigation while filtered, so the anchor cannot simply be re-read on clear. It is recomputed from the selected value against the currently visible list. Items re-mounted by the clear publish their composite indices a commit later, so the restore is deferred past that cascade: reading the registry during the clear commit pairs the selected item with a stale index and emits a wrong `onItemHighlighted`, which virtualized lists act on to scroll.

With no selection the highlight is left alone, and a selection that is no longer in the list drops the highlight rather than leaving it on whichever item now occupies that index. A clear that a controlled `inputValue` rejects does not restore.


flaviendelangle (approved):
# PR review

This PR restores the combobox highlight to the selected item when a query is cleared, fixing #5227. The change is well-structured: it consolidates highlight emission into a single `emitHighlight` helper, replaces the `allValuesRef` accumulator with the live `valuesRef` registry, extracts `findSelectionIndex`, and adds a deferred (microtask) restore that correctly waits for re-mounted items to publish composite indices. I found **no merge-blocking issues**. Typecheck, jsdom tests (263 pass, including 47 new tests), eslint, prettier, and `docs:api` are all clean; Chromium tests couldn't run locally (browser binary not installed), so the `skipIf(isJSDOM)` browser cases were not executed here.

## Bugs (0)

No findings.

## Tests (0)

No findings. The diff is additive on tests (202 → 249 test names, none removed) and covers the new paths: restore-to-selection on clear, no-selection left alone, selection-no-longer-in-list dropping the highlight, controlled `inputValue` rejecting a clear, and virtualized lists. Browser-dependent cases are correctly gated with `it.skipIf(isJSDOM)`.

## Simplifications (0)

No findings. The refactors are behavior-preserving and verified: the `requestSubmit` internal `submitOnItemClick` guard was safely moved to its only caller (`ComboboxItem` line 141); `handleSelection`'s `activeIndex` fallback was dropped because keyboard Enter now routes through `listItem.click()` with an explicit `itemValue`; removed store fields (`allValuesRef`, `isGrouped`, `onItemHighlighted`) have no external readers; `createSingleSelectionCollatorFilter` remains used by `useFilter.ts`.

## Docs (0)

No findings. The removed `@default 'none'` JSDoc and the `selectionMode: Mode` required change are on the `@internal` `AriaCombobox`; both public wrappers (`ComboboxRoot`, `AutocompleteRoot`) always pass `selectionMode` and omit it from their public surface, and `pnpm docs:api` reports no changes.

## Verdict

**Approve** — no correctness, test, or doc regressions survived verification; the change is behavior-preserving where refactored and well-covered by new tests.

One heads-up (ℹ️, not a defect): typing now resets the nearest in-popup scroll container to `scrollTop = 0` on every keystroke (`AriaCombobox.tsx:571-590`). I confirmed this cannot escape to page scroll — `getOverflowAncestors` returns `window` (a non-`HTMLElement`, so the loop breaks) rather than `documentElement` unless the body itself is an explicit overflow container. The reset is intentional ("scrolling cannot escape the popup"), just worth being aware of for non-`autoHighlight` popups where a user scrolls then keeps typing.

---

PR #5235, opened 2026-07-14
[popover] Prevent initial position transition in React 17

### LINKED ISSUE #5187: [popover] Positioner slides from viewport top-left on first open in React 17
# Bug report

## Current behavior

In React 17, opening a `Popover` for the first time via mouse click causes the positioner to animate from the viewport top-left `(0, 0)` to its target position, instead of appearing in place next to the trigger. 100% reproducible on the first open after mount

React 19 is not affected. @base-ui/react 1.4.1 + React 17 is not affected

## Expected behavior

The positioner's initial position change (from `translate(0,0)` to `translate(x,y)`) should not trigger a CSS transition. The popover should appear directly beside the trigger.

## Reproducible example

https://github.com/user-attachments/assets/e1c75f64-de80-42bc-9943-4d141f8f4611

[CodeSandbox](https://stackblitz.com/edit/jpqdbqiv?file=src%2FApp.tsx)

## Base UI version

1.6.0

Chrome

## Which OS are you using?

Mac OS

N/A.

## Additional context

Base UI guards the positioner's initial-position transition with an inline `style: { transition: 'none' }` that only lives while `transitionStatus === 'starting'` — one animation frame. The position itself is computed by `@floating-ui/react-dom`'s `useFloating` as a microtask (`computePosition().then(...)` → `flushSync(setData)`). Per the HTML spec, microtasks run before `requestAnimationFrame`, so as long as `update()` is called inside a layout effect, the position settles before `transitionStatus` is cleared and no transition fires.

In React 17, `update()` is not called inside the first layout effect. The `floatingElement` has to cross from `PopoverStore` to `FloatingRootStore` and then reach `useFloating`. Base UI switches its store implementation by React version: React 17 goes through the `use-sync-external-store/shim` polyfill, which attaches its store subscription in `useEffect` (runs after paint) rather than `useLayoutEffect`. As a result, the `FloatingRootStore.floatingElement` update that happens during the first commit cycle is not propagated to `useFloating` in the same cycle — the subscription isn't installed yet. `useFloating` only finds out in `useEffect`, which runs after the rAF that clears `transitionStatus` has already fired. By the time `computePosition` resolves and the position changes from `translate(0,0)` to `translate(x,y)`, the inline `transition: none` has already been removed, so the CSS transition fires and the popover slides in from the top-left corner.

React 19 is unaffected because the native `useSyncExternalStore` has tearing protection — store changes during render are detected immediately, so `update()` is called inside the first layout effect, the microtask runs before rAF, and the position settles while `transition: none` is still applied.

wanxiankai:
## Summary

- keep the Popover positioner's CSS transitions disabled until Floating UI has produced its first real position
- flush the positioned layout before restoring the element's original `transition-property`
- add a Chromium regression test that verifies the first open does not run a transform transition and that transitions are restored afterward

## Root cause

In React 17, the `use-sync-external-store` shim subscribes after paint. On the first open, the floating element can therefore reach the positioning hook after the one-frame `starting` transition guard has already ended. When Floating UI later changes the transform from `(0, 0)` to the anchor position, consumer CSS transitions animate that initial positioning update.

The positioner now temporarily overrides only `transition-property`, preserving the consumer's shorthand and other transition settings. Once `isPositioned` becomes true, it forces the target layout to be calculated with transitions disabled and then restores the original property without an extra React render.

Closes #5187

## Test plan

- React 17: `pnpm test:chromium PopoverPositioner --no-watch -t "does not transition from the viewport origin on first open"`
- React 19.2.5: `pnpm test:chromium Popover --no-watch` (434 passed, 1 skipped)
- React 19.2.5: `pnpm test:jsdom Popover --no-watch` (335 passed, 100 skipped)
- `pnpm typescript`
- `pnpm eslint`
- `pnpm prettier`
- `pnpm stylelint`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -168,6 +168,21 @@ jobs:
           command: pnpm test:chromium
       - store_test_results:
           path: test-results
+  test_browser_react_17:
+    <<: *default-job
+    resource_class: large
+    executor:
+      name: code-infra/mui-node-browser
+      playwright-img-version: v1.59.1-noble
+    steps:
+      - checkout
+      - code-infra/install-deps:
+          package-overrides: react@17.0.2
+      - run:
+          name: Run anchored Positioner regressions with React 17
+          command: pnpm test:chromium PopoverPositioner TooltipPositioner --no-watch
+      - store_test_results:
+          path: test-results
   test_regressions:
     <<: *default-job
     resource_class: medium+
@@ -333,6 +348,13 @@ workflows:
       - test_browser_compiler:
           <<: *default-context
           name: 'Browser tests (React Compiler)'
+  react-17:
+    when:
+      equal: [pipeline, << pipeline.parameters.workflow >>]
+    jobs:
+      - test_browser_react_17:
+          <<: *default-context
+          name: 'Browser tests (React 17)'
   react-18:
     when:
       equal: [pipeline, << pipeline.parameters.workflow >>]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,7 +182,8 @@ jobs:
           name: Run anchored Positioner regressions with React 17
           command: |
             pnpm exec cross-env TZ=UTC VITEST_ENV=chromium vitest \
-              --project @base-ui/react PopoverPositioner TooltipPositioner --no-watch
+              --project @base-ui/react PopoverPositioner TooltipPositioner --no-watch \
+              -t 'does not transition from the viewport origin on first open|preserves the popup starting-style animation on every open'
           environment:
             REACT_17_TESTS: 'true'
       - store_test_results:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,10 +177,14 @@ jobs:
     steps:
       - checkout
       - code-infra/install-deps:
-          package-overrides: react@17.0.2
+          package-overrides: react@17.0.2 @testing-library/react@12.1.5
       - run:
           name: Run anchored Positioner regressions with React 17
-          command: pnpm test:chromium PopoverPositioner TooltipPositioner --no-watch
+          command: |
+            pnpm exec cross-env TZ=UTC VITEST_ENV=chromium vitest \
+              --project @base-ui/react PopoverPositioner TooltipPositioner --no-watch
+          environment:
+            REACT_17_TESTS: 'true'
       - store_test_results:
           path: test-results
   test_regressions:
--- a/packages/react/src/combobox/positioner/ComboboxPositioner.tsx
+++ b/packages/react/src/combobox/positioner/ComboboxPositioner.tsx
@@ -115,6 +115,7 @@ export const ComboboxPositioner = React.forwardRef(function ComboboxPositioner(
 
   const element = usePositioner(componentProps, state, {
     styles: positioning.positionerStyles,
+    isPositioned: positioning.isPositioned,
     transitionStatus,
     props: elementProps,
     refs: [forwardedRef, setPositionerElement],
--- a/packages/react/src/menu/positioner/MenuPositioner.tsx
+++ b/packages/react/src/menu/positioner/MenuPositioner.tsx
@@ -273,6 +273,7 @@ export const MenuPositioner = React.forwardRef(function MenuPositioner(
 
   const element = usePositioner(componentProps, state, {
     styles: positioner.positionerStyles,
+    isPositioned: positioner.isPositioned,
     transitionStatus,
     props: elementProps,
     refs: [forwardedRef, store.useStateSetter('positionerElement')],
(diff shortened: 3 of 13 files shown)


atomiks:
# PR review

The Popover workaround passes an explicit Chromium reproduction with React 17.0.2, while the proposed `setFloating` alternative does not. However, the implementation remains Popover-specific even though the same anchored-positioning sequence affects the other popup families, and it forces synchronous layout during first open. The exact v1.6.0 repro also exposes a separate first-open `.Popup` starting-style failure that this patch does not address; current master appears to avoid it only because of unrelated post-v1.6 popup-store changes. The incomplete component coverage is merge-blocking.

## Bugs (1)

### 1. 🔴 Apply the React 17 fix to every affected anchored positioner

**Location:** `packages/react/src/popover/positioner/PopoverPositioner.tsx:95`

```tsx
const initialPositioningCompleteRef = React.useRef(false);
const initialTransitionPropertyRef = React.useRef<string | null>(null);
```

The workaround is implemented only inside `PopoverPositioner`, but Menu, Select, Combobox, Tooltip, Preview Card, Navigation Menu, and Toast use the same anchored `usePositioner` flow and can encounter the same React 17 first-mount timing. Fixing only Popover leaves the reported behavior in the other component families.

**Failure scenario:** A React 17 consumer opens a Menu, Select, Combobox, Tooltip, Preview Card, Navigation Menu, or Toast for the first time and still sees its Positioner transition from the viewport origin.

**Fix:** Move the initial-position transition handling into the shared anchored-positioning/`usePositioner` layer so every affected Positioner receives the same behavior, rather than duplicating this callback and lifecycle state component by component.

## Tests (2)

### 1. 🟠 Exercise the regression in a React 17 test job

**Location:** `packages/react/src/popover/positioner/PopoverPositioner.test.tsx:40`

```tsx
it.skipIf(isJSDOM)('does not transition from the viewport origin on first open', async () => {
```

The test captures the right browser symptom, but the repository's default suite resolves React 19, where fresh base already avoids the React 17 timing failure. Without running this assertion in an actual React 17 job, normal CI does not prove the regression remains fixed. It also covers only Popover despite the shared component scope.

**Failure scenario:** The React 17 workaround regresses, or another affected Positioner remains unfixed, while the default React 19 Chromium test stays green.

**Fix:** Run the regression under a React 17 dependency matrix and add shared or representative coverage for the other affected anchored Positioners.

### 2. 🟠 Preserve the first-open `.Popup` starting-style animation

**Location:** `packages/react/src/popover/positioner/PopoverPositioner.test.tsx:40`

The exact linked v1.6.0 React 17 repro has two first-open symptoms. In addition to the Positioner slide, `.Popup` mounts on the first open without `data-starting-style`: sampled computed styles remain `opacity: 1` and `transform: none`. On the second open it mounts with the attribute and visibly transitions from `opacity: 0` / `scale(0.9)`. This patch changes only the Positioner's `transition-property`, and the new test asserts only that the Positioner's transform does not transition. Current master does animate `.Popup` on first open, apparently because of unrelated post-v1.6 popup-store changes, so the positioner-only test does not preserve the full behavior demonstrated by the release repro.

**Failure scenario:** A later store or mount-order change reintroduces the React 17 first-open `.Popup` failure while this regression test remains green because the Positioner no longer slides.

**Fix:** In the React 17 Chromium job, reproduce the linked Popup CSS and assert on the first open that `data-starting-style` is present at mount and that computed opacity/transform progress from the starting values after positioning. Repeat the assertion after reopening to distinguish the first-mount path.

## Simplifications (1)

### 1. 🟠 Avoid forcing synchronous layout during first open

**Location:** `packages/react/src/popover/positioner/PopoverPositioner.tsx:117`

```tsx
positionerElement.getBoundingClientRect();
```

This read follows positioning/style writes and deliberately forces the browser to synchronously calculate layout before restoring `transition-property`. It adds work directly to the first-open path of every Popover instance; extending this implementation separately to every affected component would duplicate both the code and the layout cost.

**Failure scenario:** A page mounting or opening many anchored popups pays an avoidable forced-layout cost for each new Positioner, increasing first-open latency and the chance of frame drops.

**Fix:** Centralize the guard in the shared positioning layer and keep transitions disabled through the first positioned frame without a synchronous geometry read, using the repository's `useAnimationFrame` utility. If a forced layout is unavoidable, keep it shared and add a focused measurement demonstrating the cost is acceptable.

## Docs (0)

No findings.

## Verdict

**Request changes** - the Popover workaround is effective for the Positioner in React 17, but the shared anchored-component regression remains unfixed and the exact release repro's first-open Popup behavior is not covered.

---

🤖 Review generated with Codex

wanxiankai:
Thanks for the detailed review — I addressed the findings in 85c92d4.

- Moved the initial-position transition guard into the shared `usePositioner` layer and wired it to all anchored Positioners (Popover, Menu, Select, Combobox, Tooltip, Preview Card, Navigation Menu, and Toast).
- Removed the synchronous `getBoundingClientRect()` flush. The shared guard now uses `useAnimationFrame` to keep `transition-property` disabled through the first positioned paint before restoring the consumer value.
- Added representative Popover and Tooltip regressions and a dedicated React 17 Chromium job that runs both Positioner suites.
- Added first-open and reopen coverage for the Popup starting style, including `data-starting-style` at mount and opacity/transform progression.

CI is running on the new commit. Please take another look when it is green.

PR #5236, opened 2026-07-14
[navigation-menu] Fix return value for nodeChildrenContains in isOutsideMenuEvent

sai6855:
`nodeChildrenContains` falls back to `[]` instead of `false`, but it's only used as `!nodeChildrenContains` and `![] is false`, inverting the intended meaning.

### DIFF
--- a/packages/react/src/navigation-menu/utils/isOutsideMenuEvent.ts
+++ b/packages/react/src/navigation-menu/utils/isOutsideMenuEvent.ts
@@ -20,7 +20,7 @@ export function isOutsideMenuEvent({ currentTarget, relatedTarget }: Targets, pa
     ? getNodeChildren(tree.nodesRef.current, nodeId).some((node) =>
         contains(node.context?.elements.floating, relatedTarget),
       )
-    : [];
+    : false;
 
   // For nested scenarios without popupElement, we need to be more lenient
   // and only close if we're definitely outside the root


packages/react/src/navigation-menu/utils/isOutsideMenuEvent.ts
```diff
@@ -20,7 +20,7 @@ export function isOutsideMenuEvent({ currentTarget, relatedTarget }: Targets, pa
     ? getNodeChildren(tree.nodesRef.current, nodeId).some((node) =>
         contains(node.context?.elements.floating, relatedTarget),
       )
-    : [];
+    : false;
```
sai6855, line 23: Changing to false also changed type of `nodeChildrenContains` from `boolean| never[]` to `boolean`

atomiks:
# PR review

Nothing is merge-blocking. The `false` fallback correctly restores the intended boolean semantics when `tree` is `null` and fixes root-menu focus-out handling, but this regression path is not covered by a test.

## Bugs (0)

No findings.

## Tests (1)

### 1. 🟡 Add coverage for root-menu focus-out without a floating tree

**Location:** `packages/react/src/navigation-menu/utils/isOutsideMenuEvent.ts:23`

```ts
const nodeChildrenContains = tree
  ? getNodeChildren(tree.nodesRef.current, nodeId).some((node) =>
      contains(node.context?.elements.floating, relatedTarget),
    )
  : false;
```

The fix changes behavior specifically for a normal, non-nested Navigation Menu where `useFloatingTree()` returns `null`, but no test currently pins that path. An interaction test is more valuable than a helper-only unit test because it verifies that the real `NavigationMenuLink` or trigger blur path reaches this condition.

**Failure scenario:** Open a root Navigation Menu, focus a link inside its popup, then move focus to an unrelated element outside the root. The menu should close with `reason === 'focus-out'`, but a future regression to a truthy fallback could silently prevent that path from firing again.

**Fix:** Add a regression test that renders a non-nested Navigation Menu, opens it, focuses a popup link, moves focus outside the root, and asserts that it closes with `reason === 'focus-out'`.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve after nits** - the implementation is correct, with one focused regression test requested.

---

🤖 Review generated with Codex

michaldudak:
This has already been addressed in #5271

PR #5240, opened 2026-07-15
[navigation-menu] Fix frozen menu when the open trigger unmounts

### LINKED ISSUE #5239: [navigation menu] NavigationMenu can permanently freeze
# Bug report

## Current behavior

A top-level `NavigationMenu` can permanently freeze. The trigger `<ul>` is left with an inline
`pointer-events: none` that is never removed, so it intercepts every click and the whole menu (plus anything
layered under that `<ul>`) becomes unclickable. You can see `<ul ... style="pointer-events: none;">` stuck in
the DOM with no popup open.

That inline style is applied by the safe-polygon hover interaction and is meant to be a transient lock that
lasts only while a flyout is open. It becomes permanent when the `NavigationMenu.Trigger` that owns the lock
**unmounts while the lock is still active**.

This is easy to hit with the common "active item renders inline, without a trigger" pattern (a sidebar or
tab list where the currently-selected entry is rendered as plain content instead of a
`NavigationMenu.Trigger`): navigating to an item removes its trigger while its flyout lock is still active.
In real apps it looks non-deterministic because it depends on a render/commit race.

## Expected behavior

After the menu closes, the trigger `<ul>` has no residual `pointer-events: none` and the menu stays
interactive regardless of whether the trigger that opened the flyout is still mounted.

## Reproducible example

[StackBlitz reproduction](https://stackblitz.com/edit/vitejs-vite-u2zkx7uq?file=src%2Fmain.tsx)

The key is that the active item conditionally renders **without** a trigger, and the trigger is dropped
while its flyout lock is still active:

```tsx
<NavigationMenu.Root value={value} onValueChange={setValue} orientation="vertical">
  <NavigationMenu.List>
    {aIsActive ? (
      <NavigationMenu.Item value="A">
        <a href="#a">A (active — no trigger)</a>
      </NavigationMenu.Item>
    ) : (
      <NavigationMenu.Item value="A">
        <NavigationMenu.Trigger>A</NavigationMenu.Trigger>
        <NavigationMenu.Content>…</NavigationMenu.Content>
      </NavigationMenu.Item>
    )}
    <NavigationMenu.Item value="B">
      <NavigationMenu.Trigger>B</NavigationMenu.Trigger>
      <NavigationMenu.Content>…</NavigationMenu.Content>
    </NavigationMenu.Item>
  </NavigationMenu.List>
  <NavigationMenu.Portal>…</NavigationMenu.Portal>
</NavigationMenu.Root>
```

Steps:

1. Hover the **A** trigger so its flyout opens (this applies the lock, with the pointer on the trigger,
   never on the popup).
2. While A's flyout is open, in one batched update close the menu and drop A's trigger, simulating an async
   navigation completing while the pointer still rests on the trigger:
   ```tsx
   setValue(null);      // close the menu
   setAIsActive(true);  // A becomes "active" → its NavigationMenu.Trigger unmounts
   ```
3. The trigger `<ul>` is now stuck at `pointer-events: none` with no popup open. Hovering **B** does nothing → the menu is frozen.

## Base UI version

v1.6.0

Chrome

## Which OS are you using?

macOS

None

DreierF:
Fixes #5239 

The contributing guide does not have a PR section 🤔


atomiks:
# PR review

No correctness blockers. The component-scoped unmount cleanup addresses the stale list lock and is ownership-safe; the only suggestion is an optional consolidation with the existing hover-state effect. The PR’s Chromium regression passed, and the consolidated form also passed the full Navigation Menu suites in Chromium and jsdom.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (1)

### 1. ℹ️ Fold the unmount cleanup into the existing hover-state effect

**Location:** `packages/react/src/navigation-menu/trigger/NavigationMenuTrigger.tsx:582`

```tsx
React.useEffect(() => {
  return () => {
    clearSafePolygonPointerEventsMutation(hoverInteractionState);
  };
}, [hoverInteractionState]);
```

The standalone effect is correct, but the effect immediately above already owns the closed-state reset for the same `hoverInteractionState`. It can return this cleanup instead. The existing call inside `if (!open)` can then be removed: when `open` changes from `true` to `false`, React runs the preceding effect cleanup first, and unmounting runs the same cleanup.

The consolidated form passed all Navigation Menu tests: 286 passed and 17 skipped in Chromium; 278 passed and 25 skipped in jsdom.

**Failure scenario:** There is no user-visible failure in the current patch; the separate effect adds unnecessary hook work and production code for lifecycle handling that can share one effect.

**Fix:** Return `clearSafePolygonPointerEventsMutation(hoverInteractionState)` from the existing `[context, hoverInteractionState, open]` effect and remove both the standalone effect and the closed-state call.

Do not move this into `HoverInteraction.dispose()`: the shared interaction can have multiple hook consumers, so one consumer unmounting could clear a lock another still needs. `NavigationMenuTrigger` is the correct scope.

## Docs (0)

No findings.

## Verdict

**Approve** - the lifecycle fix is correct; consolidating the effect is optional.

---

🤖 Review generated with Codex

DreierF:
Applied the simplification

atomiks (approved):
Thanks for the fix @DreierF

Issue #5242, opened 2026-07-15
Make CssVars and DataAttributes enums public


superhooman:
# Feature request

## Summary

Export typed constants (enums) for CSS custom properties and data attributes exposed by Base UI components, so consumers can reference them in a statically-typed way instead of hardcoding string literals.

## Examples in other libraries

ArkUI kinda does that with their `@ark-ui/react/anatomy`. While it mainly exposes "parts", it has `_open` states, etc.

## Motivation

CSS variable and data attribute names exposed by Base UI components are currently only documented as plain strings, e.g.:
```
var(--nested-dialogs)
```
Instead, we could have something like:
```
var(${DialogPopupCssVars.nestedDialogs})
```

This would give autocomplete, compile-time safety against typos, and refactor safety if Base UI ever renames an internal var or attribute - all of which TypeScript can't currently catch, leading to silently broken styles. 
This is mainly for developers building custom styling on top of Base UI's unstyled components (CSS-in-JS, vanilla CSS with var(), or JS reading data-* attributes for conditional logic).

superhooman:
Trying to implement myself: #5243

jjenzz:
Thanks for this (and your work on the PR), I've added this topic to our agenda to discuss next week. I'll keep you posted 🙂

PR #5243, opened 2026-07-15
[all components] Export CssVariables and DataAttributes

superhooman:
Related to #5242 

35 components now export their `CssVariables`/`DataAttributes` constants publicly.

Includes component-specific aliases for borrowed Drawer and separator parts, a checked public export inventory with rendered-attribute tests, and a CSS-in-JS usage example in the styling handbook.

---

### Tree-shaking and consumer cost

The barrel-export design was checked with production ESM consumer builds using the published packages at `f25fa280` (this PR) and `0c34027c` (#5608). For the latter, the comparison added its modern constant exports to the existing component barrels instead of requiring `/metadata` imports.

- **Vite 8.2.2:** adding the exports produced byte-identical output (**0 B parsed / gzip**) for selective `Dialog`, `Menu`, `Select`, and `Button` imports from the root/component entrypoints, retaining the `Dialog` component namespace, and statically accessing parts through a namespace import. Both the existing enums and modern constants passed these comparisons.
- **Next.js 16.3.4 App Router:** clean webpack and Turbopack production builds with the modern constants re-exported from the barrels produced byte-identical client JavaScript (**0 B parsed / gzip**). The fixture included Server and Client Component imports from root/component entrypoints. No custom `optimizePackageImports` configuration was used.

The bundle checker's default fixture imports and retains the entire entrypoint namespace (`import * as everything; console.log(everything)`), so all newly exported metadata survives. That case grew by **9,873 B gzip** for the modern constants at the root entrypoint. It does not represent the cost to consumers whose unused exports are tree-shaken. Likewise, summing overlapping entrypoint deltas into a total consumer cost is misleading; the loss of constant inlining reported in #5608 did not occur in the selective-import comparisons.

A separate Vite fixture, before the public `CssVariables` rename, retaining all 199 metadata namespaces (974 entries, including derived-component aliases) produced **8,688 B gzip** without any components. Much of the size is the retained public names, string values, and namespace getter/object machinery; it should not all be attributed to lost constant inlining. Accessing only `DialogPopupCssVariables.nestedDialogs` compiled to the literal string `"--nested-dialogs"` with no retained namespace object. A fresh build after the rename also confirmed that `DialogPopupDataAttributes.open` and the handwritten `"data-open"` literal produce byte-identical output. These are measured fixtures, not a fixed cost per application.

These zero-delta results apply to the tested production tree-shaken bundles. Unbundled Node ESM and CommonJS imports still evaluate the re-exported metadata modules, even when consumers select only a component; separate metadata subpaths would isolate that module-loading cost.

Separate `/metadata` subpaths are therefore unnecessary to protect the tested production client bundles. #5608 was closed in favor of this design. This PR has been rebased onto current master and updated to the current constants, with derived-component aliases and public export tests included. After completing the missing metadata exports, the selective Vite fixtures remain byte-identical; retaining the entire root entrypoint adds **10,037 B gzip**. The current public inventory contains 205 metadata namespace exports with 982 entries, including aliases. Public export tests passed in jsdom and Chromium, alongside the package build, TypeScript, ESLint, Stylelint, API generation, documentation snapshot, and formatting checks.

### DIFF
--- a/docs/scripts/generateLlmTxt/__snapshots__/mdxToMarkdown.test.mjs.snap
+++ b/docs/scripts/generateLlmTxt/__snapshots__/mdxToMarkdown.test.mjs.snap
@@ -907,6 +907,54 @@ type AccordionPanelState = {
 
 ## Additional Types
 
+### AccordionHeaderDataAttributes
+
+| Member   | Value           | Description                                  |
+| :------- | :-------------- | :------------------------------------------- |
+| index    | \`data-index\`    | Indicates the index of the accordion item.   |
+| disabled | \`data-disabled\` | Present when the accordion item is disabled. |
+| open     | \`data-open\`     | Present when the accordion item is open.     |
+
+### AccordionItemDataAttributes
+
+| Member   | Value           | Description                                  |
+| :------- | :-------------- | :------------------------------------------- |
+| index    | \`data-index\`    | Indicates the index of the accordion item.   |
+| disabled | \`data-disabled\` | Present when the accordion item is disabled. |
+| open     | \`data-open\`     | Present when the accordion item is open.     |
+
+### AccordionPanelCssVars
+
+| Member               | Value                      | Description                   |
+| :------------------- | :------------------------- | :---------------------------- |
+| accordionPanelHeight | \`--accordion-panel-height\` | The accordion panel's height. |
+| accordionPanelWidth  | \`--accordion-panel-width\`  | The accordion panel's width.  |
+
+### AccordionPanelDataAttributes
+
+| Member        | Value                 | Description                                  |
+| :------------ | :-------------------- | :------------------------------------------- |
+| index         | \`data-index\`          | Indicates the index of the accordion item.   |
+| open          | \`data-open\`           | Present when the accordion panel is open.    |
+| orientation   | \`data-orientation\`    | Indicates the orientation of the accordion.  |
+| disabled      | \`data-disabled\`       | Present when the accordion item is disabled. |
+| startingStyle | \`data-starting-style\` | Present when the panel begins animating in.  |
+| endingStyle   | \`data-ending-style\`   | Present when the panel is animating out.     |
+
+### AccordionRootDataAttributes
+
+| Member      | Value              | Description                                 |
+| :---------- | :----------------- | :------------------------------------------ |
+| disabled    | \`data-disabled\`    | Present when the accordion is disabled.     |
+| orientation | \`data-orientation\` | Indicates the orientation of the accordion. |
+
+### AccordionTriggerDataAttributes
+
+| Member    | Value             | Description                                  |
+| :-------- | :---------------- | :------------------------------------------- |
+| panelOpen | \`data-panel-open\` | Present when the accordion panel is open.    |
+| disabled  | \`data-disabled\`   | Present when the accordion item is disabled. |
+
 ### AccordionValue
 
 \`\`\`typescript
@@ -928,7 +976,7 @@ type Orientation = 'horizontal' | 'vertical';
 - \`Accordion.Header\`: \`Accordion.Header\`, \`Accordion.Header.State\`, \`Accordion.Header.Props\`
 - \`Accordion.Trigger\`: \`Accordion.Trigger\`, \`Accordion.Trigger.State\`, \`Accordion.Trigger.Props\`
 - \`Accordion.Panel\`: \`Accordion.Panel\`, \`Accordion.Panel.State\`, \`Accordion.Panel.Props\`
-- \`Default\`: \`AccordionValue\`, \`AccordionRootState\`, \`AccordionRootProps\`, \`AccordionRootChangeEventReason\`, \`AccordionRootChangeEventDetails\`, \`AccordionItemState\`, \`AccordionItemProps\`, \`AccordionItemChangeEventReason\`, \`AccordionItemChangeEventDetails\`, \`AccordionHeaderState\`, \`AccordionHeaderProps\`, \`AccordionTriggerState\`, \`AccordionTriggerProps\`, \`AccordionPanelState\`, \`AccordionPanelProps\`
+- \`Default\`: \`AccordionPanelCssVars\`, \`AccordionHeaderDataAttributes\`, \`AccordionItemDataAttributes\`, \`AccordionPanelDataAttributes\`, \`AccordionRootDataAttributes\`, \`AccordionTriggerDataAttributes\`, \`AccordionValue\`, \`AccordionRootState\`, \`AccordionRootProps\`, \`AccordionRootChangeEventReason\`, \`AccordionRootChangeEventDetails\`, \`AccordionItemState\`, \`AccordionItemProps\`, \`AccordionItemChangeEventReason\`, \`AccordionItemChangeEventDetails\`, \`AccordionHeaderState\`, \`AccordionHeaderProps\`, \`AccordionTriggerState\`, \`AccordionTriggerProps\`, \`AccordionPanelState\`, \`AccordionPanelProps\`
 
 ## Canonical Types
 
--- a/docs/src/app/(docs)/react/components/accordion/types.md
+++ b/docs/src/app/(docs)/react/components/accordion/types.md
@@ -352,6 +352,54 @@ type AccordionPanelState = {
 
 ## Additional Types
 
+### AccordionHeaderDataAttributes
+
+| Member   | Value           | Description                                  |
+| :------- | :-------------- | :------------------------------------------- |
+| index    | `data-index`    | Indicates the index of the accordion item.   |
+| disabled | `data-disabled` | Present when the accordion item is disabled. |
+| open     | `data-open`     | Present when the accordion item is open.     |
+
+### AccordionItemDataAttributes
+
+| Member   | Value           | Description                                  |
+| :------- | :-------------- | :------------------------------------------- |
+| index    | `data-index`    | Indicates the index of the accordion item.   |
+| disabled | `data-disabled` | Present when the accordion item is disabled. |
+| open     | `data-open`     | Present when the accordion item is open.     |
+
+### AccordionPanelCssVars
+
+| Member               | Value                      | Description                   |
+| :------------------- | :------------------------- | :---------------------------- |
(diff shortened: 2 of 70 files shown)


superhooman:
So, I tried to just straight up exporting these enums as is

But I have a questions:
- Are we okay with no reexports from _derived_ components like AlertDialog (Dialog) and Autocomplete (Combobox)? Since they use the same enums as underlying components
- Should I place these exports in a separate barrel file? So it looks cleaner in the main one
- Is it even worth it given it increases bundle size? Perhaps we can publish it as a separate package?

jjenzz:
hi @superhooman, we recently changed our exports to `const`s instead of `enum`s. would you mind addressing the conflicts here to expose those instead? you might need to run `pnpm docs:api` when done too.

> Are we okay with no reexports from derived components like AlertDialog (Dialog) and Autocomplete (Combobox)? Since they use the same enums as underlying components

things might be different now after our refactor, but i would personally expect comps to have their own equivalent named exports so that it's possible for `AlertDialog` to diverge from `Dialog` (if needs be) without a breaking change. 

@atomiks what do you think?

> Should I place these exports in a separate barrel file? So it looks cleaner in the main one

exporting from `index` is fine, they'll be tree-shaken.

> Is it even worth it given it increases bundle size? Perhaps we can publish it as a separate package?

the `const` change we did aims to address some of that, but they will be tree-shaken so should be fine :)

superhooman:
Hi! 
Sorry, just noticed this

I guess this PR should be closed as exporting is being done by @atomiks in #5608

jjenzz:
@superhooman i think we agreed internally that your approach of exposing from the `index.ts` barrel files is preferred, is that correct @atomiks ?

atomiks:
Yes this approach works fine (treeshakes) after testing locally, so existing consumers don't pay the +10 KB increase reported by the bundle bot. Only those who want to import the metadata pay the cost. 

I've renamed `CssVars` to `CssVariables` as `DataAttributes` isn't `DataAttrs`; it should likely spell out the full word for consistency now that it is public

packages/react/src/index.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+  it('should have exports', () => {
+    expect(typeof BaseUI).toBe('object');
+  });
+
+  it('should not have undefined exports', () => {
+    Object.keys(BaseUI).forEach((exportKey) => {
+      const value = (BaseUI as Record<string, unknown>)[exportKey];
+      expect(Boolean(value)).toBe(true);
+    });
+  });
```
jjenzz, line 22: do we need these? i dno, just thinking they're not particularly pragmatic, and if we allow adding tests like this everywhere we will be slowing our test suite down for little gain?

packages/react/src/index.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+    expect(BaseUI.DialogPopupCssVariables.nestedDialogs).toBe('--nested-dialogs');
+    expect(BaseUI.DialogPopupDataAttributes.open).toBe('data-open');
+  });
+
+  it('should export borrowed metadata under the derived component name', () => {
+    expect(BaseUI.AlertDialogPopupCssVariables).toBe(BaseUI.DialogPopupCssVariables);
+    expect(BaseUI.AlertDialogPopupDataAttributes).toBe(BaseUI.DialogPopupDataAttributes);
+    expect(BaseUI.AutocompletePopupDataAttributes).toBe(BaseUI.ComboboxPopupDataAttributes);
+    expect(BaseUI.ContextMenuItemDataAttributes).toBe(BaseUI.MenuItemDataAttributes);
+  });
```
jjenzz: same here? this seems better covered by a lint rule rather than a test (obvs don't need to create lint rule here tho btw).

packages/react/src/index.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  it('should expose the disabled attribute rendered by Button', async () => {
+    await render(<BaseUI.Button disabled>Button</BaseUI.Button>);
+
+    expect(screen.getByRole('button')).toHaveAttribute(BaseUI.ButtonDataAttributes.disabled);
+  });
```
jjenzz, line 55: this feels like it should exist in button tests rather than overall base-ui package tests?

packages/react/src/index.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  it('should expose the attributes rendered by borrowed Drawer parts', async () => {
```
jjenzz, line 57: same here, should we move these to `Drawer` tests?

packages/react/src/index.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+    expect(screen.getByRole('button', { name: 'Close' })).toHaveAttribute(
+      BaseUI.DrawerCloseDataAttributes.disabled,
+    );
+
+    const trigger = screen.getByRole('button', { name: 'Open' });
+    await user.click(trigger);
+    expect(trigger).toHaveAttribute(BaseUI.DrawerTriggerDataAttributes.popupOpen);
+  });
+
+  it.each(['Select', 'Combobox', 'Autocomplete'] as const)(
```
jjenzz, line 78: and again, should these be Select, Combobox, and Autocomplete tests rather than base-ui tests?

jjenzz (commented):
# DX review

The public API direction is sound. Two documentation and maintenance issues should be fixed, but neither blocks the API design.

## Findings (2)

### 1. 🟠 Runtime constants are documented as nonexistent types

**Source:** [[button/types.md:41](https://github.com/mui/base-ui/blob/f99930422a4db90a1871bdd460f2eac379a064e8/docs/src/app/%28docs%29/react/components/button/types.md#L41-L49)](https://github.com/mui/base-ui/blob/f99930422a4db90a1871bdd460f2eac379a064e8/docs/src/app/%28docs%29/react/components/button/types.md#L41-L49)

**Problem:**

The generated reference places runtime constants under “Additional Types” and renders declarations such as `type ButtonDataAttributesdisabled = 'data-disabled'`. That type does not exist, and the concatenated name does not represent the public API.

**User scenario:**

A consumer consulting the API reference may try to import the documented type or use a type-only import instead of importing `ButtonDataAttributes` as a runtime value.

**Suggested change:**

> Could we update the API generator to present these exports as runtime constants rather than additional types? For example, use a “Constants” section with importable syntax that preserves the namespace and member name.
>
> Please also keep these constants out of the generated `Types:` inventories and regenerate the component references and LLM snapshots. This will make the reference agree with the actual `MenuItemDataAttributes.highlighted` usage shown in the styling handbook.

### 2. 🟡 Button has two source files for the same module

**Source:** [[ButtonDataAttributes.ts:1](https://github.com/mui/base-ui/blob/f99930422a4db90a1871bdd460f2eac379a064e8/packages/react/src/button/ButtonDataAttributes.ts#L1-L4)](https://github.com/mui/base-ui/blob/f99930422a4db90a1871bdd460f2eac379a064e8/packages/react/src/button/ButtonDataAttributes.ts#L1-L4), [[ButtonDataAttributes.tsx:1](https://github.com/mui/base-ui/blob/f99930422a4db90a1871bdd460f2eac379a064e8/packages/react/src/button/ButtonDataAttributes.tsx#L1-L4)](https://github.com/mui/base-ui/blob/f99930422a4db90a1871bdd460f2eac379a064e8/packages/react/src/button/ButtonDataAttributes.tsx#L1-L4)

**Problem:**

Both files contain the same export, while the extensionless import resolves only one of them. A later edit could update the unused file and leave the public constant stale.

**User scenario:**

A maintainer changes the existing `.tsx` file when renaming the rendered attribute, but the public barrel continues resolving the `.ts` copy.

**Suggested change:**

> Could we retain only `ButtonDataAttributes.ts` and remove the existing `.tsx` copy? The module contains no JSX, and one source file will preserve the refactor-safety guarantee this API is intended to provide.

## What is already sound

- The `DataAttributes` and `CssVariables` names are consistent and predictable.
- Derived components receive their own aliases, allowing later divergence without renaming their public exports.
- The handbook demonstrates valid selector and `var()` interpolation.
- The export inventory and rendered-attribute tests cover the important public behavior.
- I checked all 5,491 added TypeScript and documentation lines with Codespell. No spelling mistakes were found; the PR’s Vale check also passes.

## Verdict

**Approve after nits** — The API is coherent, but the generated documentation should describe the exports accurately and the duplicate Button module should be removed.

---

🤖 Review generated with Codex

jjenzz:
Separately, is it intentional that these exports only appear under "Additional Types" in the generated Markdown/LLM output? They don't appear in the component docs atm.

jjenzz:
@Janpot — `pnpm docs:api` appears to need a fix in `@mui/internal-docs-infra` for the runtime constants exported by this PR.

For example, the public API is `AutocompleteTriggerDataAttributes.pressed`, whose value is `'data-pressed'`, but the generated reference currently invents a type that is not exported:

```ts
type AutocompleteTriggerDataAttributespressed = 'data-pressed';
```

The generator recognizes component exports separately, but these string constants fall through to the raw-type formatter. In `formatRaw.mjs`, `generateFormattedCode` removes the dots from the display name and formats the result as a type alias.

**Codex proposed the following targeted fix** (a suggestion for review, not an implemented or tested infra change): handle these metadata members before stripping the dots, and emit a namespace constant declaration instead.

```js
const constant = displayName.match(
  /^(.+(?:DataAttributes|CssVariables))\.([^.]+)$/,
);

if (constant) {
  const [, namespaceName, memberName] = constant;
  const valueType = formatType(exportNode.type, {
    typeNameMap,
    externalTypesCollector,
  });

  return prettyFormat(
    `declare namespace ${namespaceName} {
      const ${memberName}: ${valueType};
    }`,
    null,
  );
}
```

This would generate:

```ts
declare namespace AutocompleteTriggerDataAttributes {
  const pressed: 'data-pressed';
}
```

The intent is to preserve accurate export information in the generated Markdown/LLM docs without adding duplicate Constants tables to the website, which already has attribute and CSS-variable tables. The fix should live in the generator so subsequent `pnpm docs:api` runs preserve it. A regression test should cover the emitted declaration and ensure ordinary component/type formatting remains unchanged.

— Posted by Codex on Jenna’s behalf; the proposed implementation above is Codex’s suggestion.

Janpot:
Trying out a [fix](https://github.com/mui/mui-public/pull/1888) in https://github.com/mui/base-ui/pull/5784

PR #4236, opened 2026-03-02
[autocomplete][docs] Add command palette search icon and stronger highlight

atomiks:
- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/base-ui/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

Fixes https://github.com/mui/base-ui/issues/4182
Fixes https://github.com/mui/base-ui/issues/4183

For the highlight, I updated the SearchBar (cmdk) search on the docs to use `--color-blue` highlight, same as the style Select in code blocks and what macOS uses for Spotlight.

Screenshot 2026-03-03 at 12 53 44 pm

The highlight background needs to have a contrast ratio of `3:1` with non-highlight backgrounds if there is no focus-only style — an outline for `:focus` would work otherwise, but we don't have a data state attribute to do this unless you use `highlightItemOnHover={false}` to differentiate `:hover` from `:focus`, similar to Chrome's omnibox. @colmtuite 

---

Command Palette uses the black highlight, same as other demos, and adds a search icon:

Screenshot 2026-03-03 at 12 54 07 pm

(imo all demos should use `--color-blue` as the highlight, but that's a different fix for the docs overhaul)

PR #5245, opened 2026-07-16
[docs][autocomplete] Add persistent search icon to command palette demo

### LINKED ISSUE #4183: [docs] Command Palette autocomplete example has no persistently visible label
# Bug report

## Current behavior

Inside the Command Palette dialog, there is an input labelled by placeholder text only. This presents an issue for people with cognitive and/or memory impairments and is a failure against [WCAG 2.2 Success Criterion 3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)

## Expected behavior

Inputs must have a persistently visible label, either in text or as an icon. The search input is fine on the commands dialog because it has a persistently visible search icon that is recognisable. However, in this case, there wouldn't be an appropriate icon for "Search for apps and commands..." so a text label would is needed.

## Reproducible example

Steps to reproduce:
1. Navigate to the [Autocomplete command palette demo](https://base-ui.com/react/components/autocomplete#command-palette)
2. Open the command palette 
3. Type into the input 
4. Note the visible label has disappeared.

## Base UI version

v1.2.0

Chrome

## Which OS are you using?

Mac OS

Keyboard

## Additional context

Inputs must have a persistently visible label. Further reading/resources:

* [Brad Frost: Float label pattern](https://bradfrost.com/blog/post/float-label-pattern/)
* [CSS Tricks: Float labels with CSS](https://css-tricks.com/float-labels-css/)
* [HTML Living Standard: The placeholder attribute](https://html.spec.whatwg.org/multipage/input.html#the-placeholder-attribute)
* [Matt Smith: Accessible floating labels](https://allthingssmitty.com/2016/09/25/accessible-floating-labels/)
* [UX Matters: Don't put hints inside text boxes in web forms](https://www.uxmatters.com/mt/archives/2010/03/dont-put-hints-inside-text-boxes-in-web-forms.php)
* [W3C: WCAG 2.2 Techniques - Technique H44 Using label elements to associate text labels with form controls](https://www.w3.org/WAI/WCAG22/Techniques/html/H44)
* [Deque: The problem with placeholders](https://www.deque.com/blog/accessible-forms-the-problem-with-placeholders/)

mmaxence:

Fixes #4183

The command palette input is labelled by its placeholder only, which disappears on type. This adds a persistent search icon to the input, as agreed in the issue. Both demo variants updated.

The icon is aria-hidden and the input keeps its aria-label. Its colour matches the placeholder grey, above 3:1 contrast in light and dark. The focus outline moves to the input row so icon and input read as one field. Scoped to the icon only, none of the highlight styling from #4236.


docs/src/app/(docs)/react/components/autocomplete/demos/command-palette/css-modules/index.tsx
```diff
@@ -94,6 +97,23 @@ export default function ExampleAutocompleteCommandPalette() {
...
+      strokeWidth="2"
+      strokeLinecap="round"
+      strokeLinejoin="round"
+      {...props}
+    >
+      <circle cx="11" cy="11" r="8" />
+      <path d="m21 21-4.3-4.3" />
+    </svg>
+  );
+}
```
aarongarciah: Let's copy the icon from `/docs/src/icons/MagnifyingGlassIcon.tsx`, which is a real 16x16 icon. Same in the Tailwind demo.

docs/src/app/(docs)/react/components/autocomplete/demos/command-palette/css-modules/index.tsx
```diff
@@ -27,12 +27,15 @@ export default function ExampleAutocompleteCommandPalette() {
               autoHighlight="always"
               keepHighlight
             >
-              <Autocomplete.Input
-                className={styles.Input}
-                aria-label="Search commands"
-                aria-describedby={shortcutsDescriptionId}
-                placeholder="Search for apps and commands…"
-              />
+              <div className={styles.InputRow}>
```
aarongarciah: The icon area is a click "dead" zone. The more straightforward fix is to use `Autocomplete.InputGroup` instead of the plain `div`.

Longer explanation by Fable 5:
> Before this change, the input element spanned the full row (its own padding: 0 0.75rem), so clicking anywhere in the strip focused it and placed the caret. Now the leading ~2.25rem (start padding + 1rem icon + 0.5rem gap) belongs to a non-focusable div: clicking the search icon no longer focuses the input, and since the default mousedown behavior moves focus off the input, it can actively blur it mid-use — clicking the icon while typing interrupts input.

> Base UI already has the part built for this: Autocomplete.InputGroup ("A wrapper for the input and its associated controls", listed in the anatomy at docs/src/app/(docs)/react/components/autocomplete/page.mdx:27 and used by the combobox hero demo). Its onMouseDown handler (handleInputPress.ts:34) prevents the default and focuses the input when pressing anywhere non-interactive inside the group, which fixes the dead zone for free, and it adds role="group" plus state attributes. A docs demo teaching a plain-div wrapper when the dedicated public part exists is also the wrong pattern for readers to copy.

docs/src/app/(docs)/react/components/autocomplete/demos/command-palette/css-modules/index.module.css
```diff
@@ -115,11 +115,43 @@
...
-.Input {
+.InputRow {
   box-sizing: border-box;
   position: relative;
   z-index: 1;
-  padding: 0 0.75rem;
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  padding-inline-start: 0.75rem;
```
aarongarciah: Let's stick to physical CSS properties as the rest of the demo (and the Tailwind demo as well).

Switching to logical properties is something we should probably do at some point, but it's better not to mix them for now.


```suggestion
  padding-left: 0.75rem;
```

docs/src/app/(docs)/react/components/autocomplete/demos/command-palette/css-modules/index.module.css
```diff
@@ -115,11 +115,43 @@
   }
 }
 
-.Input {
+.InputRow {
```
aarongarciah: Let's use `.InputGroup` instead since we're going to use `Autocomplete.InputGroup`.


```suggestion
.InputGroup {
```

aarongarciah (commented):
Thanks for working on this @mmaxence. I left inline comments on a few things we need to fix before merging.

mmaxence:
Applied all four: swapped the wrapper to `Autocomplete.InputGroup` (clicking anywhere in the strip focuses the input now), copied the 16x16 `MagnifyingGlassIcon` into both demo variants, switched to `padding-left` (and `ps-3` → `pl-3` in the Tailwind demo for the same reason), and renamed the class to `.InputGroup`.

aarongarciah (approved):
Thanks @mmaxence!

PR #5246, opened 2026-07-16
[button][docs] Announce the loading state in the loading demo

### LINKED ISSUE #4249: [button] NVDA doesn't announce loading state example correctly
# Bug report

## Current behavior

The "Loading states" variant includes a button that becomes temporarily disabled and changes its visible label from "Submit" to "Submitting". NVDA does not announce the "Submitting" text, only "Unavailable".

VoiceOver announces both the changed label and dimmed (disabled) state correctly.

## Expected behavior

All screen readers should announce "submitting".

## Reproducible example

1. On a Windows machine, turn on NVDA screen reader
2. Navigate to [Button: Loading states](https://base-ui.com/react/components/button#loading-states)
3. Activate the button
4. Note only "unavailable" is announced and not the visible "submitting" text

## Base UI version

v1.2.0

Chrome

## Which OS are you using?

Windows

NVDA

## Additional context

The button could have an `aria-live` attribute on it by default.

mmaxence:

Fixes #4249

The loading demo swaps its label from Submit to Submitting, but NVDA announces only "unavailable" when the button becomes disabled — the label change itself is never spoken.

This PR makes the changing label the button's explicit accessible name: the label text is wrapped in a `<span>` with an id (from `React.useId()`) that the button references via `aria-labelledby`. Per the [accessible name computation](https://www.w3.org/TR/accname-1.2/#computation-steps), the label swap then surfaces as a name change on the focused button, which screen readers announce — see [the NVDA Speech Viewer output](https://github.com/mui/base-ui/pull/5246#issuecomment-5493155836) for this approach.

Earlier revisions of this PR tried two other mechanisms, both abandoned:

- A visually hidden sibling `role="status"` region mirroring the loading state. It works and is the common ecosystem pattern, but it needs an extra element plus CSS, and clearing the region at the end of loading is not announced, so completion stayed silent.
- `aria-live="polite"` on the button itself. Testing on NVDA + Chrome (Assistiv Labs) confirmed it announces both transitions, but every announcement was doubled ("Submitting, Submitting"), and there is no precedent for rooting a live region on an interactive control.

The Loading states section in `page.mdx` now explains the pattern and links the spec. Both demo variants updated.

### DIFF
--- a/docs/src/app/(docs)/react/components/button/demos/loading/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/button/demos/loading/css-modules/index.tsx
@@ -7,18 +7,23 @@ export default function ExampleButton() {
   const [loading, setLoading] = React.useState(false);
 
   return (
-    <Button
-      className={styles.Button}
-      disabled={loading}
-      focusableWhenDisabled
-      onClick={() => {
-        setLoading(true);
-        setTimeout(() => {
-          setLoading(false);
-        }, 4000);
-      }}
-    >
-      {loading ? 'Submitting' : 'Submit'}
-    </Button>
+    <React.Fragment>
+      <Button
+        className={styles.Button}
+        disabled={loading}
+        focusableWhenDisabled
+        onClick={() => {
+          setLoading(true);
+          setTimeout(() => {
+            setLoading(false);
+          }, 4000);
+        }}
+      >
+        {loading ? 'Submitting' : 'Submit'}
+      </Button>
+      <span role="status" className={styles.VisuallyHidden}>
+        {loading ? 'Submitting' : null}
+      </span>
+    </React.Fragment>
   );
 }
--- a/docs/src/app/(docs)/react/components/button/demos/loading/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/button/demos/loading/css-modules/index.tsx
@@ -7,23 +7,19 @@ export default function ExampleButton() {
   const [loading, setLoading] = React.useState(false);
 
   return (
-    <React.Fragment>
-      <Button
-        className={styles.Button}
-        disabled={loading}
-        focusableWhenDisabled
-        onClick={() => {
-          setLoading(true);
-          setTimeout(() => {
-            setLoading(false);
-          }, 4000);
-        }}
-      >
-        {loading ? 'Submitting' : 'Submit'}
-      </Button>
-      <span role="status" className={styles.VisuallyHidden}>
-        {loading ? 'Submitting' : null}
-      </span>
-    </React.Fragment>
+    <Button
+      aria-live="polite"
+      className={styles.Button}
+      disabled={loading}
+      focusableWhenDisabled
+      onClick={() => {
+        setLoading(true);
+        setTimeout(() => {
+          setLoading(false);
+        }, 4000);
+      }}
+    >
+      {loading ? 'Submitting' : 'Submit'}
+    </Button>
   );
 }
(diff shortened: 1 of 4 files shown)


docs/src/app/(docs)/react/components/button/demos/loading/css-modules/index.tsx
```diff
@@ -7,18 +7,23 @@ export default function ExampleButton() {
...
+            setLoading(false);
+          }, 4000);
+        }}
+      >
+        {loading ? 'Submitting' : 'Submit'}
+      </Button>
+      <span role="status" className={styles.VisuallyHidden}>
+        {loading ? 'Submitting' : null}
+      </span>
+    </React.Fragment>
```
aarongarciah: @mmaxence do you have the capacity to investigate if it's possible to return a single React element (i.e. `<Button>`) instead of `React.Fragment`? Returning `<React.Fragment>` would be problematic in a component lib since users/LLMs will most certainly copy this example as a starting point.

mmaxence: Reworked: the demo now returns a single `<Button>` with `aria-live="polite"`, and the sibling status region is gone. The Submit / Submitting label change announces from the button's own live region, which is also what the issue author originally suggested in #4249. I skipped moving the `role="status"` span inside the button because button children are presentational per ARIA, so a nested live region is unreliable across browsers. Checked the Chrome accessibility tree: the button exposes `live: polite` and its name flips between Submit and Submitting through the cycle. The overall diff gets smaller too. A quick NVDA pass would still be a useful double check since that was the report's environment.

mmaxence:
@aarongarciah bump: reworked to the single-element approach you suggested (aria-live on the Button itself, f3107568). Let me know if anything else is needed.

docs/src/app/(docs)/react/components/button/demos/loading/css-modules/index.tsx
aarongarciah: I tested the `aria-live` approach on the button and it's problematic since NVDA announces the label twice ("Submitting, Submitting -> Submit, Submit"). I've reverted the second commit so the demo uses the original approach with the `React.Fragment` + `role="status"` separate element.

jjenzz:
Can we update the `Button` with `aria-labelledby` pointing to the "Submitting" text? Using `aria-labelledby` [forces hidden text to be included in the a11y tree](https://www.w3.org/TR/accname-1.2/#computation-steps) so NVDA will read it aloud.

CleanShot 2026-09-01 at 12 09 50

demo: https://stackblitz.com/edit/1vkkyije?file=src%2FApp.tsx

```diff
export default function ExampleButton() {
  const [loading, setLoading] = React.useState(false);
+ const labelId = React.useId();

  return (
    <Button
      className={styles.Button}
      disabled={loading}
      focusableWhenDisabled
+     aria-labelledby={labelId}
      onClick={() => {
        setLoading(true);
        setTimeout(() => {
          setLoading(false);
        }, 4000);
      }}
    >
+     <span id={labelId}>
        {loading ? 'Submitting' : 'Submit'}
+     </span>
    </Button>
  );
}
```

We'd probably need to update the copy to explain why that is necessary also (it can point to the spec).

aarongarciah:
@jjenzz if we accept the trade-off of users not receiving feedback at the end of the operation when they tab away mid-operation, it's a great approach. I’ve updated the PR.

jjenzz (approved):

aarongarciah:
Thanks @mmaxence!

PR #5247, opened 2026-07-16
[autocomplete][docs] Clarify that the item value type constraint is unconditional

### LINKED ISSUE #3853: [docs][autocomplete] Confusing TypeScript inference section

mmaxence:

Fixes #3853

The TypeScript inference section implied the `value` prop constraint only applies when using `itemToStringValue`; it always applies. Reworded to state the constraint unconditionally and tie `itemToStringValue` to type inference rather than display behaviour, following the discussion in the issue.


atomiks (approved):
Thanks for the fix @mmaxence

PR #5248, opened 2026-07-16
[all components] Use ESM metadata constants

atomiks:
Replace the `DataAttributes` and `CssVars` string enums with named ESM string constants, imported as namespaces in source. Application bundlers can then inline and tree-shake the values without a custom TypeScript-checker transform, and the metadata modules stay erasable-syntax-only.

Unblocked by mui/mui-public#1713 (merged): docs metadata extraction now supports named constants, and `pnpm docs:validate` output is unchanged.

## Note

- `@mui/internal-docs-infra` is pinned to the pkg.pr.new build of mui/mui-public#1713 (`a2333652`) because the mui-public canary publish workflow has been failing since 2026-08-25. Will be swapped to a published canary once one exists.


Janpot:
Checking the [diff](https://code-infra-dashboard.onrender.com/diff-package?package1=https://pkg.pr.new/mui/base-ui/@base-ui/react@eaf727c&package2=https://pkg.pr.new/mui/base-ui/@base-ui/react@d34678f). Here's my human review:

Imo there are several downsides to this setup
* it's considerably slower.
* it clashes with per-file transforms, which is how babel/swc/esbuild/... think about these typescript files, pure type-stripping.
* It adds critical logic in custom babel plugins, which makes it hard to move around between build tools, i.e. it confines us to the performance profile of babel, which we hope to get away from at some point.
* It requires careful configuration on the authors' side, and is easily broken by renaming things. Big tech debt.
* It conflates types with runtime, and introduces type dependent runtime behavior. e.g. an ambient `declare enum FooDataAttributes { checked = 'data-checked' }` has no runtime object at all, yet `FooDataAttributes.checked` inlines and "works" under this transform. Could be error prone.
* End-users using these enums on their end can't benefit from the same optimization.

In short, if code infra has to adopt this complexity, I first want to make sure it's the right solution to the problem. The thing is that our users already have tools in their toolchain that can inline values like this. e.g. terser will take the following

```tsx
const MyEnum = {
  f: 'foo'
}

export const y = {
 [MyEnum.f]: 'bar'
}
```

and happily produce

```tsx
export const y={foo:"bar"};
```

out of it in your app bundles if you use it. The problem we have and the reason why this is not already automatically happening is that babel typescript emits those enums as

```tsx
let MyEnum = function (MyEnum) {
  MyEnum["f"] = "foo";
  return MyEnum;
}({});
export const y = {
 [MyEnum.f]: 'bar'
}
```

and terser doesn't know how to inline this. It will produce
```tsx
export const y={[function(y){return y.f="foo",y}({}).f]:"bar"};
```

Moreover, different compilers emit different things for these enums, and exactly because they hold runtime semantics based on types, typescript suggests to keep away from them ([`erasableSyntaxOnly`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8/#the---erasablesyntaxonly-option)). So I wonder, is it possible for us to express these "enums" as plain old javascript objects with `as const`? With the right configuration, users machines can yield the same output, at no extra tech debt. (to note: not all minifiers inline these by default, but real perf-conscious users should know how to configure.)

The only real downsides I see
* we don't have the install-size win, but I don't see a good argument to optimize for this unless it's a really big win.
* we currently have the enums doing double duty as a value, and as a type. If we want to preserve that behavior, but still move to a erasable syntax, we're going to need to [emulate](https://www.typescriptlang.org/play/?#code/PTAEGEHsFtsg7ACpADgVxQEQIYBdsCCuuATgJYBGauApgM4B0udAUCzQB4qQm6gDGCOnyiwEydFjyFi5KrTqgAvKADeLUKGAAqbaEQl6NeHwDuAC2Ohcl0N0mgyi1MYahtwDaBfwAXKAByABNpAFofAIAaLx09AyMTUAsrGxo7VAxHRX4AG0g6GiC3Dy9c-ML-YLCygqComN19QwLE5PhrW3tMihoAczJ4RWx4Mmg8Ad7HeGLPTWFsXgmAZVwATxyaSpD8UPnF+F7dtY36zVimhLNLdtT0hydQYdHxg+9qGa9jIOXjzcDt7ChL4TI7rGj1AC+j2yQlwAG42JxuLxrKsUGlRHAkBkpPgiKRKNR6MovAAKNboyAAMwgMCxEgwODxskJCgAlABtADWNFW1NRlJpmPEOKZMgJ8noAF0EewuDw+PBsNB6ChsPwMXSRZIxfi5ETFOpNEiFQK0j5lGb+cLsTrpHrWfQGD4Ecb5SiKWkaoVLZ7rVrbYz7SzJYxvUFXaATR60Wk9rgfmDfbH-WJA7jxfqFAx44mNpHo3xPVH4N8Diskyo-UKAwyMw7QwxgeXfgiIWwQKBMGRsHlenXdSGDUxWHLkXxBIM+N3e5B+6LgxKDZajaAGOubQPF1n6NEzo14i0rilbN9Z5MHvB6LQgkkyDYBo94JBUiRQGe+x9NFfhBV-mEfxvU4tAPZpjGPG5Tx7PtQHMbBnFfbx0XaD850UQCfVMe9zEfe8v1ADCghnPsAHlkK2ADr0KUJUMOCJoiheCBFhWVCzNLtoLnLdmSXBQSU0ckUxpYiuIXHidzoTkeT5Glqw489uMzR06BlRF3UVZVVXVNIRPnO1xOUtRPnU9iLSrIT5L7RSG2HF1jPHdjw2TQVLNE-SlMbcMCxM4tcxbSsrWEzi9KDAzGz83oK3zezTWLZtIt+ZyaH5XTrKHbN4qimhvIc4tCKSlLgrS3inUInLYtjAiqKI4KyKscyXNSsSPOHQjdLq+A2yAA) it with a slightly less ergonomic shape. (note we're using namespaces here, but in a type erasable way)

_edit: Just noticed that the pattern with `...` in the [playground](https://www.typescriptlang.org/play/?#code/PTAEGEHsFtsg7ACpADgVxQEQIYBdsCCuuATgJYBGauApgM4B0udAUCzQB4qQm6gDGCOnyiwEydFjyFi5KrTqgAvKADeLUKGAAqbaEQl6NeHwDuAC2Ohcl0N0mgyi1MYahtwDaBfwAXKAByABNpAFofAIAaLx09AyMTUAsrGxo7VAxHRX4AG0g6GiC3Dy9c-ML-YLCygqComN19QwLE5PhrW3tMihoAczJ4RWx4Mmg8Ad7HeGLPTWFsXgmAZVwATxyaSpD8UPnF+F7dtY36zVimhLNLdtT0hydQYdHxg+9qGa9jIOXjzcDt7ChL4TI7rGj1AC+j2yQlwAG42JxuLxrKsUGlRHAkBkpPgiKRKNR6MovAAKNboyAAMwgMCxEgwODxskJCgAlABtADWNFW1NRlJpmPEOKZMgJ8noAF0EewuDw+PBsNB6ChsPwMXSRZIxfi5ETFOpNEiFQK0j5lGb+cLsTrpHrWfQGD4Ecb5SiKWkaoVLZ7rVrbYz7SzJYxvUFXaATR60Wk9rgfmDfbH-WJA7jxfqFAx44mNpHo3xPVH4N8Diskyo-UKAwyMw7QwxgeXfgiIWwQKBMGRsHlenXdSGDUxWHLkXxBIM+N3e5B+6LgxKDZajaAGOubQPF1n6NEzo14i0rilbN9Z5MHvB6LQgkkyDYBo94JBUiRQGe+x9NFfhBV-mEfxvU4tAPZpjGPG5Tx7PtQHMbBnFfbx0XaD850UQCfVMe9zEfe8v1ADCghnPsAHlkK2ADr0KUJUMOCJoiheCBFhWVCzNLtoLnLdmSXBQSU0ckUxpYiuIXHidzoTkeT5Glqw489uMzR06BlRF3UVZVVXVNIRPnO1xOUtRPnU9iLSrIT5L7RSG2HF1jPHdjw2TQVLNE-SlMbcMCxM4tcxbSsrWEzi9KDAzGz83oK3zezTWLZtIt+ZyaH5XTrKHbN4qimhvIc4tCKSlLgrS3inUInLYtjAiqKI4KyKscyXNSsSPOHQjdLq+A2yAA) prevents inlining as well, so copying the properties may be better_

brijeshb42:
With this, I can see scenarios were the end-users final bundle has more size than before where the code could be -

```jsx
import a from './a.js';
//...
style={{[a.b]: '43px', [a.c]: '100px'}}
```

but is now -

```jsx
style={{['--accordion-panel-height']: '43px', ['--accordion-panel-width']: '100px'}}
```

Janpot:
> With this, I can see scenarios were the end-users final bundle has more size than before where the code could be

👍  and this supports exactly why this is best left as a concern for the minifier stage on the end-user build pipeline. The minifier can measure and decide whether it will be an overall optimization or not, based on the usage across the full end-user bundle.

atomiks:
@Janpot @brijeshb42 thanks for checking here.

I don't think leaving the consumer to optimize is realistic, most will not do it or know how to do it.

> With this, I can see scenarios were the end-users final bundle has more size than before where the code could be -

I can't measure an actual huge realistic app, but I checked empirically by measuring a Vite app that includes React and uses 10 Base UI components. 

All numbers below are gzip deltas (each variant was built as a published package and then bundled through Vite)

| Implementation | Vite default | Vite + Terser |
| --- | ---: | ---: |
| All enum refs, normal Babel output | baseline | baseline |
| Current hybrid `master` (needs `enumSync` tests, split source of truth) | −968 B | −943 B |
| All enum refs, checker-inlined | **−1,107 B** | **−1,072 B** |
| One `as const` object per enum | −176 B | −906 B |
| Named ESM string exports | −944 B | −990 B |

"Named ESM string export" means defining each value as an ordinary JavaScript binding:

```ts
export const checked = 'data-checked';
export const unchecked = 'data-unchecked';
```

This is similar to the `REASONS`/`reason-parts.ts` file, where we could import them as a module that remains treeshakable. Would this be a decent alternative to the enums without needing to do checker-inlining?

Janpot:
> I don't think leaving the consumer to optimize is realistic, most will not do it or know how to do it.

Yes, I think I'd like to frame this more as "optimizing for statically analyzable code, which underlying tooling will be able to pick up on more and more over time is a more durable long term solution". Or "unless there is a significant win, it may be more durable to leave this an end-user tooling concern and optimize for machine-readability".

> Would this be a decent alternative to the enums without needing to do checker-inlining?

Yes, absolutely. Interesting differences in output for ~`vite`~ `esbuild` defaults:

- [as const](https://esbuild.github.io/try/#YgAwLjI4LjEALS1taW5pZnkgLS1idW5kbGUAZQBlbnRyeS5qcwBjb25zdCBNeUVudW0gPSB7CiAgZjogJ2ZvbycKfQoKZXhwb3J0IGNvbnN0IHkgPSB7CiBbTXlFbnVtLmZdOiAnYmFyJwp9) => 44B
- [import * as MyEnum](https://esbuild.github.io/try/#YgAwLjI4LjEALS1taW5pZnkgLS1idW5kbGUAZQBlbnRyeS5qcwBpbXBvcnQgKiBhcyBNeUVudW0gZnJvbSAnLi9NeUVudW0nCgpleHBvcnQgY29uc3QgeSA9IHsKIFtNeUVudW0uZl06ICdiYXInCn0AAE15RW51bS5qcwBleHBvcnQgY29uc3QgZiA9ICdmb28n) => 30B, it inlines!
- [export * as MyEnum](https://esbuild.github.io/try/#YgAwLjI4LjEALS1taW5pZnkgLS1idW5kbGUAZQBlbnRyeS5qcwBpbXBvcnQgeyBNeUVudW0gfSBmcm9tICcuL2xpYicKCmV4cG9ydCBjb25zdCB5ID0gewogW015RW51bS5mXTogJ2JhcicKfQAATXlFbnVtLmpzAGV4cG9ydCBjb25zdCBmID0gJ2ZvbycAAGxpYi5qcwBleHBvcnQgKiBhcyBNeUVudW0gZnJvbSAnLi9NeUVudW0n) => 158B, mostly in a deduped helper and calling it for the `export *` pattern.

Surprised to find it also perfectly inlines certain patterns. I don't see a reason why it couldn't inline the `export * as ...` variant the same way at some point in the future. Could be valuable submitting a fix on their end.

A few notes:

* gzip size is a good comparison, because it can optimize inlined string constants very well.
* but, deltas from a baseline for each method may not be fully representative, 1 baseline may already be more optimized to take further duplication than another, best to compare absolute values to each other.
* we have been having [trouble](https://github.com/rolldown/rolldown/issues/7874) with `export * as` and tree-shaking before.
* One interesting optimization we could look into for our own code is to resolve away those `export * as` calls and replace with `import * as` everywhere they're used. Maybe even just resolve all re-exports altogether. Need to make sure it's side-effect free though. It should have the same effect as your PR here under vite defaults. Opening https://github.com/mui/mui-public/issues/1673


_edit: just realized vite isn't built on esbuild anymore, but ok..._


_edit 2: Interestingly, `rolldown` seems to inline both patterns, not sure why you don't see a higher delta on vite defaults with the named exports pattern._ 🤔

- [as const](https://repl.rolldown.rs/#eNptUEFOxDAM/IqVy4JUde+VOCGOvIBwSFNnN5A6VePCVlX+jpvAAtJKlhLbM/aMN+VUtylPA15aTvufVPebN8pKaiMlhuf1iZYRHmDTBOA6OLgYD5qyJk14meLMUJHrN+ilUlr3KuDezDtYRqLqeF4wN2qOIQzxk1rhOX/6o+BGp2rxY1m0wYDOEz6WNmRwcxzh8EOTTVdNAjRLKO+VcFc8eJoW7qCObEdk05ZSA8cj8BnB+YAJEga0jAP0a6n2C3MkkEBjz8Cm34cJJU7s5QD77nyvSYnDydh3c8L2LUWxVb39q1VXRY5WA04opyfrMWnVwSa3lVBZJn0ILBjGxCp/AYNvlpc=)
- [import * as MyEnum](https://repl.rolldown.rs/#eNptUEFugzAQ/MrKF9oKmTtST1WPfUHpwcA6cQprhJc2CPH3LnaaBCkSEt7dmdHMLMqqclGOWjxrDtubVHmbc9VsYz/4keEFTICP+Z2mHuzoe8h0kcasoorwHFGNp8AwwyssFcFnAmj7VUJWm1GQq6iiKnmccM3V6Luu9b+khWfd4c7Eg8vOzgItWkf4Fs+wXjz90+49CdBMXfxfCU+bPXA0TFxCktQ9stFxlUNRAB8RrOswQMAOG8YW6jlu64nZE8iHpjkCm3oTE4of2EkBkvK5IiX5BtN8mwPqU/ASKiXb7VKmaKZSLQ4o3VPjMFSqhGUVpSR0KfJ07ee2SBK7/q30n1nvM7UK90fOnWEMrNY/+SypSw==)
- [export * as MyEnum](https://repl.rolldown.rs/#eNptUcFugzAM/RUrF7apSu9IO0077gvGDgGcNh04iJitCPHvM6RQmCpFSuz4Pb33PCir0kE5KvGqOUxvUum9PqhiKuvGtwwDfPTv1NUwgm19DYk+Vi5PMsoIr/NE4Skw9PAKQ0bwGce1/UohyU0rk6Mwokq57XA8qNZXVel/SQvOutNGwIOff1JKtI7wbf5eBS2wrSYZNF013yvgaZIHjpqOU4iUukY2em4d4HgEPiNYV2GAgBUWjCXk/dzNO2ZPIAdNcQY2+UQmEN+wkwDE5XNGSvw1pvg2J9SX4MVUdLbrRU+zmEyV2KDkToXDkKkUhlGYItEtyMuaz70RKXb5W8k/sd4nE1Q2tMHdqh3oBUxYNrvsNZZCIAw/E84wBlbjH9hpvB0=)

Janpot:
[Prototyped](https://github.com/mui/mui-public/pull/1674) a version of our build tool that resolves internal `export *` to their `import *` at the source. In case you want to try if it improves on the "Named ESM string exports".

```
pnpm add https://pkg.pr.new/mui/mui-public/@mui/internal-code-infra@988e42d
```

atomiks:
I tried with named ESM string exports & the Rolldown build resolves the namespace re-exports as intended. Compared with full enums using normal Babel output in the same realistic 10-comp Vite app:

| Implementation | Vite default | Vite + Terser |
| --- | ---: | ---: |
| Full enums | baseline | baseline |
| Named exports + Rolldown | **−911 B** | **−1,067 B** |
| Checker-inlined enums | **−1,107 B** | **−1,072 B** |

So it works. But a blocker is docs generation, the current pipeline only extracts `DataAttributes` and `CssVars` from enums. Aside from that, it looks functional

Janpot:
It's only a few bytes but still interesting how the inlined variant wins compressed. Did some benchmarking of the different solutions myself, and it seems to roughly reproduce.

```
                                      realistic                             full
    strategy                    raw         gz         br          raw         gz         br
                           ────────────────────────────────   ────────────────────────────────
    baseline (enums)        249,580     81,328     68,877      451,742    146,632    120,035
    esm+rolldown+inliner     −1,282 *       +7       −113       −2,742       −746 *     −768 *
    esm+rolldown             −1,102       +206        +98       −3,477 *     −335       −338
    esm+babel                −1,047       −222 *     −206       −3,137       −603       −542 
    enum+checker-inline        −598       −145       −211 *     −1,588       −534       −609
```

2 benchmarks, a 10 BUI component page, and a page with every component. Built with vite+rolldown-vite, gzip size and brotli size. Because anyone who cares about shaving off <1kb on a 100+ kb bundle would have already set up their env with best in class minifier and best in class compression.

* baseline: the enums, built with babel
* esm+rolldown+inliner: this PR, enums replaced by const exports, built with rolldown flattening export * and a inliner on top that inline constant strings exports if they match "data-*". For this I added an inliner plugin on top of the rolldown version of our build tool (js-based, no type checker). (see https://github.com/mui/mui-public/pull/1674)
* esm+rolldown: the above, without the inliner
* esm+babel: the current PR, built with babel
* enum+checker-inline: your initial version, enums but compiled with a typescript based inliner

We see on the small app the rolldown variants are roughly neutral. the babel-built consts and the checker inliner take small gz/br wins.
For the large app it's the new rolldown based inliner that seems to take the edge.

Conclusions:
1. This is deep micro-optimizing territory, like fractions of a percent. While we see that simply using a different compression algorithm can improve over-the-wire size by >10%.
2. Moving to a real bundler allows us to add optimizations that span multiple files without relying on type checking. It can stay fully within JS.
3. erasable type syntax is the real winner IMO, it's what opens up a range of optimizations on our side, but also downstream. An inliner would be more effective if it runs on the whole end-user bundle instead of our lib only.

Janpot:
> So it works. But a blocker is docs generation, the current pipeline only extracts `DataAttributes` and `CssVars` from enums. Aside from that, it looks functional

Updating in https://github.com/mui/mui-public/pull/1713

jjenzz:
@Janpot is there anything we can do to help get https://github.com/mui/mui-public/pull/1713 merged?

Janpot:
It's reviewable and mergeable. I tried it locally on this PR and it seems to work well. If you want to double-check here, you can use https://pkg.pr.new/mui/mui-public/@mui/internal-docs-infra@a2333652

jjenzz (commented):
me:

> does [https://github.com/mui/base-ui/pull/5248](https://github.com/mui/base-ui/pull/5248) address [https://github.com/mui/base-ui/issues/5242](https://github.com/mui/base-ui/issues/5242)

llm:

> No—not at its current head. It’s enabling work, but it does not satisfy the issue’s consumer-facing requirement.
> 
> The PR replaces internal enums with typed ESM constants, such as [nestedDialogs](https://github.com/mui/base-ui/blob/85fc9c2dbdccd371043934236eaf8ec328b9e0f5/packages/react/src/dialog/popup/DialogPopupCssVars.ts#L1-L5). However:
> 
> - The [`dialog` public entry point](https://github.com/mui/base-ui/blob/85fc9c2dbdccd371043934236eaf8ec328b9e0f5/packages/react/src/dialog/index.ts#L1-L11) does not export `DialogPopupCssVars` or its data attributes.
> - The [package export map](https://github.com/mui/base-ui/blob/85fc9c2dbdccd371043934236eaf8ec328b9e0f5/packages/react/package.json#L28-L71) provides no supported metadata subpaths.
> - Therefore consumers still cannot import these constants to get the autocomplete and typo protection requested in [#5242](https://github.com/mui/base-ui/issues/5242).
> 
> To close #5242, the PR still needs a deliberate public export shape, corresponding published-package type/import tests, and consumer documentation.

atomiks:
@jjenzz was going to do exports in a new PR

jjenzz:
that's fair !

jjenzz (approved):

PR #5256, opened 2026-07-16
[composite] Publish an atomic item registry

### LINKED ISSUE #4657: [Combobox][Autocomplete] Combobox.Item/Autocomplete.Item `index` breaks hover state

### LINKED ISSUE #4698: [select] Mouse selection is a no-op in dev StrictMode

atomiks:
Fixes #4657
Fixes #4698

Supersedes #5148 and #5207. The `onItemsCommit` callback that #5227 needs will land with that fix rather than ahead of it.

## Problem

`CompositeList` committed the map, elements, labels, and indexes through separate paths: the map was built in a render memo, elements and labels were written by each item's ref callback, and the array lengths were trimmed in a later parent effect. Consumers reading `elementsRef` during `onMapChange` could see a registry that didn't match the map.

Items with an explicit `index` skipped registration entirely, so non-virtualized lists couldn't resolve them in `listRef` for pointer navigation (#4657), and the length trim wiped their slots. In StrictMode the cleanup emptied the arrays without the ref callbacks re-running to refill them (#4698).

## Changes

Registration moves to the item ref callback, and one flush per commit publishes a settled, index-addressed snapshot, so elements, labels, map, and indexes are always aligned. Explicit and automatic indexes coexist: explicit slots are reserved and automatic indexes fill the gaps.

Index guessing is opt-in per consumer, and `IndexGuessBehavior` is removed. Combobox, Menu, and Select opt in for list performance; other composites retain master's initial `-1` behavior until the commit flush. A correct guess skips the corrective re-render, while a wrong guess costs exactly what non-guessing already paid because the flush corrects it before paint.

`packages/react/src/internals/composite/list` now has 100% statement, branch, function, and line coverage. Unreachable registration and document-order branches were removed, while tests now cover detached, shadow-root, and temporarily elementless items.

> [!WARNING]
> **Breaking change (`internals`):** The internal `IndexGuessBehavior` export and `indexGuessBehavior` option are replaced by `guess`.

> [!NOTE]
> GPT-5.6 may flag “Sparse explicit indexes create dead keyboard stops” in reviews. This is intentionally ignored because master is already broken for sparse explicit indexes and the scenario is unrealistic.

> [!NOTE]
> Direct ShadowRoot-child and cross-wrapper DOM reorders remain unsupported, matching master. Wait for real usage before expanding mutation observation to those cases.

## Measurements

Production build, Chromium, 20 iterations per benchmark. Master and this branch were built and benchmarked alternately, 6 samples each; mean ± SD across samples, with Welch's t-test.

| Benchmark | master | this PR | Δ | Renders |
| Slider mount (300 instances) | 30.94 ± 1.62 ms | 19.28 ± 0.93 ms | **−37.7%** | 3 → 2 |
| Combobox type, narrows to ~11 | 9.49 ± 0.39 ms | 9.03 ± 0.59 ms | −4.8% (ns) | 17 → 15 |
| Combobox open (500 items) | 6.56 ± 0.31 ms | 6.33 ± 0.45 ms | −3.6% (ns) | 4 |
| Combobox type, 500 stay mounted | 7.86 ± 1.21 ms | 6.43 ± 0.42 ms | **−18.2%** | 11 |
| Menu open (500 items) | 17.18 ± 1.10 ms | 16.21 ± 1.35 ms | −5.6% (ns) | 12 |
| Select open (500 options) | 10.02 ± 0.56 ms | 9.37 ± 0.88 ms | −6.5% (ns) | 14 |
| Menu mount (300 instances) | 22.14 ± 2.23 ms | 22.39 ± 2.50 ms | +1.1% (ns) | 2 |
| Select mount (200 instances) | 23.67 ± 2.87 ms | 23.33 ± 3.33 ms | −1.4% (ns) | 3 |

Only Slider mount and `Combobox type, 500 stay mounted` are statistically significant; Slider and narrowing Combobox also deterministically drop from 3 → 2 and 17 → 15 renders. Menu opts into `guess` like Select to avoid a corrective popup-mount render without changing initial server markup; the pre-paint flush still corrects grouped or out-of-order guesses.

### DIFF
--- a/packages/react/src/combobox/item/ComboboxItem.test.tsx
+++ b/packages/react/src/combobox/item/ComboboxItem.test.tsx
@@ -3,6 +3,7 @@ import { expect, vi } from 'vitest';
 import { Combobox } from '@base-ui/react/combobox';
 import { fireEvent, flushMicrotasks, screen, waitFor } from '@mui/internal-test-utils';
 import { createRenderer, describeConformance, isJSDOM } from '#test-utils';
+import { useComboboxRootContext } from '../root/ComboboxRootContext';
 
 describe('<Combobox.Item />', () => {
   const { render } = createRenderer();
@@ -455,6 +456,103 @@ describe('<Combobox.Item />', () => {
       await user.click(screen.getByRole('option', { name: 'five' }));
 
       await waitFor(() => expect(input).toHaveValue('five'));
+      expect(screen.queryByRole('listbox')).toBe(null);
+    });
+
+    it('registers a rendered window at its filtered indexes', async () => {
+      function ListRefProbe() {
+        const store = useComboboxRootContext();
+        const [snapshot, setSnapshot] = React.useState('');
+
+        return (
+          <button
+            type="button"
+            data-testid="refresh-list-ref"
+            onClick={() => {
+              const list = store.state.listRef.current;
+              setSnapshot(
+                Array.from(
+                  { length: list.length },
+                  (_, index) => list[index]?.textContent ?? '',
+                ).join('|'),
+              );
+            }}
+          >
+            {snapshot}
+          </button>
+        );
+      }
+
+      function WindowedItems() {
+        const items = Combobox.useFilteredItems<string>();
+        return items.slice(2, 4).map((item) => (
+          <Combobox.Item key={item} value={item}>
+            {item}
+          </Combobox.Item>
+        ));
+      }
+
+      const { user } = await render(
+        <Combobox.Root virtualized items={['one', 'two', 'three', 'four', 'five']}>
+          <Combobox.Input data-testid="input" />
+          <ListRefProbe />
+          <Combobox.Portal>
+            <Combobox.Positioner>
+              <Combobox.Popup>
+                <Combobox.List>
+                  <WindowedItems />
+                </Combobox.List>
+              </Combobox.Popup>
+            </Combobox.Positioner>
+          </Combobox.Portal>
+        </Combobox.Root>,
+      );
+
+      await user.click(screen.getByTestId('input'));
+      await waitFor(() => expect(screen.getByRole('listbox')).not.toBe(null));
+      fireEvent.click(screen.getByTestId('refresh-list-ref'));
+
+      expect(screen.getByTestId('refresh-list-ref')).toHaveTextContent('||three|four|');
+    });
+  });
+
+  describe('virtualized with explicit index', () => {
+    function VirtualizedItems() {
+      const items = Combobox.useFilteredItems<string>();
+      return items.map((item, index) => (
+        <Combobox.Item key={item} value={item} index={index}>
+          {item}
+        </Combobox.Item>
+      ));
+    }
+
+    it('selects the highlighted filtered item with the keyboard after filtering', async () => {
+      const { user } = await render(
+        <Combobox.Root virtualized items={['one', 'two', 'three']}>
+          <Combobox.Input data-testid="input" />
+          <Combobox.Portal>
+            <Combobox.Positioner>
+              <Combobox.Popup>
+                <Combobox.List>
+                  <VirtualizedItems />
+                </Combobox.List>
+              </Combobox.Popup>
+            </Combobox.Positioner>
+          </Combobox.Portal>
+        </Combobox.Root>,
+      );
+
+      const input = screen.getByTestId('input');
+      await user.click(input);
+      await waitFor(() => expect(screen.getByRole('listbox')).not.toBe(null));
+
+      await user.type(input, 't');
+      await waitFor(() => expect(screen.queryByRole('option', { name: 'one' })).toBe(null));
+
+      await user.keyboard('{ArrowDown}');
+      await user.keyboard('{Enter}');
+
+      await waitFor(() => expect(input).toHaveValue('two'));
       expect(screen.queryByRole('listbox')).toBe(null);
     });
   });
--- a/packages/react/src/combobox/item/ComboboxItem.test.tsx
+++ b/packages/react/src/combobox/item/ComboboxItem.test.tsx
@@ -516,4 +516,47 @@ describe('<Combobox.Item />', () => {
       expect(renderSpy).not.toHaveBeenCalled();
     });
   });
+
+  describe('prop: index', () => {
+    // Regression test for https://github.com/mui/base-ui/issues/4657 — explicitly
+    // indexed items in a non-virtualized list previously skipped CompositeList
+    // registration, so hover could not resolve the item in `listRef` and
+    // `data-highlighted` stayed on the previous item.
+    it('updates the highlight on hover in a non-virtualized list', async () => {
+      const items = ['apple', 'banana', 'cherry'];
+      const { user } = await render(
+        <Combobox.Root items={items} defaultOpen autoHighlight>
+          <Combobox.Input data-testid="input" />
+          <Combobox.Portal>
+            <Combobox.Positioner>
+              <Combobox.Popup>
+                <Combobox.List>
+                  {items.map((item, index) => (
+                    <Combobox.Item key={item} value={item} index={index}>
+                      {item}
+                    </Combobox.Item>
+                  ))}
+                </Combobox.List>
+              </Combobox.Popup>
+            </Combobox.Positioner>
+          </Combobox.Portal>
+        </Combobox.Root>,
+      );
+
+      await waitFor(() => expect(screen.getByRole('listbox')).not.toBe(null));
+
+      const banana = screen.getByRole('option', { name: 'banana' });
+      await user.hover(banana);
+      await waitFor(() => {
+        expect(banana).toHaveAttribute('data-highlighted');
+      });
+
+      const cherry = screen.getByRole('option', { name: 'cherry' });
+      await user.hover(cherry);
+      await waitFor(() => {
+        expect(cherry).toHaveAttribute('data-highlighted');
+      });
+      expect(banana).not.toHaveAttribute('data-highlighted');
+    });
+  });
 });
--- a/packages/react/src/combobox/item/ComboboxItem.tsx
+++ b/packages/react/src/combobox/item/ComboboxItem.tsx
@@ -48,6 +48,7 @@ function ComboboxItemInner(props: ComboboxItemInnerProps) {
 
   const textRef = React.useRef<HTMLElement | null>(null);
   const listItem = useCompositeListItem({
+    guess: true,
     index: indexProp,
     textRef,
   });
--- a/packages/react/src/combobox/item/ComboboxItem.tsx
+++ b/packages/react/src/combobox/item/ComboboxItem.tsx
@@ -8,10 +8,7 @@ import {
   useComboboxHasItemsContext,
   useComboboxDerivedItemsContext,
 } from '../root/ComboboxRootContext';
-import {
-  useCompositeListItem,
-  IndexGuessBehavior,
-} from '../../internals/composite/list/useCompositeListItem';
+import { useCompositeListItem } from '../../internals/composite/list/useCompositeListItem';
 import type { BaseUIComponentProps, HTMLProps, NonNativeButtonProps } from '../../internals/types';
 import { useRenderElement } from '../../internals/useRenderElement';
 import { ComboboxItemContext } from './ComboboxItemContext';
@@ -53,7 +50,6 @@ function ComboboxItemInner(props: ComboboxItemInnerProps) {
   const listItem = useCompositeListItem({
     index: indexProp,
     textRef,
-    indexGuessBehavior: IndexGuessBehavior.GuessFromOrder,
   });
 
   const store = useComboboxRootContext();
--- a/packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
+++ b/packages/react/src/combobox/trigger/ComboboxTrigger.test.tsx
@@ -875,6 +875,66 @@ describe('<Combobox.Trigger />', () => {
         expect(trigger).toHaveTextContent('apricot');
       });
     });
+
+    it('only matches mounted indexed item labels when typing on the focused trigger', async () => {
+      function TestComponent() {
+        const [showBanana, setShowBanana] = React.useState(true);
+
+        return (
+          <Combobox.Root>
+            <Combobox.Trigger data-testid="trigger">
+              <Combobox.Value />
+            </Combobox.Trigger>
+            <button type="button" onClick={() => setShowBanana(false)}>
+              Remove banana
+            </button>
+            <Combobox.Portal>
+              <Combobox.Positioner>
+                <Combobox.Popup>
+                  <Combobox.List>
+                    <Combobox.Item value="apple" index={0}>
+                      Apple
+                    </Combobox.Item>
+                    {showBanana && (
+                      <Combobox.Item value="banana" index={1}>
+                        Banana
+                      </Combobox.Item>
+                    )}
+                    <Combobox.Item value="blueberry" index={2}>
+                      Blueberry
+                    </Combobox.Item>
+                  </Combobox.List>
+                </Combobox.Popup>
+              </Combobox.Positioner>
+            </Combobox.Portal>
+          </Combobox.Root>
+        );
+      }
+
+      const { user } = await render(<TestComponent />);
+      const trigger = screen.getByTestId('trigger');
+
+      await user.click(trigger);
+      await screen.findByRole('listbox');
+
+      await user.click(screen.getByRole('button', { name: 'Remove banana' }));
+      await waitFor(() => {
+        expect(screen.queryByRole('option', { name: 'Banana' })).toBe(null);
+      });
+      await user.keyboard('{Escape}');
+      await waitFor(() => {
+        expect(screen.queryByRole('listbox')).toBe(null);
+      });
+
+      await act(async () => {
+        trigger.focus();
+      });
+      await user.keyboard('b');
+
+      await waitFor(() => {
+        expect(trigger).toHaveTextContent('blueberry');
+      });
+    });
   });
 
   describe('data state attributes', () => {
--- a/packages/react/src/internals/composite/index.ts
+++ b/packages/react/src/internals/composite/index.ts
@@ -6,7 +6,6 @@ export type { CompositeListContextValue } from './list/CompositeListContext';
 export { CompositeRoot } from './root/CompositeRoot';
 export { useCompositeListItem } from './list/useCompositeListItem';
 export type { UseCompositeListItemParameters } from './list/useCompositeListItem';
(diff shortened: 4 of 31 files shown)


aarongarciah (commented):
# PR review

The composite registry rewrite itself is solid — the atomic flush design, StrictMode handling, and nested-registration ownership all check out, coverage is strong, and every benchmarked path is neutral or faster. One merge-blocking issue: removing index guessing from `OTPField.Input` regresses server-rendered HTML for all existing non-indexed usage in every React version, and the docs describe a behavior (`guess` retained) that the final code no longer has. Remaining findings are validation gaps, test gaps, and doc drift.

## Bugs (4)

### 1. 🔴 OTP field SSR markup regressed for all non-indexed inputs

**Location:** `packages/react/src/otp-field/input/OTPFieldInput.tsx:66`

```tsx
const { ref: listItemRef, index } = useCompositeListItem({ index: indexProp });
```

Master passed `IndexGuessBehavior.GuessFromOrder`, so the single SSR render pass produced correct indexes 0…N−1. This PR drops guessing without replacement, so every input without an explicit `index` prop server-renders with `index === -1` — in **all** React versions, not just 17/18 Strict Mode. The pre-hydration HTML gets: duplicate `id` on every input (`getInputId(-1)` → `${id}-0`, and `<label htmlFor>` associates with nothing), `tabIndex={-1}` on all slots (field keyboard-unreachable), `autoComplete="off"` on the first slot (loses `one-time-code` SMS autofill — which typically arrives exactly in the slow-hydration window), missing `maxLength`, wrong `enterKeyHint`, and `defaultValue` characters not rendered (`value[-1] ?? ''`). The same commit deleted the `OTPFieldRoot` SSR test "renders visible inputs with unique IDs". A dev-only React 17/18 Strict Mode hydration problem is being fixed by degrading production SSR HTML for everyone who doesn't adopt the new prop.

**Failure scenario:** Existing Next.js app upgrades with `<OTPField.Input />` × 6 (current docs-demo usage, unchanged): server HTML ships six inputs sharing one duplicate id, none focusable, no OTP autofill hint, until hydration completes.

**Fix:** `useCompositeListItem({ guess: true, index: indexProp })` — identical to master when `index` is absent (guess is deterministic in the single SSR pass), deterministic when present; `index` remains the escape hatch for React 17/18 Strict Mode hydration. This is what `ComboboxItem` does and what the intermediate commit had. Otherwise this must be an explicit breaking change, not a docs footnote.

### 2. 🟠 New `index` prop is completely unvalidated

**Location:** `packages/react/src/otp-field/input/OTPFieldInput.tsx:66`, `packages/react/src/internals/composite/list/CompositeList.tsx:227`

Duplicate, out-of-range, negative, and non-integer indexes all fail silently. Duplicates: both items survive `getCompositeListSnapshot`, both publish the same index in the map, and `syncRefs` lets the last ref attachment clobber `elementsRef.current[index]` — for OTP that means two inputs with identical `id` sharing one `slotValue`, and the root's `inputCount !== length` warning never fires because the map is element-keyed. `index >= length` leaves `enterKeyHint` never `'done'` and focus routing unable to reach the input. Negative indexes are silently dropped from the registry (`index >= 0` at `CompositeList.tsx:229`), leaving the input stuck at −1. The new docs teach hand-writing `index={0}`…`index={5}`, making copy-paste collisions likely.

**Failure scenario:** User duplicates a line and ships `index={2}` twice with `length={6}`: field renders fine, digit routing and focus silently break, no warning.

**Fix:** Dev-mode warning in `OTPFieldInput` when `indexProp` is not an integer in `[0, length)`; duplicate-explicit-index detection (dev-mode) in `CompositeList` or the OTP root.

### 3. 🟡 `nextIndexRef` reset to sparse array length breaks subsequent guesses

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:86`

```tsx
nextIndexRef.current = elementsRef.current.length;
```

With explicit indexes, `length` is max index + 1 including holes (the PR's own test creates `index={100}` → length 101), not the item count master used (`sortedMap.size`). The next `guess: true` item to mount guesses 101 instead of the next free automatic slot, so the guess is always wrong and always pays the corrective re-render the mechanism exists to avoid (one pre-paint pass with e.g. a wrong `id="${rootId}-101"` on a combobox item).

**Fix:** `nextIndexRef.current = items.length` (count of connected registrations).

### 4. 🟡 `useCompositeRoot` disabled-fallback path still mixes positions with indexes

**Location:** `packages/react/src/internals/composite/root/useCompositeRoot.ts:126`

The PR correctly fixes `activeIndex` to read `map.get(activeItem)?.index` because key position ≠ index once a list has gaps — but a few lines below, `isListIndexDisabled(sortedElements, highlightedIndex, …)` and `findNonDisabledListIndex(sortedElements, …)` still index into position-ordered `Array.from(map.keys())` with item indexes and feed a position back into `onHighlightedIndexChange` as an index. Unreachable today only because no `CompositeRoot` consumer passes explicit indexes; the file now half-supports gaps.

**Fix:** Use a dense index-addressed array (or `elementsRef.current` post-flush) in the fallback path too, or comment the density assumption.

## Tests (6)

### 1. 🟠 The core "one flush per commit" contract is not pinned

**Location:** `packages/react/src/internals/composite/list/CompositeList.test.tsx:80`

The "only publishes maps that are aligned…" test asserts only `expect(snapshots.length).toBeGreaterThan(0)`. A regression back to per-registration publishing (N `onMapChange` calls for N items — the exact pre-PR behavior this PR exists to remove) still passes, since each per-item snapshot can coincidentally stay aligned.

**Fix:** Render with `{ strict: false }` and assert `snapshots.length` is exactly 1 after mount (and exactly 2 after a single item add), keeping the alignment check.

### 2. 🟡 Menu `guess: true` opt-in has zero menu-level test coverage

**Location:** `packages/react/src/menu/item/MenuItem.tsx:33` (plus checkbox-item, radio-item, link-item, submenu-trigger)

Five menu source files switch from no-guess to `guess: true` with no menu test additions. Grouped menus (`Menu.Group` wrappers) are exactly where the flat registration-order guess is wrong and must be corrected by the pre-paint flush; no menu test arrow-navigates or typeaheads across grouped items. Menus also newly inherit the React 18 Strict Mode guess-consumption quirk (discarded first render consumes a guess) that master never exposed them to.

**Failure scenario:** A menu-specific guess-correction bug (wrong highlight order or typeahead target in grouped menus) ships with a green suite.

**Fix:** One `MenuRoot`/`MenuGroup` test with grouped items asserting ArrowDown traversal order and typeahead label matching.

### 3. 🟡 `activeIndex` map-value fix has no test for its motivating scenario

**Location:** `packages/react/src/internals/composite/root/useCompositeRoot.ts:122`

Nothing in `CompositeRoot.test.tsx` renders explicit-index items, so reverting `map.get(activeItem)?.index` back to `sortedElements.indexOf(activeItem)` fails no test.

**Fix:** CompositeRoot test with an explicitly indexed `ACTIVE_COMPOSITE_ITEM` item leaving a gap, asserting the default highlighted index.

### 4. 🟡 Dynamic re-registration paths untested

**Location:** `packages/react/src/internals/composite/list/CompositeList.test.tsx:386`, `:833`

Two gaps in the riskiest mechanism of the rewrite (identity-sensitive ref callback re-running unregister/register): (a) explicit→automatic transition — removing the `index` prop after mount flips the item to guess/subscription mode; the existing test only covers explicit 0→2, so an item stuck at −1 is unpinned; (b) updating `label` (or `metadata` identity) on a mounted non-nested item — only initial mount is asserted, so a dropped republish would pass.

**Fix:** Extend the re-register test with `setProps({ index: undefined })`; add a test updating `label` on a mounted item asserting `labelsRef`/`onMapChange` republish.

### 5. 🟡 New OTP `index` prop tested only via SSR attribute snapshot

**Location:** `packages/react/src/otp-field/input/OTPFieldInput.test.tsx:76`

No interaction test with explicitly indexed inputs. Typing, arrow-key focus movement, and paste all flow through composite `listRef` slots, so an off-by-one in explicit registration renders fine but breaks input routing.

**Fix:** One interaction test on explicitly indexed inputs (type full code, assert focus advance and value distribution); pin out-of-range behavior alongside the fix for Bugs #2.

### 6. ℹ️ Minor test hygiene

`CompositeList.test.tsx:912` double-wraps StrictMode (`createRenderer` already defaults `strict: true`) and awaits sync `renderToString`/`hydrate`; the windowed `ComboboxItem.test.tsx:462` test asserts internal `store.state.listRef` plumbing rather than user-visible highlight — worth a behavioral companion assertion.

## Simplifications (4)

### 1. 🟡 `syncRefs` reaches back into the live map for data the snapshot already passed through

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:78`

```tsx
const registration = map.get(item.element)!;
```

The snapshot item carries `metadata` but not `label`/`textRef`, forcing a live-map lookup with a non-null assertion whose safety depends on "any map mutation dirties before the next resync" — fragile because the first layout effect replays `syncRefs(itemsRef.current)` against a stale snapshot while reading the live map.

**Fix:** Have `getCompositeListSnapshot` carry the whole `registration` on each item; `syncRefs` reads `item.registration.label`/`.textRef`/`.metadata` with no live-map coupling and no `!`.

### 2. 🟡 Explicit-index items now pay a registry cost that was zero on master — unbenchmarked

**Location:** `packages/react/src/internals/composite/list/useCompositeListItem.ts:68`

Master's registration effect returned early for `externalIndex != null`; items wrote `elementsRef.current[index]` directly with zero `CompositeList` work. Now every mount/unmount of an explicitly indexed item triggers a parent render + full flush (snapshot, map rebuild with per-item spread, listener walk, `onMapChange` → e.g. `SelectPositioner` `findItemIndex` scans). In a non-virtualized windowed list this fires per scroll commit. Bounded (mounted-row count is small, no DOM-position sort for explicit items), but it is the one hot path whose cost went from zero to nonzero and no benchmark covers it.

**Fix:** Add a scrolling windowed-list benchmark to confirm or dismiss; consider skipping the flush when only explicit-index registrations changed and their slots are unchanged.

### 3. ℹ️ `scheduleMapUpdate` is a per-render closure captured by long-lived callbacks

**Location:** `packages/react/src/internals/composite/list/CompositeList.tsx:40`

Captured by MutationObserver callbacks and called from stable handlers; safe today only because its body touches nothing but stable values. `useStableCallback` (matching adjacent `register`/`unregister`/`flush`) removes the stale-closure reasoning for free.

### 4. ℹ️ Published map could be typed and compared tighter

`syncRefs` provably always publishes non-null entries with numeric `index`, yet the map is typed `CompositeMetadata<Metadata> | null`, forcing downstream guard chains (`map.get(activeItem)?.index ?? -1`). Separately, `flush` publishes a fresh `Map` identity with no equality bail, so root-level `setState` consumers (Tabs/Slider/Toolbar) re-render on every flush even when contents are identical — inherited from master, but a shallow-compare bail would cap the blast radius. Both reasonable to defer.

## Docs (3)

### 1. 🟠 "React 19 can infer the index safely" is false for the shipped code

**Location:** `docs/src/app/(docs)/react/components/otp-field/page.mdx:17`, `packages/react/src/otp-field/input/OTPFieldInput.tsx:354` (JSDoc, propagated to `types.md:157`)

> Set this prop on each input when using server-side rendering to ensure deterministic slot attributes during hydration in React 17 and 18 Strict Mode. React 19 can infer the index safely in simple hydration cases.

This describes the intermediate design that kept `guess: true`. In the shipped code nothing infers anything in any React version: server markup is −1 everywhere without `index`, and conversely there is no 17/18 Strict Mode hydration mismatch either (both sides deterministically render −1). React 19 users will read this as "I can skip `index`" and ship the degraded SSR HTML from Bugs #1.

**Fix:** Resolve with Bugs #1 — either restore `guess` (making the sentence true) or rewrite prose + JSDoc to the actual contract ("without `index`, server-rendered slot attributes are placeholders until hydration; set `index` whenever server markup matters"). Re-run `pnpm docs:api` after rewording.

### 2. 🟡 Docs snippets and runnable demos teach contradictory usage

**Location:** `docs/src/app/(docs)/react/components/otp-field/page.mdx:43`, `docs/src/app/(docs)/react/components/otp-field/demos/**`

Every mdx snippet now hardcodes `index={0}`…`index={5}` on all six inputs, while every runnable demo on the same page still omits `index` (each already maps over an `index` variable — passing it is a one-liner). Copy-pasters from demos get the SSR footgun; snippet readers get 6 lines of boilerplate. Bonus: `index={1}` sits beside `aria-label="Character 2 of 6"`, inviting off-by-one edits.

**Fix:** Pick one story once Bugs #1 is resolved; if `index` stays recommended for SSR, thread it through the demos' existing map variable.

### 3. 🟡 Load-bearing internals lack (or have stale) comments

Grouped: `useCompositeListItem.ts:34` says "Guess the index from the registration order" but the guess is consumed during render (param JSDoc correctly says "render order"); `CompositeItem.tsx:32`'s `[compositeRef, ...refs]` reorder decides nested-registration ownership yet reads as an arbitrary swap someone could revert — needs one line; `CompositeList.tsx:155`'s `if (!isDirtyRef.current) syncRefs(itemsRef.current)` branch looks dead on first mount but exists for ref-object replacement and Strict Mode replay — two tests target it, zero comments explain it; and the ref callback's referential-stability requirement on `metadata`/`label`/`textRef` (unmemoized `metadata` now cycles DOM ref detach/attach every render, where master only re-ran an effect) is enforced nowhere — state it in the param JSDoc.

## Verdict

**Request changes** — the OTP field SSR regression (Bugs #1) degrades production server markup for all existing non-indexed usage and contradicts the PR's own documentation.

---

🤖 Review generated with Claude Code (Fable 5 high effort)

atomiks:
> 1. 🔴 OTP field SSR markup regressed for all non-indexed inputs

I don't know why this returned 🤦‍♂️ already saw this issue before and fixed it previously. Will check the others

Edit: seems Codex broke and was starting all new chats without a worktree, so it pulled in changes that were meant to be part of a new branch here. Removing all of that as it's unrelated (though I found that bug from here).

atomiks:
@aarongarciah

## Bugs

> 1. 🔴 OTP field SSR markup regressed for all non-indexed inputs

Fixed.

> 2. 🟠 New `index` prop is completely unvalidated

Not applicable: the OTP Field `index` prop was removed from this PR.

> 3. 🟡 `nextIndexRef` reset to sparse array length breaks subsequent guesses

Not addressed: sparse explicit indexes are intentionally out of scope, as documented in the PR body.

> 4. 🟡 `useCompositeRoot` disabled-fallback path still mixes positions with indexes

Not addressed: this is only reachable with sparse explicit indexes, which are intentionally out of scope.

## Tests

> 1. 🟠 The core "one flush per commit" contract is not pinned

Fixed.

> 2. 🟡 Menu `guess: true` opt-in has zero menu-level test coverage

Fixed.

> 3. 🟡 `activeIndex` map-value fix has no test for its motivating scenario

Not addressed as written: the position/index mismatch requires sparse indexes; dense controlled-index coverage was added instead.

> 4. 🟡 Dynamic re-registration paths untested

Fixed.

> 5. 🟡 New OTP `index` prop tested only via SSR attribute snapshot

Not applicable: the OTP Field `index` prop and its related tests were removed from this PR.

> 6. ℹ️ Minor test hygiene

Fixed.

## Simplifications

> 1. 🟡 `syncRefs` reaches back into the live map for data the snapshot already passed through

Fixed.

> 2. 🟡 Explicit-index items now pay a registry cost that was zero on master — unbenchmarked

Benchmarked in production Chromium: a 20-row window shift averaged 1.00 ± 0.06 ms on this PR versus 1.05 ± 0.05 ms on master across six alternating 20-iteration samples (−4.6%, not significant), with five renders on both. No measurable regression.

> 3. ℹ️ `scheduleMapUpdate` is a per-render closure captured by long-lived callbacks

Fixed.

> 4. ℹ️ Published map could be typed and compared tighter

Not addressed: this is an informational follow-up and is unrelated to the behavior fixed by this PR.

## Docs

> 1. 🟠 "React 19 can infer the index safely" is false for the shipped code

Fixed.

> 2. 🟡 Docs snippets and runnable demos teach contradictory usage

Fixed.

> 3. 🟡 Load-bearing internals lack (or have stale) comments

Fixed.

jjenzz (approved):
thanks for picking this up, glad we got there in the end 🎉

PR #5259, opened 2026-07-17
[scroll area] Prevent scroll snapping while dragging the thumb

### LINKED ISSUE #5209: ScrollArea with snap cannot scroll via mouse dragging smoothly
# Bug report

## Current behavior

A native (horizontal) scrollable div with scroll snapping can "smoothly" scroll via both the mouse/dragging & keyboard, and then snap to the nearest item when scrolling has stopped. In Base UI's ScrollArea, keyboard snapping works as expected, but the mouse drag scrolling forcibly snaps as demonstrated below.

https://github.com/user-attachments/assets/f1090e79-6c26-41e7-a0b8-1f0985ec1399

## Expected behavior

The mouse drag scrolling in the ScrollArea should behave like the native one, i.e. does not snap while dragging.

## Reproducible example

https://stackblitz.com/edit/esxr37a1?file=src%2FApp.tsx

## Base UI version

v1.6.0

Zen v1.21.5b (Firefox 152.0.4) 

## Which OS are you using?

NixOS

atomiks:
Closes #5209

When the viewport has CSS `scroll-snap-type`, dragging the scrollbar thumb jerks between snap points instead of scrolling smoothly. The drag handler assigns `scrollTop`/`scrollLeft` on every pointer move, and the CSS scroll snap spec forces programmatic scrolls to land on a snap point, so each move snaps. Native scrollbars suppress snapping while dragging and only snap on release.

Assigning exact pixel values to `scrollLeft` on a plain container with `scroll-snap-type: x mandatory` and snap points every 200px shows the quantization that makes the thumb unable to sit under the pointer:

| assigned | landed (`x mandatory`) | landed (`none`) |
| --- | --- | --- |
| 20 | 0 | 20 |
| 90 | 0 | 90 |
| 150 | 200 | 150 |
| 185 | 200 | 185 |

The root pointer handlers now save the viewport's inline `scroll-snap-type` and set it to `none` when a drag starts, restoring it on release. Restoring the value makes the browser re-snap to the nearest snap point, matching native behavior. The save is guarded so a second pointer during an active drag can't clobber the saved value with `none`. Track presses need the disable to run before the jump-to-click assignment as well — otherwise the initial jump quantizes to a snap point and the thumb stays offset from the pointer for the whole drag — so the guarded disable is shared with the scrollbar's `onPointerDown` and runs ahead of the jump.

[Experiment](https://deploy-preview-5259--base-ui.netlify.app/experiments/scroll-area/scroll-area-snap) puts a scroll area next to a native scroller with the same snap points for comparison.

> [!NOTE]
> The snap on release is instant rather than animated: restoring `scroll-snap-type` triggers a re-snap, which the spec requires to be non-animated. Native gestures animate that settle. Animating it is feasible as a follow-up without reimplementing snap resolution, because the browser can compute the target: restoring the snap type re-snaps synchronously (verified in Chromium), so reading the scroll position immediately afterwards yields the resolved snap target — the position can then be reverted and smooth-scrolled to that target, restoring the snap type on `scrollend`. The remaining cost is edge handling: a `prefers-reduced-motion` branch, a `scrollend` fallback for older Safari, and cancellation when a new drag or wheel interrupts the settle. If a browser instead defers the re-snap to the next frame, the read returns the unchanged position and the behavior degrades to the instant snap shipped here. The continuous tracking during the drag, which is what the issue reports, is fixed either way.


flaviendelangle (approved):
# PR review

The change disables the viewport's inline `scroll-snap-type` on drag start and restores it on release so thumb dragging tracks the pointer continuously instead of quantizing to snap points, matching native scrollbars. The implementation is sound: the save is guarded against multi-touch clobbering, disable is correctly ordered ahead of the track's jump-to-click assignment, and restore is reachable on every real release path (thumb pointer-capture delivers `pointerup`/`pointercancel`, with the scrollbar's own handlers as a fallback). Nothing is merge-blocking. Note: automated tests could not be run here (repo requires Node ≥22.22.3; environment has 22.19.0), so verification is by code analysis.

## Bugs (0)

No findings.

## Tests (1)

### 1. ℹ️ Restore path is only asserted with an inline `scroll-snap-type`, not a stylesheet-defined one

**Location:** `packages/react/src/scroll-area/thumb/ScrollAreaThumb.test.tsx:258`

```tsx
style={{ width: '100%', height: '100%', scrollSnapType: 'y mandatory' }}
```

All three unit tests set `scrollSnapType` inline, so `viewportEl.style.scrollSnapType` reads back the literal value, is saved, and is restored to that same literal. The common real-world case is a CSS class (as in the new experiment's `.ScrollAreaViewport`), where `style.scrollSnapType` is `''`; the code then saves `''`, sets inline `none`, and restores `''` to let the class value re-apply. That restore-to-empty path — the one that most consumers hit — isn't pinned by any unit test (only exercised by the manual experiment).

**Failure scenario:** A future refactor that changed the sentinel or wrote a concrete value on restore instead of `''` would silently break class-based snap (leaving `scrollSnapType: none` on the element) while all unit tests stayed green.

**Fix:** Optionally add a case rendering the viewport with a class-provided snap type and assert `viewport.style.scrollSnapType === 'none'` during drag and `=== ''` after release. jsdom can verify the inline-style bookkeeping even though it can't verify actual re-snapping.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Approve** — a focused, well-guarded fix with symmetric disable/restore and correct ordering; the only note is an optional test for the stylesheet-class restore path.

---

atomiks:
> [!NOTE]
> There is a theoretical edge case if the scrollbar thumb unmounts during an active drag—for example, because an external content or layout update removes overflow while the pointer is held down. The captured element is then removed before its `pointerup`/`pointercancel` handler can restore the viewport's original `scroll-snap-type`, so snapping could remain disabled until another drag completes.
>
> This requires an unusually timed external update, and handling it would add extra unmount/lifecycle coordination. We should wait for a real user report before adding that complexity.

PR #5260, opened 2026-07-17
[otp field] Add Input index prop

atomiks:
React 17 and 18 Strict Mode hydration can infer different OTP slot indexes between the server and client and log an attribute mismatch. This adds an opt-in index escape hatch for deterministic initial slot markup.

## Changes

- Added `index` to `OTPField.Input` while retaining inferred indexes by default.
- Added SSR hydration coverage for stable slot attributes and post-hydration keyboard navigation.
- Documented the escape hatch in the generated API table.


atomiks:
> [!NOTE]
> Closing this because the hydration mismatch is limited to React 17 and 18 development Strict Mode and corrects itself after hydration. Consumers who only need to suppress the console warning can already pass `suppressHydrationWarning` to each `<OTPField.Input>`, so adding a permanent `index` prop is unnecessary.

PR #5261, opened 2026-07-17
[toast] Fix remaining timer calculation in ToastStore

sai6855:


packages/react/src/toast/store.ts
sai6855, line 399: Delay actually tracks entire duration, instead of leftover duration. remaining is the one which have leftover duration

packages/react/src/toast/store.ts
```diff
@@ -360,7 +396,7 @@ export class ToastStore extends ReactStore<State, {}, typeof selectors> {
       if (timer.timeout) {
         timer.timeout.clear();
         const elapsed = Date.now() - timer.start;
-        const remaining = timer.delay - elapsed;
+        const remaining = timer.remaining - elapsed;
```
sai6855, line 363: ## Why the toast never dismisses

`timer.delay` holds the **entire** duration, while `timer.remaining` holds the **time left**. `pauseTimers` was calculating the remaining time from `timer.delay` instead of `timer.remaining`, so repeated hover/leave cycles kept restoring almost the full duration.

**Walkthrough** (`delay = 10s`):

| Step | Action | Elapsed this segment | Correct remaining | Buggy code gives |
| ---- | ------ | -------------------: | ----------------: | ---------------: |
| 1 | Enter after 1s, leave after 1s | 1s | 10 − 1 = 9s | 10 − 1 = 9s ✓ |
| 2 | Enter after 1s, leave after 1s | 1s | 9 − 1 = 8s | 10 − 1 = 9s ✗ |

At step 2, the remaining time should be `10 − 1 − 1 = 8s`, but the code calculated `delay − elapsed = 10 − 1 = 9s`, ignoring the 1 second already consumed in step 1. Every hover/leave cycle resets the remaining time toward the full duration, so the toast never expires.

**Fix:** On resume, sync `timer.delay = timer.remaining` so the next pause computes `delay − elapsed` against the actual remaining time.

---

atomiks (approved):

PR #5262, opened 2026-07-17
[docs][menu] Use align="start" in demos

aarongarciah:
## What changed

Adds `align="start"` to `Menu.Positioner` in all standalone menu demos, except the [arrow](https://deploy-preview-5262--base-ui.netlify.app/react/components/menu#arrow) one.

## Why

With the default `align="center"`, the popup's appear animation (`scale(0.98)` from `var(--transform-origin)`) blooms out symmetrically from behind the trigger with neither popup edge anchored to it, which looks off — most visibly in `detached-triggers-simple`, where the trigger is a small icon button. Start alignment matches menu conventions and makes the animation grow out from the trigger's corner.

## Notes for review

- Demos that already set an explicit `align` (e.g. `context-menu/with-menu`, `autocomplete/async`/`grid`, `combobox/input-inside-popup`) are untouched.
- Other popup demos were audited and intentionally left alone: popover/preview-card/tooltip/navigation-menu demos render an `Arrow` (centering is intentional), combobox/autocomplete popups match the anchor width (`--anchor-width`) so `align` has no visual effect, and context menu/menubar/nested submenus already default to `start` via `MenuPositioner`.

|before|after|
|---|---|
|image|image|

### DIFF
--- a/docs/src/app/(docs)/react/components/menu/demos/checkbox-items/css-modules/index.tsx
+++ b/docs/src/app/(docs)/react/components/menu/demos/checkbox-items/css-modules/index.tsx
@@ -14,7 +14,7 @@ export default function ExampleMenu() {
         Workspace <CaretDownIcon />
       </Menu.Trigger>
       <Menu.Portal>
-        <Menu.Positioner className={styles.Positioner} sideOffset={8}>
+        <Menu.Positioner className={styles.Positioner} sideOffset={8} align="start">
           <Menu.Popup className={styles.Popup}>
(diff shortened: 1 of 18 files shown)


colmtuite:
Why just Menu? Same applies to Popover too

colmtuite (approved):

aarongarciah:
> Why just Menu? Same applies to Popover too

@colmtuite Popover demos use arrows. They probably look better centered?

image

PR #5283, opened 2026-07-20
[form][field] Ignore controls in disabled native fieldsets

atomiks:
Native disabled fieldsets were not reflected in Base UI's field registry for composite controls whose visible element differs from the hidden native input. This could run validation, block submission, and include a value in `onFormSubmit` even though the browser omitted the control.

## Changes

- Read effective disabled state from registered native inputs while preserving visible control refs.
- Skip disabled fields during validation, invalid focus, and Form value collection.
- Add Chromium regression coverage for Checkbox and Switch, including grouped and Radio behavior.

### DIFF
--- a/packages/react/src/checkbox-group/CheckboxGroup.test.tsx
+++ b/packages/react/src/checkbox-group/CheckboxGroup.test.tsx
@@ -1180,6 +1180,36 @@ describe('<CheckboxGroup />', () => {
       expect(validate.mock.lastCall?.[1].fruits).toEqual(['apple']);
       expect(handleSubmit.mock.lastCall?.[0].fruits).toEqual(['apple']);
     });
+
+    it.skipIf(isJSDOM)(
+      'omits a group disabled by a native fieldset from validation and submission',
+      () => {
+        const handleSubmit = vi.fn();
+        const validate = vi.fn(() => 'invalid');
+
+        render(
+          <Form onFormSubmit={handleSubmit} data-testid="form">
+            <fieldset disabled>
+              <Field.Root name="fruits" validate={validate}>
+                <CheckboxGroup defaultValue={['apple']}>
+                  <Checkbox.Root value="apple" />
+                  <Checkbox.Root value="banana" />
+                </CheckboxGroup>
+              </Field.Root>
+            </fieldset>
+            <button type="submit">Submit</button>
+          </Form>,
+        );
+
+        const form = screen.getByTestId('form') as HTMLFormElement;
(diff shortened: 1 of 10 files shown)


flaviendelangle (approved):
# PR review

This PR makes Base UI's field registry honor native `<fieldset disabled>` for composite controls (checkbox/switch/radio) whose visible element differs from the hidden native input, so those fields are correctly skipped during validation, invalid-focus, and form-value collection — matching native form semantics. The implementation is clean and well-targeted; both branches of the new `isDisabled` helper (the `registeredInputs` group path via Checkbox/Radio and the `inputRef` fallback path via Switch) are exercised by the added Chromium tests. I found no merge-blocking issues. Note: I could not run the test suite or `tsc` locally because the environment's Node (v22.19.0) is below the repo's required `>=22.22.3`, so verification is by code reading only.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (1)

### 1. ℹ️ `formValues` passed to custom `validate` now omits natively-disabled fields

**Location:** `packages/react/src/field/root/useFieldValidation.ts:254`

```ts
const formValues = Array.from(formRef.current.fields.values()).reduce((acc, field) => {
  if (field.name && !field.isDisabled()) {
    acc[field.name] = field.getValue();
  }
  return acc;
}, {} as Form.Values);
```

The second argument handed to a user-supplied `validate(value, formValues)` now excludes fields disabled by an ancestor fieldset, mirroring the `onFormSubmit` payload change in `Form.tsx:133`. This is a defensible, self-consistent behavior change (the values a validator sees now match what gets submitted), but it is a subtle, user-observable shift for cross-field validation: a validator that reads another field's value while that field sits in a temporarily-disabled fieldset will now see `undefined` instead of the prior value.

**Failure scenario:** A "confirm" field whose validator compares against a `password` field grouped in a fieldset that is disabled at read time would previously receive the password value in `formValues` and now receives `undefined` — no crash, but different validation output. Worth a one-line mention in the Form/Field validation docs so the behavior is discoverable.

**Fix:** Optional — add a sentence to the public validation docs (or the `validate` JSDoc) noting that natively-disabled fields are excluded from both the submitted values and the `formValues` passed to `validate`, consistent with native form submission.

## Verdict

**Approve** - correct, native-aligned fix with matching coverage on both `isDisabled` branches; the only note is an optional docs mention of the (intentional) `formValues` exclusion, and local test/type verification was blocked by a Node version mismatch.

---

PR #5287, opened 2026-07-20
[form] Focus the first invalid field in document order

### LINKED ISSUE #5284: [form] Focus can land on the wrong invalid field on submit
# Bug report

## Current behavior

After an invalid form submission, `<Form>` moves focus to the first invalid field so the user can find and fix it. It selects that field with `invalidFields[0]`, where `invalidFields` is built by iterating the internal `formRef.current.fields` **Map in insertion order**.

That insertion order is assumed to match document order, but it drifts. A field control re-registers whenever its value changes — the registration effect (`useRegisterFieldControl`) includes the control's `value` in its dependency array — and re-registration **deletes and re-adds** the field's entry in the `fields` Map (`useFieldControlRegistration`: `fields.delete(id)` followed by `fields.set(id, …)`). In a JS `Map`, a `delete` + `set` moves the key to the **end**.

So as soon as a user interacts with a control (e.g. checks then unchecks a checkbox), that field jumps to the back of the Map. On the next invalid submit, `invalidFields[0]` is no longer the first invalid field **in the document** — focus can land on a field that appears *later* in the DOM, skipping an earlier invalid one.

Text `<input>`s hide the problem because they don't re-register on input. It shows up with controls that re-register on change, i.e. **checkbox and radio**.

Minimal sequence (two required checkboxes, A above B in the DOM, both empty):

1. Submit → focus correctly goes to **A**.
2. Check A, then uncheck A (any toggle).
3. Submit again → focus goes to **B**, even though **A** is still the first invalid field in the document.

## Expected behavior

After an invalid submit, focus should always move to the first invalid field **in document order**, regardless of whether the user has previously interacted with (toggled) any control. In the sequence above, the second submit should return focus to **A**.

## Reproducible example

```tsx
import { Form } from "@base-ui/react/form"
import { Field } from "@base-ui/react/field"
import { Checkbox } from "@base-ui/react/checkbox"

  return (
    <Form onFormSubmit={() => {}}>
      <Field.Root name="a">
        <Checkbox.Root required style={{ width: 20, height: 20, border: "1px solid" }}>
          <Checkbox.Indicator />
        </Checkbox.Root>
        <Field.Label>Checkbox A (first in DOM)</Field.Label>
        <Field.Error />
      </Field.Root>

      <Field.Root name="b">
        <Checkbox.Root required style={{ width: 20, height: 20, border: "1px solid" }}>
          <Checkbox.Indicator />
        </Checkbox.Root>
        <Field.Label>Checkbox B (second in DOM)</Field.Label>
        <Field.Error />
      </Field.Root>

      <button type="submit">Submit</button>
    </Form>
  )
}
```

Steps: click **Submit** (focus lands on A) → click A twice to check then uncheck it → click **Submit** again. Observed: focus moves to **B**. Expected: focus stays on **A**. (Watch `document.activeElement`, or give A/B different visible focus styles.)

## Base UI version

v1.4.1

Chrome 

## Which OS are you using?

Windows 11 

None required to observe the bug — it's a focus-order defect. It does, however, directly degrade the experience for screen-reader and keyboard users, who rely on focus landing on the first error.

## Additional context

Root cause appears to be that `invalidFields` is ordered by Map insertion order rather than document order, and that order is disturbed by control re-registration on value change (`useRegisterFieldControl` depends on `value`; `useFieldControlRegistration` does `fields.delete(id)` then `fields.set(id, …)`, moving the entry to the end of the Map).

**Recommended fix — select by document order.** Where the first invalid field is picked, sort the invalid fields' controls by `compareDocumentPosition` and focus the first, instead of trusting Map iteration order. This is the correct semantic for "focus the first error" and is robust regardless of the internal Map's ordering.

The change I applied and verified locally, at the point where the invalid control is selected. Base UI ships `Form` in two builds, so the same edit is made in both:

- `form/Form.js` (CommonJS build)
- `esm/form/Form.js` (ES-module build)

```diff
- focusControl(invalidFields[0].controlRef.current);
+ focusControl(
+   invalidFields
+     .map((field) => field.controlRef.current)
+     .filter(Boolean)
+     .sort((a, b) =>
+       a.compareDocumentPosition(b) & Node.DOCUMENT_POSITION_FOLLOWING ? -1 : 1,
+     )[0],
+ );
```

This selection runs on both the submit path and the external-`errors` path, so both should be updated. It resolves the repro above with no regressions.

**Alternative — preserve order at the source.** Update the existing Map entry in place on re-registration instead of `fields.delete(id)` + `fields.set(id, …)`, so insertion order keeps matching mount order. This fixes the observed case too, but still relies on Map order equalling document order in all cases, which selecting by `compareDocumentPosition` does not.

atomiks:
Closes #5284

After an invalid submit, `Form` focuses the first invalid field by iterating the internal fields Map in insertion order. That order drifts: a control re-registers whenever its value changes, and the deregister/re-register cycle deletes and re-adds the Map entry, moving the field to the end. Toggling a required checkbox and submitting again then focuses a later field instead of the first invalid one in the DOM, which degrades the experience for keyboard and screen reader users relying on focus landing on the first error.

Two changes:

- Value changes no longer deregister and re-register the field control; re-registration updates the existing Map entry in place, keeping the registry stable. Deregistration now happens only on unmount or when registration is disabled. This alone fixes the reported sequence.
- `focusFirstInvalid` selects the invalid control by document position rather than registration order, since registration order can also diverge from DOM order without any re-registration (keyed fields reordered without remounting, portals). The comparison applies only when the controls are in the same tree; for disconnected controls (e.g. separate shadow roots), where `compareDocumentPosition` is implementation-specific, the now-stable registration order is kept.

The existing contract is preserved: submission stays blocked when a field is invalid without a focusable control. The keyed-reorder regression test renders without Strict Mode, which re-runs effects when a keyed subtree moves and would re-register the fields in DOM order, masking the divergence the test pins.


flaviendelangle (approved):
# PR review

This PR fixes invalid-field focus drifting away from the first DOM error by (1) making field re-registration update the `fields` Map entry in place instead of delete+re-add, and (2) selecting the invalid control by document position rather than registration order. The logic is correct, the DOM-position comparison uses the right `compareDocumentPosition` argument order, and the disconnected-tree fallback is genuinely exercised by the shadow-root test. Nothing merge-blocking; I found only two minor observations.

## Bugs (0)

No findings.

The two behavioral changes hold up under scrutiny:
- `comesBeforeInSameTree(control, firstControl)` returns true exactly when `control` precedes `firstControl` in the same tree (`element.compareDocumentPosition(reference) & FOLLOWING` means `reference` follows `element`), so the loop keeps the DOM-earliest control. The `DOCUMENT_POSITION_DISCONNECTED` guard correctly falls back to registration order for separate roots — verified by the shadow-root test, which would focus `b` instead of `a` if the disconnected branch were mishandled.
- Splitting `useRegisterFieldControl` into a register/re-register effect plus an unmount-only cleanup effect preserves every prior invariant: disable still deregisters (`registerFieldControl(source, undefined)`), unmount still deregisters (stable `sourceRef`/`registerFieldControl` deps), and `id` changes still delete the previous entry inside `register`. The removed per-render cleanup only elided a transient deregister that was immediately overwritten.

## Tests (0)

No findings.

Coverage is strong: value-change stability, keyed-reorder DOM ordering (correctly rendered with `strict: false`, since Strict Mode's effect re-runs would re-register in DOM order and mask the divergence), and the disconnected-tree fallback each pin a distinct branch.

## Simplifications (0)

No findings.

`comesBeforeInSameTree` overlaps with `sortByDocumentPosition` in `CompositeList.tsx`, but the added `DOCUMENT_POSITION_DISCONNECTED` guard is a real semantic difference, so keeping it local is reasonable rather than duplication worth consolidating.

## Docs (0)

No findings.

## Verdict

**Approve** — correct, well-tested fix with no blocking or non-blocking issues; the only NumberField test failure is a pre-existing ICU/locale issue on `master`, unrelated to this diff.

Two optional notes for the maintainer:
- ℹ️ `Form.test.tsx` — the two new `render(<App />, ...)` / `render(<Form>...)` calls (value-change and keyed-reorder tests) omit `await`, while the rest of the file (including the line changed at ~47) uses `await render(...)`. Harmless since `render` returns a synchronously-usable result, but inconsistent.
- ℹ️ `Form.tsx:63-64` — if a still-registered field's `controlRef` ever pointed at a node detached from the document while later invalid controls were connected, the disconnected guard would keep the detached control and `.focus()` would silently no-op. Not reachable through normal mount/unmount (unmount deregisters), so no action needed — just the one edge the `compareDocumentPosition` fallback doesn't rank.

---

PR #5292, opened 2026-07-21
[popups] Focus keyboard-open item synchronously

### LINKED ISSUE #5268: [floating-ui-react] enqueueFocus module-level rafId lets concurrent callers cancel each other — keyboard-open focus hop into a roving-focus listbox is racy
# [floating-ui-react] `enqueueFocus` module-level `rafId` lets concurrent callers cancel each other — keyboard-open focus hop into a roving-focus listbox is racy

## Environment

- `@base-ui/react` 1.6.0 (vendored `floating-ui-react` utils)
- React 18.3.1 / react-dom 18.3.1
- Chromium via Playwright (`@storybook/test-runner`); observed against a production Storybook build

## Summary

`floating-ui-react/utils/enqueueFocus` keeps a single **module-level** rAF handle, and its first statement cancels whatever focus rAF is pending — regardless of which caller scheduled it:

```js
let rafId = 0;
export function enqueueFocus(el, options = {}) {
  // ...
  cancelAnimationFrame(rafId);   // cancels ANY pending caller's focus
  // ...
  const currentRafId = requestAnimationFrame(exec);
  rafId = currentRafId;
  // ...
}
```

On keyboard-open of a roving-focus popup (measured on `Select`, ArrowDown to open), two callers race:

1. `focusItem` enqueues focus into the highlighted option;
2. `FloatingFocusManager`'s initial-focus microtask calls `enqueueFocus` a moment later — **cancelling the option-focus rAF** — and re-issues focus through its own path.

Focus does settle on the option ~2–6 ms later, so real users are effectively unaffected. But there is a measurable window in which `data-highlighted` is already set (it lands synchronously with React state, ~1 ms) while DOM focus is still on the **trigger**.

## Impact

Automated keyboard tests that chain keystrokes flake with a misleading signature: a keystroke fired inside that window hits the still-focused trigger, so listbox navigation goes dead and the highlighted attribute never advances — **no finite `waitFor` on `data-highlighted` can help**, because the attribute is not the thing that's late; focus is. The signature misdirects investigation toward timeout or rAF-starvation theories (we instrumented both; neither was the cause).

Any Base UI roving-focus popup composed through these utils (Select, Menu, Combobox, …) shares the mechanism; we measured Select.

## Evidence

Instrumented Playwright harness against the built story iframe — rAF request/fire/cancel ledger + `focusin` log + `data-highlighted` mutation timeline. Local repro ~44% (30/68 runs) with a 5 s `waitFor` active on every failing run. The ledger shows:

- `data-highlighted` set ~1 ms after ArrowDown (sync with React state);
- the option-focus rAF cancelled by a later `enqueueFocus` call from the focus manager's initial-focus path;
- focus settling on the option 2–6 ms after open;
- zero rAF gaps through every failure window (no starvation).

## Suggested direction

Scope the cancellation to the caller or the element instead of the module — e.g. key pending handles per element (`WeakMap<Element, number>`), or rely solely on the returned cancel function (which already guards `rafId === currentRafId`) and drop the unconditional `cancelAnimationFrame(rafId)` preamble. Alternatively, the focus manager's initial focus could skip enqueueing when an item-focus is already pending for a node inside the floating element.

## Workaround

In tests: after keyboard-opening a roving-focus popup, assert that focus has entered the list (`expect(highlightedOption).toHaveFocus()`) before the next keystroke — asserting the focus contract instead of the visual attribute removed the flake entirely for us (0/40 post-fix, 3× full suite green).

michaldudak:
Fixes #5268.

## Summary

- preserve keyboard-open focus intent across the controlled `activeIndex` update
- focus the initial highlighted item without waiting for an animation frame
- add hook regressions for empty and preselected lists
- add a Select integration regression with animation-frame callbacks blocked

## Root cause

`useListNavigation` recorded the navigation key, but reset synchronous focus while the initial active item was being derived. The highlighted state therefore committed before focus, leaving a frame where a subsequent key event could still target the trigger.

The new one-shot ref carries keyboard-open intent until the active item is focused, then clears on consumption or close. Pointer-open and virtualized fallback behavior remain unchanged.

## Validation

- `pnpm test:jsdom useListNavigation --no-watch`
- `pnpm test:chromium useListNavigation --no-watch`
- `pnpm test:chromium SelectRoot --no-watch`
- `pnpm eslint packages/react/src/floating-ui-react/hooks/useListNavigation.ts packages/react/src/floating-ui-react/hooks/useListNavigation.test.tsx packages/react/src/select/root/SelectRoot.test.tsx`
- `pnpm prettier --check`
- `pnpm typescript --force`

### DIFF
--- a/packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
+++ b/packages/react/src/floating-ui-react/components/FloatingFocusManager.test.tsx
@@ -219,6 +219,56 @@ describe('FloatingFocusManager', () => {
         await flushMicrotasks();
         expect(screen.getByTestId('input')).toHaveFocus();
       });
+
+      test('recovers when focus falls back to the body before initial focus runs', async () => {
+        const animationFrameCallbacks: FrameRequestCallback[] = [];
+        vi.mocked(window.requestAnimationFrame).mockImplementation((callback) => {
+          animationFrameCallbacks.push(callback);
+          return animationFrameCallbacks.length;
+        });
+
+        function TestCase() {
+          const [open, setOpen] = React.useState(false);
+          const [showTransient, setShowTransient] = React.useState(true);
+          const initialFocusRef = React.useRef<HTMLButtonElement | null>(null);
+          const { refs, context } = useFloating({ open, onOpenChange: setOpen });
+
+          return (
+            <React.Fragment>
+              <button ref={refs.setReference} onClick={() => setOpen(true)}>
+                Open
+              </button>
+              <button data-testid="hide-transient" onClick={() => setShowTransient(false)}>
+                Hide transient
+              </button>
+              {open && (
+                <FloatingFocusManager
+                  context={context}
+                  initialFocus={initialFocusRef}
+                  modal={false}
+                >
+                  <div ref={refs.setFloating}>
+                    {showTransient && <button data-testid="transient">Transient</button>}
+                    <button ref={initialFocusRef}>Initial focus</button>
+                  </div>
+                </FloatingFocusManager>
+              )}
+            </React.Fragment>
+          );
+        }
+
+        render(<TestCase />);
+        fireEvent.click(screen.getByRole('button', { name: 'Open' }));
+        act(() => screen.getByTestId('transient').focus());
+        await flushMicrotasks();
+
+        fireEvent.click(screen.getByTestId('hide-transient'));
+        expect(document.body).toHaveFocus();
+
+        act(() => animationFrameCallbacks.forEach((callback) => callback(performance.now())));
+
+        expect(screen.getByRole('button', { name: 'Initial focus' })).toHaveFocus();
+      });
     });
 
     describe('prop: returnFocus', () => {
@@ -770,8 +820,15 @@ describe('FloatingFocusManager', () => {
             innerRoot?.appendChild(iframe);
 
             // Properly open, write, and close the iframe document.
-            const iframeDoc = iframe.contentWindow?.document;
+            const iframeWindow = iframe.contentWindow;
+            const iframeDoc = iframeWindow?.document;
             if (iframeDoc) {
+              // Focus scheduling is scoped to the target document's window.
+              // Mirror this suite's synchronous frame mock in the iframe realm.
+              iframeWindow.requestAnimationFrame = (callback) => {
+                callback(0);
+                return 0;
+              };
               iframeDoc.open();
               iframeDoc.write(`<div id="rootIframe"></div>`);
               iframeDoc.close();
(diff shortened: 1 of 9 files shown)


flaviendelangle (approved):
# PR review

This PR fixes a keyboard-open focus race in `useListNavigation` by carrying keyboard-open intent through the controlled `activeIndex` update via a one-shot token ref, so the highlighted item is focused synchronously in the same commit rather than a frame later. I traced every exit path of the new token, the `focusItem(deferReveal)` scheduler split, and the removed `item &&` guard; the logic is sound and the token is cleared on all reachable paths (close, in-bounds consume, out-of-bounds, retries exhausted). Nothing merge-blocking; both cited jsdom suites pass locally.

## Bugs (0)

No findings.

## Tests (0)

No findings.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Notes

### ℹ️ Removed `item &&` guard in `shouldScrollIntoView` is a safe dead-guard removal

**Location:** `packages/react/src/floating-ui-react/hooks/useListNavigation.ts:363`

```tsx
const shouldScrollIntoView = forceScrollIntoView || !isPointerModalityRef.current;
```

The dropped `item &&` referenced the `React.useMemo`-produced `itemProps` object (defined later, hence the removed `no-use-before-define` disable), which is unconditionally truthy — so the condition is unchanged and the eslint-disable is correctly gone. Confirming this was intentional, not an accidental narrowing.

### ℹ️ Token intentionally leaks one render on the `selectedIndex` path

On a keyboard open with `selectedIndex != null`, `forceSyncFocusOnOpenRef` is set during the `activeIndex == null` commit, then the effect early-returns (`selectedIndexRef.current != null`) without clearing it; it is consumed on the next render when `activeIndex` becomes the selected index. This cross-render carry is what makes the selected-item sync focus work (covered by "focuses the selected keyboard-open item"), and any close clears it — no leak into unrelated navigation. Flagging only because the one-render persistence is non-obvious for future maintainers.

## Verdict

**Approve** — focused, well-tested fix; token lifecycle is fully cleared on every reachable path and the cited test suites pass locally.

---

atomiks:
This could cause scroll-to-top bugs due to sync `.focus()`. `{ preventScroll: true }` doesn't work on Android.

Requires a popup portaled into a container in a complex layout such that the `position: fixed` fallback doesn't work

atomiks:
>  Bundle | Parsed size | Gzip size
-- | -- | --
@base-ui/react | 🔺+1.12KB(+0.25%) | 🔺+409B(+0.28%)



Given this is only a test issue (from what I gather), I think it makes sense to just drop this, especially given it has a test workaround.

michaldudak:
Yup, the bundle size increase isn't worth it.

PR #5295, opened 2026-07-21
[toast] Fix swipe direction locking for two-axis swipes

sai6855:
Fix swipe direction locking for two-axis swipes

### DIFF
--- a/packages/react/src/toast/root/ToastRoot.tsx
+++ b/packages/react/src/toast/root/ToastRoot.tsx
@@ -323,6 +323,8 @@ export const ToastRoot = React.forwardRef(function ToastRoot(
     const cancelDeltaY = clientY - swipeCancelBaselineRef.current.y;
     const cancelDeltaX = clientX - swipeCancelBaselineRef.current.x;
 
+    let resolvedLockedDirection = lockedDirection;
+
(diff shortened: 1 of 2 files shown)


packages/react/src/toast/root/ToastRoot.tsx
```diff
@@ -335,20 +337,21 @@ export const ToastRoot = React.forwardRef(function ToastRoot(
...
-          setLockedDirection(absX > absY ? 'horizontal' : 'vertical');
+          resolvedLockedDirection = absX > absY ? 'horizontal' : 'vertical';
+          setLockedDirection(resolvedLockedDirection);
         }
       }
     }
 
     let candidate: 'up' | 'down' | 'left' | 'right' | undefined;
     if (!intendedSwipeDirectionRef.current) {
-      if (lockedDirection === 'vertical') {
```
sai6855, line 345: `lockedDirection` here is refering to it's stale value not the updated one at line [338](https://github.com/mui/base-ui/pull/5295/changes#diff-80b2925d1049cfc82c92ff05e375468cb8a45eb2aff379cf40aff022c9ea4428L338).

atomiks (approved):

